PR #429, opened 2020-05-29
Update Windows Kernel Build Flags

nibanks:
Updates the Visual Studio project files for the Windows kernel mode to better match the compile & linker flags used by the internal Windows build system. This is an attempt to fix an XBOX load problem with msquic.sys that we've found internally.


src/platform/platform.kernel.vcxproj
```diff
@@ -104,4 +110,4 @@ mc.exe -um -h $(SolutionDir)build\winkernel\$(Platform)_$(Configuration)_schanne
   </ItemDefinitionGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets" />
-</Project>
+</Project>
```
anrossi, line 113: VS ate the newline at the end of the file again

nibanks: Uggggg! I will fix this.

PR #431, opened 2020-05-29
Disallow in-source builds

ThadHouse:
In source builds cause weird tree corruptions, and most bigger projects disallow them just to not have to worry about them

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,14 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
+# Disable in-source builds to prevent source tree corruption.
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1,6 +1,14 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
+# Disable in-source builds to prevent source tree corruption.
+if(" ${CMAKE_SOURCE_DIR}" STREQUAL " ${CMAKE_BINARY_DIR}")
```
nibanks, line 5: Why do you have a space at the beginning of each string?

ThadHouse: Not fully sure. I had found this years ago on the CMake docs site, and this is verbatim what it was. I'll fix it to clean it up a bit.

PR #433, opened 2020-05-29
Make CMake build compatible with more platforms

ThadHouse:
No need to select m32 or m64. Cmake does this automatically based on the toolchain.

GCC 8+ needs and explicit link to libatomic.

Use ./config instead of ./Configure for openssl.


src/platform/CMakeLists.txt
```diff
@@ -59,3 +59,13 @@ if(QUIC_TLS STREQUAL "openssl")
...
     target_link_libraries(platform kremlib evercrypt mitls quiccrypto)
 endif()
+
+if (NOT MSVC AND NOT APPLE)
+    find_library(ATOMIC NAMES atomic libatomic.so.1)
+    if (ATOMIC)
+        message(STATUS "Found libatomic: ${ATOMIC}")
+    endif()
+endif()
+
```
nibanks, line 69: nit: remove extra newline

CMakeLists.txt
```diff
@@ -173,7 +167,7 @@ if(QUIC_TLS STREQUAL "openssl")
         OUTPUT ${QUIC_BUILD_DIR}/openssl/include
         OUTPUT ${QUIC_BUILD_DIR}/openssl/lib/libcrypto.a
         OUTPUT ${QUIC_BUILD_DIR}/openssl/lib/libssl.a
-        COMMAND ./Configure linux-x86_64 enable-tls1_3 --prefix=${QUIC_BUILD_DIR}/openssl
+        COMMAND ./config enable-tls1_3 --prefix=${QUIC_BUILD_DIR}/openssl
```
nibanks, line 178: I've tried this change before. Azure Pipelines build is going to fail, because it doesn't recognize the Azure Pipelines Windows image.

nibanks:
```
[ 34%] Generating openssl/include, openssl/lib/libcrypto.a, openssl/lib/libssl.a
Operating system: x86_64-whatever-build
This system (build) is not supported. See file INSTALL for details.
CMakeFiles/OpenSSL.dir/build.make:79: recipe for target 'openssl/include' failed
```

ThadHouse: Don't we not currently support openssl on windows? Shoud we just be disabling it on Windows? We're going to have to special case Windows anyway.

nibanks: Err, I meant the Azure Pipelines **Linux** image, not Windows.

ThadHouse: I'll look into getting it on Windows to satisfy the issue you created a bit ago.

ThadHouse: https://developercommunity.visualstudio.com/content/problem/602584/openssl-build-error-when-using-pipelines.html

CMakeLists.txt
```diff
@@ -181,7 +175,7 @@ if(QUIC_TLS STREQUAL "openssl")
         OUTPUT ${QUIC_BUILD_DIR}/openssl/include
         OUTPUT ${QUIC_BUILD_DIR}/openssl/lib/libcrypto${CMAKE_STATIC_LIBRARY_SUFFIX}
         OUTPUT ${QUIC_BUILD_DIR}/openssl/lib/libssl${CMAKE_STATIC_LIBRARY_SUFFIX}
-        COMMAND ./Configure linux-x86_64 enable-tls1_3 --prefix=${QUIC_BUILD_DIR}/openssl
+        COMMAND SYSTEM=`uname -s` ./config enable-tls1_3 --prefix=${QUIC_BUILD_DIR}/openssl
```
nibanks, line 178: Should we always be doing `SYSTEM='uname -s'` or just for Azure Pipelines?

ThadHouse: In this case, its fine to do it in the command. We'd have to do a LOT more work to make cross compilation work, which is what that is normally used for.

PR #443, opened 2020-06-01
Remove flags that can't be sent during congestion control without allowance

ThadHouse:
Without removing these flags, the framing logic thinks it has data that can be sent, even though congestion control told us that we could not send certain types of frames, causing empty frames to be generated.


nibanks:
Please add a description as to what the problem was that this is fixing.

PR #448, opened 2020-06-01
Separate datapath shutdown from datapath cleanup

ThadHouse:
If the datapath threads are still running, they can receive data, that can race when we are trying to clean up worker threads. By cleaning up the datapath threads before the worker threads, we avoid this race

### DIFF
--- a/src/core/library.c
+++ b/src/core/library.c
@@ -280,9 +280,7 @@ MsQuicLibraryUninitialize(
     // Ensure all datapath workers are shut down, so new connections won't be 
     // created while we shut down worker threads.
     //
-    if (MsQuicLib.Datapath != NULL) {
-        QuicDataPathShutdownWorkers(MsQuicLib.Datapath);
-    }
+    QuicDataPathShutdown(MsQuicLib.Datapath);
 
     //
     // The library's worker pool for processing half-opened connections
--- a/src/core/library.c
+++ b/src/core/library.c
@@ -277,6 +277,14 @@ MsQuicLibraryUninitialize(
     }
 
     //
+    // Ensure all datapath workers are shut down, so new connections won't be 
+    // created while we shut down worker threads.
+    //
+    if (MsQuicLib.Datapath != NULL) {
+        QuicDataPathShutdownWorkers(MsQuicLib.Datapath);
+    }
+
+    //
     // The library's worker pool for processing half-opened connections
     // needs to be cleaned up first, as it's the last thing that can be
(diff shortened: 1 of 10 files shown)


src/core/library.c
```diff
@@ -276,6 +276,14 @@ MsQuicLibraryUninitialize(
         MsQuicLib.UnregisteredSession = NULL;
     }
 
+    //
+    // Ensure all datapath workers are shut down, so new connections won't be 
+    // created while we shut down worker threads.
+    //
+    if (MsQuicLib.Datapath != NULL) {
```
nibanks, line 283: I don't believe `MsQuicLib.Datapath` should ever be NULL. You can remove the check.

src/platform/datapath_winuser.c
```diff
@@ -770,19 +770,14 @@ QuicDataPathInitialize(
...
 _IRQL_requires_max_(PASSIVE_LEVEL)
 void
-QuicDataPathUninitialize(
+QuicDataPathShutdownWorkers(
     _In_ QUIC_DATAPATH* Datapath
     )
 {
     if (Datapath == NULL) {
         return;
     }
```
nibanks, line 779: Do we ever actually call with a NULL input? According to the SAL annotation, it's not optional. So I think we can just remove (and not add) all these checks.

src/inc/quic_datapath.h
```diff
@@ -247,8 +247,18 @@ QuicDataPathInitialize(
     _Out_ QUIC_DATAPATH* *NewDatapath
     );
 
+//
+// Shuts down all workers in a QUIC Datapath handle.
+//
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+QuicDataPathShutdownWorkers(
```
nibanks, line 255: I think I prefer just `QuicDataPathShutdown` instead.

ThadHouse:
I think we're hitting the extended shutdown time anyway. The unit tests are timing out on CI.

src/inc/quic_platform_dispatch.h
```diff
@@ -100,6 +100,12 @@ void
     _In_ QUIC_DATAPATH* Datapath
     );
 
+typedef
+void
+(*QUIC_DATAPATH_SHUTDOWN_WORKERS)(
```
nibanks, line 105: This needs to be shutdown too.

ThadHouse:
Closed in favor of #458. Will keep branch around for a day or 2.

PR #451, opened 2020-06-01
Remove use of internal APIs from sample application

ThadHouse:
The sample application should show how things can be done with the public API, without needing access to the platform project. Will make it much easier to learn how the API works.


ThadHouse:
Nevermind. This doesn't help. msquichelper is full of internal API usage as well.

PR #454, opened 2020-06-02
Add support for building in UWP mode

### LINKED ISSUE #436: Enable building for UWP user mode

Building the library for UWP user mode could be useful. The following functions and definitions don't exist in UWP mode.

`ReadNoFence64` does not exist. Need to find an alternative. Since the function is defined in a header, and just forwards to another function, might be able to see if the other function exists.

Much of the code related to setting thread name in `QuicThreadCreate` does not exist. However, because UWP is separate binaries in all cases, we can safely use `SetThreadDescription` instead for UWP.

That looks to be it.

ThadHouse:
Flags needed to be added to the build. Right now we have to force override a compiler defined macro in the platform header file. Looking to see if there is a way to do this only through the command line.

Running tests is possible, but requires installing the msix file, so not extremely feasable. I'll look into this more as well. I did add a build however so we can be sure this at leasts continues to build.

Closes #436

### DIFF
--- a/.azure/templates/build-config-user.yml
+++ b/.azure/templates/build-config-user.yml
@@ -6,11 +6,10 @@ parameters:
   arch: ''
   tls: ''
   extraBuildArgs: ''
-  extraName: ''
 
 jobs:
-- job: build_${{ parameters.platform }}_${{ parameters.arch }}_${{ parameters.tls }}_${{ parameters.extraName }}
-  displayName: Build ${{ parameters.platform }} ${{ parameters.arch }} ${{ parameters.tls }} ${{ parameters.extraName }}
+- job: build_${{ parameters.platform }}_${{ parameters.arch }}_${{ parameters.tls }}
+  displayName: Build ${{ parameters.platform }} ${{ parameters.arch }} ${{ parameters.tls }}
   pool:
     vmImage: ${{ parameters.image }}
   steps:
@@ -29,13 +28,13 @@ jobs:
     inputs:
       pwsh: true
       filePath: scripts/build.ps1
-      arguments: -Arch ${{ parameters.arch }} -Tls ${{ parameters.tls }} ${{ parameters.extraBuildArgs }}
+      arguments: -Arch ${{ parameters.arch }} -Tls ${{ parameters.tls }} -Platform ${{ parameters.platform }} ${{ parameters.extraBuildArgs }}
 
   - task: PowerShell@2
     displayName: Build Source Code (Release)
     inputs:
       pwsh: true
       filePath: scripts/build.ps1
-      arguments: -Config Release -Arch ${{ parameters.arch }} -Tls ${{ parameters.tls }} ${{ parameters.extraBuildArgs }}
+      arguments: -Config Release -Arch ${{ parameters.arch }} -Tls ${{ parameters.tls }} -Platform ${{ parameters.platform }} ${{ parameters.extraBuildArgs }}
 
   - template: ./upload-artifacts.yml
--- a/.azure/templates/build-config-user.yml
+++ b/.azure/templates/build-config-user.yml
@@ -6,10 +6,11 @@ parameters:
   arch: ''
   tls: ''
   extraBuildArgs: ''
+  extraName: ''
 
 jobs:
-- job: build_${{ parameters.platform }}_${{ parameters.arch }}_${{ parameters.tls }}
-  displayName: Build ${{ parameters.platform }} ${{ parameters.arch }} ${{ parameters.tls }}
+- job: build_${{ parameters.platform }}_${{ parameters.arch }}_${{ parameters.tls }}_${{ parameters.extraName }}
+  displayName: Build ${{ parameters.platform }} ${{ parameters.arch }} ${{ parameters.tls }} ${{ parameters.extraName }}
   pool:
     vmImage: ${{ parameters.image }}
   steps:
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -56,7 +56,7 @@ param (
     [string]$Config = "Debug",
 
     [Parameter(Mandatory = $false)]
-    [ValidateSet("x86", "x64", "arm", "arm64", "x86_uwp", "x64_uwp", "arm_uwp", "arm64_uwp")]
+    [ValidateSet("x86", "x64", "arm", "arm64")]
     [string]$Arch = "x64",
 
     [Parameter(Mandatory = $false)]
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -9,6 +9,9 @@ This script provides helpers for building msquic.
 .PARAMETER Arch
     The CPU architecture to build for.
 
+.PARAMETER Uwp
+    Set to build for UWP platform
+
 .PARAMETER Tls
     The TLS library to use.
 
@@ -57,6 +60,9 @@ param (
     [string]$Arch = "x64",
 
     [Parameter(Mandatory = $false)]
+    [switch]$Uwp = $false,
+
+    [Parameter(Mandatory = $false)]
     [ValidateSet("schannel", "openssl", "stub", "mitls")]
     [string]$Tls = "",
 
@@ -97,7 +103,7 @@ if ("" -eq $Tls) {
     }
 }
 
-if (!$IsWindows -And $Arch -like "*uwp") {
+if (!$IsWindows -And $Uwp) {
     Write-Error "[$(Get-Date)] Cannot build UWP on non windows platforms"
     exit
 }
@@ -118,8 +124,13 @@ if ($IsWindows) {
     $ArtifactsDir = Join-Path $BaseArtifactsDir "linux"
     $BuildDir = Join-Path $BaseBuildDir "linux"
 }
-$ArtifactsDir = Join-Path $ArtifactsDir "$($Arch)_$($Config)_$($Tls)"
-$BuildDir = Join-Path $BuildDir "$($Arch)_$($Tls)"
+if ($Uwp) {
+    $ArtifactsDir = Join-Path $ArtifactsDir "$($Arch)_uwp_$($Config)_$($Tls)"
+    $BuildDir = Join-Path $BuildDir "$($Arch)_uwp_$($Tls)"
+} else {
+    $ArtifactsDir = Join-Path $ArtifactsDir "$($Arch)_$($Config)_$($Tls)"
+    $BuildDir = Join-Path $BuildDir "$($Arch)_$($Tls)"
+}
 
 if ($Clean) {
     # Delete old build/config directories.
@@ -159,10 +170,6 @@ function CMake-Generate {
             "x64"   { $Arguments += "x64" }
             "arm"   { $Arguments += "arm" }
             "arm64" { $Arguments += "arm64" }
-            "x86_uwp"   { $Arguments += "Win32" }
-            "x64_uwp"   { $Arguments += "x64" }
-            "arm_uwp"   { $Arguments += "arm" }
-            "arm64_uwp" { $Arguments += "arm64" }
         }
     } else {
         $Arguments += " 'Linux Makefiles'"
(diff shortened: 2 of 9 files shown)


src/bin/CMakeLists.txt
```diff
@@ -21,6 +21,9 @@ target_link_libraries(msquic PRIVATE core platform)
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
     target_link_libraries(msquic PUBLIC
         ws2_32 schannel ntdll bcrypt ncrypt crypt32 iphlpapi advapi32)
+    if(QUIC_UWP_BUILD)
+        target_link_libraries(msquic PUBLIC OneCoreUAP)
```
nibanks, line 25: Just go ahead and add this to the list above unconditionally, assuming it works.

src/tools/attack/CMakeLists.txt
```diff
@@ -17,4 +17,7 @@ target_link_libraries(quicattack core platform)
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
     target_link_libraries(quicattack
         ws2_32 schannel ntdll bcrypt ncrypt crypt32 iphlpapi advapi32)
+    if(QUIC_UWP_BUILD)
+        target_link_libraries(quicattack OneCoreUAP)
```
nibanks, line 21: Just go ahead and add this to the list above unconditionally, assuming it works.

src/inc/quic_platform_winuser.h
```diff
@@ -776,13 +781,17 @@ QuicThreadCreate(
             ARRAYSIZE(WideName) - 1,
             Config->Name,
             _TRUNCATE);
+#if defined(QUIC_UWP_BUILD)
+        SetThreadDescription(*Thread, WideName);
+#else
         THREAD_NAME_INFORMATION ThreadNameInfo;
         RtlInitUnicodeString(&ThreadNameInfo.ThreadName, WideName);
         NtSetInformationThread(
```
nibanks, line 789: Do you get a build time or run time error from this?

src/inc/quic_platform_winuser.h
```diff
@@ -31,6 +31,11 @@
 #define WIN32_LEAN_AND_MEAN 1
 #endif
 
+#ifdef QUIC_UWP_BUILD
+#undef WINAPI_FAMILY
+#define WINAPI_FAMILY WINAPI_FAMILY_DESKTOP_APP
+#endif
```
nibanks, line 37: I want to see if we can get a better solution for this before we merge this PR.

scripts/build.ps1
```diff
@@ -154,6 +159,10 @@ function CMake-Generate {
             "x64"   { $Arguments += "x64" }
             "arm"   { $Arguments += "arm" }
             "arm64" { $Arguments += "arm64" }
+            "x86_uwp"   { $Arguments += "Win32" }
```
nibanks, line 162: nit: can you align all the `{ ... }` above too?

scripts/build.ps1
```diff
@@ -53,7 +53,7 @@ param (
     [string]$Config = "Debug",
 
     [Parameter(Mandatory = $false)]
-    [ValidateSet("x86", "x64", "arm", "arm64")]
+    [ValidateSet("x86", "x64", "arm", "arm64", "x86_uwp", "x64_uwp", "arm_uwp", "arm64_uwp")]
```
nibanks, line 56: Might it be better just to add a `$UWP` parameter instead that defaults to `$false`?

src/inc/quic_platform_winuser.h
ThadHouse: Build time. UWP has more locked down access to the Nt stuff. But at the same time, it's known to have SetThreadDescription in all cases.

src/bin/CMakeLists.txt
ThadHouse: This would break compatibility with Windows 7, since that is a UAP specific library.

src/inc/quic_platform_winuser.h
ThadHouse: I have an email out asking.

scripts/build.ps1
ThadHouse: I can do that. It will be a bit more invasive because of modifications required to the directory setting, but that's ok.

src/bin/CMakeLists.txt
nibanks: I'm fine that that, as `SetThreadDescription` is also newer than Win7. We have no goals to support < Win10.

src/inc/quic_platform_winuser.h
nibanks: What is the build error exactly?

scripts/build.ps1
nibanks: Yeah, but I think it will end up being cleaner.

src/bin/CMakeLists.txt
ThadHouse: Ok. So no goals to support Windows 8.1 either? If so, we should change everything to link to WindowsApp.lib, which has all of those APIs.

src/inc/quic_platform_winuser.h
ThadHouse: Tons of undeclared identifier errors. The low level nt headers are just completely excluded from UWP. However, in regard to the discussion above, if we don't care about anything before Windows 10, we can just use SetThreadDescription for all windows builds.

src/bin/CMakeLists.txt
ThadHouse: Ok. Turns out I still need this separate. I can use WindowsApp.lib instead of many of the other windows APIs, but GetActiveProcessor account only exists in the library for UWP, and linking to it from a desktop app will not work correctly.

scripts/build.ps1
ThadHouse: Fixed by fixing below

scripts/build.ps1
ThadHouse: Fixed

.azure/templates/build-config-user.yml
```diff
@@ -6,10 +6,11 @@ parameters:
   arch: ''
   tls: ''
   extraBuildArgs: ''
+  extraName: ''
 
 jobs:
-- job: build_${{ parameters.platform }}_${{ parameters.arch }}_${{ parameters.tls }}
-  displayName: Build ${{ parameters.platform }} ${{ parameters.arch }} ${{ parameters.tls }}
+- job: build_${{ parameters.platform }}_${{ parameters.arch }}_${{ parameters.tls }}_${{ parameters.extraName }}
+  displayName: Build ${{ parameters.platform }} ${{ parameters.arch }} ${{ parameters.tls }} ${{ parameters.extraName }}
```
nibanks, line 13: Sorry to keep asking you to change things, but architecturally, I think considering UWP a separate platform might be better; especially when considering directory structure. What do you think? I'm not sure how much of a pain it would be to change it that way though.

scripts/build.ps1
```diff
@@ -56,6 +59,9 @@ param (
     [ValidateSet("x86", "x64", "arm", "arm64")]
     [string]$Arch = "x64",
 
+    [Parameter(Mandatory = $false)]
+    [switch]$Uwp = $false,
```
nibanks, line 63: Let's keep it all caps. `$UWP` like the `$PGO` option below.

.azure/templates/build-config-user.yml
ThadHouse: If we're just considering it separate on Azure then its not too bad. In the src, its platform would still be Windows. vcpkg considers uwp a separate architecture inside of windows, which is why I went that way originally.

scripts/build.ps1
ThadHouse: Done

.azure/templates/build-config-user.yml
nibanks: I think integrating it into our PowerShell and AZP yml files as just a different platform might be best.

scripts/build.ps1
```diff
@@ -107,15 +118,22 @@ $SrcDir = Join-Path $RootDir "src"
...
+    }
 } else {
     $ArtifactsDir = Join-Path $BaseArtifactsDir "linux"
     $BuildDir = Join-Path $BaseBuildDir "linux"
 }
+
 $ArtifactsDir = Join-Path $ArtifactsDir "$($Arch)_$($Config)_$($Tls)"
 $BuildDir = Join-Path $BuildDir "$($Arch)_$($Tls)"
 
+
```
nibanks, line 136: nit: remove this extra line

.azure/templates/build-config-user.yml
nibanks: I don't think you should need the `extraName` parameter any more, right? Also, I'd like it if we could infer that `extraBuildArgs` based on the platform now.

scripts/build.ps1
```diff
@@ -107,15 +118,22 @@ $SrcDir = Join-Path $RootDir "src"
 $ArtifactsDir = $null
 $BuildDir = $null
 if ($IsWindows) {
-    $ArtifactsDir = Join-Path $BaseArtifactsDir "windows"
-    $BuildDir = Join-Path $BaseBuildDir "windows"
+    if ($IsUwp) {
+        $ArtifactsDir = Join-Path $BaseArtifactsDir "windows-uwp"
+        $BuildDir = Join-Path $BaseBuildDir "windows-uwp"
```
nibanks, line 123: I might prefer just `uwp` instead. It already stands for "Universal Windows Platform" (I think) so an extra windows seems redundant.

.azure/templates/build-config-user.yml
```diff
@@ -28,13 +28,13 @@ jobs:
     inputs:
       pwsh: true
       filePath: scripts/build.ps1
-      arguments: -Arch ${{ parameters.arch }} -Tls ${{ parameters.tls }} ${{ parameters.extraBuildArgs }}
+      arguments: -Arch ${{ parameters.arch }} -Tls ${{ parameters.tls }} -Platform ${{ parameters.platform }} ${{ parameters.extraBuildArgs }}
```
nibanks, line 31: Getting closer. Again, sorry to keep asking for more. I don't care so much for a `-Platform` parameter that's only used for UWP. I liked the `-UWP` way at the PowerShell script layer. Somehow I'd like to turn a `parameters.platform == UWP` into that other param... Ideas?

ThadHouse: Only way to do that is to create another script azure script specifically for uwp, or to conventionalize both build source code tasks in build-config-user.yml.

The platform can be nice though, because we could use that same flag for cross compilation in the future, specifically for like Pi builds.

nibanks: Ok.

scripts/build.ps1
```diff
@@ -56,6 +59,10 @@ param (
     [ValidateSet("x86", "x64", "arm", "arm64")]
     [string]$Arch = "x64",
 
+    [Parameter(Mandatory = $false)]
+    [ValidateSet("UWP", "windows", "linux")] # For future expansion
```
nibanks, line 63: Let's keep the case the same for all (lowercase).

nibanks: Also, below, if $Platform is empty, default it below to "windows" or "linux" based on the current platform. Then you can use that when building up $ArtifactsDir and $BuildDir.

ThadHouse:
Ok, everything should be good now, except for the macro redefinition. Its an odd interaction between CMake and VS.

nibanks:
> Ok, everything should be good now, except for the macro redefinition. Its an odd interaction between CMake and VS.

I agree, but I do want to hold out for a little bit in the hope that we can find a solution to the macro redefinition. If we don't have anything by tomorrow, we can merge it.

nibanks:
In quic_platform_winuser.h #ifdef out the `QUIC_COMPARTMENT_ID` stuff for UWP mode.

ThadHouse:
With that commented out, we are now just at the 2 disallowed Query APIs, which I think the right solution is to work with and make legal in UWP. I don't think that needs to be a blocker on this PR, since once its building its easier to keep it building with future changes.

PR #455, opened 2020-06-02
Don't call DisableThreadLibraryCalls with static CRT

### LINKED ISSUE #437: DisableThreadLibraryCalls should not be used with static CRT
### Describe the bug

According to https://docs.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-disablethreadlibrarycalls DisableThreadLibraryCalls should not be used if the static CRT is used. Not sure exactly what would break, but violating documented practices is never a good thing.

ThadHouse:
Closes #437

### DIFF
--- a/src/bin/winuser/dllmain.c
+++ b/src/bin/winuser/dllmain.c
@@ -34,7 +34,9 @@ DllMain(
     switch (Reason) {
 
     case DLL_PROCESS_ATTACH:
+#ifndef _MT
         DisableThreadLibraryCalls(Instance);
+#endif
         QuicPlatformSystemLoad();
         MsQuicLibraryLoad();
         break;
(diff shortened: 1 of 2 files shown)


src/bin/winuser/dllmain.c
```diff
@@ -34,7 +34,9 @@ DllMain(
     switch (Reason) {
 
     case DLL_PROCESS_ATTACH:
+#ifndef _MT
```
nibanks, line 37: Can you put a comment on this line (and in the other file) saying something like "Don't disable thread library calls with static CRT!"

ThadHouse: Done

nibanks: Just put the comment on the same line as the #ifdef please:
```
#ifndef _MT // Don't disable thread library calls with static CRT!
```

PR #461, opened 2020-06-03
Build QNS interop image

larseggert:
This builds a docker image for the quic-network-simulator (https://github.com/marten-seemann/quic-network-simulator/).

What's missing is a `run_endpoint.sh` script that configures the client and server binary for the various QNS tests. Since I'm not super-familiar with the options these things take, I'll let @nibanks or someone else add that script...

First step for #462.

### DIFF
--- a/run_endpoint.sh
+++ b/run_endpoint.sh
@@ -1,0 +1,43 @@
+#! /usr/bin/env bash
+
+# Set up the routing needed for the simulation
+/setup.sh
+
+if [ -n "$TESTCASE" ]; then
+    case "$TESTCASE" in
+    # TODO: add supported test cases here
+    "versionnegotiation"|"handshake"|"transfer"|"retry"|"resumption"|\
+    "multiconnect"|"zerortt"|"chacha20")
+        ;;
+    *)
+        exit 127
+        ;;
+    esac
+fi
+
+# The following variables are available for use:
+# - ROLE contains the role of this execution context, client or server
+# - SERVER_PARAMS contains user-supplied command line parameters
+# - CLIENT_PARAMS contains user-supplied command line parameters
+
+if [ "$ROLE" == "client" ]; then
+    # Wait for the simulator to start up.
+    /wait-for-it.sh sim:57832 -s -t 30
+    cd /downloads || exit
+
+    # TODO: add client support
+    # I am not sure if the msquic codebase has an h09 client?
+    exit 127
+
+elif [ "$ROLE" == "server" ]; then
+    case "$TESTCASE" in
+    "retry")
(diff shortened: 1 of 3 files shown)


nibanks:
Thanks @larseggert for putting this together! We'll take a look!

larseggert:
I force-pushed an update, since I figured out how to run the server. It doesn't do -28 unfortunately, so I couldn't test.

Does msquic have an h09 client? It didn't seem like based on what I saw from the various built binaries. So client functionality is disabled. If someone knows how to massage an msquic h09 client into generating requests for the various test cases, please add that?

run_endpoint.sh
```diff
@@ -0,0 +1,43 @@
...
+# - SERVER_PARAMS contains user-supplied command line parameters
+# - CLIENT_PARAMS contains user-supplied command line parameters
+
+if [ "$ROLE" == "client" ]; then
+    # Wait for the simulator to start up.
+    /wait-for-it.sh sim:57832 -s -t 30
+    cd /downloads || exit
+
+    # TODO: add client support
+    # I am not sure if the msquic codebase has an h09 client?
```
nibanks, line 29: The quicinterop tool is the client side tool that we use for `hq` testing. I'm not sure if it's generic enough for these purposes though.

run_endpoint.sh
```diff
@@ -0,0 +1,43 @@
...
+    case "$TESTCASE" in
+    "retry")
+        SERVER_PARAMS="-retry:1 $SERVER_PARAMS"
+        ;;
+    *)
+        ;;
+    esac
+
+    quicinteropserver ${SERVER_PARAMS} -root:/www -listen:* -port:443 \
+        -file:/server.crt -key:/server.key 2>&1
```
nibanks, line 42: Would it be possible to somehow dynamically generate these certs instead of checking them in? I'd prefer not to have checked in certs? Another option is to add a `-selfsign` option to quicinteropserver. We already have a helper function (`QuicPlatGetSelfSignedCert`) that could be used to easily add that functionality.

larseggert: Fixed in latest push

larseggert:
The latest push dynamically creates the cert and key

larseggert:
You also want to create a PR to add microsoft/msquic to https://github.com/marten-seemann/quic-interop-runner/blob/master/implementations.py

And set up a [Docker Hub](https://hub.docker.com/repository/create) integration, so the docker image auto-rebuilds.

run_endpoint.sh
larseggert: I never understood what options it takes, so someone who does would need to add support for it.

PR #470, opened 2020-06-04
Moves linux platform functions in public header to be inlined

### LINKED ISSUE #468: Linux functions in msquic_linux are not considered public
### Describe the bug

Unlike Windows, where functions such as  `QuicAddrSetFamily` are defined inline, on linux they are compiled as part of the platform library, and therefore not available in the DSO. However, `msquic_linux.h` is included from `msquic.h`, which means its part of the public API. This means we either need to export all of those functions from the DSO, we need to make them inline, or we need to remove them from the public API.

ThadHouse:
Because they were included from the public header, any attempt to use them before would get undefined symbols. This fixes that by inlining the functions like Windows

Closes #468


anrossi:
You'll need to add the function signatures of everything that is `inline`d into the platform/inline.c file otherwise you'll get some fun compilation errors on GCC.

ThadHouse:
This will need a rereview because of const correctness issues in QuicAddr4FromString and QuicAddr6FromString.

src/inc/msquic_linux.h
```diff
@@ -192,76 +193,258 @@ extern char *QuicOpenSslClientTrustedCert;
...
+BOOLEAN
+QuicAddr4FromString(
+    _In_z_ const char* AddrStr,
+    _Out_ QUIC_ADDR* Addr
+    )
+{
+    if (AddrStr[0] == '[') {
+        return FALSE;
+    }
+    char* AddrStrCopy = (char*)malloc(strlen(AddrStr) + 1);
```
nibanks, line 372: Don't copy. Just cast.

ThadHouse: It's unsafe to cast. If a user passes in a string literal that's an AV

nibanks: How about this then. Don't malloc a full copy. Instead, down in the  `PortStart != NULL` scope have a stack `char[]` big enough for the maximum IPv4 address (BTW fix IPv6 func below too the same general way). Copy the address into that array and call `inet_pton` onto it. No heap allocation necessary.

src/inc/msquic_linux.h
```diff
@@ -192,76 +193,257 @@ extern char *QuicOpenSslClientTrustedCert;
...
+    const char* PortStart = strchr(AddrStr, ':');
+    if (PortStart != NULL) {
+        if (strchr(PortStart+1, ':') != NULL) {
+            return FALSE;
+        }
+
+        char TmpAddrStr[16];
+        ptrdiff_t AddrLength = PortStart - AddrStr;
+        if (AddrLength >= sizeof(TmpAddrStr)) {
+            return false;
```
nibanks, line 382:
```suggestion
            return FALSE;
```

src/inc/msquic_linux.h
```diff
@@ -192,76 +193,257 @@ extern char *QuicOpenSslClientTrustedCert;
...
+    if (AddrStr[0] == '[') {
+        const char* BracketEnd = strchr(AddrStr, ']');
+        if (BracketEnd == NULL || *(BracketEnd+1) != ':') {
+            return FALSE;
+        }
+        char TmpAddrStr[41];
+
+        ptrdiff_t AddrLength = BracketEnd - AddrStr;
+        if (AddrLength >= sizeof(TmpAddrStr)) {
+            return false;
```
nibanks, line 416:
```suggestion
            return FALSE;
```

src/inc/msquic_linux.h
```diff
@@ -192,76 +193,257 @@ extern char *QuicOpenSslClientTrustedCert;
...
+    _In_z_ const char* AddrStr,
+    _Out_ QUIC_ADDR* Addr
+    )
+{
+    if (AddrStr[0] == '[') {
+        const char* BracketEnd = strchr(AddrStr, ']');
+        if (BracketEnd == NULL || *(BracketEnd+1) != ':') {
+            return FALSE;
+        }
+        char TmpAddrStr[41];
```
nibanks, line 412: Apparently the longest length is 46: https://stackoverflow.com/questions/166132/maximum-length-of-the-textual-representation-of-an-ipv6-address. But that assumes you support the IPv4-mapped IPv6 address, but I don't know if `inet_pton` does support it.

ThadHouse: I'll update it, because if it doesn't now it might in the future.

ThadHouse: I'm just going to go next highest power of 2 (64). I don't think we're too worried about stack space when this would be called, and 64 isn't bad anyway.

src/inc/msquic_linux.h
```diff
@@ -192,76 +193,257 @@ extern char *QuicOpenSslClientTrustedCert;
...
+    }
+
+    const char* PortStart = strchr(AddrStr, ':');
+    if (PortStart != NULL) {
+        if (strchr(PortStart+1, ':') != NULL) {
+            return FALSE;
+        }
+
+        char TmpAddrStr[16];
+        ptrdiff_t AddrLength = PortStart - AddrStr;
```
nibanks, line 380:
```suggestion
        size_t AddrLength = PortStart - AddrStr;
```

src/inc/msquic_linux.h
```diff
@@ -192,76 +193,257 @@ extern char *QuicOpenSslClientTrustedCert;
...
+    )
+{
+    if (AddrStr[0] == '[') {
+        const char* BracketEnd = strchr(AddrStr, ']');
+        if (BracketEnd == NULL || *(BracketEnd+1) != ':') {
+            return FALSE;
+        }
+        char TmpAddrStr[64];
+
+        ptrdiff_t AddrLength = BracketEnd - AddrStr;
```
nibanks, line 414:
```suggestion

        char TmpAddrStr[64];
        size_t AddrLength = BracketEnd - AddrStr;
```

PR #480, opened 2020-06-08
Fixes and Documentation for PGO

nibanks:
Apparently PGO wasn't actually working before, for a couple of reasons. This fixes it for x64 and adds some initial documentation.


ThadHouse:
Wouldn't we need a PDB per build specification? So we'd need all variations of x86, x64, mitls, schannel  and any other build types we want running over PGO?

nibanks:
> Wouldn't we need a PDB per build specification? So we'd need all variations of x86, x64, mitls, schannel and any other build types we want running over PGO?

Yep. (Assuming I can get this to build). For now, I have it only dynamically picking up PGO if we specify a file for it. As for other TLS configurations, Schannel is the only one we care about, plus it still works if you use the .pgd from Schannel on miTLS builds.

nibanks:
Ugg. I have no idea why it's not working now. It was building and running locally just fine. Then I pushed my PR. The PR build broke, then I tried a clean build locally and now it breaks locally too. I'm confused.

PR #481, opened 2020-06-08
Add configuration surface for resumption and 0-RTT for server.

anrossi:


nibanks:
I don't think we need to have a connection specific parameter; just a session layer one.

src/core/connection.h
```diff
@@ -141,6 +141,13 @@ typedef union QUIC_CONNECTION_STATE {
         //
         BOOLEAN UseRoundRobinStreamScheduling : 1;
 
+        //
+        // Indicates that this connection has resumption enabled and needs to
+        // keep the TLS state and transport parameters until it is done sending
+        // resumption tickets.
+        //
+        BOOLEAN ResumptionEnabled : 1;
```
nibanks, line 149: Do we need this or can we just dynamically read the current setting?

anrossi: This will be used to signal when the connection is no longer going to send any more resumption tickets, and thus, when it is safe to clean up the TLS state and the handshake transport parameters.

PR #487, opened 2020-06-10
Add API stubs for server 0-RTT support.

anrossi:


src/inc/msquic.h
```diff
@@ -105,6 +110,13 @@ typedef enum QUIC_CONNECTION_SHUTDOWN_FLAGS {
 
 DEFINE_ENUM_FLAG_OPERATORS(QUIC_CONNECTION_SHUTDOWN_FLAGS);
 
+typedef enum QUIC_SEND_RESUMPTION_FLAGS {
+    QUIC_SEND_RESUMPTION_FLAG_NONE          = 0x0000,
+    QUIC_SEND_RESUMPTION_FLAG_FINAL         = 0x0001    // Free TLS state after sending this ticket.
```
nibanks, line 115: I think we should we have a flag for 0-RTT key too.

src/tools/etw/trace.c
```diff
@@ -83,6 +83,7 @@ const char* ApiTypeStr[] = {
     "CONNECTION_CLOSE",
     "CONNECTION_SHUTDOWN",
     "CONNECTION_START",
+    "CONNECTION_SEND_RESUMPTION_TICKET"
```
anrossi, line 86: add comma

src/tools/etw/quicetw.h
```diff
@@ -25,7 +25,7 @@ typedef struct _BINDING BINDING;
 
 #define CAP_TO_32(uint64) (uint64 > UINT_MAX ? UINT_MAX : (ULONG)uint64)
 
-#define QUIC_API_COUNT 24
+#define QUIC_API_COUNT 26
```
anrossi, line 28: don't let this get out of sync

src/inc/msquic.h
anrossi: I imagined 0-RTT (read and write) being controlled by the resumption level instead. Are you suggesting that the 0-RTT resumption level (the session param) should instead control whether the server accepts 0-RTT (read) and this flag controls whether the server will send 0-RTT keys (write)?

src/inc/msquic.h
```diff
@@ -976,6 +1013,7 @@ typedef struct QUIC_API_TABLE {
     QUIC_CONNECTION_CLOSE_FN            ConnectionClose;
     QUIC_CONNECTION_SHUTDOWN_FN         ConnectionShutdown;
     QUIC_CONNECTION_START_FN            ConnectionStart;
+    QUIC_CONNECTION_SEND_RESUMPTION_FN  ConnectionSendResumptionTicket;
```
anrossi, line 1016: Don't forget to assign this.

src/inc/msquic.h
nibanks: You're right. Ignore this comment.

PR #489, opened 2020-06-10
Add base support for regression testing

ThadHouse:
No data currently, need to see what perf is like on Azure before creating test data.


scripts/regression.ps1
```diff
@@ -0,0 +1,183 @@
...
+    $pinfo.UseShellExecute = $false
+    $pinfo.Arguments = $Arguments
+    $p = New-Object System.Diagnostics.Process
+    $p.StartInfo = $pinfo
+    $p.Start() | Out-Null
+    $p.WaitForExit()
+    return $p.StandardOutput.ReadToEnd()
+}
+
+$GitPath = Join-Path $RootDir "build/PerfDataGit"
```
nibanks, line 97: Why are you putting it under the `build` folder instead of the `artifacts`?

.azure/azure-pipelines.ci.yml
```diff
@@ -188,6 +188,37 @@ stages:
       arch: x64
       tls: openssl
 
+#
+# Regression Tests
+#
+
+- stage: regression
+  displayName: Regression
```
nibanks, line 196: Let's name these performance tests instead of regression tests. Regression protection is just one part of the perf tests. Recording the results and tracking them over time is another.

.azure/azure-pipelines.ci.yml
```diff
@@ -188,6 +188,37 @@ stages:
...
+      platform: linux
+      arch: x64
+      tls: openssl
+
+- stage: performance_publish
+  displayName: Performance Results Publish
+  dependsOn:
+  - performance
+  jobs:
+  - template: ./templates/publish-performance.yml
```
nibanks, line 219: Does this really need to be a separate stage? What about making it a job of the stage above, but make it depend on all the other jobs?

.azure/templates/run-performance.yml
```diff
@@ -0,0 +1,54 @@
+# This template contains steps to run performance tests for a single configuration.
+
+parameters:
+  image: ''
+  platform: ''
+  config: 'Debug'
```
nibanks, line 6: Perf should probably run `Release` by default instead.

.azure/scripts/publish-performance-results.ps1
```diff
@@ -0,0 +1,16 @@
+# Root directory of the project.
+$RootDir = Split-Path $PSScriptRoot -Parent
+
+$GitPath = Join-Path $RootDir "artifacts/PerfDataGit"
+$ResultsPath = Join-Path $RootDir "artifacts/PerfDataResults"
+
+$env:GIT_REDIRECT_STDERR = '2>&1'
+git clone https://github.com/microsoft/msquic $GitPath
```
nibanks, line 8: Might be worth explicitly cloning the branch you want (for speed).
```suggestion
git clone --single-branch --branch https://github.com/microsoft/msquic $GitPath
```

.azure/azure-pipelines.ci.yml
ThadHouse: I can check if its possible. Not fully sure, because we actually name the jobs in the template script based off of parameters. It might work, but it would be ugly, and we'd have to manually add every new platform to the dep list. Whereas currently the dependency is really easy.

nibanks: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops&tabs=yaml#use-the-output-variable-from-a-job-in-a-condition-in-a-subsequent-job

ThadHouse: That actually doesn't help, because we'd still be explicitly naming things. In theory I should just be able to depend on the job name. I just think thats harder to maintain.

ThadHouse: I pushed it just to see what it looks like and to double check that it does work.

ThadHouse: Ok. It does work, but I have to put the depends on in `publish-performance.yml`. Because its a template, I can't put it in the root script.

PR #492, opened 2020-06-11
Resumption ticket generation/parsing logic

anrossi:
No-op stubbed TLS implementations for now.

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -2147,11 +2147,19 @@ QuicConnRecvResumptionTicket(
         ResumptionAccepted =
             QUIC_SUCCEEDED(QuicConnIndicateEvent(Connection, &Event));
 
-        QuicTraceEvent(
-            ConnServerResumeTicket,
-            "[conn][%p] Server app %b ticket.",
-            Connection,
-            ResumptionAccepted);
+
+        if (ResumptionAccepted) {
+            QuicTraceEvent(
+                ConnServerResumeTicket,
+                "[conn][%p] Server app accepted resumption ticket",
+                Connection);
+        } else {
+            QuicTraceEvent(
+                ConnError,
+                "[conn][%p] ERROR, %s.",
+                Connection,
+                "Resumption Ticket rejected by server app");
+        }
 
         QUIC_DBG_ASSERT(Offset + AppTicketLength == TicketLength);
     } else {
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -388,6 +388,10 @@ QuicConnFree(
     if (Connection->OrigDestCID != NULL) {
         QUIC_FREE(Connection->OrigDestCID);
     }
+    if (Connection->HandshakeTP != NULL) {
+        QUIC_FREE(Connection->HandshakeTP);
+        Connection->HandshakeTP = NULL;
+    }
     QuicTraceEvent(
         ConnDestroyed,
         "[conn][%p] Destroyed",
@@ -435,10 +439,17 @@ QuicConnApplySettings(
 
     if (Settings->ServerResumptionLevel > QUIC_SERVER_NO_RESUME) {
         QUIC_DBG_ASSERT(!Connection->State.Started);
-        //
-        // TODO: allocate memory for handshake TP here
-        //
-        Connection->State.ResumptionEnabled = TRUE;
+        Connection->HandshakeTP = QUIC_ALLOC_NONPAGED(sizeof(*Connection->HandshakeTP));
+        if (Connection->HandshakeTP == NULL) {
+            QuicTraceEvent(
+                AllocFailure,
+                "Allocation of '%s' failed. (%llu bytes)",
+                "handshake TP",
+                sizeof(*Connection->HandshakeTP));
+        } else {
+            QuicZeroMemory(Connection->HandshakeTP, sizeof(*Connection->HandshakeTP));
+            Connection->State.ResumptionEnabled = TRUE;
+        }
     }
 
     QuicSendApplySettings(&Connection->Send, Settings);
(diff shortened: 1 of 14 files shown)


src/core/connection.c
```diff
@@ -435,10 +439,17 @@ QuicConnApplySettings(
 
     if (Settings->ServerResumptionLevel > QUIC_SERVER_NO_RESUME) {
         QUIC_DBG_ASSERT(!Connection->State.Started);
-        //
-        // TODO: allocate memory for handshake TP here
-        //
-        Connection->State.ResumptionEnabled = TRUE;
+        Connection->HandshakeTP = QUIC_ALLOC_NONPAGED(sizeof(*Connection->HandshakeTP));
```
anrossi, line 442: A later PR will use a pool for these allocations.

nibanks: Put a TODO please so it's not forgotten.

src/core/connection.c
```diff
@@ -1903,6 +1914,273 @@ QuicConnRestart(
...
+        QUIC_TP_FLAG_INITIAL_MAX_STRM_DATA_UNI |
+        QUIC_TP_FLAG_INITIAL_MAX_STRMS_BIDI |
+        QUIC_TP_FLAG_INITIAL_MAX_STRMS_UNI);
+
+    const uint8_t* EncodedHSTP =
+        QuicCryptoTlsEncodeTransportParameters(
+            Connection,
+            &HSTPCopy,
+            &EncodedTransportParametersLength);
+
```
nibanks, line 1946: nit: no newlines between calls and null checks,

src/core/connection.c
```diff
@@ -1903,6 +1914,273 @@ QuicConnRestart(
...
+        if (ResumedTP.ActiveConnectionIdLimit > QUIC_ACTIVE_CONNECTION_ID_LIMIT ||
+            ResumedTP.InitialMaxData > Connection->Send.MaxData ||
+            ResumedTP.InitialMaxStreamDataBidiLocal > Connection->Session->Settings.StreamRecvWindowDefault ||
+            ResumedTP.InitialMaxStreamDataBidiRemote > Connection->Session->Settings.StreamRecvWindowDefault ||
+            ResumedTP.InitialMaxStreamDataUni > Connection->Session->Settings.StreamRecvWindowDefault ||
+            ResumedTP.InitialMaxUniStreams > Connection->Streams.Types[STREAM_ID_FLAG_IS_CLIENT | STREAM_ID_FLAG_IS_UNI_DIR].MaxTotalStreamCount ||
+            ResumedTP.InitialMaxBidiStreams > Connection->Streams.Types[STREAM_ID_FLAG_IS_CLIENT | STREAM_ID_FLAG_IS_BI_DIR].MaxTotalStreamCount) {
+            //
+            // Server settings have changed since the resumption ticket was
+            // encoded, so reject resumption.
```
nibanks, line 2117: Definitely log an error/warning here.

src/core/connection.c
```diff
@@ -1903,6 +1914,273 @@ QuicConnRestart(
...
+                "Resumption Ticket app data length failed to decode");
+            goto Error;
+        }
+
+        QUIC_CONNECTION_EVENT Event = { 0, };
+        Event.Type = QUIC_CONNECTION_EVENT_RESUMED;
+        Event.RESUMED.ResumptionStateLength = (uint16_t)AppTicketLength;
+        Event.RESUMED.ResumptionState = (AppTicketLength > 0) ? Ticket + Offset : NULL;
+        ResumptionAccepted =
+            QUIC_SUCCEEDED(QuicConnIndicateEvent(Connection, &Event));
```
nibanks, line 2136: At the very least, log an error/warning if the app didn't accept. But we might want a new event for the success case too.

src/core/connection.c
```diff
@@ -1903,6 +1914,273 @@ QuicConnRestart(
...
+    _In_reads_bytes_opt_(AppDataLength)
+        const uint8_t* AppResumptionData
+    )
+{
+    QUIC_STATUS Status;
+    uint32_t EncodedTransportParametersLength = 0;
+    uint8_t* TicketBuffer = NULL;
+    uint16_t AlpnLength = *(Connection->Crypto.TlsState.NegotiatedAlpn);
+
+    QUIC_TRANSPORT_PARAMETERS HSTPCopy = *Connection->HandshakeTP;
```
nibanks, line 1931: Since you don't bail if you fail to alloc in ApplySettings, this could be null.

src/core/connection.c
```diff
@@ -6101,6 +6388,20 @@ QuicConnProcessApiOperation(
...
 
+    case QUIC_API_TYPE_CONN_SEND_RESUMPTION_TICKET:
+        Status =
+            QuicConnSendResumptionTicket(
+                Connection,
+                ApiCtx->CONN_SEND_RESUMPTION_TICKET.AppDataLength,
+                ApiCtx->CONN_SEND_RESUMPTION_TICKET.ResumptionAppData);
+        ApiCtx->CONN_SEND_RESUMPTION_TICKET.ResumptionAppData = NULL;
+        if (ApiCtx->CONN_SEND_RESUMPTION_TICKET.Flags & QUIC_SEND_RESUMPTION_FLAG_FINAL) {
+            QUIC_DBG_ASSERT(QuicConnIsServer(Connection)); // TODO: handle client-side
```
nibanks, line 6399: I don't understand this TODO. Why would client ever be sending tickets?

src/core/connection.c
```diff
@@ -6101,6 +6388,20 @@ QuicConnProcessApiOperation(
...
+            Connection->State.ResumptionEnabled = FALSE;
+            QuicConnCleanupServerResumptionState(Connection); // BUG: With Async processing, this frees the memory before it's used.
```
nibanks, line 6401: I think we should do synchronous processing. Always require the data be ready to query right after.

src/core/quicdef.h
```diff
@@ -409,6 +409,12 @@ QUIC_STATIC_ASSERT(
 //
 #define QUIC_DEFAULT_SERVER_RESUMPTION_LEVEL    QUIC_SERVER_NO_RESUME
 
+//
+// Version of the wire-format for resumption tickets.
+// This needs to be incremented for each change in order or count of fields.
+//
+#define QUIC_TLS_RESUMPTION_TICKET_VERSION 1
```
nibanks, line 416: nit: tab over the `1` to match formatting

src/core/connection.c
anrossi: I imagined when client-side is implemented it would could reuse the same queued event type.
However, that doesn't make as much sense now as it did then. I'll remove the comment.

src/core/connection.c
```diff
@@ -1903,6 +1917,296 @@ QuicConnRestart(
...
+        Event.Type = QUIC_CONNECTION_EVENT_RESUMED;
+        Event.RESUMED.ResumptionStateLength = (uint16_t)AppTicketLength;
+        Event.RESUMED.ResumptionState = (AppTicketLength > 0) ? Ticket + Offset : NULL;
+        ResumptionAccepted =
+            QUIC_SUCCEEDED(QuicConnIndicateEvent(Connection, &Event));
+
+
+        if (ResumptionAccepted) {
+            QuicTraceEvent(
+                ConnServerResumeTicket,
```
nibanks, line 2153: Going to need an LTTng equivalent.

src/core/connection.c
anrossi: Agreed. I'll wire that up in the PR with the TLS changes.

ngo:
Sorry to revive this already-closed PR, but I find that this place is probably as good as any for this question. Reading the current tls_schannel.c code I can deduce that session resumption in msquic is not currently supported neither on server nor on client. I was wondering, does schannel itself currently offer such support for TLS1.3 or is it not implemented? I've done a number of tests on an up-to-date win10 with TLS1.3 support enabled in the registry, using a custom c app that uses schannel APIs directly and wasn't able to make it resume TLS1.3 sessions (whereas TLS1.2 sessions are resumed without problems using the same application). 
So, am I right in supposing that schannel currently lacks resumption support for TLS1.3 sessions?

Issue #404, opened 2020-05-24
Use platform independent enum for address family


rzikm:
Currently, the value of `AF_INET6` is system dependent. On Windows, it is 23 and on Linux, it is 10. When the library is used from a language like C#, which does not need separate compilation for different OSs, the caller needs to detect the running OS and select the correct value.

Since `AF_INET6` and related symbols are defined in OS headers, I suppose the solution would be defining a platform-independent enum for the address family.

nibanks:
In order to not have to dynamically translate (at runtime) the address family values to the correct values, I want to continue having platform specific values for these. IMO, the real problem here is that the C# abstraction layer isn't properly providing per-platform values. @jkotalik do you agree? If so, it might be better to move this issue to dotnet.

jkotalik:
What API is `AF_INET6` being passed in for? We could do the per-platform check in C# if necessary. However, from what it sounds like, maybe passing in an enum for IPV4, IPV6, etc may be better for this API?

nibanks:
Generally, QUIC is going to be a pass through for this. Eventually it will be passed down directly to the UDP socket.

rzikm:
>  IMO, the real problem here is that the C# abstraction layer isn't properly providing per-platform values. 

This is not about the current C# wrapper deep in .NET repo. I personally don't think it is ok for a cross-platform library to have different interface -- however slightly -- based on the platform, unless absolutely necessary.

This affects everyone writing code in a "portable" language like C#, Java, Python, and others capable of interop with C. It is likely that everyone will miss this difference and consider it a bug because such behavior is just unexpected, and use a platform check as a workaround for the issue.

I don't believe that avoiding runtime translation of this enum, in particular, justifies this. I would understand if it was some optional advanced feature available only on a particular system. However, address family is needed by every user.

If you still stand by your decision, could you please document any parts of the API that are platform-dependent in the way described in the original post? That way at least the unfortunate developers do not have to discover the discrepancies one by one.

nibanks:
Let's standardize around the Windows format/values.

ThadHouse:
So, I just spent a bit trying to do this, and its not as easy as it looks. Theres a lot of edge cases we'd have to handle here, and it gets especially hard both because of our helper functions, and because we don't actually ever manipulate the values in here internally. As an interop developer myself, I definitely agree with wanting this, but I think we need to discuss more if this is the right solution, or if theres another solution we should be taking that abstracts away even more.

[...]

PR #495, opened 2020-06-11
[Linux] Move from si_family in the union to sockaddr;

Maximus-:
**There is no functionality change in this PR**

This way we don't rely on the internal sockaddr structure ordering. On
darwin for example, si_family is not the first member of sockaddr. There's no
guarantee this structure will stay the same, and the proper way to do
this is to just use a sockaddr as the first member, which is guaranteed
to be the first member of both sockaddr_in and sockaddr_in6.

Note: quicping doesn't build currently.

```
/home/max/msquic/src/tools/ping/QuicPing.cpp: In function ‘void ParseClientCommand(int, char**)’:
/home/max/msquic/src/tools/ping/QuicPing.cpp:307:48: error: ‘QUIC_ADDR’ {aka ‘union QUIC_ADDR’} has no member named ‘si_family’
         case 4: PingConfig.Client.RemoteIpAddr.si_family = AF_INET; break;
                                                ^~~~~~~~~
/home/max/msquic/src/tools/ping/QuicPing.cpp:308:48: error: ‘QUIC_ADDR’ {aka ‘union QUIC_ADDR’} has no member named ‘si_family’
         case 6: PingConfig.Client.RemoteIpAddr.si_family = AF_INET6; break;
```

The fix here would be for platform-independent access to network-AF info (`QuicPlatformGetAddrAf`?). Please advise :)

Edit: quicping builds now. :)

### DIFF
--- a/src/inc/msquic_linux.h
+++ b/src/inc/msquic_linux.h
@@ -147,9 +147,9 @@ typedef struct addrinfo ADDRINFO;
 typedef sa_family_t QUIC_ADDRESS_FAMILY;
 
 typedef union QUIC_ADDR {
+    struct sockaddr Ip;
     struct sockaddr_in Ipv4;
     struct sockaddr_in6 Ipv6;
-    sa_family_t si_family;
 } QUIC_ADDR;
 
 #define FIELD_OFFSET(type, field)       ((uint32_t)(size_t)&(((type *)0)->field))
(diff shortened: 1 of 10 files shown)


ThadHouse:
We have a discussion over at https://github.com/microsoft/msquic/issues/404 for platform specifics in the public API. That one is just about the enum, but the sockaddr struct should be included in that discussion. I do agree we should look at finding some way to remove platform dependence from the public API.

ThadHouse:
For now, you can use `QuicAddrSetFamily`

Maximus-:
Ah, I didn't realize `QuicAddrGetFamily` existed :)

Will adjust quicping accordingly.

Maximus-:
Just noticed tests are failing... fixing

PR #506, opened 2020-06-13
Fix Version Negotiation Packet

nibanks:
Recent version refactoring broke the logic that builds up version negotiation packets. This PR fixes that logic.


nibanks:
@anrossi not really. The bug her was that we were sending too much. Our tests still passed because we still sent everything that's expected.

PR #507, opened 2020-06-14
Make kernel static analysis warnings fatal.

anrossi:


src/core/core.kernel.vcxproj
```diff
@@ -165,4 +166,4 @@
   </ItemDefinitionGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets" />
-</Project>
```
nibanks, line 168: You removed the newline

PR #508, opened 2020-06-15
0-RTT Server TLS implementation

anrossi:
This PR includes the final piece of the Server 0-RTT changes, the TLS implementation.
Also includes tests.


src/tools/ping/PingConnection.cpp
```diff
@@ -285,6 +287,13 @@ PingConnection::ProcessEvent(
             QuicConnection,
             (uint32_t)(ElapsedMicroseconds / 1000),
             (uint32_t)(ElapsedMicroseconds % 1000));
+
+        if (this->IsServer) {
+            if (!QUIC_SUCCEEDED(
+                    MsQuic->ConnectionSendResumptionTicket(QuicConnection, QUIC_SEND_RESUMPTION_FLAG_FINAL, 0, nullptr))) {
```
anrossi, line 293: Note: this succeeds even when 0-RTT the session parameter isn't set; investigate further why the default seems wrong.

anrossi:
Implement Stub TLS

src/inc/quic_tls.h
```diff
@@ -134,8 +154,8 @@ typedef enum QUIC_TLS_RESULT_FLAGS {
 
 typedef enum QUIC_TLS_DATA_FLAGS {
 
-    QUIC_TLS_CRYPTO_DATA,
-    QUIC_TLS_TICKET_DATA
+    QUIC_TLS_CRYPTO_DATA                = 0x01,
+    QUIC_TLS_TICKET_DATA                = 0x02
```
nibanks, line 158: Why?

src/tools/ping/PingConnection.cpp
nibanks: Can you also align the `MsQuic->Conn..` with the previous line and replaces `!QUIC_SUCCEEDED` with `QUIC_FAILED`?

src/tools/spin/spinquic.cpp
```diff
@@ -215,6 +215,33 @@ QUIC_STATUS QUIC_API SpinQuicHandleStreamEvent(HQUIC Stream, void * /* Context *
 QUIC_STATUS QUIC_API SpinQuicHandleConnectionEvent(HQUIC Connection, void * /* Context */, QUIC_CONNECTION_EVENT *Event)
 {
     switch (Event->Type) {
+    case QUIC_CONNECTION_EVENT_CONNECTED: {
+        int Selector = rand() % 3;
```
nibanks, line 219: Instead of calling `rand()` directly, can you use the helper from above (that everything else uses)?

src/tools/spin/spinquic.cpp
```diff
@@ -215,6 +215,33 @@ QUIC_STATUS QUIC_API SpinQuicHandleStreamEvent(HQUIC Stream, void * /* Context *
...
+            // Send ticket with too much data
+            //
+            DataLength = QUIC_MAX_RESUMPTION_APP_DATA_LENGTH + 1;
+        } else {
+            //
+            // Send ticket with no app data (no-op)
+            //
+        }
+        if (DataLength) {
+            Data = (uint8_t*)malloc(DataLength);
```
nibanks, line 238: Do other places in the code check for allocation failures? If this returns NULL, just set DataLength to zero.

src/inc/quic_tls.h
anrossi: Flags should start at 1 not at 0.

nibanks: I mean why does it need to be flags? Do you expect to somehow pass both types of data at the same time? If it's always going to be either/or then it doesn't need to be a flag.

src/core/crypto.c
```diff
@@ -1672,11 +1674,30 @@ QuicCryptoProcessAppData(
...
+
+    QUIC_TLS_RESULT_FLAGS ResultFlags =
+        QuicTlsProcessData(Crypto->TLS, QUIC_TLS_TICKET_DATA, AppData, &DataLength, &Crypto->TlsState);
+    if (ResultFlags & QUIC_TLS_RESULT_ERROR) {
+        return QUIC_STATUS_TLS_ERROR;
+    }
+
+    if (ResultFlags & QUIC_TLS_RESULT_PENDING) {
+        //
+        // Process async calls in-line.
```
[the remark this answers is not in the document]
anrossi: After offline discussion, this will be allowed to be async instead of in-line.

src/inc/quic_tls.h
anrossi: Good point. I thought we might want to have modifiers to the data in the future, so some flags would be exclusive, and others wouldn't. But we can make that change in the future, if we need to.

src/core/session.c
```diff
@@ -1111,6 +1111,7 @@ QuicSessionParamSet(
 
     case QUIC_PARAM_SESSION_SERVER_RESUMPTION_LEVEL: {
         if (BufferLength != sizeof(QUIC_SERVER_RESUMPTION_LEVEL) ||
+            Buffer == NULL ||
```
nibanks, line 1114: If this check is needed, it should be at a higher level. I thought the function that calls this already does this check.

anrossi: No, it appears this is left to each parameter to check individually. Which makes some sense because some parameters may want to use NULL as a sentinel value.

PR #509, opened 2020-06-16
Refactor Test Datapath Hooks

nibanks:
This does some refactoring of the test code on how it handles datapath hooks to allow for multiple simultaneous hooks to be used. It updates the test code, specifically the rebind tests to use this new hook.


nibanks:
Need to disable any test hook dependent tests in release mode.

src/core/quicdef.h
```diff
@@ -307,7 +307,7 @@ QUIC_STATIC_ASSERT(
 // The minimum number of bytes of send allowance we must have before we will
 // send another packet.
 //
-#define QUIC_MIN_SEND_ALLOWANCE                 100
+#define QUIC_MIN_SEND_ALLOWANCE                 75
```
ThadHouse, line 310: This seems unrelated, but not sure. Is this required for this change?

nibanks: It is required because of the rebinding test change. We now use the test hooks to simulate a UDP port change instead of explicitly changing sockets. The API to change socket address triggered a PING frame to be sent as well. Since that no longer happens, the next packet sent out is sometimes a lot smaller. Therefore, according to the amplification protection rules on the server (can only send at most 3x what they receive initially) the server's send allowance ended up being too small to send anything. So I had to reduce this min bar for that to get things working again.

PR #510, opened 2020-06-16
Add Database Server ASP.NET Core Source Code

ThadHouse:
No secrets are stored on GitHub, this is all safe to have publicly.


ThadHouse:
This also has PS scripts to create a new test and to create a new platform

src/perf/dbserver/.gitignore
```diff
@@ -0,0 +1,360 @@
+## Ignore Visual Studio temporary files, build results, and
+## files generated by popular Visual Studio add-ons.
+##
+## Get latest from https://github.com/github/gitignore/blob/master/VisualStudio.gitignore
```
nibanks, line 4: Why is this needed?

src/perf/dbserver/CreatePlatform.ps1
```diff
@@ -0,0 +1,37 @@
...
+This script creates a new platform on the DB Server
+
+.PARAMETER PlatformName
+    The Platform name to create
+
+.PARAMETER AuthKey
+    The Authorization Key to use
+
+#>
+
```
nibanks, line 13: nit: can we remove extra newlines (only have one) in these scripts?

src/perf/dbserver/QuicPerformanceDataServer/Controllers/PerformanceController.cs
```diff
@@ -0,0 +1,315 @@
...
+
+    [ApiController]
+    [Route("[controller]")]
+    public class PerformanceController : ControllerBase
+    {
+        private readonly PerformanceContext context;
+        private readonly ILogger<PerformanceController> logger;
+        private readonly IConfiguration configuration;
+
+        public PerformanceController(ILogger<PerformanceController> logger, PerformanceContext context,
```
nibanks, line 28: Can we try to match the style of the native code as much as possible please?

src/perf/dbserver/QuicPerformanceDataServer/Data/DbInitializer.cs
```diff
@@ -0,0 +1,99 @@
...
+            {
+                DbPlatformId = 2,
+                TestName = "loopback",
+            });
+
+            await context.SaveChangesAsync().ConfigureAwait(false);
+
+            using HttpClient client = new HttpClient();
+
+            var seedUri = new Uri("https://raw.githubusercontent.com/ThadHouse/msquic/dbseed/seeddata.json");
```
nibanks, line 64: Should this line/location be in here? Also, do we really need this initializer code checked in at all?

src/perf/dbserver/QuicPerformanceDataServer/Data/PerformanceContext.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Microsoft.EntityFrameworkCore;
+using QuicDataServer.Models.Db;
+
+namespace QuicDataServer.Data
+{
+    public class PerformanceContext : DbContext
+    {
+#pragma warning disable CS8618 // Non-nullable field is uninitialized. Consider declaring as nullable.
```
nibanks, line 8: These warning disable/enable are a bit annoying throughout. Possibly just better to ignore it in the project settings?

src/perf/dbserver/QuicPerformanceDataServer/Controllers/PerformanceController.cs
```diff
@@ -0,0 +1,315 @@
+﻿using Microsoft.AspNetCore.Http;
```
nibanks, line 1: Every code file needs the appropriate copy right. See https://docs.opensource.microsoft.com/content/releasing/copyright-headers.html#c-2

src/perf/dbserver/QuicPerformanceDataServer/Properties/serviceDependencies.MsQuicPerformanceResults - Web Deploy.json
```diff
@@ -0,0 +1,10 @@
+{
+  "dependencies": {
+    "mssql1": {
+      "resourceId": "/subscriptions/[parameters('subscriptionId')]/resourcegroups/[parameters('resourceGroup')]/providers/Microsoft.Sql/servers/quicperformanceresultsdbserver/databases/QuicPerformanceResults_db",
+      "type": "mssql.azure",
+      "connectionId": "DefaultConnection",
+      "secretStore": "AzureAppSettings"
+    }
+  }
+}
```
nibanks, line 10: Can we make GitHub happy and have all these files have a newline at the end?

src/perf/dbserver/QuicPerformanceDataServer/Controllers/PerformanceController.cs
ThadHouse: For C#, pretty much the standard is this format, and this is what happens when you auto format in VS and VS Code. I can change it, but unlike native code this is pretty standardized across all C# projects.

src/perf/dbserver/.gitignore
ThadHouse: Theres a lot of extra ignores for C# projects, especially ones published to azure. I can move most of these to the root gitignore, but usually I see subprojects like this using their own in case they want to get split out.

src/perf/dbserver/QuicPerformanceDataServer/Data/PerformanceContext.cs
ThadHouse: I don't want to ignore it completely in the project settings. There are a lot of cases where we do want this warning. Its just an inefficiency with data models created with reflection and NRTs.

src/perf/dbserver/QuicPerformanceDataServer/Properties/serviceDependencies.MsQuicPerformanceResults - Web Deploy.json
ThadHouse: Done

src/perf/dbserver/QuicPerformanceDataServer/Controllers/PerformanceController.cs
ThadHouse: Done

src/perf/dbserver/QuicPerformanceDataServer/Data/DbInitializer.cs
ThadHouse: Removed. I can always recreate it if necessary.

src/perf/dbserver/CreatePlatform.ps1
ThadHouse: Done

src/perf/dbserver/QuicPerformanceDataServer/Controllers/PerformanceController.cs
nibanks: @microsoft/quic what do other folks think? Match default C# or our code style? I tend to prefer all code in the same repo to follow the same style.

src/perf/dbserver/QuicPerformanceDataServer/Data/PerformanceContext.cs
nibanks: Hmmm... It makes the code quite ugly in my opinion. What if we just disable it for the whole file (at the top) for these data model files? Another thought, what if we put all the data model classes in a single file with the ignore?

src/perf/dbserver/.gitignore
nibanks: I thought our top level .gitignore had pretty much all of these. What exactly is missing from it that makes this one necessary?

src/perf/dbserver/QuicPerformanceDataServer/Data/PerformanceContext.cs
ThadHouse: Standard is 1 file per class, so I don't want to combine them into a single file. But I can just put the disable at the top of each one.

src/perf/dbserver/QuicPerformanceDataServer/QuicPerformanceDataServer.csproj
```diff
@@ -0,0 +1,30 @@
...
+    </PackageReference>
+    <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer" Version="3.1.5" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore.Tools" Version="3.1.5">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
+    <PackageReference Include="Swashbuckle.AspNetCore" Version="5.4.1" />
+  </ItemGroup>
+
```
nibanks, line 28: nit: remove extra newline (I know I'm a stickler for newlines. I'm OCD about them.)

src/perf/dbserver/QuicPerformanceDataServer/Data/PerformanceContext.cs
nibanks: From my experience with C# it hasn't been that uncommon to have multiple related classes in a single file. Generally, I agree one per file, but for things like this, combining seem better to me.

src/perf/dbserver/.gitignore
ThadHouse: Turns out it does. I thought It was missing the azure ignores. Removed.

src/perf/dbserver/QuicPerformanceDataServer/QuicPerformanceDataServer.csproj
ThadHouse: Done

src/perf/dbserver/QuicPerformanceDataServer/Data/PerformanceContext.cs
ThadHouse: Actually, never pushed that combine, but got a suggestion for a better way to do this without disabling the warning. So done.

src/perf/.config/dotnet-tools.json
```diff
@@ -0,0 +1,12 @@
...
+  "isRoot": true,
+  "tools": {
+    "dotnet-format": {
+      "version": "4.1.131201",
+      "commands": [
+        "dotnet-format"
+      ]
+    }
+  }
+}
```
nibanks, line 12: newline at end please.

src/perf/.editorconfig
```diff
@@ -0,0 +1,188 @@
+# editorconfig.org
```
nibanks, line 1: This file should be at the top level probably, in case we have C# code outside of perf.

PR #513, opened 2020-06-17
Test Refactoring and Improvements

nibanks:
Refactoring the test code to improve the flow and reduce duplication where possible. More PRs likely to follow after this.

### DIFF
--- a/src/inc/msquic_linux.h
+++ b/src/inc/msquic_linux.h
@@ -317,9 +317,9 @@ QuicAddrIncrement(
     )
 {
     if (Addr->Ip.sa_family == AF_INET) {
-        ((uint8_t*)&Addr->Ipv4.sin_addr)[3]++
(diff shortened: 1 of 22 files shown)


src/inc/msquic_linux.h
```diff
@@ -307,6 +307,22 @@ QuicAddrIsBoundExplicitly(
...
+inline
+void
+QuicAddrIncrement(
+    _Inout_ QUIC_ADDR * Addr
+    )
+{
+    if (Addr->Ip.sa_family == AF_INET) {
+        ((uint8_t*)&Addr->Ipv4.sin_addr)[3]++;
+    } else {
+        ((uint8_t*)&Addr->Ipv6.sin6_addr)[15]++;
```
anrossi, line 322: is overflow acceptable?

nibanks: Yes. The goal here is really just to produce a different IP from the original. I thought about randomizing too, but this was simpler.

PR #524, opened 2020-06-18
Enable building on Azure for Raspberry Pi with a cross compiler

### LINKED ISSUE #122: Build Different CPU Architectures on Linux
There is some initial cmake code to build different architectures but it is not implemented for Linux. We need to get x86, x64 and arm working, possibly arm64 (if possible).

ThadHouse:
Cross compiling for arm has a lot of issues. For C libraries things work pretty well usually, and the library will likely run on many Arm platforms. However, C++ this is not the case. Different platforms compile their base setup with incompatible flags around the C++ ABI. In addition, cross compilers are usually setup in a sub optimal configuration, limiting cpu support even further. To solve this, the volunteer work I've done we created our own cross compiler specific to the Raspberry Pi. This will work on all models of Pi, including the ArmV6 versions (The Zero and 1). Because of how the C library works, the library itself should work on any ArmV6 or ArmV7 CPU (Or V8 running 32 bit OS) with hard float capabilities. But the C++ test libraries likely would only run on the Pi. 

Resolves #122. If more build configurations are needed in the future, we can handle them with separate issues/PRs.

### DIFF
--- a/.azure/templates/build-config-user-container.yml
+++ b/.azure/templates/build-config-user-container.yml
@@ -20,13 +20,9 @@ jobs:
 
   - script: |
       wget -q https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb
-
       sudo dpkg -i packages-microsoft-prod.deb
-
       sudo apt-get update
-
       sudo add-apt-repository universe
-
       sudo apt-get install -y powershell
     displayName: Install Powershell
 
--- a/.azure/templates/build-config-user-container.yml
+++ b/.azure/templates/build-config-user-container.yml
@@ -29,13 +29,6 @@ jobs:
     displayName: Install Powershell
 
   - task: PowerShell@2
-    displayName: Prepare Build Machine
-    inputs:
-      pwsh: true
-      filePath: scripts/prepare-machine.ps1
-      arguments: -Configuration Build
-
-  - task: PowerShell@2
     displayName: Build Source Code (Debug)
     inputs:
       pwsh: true
--- a/.azure/templates/build-config-user-container.yml
+++ b/.azure/templates/build-config-user-container.yml
@@ -18,7 +18,7 @@ jobs:
   - checkout: self
     submodules: recursive
 
-  - shell: |
+  - script: |
       wget -q https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb
 
       sudo dpkg -i packages-microsoft-prod.deb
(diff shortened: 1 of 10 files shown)


ThadHouse:
This is now ready to go. I can add more explanation on how the toolchain files work if you would like. Basically, you would need one per cross platform you want to support. So if we wanted an arm64 linux build we'd need to add a toolchain file and a cross compiler for that.

.azure/templates/build-config-user-container.yml
```diff
@@ -0,0 +1,54 @@
...
+  - script: |
+      wget -q https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb
+
+      sudo dpkg -i packages-microsoft-prod.deb
+
+      sudo apt-get update
+
+      sudo add-apt-repository universe
+
+      sudo apt-get install -y powershell
```
nibanks, line 30: I don't think the extra lines are necessary.

CMakeLists.txt
```diff
@@ -238,14 +238,26 @@ include_directories(${CMAKE_CURRENT_SOURCE_DIR}/src/inc)
...
+    if (CMAKE_SYSTEM_PROCESSOR STREQUAL arm)
+            add_custom_command(
+            WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/submodules/openssl
+            OUTPUT ${QUIC_BUILD_DIR}/openssl/include
+            OUTPUT ${QUIC_BUILD_DIR}/openssl/lib/libcrypto${CMAKE_STATIC_LIBRARY_SUFFIX}
+            OUTPUT ${QUIC_BUILD_DIR}/openssl/lib/libssl${CMAKE_STATIC_LIBRARY_SUFFIX}
+            COMMAND SYSTEM=${CMAKE_HOST_SYSTEM_NAME} ./Configure linux-generic32 --cross-compile-prefix=${GNU_MACHINE}${FLOAT_ABI_SUFFIX}- enable-tls1_3 --prefix=${QUIC_BUILD_DIR}/openssl
+            COMMAND make clean
+            COMMAND make -j$$(nproc)
+            COMMAND make install_sw)
```
nibanks, line 250: tab over

ThadHouse:
This is actually still a fairly small amount of root files relative to other open source projects. Most others have a ton more. VS Code, Roslyn and .NET Core are all well above what we are at. Agree its a lot, but there isn't much that can be done. Maybe moving the cmake and scripts folders into a utilities folder. But that doesn't help a whole ton.

ThadHouse:
I tested these builds on a Pi, and they work correctly. Do we want to build Stub for Pi as well, or just openssl?

nibanks:
OpenSSL is just fine.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

PR #525, opened 2020-06-18
Add Support for PGO Azure Pipeline

nibanks:
This adds a new yaml file for a new (on demand) Azure Pipeline for building and running a PGO instrumented performance test.

Current on-going run: https://dev.azure.com/ms/msquic/_build/results?buildId=88992&view=results


scripts/build.ps1
```diff
@@ -232,6 +232,22 @@ function CMake-Build {
         if (!$DisableTools) {
             Copy-Item (Join-Path $BuildDir "obj" $Config "msquicetw.lib") $ArtifactsDir
         }
+        if ($PGO -and $Config -eq "Release") {
+            # TODO - Figure out a better way to get the path?
+            $VCToolsPath = "C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Tools\MSVC\14.26.28801\bin\Host$Arch\$Arch"
```
nibanks, line 237: Apparently these paths don't exist on AZP machines. @ThadHouse any ideas where they'd be?

nibanks:
Additionally, @ThadHouse any ideas about this error from AZP:
```
Running Loopback Test
ForEach-Object : The variable '$Matches' cannot be retrieved because it has not been set.
At C:\a\w\2\s\scripts\performance.ps1:142 char:13
+     1..10 | ForEach-Object {
```

scripts/build.ps1
ThadHouse: AZP uses enterprise VS. So `C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise` instead of Buildtools

ThadHouse:
Since thats locally, what happened is the test failed for some reason.

scripts/performance.ps1
```diff
@@ -143,31 +149,42 @@ function Run-Loopback-Test() {
...
-    $ToPublishResults.IndividualRunResults = $allRunsResults
-
-    $ResultsFolderRoot = "$Platform/loopback"
-    $ResultsFileName = "/results.json"
-
-    $NewFilePath = Join-Path $RootDir "artifacts/PerfDataResults/$ResultsFolderRoot"
-    $NewFileLocation = Join-Path $NewFilePath $ResultsFileName
-    New-Item $NewFilePath -ItemType Directory -Force
+    if ($PGO) {
+        $Artifacts = Join-Path $RootDir "\artifacts\windows\$($Arch)_$($Config)_$($Tls)"
```
ThadHouse, line 155: Should we be doing this after Run-Loopback-Test is called? That way when we add more tests, they'll automatically be PGO'd too?

nibanks: Ah, You're right. I'll fix this

PR #526, opened 2020-06-18
Publish machine information with performance results

ThadHouse:
A WIP PR, but need to know if the machine name gets reported correctly. Then will start adding things.


ThadHouse:
This has been done in #573

Issue #530, opened 2020-06-19
OpenSSL Crypto Perf is Inefficient


nibanks:
### Describe the bug

I was looking at the OpenSSL crypto code (`QuicEncrypt`, `QuicDecrypt`, `QuicHpComputeMask`) and noticed a number of inefficiencies. For one, they all allocate a new `EVP_CIPHER_CTX` on every call! Additionally, `QuicHpComputeMask` is not processing the whole buffer at once, but instead doing a loop.

### Recommended Changes

Since QUIC never uses key structs in parallel, OpenSSL abstraction layer should allocate the `EVP_CIPHER_CTX` with the key struct. Also, `QuicHpComputeMask` should process the whole input in a single operation (no loops).

PR #531, opened 2020-06-19
Perf Improvement for OpenSSL Header Key

nibanks:
One step closer for #530. This PR improves perf for the header key, though it's still using a loop; but at least the cipher context is preallocated, and reused for each call.

### DIFF
--- a/src/platform/tls_openssl.c
+++ b/src/platform/tls_openssl.c
@@ -161,6 +161,11 @@ typedef struct QUIC_HP_KEY {
     const EVP_CIPHER *Aead;
 
     //
+    // The cipher context to use for encryption/decryption.
+    //
+    EVP_CIPHER_CTX *CipherCtx;
+
+    //
     // Buffer and BufferLen of the key.
     //
     int BufferLen;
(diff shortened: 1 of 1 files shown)


src/platform/tls_openssl.c
```diff
@@ -2127,18 +2134,46 @@ QuicHpComputeMask(
...
-                Mask + i * QUIC_HP_SAMPLE_LENGTH,
-                Key->Buffer,
-                Key->BufferLen,
-                Cipher + i * QUIC_HP_SAMPLE_LENGTH,
-                Key->Aead)) {
-            return QUIC_STATUS_TLS_ERROR;
+    BOOLEAN Ret = FALSE;
+    int Len = 0;
+    uint32_t Offset = 0;
+    static const uint8_t PLAINTEXT[] = "\x00\x00\x00\x00\x00";
```
anrossi, line 2140: This can be a plain `uint8_t` array that is zero initialized.

src/platform/tls_openssl.c
```diff
@@ -2127,18 +2134,46 @@ QuicHpComputeMask(
...
+    for (uint8_t i = 0; i < BatchSize; ++i) { // TODO - Figure out how to not use a loop here!
+        if (EVP_EncryptInit_ex(Key->CipherCtx, Key->Aead, NULL, Key->Buffer, Cipher + Offset) != 1) {
+            QuicTraceEvent(
+                LibraryError,
+                "[ lib] ERROR, %s.",
+                "EVP_EncryptInit_ex failed");
+            goto Exit;
         }
+
+        if (EVP_EncryptUpdate(Key->CipherCtx, Mask + Offset, &Len, PLAINTEXT, sizeof(PLAINTEXT) - 1) != 1) {
```
anrossi, line 2151: If you make the above suggestion, you can remove the `- 1` for the terminating NULL here.

PR #533, opened 2020-06-19
Allow datapath hook failure happen in spinquic

### LINKED ISSUE #532: Release mode spinquic fails
### Describe the bug

When running spinquic in release mode, all tested OS's fail with the following error.

`D:\a\1\s\src\tools\spin\spinquic.cpp:765 MsQuic->SetParam( nullptr, QUIC_PARAM_LEVEL_GLOBAL, QUIC_PARAM_GLOBAL_TEST_DATAPATH_HOOKS, sizeof(Value), &Value) failed!`

The failure is `QUIC_STATUS_INVALID_PARAMETER` on all platforms

1. Run spinquic release mode
2. View output
3. See error

The test to run.

ThadHouse:
In release mode, datapath hooks can't be enabled. Because we want to support debug spinquic with release libmsquic, we want to allow the failure rather then conditional compile.

Closes #532

### DIFF
--- a/src/tools/spin/spinquic.cpp
+++ b/src/tools/spin/spinquic.cpp
@@ -756,7 +756,8 @@ main(int argc, char **argv)
 
         if (Settings.LossPercent != 0) {
             QUIC_TEST_DATAPATH_HOOKS* Value = &DataPathHooks;
-            if (QUIC_FAILED(MsQuic->SetParam(
+            if (QUIC_FAILED(
+                MsQuic->SetParam(
                     nullptr,
                     QUIC_PARAM_LEVEL_GLOBAL,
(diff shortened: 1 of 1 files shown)


src/tools/spin/spinquic.cpp
```diff
@@ -756,13 +756,14 @@ main(int argc, char **argv)
 
         if (Settings.LossPercent != 0) {
             QUIC_TEST_DATAPATH_HOOKS* Value = &DataPathHooks;
-            EXIT_ON_FAILURE(
-                MsQuic->SetParam(
+            if (QUIC_FAILED(MsQuic->SetParam(
                     nullptr,
```
nibanks, line 760:
```suggestion
            if (QUIC_FAILED(
                MsQuic->SetParam(
                    nullptr,
```

PR #538, opened 2020-06-19
Move All Scripts to Top Level Scripts Folder

nibanks:
Didn't really make sense to have two separate folders.


ThadHouse:
Things like run-executable and run-gtest are fine, since they could be arbitrarily used, but the azure specific ones I think should stay in .azure.

nibanks:
> Things like run-executable and run-gtest are fine, since they could be arbitrarily used, but the azure specific ones I think should stay in .azure.

The line is blurred between local and Azure for a lot of them. I still think it's best to have one folder, even for the ones that are clearly Azure only.

nibanks:
Ping @microsoft/quic

PR #540, opened 2020-06-19
Enable Application Verifier on Windows BVTs.

anrossi:


nibanks:
Still timing out after 45 min. Too long for a PR CI run. Keep increasing the timeout to see how long it actually takes, then we can decide if we might want to enable this on master commits.

nibanks:
As mentioned on IM, let's scope this to the internal CI for now.

scripts/prepare-machine.ps1
```diff
@@ -44,6 +46,18 @@ if ($IsWindows) {
...
+        where.exe appverif.exe
+        if ($LastExitCode -eq 0) {
+            appverif.exe /verify msquiccoretest.exe
+            appverif.exe /verify msquicplatformtest.exe
+            appverif.exe /verify msquictest.exe
+        } else {
+            Write-Debug "Application Verifier not installed!"
+        }
+    }
+
```
nibanks, line 60: Is it possible we could do this as a part of running the test, and not configuring the machine? The internal machines never even reboot (ideally), so these persist forever.

anrossi: It'll require the test script to run as administrator, and it'll blur the responsibilities of the scripts, but sure.  If the internal machines never reboot, why would having these settings always present be a problem?

nibanks: If you make it past of the test runner, then you can undo it when the test is done.

scripts/test.ps1
```diff
@@ -119,9 +119,14 @@ param (
...
 
     [Parameter(Mandatory = $false)]
-    [switch]$NoProgress = $false
+    [switch]$NoProgress = $false,
+
+    [Parameter(Mandatory = $false)]
+    [switch]$EnableAppVerifier = $false
 )
 
+#Requires -RunAsAdministrator
```
nibanks, line 128: I'd rather remove this. The script only requires admin if  you use the enable appverifier flag, which won't be used by most people, most of the time, so I'd still like them to be able to run the script without.

scripts/test.ps1
```diff
@@ -193,7 +198,30 @@ if ($NoProgress) {
...
+    where.exe appverif.exe
+    if ($LastExitCode -eq 0) {
+        appverif.exe /verify msquiccoretest.exe
+        appverif.exe /verify msquicplatformtest.exe
+        appverif.exe /verify msquictest.exe
+    } else {
+        Write-Debug "Application Verifier not installed!"
+        $EnableAppVerifier = $false;
+    }
+}
```
nibanks, line 212: I feel like it makes more sense to put this into the run-gtest.ps1 script (and probably run-executable.ps1 script).

scripts/test.ps1
```diff
@@ -193,7 +198,30 @@ if ($NoProgress) {
...
 
+if ($EnableAppVerifier) {
+    # Enable Application Verifier for test binaries
+    where.exe appverif.exe
+    if ($LastExitCode -eq 0) {
+        appverif.exe /verify msquiccoretest.exe
+        appverif.exe /verify msquicplatformtest.exe
+        appverif.exe /verify msquictest.exe
+    } else {
+        Write-Debug "Application Verifier not installed!"
```
nibanks, line 209: Let's use `Write-Warning` instead

scripts/test.ps1
anrossi: Is there a way to make this conditional on the appverifier flag being set?

anrossi: Also, the script still requires admin even without the appverifier flag because of the WER dump path setting. And that's not gated behind any switch as far as I can tell.

scripts/test.ps1
anrossi: That will add about ~4 seconds to each invocation however...

scripts/test.ps1
nibanks: No. There's no way to gate this. It's always on (on all platforms). That's why I think it best to run remove, and warn/error out if a particular feature is used that requires admin.

scripts/test.ps1
nibanks: So, 4 seconds for each EXE, not per test case. The total time should be unchanged right? We're just moving setting all three in batch here, to doing it individually in run-gtest.ps1.

.azure/templates/run-bvt-int.yml
```diff
@@ -26,16 +26,16 @@ jobs:
     inputs:
       pwsh: true
       filePath: scripts/prepare-machine.ps1
-      arguments: -Configuration Test
+      arguments: -Configuration Test -EnabledAppVerifier
```
nibanks, line 29: No longer necessary, right?

nibanks:
```suggestion
      arguments: -Configuration Test
```

PR #541, opened 2020-06-19
Dump Results on Error

nibanks:
We seem to be getting some failures with the new perf machines. Need more info to debug.


nibanks:
Ping @ThadHouse. This was helpful for debugging.

PR #552, opened 2020-06-25
Add cmake changes necessary for Darwin builds

Maximus-:
This adds QUIC_PLATFORM Darwin option, and sets -DQUIC_PLATFORM_DARWIN. Also fixes a typo which had `--std=c++17`, instead of `-std=c++17`.  Adds -std=c++11 which was necessary for me to build the gtest suite. Currently this has everything building for debug as well, which I will address in a later commit as a flag. 

The next PR will have introduce the platform code, and then the datapath code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,8 +100,10 @@ endif()
 
 if ("${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Windows")
     set(QUIC_PLATFORM "windows")
-else()
+elseif ("${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Linux")
     set(QUIC_PLATFORM "linux")
+elseif ("${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Darwin")
+    set(QUIC_PLATFORM "darwin")
 endif()
 
 set(QUIC_BUILD_DIR ${CMAKE_CURRENT_BINARY_DIR})
@@ -206,7 +208,13 @@ if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
 
 else()
     # Custom build flags.
-    set(QUIC_COMMON_FLAGS "-DQUIC_PLATFORM_LINUX -fms-extensions -fPIC -pthread -Wl,--no-as-needed -ldl")
+
+    set(QUIC_COMMON_FLAGS "-fms-extensions -fPIC -pthread -g -O0 -fno-inline -fno-omit-frame-pointer -glldb")
+    if (QUIC_PLATFORM STREQUAL "darwin")
+        set(QUIC_COMMON_FLAGS "${QUIC_COMMON_FLAGS} -DQUIC_PLATFORM_DARWIN -Wno-microsoft-anon-tag -Wno-tautological-constant-out-of-range-compare -Wmissing-field-initializers")
+    else()
+        set(QUIC_COMMON_FLAGS "${QUIC_COMMON_FLAGS} -DQUIC_PLATFORM_LINUX -Wl,--no-as-needed -ldl")
+    endif()
 
     if(QUIC_ENABLE_LOGGING)
         include(FindLTTngUST)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -229,15 +237,15 @@ else()
 
     if(QUIC_SANITIZE_ADDRESS)
         message(STATUS "Configuring address sanitizer")
-        set(QUIC_COMMON_FLAGS "${QUIC_COMMON_FLAGS} -fsanitize=address -fno-omit-frame-pointer -Wno-maybe-uninitialized -O0 -Og -g")
```
Maximus-, line 232: -Og actually does optimizations near -O1, so prefer -O0 over it.

Maximus-:
Noticed breakage with -glldb. Fixing now.

CMakeLists.txt
```diff
@@ -206,7 +208,13 @@ if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
 
 else()
     # Custom build flags.
-    set(QUIC_COMMON_FLAGS "-DQUIC_PLATFORM_LINUX -fms-extensions -fPIC -pthread -Wl,--no-as-needed -ldl")
+
+    set(QUIC_COMMON_FLAGS "-fms-extensions -fPIC -pthread -g -O0 -fno-inline -fno-omit-frame-pointer")
```
ThadHouse, line 212: This is going to force -g -O0 for all builds, even release which we don't want. We should let cmake use its automated handling to handle optimization flags.

Maximus-: Sounds good; will remove :)

CMakeLists.txt
ThadHouse: I'm fine with this. For asan, O0 should be fine.

CMakeLists.txt
```diff
@@ -308,7 +316,10 @@ if(QUIC_BUILD_TEST)
     set(INSTALL_GTEST OFF CACHE BOOL "Enable installation of googletest. (Projects embedding googletest may want to turn this OFF.)")
     if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
         option(gtest_force_shared_crt "Use shared (DLL) run-time lib even when Google Test is built as static lib." ON)
+    else()
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
```
ThadHouse, line 320: Why 11, and not 17? Did you do this change before fixing the bad `--std=c++17`?  I wonder if fixing that flag will make this not required.

Maximus-:
Adding the rest of my CMakeFiles.. I forgot there were more..

CMakeLists.txt
ThadHouse: Ah I didn't realize that was a different set of flags. All good, other then as long as it compiles I'd prefer 17 in great to not get any potential incompatibilities. I know there won't be any, but I always like building everything the same way.

Maximus-: No complaints there; Will adjust

CMakeLists.txt
```diff
@@ -206,7 +208,13 @@ if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
 
 else()
     # Custom build flags.
-    set(QUIC_COMMON_FLAGS "-DQUIC_PLATFORM_LINUX -fms-extensions -fPIC -pthread -Wl,--no-as-needed -ldl")
+
+    set(QUIC_COMMON_FLAGS "-fms-extensions -fPIC -pthread -fno-inline -fno-omit-frame-pointer")
```
ThadHouse, line 212: @Maximus- Were -fno-inline and -fno-omit-frame-pointer required for macOS? If so those should only be added to mac. We don't want these flags unconditionally on linux.

Maximus-: Not at all; only there for debugging. Will remove. My mistake!

ThadHouse: Got the PR in #554 already, so we're good. Just wanted to know if it was something mac required so I could add it to mac specifically.

PR #558, opened 2020-06-27
SpinQuic Explicit ALPN Improvements

nibanks:
Updates SpinQuic to not randomize the ALPN if there is only a single session; and defaults the session count to 1 if there is an explicit ALPN passed in.

### DIFF
--- a/src/tools/spin/spinquic.cpp
+++ b/src/tools/spin/spinquic.cpp
@@ -761,7 +761,9 @@ main(int argc, char **argv)
             Settings.Ports = std::vector<uint16_t>({dstPort});
         }
         TryGetValue(argc, argv, "target", &Settings.ServerName);
-        TryGetValue(argc, argv, "alpn", &Settings.AlpnPrefix);
+        if (TryGetValue(argc, argv, "alpn", &Settings.AlpnPrefix)) {
+            SessionCount = 1; // Default session count to 1 if ALPN explicitly specified.
+        }
         TryGetValue(argc, argv, "sessions", &SessionCount);
     }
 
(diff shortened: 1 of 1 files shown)


src/tools/spin/spinquic.cpp
```diff
@@ -818,9 +814,30 @@ main(int argc, char **argv)
...
+        } else {
+            QUIC_BUFFER AlpnBuffer;
+            AlpnBuffer.Length = (uint32_t)strlen(Settings.AlpnPrefix) + 1; // You can't have more than 2^8 SessionCount. :)
+            AlpnBuffer.Buffer = (uint8_t*)malloc(AlpnBuffer.Length);
+            EXIT_ON_NOT(AlpnBuffer.Buffer);
+            memcpy(AlpnBuffer.Buffer, Settings.AlpnPrefix, AlpnBuffer.Length);
+
+            for (uint32_t i = 0; i < SessionCount; i++) {
+
+                AlpnBuffer.Buffer[AlpnBuffer.Length-1] = (uint8_t)i;
```
Maximus-, line 827: Shouldn’t this be (uint8_t)(i + 0x30) so at least the ALPN is always ascii?

nibanks: There is no requirement that the ALPN be ASCII.

ThadHouse: Is it utf8, or encoding agnostic? If spec requires utf8, then this will generate invalid code points? Also, if encoding agnostic what about embedded nulls, which this could put in.

nibanks: It's a raw byte buffer. No string encoding. The fact that ASCII characters are generally used (usually without a NULL) is beside the point.

PR #560, opened 2020-06-28
Support Disable 1-RTT Encryption Extension

nibanks:
This PR is a prototype new QUIC extension that allows the client and server to negotiate the disabling of encryption post-handshake. There are two practical scenarios for this:

- Performance Testing - to be able to do A:B testing to measure the "cost of encryption".
- Trusted Environments - to disable encryption in scenarios where there is already trust. For example, a walled garden or QUIC already in an encrypted tunnel.

WIP draft spec for this available here: https://github.com/nibanks/quic-disable-encryption


src/inc/msquic.h
```diff
@@ -403,6 +403,9 @@ typedef enum QUIC_SERVER_RESUMPTION_LEVEL {
 #define QUIC_PARAM_CONN_STREAM_SCHEDULING_SCHEME        20  // QUIC_STREAM_SCHEDULING_SCHEME
 #define QUIC_PARAM_CONN_DATAGRAM_RECEIVE_ENABLED        21  // uint8_t (BOOLEAN)
 #define QUIC_PARAM_CONN_DATAGRAM_SEND_ENABLED           22  // uint8_t (BOOLEAN)
+#ifdef QUIC_API_ENABLE_INSECURE_FEATURES
```
ThadHouse, line 406: Is there a particular reason to hide this behind an ifdef? Just like the validation flag, hiding those behind a flag makes writing common libraries much harder, since even if you include what you use, if you get included after the first use of msquic.h, you still fail. I think a much better option if we want to hide these is an `msquic_unsafe.h` header, that can be included separately.

nibanks: We want to make it hard to use this. We don't want people shooting themselves in the foot with this accidentally. But I also don't want to use a separate file for it.

nibanks: I might also end up putting any flags/code to disable server certificate validation under the same flag.

ThadHouse: Ok. It just makes it extremely difficult to even write tests, because you have to be very careful about include order. Thats why QUIC_TEST_APIS has to be defined in the high level files usually, rather then helper headers. But its not the end of the world, just makes things a bit painful sometimes.

nibanks: Yeah, I prefer it this way for now.

PR #562, opened 2020-06-29
Merge macOS support (and platform renaming)

Maximus-:
This PR is kind of hefty, so let me describe the changes.

- Introduces QUIC_PLATFORM_DARWIN
- Adds macOS support as a "feature-based" datapath (i.e. datapath_kqueue.c instead of datapath_darwin.c)
   - Subsequently, this also renames `datapath_linux.c` to `datapath_epoll.c`. 
   - As well as the cmake changes necessary for this to build
- Merge darwin platform support into `platform_posix.c`. 
- Annotate most places that have "Linux" as "Linux and Darwin" or "Posix" (or both). Unclear how you guys want this to sound. There are still places in the codebase that have just "Linux". 

- `{ 0 }` is a build error on my clang. I don't know why. I moved all of the error-ing occurrences of `{ 0 }` to `{ }` to fix the build issue. This _should_ have the same behavior of full-initialization. 

I've tested this on both Linux & macOS locally. 

What's likely left (and I probably won't have time for)
- Azure pipeline support for macOS / iOS.
- Figuring out if it's worthwhile to multi-thread kqueue. (`ProcCount` > 1)


src/platform/platform_posix.c
```diff
@@ -549,7 +555,19 @@ QuicProcCurrentNumber(
     void
     )
 {
+#if defined(QUIC_PLATFORM_LINUX)
     return (uint32_t)sched_getcpu();
+#elif defined(QUIC_PLATFORM_DARWIN)
+    int cpuinfo[4];   
+    asm("cpuid"
```
Maximus-, line 562: I haven't found a better way to do this... :( 
This should probably get gated by `#if __x86_64__` for iOS.

src/bin/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
     endif()
     SET_TARGET_PROPERTIES(msquic
         PROPERTIES LINK_FLAGS "/DEF:\"${CMAKE_CURRENT_SOURCE_DIR}/winuser/msquic.def\"")
-else()
+elseif (QUIC_PLATFORM STREQUAL "linux")
     SET_TARGET_PROPERTIES(msquic
         PROPERTIES LINK_FLAGS "-Wl,--version-script=\"${CMAKE_CURRENT_SOURCE_DIR}/linux/exports.txt\"")
 endif()
```
ThadHouse, line 35: Theres a macOS equivalent of this setup. 

The file is just a single line per symbol, prefixed with `_`. Place that file in darwin/exports.txt. That file is then passed to the linker with `-exported_symbols_list ${CMAKE_CURRENT_SOURCE_DIR}/darwin/exports.txt`

src/inc/quic_trace.h
```diff
@@ -123,9 +123,1124 @@ QuicTraceRundown(
 
 #ifdef QUIC_EVENTS_STUB
 
+#define QUIC_TRACE_EVENT(x, ...)
```
nibanks, line 126: @maximus- why all this?

src/platform/platform_posix.c
```diff
@@ -26,7 +26,7 @@
 
 #define QUIC_MAX_LOG_MSG_LEN        1024 // Bytes
 
-#ifdef QUIC_PLATFORM_DISPATCH_TABLE
+#if QUIC_PLATFORM_DISPATCH_TABLE
```
nibanks, line 29: Change this back?

nibanks:
I've been playing around with the build flags trying to get around this macOS build error:
```
/Users/runner/work/1/s/src/inc/msquic_posix.h:375:36: error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
        const IN6_ADDR ZeroAddr = {0};
                                   ^
                                   {}
```
Any suggestions?

ThadHouse:
Try a double brace. Sometimes clang requires that. {{0}}

src/bin/CMakeLists.txt
nibanks: @thadhouse would you mind pushing your suggested change to this branch?

ThadHouse: I will once it builds successfully. I don't want to interfere with getting that working.

nibanks:
> Try a double brace. Sometimes clang requires that. {{0}}

It didn't work.

ThadHouse:
So clang is a lot more picky about being correct to the standard, and `{ 0 }` doesn't look to always be allowed for locals. But looking at some disassembly, 0 initialization is also achieved with just `{}` so try that.  Theres a bunch of places that needs to be updated though, we have a lot of initialization in that style.

You can also do an explicit memset, compilers do a very good job of optimizing the memset.

nibanks:
@ThadHouse  `{}` is what @maximus- had in this PR originally, but it broke all the Windows builds; so that's a no-go.

Maximus-:
Yeah I think { 0 } is incorrect for structures with more than 1 member or with nested types. Worst case scenario, instead of emitting a memset everywhere, can do something like struct a b = DEFAULT_INITIALIZER and have that behind a platform macro... :(

nibanks:
> can do something like struct a b = DEFAULT_INITIALIZER and have that behind a platform macro

I thought about that but ewwww.... I'm still hoping for a compiler flag solution here. All other platforms are fine with the current status quo...

ThadHouse:
Hmm. I wonder why disabling the warning in QUIC_COMMON_FLAGS doesn't work. Adding it to `CLANG_GCC_WARNING_FLAGS ` works weirdly

nibanks:
> Hmm. I wonder why disabling the warning in QUIC_COMMON_FLAGS doesn't work. Adding it to `CLANG_GCC_WARNING_FLAGS ` works weirdly

Can you push that to the branch?

scripts/build.ps1
```diff
@@ -175,8 +179,10 @@ function CMake-Generate {
             "arm"   { $Arguments += "arm" }
             "arm64" { $Arguments += "arm64" }
         }
-    } else {
+    } elseif ($IsLinux) {
         $Arguments += " 'Linux Makefiles'"
+    } elseif ($IsMacOS) {
```
Maximus-, line 184: I've been trying to keep everything as "Darwin" so it works on macOS + iOS, but making the distinction between macOS and iOS might be useful... unclear

ThadHouse: This is a variable built into Powershell. Building for iOS would be a cross compilation, which would need to be set up separately.

nibanks: Yeah, let's keep them separate at this level for now. If/when iOS support is added, we can revisit this decision.

src/inc/quic_trace.h
Maximus-: I didn't have a way to enable the tracing on macOS so I ended up forwarding this all to a printf call and this was just part of the hack. I could've sworn I removed it...

ThadHouse:
Should we make this run at least some tests?

Maximus-:
> Should we make this run at least some tests? Or is this not in a running state and just a building state?

It's fully functional afaict; quicinterop gives mostly the same results that I get on Linux. (Not the same interop results as the build pipeline has, but at least similar results from my local machine).

.azure/azure-pipelines.ci.yml
```diff
@@ -134,6 +134,19 @@ stages:
...
+  dependsOn: []
+  jobs:
+  # Officially supported configurations.
+  - template: ./templates/build-config-user.yml
+    parameters:
+      image: macOS-10.15
+      platform: macos
+      arch: x64
+      tls: openssl
+      extraBuildArgs: -DisableLogs
```
nibanks, line 148: In our CI, I want logging enabled. For local builds, if people don't want to use clog, they can disable.

CMakeLists.txt
```diff
@@ -116,6 +116,11 @@ if (${CMAKE_VERSION} VERSION_LESS "3.6.0")
     set(QUIC_ENABLE_LOGGING OFF)
 endif()
 
+if(QUIC_PLATFORM STREQUAL "darwin)
+    message(WARNING "macOS does not support logging currently")
+    set(QUIC_ENABLE_LOGGING OFF)
```
nibanks, line 121: Why not? CLOG was running before this.

ThadHouse: CLOG was, but we were always passing DisableLogs because lttng isn't supported.

ThadHouse: But I can fix that a different way

ThadHouse:
I've got the latest master merged into this. However, it has a major problem. It appears something broke recently, and I can't figure out how to get openssl to build. Its missing std library include headers, like stdio.h. @Maximus- any ideas or clues how to solve this?

nibanks:
> I've got the latest master merged into this. However, it has a major problem. It appears something broke recently, and I can't figure out how to get openssl to build. Its missing std library include headers, like stdio.h. @Maximus- any ideas or clues how to solve this?

Perhaps @larseggert has an idea? I'd be willing to bet it's related to his refactoring of openssl builds.

larseggert:
I will take a look tomorrow.

larseggert:
I fixed the openssl issue (and some other nits) in #753. it compiles now, but there are test failures and crashes.

larseggert:
I think I fixed this in #759 (at least that builds locally), but for some reason, the CI run for it tried to pull from the already-closed #758, failed, and cannot be restarted now. Help?

larseggert:
The schannel CI failure is due to lack of disk space, otherwise looks hopefully good now.

nibanks:
Looks like the build is good (thanks @larseggert!) but the tests still have issues. @Maximus- do you expect to have time to look into these any time soon?

larseggert:
I had a quick look, and the test failures seem to be due to `kqueue` issues. Not sure if this is of much help.

Maximus-:
I'm debugging some of them daily. There's an issue where sending a UDP packet to ::1 returns EAFNOSUPPORT, which I dont fully understand. It's on an AF_INET6 socket, but my hosts file indicates that ::1 goes to localhost.. I don't know if that's an issue since that's a V4 address. Reading kernel code in the meantime 😃

larseggert:
'localhost' is a name, and resolves to the v4 address `127.0.0.1` and the v6 address `::1`. Are you sure you are picking the right one after resolution?

Maximus-:
> `localhost` is a name, and resolves to the v4 address `127.0.0.1` and the v6 address `::1`. Are you sure you are picking the right one after resolution?

While I don't understand why the default hosts file has `::1 localhost`, I've resolved this, it was a different issue :)

larseggert:
I'm a bit lost where we are with this PR. Is there anything I can do to help this along?

nibanks:
@Maximus- was trying to fix the data path layers so that our tests stopped failing. I'm not sure what the latest status is.

Maximus-:
I essentially copied and refactored the winuser datapath. Having some strange issues that need hunting down. Been OOF and preoccupied in my free time unfortunately. Will spend some more time on it this week. quicinterop works, at the very least. :)

nibanks:
This PR has stagnated, mostly because the folks involved don't have the time necessary to complete it. I propose the following:

- Close this PR
- Create a new feature branch: `feature/mac-os`
- Push all the latest code to that branch (both main and these changes).
- Folks can start making PRs on that branch to fix individual parts.
- Once everything is green in that branch, we can create a PR to merge it to main.

@thadhouse, @Maximus- are you Ok with that?

ThadHouse:
I'm good with that.

Maximus-:
That's fine with me :-) I'll hopefully have time to revisit this after the Nov. 3rd election.

nibanks:
FYI, the branch `feature/mac-os` has been created and is up to date with mainline. Folks (@thadhouse, @maximus- @larseggert) feel free to send PRs for that branch if you want to work on the MacOS stuff. It has full Azure Pipeline support for any PR into it.

PR #567, opened 2020-06-30
Allow QuicPing server to specify number of connections allowed before shutting down

ThadHouse:
By doing this, performance tests can be cleaner to run, as they can explicitly state how many connections are expected. Also adds parity with the client

### DIFF
--- a/src/tools/ping/PingConnection.cpp
+++ b/src/tools/ping/PingConnection.cpp
@@ -47,8 +47,7 @@ PingConnection::PingConnection(
 
 PingConnection::PingConnection(
     _In_ PingTracker* Tracker,
-    _In_ HQUIC Connection,
-    _In_ int OverloadToken
+    _In_ HQUIC Connection
     ) :
     Tracker(Tracker), QuicConnection(Connection), DumpResumption(false),
     ConnectedSuccessfully(false), BytesSent(0), BytesReceived(0), DatagramLength(0),
--- a/src/tools/ping/PingConnection.cpp
+++ b/src/tools/ping/PingConnection.cpp
@@ -33,19 +33,6 @@ PingConnection::PingConnection(
 }
 
 PingConnection::PingConnection(
-    _In_ HQUIC Connection
-    ) :
-    Tracker(nullptr), QuicConnection(Connection), DumpResumption(false),
-    ConnectedSuccessfully(false), BytesSent(0), BytesReceived(0), DatagramLength(0),
-    DatagramsSent(0), DatagramsAcked(0), DatagramsLost(0), DatagramsCancelled(0),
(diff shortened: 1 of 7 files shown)


src/tools/ping/QuicPing.cpp
```diff
@@ -377,6 +378,13 @@ main(
         goto Error;
     }
 
+    //
+    // 0 to the server means infinite. For client, 0 will be coorced into
+    // DEFAULT_CLIENT_CONNECTION_COUNT
+    //
+    TryGetValue(argc, argv, "connections", &connections);
+    PingConfig.ConnectionCount = connections;
+
```
nibanks, line 387: Instead of doing it this way, put this code into `ParseCommonCommands` and in `ParseClientCommand` always default the count to `DEFAULT_CLIENT_CONNECTION_COUNT` before calling the function.

src/tools/ping/Server.cpp
```diff
@@ -11,10 +11,13 @@
 
 #include "QuicPing.h"
 
+
```
nibanks, line 14: nit: remove extra line

src/tools/ping/PingConnection.h
```diff
@@ -110,6 +121,17 @@ struct PingConnection {
...
 
+    //
+    // Constructor for incoming connection with tracker.
+    // Overload token is because otherwise this matches the first constructor
+    // which is for clients.
+    //
+    PingConnection(
+        _In_ PingTracker* Tracker,
+        _In_ HQUIC Connection,
+        _In_ int OverloadToken
```
nibanks, line 132: Instead of doing this, remove the default for `DumpResumption` in the client constructor.

src/tools/ping/PingConnection.cpp
```diff
@@ -45,6 +45,20 @@ PingConnection::PingConnection(
     MsQuic->SetCallbackHandler(Connection, (void*)QuicCallbackHandler, this);
 }
 
+PingConnection::PingConnection(
+    _In_ PingTracker* Tracker,
+    _In_ HQUIC Connection
```
nibanks, line 50: Why didn't you just modify the old constructor? Also, did you update the existing client callers to pass the non-default param now?

src/tools/ping/Server.cpp
```diff
@@ -48,7 +50,7 @@ struct PingServer {
         ) {
         switch (Event->Type) {
         case QUIC_LISTENER_EVENT_NEW_CONNECTION: {
-            auto Connection = new PingConnection(Event->NEW_CONNECTION.Connection);
+            auto Connection = new PingConnection(&Tracker, Event->NEW_CONNECTION.Connection, 42);
```
nibanks, line 53: Remove the 42

PR #570, opened 2020-07-01
Disable Send Buffering for Perf

nibanks:
When looking at a trace for some of the latest perf runs I noticed some CPU from send buffering. Let's disable that in our automated tests to see how much benefit we get.

Results from latest perf run with these changes:

Stub (i.e. no crypto):
```
Median: 14931989 kbps (+64.34%)
Master: 9085911 kbps
```
Schannel:
```
Median: 5818387 kbps (+23.08%)
Master: 4727407 kbps
```
FYI, @maolson-msft & @pravb here is the cost of send buffering. It's a lot more than I expected.

### DIFF
--- a/scripts/performance.ps1
+++ b/scripts/performance.ps1
@@ -107,7 +107,7 @@ if ($IsWindows) {
 
 # QuicPing arguments
 $ServerArgs = "-listen:* -port:4433 -selfsign:1 -peer_uni:1 -connections:$Runs"
-$ClientArgs = "-target:localhost -port:4433 -uni:1 -length:$Length"
+$ClientArgs = "-target:localhost -port:4433 -sendbuf:0 -uni:1 -length:$Length"
 if ($IsWindows) {
     # Always use the same local address and core to provide more consistent results.
     $ClientArgs += " -bind:127.0.0.1:4434 -ip:4 -core:0"


maolson-msft:
I assume "median" means median after the change and "master" means median before the change?

Also, you both suggested that we need to check it in to see the result, and also that you've already run an A/B test?

nibanks:
> I assume "median" means median after the change and "master" means median before the change?

Yes "median" is the median with the change, and "master" is the median without the change.

> Also, you both suggested that we need to check it in to see the result, and also that you've already run an A/B test?

The change had to be checked in to a branch and pushed to be able to run our automated perf pipeline on it. Since the results were positive, we merged it into master.

maolson-msft:
The schannel results are definitely interesting.

Does the throughput-testing tool actually touch its payload bytes?

nibanks:
> Does the throughput-testing tool actually touch its payload bytes?

No, it does not.

maolson-msft:
Then the test results here are suspect. Real apps will touch their payloads, which has a significant effect on the performance of copying those bytes due to CPU caching.

nibanks:
Feel free to propose a change to quicping (with a control knob) to "touch" the buffer and we can at least run an A:B test, if not onboard a run for both.

PR #573, opened 2020-07-02
Add multi system test platform infrastructure

ThadHouse:
Uses a json file to configure which tests are available. The runner then knows which tests it can run. It has a remote, located at `quic-server` that allows use to configure the remote machine properly. 

The script has a local and remote tls and arch provider, and also allows multiple operating systems to be supported as well. Part of this is still a WIP, but the general infrastructure is there.

Running local tests in theory is also supported, but will require a small bit more work to make fully working.


.azure/azure-pipelines.ci.yml
```diff
@@ -198,6 +198,28 @@ stages:
 # Performance Tests
 #
 
+- stage: remoteperformance
+  displayName: Remote Performance Testing
+  dependsOn:
+  - build_windows
+  jobs:
+  - template: ./templates/run-remote-performance-int.yml
```
nibanks, line 206: Let's get rid of the `-int`. I just refactored most of them out with my last PR.

scripts/RemoteTests.json
```diff
@@ -0,0 +1,58 @@
...
+        "TestName": "Throughput",
+        "Remote" : {
+            "Platform": "Linux",
+            "Tls": ["stub", "openssl"],
+            "Arch": ["x64", "arm"],
+            "Exe": "quicping",
+            "Arguments": {
+                "All": "-listen:* -port:4433 -peer_uni:1 -connections:10",
+                "Loopback": "-selfsign:1",
+                "Remote": "-thumbprint:$Thumbprint -machine_cert:1 -cert_store:My"
```
nibanks, line 40: This doesn't actually work on Linux. `-selfsign` should still work for Linux.
```suggestion
                "Remote": "-selfsign:1"
```

scripts/performance.ps1
```diff
@@ -278,6 +277,10 @@ function Run-Loopback-Test() {
     $serverOutput = $null
     try {
 
+        Write-Host $QuicPing
+        Write-Host $ServerArgs
+        Write-Host $ClientArgs
```
nibanks, line 282: I assume temporary? If so please remember to remove. If not, please use `Write-Debug`

scripts/remoteperformance.ps1
```diff
@@ -0,0 +1,567 @@
...
+    [switch]$Local = $false
+)
+
+Set-StrictMode -Version 'Latest'
+$PSDefaultParameterValues['*:ErrorAction'] = 'Stop'
+
+# Root directory of the project.
+$RootDir = Split-Path $PSScriptRoot -Parent
+
+Get-PSSession | Remove-PSSession
```
nibanks, line 80:
```suggestion
# Remove any previous remote PowerShell sessions.
Get-PSSession | Remove-PSSession
```

scripts/remoteperformance.ps1
```diff
@@ -0,0 +1,567 @@
...
+            $session = New-PSSession -ComputerName quic-server -ConfigurationName PowerShell.7
+        }
+    } else {
+        $session = New-PSSession -HostName "$Remote"
+    }
+
+    $RemoteAddress = $session.ComputerName
+
+    if ($null -eq $session) {
+        exit
```
nibanks, line 119: Can you use a `Write-Error` with a message instead?

scripts/remoteperformance.ps1
```diff
@@ -0,0 +1,567 @@
...
+
+        $PossibleRemoteIPs | ForEach-Object {
+            [ipaddress]$RemoteIpAddr = $_.IPAddressToString
+            $ToMaskRemoteAddress = HostToNetworkOrder($RemoteIpAddr)
+            $RemoteMasked = $ToMaskRemoteAddress -band $Mask
+
+            if ($RemoteMasked -eq $LocalSubnet) {
+                $MatchedIPs += $LocalIpAddr.IPAddressToString
+            }
+        }
```
nibanks, line 146: IMO, this logic is overly complicated. I think we should instead just rely on an ENV var, just like the thumbprint. It would reduce the complexity of this script considerably.

scripts/remoteperformance.ps1
```diff
@@ -0,0 +1,567 @@
...
+    if ($MatchedIPs.Length -ne 1) {
+        Write-Host "Failed to parse local address. Using first address"
+    }
+    $LocalAddress = $MatchedIPs[0]
+
+    Write-Host "Connected to: $RemoteAddress"
+    Write-Host "Local IP Connection $LocalAddress" 
+}
+
+function Invoke-Test-Command {
```
nibanks, line 158: What do you think about renaming this to `Invoke-TestCommand` instead?

scripts/remoteperformance.ps1
```diff
@@ -0,0 +1,567 @@
...
+            return Invoke-Command -Session $Session -ScriptBlock $ScriptBlock -AsJob -ArgumentList $ArgumentList
+        }
+        return Invoke-Command -Session $Session -ScriptBlock $ScriptBlock -ArgumentList $ArgumentList
+    }
+    
+}
+
+$OutputDir = Join-Path $RootDir "artifacts/PerfDataResults"
+New-Item -Path $OutputDir -ItemType Directory -Force | Out-Null
+
```
nibanks, line 182: nit: Remove extra newline here and anywhere else you have two together.

scripts/remoteperformance.ps1
```diff
@@ -0,0 +1,567 @@
...
+    while ($true) {
+        $CurrentResults = Receive-Job -Job $Job -Keep
+        if (![string]::IsNullOrWhiteSpace($CurrentResults)) {
+            $DidMatch = $CurrentResults -match $Matcher
+            if ($DidMatch) {
+                return $true
+            }
+        }
+        Start-Sleep -Seconds 0.1 | Out-Null
+        if ($StopWatch.ElapsedMilliseconds -gt 10000) {
```
nibanks, line 213: Put this in the `while` condition instead. Then, after the loop, return false.

scripts/remoteperformance.ps1
```diff
@@ -0,0 +1,567 @@
...
+    $RetVal = Receive-Job -Job $Job
+    return $RetVal -join "`n"
+}
+
+function Copy-Artifacts {
+    param([string]$From, [string]$To)
+    try {
+        Invoke-Test-Command $session -ScriptBlock { param($To) Remove-Item -Path "$To/*" -Recurse -Force } -ArgumentList $To
+    } catch {
+        # Ignore failure
```
nibanks, line 231: Can this comment elaborate why?

scripts/remoteperformance.ps1
```diff
@@ -0,0 +1,567 @@
...
+    $ExeName = $TestPlat.Exe
+    if ($Platform -eq "windows") {
+        $ExeName += ".exe"
+    }
+
+    
+
+    if ($IsRemote) {
+        $ConfigStr = "$($RemoteArch)_$($Config)_$($RemoteTls)"
+        return Invoke-Test-Command -Session $session -ScriptBlock { param($PathRoot, $Platform, $ConfigStr, $ExeName) Join-Path $PathRoot $Platform $ConfigStr $ExeName  } -ArgumentList $PathRoot, $Platform, $ConfigStr, $ExeName
```
nibanks, line 321: line is too long

scripts/remoteperformance.ps1
```diff
@@ -0,0 +1,567 @@
...
+}
+
+class TestPublishResult {
+    [string]$PlatformName
+    [string]$TestName
+    [string]$CommitHash
+    [double[]]$IndividualRunResults
+}
+
+$currentLoc = Get-Location
```
nibanks, line 298: This script scatters a lot of stuff at the global execution scope throughout. Can we clean it up some? If there are global variables that need to be initialized for everything, put them at the top. Otherwise all the global execution goes as the end?

.azure/templates/create-package.yml
```diff
@@ -12,12 +12,12 @@ jobs:
     inputs:
       artifactName: drop
       itemPattern: drop/win*/*_schannel/**
-      downloadPath: $(Build.TempDirectory)
+      downloadPath: $(Agent.TempDirectory)
```
nibanks, line 15: Might be worth making a quick separate PR just to make these changes.

scripts/performance-helper.psm1
```diff
@@ -0,0 +1,292 @@
+function SetGlobals {
```
nibanks, line 1: Does this script need the following, or will it inherit it from the importer?
```
```

scripts/performance-helper.psm1
```diff
@@ -0,0 +1,292 @@
...
+    } else {
+        if ($AsJob) {
+            return Invoke-Command -Session $Session -ScriptBlock $ScriptBlock -AsJob -ArgumentList $ArgumentList
+        }
+        return Invoke-Command -Session $Session -ScriptBlock $ScriptBlock -ArgumentList $ArgumentList
+    }   
+}
+
+function Wait-ForRemoteReady {
+    param ($Job, $Matcher)
```
nibanks, line 76: Can we keep a consistent format for `param`, like the function above?

scripts/remoteperformance.ps1
```diff
@@ -0,0 +1,298 @@
...
+}
+
+$LocalDataCache = LocalSetup
+
+try {
+    $Tests = Get-Tests $TestsFile
+
+    if ($null -eq $Tests) {
+        Write-Output "Tests are not valid"
+        exit
```
nibanks, line 279: Instead of calling exit, just use `Write-Error`

.azure/azure-pipelines.ci.yml
```diff
@@ -203,6 +203,23 @@ stages:
...
+    parameters:
+      localTls: schannel
+      remoteTls: schannel
+      extraArgs: '-Publish -Record'
+  - template: ./templates/run-remote-performance-int.yml
+    parameters:
+      localTls: schannel
+      remoteTls: schannel
+      extraName: '_Loopback'
```
nibanks, line 220: Can we pass this without the `_` and then add the `_` as necessary in the template?

scripts/remoteperformance.ps1
```diff
@@ -0,0 +1,298 @@
...
+    # Check both Exes
+    $RemoteExeExists = Invoke-TestCommand -Session $Session -ScriptBlock {
+        param ($RemoteExe)
+        Test-Path $RemoteExe
+    } -ArgumentList $RemoteExe
+
+    $LocalExeExists = Test-Path $LocalExe
+
+    if (!$RemoteExeExists -or !$LocalExeExists) {
+        Write-Output "Failed to Run $Test because of missing exe"
```
nibanks, line 214: Make this a `Write-Error` and move this down and replace the `return` with it. Generally, replace all the `Write-Output` for error messages with `Write-Error` so things immediately terminate.

scripts/remoteperformance.ps1
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    if ($Record) {
+        Get-RemoteFile -From ($RemoteExe + ".remote.etw") -To (Join-Path $OutputDir $Test.ToString() + ".etw")
+    }
+
+    Publish-TestResults -Test $Test `
+                        -AllRunsResults $AllRunsResults `
+                        -CurrentCommitHash $CurrentCommitHash `
+                        -OutputDir $OutputDir
```
nibanks, line 269: Only if `$Publish` is `$true`?

scripts/performance-helper.psm1
```diff
@@ -0,0 +1,414 @@
...
+
+            $WpaStackWalkProfileXml | Out-File $EtwXmlName
+            Write-Host $EtwXmlName
+            wpr.exe -start $EtwXmlName -filemode 2> $null
+        }
+
+        & $Exe ($RunArgs).Split(" ")
+
+        if ($Record -and $IsWindows) {
+            $EtwName = $Exe + ".remote.etw"
```
nibanks, line 325: The extension is `.etl` not `.etw`.

scripts/remoteperformance.ps1
ThadHouse: No, we still want to do parts of this, as this also does the printing. Internally it checks for Publish before actually writing the file.

nibanks: So it's not just publishing results? We should rename this to a more appropriate name then.

scripts/remoteperformance.ps1
ThadHouse: I feel like I would rather do it right rather then hardcoding a required env var. I don't like requiring the cert either as it makes testing harder, and after I get done with this I plan on doing some debugging into why PS remoting on those systems specifically doesn't handle self signed certs, as I feel like that might somehow be a bug. With it being a separate function in a separate file now, its much cleaner and easier to maintain.

scripts/remoteperformance.ps1
ThadHouse: So, for PS functions to not complain when they get imported as a module, you have to use an approved list of verbs. 

https://docs.microsoft.com/en-us/powershell/scripting/developer/cmdlet/approved-verbs-for-windows-powershell-commands?view=powershell-7

I can't really find a better one in this list. And it does 2 things. It publishes to the output, and it optionally publishes the results file. So in a way, it is publish anyway.

nibanks: Ok.

.azure/templates/run-remote-performance-int.yml
```diff
@@ -0,0 +1,67 @@
...
+    parameters:
+      platform: windows
+      arch: ${{ parameters.arch }}
+      tls: ${{ parameters.localTls }}
+
+  - template: ./download-artifacts.yml
+    parameters:
+      platform: windows
+      arch: ${{ parameters.arch }}
+      tls: ${{ parameters.remoteTls }}
```
nibanks, line 38: You should be able to wrap this with a conditional and only execute it when `parameters.localTls` is `ne` to `parameters.remoteTls`.

scripts/remoteperformance.ps1
```diff
@@ -0,0 +1,296 @@
...
+    try {
+        if ($IsWindows) {
+            $apipaAddr = Get-NetIPAddress 169.254.*
+            if ($null -ne $apipaAddr) {
+                # Disable all the APIPA interfaces for URO perf.
+                Write-Debug "Temporarily disabling APIPA interfaces"
+                $RetObj.apipaInterfaces = (Get-NetAdapter -InterfaceIndex $apipaAddr.InterfaceIndex) | Where-Object {$_.AdminStatus -eq "Up"}
+                $RetObj.apipaInterfaces | Disable-NetAdapter -Confirm:$false
+            }
+        }
```
nibanks, line 180: This is only necessary for loopback tests. Don't do it on remote tests.

scripts/remoteperformance.ps1
```diff
@@ -0,0 +1,296 @@
...
+    # Check both Exes
+    $RemoteExeExists = Invoke-TestCommand -Session $Session -ScriptBlock {
+        param ($RemoteExe)
+        Test-Path $RemoteExe
+    } -ArgumentList $RemoteExe
+
+    $LocalExeExists = Test-Path $LocalExe
+
+    if (!$RemoteExeExists -or !$LocalExeExists) {
+
```
nibanks, line 214: nit: remove newline

scripts/remoteperformance.ps1
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    $RemoteArguments = $Test.Remote.Arguments.GetArguments().Replace('$Thumbprint', $CertThumbprint)
+
+    $RemoteJob = Invoke-RemoteExe -Exe $RemoteExe -RunArgs $RemoteArguments
+
+    $ReadyToStart = Wait-ForRemoteReady -Job $RemoteJob
+
+    if (!$ReadyToStart) {
+
```
nibanks, line 238: nit: remove newline

scripts/remoteperformance.ps1
```diff
@@ -0,0 +1,296 @@
...
+        Stop-Job -Job $RemoteJob
+        Write-Error "Test Remote for $Test failed to start"
+    }
```
nibanks, line 241: Remove the spaces between these, and put a comment at the beginning; something to the effect of "Starting the server"

scripts/remoteperformance.ps1
```diff
@@ -0,0 +1,296 @@
...
+
+    $AllRunsResults = @()
+
+    try {
+        1..$Test.Iterations | ForEach-Object {
+            $LocalResults = Invoke-LocalExe -Exe $LocalExe -RunArgs $LocalArguments
+
+            $LocalParsedResults = Get-TestResult -Results $LocalResults -Matcher $Test.ResultsMatcher
+
+            $AllRunsResults += $LocalParsedResults
```
nibanks, line 251: nit: you don't need all these extra newlines between things. In general, don't put newlines between single line code blocks.

.azure/templates/run-remote-performance-int.yml
```diff
@@ -0,0 +1,67 @@
+# This template contains steps to run performance tests for a single configuration.
```
nibanks, line 1: Rename the file and remote the remote and int from the name

scripts/performance-helper.psm1
```diff
@@ -0,0 +1,410 @@
...
+    $Bytes = $Address.GetAddressBytes()
+    [Array]::Reverse($Bytes) | Out-Null
+    return [System.BitConverter]::ToUInt32($Bytes, 0)
+}
+
+function Get-LocalAddress {
+    param ($RemoteAddress)
+
+    $PossibleRemoteIPs = [System.Net.Dns]::GetHostAddresses($RemoteAddress) | Select-Object -Property IPAddressToString
+
```
nibanks, line 68: nit: remove newline

scripts/performance-helper.psm1
```diff
@@ -0,0 +1,410 @@
...
+    $PossibleLocalIPs = Get-NetIPAddress -AddressFamily IPv4 | Select-Object -Property IPv4Address, PrefixLength
+
+    $MatchedIPs = @()
+
```
nibanks, line 72: nit: remove newline

scripts/performance-helper.psm1
```diff
@@ -0,0 +1,410 @@
...
+    param ([string]$From, [string]$To)
+    try {
+        Invoke-TestCommand $Session -ScriptBlock {
+            param ($To)
+            Remove-Item -Path "$To/*" -Recurse -Force
+        } -ArgumentList $To
+    } catch {
+        # Ignore failure, which occurs when directory does not exist
+    }
+    # TODO Figure out how to filter this
```
nibanks, line 151: Finish this up before merging?

scripts/performance-helper.psm1
```diff
@@ -0,0 +1,410 @@
...
+class TestDefinition {
+    [string]$TestName;
+    [ExecutableSpec]$Remote;
+    [ExecutableSpec]$Local;
+    [int]$Iterations;
+    [string]$RemoteReadyMatcher;
+    [string]$ResultsMatcher;
+
+    [string]ToString() {
+        $RetString = ("{0}_{1}_{2}_{3}" -f $this.TestName,
```
nibanks, line 186: `{${$this.TestName}}_{${$this.Remote.Platform}}_{${$script:RemoteTls}}_{${$script:RemoteArch}}`

What do you think about using this format instead? I'm not a big fan of the `-f`  method.

scripts/performance-helper.psm1
```diff
@@ -0,0 +1,410 @@
...
+    $BasePath = Invoke-TestCommand -Session $Session -ScriptBlock {
+        param ($Exe)
+        if (!$IsWindows) {
+            chmod +x $Exe
+            return Split-Path $Exe -Parent
+        }
+        return $null
+    } -ArgumentList $Exe
+
+    Write-Debug "Running Remote: $Exe $RunArgs" | Out-Null
```
nibanks, line 301: Why do you have a `Write-Debug` but then have a `Out-Null`?

.azure/azure-pipelines.ci.yml
```diff
@@ -203,6 +203,23 @@ stages:
 # Performance Tests
 #
 
+- stage: remoteperformance
```
nibanks, line 206: So practically, we should be replacing the performance templates & scripts with this, right? Let's remove all those and replace them.

ThadHouse: I'll need to make sure it still handles linux. If it does we can do that.

ThadHouse: Also, this doesn't have PGO in it yet. I can work on adding that.

.azure/templates/run-remote-performance.yml
```diff
@@ -0,0 +1,74 @@
...
+    inputs:
+      contents: '**/artifacts/*'
+
+  - template: ./download-artifacts.yml
+    parameters:
+      platform: windows
+      arch: ${{ parameters.arch }}
+      tls: ${{ parameters.localTls }}
+
+  ${{ if ne(parameters.localTls, parameters.remoteTls) }}:
```
nibanks, line 40: I think you need a `-` before this, but I'm not sure.

.azure/azure-pipelines.ci.yml
nibanks: Remove the word "remote" from everything when you're done.

ThadHouse:
I think all the review stuff is done, other then removing remote. I want to wait on that, because its going to make this PR much harder to review, since the scripts will be changed rather then new. Once everything is good, I'll do the rename. Finally feature complete now though.

nibanks:
Can you make sure the perf AZP YAML is updated accordingly as well please?

ThadHouse:
Thats been done too. Its using the remote script, but I'll change that when I rename things

.azure/azure-pipelines.ci.yml
```diff
@@ -203,38 +203,92 @@ stages:
...
-  - template: ./templates/run-performance.yml
+  - template: ./templates/run-remote-performance.yml
     parameters:
       pool: MsQuic-Win-Perf
       platform: windows
-      tls: stub
-  - template: ./templates/run-performance.yml
+      localTls: schannel
+      remoteTls: schannel
+      extraArgs: '-Publish -Record'
```
nibanks, line 218: By default, we probably shouldn't use the `-Record` option.

ThadHouse: Done

ThadHouse:
Whoops. Accidentally pushed the clog submodule update. I could revert, but I'd rather just merge that fix before this.

ThadHouse:
Ok. Renamed away from remote.

PR #576, opened 2020-07-06
Fix azure build temp directory

ThadHouse:
Build.TempDirectory is not a variable, and just gets used as the folder name. Agent.TempDirectory is unique per test

### DIFF
--- a/.azure/templates/download-artifacts.yml
+++ b/.azure/templates/download-artifacts.yml
@@ -14,7 +14,7 @@ steps:
     downloadPath: $(Agent.TempDirectory)
 
 - task: DownloadBuildArtifacts@0
-  displayName: Download�CLOG Artifacts
+  displayName: Download CLOG Artifacts
   inputs:
     artifactName: drop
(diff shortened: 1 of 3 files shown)


.azure/templates/download-artifacts.yml
```diff
@@ -11,19 +11,19 @@ steps:
   inputs:
     artifactName: drop
     itemPattern: drop/${{ parameters.platform }}/${{ parameters.arch }}_*_${{ parameters.tls }}/**
-    downloadPath: $(Build.TempDirectory)
+    downloadPath: $(Agent.TempDirectory)
 
 - task: DownloadBuildArtifacts@0
-  displayName: Download�CLOG Artifacts
+  displayName: Download�CLOG Artifacts
```
nibanks, line 17: ??? Undo this ???

ThadHouse: The original commit must have used a UTF-16 marker there or something, which VS Code got confusing on when saving. Fixed.

PR #579, opened 2020-07-07
Update clog submodule to fix output overwriting

ThadHouse:
Switches the raspbian build to ninja, which will likely test this much better.

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -45,6 +45,9 @@ This script provides helpers for building msquic.
 .PARAMETER PGO
     Builds msquic with profile guided optimization support (Windows-only).
 
+.PARAMETER Generator
+    Specifies a specific cmake generator (Only supported on unix)
+
 .EXAMPLE
     build.ps1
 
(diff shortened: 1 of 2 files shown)


scripts/build.ps1
```diff
@@ -95,7 +98,10 @@ param (
     [switch]$DynamicCRT = $false,
 
     [Parameter(Mandatory = $false)]
-    [switch]$PGO = $false
+    [switch]$PGO = $false,
+
+    [Parameter(Mandatory = $false)]
+    [string]$Generator = "Linux Makefiles"
```
nibanks, line 104: Instead of defaulting for Linux, you should default to empty, and below set platform specific generators if it's empty.

scripts/build.ps1
```diff
@@ -95,12 +98,25 @@ param (
...
+
+    [Parameter(Mandatory = $false)]
+    [string]$Generator = ""
 )
 
 Set-StrictMode -Version 'Latest'
 $PSDefaultParameterValues['*:ErrorAction'] = 'Stop'
 
+if ($IsWindows -and ($Generator -ne "")) {
+    Write-Error "Generator setting not supported on windows"
```
nibanks, line 111: Why not default to `Visual Studio 16 2019`?

ThadHouse: I can default to that, but theres the arch arguments to which have to be set afterwards. I can do that by adding a second variable I guess.

nibanks: Just implicitly do the arch argument. Just let this change the generator.

ThadHouse: Fixed. Should be good to go now.

PR #584, opened 2020-07-09
Increase Max Partition Count from 64 to 128

nibanks:
The bulk of this change comes down to handling the +64 processor scenario. In this case, we need to set the proper group affinity before setting the processor index. Determining the mapping to these values is quite involved. The platform now generates and persists a global table for this information to be used efficiently later.

Only Windows user mode is updated here.


src/platform/platform_winuser.c
```diff
@@ -21,6 +21,8 @@
 uint64_t QuicPlatformPerfFreq;
 uint64_t QuicTotalMemory;
 QUIC_PLATFORM QuicPlatform = { NULL };
+QUIC_PROCESSOR_INFO* QuicProcessorInfo;
```
ThadHouse, line 24: Don't we need to free this in QuicPlatformUninitialize?

nibanks: Already done in the latest commit. :)

src/platform/platform_winuser.c
```diff
@@ -55,6 +57,142 @@ QuicPlatformSystemUnload(
...
+                        goto FindNumaNode;
+                    }
+                    ProcessorOffset += Info->Group.GroupInfo[i].ActiveProcessorCount;
+                }
+            }
+            Offset += Info->Size;
+        }
+
+        //printf("Failed to determine group!\n");
+        goto Error;
```
ThadHouse, line 156: If we hit this, we leak QuicNumaMasks

src/core/quicdef.h
```diff
@@ -105,7 +105,7 @@ typedef struct QUIC_PATH QUIC_PATH;
 //
 // Maximum number of partitions to support.
 //
-#define QUIC_MAX_PARTITION_COUNT                64
+#define QUIC_MAX_PARTITION_COUNT                128
```
ThadHouse, line 108: Any specific reasons now to limit this to 128? EPYC is now at 256 threads in a dual socket system, and over the next few years a setup like that will just get cheaper and more common.

src/platform/datapath_winuser.c
```diff
@@ -577,15 +577,14 @@ QuicDataPathInitialize(
...
     DatapathLength =
         sizeof(QUIC_DATAPATH) +
-        Sysinfo.dwNumberOfProcessors * sizeof(QUIC_DATAPATH_PROC_CONTEXT);
+        MaxProcCount * sizeof(QUIC_DATAPATH_PROC_CONTEXT);
 
     //
     // Thread affinity uses a bit mask. We can't support more cores than bits.
     //
-    QUIC_FRE_ASSERT(Sysinfo.dwNumberOfProcessors <= sizeof(DWORD_PTR) * 8);
+    QUIC_FRE_ASSERT(MaxProcCount <= sizeof(DWORD_PTR) * 8);
```
ThadHouse, line 587: Wouldn't this assertion still fail on more then 64 cores?

nibanks: Fixed.

src/core/quicdef.h
nibanks: We currently use a `uint8_t` for partition count, so `256` doesn't fix. We can change it to a `uint16_t` in a future change.

PR #585, opened 2020-07-09
Fix up issues in the PS Performance script

ThadHouse:
Timeout added to local run (configurable)
Local matcher now prints error on failure
Release mode as default
Automatically configure tls to match system if not set
Validate sets for tls and arch to enable tab completion

### DIFF
--- a/scripts/performance-helper.psm1
+++ b/scripts/performance-helper.psm1
@@ -385,13 +385,11 @@ function Get-MedianTestResults($FullResults) {
 }
 
 function Get-TestResult($Results, $Matcher) {
-    try {
-        # Unused variable on purpose
-        $Results -match $Matcher | Out-Null
+    $Found = $Results -match $Matcher
+    if ($Found) {
         return $Matches[1]
-    } catch {
-        Write-Output "Error Processing Results:`n`n$Results"
-        throw
+    } else {
+        Write-Error "Error Processing Results:`n`n$Results"
     }
 }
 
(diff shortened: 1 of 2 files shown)


scripts/performance-helper.psm1
```diff
@@ -360,7 +361,14 @@ function Invoke-LocalExe {
     $FullCommand = "$Exe $RunArgs"
     Write-Debug "Running Locally: $FullCommand"
 
-    return (Invoke-Expression $FullCommand) -join "`n"
+    $LocalJob = Invoke-Command -ScriptBlock { & $Exe ($RunArgs).Split(" ") } -AsJob
+
+    # Wait 60 seconds for the job to finish
+    Wait-Job -Job $LocalJob -Timeout 60 | Out-Null
```
nibanks, line 367: Timeout should be an (optional) input argument, all the way up to the user when it calls the script.

ThadHouse: The timeout would be per individual test, not the whole script. But that'd be fine.

scripts/performance.ps1
```diff
@@ -102,6 +111,22 @@ if ($IsWindows) {
...
+    } else {
+        $LocalTls = "openssl"
+        $RemoteTls = $LocalTls
+    }
+} elseif (($LocalTls -ne "") -xor ($RemoteTls -ne "")) {
+    Write-Error "Both TLS arguments must be set if a manual setting is done"
+}
+
+Write-Host $LocalTls
+Write-Host $RemoteTls
```
nibanks, line 128: Do we really need these?

ThadHouse: Nope. Removed

PR #587, opened 2020-07-10
User/chgray/clog just linux

chgray:


ThadHouse:
Does this depend on https://github.com/microsoft/CLOG/pull/9 if so we should merge that, and I'll update master.

PR #589, opened 2020-07-10
Refactor build system to use more modern cmake techniques

### LINKED ISSUE #566: Clean up cmake compiler argument setting
Currently, we use a mix of set(CMAKE_C_FLAGS) and target_compile_options to add compiler arguments. This causes ordering issues with flags, and adds a lot of complication to the build scripts. We should go through and switch everything to target_compiler_options. This will clean up the build scripts, make them less complicated. This will also enable propagation through the build targets, making downstream builds much cleaner.

ThadHouse:
The old method of setting CMAKE_C_FLAGS and other options like that is not recommended anymore. Its much more recommended to put any compile options needed on the lowest base library in your build setup (platform for us), and then have those settings propagate through the rest of the targets. By doing this, its much easier to enforce that builds are happening correctly, along with ensure a dependency is never accidentally missed. 

One side effect of this change for use is that warnings and errors are now enabled through the entire project, even through tools. Some smaller errors have been fixed, where others have had their individual warnings disabled in their individual projects.

We're still having to do a small amount of manual flag setting. This is because adding options to the linker target wasn't added to cmake until 3.13. We could make that the minimum, but that would require linux users running anything before ubuntu 19.04 to compile their own versions of CMake. This might not be a big deal, and AZP has CMake 3.17, so this might be ok, but I wanted to get it working with our existing minimum requirements first.

Closes #566


src/platform/unittest/DataPathTest.cpp
```diff
@@ -49,6 +49,7 @@ struct QuicAddr
     }
 
     void Resolve(QUIC_ADDRESS_FAMILY af, const char* hostname) {
+        UNREFERENCED_PARAMETER(af);
```
nibanks, line 52: For all C++ code, if the variable is unused, wrap the variable in a comment. `QUIC_ADDRESS_FAMILY /* af */, ...`

ThadHouse:
It looks like its relatively easy to install newer version of cmake. I'll work this weekend on updating the docker image used for raspbian to have cmake 3.17. Theres some other nice usability improvements int newer versions of cmake as well.

nibanks:
So, I like the changes in the top-level cmakelist.txt file, but for the most part don't care for the changes in all the other files. I don't like needing a new base cmakelists.txt file that then has to be included by everything else. I feel like it makes things even more complicated in each individual cmakelist.txt file.

ThadHouse:
Without adding that base, this pr isn't possible. We don't have a common base library to propagate all the arguments through the right way. And modern cmake depends on propagating args through targets rather then globally set variables.

I can move that base definition into the root cmake file though, but getting rid of it completely wouldn't work. If you don't want it, then we'll just close this and not do it.

Basically, it's the replacement for having to do the set(CMAKE_C_ARGS) in every file, which we already have to do. So it's not adding anything that wasn't already there, it's just doing it in a different spot.

nibanks:
I like the refactoring of flags in the main file, but I'm not sold on the other stuff. Can you split the main file changes into a new PR? We should be able to immediately take that and then discuss the rest of this some more. Like what real benefit is it giving us? What do we lose without it?

ThadHouse:
You can't just split out the main file. The way the flags are changed in the main file are incompatible with the existing method of setting arguments, since they're using a list, and not just a flat string. The only change you could pull out would be how we if check windows, but thats pretty minor.

CMakeLists.txt
```diff
@@ -154,16 +162,32 @@ set(CLOG_INCLUDE_DIRECTORY ${CMAKE_SOURCE_DIR}/build/clog)
...
 
-set(CLANG_GCC_WARNING_FLAGS -Werror -Wall -Wextra -Wformat=2 -Wno-type-limits -Wno-unknown-pragmas -Wno-unused-value CACHE INTERNAL "")
-set(MSVC_WARNING_FLAGS /WX /W4 /sdl CACHE INTERNAL "")
+if (WIN32)
+    set(QUIC_WARNING_FLAGS /WX /W4 /sdl CACHE INTERNAL "")
+    set(QUIC_COMMON_FLAGS "")
+    if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+        list(APPEND QUIC_COMMON_FLAGS /MP)
+    endif()
+
```
nibanks, line 177: nit: remove newline

CMakeLists.txt
```diff
@@ -154,16 +162,32 @@ set(CLOG_INCLUDE_DIRECTORY ${CMAKE_SOURCE_DIR}/build/clog)
 set(CMAKE_CLOG_GENERATE_FILE ${CMAKE_SOURCE_DIR}/build/clog/CLog.cmake)
 set(CMAKE_CLOG_CONFIG_FILE ${CMAKE_CURRENT_SOURCE_DIR}/src/manifest/msquic.clog_config)
 
+include(${CMAKE_CLOG_GENERATE_FILE})
+
+set(QUIC_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/src/inc)
+
```
nibanks, line 168: nit: remove empty newlines

CMakeLists.txt
```diff
@@ -39,7 +39,15 @@ message(STATUS "System name: ${CMAKE_SYSTEM_NAME}")
...
 message(STATUS "Platform version: ${CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION}")
 
-if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+if (WIN32)
+    set(QUIC_PLATFORM "windows")
+elseif (APPLE)
+    set(QUIC_PLATFORM "darwin")
+elseif (UNIX)
+    set(QUIC_PLATFORM "linux")
+endif()
```
nibanks, line 48: Let's add a `message(STATUS` for the value of `QUIC_PLATFORM`

CMakeLists.txt
```diff
@@ -331,6 +349,9 @@ if(QUIC_TLS STREQUAL "mitls")
     add_subdirectory(submodules/everest/msquic/msvc/quiccrypto)
 endif()
 
+include(src/tools/toolbase.cmake)
```
nibanks, line 352: I'm not sure if cmake supports this, but what if, instead of including this globally, we have a cmakelist.txt in src/tools that includes this, and has all the `add_subdirectory` calls. Then, we only add the src/tools subdirectory if `QUIC_BUILD`TOOLS` is set?

src/core/CMakeLists.txt
```diff
@@ -49,18 +46,16 @@ endif()
 # Allow CLOG to preprocess all the source files.
 CLOG_GENERATE_TARGET(CORE_CLOG_LIB ${SOURCES})
 
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${QUIC_C_FLAGS}")
+target_link_libraries(CORE_CLOG_LIB PRIVATE inc)
```
nibanks, line 49: What do you think about renaming all these *_CLOG_LIB libraries to *.clog instead?

src/tools/ip/client/CMakeLists.txt
```diff
@@ -1,17 +1,11 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${QUIC_CXX_FLAGS}")
-
-include_directories(${PROJECT_SOURCE_DIR}/src/tools/ip)
-
+# This does not use add_quic_tool because it doesn't use platform
```
nibanks, line 4: Does it break things to include platform though? Shouldn't the linker just ignore it?

src/tools/toolbase.cmake
```diff
@@ -0,0 +1,15 @@
...
+    list(REMOVE_AT ARGV 0)
+
+    add_executable(${targetname} ${ARGV})
+    target_link_libraries(${targetname} msquic platform inc)
+    set_property(TARGET ${targetname} PROPERTY FOLDER "tools")
+endfunction()
+
+function(quic_tool_warnings)
+    target_link_libraries(${ARGV0} warnings)
+endfunction()
```
nibanks, line 15: Personally, I'm not sold on the necessity to even have this or link in warnings for any tools; just for simplicity sake.

ThadHouse: Some tools already have warnings enabled, about 1/3 of them. Wanted to keep the status quo. I would prefer any new tools have warnings enabled, because it really is better coding practice.

CMakeLists.txt
ThadHouse: Yeah it supports that.

src/core/CMakeLists.txt
ThadHouse: Easy enough. I can do that.

src/tools/ip/client/CMakeLists.txt
ThadHouse: It doesn't technically break things. So we could add that.

src/core/unittest/CMakeLists.txt
```diff
@@ -22,10 +14,21 @@ set(SOURCES
 # Allow CLOG to preprocess all the source files.
 CLOG_GENERATE_TARGET(MSQUIC_CORETEST ${SOURCES})
```
nibanks, line 15: You missed a clog name change.

src/tools/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+    add_subdirectory(post)
+    add_subdirectory(reach)
+    add_subdirectory(sample)
+    add_subdirectory(siduck)
+    add_subdirectory(spin)
+    add_subdirectory(void)
+    if(WIN32)
+        add_subdirectory(etwlib)
+        add_subdirectory(etw)
+    endif()
```
nibanks, line 32: Remove indent?

PR #591, opened 2020-07-13
Add support for SourceLink for msquic

ThadHouse:
SourceLink is a technology that adds a mapping of source files to their permalinks on GitHub. This allows the VS Debugger to download sources directly from GitHub, rather then needing to find them locally. It does this automatically from information added into the PDB

Cmake files taken from https://github.com/microsoft/Azure-Kinect-Sensor-SDK/commit/aebb27acd8aa4a3aa6005689da3c8a7fe1413878

Currently, this is only supported by full Visual Studio, but there are issues in the vscode cpptools repo to add support to VS Code, along with I would assume support in windbg is in the pipeline.


nibanks:
This seems to require cmake? Is there a way to do this for our kernel mode VS projects?

ThadHouse:
We'd need to generate the SourceLink file using some other tool (Could probably do it in PS). And then hopefully /SOURCELINK works in kernel mode.

ThadHouse:
Sadly, this might not be usable with PGO. Might need to report a bug to the linker team.

ThadHouse:
We'll see if my trial fix works, otherwise we'll have to wait. I created the following issue for the compiler.

https://developercommunity.visualstudio.com/content/problem/1111456/sourcelink-incompatible-with-useprofile.html

ThadHouse:
Nope. This will not work with PGO.

ThadHouse:
I've unblocked this by making it only apply to debug builds.

nibanks:
With the latest changes, does kernel mode build locally without first running that PowerShell script?

ThadHouse:
I'll check. Gotta install the WDK.

PR #594, opened 2020-07-13
Switch kernel mode to use ExAllocatePool2

### LINKED ISSUE #574: Use ExAllocatePool2 for Kernel Mode
Move kernel mode allocations to [ExAllocatePool2](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-exallocatepool2)

ThadHouse:
The big change is that all memory is now zero initialized. We need to go through, and determine if this is fine, or if we have any extremely hot paths we want to remove this new behavior

Closes #574


src/inc/quic_platform_winkernel.h
```diff
@@ -257,8 +257,8 @@ extern uint64_t QuicTotalMemory;
 
 #define QUIC_POOL_TAG 'CIUQ'
 
-#define QUIC_ALLOC_PAGED(Size) ExAllocatePoolWithTag(PagedPool, Size, QUIC_POOL_TAG)
-#define QUIC_ALLOC_NONPAGED(Size) ExAllocatePoolWithTag(NonPagedPoolNx, Size, QUIC_POOL_TAG)
+#define QUIC_ALLOC_PAGED(Size) ExAllocatePool2(POOL_FLAG_PAGED, Size, QUIC_POOL_TAG)
+#define QUIC_ALLOC_NONPAGED(Size) ExAllocatePool2(POOL_FLAG_NON_PAGED, Size, QUIC_POOL_TAG)
```
anrossi, line 261: There may be a perf regression here because the QUIC code zero-initializes allocations as well. For example, see QuicConnAlloc

ThadHouse: QuicConnAlloc doesn't actually hit these paths. It hits QuicPoolAlloc, which calls ExAllocateFromLookasideListEx. I'll look for other cases that actually do though.

nibanks: By default, I want to actually use the special flag that makes this **not** zero initialize.

ThadHouse: Ok. So maybe we want to do the following. It's a bit invasive, but not bad.

1. Add QUIC_ALLOC_*_UNINITIALIZED, which has the uninitialized flag set in kernel mode. In user mode, this just calls a new QuicAllocUninitialized()
2. Change QuicAlloc to zero memory if the allocation was successful.
3. Remove all existing manual zeroing calls, switch perf sensitive loads to QUIC_ALLOC_*_UNINITIALIZED.

ThadHouse: I think its a much smarter idea to use this as the time to go through an make sure anything we want to make uninitialized explicitly be uninitialized, and make the default be zero initialized. For safety, I think thats worth it. And really isn't that much work. Theres probably only 50-75 callers of QUIC_ALLOC_*. It wouldn't take much time at all.

nibanks: Let's do this in two stages. First stage, replace the alloc function and keep the same behavior. Second, possibly look into having separate allocators for initialized and uninitialized.

ThadHouse: I'll switch these ones to uninitialized, and I'll switch the one in QuicSendBufferPoolAlloc. The test ones I'll leave zeroing, since there's little perf concerns there.

Maximus-: To clarify, it's *explicit* when you want an uninitialized allocation with this change, right? And the default stays good and is zero-ing?

nibanks: We are explicitly defaulting to uninitialized allocations in MsQuic, since the code selectively zeros out the data after allocation, when necessary.

ThadHouse: I've remove that for this PR, and those changes are going to happen in another PR. Right now, everything is going to be uninitialized, like previously. The next PR with make the default be zeroing, and we'll explicitly opt out of the places we need the extra perf and know we're not going to have a problem. That would affect user mode as well.

PR #595, opened 2020-07-14
Windows Kernel Mode Test Automation

nibanks:
Onboard Windows kernel mode testing to Azure Pipelines CI.


scripts/run-gtest.ps1
```diff
@@ -56,6 +59,9 @@ param (
     [Parameter(Mandatory = $true)]
     [string]$Path,
 
+    [Parameter(Mandatory = $false)]
+    [string]$Kernel = "",
```
anrossi, line 63: Why is this not a switch, or boolean?

nibanks: It's a path to the binaries.

PR #601, opened 2020-07-16
Public IP Lookup Protocol/Tool

nibanks:
Implements a new, simple protocol to easily lookup the public IP address of a client. The protocol is as follows:

- Server running at well-known, publicly accessible address/host.
- Client connects to server, and provides for 1 unidirectional stream to be created by the server.
- Server sends the client's IP address and port (as it sees it) back to the client on the unidirectional stream.
- Client successfully connects (securely) and then processes the stream data from the server.
- Client closes connection.

In total, the protocol takes 1 round trip to get the address and an additional round trip to close the connection:

```
Client                                                    Server
======                                                    ======

Connect
                     Initial ------------->
                                                  New Connection
                     <------------- Initial
                     <----------- Handshake
					 
                                          Send Public IP Address
                     <- - - - - - - - 1-RTT

Handshake Received (Initial)
Handshake Received (Handshake)
                     Handshake ----------->
Connection Complete
Process 1-RTT Stream Data

Shutdown Connection
                     1-RTT - - - - - - - ->

                                             Connection Shutdown
                     <- - - - - - - - 1-RTT
Connection Closed
```


src/tools/ip/quicip.cpp
```diff
@@ -0,0 +1,427 @@
...
+    case QUIC_CONNECTION_EVENT_CONNECTED:
+        break;
+    case QUIC_CONNECTION_EVENT_SHUTDOWN_INITIATED_BY_TRANSPORT:
+    case QUIC_CONNECTION_EVENT_SHUTDOWN_INITIATED_BY_PEER:
+        break;
+    case QUIC_CONNECTION_EVENT_SHUTDOWN_COMPLETE:
+        MsQuic->ConnectionClose(Connection);
+        break;
+    case QUIC_CONNECTION_EVENT_PEER_STREAM_STARTED:
+        MsQuic->SetCallbackHandler(Event->PEER_STREAM_STARTED.Stream, ClientStreamCallback, nullptr);
```
ThadHouse, line 273: need (void*) cast on the ClientStreamCallback.

nibanks: Done.

src/tools/ip/quicip.h
```diff
@@ -0,0 +1,174 @@
...
+TODO:
+
+    Don't use the QUIC_ADDR format, as it's not cross-platform. Use the same
+    format as the QUIC spec's preferred address TP.
+
+--*/
+
+#include <msquichelper.h>
+
+const QUIC_API_TABLE* MsQuic;
```
ThadHouse, line 19: Make this extern, and then add the definition to both TU's. Otherwise including this file from multiple TUs will fail to link.

src/tools/ip/quicip.h
```diff
@@ -0,0 +1,174 @@
...
+Abstract:
+
+    A client helper for resolving the public IP address.
+
+TODO:
+
+    Don't use the QUIC_ADDR format, as it's not cross-platform. Use the same
+    format as the QUIC spec's preferred address TP.
+
+--*/
```
ThadHouse, line 15: Header Guards and/or Pragma once

src/tools/ip/quicip.h
```diff
@@ -0,0 +1,174 @@
...
+
+#include <msquichelper.h>
```
ThadHouse, line 17: By doing this, we'll need to add msquichelper.h to the public headers list
https://github.com/microsoft/msquic/blob/master/src/bin/CMakeLists.txt#L41

Also means msquichelper.h needs to be updated to not include quic_platform.h, msquicp.h, and not use any functions needed by those 2 headers either.

ThadHouse:
Linux is going to need the `_strnicmp` usage switched into an ifdef and then to use `strncasecmp`

src/tools/ip/quicip.h
nibanks: Added to context instead.

src/tools/ip/quicip.h
nibanks: Fixed this another way.

src/tools/ip/quicip.h
nibanks: Done.

ThadHouse:
Those fixes help most of the way, but technically quicip.h still requires msquichelper.h, because of `QUIC_CERTIFICATE_FLAG_DISABLE_CERT_VALIDATION`. It should be usable without msquichelper.h included at all.

nibanks:
> Those fixes help most of the way, but technically quicip.h still requires msquichelper.h, because of `QUIC_CERTIFICATE_FLAG_DISABLE_CERT_VALIDATION`. It should be usable without msquichelper.h included at all. I think hiding that functionality behind a macro would be fine in this cause, since its not generally used in public code, but mainly for testing.

That's defined in `msquicp.h`. Eventually, I plan to refactor that stuff so it won't be needed. I'm happy with how it is now.

ThadHouse:
Ok. Still gotta do the linux fix, otherwise it won't build at all on linux.

nibanks:
> Ok. Still gotta do the linux fix, otherwise it won't build at all on linux.

Done. *I think*.

PR #604, opened 2020-07-16
Preallocate Initial Subranges

nibanks:
Allocating range objects in stream initialization was another significant performance hit. This adds support to include the initial allocation of subranges for each range object along with the range object itself.

### DIFF
--- a/src/core/range.c
+++ b/src/core/range.c
@@ -15,8 +15,6 @@ Abstract:
 #include "range.c.clog.h"
 #endif
 
-#define INITIAL_SUBRANGE_COUNT 8
-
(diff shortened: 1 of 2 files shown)


ThadHouse:
Any reason why this is hidden behind a flag, if #605 isn't? Seems unlikely we'd ever want that flag off at compile time.

nibanks:
> Any reason why this is hidden behind a flag, if #605 isn't? Seems unlikely we'd ever want that flag off at compile time.

For the other PR, it's more of a pain to add the flag because it touches more files. Also, for the range one, we already had a flag controlling the binary search feature. Eventually, I'd likely remove all the flags.

PR #607, opened 2020-07-17
Add Code coverage to Azure Pipelines BVTs

anrossi:


scripts/prepare-machine.ps1
```diff
@@ -40,6 +40,26 @@ if ($IsWindows) {
...
+    if ($Configuration -eq "Test") {
+        # Install OpenCppCoverage on test machines
+
+        # Install procdump on Windows if not already present.
+        if (!(Test-Path "C:\Program Files\OpenCppCoverage\OpenCppCoverage.exe")) {
+
+            Write-Host "[$(Get-Date)] Installing OpenCppCoverage..."
+
+            # Download the zip file.
+            $ExeFile = Join-Path $Env:TEMP "OpenCppCoverageSetup-x64-0.9.9.0.exe"
```
ThadHouse, line 52: Will a 64 bit copy of OpenCppCoverage work with 32 bit binaries? 

We should also add a check here either to not allow this if running a 32 bit os, or be smart enough to download the 32 bit copy on a 32 bit os.

ThadHouse: Before calling Invoke-WebRequest, add the following

`$ProgressPreference = 'SilentlyContinue'`

That will massively speed up the download. I need to do the same thing to the file copies in the perf script.

scripts/prepare-machine.ps1
```diff
@@ -40,6 +40,26 @@ if ($IsWindows) {
         reg.exe add $TlsClientKeyPath /v DisabledByDefault /t REG_DWORD /d 0 /f | Out-Null
         reg.exe add $TlsClientKeyPath /v Enabled /t REG_DWORD /d 1 /f | Out-Null
     }
+    if ($Configuration -eq "Test") {
+        # Install OpenCppCoverage on test machines
+
+        # Install procdump on Windows if not already present.
```
nibanks, line 46:
```suggestion
        # Install OpenCppCoverage on test machines.
```

scripts/run-gtest.ps1
```diff
@@ -105,7 +105,10 @@ param (
     [switch]$NoProgress = $false,
 
     [Parameter(Mandatory = $false)]
-    [switch]$EnableAppVerifier = $false
+    [switch]$EnableAppVerifier = $false,
+
+    [Parameter(Mandatory = $false)]
+    [switch]$CodeCoverage = $false
```
nibanks, line 111: Add .Parameter description at the top of the file please

scripts/run-gtest.ps1
```diff
@@ -138,6 +157,13 @@ $TestExeName = Split-Path $Path -Leaf
 $LogDir = Join-Path $RootDir "artifacts" "logs" $TestExeName (Get-Date -UFormat "%m.%d.%Y.%T").Replace(':','.')
 New-Item -Path $LogDir -ItemType Directory -Force | Out-Null
 
+# Folder for coverage files
+$CoverageDir = $null
+if ($CodeCoverage) {
+    $CoverageDir = Join-Path $RootDir "artifacts" "coverage"
+    New-Item -Path $CoverageDir -ItemType Directory -Force | Out-Null
+}
```
nibanks, line 165: We need to support running in isolated mode, where each test is a different execution of the test binary. Then merge all executions back up. It might be as simple as using the log directory as the temporary out and then merge it into this path after each test.

.azure/azure-pipelines.ci.yml
```diff
@@ -150,6 +150,13 @@ stages:
       platform: windows
       tls: schannel
       logProfile: Full.Light
+  - template: ./templates/run-bvt.yml
+    parameters:
+      pool: MsQuic-Win-Latest
+      platform: windows
+      tls: schannel
+      logProfile: Full.Light
+      codeCoverage: true
```
nibanks, line 159: This is just a temporary attempt to get AZP to show something. As we discussed before, it's probably better not to need a separate run for this.

nibanks:
We should also add coverage to spinquic.

scripts/prepare-machine.ps1
anrossi: To the first, yes it will work with 32-bit binaries.  However, that is a good point to ensure that this is a 64-bit OS, and possibly grab the 32-bit flavor in case of a 32-bit OS and test binary.

.azure/azure-pipelines.ci.yml
```diff
@@ -224,6 +218,25 @@ stages:
       platform: linux
       tls: openssl
 
+- stage: codecoverage
+  displayName: CodeCoverage
+  dependsOn:
+  - build_windows
+  - build_linux
```
nibanks, line 225: We don't actually depend on Linux right?

.azure/scripts/install-procdump.ps1
```diff
@@ -0,0 +1,44 @@
+<#
```
anrossi, line 1: whoops. delete this.

.azure/azure-pipelines.ci.yml
anrossi: not at this time, no. if we add support for linux, we could.

.azure/azure-pipelines.ci.yml
anrossi: we're going to keep them separate for now since it appears that the code coverage tools eat crashes.

.azure/azure-pipelines.ci.yml
```diff
@@ -187,6 +187,34 @@ stages:
...
+      platform: windows
+      tls: schannel
+      codeCoverage: true
+  - template: ./templates/run-spinquic.yml
+    parameters:
+      pool: MsQuic-Win-Latest
+      platform: windows
+      tls: schannel
+      codeCoverage: true
+  - template: ./templates/merge-publish-coverage.yml
```
nibanks, line 211: This needs to depend on the previous two so it only runs after they're done.

.azure/templates/merge-publish-coverage.yml
```diff
@@ -0,0 +1,36 @@
...
+  - task: Powershell@2
+    displayName: Merge coverage
+    inputs:
+      pwsh: true
+      filePath: scripts/merge-coverage.ps1
+
+  - task: PublishCodeCoverageResults@1
+    inputs:
+      codeCoverageTool: 'cobertura'
+      summaryFileLocation: '**/artifacts/coverage/msquiccoverage.xml'
```
nibanks, line 36: Needs newline at end of file

nibanks:
@anrossi a few things failed with the latest AZP run:

- The CC BVT didn't generate any test xml output: https://dev.azure.com/ms/msquic/_build/results?buildId=95369&view=logs&j=9500acbd-a9c1-59f1-8206-3772c422dbe8&t=8b269cb0-7980-58a5-cf3d-b77d857dec79
- The CC SpinQuic run failed right off the bat it seems and didn't generate any CC output: https://dev.azure.com/ms/msquic/_build/results?buildId=95369&view=logs&j=229b2163-e8cb-50c0-9a0f-5b74e8193400&t=51e5d351-59f9-57f6-550a-d4062feb52b4
- The merge CC failed to download: https://dev.azure.com/ms/msquic/_build/results?buildId=95369&view=logs&j=07fe8205-c3a8-5a49-d73c-7d5d966d2119&t=2bffa87f-a3a3-51b9-d41d-02e5a69dabed

nibanks:
I've made some progress. It's all building and passing now, but the merge isn't resulting any anything, I think because the binary paths are different.

scripts/run-executable.ps1
```diff
@@ -86,6 +92,19 @@ if (!(Test-Path $Path)) {
     Write-Error "$($Path) does not exist!"
 }
 
+# Validate the code coverage switch
+if ($CodeCoverage) {
+    if (!$IsWindows) {
+        Write-Error "-CodeCoverage switch only supported on Windows";
```
Maximus-, line 98: FWIW gcc/clang support their own coverage modes, `--coverage`, which can output coverage reports as well on Linux.

nibanks: We want to make sure we use a format that Azure Pipelines supports, which is what currently drives this PR.

anrossi: At some point in the future, if gcc/clang can output coverage in cobertura format, it would be nice to also collect linux coverage. However, at this time, AZP only supports reporting 1 coverage file per pipeline, so we'd have to merge the Windows and linux coverage, which is undesirable.

Maximus-: That's rather unfortunate. It would be nice to see how much coverage spinquic is actually getting, or how much the unit tests on their own are getting.

PR #612, opened 2020-07-17
Change clog to use a published nuget package

ThadHouse:
Much cleaner and easier to update. In the future would likely grab the headers and configuration needed as well remotely, but for now this is good


ThadHouse:
This is good to go now. Docker build is also running as well, so that won't break.

Dockerfile
```diff
@@ -11,6 +11,7 @@ FROM    source as build
 WORKDIR /src/Debug
 RUN     chmod +x /src/scripts/install-powershell-docker.sh
 RUN     /src/scripts/install-powershell-docker.sh
+ENV	    PATH="/root/.dotnet/tools:${PATH}"
```
nibanks, line 14: Can you line this up with the rest please?

ThadHouse: 😢 VS Code had made the first character a tab rather then a space... lol

src/core/CMakeLists.txt
```diff
@@ -1,8 +1,15 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
+message(STATUS ${PROJECT_SOURCE_DIR})
+file(GLOB_RECURSE TESTING_CMAKE_DIR LIST_DIRECTORIES true ${PROJECT_SOURCE_DIR}/*)
+
+message(STATUS "FILES: ${TESTING_CMAKE_DIR}")
+
+message(STATUS "${CMAKE_CURRENT_SOURCE_DIR}")
```
nibanks, line 9: Necessary?

ThadHouse: Nope

.azure/templates/build-config-user.yml
```diff
@@ -22,6 +22,13 @@ jobs:
     ${{ if eq(parameters.platform, 'windows') }}:
       path: msquic
 
+  - task: UseDotNet@2
```
nibanks, line 25: This adds nearly 2 minutes to our build time. Is it really necessary?

ThadHouse: I can disable it on windows. It is required for linux, where it only takes 6 seconds...

nibanks: Please do so.

ThadHouse: Done

.azure/templates/build-config-user.yml
```diff
@@ -22,6 +22,15 @@ jobs:
...
+
```
nibanks, line 26: nit: extra line?

scripts/build.ps1
```diff
@@ -292,6 +292,8 @@ function CMake-Build {
 #                     Main Execution                         #
 ##############################################################
 
+clog --installDirectory (Join-Path $RootDir build clog)
```
nibanks, line 295: We want to support cmake-only (no pwsh) builders. This won't help them. Should we make this part of the prepare-machine steps (and update the cmake build instructions)?

nibanks: Or make cmake do this?

ThadHouse: Done. Moved into cmake.

PR #614, opened 2020-07-18
minor bug in timer wheel implementation

tadiparthi:
simple bug. the linked list had to move forward than backwards

### DIFF
--- a/src/core/timer_wheel.c
+++ b/src/core/timer_wheel.c
@@ -109,7 +109,7 @@ QuicTimerWheelUninitialize(
                 StillInTimerWheel,
                 Connection,
                 "Still in timer wheel! Connection was likely leaked!");
-            Entry = Entry->Blink;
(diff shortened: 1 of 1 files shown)


nibanks:
@tadiparthi thanks a lot for finding this issue! Unfortunately, we can't take external contributions to the core protocol logic at this time, even for trivial changes like this (please see our[ contribution guidelines](https://github.com/microsoft/msquic/blob/master/.github/CONTRIBUTING.md), also quoted below). We're still working through our internal processes and regression testing at this time. Please open an issue for this instead and we'll have to make the change (most likely the exact same thing) ourselves.

> **Important** - We are still bringing up important regression tests for the core code. Until they are onboarded, any external contributions to the core or kernel mode files in the platform will not be accepted. This is only a temporary restriction and we are working to complete it by the end of 2020.

PR #615, opened 2020-07-18
Driver Verifier for Kernel Mode

nibanks:


nibanks:
Even with the latest attempt to fix it, verifier claims we're leaking the following allocation:
```
fffff8043de306a2 nt!VerifierExAllocatePoolEx+0x92
fffff8043b29a6f1 msquic!ExAllocateFromLookasideListEx+0x81
fffff8043b2a993f msquic!QuicConnApplySettings+0x20f
fffff8043b2a8a2d msquic!QuicConnAlloc+0x45d
fffff8043b2a8f9a msquic!QuicConnInitialize+0x2a
fffff8043b2a2285 msquic!QuicBindingCreateConnection+0x45
fffff8043b2a298e msquic!QuicBindingDeliverDatagrams+0x40e
fffff8043b2a2edf msquic!QuicBindingReceive+0x4ff
fffff8043b2f1dfa msquic!QuicDataPathSocketReceive+0xeca
```
It looks like it might happen every time. I'm not sure how/why.

ThadHouse:
https://github.com/microsoft/msquic/blob/b34365232ccae65dbf2a535db71c56748442cb76/src/core/connection.c#L2179

Should this not be inverted? On line 460 if the allocation is successful, we set ResumptionEnabled true. Or maybe we should be freeing the HandshakeTP outside of that if check?

nibanks:
> https://github.com/microsoft/msquic/blob/b34365232ccae65dbf2a535db71c56748442cb76/src/core/connection.c#L2179
> 
> Should this not be inverted? On line 460 if the allocation is successful, we set ResumptionEnabled true. Or maybe we should be freeing the HandshakeTP outside of that if check?

No, I believe that's correct. Either way, the HandshakeTP should still be freed when the connection is cleaned up.

ThadHouse:
I think I might have found something.

If the app has set `ServerResumptionLevel` to not be the default, when QuicConnAlloc calls QuicConnApplySettings, the HandshakeTP will be allocated. Then later on down the road, QuicSessionRegisterConnection will call QuicConnApplySettings, which if if has a resumption level set, will actually allocate over the existing allocation. We need to only do the allocation if an allocation does not already exist I think.

nibanks:
Weird. It actually all passed this time. Maybe my private run didn't apply binaries correctly. Looks like this is ready to go!

PR #621, opened 2020-07-21
Refactor IP Tool/Helper

nibanks:
Refactors the IP tool/helper API a bit. Allows for stubbing out logging. Allows for using private registration and function table.

### DIFF
--- a/src/tools/ip/quicip.h
+++ b/src/tools/ip/quicip.h
@@ -70,14 +70,12 @@ ClientStreamCallback(
     }
     case QUIC_STREAM_EVENT_PEER_SEND_ABORTED:
         if (!Context->Success && Context->Status == QUIC_IP_DEFAULT_STATUS) {
-            QUIC_PRINTF("Stream Peer Send Aborted, 0x%llx!\n", Event->PEER_SEND_ABORTED.ErrorCode);
-            // TODO - Cache error to report to app?
+            QUIC_PRINTF("Stream Peer Send Aborted!\n");
         }
         break;
     case QUIC_STREAM_EVENT_PEER_RECEIVE_ABORTED:
         if (!Context->Success && Context->Status == QUIC_IP_DEFAULT_STATUS) {
-            QUIC_PRINTF("Stream Peer Receive Aborted, 0x%llx!\n", Event->PEER_RECEIVE_ABORTED.ErrorCode);
-            // TODO - Cache error to report to app?
+            QUIC_PRINTF("Stream Peer Receive Aborted!\n");
         }
         break;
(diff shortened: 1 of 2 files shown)


src/tools/ip/quicip.h
```diff
@@ -17,15 +17,28 @@
 #pragma once
 
 #include <msquic.h>
+#include <msquicp.h>
```
ThadHouse, line 20: What is this using out of msquicp? (Other then the self signed security thing).

nibanks: That's it.

src/tools/ip/quicip.h
```diff
@@ -72,8 +98,20 @@ ClientConnectionCallback(
...
     switch (Event->Type) {
+    case QUIC_CONNECTION_EVENT_SHUTDOWN_INITIATED_BY_TRANSPORT:
+        if (!Context->Success && Context->Status == QUIC_IP_DEFAULT_STATUS) {
+            QUIC_PRINTF("Connection Shutdown, 0x%x!\n", Event->SHUTDOWN_INITIATED_BY_TRANSPORT.Status);
+            Context->Status = Event->SHUTDOWN_INITIATED_BY_TRANSPORT.Status;
+        }
+        break;
+    case QUIC_CONNECTION_EVENT_SHUTDOWN_INITIATED_BY_PEER:
+        if (!Context->Success && Context->Status == QUIC_IP_DEFAULT_STATUS) {
+            QUIC_PRINTF("Connection Shutdown by Peer, 0x%llx!\n", Event->SHUTDOWN_INITIATED_BY_PEER.ErrorCode);
```
nibanks, line 111: Apparently, Linux doesn't like this:
```
/home/vsts/work/1/s/src/tools/ip/quicip.h:24:56: error: format ‘%llx’ expects argument of type ‘long long unsigned int’, but argument 2 has type ‘QUIC_UINT62 {aka long unsigned int}’ [-Werror=format=]
 #define QUIC_PRINTF(fmt, ...) printf(fmt, ##__VA_ARGS__)
                                                        ^
/home/vsts/work/1/s/src/tools/ip/quicip.h:73:64:
             QUIC_PRINTF("Stream Peer Send Aborted, 0x%llx!\n", Event->PEER_SEND_ABORTED.ErrorCode);
```
What's the best way forward?

ThadHouse: #include <inttypes.h>

Then instead of `0x%llx!` use `%"PRIx64"` (The quotes are important here)

ThadHouse: Just noticed because C++ you need spaces between the end quotes and the macro

nibanks: Just decided to remove them all together.

PR #626, opened 2020-07-22
Create new Performance Execution Driver

ThadHouse:
Replacement for our performance tests currently using QuicPing. Very similar to how QuicPing performs for our throughput tests, but a bit more limited on what all it can do. Still has the relevant flags to performance testing.

Right now only has throughput, but the infrastructure can easily be expanded in the future to make it easy to add new tests.

Kernel mode is not functional, however it does build, which ensures that no changes will break the kernel mode build. Kernel mode will be finished in a separate PR.

Once this is done, I'll work on making the performance runner able to handle different test flags.


src/inc/msquic_linux.h
```diff
@@ -20,6 +20,7 @@
 #define _MSQUIC_LINUX_
 
 #include <stdlib.h>
+#include <stdio.h>
```
nibanks, line 23: Absolutely necessary?

nibanks:
What do you think about not having a separate winkernel directory, merging driver and control files and just having: appmain.cpp and drvmain.cpp?

scripts/RemoteTests.json
```diff
@@ -5,9 +5,9 @@
             "Platform": "Windows",
             "Tls": ["stub", "schannel", "mitls"],
             "Arch": ["x64", "x86", "arm", "arm64"],
-            "Exe": "quicping",
+            "Exe": "quicperf",
             "Arguments": {
-                "All": "-listen:* -port:4433 -peer_uni:1 -connections:10",
+                "All": "-TestName:Throughput -ServerMode:1 -listen:* -port:4433 -peer_uni:1 -connections:10",
```
nibanks, line 10: If you're passing the `listen` arg, there shouldn't be any reason you need a separate `ServerMode` arg.

src/inc/msquic_linux.h
ThadHouse: Yes. That's where sprintf is pulled in. We've just gotten lucky in the past that any time we've included this and used the sprintf_s macro, its been included. But I ran into a case here where I wasn't including it, and we should be including what we use.

scripts/RemoteTests.json
ThadHouse: The way the code is written, the separate tests don't share args. So when I need to know to start the server or client, I don't know that listen is even going to be an argument.

ThadHouse:
We could do an appmain and drvmain. I don't see much of an issue for that. Would we want to do that for the test bin too? Separate PR of course.

nibanks:
> We could do an appmain and drvmain. I don't see much of an issue for that. Would we want to do that for the test bin too? Separate PR of course.

I'd probably the test ones separate for now, because control is so huge.

scripts/RemoteTests.json
nibanks: Ok.

src/perf/bin/main.cpp
```diff
@@ -0,0 +1,500 @@
...
+                "[ lib] ERROR, %u, %s.",
+                 GetLastError(),
+                "OpenService failed");
+            char DriverFilePath[MAX_PATH];
+            Error =
+                GetFullPathNameA(
+                    "msquictest.sys",
+                    sizeof(DriverFilePath),
+                    DriverFilePath,
+                    nullptr);
```
nibanks, line 95: Use the updated way the test code does this.

src/perf/bin/main.cpp
```diff
@@ -0,0 +1,500 @@
...
+    ( ControlCode >> 2) & 0x00000FFF )
+
+#define IOCTL_QUIC_SEC_CONFIG \
+    QUIC_CTL_CODE(1, METHOD_BUFFERED, FILE_WRITE_DATA)
+
+#define IOCTL_QUIC_RUN_PERF \
+    QUIC_CTL_CODE(2, METHOD_BUFFERED, FILE_WRITE_DATA)
+
+#define IOCTL_QUIC_READ_DATA \
+    QUIC_CTL_CODE(3, METHOD_BUFFERED, FILE_READ_DATA)
```
nibanks, line 172: These should be in a shared header with the driver.

src/perf/bin/main.cpp
```diff
@@ -0,0 +1,500 @@
...
+        QuicEventSet(StopEvent);
+    }
+    RetVal = QuicMainStop(0);
+    QuicEventUninitialize(StopEvent);
+    return RetVal;
+}
+
+#ifdef _WIN32
+
+class QuicDriverService {
```
nibanks, line 60: We should generalize the following code and share it between the test and perf code.

src/perf/bin/main.cpp
```diff
@@ -0,0 +1,500 @@
...
+
+int
+QUIC_MAIN_EXPORT
+main(
+    _In_ int argc,
+    _In_reads_(argc) _Null_terminated_ char* argv[]
+    ) {
+    QuicPlatformSystemLoad();
+    QuicPlatformInitialize();
+    QUIC_SEC_CONFIG_PARAMS* SelfSignedConfig = QuicPlatGetSelfSignedCert(QUIC_SELF_SIGN_CERT_USER);
```
nibanks, line 458: If doing kernel mode, you need to pass the other flag into here.

src/perf/lib/PerfHelpers.h
```diff
@@ -0,0 +1,629 @@
...
+        QuicZeroMemory(Buffer, sizeof(*Buffer) + Size);
+        Buffer->Length = Size;
+        Buffer->Buffer = (uint8_t*)(Buffer + 1);
+    }
+    operator QUIC_BUFFER* () { return Buffer; }
+    ~QuicBufferScope() { if (Buffer) { delete[](uint8_t*) Buffer; } }
+};
+
+
+struct CountHelper {
```
nibanks, line 530: Looks like this is essentially a QUIC_RUNDOWN_REF, right?

ThadHouse: It was actually copied from QuicPing, as its way of counting connections. But needs a way to take an external event. Also the helper functions.

src/perf/bin/main.cpp
ThadHouse: This is exactly the way test/bin/quic_gtest.cpp does this. And that hasn't changed since I started this PR. What change exactly do you mean?

ThadHouse: Nevermind

src/perf/bin/main.cpp
ThadHouse: Done

src/inc/quic_driver_connection.h
```diff
@@ -0,0 +1,373 @@
...
+                "[ lib] ERROR, %u, %s.",
+                Error,
+                "GetFullPathName failed");
+            return false;
+        }
+    QueryService:
+        ServiceHandle =
+            OpenServiceA(
+                ScmHandle,
+                QUIC_TEST_DRIVER_NAME,
```
nibanks, line 40: This needs to be passed in as an arg, right?

src/inc/quic_driver_connection.h
```diff
@@ -0,0 +1,373 @@
...
+            QuicTraceEvent(
+                LibraryErrorStatus,
+                "[ lib] ERROR, %u, %s.",
+                 GetLastError(),
+                "OpenService failed");
+            char DriverFilePath[MAX_PATH] = {0};
+            GetModuleFileNameA(NULL, DriverFilePath, MAX_PATH);
+            char* PathEnd = strrchr(DriverFilePath, '\\');
+            if (!PathEnd ||
+                sizeof(DriverFilePath) - (PathEnd - DriverFilePath) < sizeof("msquictest.sys")) {
```
nibanks, line 52: Don't hard code msquictest.sys

ThadHouse: Fixed.

src/inc/quic_driver_connection.h
ThadHouse: No, because its a macro required to be defined. Lots of places in here depend on being able to combine that macro with another string.

nibanks: But calling it `QUIC_TEST_DRIVER_NAME` isn't great if you're using this for perf.

nibanks: Yeah, the more I think about it, I really prefer this file not to rely on any outside #defines. Please make these take the parameters as arguments. The couple of places that are combining macros, just remove the macro part and have a more generic error string.

src/inc/quic_driver_connection.h
```diff
@@ -0,0 +1,376 @@
+/*++
+
+    Copyright (c) Microsoft Corporation.
+    Licensed under the MIT License.
```
nibanks, line 4: I'm not a big fan of the file name. Let's call this quic_driver_helpers.h. Also, please add a bit of an abstract at the top.

src/inc/quic_driver_connection.h
```diff
@@ -0,0 +1,376 @@
...
+
+--*/
+
+#pragma once
+
+#define QUIC_TEST_APIS 1
```
nibanks, line 10: Does this file really need to define this?

src/inc/quic_driver_connection.h
```diff
@@ -0,0 +1,376 @@
...
+#include "quic_platform.h"
+#include "quic_trace.h"
+
+#ifdef _WIN32
```
nibanks, line 14: Instead of this, let's make the caller only include the file in _WIN32. Then we can remove this here.

src/inc/quic_driver_run.h
```diff
@@ -0,0 +1,37 @@
+/*++
+
+    Copyright (c) Microsoft Corporation.
+    Licensed under the MIT License.
+
+Abstract:
+
+    QUIC Perf Main common driver code.
```
nibanks, line 8: Let's rename this to quic_driver_main.h

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,207 @@
...
+        static_assert(sizeof(SelfSignedSecurityHash) == sizeof(SelfSignedConfig->Thumbprint));
+        QuicCopyMemory(SelfSignedSecurityHash, SelfSignedConfig->Thumbprint, 20);
+#else
+        SelfSignedParams = SelfSignedConfig;
+#endif
+        IsSelfSignedValid = true;
+    }
+
+
+
```
nibanks, line 186: nit: remove extra newlines

src/inc/quic_driver_connection.h
ThadHouse: That massively breaks the BVT tests. It depends on the empty implementations being there.

src/inc/quic_driver_connection.h
ThadHouse: Ok. I just renamed it to `QUIC_DRIVER_NAME`

src/perf/lib/PerfHelpers.h
```diff
@@ -0,0 +1,629 @@
...
+};
+
+inline
+int
+WriteOutput(
+    _In_z_ const char* format
+    ...
+    )
+{
+#ifndef _KERNEL_MODE
```
nibanks, line 184: Just always write to the buffer independent of platform.

src/inc/msquic.hpp
```diff
@@ -84,18 +184,21 @@ class UniquePtrArray {
...
         }
     }
 
     ~QuicApiTable() {
         if (QUIC_SUCCEEDED(Init)) {
-            MsQuicClose(this);
+            MsQuicClose(ApiTable);
+            ApiTable = nullptr;
+            QUIC_API_TABLE* thisTable = this;
+            QuicZeroMemory(thisTable, sizeof(*thisTable));
```
nibanks, line 201: Just setting ApiTable to nullptr is enough.

src/inc/msquic.hpp
```diff
@@ -121,14 +224,35 @@ class MsQuicRegistration {
...
+                Reg,
+                &Alpn,
+                1,
+                nullptr,
+                &Handle))) {
+            Handle = nullptr;
+        }
+    }
+
+#ifndef QUIC_SKIP_GLOBAL_CONSTRUCTORS
```
nibanks, line 254: Why?

src/inc/quic_driver_helpers.h
```diff
@@ -0,0 +1,379 @@
...
+};
+
+#else
+
+class QuicDriverService {
+public:
+    bool Initialize() { return false; }
+    void Uninitialize() { }
+    bool Start() { return false; }
+};
```
nibanks, line 323: I rather not go with this model. I prefer the caller to only reference these classes if _WIN32

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,204 @@
...
+    QUIC_SEC_CONFIG_PARAMS* SelfSignedConfig =
+        QuicPlatGetSelfSignedCert(
+            TestingKernelMode ?
+                QUIC_SELF_SIGN_CERT_MACHINE :
+                QUIC_SELF_SIGN_CERT_USER);
+    if (SelfSignedConfig) {
+#ifdef _KERNEL_MODE
+        static_assert(sizeof(SelfSignedSecurityHash) == sizeof(SelfSignedConfig->Thumbprint));
+        QuicCopyMemory(SelfSignedSecurityHash, SelfSignedConfig->Thumbprint, 20);
```
nibanks, line 178:
```suggestion
        QuicCopyMemory(SelfSignedSecurityHash, SelfSignedConfig->Thumbprint, sizeof(SelfSignedConfig->Thumbprint));
```

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,204 @@
...
+    ) {
+    QuicPlatformSystemLoad();
+    QuicPlatformInitialize();
+
+    bool TestingKernelMode = false;
+    bool KeyboardWait = false;
+
+    for (int i = 0; i < argc; ++i) {
+        if (strcmp("--kernel", argv[i]) == 0) {
+            TestingKernelMode = true;
```
nibanks, line 164: Just put the code that checks for _WIN32 here and error out if they're trying to run on a platform that doesn't support this.

src/perf/lib/PerfHelpers.h
```diff
@@ -0,0 +1,245 @@
...
+#else
+#include <new.h>
+#endif
+
+#ifdef _KERNEL_MODE
+extern uint8_t SelfSignedSecurityHash[20];
+#else
+extern QUIC_SEC_CONFIG_PARAMS* SelfSignedParams;
+#endif
+extern bool IsSelfSignedValid;
```
nibanks, line 44: Don't have global variables. Store them in the actual struct that uses them.

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,204 @@
...
+            KeyboardWait = true;
+        }
+    }
+
+    QUIC_SEC_CONFIG_PARAMS* SelfSignedConfig =
+        QuicPlatGetSelfSignedCert(
+            TestingKernelMode ?
+                QUIC_SELF_SIGN_CERT_MACHINE :
+                QUIC_SELF_SIGN_CERT_USER);
+    if (SelfSignedConfig) {
```
nibanks, line 175: If this fails, you should error and bail.

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,204 @@
...
+    int RetVal = 0;
+    if (TestingKernelMode) {
+#ifdef _WIN32
+        RetVal = QuicKernelMain(argc, argv, KeyboardWait, SelfSignedParams);
+#else
+        printf("Cannot run kernel mode tests on non windows platforms\n");
+        RetVal = QUIC_RUN_INVALID_MODE;
+#endif
+    } else {
+        RetVal = QuicUserMain(argc, argv, KeyboardWait);
```
nibanks, line 194: Just pass SelfSignedParams along regardless. Like I say later, don't rely on global variables. Globals == Bad

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    if (SelfSignedParams) {
+        QuicPlatFreeSelfSignedCert(SelfSignedParams);
+    }
+    QuicPlatformUninitialize();
```
nibanks, line 200: nit: newline before QuicPlatformUninitialize

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,204 @@
...
+    printf("Ready For Connections!\n\n");
+    fflush(stdout);
+    if (KeyboardWait) {
+        printf("Press enter to exit\n");
+        getchar();
+        QuicEventSet(StopEvent);
+    }
+    RetVal = QuicMainStop(0);
+    QuicEventUninitialize(StopEvent);
+    return RetVal;
```
nibanks, line 59: Either have no spaces between code blocks or spaces between all blocks.

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,204 @@
...
+#define QUIC_TEST_APIS 1
+#include "quic_driver_main.h"
+#include "PerfHelpers.h"
+#include <quic_trace.h>
+
+#ifdef _WIN32
+
+//
+// Name of the driver service for quicperf.sys.
+// Must be defined before quic_driver_helpers.h is included
```
nibanks, line 25: I know it's a bigger change, but I still greatly dislike reliance on global #defines or variables. Pass things as input please. It makes the code much more readable.

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+Abstract:
+
+    QUIC Perf Main execution runner.
+
+--*/
+
+#ifdef QUIC_CLOG
+#include "appmain.cpp.clog.h"
+#endif
```
nibanks, line 14: I believe this will need to come **after** #include <quic_trace.h>

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,204 @@
...
+    _In_reads_(argc) _Null_terminated_ char* argv[],
+    _In_ bool KeyboardWait,
+    _In_ QUIC_SEC_CONFIG_PARAMS* SelfSignedParams
+    ) {
+    size_t TotalLength = 0;
+
+    // Get total length
+    for (int i = 0; i < argc; ++i) {
+        //
+        // Length of string, plus null terminator, plus length
```
nibanks, line 76: Why do you need a length? Generally a double null terminator indicates the end of a list like this:
```
-kernel\0-kbwait\0\0
```

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,204 @@
...
+        printf("Failed to allocate space for output buffer\n");
+        QuicFree(Data);
+        return QUIC_RUN_FAILED_TEST_INITIALIZE;
+    }
+
+    QuicDriverService DriverService;
+    QuicDriverClient DriverClient;
+    DriverService.Initialize();
+    DriverService.Start();
+    DriverClient.Initialize(SelfSignedParams);
```
nibanks, line 120: Check for failures and bail.

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+    DWORD OutBufferWritten = 0;
+    bool RunSuccess =
+        DriverClient.Read(
+            IOCTL_QUIC_READ_DATA,
+            OutBuffer,
+            OutBufferSize,
+            &OutBufferWritten);
+
+    if (RunSuccess) {
```
nibanks, line 138: no spaces between getting status and checking it. Same throughout the code

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,204 @@
...
+        printf("%s", OutBuffer);
+    }
+
+    QuicFree(Data);
+    QuicFree(OutBuffer);
+
+    return RunSuccess ? QUIC_RUN_SUCCESS : QUIC_RUN_STOP_FAILURE;
```
nibanks, line 145: It's not actually stop failure though, right? Also, why even return a status here? Generally, if the caller isn't going to do anything special with a specific error, and just needs to know success/fail, return a bool. If they aren't going to do anything special with a failure, return void. For most of this stuff, you can just print something at the point of failure and bail.

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,204 @@
...
+#endif
+
+int
+QUIC_MAIN_EXPORT
+main(
+    _In_ int argc,
+    _In_reads_(argc) _Null_terminated_ char* argv[]
+    ) {
+    QuicPlatformSystemLoad();
+    QuicPlatformInitialize();
```
nibanks, line 157: I believe one or both of these return status codes. Please validate them.

src/inc/quic_driver_helpers.h
```diff
@@ -0,0 +1,379 @@
...
+        _In_ uint32_t TimeoutMs = 30000
+        ) {
+        return Run(IoControlCode, (void*)&Data, sizeof(Data), TimeoutMs);
+    }
+    bool Read(
+        _In_ uint32_t IoControlCode,
+        _Out_writes_bytes_opt_(OutBufferSize)
+            void* OutBuffer,
+        _In_ uint32_t OutBufferSize,
+        _Out_ LPDWORD OutBufferWritten, // Needs to be LPDWORD to match DeviceIoControl
```
nibanks, line 253: Instead of requiring an LPBDWORD, let's make this an optional output uint32_t* parameter. Then, internally use a private DWORD for getting the length; and only set the optional output parameter if specified by the caller.

src/inc/quic_driver_main.h
```diff
@@ -0,0 +1,37 @@
...
+#include "quic_platform.h"
+
+#define QUIC_RUN_SUCCESS 0
+#define QUIC_RUN_MISSING_TEST_TYPE -1
+#define QUIC_RUN_FAILED_QUIC_OPEN -2
+#define QUIC_RUN_FAILED_TEST_INITIALIZE -3
+#define QUIC_RUN_MISSING_NECESSARY_ARGUMENTS -4
+#define QUIC_RUN_UNKNOWN_TEST_TYPE -5
+#define QUIC_RUN_INVALID_MODE -6
+#define QUIC_RUN_STOP_FAILURE -7
```
nibanks, line 23: Why define your error codes and not just use QUIC_STATUS?

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,204 @@
...
+    _In_ bool KeyboardWait
+    ) {
+    QUIC_EVENT StopEvent;
+    QuicEventInitialize(&StopEvent, true, false);
+
+    int RetVal = QuicMainStart(argc, argv, StopEvent);
+    if (RetVal != 0) {
+        return RetVal;
+    }
+    printf("Ready For Connections!\n\n");
```
nibanks, line 50: I prefer a more general "Started" or "Running" status.

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,204 @@
...
+int
+QuicKernelMain(
+    _In_ int argc,
+    _In_reads_(argc) _Null_terminated_ char* argv[],
+    _In_ bool KeyboardWait,
+    _In_ QUIC_SEC_CONFIG_PARAMS* SelfSignedParams
+    ) {
+    size_t TotalLength = 0;
+
+    // Get total length
```
nibanks, line 73: style: fix comment to have extra // before/after

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,204 @@
...
+    for (int i = 0; i < argc; ++i) {
+        //
+        // Length of string, plus null terminator, plus length
+        //
+        TotalLength += strlen(argv[i]) + sizeof(size_t) + 1;
+    }
+
+    TotalLength += sizeof(TotalLength);
+
+    char* Data = static_cast<char*>(QuicAlloc(TotalLength));
```
nibanks, line 83: Do not call QuicAlloc. QUIC_ALLOC_NONPAGED instead.

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,204 @@
...
+        QuicCopyMemory(DataCurrent, &ArgLen, sizeof(ArgLen));
+        DataCurrent += sizeof(ArgLen);
+        QuicCopyMemory(DataCurrent, argv[i], ArgLen);
+        DataCurrent += ArgLen;
+        DataCurrent[0] = '\0';
+        ++DataCurrent;
+    }
+
+    QUIC_DBG_ASSERT(DataCurrent == (Data + TotalLength));
+    QUIC_DBG_ASSERT(TotalLength <= UINT_MAX);
```
nibanks, line 106: Actually validate this. Generally, asserts are for code that should be completely impossible to hit. With enough parameters, at least this length check could fail. Also, only use DBG asserts if it's computationally expensive to always check. But as I said, do the real validation here.

src/perf/bin/appmain.cpp
nibanks: Or use new/free

src/perf/bin/drvmain.cpp
```diff
@@ -0,0 +1,604 @@
...
+Abstract:
+
+    QUIC Kernel Mode Performance Driver
+
+--*/
+
+#include <quic_platform.h>
+#include "msquic.h"
+#include <quic_driver_main.h>
+#include "perfioctls.h"
```
nibanks, line 15: let's put a _ between the words of this file name: "perf_ioctls.h"

src/perf/bin/drvmain.cpp
```diff
@@ -0,0 +1,604 @@
...
+#define QUIC_PERF_TAG 'frPQ' // QPrf
+
+DECLARE_CONST_UNICODE_STRING(QuicPerfCtlDeviceName, L"\\Device\\quicperformance");
+DECLARE_CONST_UNICODE_STRING(QuicPerfCtlDeviceSymLink, L"\\DosDevices\\quicperformance");
+
+typedef struct QUIC_DEVICE_EXTENSION {
+    EX_PUSH_LOCK Lock;
+
+    _Guarded_by_(Lock)
+        LIST_ENTRY ClientList;
```
nibanks, line 31: style: don't tab over LIST_ENTRY ClientList. Match the rest of the variables.

src/perf/bin/drvmain.cpp
```diff
@@ -0,0 +1,604 @@
...
+    ULONG ClientListSize;
+
+} QUIC_DEVICE_EXTENSION;
+
+WDF_DECLARE_CONTEXT_TYPE_WITH_NAME(QUIC_DEVICE_EXTENSION, QuicPerfCtlGetDeviceContext);
+
+typedef struct QUIC_TEST_CLIENT
+{
```
nibanks, line 39:
```suggestion
typedef struct QUIC_TEST_CLIENT {
```

src/perf/bin/drvmain.cpp
```diff
@@ -0,0 +1,604 @@
...
+NTSTATUS
+QuicPerfCtlInitialize(
+    _In_ WDFDRIVER Driver
+    );
+
+_IRQL_requires_max_(PASSIVE_LEVEL)
+VOID
+QuicPerfCtlUninitialize(
+
+    );
```
nibanks, line 71:
```suggestion
_IRQL_requires_max_(PASSIVE_LEVEL)
void
QuicPerfCtlUninitialize(
    void
    );
```

src/perf/bin/drvmain.cpp
```diff
@@ -0,0 +1,604 @@
...
+
+    RegistryPath - represents the driver specific path in the Registry.
+    The function driver can use the path to store driver related data between
+    reboots. The path does not store hardware instance specific data.
+
+Return Value:
+
+    A success status as determined by NT_SUCCESS macro, if successful.
+
+--*/
```
nibanks, line 135: I know other kernel specific code has this, but let's not do it here. It's always been a style diversion from the rest of the code anyways.

src/perf/bin/drvmain.cpp
```diff
@@ -0,0 +1,604 @@
...
+extern "C"
+INITCODE
+_Function_class_(DRIVER_INITIALIZE)
+_IRQL_requires_same_
+_IRQL_requires_(PASSIVE_LEVEL)
+NTSTATUS
+DriverEntry(
+    _In_ PDRIVER_OBJECT DriverObject,
+    _In_ PUNICODE_STRING RegistryPath
+)
```
nibanks, line 110:
```suggestion
    )
```

src/perf/bin/drvmain.cpp
```diff
@@ -0,0 +1,604 @@
...
+    QuicPlatformUninitialize();
+    QuicPlatformSystemUnload();
+}
+
+_No_competing_thread_
+INITCODE
+NTSTATUS
+QuicPerfCtlInitialize(
+    _In_ WDFDRIVER Driver
+)
```
nibanks, line 246:
```suggestion
    )
```

src/perf/bin/drvmain.cpp
```diff
@@ -0,0 +1,604 @@
...
+        WdfDeviceInitFree(DeviceInit);
+    }
+
+    return Status;
+}
+
+_IRQL_requires_max_(PASSIVE_LEVEL)
+VOID
+QuicPerfCtlUninitialize(
+)
```
nibanks, line 369:
```suggestion
    )
```

nibanks: And fix the rest below

src/perf/bin/drvmain.cpp
```diff
@@ -0,0 +1,604 @@
...
+    //
+    // Initialize the device control interface.
+    //
+    Status = QuicPerfCtlInitialize(Driver);
+    if (!NT_SUCCESS(Status)) {
+        QuicTraceEvent(
+            LibraryErrorStatus,
+            "[ lib] ERROR, %u, %s.",
+            Status,
+            "QuicPerfCtlInitialize failed");
```
nibanks, line 188: Unnecessary, as the function will log at the actual point of failure. In general, we only log function failure if it's an external call.

src/perf/bin/drvmain.cpp
```diff
@@ -0,0 +1,604 @@
...
+        if (QuicPerfCtlExtension->ClientListSize >= 1) {
+            QuicTraceEvent(
+                LibraryError,
+                "[ lib] ERROR, %s.",
+                "Already have max clients");
+            Status = STATUS_TOO_MANY_SESSIONS;
+            break;
+        }
+
+        QUIC_TEST_CLIENT* Client = QuicPerfCtlGetFileContext(FileObject);
```
nibanks, line 415: Don't call it a QUIC_TEST_CLIENT. At the very least, call it a QUIC_DRIVER_CLIENT.

src/perf/bin/drvmain.cpp
```diff
@@ -0,0 +1,604 @@
...
+} QUIC_DEVICE_EXTENSION;
+
+WDF_DECLARE_CONTEXT_TYPE_WITH_NAME(QUIC_DEVICE_EXTENSION, QuicPerfCtlGetDeviceContext);
+
+typedef struct QUIC_TEST_CLIENT
+{
+    LIST_ENTRY Link;
+    bool TestFailure;
+
+} QUIC_TEST_CLIENT;
```
nibanks, line 43: You should have the stop event in here, to (in the future) allow for multiple user mode clients and each individually be stopped.

src/perf/bin/drvmain.cpp
```diff
@@ -0,0 +1,604 @@
...
+    _In_ WDFFILEOBJECT /* FileObject */
+)
+{
+    PAGED_CODE();
+}
+
+PAGEDX
+_Use_decl_annotations_
+VOID
+QuicPerfCtlEvtFileCleanup(
```
nibanks, line 463: Once you put the stop event in the client struct, you should set the event in here and the IoCanceled function below.

src/perf/bin/perfioctls.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
```
nibanks, line 1: Make sure all files have copyright, and at least some small abstract.

src/perf/lib/PerfBase.h
```diff
@@ -0,0 +1,56 @@
+/*++
```
nibanks, line 1: Let's be consistent in file naming schemes here. I think "perf_base.h" would be more consistent.

nibanks: Or rename perfioctls.h to PerfIoctls.h

src/perf/lib/PerfBase.h
```diff
@@ -0,0 +1,56 @@
...
+    QUIC_STATUS
+    Init(
+        _In_ int argc,
+        _In_reads_(argc) _Null_terminated_ char* argv[]
+        ) = 0;
+
+    //
+    // Start the runner. The StopEvent can be triggered to stop early
+    // Passed here rather then Wait so we can synchronize off of it.
+    // This event must be kept alive until Wait is called.
```
nibanks, line 39: In general, try to fill comments all the way out to col 80 before starting a new line. Don't just use a new line for each separate thought.

src/perf/lib/PerfHelpers.h
```diff
@@ -0,0 +1,294 @@
...
+    int rval = vprintf(format, args);
+    va_end(args);
+    return rval;
+#else
+    UNREFERENCED_PARAMETER(format);
+    return 0;
+#endif
+}
+
+struct MsQuicListener {
```
nibanks, line 67: Let's name these so they are explicitly Perf specific. PerfListener perhaps.

src/perf/lib/PerfHelpers.h
```diff
@@ -0,0 +1,294 @@
...
+        if (Handle) {
+            MsQuic->ListenerClose(Handle);
+        }
+    }
+
+    QUIC_STATUS
+    Start(
+        _In_ QUIC_ADDR* Address,
+        _In_ QUIC_LISTENER_CALLBACK_HANDLER ShadowHandler,
+        _In_ void* ShadowContext) {
```
nibanks, line 100: Instead of Shadow in the name, I prefer either you prefix with _ or you use the same name, and when you set it, do something like `this->Handler = Handler`.

nibanks: Also, this might make more sense to pass these in the constructor anyways.

src/perf/lib/ThroughputClient.cpp
```diff
@@ -0,0 +1,410 @@
...
+    QUIC Perf Throughput Client Implementation.
+
+--*/
+
+#ifdef QUIC_CLOG
+#include "ThroughputClient.cpp.clog.h"
+#endif
+
+#ifndef _KERNEL_MODE
+#define QUIC_TEST_APIS 1
```
nibanks, line 17: Why does client need QUIC_TEST_APIS?

src/perf/lib/ThroughputClient.cpp
```diff
@@ -0,0 +1,410 @@
...
+    const char* localAddress;
+    if (TryGetValue(argc, argv, "bind", &localAddress)) {
+        if (!ConvertArgToAddress(localAddress, 0, &LocalIpAddr)) {
+            WriteOutput("Failed to decode IP address: '%s'!\nMust be *, a IPv4 or a IPv6 address.\n", localAddress);
+            return QUIC_STATUS_INVALID_PARAMETER;
+        }
+    }
+
+    // TODO: Core, since we need to support kernel mode
+    #if _WIN32 && !_KERNEL_MODE
```
nibanks, line 70: As we discussed on our call, use QUIC_COMPARTMENT_ID for all this.

src/perf/lib/ThroughputClient.cpp
```diff
@@ -0,0 +1,410 @@
...
+    uint32_t ioCount = UseSendBuffer ? THROGHTPUT_DEFAULT_SEND_COUNT_BUFFERED : THROUGHPUT_DEFAULT_SEND_COUNT_NONBUFFERED;
+    TryGetValue(argc, argv, "iocount", &ioCount);
+    IoCount = ioCount;
+
+    size_t Len = strlen(Target);
+    TargetData.reset(new char[Len + 1]);
+    QuicCopyMemory(TargetData.get(), Target, Len);
+    TargetData[Len] = '\0';
+
+    RawIoBuffer = new uint8_t[IoSize];
```
nibanks, line 103: Make sure to validate all allocations.

src/perf/lib/ThroughputClient.cpp
```diff
@@ -0,0 +1,410 @@
...
+            QuicBuffer.Length = IoSize;
+        } else {
+            Flags |= QUIC_SEND_FLAG_FIN;
+            QuicBuffer.Length = (uint32_t)BytesLeftToSend;
+        }
+    }
+};
+
+struct ShutdownWrapper {
+    HQUIC ConnHandle{nullptr};
```
nibanks, line 133: Like you do in SendRequest above, for all initializations, put a space between the variable and the {...} stuff.

src/perf/lib/ThroughputClient.cpp
```diff
@@ -0,0 +1,419 @@
...
+    Port = THROUGHPUT_DEFAULT_PORT;
+    TryGetValue(argc, argv, "port", &Port);
+
+    const char* Target;
+    if (!TryGetValue(argc, argv, "target", &Target)) {
+        WriteOutput("Must specify '-target' argument!\n");
+        return QUIC_STATUS_INVALID_PARAMETER;
+    }
+
+    uint16_t ip;
```
nibanks, line 45: Let's fix up the casing for these variable to match our style. I know you probably copied from quicping and it has the issue already...

src/perf/lib/ThroughputClient.cpp
```diff
@@ -0,0 +1,419 @@
...
+            StrmData,
+            &StrmData->Stream.Handle);
+
+    if (QUIC_FAILED(Status)) {
+        WriteOutput("Failed StreamOpen 0x%x\n", Status);
+        StreamDataAllocator.Free(StrmData);
+        return Status;
+    }
+
+    Shutdown.StreamHandle = StrmData->Stream.Handle;
```
nibanks, line 258: You shouldn't actually need to shutdown the stream explicitly; just the connection. The caveat to this: if the stream hasn't been successfully started, you need to just close it.

src/perf/lib/ThroughputClient.cpp
```diff
@@ -0,0 +1,419 @@
...
+                    MsQuic->StreamSend(
+                        StrmData->Stream.Handle,
+                        &Req->QuicBuffer,
+                        1,
+                        Req->Flags,
+                        Req);
+
+                if (QUIC_SUCCEEDED(Status)) {
+                    Req = nullptr;
+                }
```
nibanks, line 385:
```suggestion
                if (QUIC_SUCCEEDED(
                    MsQuic->StreamSend(
                        StrmData->Stream.Handle,
                        &Req->QuicBuffer,
                        1,
                        Req->Flags,
                        Req)) {
                    Req = nullptr;
                }
```

src/perf/lib/ThroughputClient.cpp
```diff
@@ -0,0 +1,419 @@
...
+        }
+        if (Req) {
+            SendRequestAllocator.Free(Req);
+        }
+        break;
+    }
+    case QUIC_STREAM_EVENT_PEER_SEND_ABORTED:
+    case QUIC_STREAM_EVENT_PEER_RECEIVE_ABORTED:
+
+        MsQuic->StreamShutdown(
```
nibanks, line 396:
```suggestion
    case QUIC_STREAM_EVENT_PEER_RECEIVE_ABORTED:
        MsQuic->StreamShutdown(
```

src/perf/lib/ThroughputClient.cpp
```diff
@@ -0,0 +1,419 @@
...
+
+        WriteOutput("[%p][%llu] Closed [%s] after %u.%u ms. (TX %llu bytes @ %u kbps).\n",
+            StrmData->Connection,
+            GetStreamID(MsQuic, StreamHandle),
+            "Complete",
+            (uint32_t)(ElapsedMicroseconds / 1000),
+            (uint32_t)(ElapsedMicroseconds % 1000),
+            StrmData->BytesCompleted, SendRate);
+
+        StreamDataAllocator.Free(StrmData);
```
nibanks, line 414: I assume it is, but make sure this is closing the stream.

src/perf/lib/ThroughputServer.cpp
```diff
@@ -0,0 +1,193 @@
...
+        // Add a single item so we can wait on the Count Helper
+        //
+        RefCount.AddItem();
+    }
+    return QUIC_STATUS_SUCCESS;
+}
+
+QUIC_STATUS
+ThroughputServer::Wait(
+    _In_ int Timeout
```
nibanks, line 91: Does this wait function really need to return a status?

src/perf/lib/ThroughputServer.cpp
```diff
@@ -0,0 +1,193 @@
...
+                return ((ConnectionData*)Context)->Server->
+                    ConnectionCallback(
+                        Conn,
+                        Event,
+                        (ConnectionData*)Context);
+            };
+        MsQuic->SetCallbackHandler(
+            Event->NEW_CONNECTION.Connection,
+            (void*)Handler,
+            new ConnectionData{ this });
```
nibanks, line 121: Can this not be combined into a single statement, like you've done other places?

src/perf/lib/quicmain.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+const QuicApiTable* MsQuic{nullptr};
+#ifdef _KERNEL_MODE
+uint8_t SelfSignedSecurityHash[20];
+#else
+QUIC_SEC_CONFIG_PARAMS* SelfSignedParams{nullptr};
+#endif
+bool IsSelfSignedValid{ false };
+
+PerfBase* TestToRun{nullptr};
```
nibanks, line 33: Whenever possible do not use globals. Especially in general mode, you cannot have a global destructor executed.

src/perf/lib/quicmain.cpp
```diff
@@ -0,0 +1,105 @@
...
+            }
+        }
+    }
+
+    if (TestToRun != nullptr) {
+        delete TestToRun;
+    }
+    if (MsQuic != nullptr) {
+        delete MsQuic;
+    }
```
nibanks, line 86: delete is supposed to handle nullptr. Don't check for it here. Just call delete.

src/perf/lib/quicmain.cpp
```diff
@@ -0,0 +1,105 @@
...
+        return QUIC_RUN_SUCCESS;
+    }
+
+    QUIC_STATUS Status = TestToRun->Wait(Timeout);
+    delete TestToRun;
+    delete MsQuic;
+    if (QUIC_SUCCEEDED(Status)) {
+        return QUIC_RUN_SUCCESS;
+    }
+    return QUIC_RUN_STOP_FAILURE;
```
nibanks, line 104:
```suggestion
    return QUIC_SUCCEEDED(Status) ? QUIC_RUN_SUCCESS : QUIC_RUN_STOP_FAILURE;
```

src/perf/bin/drvmain.cpp
ThadHouse: This was copied straight from tests control.cpp. Would you like me to fix that there in this PR, or separate one?

src/perf/bin/drvmain.cpp
ThadHouse: Copied straight from the BVT code. We should create an issue to fix the issues seen here over there too.

src/perf/bin/drvmain.cpp
ThadHouse: Yeah I'll add that when I finish up kernel mode.

src/perf/lib/PerfHelpers.h
ThadHouse: It has to be passed here, since because of how construction works I don't have access to these initially.

src/perf/lib/ThroughputClient.cpp
ThadHouse: Already fixed. Also disabling `throwing new` is an internal compiler flag. Because of how construction works, if new is ever used and fails to allocate, but doesn't throw, the constructor will AV. Unless you do what I do during the pool allocator and placement new.

src/perf/lib/ThroughputClient.cpp
ThadHouse: It is. The destructor will close the stream.

src/perf/lib/ThroughputServer.cpp
ThadHouse: Already done.

src/perf/lib/quicmain.cpp
ThadHouse: Destructors won't run, which is why I'm using a pointer and not a unique ptr. I can't realistically pass state back here. In user mode I technically can, but in kernel mode I'd just have the same thing 1 layer up.

Also, MsQuic HAS to be global because of all the headers that expect it to be global.

src/perf/lib/quicmain.cpp
ThadHouse: Already fixed

src/perf/lib/quicmain.cpp
nibanks: I agree some pointers need to be global. I'm just saying whenever possible do not have a global.

src/perf/lib/ThroughputServer.cpp
ThadHouse: I want the API to return a status, because its possible there could be error cases in the not too distant future.

src/perf/lib/ThroughputClient.cpp
ThadHouse: It doesn't, but kernel mode does need quic_platform. So did just that by itself instead.

src/perf/lib/ThroughputClient.cpp
ThadHouse: Done.

src/inc/msquic.hpp
ThadHouse: Because otherwise I have to define an extern Registration somewhere, because these are using globals. And fixing that requires massively changing the BVTs.

src/inc/msquic.hpp
ThadHouse: Its not actually, because there's 2 copies of the api table, one as part of the inheritance. I'd like to clear that one too just because it makes catching use after free bugs easier.

src/inc/quic_driver_helpers.h
ThadHouse: That requires massive changes to the BVTs

src/perf/bin/appmain.cpp
ThadHouse: I still need the ifdef below, otherwise it won't link, but I'll put the error condition here.

src/perf/bin/appmain.cpp
ThadHouse: This ones actually for a different reason. Because its a macro, I can concatinate it with other macros to get the other versions of it I need later on. Without this, the code in driver helpers actually becomes a lot more complicated, as it will then have to manually concatinate strings in memory. I can do that, but that will also affect BVTs

src/inc/msquic.hpp
nibanks: Oh, I didn't see that. Definitely remove the inheritance.

src/inc/msquic.hpp
nibanks: Ok.

src/perf/bin/appmain.cpp
ThadHouse: I'm going to need to allocate an array on the other side, and having an explicit length means I can do it in 1 loop rather then 2. Also I've never seen double null indicating the end up a list.

src/perf/bin/appmain.cpp
ThadHouse: We'll want to do the same thing in the BVTs at some point then.

src/perf/bin/appmain.cpp
ThadHouse: I had fixed most of these.

src/perf/bin/appmain.cpp
ThadHouse: It's int because this is what is returned from main. Its a forwarder basically. I can make it bool and then map it on the other side, but thats really not much cleaner.

src/perf/bin/appmain.cpp
ThadHouse: We don't do that in much of our test code, but I'll do it here.

src/perf/lib/PerfBase.h
```diff
@@ -0,0 +1,56 @@
...
+    Licensed under the MIT License.
+
+Abstract:
+
+    QUIC Perf Base blass declaration. Defines the base class for all perf
+    executions.
+
+--*/
+
+
```
nibanks, line 13: nit: only single newline between blocks

src/inc/quic_driver_main.h
ThadHouse: Because theres not a QUIC_STATUS that maps to most of these. Most would just be QUIC_STATUS_INTERNAL_ERROR. I can switch them over though if thats cleaner.

src/perf/bin/drvmain.cpp
```diff
@@ -0,0 +1,575 @@
...
+Routine Description:
+
+    QuicTestDriverUnload will clean up any resources that were allocated for
+    this driver.
+
+Arguments:
+
+    Driver - Handle to a framework driver object created in DriverEntry
+
+--*/
```
nibanks, line 196: Remove all these comment blocks

src/perf/lib/PerfHelpers.h
```diff
@@ -0,0 +1,283 @@
...
+template <class _Ty>
+using QuicRemoveReferenceT = typename QuicRemoveReference<_Ty>::type;
+
+template <class _Ty>
+constexpr _Ty&& QuicForward(
+    QuicRemoveReferenceT<_Ty>& _Arg) noexcept { // forward an lvalue as either an lvalue or an rvalue
+    return static_cast<_Ty&&>(_Arg);
+}
+
+
```
nibanks, line 206: nit: extra newline

src/inc/msquic.hpp
ThadHouse: Why? It makes it so I don't have to redefine all the function pointers manually. Fairly common thing to wrap a C struct like this in C++.

src/perf/lib/ThroughputServer.cpp
```diff
@@ -0,0 +1,211 @@
...
+        QuicTraceLogInfo(
+            ConnectionEventPeerShutdown,
+            "[ conn] Connection Shutdown By Peer (%p).",
+            ConnectionHandle);
+        break;
+    case QUIC_CONNECTION_EVENT_SHUTDOWN_COMPLETE:
+        QuicTraceLogInfo(
+            ConnectionEventShutdownComplete,
+            "[ conn] Connection Shutdown Complete (%p).",
+            ConnectionHandle);
```
nibanks, line 161: Remember that we have logging for all this stuff already in msquic. You shouldn't need to duplicate it. Also, prefix your events with Perf, so they don't collide with any in the core.

ThadHouse: Ok, so our discussion yesterday should have been just removing the prints rather then adding tracing?

src/perf/lib/ThroughputClient.cpp
```diff
@@ -0,0 +1,409 @@
...
+    const char* LocalAddress;
+    if (TryGetValue(argc, argv, "bind", &LocalAddress)) {
+        if (!ConvertArgToAddress(LocalAddress, 0, &LocalIpAddr)) {
+            WriteOutput("Failed to decode IP address: '%s'!\nMust be *, a IPv4 or a IPv6 address.\n", LocalAddress);
+            return QUIC_STATUS_INVALID_PARAMETER;
+        }
+    }
+
+    // TODO: Core, since we need to support kernel mode
+    #ifdef QUIC_COMPARTMENT_ID
```
nibanks, line 72:
```suggestion
#ifdef QUIC_COMPARTMENT_ID
```

src/perf/lib/ThroughputClient.cpp
```diff
@@ -0,0 +1,409 @@
...
+            return QUIC_STATUS_INVALID_PARAMETER;
+        } else {
+            WriteOutput("Running in Compartment %d\n", CompartmentId);
+        }
+    }
+
+    uint8_t CpuCore;
+    if (TryGetValue(argc, argv, "core",  &CpuCore)) {
+        QuicSetCurrentThreadAffinityMask((DWORD_PTR)(1ull << CpuCore));
+    }
```
nibanks, line 87: Don't include this in the comparment ID ifdef

ThadHouse: We don't have a linux equivalent yet, so this is a way to hide it.

nibanks: Just wrap it in an #ifdef QuicSetCurrentThreadAffinityMask

src/perf/lib/ThroughputServer.cpp
nibanks: All depends on if the log duplicates something in msquic already.

src/perf/bin/appmain.cpp
ThadHouse: Ok. Just curious whats wrong with using QuicAlloc?

nibanks: It's not the cross-platform abstraction we use for allocations. The fact that Linux and Windows user modes happen to have a function called that is just coincidence and may change in the future.

src/perf/bin/drvmain.cpp
ThadHouse: This was renamed to use capitals to match PerfHelpers and other things.

src/perf/bin/drvmain.cpp
ThadHouse: This will need to be fixed in the BVTs then at some point too.

src/perf/lib/PerfHelpers.h
ThadHouse: I still don't have a buffer yet. I want to do it right, so I'll do that when I do kernel mode. But when I do I'll use the buffer for both.

src/inc/msquic.hpp
nibanks: Why can't you just have an operator that return the `table`?

ThadHouse: I'll find a day this month to fix this, and fix up the BVTs to allow me to fix this. Its not hard, just time consuming and I didn't want to be changing the BVTs at the same time. Once I do, this will be a lot more correct.

src/perf/bin/appmain.cpp
ThadHouse: Fixed.

src/inc/quic_driver_helpers.h
ThadHouse: Once this is in, I'll do fix this in the BVTs. I'm creating an issue for this and the other issues that  will involve changing the BVTs.

CMakeLists.txt
```diff
@@ -381,4 +381,6 @@ if(QUIC_BUILD_TEST)
     add_subdirectory(src/platform/unittest)
     add_subdirectory(src/test/lib)
     add_subdirectory(src/test/bin)
+    add_subdirectory(src/perf/lib)
+    add_subdirectory(src/perf/bin)
```
nibanks, line 385: Let's make "perf" it's own thing. Not test or tools.

src/inc/msquic.hpp
```diff
@@ -52,18 +52,118 @@ struct QuicAddr {
...
-    T* get() { return ptr; }
-    const T* get() const { return ptr; }
+    UniquePtr() noexcept = default;
+
+    explicit UniquePtr(T* _ptr) :
+        ptr{_ptr}
+    {
+    }
+
+    UniquePtr(const UniquePtr& other) = delete;
```
nibanks, line 63:
```suggestion
    UniquePtr() noexcept = default;
    explicit UniquePtr(T* _ptr) : ptr{_ptr} { }
    UniquePtr(const UniquePtr& other) = delete;
```

nibanks: Ditto below.

src/inc/msquic.hpp
nibanks: Ok.

src/perf/bin/appmain.cpp
nibanks: Ok.

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,228 @@
...
+QuicUserMain(
+    _In_ int argc,
+    _In_reads_(argc) _Null_terminated_ char* argv[],
+    _In_ bool KeyboardWait,
+    _In_ PerfSelfSignedConfiguration* SelfSignedConfig
+    ) {
+    QUIC_EVENT StopEvent;
+    QuicEventInitialize(&StopEvent, true, false);
+
+    QUIC_STATUS RetVal = QuicMainStart(argc, argv, StopEvent, SelfSignedConfig);
```
nibanks, line 46: Instead of `RetVal` let's use `Status` since that's what we use pretty much throughout the code base.

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+QUIC_STATUS
+QuicUserMain(
+    _In_ int argc,
+    _In_reads_(argc) _Null_terminated_ char* argv[],
+    _In_ bool KeyboardWait,
+    _In_ PerfSelfSignedConfiguration* SelfSignedConfig
+    ) {
+    QUIC_EVENT StopEvent;
+    QuicEventInitialize(&StopEvent, true, false);
```
nibanks, line 44: Might be worth making a C++ wrapper for QUIC_EVENT in msquic.hpp (but only if the platform is included before it).

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,228 @@
...
+    // Get total length
+    //
+    for (int i = 0; i < argc; ++i) {
+        //
+        // Length of string, plus null terminator, plus length
+        //
+        TotalLength += strlen(argv[i]) + 1;
+    }
+
+    TotalLength += sizeof(argc);
```
nibanks, line 86: Just initialize `TotalLength` to this above, instead of zero.

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,228 @@
...
+    size_t TotalLength = 0;
+
+    //
+    // Get total length
+    //
+    for (int i = 0; i < argc; ++i) {
+        //
+        // Length of string, plus null terminator, plus length
+        //
+        TotalLength += strlen(argv[i]) + 1;
```
nibanks, line 83:
```suggestion
        TotalLength += strlen(argv[i]) + 1;
```

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,228 @@
...
+    if (!Data) {
+        printf("Failed to allocate arguments to pass\n");
+        return QUIC_STATUS_OUT_OF_MEMORY;
+    }
+
+    char* DataCurrent = Data;
+
+    QuicCopyMemory(DataCurrent, &argc, sizeof(TotalLength));
+
+    DataCurrent += sizeof(argc);
```
nibanks, line 98:
```suggestion
    char* DataCurrent = Data;

    *((int*)DataCurrent) = argc;
    DataCurrent += sizeof(argc);
```
Personally, I prefer this approach that a direct call to memcpy. Also, you had a bug above because `sizeof(TotalLength)` does not equal `sizeof(argc)`. This pattern makes that type of bug impossible.

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,228 @@
...
+        ++DataCurrent;
+    }
+
+    QUIC_DBG_ASSERT(DataCurrent == (Data + TotalLength));
+
+    if (TotalLength > UINT_MAX) {
+        printf("Too many arguments to pass to the driver\n");
+        QUIC_FREE(Data);
+        return QUIC_STATUS_OUT_OF_MEMORY;
+    }
```
nibanks, line 114: This should be moved up to right after you calculate `TotalLength`.

src/perf/bin/appmain.cpp
```diff
@@ -0,0 +1,228 @@
...
+        RetVal = QUIC_STATUS_INTERNAL_ERROR;
+        goto Exit;
+    }
+
+    SelfSignedConfig.SelfSignedParams = SelfSignedParams;
+
+    if (TestingKernelMode) {
+#ifdef _WIN32
+        RetVal = QuicKernelMain(argc, argv, KeyboardWait, SelfSignedParams);
+#endif
```
nibanks, line 214:
```suggestion
#ifdef _WIN32
        RetVal = QuicKernelMain(argc, argv, KeyboardWait, SelfSignedParams);
#else
        QUIC_FRE_ASSERT(FALSE);
#endif
```

src/perf/bin/drvmain.cpp
```diff
@@ -0,0 +1,563 @@
...
+
+_Function_class_(EVT_WDF_DRIVER_UNLOAD)
+_IRQL_requires_same_
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+QuicPerfDriverUnload(
+    _In_ WDFDRIVER Driver
+    )
+{
+    UNREFERENCED_PARAMETER(Driver);
```
nibanks, line 186: Instead of this, just comment out the Driver variable name `/* Driver */` above.

src/perf/bin/drvmain.cpp
```diff
@@ -0,0 +1,563 @@
...
+{
+    NTSTATUS Status = STATUS_SUCCESS;
+
+    PAGED_CODE();
+
+    KeEnterGuardedRegion();
+    ExfAcquirePushLockExclusive(&QuicPerfCtlExtension->Lock);
+
+    do
+    {
```
nibanks, line 364:
```suggestion
    do {
```

src/perf/lib/ThroughputClient.cpp
```diff
@@ -0,0 +1,369 @@
...
+}
+
+struct ShutdownWrapper {
+    HQUIC ConnHandle {nullptr};
+    ~ShutdownWrapper() {
+        if (ConnHandle) {
+            MsQuic->ConnectionShutdown(ConnHandle, QUIC_CONNECTION_SHUTDOWN_FLAG_NONE, 0);
+        }
+    }
+};
```
nibanks, line 121: I feel like this doesn't need to be in this file, and possibly, instead, in msquic.hpp.

src/perf/lib/ThroughputServer.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+--*/
+
+#ifdef QUIC_CLOG
+#include "ThroughputServer.cpp.clog.h"
+#endif
+
+#ifndef _KERNEL_MODE
+#define QUIC_TEST_APIS 1
+#endif
```
nibanks, line 18: Why? I don't see any other kernel mode specific code below.

src/perf/lib/quicmain.cpp
```diff
@@ -0,0 +1,95 @@
...
+    Licensed under the MIT License.
+
+Abstract:
+
+    QUIC Perf Main execution engine.
+
+--*/
+
+
+
```
nibanks, line 13: nit: only one separating newline please

src/perf/lib/quicmain.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include "ThroughputClient.h"
+
+#include "quic_trace.h"
+#ifdef QUIC_CLOG
+#include "quicmain.cpp.clog.h"
+#endif
+
+const QuicApiTable* MsQuic {nullptr};
+
+PerfBase* TestToRun {nullptr};
```
nibanks, line 29: All global variable auto-initialize to zero'd out. No need to explicitly set nullptr.

src/perf/lib/quicmain.cpp
```diff
@@ -0,0 +1,95 @@
...
+    _In_ PerfSelfSignedConfiguration* SelfSignedConfig
+    ) {
+    const char* TestName = GetValue(argc, argv, "TestName");
+    uint8_t ServerMode = 0;
+    TryGetValue(argc, argv, "ServerMode", &ServerMode);
+
+    if (!TestName) {
+        WriteOutput("Must have a TestName specified. Ex: -TestName:Throughput\n");
+        return QUIC_STATUS_INVALID_PARAMETER;
+    }
```
nibanks, line 45:
```suggestion
    const char* TestName = GetValue(argc, argv, "TestName");
    if (!TestName) {
        WriteOutput("Must have a TestName specified. Ex: -TestName:Throughput\n");
        return QUIC_STATUS_INVALID_PARAMETER;
    }

    uint8_t ServerMode = 0;
    TryGetValue(argc, argv, "ServerMode", &ServerMode);
```

src/perf/bin/appmain.cpp
ThadHouse: That pattern is undefined behavior though. It has the potential for unaligned access, which ARM is especially sensitive about, even on Windows Arm.

src/perf/bin/drvmain.cpp
ThadHouse: We'll need to fix this in the test code then at some point too.

src/perf/lib/ThroughputClient.cpp
ThadHouse: This isn't general purpose though. Its specific to how I'm initializing things. Sometimes you might need other types of handles with an RAII wrapper.

src/perf/lib/ThroughputServer.cpp
ThadHouse: If you have this defined in kernel mode, it fails to build, because of what gets included behind this flag.

src/perf/bin/appmain.cpp
ThadHouse: Maybe when I go through and clean up the C++ wrappers completely I'll do that.

src/perf/bin/appmain.cpp
nibanks: Why is it any different from accessing a packed struct? This pattern is used all over the Windows code base.

CMakeLists.txt
```diff
@@ -359,6 +360,12 @@ if(QUIC_BUILD_TOOLS)
     add_subdirectory(src/tools)
 endif()
 
+# Tool code
```
nibanks, line 363:
```suggestion
# Performance code
```

scripts/build.ps1
```diff
@@ -30,6 +30,9 @@ This script provides helpers for building msquic.
 .PARAMETER DisableTest
     Don't build the test directory.
 
+.PARAMETER DisablePerf
+    Don't build the perf directory.
```
nibanks, line 34: The perf pipeline should now be able to disable test and tool code from building.

src/perf/bin/appmain.cpp
ThadHouse: Its more the Arm Linux builds I'm worried about. The windows compiler goes out of its way to try and make things like this work. Gcc doesn't do that. In this case its fine, because its the first item added. But if it happened at the end that would cause arm linux to segfault.

PR #633, opened 2020-07-23
Code Coverage Core Only

nibanks:
This updates code coverage to only look at the core code for now, which is what I'm most interested in (for the time being). Also, it enables some logging for spinquic to get some code coverage in the logging/tracing code paths. From my local execution, this brings coverage up to 80%. Not a bad start!

### DIFF
--- a/scripts/run-gtest.ps1
+++ b/scripts/run-gtest.ps1
@@ -264,7 +264,7 @@ function Start-TestExecutable([String]$Arguments, [String]$OutputDir) {
         } elseif ($CodeCoverage) {
             $CoverageOutput = Join-Path $OutputDir $CoverageName
             $pinfo.FileName = "C:\Program Files\OpenCppCoverage\OpenCppCoverage.exe"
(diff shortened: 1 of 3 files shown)


scripts/run-gtest.ps1
```diff
@@ -264,7 +264,7 @@ function Start-TestExecutable([String]$Arguments, [String]$OutputDir) {
         } elseif ($CodeCoverage) {
             $CoverageOutput = Join-Path $OutputDir $CoverageName
             $pinfo.FileName = "C:\Program Files\OpenCppCoverage\OpenCppCoverage.exe"
-            $pinfo.Arguments = "--modules=$(Split-Path $Path -Parent) --cover_children --sources src\core --sources src\inc --sources src\platform --excluded_sources unittest --working_dir $($OutputDir) --export_type binary:$($CoverageOutput) -- $($Path) $($Arguments)"
```
anrossi, line 267: Why shouldn't we keep coverage for platform?  We should make sure that changes in platform are still covered by the unittests.

nibanks: I want to start with a smaller scope here. Scoping to just the core also provides a simpler view of code coverage by default. With platform included, we're getting duplication of these files for ever test application, which bloats the coverage gaps. I don't want to enable platform until we can eliminate that bloat.

Issue #611, opened 2020-07-17
Official Windows User Mode Binary Distribution


ThadHouse:
We should start looking into ways to do distribution for msquic. Right now, these are the ideas floating around.

1.  ~~vcpkg. This is becoming a lot more popular, and relatively easy to set up. I've done this before for other projects.~~

2. NuGet. Best to publish to the NuGet.org, that way no repo has to be added. This will only work from Visual Studio, but will make it easy to use in VS.

3. Keep making sure cmake is installable. This will enable the standard cmake workflow, and also allow making a submodule out of cmake. I also want to make external_project in cmake work, it shouldn't be any harder

4. Distribute the headers and shared object in a zip file on the releases page.

It might also be nice to distribute through the normal package managers as well, we could likely hook into the same repos pwsh and dotnet core use if we wanted to go this route.

nibanks:
@thhous-msft what are we tracking with this issue now? Should we close it?

nibanks:
Closing as complete.

PR #635, opened 2020-07-23
Add zip archive distribution for libraries

ThadHouse:
Takes all artifacts and zips them into a zip with the correct headers and binaries.

Partial work on #611, still much more to go.

### DIFF
--- a/.azure/azure-pipelines.ci.yml
+++ b/.azure/azure-pipelines.ci.yml
@@ -359,3 +359,16 @@ stages:
         arguments: -Branch $(Build.SourceBranch)
       env:
         AzDO_PAT: $(AzDO_PAT)
+
+#
+# Distribution
+#
+
+- stage: distribution
+  displayName: Distribution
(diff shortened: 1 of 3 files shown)


.azure/azure-pipelines.ci.yml
```diff
@@ -359,3 +359,16 @@ stages:
...
+#
+# Distribution
+#
+
+- stage: distribution
+  displayName: Distribution
+  dependsOn:
+  - build_windows
+  - build_linux
+  - build_winkernel
```
nibanks, line 372: No need to worry about kernel mode stuff.

nibanks: Remove the dependency on build_winkernel too; and all the winkernel references in the distribution script.

nibanks:
I feel like this should only run on master/release branches. Not PRs.

ThadHouse:
Done, this now only runs on non prs.

PR #636, opened 2020-07-23
Fix cmake installs since we added clog.

ThadHouse:
This uses a tracepoint provider for lttng, which is a common practice, coreclr does the same thing.


scripts/package-distribution.ps1
```diff
@@ -82,8 +82,7 @@ foreach ($Build in $AllBuilds) {
     } else {
         $Binaries += Join-Path $ArtifactsDir "libmsquic.so"
         #Temporary until we fix CLOG
-        $Binaries += Join-Path $ArtifactsDir "libPLATFORM_CLOG_LIB.so"
-        $Binaries += Join-Path $ArtifactsDir "libCORE_CLOG_LIB.so"
+        $Binaries += Join-Path $ArtifactsDir "libmsquictraceprovider.so"
```
nibanks, line 85: Can this be named `libmsquic.traceprovider.so` or (since I think it's only necessary when we use lttng) `libmsquic.lttng.so`?

src/platform/platform_linux.c
```diff
@@ -63,6 +66,60 @@ QuicPlatformSystemLoad(
...
+    // https://github.com/dotnet/coreclr/blob/ed5dc831b09a0bfed76ddad684008bebc86ab2f0/src/pal/src/misc/tracepointprovider.cpp#L106
+    //
+
+    int fShouldLoad = 1;
+    // Check if loading the LTTng providers should be disabled.
+    char *disableValue = getenv("QUIC_LTTng");
+    if (disableValue != NULL)
+    {
+        fShouldLoad = strtol(disableValue, NULL, 10);
+    }
```
nibanks, line 80:
```suggestion
    if (disableValue != NULL) {
        fShouldLoad = strtol(disableValue, NULL, 10);
    }
```

src/platform/platform_linux.c
```diff
@@ -63,6 +66,60 @@ QuicPlatformSystemLoad(
     void
     )
 {
+    //
+    // Following code is modified from coreclr.
+    // https://github.com/dotnet/coreclr/blob/ed5dc831b09a0bfed76ddad684008bebc86ab2f0/src/pal/src/misc/tracepointprovider.cpp#L106
+    //
+
+    int fShouldLoad = 1;
```
nibanks, line 74: None of these variables follow our style.

scripts/package-distribution.ps1
ThadHouse: Yeah I'll do that.

scripts/prepare-machine.ps1
```diff
@@ -21,7 +21,7 @@ param (
     [string]$Configuration
 )
 
-#Requires -RunAsAdministrator
+##Requires -RunAsAdministrator
```
nibanks, line 24: Was this change on purpose?

ThadHouse: Nope. Fixed.

src/bin/CMakeLists.txt
```diff
@@ -34,9 +34,19 @@ set(PUBLIC_HEADERS
...
-#install(FILES ${CMAKE_BINARY_DIR}/msquic-config.cmake DESTINATION ${msquic_dest})
-#install(EXPORT msquic DESTINATION ${msquic_dest})
+configure_file(msquic-config.cmake.in ${CMAKE_BINARY_DIR}/msquic-config.cmake)
+
+install(FILES ${CMAKE_BINARY_DIR}/msquic-config.cmake DESTINATION ${msquic_dest})
+
+install(EXPORT msquic DESTINATION ${msquic_dest})
+
+if(WIN32)
+    add_library(msquic.lttng INTERFACE)
```
nibanks, line 47: Why are you doing this on Windows? There is no lttng on Windows.

src/platform/platform_linux.c
```diff
@@ -63,6 +66,61 @@ QuicPlatformSystemLoad(
     void
     )
 {
+    //
+    // Following code is modified from coreclr.
+    // https://github.com/dotnet/coreclr/blob/ed5dc831b09a0bfed76ddad684008bebc86ab2f0/src/pal/src/misc/tracepointprovider.cpp#L106
+    //
+
+    int ShouldLoad = 1;
+    // Check if loading the LTTng providers should be disabled.
```
nibanks, line 75: Comments in this function don't follow our style.

src/bin/CMakeLists.txt
ThadHouse: I need a target on windows, since downstream executables use this target to make loading slightly easier. This just creates an empty target, so every downstream use doesn't require an if check.

nibanks: Ok. Though, I expect this will eventually move into clog somehow...

PR #643, opened 2020-07-24
Fix server disconnect failure

nibanks:
Fixes a bug I saw in a test failure. The code didn't wait enough and ended up failing. Ideally, we'd have some better way to do this than just sleep and hope...


anrossi:
Couldn't the server set an event when the shutdown is complete on its end, and the client waits on that?

nibanks:
> Couldn't the server set an event when the shutdown is complete on its end, and the client waits on that?

Possibly, but I wanted to try a quick solution. Feel free to try your suggestion.

PR #644, opened 2020-07-24
Make QuicProcCurrentNumber thread group aware

### LINKED ISSUE #616: QuicProcCurrentNumber does not take into account processor groups
`GetCurrentProcessorNumber` returns the processor number relative to the current processor group. This means places in the code where we are using the current processor number to index will only access the indices of the first group.

`KeGetCurrentProcessorIndex` isn't documented, but might have the same issue as well.

ThadHouse:
Still need to double check if kernel mode has the same issue.

Closes #616


src/platform/platform_winuser.c
```diff
@@ -67,6 +68,14 @@ QuicProcessorInfoInit(
     uint8_t* Buffer = NULL;
     uint32_t Offset;
 
+    //
+    // dwNumberOfProcessors is specified to be number of processors in current
+    // group. Processor groups are specified to be equal length.
+    //
+    SYSTEM_INFO SysInfo;
+    GetSystemInfo(&SysInfo);
+    QuicProcessorsPerGroup = (uint8_t)SysInfo.dwNumberOfProcessors;
```
nibanks, line 77: Can you not get this info from the data below?

ThadHouse: I'd have to count it manually, to find the count of group 0. I could do that if you want though, but would be a little wonky

src/inc/quic_platform_winuser.h
```diff
@@ -710,10 +710,22 @@ typedef struct {
...
+
+_IRQL_requires_max_(DISPATCH_LEVEL)
+inline
+DWORD
+QuicProcCurrentNumber(
+    void
+    ) {
+    PROCESSOR_NUMBER ProcNumber;
+    GetCurrentProcessorNumberEx(&ProcNumber);
+    return (ProcNumber.Group * QuicProcessorsPerGroup) + ProcNumber.Number;
```
nibanks, line 726: Multiplication is not super efficient. I think it would be better to have a per-group lookup table that returns the base index.

src/platform/platform_winuser.c
nibanks: Ok. I guess I can live with this then.

nibanks:
This looks great! Thanks for putting it together, but I think we need the same thing for kernel mode. So one of the following needs to be done:

1. Definitely confirm that kernel mode doesn't need this.
2. Don't resolve the issue with this PR.
3. Add kernel mode to this PR.

PR #652, opened 2020-07-28
Skip calling CLOG if logging is disabled

ThadHouse:
Adds an unnecessary dependency if logging is disabled anyway.

The unreferenced parameter changes are needed to allow -DisableLogs on windows, otherwise the windows build fails.


src/platform/datapath_winuser.c
```diff
@@ -2402,6 +2402,9 @@ QuicSendContextComplete(
     _In_ ULONG IoResult
     )
 {
+#ifdef QUIC_EVENTS_STUB
```
nibanks, line 2405: Don't wrap in ifdef.

CMakeLists.txt
```diff
@@ -148,21 +148,33 @@ set(QUIC_OUTPUT_DIR ${QUIC_BUILD_DIR}/bin/$<IF:$<CONFIG:Debug>,Debug,Release> CA
...
-set(CMAKE_CLOG_OUTPUT_DIRECTORY ${QUIC_BUILD_DIR}/inc)
-set(CMAKE_CLOG_SIDECAR_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/src/manifest)
-set(CLOG_INCLUDE_DIRECTORY ${CMAKE_SOURCE_DIR}/build/clog)
-set(CMAKE_CLOG_GENERATE_FILE ${CMAKE_SOURCE_DIR}/build/clog/CLog.cmake)
-set(CMAKE_CLOG_CONFIG_FILE ${CMAKE_CURRENT_SOURCE_DIR}/src/manifest/msquic.clog_config)
-include(${CMAKE_CLOG_GENERATE_FILE})
+if (QUIC_ENABLE_LOGGING)
+    execute_process(COMMAND clog --installDirectory ${CMAKE_SOURCE_DIR}/build/clog)
+
+    set(CMAKE_CLOG_OUTPUT_DIRECTORY ${QUIC_BUILD_DIR}/inc)
```
nibanks, line 161: The point of taking clog into master was to always run clog everywhere. We don't want to disable it.

ThadHouse: Is eventually the goal that it will be required, and quic_trace.h would be gone? Right now its includes are behind QUIC_CLOG.

nibanks: It's still a WIP so I'm not positive what the final product will look like.

nibanks: Ok. I thought about this, I'm fine if we do this, but I want all our CI builds (including the pi) to not disable logs then. We need the build coverage for clog.

src/core/CMakeLists.txt
```diff
@@ -45,8 +45,7 @@ endif()
 
 # Allow CLOG to preprocess all the source files.
 CLOG_GENERATE_TARGET(core.clog ${SOURCES})
-
-target_link_libraries(core.clog PRIVATE inc)
+CLOG_ADD_ARGS(core.clog)
```
nibanks, line 48: Do we ever call `CLOG_GENERATE_TARGET` but not `CLOG_ADD_ARGS`? If not, we should have one wrapper function that calls clogs and does the `target_link_libraries`.

ThadHouse: We actually can't get lttng working on the Pi. That was actually my use case for disabling this. Same with mac, which doesn't have logging currently.

As for the function, I want to change that up a bit anyway.

ThadHouse:
I also want to use this function to put the clog targets into a clog folder using the set property thing.

Issue #634, opened 2020-07-23
Improve Code Coverage


nibanks:
Now that we have code coverage enabled there are a few places we can easily improve test coverage for.

- [x] Get/Set all parameters for all types of objects
- [x] Trigger the Settings updated code path by setting a registry key
- [x] Test out multiple registration execution modes
- [x] Test Server IP load balancing mode
- [ ] Packet drops in validating received initial token
- [x] Trace rundowns in the middle of execution

PR #659, opened 2020-07-30
More Code Coverage

nibanks:
- Updates spinquic to get parameters randomly now.
- Rework a binding code path where the `else` was pointless because it was after a `return`.

Related to #634.

### DIFF
--- a/src/tools/spin/spinquic.cpp
+++ b/src/tools/spin/spinquic.cpp
@@ -117,6 +117,7 @@ typedef enum {
     SpinQuicAPICallStreamClose,
     SpinQuicAPICallSetParamSession,
     SpinQuicAPICallSetParamConnection,
+    SpinQuicAPICallGetParam,
     SpinQuicAPICallDatagramSend,
     SpinQuicAPICallCount    // Always the last element
 } SpinQuicAPICall;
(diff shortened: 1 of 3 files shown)


src/tools/spin/spinquic.cpp
```diff
@@ -442,6 +443,30 @@ void SpinQuicSetRandomConnectionParam(HQUIC Connection)
...
+    QUIC_PARAM_CONN_DISABLE_1RTT_ENCRYPTION + 1
+};
+
+void SpinQuicGetRandomParam(HQUIC Connection)
+{
+    QUIC_PARAM_LEVEL Level = (QUIC_PARAM_LEVEL)GetRandom(5);
+    uint32_t Param = (uint32_t)GetRandom(ParamCounts[Level]);
+
+    uint8_t OutBuffer[512];
+    uint32_t OutBufferLength = (uint32_t)GetRandom(sizeof(OutBuffer) + 1);
```
anrossi, line 460: Is the goal to have `OutBufferLength` potentially be 1 more than the size of the buffer?  I think you want the `+1` outside the `GetRandom` call.

nibanks: No. That's not the point. The goal is for it to be between 0 and `sizeof(OutBuffer)` **inclusively**.

nibanks:
Looks like this ups code coverage to ~81%: https://dev.azure.com/ms/msquic/_build/results?buildId=98084&view=results

PR #675, opened 2020-08-03
QUIC Network Simulation Logging Support

nibanks:
Updates the QUIC network simulation script to collect logs.


scripts/run_endpoint.sh
```diff
@@ -41,3 +46,8 @@ elif [ "$ROLE" == "server" ]; then
     quicinteropserver ${SERVER_PARAMS} -root:/www -listen:* -port:443 \
```
nibanks, line 46: Not sure how to stop this app from running "when the test is done" though.

scripts/run_endpoint.sh
```diff
@@ -41,3 +55,5 @@ elif [ "$ROLE" == "server" ]; then
...
         -file:/server.crt -key:/server.key 2>&1
 fi
+
+trap finish EXIT
```
nibanks, line 59: Ok. Well the logs start just fine, but it seems this `trap` isn't working as expected. I never get any output from the server when it's shutdown.

PR #676, opened 2020-08-04
Add ECN support to the Linux datapath 

larseggert:
**Not ready for merging yet!** I started the PR so that folks could take a look and for the CI pipeline to kick in, to tell me if I broke the Windows build.

This set of patches makes msquic send ECT(0)-marked packets. It also extracts the ECN information from received packets, counts them per packet space, and uses those counters to send ACK frames with ECN information.

There is currently no logic for any ECN validation, and therefore no support for disabling ECN when that validation would fail.

There is also a (minor?) issue (see the `FIXME`) that causes ECN marks on 0-RTT packets to not be counted.

### DIFF
--- a/src/platform/unittest/DataPathTest.cpp
+++ b/src/platform/unittest/DataPathTest.cpp
@@ -74,7 +74,7 @@ struct DataRecvContext {
     QUIC_EVENT ClientCompletion;
 };
 
-struct DataPathTest : public ::testing::TestWithParam<int32_t>
+struct DataPathTest : public ::testing::TestWithParam<std::tuple<int, QUIC_ECN_TYPE>>
 {
 protected:
     static volatile uint16_t NextPort;
@@ -120,7 +120,7 @@ protected:
     QuicAddr
     GetNewLocalAddr(bool randomPort = true)
     {
-        int addressFamily = GetParam();
+        int addressFamily = std::get<0>(GetParam());
 
         if (addressFamily == 4) {
             return GetNewLocalIPv4(randomPort);
@@ -170,10 +170,11 @@ protected:
     }
 
     static void
-    DataRecvCallback(
+    DoDataRecvCallback(
         _In_ QUIC_DATAPATH_BINDING* binding,
         _In_ void * recvContext,
-        _In_ QUIC_RECV_DATAGRAM* recvBufferChain
+        _In_ QUIC_RECV_DATAGRAM* recvBufferChain,
+        _In_ bool checkECN
         )
     {
         DataRecvContext* RecvContext = (DataRecvContext*)recvContext;
@@ -187,8 +188,10 @@ protected:
 
             if (recvBuffer->Tuple->LocalAddress.Ipv4.sin_port == RecvContext->ServerAddress.Ipv4.sin_port) {
 
+                QUIC_ECN_TYPE ecn = (QUIC_ECN_TYPE)recvBuffer->TypeOfService;
+
                 auto ServerSendContext =
-                    QuicDataPathBindingAllocSendContext(binding, QUIC_ECN_NON_ECT, 0);
+                    QuicDataPathBindingAllocSendContext(binding, ecn, 0);
                 ASSERT_NE(nullptr, ServerSendContext);
 
                 auto ServerDatagram =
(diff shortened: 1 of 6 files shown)


nibanks:
Thanks @larseggert for starting this PR. One problem though, is that our contributing guidelines don't allow for external developers to modify the core files:

https://github.com/microsoft/msquic/blob/master/.github/CONTRIBUTING.md
> **Important** - We are still bringing up important regression tests for the core code. Until they are onboarded, any external contributions to the [core](../src/core) or kernel mode files in the [platform](../src/platform) will not be accepted. This is only a **temporary restriction** and we are working to complete it by the end of 2020.

Let's work together and I will see if I can get the core parts updated to match any changes you make in the platform.

nibanks:
@larseggert, please merge master and I *think* all your core dependencies will be there now.

larseggert:
@nibanks thanks for adding the core stuff!

One additional addition is needed to core, but I didn't know what form you'd like it to take, hence I didn't do a strawman: In order to disable ECN when validation fails, the datapath API *either* needs functions that the core can call that will turn off (and on) ECN-marking for a socket (i.e., do a `setsockopt`) *or* on TX, the desired value of the TOS byte for each buffer should be passed down alongside is, to be set with ancillary data, *or* there needs to be a connection-level flag that can be accessed from inside the datapath indicating whether ECT(0) should be marked at any given time or not.

(There may be other options.)

Either of these would work, but I don't know which one you'd prefer?

nibanks:
@larseggert I think my preference would be to have the send path take the additional argument to set the ECN per-send.

larseggert:
OK, so when you get a chance, add that? :-)

larseggert:
(Let's see if this does something...)

src/core/packet_space.h
```diff
@@ -104,6 +104,11 @@ typedef struct QUIC_PACKET_SPACE {
     //
     BOOLEAN AwaitingKeyPhaseConfirmation: 1;
 
+    //
+    // ECN marks of received packets.
+    //
+    QUIC_ACK_ECN_EX RecvECN;
```
nibanks, line 110: This is already being tracked in the ACK tracker. This should be unnecessary.

larseggert: Good catch. Fixed.

src/core/connection.c
```diff
@@ -4772,6 +4772,27 @@ QuicConnRecvDatagramBatch(
...
+                    case QUIC_ECN_ECT_1:
+                        PacketSpace->AckTracker.ReceivedECN.ECT_1_Count++;
+                        break;
+                    case QUIC_ECN_ECT_0:
+                        PacketSpace->AckTracker.ReceivedECN.ECT_0_Count++;
+                        break;
+                    case QUIC_ECN_NON_ECT:
+                        break;
+                }
+            }
```
nibanks, line 4794: I'm also already setting this in the "ack packet" code path" in the ack_tracker.c.

larseggert: Oh yes

larseggert:
@nibanks I think this might be baked now

larseggert:
I will need to hook in with the changes from #685 still.

nibanks:
@larseggert the change looks good, but we at least need a unit test (currently scoped to Linux) that send a datagram over loopback with a particular ECN marking and validates it's received correctly. Could you add a simple test case to .\src\platform\unittest\DataPathTest.cpp?

larseggert:
I can take a stab tomorrow.

larseggert:
The failing tests here seem to be due to the fact that the Windows datapath doesn't do ECN yet.

src/platform/unittest/DataPathTest.cpp
```diff
@@ -429,10 +455,12 @@ TEST_P(DataPathTest, DataRebind)
 
     QuicEventInitialize(&RecvContext.ClientCompletion, FALSE, FALSE);
 
+    QUIC_ECN_TYPE ecn = std::get<1>(GetParam());
```
nibanks, line 458: I'd prefer not to reuse the DataRebind test, and just make a new one.

larseggert: Ok. I will get to this next week.

nibanks:
Looks good. Since the tests don't pass on Windows, we need to add an ignore filter to not run them. Please update azure-pipelines.yml to add a negative filter to the `extraArgs` for this.

larseggert:
With `azure-pipelines.yml` you mean `./submodules/wil/scripts/azure-pipelines.yml`? I don't quite understand what you want me to add where?

nibanks:
mitls already has a couple of negative filters:
```
extraArgs: -Filter -*Unreachable/0:CryptTest/CryptTest.Encryption/2:TlsTest.CertificateError
```
They are `:` delaminated. Just add your new test to the end of this one and schannel.

larseggert:
So the file is `.azure/azure-pipelines.ci.yml`?

nibanks:
Yes, sorry if that wasn't clear.

PR #684, opened 2020-08-05
Fix build with clang under Linux

larseggert:
Use `CMAKE_C_COMPILER` and `CMAKE_CXX_COMPILER` when building openssl.

Also add to `QUIC_WARNING_FLAGS` to suppress some additional warnings that clang's `-Wextra` emits (and that `-Werror` then aborts over.) Apply the modified `QUIC_WARNING_FLAGS` to some parts of the build that didn't use them so far.

(It would be possible to add `-Weverything` when building with clang, to get even more warnings - but those would then all need to be checked and suppressed...)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -198,10 +198,7 @@ else()
         -Wno-unknown-pragmas -Wno-unused-value -Wno-multichar -Wno-switch
         CACHE INTERNAL "")
     if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-        list(APPEND QUIC_WARNING_FLAGS -Wno-missing-braces
-             -Wno-microsoft-anon-tag -Wno-enum-conversion -Wno-reorder
-             -Wno-tautological-constant-out-of-range-compare
-             -Wno-format-nonliteral -Wno-unused-private-field)
+        list(APPEND QUIC_WARNING_FLAGS -Wno-missing-braces -Wno-microsoft-anon-tag)
     endif()
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -194,7 +194,15 @@ if (WIN32)
 else()
     set(QUIC_COMMON_FLAGS "")
     set(QUIC_COMMON_DEFINES "")
-    set(QUIC_WARNING_FLAGS -Werror -Wall -Wextra -Wformat=2 -Wno-type-limits -Wno-unknown-pragmas -Wno-unused-value -Wno-multichar CACHE INTERNAL "")
+    set(QUIC_WARNING_FLAGS -Werror -Wall -Wextra -Wformat=2 -Wno-type-limits
+        -Wno-unknown-pragmas -Wno-unused-value -Wno-multichar -Wno-switch
+        CACHE INTERNAL "")
+    if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+        list(APPEND QUIC_WARNING_FLAGS -Wno-missing-braces
+             -Wno-microsoft-anon-tag -Wno-enum-conversion -Wno-reorder
+             -Wno-tautological-constant-out-of-range-compare
+             -Wno-format-nonliteral -Wno-unused-private-field)
+    endif()
 endif()
 
 if(WIN32)
--- a/src/perf/lib/CMakeLists.txt
+++ b/src/perf/lib/CMakeLists.txt
@@ -23,7 +23,7 @@ target_include_directories(perflib PUBLIC ${CMAKE_CURRENT_LIST_DIR})
 if (MSVC)
     target_compile_options(perflib PUBLIC /wd4459)
 else()
-    target_compile_options(perflib PUBLIC -Wno-switch)
+    target_compile_options(perflib PUBLIC ${QUIC_WARNING_FLAGS})
 endif()
 
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
--- a/src/perf/lib/CMakeLists.txt
+++ b/src/perf/lib/CMakeLists.txt
@@ -22,8 +22,6 @@ target_include_directories(perflib PUBLIC ${CMAKE_CURRENT_LIST_DIR})
 
 if (MSVC)
     target_compile_options(perflib PUBLIC /wd4459)
-else()
-    target_compile_options(perflib PUBLIC ${QUIC_WARNING_FLAGS})
 endif()
 
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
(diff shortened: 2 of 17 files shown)


CMakeLists.txt
```diff
@@ -194,7 +194,15 @@ if (WIN32)
 else()
     set(QUIC_COMMON_FLAGS "")
     set(QUIC_COMMON_DEFINES "")
-    set(QUIC_WARNING_FLAGS -Werror -Wall -Wextra -Wformat=2 -Wno-type-limits -Wno-unknown-pragmas -Wno-unused-value -Wno-multichar CACHE INTERNAL "")
+    set(QUIC_WARNING_FLAGS -Werror -Wall -Wextra -Wformat=2 -Wno-type-limits
+        -Wno-unknown-pragmas -Wno-unused-value -Wno-multichar -Wno-switch
```
ThadHouse, line 198: I think I would prefer to not have `Wno-switch` be global. This should only be affecting the perf lib anyway. I found this with the mac build, and I'll look there for what our solution was.

src/test/lib/CMakeLists.txt
```diff
@@ -31,4 +31,6 @@ set_property(TARGET testlib PROPERTY FOLDER "tests")
 
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
     add_dependencies(platform MsQuicEtw)
+else()
+    target_compile_options(testlib PRIVATE ${QUIC_WARNING_FLAGS})
```
ThadHouse, line 35: This is actually already here because of the target_link_libraries warnings above. Did this actually do anything different?

CMakeLists.txt
ThadHouse: Nevermind, turns out for Mac we also just had to globally add this. So I'm fine with it

src/perf/lib/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ target_include_directories(perflib PUBLIC ${CMAKE_CURRENT_LIST_DIR})
 if (MSVC)
     target_compile_options(perflib PUBLIC /wd4459)
 else()
-    target_compile_options(perflib PUBLIC -Wno-switch)
+    target_compile_options(perflib PUBLIC ${QUIC_WARNING_FLAGS})
```
ThadHouse, line 26: Just like below, warnings are already added with the target link libraries warning above. So we should just be able to completely remove this else branch now.

larseggert: I can double-check tomorrow, but I did see errors without these changes.

ThadHouse:
Definitely no on `-Weverything`. I really don't want to have to go and change anything anytime a new compiler comes out. Especially since suppressions of unknown warnings are errors in older compilers.

ThadHouse:
Also will need merge conflicts fixed because of merging #683

src/perf/lib/CMakeLists.txt
larseggert: I checked, and if i take this out, I get the following compile error:
```
In file included from /vagrant/src/perf/lib/PerfHelpers.h:32:
/vagrant/src/inc/msquic.hpp:239:11: error: field 'Handle' will be initialized after field 'CloseAllConnectionsOnDelete' [-Werror,-Wreorder]
        : Handle(nullptr), CloseAllConnectionsOnDelete(false) {
          ^
1 error generated.
```

src/test/lib/CMakeLists.txt
larseggert: I checked, and if I take this out, I get the following compile error:
```
In file included from /vagrant/src/test/MsQuicTests.h:24:
/vagrant/src/inc/msquic.hpp:239:11: error: field 'Handle' will be initialized after field 'CloseAllConnectionsOnDelete' [-Werror,-Wreorder]
        : Handle(nullptr), CloseAllConnectionsOnDelete(false) {
          ^
/vagrant/src/inc/msquic.hpp:262:11: error: field 'Handle' will be initialized after field 'CloseAllConnectionsOnDelete' [-Werror,-Wreorder]
        : Handle(nullptr), CloseAllConnectionsOnDelete(false) {
          ^
/vagrant/src/inc/msquic.hpp:278:11: error: field 'Handle' will be initialized after field 'CloseAllConnectionsOnDelete' [-Werror,-Wreorder]
        : Handle(nullptr), CloseAllConnectionsOnDelete(false) {
          ^
```

larseggert:
I rebased and verified that the two additions you asked about are in fact needed to make the build work.

larseggert:
I took a slightly different approach, and decided to simply fix most warnings that clang complained about, rather than suppressing them.

nibanks:
@larseggert, lots of build errors in Windows (user & kernel) with this:

https://dev.azure.com/ms/msquic/_build/results?buildId=99761&view=logs&j=1204e69e-9c52-5fc3-715c-177f7fdb0a3a&t=b2eebfab-7505-58e7-4497-5b1e11c53393

larseggert:
I wonder if Windows needs a `#include <inttypes.h>` for `PRIu64`?

nibanks:
> I wonder if Windows needs a `#include <inttypes.h>` for `PRIu64`?

I'm not sure. What error are you looking at?

larseggert:
`##[error]src\perf\lib\ThroughputClient.cpp(372,30): Error C2065: 'PRIu64': undeclared identifier`

larseggert:
OK, I pushed two fixes that hopefully will be enough

nibanks:
I don't think that's defined on Windows. According to this Stackoverflow question (https://stackoverflow.com/questions/16859500/mmh-who-are-you-priu64) it doesn't look like there's an easy solution here. What issue do you get when you just use `%llu` (which we use all over the place)?

larseggert:
The error is
```
/vagrant/src/perf/lib/ThroughputClient.cpp:376:13: error: format specifies type 'unsigned long long' but the argument has type 'uint64_t'
      (aka 'unsigned long') [-Werror,-Wformat]
            GetStreamID(MsQuic, StreamHandle),
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

ThadHouse:
That header does exist on Windows. Anthony ran into the same issue in another place. The current azure error seems unrelated.

larseggert:
Well, there are bunch of Azure errors, but one of them is
```
##[error]src\perf\lib\ThroughputClient.cpp(12,10): Error C1083: Cannot open include file: 'inttypes.h': No such file or directory
```

larseggert:
Anyone know what that failing "spinquic stub" is? The error doesn't really makes this clear (to me).

larseggert:
It might make sense to add clang to the CI matrix, too.

PR #686, opened 2020-08-05
Perf Refactoring

nibanks:
Some fairly minor refactoring in preparation for adding some initial RPS test cases.


src/perf/lib/quicmain.cpp
```diff
@@ -65,9 +60,9 @@ QuicMainStart(
 
     if (IsValue(TestName, "Throughput")) {
         if (ServerMode) {
-            TestToRun = new ThroughputServer{SelfSignedConfig};
```
ThadHouse, line 68: Keep this as braced based initialization. Its safer and more correct.

src/perf/lib/quicmain.cpp
```diff
@@ -65,9 +60,9 @@ QuicMainStart(
...
+            TestToRun = new ThroughputServer(SelfSignedConfig);
         } else {
-            TestToRun = new ThroughputClient{};
```
ThadHouse, line 70: I would prefer to not get rid of these. The errors if something changes is a bit cleaner.

ThadHouse:
Kernel code needs to be fixed because of the main header change.

src/perf/lib/quicmain.cpp
ThadHouse: In addition, my proposed C++ styleguide changes say to always used brace based initialization, since its more uniform and avoids some fun parsing bugs.

nibanks: The rest of the code base has always used parenthesis for constructors. I want to keep that consistent. Additionally, as far as I can tell, that is the normal way to call C++ constructors. I'd never seen them called your way before this. The current pattern we use is `{ ... }` for C struct initializations (no constructor) and `(...)` for C++ constructors.

src/perf/lib/quicmain.cpp
nibanks: Either way it's just calling the default constructor. I don't see how errors would be different. Personally, I prefer no parenthesis if there are no arguments.

src/perf/lib/quicmain.cpp
ThadHouse: Brace based initialization was added in C++ 11. It changes a few things, mainly it won't allow implicit conversions. Its also uniform since it doesn't matter what type of initialization is being done. All the code I committed in here used it, even for member initialization. So if we want to change it, which I don't think we should, we'd need to change a lot more then this.

nibanks: All the test code (a lot more code than this) uses the other pattern. Where, beyond class declaration of member variables, is this pattern actually used?

ThadHouse: Theres a few, mainly in initialization lists. I was actually planning on going through at some point and PR changing all the initialization to use it. Just hadn't gotten to it yet.

https://github.com/microsoft/msquic/blob/master/src/inc/msquic.hpp#L58

ThadHouse: But its not the end of the world to not use it. Just going to just be an extremely hard habit to break, especially since most actual C++ projects have switched.

nibanks: IMO, using the braces makes it hard to differentiate if this is calling a constructor or it's just initializing the member variables of a struct. You either have to know if it's a class or a C struct; or you have to go look at the declaration. I prefer not to use this form of initialization primarily for that reason.

src/perf/lib/quicmain.cpp
ThadHouse: I guess. I'm just always a fan of being more explicit, But its not a big deal.

src/perf/lib/quicmain.cpp
ThadHouse: Sure. I just came from a mostly C++ perspective, where not having to think if you're initializing on the stack or with new (since paranthesis don't work with stack) is much easier. Never thought about the mixing C and C++, especially since the brace initialization in C isn't always fully standard, and everything I've worked on in the past has explicitly cared about being standards compliant.

PR #689, opened 2020-08-06
With some fixes, we can remove "-Wno-unused-value"

larseggert:
Also fix a bug (== instead of =) found by not suppressing those warnings anymore.

### DIFF
--- a/src/inc/quic_platform.h
+++ b/src/inc/quic_platform.h
@@ -68,30 +68,28 @@ typedef struct QUIC_SINGLE_LIST_ENTRY {
 //
 #ifdef _WIN32
 
-#define POOL_TAG_QUIC 'QUIC'
-#define POOL_TAG_QCon 'QCon'
-#define POOL_TAG_QCTP 'QCTP'
-#define POOL_TAG_QStm 'QStm'
-#define POOL_TAG_QSBf 'QSBf'
-#define POOL_TAG_QSFM 'QSFM'
-#define POOL_TAG_QDta 'QDta'
-#define POOL_TAG_QTst 'QTst'
-#define POOL_TAG_QPrf 'QPrf'
-#define POOL_TAG_QTol 'QTol'
+#define POOL_TAG_QUIC 'CIUQ'
+#define POOL_TAG_QCon 'noCQ'
+#define POOL_TAG_QCTP 'PTCQ'
+#define POOL_TAG_QStm 'mtSQ'
+#define POOL_TAG_QSBf 'fBSQ'
+#define POOL_TAG_QSFM 'MFSQ'
+#define POOL_TAG_QDta 'atDQ'
+#define POOL_TAG_QTst 'tsTQ'
+#define POOL_TAG_QPrf 'frPQ'
+#define POOL_TAG_QTol 'loTQ'
 
-#define POOL_TAG(x) POOL_TAG_##x
-
-#define QUIC_POOL_TAG(a, b, c, d) POOL_TAG(d ## c ## b ## a)
+#define QUIC_POOL_TAG(a, b, c, d) POOL_TAG_ ##a ## b ## c ## d
 
 #else
 
 #define POOL_TAG_QUOTE(x) ((#x)[0])
 
 #define QUIC_POOL_TAG(a, b, c, d) \
-    ((POOL_TAG_QUOTE(a) << 24) + \
-     (POOL_TAG_QUOTE(b) << 16) + \
-     (POOL_TAG_QUOTE(c) << 8) + \
-     POOL_TAG_QUOTE(d))
+    ((POOL_TAG_QUOTE(d) << 24) + \
+     (POOL_TAG_QUOTE(c) << 16) + \
+     (POOL_TAG_QUOTE(b) << 8) + \
+     POOL_TAG_QUOTE(a))
 
 #endif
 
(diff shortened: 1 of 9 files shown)


src/tools/interopserver/InteropServer.cpp
```diff
@@ -187,7 +187,8 @@ HttpRequest::Process()
     }
 
     if (strcmp("/", PathStart) == 0) {
-        PathStart = "/index.html";
+        char index[] = "/index.html";
+        PathStart = index;
```
nibanks, line 191: `PathStart` outlives the scope of `index`. Isn't that a bug?

larseggert: I don't think the memory is released from the stack before the end of the function. But I can move it outside, to make it clearer.

src/inc/quic_platform.h
```diff
@@ -67,16 +67,16 @@ typedef struct QUIC_SINGLE_LIST_ENTRY {
...
-#define QUIC_POOL_CONN      'noCQ'  // QCon - QUIC connection
-#define QUIC_POOL_TP        'PTCQ'  // QCTP - QUIC connection transport parameters
-#define QUIC_POOL_STREAM    'mtSQ'  // QStm - QUIC stream
-#define QUIC_POOL_SBUF      'fBSQ'  // QSBf - QUIC stream buffer
-#define QUIC_POOL_META      'MFSQ'  // QSFM - QUIC sent frame metedata
-#define QUIC_POOL_DATA      'atDQ'  // QDta - QUIC datagram buffer
-#define QUIC_POOL_TEST      'tsTQ'  // QTst - QUIC test code
-#define QUIC_POOL_PERF      'frPQ'  // QPrf - QUIC perf code
-#define QUIC_POOL_TOOL      'loTQ'  // QTol - QUIC tool code
+#define QUIC_POOL_GENERIC   *(uint32_t*)"CIUQ"  // QUIC - Generic QUIC
```
ThadHouse, line 70: These worry me. The generated assembly is not the same, at least without optimization. 

`mov     edi, 1128879441` for the multichar

```
.LC0:
        .string "noCQ"

        mov     eax, OFFSET FLAT:.LC0
        mov     eax, DWORD PTR [rax]
        mov     edi, eax
```

I know that they realistically are the same, but I think I would prefer the warning suppression and identical assembly between debug and release. @nibanks what do you think?

nibanks: I tend to agree.

nibanks: OR, we can have a `QUIC_TAG` macro that abstracts difference between platforms here.

larseggert: You could define them as hex constants instead of strings...

ThadHouse: Linux was where I was seeing the assembly difference. I don't think either one is standard, but I'd prefer the multichar over type punning a string, as its implementation defined instead of potentially being UB.

nibanks: I prefer to keep Windows as it is, as this is the standard mechanism for how Windows (internally) defines pool tags.

larseggert: E.g., 0x74735451 instead of 'tsTQ'

ThadHouse: That'd I'd be more fine with. But keep windows the multichars. I like the macro idea.

larseggert: I implemented the macro idea and left the multichar for Windows (which makes the non-Windows macro quite ugly...)

src/inc/quic_platform.h
```diff
@@ -66,17 +66,44 @@ typedef struct QUIC_SINGLE_LIST_ENTRY {
...
+#define POOL_TAG_QCTP 'PTCQ'
+#define POOL_TAG_QStm 'mtSQ'
+#define POOL_TAG_QSBf 'fBSQ'
+#define POOL_TAG_QSFM 'MFSQ'
+#define POOL_TAG_QDta 'atDQ'
+#define POOL_TAG_QTst 'tsTQ'
+#define POOL_TAG_QPrf 'frPQ'
+#define POOL_TAG_QTol 'loTQ'
+
+#define QUIC_POOL_TAG(a, b, c, d) POOL_TAG_ ##a ## b ## c ## d
```
nibanks, line 82: Not  a super fan of this. Is there no way to concatenate single quoted characters?

larseggert: Nope. See https://gcc.gnu.org/onlinedocs/cpp/Stringizing.html, which says:
> There is no way to convert a macro argument into a character constant.

larseggert: We could also decide to just leave the `-Wno-multichar` in place...

nibanks: Ok.

ThadHouse, line 82: `##a ## b ## c ## d` Is a having a different spacing from the b, c and d expected?

larseggert: There should be a space after the a for consistency, but functionally it doesn't matter.

nibanks: Actually Lars, the more I think about it, let's just ignore the warning and revert all this please.

larseggert: ACK

PR #691, opened 2020-08-06
Refactored Perf Code

nibanks:
Refactoring while I'm working on RPS code. Figured it'd be easier to review this separately.


src/perf/lib/quicmain.cpp
```diff
@@ -37,18 +37,21 @@ QUIC_STATUS
...
     ) {
-    const char* TestName = GetValue(argc, argv, "TestName");
-    if (!TestName) {
-        WriteOutput("Must have a TestName specified. Ex: -TestName:Throughput\n");
+    if (argc < 3 || IsArg(argv[1], "?") || IsArg(argv[1], "help") || !IsArg(argv[2], "TestName")) {
         PrintHelp();
         return QUIC_STATUS_INVALID_PARAMETER;
     }
 
+    const char* TestName = GetValue(argc, argv, "TestName");
```
ThadHouse, line 48: Won't this collide with https://github.com/microsoft/msquic/pull/691/files#diff-a3358465a6163ede94a3c2edac7ee7bcR54

Since this will make argv[0] not be the executable name?

ThadHouse: Whoops, meant this to be on the line below.

src/perf/lib/ThroughputServer.cpp
```diff
@@ -47,25 +45,16 @@ ThroughputServer::Init(
...
-        return Listener.GetInitStatus();
-    }
-
-    uint16_t port = THROUGHPUT_DEFAULT_PORT;
-    TryGetValue(argc, argv, "port", &port);
-
-    const char* localAddress = nullptr;
-    if (!TryGetValue(argc, argv, "listen", &localAddress)) {
-        WriteOutput("Server mode must have -listen\n");
+    if (argc > 0 && (IsArg(argv[1], "?") || IsArg(argv[1], "help"))) {
```
ThadHouse, line 48: This check would need to be `argc > 1` to access index 1. But I think the access should be index 0, see my comment below.

src/perf/lib/quicmain.cpp
nibanks: No, the goal here is to support the following:
```
quicping -help
```
and
```
quicping -TestName:Throughput -help
```
They each hit different `PrintHelp()` functions.

ThadHouse: Ok. Then my other comment will definitely need to be fixed then. There were 2 places it happened too.

src/perf/lib/ThroughputServer.cpp
nibanks: Ah, yes. It should be `argv[0]`.

PR #692, opened 2020-08-06
Recreate perf database to allow running perf tests with changing arguments

ThadHouse:
Allows a matrix like setup to compare specific flags.

WIP, the scripts still need to be updated partially.


ThadHouse:
This is good to go now. Once its done, I'll merge it into the RPS PR and get that updated as well.

PR #693, opened 2020-08-06
Switch all C++ new in kernel mode to nothrow new

ThadHouse:
By default, even in kernel mode, new will immediately call the constructor, before you even have a chance to check for null. nothrow new checks for null before calling the constructor, so the checks are actually valid.

Only needed to happen in the perf and test libs that are ran in kernel mode.

We will need to be sure this happens for all kernel mode allocations in the future.


ThadHouse:
The failure isn't related. Its one of the random tests thats been failing often.

PR #694, opened 2020-08-06
Initial RPS Perf Code

nibanks:


src/perf/lib/ThroughputClient.cpp
```diff
@@ -47,6 +47,8 @@ ThroughputClient::ThroughputClient(
     QuicZeroMemory(&LocalIpAddr, sizeof(LocalIpAddr));
     if (Session.IsValid()) {
         Session.SetAutoCleanup();
+        Session.SetDisconnectTimeout(RPS_DEFAULT_DISCONNECT_TIMEOUT);
```
ThadHouse, line 50: We should either have a separate timeout for throughput, or not scope this.

nibanks: I don't understand?

ThadHouse:
Once https://github.com/microsoft/msquic/pull/693 comes in, we'll want to do a similar thing here. We should also mark all the c++ functions here as noexcept. The compiler can generate more performant code if that happens.

src/perf/lib/ThroughputClient.cpp
nibanks: Oh, this was put in the wrong file. I will fix.

ThadHouse: Since this is in Throughput, but is using an RPS default. Either just change RPS_DEFAULT_DISCONNECT_TIMEOUT to PERF_DEFAULT_DISCONNECT_TIMEOUT, or make a separate one just for throughput. Would probably prefer PERF_DEFAULT though.

ThadHouse: Or that works too 👍

PR #695, opened 2020-08-07
Fix Schannel TLS alerts not being read correctly.

### LINKED ISSUE #255: Schannel Alert Code Doesn't Work
### Describe the bug

The code to get the TLS alert from schannel when there is an error doesn't seem to work. It is always zero.

The easiest way to repro is to set up a server with a self-signed (or otherwise invalid) certificate and then use the schannel client to connect. You'll get output that looks like this:

```
[Microsoft-Quic][ tls][0xFFFFCF06DE2AF010] ERROR, 2148074277, Accept/InitializeSecurityContext.
[Microsoft-Quic][conn][0xFFFFCF06DE2AF010] ERROR, 0, Received alert from TLS.
```

We should get a non-zero alert code from schannel.

anrossi:
The TLS alert buffer was being provided to Schannel incorrectly, causing the alert to always be zero.
A new TLS test was introduced to the platform unittests to cover this as well.
Adds a script to make it easier to run quicping.exe for repro and debugging purposes.

Fixes #255

### DIFF
--- a/.azure/azure-pipelines.ci.yml
+++ b/.azure/azure-pipelines.ci.yml
@@ -217,17 +217,19 @@ stages:
       platform: windows
       tls: mitls
       logProfile: Full.Light
-      extraArgs: -Filter -*Unreachable/0:CryptTest/CryptTest.Encryption/2
+      extraArgs: -Filter -*Unreachable/0:CryptTest/CryptTest.Encryption/2:TlsTest.CertificateError
   - template: ./templates/run-bvt.yml
     parameters:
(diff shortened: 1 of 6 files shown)


nibanks:
Looks great! Thanks! But every other TLS implementation fails the test. We need to get some solution to that, even if it's explicitly skipping this test in the CI or not compiling it in for these.

anrossi:
@nibanks I was hoping at least mitls would pass. That's disappointing.  Could add a filter so that this test only runs on Schannel BVTs?

PR #699, opened 2020-08-08
Cleanup Logging Script Args

nibanks:


scripts/run-executable.ps1
```diff
@@ -24,9 +24,6 @@ This script runs an executable and collects and logs or process dumps as necessa
 .PARAMETER LogProfile
```
ThadHouse, line 24: We should have this parameter name match between here and the log script.

nibanks: I explicitly didn't want them to be the same. A `-Profile` flag here be too generic.

PR #702, opened 2020-08-10
Add sidechannel to perf runner to enable stopping the tests safely internally

ThadHouse:
Only currently implemented in user mode, kernel mode will be different but done when kernel mode is finished


src/perf/bin/appmain.cpp
```diff
@@ -41,7 +45,24 @@ QuicUserMain(
...
-    QUIC_STATUS Status = QuicMainStart(argc, argv, &StopEvent, SelfSignedConfig);
+    uint8_t ServerMode = 0;
+    TryGetValue(argc, argv, "ServerMode", &ServerMode);
+
+    QUIC_STATUS Status;
+    QUIC_DATAPATH* Datapath;
+    QUIC_DATAPATH_BINDING* Binding;
+
+    if (ServerMode) {
+        QuicDataPathInitialize(256, DatapathReceiveUserMode, DatapathUnreachable, &Datapath);
```
nibanks, line 56: You should just be able to pass 0 instead of 256. You don't need any callback context.

src/perf/bin/appmain.cpp
```diff
@@ -41,7 +45,24 @@ QuicUserMain(
...
+        QUIC_ADDR LocalAddress;
+        QuicZeroMemory(&LocalAddress, sizeof(LocalAddress));
+        QuicAddrSetPort(&LocalAddress, 9999);
+        QuicAddrSetFamily(&LocalAddress, AF_INET);
```
nibanks, line 60: No need to zero init since you're setting port and family. BTW, set family to AF_UNSPEC. No need to hard code to V4. Never know when we want to use V6 instead.

src/perf/bin/appmain.cpp
```diff
@@ -41,7 +45,24 @@ QuicUserMain(
...
+
+        QuicDataPathBindingCreate(Datapath, &LocalAddress, nullptr, &StopEvent, &Binding);
```
nibanks, line 62: Can you check for errors please?

PR #705, opened 2020-08-10
Fix compilation with VS 16.7

### LINKED ISSUE #704: VS2019 16.7.0 build failure
I encounter new problem building msquic.kernel.sln possibly after I upgrade VS2019 to 16.7.0.

msquic\src\platform\datapath_winkernel.c(2202): error C6011: Dereferencing NULL pointer 'Binding'. 
msquic\src\platform\hashtable.c(731): error C6387: 'ContextPtr->ChainHead' could be '0':  this does not adhere to the specification for the function 'QuicListIsEmpty'.

ThadHouse:
Closes #704

Found 1 actual use after free bug, 2 potential things we probably should be debug asserting, and then a few uninitialized variables that very much seem like false positives.

### DIFF
--- a/src/core/lookup.c
+++ b/src/core/lookup.c
@@ -55,7 +55,10 @@ QuicLookupUninitialize(
         for (uint8_t i = 0; i < Lookup->PartitionCount; i++) {
             QUIC_PARTITIONED_HASHTABLE* Table = &Lookup->HASH.Tables[i];
             QUIC_DBG_ASSERT(Table->Table.NumEntries == 0);
+#pragma warning(push)
+#pragma warning(disable:6001)
             QuicHashtableUninitialize(&Table->Table);
+#pragma warning(pop)
             QuicDispatchRwLockUninitialize(&Table->RwLock);
         }
         QUIC_FREE(Lookup->HASH.Tables);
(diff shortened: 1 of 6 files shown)


src/core/lookup.c
```diff
@@ -55,7 +55,10 @@ QuicLookupUninitialize(
         for (uint8_t i = 0; i < Lookup->PartitionCount; i++) {
             QUIC_PARTITIONED_HASHTABLE* Table = &Lookup->HASH.Tables[i];
             QUIC_DBG_ASSERT(Table->Table.NumEntries == 0);
+#pragma warning(push)
+#pragma warning(disable:6001)
```
nibanks, line 59: Instead of pragma ignores, I'd prefer DBG asserts. What is the error for?

src/platform/datapath_winkernel.c
```diff
@@ -2195,10 +2195,11 @@ QuicDataPathBindingReturnRecvDatagrams(
         QuicPoolFree(BatchedInternalContext->OwningPool, BatchedInternalContext);
     }
 
-    if (DataIndications != NULL) {
+    if (DataIndications != NULL && Binding != NULL) {
         //
         // Return the datagram indications back to Wsk.
         //
+        QUIC_DBG_ASSERT(Binding != NULL);
```
nibanks, line 2202: Assert is unnecessary with the check above.

nibanks: Actually, remove the if check instead. Keep the assert.

src/core/lookup.c
ThadHouse: Uninitialized memory. So can't really assert it.

src/platform/datapath_winkernel.c
ThadHouse: Ah thought I had done that. Whoops. Easy fix.

src/core/lookup.c
ThadHouse: It's assuming that Directory in the union in QUIC_HASHTABLE  is uninitialized.

PR #706, opened 2020-08-11
Add siduck handler to interop server

nibanks:
Integrates the server side of the siduck protocol into quicinterop: https://tools.ietf.org/html/draft-pardue-quic-siduck-00

### DIFF
--- a/src/tools/interopserver/InteropServer.h
+++ b/src/tools/interopserver/InteropServer.h
@@ -12,6 +12,9 @@
 extern const QUIC_API_TABLE* MsQuic;
 extern QUIC_SEC_CONFIG* SecurityConfig;
 
+const QUIC_BUFFER QuackBuffer = { sizeof("quack") - 1, (uint8_t*)"quack" };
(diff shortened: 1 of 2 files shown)


src/tools/interopserver/InteropServer.h
```diff
@@ -196,7 +253,12 @@ struct HttpServer {
         ) {
         if (Event->Type == QUIC_LISTENER_EVENT_NEW_CONNECTION) {
             Event->NEW_CONNECTION.SecurityConfig = SecurityConfig;
-            new HttpConnection(Event->NEW_CONNECTION.Connection);
+            if (Event->NEW_CONNECTION.Info->NegotiatedAlpnLength >= 6 &&
+                !memcmp(Event->NEW_CONNECTION.Info->NegotiatedAlpn, "siduck", 6)) {
```
ThadHouse, line 257: Can you not use strcmp here? Are they not null terminated?

nibanks: Nope. ALPN is **not** null terminated. That's why the length is also provided.

PR #709, opened 2020-08-11
Fix schannel resumption tests to run faster.

### LINKED ISSUE #571: Schannel Handshake Tests Resume=1 Take 2 Extra Seconds
### Describe the bug

I was just manually running the latest schannel tests and noticed all the Resume=1 test cases seem to take an extra 2 seconds. They shouldn't need all that time. I suspect a bug somewhere in there.

anrossi:
The resumption tests on Schannel take 2 seconds longer than the other TLS platforms because there's no clear indication when the resumption ticket has been received, so it retries until timeout.
This fixes that retry loop by setting a boolean when the test can be reasonably certain the resumption ticket has arrived.
Fixes #571

### DIFF
--- a/src/platform/tls_schannel.c
+++ b/src/platform/tls_schannel.c
@@ -335,6 +335,7 @@ typedef struct QUIC_TLS {
     BOOLEAN PeerTransportParamsReceived : 1;
     BOOLEAN HandshakeKeyRead : 1;
     BOOLEAN ApplicationKeyRead : 1;
+    BOOLEAN TicketReceived : 1;
 
     QUIC_TLS_SESSION* TlsSession;
 
@@ -2222,6 +2223,17 @@ QuicTlsProcessData(
         goto Error;
     }
 
+    if (!TlsContext->IsServer && State->BufferOffset1Rtt > 0 &&
+        State->ReadKeys[QUIC_PACKET_KEY_HANDSHAKE] == NULL) {
+        //
+        // Schannel currently sends the NST with server_finished.
+        // We need to wait for the handshake to be confirmed before
+        // setting the flag, since we don't know if we're received
+        // the ticket yet.
+        //
+        TlsContext->TicketReceived = TRUE;
+    }
+
     QuicTraceLogConnVerbose(
         SchannelProcessingData,
         TlsContext->Connection,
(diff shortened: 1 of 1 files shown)


src/platform/tls_schannel.c
```diff
@@ -2222,6 +2223,17 @@ QuicTlsProcessData(
         goto Error;
     }
 
+    if (!TlsContext->IsServer && State->BufferOffset1Rtt > 0 &&
+        State->ReadKeys[QUIC_PACKET_KEY_HANDSHAKE] == NULL) {
```
nibanks, line 2227: Why are you checking the ReadKey?

anrossi: Ideally, I'd like to read `Connection->State.HandshakeConfirmed` but that's not exposed to the TLS layer.
So I checked what changes in the TLS state when `QuicCryptoOnHandshakeConfirmed` is called, and the ReadKeys for Handshake are set to NULL.

src/platform/tls_schannel.c
```diff
@@ -2222,6 +2223,17 @@ QuicTlsProcessData(
...
+        //
+        // Schannel currently sends the NST with server_finished.
+        // We need to wait for the handshake to be confirmed before
+        // setting the flag, since we don't know if we're received
```
anrossi, line 2231: Grammar better.

src/platform/tls_schannel.c
nibanks: miTLS code doesn't wait for handshake confirmed and it passes. Why are you trying to impose that restriction here?

anrossi: miTLS also gets a notification when the ticket is received. Since Schannel doesn't provide that interface we have to guess at a milestone where we can be certain the ticket has been received.

nibanks: At least for now, the only 1-RTT crypto data exchanged is NST. So when a client receives 1-RTT crypto data is MUST be a ticket. Also, miTLS always indicates inline when we get the ticket. So, removing this additional check is equivalent to what miTLS does already. When we do get an indication from Schannel I expect it to be in response to a call to them and also be equivalent to this.

anrossi: When we get notifications from Schannel this check will go away and things will work appropriately. This is just a temporary workaround.

Issue #606, opened 2020-07-16
Support Sanitizers


Maximus-:
### Describe the feature you'd like supported
- [ ] Add support for Thread Sanitizer (TSAN) build configuration on the Linux platform (It currently does not support Windows in any capacity).
- [x] Add support for Windows ASAN 64bit https://devblogs.microsoft.com/cppblog/asan-for-windows-x64-and-debug-build-support/ (only available in VS 2019 Preview 16.7 at the moment)

### Proposed solution
Thread Sanitizer helps identify race conditions while in testing. The required change is adding -fsanitize=thread instead of -fsanitize=address. 

### Additional context
There will certainly be a bit of noise produced here. Last time I ran TSAN against MsQuic, a few things popped up that were "best efforts", so the race wasn't important and couldn't cause any memory corruption or otherwise harmful behavior. Ideally you can tag some of these false positives with `__attribute__((no_sanitize("thread"))` (this is per-function unfortunately...).

Reminder as well, that any errors found with TSAN *will not cause the program to terminate*. It will just output to stderr. 

More info here - https://github.com/google/sanitizers/wiki/ThreadSanitizerCppManual

nibanks:
@maximus- any hope for a PR from you for these?

Maximus-:
I'm happy to do it, but it may not happen until next week :) Although the change should be tiny; All it needs is /fsanitize=address passed to `cl`, I believe MSVC should handle the linkage to the ASAN runtime libraries automatically.

nibanks:
@maximus- is TSAN still important? Or should we close this bug now that ASAN is integrated on Windows?

Maximus-:
> @Maximus- is TSAN still important? Or should we close this bug now that ASAN is integrated on Windows?

MsQuic is a high performance, multi-threaded networking stack. I definitely think TSAN is worth investing in. Might need some annotations around the codebase to ignore or silence the false positives, but otherwise it's basically free race condition detection.

nibanks:
@maximus- so what work is left for TSAN support integration?

Maximus-:
To pass `-fsanitize=thread` to the compiler in your Linux configuration. It's not compatible with ASAN, so needs to be a mutually exclusive build option.

nibanks:
I'm going to close this with just ASAN support for now. We can investigate possible tsan support in the future.

PR #712, opened 2020-08-12
More sanitizers

larseggert:
And fixes for some things they found.


src/core/packet.h
```diff
@@ -445,7 +445,8 @@ QuicPacketEncodeLongHeaderV1(
     }
     *PayloadLengthOffset = (uint16_t)(HeaderBuffer - Buffer);
     HeaderBuffer += sizeof(uint16_t); // Skip PayloadLength.
-    *(uint32_t*)HeaderBuffer = QuicByteSwapUint32(PacketNumber);
+    const uint32_t tmp = QuicByteSwapUint32(PacketNumber);
+    memcpy(HeaderBuffer, &tmp, sizeof(tmp));
```
nibanks, line 449:
```suggestion
    PacketNumber = QuicByteSwapUint32(PacketNumber);
    memcpy(HeaderBuffer, &PacketNumber, sizeof(PacketNumber));
```

src/inc/quic_hashtable.h
```diff
@@ -246,6 +246,9 @@ QuicHashTableContract(
 // Simple helper hash function.
 //
 inline
+#if defined(__clang__)
+    __attribute__((no_sanitize("unsigned-integer-overflow")))
+#endif
```
nibanks, line 251: I don't like #ifdef (some platform specific thing) in common code. I prefer some platform abstraction layer, that is no-op on platforms it doesn't apply to. For instance:
```C
#if defined(__clang__)
#define QUIC_NO_SANITIZE(X) __attribute__((no_sanitize(X)))
#else
#define QUIC_NO_SANITIZE(X) 
#endif
```

src/inc/quic_var_int.h
```diff
@@ -65,13 +65,16 @@ QuicVarIntEncode(
         Buffer[0] = (uint8_t)Value;
         return Buffer + sizeof(uint8_t);
     } else if (Value < 0x4000) {
-        *(uint16_t*)Buffer = QuicByteSwapUint16((0x40 << 8) | (uint16_t)Value);
```
nibanks, line 68: I'm not a big fan of using temporary variables... I'm wondering if just writing out each individual byte assignment might be better.

src/inc/quic_hashtable.h
larseggert: The problem is that "X" is compiler-specific.

src/inc/quic_var_int.h
larseggert: I checked for quant, and this gets optimized away completely.

nibanks: Ok.

src/core/range.c
```diff
@@ -259,6 +259,7 @@ QuicRangeAddRange(
     *RangeUpdated = FALSE;
 
 #if QUIC_RANGE_USE_BINARY_SEARCH
+    // FIXME: runtime error: unsigned integer overflow: 0 - 1 cannot be represented in type 'unsigned int'
```
nibanks, line 262: This code relied on underflow. I think we can just annotate/ignore this function.

src/inc/quic_hashtable.h
larseggert: Which file should the `QUIC_NO_SANITIZE` define go into?

larseggert:
Instead of using `tmp`, another option would be to overwrite `Value`?

src/inc/quic_var_int.h
larseggert: Other option would be to overwrite `Value`?

nibanks:
> Instead of using `tmp`, another option would be to overwrite `Value`?

Well, I don't think it matters to much either way. The problem with `Value` is that it's a `uint64_t` and not the type specific you have for `tmp` already.

larseggert:
Well, I would `memcpy` the relevant range out of the `uint64_t`, but I'm not convinced that is cleaner.

ThadHouse:
Yeah I'm not a fan of that. Gets especially weird with endianness.

src/core/library.h
```diff
@@ -343,8 +343,10 @@ QuicCidNewRandomSource(
         *Data = PartitionID;
         Data++;
 
-        QuicCopyMemory(Data, Prefix, PrefixLength);
-        Data += PrefixLength;
+        if (PrefixLength) {
+            QuicCopyMemory(Data, Prefix, PrefixLength);
+            Data += PrefixLength;
+        }
```
nibanks, line 349: So, does clang not like passing `NULL, 0` to `memcpy`? `memcpy` is practically a `for` loop and if length is zero it would never read from the `NULL` being passed in. This effectively just adds another (irrelevant) NULL check IMO.

src/inc/quic_hashtable.h
nibanks: It would go in each individual platform header.

src/core/library.h
ThadHouse: Passing null to memcpy is UB, even if 0 length is passed. Standard 7.1.4

larseggert:
The CI fails are due to a new `FIXME` assert in `src/core/stream_send.c` that - I think - is an actual bug.

larseggert:
And also: `There is not enough space on the disk. : 'C:\a\_diag\pages\e303fa06-3494-451b-9145-6fe729f28706_1.log'`

src/core/library.c
```diff
@@ -1352,7 +1352,8 @@ QuicLibraryGetWorker(
     QUIC_DBG_ASSERT(MsQuicLib.WorkerPool != NULL);
     return
         &MsQuicLib.WorkerPool->Workers[
-            MsQuicLib.NextWorkerIndex++ % MsQuicLib.WorkerPool->WorkerCount];
```
nibanks, line 1355: Why? Just let this roll over?

src/inc/quic_versions.h
```diff
@@ -14,28 +14,28 @@
 //
 // The QUIC version numbers, in network byte order.
 //
-#define QUIC_VERSION_VER_NEG    0x00000000      // Version for 'Version Negotiation'
-#define QUIC_VERSION_1          0x01000000      // First official version
-#define QUIC_VERSION_MS_1       0x0000cdab      // First Microsoft version (currently same as latest draft)
-#define QUIC_VERSION_DRAFT_27   0x1b0000ff      // IETF draft 27
-#define QUIC_VERSION_DRAFT_28   0x1c0000ff      // IETF draft 28
-#define QUIC_VERSION_DRAFT_29   0x1d0000ff      // IETF draft 29
+#define QUIC_VERSION_VER_NEG    0x00000000U      // Version for 'Version Negotiation'
```
nibanks, line 17: nit: can you fix the comments back to where they were?

src/core/library.c
larseggert: Sanitizer warns when the `uint8_t` exceeds 255. Alternative is to suppress that warning?

src/core/stream_send.c
```diff
@@ -400,6 +400,8 @@ QuicStreamCompleteSendRequest(
             SendRequest->InternalBuffer.Length);
     }
 
+    // FIXME: this underflows when running spinquic
+    QUIC_DBG_ASSERT(Stream->Connection->SendBuffer.PostedBytes >= SendRequest->TotalLength);
```
nibanks, line 404: This is interesting. Let me look into this some... Can you open an issue for this?

src/core/library.c
nibanks: I think that'd be my preference in this particular case.

src/inc/quic_versions.h
larseggert: You mean the indentation?

nibanks: Yes. Just delete a space before it please.

src/core/stream_send.c
larseggert: Done: #719

larseggert:
@nibanks any idea why the CI takes forever? Azure datacenter on fire?

larseggert:
And of course it immediately completes as soon as I write this. I guess the ML detected that I am complaining about Azure and increased the priority of the run.

nibanks:
> And of course it immediately completes as soon as I write this. I guess the ML detected that I am complaining about Azure and increased the priority of the run.

Of course!

CMakeLists.txt
```diff
@@ -103,7 +103,7 @@ option(QUIC_BUILD_TOOLS "Builds the tools code" ON)
 option(QUIC_BUILD_TEST "Builds the test code" ON)
 option(QUIC_BUILD_PERF "Builds the perf code" ON)
 option(QUIC_ENABLE_LOGGING "Enables logging" ON)
-option(QUIC_SANITIZE_ADDRESS "Enables address sanitizer" OFF)
+option(QUIC_ENABLE_SANITIZERS "Enables sanitizers" OFF)
```
Maximus-, line 106: This conflicts a bit with #606, which I should eventually add :)

Maybe QUIC_ENABLE_SANITIZER=thread, QUIC_ENABLE_SANITIZER=address?

larseggert: I'm in favor of merging now and letting you change things once #606 is ready :-)

PR #714, opened 2020-08-12
Onboard clang into build scripts and CI

### LINKED ISSUE #696: Onboard a Linux Clang build to CI
Clang is becoming more and more popular as a compiler on linux, and its a bit more particular then GCC. We should onboard a CI build using clang to ensure the build doesn't break there for external users.

ThadHouse:
Closes #696

### DIFF
--- a/.azure/azure-pipelines.ci.yml
+++ b/.azure/azure-pipelines.ci.yml
@@ -128,19 +128,25 @@ stages:
   - template: ./templates/build-config-user.yml
     parameters:
       image: ubuntu-latest
-      platform: linux
-      arch: x64
-      tls: openssl
-      extraBuildArgs: -Clang
-      skipArtifacts: 'true'
-  - template: ./templates/build-config-user.yml
-    parameters:
-      image: ubuntu-latest
       container: raspbian
       platform: linux
       arch: arm
       tls: openssl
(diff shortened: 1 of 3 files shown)


.azure/templates/build-config-user.yml
```diff
@@ -8,6 +8,7 @@ parameters:
   tls: ''
   config: 'Debug,Release'
   extraBuildArgs: ''
+  skipArtifacts: ''
```
nibanks, line 11: Default to `false` instead?

.azure/azure-pipelines.ci.yml
```diff
@@ -134,6 +134,20 @@ stages:
       tls: openssl
       extraBuildArgs: -DisableLogs -Generator Ninja -ToolchainFile cmake/toolchains/arm-pi-gnueabihf.toolchain.cmake
 
+# Unsupported currently.
+- stage: build_linux_clang
```
nibanks, line 138: I'd prefer to not make a separate stage...

ThadHouse: Then I have to modify the build-config-user.yml to have an extra string I can use to modify the job name. I can do that if you think that'd be cleaner.

nibanks: Yeah, I think I prefer that way.

Issue #698, opened 2020-08-08
Assert in QuicCryptoInitialize on Server Connection


nibanks:
### Describe the bug

The most recent Windows (schannel) spinquic hit this assert:

https://dev.azure.com/ms/msquic/_build/results?buildId=100196&view=results

```
0:013> kn
 # Child-SP          RetAddr           Call Site
00 000000c1`3bffe498 00007ffb`f201930c ntdll!ZwWaitForMultipleObjects+0x14
01 000000c1`3bffe4a0 00007ffb`f201920e KERNELBASE!WaitForMultipleObjectsEx+0xec
02 000000c1`3bffe790 00007ffb`f3ffe5ac KERNELBASE!WaitForMultipleObjects+0xe
03 000000c1`3bffe7d0 00007ffb`f3ffe01e kernel32!WerpLaunchAeDebug+0x34cc
04 000000c1`3bffe8f0 00007ffb`f20f508a kernel32!WerpLaunchAeDebug+0x2f3e
05 000000c1`3bffe930 00007ffb`f485bc8d KERNELBASE!UnhandledExceptionFilter+0x3da
06 000000c1`3bffea50 00007ffb`f4843176 ntdll!memset+0x18cd
07 000000c1`3bffea90 00007ffb`f485869f ntdll!_C_specific_handler+0x96
08 000000c1`3bffeb00 00007ffb`f47e836c ntdll!_chkstk+0x12f
09 000000c1`3bffeb30 00007ffb`f485719e ntdll!RtlRaiseException+0x3cc
0a 000000c1`3bfff240 00007ffb`d12ac00e ntdll!KiUserExceptionDispatcher+0x2e
0b 000000c1`3bfff950 00007ffb`d1271d46 msquic!QuicCryptoInitialize+0x27e [D:\a\1\msquic\src\core\crypto.c @ 156] 
0c 000000c1`3bfffa20 00007ffb`d126b4d7 msquic!QuicConnInitializeCrypto+0x36 [D:\a\1\msquic\src\core\connection.c @ 2155] 
0d 000000c1`3bfffa60 00007ffb`d12675df msquic!QuicConnDrainOperations+0x127 [D:\a\1\msquic\src\core\connection.c @ 6550] 
0e 000000c1`3bfffb80 00007ffb`d1266307 msquic!QuicWorkerProcessConnection+0x24f [D:\a\1\msquic\src\core\worker.c @ 482] 
[...]

PR #715, opened 2020-08-12
Add assertions to enqueue connection operation to help solve 698

ThadHouse:
We haven't seen it again, but when it does happen, having the assert here will make it much easier to detect the source. When it happens on the other side of the queue, its hard to tell what the source actually is. Related to #698.

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -665,14 +665,13 @@ QuicConnQueueOper(
     _In_ QUIC_OPERATION* Oper
     )
 {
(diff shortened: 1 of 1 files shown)


src/core/connection.c
```diff
@@ -666,6 +666,13 @@ QuicConnQueueOper(
     )
 {
     if (QuicOperationEnqueue(&Connection->OperQ, Oper)) {
+#if DEBUG
```
nibanks, line 669: I think it's better to put this **before** the enqueue call.

PR #721, opened 2020-08-13
fix decoder to compile

chgray:
unclear what broke in the updated one, this just goes back to the previous version while we sort it out


ThadHouse:
This won't fully work, as the names being used had changed in the QUIC code. What's the compilation error? I'll take a look when I get back home.

We also need to add some sort of way to test that compile. Especially since it happens at runtime.

nibanks:
> We also need to add some sort of way to test that compile. Especially since it happens at runtime.

Yeah, when clog runs at build time, it should try to compile the code and spit out any errors.

ThadHouse:
This will be replaced by https://github.com/microsoft/CLOG/pull/58 and https://github.com/microsoft/CLOG/pull/59 which correctly solve the issue.

PR #722, opened 2020-08-13
Start on implementing kernel mode perf

ThadHouse:


ThadHouse:
PR CI Perf is not enabled yet, but I want to ensure it's not going to bring down the perf systems before adding that. So currently only the Perf Pipeline will run kernel mode perf. But the code should be basically complete.

.azure/azure-pipelines.perf.yml
```diff
@@ -18,6 +18,7 @@ parameters:
   - PGO
   - Record
   - RecordQUICTrace
+  - Kernel
```
nibanks, line 21: I don't want to have to pick between kernel and everything else. Ideally it should either always be included to a separate parameter.

.azure/templates/build-config-winkernel.yml
```diff
@@ -2,6 +2,7 @@
 
 parameters:
   arch: ''
+  skipDebug: false
```
nibanks, line 5: Do this the same way usemode does `config: 'Debug,Release'`

src/perf/bin/appmain.cpp
```diff
@@ -149,41 +121,72 @@ QuicKernelMain(
...
         QUIC_FREE(Data);
         QUIC_FREE(OutBuffer);
+
+        RunSuccess =
+        DriverClient.Read(
+            IOCTL_QUIC_READ_DATA,
+            OutBuffer,
+            OutBufferSize,
+            &OutBufferWritten,
+            180000);
```
nibanks, line 158: nit: tab over

src/perf/bin/drvmain.cpp
```diff
@@ -490,18 +502,136 @@ QuicPerfCtlEvtIoCanceled(
     WdfRequestComplete(Request, Status);
 }
 
+NTSTATUS
+QuicPerfCtlSetSecurityConfig(
+    _Inout_ QUIC_DRIVER_CLIENT* Client,
+    _In_ const QUIC_CERTIFICATE_HASH* CertHash
+)
```
nibanks, line 509: nit: tab over

src/perf/bin/drvmain.cpp
```diff
@@ -490,18 +502,136 @@ QuicPerfCtlEvtIoCanceled(
...
+        char Data;
+    };
+    QUIC_CERTIFICATE_HASH CertHash;
+} QUIC_IOCTL_PARAMS;
+
 static_assert(
     QUIC_PERF_MAX_IOCTL_FUNC_CODE + 1 == (sizeof(QUIC_IOCTL_BUFFER_SIZES) / sizeof(size_t)),
     "QUIC_IOCTL_BUFFER_SIZES must be kept in sync with the IOTCLs");
 
+QUIC_THREAD_CALLBACK(WaitThreadCb, Context)
```
nibanks, line 536: I don't understand what's going on in this thread from it's name. Can you improve it a bit?

src/bin/winkernel/msquic.kernel.vcxproj
```diff
@@ -119,7 +119,7 @@
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
     <Link>
-      <AdditionalOptions>%(AdditionalOptions) /FORCE:PGOREPRO /USEPROFILE:PGD=$(SolutionDir)src\bin\winkernel\pgo_x64\msquic.pgd</AdditionalOptions>
+      <AdditionalOptions>%(AdditionalOptions) /FORCE:PGOREPRO</AdditionalOptions>
```
nibanks, line 122: Why?

src/inc/quic_driver_helpers.h
```diff
@@ -82,7 +82,7 @@ class QuicDriverService {
                     DriverFilePath,
                     nullptr,
                     nullptr,
-                    "msquic\0",
+                    "msquicpriv\0",
```
nibanks, line 85: This is going to break Windows tests as they depend on the real msquic.

src/perf/lib/ThroughputClient.cpp
```diff
@@ -377,13 +378,17 @@ ThroughputClient::StreamCallback(
...
         uint32_t SendRate = (uint32_t)((StrmData->BytesCompleted * 1000 * 1000 * 8) / (1000 * ElapsedMicroseconds));
 
-        WriteOutput("[%p][%llu] Closed [%s] after %u.%u ms. (TX %llu bytes @ %u kbps).\n",
-            StrmData->Connection,
-            (unsigned long long)GetStreamID(MsQuic, StreamHandle),
-            "Complete",
-            (uint32_t)(ElapsedMicroseconds / 1000),
-            (uint32_t)(ElapsedMicroseconds % 1000),
-            (unsigned long long)StrmData->BytesCompleted, SendRate);
+        if (StrmData->BytesCompleted != 0) {
```
nibanks, line 381: It's an acceptable scenario to open and close a stream with no data. Not necessarily a failure.

src/test/MsQuicTests.h
```diff
@@ -300,9 +300,9 @@ LogTestFailure(
 //
 
 //
-// Name of the driver service for msquictest.sys.
+// Name of the driver service for msquictestpriv.sys.
 //
-#define QUIC_DRIVER_NAME   "msquictest"
+#define QUIC_DRIVER_NAME   "msquictestpriv"
```
nibanks, line 305: Also break Windows tests.

src/bin/winkernel/msquic.kernel.vcxproj
ThadHouse: Because otherwise this would fail to build on my local system, saying the pgd file didn't exist.

src/inc/quic_driver_helpers.h
ThadHouse: Hmm. Thats an interesting issue. We'll have to have a separate user mode executable too. I can do that easily enough.

src/perf/lib/ThroughputClient.cpp
ThadHouse: Even for throughput? I'd consider that an error. If not, I'll need to find another way to detect failure. Especially since I wouldn't be able to use Shutdown initiated by transport.

PR #723, opened 2020-08-14
Create initial performance counters.

anrossi:
Create initial performance counters for MsQuic and expose them via a global `GetParam()`.


src/core/inline.c
```diff
@@ -171,6 +171,14 @@ QuicPartitionIndexDecrement(
     uint8_t Decrement
     );
 
+_IRQL_requires_max_(DISPATCH_LEVEL)
+inline
```
nibanks, line 175: Not needed in the actual inline file

src/core/library.c
```diff
@@ -663,6 +663,54 @@ QuicLibraryGetGlobalParam(
...
+        uint64_t MergedPerfCounters[QUIC_PERF_COUNTER_MAX];
+        memcpy(MergedPerfCounters, MsQuicLib.PerProc[0].PerfCounters, sizeof(MergedPerfCounters));
+
+        for (int ProcIndex = 1; ProcIndex < MsQuicLib.PartitionCount; ++ProcIndex) {
+            for (int CounterIndex = 0; CounterIndex < QUIC_PERF_COUNTER_MAX; ++CounterIndex) {
+                switch (CounterIndex) {
+                    case ConnectionsActive:
+                        MergedPerfCounters[CounterIndex] =
+                            ((int64_t)MergedPerfCounters[CounterIndex]) +
+                            (int64_t)MsQuicLib.PerProc[ProcIndex].PerfCounters[CounterIndex];
```
nibanks, line 688: I think it's just best to do this regardless... actually can't you just add them ass uint64_t and the wrap around addition works out?

src/core/library.c
```diff
@@ -663,6 +663,54 @@ QuicLibraryGetGlobalParam(
...
+                        break;
+                    default:
+                        MergedPerfCounters[CounterIndex] +=
+                            MsQuicLib.PerProc[ProcIndex].PerfCounters[CounterIndex];
+                        break;
+                }
+            }
+        }
+
+        QUIC_DBG_ASSERT((int64_t)MergedPerfCounters[ConnectionsActive] >= 0);
```
nibanks, line 698: Dont' assert this. It's theoretically possible, depending on timings that a value could be less than zero. Instead, if it is less than zero, set it to zero.

src/core/library.c
```diff
@@ -663,6 +663,54 @@ QuicLibraryGetGlobalParam(
...
+
+        if (Buffer == NULL) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+
+        uint64_t MergedPerfCounters[QUIC_PERF_COUNTER_MAX];
+        memcpy(MergedPerfCounters, MsQuicLib.PerProc[0].PerfCounters, sizeof(MergedPerfCounters));
+
+        for (int ProcIndex = 1; ProcIndex < MsQuicLib.PartitionCount; ++ProcIndex) {
```
nibanks, line 682: Talked on IM, this should be per-proc, not per-partition.

src/core/library.c
```diff
@@ -663,6 +663,54 @@ QuicLibraryGetGlobalParam(
...
+            Status = QUIC_STATUS_BUFFER_TOO_SMALL;
+            break;
+        }
+
+        if (Buffer == NULL) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+
+        uint64_t MergedPerfCounters[QUIC_PERF_COUNTER_MAX];
```
nibanks, line 679: Don't use the temporary variable. Use the output buffer directly. Then, instead of looping over `QUIC_PERF_COUNTER_MAX` counters when summing, loop over `*BufferLength / sizeof(uint64_t)`'s worth. Then you don't need to copy over at the end.

src/core/library.h
```diff
@@ -299,6 +304,19 @@ QuicPartitionIndexDecrement(
     }
 }
 
+_IRQL_requires_max_(DISPATCH_LEVEL)
+inline
+void
+QuicPerfCounterAdd(
```
nibanks, line 310: Might be worth having a `#define` for a `QuicPerfCounterIncrement` and `QuicPerfCounterDecrement` but I leave it up to you.

src/core/loss_detection.c
```diff
@@ -766,6 +766,10 @@ QuicLossDetectionRetransmitFrames(
         QuicSentPacketPoolReturnPacketMetadata(&Connection->Worker->SentPacketPool, Packet);
     }
 
+    if (NewDataQueued) {
+        QuicPerfCounterAdd(PacketsRetransmitted, 1);
```
nibanks, line 770: I thought we already had a place where we calculated retransmitted bytes for the connection... If so, we should use that to increment this probably.

src/inc/msquic.h
```diff
@@ -295,6 +295,18 @@ typedef struct QUIC_LISTENER_STATISTICS {
     } Binding;
 } QUIC_LISTENER_STATISTICS;
 
+typedef enum QUIC_PERFORMANCE_COUNTERS {
+    ConnectionsCreated,
```
nibanks, line 299: All out enums are all caps and prefixed with the type. So `QUIC_PERF_COUNT_CONNECTIONS_CREATED` maybe?

src/platform/unittest/TlsTest.cpp
```diff
@@ -932,4 +932,93 @@ TEST_P(TlsTest, PacketEncryptionPerf)
...
+    )
+{
+    uint64_t Start, End, Accumulator = 0;
+    QUIC_DISPATCH_LOCK Lock;
+    uint64_t Counter = 0;
+
+    QuicDispatchLockInitialize(&Lock);
+
+    for (uint64_t j = 0; j < LoopCount; ++j) {
+        Start = QuicTimeUs64();
```
nibanks, line 946: If you put the start/stop time in the loop you end up skewing the results I believe. Especially if things are really fast. Try putting it outside the loop and measure the time for the whole loop.

nibanks:
I want `QuicTraceRundown` to dump all the stats to an ETW event eventually. Either do that in this PR or at least put a TODO in there.

src/platform/unittest/TlsTest.cpp
anrossi: But then you collect the time for initializing, comparing, and incrementing the counter, as well as the jump instruction.

src/core/connection.c
```diff
@@ -2443,6 +2452,7 @@ QuicConnHandshakeConfigure(
 
     Connection->State.Started = TRUE;
     Connection->Stats.Timing.Start = QuicTimeUs64();
+    QuicPerfCounterAdd(ConnectionsActive, 1);
```
nibanks, line 2455: I think it's better to increment this at the same point of create and decrement on free.

src/core/inline.c
```diff
@@ -171,6 +171,14 @@ QuicPartitionIndexDecrement(
...
+void
```
nibanks, line 176:
```suggestion
_IRQL_requires_max_(DISPATCH_LEVEL)
void
```

src/core/loss_detection.c
anrossi: There's a per-connection counter for retransmittable packets, but that's just packets eligible for retransmission.

src/core/library.c
```diff
@@ -663,6 +663,49 @@ QuicLibraryGetGlobalParam(
...
+            }
+        }
+
+        if((int64_t)Counters[QUIC_PERF_COUNTER_CONN_ACTIVE] < 0) {
+            Counters[QUIC_PERF_COUNTER_CONN_ACTIVE] = 0;
+        }
+
+        Status = QUIC_STATUS_SUCCESS;
+        break;
+        }
```
nibanks, line 707:
```suggestion
    }
```

src/core/library.h
```diff
@@ -305,6 +310,19 @@ QuicPartitionIndexDecrement(
...
+    _In_ QUIC_PERFORMANCE_COUNTERS Type,
+    _In_ int64_t Value
+    )
+{
+    QUIC_DBG_ASSERT(Type < QUIC_PERF_COUNTER_MAX);
+    uint8_t ProcIndex = QuicLibraryGetCurrentPartition();
```
nibanks, line 322: Change to current proc index.

src/inc/msquic.h
```diff
@@ -353,6 +365,7 @@ typedef enum QUIC_SERVER_RESUMPTION_LEVEL {
 #define QUIC_PARAM_GLOBAL_RETRY_MEMORY_PERCENT          0   // uint16_t
 #define QUIC_PARAM_GLOBAL_SUPPORTED_VERSIONS            1   // uint32_t[] - network byte order
 #define QUIC_PARAM_GLOBAL_LOAD_BALACING_MODE            2   // uint16_t - QUIC_LOAD_BALANCING_MODE
+#define QUIC_PARAM_GLOBAL_PERF_COUNTERS                 3   // uint64_t[] - Array size is QUIC_PERF_COUNTER_MAX
```
nibanks, line 368: Add some test code to query it?

src/core/library.c
anrossi: Doesn't this confuse with the closing brace for the enclosing `switch` statement?

src/core/connection.c
anrossi: But won't that count connections that never do anything?

src/core/api.c
```diff
@@ -944,6 +944,11 @@ MsQuicStreamSend(
     SendRequest->TotalLength = TotalLength;
     SendRequest->ClientContext = ClientSendContext;
 
+    uint32_t TotalBytesSent = 0;
+    for (uint32_t i = 0; i < BufferCount; ++i) {
+        TotalBytesSent += Buffers[i].Length;
+    }
+
```
nibanks, line 951: Put this stuff in QuicStreamSendFlush instead. BTW, `TotalLength` represents what you want already.

src/inc/msquic.h
```diff
@@ -295,6 +295,26 @@ typedef struct QUIC_LISTENER_STATISTICS {
     } Binding;
 } QUIC_LISTENER_STATISTICS;
 
+typedef enum QUIC_PERFORMANCE_COUNTERS {
```
nibanks, line 298: Let's put some definitions in comments after each event.

src/inc/msquic.h
```diff
@@ -295,6 +295,26 @@ typedef struct QUIC_LISTENER_STATISTICS {
...
+    QUIC_PERF_COUNTER_CONN_APP_REJECT,
+    QUIC_PERF_COUNTER_CONN_RESUMED,
+    QUIC_PERF_COUNTER_CONN_ACTIVE,
+    QUIC_PERF_COUNTER_CONN_CLOSED_ERROR,
+    QUIC_PERF_COUNTER_CONN_NO_ALPN,
+    QUIC_PERF_COUNTER_PKTS_SUSPECTED_LOST,
+    QUIC_PERF_COUNTER_PKTS_DROPPED,
+    QUIC_PERF_COUNTER_PKTS_DECRYPTION_FAIL,
+    QUIC_PERF_COUNTER_DGRAM_RECV,
+    QUIC_PERF_COUNTER_DGRAM_SEND,
```
nibanks, line 310: I also want a send/recv event that increments by 1 for each send call or receive event.

src/core/crypto.c
```diff
@@ -1417,6 +1417,10 @@ QuicCryptoProcessTlsCompletion(
             "Indicating QUIC_CONNECTION_EVENT_CONNECTED (Resume=%hhu)",
             Event.CONNECTED.SessionResumed);
         (void)QuicConnIndicateEvent(Connection, &Event);
+        if (Crypto->TlsState.SessionResumed) {
+            QuicPerfCounterIncrement(QUIC_PERF_COUNTER_CONN_RESUMED);
+        }
+        Connection->Stats.ResumptionSucceeded = Crypto->TlsState.SessionResumed;
```
nibanks, line 1423: Squeezing in a fix too?

src/core/connection.c
anrossi: It's more consistent to do it the way proposed by Nick.

src/core/crypto.c
anrossi: Yeah :)
Could make it a separate PR for clarity.

src/core/datagram.c
```diff
@@ -409,6 +409,7 @@ QuicDatagramSendFlush(
     QUIC_SEND_REQUEST* ApiQueue = Datagram->ApiQueue;
     Datagram->ApiQueue = NULL;
     QuicDispatchLockRelease(&Datagram->ApiQueueLock);
+    uint32_t TotalBytesSent = 0;
```
anrossi, line 412: This could be more than 32-bits

src/core/library.c
```diff
@@ -505,6 +505,58 @@ QuicLibApplyLoadBalancingSetting(
...
+
+    if (QUIC_PERF_COUNTER_CONN_QUEUE_DEPTH < CountersPerBuffer &&
+        (int64_t)Counters[QUIC_PERF_COUNTER_CONN_QUEUE_DEPTH] < 0) {
+        Counters[QUIC_PERF_COUNTER_CONN_QUEUE_DEPTH] = 0;
+    }
+
+    if (QUIC_PERF_COUNTER_OPER_QUEUE_DEPTH < CountersPerBuffer &&
+        (int64_t)Counters[QUIC_PERF_COUNTER_OPER_QUEUE_DEPTH] < 0) {
+        Counters[QUIC_PERF_COUNTER_OPER_QUEUE_DEPTH] = 0;
+    }
```
nibanks, line 557: I'm thinking it's still better to just have a big loop and not be type specific. It scales better.

src/core/operation.c
```diff
@@ -179,6 +183,7 @@ QuicOperationDequeue(
 #if DEBUG
         Oper->Link.Flink = NULL;
 #endif
+        QuicPerfCounterDecrement(QUIC_PERF_COUNTER_OPER_QUEUE_DEPTH);
```
nibanks, line 186: Maybe put this outside the lock, only if `Oper` isn't null.

src/core/connection.c
```diff
@@ -1315,6 +1319,10 @@ QuicConnOnShutdownComplete(
 
     if (Connection->State.ExternalOwner == FALSE) {
 
+        if (Connection->State.Started && !Connection->State.Connected) {
+            QuicPerfCounterIncrement(QUIC_PERF_COUNTER_CONN_HANDSHAKE_FAIL);
```
nibanks, line 1323: In order to have a single place to increment this, let's just do it on free, if started but not connected.

src/core/connection.c
```diff
@@ -359,6 +361,8 @@ QuicConnFree(
 #if DEBUG
     InterlockedDecrement(&MsQuicLib.ConnectionCount);
 #endif
+    QuicPerfCounterDecrement(QUIC_PERF_COUNTER_CONN_ACTIVE);
+    QuicPerfCounterDecrement(QUIC_PERF_COUNTER_CONN_CONNECTED);
```
nibanks, line 365: Only decrement this if it was connected, right?

src/core/connection.c
```diff
@@ -1522,6 +1534,7 @@ QuicConnTryClose(
         if (ResultQuicStatus) {
             Connection->CloseStatus = (QUIC_STATUS)ErrorCode;
             Connection->CloseErrorCode = QUIC_ERROR_INTERNAL_ERROR;
+            QuicPerfCounterIncrement(QUIC_PERF_COUNTER_CONN_CLOSED_ERROR);
```
nibanks, line 1537: Let's call this event _SHUTDOWN_ERROR

nibanks: Though, I'm not convinced of the usefulness of this event, because connection idle is an "error" but will be the normal way HTTP connections are shutdown.

src/core/stream.c
```diff
@@ -225,6 +225,7 @@ QuicStreamStart(
     }
 
     Stream->Flags.Started = TRUE;
+    QuicPerfCounterIncrement(QUIC_PERF_COUNTER_STRM_ACTIVE);
```
nibanks, line 228: Let's have this be from alloc to free.

src/core/worker.c
```diff
@@ -205,6 +205,8 @@ QuicWorkerQueueConnection(
 
     QuicDispatchLockRelease(&Worker->Lock);
 
+    QuicPerfCounterIncrement(QUIC_PERF_COUNTER_CONN_QUEUE_DEPTH);
```
nibanks, line 208: This isn't the right place. Notice above where the connection is actually added to the list and the ref count is added. This perf counter should pretty much match up with ref count add/release calls.

src/manifest/MsQuicEtw.man
```diff
@@ -3343,6 +3363,11 @@
             id="Etw.ApiWaitOperation"
             value="[ api] Waiting on operation"
             />
+
```
nibanks, line 3366: nit: remove extra newline

src/test/bin/winkernel/control.cpp
```diff
@@ -951,6 +951,10 @@ QuicTestCtlEvtIoDeviceControl(
                 Params->Family));
         break;
 
+    case IOCTL_QUIC_RUN_VALIDATE_GET_PERF_COUNTERS:
```
nibanks, line 954: You're missing the length (of zero) in the array above this function.

src/inc/msquic.h
```diff
@@ -295,6 +295,34 @@ typedef struct QUIC_LISTENER_STATISTICS {
...
 } QUIC_LISTENER_STATISTICS;
 
+typedef enum QUIC_PERFORMANCE_COUNTERS {
+    QUIC_PERF_COUNTER_CONN_CREATED,         // Total connections ever allocated.
+    QUIC_PERF_COUNTER_CONN_HANDSHAKE_FAIL,  // Total connections that failed during handshake.
+    QUIC_PERF_COUNTER_CONN_APP_REJECT,      // Total connections rejected by the application.
+    QUIC_PERF_COUNTER_CONN_RESUMED,         // Total connections resumed.
+    QUIC_PERF_COUNTER_CONN_ACTIVE,          // Connections currently allocated.
+    QUIC_PERF_COUNTER_CONN_CONNECTED,       // Connections currently in the connected state.
+    QUIC_PERF_COUNTER_CONN_CLOSED_ERROR,    // Total connections closed with a protocol error.
```
nibanks, line 305: As I mentioned elsewhere, let's use the word shutdown instead of close (since that's the event generated that has the error).

src/manifest/MsQuicEtw.man
```diff
@@ -1779,6 +1790,15 @@
               symbol="QuicApiWaitOperation"
               value="14"
               />
+          <event
+              keywords="ut:LowVolume"
+              level="win:Informational"
+              message="$(string.Etw.PerfCountersRundown)"
+              opcode="win:Info"
+              symbol="QuicPerfCountersRundown"
```
nibanks, line 1798: Add the necessary changes to quicetw and etwlib for these (mostly just header changes).

src/core/library.c
anrossi: Well some of these counter might actually be unsigned values that large. We wouldn't want to zero those out.

src/core/operation.c
anrossi: The conundrum was whether adding another if-statement hurt performance more than doing the interlocked operation under lock.  The if-statement is probably faster by far.

src/core/connection.c
anrossi: Should it be scoped to non-idle errors? Or only app-level errors?

nibanks: I don't know... Maybe we call this _CONN_PROTOCOL_ERRORS instead and scope this to a subset of error codes?

src/core/operation.c
nibanks: Yeah, the if will be faster, especially once trained with PGO.

src/core/library.c
nibanks: You think some of these counters will be greater than 2^63? That's [nearly 300 years](https://www.bing.com/search?q=2%5E63+nanoseconds+to+years&qs=n&form=QBRE&sp=-1&pq=2%5E63+nanoseconds+to+years&sc=5-25&sk=&cvid=F6D98D58B097428CAF2BAF6DF4AF2EB4) if the counter is incremented once every nanosecond...

anrossi: I was thinking that the quic stream data counters are 62 bits, and summing them up might overflow a signed value. But even with a 100Gb NIC, that would take 23 years. Point taken.

src/core/worker.c
anrossi: Since, in the other place, we decided to not have the counter increment inside the lock, I'm going to set a boolean when the connection is queued, and wrap this counter in an if-statement.

src/inc/msquic.h
anrossi: I think the above comment suggested QUIC_PERF_COUNTER_CONN_PROTOCOL_ERRORS. I'm going with this unless there's objections.

src/core/connection.c
```diff
@@ -1522,6 +1536,9 @@ QuicConnTryClose(
         if (ResultQuicStatus) {
             Connection->CloseStatus = (QUIC_STATUS)ErrorCode;
             Connection->CloseErrorCode = QUIC_ERROR_INTERNAL_ERROR;
+            if (ErrorCode != QUIC_STATUS_CONNECTION_IDLE) {
```
nibanks, line 1539: I feel like we shouldn't include TIMEOUT in this case either. It's not a protocol error. It just means one side of the connection is no longer reachable.

src/core/library.c
```diff
@@ -505,6 +505,28 @@ QuicLibApplyLoadBalancingSetting(
...
+    int64_t* const Counters = (int64_t*)Buffer;
+    memcpy(Buffer, MsQuicLib.PerProc[0].PerfCounters, BufferLength);
+
+    for (uint32_t ProcIndex = 1; ProcIndex < MsQuicLib.ProcessorCount; ++ProcIndex) {
+        for (uint32_t CounterIndex = 0; CounterIndex < CountersPerBuffer; ++CounterIndex) {
+            Counters[CounterIndex] += MsQuicLib.PerProc[ProcIndex].PerfCounters[CounterIndex];
+
+            if (Counters[CounterIndex] < 0) {
+                Counters[CounterIndex] = 0;
+            }
```
nibanks, line 525: You need to do this only after summing them all up.

anrossi: I was trying to save another loop over the PerProc objects.
If I invert the loops so instead the outer loop is over the counters, and the inner loop is over the PerProc, I could do that. But I think that approach would hurt cache locality. Do you have an opinion against a second loop?

src/manifest/clog.sidecar
```diff
@@ -22907,6 +22907,34 @@
...
       }
+    },
+    "PerfCountersRundown": {
+      "ModuleProperites": {},
+      "TraceString": "[ lib] Perf counters Rundown, Counters=%!CID!",
+      "UniqueId": "PerfCountersRundown",
+      "splitArgs": [
+        {
+          "VariableInfo": {
+            "UserSpecifiedUnModified": "\r\n            CLOG_BYTEARRAY(sizeof(PerfCounters), PerfCounters)",
```
nibanks, line 22918: I don't think this line is necessary any longer. Can you update your clog and delete this and see if things still build?

src/manifest/clog.sidecar
```diff
@@ -22956,7 +22984,7 @@
...
...t = value[2] | ((ushort)value[3] << 8);\r\n\r\n            Span<byte> sa2 = value;\r\n\r\n            string msg = \"\";\r\n\r\n            switch (si_family)\r\n            {\r\n                case 0:  //<--unspecified\r\n                    msg += $\"*:{sin_port}\";\r\n                    break;\r\n                case 2:  //< --v4\r\n                    msg += $\"{new IPAddress(sa2.Slice(4, 4)).ToString()}:{sin_port}\";\r\n                    break;\r\n                case 10:  //<--v6 (linux)\r\n                case 23: //<--v6\r\n                    msg += $\"[{new IPAddress(sa2.Slice(8, 16)).ToString()}]:{sin_port}\";\r\n                    break;\r\n                default:\r\n                    throw new NotSupportedException(\"Invalid SI_FAMILY : \" + si_family);\r\n            }\r\n            \r\n            return msg;\r\n        }\r\n\r\n        public static string CID(byte [] value)\r\n        {\r\n            StringBuilder hex = new StringBuilder(value.Length * 2);\r\n            foreach (byte v in value)\r\n            {\r\n                hex.AppendFormat(v.ToString(\"X2\"));\r\n            }\r\n            return hex.ToString();\r\n        }\r\n    }\r\n}\r\n",
```
nibanks, line 22987: Ditto for updating clog and reverting this change.

nibanks:
You've got a conflict with the sidecar too. Need to merge master

src/manifest/clog.sidecar
anrossi: I merged master locally (accept incoming), and when I re-ran clog this change went away. so I don't think I need it.

src/core/library.c
nibanks: Just make a second loop.

src/core/connection.c
```diff
@@ -1339,6 +1349,10 @@ QuicConnOnShutdownComplete(
         (void)QuicConnIndicateEvent(Connection, &Event);
 
         Connection->ClientCallbackHandler = NULL;
+
+        if (Connection->State.Started && !Connection->State.Connected) {
+            QuicPerfCounterIncrement(QUIC_PERF_COUNTER_CONN_HANDSHAKE_FAIL);
+        }
```
anrossi, line 1355: Delete this.

nibanks:
@anrossi I fixed a bug in the increment counter function where you were using proc count instead of current proc. I also added a helper to dump the counters to the console and added a call in spinquic right before it closes the top level handle.
```
PS F:\msquic> .\scripts\spin.ps1
Perf Counters:
  CONN_CREATED:         68779
  CONN_HANDSHAKE_FAIL:  3615
  CONN_APP_REJECT:      0
  CONN_ACTIVE:          2
  CONN_CONNECTED:       0
  CONN_PROTOCOL_ERRORS: 33648
  CONN_NO_ALPN:         0
  STRM_ACTIVE:          0
  PKTS_SUSPECTED_LOST:  431
  PKTS_DROPPED:         6466
  PKTS_DECRYPTION_FAIL: 3123
  UDP_RECV:             148087
  UDP_SEND:             181068
  UDP_RECV_BYTES:       115820797
  UDP_SEND_BYTES:       117027331
  UDP_RECV_EVENTS:      148087
  UDP_SEND_CALLS:       135821
  APP_SEND_BYTES:       6653066828
  APP_RECV_BYTES:       12770306
  CONN_QUEUE_DEPTH:     0
  OPER_QUEUE_DEPTH:     5895
  OPER_QUEUED:          998791
  OPER_COMPLETED:       1002424
```
Notice, `OPER_QUEUE_DEPTH` is non-zero. Not a good sign... We should probably assert (at least in spinquic) that all the "current" or "depth" counters are zero at the end of the run. Otherwise we have a counter bug or a leak somewhere.

src/core/binding.c
```diff
@@ -1515,6 +1531,9 @@ QuicBindingReceive(
     if (ReleaseChain != NULL) {
         QuicDataPathBindingReturnRecvDatagrams(ReleaseChain);
     }
+    QuicPerfCounterAdd(QUIC_PERF_COUNTER_UDP_RECV, TotalChainLength);
```
nibanks, line 1534: nit: newline before this

src/core/library.c
```diff
@@ -313,6 +343,16 @@ MsQuicLibraryUninitialize(
     QUIC_TEL_ASSERT(MsQuicLib.ConnectionCount == 0);
 #endif
 
+#if DEBUG
+    uint64_t PerfCounters[QUIC_PERF_COUNTER_MAX];
+    QuicLibrarySumPerfCounters((uint8_t*)PerfCounters, sizeof(PerfCounters));
+    QUIC_DBG_ASSERT(PerfCounters[QUIC_PERF_COUNTER_CONN_ACTIVE] == 0);
```
nibanks, line 349: Let's put a comment along the lines of "All these are current active counts and should be zero on cleanup"

src/core/operation.c
```diff
@@ -199,9 +207,12 @@ QuicOperationQueueClear(
     QuicListMoveItems(&OperQ->List, &OldList);
     QuicDispatchLockRelease(&OperQ->Lock);
 
+    int64_t OperationsDequeued = 0;
+
     while (!QuicListIsEmpty(&OldList)) {
         QUIC_OPERATION* Oper =
             QUIC_CONTAINING_RECORD(QuicListRemoveHead(&OldList), QUIC_OPERATION, Link);
+            --OperationsDequeued;
```
nibanks, line 215: nit: tab left

src/core/stream_recv.c
```diff
@@ -777,6 +777,7 @@ QuicStreamRecvFlush(
                 Stream->Connection->Registration->AppName,
                 Status, 0);
         }
+        QuicPerfCounterAdd(QUIC_PERF_COUNTER_APP_RECV_BYTES, Event.RECEIVE.TotalBufferLength);
```
nibanks, line 780: Move to completion code path.

src/core/stream_send.c
```diff
@@ -587,6 +589,7 @@ QuicStreamSendFlush(
     if (Start) {
         (void)QuicStreamStart(Stream, QUIC_STREAM_START_FLAG_ASYNC, FALSE);
     }
+    QuicPerfCounterAdd(QUIC_PERF_COUNTER_APP_SEND_BYTES, TotalBytesSent);
```
nibanks, line 592: nit: newline before

src/tools/spin/spinquic.cpp
```diff
@@ -927,6 +927,10 @@ main(int argc, char **argv)
         MsQuic->RegistrationClose(Registration);
         Registration = nullptr;
 
+        QuicSleep(4000);
```
nibanks, line 930: Remove sleep and maybe dump

anrossi: I'll remove the sleep, but keep the dump.

src/core/worker.c
```diff
@@ -656,15 +666,18 @@ QUIC_THREAD_CALLBACK(QuicWorkerThread, Context)
         QuicConnRelease(Connection, QUIC_CONN_REF_WORKER);
```
nibanks, line 666: Need to clean up the connection queue depth counter here too.

PR #736, opened 2020-08-19
Enable GitHub Code Scanning

larseggert:


nibanks:
@larseggert so what am I supposed to do with this to onboard it?

larseggert:
Once it's onboarded, you will see issue pop up under the "security tab". Once merged, I think it should start running automatically.

PR #413, opened 2020-05-27
Update OpenSSL submodule to new branch

anrossi:

PR #754, opened 2020-08-21
OpenSSL master-alpha4-quic-support

larseggert:
Similar to #413 but with a newer branch and rebased

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,7 +4,7 @@
 [submodule "submodules/openssl"]
 	path = submodules/openssl
 	url = https://github.com/akamai/openssl.git
-	branch = master-alpha2-quic-support
+	branch = master-alpha4-quic-support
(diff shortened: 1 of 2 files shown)


larseggert:
But a test failure:
```
/vagrant/src/platform/unittest/TlsTest.cpp:69: Failure
Expected: (nullptr) != (SelfSignedCertParams), actual: (nullptr) vs NULL
```

larseggert:
OK, that test failure was fixed by 481f66d. There is another one now:
```
/vagrant/src/platform/unittest/TlsTest.cpp:730: Failure
Value of: Result & QUIC_TLS_RESULT_ERROR
  Actual: false
Expected: true
[  FAILED  ] TlsTest.CertificateError (12 ms)
```

larseggert:
I think someone who knows this code needs to figure out what this is about - @anrossi?

PR #760, opened 2020-08-24
Fix Bad TLS Events

nibanks:
I fixes the TLS created events to give more context, and in the schannel case, properly pass the right argument (not null).

@chgray, I updated the strings and rand clog in dev mode for all the affected TLS configurations, but it seems clog only updated stub for some reason.


nibanks:
@chgray it seems I had to do a clean build for some reason to get clog to update the other events. I opened https://github.com/microsoft/CLOG/issues/63 tom track this.

PR #761, opened 2020-08-24
Fix Incorrect numa node initialization with 2 numa nodes but 1 group

ThadHouse:
We were only matching the group, but not the specific processor in the group. Use a mask to find the specific processor to match.

### DIFF
--- a/src/platform/platform_winuser.c
+++ b/src/platform/platform_winuser.c
@@ -207,7 +207,7 @@ QuicProcessorInfoInit(
             if (Info->Relationship == RelationGroup) {
                 uint32_t ProcessorOffset = 0;
                 for (WORD i = 0; i < Info->Group.ActiveGroupCount; ++i) {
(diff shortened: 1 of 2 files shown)


src/platform/platform_winuser.c
```diff
@@ -210,6 +210,7 @@ QuicProcessorInfoInit(
                     if (Index - ProcessorOffset < Info->Group.GroupInfo[i].ActiveProcessorCount) {
                         QuicProcessorInfo[Index].Group = i;
                         QuicProcessorInfo[Index].Index = Index - ProcessorOffset;
+                        QuicProcessorInfo[Index].MaskInGroup = 1ull << (Index - ProcessorOffset);
```
nibanks, line 213: Should we debug assert the diff is less than 64?

ThadHouse: Probably

PR #762, opened 2020-08-24
Add Release Info for v1.0

nibanks:


nibanks:
@pravb or @dahavey any comments?

nibanks:
FYI, I'm merging this at EoD if there is no additional feedback.

Issue #537, opened 2020-06-19
Stub TLS on linux leaks a stream in AbortiveShutdown tests


anrossi:
### Describe the bug

The AbortiveShutdown tests on Linux, using Stub TLS, leak a QUIC stream.

### Steps to reproduce the behavior

1. Build with the `-SanitizeAddress` and `-Tls stub` switches
2. Run the AbortiveShutdown tests with `Filter *AbortiveShutdown*`

### Expected vs actual behavior

LeakAnalyzer shouldn't be throwing errors for these test cases.
Also, AZP should detect LeakAnalyzer failures and fail the run instead of marking the run as successful.  Perhaps gtest has an option to fail when LeakAnalyzer throws a warning?

ThadHouse:
Just to double check @anrossi were you seeing a stream leak or were you seeing the following?


```

/home/thad/msquic/src/platform/tls_stub.c:1880:48: runtime error: store to misaligned address 0x61b00003c454 for type 'uint64_t', which requires 8 byte alignment
0x61b00003c454: note: pointer points here
  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00
              ^
/home/thad/msquic/src/platform/tls_stub.c:1881:48: runtime error: store to misaligned address 0x61b00003c45c for type 'uint64_t', which requires 8 byte alignment
0x61b00003c45c: note: pointer points here
  e7 b3 53 72 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00
              ^
/home/thad/msquic/src/platform/tls_stub.c:1903:9: runtime error: load of misaligned address 0x61b00001ccb4 for type 'uint64_t', which requires 8 byte alignment
0x61b00001ccb4: note: pointer points here
  00 00 00 00 b4 89 d1 c1  e7 b3 53 72 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00
              ^

[08/25/2020 08:49:01] 192 test(s) run. 0 test(s) failed.
```

Trying to repro this, I'm not seeing a leak, but I am seeing those messages.

PR #764, opened 2020-08-25
Fix unaligned read and write ASAN error in stub TLS

ThadHouse:
The pointer punning was unaligned, and ASAN warns about this.

This might be whats seen in #537 but I'm not sure. That issue specifically doesn't reproduce for me.


src/platform/tls_stub.c
```diff
@@ -1900,7 +1902,9 @@ QuicDecrypt(
     UNREFERENCED_PARAMETER(AuthDataLength);
     UNREFERENCED_PARAMETER(AuthData);
     uint16_t PlainTextLength = BufferLength - QUIC_ENCRYPTION_OVERHEAD;
-    if (*(uint64_t*)(Buffer + PlainTextLength) != Key->Secret) {
+    uint64_t ReadVal = 0;
+    QuicCopyMemory(&ReadVal, Buffer + PlainTextLength, sizeof(ReadVal));
+    if (ReadVal != Key->Secret) {
```
nibanks, line 1907: Use a mem compare instead?

src/platform/tls_stub.c
```diff
@@ -1877,8 +1877,10 @@ QuicEncrypt(
     UNREFERENCED_PARAMETER(AuthDataLength);
     UNREFERENCED_PARAMETER(AuthData);
     uint16_t PlainTextLength = BufferLength - QUIC_ENCRYPTION_OVERHEAD;
-    ((uint64_t*)(Buffer + PlainTextLength))[0] = Key->Secret;
-    ((uint64_t*)(Buffer + PlainTextLength))[1] = 0;
+    uint64_t WriteVal = Key->Secret;
+    QuicCopyMemory(Buffer + PlainTextLength, &WriteVal, sizeof(WriteVal));
```
nibanks, line 1881: You shouldn't need a temporary variable.

src/platform/tls_stub.c
```diff
@@ -1877,8 +1877,10 @@ QuicEncrypt(
...
+    WriteVal = 0;
+    QuicCopyMemory(Buffer + PlainTextLength + sizeof(WriteVal), &WriteVal, sizeof(WriteVal));
```
nibanks, line 1883: Just use a mem set instead? Or QuicZeroMemory?

src/platform/tls_stub.c
ThadHouse: I needed the temp for the 2nd value. But if I use zero, I can get rid of it.

nibanks:
@ThadHouse we've consistently been seeing some failures with this now:

image

I ran it three times and can't get a clean run. Can you take a look?

PR #765, opened 2020-08-25
Add RPS Helper Script

nibanks:


ThadHouse:
Should we do all the configuration stuff for this. Since this one currently needs to be in the directory containing the tool. I'm fine with it for now if we just need something simple.

nibanks:
> Should we do all the configuration stuff for this? Since this one currently needs to be in the directory containing the tool. I'm fine with it for now if we just need something simple.

For now, just this.

PR #777, opened 2020-08-26
QUIC Configuration Interface

nibanks:
Refactors much of the MsQuic control path APIs.


src/inc/msquic.h
```diff
@@ -77,24 +74,32 @@ typedef enum QUIC_EXECUTION_PROFILE {
     QUIC_EXECUTION_PROFILE_LOW_LATENCY,         // Default
     QUIC_EXECUTION_PROFILE_TYPE_MAX_THROUGHPUT,
     QUIC_EXECUTION_PROFILE_TYPE_SCAVENGER,
-    QUIC_EXECUTION_PROFILE_TYPE_REAL_TIME
+    QUIC_EXECUTION_PROFILE_TYPE_REAL_TIME,
+    QUIC_EXECUTION_PROFILE_TYPE_INTERNAL        // For internal use, equal to LOW_LATENCY
```
nibanks, line 78: Hide from API (make internal).

src/inc/msquic.h
```diff
@@ -197,6 +202,27 @@ typedef struct QUIC_REGISTRATION_CONFIG { // All fields may be NULL/zero.
...
+    _In_opt_ void* Context,
+    _In_ QUIC_STATUS Status
+    );
+
+typedef QUIC_CREDENTIAL_LOAD_COMPLETE *QUIC_CREDENTIAL_LOAD_COMPLETE_HANDLER;
+
+typedef struct QUIC_CREDENTIAL_CONFIG {
+    QUIC_CREDENTIAL_TYPE Type;
+    QUIC_CREDENTIAL_FLAGS Flags;
+    void* Creds;
```
nibanks, line 220: Possibly replace with a pointer to a union (`QUIC_CERTIFICATE`?) of all other structs below?

src/inc/msquic.h
```diff
@@ -580,92 +603,62 @@ void
...
+    _Outptr_ _At_(*Configuration, __drv_allocatesMem(Mem)) _Pre_defensive_
+        HQUIC* Configuration
     );
 
 //
-// Closes an existing session. This function synchronizes the cleanup of all
-// child objects (listeners and connections). It does this by blocking until
-// all those child objects have been closed by the application.
-// N.B. This function will deadlock if called in any MsQuic callbacks.
+// Closes an existing configuration. N.B. This function will deadlock if called
```
nibanks, line 642: TODO - Won't actually deadlock.

src/inc/msquic.h
```diff
@@ -889,12 +884,25 @@ _IRQL_requires_max_(PASSIVE_LEVEL)
 QUIC_STATUS
 (QUIC_API * QUIC_CONNECTION_START_FN)(
     _In_ _Pre_defensive_ HQUIC Connection,
+    _In_ _Pre_defensive_ HQUIC Configuration,
```
nibanks, line 887: Make sure this checks the configuration has been loaded.

src/tools/sample/sample.cpp
```diff
@@ -328,53 +430,72 @@ ClientConnectionCallback(
...
+bool
+ClientLoadConfiguration(
+    bool Unsecure
+    )
+{
+    QUIC_SETTINGS Settings{0};
+    Settings.IdleTimeoutMs = IdleTimeoutMs;
+    Settings.IsSet.IdleTimeoutMs = TRUE;
+
+    QUIC_CREDENTIAL_CONFIG CredConfig;
```
nibanks, line 442: Zero out

src/core/library.c
```diff
@@ -164,6 +151,27 @@ MsQuicLibraryReadSettings(
...
 }
 
+_IRQL_requires_max_(PASSIVE_LEVEL)
+_Function_class_(QUIC_STORAGE_CHANGE_CALLBACK)
+void
+MsQuicLibraryReadSettings(
+    _In_opt_ void* Context
+    )
+{
+    QuicSettingsSetDefault(&MsQuicLib.Settings);
```
nibanks, line 161: Make sure this doesn't overwrite any app set value.

src/platform/tls_schannel.c
```diff
@@ -816,221 +808,165 @@ const static UNICODE_STRING QuicTlsPackageName = RTL_CONSTANT_STRING(L"Schannel"
...
+    Credentials->dwFlags |= SCH_USE_STRONG_CRYPTO;
+    if (CredConfig->Flags & QUIC_CREDENTIAL_FLAG_NO_CERTIFICATE_VALIDATION) {
+        Credentials->dwFlags |= SCH_CRED_MANUAL_CRED_VALIDATION;
+    }
+    if (CredConfig->Flags & QUIC_CREDENTIAL_FLAG_ENABLE_OCSP) {
+        Credentials->dwFlags |= SCH_CRED_SNI_ENABLE_OCSP;
+    }
+    if (CredConfig->Flags & QUIC_CREDENTIAL_FLAG_CLIENT) {
+        Credentials->dwFlags |= SCH_CRED_NO_DEFAULT_CREDS;
+        Credentials->pTlsParameters->grbitDisabledProtocols = (DWORD)~SP_PROT_TLS1_3_CLIENT;
```
anrossi, line 906: If you're using the same configuration for server and client, this might cause problems

src/platform/tls_schannel.c
```diff
@@ -1445,22 +1273,17 @@ QuicTlsInitialize(
...
         (uint16_t)(Config->LocalTPLength - FIELD_OFFSET(SEND_GENERIC_TLS_EXTENSION, Buffer));
 
-    //
-    // Associate the existing security config with this TLS context.
-    //
-    TlsContext->SecConfig = QuicTlsSecConfigAddRef(Config->SecConfig);
-
     State->EarlyDataState = QUIC_TLS_EARLY_DATA_UNSUPPORTED; // 0-RTT not currently supported.
+    if (Config->ResumptionTicketBuffer != NULL) {
+        QUIC_FREE(Config->ResumptionTicketBuffer);
```
anrossi, line 1278: should this be set NULL here to prevent double-free?

src/platform/tls_schannel.c
nibanks: They can't be the same

src/platform/tls_schannel.c
nibanks: No. The contract is that success from this function indicates ownership of the buffer.

ThadHouse: Does this come from externally allocated code? If so, this is not safe to blindly free, and we either need to heavily document exactly which allocator must be used for this, find a way where we don't take ownership, or add a way to get a properly allocated buffer back.

nibanks: No. This is internal to internal code.

src/core/crypto.c
```diff
@@ -1371,37 +1384,17 @@ QuicCryptoProcessTlsCompletion(
...
-            QuicTlsSecConfigRelease(SecConfig);
-        }
-
         QUIC_DBG_ASSERT(Crypto->TlsState.NegotiatedAlpn != NULL);
         if (!QuicConnIsServer(Connection)) {
             //
             // Currently, NegotiatedAlpn points into TLS state memory, which
             // doesn't live as long as the connection. Update it to point to the
-            // session state memory instead.
+            // configuration state memory instead.
```
anrossi, line 1392: is this still correct?

nibanks: Yes.

anrossi (approved):

PR #782, opened 2020-08-27
Switch CFLAGS based on CMAKE_BUILD_TYPE

larseggert:
And some other CFLAGS-related cleanup.

Need to check if this breaks Windows...


CMakeLists.txt
```diff
@@ -115,14 +129,9 @@ option(QUIC_CODE_CHECK "Run static code checkers" OFF)
...
-
-if ("${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Windows")
-    set(QUIC_PLATFORM "windows")
-elseif ("${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Linux")
-    set(QUIC_PLATFORM "linux")
-elseif ("${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Darwin")
-    set(QUIC_PLATFORM "darwin")
+elseif (QUIC_PLATFORM STREQUAL "darwin")
+    message(WARNING "Logging unsupported on ${QUIC_PLATFORM}.")
+    set(QUIC_ENABLE_LOGGING OFF)
```
nibanks, line 134: Logging is supported on darwin. Let's remove this.

nibanks:
@larseggert what exactly do these changes give us?

CMakeLists.txt
```diff
@@ -35,9 +35,23 @@ endif()
 
 project(msquic)
 
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_C_STANDARD 11)
```
ThadHouse, line 39: I'd prefer to not use these globals, they cause dependency issues. Instead, if we want to set standard this way (I'm fine with it) set it as a public target property on the `inc` target. That way its done the correct way.

larseggert:
@nibanks what this gives us is debug symbols and/or optimized code for build types that should have that

CMakeLists.txt
larseggert: If I do something like this
```
set_property(TARGET inc PROPERTY CXX_STANDARD 17)
```
I get 
```
CMake Error at src/inc/CMakeLists.txt:15 (set_property):
  INTERFACE_LIBRARY targets may only have whitelisted properties.  The
  property "CXX_STANDARD" is not allowed.
```

larseggert: Ah, but this seems to work:
```
target_compile_features(inc INTERFACE cxx_std_17)
target_compile_features(inc INTERFACE c_std_11)
```

CMakeLists.txt
```diff
@@ -292,6 +295,18 @@ if(WIN32)
 else()
     # Custom build flags.
 
+    if (NOT CMAKE_SYSTEM_PROCESSOR STREQUAL arm)
+        set(MARCH -march=native)
```
ThadHouse, line 299: I definitely do not want this to be the default. We want to be able to distribute binaries that work on older CPUs. We don't control what CPU level is used by CI.

I'm not opposed to adding a way to set this with a flag, as I definitely understand what its there for.

CMakeLists.txt
```diff
@@ -292,6 +295,18 @@ if(WIN32)
...
+    endif()
+    set(CMAKE_C_FLAGS_DEBUG "-Og -ggdb")
+    set(CMAKE_C_FLAGS_MINSIZEREL "-Os -DNDEBUG")
+    set(CMAKE_C_FLAGS_RELWITHDEBINFO "-Ofast ${MARCH} -ggdb -DNDEBUG")
+    set(CMAKE_C_FLAGS_RELEASE "-Ofast ${MARCH} -DNDEBUG")
+    set(CMAKE_CXX_FLAGS_DEBUG ${CMAKE_C_FLAGS_DEBUG})
+    set(CMAKE_CXX_FLAGS_MINSIZEREL ${CMAKE_C_FLAGS_MINSIZEREL})
+    set(CMAKE_CXX_FLAGS_RELWITHDEBINFO ${CMAKE_C_FLAGS_RELWITHDEBINFO})
+    set(CMAKE_CXX_FLAGS_RELEASE ${CMAKE_C_FLAGS_RELEASE})
```
ThadHouse, line 308: Just to double check, these won't break the macos or clang build, right?

CMakeLists.txt
```diff
@@ -328,7 +343,7 @@ else()
 
     if(QUIC_ENABLE_SANITIZERS)
         message(STATUS "Configuring sanitizers")
-        list(APPEND QUIC_COMMON_FLAGS -fsanitize=address,leak,undefined -fsanitize-address-use-after-scope -Og -g3 -fno-omit-frame-pointer -fno-optimize-sibling-calls)
+        list(APPEND QUIC_COMMON_FLAGS -fsanitize=address,leak,undefined -fsanitize-address-use-after-scope -Og -ggdb -fno-omit-frame-pointer -fno-optimize-sibling-calls)
```
ThadHouse, line 346: I think we would prefer -ggdb3 so we would still have macros. Same with the -ggdb above.

CMakeLists.txt
larseggert: CI didn't fail, I think

CMakeLists.txt
larseggert: ACK

CMakeLists.txt
larseggert: Should I just remove `-march=native` from the PR? Or maybe make it conditional on some flag (`QUIC_OPTIMIZE_MAX` or something like that)?

ThadHouse: I think its a good idea to have it, so a conditional flag is fine. Its just not a good deal.

CMakeLists.txt
Maximus-: -Og is most likely not what we want also (I made this mistake before, and eventually we reverted). -Og is equivalent to -O1 and some -g, so some variables are still optimized out. -O0 -g3 -ggdb is much better :)

PR #785, opened 2020-08-28
Perf Protocol Refactor

nibanks:
Refactors perf code to have a common, generic server.


src/perf/lib/PerfServer.cpp
```diff
@@ -0,0 +1,280 @@
...
+    ) : SelfSignedConfig{SelfSignedConfig} {
+    if (Session.IsValid()) {
+        Session.SetAutoCleanup();
+        Session.SetPeerBidiStreamCount(PERF_DEFAULT_STREAM_COUNT);
+        Session.SetPeerUnidiStreamCount(PERF_DEFAULT_STREAM_COUNT);
+        Session.SetDisconnectTimeout(PERF_DEFAULT_DISCONNECT_TIMEOUT);
+        Session.SetIdleTimeout(PERF_DEFAULT_IDLE_TIMEOUT);
+    }
+    if (DataBufferBuffered) {
+        QUIC_FREE(DataBufferBuffered);
```
ThadHouse, line 47: Why are you doing this in the constructor? They're already nullptr. Did you mean this to be in a destructor?

nibanks: Oops. I meant to put it in the destructor!

src/perf/lib/PerfServer.h
```diff
@@ -0,0 +1,96 @@
...
+#include "PerfBase.h"
+#include "PerfCommon.h"
+
+class PerfServer : public PerfBase {
+public:
+    PerfServer(
+        _In_ PerfSelfSignedConfiguration* SelfSignedConfig
+        );
+
+    ~PerfServer();
```
ThadHouse, line 25: Mark this as override please. `~PerfServer() override;`

PR #791, opened 2020-08-31
Limit number of cores in RPS Client test, fix RPS thread out of range bug

### LINKED ISSUE #792: RPS Performance Tests read processor array out of range.
We pass a uint8_t for processor index without also assuring that the passed index is within the size of the processor array. This causes out of range processor array reading.

ThadHouse:
Closes #792

### DIFF
--- a/src/perf/lib/RpsClient.cpp
+++ b/src/perf/lib/RpsClient.cpp
@@ -136,10 +136,17 @@ RpsClient::Start(
     }
 
     QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
+    uint32_t ActiveProcCount = QuicProcActiveCount();
+    if (ActiveProcCount >= 8) {
+        //
+        // If we have enough cores, leave 2 cores for OS overhead
+        //
+        ActiveProcCount -= 2;
+    }
     for (uint32_t i = 0; i < ConnectionCount; ++i) {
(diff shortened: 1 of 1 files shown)


src/perf/lib/RpsClient.cpp
```diff
@@ -136,10 +136,17 @@ RpsClient::Start(
     }
 
     QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
+    uint32_t ActiveProcCount = QuicProcActiveCount();
```
nibanks, line 139: I think you should only need to leave 1 core free; specifically core 0 (so add one to the index below).

nibanks: Actually, now I realize you're trying to explicitly leave cores free in the second group. But one should still be enough.

ThadHouse: The reason I went 2 was because of hyperthreading. Usually the hyperthread core is the 2nd one in a pair. So pulling 2 in general removes an entire physical core and it's hyperthread pair

nibanks: Ok.

ThadHouse:
Closes #792

ThadHouse:
This doesn't hurt perf, and is within the margin of error, so we're good.

PR #794, opened 2020-08-31
Don't continue perf execution on error

ThadHouse:


.azure/templates/run-performance.yml
```diff
@@ -61,7 +61,7 @@ jobs:
     condition: contains('${{ parameters.testTypes }}', 'Remote')
     displayName: Run Performance Test
     timeoutInMinutes: 45
-    continueOnError: true
+    continueOnError: false
```
nibanks, line 64: This is the default. Just delete the line instead.

PR #796, opened 2020-09-01
Default CLOG Off in CMake

nibanks:
This simplifies the default, generic cmake build requirements. It *should* allow for automated tools like lgtm.com to compile the project by default.


nibanks:
> You would also need to change the default in CMakeLists.txt for that flag to off. Currently it's on.

Forgot to hit save.

PR #797, opened 2020-09-01
Update to CLOG 0.1.8

ThadHouse:
Fixes output of characters, and also properly updates sidecar

### DIFF
--- a/src/manifest/clog.sidecar
+++ b/src/manifest/clog.sidecar
@@ -21918,7 +21918,7 @@
         "MacroConfigurations": {},
         "Version": 1,
         "CustomTypeClogCSharpFile": "DEFAULTS",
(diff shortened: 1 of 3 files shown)


src/manifest/clog.sidecar
```diff
@@ -24635,4 +24636,4 @@
       }
     ]
   }
-}
+}
```
nibanks, line 24639: fyi: removed newline

ThadHouse: Yes. The generated file doesnt have the new line. The old one must have had it because I think I edited it manually.

ThadHouse: I'll look into making clog add a new line at end of file.

PR #801, opened 2020-09-01
Close connection on AEAD integrity failure

### LINKED ISSUE #557: Keep Track of Failed Packet Decryptions
### Describe the bug

According to the spec, we should be keeping track of the number of failed decryption attempts, and when that reaches a certain limit (across all key updated; see https://github.com/quicwg/base-drafts/pull/3789) we should kill the connection immediately.

ThadHouse:
Updates to the spec in https://github.com/quicwg/base-drafts/commit/71a1ff22aef4a3f948b032447dd5bb36e7314e18 require a connection to be closed on too many failed packet decryptions globally across a connection.

The lowest limit was chosen, at 2^23.5 (rounded down)

Closes #557

Still need to add tests

### DIFF
--- a/src/core/quicdef.h
+++ b/src/core/quicdef.h
@@ -409,6 +409,13 @@ QUIC_STATIC_ASSERT(
 //
 #define QUIC_TLS_RESUMPTION_TICKET_VERSION      1
 
+//
+// The AEAD Integrity limit for maximum failed decryption packets over the
+// lifetime of a connection. Set to the lowest limit, which is for
+// AEAD_AES_128_CCM at 2^23.5 (rounded down)
+//
+#define QUIC_AEAD_INTEGRITY_LIMIT 0xB503F3
+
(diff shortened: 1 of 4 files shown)


src/core/quicdef.h
```diff
@@ -409,6 +409,13 @@ QUIC_STATIC_ASSERT(
 //
 #define QUIC_TLS_RESUMPTION_TICKET_VERSION      1
 
+//
+// The AEAD Integrity limit for maximum failed decryption packets over the
+// lifetime of a connection. Set to the lowest limit, which is for
+// AEAD_AES_128_CCM at 2^23.5 (rounded down)
+//
+#define QUIC_AEAD_INTEGRITY_LIMIT 0xB503F3
```
nibanks, line 417: nit: tab over and change the value to decimal to I can understand the value when reading it?

PR #807, opened 2020-09-03
Fix Wrap Around Bug with Partition Count

### LINKED ISSUE #808: msquic.sys bugcheck on 256 processor system
### Describe the bug

On a machine with 256 processors a bug check is hit at https://github.com/microsoft/msquic/blob/main/src/core/connection.c#L69 as `MsQuicLib.PartitionCount == 0`.

`MsQuicLib.PartitionCount` should not be zero avoiding the bug check.

nibanks:
Fixes #808, a bug where `(uint8_t)(256)` results in us setting `PartitionCount` to zero.

### DIFF
--- a/src/core/library.c
+++ b/src/core/library.c
@@ -200,7 +200,7 @@ MsQuicLibraryInitialize(
     //
     MsQuicLib.ProcessorCount = (uint16_t) QuicProcActiveCount();
     QUIC_FRE_ASSERT(MsQuicLib.ProcessorCount > 0);
-    MsQuicLib.PartitionCount = (uint8_t)MsQuicLib.ProcessorCount;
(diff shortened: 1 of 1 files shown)


ThadHouse (commented):
Wouldn't this mean we have a bug somewhere else? Since we should be able to have 256 partitions, 0-255? Does this value indicate the size of the array, or is this used to actually index the array?

Nevermind realized what this is actually having an issue with.

ThadHouse (approved):

anrossi (approved):

nibanks:
Actually, I stand corrected. `QUIC_PARAM_CONN_IDEAL_PROCESSOR` is a `uint8_t`. We should change this soon.

ThadHouse:
I just created #809 to track this.

PR #811, opened 2020-09-04
Settings Refactor

nibanks:
This refactors how settings are set via the API on a session object. Instead of having individual parameters for each setting, the app can (optionally) pass all settings into `SessionOpen` now.


src/core/settings.c
```diff
@@ -107,86 +107,204 @@ QuicSettingsCopy(
...
+void
+QuicSettingApply(
+    _Inout_ QUIC_SETTINGS* Settings,
+    _In_range_(FIELD_OFFSET(QUIC_SETTINGS, MaxBytesPerKey), >=)
+        uint32_t NewSessionSize,
+    _In_reads_bytes_(NewSessionSize)
+        const QUIC_SETTINGS* NewSettings
+    )
+{
+    // TODO - Input validation
```
nibanks, line 200: We need to validate parameters (that need it). Should we fail the call or silently ignore invalid parameters?

anrossi: I think fail the call. Tell consumers up front when they have bugs.

ThadHouse: Why don't we put a status variable in the settings struct, and if it fails we put the error code in there for the ones that fail. That way a user can know which ones fail. And then return a failed status code. But I'd continue looping and set the rest.

src/core/session.c
```diff
@@ -189,6 +192,12 @@ MsQuicSessionOpen(
         goto Error;
     }
 
+    if (Settings != NULL &&
+        SettingsSize < (uint32_t)FIELD_OFFSET(QUIC_SETTINGS, MaxBytesPerKey)) {
```
anrossi, line 196: why not `sizeof`?

nibanks: Because I don't want this size to change when we add a new setting.

anrossi: Is the goal that newer MsQuic versions can work with clients that are targeting an older version?

src/inc/msquic.hpp
```diff
@@ -181,124 +181,124 @@ class UniquePtrArray {
...
-    QUIC_STATUS Init;
-    const QUIC_API_TABLE* ApiTable{nullptr};
+class MsQuicApi : public QUIC_API_TABLE {
+    const QUIC_API_TABLE* ApiTable {nullptr};
+    QUIC_STATUS InitStatus;
 public:
-    QuicApiTable() noexcept {
-        if (QUIC_SUCCEEDED(Init = MsQuicOpen(&ApiTable))) {
+    MsQuicApi() noexcept {
+        if (QUIC_SUCCEEDED(InitStatus = MsQuicOpen(&ApiTable))) {
```
anrossi, line 189: Doesn't assignment within an if-statement cause C++ compilers to throw warnings?

ThadHouse: No, it doesn't. We do it in a ton of C++ places too.

src/core/settings.c
nibanks: I don't understand you. Why would we put a status variable in the struct? We could return an "invalid parameter" error if we want from the SessionOpen call.

src/core/session.c
nibanks: Yes. The scenario is app A compiles and ships against MsQuic 1.0.0. We then ship MsQuic 1.1.0 (or 1.0.1) which is guaranteed not to be a breaking change; and therefor the app should be able to drop in the new msquic.dll (or .so) without updating the app. In order to support adding new settings not being a breaking change, this check here only requires the original size. Anything beyond this is optional.

src/core/settings.c
ThadHouse: So you could know which one failed, rather then just one failed. Hardest part about multi setting apps is knowing what failed. Debugging that sucks, especially for an end user.

nibanks: How is putting a single Status variable in the struct any different from returning a status?

ThadHouse: Because you could loop through the array afterwards and see which one fails. Rather then just getting a generic failure, and having no way to debug other then to just start setting less and less settings until you know which one actually failed.

ThadHouse: Nevermind. I thought this was passing an array. Forgot this was a bit field and a giant struct.

src/test/bin/winkernel/control.cpp
```diff
@@ -267,13 +266,15 @@ QuicTestCtlEvtFileCreate(
         RtlZeroMemory(Client, sizeof(QUIC_TEST_CLIENT));
         KeInitializeEvent(&Client->SecConfigComplete, NotificationEvent, FALSE);
 
-        const QUIC_REGISTRATION_CONFIG RegConfig = { "MsQuicBvt", QUIC_EXECUTION_PROFILE_LOW_LATENCY };
-        Status = MsQuic->RegistrationOpen(&RegConfig, &Client->Registration);
-        if (QUIC_FAILED(Status)) {
+        Registration = new MsQuicRegistration("MsQuicBVT");
```
anrossi, line 269: No setting the execution profile?

src/tools/ping/Server.cpp
```diff
@@ -78,79 +78,35 @@ struct PingServer {
 
 void QuicPingServerRun()
 {
+    QUIC_SETTINGS Settings{0};
```
anrossi, line 81: Why not use the `.hpp` interface?

anrossi (approved):
Looks good to me!

src/test/bin/winkernel/control.cpp
nibanks: Just using the default.

src/tools/ping/Server.cpp
nibanks: No. I don't want to add a new dependency here.

Issue #806, opened 2020-09-03
Unaligned read and write in datagram.c


ThadHouse:
The read happens https://github.com/microsoft/msquic/blob/main/src/core/datagram.c#L109

The write happens https://github.com/microsoft/msquic/blob/main/src/core/datagram.c#L119

Seen in https://dev.azure.com/ms/msquic/_build/results?buildId=106714&view=logs&jobId=4e08e7db-7517-5f8d-8ec9-334a70dc8fc2&j=4e08e7db-7517-5f8d-8ec9-334a70dc8fc2&t=a1521c54-938f-5e18-a5ca-59dc8be93bbb

ThadHouse:
Closed by #814

PR #814, opened 2020-09-08
Pack QUIC_SENT_PACKET_METADATA manually to remove the need for pragma pack

ThadHouse:
This causes everything to be aligned correctly, removing potential bugs and performance penalties for unaligned reads and writes.

This commit currently loses 1 byte per packet, but gains 4 per frame. However, with a future optimization for StreamOffset, we can actually lose 4 bytes per frame, which will be a net win.

Solves #806 and many more silent packed bugs.

### DIFF
--- a/src/core/sent_packet_metadata.h
+++ b/src/core/sent_packet_metadata.h
@@ -15,9 +15,6 @@ typedef struct QUIC_STREAM QUIC_STREAM;
 #define QUIC_SENT_FRAME_FLAG_STREAM_OPEN    0x01    // STREAM frame opened stream
 #define QUIC_SENT_FRAME_FLAG_STREAM_FIN     0x02    // STREAM frame included FIN bit
 
-#pragma pack(push)
-#pragma pack(1)
-
 //
 // Tracker for a sent frame.
 //
@@ -39,11 +36,6 @@ typedef struct QUIC_SENT_FRAME_METADATA {
         } CRYPTO;
         struct {
             QUIC_STREAM* Stream;
-            //
-            // TODO- optimization: encode in 32 bits.
-            //
-            uint64_t Offset;
-            uint16_t Length;
         } STREAM;
         struct {
             QUIC_STREAM* Stream;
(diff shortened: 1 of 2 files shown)


src/core/sent_packet_metadata.h
```diff
@@ -105,12 +106,21 @@ typedef struct QUIC_SENT_PACKET_METADATA {
     //
     // Frames included in this packet.
     //
-    uint8_t FrameCount;
+    uint8_t FrameCount  : 4;
+    uint8_t PathId      : 4;
```
nibanks, line 110: This isn't going to work. PathId is a constantly rotating value. We might have a max of X paths at one time, but paths come and go and this value constantly increases.

src/core/sent_packet_metadata.h
```diff
@@ -67,6 +59,16 @@ typedef struct QUIC_SENT_FRAME_METADATA {
...
+    //
+    // The following to fields are for STREAM. However, if they were in stream
+    // they force the union to completely contain them, which doesn't allow the
+    // Type and Flags fields to be packed nicely.
+    //
+    //
+    // TODO- optimization: encode in 32 bits.
+    //
+    uint64_t StreamOffset;
+    uint16_t StreamLength;
```
nibanks, line 71: Really, we only need ~12 bits here if we assert max packet size is < 4096.

nibanks:
If this is really the direction we want to go to fix these unaligned read/writes we should just try removing the packed as is and see how memory usage changes in tests like RPS. See what peak RPS is for 16K payload before and after this change.

ThadHouse:
> If this is really the direction we want to go to fix these unaligned read/writes we should just try removing the packed as is (no other changes for now, or at least no extra bit squashing) and see how memory usage changes in tests like RPS. See what peak RPS is for 16K payload before and after this change.

Sounds good. I'll leave the StreamOffset change in, since that isn't bit packing but does actually make a large noticeable difference.

nibanks (approved):

ThadHouse:
This looks to be well within the margin of error for perf. Assuming we can do the 32 bit optimization for offset, we'll even gain a lot back. But for now this looks to not hurt perf. Which makes sense, since the cost is 4 bytes per frame.

ThadHouse:
Just to leave as a note, the only packed struct remaining in production source code is the QUIC packet headers themselves, which need to be packed on purpose. There are a few in tests, and a few in the ETW logging, but otherwise we're back at a safe point.

PR #815, opened 2020-09-08
Update Partition size to 16 bits.

### LINKED ISSUE #809: Support more then 255 processor cores in public API
Currently, all public APIs dealing with processor counts and ideal processor numbers use an 8 bit integer, which limits to 255 processors in most cases. Systems are quickly getting more processor cores then that, and in the next few years it will be something we actively want to support. However, changing the variable size would be a breaking change, requiring new APIs to fix. We should at least make the public API allow passing and receiving at least a 16 bit integer, if not a 32 bit, so we can increase supported processor count and not break the API.

ThadHouse:
This will allow us to support up to 65k processors, which should be good for the forseeable future.

Needs to be tested a lot, especially if we can access that 256 core system again.

Closes #809

### DIFF
--- a/src/core/lookup.c
+++ b/src/core/lookup.c
@@ -79,7 +79,7 @@ _IRQL_requires_max_(DISPATCH_LEVEL)
 BOOLEAN
 QuicLookupCreateHashTable(
     _In_ QUIC_LOOKUP* Lookup,
-    _In_range_(>, 0) uint8_t PartitionCount
+    _In_range_(>, 0) uint16_t PartitionCount
     )
 {
     QUIC_DBG_ASSERT(Lookup->LookupTable == NULL);
@@ -90,8 +90,8 @@ QuicLookupCreateHashTable(
 
     if (Lookup->HASH.Tables != NULL) {
 
-        uint8_t Cleanup = 0;
-        for (uint8_t i = 0; i < PartitionCount; i++) {
+        uint16_t Cleanup = 0;
+        for (uint16_t i = 0; i < PartitionCount; i++) {
             if (!QuicHashtableInitializeEx(&Lookup->HASH.Tables[i].Table, QUIC_HASH_MIN_SIZE)) {
                 Cleanup = i;
                 break;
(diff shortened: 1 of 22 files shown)


src/core/lookup.c
```diff
@@ -360,8 +360,9 @@ QuicLookupFindConnectionByLocalCidInternal(
         // partitioned hash table array, and look up the connection in that
         // hash table.
         //
-        QUIC_STATIC_ASSERT(MSQUIC_CID_PID_LENGTH == 1, "The code below assumes 1 byte");
-        uint32_t PartitionIndex = CID[MsQuicLib.CidServerIdLength];
+        QUIC_STATIC_ASSERT(MSQUIC_CID_PID_LENGTH == 2, "The code below assumes 2 bytes");
+        uint32_t PartitionIndex;
```
nibanks, line 364: I think we can just change this to a `uint16_t`.
```suggestion
        uint16_t PartitionIndex;
```

src/core/lookup.c
```diff
@@ -478,8 +479,9 @@ QuicLookupInsertLocalCid(
         //
         // Insert the source connection ID into the hash table.
         //
-        QUIC_STATIC_ASSERT(MSQUIC_CID_PID_LENGTH == 1, "The code below assumes 1 byte");
-        uint32_t PartitionIndex = SourceCid->CID.Data[MsQuicLib.CidServerIdLength];
+        QUIC_STATIC_ASSERT(MSQUIC_CID_PID_LENGTH == 2, "The code below assumes 2 bytes");
+        uint32_t PartitionIndex;
```
nibanks, line 483: Ditto here and below.

src/core/worker.c
```diff
@@ -40,15 +40,15 @@ QUIC_STATUS
...
+    _In_ uint16_t IdealProcessor,
     _Inout_ QUIC_WORKER* Worker
     )
 {
     QUIC_STATUS Status;
 
     QuicTraceEvent(
         WorkerCreated,
-        "[wrkr][%p] Created, IdealProc=%hhu Owner=%p",
+        "[wrkr][%p] Created, IdealProc=%d Owner=%p",
```
nibanks, line 51:
```suggestion
        "[wrkr][%p] Created, IdealProc=%hu Owner=%p",
```

src/inc/msquic.h
```diff
@@ -372,7 +372,7 @@ typedef struct QUIC_SETTINGS {
     uint8_t DatagramReceiveEnabled  : 1;
     uint8_t ServerResumptionLevel   : 2;    // QUIC_SERVER_RESUMPTION_LEVEL
     uint8_t RESERVED                : 3;
-    uint8_t MaxPartitionCount;              // Global only
+    uint16_t MaxPartitionCount;              // Global only
```
nibanks, line 375:
```suggestion
    uint16_t MaxPartitionCount;             // Global only
```

nibanks (approved):

PR #816, opened 2020-09-09
Limit RPS tests to 8 cores less then max on systems with more then 60 cores.

ThadHouse:
It definitely seems to make the RPS tests more stable. Also, the threshold is high enough where it won't trigger on the lab machines.


nibanks (approved):

ThadHouse:
This still hit the RPS networking hang, but for the tests that actually ran, this seems to be about the magic number that fixes the timeouts, but does not affect performance. RPS was basically identical in User Mode. I'll rerun kernel on Monday.

ThadHouse:
Replaced by #838 which properly handles the extra cores.

PR #819, opened 2020-09-09
Clean up packet space in QuicConnFree

ThadHouse:
We have a leak, but its hard to detect where. By asserting that its already been cleaned up, we should be able to debug a bit easier

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -304,6 +304,9 @@ QuicConnFree(
                 AllStreamsLink);
         QUIC_DBG_ASSERTMSG(Stream != NULL, "Stream was leaked!");
     }
+    for (uint32_t i = 0; i < ARRAYSIZE(Connection->Packets); i++) {
+        QUIC_FRE_ASSERT(Connection->Packets[i] == NULL);
+    }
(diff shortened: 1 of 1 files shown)


nibanks (approved):

src/core/connection.c
```diff
@@ -304,6 +304,9 @@ QuicConnFree(
                 AllStreamsLink);
         QUIC_DBG_ASSERTMSG(Stream != NULL, "Stream was leaked!");
     }
+    for (uint32_t i = 0; i < ARRAYSIZE(Connection->Packets); i++) {
+        QUIC_FRE_ASSERT(Connection->Packets[i] == NULL);
```
nibanks, line 308: Actually, use a debug assert. We don't want a leak to actually bugcheck a production system.

ThadHouse: Its already in a #Debug

ThadHouse:
I fixed the actual bug by cleaning up the packet space in QuicConnFree in addition to QuicConnUninitialize. It does check, so this should still be ok.

nibanks (approved):

PR #828, opened 2020-09-11
Copy kernel mode data buffers if too much data is currently cached

ThadHouse:
When testing high throughput performance, we'd hit a point where msquic would be holding on to too many WSK buffers, which would cause resource exhaustion and dropped packets at the networking layer. Now, we'll only hold onto a limited set of buffers at one time, and if we receive more we'll grab a copy.

I need to remove all the yaml changes, and a few code cleanups, but the idea is there.


src/platform/datapath_winkernel.c
```diff
@@ -341,15 +355,17 @@ typedef struct QUIC_DATAPATH_PROC_CONTEXT {
 
     //
     // Pool of receive datagram contexts and buffers to be shared by all sockets
-    // on this core.
+    // on this core. Idx 0 is regular, Idx 1 is URO.
```
nibanks, line 358: nit: spell out `Index` here and below.

src/platform/datapath_winkernel.c
```diff
@@ -2034,8 +2101,34 @@ QuicDataPathSocketReceive(
                     goto Drop;
                 }
 
+                if (RecvContext->ProcContext->OutstandingStoredBytes > 256000) {
```
nibanks, line 2104: Create a macro for the magic number.

src/platform/datapath_winkernel.c
```diff
@@ -2034,8 +2101,34 @@ QuicDataPathSocketReceive(
...
+                    //
+                    // Perform a copy
+                    //
+                    RecvContext->IsCopiedBuffer = TRUE;
+                    RecvContext->BufferPoolIndex = DataLength > 4096 ? 1 : 0;
+                    RecvContext->DataBufferStart =
+                        (uint8_t*)QuicPoolAlloc(
+                            &RecvContext->ProcContext->RecvBufferPools[RecvContext->BufferPoolIndex]);
+                    if (RecvContext->DataBufferStart == NULL) {
```
nibanks, line 2113: Idea: should we pend if we can't alloc to copy? Or just drop

src/platform/datapath_winkernel.c
```diff
@@ -2087,8 +2187,6 @@ QuicDataPathSocketReceive(
                     Binding->Datapath->DatagramStride);
         }
 
-        continue;
-
```
nibanks, line 2091: Why did you remove this? If we're pending, shouldn't we be doing this?

src/platform/datapath_winkernel.c
```diff
@@ -2034,8 +2101,34 @@ QuicDataPathSocketReceive(
...
+                        goto Drop;
+                    }
+                    CurrentCopiedBuffer = RecvContext->DataBufferStart;
+                } else {
+                    RecvContext->IsCopiedBuffer = FALSE;
+                    RecvContext->DataIndication = DataIndication;
+                    QUIC_DBG_ASSERT(DataIndication->Next == NULL);
+                    RecvContext->DataIndicationSize = (int32_t)DataLength;
+                    InterlockedAdd64(
+                        &RecvContext->ProcContext->OutstandingStoredBytes,
```
nibanks, line 2127: I want to increment/decrement OutstandingStoredBytes regardless of if the bytes are pending or copied. I also want to be able to use this as an upper limit for number of bytes we're willing to pend before we start dropping.

src/platform/datapath_winkernel.c
ThadHouse: Since thats an OOM scenario, I'd drop. That way going OOM couldn't be an attack vector to lock up all kernel NBLs.

src/platform/datapath_winkernel.c
ThadHouse: No, we still need to hit one of the checks in the drop case to release the buffers we copied.

src/platform/datapath_winkernel.c
ThadHouse: We have to have this variable be a pending buffer only list. Otherwise as soon as we cross the limit, it would basically be impossible to get out of, because if the pending buffers get released, that might not be enough to bring the stored bytes below the limit, so everything stays copy forever, even though we aren't actually pending anything.

I'll rename it to OutstandingPendingBytes, and if we want to add a separate OutstandingBytesInFlight we can easily enough.

nibanks: That assumes we don't drain any outstanding pending or copied data before receiving more. That's not a valid assumption necessarily.

ThadHouse: Well its not just draining any. To get below the threshold you'd have to drain the entire thing. Remember the pending buffers are going to be at the front of the queue, since they were received first. So they'll get returned first, while the copied ones are still pending. Then if any more packets are received before more are processed, they'll be copied instead of pended because we'll still be above the limit in flight, even though in flight is all copied packets.

src/platform/datapath_winkernel.c
```diff
@@ -1786,27 +1825,57 @@ QuicDataPathBindingGetRemoteAddress(
...
+    }
+
+    QuicPoolFree(
+        &Context->ProcContext->RecvDatagramPools[Context->DatagramPoolIndex],
+        Context);
+    return DataIndication;
+}
+
+
+
```
nibanks, line 1878: nit: one space please

nibanks (approved):

PR #830, opened 2020-09-12
Handshakes Per Second Perf Test

nibanks:
Adds a new handshakes per second test with the goal of measuring the maximum handshakes that can be completed per second.

### DIFF
--- a/scripts/RemoteTests.json
+++ b/scripts/RemoteTests.json
@@ -179,5 +179,36 @@
         "RemoteReadyMatcher": "Started!",
         "ResultsMatcher": "Result: (.*) RPS",
         "Units": "RPS"
+    },
+    {
+        "TestName": "HPS",
+        "Remote": {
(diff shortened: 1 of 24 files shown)


ThadHouse:
This will require database updates and updates to the performance script before it will run properly. Since its a completely new test rather then updating an existing test.

nibanks:
> This will require database updates and updates to the performance script before it will run properly. Since its a completely new test rather then updating an existing test. I'll work on that this week.

But shouldn't I at least be able to run the test? I understand publishing won't work yet, but what about running?

ThadHouse:
https://github.com/microsoft/msquic/blob/ec6a3c00b7bfbf630536ebcdd7363df62916f1e0/scripts/performance-helper.psm1#L640

We should make this a write info rather then a write error. That will allow it to run.

ThadHouse (approved):

PR #834, opened 2020-09-14
Update Windows Usermode Datapath to Use Processor Context Index

nibanks:
No longer use current processor index to allocate receive datagram buffers, but the current proc context's index.
Also fixed up a number of things that assumed 1 byte processor index.

### DIFF
--- a/src/platform/datapath_winuser.c
+++ b/src/platform/datapath_winuser.c
@@ -227,7 +227,7 @@ typedef struct QUIC_DATAPATH_BINDING {
     //
     // The index of the affinitized receive processor for a connected socket.
     //
-    UINT8 ConnectedProcessorAffinity;
+    uint16_t ConnectedProcessorAffinity;
 
     //
     // Parent datapath.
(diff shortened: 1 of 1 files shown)


src/platform/datapath_winuser.c
```diff
@@ -565,7 +565,12 @@ QuicDataPathInitialize(
     WSADATA WsaData;
     QUIC_DATAPATH* Datapath;
     uint32_t DatapathLength;
+
     uint32_t MaxProcCount = QuicProcActiveCount();
+    QUIC_DBG_ASSERT(MaxProcCount <= UINT16_MAX);
+    if (MaxProcCount > UINT16_MAX) {
+        MaxProcCount = UINT16_MAX;
+    }
```
ThadHouse, line 573: Need to be 1 less then UINT16_MAX here. Otherwise all the loops with uint16_t control variables will infinite loop.

nibanks: Ah, good catch.

ThadHouse (approved):

PR #838, opened 2020-09-15
Fix perf crashes in RPS and HPS tests

ThadHouse:
Each had a different style of race that was either resulting in a use after free (HPS) or a double free (RPS)

Fixing the double free requires an API change, so that in a shutdown complete event, it can be determined if it was app
initiated or connection initiated


src/core/connection.h
```diff
@@ -147,6 +147,12 @@ typedef union QUIC_CONNECTION_STATE {
         //
         BOOLEAN ResumptionEnabled : 1;
 
+        //
+        // Indicates that an app shutdown from a non worker thread is in progress.
+        // Received by the QUIC_CONNECTION_EVENT_SHUTDOWN_COMPLETE event.
+        //
+        BOOLEAN AppShutdownInProgress: 1;
```
nibanks, line 154: "Shutdown" is a different concept. It's in Close right now. Call it `AppCloseInProgress` instead.

src/inc/msquic.h
```diff
@@ -786,6 +786,7 @@ typedef struct QUIC_CONNECTION_EVENT {
         struct {
             BOOLEAN HandshakeCompleted;
             BOOLEAN PeerAcknowledgedShutdown;
+            BOOLEAN AppShutdownInProgress;
```
nibanks, line 789: Let's go ahead and make this a bit field now that we have several.

src/perf/lib/HpsClient.h
```diff
@@ -33,7 +33,7 @@ struct HpsWorkerContext {
         QuicZeroMemory(&LocalAddrs, sizeof(LocalAddrs));
         QuicEventInitialize(&WakeEvent, FALSE, TRUE);
     }
-    ~HpsWorkerContext() {
+    void WaitForWorker() {
         if (ThreadStarted) {
```
nibanks, line 37: Don't do it this way. Instead make a new WaitForWorker function that just sets the event and waits on the thread. Keep the destructor as is.

src/inc/msquic.h
```diff
@@ -784,8 +784,9 @@ typedef struct QUIC_CONNECTION_EVENT {
             QUIC_UINT62 ErrorCode;
         } SHUTDOWN_INITIATED_BY_PEER;
         struct {
-            BOOLEAN HandshakeCompleted;
-            BOOLEAN PeerAcknowledgedShutdown;
+            BOOLEAN HandshakeCompleted          : 1;
+            BOOLEAN PeerAcknowledgedShutdown    : 1;
+            BOOLEAN AppCloseInProgress       : 1;
```
nibanks, line 789: nit: spacing

nibanks (approved):

nibanks (approved):

ThadHouse:
Closing just so I can run perf pipeline. Will reopen.

nibanks (approved):

PR #841, opened 2020-09-16
Refactor Ideal Processor Usage

nibanks:
Recent perf tests seem to indicate setting ideal processor actually hurts performance. This change refactors how we use that thread config flag, and removes it for default configurations.

### DIFF
--- a/src/inc/quic_platform_winkernel.h
+++ b/src/inc/quic_platform_winkernel.h
@@ -695,10 +695,6 @@ QuicSleep(
 // Create Thread Interfaces
 //
 
-#define QUIC_THREAD_FLAG_SET_IDEAL_PROC     0x0001
-#define QUIC_THREAD_FLAG_SET_AFFINITIZE     0x0002
-#define QUIC_THREAD_FLAG_HIGH_PRIORITY      0x0004
-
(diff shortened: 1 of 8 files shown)


src/inc/quic_platform_winkernel.h
```diff
@@ -752,56 +750,57 @@ QuicThreadCreate(
...
+                ThreadGroupInformation,
+                &Info.NumaNode.GroupMask,
+                sizeof(GROUP_AFFINITY));
+        QUIC_DBG_ASSERT(QUIC_SUCCEEDED(Status));
+        if (QUIC_FAILED(Status)) {
+            goto Cleanup;
+        }
+    }
+    if (Config->Flags & QUIC_THREAD_FLAG_SET_IDEAL_PROC) &&
+        !(Config->Flags & QUIC_THREAD_FLAG_SET_AFFINITIZE) {
```
nibanks, line 803: Maybe we don't need to include the !QUIC_THREAD_FLAG_SET_AFFINITIZE?

nibanks: Nevermind.

ThadHouse (approved):

PR #842, opened 2020-09-16
Add multifile support to quicinterop

anrossi:


src/tools/interop/interop.cpp
```diff
@@ -24,7 +24,7 @@ const QUIC_API_TABLE* MsQuic;
 HQUIC Registration;
 int EndpointIndex = -1;
 uint32_t TestCases = QuicTestFeatureAll;
-uint32_t WaitTimeoutMs = 5000;
+uint32_t WaitTimeoutMs = 10000;
```
nibanks, line 27: This really necessary?

src/tools/interop/interop.cpp
```diff
@@ -119,8 +119,8 @@ uint32_t CurrentThreadCount;
 
 uint16_t CustomPort = 0;
 
-const char* UrlPath = "/";
 bool CustomUrlPath = false;
+std::vector<const char*> Urls;
```
nibanks, line 123: If you're using `vector`, maybe worth while to use `string` too?

src/tools/interop/interop.cpp
```diff
@@ -1002,13 +1050,42 @@ main(
     TryGetValue(argc, argv, "timeout", &WaitTimeoutMs);
     TryGetValue(argc, argv, "version", &InitialVersion);
     TryGetValue(argc, argv, "port", &CustomPort);
-    if (TryGetValue(argc, argv, "urlpath", &UrlPath)) {
-        if (UrlPath[0] != '/' || strlen(UrlPath) == 1) {
-            printf("Invalid UrlPath! Must begin with '/'!\n");
-            Status = QUIC_STATUS_INVALID_PARAMETER;
-            goto Error;
+    //
+    // Get URLs from commandline.
```
nibanks, line 1054: Maybe put the following code in a helper function?

nibanks (approved):

src/tools/interop/interop.cpp
anrossi: I was wondering if part of the reason the high RTT/loss tests fail is that the file doesn't finish transferring within 5 seconds. Some of those files are over 25 MB.

src/tools/interop/interop.cpp
anrossi: What benefit would there be, though?

nibanks: you don't have to manage the memory?

src/tools/interop/interop.cpp
nibanks: Ok.

PR #852, opened 2020-09-21
Add platform independent enum for address family

### LINKED ISSUE #404: Use platform independent enum for address family

ThadHouse:
To make interop easier, and to be a better OSS library, use a QUIC defined enum for AddressFamily rather then the platform specific ones.

Values have been standardized to Windows values to make windows free.

Closes #404


src/test/lib/ApiTest.cpp
```diff
@@ -600,7 +600,7 @@ void QuicTestValidateConnection()
             QUIC_STATUS_INVALID_PARAMETER,
             MsQuic->ConnectionStart(
                 Connection.Handle,
-                AF_MAX,
+                (QUIC_ADDRESS_FAMILY)AF_MAX,
```
nibanks, line 603: Maybe we should just use something like `127` instead of `AF_MAX` here.

nibanks (approved):

nibanks:
If Windows code passes `AF_INET` will it get a compile break?

ThadHouse:
Yes it will be a breaking change, so something we'd need to go in and update.

nibanks:
> Yes it will be a breaking change, so something we'd need to go in and update.

I'd like to avoid that if possible. Can we "just make it work" as is?

ThadHouse:
No, not really. ConnectionStart takes a QUIC_ADDRESS_FAMILY, but I can't redefine AF_INET and friends if I make QUIC_ADDRESS_FAMILY a new enum rather then just a typedef. I can make QUIC_ADDRESS_FAMILY_INET and the new items macros, but thats really ugly and I don't like that.

nibanks:
> No, not really. ConnectionStart takes a QUIC_ADDRESS_FAMILY, but I can't redefine AF_INET and friends if I make QUIC_ADDRESS_FAMILY a new enum rather then just a typedef. I can make QUIC_ADDRESS_FAMILY_INET and the new items macros, but thats really ugly and I don't like that.

So, I guess the simplest solution will just be to add C casts in the existing Windows code?

ThadHouse:
Yeah. This is part of where this PR is really ugly. QUIC_ADDR is storing values as an ADDRESS_FAMILY, but we're kind of expecting them as a QUIC_ADDRESS_FAMILY. On windows this isn't a huge deal except for casts, but on linux this is going to get really funky.

One thing I did notice though is that a QUIC_ADDR is never actually passed up or down directly to the socket in any data path, since it always uses the msg structs. So, instead of typedeffing QUIC_ADDR per platform, it looks like we could actually make a platform independent struct all the way down. That would solve a lot of these issues, and I think might actually be the best solution, better then this PR even, which is really kind of hacky. I'll try that out to see if I am correct on QUIC_ADDR never actually being passed.

nibanks (approved):

nibanks (approved):

nibanks (approved):

PR #853, opened 2020-09-22
Use certs provided by interop framework

larseggert:
**FIXME:** There doesn't seem to be a way to specify to use `/certs/ca.pem` for certificate verification by the `quicinterop` client?


anrossi:
I think the reason for that is that we don't have a unified way (yet) to validate the server certificate, so the interopclient just disables certificate validation.

anrossi (approved):
Looks good to me!

nibanks (approved):

PR #854, opened 2020-09-22
Improve QNS Pipeline

nibanks:
- Deletes some old, unnecessary pipeline files.
- Updates QNS pipeline/template to not run measurements by default
- Also cleans up how the `build` arg works for the template, to reduce complexity.


.azure/azure-pipelines.qns.yml
```diff
@@ -48,10 +52,11 @@ jobs:
     ${{ if parameters.build }}:
       dependsOn: publish_docker
     timeout: ${{ parameters.timeout }}
+    measurements: ${{ parameters.measurements }}
+    ${{ if parameters.build }}:
```
anrossi, line 56: This doesn't make sense unless the client or server is msquic.

anrossi (approved):

PR #856, opened 2020-09-23
Improving Interop Runner Logging

nibanks:
Replaces streaming logging with normal start/stop model. Client seems to be working but server is not quite yet.


scripts/run_endpoint.sh
```diff
@@ -81,5 +99,10 @@ elif [ "$ROLE" == "server" ]; then
     esac
 
     quicinteropserver ${SERVER_PARAMS} -root:/www -listen:* -port:443 \
-        -file:/certs/cert.pem -key:/certs/priv.key 2>&1
+        -file:/certs/cert.pem -key:/certs/priv.key &
```
nibanks, line 102: For some reason, the `&` doesn't seem to be working as expected. The `getchar()` call in there isn't blocking as expected. Anyone have any ideas why that'd be?

.azure/templates/run-qns.yml
```diff
@@ -35,7 +35,8 @@ jobs:
 
         - script: |
             sudo apt-get install -y python3-setuptools
-            git clone "https://github.com/marten-seemann/quic-interop-runner.git"
+            git clone --single-branch --branch separate-docker-compose-runs \
```
nibanks, line 38: This is temporary until @marten-seemann merges these changes.

scripts/run_endpoint.sh
ThadHouse: I know we ran into the same issue in the perf setup, where when running in the background stdin wouldn't block. We solved that with the udp explicit shutdown function.

nibanks: Ok. Just pushed a possible solution (`-noexit`).

nibanks:
Closing this PR since it's not making any progress. I'll keep the branch around though, just in case...

PR #866, opened 2020-09-28
MsQuic Deployment Documentation

nibanks:
This adds some deployment documentation on various topics that anyone deploying an MsQuic based solution should be aware of.


docs/Deployment.md
```diff
@@ -0,0 +1,94 @@
...
+On Windows, these settings can set via the registry. The main registry path for the keys is:
+
+> HKLM:\System\CurrentControlSet\Services\MsQuic\Parameters
+
+The settings can also be set per "app-name" (as indicated in `RegistrationOpen`):
+
+> HKLM:\System\CurrentControlSet\Services\MsQuic\Parameters\app-name
+
+The `DWord` type should be used for all 32-bit or less types. For 64-bit types, `DWord` or `QWord` may be used. If invalid types or values are used, they will be ignored and the built-in default will be used instead.
+
```
nibanks, line 32: Let's add an example PowerShell command to set a value.

docs/Deployment.md
```diff
@@ -0,0 +1,94 @@
...
+| Max Partition Count      | uint16_t | MaxPartitionCount      | The maximum processor count used for partitioning work in MsQuic                                      | Yes                 |
+| Max Operations per Drain | uint8_t  | MaxOperationsPerDrain  | The maximum number of operations to drain per connection quantum                                      | No                  |
+| Retry Memory Limit       | uint16_t | RetryMemoryFraction    | The percentage of available memory usable for handshake<br>connections before stateless retry is used | No                  |
+| Max Worker Queue Delay   | uint32_t | MaxWorkerQueueDelayMs  | The maximum queue delay (in ms) allowed for a worker thread                                           | No                  |
+| Max Stateless Operations | uint32_t | MaxStatelessOperations | The maximum number of stateless operations that may be queued at<br>any one time                      | No                  |
+| Initial Window Size      | uint32_t | InitialWindowPackets   | The size (in packets) of the initial congestion window for a<br>connection                            | No                  |
```
ThadHouse, line 16: We shouldn't explicitly put breaks here, and instead depend on the renderer to do this correctly. On my system this looks REALLY bad.
DeploymentMarkdown

docs/Deployment.md
```diff
@@ -0,0 +1,94 @@
...
+
+On Windows, these settings can set via the registry. The main registry path for the keys is:
+
+> HKLM:\System\CurrentControlSet\Services\MsQuic\Parameters
+
+The settings can also be set per "app-name" (as indicated in `RegistrationOpen`):
+
+> HKLM:\System\CurrentControlSet\Services\MsQuic\Parameters\app-name
+
+The `DWord` type should be used for all 32-bit or less types. For 64-bit types, `DWord` or `QWord` may be used. If invalid types or values are used, they will be ignored and the built-in default will be used instead.
```
ThadHouse, line 31: Probably all caps for DWORD and QWORD. The casing here just looks really odd especially compared to all other documentation.

docs/Deployment.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Windows
+
+In order to configure the Windows firewall to allow inbound QUIC traffic efficiently, use a command such as the one below.
+
+```PowerShell
+New-NetFirewallRule -DisplayName "Allow QUIC" -Direction Inbound -Protocol UDP -LocalPort 433 -Action Allow -LocalOnlyMapping $true
+```
+
+Note the use of the `-LocalOnlyMapping $true` argument. This is a performance optimizing feature that should be used for UDP based protocols (like QUIC).
```
ThadHouse, line 53: Might be good to explain what this does rather then just saying its a performance optimization. Just saying set this flag for better perf isn't usually helpful.

nibanks: I added a link to MSDN. They don't really same much either. I can't actually tell you what this does myself...

ThadHouse: Ok. Well at least now it points to where info would be.

ThadHouse (approved):

docs/Deployment.md
```diff
@@ -0,0 +1,99 @@
...
+| Max Operations per Drain | uint8_t  | MaxOperationsPerDrain  | The maximum number of operations to drain per connection quantum                                      | No                  |
+| Retry Memory Limit       | uint16_t | RetryMemoryFraction    | The percentage of available memory usable for handshake connections before stateless retry is used    | No                  |
+| Max Worker Queue Delay   | uint32_t | MaxWorkerQueueDelayMs  | The maximum queue delay (in ms) allowed for a worker thread                                           | No                  |
+| Max Stateless Operations | uint32_t | MaxStatelessOperations | The maximum number of stateless operations that may be queued at any one time                         | No                  |
+| Initial Window Size      | uint32_t | InitialWindowPackets   | The size (in packets) of the initial congestion window for a connection                               | No                  |
+|                          |          |                        |                                                                                                       |                    |
+
+**TODO** - Finish list above
```
pravb, line 19: We should add a line that we have picked the best defaults and changing these should only be done after due diligence and A/B testing.

docs/Deployment.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Windows
+
+On Windows, these settings can set via the registry. The main registry path for the keys is:
+
```
pravb, line 24: please mention what happens on build to build upgrades. also mention that these only impact new connections?

docs/Deployment.md
```diff
@@ -0,0 +1,99 @@
...
+```PowerShell
+Enable-TlsCipherSuite -Name TLS_CHACHA20_POLY1305_SHA256
+```
+
+# Firewall
+
+## Windows
+
+In order to configure the Windows firewall to allow inbound QUIC traffic efficiently, use a command such as the one below.
+
```
pravb, line 53: this should only be done when not using IIS? Please clarify when this should be done and when not.

docs/Deployment.md
```diff
@@ -0,0 +1,99 @@
...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|  First octet  |             Server ID (X=8..152)              |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                        Any (0..152-X)                         |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+```
+
+`Server ID` is 4 bytes long, as encodes the complete IPv4 address OR the last 4 bytes of the IPv6 address.
+
+This encoding is **not enabled by default**. To configure it, set the `LoadBalancingMode` setting to `1`. **Note** - The server must be restarted for this setting to take effect.
```
pravb, line 76: add that a LB that supports QUIC is needed for this to work.

docs/Deployment.md
```diff
@@ -0,0 +1,99 @@
+# Deploying MsQuic
+
```
pravb, line 2: i would like to see some discussion of keep-alives and when and how to configure them. i'd also like you to talk about migration support on server and how that can be enabled in conjunction with a load balancer that supports it.

docs/InteropServerSetup.md
```diff
@@ -37,6 +37,8 @@ Examples:
   quicinteropserver -listen:* -retry:1 -thumbprint:175342733b39d81c997817296c9b691172ca6b6e -root:c:\temp
 ```
 
+Please see [Deployment.md](Deployment.md) for additional deployment considerations.
+
```
pravb, line 41: there should be a section on diagnostics - perfmon / tracing instructions etc.

nibanks:
@pravb I will make a new PR to address all these great comments. Thanks!

PR #878, opened 2020-10-01
Some Attack Tool Fixes and Improvements

nibanks:
- Fix the cred config args.
- Pre-generate the CRYPTO frame once.

### DIFF
--- a/src/tools/attack/attack.cpp
+++ b/src/tools/attack/attack.cpp
@@ -27,14 +27,16 @@
 #define ATTACK_PORT_DEFAULT 443
 
 static QUIC_DATAPATH* Datapath;
+static PacketWriter* Writer;
 
 static uint32_t AttackType;
 static const char* ServerName;
 static const char* IpAddress;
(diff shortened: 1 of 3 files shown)


src/tools/attack/attack.cpp
```diff
@@ -27,14 +27,16 @@
...
 
 static uint32_t AttackType;
 static const char* ServerName;
 static const char* IpAddress;
 static QUIC_ADDR ServerAddress;
-static const char* Alpn = "h3-31";
 static uint64_t TimeoutMs = ATTACK_TIMEOUT_DEFAULT_MS;
 static uint32_t ThreadCount = ATTACK_THREADS_DEFAULT;
+static const char* Alpn = "h3-29";
+static uint32_t Version = QUIC_VERSION_DRAFT_29;
```
nibanks, line 39: FYI, defaulting to -29 since that's the most supported version out there right now.

ThadHouse (approved):

PR #883, opened 2020-10-02
Handle receipt of a Version Negotiation packet better.

anrossi:


src/core/connection.c
```diff
@@ -3116,7 +3116,7 @@ QuicConnRecvHeader(
 
     if (!Packet->IsShortHeader) {
         if (Packet->Invariant->LONG_HDR.Version != Connection->Stats.QuicVersion) {
-            if (Packet->Invariant->LONG_HDR.Version == QUIC_VERSION_VER_NEG) {
+            if (Packet->Invariant->LONG_HDR.Version != QUIC_VERSION_LATEST) {
```
nibanks, line 3119: This isn't how this is supposed to work. Only `QUIC_VERSION_VER_NEG` is a version negotiation packet. It will contain a list of server supported version numbers. If that list includes `Connection->Stats.QuicVersion` then we should just ignore/drop the packet.

src/core/connection.c
```diff
@@ -2856,16 +2857,16 @@ QuicConnRecvVerNeg(
         //
         QuicConnCloseLocally(
```
nibanks, line 2858: You need to close regardless of SupportedVersion being zero or not, because we don't support version negotiation yet.

anrossi: Ah, but then we'll continue to fail the handshake corruption tests in QNS.

nibanks: There's not much else you can do if you receive a VN packet that doesn't include your requested version.

nibanks (approved):

PR #885, opened 2020-10-02
Some Initial Diagnostics Documentation

nibanks:
Still very minimal. Lots still left to add.

### DIFF
--- a/docs/Diagnostics.md
+++ b/docs/Diagnostics.md
@@ -1,4 +1,75 @@
-# Diagnosing Issues using MsQuic
+# Diagnosing Issues with MsQuic
 
-TODO
+This document describes various ways to debug and diagnose issues when using MsQuic.
 
+# Logging
+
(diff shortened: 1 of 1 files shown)


docs/Diagnostics.md
```diff
@@ -1,4 +1,75 @@
...
+
+```
+lttng stop msquic
+```
+
+To convert the trace, you can use the following commands:
+
+```
+babeltrace --names all ./msquic_lttng/* > quic.babel.txt
+clog2text_lttng -i quic.babel.txt -s clog.sidecar -o quic.log --showTimestamp --showCpuInfo
```
ThadHouse, line 57: Probably need it documented that you need clog2text_lttng installed. Just pointing to prepare-machine.ps1 is probably fine.

nibanks: Yeah, there's a lot more that needs to be added. For now, I don't want to add any more. If you're feeling up to it, feel free to add more commits to this PR/branch.

PR #887, opened 2020-10-05
Work Around BCrypt IV Issues

nibanks:
Apparently, even though the IV is 12 bytes, BCrypt will read/write the buffer up to the block size (!) so we need to make sure we allocate enough space on the stack.

### DIFF
--- a/src/core/binding.h
+++ b/src/core/binding.h
@@ -466,7 +466,7 @@ QuicRetryTokenDecrypt(
     //
     QuicCopyMemory(Token, TokenBuffer, sizeof(QUIC_RETRY_TOKEN_CONTENTS));
 
-    uint8_t Iv[QUIC_IV_LENGTH];
+    uint8_t Iv[QUIC_MAX_IV_LENGTH];
     if (MsQuicLib.CidTotalLength >= sizeof(Iv)) {
         QuicCopyMemory(Iv, Packet->DestCid, sizeof(Iv));
         for (uint8_t i = sizeof(Iv); i < MsQuicLib.CidTotalLength; ++i) {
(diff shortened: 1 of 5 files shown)


src/core/binding.h
```diff
@@ -466,14 +466,14 @@ QuicRetryTokenDecrypt(
...
-            Iv[i % sizeof(Iv)] ^= Packet->DestCid[i];
+    uint8_t Iv[QUIC_MAX_IV_LENGTH];
+    if (MsQuicLib.CidTotalLength >= QUIC_IV_LENGTH) {
+        QuicCopyMemory(Iv, Packet->DestCid, QUIC_IV_LENGTH);
+        for (uint8_t i = QUIC_IV_LENGTH; i < MsQuicLib.CidTotalLength; ++i) {
+            Iv[i % QUIC_IV_LENGTH] ^= Packet->DestCid[i];
         }
     } else {
-        QuicZeroMemory(Iv, sizeof(Iv));
+        QuicZeroMemory(Iv, QUIC_IV_LENGTH);
```
anrossi, line 476: Maybe we should zero the whole thing in case the garbage is being used internally?

nibanks: Everything is working as is already. The only way we knew about this was some Windows ASAN code @maximus- was playing around with. I'd prefer to **not** add any more CPU cycles if we don't have to, especially considering crypto is the biggest bottleneck already.

anrossi (approved):

src/inc/quic_crypt.h
```diff
@@ -40,6 +40,11 @@ typedef struct QUIC_HASH QUIC_HASH;
 //
 #define QUIC_IV_LENGTH 12
 
+//
+// The maximum buffer length of the IV need by the platform layer.
+//
+#define QUIC_MAX_IV_LENGTH 48
```
nibanks, line 46: Should we special case this to the Windows platform, since BCrypt is the only one that has this behavior(bug)? Everywhere else, `QUIC_IV_LENGTH` suffices.

anrossi: I would prefer that since this is a platform specific workaround.

PR #890, opened 2020-10-07
Version Negotiation Cleanup

nibanks:
- Fixes an issue where the server would only send a Version Negotiation packet if it matching V1's CID requirements. V2 might have different requirements...
- Some other general clean up of code, logging and comments.

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -2828,11 +2828,13 @@ QuicConnRecvVerNeg(
         }
     }
 
-    if (SupportedVersion != 0) {
-        //
-        // TODO - Remove once version negotiation extension support is added.
-        //
+    if (SupportedVersion != 0) { // TODO - Remove once version negotiation extension support is added.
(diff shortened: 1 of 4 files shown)


src/core/connection.c
```diff
@@ -2820,11 +2814,8 @@ QuicConnRecvVerNeg(
         // Check to see if this is the current version.
         //
         if (ServerVersionList[i] == Connection->Stats.QuicVersion) {
-            QuicTraceLogConnVerbose(
-                InvalidVerNeg,
-                Connection,
-                "Dropping version negotation that includes the current version");
-            goto Exit;
+            QuicPacketLogDrop(Connection, Packet, "Version Negotation that includes the current version");
```
anrossi, line 2817: shouldn't this be an error?

anrossi (approved):

src/core/connection.c
nibanks: No. Just a packet drop. Anything bad with an unauthenticated packet is just a drop.

PR #892, opened 2020-10-08
Ensure sending stream data also includes outstanding ACKs.

### LINKED ISSUE #865: Sending application data does not override ACK delay
This means that when an application sends data, it's not sending any pending ACKs, which are waiting for the ACK delay timer to expire before being sent.
This was discovered via interop client and could be affecting pass rates.

anrossi:
Fixes #865.

### DIFF
--- a/src/core/connection.h
+++ b/src/core/connection.h
@@ -254,6 +254,8 @@ typedef struct QUIC_CONN_STATS {
 
         uint64_t TotalBytes;            // Sum of UDP payloads
         uint64_t TotalStreamBytes;      // Sum of stream payloads
+
+        uint64_t AcksReceived;          // Count of receive ACK frames.
     } Recv;
 
     struct {
--- a/src/core/connection.h
+++ b/src/core/connection.h
@@ -251,11 +251,10 @@ typedef struct QUIC_CONN_STATS {
         uint64_t DuplicatePackets;
         uint64_t DecryptionFailures;    // Count of packets that failed to decrypt.
         uint64_t ValidPackets;          // Count of packets that successfully decrypted or had no encryption.
+        uint64_t ValidAckFrames;        // Count of receive ACK frames.
 
         uint64_t TotalBytes;            // Sum of UDP payloads
         uint64_t TotalStreamBytes;      // Sum of stream payloads
-
-        uint64_t AcksReceived;          // Count of receive ACK frames.
     } Recv;
 
     struct {
--- a/src/test/MsQuicTests.h
+++ b/src/test/MsQuicTests.h
@@ -588,5 +588,6 @@ typedef struct {
 
 #define IOCTL_QUIC_RUN_ACK_SEND_DELAY \
     QUIC_CTL_CODE(46, METHOD_BUFFERED, FILE_WRITE_DATA)
+    // int - Family
 
 #define QUIC_MAX_IOCTL_FUNC_CODE 46
--- a/src/test/MsQuicTests.h
+++ b/src/test/MsQuicTests.h
@@ -237,6 +237,7 @@ QuicTestReceiveResumeNoData(
 
 void
 QuicTestAckSendDelay(
+    _In_ int Family
     );
 
 //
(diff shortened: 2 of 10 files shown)


src/core/connection.h
```diff
@@ -254,6 +254,8 @@ typedef struct QUIC_CONN_STATS {
 
         uint64_t TotalBytes;            // Sum of UDP payloads
         uint64_t TotalStreamBytes;      // Sum of stream payloads
+
+        uint64_t AcksReceived;          // Count of receive ACK frames.
```
nibanks, line 258: Don't put Received in the name. Maybe, call it `ValidAckFrames` and put it after `ValidPackets`.

src/test/MsQuicTests.h
```diff
@@ -235,6 +235,10 @@ QuicTestReceiveResumeNoData(
     _In_ QUIC_RECEIVE_RESUME_SHUTDOWN_TYPE ShutdownType
     );
 
+void
+QuicTestAckSendDelay(
+);
```
nibanks, line 240: nit: tab over
```suggestion
    );
```

src/test/lib/DataTest.cpp
```diff
@@ -1857,3 +1857,302 @@ QuicTestReceiveResumeNoData(
...
+                nullptr);
+            if (QUIC_FAILED(Status)) {
+                TEST_FAILURE("Server failed to send to send data back 0x%x", Status);
+            }
+            break;
+        default:
+            break;
+        }
+    } else {
+        if(TestContext->ClientStream.Handle != QuicStream) {
```
nibanks, line 1912:
```suggestion
        if (TestContext->ClientStream.Handle != QuicStream) {
```

src/test/lib/DataTest.cpp
```diff
@@ -1857,3 +1857,302 @@ QuicTestReceiveResumeNoData(
...
+        case QUIC_STREAM_EVENT_RECEIVE:
+            Event->RECEIVE.TotalBufferLength = 0;
+            Status = MsQuic->StreamSend(
+                QuicStream,
+                TestContext->SendBuffer.Buffers,
+                TestContext->SendBuffer.BufferCount,
+                QUIC_SEND_FLAG_FIN,
+                nullptr);
+            if (QUIC_FAILED(Status)) {
+                TEST_FAILURE("Server failed to send to send data back 0x%x", Status);
```
nibanks, line 1905: `TEST_FAILURE` doesn't return. You should put explicit return calls after each of them in this function.

src/test/lib/DataTest.cpp
```diff
@@ -1857,3 +1857,302 @@ QuicTestReceiveResumeNoData(
...
+    uint32_t AckDelayMs = 1000;
+    MsQuicRegistration Registration;
+    TEST_TRUE(Registration.IsValid());
+
+    MsQuicAlpn Alpn("MsQuicTest");
+
+    MsQuicSettings Settings{};
+    Settings.SetIdleTimeoutMs(TimeoutMs);
+    Settings.MaxAckDelayMs = AckDelayMs;
+    Settings.IsSet.MaxAckDelayMs = true;
```
nibanks, line 2033: I feel like we should have a rule: If you add code to set a setting that doesn't have a helper yet, you get to add the helper. :D

src/test/lib/DataTest.cpp
```diff
@@ -1857,3 +1857,302 @@ QuicTestReceiveResumeNoData(
...
+    Settings.SetPeerBidiStreamCount(1);
+
+    MsQuicConfiguration ServerConfiguration(Registration, Alpn, Settings, SelfSignedCredConfig);
+    TEST_TRUE(ServerConfiguration.IsValid());
+
+    MsQuicCredentialConfig ClientCredConfig;
+    MsQuicConfiguration ClientConfiguration(Registration, Alpn, Settings, ClientCredConfig);
+    TEST_TRUE(ClientConfiguration.IsValid());
+
+    QUIC_ADDRESS_FAMILY QuicAddrFamily = (true) ? QUIC_ADDRESS_FAMILY_INET : QUIC_ADDRESS_FAMILY_INET6;
```
nibanks, line 2043: A TODO to add family as input to this test?

src/test/lib/DataTest.cpp
anrossi: Good rule!

nibanks (approved):

ThadHouse (approved):

PR #893, opened 2020-10-08
More Version Negotiation Improvements

nibanks:
- Fixes some of the logic around calls/usage of the `QuicIsVersionSupported` function. It no longer includes the version negotiation. Updates the one caller where it needed that to be included. All the rest were bugs.
- Updates the logic to validate the minimum UDP length before sending a VN packet.

### DIFF
--- a/src/core/quicdef.h
+++ b/src/core/quicdef.h
@@ -34,7 +34,7 @@ typedef struct QUIC_PATH QUIC_PATH;
 #define QUIC_MIN_INITIAL_PACKET_LENGTH          1200
 
 //
-// The minimum UDP payload size across all supported versions. Used to decided
+// The minimum UDP payload size across all supported versions. Used to decide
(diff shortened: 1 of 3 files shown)


src/core/quicdef.h
```diff
@@ -28,10 +28,18 @@ typedef struct QUIC_PATH QUIC_PATH;
...
 
 //
-// The minimum QUIC Packet Size (UDP payload size) for initial QUIC packets.
+// The minimum (version 1) QUIC Packet Size (UDP payload size) for initial QUIC
+// packets.
 //
 #define QUIC_MIN_INITIAL_PACKET_LENGTH          1200
 
+//
+// The minimum UDP payload size across all supported versions. Used to decided
```
anrossi, line 37: decide

anrossi (approved):

PR #903, opened 2020-10-12
Properly count protocol errors in perf counters.

### LINKED ISSUE #889: For performance counters, only treat known protocol errors as protocol errors
Currently, anything not timeout or idle is considered a protocol error to the perf counters. This is not ideal, and instead we should only count actual defined protocol errors.

ThadHouse:
Before, anything not a few small errors was counted. Now, only things that are explicitly protocol errors are counted as protocol errors.

Closes #889

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -1567,17 +1567,12 @@ QuicConnTryClose(
         if (ResultQuicStatus) {
             Connection->CloseStatus = (QUIC_STATUS)ErrorCode;
             Connection->CloseErrorCode = QUIC_ERROR_INTERNAL_ERROR;
-            //
-            // Errors 0x1-0xF are quic errors, as are 0x100-0xFF. For future proofing,
-            // consider the whole range of 0x1-0x1FF as a protocol error.
-            //
-            if (ErrorCode > 0 &&
-                ErrorCode <= 0x1FF) {
-                QuicPerfCounterIncrement(QUIC_PERF_COUNTER_CONN_PROTOCOL_ERRORS);
-            }
         } else {
             Connection->CloseStatus = QuicErrorCodeToStatus(ErrorCode);
             Connection->CloseErrorCode = ErrorCode;
+            if (QuicErrorIsProtocolError(ErrorCode)) {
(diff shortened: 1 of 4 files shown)


src/core/connection.c
```diff
@@ -1567,8 +1567,12 @@ QuicConnTryClose(
         if (ResultQuicStatus) {
             Connection->CloseStatus = (QUIC_STATUS)ErrorCode;
             Connection->CloseErrorCode = QUIC_ERROR_INTERNAL_ERROR;
-            if (ErrorCode != QUIC_STATUS_CONNECTION_IDLE &&
-                ErrorCode != QUIC_STATUS_CONNECTION_TIMEOUT) {
+            //
+            // Errors 0x1-0xF are quic errors, as are 0x100-0xFF. For future proofing,
+            // consider the whole range of 0x1-0x1FF as a protocol error.
+            //
+            if (ErrorCode > 0 &&
```
nibanks, line 1574: Notice, we're in the `if (ResultQuicStatus)` block, so `ErrorCode` is not a QUIC protocol error code. Additionally, not all QUIC protocol error codes should be treated as a `QUIC_PERF_COUNTER_CONN_PROTOCOL_ERRORS`. For now, I think 3 through E (inclusive) should be considered protocol errors, as far as this perf counter goes. We should make an inline function `QuicErrorIsProtocolError(...)` right after the error code definitions in frame.h, and put a comment by the function indicating this is for performance counter.

ThadHouse: Ok. So this should be happening in the else side of that? Adding the function.

src/core/frame.h
```diff
@@ -89,6 +89,17 @@
 #define QUIC_ERROR_CRYPTO_HANDSHAKE_FAILURE         QUIC_ERROR_CRYPTO_ERROR(40)  // TLS error code for 'handshake_failure'
 #define QUIC_ERROR_CRYPTO_NO_APPLICATION_PROTOCOL   QUIC_ERROR_CRYPTO_ERROR(120) // TLS error code for 'no_application_protocol'
 
+inline
+BOOLEAN
+QuicErrorIsProtocolError(
```
nibanks, line 94: Put a comment here indicating this is used for performance counters only.

nibanks (approved):
Just add the comment and this is good.

PR #908, opened 2020-10-13
Don't Hardcode Interop Server Name

nibanks:
Recent change in the interop runner broke our hardcoded server name

### DIFF
--- a/scripts/run_endpoint.sh
+++ b/scripts/run_endpoint.sh
@@ -61,14 +61,21 @@ if [ "$ROLE" == "client" ]; then
         ;;
     esac
 
+    # Figure out the server name from the first request. This assumes all URLS
+    # point to the same server.
+    REQS=($REQUESTS)
+    REQ=${REQS[0]}
+    SERVER=$(echo $REQ | cut -d'/' -f3)
+    echo "Connecting to $SERVER"
+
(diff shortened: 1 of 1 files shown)


ThadHouse (approved):

scripts/run_endpoint.sh
```diff
@@ -61,14 +61,21 @@ if [ "$ROLE" == "client" ]; then
...
+    # point to the same server.
+    REQS=($REQUESTS)
+    REQ=${REQS[0]}
+    SERVER=$(echo $REQ | cut -d'/' -f3)
+    echo "Connecting to $SERVER"
+
     if [ "$TESTCASE" == "multiconnect" ]; then
         for REQ in $REQUESTS; do
-            quicinterop ${CLIENT_PARAMS} -custom:server -port:443 -urls:"$REQ" -version:-16777187
+            quicinterop ${CLIENT_PARAMS} -custom:$SERVER -port:443 -urls:"$REQ" -version:-16777187
```
anrossi, line 73: Why are we setting the version explicitly?

anrossi (approved):
Looks good!

scripts/run_endpoint.sh
nibanks: We're hard coding to -29, since that's what the interop runner implementations assume.

Issue #752, opened 2020-08-20
OOM Crashes


Maximus-:
I implemented the faulting-heap that returns NULL ~1/100th of the time. The patch is attached. These issues are only relevant in OOM scenarios or other resiliency support scenarios. 

There's a few crashes here so I don't spam the issue tracker. 

These are spinquic:
```
Stopped reason: SIGSEGV
QuicLibraryGetParam (Handle=0x61e000057080, Level=QUIC_PARAM_LEVEL_LISTENER, Param=0x0, BufferLength=0x7fffe7059ad0, Buffer=0x0) at /home/max/msquic/src/core/library.c:940
940             Registration = Session->Registration;
gdb-peda$ bt
#0  QuicLibraryGetParam (Handle=0x61e000057080, Level=QUIC_PARAM_LEVEL_LISTENER, Param=0x0, BufferLength=0x7fffe7059ad0, Buffer=0x0) at /home/max/msquic/src/core/library.c:940
#1  0x00007ffff7c4223c in QuicConnProcessApiOperation (Connection=0x61e000057080, ApiCtx=0x7fffe7059870) at /home/max/msquic/src/core/connection.c:6498
#2  0x00007ffff7c42ce4 in QuicConnDrainOperations (Connection=0x61e000057080) at /home/max/msquic/src/core/connection.c:6602
#3  0x00007ffff7c1fbb6 in QuicWorkerProcessConnection (Worker=0x620000000ac8, Connection=0x61e000057080) at /home/max/msquic/src/core/worker.c:492
#4  0x00007ffff7c1dbfb in QuicWorkerThread (Context=0x620000000ac8) at /home/max/msquic/src/core/worker.c:580
#5  0x00007ffff7859fa3 in start_thread (arg=<optimized out>) at pthread_create.c:486
#6  0x00007ffff77644cf in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

```
#0  __GI_raise (sig=sig@entry=0x6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff768d535 in __GI_abort () at abort.c:79
#2  0x00007ffff7c91deb in quic_bugcheck () at /home/max/msquic/src/platform/platform_linux.c:60
[...]

PR #909, opened 2020-10-13
Support Random Memory Allocation Failures (For Testing)

nibanks:
Related to #752.

- Adds a build flag to enable 1/100 allocation failure chance.
- Ran spinquic a few times and fixed a few bugs I found. Still not 100% fixed. Some asserts hit, but I'm not sure those asserts are valid in the case of OOM. Spinquic also hung on shutdown (which I haven't investigated yet).
- I only ran this on Windows.


Maximus- (approved):

anrossi:
This looks good, but we have AppVerifier support that accomplishes the same thing. Why not leverage that?

nibanks:
> This looks good, but we have AppVerifier support that accomplishes the same thing. Why not leverage that?

This is just meant for local execution, so you can definitely use AppVerifier when you run this.

PR #910, opened 2020-10-13
Remove a lot of options from openssl to decrease binary size

ThadHouse:
Binary size on windows is down to 2.7MB.

### DIFF
--- a/.azure/templates/build-config-user.yml
+++ b/.azure/templates/build-config-user.yml
@@ -26,7 +26,7 @@ jobs:
 
   - task: Cache@2
     inputs:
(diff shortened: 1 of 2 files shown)


nibanks:
Here's the diff between the CI built binaries:
```
Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a---          10/13/2020  4:21 PM        2866176 msquic.new.dll
-a---          10/13/2020  4:22 PM        3458560 msquic.old.dll
```

ThadHouse:
Ok. And thats pretty much everything I can remove. I'll play around with tomorrow to see if I can not do any of this. I'm thinking the algo's are not actually that big, and it might be worth having them just in case.

nibanks (approved):

PR #912, opened 2020-10-14
Pad Server Initial Packets

nibanks:
For https://github.com/quicwg/base-drafts/pull/4188. Ensures server side fails if MTU is too small.

### DIFF
--- a/src/core/packet_builder.c
+++ b/src/core/packet_builder.c
@@ -253,12 +253,12 @@ QuicPacketBuilderPrepare(
                     QuicAddrGetFamily(&Builder->Path->RemoteAddress),
                     QUIC_INITIAL_PACKET_LENGTH);
 
-            if (Builder->MinimumDatagramLength > Builder->DatagramLength) {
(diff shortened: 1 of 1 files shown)


src/core/packet_builder.c
```diff
@@ -243,17 +243,24 @@ QuicPacketBuilderPrepare(
...
             //
-            // Make sure to pad the packet if client Initial packets.
+            // Make sure to pad Initial packets.
             //
             Builder->MinimumDatagramLength =
                 MaxUdpPayloadSizeForFamily(
                     QuicAddrGetFamily(&Builder->Path->RemoteAddress),
                     QUIC_INITIAL_PACKET_LENGTH);
 
+            if ((uint32_t)Builder->MinimumDatagramLength > Builder->Datagram->Length) {
```
anrossi, line 256: Doesn't `Builder->Datagram->Length` refer to the length of the buffer?

nibanks: The allocated length of the datagram.

anrossi (approved):

PR #916, opened 2020-10-15
Fix interop client to support resumption tickets and zeroRTT.

anrossi:
The interop client support of 0-RTT and Resumption was broken when the refactor to remove the Session layer was completed.
This change adds that support back.

### DIFF
--- a/src/inc/msquic.h
+++ b/src/inc/msquic.h
@@ -522,7 +522,7 @@ typedef enum QUIC_PARAM_LEVEL {
 #ifdef QUIC_API_ENABLE_INSECURE_FEATURES
 #define QUIC_PARAM_CONN_DISABLE_1RTT_ENCRYPTION         15  // uint8_t (BOOLEAN)
 #endif
-#define QUIC_PARAM_CONN_RESUMPTION_STATE                16  // uint8_t[]
+#define QUIC_PARAM_CONN_RESUMPTION_TICKET               16  // uint8_t[]
 
 //
 // Parameters for QUIC_PARAM_LEVEL_TLS.
--- a/src/tools/interop/interop.cpp
+++ b/src/tools/interop/interop.cpp
@@ -415,9 +415,7 @@ public:
         QuicEventUninitialize(QuackAckReceived);
         QuicEventUninitialize(ConnectionComplete);
         delete [] NegotiatedAlpn;
-        if (ResumptionTicket) {
-            delete [] ResumptionTicket;
-        }
+        delete [] ResumptionTicket;
     }
     bool SetKeepAlive(uint32_t KeepAliveMs) {
         QUIC_SETTINGS Settings{0};
--- a/src/tools/interop/interop.cpp
+++ b/src/tools/interop/interop.cpp
@@ -342,7 +342,10 @@ class InteropConnection {
     QUIC_EVENT RequestComplete;
     QUIC_EVENT QuackAckReceived;
     QUIC_EVENT ShutdownComplete;
+    QUIC_EVENT TicketReceived;
     char* NegotiatedAlpn;
+    const uint8_t* ResumptionTicket;
+    uint32_t ResumptionTicketLength;
 public:
     bool VersionUnsupported : 1;
     bool Connected : 1;
@@ -352,6 +355,8 @@ public:
         Configuration(Configuration),
         Connection(nullptr),
         NegotiatedAlpn(nullptr),
+        ResumptionTicket(nullptr),
+        ResumptionTicketLength(0),
         VersionUnsupported(false),
         Connected(false),
         Resumed(false),
@@ -361,6 +366,7 @@ public:
         QuicEventInitialize(&RequestComplete, TRUE, FALSE);
         QuicEventInitialize(&QuackAckReceived, TRUE, FALSE);
         QuicEventInitialize(&ShutdownComplete, TRUE, FALSE);
+        QuicEventInitialize(&TicketReceived, TRUE, FALSE);
 
         VERIFY_QUIC_SUCCESS(
             MsQuic->ConnectionOpen(
@@ -403,11 +409,15 @@ public:
         Streams.clear();
         Shutdown();
         MsQuic->ConnectionClose(Connection);
+        QuicEventUninitialize(TicketReceived);
         QuicEventUninitialize(ShutdownComplete);
         QuicEventUninitialize(RequestComplete);
         QuicEventUninitialize(QuackAckReceived);
         QuicEventUninitialize(ConnectionComplete);
         delete [] NegotiatedAlpn;
+        if (ResumptionTicket) {
+            delete [] ResumptionTicket;
+        }
     }
     bool SetKeepAlive(uint32_t KeepAliveMs) {
         QUIC_SETTINGS Settings{0};
@@ -434,6 +444,16 @@ public:
                     QUIC_PARAM_CONN_SETTINGS,
                     sizeof(Settings),
                     &Settings));
+    }
+    bool SetResumptionTicket(const uint8_t* Ticket, uint32_t TicketLength) {
+        return
+            QUIC_SUCCEEDED(
+                MsQuic->SetParam(
+                    Connection,
+                    QUIC_PARAM_LEVEL_CONNECTION,
+                    QUIC_PARAM_CONN_RESUMPTION_STATE,
+                    TicketLength,
+                    Ticket));
     }
     bool ConnectToServer(const char* ServerName, uint16_t ServerPort) {
         if (QUIC_SUCCEEDED(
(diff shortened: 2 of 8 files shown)


src/tools/interop/interop.cpp
```diff
@@ -757,14 +783,26 @@ RunInteropTest(
     case Resumption:
     case StatelessRetry:
     case PostQuantum: {
+        const uint8_t* ResumptionTicket = nullptr;
+        uint32_t ResumptionTicketLength = 0;
         if (Feature == Resumption) {
             InteropConnection Connection(Configuration);
             if (!Connection.ConnectToServer(Endpoint.ServerName, Port) ||
                 !Connection.WaitForTicket()) {
+                printf("Failed waiting for ticket\n");
```
anrossi, line 792: Delete this.

src/tools/interop/interop.cpp
```diff
@@ -757,14 +783,26 @@ RunInteropTest(
...
+                break;
+            }
+            if (!Connection.GetResumptionTicket(&ResumptionTicket, &ResumptionTicketLength)) {
+                printf("Failed to get resumption ticket\n");
```
anrossi, line 796: Delete this

src/tools/interop/interop.cpp
```diff
@@ -757,14 +783,26 @@ RunInteropTest(
...
                 break;
             }
         }
         InteropConnection Connection(Configuration, false, Feature == PostQuantum);
+        if (Feature == Resumption) {
+            if (!Connection.SetResumptionTicket(ResumptionTicket, ResumptionTicketLength)) {
+                printf("Failed to SET resumption ticket\n");
```
anrossi, line 803: Delete this

src/tools/interop/interop.cpp
```diff
@@ -647,6 +662,17 @@ class InteropConnection {
...
+        case QUIC_CONNECTION_EVENT_RESUMPTION_TICKET_RECEIVED:
+            if (Event->RESUMPTION_TICKET_RECEIVED.ResumptionTicketLength) {
+                pThis->ResumptionTicketLength = Event->RESUMPTION_TICKET_RECEIVED.ResumptionTicketLength;
+                pThis->ResumptionTicket = new uint8_t[pThis->ResumptionTicketLength];
+                memcpy(
+                    (uint8_t*)pThis->ResumptionTicket,
+                    Event->RESUMPTION_TICKET_RECEIVED.ResumptionTicket,
+                    Event->RESUMPTION_TICKET_RECEIVED.ResumptionTicketLength);
+            }
+            QuicEventSet(pThis->TicketReceived);
```
anrossi, line 674: Should this only be set in the case where the resumption ticket length is non-zero? The benefit is it might help us catch implementation bugs, but it'll look like the test timed out.

src/tools/interop/interop.cpp
```diff
@@ -435,6 +445,16 @@ class InteropConnection {
                     sizeof(Settings),
                     &Settings));
     }
+    bool SetResumptionTicket(const uint8_t* Ticket, uint32_t TicketLength) {
+        return
+            QUIC_SUCCEEDED(
+                MsQuic->SetParam(
+                    Connection,
+                    QUIC_PARAM_LEVEL_CONNECTION,
+                    QUIC_PARAM_CONN_RESUMPTION_STATE,
```
anrossi, line 454: I don't like this name. I want to rename it to something like "RESUMPTION TICKET". Thoughts?

nibanks: Yeah, might be good.

nibanks: It should probably just match the recv callback event for it.

src/tools/interop/interop.cpp
```diff
@@ -403,11 +409,15 @@ class InteropConnection {
...
         Shutdown();
         MsQuic->ConnectionClose(Connection);
+        QuicEventUninitialize(TicketReceived);
         QuicEventUninitialize(ShutdownComplete);
         QuicEventUninitialize(RequestComplete);
         QuicEventUninitialize(QuackAckReceived);
         QuicEventUninitialize(ConnectionComplete);
         delete [] NegotiatedAlpn;
+        if (ResumptionTicket) {
+            delete [] ResumptionTicket;
```
nibanks, line 419: delete allows a nullptr to be passed in. No need to check before you call it.

src/tools/interop/interop.cpp
nibanks: Why would a nonzero length ever be received? I don't think you should special case this.

src/tools/interop/interop.cpp
```diff
@@ -596,6 +602,15 @@ class InteropConnection {
         }
         return false;
     }
+    bool GetResumptionTicket(const uint8_t*& Ticket, uint32_t& TicketLength) {
+        if (!WaitForTicket() || ResumptionTicket == nullptr) {
+            return false;
+        }
+        Ticket = new uint8_t[ResumptionTicketLength];
```
nibanks, line 609: I don't think we need to copy. Just return the pointer and length.

nibanks: Oh, I see, you're freeing the connection before using this. In that case, return the buffer length, and null out the local pointer.

src/tools/interop/interop.cpp
```diff
@@ -784,19 +816,28 @@ RunInteropTest(
                 Success = Connection.SendHttpRequests();
             }
         }
+        if (ResumptionTicket) {
```
nibanks, line 819: No need for if check

src/tools/interop/interop.cpp
```diff
@@ -784,19 +816,28 @@ RunInteropTest(
...
+        uint32_t ResumptionTicketLength = 0;
         if (Feature == ZeroRtt) {
             InteropConnection Connection(Configuration);
             if (!Connection.ConnectToServer(Endpoint.ServerName, Port) ||
-                !Connection.WaitForTicket()) {
+                !Connection.WaitForTicket() ||
+                !Connection.GetResumptionTicket(ResumptionTicket, ResumptionTicketLength)) {
                 break;
             }
         }
```
nibanks, line 836: Might be worth just making a helper function for this to reduce duplication.

anrossi: If we have a third test case that needs to do this, I'll agree. But for now, this isn't so bad.

src/inc/msquic.h
```diff
@@ -522,7 +522,7 @@ typedef enum QUIC_PARAM_LEVEL {
 #ifdef QUIC_API_ENABLE_INSECURE_FEATURES
 #define QUIC_PARAM_CONN_DISABLE_1RTT_ENCRYPTION         15  // uint8_t (BOOLEAN)
 #endif
-#define QUIC_PARAM_CONN_RESUMPTION_STATE                16  // uint8_t[]
+#define QUIC_PARAM_CONN_RESUMPTION_TICKET               16  // uint8_t[]
```
nibanks, line 525: Double check there was no Windows code using this already.

src/tools/spin/spinquic.cpp
```diff
@@ -374,7 +374,7 @@ void SpinQuicSetRandomConnectionParam(HQUIC Connection)
     case QUIC_PARAM_CONN_DISABLE_1RTT_ENCRYPTION:                   // uint8_t (BOOLEAN)
         Helper.SetUint8(QUIC_PARAM_CONN_DISABLE_1RTT_ENCRYPTION, (uint8_t)GetRandom(2));
         break;
-    case QUIC_PARAM_CONN_RESUMPTION_STATE:                          // uint8_t[]
+    case QUIC_PARAM_CONN_RESUMPTION_TICKET:                          // uint8_t[]
```
nibanks, line 377: Fix comment alignment

src/tools/interop/interop.cpp
```diff
@@ -596,6 +602,15 @@ class InteropConnection {
...
         return false;
     }
+    bool GetResumptionTicket(const uint8_t*& Ticket, uint32_t& TicketLength) {
+        if (!WaitForTicket() || ResumptionTicket == nullptr) {
```
nibanks, line 606: ` || ResumptionTicket == nullptr` should be unnecessary, no?

src/tools/interop/interop.cpp
nibanks: You didn't delete this if check.

src/tools/interop/interop.cpp
anrossi: If this gets called twice, the second call will return null. Why not just fail the call?

src/inc/msquic.h
anrossi: I searched and there's no usages of this.

src/tools/interop/interop.cpp
anrossi: Missed one!

nibanks (approved):

PR #919, opened 2020-10-15
Properly close RPS tests upon failure

ThadHouse:
Previously, if RPS tests failed to initialize, the code would not close existing connections, causing registration close to hang. This will now explicitly close all connections on error, fixing this issue.


src/perf/lib/RpsClient.cpp
```diff
@@ -238,8 +241,15 @@ RpsClient::Start(
...
 
-
     return QUIC_STATUS_SUCCESS;
+
+Error:
+    for (uint32_t i = 0; i < ConnectionCount; i++) {
+        if (Connections[i] != nullptr) {
+            MsQuic->ConnectionClose(Connections[i]);
+        }
+    }
```
nibanks, line 251: Should we just put this in the destructor and remove it from the end of wait as well?

nibanks: That, plus don't even use a temporary variable `Connection`. Put it directly in `Connections[i]`, then we can return on failure and not need to close explicitly.

src/perf/lib/RpsClient.h
```diff
@@ -24,6 +24,13 @@ class RpsClient : public PerfBase {
     }
 
     ~RpsClient() override {
+        if (Connections != nullptr) {
```
nibanks, line 27: Now that I look at this, I guess Connections could be an array of `ConnectionScope` and then you don't have to do anything manually, right?

src/perf/lib/RpsClient.cpp
```diff
@@ -114,13 +114,13 @@ RpsClient::Start(
...
 
-    Connections = UniquePtr<HQUIC[]>(new(std::nothrow) HQUIC[ConnectionCount]);
+    Connections = UniquePtr<ConnectionScope[]>(new(std::nothrow) ConnectionScope[ConnectionCount]);
     if (!Connections.get()) {
         return QUIC_STATUS_OUT_OF_MEMORY;
     }
 
     for (uint32_t i = 0; i < ConnectionCount; i++) {
-        Connections[i] = nullptr;
+        Connections[i].Handle = nullptr;
```
nibanks, line 123: This for loop is unnecessary, no? It should call the constructor, which will null out the handle pointers.

src/perf/bin/drvmain.cpp
```diff
@@ -107,6 +107,12 @@ void __cdecl operator delete[] (_In_opt_ void* Mem) {
     }
 }
 
+void __cdecl operator delete[] (_In_opt_ void* Mem, _In_opt_ size_t) {
```
nibanks, line 110: Why is this just now needed?

src/perf/lib/RpsClient.cpp
```diff
@@ -145,7 +143,7 @@ RpsClient::Start(
                 Registration,
                 Handler,
                 this,
-                &Connection);
+                &Connections[i].Handle);
```
nibanks, line 146: If the HQUIC operator returned a HQUIC& then the .Handle wouldn't be necessary here, right?

src/perf/bin/drvmain.cpp
ThadHouse: Because the compiler used this overload since its not trivially destructible. Its allowed to pick whichever one.

src/perf/lib/RpsClient.cpp
ThadHouse: I would prefer not. Not a big fan of non const references.

src/perf/lib/RpsClient.cpp
ThadHouse: In theory yes, but I'm not sure I 100% trust that kernel mode will do that. I can double check.

src/perf/lib/RpsClient.cpp
nibanks: Ok. We can keep as is for now.

nibanks (approved):

PR #920, opened 2020-10-15
Properly set thread affinity for throughput tests

### LINKED ISSUE #917: User Mode Download Performance is Significantly Lower than Upload
### Describe the bug

For some reason, download perf is a lot worse than upload for user mode. We need to investigate why.

image

ThadHouse:
We were not setting this before, causing low performance for download performance tests. 

Should fix #917

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -65,7 +65,7 @@ QuicConnAlloc(
     )
 {
     BOOLEAN IsServer = Datagram != NULL;
-    uint32_t CurProcIndex = QuicProcCurrentNumber() & 0xFFFFFFFE;
+    uint32_t CurProcIndex = QuicProcCurrentNumber();
 
     //
     // For client, the datapath partitioning info is not known yet, so just use
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -65,7 +65,7 @@ QuicConnAlloc(
     )
 {
     BOOLEAN IsServer = Datagram != NULL;
-    uint32_t CurProcIndex = QuicProcCurrentNumber();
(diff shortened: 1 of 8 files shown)


ThadHouse:
Doing this slightly differently. Realizing we shouldn't affinitize this to the core, but instead should just affinitize it to thread group 0.

src/perf/lib/ThroughputClient.cpp
```diff
@@ -26,7 +26,6 @@ PrintHelp(
         "  -target:<####>              The target server to connect to.\n"
 #if _WIN32
         "  -comp:<####>                The compartment ID to run in.\n"
-        "  -core:<####>                The CPU core to use for the main thread.\n"
```
nibanks, line 29: I still think it'd be nice to support this flag for now.

nibanks (approved):

nibanks:
Looks like Windows builds don't like `Affinity.Mask = (KAFFINITY)(~0ull);` Maybe try `Affinity.Mask = (KAFFINITY)(-1);` instead?

src/core/connection.c
```diff
@@ -65,7 +65,7 @@ QuicConnAlloc(
     )
 {
     BOOLEAN IsServer = Datagram != NULL;
-    uint32_t CurProcIndex = QuicProcCurrentNumber();
+    uint32_t CurProcIndex = QuicProcCurrentNumber() & 0xFFFFFFFE;
```
nibanks: No. We cannot do this.

ThadHouse: I know, I was just trying it.

nibanks (approved):

PR #921, opened 2020-10-15
Fix OpenSSL Hash/HMAC Performance

nibanks:
Previously all Hash abstraction layer actions with OpenSSL were taking all CPU in HPS tests. The new code seems to fix that.

### DIFF
--- a/src/platform/tls_openssl.c
+++ b/src/platform/tls_openssl.c
@@ -1994,15 +1994,27 @@ QuicHashCompute(
     HMAC_CTX* HashContext = (HMAC_CTX*)Hash;
 
     if (!HMAC_Init_ex(HashContext, NULL, 0, NULL, NULL)) {
+        QuicTraceEvent(
+            LibraryError,
+            "[ lib] ERROR, %s.",
+            "HMAC_Init_ex(NULL) failed");
         return QUIC_STATUS_INTERNAL_ERROR;
     }
 
     if (!HMAC_Update(HashContext, Input, InputLength)) {
+        QuicTraceEvent(
+            LibraryError,
+            "[ lib] ERROR, %s.",
+            "HMAC_Update failed");
         return QUIC_STATUS_INTERNAL_ERROR;
     }
 
     uint32_t ActualOutputSize = OutputLength;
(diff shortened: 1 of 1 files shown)


src/platform/tls_openssl.c
```diff
@@ -2001,27 +1993,33 @@ QuicHashCompute(
     _Out_writes_all_(OutputLength) uint8_t* const Output
     )
 {
-    QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
+    HMAC_CTX* HashContext = (HMAC_CTX*)Hash;
 
-    if (!EVP_DigestSignInit(Hash->HashContext, NULL, Hash->Md, NULL, Hash->HmacKey)) {
-        Status = QUIC_STATUS_INTERNAL_ERROR;
-        goto Error;
+    if (!HMAC_Init_ex(HashContext, NULL, 0, NULL, NULL)) {
```
anrossi, line 1998: Won't this call will overwrite your previous call to `HMAC_Init_ex` which provides the key/"salt"?

anrossi: Yeah this should be unnecessary because `QuicHashCreate` has already initialized the hash. Unless you reuse the hash, in which case, then you need to store the key and reinitialize with it every time.

nibanks: Per my understanding of the API, you can call `HMAC_Init_ex` with all NULLs to re-initialize, so that you can use HashCompute multiple different times.

nibanks: And by re-initialize, it goes back to the state you last called `HMAC_Init_ex` with the key.

anrossi: In most cases, you're double-initializing here, then.

nibanks: Yes, but it's the only option to make this function generically support multiple calls.

anrossi: Ok. Let's go with this for now, and I'll do some further investigating to find the correct new APIs we should use which don't have performance problems.

anrossi (approved):

PR #925, opened 2020-10-16
Rename bin artifact name to artifacts

ThadHouse:
Because of IIS, bin is a disallowed folder, so rename to artifacts so files can be properly downloaded.

https://github.com/microsoft/azure-pipelines-tasks/issues/13322


nibanks (approved):

ThadHouse:
Did test that we could actually grab the files.

PR #927, opened 2020-10-19
Add CONTRIBUTING-zh.md

CH-ND-N:
I have added the Chinese version of CONTRIBUTING.md. Hope you find my first contribution helpful.


nibanks:
@CH-ND-N thanks for the contribution, but we don't have a way to easily review or maintain this addition. Additionally, we definitely can't scale this up to more/other languages. So we are going to have to stick with English only documentation. If the existing documentation is difficult to understand for non-native speakers, feel free to open issues or suggest PRs to improve the text. Thanks again!

CH-ND-N:
Okay, sir
I would make my progress in that direction. It was my first repo which I had many expectations with. Thanks for guiding me.

PR #928, opened 2020-10-19
Document perf counters in Diagnostics.md

anrossi:


docs/Diagnostics.md
```diff
@@ -72,4 +72,77 @@ First off, you're going to need xperf and wpa. Installing the [Windows ADK](http
...
+uint32_t BufferLength = sizeof(Counters);
+MsQuic->GetParam(
+    nullptr,
+    QUIC_PARAM_LEVEL_GLOBAL,
+    QUIC_PARAM_GLOBAL_PERF_COUNTERS,
+    &BufferLength,
+    Counters);
+```
+
+Each of the parameters available is described here:
```
nibanks, line 88: I think a 2 column table would be nicer for these.

nibanks:
Can you also add some information about how these are exposed on the latest Windows builds via perfmon?

nibanks:
Also, should we call out that these are collected by ETW rundowns? So the latest values will always be included at the start of any trace?

docs/Diagnostics.md
```diff
@@ -72,4 +72,77 @@ First off, you're going to need xperf and wpa. Installing the [Windows ADK](http
 
 ## Counters
 
+To assist investigations into running systems, MsQuic has a number of performance counters that are updated during runtime. These counters are exposed as a global `GetParam` parameter.
+Sample code demonstrating how to query the performance counters:
+```c
+uint64_t Counters[QUIC_PERF_COUNTER_MAX];
+uint32_t BufferLength = sizeof(Counters);
+MsQuic->GetParam(
+    nullptr,
```
nibanks, line 81:
```suggestion
    NULL,
```
Let's use C syntax instead.

anrossi:
> Also, should we call out that these are collected by ETW rundowns? So the latest values will always be included at the start of any trace (at least on Windows)?

Are the counters in the rundown kernel mode only, or do they include the values of user mode processes?

nibanks:
> > Also, should we call out that these are collected by ETW rundowns? So the latest values will always be included at the start of any trace (at least on Windows)?
> 
> Are the counters in the rundown kernel mode only, or do they include the values of user mode processes?

Rundown is for all processes.

docs/Diagnostics.md
```diff
@@ -72,4 +72,52 @@ First off, you're going to need xperf and wpa. Installing the [Windows ADK](http
...
+    QUIC_PARAM_LEVEL_GLOBAL,
+    QUIC_PARAM_GLOBAL_PERF_COUNTERS,
+    &BufferLength,
+    Counters);
+```
+
+Each of the counters available is described here:
+Counter | Description
+--------|------------
+QUIC_PERF_COUNTER_CONN_CREATED | Total connections ever allocated.
```
nibanks, line 91: Some of these have periods at the end, and some don't. Let's just remove all the periods.

docs/Diagnostics.md
```diff
@@ -72,4 +72,52 @@ First off, you're going to need xperf and wpa. Installing the [Windows ADK](http
...
+QUIC_PERF_COUNTER_APP_SEND_BYTES | Total bytes sent by applications.
+QUIC_PERF_COUNTER_APP_RECV_BYTES | Total bytes received by applications.
+QUIC_PERF_COUNTER_CONN_QUEUE_DEPTH | Current connections queued for processing.
+QUIC_PERF_COUNTER_CONN_OPER_QUEUE_DEPTH | Current connection operations queued.
+QUIC_PERF_COUNTER_CONN_OPER_QUEUED | Total connection operations queued ever.
+QUIC_PERF_COUNTER_CONN_OPER_COMPLETED | Total connection operations processed ever.
+QUIC_PERF_COUNTER_WORK_OPER_QUEUE_DEPTH | Current worker operations queued.
+QUIC_PERF_COUNTER_WORK_OPER_QUEUED | Total worker operations queued ever.
+QUIC_PERF_COUNTER_WORK_OPER_COMPLETED | Total worker operations processed ever.
+QUIC_PERF_COUNTER_MAX | Not a perf counter; just the count of counters.
```
nibanks, line 118: No need to include this.

nibanks (approved):

PR #931, opened 2020-10-20
Base work of adding RPS Latency results to performance testing

ThadHouse:
Adds a context per stream so we can record start times and other metadata per stream.

With this, currently just average latency across the whole test is computed. This number is outputted by the runner, but is not uploaded to the Database. I want the full RPS latency test setup before doing that. But this both gives us some verification numbers and a good baseline.

As an example, the current pipeline averages are about:

```
Running Test RPS_Windows_x64_schannel_Default
Test Run Took 00:00:13.7903961
Run 1: 674038 RPS: 2962 Avg Latency us
Test Run Took 00:00:13.8456679
Run 2: 670415 RPS: 2997 Avg Latency us
Test Run Took 00:00:14.0215841
Run 3: 717004 RPS: 2845 Avg Latency us
Test Run Took 00:00:14.0699401
Run 4: 661930 RPS: 3057 Avg Latency us
Test Run Took 00:00:14.0180998
Run 5: 708703 RPS: 2861 Avg Latency us
```


ThadHouse:
Closign so I can work on this more. Will reopen later.

ThadHouse:
After a bunch of testing a few different methods of testing RPS, it does seem like just our standard RPS test, but with limiting the number of connections to 60, gets us a pretty consistent CPU utilization and RPS rate for testing latency. I'm going to try a few more ideas.

But it does look like adding latency to general RPS tests does not causes performance to drop, so I'm just going to add it to the general tests as well, and work on getting those numbers in the DB. Since they are useful to have for other scenarios.

src/perf/lib/PerfCommon.h
```diff
@@ -24,6 +24,7 @@
 #define TPUT_DEFAULT_IDLE_TIMEOUT           (1 * 1000)
 
 #define RPS_MAX_CLIENT_PORT_COUNT           256
+#define RPS_MAX_REQUESTS_PER_SECOND         1500000 // 1.5 million RPS max as a guess
```
nibanks, line 27: I don't like depending on things like this. What happens when this ends up being not enough?

ThadHouse: It will only record up to this much. Any requests more then this just won't have latency numbers. There is logic to make sure we don't end up writing out of range.

src/perf/lib/RpsClient.cpp
```diff
@@ -268,13 +301,22 @@ RpsClient::ConnectionCallback(
...
     ) {
     switch (Event->Type) {
     case QUIC_STREAM_EVENT_RECEIVE:
         if (Event->RECEIVE.Flags & QUIC_RECEIVE_FLAG_FIN) {
-            InterlockedIncrement64((int64_t*)&CompletedRequests);
+            uint64_t ToPlaceIndex = (uint64_t)InterlockedIncrement64((int64_t*)&CompletedRequests) - 1;
+            uint64_t EndTime = QuicTimeUs64();
+            uint64_t Delta = QuicTimeDiff64(StrmContext->StartTime, EndTime);
+            if (ToPlaceIndex < MaxLatencyIndex) {
+                if (Delta > 0xFFFFFFFF) {
```
nibanks, line 315: Use UINT32_MAX instead.

src/perf/lib/quicmain.cpp
```diff
@@ -19,18 +19,18 @@
 #include "quicmain.cpp.clog.h"
 #endif
 
-const MsQuicApi* MsQuic;
+const MsQuicApi* MsQuic = nullptr;
```
nibanks, line 22: Unnecessary, as is the one below. All globals are zero init'ed.

ThadHouse: Fixed

src/perf/lib/RpsClient.cpp
ThadHouse: Fixed

nibanks (approved):

Issue #935, opened 2020-10-21
Diagnosability improvements to identify operating (normal vs defensive) mode


genghiskMSFT:
### Describe the feature you'd like supported
Diagnosability mechanisms to help identify when in defensive mode.

msquic has protection against DoS attacks, in particular [QuicBindingShouldRetryConnection](https://github.com/microsoft/msquic/blob/main/src/core/binding.c#L1201) which has a memory-based threshold to determine whether it is inundated with requests, and therefore should drop the packet/not respond to a request.

### Proposed solution

The benefits of this feature are to be able to diagnose slow-to-respond or completely unresponsive connections as due to the implementation being in defensive mode. This is useful when investigating support cases from customers.

### Additional context

The indicator of operating mode (normal, defensive, any future operating modes) should be easily observable without needing to understand the details of the implementation or even the QUIC protocol. It should be as easy as dumping a global variable or a connection-related descriptor. In debugging environments like WinDbg, a kernel extension would be nice. For other scenarios, logged events may be of better use - e.g. when a new connection comes in but the implementation chooses to drop the packet, log an event using OS-provided mechanisms (e.g. ETW) such that the event can be configured/filtered.

nibanks:
I think it would definitely be useful to add both:

- Add ETW event to indicate the current "Retry" state (on or off) that's also fired during the ETW rundown.
- Update the debugger extension to expose this state in the `!quiclib` command.

PR #936, opened 2020-10-21
Add Retry Send State to Debugger Extension

nibanks:
The debugger extension part of #935.

### DIFF
--- a/src/tools/dbg/quictypes.h
+++ b/src/tools/dbg/quictypes.h
@@ -380,13 +380,17 @@ struct CidStr {
     char Data[256];
 
     CidStr(ULONG64 Addr, UCHAR Length) {
-        for (UCHAR i = 0; i < Length; i++) {
-            UCHAR Byte;
-            ReadTypeAtAddr(Addr + i, &Byte);
-            Data[i * 2] = QuicHalfByteToStr(Byte >> 4);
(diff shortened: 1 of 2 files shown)


src/tools/dbg/quictypes.h
```diff
@@ -1440,6 +1453,24 @@ struct QuicLibrary : Struct {
...
+    }
+
+    UINT64 TotalMemory() {
+        UINT64 QuicTotalMemory;
+        ReadTypeAtAddr<UINT64>(GetExpression("msquic!QuicTotalMemory"), &QuicTotalMemory);
+        return QuicTotalMemory;
+    }
+
+    UINT64 RetryHandshakeMemoryLimit() {
+        return (GetSettings().RetryMemoryLimit() * TotalMemory()) / UINT16_MAX;
```
anrossi, line 1467: why divide by 64k?

anrossi (approved):

src/tools/dbg/quictypes.h
nibanks: See https://github.com/microsoft/msquic/blob/main/src/core/binding.c#L1201. RetryMemoryLimit is a fracton of UINT16_MAX.

anrossi: Ah, makes sense

PR #938, opened 2020-10-21
Reorder performance read cancellation request in kernel mode

ThadHouse:
There was a tight race window where a cancellation could happen after the request is already completed.

### DIFF
--- a/src/perf/bin/drvmain.cpp
+++ b/src/perf/bin/drvmain.cpp
@@ -41,6 +41,7 @@ typedef struct QUIC_DRIVER_CLIENT {
     QUIC_THREAD Thread;
     bool Canceled;
     bool CleanupHandleCancellation;
+    QUIC_LOCK CleanupLock;
 } QUIC_DRIVER_CLIENT;
 
 WDF_DECLARE_CONTEXT_TYPE_WITH_NAME(QUIC_DRIVER_CLIENT, QuicPerfCtlGetFileContext);
@@ -402,6 +403,7 @@ QuicPerfCtlEvtFileCreate(
         }
 
         RtlZeroMemory(Client, sizeof(QUIC_DRIVER_CLIENT));
+        QuicLockInitialize(&Client->CleanupLock);
 
         //
         // Insert into the client list
@@ -473,6 +475,7 @@ QuicPerfCtlEvtFileCleanup(
         if (Client->Thread != nullptr) {
             QuicThreadWait(&Client->Thread);
             QuicThreadDelete(&Client->Thread);
+            Client->Thread = nullptr;
         }
         QuicEventUninitialize(Client->StopEvent);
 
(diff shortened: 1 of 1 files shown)


src/perf/bin/drvmain.cpp
```diff
@@ -637,12 +638,11 @@ QuicPerfCtlReadPrints(
             QuicThreadDelete(&Client->Thread);
             Client->Thread = nullptr;
         }
+        WdfRequestUnmarkCancelable(Request);
```
nibanks, line 641: Does this return an error that might indicate it's already been canceled, and therefore you shouldn't then complete it again?

nibanks: Also, I see this:
> the driver must not call WdfRequestUnmarkCancelable after EvtRequestCancel calls WdfRequestComplete.
https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdfrequest/nf-wdfrequest-wdfrequestunmarkcancelable#remarks

src/perf/bin/drvmain.cpp
```diff
@@ -585,9 +602,15 @@ QUIC_THREAD_CALLBACK(PerformanceWaitForStopThreadCb, Context)
         return;
     }
 
-    WdfRequestUnmarkCancelable(Request);
+    WdfObjectAcquireLock(Request);
+    Status = WdfRequestUnmarkCancelable(Request);
+    Client->CleanupHandleCancellation = TRUE;
+    WdfObjectReleaseLock(Request);
```
nibanks, line 608: Calling out while holding a lock? Generally a really bad idea...

ThadHouse: The driver framework is designed to do this. It's an option to enable automatic synchronization of all callbacks, but we don't need quite that deep. But we absolutely do need to synchronize between out thread and a call to cancel.

src/perf/bin/drvmain.cpp
```diff
@@ -585,9 +606,16 @@ QUIC_THREAD_CALLBACK(PerformanceWaitForStopThreadCb, Context)
         return;
     }
 
-    WdfRequestUnmarkCancelable(Request);
+    QuicLockAcquire(&Client->CleanupLock);
```
nibanks, line 609: I still don't like holding a lock while you call into here. You should be able to achieve synchronization without doing this.

ThadHouse: The call to WdfRequestUnmarkCancelable basically needs to happen with some sort of synchronize with setting the variable.

According to the docs, normally you use the Wdf automatic synchronization for this, which calls the whole cancel function under a lock, but I don't want to enable that. But this function is designed to be called under a lock, because it is 100% a race condition if you don't have it.

nibanks: I'm still not convinced but I haven't had a time to look in depth here.

nibanks: God this stuff is complicated. I was going to propose a different solution that didn't require a lock to be held while you call, but then read the docs more and realized it wouldn't work. Assuming you've tested canceling and this solution generally works, I'm fine with it for now. (it might not solve all edge conditions, but it's just tool/test code, so I don't care too much).

nibanks (approved):

PR #940, opened 2020-10-21
Simple Client Version Negotiation

nibanks:
Brings back the simple version negotiation code on the client. Instead of just killing the connection in response to the VN packet, we might as well retry the matching version.


anrossi (approved):

anrossi (approved):

src/core/crypto.c
```diff
@@ -342,14 +342,14 @@ QuicCryptoReset(
     Crypto->RecoveryEndOffset = 0;
     Crypto->InRecovery = FALSE;
 
-    if (ResetTls) {
```
anrossi, line 345: What are the ramifications of not having this?

nibanks: This used to be here because we needed to regenerate the QUIC TP for version negotiation, so that we encoded the received versions (to later authenticate). But since we're not doing that, this is unnecessary. We can just retransmit exiting CRYPTO payload, just like RETRY scenarios.

PR #941, opened 2020-10-21
Fix Negotiation of Disable 1-RTT Encryption

nibanks:
Recent change to prefetch QUIC TP introduced a bug in the negotiation of the disable 1-RTT encryption extension. Server ended up disabling encryption even if the client didn't also negotiate this. This PR updates the code paths to handle the server side setting it's option after the client's TPs have already been received.

### DIFF
--- a/src/perf/lib/PerfServer.cpp
+++ b/src/perf/lib/PerfServer.cpp
@@ -104,15 +104,12 @@ PerfServer::ListenerCallback(
     switch (Event->Type) {
     case QUIC_LISTENER_EVENT_NEW_CONNECTION: {
         BOOLEAN value = TRUE;
-        if (QUIC_FAILED(
-            MsQuic->SetParam(
-                Event->NEW_CONNECTION.Connection,
-                QUIC_PARAM_LEVEL_CONNECTION,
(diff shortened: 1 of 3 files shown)


src/perf/lib/PerfServer.cpp
```diff
@@ -104,15 +104,12 @@ PerfServer::ListenerCallback(
...
-        if (QUIC_FAILED(
-            MsQuic->SetParam(
-                Event->NEW_CONNECTION.Connection,
-                QUIC_PARAM_LEVEL_CONNECTION,
-                QUIC_PARAM_CONN_DISABLE_1RTT_ENCRYPTION,
-                sizeof(value),
-                &value))) {
-            WriteOutput("MsQuic->SetParam (CONN_DISABLE_1RTT_ENCRYPTION) failed!\n");
-        }
+        MsQuic->SetParam(
```
anrossi, line 107: why remove the check for success?

anrossi (approved):

src/perf/lib/PerfServer.cpp
nibanks: Because it now fails if the client didn't enable it first.

PR #944, opened 2020-10-23
Disallow 0 sizes in tput tests, make 0 in all perf tests an error

ThadHouse:
0 size for upload or download doesn't make sense, as that case is handled by RPS, and you get no data (other then a single useless latency number) out of in in tput.

Then make 0 for all 3 tests an error, so 0's don't accidentally get uploaded to the server.

Also makes the test runner error if a 0 is received


anrossi (approved):

ThadHouse:
Failures are unrelated.

PR #950, opened 2020-10-28
Refactor ISB Indications

nibanks:
In high BDP scenarios (e.g. 60+ ms RTT and 2+ Gbps links), tests showed that the calculation for ISB (ideal send buffer, or how much send data to keep queued) was resulting in the connection becoming application limited. This was caused by the way we tried to limit the number of these indications up to the app. We ended up not indicating the latest values because they hadn't reached "the threshold" yet.

This changes how we calculate and use the indication threshold. A 1.5x growth rate is used for calculating the thresholds and the ISB is rounded up to the next threshold so there is never underestimation.

### DIFF
--- a/src/core/send_buffer.c
+++ b/src/core/send_buffer.c
@@ -196,11 +196,11 @@ QuicGetNextIdealBytes(
     uint32_t Threshold = QUIC_DEFAULT_IDEAL_SEND_BUFFER_SIZE;
 
     //
-    // We calculate the theshold as an exponential grown based on the default
-    // up to the max, rounding up to the next threshold when equal.
+    // We calculate the threshold as an exponential growth from the default
+    // value up to the max, rounding up to the next threshold when equal.
     //
     while (Threshold <= BaseValue) {
-        uint32_t NextThreshold = (Threshold * 3) / 2; // 1.5x growth
+        uint32_t NextThreshold = Threshold + (Threshold / 2); // 1.5x growth
         if (NextThreshold > QUIC_MAX_IDEAL_SEND_BUFFER_SIZE) {
             Threshold = QUIC_MAX_IDEAL_SEND_BUFFER_SIZE;
             break;
--- a/src/core/send_buffer.c
+++ b/src/core/send_buffer.c
@@ -187,34 +187,63 @@ QuicSendBufferFill(
     }
 }
 
+_IRQL_requires_max_(DISPATCH_LEVEL)
+uint32_t
+QuicGetNextIdealBytes(
+    _In_ uint32_t BaseValue
+    )
+{
+    uint32_t Threshold = QUIC_DEFAULT_IDEAL_SEND_BUFFER_SIZE;
+
+    //
+    // We calculate the theshold as an exponential grown based on the default
+    // up to the max, rounding up to the next threshold when equal.
+    //
+    while (Threshold <= BaseValue) {
+        uint32_t NextThreshold = (Threshold * 3) / 2; // 1.5x growth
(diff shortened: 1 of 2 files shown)


src/core/send_buffer.c
```diff
@@ -228,17 +257,15 @@ QuicSendBufferConnectionAdjust(
...
+
     //
     // TODO: Currently, IdealBytes only grows and never shrinks. Add appropriate
     // shrinking logic.
     //
-    if (Connection->CongestionControl.BytesInFlightMax >
-        QUIC_IDEAL_SEND_BUFFER_THRESHOLD(Connection->SendBuffer.IdealBytes)) {
-        Connection->SendBuffer.IdealBytes =
-            min(2 * Connection->CongestionControl.BytesInFlightMax, QUIC_MAX_IDEAL_SEND_BUFFER_SIZE);
+    if (NewIdealBytes < Connection->SendBuffer.IdealBytes) {
```
ThadHouse, line 267: Is the direction wrong here? Or is this handling the shrinking case, and the comment is no longer appropriate?

nibanks: You're right. Wrong direction.

src/core/send_buffer.c
```diff
@@ -187,34 +187,63 @@ QuicSendBufferFill(
...
+    )
+{
+    uint32_t Threshold = QUIC_DEFAULT_IDEAL_SEND_BUFFER_SIZE;
+
+    //
+    // We calculate the theshold as an exponential grown based on the default
+    // up to the max, rounding up to the next threshold when equal.
+    //
+    while (Threshold <= BaseValue) {
+        uint32_t NextThreshold = (Threshold * 3) / 2; // 1.5x growth
```
maolson-msft, line 203: x + x / 2 replaces a mult with a [cheaper] addition.

src/core/send_buffer.c
```diff
@@ -187,34 +187,63 @@ QuicSendBufferFill(
...
+    _In_ uint32_t BaseValue
+    )
+{
+    uint32_t Threshold = QUIC_DEFAULT_IDEAL_SEND_BUFFER_SIZE;
+
+    //
+    // We calculate the theshold as an exponential grown based on the default
+    // up to the max, rounding up to the next threshold when equal.
+    //
+    while (Threshold <= BaseValue) {
```
maolson-msft, line 202: I think "==" makes more sense, because we want to avoid underestimation, as opposed to wanting overestimation.

src/core/send_buffer.c
```diff
@@ -187,34 +187,63 @@ QuicSendBufferFill(
...
+_IRQL_requires_max_(DISPATCH_LEVEL)
+uint32_t
+QuicGetNextIdealBytes(
+    _In_ uint32_t BaseValue
+    )
+{
+    uint32_t Threshold = QUIC_DEFAULT_IDEAL_SEND_BUFFER_SIZE;
+
+    //
+    // We calculate the theshold as an exponential grown based on the default
```
maolson-msft, line 199: Does this sentence make sense?

src/core/send_buffer.c
maolson-msft: Sorry, I meant "<" :) And actually, it looks like your design requires "<=", since BaseValue (when it's BytesInFlightMax) is limited by the threshold value.

src/core/send_buffer.c
maolson-msft: Eh, good enough. Just fix typo "theshold".

maolson-msft (approved):

ThadHouse (approved):

PR #953, opened 2020-10-29
Print out clog error messages at end of prepare-machine.ps1

### LINKED ISSUE #780: Print clog install errors at the end of prepare-machine.ps1
Prepare-machine installs clog at the beginning, and then installs a bunch of other dependencies on linux. This causes the clog failed to install message to be buried. We should instead also print this message at the end of the script. I don't want to fail it because of -DisableLogs.

ThadHouse:
Closes #780

### DIFF
--- a/scripts/prepare-machine.ps1
+++ b/scripts/prepare-machine.ps1
@@ -35,7 +35,7 @@ $NuGetPath = Join-Path $RootDir "nuget"
 $ClogVersion = "0.1.8"
 $ClogDownloadUrl = "https://github.com/microsoft/CLOG/releases/download/v$ClogVersion"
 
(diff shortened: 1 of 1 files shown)


scripts/prepare-machine.ps1
```diff
@@ -152,3 +153,8 @@ if ($IsWindows) {
         }
     }
 }
+
+foreach ($errMsg in $MessagesAtEnd) {
+   Write-Warning "Clog could not be installed. Building with logs will not work"
```
nibanks, line 158: This line should be a message you add to `MessagesAtEnd` above.

nibanks (approved):

PR #958, opened 2020-11-03
Build configuration updates to WPA tool

ThadHouse:
This does a few things. Enables FxCopAnalyzers which check code style. Also enables nullable warnings. Finally turns on warnings as errors, so we make sure code quality stays sound.

Also does a few syntax improvements enabled in C# 8 (which is valid in NS 2.1)

PR rather then direct commit so it can be reviewed (Especially the name warning supression), and so it doesn't cause merge conflicts with active code changes.


src/tools/wpa/ETWTraceEvent.cs
```diff
@@ -9,7 +9,11 @@
 
 namespace QuicEventDataSource
 {
+#pragma warning disable CA1036 // Override methods on comparable types
```
nibanks, line 12: Can you see if we can just remove the compare instead?

src/tools/wpa/QuicEventDataSource.cs
```diff
@@ -17,15 +18,22 @@ namespace QuicEventDataSource
         "QUIC Event",
         "Events generated by MsQuic")]
     [FileDataSource("etl", "Event Trace Log")]
+#pragma warning disable CA1724 // Class name should not match namespace name
```
nibanks, line 21: Go head and rename the namespace. `Quic` or `MsQuic` works for me, but I don't really care what.

src/tools/wpa/Tables/EventStats.cs
```diff
@@ -34,6 +34,16 @@ public sealed class EventStats
...
         {
+            if (tableBuilder == null)
+            {
+                throw new ArgumentNullException(nameof(tableBuilder));
+            }
+
+            if (tableData == null)
+            {
+                throw new ArgumentNullException(nameof(tableData));
+            }
```
nibanks, line 45: Why are you adding these?

src/tools/wpa/QuicEventDataSource.cs
```diff
@@ -17,15 +18,22 @@ namespace QuicEventDataSource
...
 
         protected override ICustomDataProcessor CreateProcessorCore(
             IEnumerable<IDataSource> dataSources,
             IProcessorEnvironment processorEnvironment,
             ProcessorOptions options)
         {
+            if (applicationEnvironment == null)
+            {
+                throw new InvalidOperationException("SetApplicationEnvironmentCore must be called before CreateProcessorCore");
+            }
```
nibanks, line 35: Why are you adding this?

src/tools/wpa/ETWTraceEvent.cs
ThadHouse: Its not that, its because its a struct. FxCop basically requires all structs to implement comparable. But this doesn't make sense for this one as much.

src/tools/wpa/Tables/EventStats.cs
ThadHouse: Because this is a public function, and FxCop requires public functions to explicitly null check their variables. It throws an error otherwise. The reason is because the stacktrace for actually dereferencing null is harder to read then explicitly null checking.

src/tools/wpa/QuicEventDataSource.cs
ThadHouse: Because this variable is nullable, since its set after construction. So at this point, it can be null. But the constructor we're passing it into requires the variable to be non null. It would be broken if that happened as well, as late calling the set function would do nothing.

src/tools/wpa/Tables/EventStats.cs
nibanks: I prefer to global ignore that error. I don't want to add all these null checks.

src/tools/wpa/ETWTraceEvent.cs
```diff
@@ -7,9 +7,13 @@
 using Microsoft.Diagnostics.Tracing;
 using Microsoft.Performance.SDK.Extensibility;
 
-namespace QuicEventDataSource
+namespace MsQuicEtw
```
nibanks, line 10: Don't include ETW in the name. That's not going to be correct when we add LTTng support.

src/tools/wpa/QuicEventDataSource.cs
nibanks: Well, don't make it nullable. I don't want to add extra error checking code to this project. I want it streamlined.

src/tools/wpa/ETWTraceEvent.cs
ThadHouse: I don't want it to just be MsQuic, because thats not what it is. `MsQuicTracing`?

src/tools/wpa/ETWTraceEvent.cs
nibanks: Can we ignore these globally instead?

ThadHouse: Yeah these I can globally disable.

nibanks (approved):

PR #963, opened 2020-11-04
Initial WPA Plugin Support

nibanks:
This adds the initial support for a WPA plugin that can process MsQuic ETW events. It's structured so that LTTng can be added in the future (just need to implement a new `QuicEventBase` wrapper for it, and a few other things). Currently it has a simple table that just shows stats for number of each type of QUIC event (by ID). Future changes will replace this with the actual tables we have support for internally.


src/tools/wpa/DataModel/QuicEtwEvent.cs
```diff
@@ -0,0 +1,21 @@
...
+//
+
+using System;
+using Microsoft.Diagnostics.Tracing;
+
+namespace MsQuicTracing.DataModel
+{
+    public class QuicEtwEvent : QuicEventBase
+    {
+#pragma warning disable CS8618
```
ThadHouse, line 13: We really should not be disabling this warning. I would much rather prefer make it nullable, and then check for null in the other 2 properties.

ThadHouse:
```
    public class QuicEtwEvent : QuicEventBase
    {
        [DisallowNull]
        public TraceEvent? Event { get; set; }

        public override Guid Provider => Event?.ProviderGuid ?? throw new InvalidOperationException("Event must be set before calling provider");

        public override ushort ID => (ushort?)Event?.ID ?? throw new InvalidOperationException("Event must be set before calling ID");
    }
```

This is the class properly done to work correctly.

nibanks: We're going to end up with a lot of properties here. I really don't want to have to add a ton of nullable checks, both for code cleanliness and performance. A dozen or two extra null checks, per event, for 10 million events makes a difference.

.azure/templates/build-config-wpa.yml
```diff
@@ -0,0 +1,30 @@
+# This template contains steps to build MsQuic WPA plugin.
+
+jobs:
+- job: build_wpa
+  displayName: Any CPU
```
ThadHouse, line 5: Does this need to be `Any CPU`?  Wouldn't something like `Build` be better?

src/tools/wpa/DataModel/QuicEtwEvent.cs
ThadHouse: Performance is not a great argument, as the runtime will optimize a lot of that away correctly. And code cleanliness is still the worst argument ever. Correct code is ALWAYS better then clean code. Unconditionally.

.azure/templates/build-config-wpa.yml
nibanks: Yeah, so it matches everything else:

image

src/tools/wpa/DataModel/QuicEtwEvent.cs
nibanks: Once we have this all up and running, we can then try switching it out to this model to see how it performs. For now, I want to keep it as is.

ThadHouse: And I don't. Its wrong,.

nibanks: I understand. I completely disagree that performance is not a great argument. It is my primary goal (and concern) moving to this new plugin model. Things already take a considerable time with the old one for large files. Until we can prove something will not affect perf I will **always** go with the more performant design. As I said, we can revisit in the future.

ThadHouse: Then you will not be getting an approval from me.

nibanks: Slight refactor to eliminate the problem (at least at this layer).

ThadHouse: Ah. Thats a much better solution.

ThadHouse (approved):

ThadHouse (approved):
The thing I wanted to check can't be removed still, so still good.

PR #970, opened 2020-11-05
Refactor ETW Keywords

nibanks:
Refactors how we use the keywords field for our ETW events. This updates them to encode the object type in the low byte (5 bits to be exact) so that they can be easily/efficiently extracted when being post-processed by WPA.


nibanks:
Going to try another direction.

PR #972, opened 2020-11-05
Complete ChaCha20-Poly1305 support.

### LINKED ISSUE #947: ChaCh20-Poly1305 Support
### Describe the bug

Per TLS 1.3 spec, ChaCha20-Poly1205 is required to be supported. Currently it's unsupported for Schannel and OpenSSL TLS abstractions. If enabled for underlying Schannel crypto (per global config) it will cause the QUIC connection to fail.

The work to complete support needs to be added (mostly around creating keys and doing header encryption).

anrossi:
Fixes #947


src/platform/tls_schannel.c
```diff
@@ -675,6 +695,37 @@ QuicTlsLibraryInitialize(
         goto Error;
     }
 
+    Status =
```
nibanks, line 698: The code below needs to be resilent to being run on older OSes. So we can't fail initializing the library if ChaCha/Poly isn't available

src/platform/tls_schannel.c
```diff
@@ -373,6 +384,13 @@ typedef struct QUIC_TLS {
 
 } QUIC_TLS;
 
+typedef struct QUIC_HP_KEY {
+    BCRYPT_KEY_HANDLE Key;
+    BCRYPT_AUTHENTICATED_CIPHER_MODE_INFO* Info;
```
nibanks, line 389: Why can't this be on the stack instead?

src/platform/tls_schannel.c
```diff
@@ -3045,19 +3163,44 @@ QuicHpComputeMask(
...
-            0,
-            Mask,
-            QUIC_HP_SAMPLE_LENGTH * BatchSize,
-            &TempSize,
-            0));
+    QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
+    if (Key->Aead == QUIC_AEAD_CHACHA20_POLY1305) {
+        uint8_t Zero[5] = {0};
+        Key->Info->cbNonce = QUIC_HP_SAMPLE_LENGTH;
+        for (uint32_t i = 0, Offset = 0; i < BatchSize; ++i, Offset += QUIC_HP_SAMPLE_LENGTH) {
```
nibanks, line 3170: Is there no way to do the batch in a single call?

src/platform/tls_schannel.c
```diff
@@ -3045,19 +3163,44 @@ QuicHpComputeMask(
...
+                NtStatusToQuicStatus(
+                BCryptEncrypt(
+                    Key->Key,
+                    Zero,
+                    sizeof(Zero),
+                    Key->Info,
+                    NULL,   //Key->Info->pbNonce,
+                    0,      //QUIC_HP_SAMPLE_LENGTH,
+                    Mask + Offset, 
+                    QUIC_HP_SAMPLE_LENGTH, // This might fail because the Tag won't fit...
```
nibanks, line 3182: Why wouldn't the tag fit?

src/platform/tls_schannel.c
anrossi: Having on the stack would incur performance cost (larger stack frame) for AES keys. This way, it's only a cost for ChaCha20-Poly1305 keys.

src/platform/tls_schannel.c
anrossi: No. ChaCha20 is a streaming cipher, not a block cipher. The QUIC-TLS spec uses it to get 5 bytes of the stream, for a given unique IV (the packet sample). Because each header uses a different IV, it produces a different stream per packet. Which makes it not possible to do a single call.

src/platform/tls_schannel.c
anrossi: The tag is 16 bytes, and QUIC_HP_SAMPLE_LENGTH is 16 bytes, but the output is length of Mask + length of Tag, which is 21 bytes.
We could solve this by making the sample length larger, and discarding the unused parts on AES.  Ideally, we'd be able to compute ChaCha20 without the Poly1305 tag, and this won't be a problem.

src/platform/tls_schannel.c
nibanks:
```
typedef struct QUIC_HP_KEY {
    BCRYPT_KEY_HANDLE Key;
    QUIC_AEAD_TYPE Aead;
    BCRYPT_AUTHENTICATED_CIPHER_MODE_INFO Info[0];
} QUIC_HP_KEY;
```

src/platform/tls_schannel.c
```diff
@@ -2635,7 +2691,44 @@ QuicParseTrafficSecrets(
...
+        default:
+            QuicTraceEvent(
+                TlsError,
+                "[ tls][%p] ERROR, %s.",
+                TlsContext->Connection,
+                "Unsupported AES key size");
+            return FALSE;
+        }
+    } else if (wcscmp(TrafficSecrets->SymmetricAlgId, BCRYPT_CHACHA20_POLY1305_ALGORITHM) == 0) {
+        switch (TrafficSecrets->KeySize) {
```
nibanks, line 2719: We should fail if we don't have a algorithm handle (failed at load time).

src/platform/tls_openssl.c
```diff
@@ -89,6 +89,11 @@ typedef struct QUIC_TLS {
 
 } QUIC_TLS;
 
+typedef struct QUIC_HP_KEY {
+    EVP_CIPHER_CTX* CipherCtx;
```
nibanks, line 93: Is there, by chance, a way to trivially query the AEAD type from this pointer? That way we wouldn't need the extra object/indirection.

src/platform/tls_openssl.c
```diff
@@ -1844,14 +1860,14 @@ QuicHpKeyCreate(
         Aead = EVP_aes_256_ecb();
         break;
     case QUIC_AEAD_CHACHA20_POLY1305:
-        Aead = EVP_chacha20_poly1305();
+        Aead = EVP_chacha20();
```
nibanks, line 1863: So we don't need the poly1305 for header protection?

src/platform/tls_schannel.c
```diff
@@ -675,6 +693,39 @@ QuicTlsLibraryInitialize(
...
+            BCRYPT_CHACHA20_POLY1305_ALGORITHM,
+            MS_PRIMITIVE_PROVIDER,
+            BCRYPT_PROV_DISPATCH);
+    if (!NT_SUCCESS(Status)) {
+        QuicTraceEvent(
+            LibraryErrorStatus,
+            "[ lib] ERROR, %u, %s.",
+            Status,
+            "Open ChaCha20-Poly1305 algorithm");
+        // ChaCha20-Poly1305 may not be supported on older OSes, so don't treat
```
nibanks, line 708: Fix comment formatting

src/platform/tls_schannel.c
```diff
@@ -702,14 +753,58 @@ QuicTlsLibraryInitialize(
...
+            BCRYPT_CHACHA20_POLY1305_ALGORITHM,
+            MS_PRIMITIVE_PROVIDER,
+            0);
+    if (!NT_SUCCESS(Status)) {
+        QuicTraceEvent(
+            LibraryErrorStatus,
+            "[ lib] ERROR, %u, %s.",
+            Status,
+            "Open ChaCha20-Poly1305 algorithm");
+        // ChaCha20-Poly1305 may not be supported on older OSes, so don't treat
```
nibanks, line 776: fix comment formatting

src/platform/tls_openssl.c
anrossi: It's unlikely; the struct is opaque by design.  There might be an API to query the object type, but that would probably be slower than a double indirection, wouldn't it?

src/platform/tls_openssl.c
anrossi: Nope. According to the [QUIC-TLS spec](https://tools.ietf.org/html/draft-ietf-quic-tls-31#section-5.4.4) only the raw ChaCha20 function is used for header protection.

src/platform/tls_schannel.c
```diff
@@ -214,6 +214,17 @@ typedef struct _SecPkgContext_SessionInfo
...
 
+//
+// Defines until BCrypt.h updates
+//
+#ifndef BCRYPT_CHACHA20_POLY1305_ALGORITHM
+#define BCRYPT_CHACHA20_POLY1305_ALGORITHM L"CHACHA20_POLY1305"
+#endif
+
+#ifndef BCRYPT_CHACHA20_POLY1305_ALG_HANDLE
+#define BCRYPT_CHACHA20_POLY1305_ALG_HANDLE ((BCRYPT_ALG_HANDLE) 0x000003A1)
```
anrossi, line 225: Remove this.

src/platform/tls_mitls.c
```diff
@@ -21,7 +21,7 @@
 
 uint16_t QuicTlsTPHeaderSize = 0;
 
-#define QUIC_SUPPORTED_CIPHER_SUITES        "TLS_AES_128_GCM_SHA256"
+#define QUIC_SUPPORTED_CIPHER_SUITES        "TLS_AES_128_GCM_SHA256:TLS_CHACHA20_POLY1305_SHA256"
```
nibanks, line 24: We should match the same everywhere. Prefix this with `TLS_AES_256_GCM_SHA384`?

anrossi: Does mitls support that?

nibanks: It _should_.

nibanks (approved):

PR #973, opened 2020-11-05
WPA Plugin - Worker Table Support

nibanks:
Now the Worker activity graph is working mostly. This adds initial parsing of event payload; specifically a couple of the worker events.


src/tools/wpa/DataModel/QuicEtwEventPayload.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
```
ThadHouse, line 1: Missing copyright.

src/tools/wpa/DataModel/QuicState.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        internal void AddEvent(QuicEvent quicEvent)
+        {
+            if (!ObjectEventCounts.ContainsKey(quicEvent.ObjectType))
+            {
+                ObjectEventCounts.Add(quicEvent.ObjectType, 1);
+            }
+            else
+            {
+                ObjectEventCounts[quicEvent.ObjectType]++;
```
ThadHouse, line 27: This is actually going to result in 3 hash lookups in this case. Since it seems like this is a hot path, do the following to bring it down to 2.

if (ObjectEventCounts.TryGetValue(quicEvent.ObjectType, out var eventCount)) 
{
   ObjectEventCounts[quicEvent.ObjectType] = eventCount + 1;
}
else
{
  ObjectEventCounts.Add(quicEvent.ObjectType, 1);
}

nibanks: I plan to remove all this code soon, so don't worry about it.

ThadHouse (approved):

PR #974, opened 2020-11-06
Add support for embedding pull tags in windows user mode alloc

ThadHouse:
Only in debug mode, but will both make debugging easier, and will help make sure all the pool tags are correct.

Found a few bugs in which pool tags were used, so those are fixed.

### DIFF
--- a/src/platform/platform_winuser.c
+++ b/src/platform/platform_winuser.c
@@ -423,6 +423,16 @@ QuicRandom(
 
 #endif
 
+#ifdef DEBUG
+#ifdef _WIN64
+// 64 bit
+#define AllocOffset 16
+#else
+// 32 bit
+#define AllocOffset 8
+#endif
+#endif
+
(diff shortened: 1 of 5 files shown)


src/platform/platform_winuser.c
```diff
@@ -423,6 +423,16 @@ QuicRandom(
...
 
+#ifdef DEBUG
+#ifdef _WIN64
+// 64 bit
+#define AllocOffset 16
+#else
+// 32 bit
+#define AllocOffset 8
+#endif
+#endif
```
nibanks, line 434: Can you just use `sizeof(size_t)`?

nibanks: Or a define of something else instead of hard coded?

ThadHouse: Done

nibanks (approved):

PR #975, opened 2020-11-06
Various fixes to make clang-tidy happier

larseggert:
This is a little all over the place, but consolidates a bunch of nit fixes to `core` and `platform` that make `clang-tidy` happier. There are still some issues it finds, incl. some zero-pointer accesses, etc.

Are you interested in fixes like this?

### DIFF
--- a/src/core/binding.c
+++ b/src/core/binding.c
@@ -318,13 +318,17 @@ QuicBindingRegisterListener(
 
         if (NewFamily > ExistingFamily) {
             break; // End of possible family matches. Done searching.
-        } else if (NewFamily != ExistingFamily) {
+        }
+
+        if (NewFamily != ExistingFamily) {
             continue;
         }
 
         if (!NewWildCard && ExistingWildCard) {
             break; // End of specific address matches. Done searching.
-        } else if (NewWildCard != ExistingWildCard) {
+        }
+
+        if (NewWildCard != ExistingWildCard) {
             continue;
         }
 
@@ -956,7 +960,7 @@ QuicBindingProcessStatelessOperation(
                 sizeof(Token),
                 (uint8_t*)&Token,
                 (uint16_t)SendDatagram->Length,
-                (uint8_t*)SendDatagram->Buffer);
+                SendDatagram->Buffer);
         QUIC_DBG_ASSERT(SendDatagram->Length != 0);
 
         QuicTraceLogVerbose(
@@ -1112,7 +1116,8 @@ QuicBindingPreprocessDatagram(
             if (Packet->DestCidLen == 0) {
                 QuicPacketLogDrop(Binding, Packet, "Zero length DestCid");
                 return FALSE;
-            } else if (Packet->DestCidLen < QUIC_MIN_INITIAL_CONNECTION_ID_LENGTH) {
+            }
+            if (Packet->DestCidLen < QUIC_MIN_INITIAL_CONNECTION_ID_LENGTH) {
                 QuicPacketLogDrop(Binding, Packet, "Less than min length CID on non-exclusive binding");
                 return FALSE;
             }
@@ -1467,7 +1472,9 @@ QuicBindingDeliverDatagrams(
                 QuicBindingQueueStatelessOperation(
                     Binding, QUIC_OPER_TYPE_RETRY, DatagramChain);
 
-        } else if (!DropPacket) {
+        }
+
+        if (!DropPacket) {
             Connection = QuicBindingCreateConnection(Binding, DatagramChain);
         }
     }
@@ -1618,7 +1625,6 @@ QuicBindingReceive(
         //
         if (!QuicBindingDeliverDatagrams(Binding, SubChain, SubChainLength)) {
             *ReleaseChainTail = SubChain;
-            ReleaseChainTail = SubChainTail;
         }
     }
 
--- a/src/core/binding.c
+++ b/src/core/binding.c
@@ -1625,6 +1625,7 @@ QuicBindingReceive(
         //
         if (!QuicBindingDeliverDatagrams(Binding, SubChain, SubChainLength)) {
             *ReleaseChainTail = SubChain;
+            ReleaseChainTail = SubChainTail; // NOLINT
         }
     }
 
--- a/src/core/binding.c
+++ b/src/core/binding.c
@@ -1625,7 +1625,7 @@ QuicBindingReceive(
         //
         if (!QuicBindingDeliverDatagrams(Binding, SubChain, SubChainLength)) {
             *ReleaseChainTail = SubChain;
-            ReleaseChainTail = SubChainTail; // NOLINT
+            ReleaseChainTail = SubChainTail; // cppcheck-suppress unreadVariable; NOLINT
         }
     }
 
--- a/src/core/configuration.c
+++ b/src/core/configuration.c
@@ -358,7 +358,7 @@ MsQuicConfigurationLoadCredential(
 _IRQL_requires_max_(PASSIVE_LEVEL)
 void
 QuicConfigurationTraceRundown(
-    _In_ QUIC_CONFIGURATION* Configuration // NOLINT
+    _In_ QUIC_CONFIGURATION* Configuration
     )
 {
     QuicTraceEvent(
--- a/src/core/configuration.c
+++ b/src/core/configuration.c
@@ -358,7 +358,7 @@ MsQuicConfigurationLoadCredential(
 _IRQL_requires_max_(PASSIVE_LEVEL)
 void
 QuicConfigurationTraceRundown(
-    _In_ QUIC_CONFIGURATION* Configuration
+    _In_ QUIC_CONFIGURATION* Configuration // NOLINT
     )
 {
     QuicTraceEvent(
@@ -408,34 +408,24 @@ QuicConfigurationParamGet(
         void* Buffer
     )
 {
-    QUIC_STATUS Status;
-
-    switch (Param) {
-    case QUIC_PARAM_CONFIGURATION_SETTINGS:
+    if (Param == QUIC_PARAM_CONFIGURATION_SETTINGS) {
 
         if (*BufferLength < sizeof(QUIC_SETTINGS)) {
             *BufferLength = sizeof(QUIC_SETTINGS);
-            Status = QUIC_STATUS_BUFFER_TOO_SMALL; // TODO - Support partial
-            break;
+            return QUIC_STATUS_BUFFER_TOO_SMALL; // TODO - Support partial
         }
 
         if (Buffer == NULL) {
-            Status = QUIC_STATUS_INVALID_PARAMETER;
-            break;
+            return QUIC_STATUS_INVALID_PARAMETER;
         }
 
         *BufferLength = sizeof(QUIC_SETTINGS);
         QuicCopyMemory(Buffer, &Configuration->Settings, sizeof(QUIC_SETTINGS));
 
-        Status = QUIC_STATUS_SUCCESS;
-        break;
-
-    default:
-        Status = QUIC_STATUS_INVALID_PARAMETER;
-        break;
+        return QUIC_STATUS_SUCCESS;
     }
 
-    return Status;
+    return QUIC_STATUS_INVALID_PARAMETER;
 }
 
 _IRQL_requires_max_(PASSIVE_LEVEL)
@@ -448,14 +438,10 @@ QuicConfigurationParamSet(
         const void* Buffer
     )
 {
-    QUIC_STATUS Status;
-
-    switch (Param) {
-    case QUIC_PARAM_GLOBAL_SETTINGS:
+    if (Param == QUIC_PARAM_GLOBAL_SETTINGS) {
 
         if (BufferLength != sizeof(QUIC_SETTINGS)) {
-            Status = QUIC_STATUS_INVALID_PARAMETER; // TODO - Support partial
-            break;
+            return QUIC_STATUS_INVALID_PARAMETER; // TODO - Support partial
         }
 
         QuicTraceLogInfo(
@@ -468,19 +454,13 @@ QuicConfigurationParamSet(
                 TRUE,
                 BufferLength,
                 (QUIC_SETTINGS*)Buffer)) {
-            Status = QUIC_STATUS_INVALID_PARAMETER;
-            break;
+            return QUIC_STATUS_INVALID_PARAMETER;
         }
 
         QuicSettingsDumpNew(BufferLength, (QUIC_SETTINGS*)Buffer);
 
-        Status = QUIC_STATUS_SUCCESS;
-        break;
-
-    default:
-        Status = QUIC_STATUS_INVALID_PARAMETER;
-        break;
+        return QUIC_STATUS_SUCCESS;
     }
 
-    return Status;
+    return QUIC_STATUS_INVALID_PARAMETER;
 }
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -2400,8 +2400,7 @@ QuicConnValidateTransportParameterDraft27CIDs(
             return FALSE;
         }
 
-        if (
-            memcmp(
+        if (memcmp(
                 Connection->PeerTransportParams.OriginalDestinationConnectionID,
                 Connection->OrigDestCID->Data,
                 Connection->OrigDestCID->Length) != 0) {
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -1612,7 +1612,7 @@ QuicConnTryClose(
             QuicConnLogStatistics(Connection);
         }
 
-        if (Flags & QUIC_CLOSE_APPLICATION) { // NOLINT
+        if (Flags & QUIC_CLOSE_APPLICATION) {
             QuicTraceEvent(
                 ConnAppShutdown,
                 "[conn][%p] App Shutdown: %llu (Remote=%hhu)",
@@ -2033,7 +2033,7 @@ QuicConnRecvResumptionTicket(
         ResumptionAccepted =
             QUIC_SUCCEEDED(QuicConnIndicateEvent(Connection, &Event));
 
-        if (ResumptionAccepted) { // NOLINT
+        if (ResumptionAccepted) {
             QuicTraceEvent(
                 ConnServerResumeTicket,
                 "[conn][%p] Server app accepted resumption ticket",
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -3130,22 +3130,20 @@ QuicConnRecvHeader(
 
     QUIC_FRE_ASSERT(QuicIsVersionSupported(Connection->Stats.QuicVersion));
 
-#if DEBUG
-    if (!Packet->IsShortHeader) {
-        if (Connection->State.ShareBinding) {
-            QUIC_DBG_ASSERT(Packet->DestCidLen >= QUIC_MIN_INITIAL_CONNECTION_ID_LENGTH);
-        } else {
-            QUIC_DBG_ASSERT(Packet->DestCidLen == 0);
-        }
-    }
-#endif
-
     //
     // Begin non-version-independent logic. When future versions are supported,
     // there may be some switches based on packet version.
     //
 
     if (!Packet->IsShortHeader) {
+#if DEBUG
+        if (Connection->State.ShareBinding) {
+            QUIC_DBG_ASSERT(Packet->DestCidLen >= QUIC_MIN_INITIAL_CONNECTION_ID_LENGTH);
+        } else {
+            QUIC_DBG_ASSERT(Packet->DestCidLen == 0);
+        }
+#endif
+
         if (Packet->LH->Type == QUIC_RETRY) {
             QuicConnRecvRetry(Connection, Packet);
             return FALSE;
@@ -4916,6 +4914,7 @@ QuicConnRecvDatagrams(
             Batch,
             Cipher,
             &RecvState);
+        // BatchCount = 0; // Value stored to 'BatchCount' is never read [clang-analyzer-deadcode.DeadStores]
     }
 
     if (RecvState.ResetIdleTimeout) {
@@ -5419,16 +5418,12 @@ QuicConnParamSet(
         Connection->CloseReasonPhrase =
             QUIC_ALLOC_NONPAGED(BufferLength, QUIC_POOL_CLOSE_REASON);
 
-        if (Connection->CloseReasonPhrase != NULL) {
-            if (Buffer) {
-                QuicCopyMemory(
-                    Connection->CloseReasonPhrase,
-                    Buffer,
-                    BufferLength);
-            }
-
+        if (Buffer && Connection->CloseReasonPhrase != NULL) {
+            QuicCopyMemory(
+                Connection->CloseReasonPhrase,
+                Buffer,
+                BufferLength);
             Status = QUIC_STATUS_SUCCESS;
-
         } else {
             Status = QUIC_STATUS_OUT_OF_MEMORY;
         }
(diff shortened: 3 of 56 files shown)


src/core/binding.c
```diff
@@ -318,13 +318,17 @@ QuicBindingRegisterListener(
 
         if (NewFamily > ExistingFamily) {
             break; // End of possible family matches. Done searching.
-        } else if (NewFamily != ExistingFamily) {
```
nibanks, line 321: Does clang actually complain about this?

src/core/binding.c
```diff
@@ -1618,7 +1625,6 @@ QuicBindingReceive(
         //
         if (!QuicBindingDeliverDatagrams(Binding, SubChain, SubChainLength)) {
             *ReleaseChainTail = SubChain;
-            ReleaseChainTail = SubChainTail;
```
nibanks, line 1621: I guess I understand that this is unnecessary, but I fear that if any other code added after this that does manipulation of the chain might result in bugs if we remove this. What do you think?

src/core/configuration.c
```diff
@@ -358,7 +358,7 @@ MsQuicConfigurationLoadCredential(
 _IRQL_requires_max_(PASSIVE_LEVEL)
 void
 QuicConfigurationTraceRundown(
-    _In_ QUIC_CONFIGURATION* Configuration
+    _In_ QUIC_CONFIGURATION* Configuration // NOLINT
```
nibanks, line 361: What does the `// NOLINT` do here?

src/core/connection.c
```diff
@@ -2389,14 +2389,18 @@ QuicConnValidateTransportParameterDraft27CIDs(
...
                 ConnError,
                 "[conn][%p] ERROR, %s.",
                 Connection,
                 "Peer provided incorrect length of original destination CID in TP");
             return FALSE;
-        } else if (
+        }
+
+        if (
             memcmp(
```
nibanks, line 2404:
```suggestion
        if (memcmp(
```

src/core/connection.c
```diff
@@ -4911,7 +4917,6 @@ QuicConnRecvDatagrams(
             Batch,
             Cipher,
             &RecvState);
-        BatchCount = 0;
```
nibanks, line 4914: Again, I don't like removing lines like this that could result in bugs in the future.

src/core/connection.c
```diff
@@ -5416,10 +5421,12 @@ QuicConnParamSet(
             QUIC_ALLOC_NONPAGED(BufferLength, QUIC_POOL_CLOSE_REASON);
 
         if (Connection->CloseReasonPhrase != NULL) {
```
nibanks, line 5423: Combine this if and the buffer if statements?

src/core/crypto.c
```diff
@@ -201,7 +201,7 @@ QuicCryptoInitialize(
 Exit:
 
     if (QUIC_FAILED(Status)) {
-        for (uint8_t i = 0; i < QUIC_PACKET_KEY_COUNT; ++i) {
+        for (size_t i = 0; i < QUIC_PACKET_KEY_COUNT; ++i) {
```
nibanks, line 204: Why?

src/core/crypto_tls.c
```diff
@@ -155,6 +155,9 @@ QuicCryptoTlsReadSniExtension(
     _Inout_ QUIC_NEW_CONNECTION_INFO* Info
     )
 {
+
+    UNREFERENCED_PARAMETER(Connection);
+
```
nibanks, line 160:
```suggestion
    UNREFERENCED_PARAMETER(Connection);
```

src/core/library.c
```diff
@@ -206,7 +206,6 @@ MsQuicLibraryInitialize(
             LibraryStorageOpenFailed,
             "[ lib] Failed to open global settings, 0x%x",
             Status);
-        Status = QUIC_STATUS_SUCCESS; // Non-fatal, as the process may not have access
```
nibanks, line 209: We should at least keep the comment.

src/core/range.c
```diff
@@ -278,7 +280,7 @@ QuicRangeAddRange(
                     QuicRangeCompare(&Key, Sub) == 0) {
                 --i;
             }
-            Sub = QuicRangeGet(Range, i);
+            QuicRangeGet(Range, i);
```
nibanks, line 283: This doesn't look right, does it?

nibanks: Well, I guess it's never being read, so the whole line should be removed?

src/core/binding.c
ThadHouse: I'm definitely not a big fan of replacing all else ifs with else statements, which it looks like this PR does a lot of.

src/core/configuration.c
```diff
@@ -408,34 +408,24 @@ QuicConfigurationParamGet(
         void* Buffer
     )
 {
-    QUIC_STATUS Status;
-
-    switch (Param) {
-    case QUIC_PARAM_CONFIGURATION_SETTINGS:
+    if (Param == QUIC_PARAM_CONFIGURATION_SETTINGS) {
```
ThadHouse, line 411: Isn't this a place where we're likely to add more options to the switch statement in the future? 

Also, I know multi returns is fine in C, but its definitely changing from what is mostly our standard. Definitely just want to double check that we're ok with this.

src/core/binding.c
nibanks: I actually don't mind those changes. After looking at the code a while, I think I might like it better.

src/core/configuration.c
nibanks: I am fine with these changes. If we do need to add more in the future, it's easy enough to change then.

src/core/crypto.c
ThadHouse: My guess is since its indexing, it prefers the size that can index into an array. The generated assembly is the same, as the variable is stuck in a register anyway, so I think this is ok. Less chance for accidental rollover bugs in the future, even though I know QUIC_PACKET_KEY_COUNT likely will not change.

src/core/connection.c
```diff
@@ -1612,7 +1612,7 @@ QuicConnTryClose(
             QuicConnLogStatistics(Connection);
         }
 
-        if (Flags & QUIC_CLOSE_APPLICATION) {
+        if (Flags & QUIC_CLOSE_APPLICATION) { // NOLINT
```
ThadHouse, line 1615: Why is it flagging so many if statements with // NOLINT? This flags one _maybe_ I can understand, but the one below on line 2036 is odd to me.

src/core/range.c
ThadHouse: Its unconditionally overwritten on line 294, so I understand why its flagged. If that is the case, is this call then needed?

src/core/stream.c
```diff
@@ -572,7 +572,6 @@ QuicStreamParamGet(
 
         if (!Stream->Flags.Started ||
             !Stream->Flags.LocalCloseAcked) {
-            Status = QUIC_STATUS_INVALID_STATE;
```
nibanks, line 575: Instead of removing this, add a `break` afterwards.

src/platform/datapath_linux.c
```diff
@@ -664,19 +664,21 @@ QuicDataPathPopulateTargetAddress(
...
 
-            return;
-        } else {
-            Address->Ipv6 = *SockAddrIn6;
             return;
         }
-    } else if (AddrInfo->ai_addr->sa_family == AF_INET) {
+        Address->Ipv6 = *SockAddrIn6;
+        return;
+
```
nibanks, line 671: nit: remove extra newline

src/platform/platform_linux.c
```diff
@@ -13,16 +13,16 @@
...
+#include "quic_trace.h"
+#include <dlfcn.h>
+#include <fcntl.h>
 #include <limits.h>
 #include <sched.h>
-#include <fcntl.h>
 #include <syslog.h>
-#include <dlfcn.h>
-#include "quic_trace.h"
-#include "quic_platform_dispatch.h"
```
nibanks, line 25: Don't remove this header.

src/platform/platform_linux.c
```diff
@@ -146,7 +146,7 @@ QuicPlatformInitialize(
 #ifdef QUIC_PLATFORM_DISPATCH_TABLE
     QUIC_FRE_ASSERT(PlatDispatch != NULL);
 #else
-    RandomFd = open("/dev/urandom", O_RDONLY);
+    RandomFd = open("/dev/urandom", O_RDONLY|O_CLOEXEC);
```
nibanks, line 149: I don't completely understand the intricacies here. Why is `O_CLOEXEC` being added?

Maximus-: Not sure why it was added here by Lars, but it's definitely good hygiene for programs that fork. That being said, I don't know what MsQuics expected behavior is in a process that forks ;P

src/core/binding.c
larseggert: The `else if` -> `if` changes address the [readability-else-after-return](https://clang.llvm.org/extra/clang-tidy/checks/readability-else-after-return.html) and similar warning. I prefer it, but we can also simply disable those checks.

src/core/binding.c
larseggert: Leave it, but comment it out, explaining why? Or `// NOLINT` the line?

src/core/configuration.c
larseggert: It stops a warning that `Configuration` is unused, because on Linux `QuicTraceEvent` is a no-op.

src/core/crypto.c
larseggert: Because `QUIC_PACKET_KEY_COUNT` is an `enum`, i.e., an `int`. Could also use `int`.

src/core/connection.c
larseggert: Because `QUIC_CLOSE_APPLICATION` is a (signed) `int`, and bitops on signed values are undefined.

src/platform/platform_linux.c
larseggert: https://clang.llvm.org/extra/clang-tidy/checks/android-cloexec-open.html

Can also just disable this check.

src/core/range.c
larseggert: I wasn't sure if the call has side effects. I guess not.

src/platform/datapath_linux.c
```diff
@@ -13,14 +13,14 @@
...
+#define _GNU_SOURCE // NOLINT
 #include "platform_internal.h"
-#include <sys/epoll.h>
-#include <sys/eventfd.h>
+#include "quic_platform_dispatch.h"
+#include <arpa/inet.h>
 #include <inttypes.h>
 #include <linux/in6.h>
-#include <arpa/inet.h>
-#include "quic_platform_dispatch.h"
```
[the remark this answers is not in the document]
larseggert: It's not removed, just reordered

src/platform/platform_linux.c
larseggert: Not removed, just reordered

src/core/connection.c
larseggert: The one on 2036 (and many others) is because both paths of the `if` have the same code - `QuicTraceEvent` is a no-op on Linux.

I can comment out this check instead?

src/core/binding.c
nibanks: I'd say `// NOLINT` the line.

src/core/configuration.c
nibanks: I'd prefer `UNREFERENCED_PARAMETER` instead, since that's what we do everywhere else.

src/platform/platform_linux.c
nibanks: I don't have the expertise here so I'll have to you defer to whatever you guys think is best. I believe I originally grabbed this code from stackoverflow.

nibanks:
Windows build is breaking:
```
D:\a\1\msquic\src\core\sent_packet_metadata.c(66,1): error C2220: the following warning is treated as an error [D:\a\1\msquic\build\windows\x86_schannel\src\core\core.vcxproj]
D:\a\1\msquic\src\core\sent_packet_metadata.c(66,1): warning C4267: 'initializing': conversion from 'size_t' to 'uint16_t', possible loss of data [D:\a\1\msquic\build\windows\x86_schannel\src\core\core.vcxproj]
```

src/core/connection.c
nibanks: The logging logic here is important. What's the correct way to fix this logic? Why is it treating `QUIC_CLOSE_APPLICATION` as a signed int?

src/core/connection.c
nibanks: I think I'd rather NOLINT these type of things.

src/core/connection.c
```diff
@@ -5811,9 +5812,9 @@ QuicConnParamGet(
         Stats->Misc.KeyUpdateCount = Connection->Stats.Misc.KeyUpdateCount;
 
         if (Param == QUIC_PARAM_CONN_STATISTICS_PLAT) {
-            Stats->Timing.Start = QuicTimeUs64ToPlat(Stats->Timing.Start);
-            Stats->Timing.InitialFlightEnd = QuicTimeUs64ToPlat(Stats->Timing.InitialFlightEnd);
-            Stats->Timing.HandshakeFlightEnd = QuicTimeUs64ToPlat(Stats->Timing.HandshakeFlightEnd);
+            Stats->Timing.Start = QuicTimeUs64ToPlat(Stats->Timing.Start); // cppcheck-suppress selfAssignment
+            Stats->Timing.InitialFlightEnd = QuicTimeUs64ToPlat(Stats->Timing.InitialFlightEnd); // cppcheck-suppress selfAssignment
+            Stats->Timing.HandshakeFlightEnd = QuicTimeUs64ToPlat(Stats->Timing.HandshakeFlightEnd); // cppcheck-suppress selfAssignment
```
nibanks, line 5817: What's up with these?

src/core/crypto.c
```diff
@@ -105,6 +105,7 @@ QuicCryptoInitialize(
             QUIC_MAX_TLS_SERVER_SEND_BUFFER : QUIC_MAX_TLS_CLIENT_SEND_BUFFER;
     uint16_t InitialRecvBufferLength =
         QuicConnIsServer(Connection) ?
+            // cppcheck-suppress duplicateValueTernary
```
nibanks, line 108: Can we just disable this warning/error globally instead?

src/core/packet.h
```diff
@@ -307,6 +307,7 @@ QuicPktNumEncode(
     )
 {
     for (uint8_t i = 0; i < PacketNumberLength; i++) {
+        // cppcheck-suppress objectIndex
```
nibanks, line 310: What's up here?

src/core/registration.c
```diff
@@ -243,37 +243,35 @@ MsQuicRegistrationShutdown(
         QUIC_TRACE_API_REGISTRATION_SHUTDOWN,
         Handle);
 
-    if (Handle && Handle->Type == QUIC_HANDLE_TYPE_REGISTRATION) {
```
nibanks, line 246: Why are you removing this?

src/core/send.c
```diff
@@ -118,7 +118,7 @@ _IRQL_requires_max_(DISPATCH_LEVEL)
 void
 QuicSendQueueFlush(
     _In_ QUIC_SEND* Send,
-    _In_ QUIC_SEND_FLUSH_REASON Reason
+    _In_ QUIC_SEND_FLUSH_REASON Reason // NOLINT
```
nibanks, line 121: I prefer `UNREFERENCED_PARAMETER` instead of `NOLINT` for variables only used in tracing.

src/core/stream.c
```diff
@@ -345,7 +345,7 @@ QuicStreamClose(
 _IRQL_requires_max_(PASSIVE_LEVEL)
 void
 QuicStreamTraceRundown(
-    _In_ QUIC_STREAM* Stream
+    _In_ QUIC_STREAM* Stream // NOLINT
```
nibanks, line 348: `UNREFERENCED_PARAMETER`

src/core/configuration.c
larseggert: I can make that change (also elsewhere). But don't you think it's confusing, since the parameter is actually referenced when `QuicTraceEvent` isn't a no-op?

src/core/connection.c
larseggert: The issue here is that both paths of the `if` are identical, because `QuicTraceEvent` is a no-op. That is what the `NOLINT` suppresses.

(I was wrong on the bitops thing; must have looked at some other code piece.)

src/core/connection.c
larseggert: `QuicTimeUs64ToPlat` is a no-op, so this is a pure (useless) self-assignment.

src/core/packet.h
larseggert: The warning is `The address of local variable 'PacketNumber' might be accessed at non-zero index. [objectIndex]`. 
Can also disable this globally?

src/core/registration.c
larseggert: Because it is asserted to be true in lines 233-234.

src/core/send.c
larseggert: See above. Am fine with this, but I felt it was confusing, because the variable is referenced when the trace call isn't a no-op.

src/platform/platform_linux.c
larseggert: There is IMO no downside to this change. @Maximus- ?

larseggert:
I don't understand what the `cmake` error on Windows is about.

src/core/registration.c
nibanks: They are debug asserts. They don't fire in release builds. We need these checks still.

src/core/packet.h
nibanks: Probably.

src/core/send.c
nibanks: `// NOLINT` isn't going to help cross-platform if logging is disabled, but `UNREFERENCED_PARAMETER` will. BTW, I don't want to worry/care/optimize-for the no-logging scenario since it's not the norm.

nibanks:
> I don't understand what the `cmake` error on Windows is about.

```D:\a\1\msquic\src\core\sent_packet_metadata.c(72,31): warning C4267: 'function': conversion from 'size_t' to 'uint32_t', possible loss of data [D:\a\1\msquic\build\windows\x64_schannel\src\core\core.vcxproj]
```
This error? The problem is that you're passing a `size_t` to a function (`QuicPoolInitialize`) that takes a `uint32_t` without an explicit cast.

larseggert:
Ah, so the error is two screens up in the log - cunning! :-)

larseggert:
@nibanks I think this is ready. The one failing run seems to be because clog didn't get downloaded for some reason.

src/core/unittest/FrameTest.cpp
```diff
@@ -173,7 +173,7 @@ TEST_P(AckFrameTest, DecodeAckFrameFail) {
         Buffer[2] = 1; // ACK Delay
         Buffer[3] = 0; // ACK range count
         Buffer[4] = 5; // First ACK range
-        for (auto TestValue : {64, 255}) {
+        for (uint8_t TestValue = 64; TestValue <= 255; TestValue++) {
```
nibanks, line 176: Why?

larseggert: Because otherwise it's an int, and then the Windows compiler throws an error about "loss of precision"

nibanks:
Looks like you broke some things, because various tests are all failing.

- SpinQuic is hanging
- Perf seems to crash

ThadHouse:
The perf crash is the following

```
0000008f`060ff820 00007ff6`c269653d     quicperf!_invoke_watson+0x18
0000008f`060ff850 00007ff6`c2696559     quicperf!_invalid_parameter+0xad
0000008f`060ff890 00007ff6`c2699b63     quicperf!_invalid_parameter_noinfo+0x19
0000008f`060ff8d0 00007ff6`c2699e9e     quicperf!__acrt_stdio_parse_mode<char>+0x2ab
(Inline Function) --------`--------     quicperf!common_openfile+0xc
0000008f`060ff900 00007ff6`c268a880     quicperf!_openfile+0x26
(Inline Function) --------`--------     quicperf!__crt_char_traits<char>::open_file+0xe
0000008f`060ff960 00007ff6`c254b9e1     quicperf!common_fsopen<char>+0x90
0000008f`060ff9a0 00007ff6`c2545d55     quicperf!QuicTlsGenerateSelfSignedCert+0x2d1
(Inline Function) --------`--------     quicperf!QuicPlatGetSelfSignedCert+0xd8
0000008f`060ffa30 00007ff6`c26819dc     quicperf!main+0x475
(Inline Function) --------`--------     quicperf!invoke_main+0x22
0000008f`060ffbd0 00007ff8`e68c4d0e     quicperf!__scrt_common_main_seh+0x10c
0000008f`060ffc10 00007ff8`e7315deb     KERNEL32!BaseThreadInitThunk+0x1e
0000008f`060ffc40 00000000`00000000     ntdll!RtlUserThreadStart+0x2b
```

src/core/unittest/FrameTest.cpp
ThadHouse: Isn't this going to infinite loop? Since 255 is true, then rollover will happen and the condition will still be true.

src/platform/selfsign_openssl.c
```diff
@@ -216,7 +216,7 @@ QuicTlsGenerateSelfSignedCert(
     fclose(Fd);
     Fd = NULL;
 
-    Fd = fopen(CertFileName, "wb");
+    Fd = fopen(CertFileName, "wbe");
```
ThadHouse, line 219: This is also called from windows, which doesn't accept the e flag. Since the e flag is specific only to gclibc, I would very much prefer this not be in here at all, because otherwise mac will break when we get to it too.

src/platform/platform_linux.c
ThadHouse: I worry that this would break macOS when we get to that PR.

src/platform/selfsign_openssl.c
larseggert: Darwin has support for `e`; but I have removed it anyway in the latest push.

larseggert:
So I think I fixed the crashes that I introduced. I will note that `spinquic` still crashes after some time, but it also does that when compiled from the current `main` branch. FWIW, here is the stack trace for those crashes:
```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:49
#1  0x00007f9bf1de6864 in __GI_abort () at abort.c:79
#2  0x0000565532153fba in quic_bugcheck () at ../src/platform/platform_linux.c:60
#3  0x0000565532152d91 in ServerSpin (Context=<optimized out>) at ../src/tools/spin/spinquic.cpp:625
#4  0x00007f9bf1fb3590 in start_thread (arg=0x7f9beb45e640) at pthread_create.c:463
#5  0x00007f9bf1ed9223 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```
```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:49
#1  0x00007f6efa25d864 in __GI_abort () at abort.c:79
#2  0x0000563c5f37ffba in quic_bugcheck () at ../src/platform/platform_linux.c:60
#3  0x0000563c5f37eeb1 in ServerSpin (Context=<optimized out>) at /usr/include/x86_64-linux-gnu/bits/byteswap.h:37
#4  0x00007f6efa42a590 in start_thread (arg=0x7f6ef38d5640) at pthread_create.c:463
#5  0x00007f6efa350223 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

nibanks:
It seems both Windows OpenSSL and Windows miTLS BVTs are failing to run and just timing out (which is incorrectly showing as a pass, btw).

larseggert:
Sigh. Is there a way to debug why it hangs based on the CI log?

nibanks:
> Sigh. Is there a way to debug why it hangs based on the CI log? I don't have a Windows box or any idea how to replicate this on one.
> 
> (And what's BVT?)

No logs or dumps were uploaded for the tests unfortunately (I think because they're not getting marked as a hard fail for some reason). Also BVT means "Build Verification Test".

larseggert:
@nibanks:
```
##[error]There is not enough space on the disk. : 'C:\a\_diag\pages\ab09ace5-9714-4291-b529-adb1a8db6aa3_1.log'
```

larseggert:
@nibanks also, is it possible to get more output from the BVT runs? i have really no idea how to debug the hangs.

nibanks:
> @nibanks also, is it possible to get more output from the BVT runs? i have really no idea how to debug the hangs.

I'm not sure how to expose build output publicly. There is no reason it needs to be hidden. Is there a way easily to split this PR into smaller changes? Or at least individual commits you can push through the CI one at a time?

larseggert:
I can try, but it will be some days until I can

larseggert:
Closing this in favor of #1085

PR #976, opened 2020-11-06
WPA Connection Table Support

nibanks:
Adds the connection worker table.


src/tools/wpa/DataModel/QuicConnection.cs
```diff
@@ -0,0 +1,234 @@
...
+                        BytesReceived = payload!.RecvTotalBytes;
+                        break;
+                    }
+                default:
+                    break;
+            }
+
+            FinalTimeStamp = evt.TimeStamp;
+
+            if (Worker != null)
```
ThadHouse, line 226: You don't need to change this unless you want to, but this call and others that you want to either null check or ignore can be done like `Worker?.OnConnectionEvent(evt);` The `?` will do a null check, and it does it a bit faster too. In multi threaded contexts it much more important, but since we're not its fine, and its just a slight cleanup if you want it in the future.

src/tools/wpa/DataModel/QuicObjectSet.cs
```diff
@@ -77,9 +75,9 @@ public override int GetHashCode()
 
     public sealed class QuicObjectSet<T> where T : class, IQuicObject
     {
-        private Dictionary<QuicObjectKey, T> activeTable = new Dictionary<QuicObjectKey, T>();
+        internal Dictionary<QuicObjectKey, T> activeTable = new Dictionary<QuicObjectKey, T>();
```
ThadHouse, line 78: If these are going to be internal but not properties, make them `readonly`

ThadHouse (approved):
Good other then the readonly thing.

src/tools/wpa/DataModel/QuicConnection.cs
nibanks: Ok. I will make the changes.

src/tools/wpa/DataModel/QuicObjectSet.cs
ThadHouse: Realistically there are some other fields that could use readonly as well in other places, but I'm fine doing that in a later cleanup.

PR #984, opened 2020-11-07
Finish Porting over WPA Plugin Code

nibanks:
This finishes up porting the rest of the internal WPA plugin for MsQuic. It also removes the need for the `etwlib` code (which I have now removed). With this change, the internal WPA plugin can now be removed.


ThadHouse (approved):

ThadHouse:
build.ps1 also depended on etwlib, so that will need to be patched as well.

PR #985, opened 2020-11-08
WPA Datapath Throughput Batching Table

nibanks:
- Adds parsing for datapath events.
- Adds table for throughput batching in "bytes per event".

### DIFF
--- a/src/tools/wpa/DataModel/QuicEventPayload.cs
+++ b/src/tools/wpa/DataModel/QuicEventPayload.cs
@@ -4,6 +4,7 @@
 //
 
 using System;
+using System.Net;
 using System.Runtime.InteropServices;
 
 namespace MsQuicTracing.DataModel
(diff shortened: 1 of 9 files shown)


src/tools/wpa/DataModel/QuicEventPayload.cs
```diff
@@ -279,4 +304,36 @@ public class QuicStreamOutFlowBlockedPayload
...
+
+    public class QuicDatapathSendPayload
+    {
+        public uint TotalSize { get; protected set; }
+
+        public byte BufferCount { get; protected set; }
+
+        public ushort SegmentSize { get; protected set; }
+
+#pragma warning disable CS8618 // Non-nullable field is uninitialized.
```
ThadHouse, line 318: A cleaner way to disable these is to set to `null!`

`public IPEndPoint RemoteAddress { get; protected set; } = null!` 

Its cleaner imo.

ThadHouse (approved):

src/tools/wpa/DataModel/QuicEventPayload.cs
nibanks: What I really want is to essentially declare all these classes as abstract and require the derived classes' constructors to implement all these. Additionally, I only want the derived classes to have to implement the constructor and not redefine the types (like you have to do for interfaces, right?).

ThadHouse: Theres no way to do that in c#. If it's defined in the base class, it has to be fully initialized in the base class code only. That analysis doesn't go up to derived classes. And if you make the property abstract, you'd have to define it in every derived class.

PR #990, opened 2020-11-09
Fail build if throughput up test has a 5% or more performance drop

### LINKED ISSUE #949: Fail Perf Tests for Large Deviations

Some of the perf tests are stable enough (upload) that we should be able to flag any significant deviations pretty easily. We should update these tests to fail if they regress by some unacceptable threshold.

ThadHouse:
Closes #949. Once we can stabilize the rests of the tests, we should make those failures as well.

### DIFF
--- a/scripts/performance-helper.psm1
+++ b/scripts/performance-helper.psm1
@@ -538,6 +538,9 @@ function Write-Failures() {
     }
 }
 
+# Fail loopback tests if < 50%
+$LocalRegressionThreshold = -50.0
+
 function Get-LatestThroughputRemoteTestResults([ThroughputRequest]$Request) {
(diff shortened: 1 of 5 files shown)


scripts/performance-helper.psm1
```diff
@@ -566,6 +580,8 @@ class ThroughputTestPublishResult {
     }
 }
 
+$ThroughputUpRegressionThreshold = -5.0
```
nibanks, line 583: This should be in the json, per test, to say what the "failure threshold" is. For throughput up, 5% is good. For RPS, 50% seems to be required.

nibanks (approved):

nibanks (approved):

PR #995, opened 2020-11-09
Add Delay Send API Flag

nibanks:
Adds a new API send flag for StreamSend that allows the app to indicate that the send shouldn't be done immediately. This is meant to be used as a hint to QUIC to indicate "more is coming". This allows the app to control the amount of "send batching" that occurs at the QUIC/UDP layer.


src/core/send.c
```diff
@@ -153,11 +154,20 @@ QuicSendQueueFlushForStream(
         QuicStreamAddRef(Stream, QUIC_STREAM_REF_SEND);
     }
 
-    if (Stream->Connection->State.Started) {
+    //
+    // TODO - If send was delayed by the app, under what conditions should we
+    // ignore that signal and queue anyways?
```
nibanks, line 159: I don't think we need to do this for this PR. This would be purely an optimization or performance improvement on top of this feature.

maolson-msft: Is this the case where a send was delayed, but then a flush still happened for whatever reason? I think it makes sense to go ahead and send the data in this case, right? [if it's as simple as that, then I don't see why this should be deferred; but if there needs to be a more subtle set of conditions, then I agree with deferral]

nibanks: I think you're asking/clarifying the following (which is true): If **anything** else causes a send to flush (on the entire connection) we're going to end up flushing the stream data too. This comment is just here to say that if we have nothing else to send and the app keeps queuing these delayed sends, it's probably a good idea to send anyways once we reach some threshold (i.e. 10k queued or something).

anrossi: Why not just send when we have a "full batch"?

anrossi: In this case, a "Full batch" would be either the max encryption batch size, or the max USO size.

nibanks: Two things:

1. There is really no such thing as a "full batch". We theoretically have a maximum batch size, but it's kind of dynamic.
2. We don't currently track "how much unsent data is currently queued", only "how much data is queued (some or all of which might be inflight already)".

Those two things make it hard to add some better logic here without further changes.

src/core/send.c
```diff
@@ -317,7 +327,8 @@ BOOLEAN
 QuicSendSetStreamSendFlag(
     _In_ QUIC_SEND* Send,
     _In_ QUIC_STREAM* Stream,
-    _In_ uint32_t SendFlags
+    _In_ uint32_t SendFlags,
+    _In_ BOOLEAN DelaySend
```
anrossi, line 331: why does this need to be a separate field instead of just another `SendFlag`?

nibanks: This isn't a flag because flags represent something to send. The design here is to queue all things that need to be sent, but don't actually flush the data yet. If something else results in a flush, go ahead and process all the flags then.

anrossi: Ok, so this means that if something else needs to be sent (ACK, Ping, etc.) it won't be delayed by stream data that is delayed?

nibanks: Notice this function is stream specific. Nothing connection-wide goes through here, so those definitely trigger an immediate flush. Non-data stream frames will also trigger immediate flush too.

src/test/lib/EventTest.cpp
```diff
@@ -17,6 +17,9 @@
 #define QUIC_EVENT_ACTION_SHUTDOWN_CONNECTION   1
 #define QUIC_EVENT_ACTION_SHUTDOWN_STREAM       2
 
+const uint8_t StreamPayload[64] = {0x1};
```
anrossi, line 20: This will only set the first byte to `0x1`, it won't initialize the entire array to `0x1`. Just checking assumptions.

nibanks: I don't care really. :) I just put something here.

ThadHouse:
Based on a suggestion from .NET, I feel like allowing a 0 length send with the flush flag is a good idea to add. A method to explicitly flush without checking that you're at the end of your data actively while looping is a good idea to have, especially for high level wrappers. Not worth adding a new API to do in C, especially if we could just plumb 0 length send to work.

src/core/send.c
ThadHouse: Before we can merge, we need to make a decision on this TODO I think.

ThadHouse (approved):

PR #998, opened 2020-11-09
Add a minimum version to cmake, and add notes in the readme how to update

ThadHouse:


docs/BUILD.md
```diff
@@ -129,6 +129,8 @@ The following section details how to build MsQuic purely with CMake commands.
 
 Note that you will need to disable logging if building with CMake exclusively. Logging enabled requires .NET Core and at least the configuration from prepare-machine.ps1 in order to build.
 
+Note at minimum CMake 3.17 is required. The default installed with Ubuntu 18.04 is too old, and will not correctly build. Instructions for installing a newer version can be found here. https://apt.kitware.com/
```
nibanks, line 132: Don't include "The default installed with Ubuntu 18.04 is too old, and will not correctly build." Also, can you also include this in the cmake only instructions at the end of this file?

ThadHouse: This is in the CMake only section. I'll add it to the normal section too.

nibanks (approved):

PR #1006, opened 2020-11-10
Move Plugin Solutions

nibanks:
Just moves the VS solution files to the plugin folder.


ThadHouse (approved):

ThadHouse:
Moving the solution dir breaks the artifact locations. Should be an easy enough fix, to just go ../..

PR #1020, opened 2020-11-11
Add SSLKEYLOGFILE support to quicinterop client

### LINKED ISSUE #873: Output SSLKEYLOGFILE for interop client (and maybe server)
This will enable QNS/wireshark to decode the handshake.
Add a new private GetParam parameter to enable exporting the TLS secrets, and update the interop client/server to have an option to export these.
See [here for file format](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/Key_Log_Format)

anrossi:
Schannel and OpenSSL TLS libraries are currently supported, and mitls only exports application traffic secrets.
quicinteropserver could also be updated to support SSLKEYLOGFILE using this mechanism.
Enabled by build flag `QUIC_TLS_SECRETS_SUPPORT`.
Fixes #873


src/core/connection.c
```diff
@@ -6399,3 +6423,34 @@ QuicConnDrainOperations(
...
+        memcpy(Connection->SslKeyLogBuffer, Line, min(LineLen, Connection->SslKeyLogBytesRemaining));
+        if (LineLen > Connection->SslKeyLogBytesRemaining) {
+             LineLen = Connection->SslKeyLogBytesRemaining;
+        }
+        Connection->SslKeyLogBytesRemaining -= (uint32_t)LineLen;
+        Connection->SslKeyLogBuffer += (uint32_t)LineLen;
+        //
+        // Null-terminate buffer.
+        //
+        *Connection->SslKeyLogBuffer = 0;
```
anrossi, line 6453: This writes past the end of the buffer

src/core/connection.c
```diff
@@ -6399,3 +6423,34 @@ QuicConnDrainOperations(
...
+        size_t LineLen = strlen(Line);
+        //
+        // Append newline on previous log line.
+        //
+        if (Connection->SslKeyLogBytesRemaining > 0) {
+            *Connection->SslKeyLogBuffer = '\n';
+            Connection->SslKeyLogBuffer++;
+            Connection->SslKeyLogBytesRemaining--;
+        }
+        memcpy(Connection->SslKeyLogBuffer, Line, min(LineLen, Connection->SslKeyLogBytesRemaining));
```
anrossi, line 6444: change `min(LineLen, Connection->SslKeyLogBytesRemaining)` to  `min(LineLen, Connection->SslKeyLogBytesRemaining - 1)` to account for NULL at the end.

src/core/connection.c
```diff
@@ -6399,3 +6423,34 @@ QuicConnDrainOperations(
...
+    _In_ QUIC_CONNECTION* Connection,
+    _In_z_ const char* Line
+    )
+{
+    if (Connection->SslKeyLogBuffer != NULL && Connection->SslKeyLogBytesRemaining > 0) {
+        size_t LineLen = strlen(Line);
+        //
+        // Append newline on previous log line.
+        //
+        if (Connection->SslKeyLogBytesRemaining > 0) {
```
anrossi, line 6439: Change this to `> 1` to give room for NULL, and skip the memcpy in an `else` clause.

CMakeLists.txt
```diff
@@ -126,6 +126,7 @@ option(QUIC_CODE_CHECK "Run static code checkers" OFF)
 option(QUIC_OPTIMIZE_LOCAL "Optimize code for local machine architecture" OFF)
 option(QUIC_CI "CI Specific build optimizations" OFF)
 option(QUIC_RANDOM_ALLOC_FAIL "Randomly fails allocation calls" OFF)
+option(QUIC_WIRESHARK_SUPPORT "Enable Wireshark to decode packets" OFF)
```
nibanks, line 129: I think I'd prefer to just use SSLKEYLOGFILE instead of WIRESHARK. Other tools can/may use the file as well.

src/core/connection.c
```diff
@@ -5626,6 +5626,30 @@ QuicConnParamSet(
...
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+
+        if (Connection->State.Started) {
+            Status = QUIC_STATUS_INVALID_STATE;
+            break;
+        }
+
+        Connection->SslKeyLogBuffer = (uint8_t*)Buffer;
```
nibanks, line 5642: Instead of a generic buffer, I'd rather standardize this around a type specific struct that contains arrays for each key type. Then, the connection has the (optional) pointer to the struct. If it's set, then it will write the struct arrays, per key.

src/platform/tls_openssl.c
```diff
@@ -674,6 +696,9 @@ QuicTlsSecConfigCreate(
 
         SSL_CTX_set_max_early_data(SecurityConfig->SSLCtx, UINT32_MAX);
         SSL_CTX_set_client_hello_cb(SecurityConfig->SSLCtx, QuicTlsClientHelloCallback, NULL);
+#ifdef QUIC_WIRESHARK_SUPPORT
+        SSL_CTX_set_keylog_callback(SecurityConfig->SSLCtx, QuicTlsSslKeyLogCallback);
```
nibanks, line 700: We should be able to extract this info ourselves when we create the keys, right?

src/core/connection.c
```diff
@@ -5626,6 +5626,29 @@ QuicConnParamSet(
...
+        if (Connection->State.Started) {
+            Status = QUIC_STATUS_INVALID_STATE;
+            break;
+        }
+
+        Connection->SslKeyLog = (QUIC_SSLKEYLOG*)Buffer;
+
+        QuicZeroMemory(Connection->SslKeyLog, sizeof(*Connection->SslKeyLog));
+
+        Status = QUIC_STATUS_SUCCESS;
```
nibanks, line 5646: nit
```suggestion
        Connection->SslKeyLog = (QUIC_SSLKEYLOG*)Buffer;
        QuicZeroMemory(Connection->SslKeyLog, sizeof(*Connection->SslKeyLog));
        Status = QUIC_STATUS_SUCCESS;
```

src/core/crypto.c
```diff
@@ -1343,6 +1346,22 @@ QuicCryptoProcessTlsCompletion(
     }
 
     if (ResultFlags & QUIC_TLS_RESULT_DATA) {
+#ifdef QUIC_SSLKEYLOG_SUPPORT
+        //
+        // Parse the client initial to populate the SSLKEYLOG with the
+        // ClientRandom
+        //
+        if (!QuicConnIsServer(Connection) &&
```
nibanks, line 1354: Only if `Connection->SslKeyLog` is set, right?

src/inc/msquicp.h
```diff
@@ -65,6 +65,26 @@ typedef struct QUIC_PRIVATE_TRANSPORT_PARAMETER {
...
+typedef struct QUIC_SSLKEYLOG {
+    uint8_t ClientRandom[QUIC_SSLKEYLOG_MAX_SECRET_LEN];
+    uint8_t ClientEarlyTrafficSecret[QUIC_SSLKEYLOG_MAX_SECRET_LEN];
+    uint8_t ClientHandshakeTrafficSecret[QUIC_SSLKEYLOG_MAX_SECRET_LEN];
+    uint8_t ServerHandshakeTrafficSecret[QUIC_SSLKEYLOG_MAX_SECRET_LEN];
+    uint8_t ClientTrafficSecret0[QUIC_SSLKEYLOG_MAX_SECRET_LEN];
+    uint8_t ServerTrafficSecret0[QUIC_SSLKEYLOG_MAX_SECRET_LEN];
+    uint8_t EarlyExporterSecret[QUIC_SSLKEYLOG_MAX_SECRET_LEN];
+    uint8_t ExporterSecret[QUIC_SSLKEYLOG_MAX_SECRET_LEN];
+    uint8_t ClientRandomLength;
```
nibanks, line 78: Is random not a fixed length?

src/inc/msquicp.h
```diff
@@ -65,6 +65,26 @@ typedef struct QUIC_PRIVATE_TRANSPORT_PARAMETER {
...
+    uint8_t ClientEarlyTrafficSecretLength;
```
nibanks, line 79: Are all the secrets same length (if present)? If so, might be best to have one length and bitfield for which are set?

src/inc/msquicp.h
```diff
@@ -78,6 +98,7 @@ typedef struct QUIC_PRIVATE_TRANSPORT_PARAMETER {
 #define QUIC_PARAM_CONN_FORCE_KEY_UPDATE                0x80000001  // No payload
 #define QUIC_PARAM_CONN_FORCE_CID_UPDATE                0x80000002  // No payload
 #define QUIC_PARAM_CONN_TEST_TRANSPORT_PARAMETER        0x80000003  // QUIC_PRIVATE_TRANSPORT_PARAMETER
+#define QUIC_PARAM_CONN_SSLKEYLOG_BUFFER                0x80000004  // Set-only, sizeof(QUIC_SSLKEYLOG)
```
nibanks, line 101: I don't think we call out "set-only" in any of the other comments. Just put `QUIC_SSLKEYLOG` here,

src/tools/interop/interop.cpp
```diff
@@ -353,6 +355,8 @@ class InteropConnection {
     char* NegotiatedAlpn;
     const uint8_t* ResumptionTicket;
     uint32_t ResumptionTicketLength;
+    uint8_t* SslKeyLogBuffer;
```
nibanks, line 358: Need to fix up this type, and the rest of the code.

src/platform/tls_openssl.c
```diff
@@ -10,6 +10,7 @@
 --*/
 
 #include "platform_internal.h"
+#include "msquicp.h"
```
nibanks, line 13: Can you add this to platform_internal.h?

src/platform/tls_openssl.c
```diff
@@ -87,6 +88,13 @@ typedef struct QUIC_TLS {
     QUIC_CONNECTION* Connection;
     QUIC_TLS_RECEIVE_TP_CALLBACK_HANDLER ReceiveTPCallback;
 
+#ifdef QUIC_SSLKEYLOG_SUPPORT
+    //
+    // Struct to log TLS traffic secrets.
```
nibanks, line 93: Indicate it's optional?

CMakeLists.txt
```diff
@@ -226,6 +227,10 @@ if(QUIC_RANDOM_ALLOC_FAIL)
     list(APPEND QUIC_COMMON_DEFINES QUIC_RANDOM_ALLOC_FAIL=1 QUIC_DISABLE_MEM_POOL=1)
 endif()
 
+if(QUIC_TLSSECRETS_SUPPORT)
```
nibanks, line 230: I think it needs a _ between TLS and secrets. Otherwise it's hard to read.

src/core/crypto_tls.c
```diff
@@ -608,7 +617,11 @@ QuicCryptoTlsReadInitial(
                 Connection,
                 Buffer + TLS_MESSAGE_HEADER_LENGTH,
                 MessageLength,
-                Info);
+                Info
+#ifdef QUIC_TLS_SECRETS_SUPPORT
+                , Connection->TlsSecrets
+#endif
```
nibanks, line 623: Nothing in these functions ever reads from `Connection` (I'm pretty sure). It's only passed in for logging reasons. I'd prefer to keep it that way, so we could more easily unit test these functions in the future. So can we just pass in this (optional) pointer please to `QuicCryptoTlsReadInitial`?

src/inc/msquicp.h
```diff
@@ -65,6 +65,29 @@ typedef struct QUIC_PRIVATE_TRANSPORT_PARAMETER {
     const uint8_t* Buffer;
 } QUIC_PRIVATE_TRANSPORT_PARAMETER;
 
+#define QUIC_TLS_SECRETS_MAX_SECRET_LEN 64
```
nibanks, line 68: Probably worth a comment above all this explaining what these are for.

src/platform/tls_mitls.c
```diff
@@ -2226,6 +2236,39 @@ QuicPacketKeyCreate(
...
 
+#ifdef QUIC_TLS_SECRETS_SUPPORT
+    if (TlsContext->TlsSecrets != NULL) {
+        switch (KeyType) {
+        case QUIC_PACKET_KEY_1_RTT: {
+            quic_secret ClientReadSecret, ServerReadSecret;
+            if (FFI_mitls_quic_get_record_secrets(
+                TlsContext->miTlsState,
+                &ClientReadSecret,
+                &ServerReadSecret)) {
```
nibanks, line 2247:
```suggestion
                    TlsContext->miTlsState,
                    &ClientReadSecret,
                    &ServerReadSecret)) {
```

src/tools/interop/interop.cpp
```diff
@@ -795,6 +923,9 @@ RunInteropTest(
                 !Connection.GetResumptionTicket(ResumptionTicket, ResumptionTicketLength)) {
                 break;
             }
+            if (SslKeyLogFile) {
+                Connection.WriteSslKeyLogFile(SslKeyLogFile);
+            }
```
nibanks, line 928: There's got to be a better way instead of doing this instead of copying this all over the place. Maybe pass the `SslKeyLogFile` to the `Connection` and on cleanup if it's set, write the file?

Though, that brings up another point. What synchronization are you doing around the file. If you open a lot of connection in parallel, they'd all be writing to the file from multiple parallel threads, right? That would result in a big jumble.

src/inc/msquichelper.h
```diff
@@ -494,4 +509,101 @@ FreeServerConfiguration(
...
+    if (File == nullptr) {
+        printf("Failed to open sslkeylogfile %s\n", FileName);
+        return;
+    }
+    if (fseek(File, 0, SEEK_END) == 0 && ftell(File) == 0) {
+        fprintf(File, "# TLS 1.3 secrets log file, generated by quicinterop\n");
+    }
+    char ClientRandomBuffer[(2 * sizeof(QUIC_TLS_SECRETS::ClientRandom)) + 1] = {0};
+    char TempHexBuffer[(2 * QUIC_TLS_SECRETS_MAX_SECRET_LEN) + 1] = {0};
+    if (TlsSecrets.IsSet.ClientRandom) {
```
nibanks, line 538: Does it make sense to do anything in this function at all if `ClientRandom` isn't set?

nibanks (approved):

src/inc/msquichelper.h
anrossi: Well, you can manually populate the CLIENT_RANDOM if you're looking at a wireshark capture of only one connection.

PR #1021, opened 2020-11-11
Initial Command Line Wrapper for WPA Plugin

nibanks:
This just adds a **very** trivial wrapper to the WPA plugin. Eventually, all quicetw native code will be migrated to this.


ThadHouse:
Once .NET 5 hits AZP, we'll likely want to switch this to that, because it makes single file easier. But for now 3.1 is fine. want to

ThadHouse (approved):

nibanks:
This won't stay 1 file for long.

ThadHouse:
Not 1 file like that. With .NET 5 you can make a single file executable. So its easier to run.

PR #1029, opened 2020-11-13
Fix the key string for TestName

iriszero:
When `test` is used, at least in Ubuntu 18.04 LTS, the line 111-116 failed to parse the `argv` and get the input value.

### DIFF
--- a/src/perf/lib/quicmain.cpp
+++ b/src/perf/lib/quicmain.cpp
@@ -69,7 +69,11 @@ QuicMainStart(
         return QUIC_STATUS_INVALID_PARAMETER;
     }
 
-    const char* TestName = GetValue(argc, argv, "test");
+    const char* TestName = GetValue(argc, argv, "TestName");
+    if (TestName == nullptr) {
+        TestName = GetValue(argc, argv, "test");
+    }
+
     ServerMode = TestName == nullptr;
 
     QUIC_STATUS Status;


src/perf/lib/quicmain.cpp
```diff
@@ -69,7 +69,7 @@ QuicMainStart(
         return QUIC_STATUS_INVALID_PARAMETER;
     }
 
-    const char* TestName = GetValue(argc, argv, "test");
+    const char* TestName = GetValue(argc, argv, "TestName");
```
nibanks, line 72: I don't mind having `test` to be an alias to `TestName` here, but we need to support both. For instance, if `TestName` isn't present, then check for `test`.

iriszero: > I don't mind having `test` to be an alias to `TestName` here, but we need to support both. For instance, if `TestName` isn't present, then check for `test`.

That would be better:
a simple short circuit may do?
```
const char* TestName = GetValue(argc, argv, "TestName") || GetValue(argc, argv, "test");
```

nibanks: Essentially, but I think you'll need separate statements.

iriszero: @nibanks Please check out 00bd717, linked below.

src/perf/lib/quicmain.cpp
```diff
@@ -69,9 +69,15 @@ QuicMainStart(
         return QUIC_STATUS_INVALID_PARAMETER;
     }
 
-    const char* TestName = GetValue(argc, argv, "test");
-    ServerMode = TestName == nullptr;
-
+    const char* TestName;
+    
+    if ((TestName = GetValue(argc, argv, "TestName")) == nullptr) {
```
nibanks, line 74: Instead, simply add a line after 72 to get test if Test name variable is null.

nibanks (approved):

PR #1032, opened 2020-11-13
Fix outputting SSLKEYLOGFILE in QNS

anrossi:


scripts/run_endpoint.sh
```diff
@@ -69,6 +69,7 @@ if [ "$ROLE" == "client" ]; then
     REQ=${REQS[0]}
     SERVER=$(echo $REQ | cut -d'/' -f3 | cut -d':' -f1)
     echo "Connecting to $SERVER"
+    echo "Client params (before files):$CLIENT_PARAMS"
```
anrossi, line 72: I am leaving this in because whenever we have to debug this kind of thing, the first change is to add it (or something like it) back into the file, and having it there will reduce the debugging time by 1 round trip.

nibanks (approved):

PR #1036, opened 2020-11-15
Refactoring WPA Plugin

nibanks:
Refactoring the WPA plugin so it can be better used with the command line tool (e.g. converting logs to text).


src/plugins/wpa/DataModel/ETW/QuicEtwEvent.cs
```diff
@@ -43,28 +46,79 @@ internal enum QuicEtwEventOpcode : byte
...
+        {
+            return PointerSize == 8 ? ReadValue<ulong>() : ReadValue<uint>();
+        }
+
+        internal string ReadString()
+        {
+            var chars = new List<char>();
+            while (true)
+            {
+                byte c = ReadValue<byte>();
```
ThadHouse, line 86: Are these strings guaranteed to be ASCII only? If not, might be better to go into a List<byte>, then do Encoding.UTF8.GetString() to convert the UTF8 into UTF16. 

Also, I'd probably try catch ArgumentException here and return an empty string or "Invalid String" here if something goes wrong with this input. Try Catches are free in C# (Well the try part assuming nothing is caught is), and string parsing is definitely sensitive. But less important then the conversion part above.

src/plugins/wpa/DataModel/QuicEvent.cs
```diff
@@ -125,36 +134,83 @@ public enum QuicEventId : ushort
...
+    //
+    // The base class for all QUIC events.
+    //
+    public class QuicEvent : IKeyedDataType<Guid>
     {
-        public abstract Guid Provider { get; }
+        //
+        // Global configuration to control how parsing works. Defaults to WPA filter mode.
+        //
+        public static QuicEventParseMode ParseMode = QuicEventParseMode.WpaFilter;
```
ThadHouse, line 145: If you make this a property, you can remove 2211. You can preassign get and set properties just fine.

`public static QuicEventParseMode ParseMode { get; set; } = QuicEventParseMode.WpaFilter;`

src/plugins/wpa/DataModel/ETW/QuicEtwEvent.cs
ThadHouse: Actually don't worry about the try catch part. If we're going to do any of that, we should just do that in TryCreate, and return null if we catch, so we'll catch all parsing errors.

ThadHouse: All these guaranteed to be ASCII strings then?

nibanks: Yes.
```
            <data
                inType="win:AnsiString"
                name="ErrStr"
                />
```

PR #1041, opened 2020-11-18
Lots of WPA Improvements

nibanks:


src/plugins/wpa/Tables/QuicTxBlockedTable.cs
```diff
@@ -112,6 +119,20 @@ public static bool IsDataAvailable(IDataExtensionRetrieval tableData)
             return quicState != null && quicState.DataAvailableFlags.HasFlag(QuicDataAvailableFlags.ConnectionFlowBlocked);
         }
 
+        internal struct Data
```
ThadHouse, line 122: TBH, because of how internally I suspect WPA works, this might actually be much more performant as a class. If there ends up being a lot of objects. Not a big deal now, something to consider if this becomes an issue.

ThadHouse (approved):

anrossi (approved):

src/plugins/wpa/Tables/QuicTxBlockedTable.cs
nibanks: Seems ok as is. We can adjust later if necessary.

PR #1042, opened 2020-11-18
Add printout of total time executing to quicinterop

anrossi:


src/tools/interop/interop.cpp
```diff
@@ -1100,6 +1103,11 @@ RunInteropTests()
         PrintTestResults((uint32_t)EndpointIndex);
     }
     printf("\n");
+    printf(
+        "Total execution time: %u.%us\n",
```
nibanks, line 1107:
```suggestion
        "Total execution time: %u.%03us\n",
```
This is necessary so 3/1000 doesn't get printed as `.3`

nibanks (approved):

PR #1047, opened 2020-11-18
Pacing Improvements

nibanks:
I've been playing around with improving/simplifying the pacing logic. I've done the following:

- Removed the min send allowance for pacing. Pacing was being too aggressive from my testing.
- Reduced the pacing timer interval to increase spreading.
  - Windows timers can't actually achieve this reliably, but if don't actually end up waiting (because the worker has other work in the meantime) when we do get around to evaluating the timer wheel, we can see that the timer elapsed sooner.
- Reset the last send time after we get unblocked on congestion control. If we're blocked on congestion control, that means we probably paced too fast or CWND dropped. Either way, we whenever we finally get unblocked we shouldn't use the last send time (which was likely quite a while ago) to calculate the burst size, as it can result in quite a large burst.

I'm keeping this as a draft for now, because while my testing on my home network (10 Mbps upload cap) seems to do better with this, I'd like to onboard our network-limited testing (w/ duonic) to do more complete A/B testing before I merge this.


src/core/congestion_control.c
```diff
@@ -160,59 +160,38 @@ QuicCongestionControlGetSendAllowance(
...
-        // everything we can.
+        // We're not in the necessary state to pace.
         //
         SendAllowance = Cc->CongestionWindow - Cc->BytesInFlight;
 
     } else {
+
         //
-        // Try to pace: if the window and RTT are large enough, the window can
-        // be split into chunks which are spread out over the RTT.
```
maolson-msft, line 173: The "split into chunks which are spread out over the RTT" bit is essential for understanding what our "pacing" is doing and not confusing it with the other type of "pacing", so let's keep it.

src/core/congestion_control.c
```diff
@@ -160,59 +160,38 @@ QuicCongestionControlGetSendAllowance(
...
-        // SendAllowance will be set to the size of the next chunk.
+        // We are pacing, so calculate the current send allowance based on the
```
maolson-msft, line 176: Change "based on" to "as".

nibanks:
I flipped this PR to ready for review. The data from my A:B test definitely shows improvements.

Original Code:

image

New Code:

image

Rate Diff Between Original and New:

image

src/core/congestion_control.c
```diff
@@ -160,59 +160,39 @@ QuicCongestionControlGetSendAllowance(
...
+        // Since the window grows via ACK feedback and since we defer packets
+        // when pacing, using the current window to calculate the pacing
+        // interval is not quite as aggressive as we'd like. Instead, use the
+        // predicted window of the next round trip.
+        //
+        uint64_t EstimatedWnd = QuicCongestionControlPredictNextWindow(Cc);
 
-            SendAllowance =
-                (uint32_t)((EstimatedWnd * TimeSinceLastSend) / Connection->Paths[0].SmoothedRtt);
-            if (SendAllowance < MinChunkSize) {
```
nibanks, line 207: My only thought/concern on this PR is that if we flush send almost immediately after getting an ACK when we were CC blocked, we might end up having a really small SendAllowance. Is that Ok? Or should we keep a 1 MTU minimum?

maolson-msft: Can you elaborate?

nibanks: Ok, assume at t=0 we're currently blocked on CC and we receive an ACK for inflight data. That reduces inflight below cwnd and now we're no longer blocked on CC, so we queue a "flush send" operation. We also reset the last send time to Now(). The flush operation is then processed a few microseconds later. We then calculate send allowance based on the elapsed time of Now() - LastSend, multiple that by the calculated rate. It might result in a very small value (perhaps << MTU).

ThadHouse (approved):

src/core/congestion_control.c
nibanks: Data looks good. I will continue to test and possibly improve this area in the future. For now, I leave it as is.

PR #1048, opened 2020-11-19
QuicPerf Throughput Improvements

nibanks:
- Adds pacing disable option
- Adds ability to upload/download based on time instead of length

### DIFF
--- a/src/perf/lib/ThroughputClient.cpp
+++ b/src/perf/lib/ThroughputClient.cpp
@@ -34,9 +34,9 @@ PrintHelp(
         "  -encrypt:<0/1>              Enables/disables encryption. (def:1)\n"
         "  -sendbuf:<0/1>              Whether to use send buffering. (def:1)\n"
         "  -pacing:<0/1>               Whether to use pacing. (def:1)\n"
-        "  -upload:<####>              The length of data to send. (def:0)\n"
-        "  -download:<####>            The length of data to request/receive. (def:0)\n"
-        "  -timed:<####>               The indicates the upload/download arg time (ms). (def:0)\n"
+        "  -timed:<0/1>                The indicates the upload/download arg time (ms). (def:0)\n"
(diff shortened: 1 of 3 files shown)


src/perf/lib/ThroughputClient.cpp
```diff
@@ -33,8 +33,10 @@ PrintHelp(
         "  -ip:<0/4/6>                 A hint for the resolving the hostname to an IP address. (def:0)\n"
         "  -encrypt:<0/1>              Enables/disables encryption. (def:1)\n"
         "  -sendbuf:<0/1>              Whether to use send buffering. (def:1)\n"
+        "  -pacing:<0/1>               Whether to use pacing. (def:1)\n"
         "  -upload:<####>              The length of data to send. (def:0)\n"
         "  -download:<####>            The length of data to request/receive. (def:0)\n"
+        "  -timed:<####>               The indicates the upload/download arg time (ms). (def:0)\n"
```
ThadHouse, line 39: Took me a second with this API to figure out how to differenciate between timed up and timed down. I wonder if instead of just a "timed" flag, we should have a "timedup" and "timeddown" flag separate. Just to make the caller a bit cleaner. Either that or a flag for upload vs download, then a separate flag for either length or timed.

anrossi: I'm in favor of a flag for upload/download and then a param for length or timed.

nibanks: I really prefer this design. Either way you slice it, you need at least three parameters/variables to store the state. I just prefer this arg model.

ThadHouse: Its just confusing on what you put for the upload/download parameter in timed mode, so if you want to go with that please expand the help to explain how that works. Because its very non obvious.

nibanks: Just pushed improved help text. Please let me know if it works for you.

anrossi (approved):

ThadHouse (approved):

PR #1050, opened 2020-11-20
Use clog 0.1.9

ThadHouse:
This enables using .NET 5 as well as .NET Core 3.1. Docs have been updated to note this


nibanks (approved):

ThadHouse:
The azure build was really weird here. It basically didn't run any tests at all. Reruning just to ensure things are not horribly broken.

/azp run

PR #1052, opened 2020-11-20
Make QNS templates use latest ubuntu image

ThadHouse:
Default is Ubuntu 16.04, which has Python 3.5, which isn't supported anymore by a dependency of the QNS runner. Just update to the latest to solve the issue


nibanks:
Please run the QNS job (building a custom image) on this before we merge.

nibanks (approved):

PR #1056, opened 2020-11-21
Add a missing dependency: Threads

iriszero:
In a custom kernel (variants of Linux), `quicperf` gave me a following message:

```
/msquic/build/bin/Debug/quicperf: ignoring missing symbol pthread_attr_setschedparam
```
, which was solved by the attached fix.

Despite the running environment differing, I think it is still `msquic`'s responsibility to ensure seamless linking.
On an Ubuntu kernel, there occured no such message; maybe the kernel supplied an appropriate `pthread` library to the executable (maybe by default).

I will leave other possible, lurking dependency issues to others.

### DIFF
--- a/src/perf/bin/CMakeLists.txt
+++ b/src/perf/bin/CMakeLists.txt
@@ -12,4 +12,5 @@ add_executable(quicperf ${SOURCES} histogram/hdr_histogram.c)
 
 set_property(TARGET quicperf PROPERTY FOLDER "perf")
 
-target_link_libraries(quicperf inc warnings perflib msquic platform perfbin.clog)
+find_package(Threads)
+target_link_libraries(quicperf inc warnings perflib msquic platform perfbin.clog ${CMAKE_THREAD_LIBS_INIT})


ThadHouse:
The `inc` target actually already includes threads.

https://github.com/microsoft/msquic/blob/main/src/inc/CMakeLists.txt#L40

And does propagate correctly. Without this, the library doesn't work even in ubuntu. I wonder if for your platform the ordering matters. Can you try moving the inc and warning dependencies to the end of the list rather then the beginning? I worry that including the dependency twice will cause issues.

iriszero:
You meant
```
target_link_libraries(quicperf perflib msquic platform perfbin.clog inc warnings)
```
this way?

Still it gives the same message.

For the record, it does not stop the execution. Just the message.

ThadHouse:
Yeah. That's weird, as the target should exactly match what the variable you are setting does.

Can you print out the contents of ${CMAKE_THREAD_LIBS_INIT} on your system? I'm curious what it's actually setting.

iriszero:
Right below the modified part:

`-pthread`

iriszero:
Maybe we should look at the transitivity of `target_link_libraries`?

For example,
https://cmake.org/cmake/help/v3.0/prop_tgt/INTERFACE_LINK_LIBRARIES.html#prop_tgt:INTERFACE_LINK_LIBRARIES

iriszero:
Oh, my bad. I had a confusion. Thank you for your vigilance.

Required (or "missing") is `pthread_attr_setschedparam` in `quicperf`,
however, there is no such symbol in the pthread library supplied.

I was confused with `pthread_attr_setschedpolicy` and `pthread_setschedparam` from `libpthread.so`.  

My Google Search showed a documentation that the missing function exists in `<pthread.h>`: https://www.man7.org/linux/man-pages/man3/pthread_attr_setschedparam.3.html, which wasn't the case in either my Linux or custom kernel.

```
➜ ldconfig -p | grep pthread.so

	libpthread.so.0 (libc6,x32, OS ABI: Linux 3.4.0) => /libx32/libpthread.so.0
	libpthread.so.0 (libc6,x86-64, OS ABI: Linux 3.2.0) => /lib/x86_64-linux-gnu/libpthread.so.0
	libpthread.so.0 (libc6, OS ABI: Linux 3.2.0) => /lib/i386-linux-gnu/libpthread.so.0
	libpthread.so.0 (libc6, OS ABI: Linux 3.2.0) => /lib32/libpthread.so.0
	libpthread.so (libc6,x32, OS ABI: Linux 3.4.0) => /libx32/libpthread.so
	libpthread.so (libc6,x86-64, OS ABI: Linux 3.2.0) => /lib/x86_64-linux-gnu/libpthread.so
	libpthread.so (libc6, OS ABI: Linux 3.2.0) => /lib32/libpthread.so
```

```
/lib/x86_64-linux-gnu  
➜ nm -D libpthread.so.0 | grep sched
000000000000b1a0 T pthread_attr_getschedpolicy
000000000000b1b0 T pthread_attr_setschedpolicy
000000000000adb0 T pthread_getschedparam
000000000000af30 T pthread_setschedparam
000000000000b070 T pthread_setschedprio
                 U __sched_getparam
                 U __sched_get_priority_max
                 U __sched_get_priority_min
                 U __sched_getscheduler
                 U sched_setparam
                 U __sched_setscheduler
                 U sched_yield
```


Instead I found `pthread_attr_setschedparam` in `libmsquic`.  
How can I specify that `quicperf` depends on `libmsquic`, rather than `${CMAKE_THREAD_LIBS_INIT}`,  here?

ThadHouse:
libmsquic.so does not export any functions other then MsQuicOpen and MsQuicClose. But we also don't define `pthread_attr_setschedparam` in msquic. Instead, its just called at https://github.com/microsoft/msquic/blob/49d86b7a7f1f161b5b72a924d88686d0e5e01489/src/platform/platform_linux.c#L726 

I'm not sure where you're seeing its in libmsquic.so. Its a dependency of libmsquic.so, but shouldn't be contained in libmsquic.so.

iriszero:
Hmm, that's right.
Please ignore the above comment.

Perhaps, it's the Ubuntu version?

On my Ubuntu 18.04.5 LTS
```
jason-huh in msquic/build/bin/Debug at {...} on  main [⇣] using ☁️  default/ 
➜ nm -gD /lib/x86_64-linux-gnu/libpthread.so.0 | grep sched
00000000000092b0 T pthread_attr_getinheritsched
0000000000009210 T pthread_attr_getschedparam
0000000000009280 T pthread_attr_getschedpolicy
00000000000092c0 T pthread_attr_setinheritsched
0000000000009220 T pthread_attr_setschedparam
0000000000009290 T pthread_attr_setschedpolicy
0000000000008df0 T pthread_getschedparam
0000000000008f20 T pthread_setschedparam
0000000000009050 T pthread_setschedprio
                 U __sched_getparam
                 U __sched_get_priority_max
                 U __sched_get_priority_min
                 U __sched_getscheduler
                 U sched_setparam
                 U __sched_setscheduler
                 U sched_yield

jason-huh in msquic/build/bin/Debug at {...} on  main [⇣] using ☁️  default/ 
➜ ldconfig -p | grep pthread.so
	libpthread.so.0 (libc6,x86-64, OS ABI: Linux 3.2.0) => /lib/x86_64-linux-gnu/libpthread.so.0

jason-huh in msquic/build/bin/Debug at {...} on  main [⇣] using ☁️  default/ 
➜ nm -gD /lib/x86_64-linux-gnu/libpthread.so.0 | grep sched
00000000000092b0 T pthread_attr_getinheritsched
0000000000009210 T pthread_attr_getschedparam
0000000000009280 T pthread_attr_getschedpolicy
00000000000092c0 T pthread_attr_setinheritsched
0000000000009220 T pthread_attr_setschedparam
0000000000009290 T pthread_attr_setschedpolicy
0000000000008df0 T pthread_getschedparam
0000000000008f20 T pthread_setschedparam
0000000000009050 T pthread_setschedprio
                 U __sched_getparam
                 U __sched_get_priority_max
                 U __sched_get_priority_min
                 U __sched_getscheduler
                 U sched_setparam
                 U __sched_setscheduler
                 U sched_yield
```

On my Ubuntu 20.04.1 LTS (on which my custom kernel is also based)
```
➜ ldconfig -p | grep pthread.so
	libpthread.so.0 (libc6,x32, OS ABI: Linux 3.4.0) => /libx32/libpthread.so.0
	libpthread.so.0 (libc6,x86-64, OS ABI: Linux 3.2.0) => /lib/x86_64-linux-gnu/libpthread.so.0
	libpthread.so.0 (libc6, OS ABI: Linux 3.2.0) => /lib/i386-linux-gnu/libpthread.so.0
	libpthread.so.0 (libc6, OS ABI: Linux 3.2.0) => /lib32/libpthread.so.0
	libpthread.so (libc6,x32, OS ABI: Linux 3.4.0) => /libx32/libpthread.so
	libpthread.so (libc6,x86-64, OS ABI: Linux 3.2.0) => /lib/x86_64-linux-gnu/libpthread.so
	libpthread.so (libc6, OS ABI: Linux 3.2.0) => /lib32/libpthread.so

~ 
➜ nm -gD /lib/x86_64-linux-gnu/libpthread.so.0 | grep sched

000000000000b1a0 T pthread_attr_getschedpolicy
000000000000b1b0 T pthread_attr_setschedpolicy
000000000000adb0 T pthread_getschedparam
000000000000af30 T pthread_setschedparam
000000000000b070 T pthread_setschedprio
                 U __sched_getparam
                 U __sched_get_priority_max
                 U __sched_get_priority_min
                 U __sched_getscheduler
                 U sched_setparam
                 U __sched_setscheduler
                 U sched_yield
```

You see `pthread_attr_setschedparam` is missing in the 20.04. Any thoughts?

ThadHouse:
That's weird that pthread would remove that function, as we do use it. We'd have to look into the potential for an alternate function. But that definitely would be a much different fix then this PR. Currently we build with 18.04, so we do still have access to that function.

iriszero:
@ThadHouse Agreed.
Also, I remember I did have to do some extra stuff to build in 20.04. (which, unfortunately, is missing in my memory :D)

So, I'd conclude:

1. This issue should be renamed and stated differently; or, for clarity, we can open a new one, referencing this.
2. We should state in the build docs that only Ubuntu 18.04 is supported, or at least that there's a pending issue with 20.04 or `pthread` in it.

ThadHouse:
I'd probably prefer closing this and creating an issue so we can look into what way we want to fix that.

PR #1065, opened 2020-11-30
Decrease Pacing Interval and Disablement Threshold

nibanks:
In the low RTT test cases, "pacing" perf was bad because pacing was actually disabled. This PR reduces the interval from 2 ms to 1 ms and decreases the RTT threshold at which we disable pacing from 15 ms to the same 1 ms. Wan perf results were improved in the affected scenario.

For some reason, enabling pacing resulted in a lower min RTT getting measured in the CPU limited throughput tests, which then resulted in FC not growing as much / as fast. This then meant FC was limiting throughput. To fix this we changed to smoothed RTT instead of min RTT.

### DIFF
--- a/.azure/azure-pipelines.ci.yml
+++ b/.azure/azure-pipelines.ci.yml
@@ -143,6 +143,18 @@ stages:
       platform: windows
       arch: x64
       tls: openssl
+  - template: ./templates/build-config-user.yml
+    parameters:
+      image: windows-latest
+      platform: windows
+      arch: arm64
+      tls: openssl
+  - template: ./templates/build-config-user.yml
+    parameters:
+      image: windows-latest
+      platform: windows
+      arch: x86
+      tls: openssl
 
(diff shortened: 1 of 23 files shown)


nibanks:
Hmm... The PR run seems to indicate low queue lengths decreased perf because of this. I will need to look at the logs...

nibanks:
Looks like it might have just been random difference between runs. Looking at the logs, the low RTT issues are from pacing being disabled. I'm trying to reduce that threshold so pacing stays enabled to see how things go.

maolson-msft:
The fact that the FC window is limited by using MinRTT instead of a bloated RTT is by design. The FC window should not be grown on account of bufferbloat. IM me if you want to discuss.

maolson-msft (approved):

ThadHouse (approved):

ThadHouse (approved):

PR #1076, opened 2020-12-02
Fix ChaCha20 bug in mitls and update logging.

anrossi:


scripts/run_endpoint.sh
```diff
@@ -6,7 +6,7 @@
 if [ -n "$TESTCASE" ]; then
     case "$TESTCASE" in
     # TODO: add supported test cases here
-    "versionnegotiation"|"handshake"|"transfer"|"retry"|"multiconnect"|"keyupdate")
+    "versionnegotiation"|"handshake"|"transfer"|"retry"|"multiconnect"|"keyupdate"|"chacha20")
```
nibanks, line 9: For the time being, can we just claim "server" support instead?

nibanks: And just append ChaCha/Poly to the cipher suites?

src/platform/unittest/CryptTest.cpp
```diff
@@ -10,6 +10,7 @@
 #include "quic_tls.h"
 
 #include "msquichelper.h"
+#include <iomanip>
```
nibanks, line 13: necessary?

scripts/run_endpoint.sh
anrossi: I looked at the QNS code, and it expects the client to drive the chacha-poly negotiation, and it should only send a single cipher suite. If there's multiple, even if one of them is chacha-poly, it will fail the test.

src/platform/unittest/CryptTest.cpp
anrossi: Not anymore.

scripts/run_endpoint.sh
anrossi: Oh, I see what you're asking. I'll do that.

nibanks (approved):

PR #1081, opened 2020-12-03
Cherry Pick: Fix nullptr deref in test stream creation on connection failure. (#1080)

ThadHouse:
If the stream in FromConnectionHandle returns null, that was never checked, and a bugcheck would occur.

Also fixes a memory leak in FromConnectionHandle when an OOM occurs.

Cherry picked from main.


nibanks:
I don't think this is going to meet the bar. We can discuss more in our meeting today, but I'm already getting a lot of pushback for the current commits we've CPed.

ThadHouse:
Ok. This one seems good because it is a bug, and not a feature. Even in test code, we'll likely keep getting that bug occasionally.

PR #1087, opened 2020-12-07
TCP WAN Perf Support via netput

nibanks:
Adds `netput.exe` support to our WAN perf script.

Currently, netput is an unencrypted TCP app that doesn't quite produce an apples to apples comparison. My short term goal will be to add TLS encryption support to netput. My long term goal will be to produce a cross-platform TLS/TCP platform datapath abstraction layer to use for more accurate (secured) TCP performance comparisons.

### DIFF
--- a/scripts/emulated-performance.ps1
+++ b/scripts/emulated-performance.ps1
@@ -86,6 +86,9 @@ param (
     [Int32]$NumIterations = 1,
 
     [Parameter(Mandatory = $false)]
+    [Int32[]]$UseTcp = 0,
+
+    [Parameter(Mandatory = $false)]
     [ValidateSet("None", "Datapath.Light", "Datapath.Verbose", "Performance.Light", "Performance.Verbose")]
     [string]$LogProfile = "None"
 )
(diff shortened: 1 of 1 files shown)


scripts/emulated-performance.ps1
```diff
@@ -128,21 +131,50 @@ if ($IsWindows) {
     $QuicPerf = Join-Path $RootDir "/artifacts/bin/linux/$($Arch)_$($Config)_$($Tls)/quicperf"
 }
 
+# Path to the netput exectuable.
+$NetPut = "C:\Windows\System32\netput.exe"
```
ThadHouse, line 135: Shouldn't this be in PATH by default, so just `netput.exe` is accessible? Windows isn't required to be on the C drive, even though it usually is.

nibanks: It didn't work without it.

ThadHouse: Huh, thats weird. I wonder if System32 isn't in PATH on those systems. Oh well not a big deal.

ThadHouse (approved):

PR #1090, opened 2020-12-08
Use Max RTT for Pacing

nibanks:


nibanks:
Going to add more emulated network testing before we make any changes.

PR #1093, opened 2020-12-10
Support Password Protected Certificate Files

nibanks:
Adds a new certificate type for protected certificate files.


anrossi (approved):

ThadHouse:
The reason this isn't working is you have to also set a callback, and then assign the password from the user data in that callback.

https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_set_default_passwd_cb_userdata.html

nibanks:
> The reason this isn't working is you have to also set a callback, and then assign the password from the user data in that callback.
> 
> https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_set_default_passwd_cb_userdata.html

Not according to this example I found: https://www.ibm.com/support/knowledgecenter/SSB23S_1.1.0.15/gtps7/s5sple1.html

I think the problem was actually with the type of certificate that was trying to be used here. For the time being, I'll change this PR to draft.

ThadHouse:
I think thats specific to IBM's openssl build. That looks like a fork, likely designed to make that easier. Since the actual openssl docs state the callback is needed.

I think even with the windows cert store change, this is still useful.

ThadHouse:
Doing digging, the built in setup does support a default callback. So very likely a cert issue. I merged main into this PR, and added explicitly pkey support for windows. This will enable non exportable certificates to be used, as long as the user can access the private key and pass it in. I want to build some certs and test both of these scenarios.

Theres also a 3rd scenario we could add, although I'm not sure we want to. Support loading the pkey for windows from file, rather then binary blob, using openssls explicit functionality. Thats a little more secure then what is currently here, but does add a bit more to the if statements.

PR #1095, opened 2020-12-10
Update WAN Perf to use Length instead of Duration

nibanks:


.azure/azure-pipelines.ci.yml
```diff
@@ -253,7 +253,7 @@ stages:
       jobName: wanperf_30Mbps
       displayName: 30 Mbps
       iterations: 3
-      durationMs: 10000
+      length: 37500000
```
maolson-msft, line 256: How were the lengths here chosen?

nibanks: (Rate / 8) * "Previous Duration"

nibanks: I wasn't really trying to change functionality yet.

PR #1099, opened 2020-12-11
Official Version 1 Support

nibanks:
Adds support for the official version 1 of QUIC! Removes support for some of the old draft versions that aren't practically deployed to minimize support costs.


src/core/connection.c
```diff
@@ -2559,17 +2552,10 @@ QuicConnProcessPeerTransportParameters(
         }
 
         //
-        // Version draft-28 and later fully validate all exchanged connection IDs.
-        // Version draft-27 only validates in the Retry scenario.
+        // Fully validate all exchanged connection IDs.
         //
-        if (Connection->Stats.QuicVersion == QUIC_VERSION_DRAFT_27) {
-            if (!QuicConnValidateTransportParameterDraft27CIDs(Connection)) {
```
nibanks, line 2566: Remove this function too

src/core/binding.c
```diff
@@ -1432,12 +1432,8 @@ QuicBindingDeliverDatagrams(
         // Only Initial (version specific) packets are processed from here on.
         //
         switch (Packet->Invariant->LONG_HDR.Version) {
-        case QUIC_VERSION_DRAFT_27:
```
ThadHouse, line 1435: http didn't depend on any of these versions, right? We won't have to make changes in there by removing these?

nibanks: Correct. They have no QUIC version specific code.

ThadHouse:
This probably isn't a blocker, but all the Stub handshake tests are failing. Other then that lgtm.

ThadHouse (approved):

nibanks:
> This probably isn't a blocker, but all the Stub handshake tests are failing. Other then that lgtm.

I'm looking into them. I think there was always a bug, but the change in TLS code point is now triggering it. I'll try to fix before I merge. Thanks for the review!

ThadHouse (approved):

src/core/crypto_tls.c
```diff
@@ -387,7 +386,21 @@ QuicCryptoTlsReadExtensions(
                 return Status;
             }
 
-        } else if (ExtType == TlsExt_QuicTransportParameters) {
+        } else if (
+            Connection->Stats.QuicVersion != QUIC_VERSION_DRAFT_29 &&
```
anrossi, line 390: we're sure we'll never get into the code path except when the version is 29 or 1?

anrossi (approved):

PR #1101, opened 2020-12-11
Wait on Separate Client Deleted Event

### LINKED ISSUE #1100: Race in QuicTestClientDisconnect on cleanup
In QuicTestClientDisconnect, at the end of the test we call WaitForShutdownComplete() on the client

https://github.com/microsoft/msquic/blob/main/src/test/lib/DataTest.cpp#L694

In the connection itself, `AutoDelete` is set to true, which causes the shutdown callback to delete the connection.

https://github.com/microsoft/msquic/blob/main/src/test/lib/TestConnection.cpp#L772

This happens just a few lines after `QuicEventSet(EventShutdownComplete)` is called. The deletion deletes the event, while its still possible the other thread has not started back up, and is still being waited on. If this happens, the event deletion fails, which triggers an assertion that the event did not uninitialize properly.

This seems to only be an issue on Linux, as Windows does seem to behave if an event is deleted while being waited on, as likely this is a kernel counted object.

https://dev.azure.com/ms/msquic/_build/results?buildId=127592&view=results

nibanks:
Fixes #1100. Add a new event to wait for the client to be deleted to avoid race condition.

### DIFF
--- a/src/test/lib/DataTest.cpp
+++ b/src/test/lib/DataTest.cpp
@@ -615,6 +615,8 @@ QuicTestClientDisconnect(
     PingStats ClientStats(UINT64_MAX - 1, 1, 1, TRUE, TRUE, FALSE, FALSE, TRUE,
         StopListenerFirst ? QUIC_STATUS_CONNECTION_TIMEOUT : QUIC_STATUS_ABORTED);
 
+    EventScope EventClientDeleted(true);
+
     MsQuicRegistration Registration;
     TEST_TRUE(Registration.IsValid());
 
@@ -640,7 +642,6 @@ QuicTestClientDisconnect(
         TEST_QUIC_SUCCEEDED(Listener.GetLocalAddr(ServerLocalAddr));
 
         TestConnection* Client;
-        EventScope EventClientDeleted(true);
         {
             UniquePtr<TestConnection> Server;
             ServerAcceptContext ServerAcceptCtx((TestConnection**)&Server);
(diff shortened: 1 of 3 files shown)


nibanks:
Well, I guess I was wrong. It did cause some test failures by removing it. We weren't waiting to clean up the listener until after the client shutdown now, which changed the behavior.

nibanks:
Looks like the tests pass on the latest iteration. @thadhouse or @anrossi any comments?

ThadHouse:
This has a race in the other direction, if a timeout happens. Thats less likely, but still there as a possibility.  Potentially not worth worrying about, as a timeout is a test failure.

nibanks:
> This has a race in the other direction, if a timeout happens. Thats less likely, but still there as a possibility. Potentially not worth worrying about, as a timeout is a test failure.

How? If there's a timeout, it will fail, then the registration will end up waiting (perhaps forever if there's a deadlock).

ThadHouse:
The event will get unwound before the registration wait. So the connection can end up finishing between when the event gets deleted and the registration shutdown happens. Which would cause set to happen on a deleted event.

nibanks:
> The event will get unwound before the registration wait. So the connection can end up finishing between when the event gets deleted and the registration shutdown happens. Which would cause set to happen on a deleted event.

Ah! Good call. I'll move the event scope up, above registration.

ThadHouse (approved):

PR #1104, opened 2020-12-15
Move perf tracking to branch on repo

ThadHouse:
Currently, all performance data gets uploaded to a database, and then displayed using PowerBi. The PowerBi dashboard is problematic, as is  adding anything new to the database. This setup will make storing arbitrary data easier, especially across branches.

This PR adds a new post perf step, that will grab all the perf results from each run, and merge them together. It does not currently publish to the new perf branch, but it will check that the configuration for doing that is correct.

### DIFF
--- a/.azure/templates/post-process-performance.yml
+++ b/.azure/templates/post-process-performance.yml
@@ -25,10 +25,20 @@ jobs:
       downloadPath: msquic/artifacts/PerfDataResults
 
   - task: Powershell@2
-    displayName: Merge coverage
+    displayName: Merge coverage (Release)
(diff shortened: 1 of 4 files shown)


.azure/templates/post-process-performance.yml
```diff
@@ -0,0 +1,46 @@
...
+    displayName: Download Perf Artifacts
+    inputs:
+      artifactName: performance
+      downloadPath: artifacts/PerfDataResults
+
+  - task: Powershell@2
+    displayName: Merge coverage
+    inputs:
+      pwsh: true
+      filePath: merge-performance.ps1
```
nibanks, line 31: This is actually in the `msquic` directory, not the local one I think.

nibanks (approved):

PR #1113, opened 2020-12-18
Fixes TLS 0-RTT Freeing Issues

nibanks:
Fixes a couple issues related to freeing the resumption ticket buffer:

- The wrong Tag was being used in free.
- miTLS wasn't freeing the buffer at all.


ThadHouse (approved):

ThadHouse:
It looks like we might have a mismatch in allocations depending on the tls provider.

ThadHouse:
Were you seeing this as a bug? MiTLS stores the resumption ticket as SerializedTicket, which we do free. We do have a bug in stub and schannel, but actually in the other direction.

ThadHouse:
I updated this PR to revert the mitls changes. They were getting freed correctly, just in a roundabout way that looked confusing. The tests were failing with a double free.

ThadHouse (approved):

ThadHouse (approved):

PR #1123, opened 2020-12-24
Weekly Release Build

### LINKED ISSUE #1024: Enable auto triggering weekly or monthly builds for release branches
Release branches can often go months or even years without commits. Because AZP is often changing, we should run builds occasionally so we can detect early if an AZP change is going to break the build.

nibanks:
Closed #1024. Schedules a weekly run (Mon morning, midnight UTC) for release branches to make sure they still build.

### DIFF
--- a/.azure/azure-pipelines.ci.yml
+++ b/.azure/azure-pipelines.ci.yml
@@ -27,6 +27,14 @@ pr:
     - docs/*
     - README.md
 
+# Periodically rebuild release branches to make sure they are functional.
+schedules:
+- cron: "0 0 * * Mon"
+  displayName: Weekly Release Build
+  branches:
+    include:
+    - release/*
+
 name: 0.$(Date:yyyy).$(Date:MM).$(DayOfMonth).$(Rev:rr).0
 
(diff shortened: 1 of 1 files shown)


ThadHouse:
Do you know if there's a way for force an email on specific task failures? That'd be something super helpful to know if the build stages fail. If azp doesn't support it, we can do something manually.

ThadHouse:
Also not directly related, what do we want to do about non serviceable releases? We probably don't want those building weekly if we're not going to update them.

nibanks:
> Do you know if there's a way for force an email on specific task failures? That'd be something super helpful to know if the build stages fail. If azp doesn't support it, we can do something manually.

I already get emails for all main runs, success or fail. I haven't spent any time trying to configure that yet. Something to look into...

> Also not directly related, what do we want to do about non serviceable releases? We probably don't want those building weekly if we're not going to update them.

I was thinking we might have a different identifier we key off of in the branch name for those. Not sure yet.

ThadHouse (approved):

Issue #1141, opened 2021-01-06
Java Support


nibanks:
### Describe the feature you'd like supported

Related to #952, as a first step to support Android we need to support the Java run time.

nibanks:
FYI, here's a repo with a good start we can work from: https://github.com/wkgcass/msquic-java (thanks @wkgcass!).

NewLunarFire:
As part of my work, I have developped a JNI for windows (unfortunately wkgcass' works only on Linux). I cannot directly contribute the code I have made for my JNI (I'm working on it) but maybe I can help shape the Java integration from my experience with the library. Also, further down the line, I would prefer using an "official" quic java library rather than maintain my own JNI.

nibanks:
@newlunarfire we'd love any help you'd like to contribute!

PR #1142, opened 2021-01-06
Platform Custom Thread Context

nibanks:
First step towards #1141.


ThadHouse:
We would need a winuser implementation of this as well. I'm ok without winkernel, as that will never be used, but winuser will definitely be used.

src/inc/quic_platform_linux.h
```diff
@@ -633,6 +633,28 @@ typedef struct QUIC_THREAD_CONFIG {
     void* Context;
 } QUIC_THREAD_CONFIG;
 
+#ifdef QUIC_USE_CUSTOM_THREAD_CONTEXT
+
+//
+// Extension point that allows additional platform specific logic to be executed
+// for every thread created. The platform must define QUIC_USE_CUSTOM_THREAD_CONTEXT
+// and implement the QuicThreadCustomStart function. QuicThreadCustomStart MUST
+// call the Callback passed in.
```
ThadHouse, line 642: This needs to also document that the implementation must free `CustomContext` after exiting. Since they have access to platform, its probably a good idea to use QUIC_ALLOC and QUIC_FREE instead of malloc and free. We can make a tag, and document what the free call needs to be.

src/inc/quic_platform_linux.h
```diff
@@ -633,6 +633,29 @@ typedef struct QUIC_THREAD_CONFIG {
...
 } QUIC_THREAD_CONFIG;
 
+#ifdef QUIC_USE_CUSTOM_THREAD_CONTEXT
+
+//
+// Extension point that allows additional platform specific logic to be executed
+// for every thread created. The platform must define QUIC_USE_CUSTOM_THREAD_CONTEXT
+// and implement the QuicThreadCustomStart function. QuicThreadCustomStart MUST
+// call the Callback passed in. QuicThreadCustomStart MUST also free
+// CustomContext before returning.
```
ThadHouse, line 643: Should probably also say must be freed with `QUIC_FREE(CustomContext, QUIC_POOL_CUSTOM_THREAD);` that way you don't have to dig into the impl to figure that out.

src/inc/quic_platform_linux.h
```diff
@@ -633,6 +633,29 @@ typedef struct QUIC_THREAD_CONFIG {
...
+//
+
+typedef struct QUIC_THREAD_CUSTOM_CONTEXT {
+    LPTHREAD_START_ROUTINE Callback;
+    void* Context;
+} QUIC_THREAD_CUSTOM_CONTEXT;
+
+void*
```
ThadHouse, line 651: I feel like we should have a `typedef QUIC_CUSTOM_THREAD_RETURN_TYPE void*` here and `typedef QUIC_CUSTOM_THREAD_RETURN_TYPE DWORD` in windows. That way the implementing code can just say `QUIC_CUSTOM_THREAD_RETURN_TYPE QuicThreadCustomStart` for their implementation, rather then needing to ifdef. Since in most cases, the other side of this will not be operating system specific.

ThadHouse (approved):

PR #1143, opened 2021-01-06
Add support for running tests as a full matrix rather then a default and changing 1 variable.

ThadHouse:
Necessary for long run RPS tests.

Do not merge currently, just running a test script currently.


ThadHouse:
This is ready to go, nothing uses the matrix, but a full matrix RPS test setup will use it.

nibanks (approved):

PR #1146, opened 2021-01-07
Add full RPS matrix, and a pipeline to run it.

ThadHouse:
Pipeline is manual only currently, and test suite likely needs more, but adding more is easy once this is going.

### DIFF
--- a/.azure/azure-pipelines.periodic.yml
+++ b/.azure/azure-pipelines.periodic.yml
@@ -60,3 +60,4 @@ stages:
       kernelMode: -Kernel -TestsFile FullRunRPS.json
       extraName: 'Kernel'
       testTypes: 'Remote'
+      timeout: 300
--- a/.azure/azure-pipelines.periodic.yml
+++ b/.azure/azure-pipelines.periodic.yml
@@ -36,9 +36,9 @@ schedules:
       config: Release
       extraBuildArgs: -DisableTest -DisableTools
   - template: ./templates/build-config-winkernel.yml
-      parameters:
-        arch: x64
-        config: Release
+    parameters:
+      arch: x64
+      config: Release
 
 #
 # Performance Tests
--- a/.azure/azure-pipelines.periodic.yml
+++ b/.azure/azure-pipelines.periodic.yml
@@ -6,9 +6,6 @@
 trigger: none
 pr: none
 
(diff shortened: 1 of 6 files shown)


.azure/azure-pipelines.periodic.yml
```diff
@@ -0,0 +1,54 @@
...
+  - template: ./templates/build-config-winkernel.yml
+      parameters:
+        arch: x64
+        config: Release
+
+#
+# Performance Tests
+#
+
+- stage: performance
```
nibanks, line 39: Rename this to something RPS related instead?

scripts/FullRunRPS.json
```diff
@@ -0,0 +1,56 @@
...
+        {
+            "TestName": "RPS",
+            "Local": {
+                "Platform": "Windows",
+                "Tls": ["stub", "schannel", "openssl", "mitls"],
+                "Arch": ["x64", "x86", "arm", "arm64"],
+                "Exe": "quicperf",
+                "Arguments": "-test:RPS -target:$RemoteAddress"
+            },
+            "Variables": [
```
nibanks, line 17: Eventually I'll probably want to add more than just these, but this is a good start. We can see how long it takes to run and go from there.

.azure/azure-pipelines.periodic.yml
```diff
@@ -0,0 +1,54 @@
+#
+# Continuous Integration (CI)
+# This pipeline builds and runs MsQuic performance RPS long run tests.
+#
+
+trigger: none
+pr: none
```
nibanks, line 7: Going to need to add the CRON trigger too.

.azure/azure-pipelines.periodic.yml
```diff
@@ -0,0 +1,62 @@
...
+
+name: 0.$(Date:yyyy).$(Date:MM).$(DayOfMonth).$(Rev:rr).0
+stages:
+
+schedules:
+- cron: "0 0 * * Sun"
```
nibanks, line 13: This is the same day as the Release builds, right? Maybe we do this Saturday so as to not conflict?

nibanks (approved):

.azure/azure-pipelines.periodic.yml
ThadHouse: Nope. Release builds are Mon, so sunday evening. These will run saturday evening.

PR #1149, opened 2021-01-10
Initial TCP Support

nibanks:
Adds TCP support to the data path abstraction layer. This is so that we can add TCP support to our performance tool to compare with QUIC. Only supported on Windows user mode. Because it no longer abstracts just a UDP binding, renames the data path binding to socket.

### DIFF
--- a/src/core/library.c
+++ b/src/core/library.c
@@ -191,6 +191,11 @@ MsQuicLibraryInitialize(
     QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
     BOOLEAN PlatformInitialized = FALSE;
     uint32_t DefaultMaxPartitionCount = QUIC_MAX_PARTITION_COUNT;
+    const QUIC_DATAPATH_CALLBACKS DatapathCallbacks = {
+        NULL,
+        QuicBindingReceive,
+        QuicBindingUnreachable
+    };
 
     Status = QuicPlatformInitialize();
     if (QUIC_FAILED(Status)) {
@@ -283,8 +288,7 @@ MsQuicLibraryInitialize(
     Status =
         QuicDataPathInitialize(
             sizeof(QUIC_RECV_PACKET),
-            QuicBindingReceive,
-            QuicBindingUnreachable,
+            &DatapathCallbacks,
             &MsQuicLib.Datapath);
     if (QUIC_FAILED(Status)) {
         QuicTraceEvent(
--- a/src/core/library.c
+++ b/src/core/library.c
@@ -1493,7 +1493,7 @@ _IRQL_requires_max_(DISPATCH_LEVEL)
 QUIC_WORKER*
 QUIC_NO_SANITIZE("implicit-conversion")
 QuicLibraryGetWorker(
-    _In_ const _In_ QUIC_RECV_DATAGRAM* Datagram
+    _In_ const _In_ QUIC_RECV_DATA* Datagram
     )
 {
     QUIC_DBG_ASSERT(MsQuicLib.StatelessRegistration != NULL);
--- a/src/core/library.c
+++ b/src/core/library.c
@@ -191,8 +191,7 @@ MsQuicLibraryInitialize(
     QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
     BOOLEAN PlatformInitialized = FALSE;
     uint32_t DefaultMaxPartitionCount = QUIC_MAX_PARTITION_COUNT;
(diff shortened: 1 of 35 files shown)


src/inc/quic_datapath.h
```diff
@@ -306,63 +361,89 @@ QuicDataPathResolveAddress(
...
     _In_opt_ const QUIC_ADDR* LocalAddress,
     _In_opt_ const QUIC_ADDR* RemoteAddress,
-    _In_opt_ void* RecvCallbackContext,
-    _Out_ QUIC_DATAPATH_BINDING** Binding
+    _In_opt_ void* CallbackContext,
+    _Out_ QUIC_SOCKET** Socket
+    );
+
+//
+// Creates a TCP socket for the (optional) (optional) given local address and
```
ThadHouse, line 382: (optional) (optional) is a bit weird here. Also doesn't match the UDP docs.

nibanks: Hmm, I think a copy/paste bug. I'll fix it up.

src/core/library.c
```diff
@@ -191,6 +191,10 @@ MsQuicLibraryInitialize(
     QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
     BOOLEAN PlatformInitialized = FALSE;
     uint32_t DefaultMaxPartitionCount = QUIC_MAX_PARTITION_COUNT;
+    const QUIC_UDP_DATAPATH_CALLBACKS DatapathCallbacks = {
```
anrossi, line 194: Even if these are const, are they still stack allocated? shouldn't this be global?

nibanks: No need. The datapath copies the contents.

src/core/quicdef.h
```diff
@@ -234,7 +234,7 @@ QUIC_STATIC_ASSERT(IS_POWER_OF_TWO(QUIC_MAX_RANGE_DECODE_ACKS), L"Must be power
 // Path MTU discovery will always start with/initialize with the smallest
 // allowable MTU for QUIC (1280 bytes).
 //
-#define QUIC_DEFAULT_PATH_MTU                   QUIC_MIN_MTU
+#define QUIC_DEFAULT_PATH_MTU                   1280    // TODO - Use 1200 instead
```
ThadHouse, line 237: Is this TODO going to be fixed in this PR, or for later?

ThadHouse (approved):

src/platform/datapath_winuser.c
```diff
@@ -573,11 +747,27 @@ QuicDataPathInitialize(
...
+    }
+    if (TcpCallbacks != NULL) {
+        if (TcpCallbacks->Accept == NULL ||
+            TcpCallbacks->Connect == NULL ||
+            TcpCallbacks->Receive == NULL) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            Datapath = NULL;
+            goto Exit;
+        }
+    }
```
anrossi, line 770: enforce that specifying both Udp and Tcp callbacks should fail. A socket cannot be both TCP and UDP, right?

src/platform/datapath_winuser.c
```diff
@@ -606,14 +796,21 @@ QuicDataPathInitialize(
     }
 
     RtlZeroMemory(Datapath, DatapathLength);
-    Datapath->RecvHandler = RecvCallback;
-    Datapath->UnreachableHandler = UnreachableCallback;
+    if (UdpCallbacks) {
+        Datapath->UdpHandlers = *UdpCallbacks;
```
anrossi, line 800: This requires the upper layer (QUIC core) to keep this structure alive for the duration of the datapath layer, which kind of breaks the layering abstraction between them.  Shouldn't this copy the structure/pointers locally?

ThadHouse: The deref will copy the struct, keeping a local copy of the pointers.

anrossi: Ah, I (incorrectly) assumed it was a pointer-pointer. You're right.

anrossi (approved):

src/core/quicdef.h
nibanks: Later.

src/platform/datapath_winuser.c
nibanks: This is the datapath layer, not socket layer. You can have one datapath handle that you use to create both UDP and TCP sockets.

PR #1154, opened 2021-01-11
Rename quicperf driver to quicperfdrv, add private instance

ThadHouse:
Changing the name makes symbols not interfere between executable and driver, and we need a version that links into shipping msquic for internel tests

### DIFF
--- a/scripts/performance-helper.psm1
+++ b/scripts/performance-helper.psm1
@@ -342,7 +342,7 @@ function Invoke-RemoteExe {
     } -ArgumentList $Exe
 
     if ($Kernel) {
-        $RunArgs = "--kernel --privateLibrary $RunArgs"
+        $RunArgs = "--kernelPriv $RunArgs"
     }
 
     Write-Debug "Running Remote: $Exe $RunArgs"
--- a/scripts/performance-helper.psm1
+++ b/scripts/performance-helper.psm1
@@ -310,12 +310,12 @@ function Get-ExeName {
 function Remove-PerfServices {
     if ($IsWindows) {
         Invoke-TestCommand -Session $Session -ScriptBlock {
-            if ($null -ne (Get-Service -Name "quicperf" -ErrorAction Ignore)) {
+            if ($null -ne (Get-Service -Name "quicperfdrvpriv" -ErrorAction Ignore)) {
                 try {
-                    net.exe stop quicperf /y | Out-Null
+                    net.exe stop quicperfdrvpriv /y | Out-Null
                 }
                 catch {}
-                sc.exe delete quicperf /y | Out-Null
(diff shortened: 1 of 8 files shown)


scripts/performance-helper.psm1
```diff
@@ -342,7 +342,7 @@ function Invoke-RemoteExe {
     } -ArgumentList $Exe
 
     if ($Kernel) {
-        $RunArgs = "--kernel $RunArgs"
+        $RunArgs = "--kernel --privateLibrary $RunArgs"
```
nibanks, line 345: Another option is have `--kernel` and `--kernelPriv` options.

src/perf/bin/appmain.cpp
```diff
@@ -23,7 +23,8 @@
 #include "PerfIoctls.h"
 #include "quic_driver_helpers.h"
 
-#define QUIC_DRIVER_NAME   "quicperf"
+#define QUIC_DRIVER_NAME   "quicperfdrv"
+#define QUIC_DRIVER_NAME_PRIVATE    "quicperfdrvpriv"
```
nibanks, line 27:
```suggestion
#define QUIC_DRIVER_NAME                   "quicperfdrv"
#define QUIC_DRIVER_NAME_PRIVATE    "quicperfdrvpriv"
```

nibanks (approved):

scripts/performance-helper.psm1
ThadHouse: That likely is better. I would likely change the tests at the same time as well so they match. This was pulled straight from the tests.

nibanks: I leave it up to you to do now or later.

src/perf/bin/appmain.cpp
```diff
@@ -371,6 +371,9 @@ main(
         if (strcmp("--kernel", argv[i]) == 0) {
```
nibanks, line 371: Shouldn't you be checking for `--kernelPriv` too here?

src/test/bin/quic_gtest.cpp
```diff
@@ -985,9 +985,9 @@ int main(int argc, char** argv) {
     for (int i = 0; i < argc; ++i) {
         if (strcmp("--kernel", argv[i]) == 0) {
```
nibanks, line 986: Shouldn't you be checking for `--kernelPriv` here too?

ThadHouse: No need. `--kernel` strcmp will match `--kernelPriv`

nibanks: strcmp compares to the null terminator. They won't be equal.

PR #1160, opened 2021-01-14
TCP/TLS Performance Interface

nibanks:
Adds several helper classes to implement TCP/TLS support for performance comparisons.


src/perf/lib/Tcp.cpp
```diff
@@ -0,0 +1,724 @@
...
+    const QUIC_ADDR* RemoteAddress,
+    const QUIC_ADDR* LocalAddress,
+    void* Context) :
+    Server(false), Initialized(false), IndicateAccept(false), IndicateConnect(false),
+    IndicateDisconnect(false), StartTls(false),
+    Engine(Engine), Worker(nullptr), Socket(nullptr), SecConfig(nullptr), Tls(nullptr),
+    ReceiveData(nullptr), SendData(nullptr), BatchedSendData(nullptr),
+    BufferedDataLength(0), Context(Context)
+{
+    Entry.Flink = NULL;
```
ThadHouse, line 289: Why is this not using CxPlatListInitializeHead? It looks like CxPlatListInsertTail would assert in this case, because Blink would be garbage, and fail QuicListEntryValidate (Also I need to fix QuicListEntryValidate)

ThadHouse: If this doesn't need to be circular, why not use CXPLAT_SLIST_ENTRY?

ThadHouse: Wait nevermind I'm misunderstanding what this does. This is the entry, not the list.

src/perf/lib/Tcp.cpp
```diff
@@ -0,0 +1,724 @@
...
+            CxPlatCopyMemory(StreamFrame->Data, NextSendData->Buffer + Offset, StreamLength);
+            Offset += StreamLength;
+
+            if (!EncryptFrame(Frame)) {
+                FreeSendBuffer(SendBuffer);
+                goto Exit;
+            }
+
+            SendBuffer->Length = sizeof(TcpFrame) + Frame->Length + CXPLAT_ENCRYPTION_OVERHEAD;
+            FinalzieSendBuffer(SendBuffer);
```
ThadHouse, line 660: `Finalize`, not `Finalzie`

src/platform/datapath_winuser.c
```diff
@@ -2523,9 +2527,11 @@ CxPlatSocketStartAccept(
     return Status;
 }
 
+extern uint32_t* CxPlatProcessorGroupOffsets;
```
ThadHouse, line 2530: Exists in quic_platform_winuser.h, does not need to be defined here.

ThadHouse (approved):

PR #1169, opened 2021-01-18
Fix incremental issues in CMake build so rebuilds don't consistently happen

### LINKED ISSUE #1166: Fix incremental builds in CMake
Currently, cmake is handling a few files incorrectly, which is causing relinks and rebuilds when not fully necessary. I've found 2 so far, could be more. 

file(Write) for ETW rc file. https://github.com/microsoft/msquic/blob/main/CMakeLists.txt#L218 This causes that file to always be touched, which causes the resource compiler to have to be reran every build.

sourcelink using file write rather then file generation https://github.com/microsoft/msquic/blob/main/CMakeLists.txt#L218
This causes every link to need to be reran, since the sourcelink file will have been touched.

ThadHouse:
Issues with source link and the resource manifest generation were causing many relinks to occur when not necessary.

Closes #1166


CMakeLists.txt
```diff
@@ -215,19 +215,20 @@ endif()
...
     add_custom_command(
         OUTPUT ${QUIC_BUILD_DIR}/inc/MsQuicEtw.h
         OUTPUT ${QUIC_BUILD_DIR}/inc/MsQuicEtw.rc
         DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/src/manifest/MsQuicEtw.man
         COMMAND mc.exe -um -h ${QUIC_BUILD_DIR}/inc -r ${QUIC_BUILD_DIR}/inc ${CMAKE_CURRENT_SOURCE_DIR}/src/manifest/MsQuicEtw.man)
-    add_custom_target(MsQuicEtw
-        DEPENDS ${QUIC_BUILD_DIR}/inc/MsQuicEtw.h
-        DEPENDS ${QUIC_BUILD_DIR}/inc/MsQuicEtw.rc)
+    add_custom_target(MsQuicEtw_HeaderBuild
+            DEPENDS ${QUIC_BUILD_DIR}/inc/MsQuicEtw.h)
```
nibanks, line 225:
```suggestion
    add_custom_target(MsQuicEtw_HeaderBuild
        DEPENDS ${QUIC_BUILD_DIR}/inc/MsQuicEtw.h)
```

nibanks (approved):
1 minor nit, but otherwise looks ok. I don't understand some of it, but since the CI passed and I assume you tested it out, I approve.

nibanks (approved):

PR #1173, opened 2021-01-19
Version Negotiation Extension

anrossi:


src/inc/msquic.h
```diff
@@ -396,13 +396,16 @@ typedef struct QUIC_SETTINGS {
             uint64_t MigrationEnabled           : 1;
             uint64_t DatagramReceiveEnabled     : 1;
             uint64_t ServerResumptionLevel      : 1;
-            uint64_t RESERVED                   : 38;
+            uint64_t DesiredVersionsList     : 1;
```
anrossi, line 399: fix this

src/inc/msquic.h
```diff
@@ -396,13 +396,16 @@ typedef struct QUIC_SETTINGS {
...
+            uint64_t RESERVED                   : 37;
         } IsSet;
     };
 
     uint64_t MaxBytesPerKey;
     uint64_t HandshakeIdleTimeoutMs;
     uint64_t IdleTimeoutMs;
+    uint32_t* DesiredVersionsList;
```
anrossi, line 407: Move to bottom.

src/core/settings.h
```diff
@@ -4,14 +4,84 @@
...
+    uint16_t LoadBalancingMode;             // Global only
+    uint8_t MaxOperationsPerDrain;
+    uint8_t SendBufferingEnabled    : 1;
+    uint8_t PacingEnabled           : 1;
+    uint8_t MigrationEnabled        : 1;
+    uint8_t DatagramReceiveEnabled  : 1;
+    uint8_t ServerResumptionLevel   : 2;    // QUIC_SERVER_RESUMPTION_LEVEL
+    uint8_t RESERVED                : 2;
+    uint32_t* GeneratedCompatibleVersionsList;
+    uint32_t GeneratedCompatibleVersionsListLength;
```
anrossi, line 74: Do we have a convention on Length as in the number of entries, vs. Length as the number of bytes?

src/core/core.kernel.vcxproj
```diff
@@ -102,6 +103,7 @@
     <ClInclude Include="stream_set.h" />
     <ClInclude Include="timer_wheel.h" />
     <ClInclude Include="transport_params.h" />
+    <ClCompile Include="version_neg.h" />
```
nibanks, line 106:
```suggestion
    <ClInclude Include="version_neg.h" />
```

src/core/unittest/TransportParamTest.cpp
```diff
@@ -34,6 +34,12 @@ void CompareTransportParams(
     COMPARE_TP_FIELD(IDLE_TIMEOUT, IdleTimeout);
     COMPARE_TP_FIELD(MAX_ACK_DELAY, MaxAckDelay);
     COMPARE_TP_FIELD(ACTIVE_CONNECTION_ID_LIMIT, ActiveConnectionIdLimit);
+    if (A->Flags & QUIC_TP_FLAG_VERSION_NEGOTIATION) {
+        ASSERT_EQ(A->VersionNegotiationInfoLength, B->VersionNegotiationInfoLength);
+        ASSERT_EQ(
+            memcmp(A->VersionNegotiationInfo, B->VersionNegotiationInfo, A->VersionNegotiationInfoLength),
```
nibanks, line 40:
```suggestion
            memcmp(A->VersionNegotiationInfo, B->VersionNegotiationInfo, (size_t)A->VersionNegotiationInfoLength),
```

src/core/crypto_tls.c
```diff
@@ -1629,6 +1649,26 @@ QuicCryptoTlsDecodeTransportParameters(
...
+                QuicTraceEvent(
+                    ConnErrorStatus,
+                    "[conn][%p] ERROR, %u, %s.",
+                    Connection,
+                    Length,
+                    "Invalid length of QUIC_TP_ID_VERSION_NEGOTIATION_EXT");
+                goto Exit;
+            }
+            TransportParams->Flags |= QUIC_TP_FLAG_VERSION_NEGOTIATION;
+            TransportParams->VersionNegotiationInfo = (uint8_t*)TPBuf + Offset; // Review: Allocate and copy?
```
nibanks, line 1663: Yes, it needs to be copied.

src/core/settings.h
nibanks: Generally I think we do number of entries.

src/core/version_neg.c
```diff
@@ -0,0 +1,384 @@
...
+#endif
+
+typedef struct QUIC_COMPATIBLE_VERSION_MAP {
+    const uint32_t OriginalVersion;
+    const uint32_t CompatibleVersion;
+} QUIC_COMPATIBLE_VERSION_MAP;
+
+const QUIC_COMPATIBLE_VERSION_MAP CompatibleVersionsMap[] = {
+    {QUIC_VERSION_DRAFT_29, QUIC_VERSION_DRAFT_29},
+    {QUIC_VERSION_MS_1, QUIC_VERSION_MS_1},
```
nibanks, line 24: We should just know versions are always compatible with themselves. No need to increase the size of this array for that.

src/core/version_neg.c
```diff
@@ -0,0 +1,384 @@
...
+    QUIC_VERSION_MS_1,
+    QUIC_VERSION_DRAFT_29
+};
+const uint32_t DefaultSupportedVersionsListLength = ARRAYSIZE(DefaultSupportedVersionsList);
+
+const uint32_t DefaultCompatibleVersionsList[] = {
+    QUIC_VERSION_1,
+    QUIC_VERSION_MS_1
+};
+const uint32_t DefaultCompatibleVersionsListLength = ARRAYSIZE(DefaultCompatibleVersionsList);
```
nibanks, line 44: I feel like we should always calculate this, so we don't have to depend on the developer correctly updating it as the default supported list and compatibility list change over time.

src/core/version_neg.c
```diff
@@ -0,0 +1,384 @@
...
+                AllocFailure,
+                "Allocation of '%s' failed. (%llu bytes)",
+                "Server Version Negotiation Info",
+                VNILen);
+            return NULL;
+        }
+        VNIBuf = VersionNegotiationInfo;
+
+        CxPlatCopyMemory(VNIBuf, &Connection->Stats.QuicVersion, sizeof(Connection->Stats.QuicVersion)); // TODO: test that this is getting set
+        VNIBuf += 4;
```
nibanks, line 304: Use sizeof(...) instead of these constants.

src/core/version_neg.h
```diff
@@ -0,0 +1,100 @@
...
+    This file contains definitions for version negotiation.
+
+--*/
+
+#pragma once
+
+//
+// This list is the versions that the server advertises support for.
+//
+extern const uint32_t DefaultSupportedVersionsList[];
```
nibanks, line 17: It's a bit more painful to update, but I prefer to put the size here as a constant, and then we don't need the length field.

nibanks (commented):
Some comments from a quick review.

src/core/settings.h
anrossi: Ok, that's the pattern I've been following here.

src/core/binding.c
```diff
@@ -816,12 +825,10 @@ QuicBindingProcessStatelessOperation(
         CxPlatRandom(sizeof(uint8_t), &RandomValue);
         VerNeg->Unused = 0x7F & RandomValue;
 
-        memcpy(Buffer, &Binding->RandomReservedVersion, sizeof(uint32_t));
```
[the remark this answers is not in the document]
anrossi: Why? It makes it harder for the server to validate the client's packet. Also, Where should it go? First or last? In the case the app sets a list of supported versions, where should it go?

src/core/configuration.c
```diff
@@ -421,7 +424,7 @@ QuicConfigurationParamGet(
         }
 
         *BufferLength = sizeof(QUIC_SETTINGS);
-        CxPlatCopyMemory(Buffer, &Configuration->Settings, sizeof(QUIC_SETTINGS));
+        CxPlatCopyMemory(Buffer, &Configuration->Settings, sizeof(QUIC_SETTINGS)); // TODO: how to copy back out DesiredVersionsList
```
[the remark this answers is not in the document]
anrossi: This is a discussion point: Is it acceptable to just pass out the pointer to the internal list?  Should we allocate memory for it? If so, how does the app free that memory?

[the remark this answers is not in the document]
nibanks: If the list is constant forever, you can return it. If it might go away, then we can't.

src/core/crypto_tls.c
anrossi: The transport parameter doesn't outlive the packet does it? We process it and then we're done?

src/core/binding.c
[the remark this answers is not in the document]
nibanks: Because it greases the feature (VN). You can put it where ever makes it easier. You can just always put it first and know the validation logic and just "make sure the first is a reserved value". We wouldn't care which one exactly.

src/core/version_neg.c
anrossi: It makes the implementation of the loops cleaner by having the array like this.

src/core/crypto_tls.c
nibanks: It lives for the lifetime of the connection I believe.

src/core/version_neg.c
nibanks: Perhaps, but I'd really prefer not having this constant. It shouldn't be _that_ hard to fix the loops to accept themselves as compatibile.

src/core/version_neg.c
anrossi: Should it live on the MsQuicLib object for the default list? The benefit here is that it saves memory and computation time for the vast majority of apps. (and it's read-only memory)

nibanks: Feel free to save it where ever you think best. Calculating it once for default apps at library start up is ~free.

src/core/configuration.c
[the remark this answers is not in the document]
anrossi: The lifetime is the same as the lifetime of the Configuration object.

pravb:
@anrossi @nibanks what's holding this PR up?

src/core/connection.c
```diff
@@ -3801,7 +4027,14 @@ QuicConnRecvFrames(
             } else if (Status == QUIC_STATUS_OUT_OF_MEMORY) {
                 return FALSE;
             } else {
-                if (Status != QUIC_STATUS_INVALID_STATE) {
+                if (Status == QUIC_STATUS_VER_NEG_ERROR) {
+                    // TODO: how to keep the datagram around long enough for the queued operation?
```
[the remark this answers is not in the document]
anrossi: We need to discuss solutions to this, since it's messy to plumb this specific case all the way to `QuicConnRecvDatagrams`
Maybe a flag on a given datagram to indicate it shouldn't be freed, and promising that it'll be returned later?

anrossi:
@pravb I'm still cleaning up feedback. But the PR probably can't be merged right away because the Version Negotiation Extension is still subject to incompatible changes in the future, and those could break older MsQuic clients/servers using the current implementation.

src/test/lib/HandshakeTest.cpp
```diff
@@ -885,24 +891,452 @@ QuicTestVersionNegotiation(
...
+    ClientSettings.SetIdleTimeoutMs(3000);
+
+    MsQuicSettings ServerSettings;
+    ServerSettings.SetDesiredVersionsList(ServerVersions, ServerVersionsLength);
+    ServerSettings.SetIdleTimeoutMs(3000);
+
+    MsQuicSettings ClearVNSettings;
+    ClearVNSettings.SetDesiredVersionsList(nullptr, 0);
+
+    TEST_QUIC_SUCCEEDED(
```
anrossi, line 933: The server settings being global makes an unavoidable test problem: the server settings will persist across test cases if one of the version negotiation tests that set the server version fails.

ThadHouse: We'll likely need to create an RAII wrapper for any tests that modify the global VN data, so on error or passing it will get reset back to stock.

anrossi: gtest has a nifty thing where I could put all the VN tests into their own test class witch a cleanup function run on the end of every test case, regardless of pass/fail. The problem exists primarily for kernel mode tests, where the driver doesn't unload.

ThadHouse: Yeah we can do that too. Either that or just any time we set it in a test we do it through a wrapper struct. Both would end up working about the same.

anrossi: But how does the RAII wrapper work for kernel? The test doesn't throw exceptions in kernel; all the macros just set a global to false and return immediately.

ThadHouse: When it returns, the stack will still unwind just like a normal function return. So it will call all destructors for stack allocated variables. Nothing special would need to happen, since that happens on returns anyway.

src/core/connection.c
```diff
@@ -2487,7 +2515,182 @@ QuicConnProcessPeerTransportParameters(
         Connection->SourceCidLimit = QUIC_TP_ACTIVE_CONNECTION_ID_LIMIT_DEFAULT;
     }
 
+    QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
     if (!FromResumptionTicket) {
+        if (Connection->Settings.VersionNegotiationExtEnabled) {
```
nibanks, line 2520: The following logic would probably be best put into its own function.

src/core/connection.c
```diff
@@ -2882,6 +3111,7 @@ QuicConnRecvRetry(
         return;
     }
 
+    // Should this use the app-set list of versions?
```
nibanks, line 3114: Dunno? Should we? Either way, remove this comment.

src/core/connection.c
```diff
@@ -3840,7 +4082,20 @@ QuicConnRecvFrames(
...
+                            QUIC_OPER_TYPE_VERSION_NEGOTIATION,
+                            CxPlatDataPathRecvPacketToRecvData(Packet))) {
+                        //
+                        // Packets that elicit a VN response don't get counted as received
+                        // if they're deferred. Update that count here to ensure it gets counted.
+                        //
+                        Connection->Stats.Recv.TotalPackets += 2;
+                        Packet->ReleaseDeferred = TRUE;
+                    }
+                    QuicConnTransportError(Connection, QUIC_ERROR_VERSION_NEGOTIATION_ERROR);
```
nibanks, line 4097: Does this send a connection close in response? You need to make sure it doesn't.

src/core/connection.c
```diff
@@ -3840,7 +4082,20 @@ QuicConnRecvFrames(
...
+                    if (QuicBindingQueueStatelessOperation(
+                            Connection->Paths[0].Binding,
+                            QUIC_OPER_TYPE_VERSION_NEGOTIATION,
+                            CxPlatDataPathRecvPacketToRecvData(Packet))) {
+                        //
+                        // Packets that elicit a VN response don't get counted as received
+                        // if they're deferred. Update that count here to ensure it gets counted.
+                        //
+                        Connection->Stats.Recv.TotalPackets += 2;
+                        Packet->ReleaseDeferred = TRUE;
```
nibanks, line 4095: We don't need to do this. For one, we don't even know if the packet was completely valid...

src/core/connection.c
```diff
@@ -5781,7 +6005,7 @@ QuicConnParamGet(
         }
 
         *BufferLength = sizeof(QUIC_SETTINGS);
-        *(QUIC_SETTINGS*)Buffer = Connection->Settings;
+        CxPlatCopyMemory(Buffer, &Connection->Settings, *BufferLength); // TODO: How to copy out DesiredVersionsList?
```
nibanks, line 6008: Fix this. At least remove comment.

src/core/connection.c
```diff
@@ -6102,9 +6341,9 @@ QuicConnApplyNewSettings(
     }
 
     if (OverWrite) {
-        QuicSettingsDumpNew(NewSettingsSize, NewSettings);
+        QuicSettingsDumpNew(NewSettingsSize, (QUIC_SETTINGS_INTERNAL*)NewSettings); // Review
     } else {
-        QuicSettingsDump(&Connection->Settings); // TODO - Really necessary?
+        QuicSettingsDump((QUIC_SETTINGS*)&Connection->Settings); // TODO - Really necessary?
```
nibanks, line 6346: Finish this stuff up and remove comments.

src/core/crypto_tls.c
```diff
@@ -1074,6 +1081,19 @@ QuicCryptoTlsEncodeTransportParameters(
...
             "TP: Disable 1-RTT Encryption");
     }
+    if (TransportParams->Flags & QUIC_TP_FLAG_VERSION_NEGOTIATION) {
+        TPBuf =
+            TlsWriteTransportParam(
+                QUIC_TP_ID_VERSION_NEGOTIATION_EXT,
+                (uint16_t)TransportParams->VersionNegotiationInfoLength,
+                TransportParams->VersionNegotiationInfo,
+                TPBuf);
+        QuicTraceLogConnVerbose(
```
nibanks, line 1091: I think we should add more encode/decode tracing for the full TP details.

src/core/library.c
```diff
@@ -673,13 +705,14 @@ QuicLibrarySetGlobalParam(
...
                 TRUE,
+                TRUE,
                 BufferLength,
                 (QUIC_SETTINGS*)Buffer)) {
             Status = QUIC_STATUS_INVALID_PARAMETER;
             break;
         }
 
-        QuicSettingsDumpNew(BufferLength, (QUIC_SETTINGS*)Buffer);
+        QuicSettingsDumpNew(BufferLength, (QUIC_SETTINGS_INTERNAL*)Buffer); // Review
```
nibanks, line 715: Remove/fix comment.

src/core/settings.h
```diff
@@ -4,24 +4,97 @@
     Licensed under the MIT License.
 
 --*/
+typedef struct QUIC_SETTINGS_INTERNAL {
```
nibanks, line 7: How are we going to deal with the problem of this getting out of sync with the public type? I don't like the idea of a copy + more... Need to figure out a better way.

nibanks (commented):
Some initial comments. Review still incomplete.

src/core/settings.h
anrossi: You just have to manually keep them in sync? I have a static assert to keep them in sync.

src/core/connection.c
anrossi: How about informing the app they need to provide a buffer large enough to store the `DesiredVersionsList` as well as the `QUIC_SETTINGS`?

src/core/settings.h
anrossi: We could have an internal type that just contains the public type as an anonymous member and adds additional fields?
But then the internal fields would have to have public flags. Or a separate `InternalFlags` field. 
What do you think of this?
```c
typedef struct QUIC_SETTINGS_INTERNAL {
    QUIC_SETTINGS; // Not all compilers may support this
    union {
        uint64_t InternalSetFlags;
        struct {
            uint64_t GeneratedCompatabilitylist :1;
        } InternalSet;
    }
    uint32_t* GeneratedCompatabilityList;
    uint32_t GeneratedCompatabililtyListLength;
} QUIC_SETTINGS_INTERNAL;
```

src/core/crypto_tls.c
anrossi: I need to sync up with @ThadHouse on how to add a new decoder to CLOG, because it would be nice to log the version lists, at least.

src/core/connection.c
anrossi: Ok, how about we just increment by 1 so when the `TotalPackets` counter is decremented (because an error was returned), it doesn't underflow?

src/core/connection.c
anrossi: It doesn't appear to. But it eventually calls into `QuicStreamRecvShutdown` which queues up a stop sending frame. Since the server is sending a version negotiation packet (to do incompatible version negotiation), it should close the connection silently and immediately, correct?

src/core/crypto_tls.c
anrossi: Added some more tracing in the VNI TP decoding logic in version_neg.c.

src/core/settings.h
nibanks: Yeah, I was also considering this design. Thoughts?

anrossi: Struct packing will be worse than the current copy&add, since the struct will have 8-byte alignment, and as fields are added (like the new fields I just added) it will have some padding between it and the internal-only fields. We can work to minimize that as much as possible, but the possibility that some padding is present is unavoidable.

anrossi: C++ doesn't allow anonymous embedded struct. (So the unit tests and the attack tool don't build because they include the internal headers directly.)

ThadHouse: Is there a reason it needs to be anonymous? Thats the only part c++ doesn't like iirc.

nibanks: There are ways to make C++ happy (if C++, inherit; if C, anonymous).

src/core/connection.c
```diff
@@ -2520,26 +2520,16 @@ QuicConnProcessPeerVersionNegotiationTP(
...
             }
         }
-        if (CurrentVersionIndex == SupportedVersionsLength) {
-            //
-            // Current version not supported, start incompatible version negotiation.
-            //
-            CXPLAT_DBG_ASSERTMSG(FALSE, "Should be caught in binding layer");
-            return QUIC_STATUS_VER_NEG_ERROR;
-        }
+        CXPLAT_FRE_ASSERTMSG(CurrentVersionIndex != SupportedVersionsLength, "Incompatible Version Negotation should happen in binding layer");
```
nibanks, line 2523: Does this need to be a FRE assert? In other words, are you confident to say this won't happen, and if it does, Windows should bugcheck?

src/tools/etw/quicetw.h
```diff
@@ -336,6 +336,11 @@ typedef enum QUIC_EVENT_ID_CONNECTION {
     EventId_QuicConnQueueSendFlush,
     EventId_QuicConnOutFlowStreamStats,
     EventId_QuicConnPacketStats,
+    EventId_QuicConnServerResumeTicket,
+    EventId_QuicConnClientCompatibleVersionList,
+    EventId_QuicConnClientReceivedVersionList,
+    EventId_QuicConnServerSupportedVersionList,
+
```
nibanks, line 343: nit: extra/unnecessary newline

src/core/binding.c
```diff
@@ -770,13 +770,23 @@ QuicBindingProcessStatelessOperation(
...
+            SupportedVersionsLength = ARRAYSIZE(DefaultSupportedVersionsList);
+        }
+
         const uint16_t PacketLength =
             sizeof(QUIC_VERSION_NEGOTIATION_PACKET) +               // Header
             RecvPacket->SourceCidLen +
             sizeof(uint8_t) +
             RecvPacket->DestCidLen +
             sizeof(uint32_t) +                                      // One random version
-            ARRAYSIZE(QuicSupportedVersionList) * sizeof(uint32_t); // Our actual supported versions
```
nibanks, line 779: Worth keeping this comment?

src/core/settings.h
anrossi: I'm just going to remove this.

src/manifest/MsQuicEtw.man
```diff
@@ -2599,6 +2614,33 @@
...
+              />
+          <event
+              keywords="ut:Connection"
+              level="win:Informational"
+              message="$(string.Etw.ConnServerSupportedVersionList)"
+              opcode="Connection"
+              symbol="QuicConnServerSupportedVersionList"
+              template="tid_CONN_VNL"
+              value="5178"
+              />
```
nibanks, line 2643: These can probable have the `ut:LowVolume` keywords as well, as they shouldn't be particular common.

anrossi: There'll be 2 per connection, at most (server side and client side). Is that low volume?

nibanks: Yeah.

nibanks (approved):

src/core/connection.c
```diff
@@ -2465,8 +2494,178 @@ QuicConnValidateTransportParameterCIDs(
...
+        for (; CurrentVersionIndex < SupportedVersionsLength; ++CurrentVersionIndex) {
+            if (Connection->Stats.QuicVersion == SupportedVersions[CurrentVersionIndex]) {
+                break;
+            }
+        }
+        if (CurrentVersionIndex == SupportedVersionsLength) {
+            //
+            // Current version not supported, start incompatible version negotiation.
+            //
+            CXPLAT_DBG_ASSERTMSG(FALSE,"Incompatible Version Negotation should happen in binding layer");
```
ThadHouse, line 2527: Space after comma. Also, the comment here seems confusing mixed with the assert. Nothing should be "started" after a possible assert IMO.

src/core/connection.c
```diff
@@ -2488,6 +2687,14 @@ QuicConnProcessPeerTransportParameters(
     }
 
     if (!FromResumptionTicket) {
+        if (Connection->Settings.VersionNegotiationExtEnabled &&
+            Connection->PeerTransportParams.Flags & QUIC_TP_FLAG_VERSION_NEGOTIATION) {
+            QUIC_STATUS Status = QuicConnProcessPeerVersionNegotiationTP(Connection);
+            if (QUIC_FAILED(Status)) {
```
ThadHouse, line 2693: We have an error case label here. Is this not expected to be an error case, but actually a special early return case, I'd like a comment here and potentially at the end of the function incase we ever do an allocation or something in here.

src/core/connection.c
```diff
@@ -5532,7 +5762,7 @@ QuicConnParamSet(
         }
 
         QuicConnOnQuicVersionSet(Connection);
-        QuicConnProcessPeerTransportParameters(Connection, TRUE);
+        (void)QuicConnProcessPeerTransportParameters(Connection, TRUE);
```
ThadHouse, line 5765: Do we really not care about the outcome of this function?

src/core/crypto.c
```diff
@@ -343,7 +343,7 @@ QuicCryptoInitializeTls(
 
     Crypto->ResumptionTicket = NULL; // Owned by TLS now.
     Crypto->ResumptionTicketLength = 0;
-    QuicCryptoProcessData(Crypto, !IsServer);
+    (void)QuicCryptoProcessData(Crypto, !IsServer);
```
ThadHouse, line 346: Same ignoring return value question here. If this is valid, maybe a comment explaining why it's getting ignored. Usually I see something like this as a code smell.

src/core/settings.h
```diff
@@ -5,6 +5,7 @@
 
 --*/
 
+
```
ThadHouse, line 8: Nit. Extra new line?

src/core/connection.c
anrossi: I'll move the comment closer to the return value. The assert is to help developers not make any mistakes, and the if-check is to make sure production systems aren't brought down by hitting this path.

src/core/settings.h
anrossi: 😱

src/core/connection.c
anrossi: you think it would be better to add a `goto Error` and put an `Error` label at the bottom?

ThadHouse: Theres already an error label at the bottom of this function. This early return is going to bypass that.

src/core/connection.c
anrossi: Yeah, in the case where `FromResumption` is `TRUE`, none of the failure paths execute. I made this into a debug assert just in case to help future developers not break this path.

ThadHouse (approved):

PR #1175, opened 2021-01-20
Make pool allocators on linux actually use pool

ThadHouse:
Previously, just malloc was being used. Switches the allocation to both be inlined, and to use a pool allocator.

On Windows, calling CxPlatPoolFree after CxPlatPoolUninitalize will result in a memory leak (assuming the pool still has memory), however in Linux it will crash. However, both of these are bugs, so I don't believe the difference will matter.

### DIFF
--- a/src/platform/platform_linux.c
+++ b/src/platform/platform_linux.c
@@ -200,67 +200,6 @@ CxPlatFree(
 }
 
 void
-CxPlatPoolInitialize(
-    _In_ BOOLEAN IsPaged,
-    _In_ uint32_t Size,
-    _In_ uint32_t Tag,
-    _Inout_ CXPLAT_POOL* Pool
-    )
-{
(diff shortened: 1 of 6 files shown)


src/platform/platform_linux.c
```diff
@@ -199,67 +199,6 @@ CxPlatFree(
...
-void
-CxPlatPoolInitialize(
-    _In_ BOOLEAN IsPaged,
-    _In_ uint32_t Size,
-    _In_ uint32_t Tag,
-    _Inout_ CXPLAT_POOL* Pool
-    )
-{
-#ifdef CX_PLATFORM_DISPATCH_TABLE
-    PlatDispatch->PoolInitialize(IsPaged, Size, Pool);
```
nibanks, line 211: Looks like we don't use these Pool* dispatch table functions anymore. Can you remove them too please?

nibanks (approved):

PR #1176, opened 2021-01-20
Switch linux to using windows values for address family

### LINKED ISSUE #404: Use platform independent enum for address family

ThadHouse:
Will make high level wrappers much easier to implement. Closes #404 

Note that this is an ABI breaking change, so cannot be done as part of a 1.x release.

### DIFF
--- a/src/inc/msquic_linux.h
+++ b/src/inc/msquic_linux.h
@@ -153,9 +153,9 @@ typedef sa_family_t QUIC_ADDRESS_FAMILY;
 //
 // Defines match windows values.
 //
-#define QUIC_ADDRESS_FAMILY_UNSPEC AF_UNSPEC
-#define QUIC_ADDRESS_FAMILY_INET AF_INET
-#define QUIC_ADDRESS_FAMILY_INET6 AF_INET6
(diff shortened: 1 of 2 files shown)


nibanks:
> Note that this is an ABI breaking change, so cannot be done as part of a 1.x release.

I have been thinking about this, and I don't really want to rev to 2.0 because of this. I think we can get away with **not** doing that since we only claimed "preview" support for Linux in 1.0. We can confirm with the .NET folks to see if they're Ok with making this change an not rev'ing the major version.

@jkotalik, @Tratcher or @scalablecory any comments here?

Tratcher:
I don't think we've shipped anything that we care about breaking, it's better to get these issues sorted out now. We control the dependency anyways, right @jkotalik, so we can fix it when we update?

nibanks (approved):

PR #1182, opened 2021-01-22
TCP Perf: Support Abort Stream

nibanks:
Adds the ability to abort a stream to the TCP logic. This is required for "timed` performance testing, which is what wanperf uses. Also fixes a few other TCP bugs found while debugging WAN perf.

### DIFF
--- a/.azure/azure-pipelines.wanperf.yml
+++ b/.azure/azure-pipelines.wanperf.yml
@@ -52,7 +52,7 @@ parameters:
 - name: extraArgs
   type: string
   displayName: Extra Args
-  default: ''
+  default: 'none'
 
 stages:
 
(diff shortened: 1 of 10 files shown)


.azure/azure-pipelines.wanperf.yml
```diff
@@ -92,3 +96,5 @@ stages:
       rttMs: ${{ parameters.rttMs }}
       pacing: ${{ parameters.pacing }}
       logProfile: ${{ parameters.logProfile }}
+      ${{ if ne(parameters.extraArgs, 'none') }}:
```
ThadHouse, line 99: Any specific reason for defaulting to `none` and checking it its not that? All the other extra args we use just leave it empty and pass the empty value thorough.

ThadHouse (approved):

.azure/azure-pipelines.wanperf.yml
nibanks: AZP doesn't allow empty input fields. You're required to put a value in it, so I use `none` as that value by default.

PR #1185, opened 2021-01-24
Custom Commit Filtering

nibanks:
Updates the JS code to support a custom commit count filtering for the commit data generated by the CI pipeline. This theoretically allows us to update the CI pipeline to generate data files with all (or probably capped to a larger number of commits) data and then have our JS filter (possibly dynamically controlled by the user) to a smaller number of commits to actually display.

Practically, this change introduces two variables (`summaryChartCommits` and `detailChartCommits`) that control the current number of commits to actually display in the respective charts.


nibanks:
Superseded by new dashboard work.

PR #1187, opened 2021-01-25
Add CI build that runs static code checkers

### LINKED ISSUE #1106: Add a build with QUIC_CODE_CHECK to CI
Now that the `QUIC_CODE_CHECK` flag for `cmake` does something, someone who knows how the Azure CI stuff works could add a run that enables this to the CI setup.

larseggert:
Fixes #1106 (when ready)

### DIFF
--- a/.azure/azure-pipelines.ci.yml
+++ b/.azure/azure-pipelines.ci.yml
@@ -232,6 +232,7 @@ stages:
       tls: openssl
       extraBuildArgs: -CodeCheck
       config: Debug
+      extraName: 'codecheck'
 
 - stage: build_linux_nontest
   displayName: Build Linux - Non Tested
--- a/.azure/azure-pipelines.ci.yml
+++ b/.azure/azure-pipelines.ci.yml
@@ -224,15 +224,6 @@ stages:
       tls: stub
       extraBuildArgs: -SanitizeAddress
       config: Debug
-  - template: ./templates/build-config-user.yml
-    parameters:
-      image: ubuntu-latest
-      platform: linux
-      arch: x64
-      tls: openssl
-      extraBuildArgs: -CodeCheck
-      config: Debug
-      extraName: 'codecheck'
 
 - stage: build_linux_nontest
   displayName: Build Linux - Non Tested
@@ -265,6 +256,15 @@ stages:
       tls: openssl
       ubuntuVersion: 18.04
       extraBuildArgs: -DisableLogs -Generator Ninja -ToolchainFile cmake/toolchains/arm-pi-gnueabihf.toolchain.cmake
+  - template: ./templates/build-config-user.yml
+    parameters:
+      image: ubuntu-latest
+      platform: linux
+      arch: x64
+      tls: openssl
+      extraBuildArgs: -CodeCheck
+      config: Debug
+      extraName: 'codecheck'
 
 #
 # Performance Tests
--- a/.azure/azure-pipelines.ci.yml
+++ b/.azure/azure-pipelines.ci.yml
@@ -224,6 +224,14 @@ stages:
       tls: stub
       extraBuildArgs: -SanitizeAddress
       config: Debug
+  - template: ./templates/build-config-user.yml
+    parameters:
+      image: ubuntu-latest
+      platform: linux
+      arch: x64
+      tls: openssl
+      extraBuildArgs: -CodeCheck
+      config: Debug
 
 - stage: build_linux_nontest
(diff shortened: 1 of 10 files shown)


larseggert:
Can anyone tell if a new CI build built with the `QUIC_CODE_CHECK=on` option? (I doesn't look like it to me.)

.azure/azure-pipelines.ci.yml
```diff
@@ -224,6 +224,14 @@ stages:
...
       extraBuildArgs: -SanitizeAddress
       config: Debug
+  - template: ./templates/build-config-user.yml
+    parameters:
+      image: ubuntu-latest
+      platform: linux
+      arch: x64
+      tls: openssl
+      extraBuildArgs: -CodeCheck
+      config: Debug
```
nibanks, line 234: You need to use the `extraName` parameter to make the job name unique, otherwise AZP complains.
```suggestion
      config: Debug
      extraName: 'codecheck'
```

.azure/azure-pipelines.ci.yml
```diff
@@ -224,6 +224,15 @@ stages:
       tls: stub
       extraBuildArgs: -SanitizeAddress
       config: Debug
+  - template: ./templates/build-config-user.yml
```
nibanks, line 227: This should be moved down into the "Build Linux - Non Tested" stage so that the rest of the pipeline (perf & tests) is not delayed by this job.

nibanks:
Looks like the new build passed. @larseggert if you could just move the job to the other stage as I mentioned, we can go ahead and merge this.

larseggert:
@nibanks I moved it.

I think I am actually not done here yet though. For this build, I need to install `clang-tidy` and/or `cppcheck`. in `./templates/build-config-user.yml`, it seems that `prepare-machine.ps1` is called with `-Configuration Build` - should I add the corresponding packages to `apt-get` in `prepare-machine.ps1`? Or is there a better approach?

nibanks:
> @nibanks I moved it.
> 
> I think I am actually not done here yet though. For this build, I need to install `clang-tidy` and/or `cppcheck`. Since `./templates/build-config-user.yml` is used and it seems that calls `prepare-machine.ps1` with `-Configuration Build` - should I add the corresponding packages to `apt-get` in `prepare-machine.ps1`? Or is there a better approach?

So long as the overhead is not large (a few seconds) then that would be fine. Otherwise, we're going to need someway to special case these dependencies.

larseggert:
Let me add it and we'll see what the impact is.

larseggert:
OK, that did it. The build is now failing with a `bugprone-branch-clone` message, which was committed after I submitted #975.

Do you want me to fix/suppress any such issues in this PR?

nibanks:
> OK, that did it. The build is now failing with a `bugprone-branch-clone` message, which was committed after I submitted #975.
> 
> Do you want me to fix/suppress any such issues in this PR?

Yes, please fix any issues. We don't want to take this until it's passing.

src/inc/quic_datapath.h
```diff
@@ -205,7 +205,7 @@ CxPlatDataPathRecvPacketToRecvData(
 //
 CXPLAT_RECV_PACKET*
 CxPlatDataPathRecvDataToRecvPacket(
-    _In_ const CXPLAT_RECV_DATA* const RecvData
```
nibanks, line 208: For this file, the names are all correct. If the Linux datapath file is out of sync (and this clang is complaining) then that is where the fixes should be.

src/platform/selfsign_openssl.c
```diff
@@ -17,14 +17,14 @@
...
-#include "openssl/ssl.h"
 #include "openssl/err.h"
 #include "openssl/kdf.h"
 #include "openssl/pem.h"
+#include "openssl/pem.h"
 #include "openssl/rsa.h"
 #include "openssl/ssl.h"
+#include "openssl/ssl.h"
 #include "openssl/x509.h"
-#include "openssl/pem.h"
```
nibanks, line 27: Looks like these were actually just duplicates and can be removed.

src/inc/quic_datapath.h
nibanks: We recently did a lot of refactoring and I only ensured the header was correct and didn't have a chance to go fix the implementations yet. Feel free to just fix the absolute minimum to make this happy.

larseggert:
@nibanks the latest version has your desired changes. I will point out though that the other datapath implementations are (still) using parameter names that are different from those in `quic_datapath.h`.

PR #1190, opened 2021-01-26
Custom Certificate Validation

### LINKED ISSUE #1063: Remote certificate validation
In .NET we have a callback that allows checking the remote certificate during the TLS handshake. I'd like to see a connection event that we can plug into to do this.

```
bool RemoteCertificateValidationCallback(object sender, X509Certificate? certificate, X509Chain? chain, SslPolicyErrors sslPolicyErrors)
```

If user returns true, accept the cert. If false, reject. Beyond allowing users finer control over validation, we use this often during testing with a blanket `return true` to allow self-signed certs that aren't trusted.

@bartonjs what (CAPI, OpenSSL) structures would we need MsQuic to pass us in order to light this up?

I don't feel that this is P0 for us, but would be very useful. CC @samsp-msft

### LINKED ISSUE #1150: OpenSSL Client does not validate certificate chain
OpenSSL never checks the certificate chain, even without the `QUIC_CREDENTIAL_FLAG_NO_CERTIFICATE_VALIDATION` flag not set.   This needs to be something implemented before we can consider openssl to be shippable.

nibanks:
Includes mostly just the beginnings of the proposed API changes.

First part of #1063. Also fixes #1150.


src/inc/msquic.h
```diff
@@ -817,6 +820,9 @@ typedef struct QUIC_CONNECTION_EVENT {
             uint32_t ResumptionTicketLength;
             const uint8_t* ResumptionTicket;
         } RESUMPTION_TICKET_RECEIVED;
+        struct {
+            void* PeerCertificate; // TODO - Type
```
nibanks, line 824: @scalablecory, @bartonjs, @anrossi, @ThadHouse I need some help figuring out exactly how we should indicate the "peer's certificate" up to the app here. I can think of a couple possible options:

1. Byte buffer, wire-format certificate chain.
2. Platform independent structure with functions to abstract working with the platform-specific certificate object.
3. A void* to an actual platform specific object, therefore requiring the app to already "know" the type.

1 seems the cleanest approach in my opinion, but might require the most work on the app. Thoughts?

nibanks: An additional point of complexity: With OpenSSL, we get a callback from OpenSSL with the certificate context/pointer and as far as I can tell, it only lives for the lifetime of the callback. So, if we want to support the app being able to return PENDING from this (which I think we **really** do) we can't really return OpenSSL's object without somehow copying it or the data it represents.

bartonjs: Assuming that all that this layer wants to do is say "I got a certificate here, do what you want to do with it?", pbData/cbData seems the best to me.

nibanks: Just to be clear @bartonjs you're vote is for (1)?

bartonjs: > Just to be clear @bartonjs you're vote is for (1)?

Yep.  The wire format is almost certainly the "raw"/DER-encoded X.509 public key certificate.  If there's some protocol-specific structure around it I'd skip that part and make the X.509 DER data pointer be what gets shared out.

anrossi:
I prefer (1) as well. Yes, there's a little more memory usage and complexity, but any app  that wants to work with a certificate will be able to handle the plain DER/ASN.1 certificate bytes.

@Nibanks, you mention the whole chain, however. Do you want to split out the leaf certificate and present the chain separately? Or just one whole blob, in the TLS format (3byte-length,certificate data,...)?

src/inc/msquic.h
anrossi: I prefer (1) as well. Yes, there's a little more memory usage and complexity, but any app  that wants to work with a certificate will be able to handle the plain DER/ASN.1 certificate bytes.

@nibanks, you mention the whole chain, however. Do you want to split out the leaf certificate and present the chain separately? Or just one whole blob, in the TLS format (3byte-length,certificate data,...)?

nibanks: I think I prefer (1) as well, except in the case of non-cross-platform clients (http.sys). If Schannel gives us back a `PCERT_CONTEXT` and then we have to convert it to the raw encoding, just for HTTP to convert it back, that seems a bit silly.

And to your question @anrossi I'd return the whole chain in one blob like TLS.

src/platform/tls_openssl.c
```diff
@@ -32,6 +32,8 @@
 
 uint16_t CxPlatTlsTPHeaderSize = 0;
 
+const size_t OpenSslFilePrefixLength = sizeof("..\\..\\..\\..\\..\\..\\submodules");
```
ThadHouse, line 35: This isn't actually going to need to be a constant in the code, right? If so, I'd MUCH prefer to find another way, maybe to inject it.

nibanks: If you have a better solution, please let me know. This works and gives all the necessary info now:
```
 92 [0]40A4.4CF8::2021/01/29-09:54:48.523204500 [Microsoft-Quic][conn][0x1ED4AFF300] TLS handshake error: error:0A000086:SSL routines::certificate verify failed, file:openssl\ssl\statem\statem_clnt.c:1869
```

ThadHouse: I'd at least like to a check to make sure we're not going out of bounds if that length changes somehow.

src/platform/tls_mitls.c
```diff
@@ -1548,6 +1548,20 @@ CxPlatTlsOnCertVerify(
...
+                TlsContext->Connection)) {
+            QuicTraceEvent(
+                TlsError,
+                "[ tls][%p] ERROR, %s.",
+                TlsContext->Connection,
+                "Custom certificate validation failed");
+        } else {
+            Result = TRUE;
+        }
+        goto Error;
```
ThadHouse, line 1562: Since this is now hit in both error and non error cases, we likely should rename this to `Exit`

ThadHouse (approved):
Looks good, other then my 2 comments.

src/platform/tls_mitls.c
nibanks: I think it's fine. Most of the places go there on error, so I'd just prefer to keep it as is.

PR #1192, opened 2021-01-26
Fix race in pending send data resulting in a corrupted socket context

### LINKED ISSUE #1191: SIGSEGV on Ubuntu/CentOS7
### Describe the bug

While transferring large number of buffers, occasionally SIGSEGV happens.

1. Build tools in this repo
2. Run interopserver at one server with `-upload` option: `./quicinteropserver -listen:* -retry:1 -file:server.cert -key:server.key -root $(pwd) -upload:"$(pwd)/upload"`
3. Run `post` tool on another VM: `./quicpost -server:<server> -ip:<ip> -port:4433 -file:data.txt`
4. Make sure that `data.txt` is large (I have 1.2Gb file).. it will force `post` to send ~thousands of buffers. Sometimes everything works fine, however pretty frequently it crashes the client.

File transfer is complete.

Updated bug with info from Debug build:

Full stack:
```
[Switching to Thread 0x7ffff560d700 (LWP 29788)]
0x00007ffff6646387 in raise () from /lib64/libc.so.6
Missing separate debuginfos, use: debuginfo-install libatomic-4.8.5-44.el7.x86_64 libgcc-4.8.5-44.el7.x86_64 libstdc++-4.8.5-39.el7.x86_64
(gdb) bt
#0  0x00007ffff6646387 in raise () from /lib64/libc.so.6
#1  0x00007ffff6647a78 in abort () from /lib64/libc.so.6
#2  0x00007ffff78ab561 in quic_bugcheck () at /git/msquic/src/platform/platform_linux.c:59
#3  0x00007ffff78aa48a in CxPlatSocketSend (Binding=0x7fffec001e10, LocalAddress=0x0,
    RemoteAddress=0x7fffe8190290, SendContext=0x7fffe8190270)
    at /git/msquic/src/platform/datapath_linux.c:2097
#4  0x00007ffff78aa563 in CxPlatSocketContextSendComplete (
    SocketContext=SocketContext@entry=0x7fffec002050, ProcContext=ProcContext@entry=0x605c38)
    at /git/msquic/src/platform/datapath_linux.c:1449
#5  0x00007ffff78aa71e in CxPlatSocketContextProcessEvents (EventPtr=<optimized out>,
    ProcContext=ProcContext@entry=0x605c38, Events=4)
    at /git/msquic/src/platform/datapath_linux.c:1555
#6  0x00007ffff78aa7f8 in CxPlatDataPathWorkerThread (Context=0x605c38)
    at /git/msquic/src/platform/datapath_linux.c:2218
#7  0x00007ffff69e5ea5 in start_thread () from /lib64/libpthread.so.0
#8  0x00007ffff670e96d in clone () from /lib64/libc.so.6
```

ThadHouse:
Previously, there was no synchronization between adding items to the pending send list, and removing them. This was causing memory corruption when pending sends would occur. While reviewing this, an issue where pending sends would result in sends occurring out of order. Both the synchronization and out of order sending issue have been fixed by this PR. The code is also easier to understand, in addition to reducing memory usage per send.

Closes #1191


src/platform/datapath_linux.c
```diff
@@ -1455,11 +1463,13 @@ CxPlatSocketContextSendComplete(
         if (QUIC_FAILED(Status)) {
             goto Exit;
```
logachev, line 1464: If this happens, lock remains in acquired state

src/platform/datapath_linux.c
```diff
@@ -182,6 +182,11 @@ typedef struct CXPLAT_SOCKET_CONTEXT {
     //
     CXPLAT_LIST_ENTRY PendingSendContextHead;
 
+    //
+    // Lock around the PendingSendContext list.
+    //
+    CXPLAT_LOCK PendingSendContextLock;
```
nibanks, line 188: Rename this to something like `SendLock` now?

src/platform/datapath_linux.c
```diff
@@ -1364,51 +1381,53 @@ CxPlatSocketContextPendSend(
...
+
+Exit:
     return QUIC_STATUS_SUCCESS;
 }
 
+//
+// N.B. Caller must be holding SocketContext->PendingSendContextLock Lock.
+//
 QUIC_STATUS
 CxPlatSocketContextSendComplete(
```
nibanks, line 1424: Can we get a design where this doesn't need the lock to be held, and instead it only grabs where necessary? I don't like this whole thing requiring the lock; and especially don't like where it selectively unlocks as necessary.

ThadHouse: There is going to be no way to avoid the unlock and lock in the middle of this. We need to do the epoll part under the lock, in case it races with the worker thread. I can look a bit more into making the caller not hold the lock though.

nibanks: At the very least, I don't want the caller to grab the lock. There is no need. Let this function handle it. On the topic of epoll, I don't see why the lock needs to be held for that. Generally, you should never be required to hold any lock and call into some other layer/component.

src/platform/datapath_linux.c
ThadHouse: Now that this is only around pending sends and nothing else, likely best to keep it this way.

src/platform/datapath_linux.c
ThadHouse: Done.

src/platform/datapath_linux.c
```diff
@@ -2114,22 +2084,64 @@ CxPlatSocketSend(
...
+                    "[data][%p] ERROR, %u, %s.",
+                    SocketContext->Binding,
+                    errno,
+                    "epoll_ctl failed");
+                Status = errno;
+                goto Exit;
+            }
+            if (!IsPendedSend) {
+                CxPlatLockAcquire(&SocketContext->PendingSendContextLock);
                 CxPlatSocketContextPendSend(
```
nibanks, line 2136: You must pend before doing epoll, otherwise the epoll thread might wake before you get a chance.

logachev:
Another crash callstack I see:

```
#0  QuicBindingPreprocessDatagram (Binding=Binding@entry=0x7fffaaaaaaaa,
    Datagram=Datagram@entry=0x7fffe404bda8,
    ReleaseDatagram=ReleaseDatagram@entry=0x7ffff4e0ad5f "\001")
    at /git/msquic/src/core/binding.c:1081
#1  0x00007ffff7884452 in QuicBindingReceive (Socket=<optimized out>,
    RecvCallbackContext=0x7fffaaaaaaaa, DatagramChain=0x0) at /git/msquic/src/core/binding.c:1559
#2  0x00007ffff78a8317 in CxPlatSocketContextRecvComplete (
    SocketContext=SocketContext@entry=0x7fffec0021b8, ProcContext=ProcContext@entry=0x605d18,
    BytesTransferred=41) at /git/msquic/src/platform/datapath_linux.c:1323
#3  0x00007ffff78a991c in CxPlatSocketContextProcessEvents (EventPtr=<optimized out>,
    ProcContext=ProcContext@entry=0x605d18, Events=1)
    at /git/msquic/src/platform/datapath_linux.c:1520
#4  0x00007ffff78a9ab0 in CxPlatDataPathWorkerThread (Context=0x605d18)
    at /git/msquic/src/platform/datapath_linux.c:2259
#5  0x00007ffff69e4ea5 in start_thread () from /lib64/libpthread.so.0
#6  0x00007ffff670d96d in clone () from /lib64/libc.so.6
```

src/platform/datapath_linux.c
```diff
@@ -2114,22 +2084,64 @@ CxPlatSocketSend(
...
+                QuicTraceEvent(
+                    DatapathErrorStatus,
+                    "[data][%p] ERROR, %u, %s.",
+                    SocketContext->Binding,
+                    errno,
+                    "epoll_ctl failed");
+                Status = errno;
+                goto Exit;
+            }
             SendPending = TRUE;
```
nibanks, line 2143: Need to set this above, even in the case of failure, so we don't free the send context on failure here, after we put it in the list already.

ThadHouse: Fixed.

src/platform/datapath_linux.c
```diff
@@ -1409,60 +1374,71 @@ CxPlatSocketContextSendComplete(
...
-            break;
+        if (!CxPlatListIsEmpty(&SocketContext->PendingSendContextHead)) {
+            SendContext =
+                CXPLAT_CONTAINING_RECORD(
+                    SocketContext->PendingSendContextHead.Flink,
+                    CXPLAT_SEND_DATA,
+                    PendingSendLinkage);
+        } else {
+            SendContext = NULL;
         }
```
nibanks, line 1437: This should just go inside the previous `if` statement. If `CxPlatSocketSendInternal` returns pending, we should just be bailing out anyways.

src/platform/datapath_linux.c
```diff
@@ -1409,60 +1374,71 @@ CxPlatSocketContextSendComplete(
...
+            &SockFdEpEvt);
+    if (Ret != 0) {
+        Status = Ret;
+        QuicTraceEvent(
+            DatapathErrorStatus,
+            "[data][%p] ERROR, %u, %s.",
+            SocketContext->Binding,
+            Status,
+            "epoll_ctl failed");
+        goto Exit;
```
nibanks, line 1400: I know it might have been doing this way before, but since nothing actually happens at `Exit` we should just `return`. Same everywhere else below.

src/platform/datapath_linux.c
```diff
@@ -1409,60 +1374,71 @@ CxPlatSocketContextSendComplete(
...
-                CxPlatListRemoveHead(&SocketContext->PendingSendContextHead),
-                CXPLAT_SEND_DATA,
-                PendingSendLinkage);
+    CxPlatLockAcquire(&SocketContext->PendingSendContextLock);
+    if (!CxPlatListIsEmpty(&SocketContext->PendingSendContextHead)) {
+        SendContext =
+        CXPLAT_CONTAINING_RECORD(
+            SocketContext->PendingSendContextHead.Flink,
+            CXPLAT_SEND_DATA,
+            PendingSendLinkage);
```
nibanks, line 1409: nit: indent

nibanks (commented):
Almost there!

src/platform/datapath_linux.c
```diff
@@ -1409,59 +1374,69 @@ CxPlatSocketContextSendComplete(
...
+        CxPlatLockAcquire(&SocketContext->PendingSendContextLock);
+        if (Status != QUIC_STATUS_PENDING) {
+            CxPlatListRemoveHead(&SocketContext->PendingSendContextHead);
+            CxPlatSendDataFree(SendContext);
+            if (!CxPlatListIsEmpty(&SocketContext->PendingSendContextHead)) {
+            SendContext =
+                CXPLAT_CONTAINING_RECORD(
+                    SocketContext->PendingSendContextHead.Flink,
+                    CXPLAT_SEND_DATA,
+                    PendingSendLinkage);
```
nibanks, line 1433: nit: indent :)

nibanks (approved):
Beautiful! Ship it!

src/platform/datapath_linux.c
```diff
@@ -150,11 +145,6 @@ typedef struct CXPLAT_SOCKET_CONTEXT {
 #define QUIC_SOCK_EVENT_SOCKET  1
```
Anas198900, line 150: Hi

Anas198900, line 150: 150

src/platform/datapath_linux.c
Anas198900: No

src/platform/datapath_linux.c
Anas198900: Ok

PR #1195, opened 2021-01-27
Deferred Certificate Validation Support

nibanks:
Exposes the Schannel TLS feature to do "deferred certificate validation" and exposes those results to the app.

### DIFF
--- a/src/inc/msquic.h
+++ b/src/inc/msquic.h
@@ -74,12 +74,12 @@ typedef enum QUIC_EXECUTION_PROFILE {
     QUIC_EXECUTION_PROFILE_LOW_LATENCY,         // Default
     QUIC_EXECUTION_PROFILE_TYPE_MAX_THROUGHPUT,
     QUIC_EXECUTION_PROFILE_TYPE_SCAVENGER,
-    QUIC_EXECUTION_PROFILE_TYPE_REAL_TIME
+    QUIC_EXECUTION_PROFILE_TYPE_REAL_TIME,
 } QUIC_EXECUTION_PROFILE;
 
 typedef enum QUIC_LOAD_BALANCING_MODE {
     QUIC_LOAD_BALANCING_DISABLED,               // Default
-    QUIC_LOAD_BALANCING_SERVER_ID_IP            // Encodes IP address in Server ID
+    QUIC_LOAD_BALANCING_SERVER_ID_IP,           // Encodes IP address in Server ID
 } QUIC_LOAD_BALANCING_MODE;
 
 typedef enum QUIC_CREDENTIAL_TYPE {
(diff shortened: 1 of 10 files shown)


src/inc/msquic.h
```diff
@@ -91,11 +91,12 @@ typedef enum QUIC_CREDENTIAL_TYPE {
...
-    QUIC_CREDENTIAL_FLAG_CLIENT                     = 0x00000001, // Lack of client flag indicates server.
-    QUIC_CREDENTIAL_FLAG_LOAD_ASYNCHRONOUS          = 0x00000002,
-    QUIC_CREDENTIAL_FLAG_NO_CERTIFICATE_VALIDATION  = 0x00000004,
-    QUIC_CREDENTIAL_FLAG_ENABLE_OCSP                = 0x00000008
+    QUIC_CREDENTIAL_FLAG_NONE                           = 0x00000000,
+    QUIC_CREDENTIAL_FLAG_CLIENT                         = 0x00000001, // Lack of client flag indicates server.
+    QUIC_CREDENTIAL_FLAG_LOAD_ASYNCHRONOUS              = 0x00000002,
+    QUIC_CREDENTIAL_FLAG_NO_CERTIFICATE_VALIDATION      = 0x00000004,
+    QUIC_CREDENTIAL_FLAG_ENABLE_OCSP                    = 0x00000008, // Schannel only currently
+    QUIC_CREDENTIAL_FLAG_DEFER_CERTIFICATE_VALIDATION   = 0x00000010, // Schannel only currently
```
ThadHouse, line 99: We should be consistent about trailing commas, especially in the public headers. The standards for both C and C++ allow both, so I don't really care which one, but we should be consistent. I like having them, because it makes diffs adding more values cleaner.

nibanks: Agree having them is nicer, that's why I put it there.

ThadHouse: I agree. So we should make an issue to fixup the other uses, at least in the public headers for now, and more potentially later. Likely not worth doing in this PR.

nibanks: I'll just do it here. When everything else is signed off, I'll push that change and merge.

ThadHouse (approved):

src/platform/unittest/TlsTest.cpp
```diff
@@ -820,6 +866,47 @@ TEST_F(TlsTest, CertificateError)
     }
 }
 
+#ifdef _WIN32
+TEST_F(TlsTest, DeferredCertificateValidationAllow)
```
anrossi, line 870: Does the AZP YAML need to be updated to run this test on Schannel builds?

anrossi (approved):

src/platform/unittest/TlsTest.cpp
nibanks: Nope. But related to this, I'm going to slightly modify how this work, and make this test always run, but just exit silently if the sec config couldn't be created (i.e. unsupported by the platform).

PR #1196, opened 2021-01-27
Remove use of POOL_FLAG_UNINITIALIZED from non lookaside list allocations

### LINKED ISSUE #1194: Remove use of POOL_FLAG_UNINITIALIZED
POOL_FLAG_UNINITIALIZED is now considered unacceptable in kernel code except in exceptional performance scenarios. We need to have actual numbers of performance impact to continue use.

ThadHouse:
This parameter has a high bar to allow being used, and the recommendation is to not use it. In the lookaside list function, it's ok since those are reused anyway, but for generic allocations keep everything the way it is.

For now, this will keep any explicit zeroing as well. This doesn't seem to result in any performance regressions, but is something we can look at.

Closes #1194

### DIFF
--- a/src/inc/quic_platform_winkernel.h
+++ b/src/inc/quic_platform_winkernel.h
@@ -264,8 +264,8 @@ CxPlatLogAssert(
 
 extern uint64_t CxPlatTotalMemory;
 
-#define CXPLAT_ALLOC_PAGED(Size, Tag) ExAllocatePool2(POOL_FLAG_PAGED | POOL_FLAG_UNINITIALIZED, Size, Tag)
-#define CXPLAT_ALLOC_NONPAGED(Size, Tag) ExAllocatePool2(POOL_FLAG_NON_PAGED | POOL_FLAG_UNINITIALIZED, Size, Tag)
+#define CXPLAT_ALLOC_PAGED(Size, Tag) ExAllocatePool2(POOL_FLAG_PAGED, Size, Tag)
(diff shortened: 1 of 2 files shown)


src/inc/quic_platform_winkernel.h
```diff
@@ -264,8 +264,8 @@ CxPlatLogAssert(
 
 extern uint64_t CxPlatTotalMemory;
 
-#define CXPLAT_ALLOC_PAGED(Size, Tag) ExAllocatePool2(POOL_FLAG_PAGED | POOL_FLAG_UNINITIALIZED, Size, Tag)
-#define CXPLAT_ALLOC_NONPAGED(Size, Tag) ExAllocatePool2(POOL_FLAG_NON_PAGED | POOL_FLAG_UNINITIALIZED, Size, Tag)
+#define CXPLAT_ALLOC_PAGED(Size, Tag) ExAllocatePool2(POOL_FLAG_PAGED, Size, Tag)
```
nibanks, line 267: I know it's going to be more work, but we should do two things:

1. Update all the platform allocations to standardize the zero initialize.
2. Remove any unnecessary/additional zero'ing that is still there.

ThadHouse: Yeah I was planning to do that. Agreed yes its a lot of work.

ThadHouse:
I think we've decided to punt on this. We have a lot of test coverage to make sure we did this right.

PR #1204, opened 2021-01-29
Port Control Protocol Support

nibanks:
Adds support for the Port Control Protocol to the platform abstraction layer, which allows for configuring (where enabled) NATs to open up specific ports.


Anas198900 (approved):

src/platform/datapath_winuser.c
```diff
@@ -269,6 +269,11 @@ typedef struct CXPLAT_SOCKET {
     //
     uint8_t Internal : 1;
 
+    //
+    // Flag indicates the binding is being used for PCP.
+    //
+    BOOLEAN PcpBinding : 1;
```
ThadHouse, line 275: Since all the other flags use uint8_t here, we should probably use that here as well. I'm not sure I trust the compiler to combine these, even if BOOLEAN is mapped to uint8_t.

src/platform/datapath_winuser.c
```diff
@@ -1048,6 +1053,105 @@ CxPlatDataPathIsPaddingPreferred(
     return !!(Datapath->Features & CXPLAT_DATAPATH_FEATURE_SEND_SEGMENTATION);
 }
 
+_IRQL_requires_max_(PASSIVE_LEVEL)
+_Success_(QUIC_SUCCEEDED(return))
+QUIC_STATUS
+CxPlatDataPathGetGatewayAddresses(
```
ThadHouse, line 1059: Should this have some sort of SAL annotation that it allocates GatewayAddresses?

src/platform/datapath_winuser.c
```diff
@@ -1048,6 +1053,105 @@ CxPlatDataPathIsPaddingPreferred(
...
+    do {
+        Error =
+            GetAdaptersAddresses(
+                AF_UNSPEC,
+                Flags,
+                NULL,
+                AdapterAddresses,
+                &AdapterAddressesSize);
+        if (Error == ERROR_BUFFER_OVERFLOW) {
+            free(AdapterAddresses);
```
ThadHouse, line 1088: CXPLAT_FREE(AdapterAddresses, QUIC_POOL_DATAPATH_ADDRESSES);

ThadHouse: Also, in CXPLAT_FREE, we don't actually allow passing in NULL, as it dereferences the value to get the tag. So you need to protect this with an if not NULL.

src/platform/datapath_winuser.c
```diff
@@ -1048,6 +1053,105 @@ CxPlatDataPathIsPaddingPreferred(
...
+        goto Exit;
+    }
+
+    CxPlatZeroMemory(*GatewayAddresses, Index * sizeof(QUIC_ADDR));
+    *GatewayAddressesCount = Index;
+    Index = 0;
+
+    for (PIP_ADAPTER_ADDRESSES Iter = AdapterAddresses; Iter != NULL; Iter = Iter->Next) {
+        for (PIP_ADAPTER_GATEWAY_ADDRESS_LH Iter2 = Iter->FirstGatewayAddress; Iter2 != NULL; Iter2 = Iter2->Next) {
+            memcpy(&(*GatewayAddresses)[Index], Iter2->Address.lpSockaddr, sizeof(QUIC_ADDR));
```
ThadHouse, line 1143: We should use CxPlatCopyMemory here, since its more consistent with the rest of our codebase.

src/platform/datapath_winuser.c
nibanks: That's unfortunate. Are you sure we don't any other callers that assumed `CXPLAT_FREE(nullptr)` was acceptable in the code base? `free(nullptr)` and `delete nullptr` have always been acceptable.

src/platform/pcp.c
```diff
@@ -0,0 +1,574 @@
...
+            &GatewayAddresses,
+            &GatewayAddressesCount);
+    if (QUIC_FAILED(Status)) {
+        goto Exit;
+    }
+    CXPLAT_DBG_ASSERT(GatewayAddresses != NULL);
+    CXPLAT_DBG_ASSERT(GatewayAddressesCount != 0);
+
+    PcpContextSize = sizeof(CXPLAT_PCP) + (GatewayAddressesCount * sizeof(CXPLAT_SOCKET*));
+    PcpContext = (CXPLAT_PCP*)CXPLAT_ALLOC_PAGED(PcpContextSize, QUIC_POOL_PCP);
```
ThadHouse, line 165: I'm assuming PAGED is fine here, but then likely couldn't GatewayAddresses also be PAGED. Currently its allocated NONPAGED. Since PAGED memory is less of a system resources issues. (Not in user mode, but still).

src/platform/datapath_winuser.c
ThadHouse: We can fixup CXPLAT_FREE to make that work. Its just currently in debug mode it does not work.

src/platform/pcp.c
```diff
@@ -0,0 +1,574 @@
...
+Exit:
+
+    if (QUIC_FAILED(Status)) {
+        if (PcpContext != NULL) {
+            CxPlatPcpUninitialize(PcpContext);
+        }
+    }
+
+    if (GatewayAddresses != NULL) {
+        CXPLAT_FREE(GatewayAddresses, QUIC_POOL_DATAPATH_ADDRESSES);
```
ThadHouse, line 208: I wonder if we should have a function in the datapath to free this. There are not many other places in code where we depend on the caller to know how to free the memory. It just gets messy.

src/platform/pcp.c
nibanks: Actually, this needs to be nonpaged. I'll change it. Good find. (Not that we support kernel yet)

src/platform/pcp.c
nibanks: What if I put the function declaration in an internal header (not quic_datapath.h)?

src/platform/datapath_winuser.c
ThadHouse: Also, we should verify that ExFreePoolWithTag also has that same contract. I think it does, but the docs don't say.

src/platform/pcp.c
```diff
@@ -0,0 +1,574 @@
...
+    CXPLAT_PCP_EVENT Event;
+    CxPlatCopyMemory(Event.FAILURE.Nonce, Response->MAP.MappingNonce, CXPLAT_PCP_NONCE_LENGTH);
+    QUIC_ADDR InternalAddress;
+    CxPlatCopyMemory(&InternalAddress, &Datagram->Tuple->LocalAddress, sizeof(QUIC_ADDR));
+    InternalAddress.Ipv6.sin6_port = Response->MAP.InternalPort;
+    QUIC_ADDR ExternalAddress = {0};
+    QUIC_ADDR RemotePeerAddress = {0};
+
+    if (Response->ResultCode != PCP_RESULT_SUCCESS) {
+        Event.Type = CXPLAT_PCP_EVENT_FAILURE;
```
ThadHouse, line 273: Either we should have an extra space at the top of this if to match the other cases, or remove the space at the top of the other spaces. Your choice.

src/platform/pcp.c
ThadHouse: In all honesty it doesn't really matter, just something I noticed. Theres probably other places we do this, theyre just rarer. This is fine.

ThadHouse (approved):

PR #1208, opened 2021-02-02
Refactor Certificate Received Code Paths

### LINKED ISSUE #1063: Remote certificate validation

nibanks:
Now MsQuic only has one event that can be used by the app for multiple purposes. The new flags were slightly refactored as well.

Fixes #1063.

### DIFF
--- a/src/inc/msquic.h
+++ b/src/inc/msquic.h
@@ -96,8 +96,8 @@ typedef enum QUIC_CREDENTIAL_FLAGS {
     QUIC_CREDENTIAL_FLAG_LOAD_ASYNCHRONOUS              = 0x00000002,
     QUIC_CREDENTIAL_FLAG_NO_CERTIFICATE_VALIDATION      = 0x00000004,
     QUIC_CREDENTIAL_FLAG_ENABLE_OCSP                    = 0x00000008, // Schannel only currently
-    QUIC_CREDENTIAL_FLAG_DEFER_CERTIFICATE_VALIDATION   = 0x00000010, // Schannel only currently
-    QUIC_CREDENTIAL_FLAG_CUSTOM_CERTIFICATE_VALIDATION  = 0x00000020,
+    QUIC_CREDENTIAL_FLAG_INDICATE_CERTIFICATE_RECEIVED  = 0x00000010,
+    QUIC_CREDENTIAL_FLAG_DEFER_CERTIFICATE_VALIDATION   = 0x00000020, // Schannel only currently
 } QUIC_CREDENTIAL_FLAGS;
 
 DEFINE_ENUM_FLAG_OPERATORS(QUIC_CREDENTIAL_FLAGS);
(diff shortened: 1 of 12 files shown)


src/inc/msquic.h
```diff
@@ -761,8 +761,7 @@ typedef enum QUIC_CONNECTION_EVENT_TYPE {
     QUIC_CONNECTION_EVENT_DATAGRAM_SEND_STATE_CHANGED       = 12,
     QUIC_CONNECTION_EVENT_RESUMED                           = 13,   // Server-only; provides resumption data, if any.
     QUIC_CONNECTION_EVENT_RESUMPTION_TICKET_RECEIVED        = 14,   // Client-only; provides ticket to persist, if any.
-    QUIC_CONNECTION_EVENT_PEER_CERTIFICATE_VALIDATION       = 15,   // Only with QUIC_CREDENTIAL_FLAG_DEFER_CERTIFICATE_VALIDATION set
-    QUIC_CONNECTION_EVENT_PEER_CERTIFICATE_RECEIVED         = 16,   // Only with QUIC_CREDENTIAL_FLAG_CUSTOM_CERTIFICATE_VALIDATION set
+    QUIC_CONNECTION_EVENT_PEER_CERTIFICATE_RECEIVED         = 15,   // Only with QUIC_CREDENTIAL_FLAG_DEFER_CERTIFICATE_VALIDATION|QUIC_CREDENTIAL_FLAG_CUSTOM_CERTIFICATE_VALIDATION set
```
anrossi: does this mean for either flag, or only both ORed together?

nibanks: I've updated the comment already. The INDICATE_CERT_RECV flag controls this behavior.

src/platform/tls_stub.c
```diff
@@ -926,33 +926,34 @@ CxPlatTlsClientProcess(
...
-                            "[ tls][%p] ERROR, %s.",
-                            TlsContext->Connection,
-                            "Custom certificate validation failed");
-                        *ResultFlags |= CXPLAT_TLS_RESULT_ERROR;
-                        State->AlertCode = 42; // bad_certificate
-                        break;
-                    }
-
-                } else {
+                if (!CxPlatCertValidateChain(
```
anrossi, line 929: Real cert validation in Stub?!

anrossi (approved):

PR #1213, opened 2021-02-05
New Session Ticket Encryption Key API

nibanks:
Refactors the NST key encryption interface. Now it's set via `SetParam` instead of via the credential config upfront. Not currently working anywhere but on miTLS. Schannel kernel bring up is imminent. User mode is a bit farther off it seems. WIP.


src/platform/tls_schannel.c
```diff
@@ -1461,6 +1501,49 @@ CxPlatTlsSecConfigDelete(
...
+        Key[i].KeyingMaterialSize = KeyConfig[i].MaterialLength;
+        CxPlatCopyMemory(Key[i].KeyingMaterial, KeyConfig[i].Material, KeyConfig[i].MaterialLength);
+        CxPlatCopyMemory(Key[i].KeyId, KeyConfig[i].Id, sizeof(KeyConfig[i].Id));
+    }
+
+    SecPkgCred_SessionTicketKeys Keys;
+    Keys.cSessionTicketKeys = KeyCount;
+    Keys.pSessionTicketKeys = Key;
+    SECURITY_STATUS SecStatus =
+        SetCredentialsAttributesW(
```
nibanks, line 1530: Currently fails on user mode with `SEC_E_UNSUPPORTED_FUNCTION`.

nibanks:
Looks like the new APIs are working on kernel mode now. User mode still fails. So I need to do the following:

1. Add a disable NST macro and wrap the functionality that doesn't support it appropriately.
2. Add NST resumption specific (not depending on 0-RTT) tests.

src/platform/tls_schannel.c
```diff
@@ -1514,6 +1554,49 @@ CxPlatTlsSecConfigDelete(
...
+{
+    if (KeyCount > SCHANEL_MAX_TICKET_KEY_COUNT) {
+        return QUIC_STATUS_INVALID_PARAMETER;
+    }
+
+    SecPkgCred_SessionTicketKey Key[SCHANEL_MAX_TICKET_KEY_COUNT];
+    for (uint8_t i = 0; i < KeyCount; ++i) {
+        Key[i].TicketInfoVersion = SESSION_TICKET_INFO_V0;
+        Key[i].KeyingMaterialSize = KeyConfig[i].MaterialLength;
+        CxPlatCopyMemory(Key[i].KeyingMaterial, KeyConfig[i].Material, KeyConfig[i].MaterialLength);
```
anrossi, line 1575: Check that the key length passed in doesn't overrun the buffer

src/test/lib/ApiTest.cpp
```diff
@@ -143,86 +141,176 @@ void QuicTestValidateConfiguration()
...
-    LocalConfiguration = nullptr;
+        TEST_QUIC_SUCCEEDED(
+            MsQuic->ConfigurationLoadCredential(
+                LocalConfiguration,
+                &SelfSignedCredConfig));
+
+        QUIC_TICKET_KEY_CONFIG KeyConfig;
+        CxPlatZeroMemory(&KeyConfig, sizeof(KeyConfig));
+        KeyConfig.MaterialLength = 64;
+        TEST_QUIC_SUCCEEDED_OR_NOT_SUPPORTED(
```
anrossi, line 272: Won't this mask errors when support is added for other TLS?

nibanks: Only NOT_SUPPORTED. I essentially want to say NOT_SUPPORTED is an acceptable return from here; essentially treating the feature as optional as far as the tests are concerned.

anrossi (approved):

PR #1217, opened 2021-02-09
Add support for RSS on linux

ThadHouse:
We had a thread per socket context, so we only needed to configure the bpf for rss to get it working

### DIFF
--- a/scripts/performance-helper.psm1
+++ b/scripts/performance-helper.psm1
@@ -514,7 +514,7 @@ function Invoke-LocalExe {
     $LocalExtraFile = Join-Path $BasePath "ExtraRunFile.txt"
     $RunArgs = """--extraOutputFile:$LocalExtraFile"" $RunArgs"
     $TimeoutMs = ($Timeout - 5) * 1000;
-    $RunArgs = "--watchdog:$TimeoutMs $RunArgs"
+    $RunArgs = "-watchdog:$TimeoutMs $RunArgs"
 
     $FullCommand = "$Exe $RunArgs"
     Write-Debug "Running Locally: $FullCommand"
@@ -535,6 +535,14 @@ function Invoke-LocalExe {
         New-ItemProperty -Path $WerDumpRegPath -Name DumpFolder -PropertyType ExpandString -Value $LogDir -Force | Out-Null
     }
 
+    if (!$IsWindows) {
+        mkdir logs
+        lttng create msquic -o=./logs
+        lttng enable-event --userspace CLOG_*
+        lttng add-context --userspace --type=vpid --type=vtid
+        lttng start
+    }
+
     $Stopwatch =  [system.diagnostics.stopwatch]::StartNew()
 
     $LocalJob = Start-Job -ScriptBlock { & $Using:Exe ($Using:RunArgs).Split(" ") }
@@ -542,6 +550,11 @@ function Invoke-LocalExe {
     # Wait for the job to finish
     Wait-Job -Job $LocalJob -Timeout $Timeout | Out-Null
     Stop-Job -Job $LocalJob | Out-Null
+
+    if (!$IsWindows) {
+        lttng stop msquic
+        babeltrace --names all ./logs/* > quic.babel.txt
+    }
 
     $RetVal = Receive-Job -Job $LocalJob
 
--- a/scripts/performance-helper.psm1
+++ b/scripts/performance-helper.psm1
@@ -226,7 +226,6 @@ function Wait-ForRemote {
     $Socket = New-Object System.Net.Sockets.UDPClient
     for ($i = 0; $i -lt 120; $i++) {
         $Socket.Send(@(1), 1, $RemoteAddress, 9999) | Out-Null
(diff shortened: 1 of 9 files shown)


src/perf/lib/RpsClient.cpp
```diff
@@ -397,7 +397,9 @@ RpsConnectionContext::StreamCallback(
     case QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE:
         Worker->Client->StreamContextAllocator.Free(StrmContext);
         MsQuic->StreamClose(StreamHandle);
-        Worker->QueueSendRequest();
+        if (CxPlatTimeMs32() < (Worker->Client->RunTime + 3000)) {
```
nibanks: This is wrong. There should be a diff between times here. Also 3000 ms is too much IMO. 100 ms might be enough.

src/perf/lib/RpsClient.cpp
```diff
@@ -397,7 +399,10 @@ RpsConnectionContext::StreamCallback(
     case QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE:
         Worker->Client->StreamContextAllocator.Free(StrmContext);
         MsQuic->StreamClose(StreamHandle);
-        Worker->QueueSendRequest();
+        if ((CxPlatTimeMs32() - Worker->Client->StartTime)
```
nibanks, line 402: Use `CxPlatTimeDiff32` instead.

scripts/performance-helper.psm1
```diff
@@ -224,8 +224,15 @@ function Wait-ForRemote {
     param ($Job)
     # Ping sidechannel socket on 9999 to tell the app to die
     $Socket = New-Object System.Net.Sockets.UDPClient
-    $Socket.Send(@(1), 1, $RemoteAddress, 9999) | Out-Null
-    Wait-Job -Job $Job -Timeout 120 | Out-Null
+    for ($i = 0; $i -lt 120; $i++) {
+        $Socket.Send(@(1), 1, $RemoteAddress, 9999) | Out-Null
+        $Socket.Send(@(1), 1, $RemoteAddress, 9999) | Out-Null
```
nibanks, line 229: Why two? Seems unnecessary.

src/perf/lib/RpsClient.cpp
```diff
@@ -397,7 +399,10 @@ RpsConnectionContext::StreamCallback(
     case QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE:
         Worker->Client->StreamContextAllocator.Free(StrmContext);
         MsQuic->StreamClose(StreamHandle);
-        Worker->QueueSendRequest();
+        if (CxPlatTimeDiff32(Worker->Client->StartTime, CxPlatTimeMs32()) 
+            < (Worker->Client->RunTime + 3000)) {
```
nibanks, line 403: Still think `3000` is too much. No more than `1000` but I think `100` is also good.

src/platform/platform_linux.c
```diff
@@ -734,6 +734,27 @@ CxPlatThreadCreate(
...
+    CPU_ZERO(&CpuSet);
+    CPU_SET(ProcessorIndex, &CpuSet);
+
+    if (!pthread_setaffinity_np(Thread, sizeof(CpuSet), &CpuSet)) {
+        QuicTraceEvent(
+            LibraryError,
+            "[ lib] ERROR, %s.",
+            "pthread_setaffinity_np failed");
+    }
+    return QUIC_STATUS_SUCCESS;
```
nibanks, line 755: nit: newline before this

src/platform/platform_linux.c
```diff
@@ -734,6 +734,27 @@ CxPlatThreadCreate(
...
 
+QUIC_STATUS
+CxPlatSetCurrentThreadProcessorAffinity(
+    _In_ uint16_t ProcessorIndex
+    )
+{
+    pthread_t Thread;
+    cpu_set_t CpuSet;
+
+    Thread = pthread_self();
```
nibanks, line 745: nit:
```suggestion
    cpu_set_t CpuSet;
    pthread_t Thread = pthread_self();
```

src/platform/datapath_linux.c
```diff
@@ -1616,6 +1658,11 @@ CxPlatSocketCreateUdp(
         }
     }
 
+    Status = CxPlatSocketConfigureRss(&Binding->SocketContexts[0], SocketCount);
+    if (QUIC_FAILED(Status)) {
+        goto Exit;
+    }
```
nibanks, line 1664: Is this universally supported? What happens if it's not?

scripts/performance-helper.psm1
ThadHouse: I thought it was failing on the remote side, not the client side, so I just wanted to make sure they would send. I'll bring it back down to 1 now.

nibanks (approved):

src/platform/datapath_linux.c
ThadHouse: Probably a good point to not make this fail. Removed the status check, and added a note on why we're ignoring the status.

PR #1226, opened 2021-02-11
macOS Support

ThadHouse:
Basically restarted the macOS datapath off of the linux datapath rather then the windows datapath. Theyre a lot more similar in how they work, so makes sense to handle it this way.


ThadHouse:
This also includes all the changes from main, so we might want to merge those in manually, and then rebase this PR.

nibanks (approved):

src/platform/platform_posix.c
```diff
@@ -776,8 +854,20 @@ CxPlatCurThreadID(
...
+
     CXPLAT_STATIC_ASSERT(sizeof(pid_t) <= sizeof(uint32_t), "PID size exceeds the expected size");
     return syscall(__NR_gettid);
+
+#elif defined(CX_PLATFORM_DARWIN)
+
+    uint64_t tid;
+    pthread_threadid_np(NULL, &tid);
+    // XXX: check for failure
+    return (uint32_t)tid;
```
wfurt, line 868: This may be problematic when it wraps IMHO. Did you consider changing CxPlatCurThreadID() to return uint64_t? That would cover the Linux as well.

src/platform/platform_posix.c
```diff
@@ -514,7 +516,11 @@ CxPlatGetAbsoluteTime(
 
     CxPlatZeroMemory(Time, sizeof(struct timespec));
 
+#if defined(CX_PLATFORM_LINUX)
     ErrorCode = clock_gettime(CLOCK_MONOTONIC, Time);
+#elif defined(CX_PLATFORM_DARWIN)
```
wfurt, line 521: I'm not sure why you need this. clock_gettime() should work on macOS just fine AFAIK.

src/platform/platform_posix.c
ThadHouse: I will check these out tomorrow morning. The one who originally did the mac port did this, so I'll double check everything tomorrow. I'll check the other comment, and any other ones found as well.

src/platform/platform_posix.c
Maximus-: If I recall correctly, the issue I was experiencing was that clock_gettime was actually *not* monotonic (Seems I wasn't the only one: https://github.com/mobile-shell/mosh/pull/1116), which was breaking some of the pthread event waiters. I feel like my initial PR used mach_absolute_time...

wfurt: I would add note there in that case as it is not obvious. .NET runtime uses clock_gettime AFAIK and I'm not aware of any issue. It may not use assert or use that output in the same way.

src/platform/platform_posix.c
wfurt: maybe add same assert as above. So if it wraps it would be visible. I don't know how this is consumed and probably does not need to addressed right now.

Anas198900 (approved):

src/platform/platform_posix.c
```diff
@@ -17,6 +17,11 @@
 #include "quic_platform.h"
 #include "quic_platform_dispatch.h"
 #include "quic_trace.h"
+#ifdef CX_PLATFORM_LINUX
+#include <unistd.h>
+#include <sys/types.h>
+#else
+#endif
```
nibanks, line 24: Is this `#else` necessary?

src/platform/platform_posix.c
```diff
@@ -519,11 +524,14 @@ CxPlatGetAbsoluteTime(
...
+    //
+    // timespec_get is used on darwin, as CLOCK_MONOTONIC isn't actually
+    // monotonic according to our tests.
+    //
+    ErrorCode = timespec_get(Time, TIME_UTC);
 #endif // CX_PLATFORM_DARWIN
 
-    CXPLAT_DBG_ASSERT(ErrorCode == 0);
-    UNREFERENCED_PARAMETER(ErrorCode);
+    CXPLAT_DBG_ASSERT(ErrorCode != 0);
```
nibanks, line 534: You changed this assert to be the opposite of what it was? Is that not going to break Linux?

src/manifest/clog.sidecar
```diff
@@ -10436,6 +10444,7 @@
         "MacroConfiguration": {
           "linux": "lttng_plus",
           "stubs": "stubs",
+          "macos": "STDOUT",
```
nibanks, line 10447: I'm not convinced we practically want to do this.... the spew just for tests will be massive.

CMakeLists.txt
```diff
@@ -336,10 +336,17 @@ else()
...
 
     if(QUIC_ENABLE_LOGGING)
-        message(STATUS "Configuring for LTTng tracing")
-        set(CMAKE_CLOG_CONFIG_PROFILE linux)
-        list(APPEND QUIC_COMMON_DEFINES QUIC_CLOG)
-        include(FindLTTngUST)
+        if (APPLE)
+            message(STATUS "Configuring for macos tracing")
+            set(CMAKE_CLOG_CONFIG_PROFILE macos)
+            #set(CMAKE_CLOG_CONFIG_PROFILE stubs)
```
nibanks, line 342: Might be better to default to `stubs` instead of `macos` if we are going to use STDOUT for `macos`.

CMakeLists.txt
```diff
@@ -357,7 +364,7 @@ else()
 
     if(QUIC_ENABLE_SANITIZERS)
         message(STATUS "Configuring sanitizers")
-        list(APPEND QUIC_COMMON_FLAGS -fsanitize=address,leak,undefined -fsanitize-address-use-after-scope -Og -ggdb3 -fno-omit-frame-pointer -fno-optimize-sibling-calls)
+        list(APPEND QUIC_COMMON_FLAGS -fsanitize=address,leak,undefined -fsanitize-address-use-after-scope -O0 -ggdb3 -fno-omit-frame-pointer -fno-optimize-sibling-calls)
```
nibanks, line 367: Why this change?

src/platform/platform_posix.c
ThadHouse: Realized that so just fixed it.

CMakeLists.txt
```diff
@@ -478,6 +485,10 @@ if(QUIC_TLS STREQUAL "openssl")
                 linux-armv4 -DL_ENDIAN
                 --cross-compile-prefix=${GNU_MACHINE}${FLOAT_ABI_SUFFIX}-)
             list(APPEND OPENSSL_CONFIG_FLAGS -latomic)
+        elseif(CX_PLATFORM STREQUAL "darwin")
+            # need to build with Apple's compiler
+            # also, "no-shared" breaks the build for master-alpha4-quic-support
```
nibanks, line 490: Is this comment still valid?

CMakeLists.txt
ThadHouse: I'm not sure, Max did it.

src/core/crypto.c
```diff
@@ -599,6 +599,7 @@ QuicCryptoWriteOneFrame(
     CXPLAT_DBG_ASSERT(CryptoOffset >= (Crypto->TlsState.BufferTotalLength - Crypto->TlsState.BufferLength));
 
     QUIC_CONNECTION* Connection = QuicCryptoGetConnection(Crypto);
+
```
nibanks, line 602: nit: remove this extra line

CMakeLists.txt
ThadHouse: No. Fixed

src/inc/quic_platform.h
```diff
@@ -147,9 +147,9 @@ DEFINE_ENUM_FLAG_OPERATORS(CXPLAT_THREAD_FLAGS);
 #elif _WIN32
 #define CX_PLATFORM_TYPE 2
 #include <quic_platform_winuser.h>
-#elif CX_PLATFORM_LINUX
+#elif CX_PLATFORM_LINUX || CX_PLATFORM_DARWIN
 #define CX_PLATFORM_TYPE 3
-#include <quic_platform_linux.h>
+#include <quic_platform_posix.h>
 #else
```
nibanks, line 153: What do you think about defining different `CX_PLATFORM_TYPE` values here instead, but both still including *posix.h?

src/platform/platform_posix.c
```diff
@@ -771,13 +858,26 @@ CxPlatThreadWait(
...
+    CXPLAT_STATIC_ASSERT(sizeof(pid_t) <= sizeof(CXPLAT_THREAD_ID), "PID size exceeds the expected size");
+    return gettid();
+
+#elif defined(CX_PLATFORM_DARWIN)
+
+    uint64_t Tid;
+    int Res = pthread_threadid_np(NULL, &Tid);
+    CXPLAT_DBG_ASSERT(Res == 0);
+    CXPLAT_DBG_ASSERT(Tid <= UINT32_MAX);
+    return (uint32_t)Tid;
```
nibanks, line 877:
```suggestion
    return (CXPLAT_THREAD_ID)Tid;
```

src/platform/platform_posix.c
```diff
@@ -771,13 +858,26 @@ CxPlatThreadWait(
...
     void
     )
 {
-    CXPLAT_STATIC_ASSERT(sizeof(pid_t) <= sizeof(uint32_t), "PID size exceeds the expected size");
-    return syscall(__NR_gettid);
+
+#if defined(CX_PLATFORM_LINUX)
+
+    CXPLAT_STATIC_ASSERT(sizeof(pid_t) <= sizeof(CXPLAT_THREAD_ID), "PID size exceeds the expected size");
+    return gettid();
```
nibanks, line 869: Why change this from what it was?

ThadHouse: Because we shouldn't be using syscalls directly, and especially not the NR versions of the syscalls. gettid returns the exact same value, but actually in a posix compatible way.

src/inc/quic_platform.h
```diff
@@ -149,7 +149,10 @@ DEFINE_ENUM_FLAG_OPERATORS(CXPLAT_THREAD_FLAGS);
 #include <quic_platform_winuser.h>
 #elif CX_PLATFORM_LINUX
 #define CX_PLATFORM_TYPE 3
-#include <quic_platform_linux.h>
+#include <quic_platform_posix.h>
+elif CX_PLATFORM_DARWIN
```
nibanks, line 153: `#elif`

CMakeLists.txt
Maximus-: I changed it to -O0 because -Og is almost *never* what you want. -Og is "optimize for debuggability, in production", but closer to -O1, so real optimizations do happen which make debugging harder. 

Also nit: on macOS, it's -glldb for optimize debugging for lldb =) since gdb on macOS is mostly broken.

PR #1227, opened 2021-02-11
Telemetry Assert Events

nibanks:
Adds, off by default, support for telemetry asserts to trigger a trace event.

### DIFF
--- a/src/bin/winkernel/msquicpriv.kernel.vcxproj
+++ b/src/bin/winkernel/msquicpriv.kernel.vcxproj
@@ -116,12 +116,12 @@
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)'=='Debug'">
     <ClCompile>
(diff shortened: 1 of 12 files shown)


src/bin/winkernel/msquicpriv.kernel.vcxproj
```diff
@@ -116,12 +116,12 @@
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)'=='Debug'">
     <ClCompile>
-      <PreprocessorDefinitions>QUIC_EVENTS_MANIFEST_ETW;QUIC_LOGS_MANIFEST_ETW;SECURITY_KERNEL;SECURITY_WIN32;_DEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
ThadHouse, line 119: Do we need to define this in all the other kernel projects too? Otherwise the base static libraries won't have this defined, which is where we want that.

nibanks: I was thinking about it, but I think this is good for now. This is at least 98% of what we want.

nibanks: Updated core now.

ThadHouse (approved):

PR #1228, opened 2021-02-11
Periodically Snap-shot Performance Counters

nibanks:
Periodically (currently hard coded to once ever 30 seconds) the library will snap shot the performance counters. Then it can compare the new snap shot to the previous values and have some heuristics to alert the system that something bad might be happening.

### DIFF
--- a/src/core/library.h
+++ b/src/core/library.h
@@ -361,7 +361,7 @@ QuicPerfCounterTrySnapShot(
         return; // Not time to resample yet.
     }
 
-    if (TimeLast !=
+    if ((int64_t)TimeLast !=
         InterlockedCompareExchange64(
             (int64_t*)&MsQuicLib.PerfCounterSamplesTime,
             (int64_t)TimeNow,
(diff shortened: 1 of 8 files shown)


src/core/library.h
```diff
@@ -236,6 +236,12 @@ typedef struct QUIC_LIBRARY {
     const uint32_t* DefaultCompatibilityList;
     uint32_t DefaultCompatibilityListLength;
 
+    //
+    // Last sample of the performance counters
+    //
+    uint64_t PerfCounterSamplesTime;
+    int64_t PerfCounterSamples[QUIC_PERF_COUNTER_MAX];
```
nibanks, line 243: Probably worth wrapping this, and everything else in the telemetry assert define check.

nibanks: Never mind. Eventually I may want to DoS protection stuff here too.

anrossi (approved):

ThadHouse (approved):

ThadHouse (approved):

anrossi (approved):

PR #1233, opened 2021-02-12
Fix distribution of linux binaries

ThadHouse:
The header file name changed. Also updates the include guard.

### DIFF
--- a/scripts/package-distribution.ps1
+++ b/scripts/package-distribution.ps1
@@ -68,7 +68,7 @@ foreach ($Build in $AllBuilds) {
     if ($Platform -eq "windows" -or $Platform -eq "uwp") {
         $Headers += Join-Path $HeaderDir  "msquic_winuser.h"
     } else {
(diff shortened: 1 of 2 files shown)


scripts/package-distribution.ps1
```diff
@@ -68,7 +68,7 @@ foreach ($Build in $AllBuilds) {
     if ($Platform -eq "windows" -or $Platform -eq "uwp") {
         $Headers += Join-Path $HeaderDir  "msquic_winuser.h"
     } else {
-        $Headers += Join-Path $HeaderDir  "msquic_linux.h"
+        $Headers += Join-Path $HeaderDir  "msquic_posix.h"
```
nibanks, line 71: .lttng.so is also being grabbed below, and that won't work on macos.

ThadHouse: That already does a check if the binary exists or not, so that won't fail. But, the library name I think will. Let me check.

nibanks (approved):

PR #1238, opened 2021-02-13
Support Spurious Congestion Event

### LINKED ISSUE #1237: Reset Congestion Control State on Spurious Loss Detection
### Describe the bug

When a congestion event is determined to be spurious, we leave the CC window in the reduced state. We should revert the window in this situation.

1. Run WAN perf test with random reordering.
2. See goodput drop significantly.

nibanks:
Allows for congestion control to recover from spurious congestion events. Closed #1237. Latest data seems quite promising:
```
RttMs, BottleneckMbps, BottleneckBufferPackets, RandomLossDenominator, RandomReorderDenominator, ReorderDelayDeltaMs, Tcp, DurationMs, Pacing, RateKbps, RawRateKbps1, RawRateKbps2, RawRateKbps3
20, 200, 73, 0, 10000, 10, 0, 10000, 1, 115620, 114405, 112321, 120133
20, 200, 367, 0, 10000, 10, 0, 10000, 1, 180087, 182338, 178573, 179351
```
Compared to the corresponding latest main values of: `32739` and `91341`.


src/core/loss_detection.c
```diff
@@ -1238,6 +1238,16 @@ QuicLossDetectionProcessAckBlocks(
 
                 QuicLossValidate(LossDetection);
             }
+
+            if (LossDetection->LostPackets == NULL) {
+                //
+                // All previously considered lost packets are been found to be
```
maolson-msft, line 1244:
```suggestion
                // All previously considered lost packets were found to be
```

src/core/loss_detection.c
```diff
@@ -1238,6 +1238,16 @@ QuicLossDetectionProcessAckBlocks(
...
+                // spuriously lost. Inform congestion control so that it might
```
maolson-msft, line 1245:
```suggestion
                // spuriously lost. Inform congestion control.
```

src/core/loss_detection.c
```diff
@@ -1238,6 +1238,16 @@ QuicLossDetectionProcessAckBlocks(
...
+                // rollback the previous congestion event effects.
```
maolson-msft, line 1246:
```suggestion

```

src/core/congestion_control.h
```diff
@@ -170,4 +175,14 @@ QuicCongestionControlOnDataLost(
     _In_ uint64_t LargestPacketNumberSent,
     _In_ uint32_t NumRetransmittableBytes,
     _In_ BOOLEAN PersistentCongestion
-    );
\ No newline at end of file
+    );
+
+//
+// Called when all recently considered lost data was found to be actually
```
maolson-msft, line 181:
```suggestion
// Called when all recently considered lost data was actually
```

src/core/congestion_control.c
```diff
@@ -538,3 +547,27 @@ QuicCongestionControlOnDataLost(
...
+QuicCongestionControlOnSpuriousCongestionEvent(
+    _In_ QUIC_CONGESTION_CONTROL* Cc
+    )
+{
+    BOOLEAN PreviousCanSendState = QuicCongestionControlCanSend(Cc);
+
+    //
+    // Revert to previous state.
+    //
+    Cc->WindowMax = Cc->PrevWindowMax;
```
maolson-msft, line 562: How are we guaranteed at this point that the PrevXxx variables were initialized? It seems pretty indirect: we had a nonempty LostPackets list, but it's not the QuicCongestionControlOnCongestionEvent function that puts packets in that list. And I see QuicCongestionControlOnCongestionEvent isn't even always called by QuicCongestionControlOnDataLost, which in turn is called by a function (QuicLossDetectionDetectAndHandleLostPackets) that doesn't add packets to that list. I.e. this assumption is very complicated.

nibanks: I added a return at the beginning if somehow we weren't in recovery.

maolson-msft (approved):

PR #1239, opened 2021-02-16
CXPLAT_TEL_ASSERT Log Only

nibanks:
Updates the recently refactored CXPLAT_TEL_ASSERT to make sure it doesn't crash in release mode, just log.

### DIFF
--- a/src/inc/quic_platform_winkernel.h
+++ b/src/inc/quic_platform_winkernel.h
@@ -183,7 +183,7 @@ CxPlatLogAssert(
 #define CXPLAT_WIDE_STRING(_str) L##_str
 
 #define CXPLAT_ASSERT_NOOP(_exp, _msg) \
-    (CXPLAT_ANALYSIS_ASSUME(_exp), ((!(_exp)) ? FALSE : TRUE))
+    (CXPLAT_ANALYSIS_ASSUME(_exp), 0)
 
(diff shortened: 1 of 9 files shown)


src/inc/quic_platform_winkernel.h
```diff
@@ -162,78 +180,70 @@ CxPlatLogAssert(
...
+#define CXPLAT_WIDE_STRING(_str) L##_str
 
-#define QUIC_ASSERT_ACTION(_exp) \
+#define CXPLAT_ASSERT_NOOP(_exp, _msg) \
+    (CXPLAT_ANALYSIS_ASSUME(_exp), \
     ((!(_exp)) ? \
         (CxPlatLogAssert(__FILE__, __LINE__, #_exp), \
-         __annotation(L"Debug", L"AssertFail", QUIC_WIDE_STRING(#_exp)), \
+         __annotation(L"Debug", L"AssertFail", _msg), \
          DbgRaiseAssertionFailure(), FALSE) : \
```
anrossi, line 190: is `DbgRaiseAssertionFailure` platform specific? Maybe it should be moved into the cxplat library? (in a later PR)

nibanks: I believe it's platform specific. But I don't think we need to add an abstraction for it at this time.

anrossi (approved):

ThadHouse (approved):

Issue #110, opened 2020-02-06
Querying the handshake results


Tratcher:
Compare to [SslStream](https://docs.microsoft.com/en-us/dotnet/api/system.net.security.sslstream?view=netcore-3.1#properties).

When the handshake completes the caller may need to know the details of what was negotiated. This information is used for auditing, compatibility checks, and migrations. E.g. "Can I disable this old protocol now or are clients still using it?"

- [x] ALPN
- [x] SNI
- [ ] Server certificate
- [ ] Client certificate
- [ ] Any suppressed remote certificate validation errors
- [x] TLS protocol version (by the time TLS 1.4 gets started)
- [x] CipherAlgorithm
- [x] CipherStrength
- [x] HashAlgorithm
- [x] HashStrength
- [x] KeyExchangeAlgorithm
- [x] KeyExchangeStrength
- [x] CipherSuite

ThadHouse:
For Schannel, CipherSuite is grabbed with the following C# code.

https://github.com/dotnet/runtime/blob/f5874b08b53665ce950b76b628deb63abecaee85/src/libraries/System.Net.Security/src/System/Net/Security/SslStreamPal.Windows.cs#L326

The code for linux openssl is here.
https://github.com/dotnet/runtime/blob/1d9e50cb4735df46d3de0cee5791e97295eaf588/src/libraries/System.Net.Security/src/System/Net/Security/SslConnectionInfo.Linux.cs#L11

ThadHouse:
For an API, based on the existing .NET types, we're looking at something like this struct, which can be read using GetParam at the connection level.

```
struct QUIC_HANDSHAKE_INFO {
    uint32_t TlsProtocolVersion;
    uint32_t CipherAlgorithm;
    uint32_t CipherStrength;
    uint32_t Hash;
    uint32_t HashStrength;
    uint32_t KeyExchangeAlgorithm;
    uint32_t KeyExchangeStrength;
    // All fields above this map directly to SecPkgContext_ConnectionInfo from schannel. Openssl can be mapped to these easily.
    uint32_t CipherSuite; // Maps to dwCipherSuite in SecPkgContext_CipherInfo 
}
```

SNI, Client Cert, Server Cert would likely need to be passed during the QUIC_CONNECTION_EVENT_CONNECTED event as additional parameters.

ThadHouse:
This has been completed, other then a getter API for getting server and client certificate. However, for these, we think it would be better to use the events received to get these, and store them for later use if necessary.

PR #1241, opened 2021-02-16
Add support for querying handshake info

ThadHouse:
This works with OpenSSL and Schannel. Schannel values are used as the values returned from all platforms, as this matches what .NET would expect.

Completes a large chunk of #110 but getting SNI and certificates still needs to be investigated.


src/platform/tls_openssl.c
```diff
@@ -1260,6 +1260,118 @@ CxPlatTlsParamSet(
     return QUIC_STATUS_NOT_SUPPORTED;
 }
 
+//
+// Mappings taken from the following .NET definitions.
+// https://github.com/dotnet/runtime/blob/69425a7e6198ff78131ad64f1aa3fc28202bfde8/src/libraries/System.Net.Security/src/System/Net/Security/SslConnectionInfo.Linux.cs
+// https://github.com/dotnet/runtime/blob/69425a7e6198ff78131ad64f1aa3fc28202bfde8/src/libraries/Native/Unix/System.Security.Cryptography.Native/pal_ssl.c
+// https://github.com/dotnet/runtime/blob/1d9e50cb4735df46d3de0cee5791e97295eaf588/src/libraries/System.Net.Security/src/System/Net/Security/TlsCipherSuiteData.Lookup.cs
+//
```
nibanks, line 1268: Put these comments inside the function below?

src/platform/tls_openssl.c
```diff
@@ -1260,6 +1260,118 @@ CxPlatTlsParamSet(
...
+            break;
+    }
+
+    return Status;
+}
+
+static
+uint32_t
+CxPlatMapVersion(
+    _In_ const char* Version,
```
nibanks, line 1315:
```suggestion
    _In_z_ const char* Version,
```

src/platform/tls_openssl.c
```diff
@@ -1260,6 +1260,118 @@ CxPlatTlsParamSet(
...
+    _In_ BOOLEAN IsServer
+    )
+{
+    //
+    // protocolVersion points to a static ASCII string that's one of:
+    //     TLSv1
+    //     TLSv1.1
+    //     TLSv1.2
+    //     TLSv1.3
```
nibanks, line 1324: Anything less than TLSv1.3 is unsupported and I'd prefer not to have code that allows for it at all. We should either return an error or some "unsupported" version value.

src/platform/tls_openssl.c
```diff
@@ -1260,6 +1260,118 @@ CxPlatTlsParamSet(
...
+            {
+                return IsServer ? 0x40 : 0x80;
+            }
+            else if (Version[5] == '.' && Version[6] != '\0' && Version[7] == '\0')
+            {
+                switch (Version[6])
+                {
+                    case '1': return IsServer ? 0x100 : 0x200;
+                    case '2': return IsServer ? 0x400 : 0x800;
+                    case '3': return IsServer ? 0x1000 : 0x2000;
```
nibanks, line 1349: I don't like all the magic numbers. We should define them in msquic.h. We can wrap the definition with `#ifndef XXX` if they're already defined for Windows.

src/platform/unittest/TlsTest.cpp
```diff
@@ -738,6 +738,44 @@ TEST_F(TlsTest, Handshake)
...
+    QUIC_HANDSHAKE_INFO HandshakeInfo;
+    CxPlatZeroMemory(&HandshakeInfo, sizeof(HandshakeInfo));
+    uint32_t HandshakeInfoLen = sizeof(HandshakeInfo);
+    QUIC_STATUS Status =
+        CxPlatTlsParamGet(
+            ClientContext.Ptr,
+            QUIC_PARAM_TLS_HANDSHAKE_INFO,
+            &HandshakeInfoLen,
+            &HandshakeInfo);
+    ASSERT_TRUE(QUIC_SUCCEEDED(Status));
```
nibanks, line 757: Let's validate as much of the actual return data as possible (like TLS version).

src/inc/msquic.h
```diff
@@ -276,6 +276,59 @@ typedef struct QUIC_NEW_CONNECTION_INFO {
...
+#define QUIC_TLS1_3_SERVER      SP_PROT_TLS1_3_SERVER
+#else
+#define QUIC_TLS1_3_SERVER      0x00001000
+#endif
+#ifdef SP_PROT_TLS1_3_CLIENT
+#define QUIC_TLS1_3_CLIENT      SP_PROT_TLS1_3_CLIENT
+#else
+#define QUIC_TLS1_3_CLIENT      0x00002000
+#endif
+#define QUIC_TLS1_3             (QUIC_TLS1_3_SERVER | QUIC_TLS1_3_CLIENT)
```
nibanks, line 316: Do we _really_ need to expose the protocol as server or client? I really don't like this, and would prefer to expose just TLS1.3.

src/platform/tls_openssl.c
```diff
@@ -1260,6 +1260,94 @@ CxPlatTlsParamSet(
...
+    HandshakeInfo->KeyExchangeAlgorithm = 0;
+    HandshakeInfo->KeyExchangeStrength = 0;
+    HandshakeInfo->HashStrength = 0;
+
+    switch (HandshakeInfo->CipherSuite) {
+        case QUIC_CIPHER_SUITE_TLS_AES_128_GCM_SHA256:
+            HandshakeInfo->CipherAlgorithm = QUIC_ALG_AES_128;
+            HandshakeInfo->CipherStrength = 128;
+            HandshakeInfo->Hash = QUIC_ALG_SHA_256;
+
```
nibanks, line 1285: nit: extra newline?

src/platform/tls_openssl.c
```diff
@@ -1260,6 +1260,94 @@ CxPlatTlsParamSet(
...
+            }
+            if (Version[5] == '.' && Version[6] != '\0' && Version[7] == '\0')
+            {
+                switch (Version[6])
+                {
+                    case '3': return IsServer ? QUIC_TLS1_3_SERVER : QUIC_TLS1_3_CLIENT;
+                }
+            }
+        }
+    }
```
nibanks, line 1345: Formatting of `{`

src/inc/msquic.h
ThadHouse: Schannel exposes it this way. We could map it internally if that would be preferred.

nibanks: I don't particularly care how schannel maps it. :) I prefer to just map the TLS protocol used, independent of role.

src/inc/msquic.h
```diff
@@ -276,6 +276,54 @@ typedef struct QUIC_NEW_CONNECTION_INFO {
...
 
+#define QUIC_CIPHER_SUITE_TLS_AES_128_GCM_SHA256          0x1301
+#define QUIC_CIPHER_SUITE_TLS_AES_256_GCM_SHA384          0x1302
+#define QUIC_CIPHER_SUITE_TLS_CHACHA20_POLY1305_SHA256    0x1303
+
+#ifdef CALG_AES_128
+#define QUIC_ALG_AES_128       CALG_AES_128
+#else
+#define QUIC_ALG_AES_128       0x660E
+#endif
```
nibanks, line 287: I'm not a super big fan of redefining these all in this way. What about if we just used the values directly and defined them if they don't exist (thus standardizing on the existing Windows value):
```
#ifndef CALG_AES_128
#define CALG_AES_128 0x660E
#endif
```

ThadHouse: That gets a bit weird, because some of them don't have a matching value in schannel, and I don't want to collide with them when they add chacha20.

src/platform/tls_schannel.c
```diff
@@ -2495,7 +2531,7 @@ CxPlatTlsParamGet(
     _In_ CXPLAT_TLS* TlsContext,
     _In_ uint32_t Param,
     _Inout_ uint32_t* BufferLength,
-    _Inout_updates_bytes_opt_(*BufferLength)
+    _Out_writes_bytes_opt_(*BufferLength)
```
nibanks, line 2534: In the case of `QUIC_PARAM_TLS_SCHANNEL_CONTEXT_ATTRIBUTE_W`, `_Inout_updates_bytes_opt_` is correct. It both reads and writes (i.e. updated) the data stored there.

ThadHouse: So we should update QUIC_GET_PARAM_FN to be _Inout_updates_bytes_opt_. The SAL annotation should match all the way down.

src/inc/msquic.h
nibanks: Once it's already defined by Windows, then your `#ifndef` gets disabled. Also, why isn't there a value for ChaCha20? There has to be one, right? What is Schannel returning in this case (they supposedly support ChaCha20)?

ThadHouse: I couldn't find a single ChaCha20 definition in schannels headers.

ThadHouse: If the ifdef gets disabled to a different value, thats a silent ABI breaking change. Definitely NOT ok with that.

nibanks: I agree we shouldn't invent a new value ourselves. We need to find the Windows value.

src/platform/tls_schannel.c
nibanks: Leave that to a separate PR and just undo this change for now?

ThadHouse: Ok made #1242  to track it.

src/inc/msquic.h
ThadHouse: Also, just like we did for address family, if you want to be a good cross platform API, building your own abstraction level really is much better from a usability perspective. The bit of defines here is worth it IMO.

nibanks: Well, if you use the address family as the example, these should go in the platform specific headers.

ThadHouse: Ok, I'm fine with that.

src/inc/msquic.h
```diff
@@ -276,6 +276,52 @@ typedef struct QUIC_NEW_CONNECTION_INFO {
...
+//
+typedef struct QUIC_HANDSHAKE_INFO {
+    int32_t TlsProtocolVersion;
+    int32_t CipherAlgorithm;
+    int32_t CipherStrength;
+    int32_t Hash;
+    int32_t HashStrength;
+    int32_t KeyExchangeAlgorithm;
+    int32_t KeyExchangeStrength;
+    int32_t CipherSuite;
```
nibanks, line 322: Personally, I don't like that these are just random `int32_t`. I'd much rather `enum`s for most of these. Just because Schannel does it this way doesn't make it right... Thoughts?

src/inc/quic_tls.h
```diff
@@ -417,7 +417,7 @@ CxPlatTlsParamGet(
     _In_ CXPLAT_TLS* TlsContext,
     _In_ uint32_t Param,
     _Inout_ uint32_t* BufferLength,
-    _Inout_updates_bytes_opt_(*BufferLength)
+    _Out_writes_bytes_opt_(*BufferLength)
```
nibanks, line 420: Let's just not change any of these annotations for now and leave fixing these up for that other issue.

src/platform/tls_openssl.c
```diff
@@ -1260,6 +1260,74 @@ CxPlatTlsParamSet(
...
+    //
+    // protocolVersion points to a static ASCII string that's one of:
+    //     TLSv1
+    //     TLSv1.1
+    //     TLSv1.2
+    //     TLSv1.3
+    //     SSLv2
+    //     SSLv3
+    //     unknown
+    // Regardless, it's null terminated.
```
nibanks, line 1322: This huge comment seems unnecessary now. Maybe simplify it?

src/inc/msquic.h
ThadHouse: Moving to enums made this be solved in a different way. So we're good there.

src/inc/msquic.h
ThadHouse: Good idea. Done. It also implicitly solved the definition issue.

src/platform/tls_openssl.c
```diff
@@ -1269,11 +1326,68 @@ CxPlatTlsParamGet(
...
+            if (Buffer == NULL) {
+                Status = QUIC_STATUS_INVALID_PARAMETER;
+                break;
+            }
+
+            const uint8_t* NegotiatedAlpn;
+            unsigned int NegotiatedAlpnLen = 0;
+            SSL_get0_alpn_selected(TlsContext->Ssl, &NegotiatedAlpn, &NegotiatedAlpnLen);
+            if (NegotiatedAlpnLen <= 0) {
+                Status = QUIC_STATUS_NOT_SUPPORTED;
```
nibanks, line 1371: ALPN is always supported. I think INVALID_STATE would most likely be the reason. It would probably be worth a log on failure as well.

src/platform/tls_openssl.c
```diff
@@ -1269,11 +1326,68 @@ CxPlatTlsParamGet(
...
+            }
+
+            QUIC_HANDSHAKE_INFO* HandshakeInfo = (QUIC_HANDSHAKE_INFO*)Buffer;
+            HandshakeInfo->TlsProtocolVersion =
+                CxPlatMapVersion(
+                    SSL_get_version(TlsContext->Ssl));
+
+            const SSL_CIPHER* Cipher = SSL_get_current_cipher(TlsContext->Ssl);
+            if (Cipher == NULL) {
+                Status = QUIC_STATUS_NOT_SUPPORTED;
```
nibanks, line 1352: Again, I don't think NOT_SUPPORTED is right here. Can you see why this would ever fail? Also, more logging would be nice I think.

src/platform/tls_schannel.c
```diff
@@ -215,6 +215,42 @@ typedef struct _SecPkgContext_SessionInfo
 #define SECPKG_ATTR_SESSION_INFO         0x5d   // returns SecPkgContext_SessionInfo
 #define SECPKG_ATTR_CERT_CHECK_RESULT_INPROC 0x72 // returns SecPkgContext_CertificateValidationResult, use only after SSPI handshake loop
 
+typedef unsigned int ALG_ID;
+
+typedef struct _SecPkgContext_CipherInfo
+{
+
```
nibanks, line 222: I know this was probably copy/paste, but (nit) lets remove this extra space.

src/platform/tls_schannel.c
```diff
@@ -2527,9 +2563,121 @@ CxPlatTlsParamGet(
...
+            HandshakeInfo->CipherAlgorithm = ConnInfo.aiCipher;
+            HandshakeInfo->CipherStrength = ConnInfo.dwCipherStrength;
+            HandshakeInfo->Hash = ConnInfo.aiHash;
+            HandshakeInfo->HashStrength = ConnInfo.dwHashStrength;
+            HandshakeInfo->KeyExchangeAlgorithm = ConnInfo.aiExch;
+            HandshakeInfo->KeyExchangeStrength = ConnInfo.dwExchStrength;
+            HandshakeInfo->CipherSuite = CipherInfo.dwCipherSuite;
+
+            break;
```
nibanks, line 2631:
```suggestion
            HandshakeInfo->CipherSuite = CipherInfo.dwCipherSuite;
            break;
```

nibanks (approved):
Looks good. Just a few minor comments.

nibanks (approved):

PR #1245, opened 2021-02-17
RACK based Loss Detection Only

nibanks:
- Disables FACK based loss detection.


maolson-msft:
Why also change the time threshold? I think you should just change one thing at a time.

nibanks:
> Why also change the time threshold? I think you should just change one thing at a time.

I can split the changes out.

maolson-msft:
> > Why also change the time threshold? I think you should just change one thing at a time.
> 
> I can split the changes out.

Thanks. We didn't discuss changing the time threshold so it's not clear what your thinking is there. I think you should test just removing the packet threshold against your WAN perf testcases.

maolson-msft (approved):

nibanks:
Going to close for now. I don't have really strong confidence in this. For instance, in DC environments (very low RTT), FACK might be very useful.

PR #1251, opened 2021-02-19
Cherry-Pick: New Stream Start Flag for Shutdown on Failure (#1248)

anrossi:


nibanks:
Not going to take this for 1.0 after all.

PR #1253, opened 2021-02-19
Fix Client Certificate Test

anrossi:


nibanks:
So, the client _was_ using a server certificate before, and it was working?! Did it have the proper EKU, or did we just disable cert validation so it "just worked" because _a_ certificate was supplied?

anrossi:
> So, the client _was_ using a server certificate before, and it was working?! Did it have the proper EKU, or did we just disable cert validation so it "just worked" because _a_ certificate was supplied?

Yes, the client was using the server certificate before and was working, The server side doesn't have certificate validation disabled, so I don't believe that's why it was working.

nibanks (approved):

PR #1254, opened 2021-02-19
Add unreachable support to datapath_kqueue, add synchronous case to datapath_epoll

ThadHouse:
The synchronous cases can sometimes occur, theyre likely rare but still probably good to handle.

kqueue doesn't give a special event type for socket error, but does give a recvmsg with an error in that case.

I want to do some double checking that this works, but initial testing looks like it does.

### DIFF
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -2218,6 +2218,19 @@ CxPlatSocketSendInternal(
                 SocketContext->Binding,
                 Status,
                 "sendmsg failed");
+
+            //
+            // Unreachable events can sometimes come synchronously.
+            // Send unreachable notification to MsQuic if any related
+            // errors were received.
+            //
+            if (Status == ECONNREFUSED ||
+                Status == EHOSTUNREACH ||
+                Status == ENETUNREACH) {
+                SocketContext->Binding->Datapath->UdpHandlers.Unreachable(
+                    SocketContext->Binding,
+                    SocketContext->Binding->ClientContext,
+                    &SocketContext->Binding->RemoteAddress);
             goto Exit;
         }
     }
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -2231,6 +2231,7 @@ CxPlatSocketSendInternal(
                     SocketContext->Binding,
                     SocketContext->Binding->ClientContext,
                     &SocketContext->Binding->RemoteAddress);
+            }
             goto Exit;
         }
     }
--- a/src/platform/datapath_kqueue.c
+++ b/src/platform/datapath_kqueue.c
@@ -1361,47 +1361,6 @@ CxPlatSocketContextProcessEvents(
         return;
     }
 
-    // TODO figure out what these mean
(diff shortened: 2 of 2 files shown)


src/platform/datapath_epoll.c
```diff
@@ -2218,6 +2218,19 @@ CxPlatSocketSendInternal(
...
+
+            //
+            // Unreachable events can sometimes come synchronously.
+            // Send unreachable notification to MsQuic if any related
+            // errors were received.
+            //
+            if (Status == ECONNREFUSED ||
+                Status == EHOSTUNREACH ||
+                Status == ENETUNREACH) {
+                SocketContext->Binding->Datapath->UdpHandlers.Unreachable(
```
nibanks, line 2230: I think it might be Ok, but this is the first case where we might have an inline indication/upcall.

src/platform/datapath_kqueue.c
```diff
@@ -1412,13 +1371,29 @@ CxPlatSocketContextProcessEvents(
...
+                    // errors were received.
+                    //
+                    if (ErrNum == ECONNREFUSED ||
+                        ErrNum == EHOSTUNREACH ||
+                        ErrNum == ENETUNREACH) {
+                        SocketContext->Binding->Datapath->UdpHandlers.Unreachable(
+                            SocketContext->Binding,
+                            SocketContext->Binding->ClientContext,
+                            &SocketContext->Binding->RemoteAddress);
+            }
```
nibanks, line 1396: nit: fix indentation

src/platform/datapath_epoll.c
ThadHouse: It is. I didn't see an issue with is, as it just queues the message. However, I'll have to see how that interacts with this send then resulting in an error. Its possible this isn't necessary depending on what happens upstream if send actually fails.

ThadHouse: A failure in CxPlatSocketSend is just straight up ignored, so we likely do need something like this. The easiest way to reproduce this was to send 2 packets in a row. 1st packet would act like it sent, and queue the failure into epoll/kqueue. 2nd packet would actually error during the send.

nibanks: I'm not worried about the send call failing. (I'm not really worried at all actually) I'm just pointing out that before this, we never had a scenario, that I know of, where a recv-type indication/up-call would happen inline/inside a send down-call.

Maximus-: I'm curious if it's better to queue up a failed send and wait for an event from the kernel indicating that there's available buffer. In the case that you don't queue it up, you are possibly wasting precious throughput, but if you do, you run the risk of the recovery logic sending the packet twice; This is in the context of the Linux datapath vs Winuser, since linux datapath used to queue up failed sends iirc

ThadHouse: We do queue up failed sends if its reported that they would block. This logic is specifically the send returning unreachable or one of its friends. If the socket returns would block, it will queue, and then unreachable might come later.

nibanks (approved):

PR #1255, opened 2021-02-19
Update Deployment.md

clockzhong:
"visiable" should be "visible", right?


nibanks (approved):
Thanks for spotting this!

PR #1258, opened 2021-02-19
Add support for validating certificates in openssl using the windows cert store

### LINKED ISSUE #1202: Accept Valid Certificates on OpenSSL
Right now, unless the options to skip validation are enabled, OpenSSL will not accept any certificates (after #1190 is merged) by default (peer certificate validation enabled). We should look into making this work. On Linux, the cert files are fine, however on Windows, we likely would like a way to check against the windows certificate store.

Here's how to enable it on Linux
https://mta.openssl.org/pipermail/openssl-users/2018-September/008691.html

Here's some samples for loading from the Windows certificate store.
https://stackoverflow.com/questions/9507184/can-openssl-on-windows-use-the-system-certificate-store

Although for Windows, I want to try using a custom verify callback that we could pass the certificate to Windows directly and get that validated, rather then needing to load everything in.

ThadHouse:
Closes #1202


src/platform/tls_openssl.c
```diff
@@ -202,16 +202,29 @@ CxPlatTlsAlpnSelectCallback(
...
 CxPlatTlsCertificateVerifyCallback(
     int preverify_ok,
     X509_STORE_CTX *x509_ctx
     )
 {
+    X509* Cert = X509_STORE_CTX_get0_cert(x509_ctx);
     SSL *Ssl = X509_STORE_CTX_get_ex_data(x509_ctx, SSL_get_ex_data_X509_STORE_CTX_idx());
     CXPLAT_TLS* TlsContext = SSL_get_app_data(Ssl);
 
+    if (!preverify_ok) {
```
ThadHouse, line 222: Is there every a case where if openssl says the certificate is valid where we still want to check the windows store?

src/platform/cert_capi_openssl.c
```diff
@@ -39,6 +39,63 @@
...
+    if (OpenSSLCertLength <= 0) {
+        QuicTraceEvent(
+            LibraryError,
+            "[ lib] ERROR, %s.",
+            "i2d_X509 failed");
+        goto Exit;
+    }
+
+    CertContext =
+        (PCCERT_CONTEXT)CertCreateContext(
```
nibanks, line 64: Personal preference:
```suggestion
        (PCCERT_CONTEXT)
        CertCreateContext(
```

src/platform/cert_capi_openssl.c
```diff
@@ -290,4 +347,14 @@ CxPlatTlsExtractPrivateKey(
     UNREFERENCED_PARAMETER(X509Cert);
     return QUIC_STATUS_NOT_SUPPORTED;
 }
+BOOLEAN
+CxPlatTlsVerifyCertificate(
```
nibanks, line 351: Isn't this already defined above?

ThadHouse: This is for Unix.

nibanks (approved):

PR #1259, opened 2021-02-20
fix posix QuicAddrToString

wfurt:
QuicAddrToString has two issues:

1. Addr->Ip.sa_family is either QUIC_ADDRESS_FAMILY_INET or QUIC_ADDRESS_FAMILY_INET6. Passing it directly to `inet_ntop()` does not work on platforms where QUIC_ADDRESS_FAMILY_INET6 differs from AF_INET6.
2. The underlying socket structures are:
```c
 struct sockaddr_in {
               sa_family_t    sin_family; /* address family: AF_INET */
               in_port_t      sin_port;   /* port in network byte order */
               struct in_addr sin_addr;   /* internet address */
 };
 struct sockaddr_in6 {
               sa_family_t     sin6_family;   /* AF_INET6 */
               in_port_t       sin6_port;     /* port number */
               uint32_t        sin6_flowinfo; /* IPv6 flow information */
               struct in6_addr sin6_addr;     /* IPv6 address */
               uint32_t        sin6_scope_id; /* Scope ID (new in 2.4) */
 };
```

since the v6 has flowinfo, passing `&Addr->Ipv4.sin_addr` gives wrong pointer for IPv6. 

In process of writing test, I discover that `QuicAddr6FromString` did not work properly either for IPv6 case with port. Existing code would leave leading '[' in TmpAddrStr and  `inet_pton` would fail. 

I added basic test for the functions. Since it did not seem to fit anywhere else, I added new test class for helper functions.

### DIFF
--- a/src/platform/unittest/CMakeLists.txt
+++ b/src/platform/unittest/CMakeLists.txt
@@ -5,6 +5,7 @@ set(SOURCES
     main.cpp
     CryptTest.cpp
     DataPathTest.cpp
+    Platform.cpp
     # StorageTest.cpp
     TlsTest.cpp
 )
--- a/src/platform/unittest/CMakeLists.txt
+++ b/src/platform/unittest/CMakeLists.txt
@@ -5,7 +5,7 @@ set(SOURCES
     main.cpp
     CryptTest.cpp
     DataPathTest.cpp
-    Platform.cpp
+    PlatformTest.cpp
     # StorageTest.cpp
     TlsTest.cpp
 )
--- a/src/platform/unittest/PlatformTest.cpp
+++ b/src/platform/unittest/PlatformTest.cpp
@@ -13,7 +13,7 @@ Abstract:
 
 #include "msquic.h"
 #ifdef QUIC_CLOG
-#include "Platform.cpp.clog.h"
+#include "PlatformTest.cpp.clog.h"
 #endif
 
 struct PlatformTest : public ::testing::TestWithParam<int32_t>
--- a/src/platform/unittest/PlatformTest.cpp
+++ b/src/platform/unittest/PlatformTest.cpp
@@ -25,7 +25,7 @@ TEST(PlatformTest, QuicAddrParsing)
     struct TestEntry {
         const char* Input;
         int Family;
-        int Port;
+        unsigned short Port;
     };
 
     TestEntry TestData[] = {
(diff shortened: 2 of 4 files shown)


src/platform/unittest/CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ set(SOURCES
     main.cpp
     CryptTest.cpp
     DataPathTest.cpp
+    Platform.cpp
```
ThadHouse, line 8: Please call this PlatformTest.cpp, not Platform.cpp please.

wfurt: make sense. updated.

wfurt:
is it expected top run the tests explicitly @ThadHouse ? If so, I can do that on next changes.

ThadHouse:
For people not on the project, azure will not run tests automatically. They have to be explicitly started by one of us.

wfurt:
ok. seems like this change broke windows build. I'm looking into it.

ThadHouse:
It looks like we don't have a QuicAddr6FromString or QuicAddr4FromString in the windows headers. Thats probably something we should have, since we would prefer to have the same API in both OS's. I'll leave that up to @nibanks but since thats the public header I'd prefer them have the same API's

wfurt:
let me know @@nibanks. I can either add the missing functions for Windows or we can make the test platform specific.

nibanks:
> let me know @@nibanks. I can either add the missing functions for Windows or we can make the test platform specific.

`QuicAddrFromString` is the platform abstract function we expose. Can you please just update your tests to call that instead of the Linux IP-version specific variants separately?

src/platform/unittest/PlatformTest.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+    QUIC_ADDR Addr;
+    QUIC_ADDR_STR AddrStr = { 0 };
+
+    for (int i = 0; i < (int)(sizeof(TestData) / sizeof(struct TestEntry)); i++) {
+        CxPlatZeroMemory(&Addr, sizeof(QUIC_ADDR));
+        TestEntry* entry = &TestData[i];
+
+        ASSERT_TRUE(QuicAddrFromString(entry->Input, entry->Port, &Addr));
+        if (entry->Family == QUIC_ADDRESS_FAMILY_INET) {
```
nibanks, line 47: Don't access internal fields of the addr struct directly. Use the QuicAddrGet* functions instead. We don't provide any guarantees for the struct fields not changing.

wfurt: ok. easy to fix. I assume it is ok to do that inside of the platform code e.g. platform_kqueue, right?

nibanks: Yes. Platform specific code is free to access platform specific fields.

nibanks (approved):
LTGM. Assuming AZP passes, it's good to go. Thanks!

PR #1260, opened 2021-02-20
improve datapath on macOS

wfurt:
with this change all `AppData/*` _could_ pass on 10.15/x64. However, when running as batch the run randomly hangs. 

As far as I can tell, the troubles with PKTINFO are only when connected to IPv4 endpoint. For a while I was on path to avoid connect() and mimic the behavior in the platform path. But then I realized that at that point also local endpoint is resolved to IPv4 and the duality really does not matter. All the "wildcard" sockets work and there is no need to create separate socket for each address family. With that the change is reasonably small and sufficient for all existing tests. 

At this point I feel the hanging tests is separate issue. It could be some missed kqueue processing as I saw some notes in the file. However I got this alert several times
```
(lldb) bt
* thread #29, stop reason = signal SIGABRT
  * frame #0: 0x00007fff724c233a libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff7257ee60 libsystem_pthread.dylib`pthread_kill + 430
    frame #2: 0x00007fff72449808 libsystem_c.dylib`abort + 120
    frame #3: 0x000000010096f1e9 libmsquic.dylib`quic_bugcheck at platform_posix.c:62:5
    frame #4: 0x000000010096f7b0 libmsquic.dylib`CxPlatEventSet(Event=<unavailable>) at platform_posix.c:396:5
    frame #5: 0x000000010094b014 libmsquic.dylib`QuicConnDrainOperations(Connection=0x000000010103b400) at connection.c:6594:13
    frame #6: 0x00000001009686a4 libmsquic.dylib`QuicWorkerProcessConnection(Worker=0x0000000104824d18, Connection=0x000000010103b400) at worker.c:488:9
    frame #7: 0x0000000100967b6d libmsquic.dylib`QuicWorkerThread(Context=0x0000000104824d18) at worker.c:576:13
    frame #8: 0x00007fff7257f109 libsystem_pthread.dylib`_pthread_start + 148
    frame #9: 0x00007fff7257ab8b libsystem_pthread.dylib`thread_start + 15
```
and that may suggest some more fundamental problem. 
When the test hang, it generally stays in `CxPlatEventWaitForever`
```
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
  * frame #0: 0x00007fff724be882 libsystem_kernel.dylib`__psynch_cvwait + 10
    frame #1: 0x00007fff7257f425 libsystem_pthread.dylib`_pthread_cond_wait + 698
    frame #2: 0x000000010096f861 libmsquic.dylib`CxPlatEventWaitForever(Event=0x00000001021276b0) at platform_posix.c:426:9
    frame #3: 0x000000010093ec4e libmsquic.dylib`MsQuicSetParam(Handle=0x000000010281cc00, Level=QUIC_PARAM_LEVEL_CONNECTION, Param=4, BufferLength=112, Buffer=0x00007ffeefbfefe0) at api.c:1365:5
    frame #4: 0x0000000100087bc8 msquictest`TestConnection::SetPeerUnidiStreamCount(this=<unavailable>, value=<unavailable>) at TestConnection.cpp:486:12
    frame #5: 0x0000000100070cdd msquictest`NewPingConnection(Registration=0x00007ffeefbff2c0, ClientStats=0x00007ffeefbff270, UseSendBuffer=<unavailable>) at DataTest.cpp:312:21
```

it is curious that if I attach debugger, break and continue execution everything works as expected. 
I'm going to investigate.


src/platform/datapath_kqueue.c
```diff
@@ -721,126 +724,208 @@ CxPlatSocketContextInitialize(
...
-            "fcntl(F_GETFL) failed");
+            "setsockopt(SO_REUSEADDR) failed");
         goto Exit;
     }
 
-    Flags |= O_NONBLOCK;
+
+    //
+    // bind() to local port if we need to. This is not necessary if we call connect afterward and there is no ask for particular
+    // source address or port. connect() will resolve that together in single system call.
```
nibanks, line 776: nit:
```suggestion
    //
    // bind() to local port if we need to. This is not necessary if we call connect afterward and there is no ask for particular
    // source address or port. connect() will resolve that together in single system call.
    //
```

src/platform/datapath_kqueue.c
```diff
@@ -721,126 +724,208 @@ CxPlatSocketContextInitialize(
...
+        CxPlatCopyMemory(&MappedAddress, &Binding->LocalAddress, sizeof(MappedAddress));
+        if (MappedAddress.Ipv6.sin6_family == QUIC_ADDRESS_FAMILY_INET6) {
+            MappedAddress.Ipv6.sin6_family = AF_INET6;
+        }
+
+        if (ForceIpv4) {
+            MappedAddress.Ipv4.sin_family = AF_INET;
+            MappedAddress.Ipv4.sin_port = Binding->LocalAddress.Ipv4.sin_port;
+            // assume wildcard address for now!
+            CXPLAT_DBG_ASSERT(QuicAddrIsWildCard(&Binding->LocalAddress));
```
nibanks, line 787: Why do we have to assume wildcard?

src/platform/datapath_kqueue.c
```diff
@@ -721,126 +724,208 @@ CxPlatSocketContextInitialize(
...
+        }
+
+        Result =
+            bind(
+                SocketContext->SocketFd,
+                &MappedAddress.Ip,
+                ForceIpv4 ? sizeof(struct sockaddr_in) : sizeof(struct sockaddr_in6));
+
```
nibanks, line 795: nit: please remove extra newlines between function calls and error checks. This exists in at least one more place below.

src/platform/datapath_kqueue.c
```diff
@@ -721,126 +724,208 @@ CxPlatSocketContextInitialize(
...
+    //
+    // connect to RemoteAddress if provided.
+    //
+    if (RemoteAddress != NULL) {
+        CxPlatZeroMemory(&MappedAddress, sizeof(MappedAddress));
+        CxPlatConvertToMappedV6(RemoteAddress, &MappedAddress);
+
+        if (ForceIpv4)
+        {
```
nibanks, line 816:
```suggestion
        if (ForceIpv4) {
```

src/platform/datapath_kqueue.c
```diff
@@ -721,126 +724,208 @@ CxPlatSocketContextInitialize(
...
         goto Exit;
     }
 
     //
-    // Set socket option to receive ancillary data about the incoming packets.
+    // Set DON'T FRAG socket option.
     //
+    // IP_DONTFRAG is not supported on macOS.
+    // This may be re-visited on other kqueue systems like FreeBSD.
+    // IPv6 does not support fragmentation so no work there.
```
nibanks, line 918:
```suggestion
    // IPv6 does not support fragmentation so no work there.
    //
```

src/platform/datapath_kqueue.c
```diff
@@ -854,25 +939,6 @@ CxPlatSocketContextInitialize(
...
-    // if (Result == SOCKET_ERROR) {
-    //     Status = errno;
-    //     QuicTracgdfgfdeEvent(
-    //         DatapathErrorStatus,
-    //         "[data][%p] ERROR, %u, %s.",
-    //         Binding,
-    //         Status,
-    //         "setsockopt(IP_PKTINFO) failed");
-    //     goto Exit;
-    // }
```
nibanks, line 874: If this isn't necessary, let's just get rid of it? Same for below?

src/platform/datapath_kqueue.c
```diff
@@ -1222,11 +1169,12 @@ CxPlatSocketContextRecvComplete(
                 LocalAddr->Ipv4.sin_port = SocketContext->Binding->LocalAddress.Ipv6.sin6_port;
                 LocalAddr->Ipv6.sin6_scope_id = PktInfo->ipi_ifindex;
                 FoundLocalAddr = TRUE;
-            } else if (CMsg->cmsg_type == IP_TOS) {
+            } else if (CMsg->cmsg_type == IP_TOS || CMsg->cmsg_type == IP_RECVTOS) {
                 RecvPacket->TypeOfService = *(uint8_t *)CMSG_DATA(CMsg);
                 FoundTOS = TRUE;
             }
         }
+
```
nibanks, line 1177: nit: remove extra newline

src/platform/datapath_kqueue.c
```diff
@@ -721,126 +724,208 @@ CxPlatSocketContextInitialize(
...
+                Binding,
+                Status,
+                "setsockopt(IPV6_V6ONLY) failed");
+            goto Exit;
+        }
+    }
+
+    //
+    // The port may be shared across processors.
+    // Even if not, this is probably cheap.
```
nibanks, line 752: I don't remember if this was already here in the code, but if we aren't sharing it with ourselves (i.e. in the same process) then we shouldn't share it with others by default.

src/platform/datapath_kqueue.c
```diff
@@ -721,126 +724,208 @@ CxPlatSocketContextInitialize(
...
             Status,
-            "fcntl(F_GETFL) failed");
+            "setsockopt(SO_REUSEADDR) failed");
         goto Exit;
     }
 
-    Flags |= O_NONBLOCK;
+
+    //
+    // bind() to local port if we need to. This is not necessary if we call connect afterward and there is no ask for particular
```
nibanks, line 775: nit: let's try to keep an 80 char line limit if possible

ThadHouse:
Currently, tests are disabled on AZP. I'd like to see how they run. They can be enabled by uncommenting the configuration at 
https://github.com/microsoft/msquic/blob/main/.azure/azure-pipelines.ci.yml#L455

For the extra args, there are too many in there, so just replace that with 

```
extraArgs: -Filter -*.Tcp*
```

Everything there should run successfully.

wfurt:
> For the extra args, there are too many in there, so just replace that with
> 
> ```
> extraArgs: -Filter -*.Tcp*
> ```
> 
> Everything there should run successfully.

The problem I see seems to be related to threading. CxPlatEventSet() randomly fails with EINVAL on pthread_cond_wait() or we fail to wake up on event. I will try to chase that down this week. 
It may be interesting to see if CI has same problem. 

On good runs I get 
```
[----------] Global test environment tear-down
[==========] 508 tests from 3 test suites ran. (155353 ms total)
[  PASSED  ] 508 tests.
```

ThadHouse:
We've seen that failure randomly on linux too. I made #1268  to track the bug, and #1269 to make debugging easier, since often times the failure reason has been swallowed by the time the bugcheck hits.

src/platform/datapath_kqueue.c
wfurt: this is leftover from existing code (just moved). I don't think OS will distinguish between sharing with same process vs sharing with others. This simply allows second socket to bind with the same parameters. This may not be needed here as it seems like the purpose on Linux is load balancing on multiple cores.

src/platform/datapath_kqueue.c
wfurt: I did not see any test doing otherwise. So this reflects testing I did so far. 
If it is not wildcard we should probably assert that the address is either IPv4 or mapped IPv4.
I can fix that in next update.

wfurt:
good to know @ThadHouse. I'm yet to hit it on Linux but I did not do too many runs. It is somewhat easy to hit it on macOS for me so I'll investigate. It may or may not have same root cause with the hanging tests.

src/platform/datapath_kqueue.c
ThadHouse: I know on TCP this is used a lot more often so the port can be reused without waiting for the socket to timeout, even in normal processes. For UDP however, I don't think its an issue, and we can likely remove this. Let me check real quick.

ThadHouse: Ok. So for UDP its for multicast support, so we don't need it. So I don't think its needed here, or for client sockets on the linux side, but thats a separate issue.

ThadHouse:
I suspect they are related. I just merged #1269 into main, so if you merge that into this PR that should make debugging that a bit easier.

wfurt:
I did sync up up and I'll findings. I'm not sure if this PR is best way how to communicate but I saw another assert:
```
(lldb) bt
* thread #25, stop reason = signal SIGABRT
  * frame #0: 0x00007fff724c233a libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff7257ee60 libsystem_pthread.dylib`pthread_kill + 430
    frame #2: 0x00007fff72449808 libsystem_c.dylib`abort + 120
    frame #3: 0x000000010096f1c9 libmsquic.dylib`quic_bugcheck at platform_posix.c:62:5
    frame #4: 0x000000010094d6be libmsquic.dylib`QuicCryptoCustomCertValidationComplete(Crypto=<unavailable>, Result=<unavailable>) at crypto.c:1634:5
    frame #5: 0x000000010094a413 libmsquic.dylib`QuicConnParamSet(Connection=<unavailable>, Param=<unavailable>, BufferLength=1, Buffer=0x00007ffeefbff1cf) at connection.c:5790:9
    frame #6: 0x000000010094ac74 libmsquic.dylib`QuicConnProcessApiOperation(Connection=0x000000010406aa00, ApiCtx=0x00007ffeefbff0e8) at connection.c:6476:13
    frame #7: 0x000000010094b014 libmsquic.dylib`QuicConnDrainOperations(Connection=0x000000010406aa00) at connection.c:6594:13
    frame #8: 0x00000001009686a4 libmsquic.dylib`QuicWorkerProcessConnection(Worker=0x000000010402c778, Connection=0x000000010406aa00) at worker.c:488:9
    frame #9: 0x0000000100967b6d libmsquic.dylib`QuicWorkerThread(Context=0x000000010402c778) at worker.c:576:13
    frame #10: 0x00007fff7257f109 libsystem_pthread.dylib`_pthread_start + 148
    frame #11: 0x00007fff7257ab8b libsystem_pthread.dylib`thread_start + 15
```

I'm still trying to understand internal state processing but on the given connection `CertValidationPending` is not set.
```
(lldb) p Connection->State
(QUIC_CONNECTION_STATE) $2 = {
  Flags = 24723
   = {
    Allocated = '\x01'
    Initialized = '\x01'
    Started = '\0'
    Connected = '\0'
    ClosedLocally = '\x01'
    ClosedRemotely = '\0'
    AppClosed = '\0'
    HandleShutdown = '\x01'
    HandleClosed = '\0'
    Uninitialized = '\0'
    Freed = '\0'
    HeaderProtectionEnabled = '\0'
    Disable1RttEncrytion = '\0'
    ExternalOwner = '\x01'
    Registered = '\x01'
    GotFirstServerResponse = '\0'
    HandshakeUsedRetryPacket = '\0'
    HandshakeConfirmed = '\0'
    ListenerAccepted = '\0'
    LocalAddressSet = '\0'
    RemoteAddressSet = '\0'
    PeerTransportParameterValid = '\0'
    UpdateWorker = '\0'
    ShutdownCompleteTimedOut = '\0'
    SendShutdownCompleteNotif = '\0'
    ShareBinding = '\0'
    TestTransportParameterSet = '\0'
    UseRoundRobinStreamScheduling = '\0'
    ResumptionEnabled = '\0'
    AppCloseInProgress = '\0'
  }
}
``` 
I'm not sure if the `ClosedLocally` really matters as I would assume but the associate crypto is 
```
(lldb) p Connection->Crypto
(QUIC_CRYPTO) $3 = {
  Initialized = '\0'
  InRecovery = '\0'
  TlsDataPending = '\0'
  TlsCallPending = '\0'
  CertValidationPending = '\0'
  TLS = 0x0000000000000000
  TlsState = {
    HandshakeComplete = '\0'
    SessionResumed = '\0'
    EarlyDataState = CXPLAT_TLS_EARLY_DATA_UNKNOWN
    ReadKey = QUIC_PACKET_KEY_INITIAL
    WriteKey = QUIC_PACKET_KEY_INITIAL
    AlertCode = 0
    BufferLength = 0
    BufferAllocLength = 0
    BufferTotalLength = 0
    BufferOffsetHandshake = 0
    BufferOffset1Rtt = 0
    Buffer = 0x0000000000000000
    NegotiatedAlpn = 0x0000000000000000
    ReadKeys = {
      [0] = 0x0000000000000000
      [1] = 0x0000000000000000
      [2] = 0x0000000000000000
      [3] = 0x0000000000000000
      [4] = 0x0000000000000000
      [5] = 0x0000000000000000
    }
    WriteKeys = {
      [0] = 0x0000000000000000
      [1] = 0x0000000000000000
      [2] = 0x0000000000000000
      [3] = 0x0000000000000000
      [4] = 0x0000000000000000
      [5] = 0x0000000000000000
    }
  }
  ResultFlags =
  MaxSentLength = 0
  UnAckedOffset = 0
  NextSendOffset = 0
  RecoveryNextOffset = 0
  RecoveryEndOffset = 0
  SparseAckRanges = {
    SubRanges = 0x0000000000000000
    UsedLength = 0
    AllocLength = 0
    MaxAllocSize = 0
    PreAllocSubRanges = {
      [0] = (Low = 0, Count = 0)
      [1] = (Low = 0, Count = 0)
      [2] = (Low = 0, Count = 0)
      [3] = (Low = 0, Count = 0)
      [4] = (Low = 0, Count = 0)
      [5] = (Low = 0, Count = 0)
      [6] = (Low = 0, Count = 0)
      [7] = (Low = 0, Count = 0)
    }
  }
  RecvTotalConsumed = 0
  RecvEncryptLevelStartOffset = 0
  RecvBuffer = {
    CopyOnDrain = '\0'
    ExternalBufferReference = '\0'
    OldBuffer = 0x0000000000000000
    Buffer = 0x0000000000000000
    PreallocatedBuffer = 0x0000000000000000
    AllocBufferLength = 0
    VirtualBufferLength = 0
    BaseOffset = 0
    BufferStart = 0
    WrittenRanges = {
      SubRanges = 0x0000000000000000
      UsedLength = 0
      AllocLength = 0
      MaxAllocSize = 0
      PreAllocSubRanges = {
        [0] = (Low = 0, Count = 0)
        [1] = (Low = 0, Count = 0)
        [2] = (Low = 0, Count = 0)
        [3] = (Low = 0, Count = 0)
        [4] = (Low = 0, Count = 0)
        [5] = (Low = 0, Count = 0)
        [6] = (Low = 0, Count = 0)
        [7] = (Low = 0, Count = 0)
      }
    }
  }
  ResumptionTicket = 0x0000000000000000
  ResumptionTicketLength = 0
}
```

ThadHouse:
For that specific test, a failure to connect will result in that error. Would need to look deeper to see why it failed to connect.

wfurt:
ok. There still may be something wrong with networking. It was just surprising to see see assert.

wfurt:
it looks like memory corruption to me (unless I miss something).
This is from the hanging case. 

```
(lldb) p Oper.API_CALL.Context->Completed
(CXPLAT_EVENT *) $22 = 0x00007ffeefbfefb0
(lldb) p *((CXPLAT_EVENT_OBJECT*)Oper.API_CALL.Context->Completed)
(CXPLAT_EVENT_OBJECT) $24 = {
  Mutex = (__sig = 4422912000, __opaque = "\x90@�\a")
  Cond = (__sig = 0, __opaque = "\x01")
  Signaled = 'p'
  AutoReset = '�'
}
```

since `Signaled` is BOOL, I see no way how that should have values different than 0 or 1.

ThadHouse:
@wfurt Do you know which operation that is? Oper.Type and Oper.API_CALL.Context->Type.

wfurt:
```
(lldb) p Oper.Type
(QUIC_OPERATION_TYPE) $2 = QUIC_OPER_TYPE_API_CALL
(lldb) p Oper.API_CALL.Context->Type
(QUIC_API_TYPE) $3 = QUIC_API_TYPE_SET_PARAM
```
I added extra members to CXPLAT_EVENT_OBJECT and I set them only in CxPlatEventInitialize(). But I still get garbage. So it is either overwritten or freed IMHO. I'll keep digging. 
```
(CXPLAT_EVENT_OBJECT) $1 = {
  Sentinel0 = 47223600
  Mutex = (__sig = 47223120, __opaque = "")
  Cond = (__sig = 4295428440, __opaque = "o")
  Sentinel1 = '�'
  Signaled = '�'
  AutoReset = '�'
  Sentinel2 = '�'
}
```
I can also try to run it under valgrind on Linux. (unless you already did). Valgrind on macOS sucks and really works only on somewhat old versions.

ThadHouse:
Interesting. It might also be worth trying https://github.com/microsoft/msquic/pull/1273

I wonder if macOS is a bit more finicky about spurious wakeups.

wfurt:
ok, I will. I also noticed 
```
#if defined(CX_PLATFORM_LINUX)
    Result = pthread_condattr_setclock(&Attr, CLOCK_MONOTONIC);
    CXPLAT_FRE_ASSERT(Result == 0);
#endif // CX_PLATFORM_LINUX
```
but I did not yet get to checking if the time works correctly on macOS

wfurt:
#1273  looks very promising @ThadHouse. I did several runs without any threading issue.  I'll wrap the test runs in a loop and I'll let it burn for a while. I see occasional networking failure but that can be investigated separately. I'll work on PR feedback and perhaps CI test run.

ThadHouse:
Interesting. I wonder if mac is mor prone to spurious wakeup issues, as thats the only way I can envision that happening. But if it works, thats good to know.

wfurt:
It may. I can add counter there temporarily to see if there is correlation with spurious wakeups. I can also do simple experiment to detect use-after-free

src/platform/datapath_kqueue.c
wfurt: updated. While there is no test case, I added verification that if LocalAddress is not wildcard, it is IPv4 in some form.

src/platform/datapath_kqueue.c
wfurt: this was originally in the code. removed since all the tests are passing without it.

wfurt:
This should be ready for another pass. 
- updated style
- sync up with main to pick up #1273
- enable CI for macOS
- address feedback

.azure/azure-pipelines.ci.yml
```diff
@@ -452,14 +452,13 @@ stages:
...
-  #     tls: openssl
-  #     logProfile: None
-  #     extraArgs: -Filter -*DataECT0*:TlsTest.CertificateError:*.Tcp*:*DataPathTest.UdpData*:ParameterValidation.Validate*:*ParameterValidation/*:*Basic/WithFamilyArgs*:*AppData*:Misc.ServerDisconnect
+  - template: ./templates/run-bvt.yml
+     parameters:
+       image: macOS-10.15
+       platform: macos
+       tls: openssl
+       logProfile: None
+       extraArgs: -Filter -*.Tcp*
```
nibanks, line 461:
```suggestion
  - template: ./templates/run-bvt.yml
    parameters:
      image: macOS-10.15
      platform: macos
      tls: openssl
      logProfile: None
      extraArgs: -Filter -*.Tcp*
```
Looks like the spacing is off and AZP is complaining.

src/inc/quic_platform_posix.h
```diff
@@ -559,21 +559,24 @@ typedef struct CXPLAT_EVENT_OBJECT {
...
+    char Sentinel1;
+
     BOOLEAN Signaled;
 
     //
     // Denotes if the event object should be auto reset after it's signaled.
     //
 
     BOOLEAN AutoReset;
+    char Sentinel2;
```
nibanks, line 579: Can these be removed? I assume they were there for testing?

src/platform/datapath_kqueue.c
```diff
@@ -1944,6 +1875,7 @@ CxPlatSocketSendInternal(
 
     if (SentByteCount < 0) {
         if (errno == EAGAIN || errno == EWOULDBLOCK) {
+        printf("%s:%d: blocking ???? \n", __func__, __LINE__);
```
nibanks, line 1878: Need to remove these printfs

nibanks:
Looks like a lot MacOS test failures: https://dev.azure.com/ms/msquic/_build/results?buildId=141624&view=logs&j=e2236ac7-b409-5bb1-0b69-b4a92122e36a&t=0d0e0a7c-7680-52dc-f392-89c1db582d99

ThadHouse:
Some tests are likely going to need https://github.com/microsoft/msquic/pull/1254 to pass. But not that many failures.

src/inc/quic_platform_posix.h
wfurt: oops. yes. that should go.

wfurt:
I saw some similar failures locally. They seems to be transient and I'm trying to investigate. Generally I never seen them running just the specific test. So may be related to overall load, buffers, sockets or another race condition.

src/inc/msquic_posix.h
```diff
@@ -186,7 +186,11 @@ typedef union QUIC_ADDR {
 extern "C" {
 #endif
 
+#if defined(CX_PLATFORM_DARWIN)
+#define QUIC_LOCALHOST_FOR_AF(Af) ((Af == QUIC_ADDRESS_FAMILY_INET) ? "localhost" : "::1")
```
ThadHouse, line 190: Its `ip6-localhost` not valid on macOS? Interesting, but not super shocking.

wfurt: no. it seems like on linux it is in /etc/hosts but not macOS. This can be also distro specific.

wfurt:
```
[  FAILED  ] Basic/WithFamilyArgs.RebindPort/0, where GetParam() = v4
[  FAILED  ] Basic/WithFamilyArgs.RebindAddr/0, where GetParam() = v4
```
are still failing. From quick look it did not look like actual bind() failure. I'm still looking but I wanted to make checkpoint.

ThadHouse:
I realized we should probably enable interop testing too.

https://github.com/microsoft/msquic/blob/main/.azure/azure-pipelines.ci.yml#L560

But it looks like most of the tests are passing. Thats sweet.

src/inc/msquic_posix.h
ThadHouse: Since this is a public header, I wonder if we should change these on all posix platforms to use `127.0.0.1` for IPv4 and `::1` for IPv6. Since those actually are standardized. @nibanks opinion?

ThadHouse: Or at minimum, switch ipv6 to `::1` on both platforms.

wfurt:
yah. I still see some flakiness locally but at least there is some baseline. I can look into Interop tomorrow.

src/inc/msquic_posix.h
wfurt: Is it public? I could not find any reference to this besides tests. 'localhost' is pretty standard IMHO but I don't think it its guarantied to exist. In case of the test, you would get name resolution failure - and that can generally happen for any name IMHO. 
'localhost' can also resolve to ::1 on certain systems:
https://www.freedesktop.org/software/systemd/man/systemd-resolved.service.html

ThadHouse: The header resolves publicly, but its probably not too big of a deal. At least for IPv4 leaving localhost is fine, but I think ::1 would be better for IPv6 since that is more likely to work across distros. I don't want to get into a case where our tests don't work on certain distros, and I'd like to avoid the ifdef if possible too.

ThadHouse:
Just for another datapoint, I grabbed the latest builds and ran them on a mac running Big Sur, and all the tests passed. Even running in 1 go and not a test at a time.

src/inc/msquic_posix.h
nibanks: Don't change it to an IP address. The whole point it to check name resolution as well.

ThadHouse: So then what do we do? If ip6-localhost isn't in hosts on macOS, do we need to modify the tests to configure that? That's difficult to do without requiring admin.

nibanks: So neither `localhost` or `ip6-localhost` work on mac for v6?

ThadHouse: It looks like localhost does, but not ip6-localhost

nibanks: If `localhost` works for both v4 & v6, then just use that. That's how it works on Windows.

wfurt: I did more testing and `ping6 localhost` work on machine connected to corp but not my home local machine. So it seems like it depends on your environment.(e.g. name server configuration)  Also if you use 'localhost' there will be ambiguity about the address family as you may ask for QUIC_LOCALHOST_FOR_AF(QUIC_ADDRESS_FAMILY_INET6) but end up with IPv4 connection or vice versa. 

One more note that because this is string, it will still go through resolver logic to resolve it into binary address format.

wfurt: Let me know @nibanks how you want to proceed. We can fork this to separate issue as somewhat unrelated to the socket changes. 

I could pass all tests on my machine with 'loopback'. We may get lucky and that _may_ work for BVT as well. We just need to be aware of the fact that that is not generally available. (and the tests may not do what is intended) 

I could also roll-back this change and disable impacted tests for now.

nibanks: I don't care too much either way. Let's just get in something that works for now.

wfurt: ok. Using "::1" string passed BVT. I can try if "localhost" works for BVT. if not, I'll roll back to current state.

wfurt: macOS BVT passed with localhost so I'll leave as such.  I also check FreeBSD and ip6-localhost is not there either. So it really seems like Linux-ism more so than posix.

nibanks (approved):

ThadHouse (approved):

src/inc/msquic_posix.h
Anas198900: > Its `ip6-localhost` not valid on macOS? Interesting, but not super shocking.

PR #1261, opened 2021-02-20
prepare-machine: Detect default PS and warn dev to install the latest

obastemur:
`IsWindows` is not defined within the default PS.

### DIFF
--- a/scripts/prepare-machine.ps1
+++ b/scripts/prepare-machine.ps1
@@ -64,6 +64,13 @@ $ClogDownloadUrl = "https://github.com/microsoft/CLOG/releases/download/v$ClogVe
 
 $MessagesAtEnd = New-Object Collections.Generic.List[string]
 
+if ($PSVersionTable.PSVersion.Major -lt 7) {
+
+    Write-Error ("`nPowerShell v7.x is needed for this script to work. " +
+                 "Please visit https://github.com/microsoft/msquic/blob/main/docs/BUILD.md#powershell-usage")
+    exit
+}
+
 if ($InitSubmodules) {
 
     # Default TLS based on current platform.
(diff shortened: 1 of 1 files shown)


scripts/prepare-machine.ps1
```diff
@@ -64,6 +64,14 @@ $ClogDownloadUrl = "https://github.com/microsoft/CLOG/releases/download/v$ClogVe
 
 $MessagesAtEnd = New-Object Collections.Generic.List[string]
 
+# Detect PS by feature
+if (!(Test-Path Variable:\IsWindows)) {
```
ThadHouse, line 68: I think I would prefer an explicit version check rather then just checking for that variable. Otherwise pwsh 6 would work, which we don't do any testing on.

Just checking that $PSVersionTable.PSVersion.Major is >= 7 would probably be the best.

obastemur: > `Otherwise pwsh 6 would work, which we don't do any testing on.` 

Makes sense if it's not tested there. Updated.

scripts/prepare-machine.ps1
```diff
@@ -64,6 +64,13 @@ $ClogDownloadUrl = "https://github.com/microsoft/CLOG/releases/download/v$ClogVe
 
 $MessagesAtEnd = New-Object Collections.Generic.List[string]
 
+if ($PSVersionTable.PSVersion.Major -lt 7) {
+
+    Write-Error ("`nPowerShell v7.x is needed for this script to work. " +
+                 "Please visit https://github.com/microsoft/msquic/blob/main/docs/BUILD.md#powershell-usage")
+    exit
```
nibanks, line 71: The `exit` is unnecessary because of `$PSDefaultParameterValues['*:ErrorAction'] = 'Stop'` above.

nibanks:
```suggestion
```

nibanks (approved):

ThadHouse (approved):

PR #1262, opened 2021-02-20
Add Kernel mode test for client certificates.

anrossi:
Adds kernel mode test support for Schannel client certificate support.


src/test/MsQuicTests.h
```diff
@@ -13,7 +13,9 @@
 
 //#define QUIC_COMPARTMENT_TESTS 1
 
-extern QUIC_CREDENTIAL_CONFIG SelfSignedCredConfig;
+extern QUIC_CREDENTIAL_CONFIG ServerCredConfig;
+extern QUIC_CREDENTIAL_CONFIG ServerCredConfigClientAuth;
+extern QUIC_CREDENTIAL_CONFIG ClientCertCredConfig;
```
nibanks, line 18: I think we should still have some indication in the name that these can't be validated (which the `SelfSigned` did before).

Anas198900 (approved):

.azure/azure-pipelines.ci.yml
```diff
@@ -406,7 +406,7 @@ stages:
       platform: windows
       tls: schannel
       logProfile: Full.Light
-      extraArgs: -Kernel
+      extraArgs: -Kernel -Filter -Handshake/WithHandshakeArgs6.ConnectClientCertificate/*
```
nibanks, line 409: Try this (same for above):
```suggestion
      extraArgs: -Kernel -Filter -*Handshake/WithHandshakeArgs6.ConnectClientCertificate/*
```

nibanks (approved):

PR #1263, opened 2021-02-21
Fix memory leak in ResumptionTicketTest

anrossi:


src/core/unittest/TicketTest.cpp
```diff
@@ -103,6 +103,26 @@ struct TransportParamsScope {
...
 
+struct TicketScope {
+    uint8_t* p = nullptr;
+
+    ~TicketScope() {
+        reset();
+    }
+
+    uint8_t** operator&() {
+        reset();
```
nibanks, line 114: Why the reset here? I don't like this hidden side effect.

anrossi: I believe `wil::unique_ptr` uses the same contract, to prevent leaked memory. But I'll change it up.

ThadHouse: ComPtr uses a GetAddressOf() function to give this behavior and then just an operator uint8_t* to get the value without reset if necessary. That'd probably be what I'd do.

anrossi: Confirmed this is the way `wistd::unique_any_t` works, which is a base-class for `wistd::unique_ptr`: https://github.com/microsoft/wil/blob/ac455d2c2e3018dc37aafcdf0ffcc17adcd5739b/include/wil/resource.h#L319

src/core/unittest/TicketTest.cpp
```diff
@@ -103,6 +103,28 @@ struct TransportParamsScope {
...
 };
 
+struct TicketScope {
+    uint8_t* p = nullptr;
+
+    ~TicketScope() {
+        reset();
+    }
+
+    uint8_t** operator&() = delete;
```
ThadHouse, line 113: This operator isn't defined anyway, so there is no reason to delete it. the default operator& will return a TicketScope*.

anrossi: Why not explicitly discourage the usage of `(uint8_t**) operator&()`? Won't that be covered by this?

nibanks (approved):

PR #1264, opened 2021-02-21
Build requires cmake 3.19

obastemur:
On Windows (although the issue is not platform specific), when the `openssl` Tls option is chosen. CMakeLists takes the path that one of the `PROPERTY FOLDER` lines are touched. CMAKE versions under 3.19 doesn't allow FOLDER property to be used.

Repro Steps : Try `scripts/build.ps1 -Tls openssl` with cmake 3.18.xx


ThadHouse:
If we're going to do this, we should also bump the minimum cmake version in CMakeLists.txt. But I'm a bit confused on why only one in openssl is breaking. I don't see anything new in the 3.19 release notes, and we use PROPERTY FOLDER in many places outside of openssl as well. 

I would probably prefer to keep 3.16 as a minimum, and then conditionally checking for 3.19 or greater when setting the specific folder property that is broken.

ThadHouse:
I realized this was a mistake and that folder property shouldn't be there, so replacing this with #1292, which just removes the error case entirely.

PR #1274, opened 2021-02-23
Loss Detection Immediately

nibanks:
See https://github.com/quicwg/base-drafts/issues/4830 for more details about the issue. This ensures that if we get unblocked from amplification protection logic, we trigger loss detection immediately (if necessary) instead of queuing it up, so that we can send lost packets first (and not application payload).


anrossi (approved):

anrossi:
Can you write a test that gets into this state to ensure that loss detection runs?

nibanks:
> Can you write a test that gets into this state to ensure that loss detection runs?

Since it requires packet loss, writing a test to hit it consistently is hard. Random loss during the handshake, a long with application data would work eventually.

PR #1276, opened 2021-02-24
Fix CRYPTO fragmentation handshake failure on client

### LINKED ISSUE #1270: Packet Loss and/or CRYPTO Fragmentation Handshake Failure on Client
### Describe the bug

One our random loss handshake tests discovered a scenario where certain scenarios and packet loss patterns resulted in a case where the client ended up failing the handshake because of `SEC_E_EXT_BUFFER_TOO_SMALL` getting returned from Schannel. This comes down to the QUIC code making the (wrong) assumption that the QUIC Transport Parameters would be delivered back to us as they are received. Instead, it seems Schannel waits for the entire flight of HANDSHAKE key encrypted data to be received and processed before indicating the TP back to us. This then results in some scenarios were the last message in that flight is delivered to Schannel but is not big enough to contain the output data from Schannel.

We need to make sure we always provide a buffer upfront to schannel that is big enough for the received TP.

anrossi:
Schannel client uses the input buffer to receive the transport parameters from the server. In the case of heavy fragmentation, the buffer might not be large enough to hold the transport parameters.  This change uses the information from Schannel to allocate sufficient buffer space to hold the transport parameters and calls Schannel again.
Fixes #1270


src/platform/unittest/TlsTest.cpp
```diff
@@ -974,6 +974,14 @@ TEST_F(TlsTest, HandshakeFragmented)
     DoHandshake(ServerContext, ClientContext, 200);
 }
 
+TEST_F(TlsTest, HandshakeVeryFragmented)
+{
+    TlsContext ServerContext, ClientContext;
+    ServerContext.InitializeServer(ServerSecConfig, false, 256);
+    ClientContext.InitializeClient(ClientSecConfigNoCertValidation);
+    DoHandshake(ServerContext, ClientContext, 1);
```
nibanks, line 982: So this didn't repro originally? Can you try a much bigger TP size (> 1500) on both client and server? My only other guess would be that the difference is related to how QUIC core code buffers until it has a complete message before it calls. We might need to duplicate that logic here to test it.

nibanks (approved):

PR #1280, opened 2021-02-25
Implement UDP send buffering support in Linux

ThadHouse:
By using sendmmsg, we can send multiple buffers at once. GSO can be added later to improve this even more.

### DIFF
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -28,10 +28,7 @@ Environment:
 CXPLAT_STATIC_ASSERT((SIZEOF_STRUCT_MEMBER(QUIC_BUFFER, Length) <= sizeof(size_t)), "(sizeof(QUIC_BUFFER.Length) == sizeof(size_t) must be TRUE.");
 CXPLAT_STATIC_ASSERT((SIZEOF_STRUCT_MEMBER(QUIC_BUFFER, Buffer) == sizeof(void*)), "(sizeof(QUIC_BUFFER.Buffer) == sizeof(void*) must be TRUE.");
 
-//
-// TODO: Support batching.
-//
-#define CXPLAT_MAX_BATCH_SEND 1
+#define CXPLAT_MAX_BATCH_SEND 7
 
 //
 // A receive block to receive a UDP packet over the sockets.
@@ -2076,6 +2073,7 @@ CxPlatSocketSendInternal(
     struct in6_pktinfo *PktInfo6 = NULL;
     BOOLEAN SendPending = FALSE;
     uint32_t ProcNumber;
+    size_t SendCount;
 
     static_assert(CMSG_SPACE(sizeof(struct in6_pktinfo)) >= CMSG_SPACE(sizeof(struct in_pktinfo)), "sizeof(struct in6_pktinfo) >= sizeof(struct in_pktinfo) failed");
     char ControlBuffer[CMSG_SPACE(sizeof(struct in6_pktinfo)) + CMSG_SPACE(sizeof(int))] = {0};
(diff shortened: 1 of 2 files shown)


src/platform/datapath_epoll.c
```diff
@@ -2112,42 +2110,52 @@ CxPlatSocketSendInternal(
...
-            CMsg->cmsg_level = IPPROTO_IPV6;
-            CMsg->cmsg_type = IPV6_PKTINFO;
-            CMsg->cmsg_len = CMSG_LEN(sizeof(struct in6_pktinfo));
-            PktInfo6 = (struct in6_pktinfo*) CMSG_DATA(CMsg);
-            PktInfo6->ipi6_ifindex = LocalAddress->Ipv6.sin6_scope_id;
-            PktInfo6->ipi6_addr = LocalAddress->Ipv6.sin6_addr;
+    struct mmsghdr Mhdrs[CXPLAT_MAX_BATCH_SEND];
+
+
+    for (SendCount = 0; SendCount < SendData->BufferCount; SendCount++) {
```
nibanks, line 2116:
```suggestion
    struct mmsghdr Mhdrs[CXPLAT_MAX_BATCH_SEND];
    for (SendCount = 0; SendCount < SendData->BufferCount; SendCount++) {
```

src/platform/datapath_epoll.c
```diff
@@ -2171,7 +2179,7 @@ CxPlatSocketSendInternal(
         }
     }
 
-    SentByteCount = sendmsg(SocketContext->SocketFd, &Mhdr, 0);
+    SentByteCount = sendmmsg(SocketContext->SocketFd, Mhdrs, (unsigned int)SendCount, 0);
```
nibanks, line 2182: So `sendmmsg` takes individual message headers for each packet, and not just one for all?

ThadHouse: Yes. Sendmmsg and recvmmsg are just fast wrappers that can do multiple sendmsg or recvmsg in 1 system call. Once in the kernel, each message is treated completely separately.

src/platform/datapath_epoll.c
```diff
@@ -2112,127 +2136,128 @@ CxPlatSocketSendInternal(
...
-            struct epoll_event SockFdEpEvt = {
-                .events = EPOLLIN | EPOLLOUT | EPOLLET,
-                .data = {
-                    .ptr = &SocketContext->EventContexts[QUIC_SOCK_EVENT_SOCKET]
+        CXPLAT_FRE_ASSERT(SuccessfullySentMessages != 0);
+
+        if (SuccessfullySentMessages < 0) {
+            if (errno == EAGAIN || errno == EWOULDBLOCK) {
+                if (!IsPendedSend) {
+                    // TODO log sends as pending so byte counts line up.
```
nibanks, line 2200: I was thinking about this, and I feel we should log sends immediately as they are passed down from the caller. Abstract away if they're pending or not. Just log it and be done with it.

src/platform/datapath_epoll.c
```diff
@@ -2069,25 +2071,47 @@ CxPlatSocketSendInternal(
...
+                RemoteAddress);
+            SendPending = TRUE;
+        }
+        CxPlatLockRelease(&SocketContext->PendingSendContextLock);
+        if (SendPending) {
+            Status = QUIC_STATUS_PENDING;
+            goto Exit;
+        }
+    }
+
```
nibanks, line 2109: nit: remove extra newline

src/platform/datapath_epoll.c
```diff
@@ -2069,25 +2071,47 @@ CxPlatSocketSendInternal(
...
-    ssize_t SentByteCount = 0;
     QUIC_ADDR MappedRemoteAddress = {0};
     struct cmsghdr *CMsg = NULL;
     struct in_pktinfo *PktInfo = NULL;
     struct in6_pktinfo *PktInfo6 = NULL;
     BOOLEAN SendPending = FALSE;
     uint32_t ProcNumber;
+    size_t TotalMessagesCount;
+
+    CXPLAT_DBG_ASSERT(Socket != NULL && RemoteAddress != NULL && SendData != NULL);
```
nibanks, line 2082: Worth asserting that `SendData->SentMessagesCount` is <= than max?

src/platform/datapath_epoll.c
ThadHouse: We do. The issue with pending is since this path is sent again, you'll get the same log multiple times, so your sends won't line up. This todo was to add another log that a send was pended, so everything could be correlated.

I could also just make pending sends not log when going through the top of this function. That might be better.

src/platform/datapath_epoll.c
nibanks: Actually, now that I think of it, does it make sense for this to be called with Count == Max?

src/platform/datapath_epoll.c
```diff
@@ -2080,13 +2080,35 @@ CxPlatSocketSendInternal(
...
+    }
+
+    if (!IsPendedSend) {
+        QuicTraceEvent(
+            DatapathSend,
+            "[data][%p] Send %u bytes in %hhu buffers (segment=%hu) Dst=%!ADDR!, Src=%!ADDR!",
+            Socket,
+            TotalSize,
+            SendData->BufferCount,
+            SendData->Buffers[SendData->SentMessagesCount].Length,
```
nibanks, line 2107: Is it guaranteed that Count == 0 when IsPendedSend is FALSE? Worth another DBG assert?

src/platform/datapath_epoll.c
nibanks: Yeah, that's what I was getting at.

src/platform/datapath_epoll.c
ThadHouse: No, this should always be called with < max.

src/platform/datapath_epoll.c
ThadHouse: Yeah I can add that. The only way for a non pended send to get > 0 here is if the core code sets that somehow, but worth checking.

src/platform/datapath_epoll.c
```diff
@@ -2080,7 +2080,10 @@ CxPlatSocketSendInternal(
     size_t TotalMessagesCount;
 
     CXPLAT_DBG_ASSERT(Socket != NULL && RemoteAddress != NULL && SendData != NULL);
-    CXPLAT_DBG_ASSERT(SendData->SentMessagesCount <= CXPLAT_MAX_BATCH_SEND);
+    CXPLAT_DBG_ASSERT(SendData->SentMessagesCount < CXPLAT_MAX_BATCH_SEND);
+    if (!IsPendedSend) {
+        CXPLAT_DBG_ASSERT(SendData->SentMessagesCount == 0);
+    }
```
nibanks, line 2086:
```suggestion
    CXPLAT_DBG_ASSERT(IsPendedSend || SendData->SentMessagesCount == 0);
```

nibanks (approved):

PR #1281, opened 2021-02-25
Use no wait flag when cancelling lttng

### LINKED ISSUE #679: Improve Linux Test Runtime with Logs
Recently we fully enabled LTTng based logs on Linux (thanks to CLOG). This seems to have caused the runtimes to go from 7-8 minutes all the way up to 23-27 minutes. We need to investigate where the overhead is coming from. Is it starting/stopping/canceling the logging? Is it from runtime log collection?

ThadHouse:
By default, destroy will ensure all trace data has been finalized before returning. When cancelling, we don't need this behavior, so pass the no wait flag, which massively reduces the amount of time needed for logging.

Closes #679

### DIFF
--- a/scripts/log.ps1
+++ b/scripts/log.ps1
@@ -141,7 +141,7 @@ function Log-Cancel {
         if (!(Test-Path $TempDir)) {
             Write-Debug "LTTng session ($InstanceName) not currently running"
         } else {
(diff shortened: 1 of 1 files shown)


ThadHouse:
Master builds take about 30 minutes, and this took about 20 minutes. So this is a 33% time save.

nibanks (approved):

PR #1288, opened 2021-02-26
Fix a Few Casting Errors

### LINKED ISSUE #1287: Posix headers define incorrect return values for atomic operations
InterlockedCompareExchange64 in quic_platform_posix.h should be returning uint64_t, not short. There are a few other duplicates of this bug there as well.

nibanks:
Fixes #1287.

### DIFF
--- a/src/core/congestion_control.c
+++ b/src/core/congestion_control.c
@@ -434,7 +434,7 @@ QuicCongestionControlOnDataAcknowledged(
         // DeltaT < ~2.5M (about 30min).
         //
 
-        int64_t DeltaT = TimeInCongAvoid - Cc->KCubic + US_TO_MS(SmoothedRtt);
+        int64_t DeltaT = TimeInCongAvoid - (uint64_t)Cc->KCubic + (uint64_t)US_TO_MS(SmoothedRtt);
 
         int64_t CubicWindow =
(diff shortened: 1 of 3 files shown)


src/core/congestion_control.c
```diff
@@ -434,7 +434,7 @@ QuicCongestionControlOnDataAcknowledged(
         // DeltaT < ~2.5M (about 30min).
         //
 
-        int64_t DeltaT = TimeInCongAvoid - Cc->KCubic + US_TO_MS(SmoothedRtt);
+        int64_t DeltaT = TimeInCongAvoid - (uint64_t)Cc->KCubic + (uint64_t)US_TO_MS(SmoothedRtt);
```
nibanks, line 437: Actually, I'm not sure this should ever be negative....

nibanks: Supposedly CubicWindow below could be negative.

nibanks: @maolson-msft would you mind commenting on this logic? MacOS is complaining about integer overflow. I think just casting all these to `int64_t` before the arithmetic would suffice. Thoughts?

ThadHouse (approved):
I'm good with at least the CompareExchange changes. Definitely would like another opinion from someone more knowledgeable in CC for the CC change.

maolson-msft (approved):

PR #1289, opened 2021-02-26
Start adding support for arm64 builds on macOS

ThadHouse:
Only stub builds currently, as theres a bit of a configuration issue with openssl on the azure machines I still need to work out. But I don't want these to get stale, so PRing until I have a chance to look.


scripts/build.ps1
```diff
@@ -95,8 +95,8 @@ param (
     [string]$Config = "Debug",
 
     [Parameter(Mandatory = $false)]
-    [ValidateSet("x86", "x64", "arm", "arm64")]
-    [string]$Arch = "x64",
+    [ValidateSet("x86", "x64", "arm", "arm64", "current")]
+    [string]$Arch = "current",
```
nibanks, line 99: I'd prefer to not add support for this feature. Why is it necessary?

ThadHouse: Because we want to be able to support natively building on arm64 platforms. If we hardcode x86_64, that won't "Just work", and will require explicit configuration.

nibanks: I don't understand why passing `arm64` in for that doesn't work.

ThadHouse: It does, but what happens when you're building on an arm64 mac. You won't get binaries that run on your current platform. Users shouldn't have to pass something in to get a build for their current platform.

nibanks: For now, I prefer to continue with the current model. We can discuss changing the default behavior of build to "current" separately from this PR.

ThadHouse: No, this is the PR to do it in, because this is actually where the problem is going to start. We can't just punt things off, that is just saying it will never get done.

ThadHouse: Let me see if that works with the validate set.

ThadHouse: Ah that does work. I didn't realized validate set worked with an empty string, but it does, since we do it with Tls. I like empty better.

nibanks (approved):

PR #1299, opened 2021-02-27
Inline allocate event objects on posix

ThadHouse:
Previously, each event required an allocation. This was because the API for events required an opaque object. Use a macro to work around this limitation, and inline allocate the events.


src/platform/platform_posix.c
```diff
@@ -337,101 +337,83 @@ CxPlatEventInitialize(
     _In_ BOOLEAN InitialState
     )
 {
-    CXPLAT_EVENT_OBJECT* EventObj = NULL;
     pthread_condattr_t Attr = {0};
     int Result;
 
-    EventObj = CXPLAT_ALLOC_NONPAGED(sizeof(CXPLAT_EVENT_OBJECT), QUIC_POOL_EVENT);
+    Event->AutoReset = !ManualReset;
```
nibanks, line 343: Any reason not to inline all of this in the header (function or macro, I don't care which)?

ThadHouse: Would we inline all 5 functions, or just the 3 simple ones? Theres no reason not to inline these, especially without the allocation anymore.

nibanks: Everything

nibanks (approved):

nibanks:
@thadhouse any idea why Linux RPS perf seemed to tank?
https://dev.azure.com/ms/msquic/_build/results?buildId=143182&view=logs&j=eeba969b-e7b9-55a4-4ca4-7a01a11088a4&t=27694ef8-6adb-5d5e-460f-1f02287fe195

ThadHouse:
That doesn't make any sense. I'll try rerunning that to see if it's consistently bad. It wasnt that bad without inlining everything. And I don't see how this would have tanked RPS.

ThadHouse:
A rerun is better, only about 10% worse and not 60%, but I'm going to revert making all the functions inline and see what that does. Because I remember the original change not changing anything.

ThadHouse:
Ok, this is being really wonky. I'll do some investigation tomorrow. I wouldn't expect a PR like this to affect perf that much at all. I need to see if this is specific to this PR, or an infra issue.

Issue #1300, opened 2021-02-27
Look into number of small allocations


ThadHouse:
I played around with mimalloc, which can give stats on what type of allocations occurred. For a spinquic linux run, we got the following results.

```
heap stats:     peak      total      freed    current       unit      count  
normal   1:     1.1 mb    13.6 mb    13.6 mb       0 b        8 b      1.7 m   ok
normal   4:    23.3 mb   121.8 mb   121.8 mb       0 b       32 b      3.9 m   ok
normal   6:    13.9 mb   108.3 mb   108.3 mb       0 b       48 b      2.3 m   ok
normal   8:    54.0 mb   150.2 mb   150.2 mb       0 b       64 b      2.4 m   ok
normal   9:    35.2 mb   155.5 mb   155.5 mb       0 b       80 b      2.0 m   ok
normal  10:     1.0 mb    16.3 mb    16.3 mb       0 b       96 b    179.0 k   ok
normal  11:     5.8 mb     1.2 gb     1.2 gb       0 b      112 b     11.7 m   ok
normal  12:   836.0 kb    91.5 mb    91.5 mb       0 b      128 b    750.3 k   ok
normal  13:    83.4 kb    83.4 kb    83.4 kb       0 b      160 b      532     ok
normal  14:    57.7 kb   469.2 kb   469.2 kb       0 b      192 b      2.4 k   ok
normal  15:    14.4 kb    35.3 kb    35.3 kb       0 b      224 b      161     ok
normal  16:    13.5 kb    18.3 kb    18.3 kb       0 b      256 b       73     ok
normal  17:    26.9 kb   728.1 kb   728.1 kb       0 b      320 b      2.3 k   ok
normal  18:    28.6 kb    28.6 kb    28.6 kb       0 b      384 b       76     ok
normal  19:   602.3 mb   781.3 mb   781.3 mb       0 b      448 b      1.8 m   ok
normal  20:    36.1 kb   996.3 kb   996.3 kb       0 b      512 b      1.9 k   ok
normal  21:    43.9 kb    80.9 kb    80.9 kb       0 b      640 b      129     ok
normal  22:     2.5 mb     2.5 mb     2.5 mb       0 b      768 b      3.4 k   ok
normal  23:     896 b     17.5 kb    17.5 kb       0 b      896 b       20     ok
normal  24:    68.2 kb     1.8 mb     1.8 mb       0 b      1.0 kb     1.9 k   ok
[...]

PR #1303, opened 2021-03-01
Small buffer optimize tls negotiated ALPN

ThadHouse:
The TLS negotiated ALPN is often very small in size (probably h3), however this was always allocating, and this allocation was always very small and stays live for a long time. Add a small buffer optimization to remove this allocation in the normal case.

Part of #1300


src/inc/quic_tls.h
```diff
@@ -30,6 +30,11 @@ typedef struct CXPLAT_TLS_SECRETS CXPLAT_TLS_SECRETS;
 #define TLS_EXTENSION_TYPE_QUIC_TRANSPORT_PARAMETERS_DRAFT          0xffa5  // Host Byte Order
 #define TLS_EXTENSION_TYPE_QUIC_TRANSPORT_PARAMETERS                0x0039  // Host Byte Order
 
+//
+// The small buffer size for an ALPN to avoid allocations in the general case.
+//
+#define TLS_SMALL_ALPN_BUFFER_SIZE  16
```
ThadHouse, line 36: Should this be 8 or 16 bytes? 8 will still fit `h3` and `h3-29`. Do we imagine any ALPN's longer the 7 bytes to happen in the real world?

ThadHouse:
This PR was mostly created to investigate the perf impact. It might not be something we take.

src/inc/quic_tls.h
```diff
@@ -291,6 +296,8 @@ typedef struct CXPLAT_TLS_PROCESS_STATE {
     //
     uint8_t* Buffer;
 
+    uint8_t SmallAlpnBuffer[TLS_SMALL_ALPN_BUFFER_SIZE];
```
nibanks, line 299: Please put a comment.

src/inc/quic_tls.h
nibanks: I'm fine with 16.

nibanks (approved):
LGTM. Just a few comments.

ThadHouse:
I realized I should probably add a test for this, since no tests will actually hit the long alpn case. I'll do that.

src/test/lib/HandshakeTest.cpp
```diff
@@ -2075,3 +2075,101 @@ QuicTestConnectClientCertificate(
...
+
+void
+QuicTestInvalidAlpnLengths(
+    void
+    )
+{
+    int Lengths[] = { 0, QUIC_MAX_ALPN_LENGTH + 1 };
+    char AlpnBuffer[QUIC_MAX_ALPN_LENGTH + 3]; // + 3 so it can always be 0 terminated
+    for (int Len = 0; Len < (int)ARRAYSIZE(Lengths); Len++)
+    {
```
nibanks, line 2087: nit: formatting (below as well)
```suggestion
    for (int Len = 0; Len < (int)ARRAYSIZE(Lengths); Len++) {
```

Issue #929, opened 2020-10-19
Dynamically Scale Immediate ACK Threshold Up


nibanks:
### Describe the feature you'd like supported

Right now MsQuic queues an immediate ACK for every (hard-coded) 2 received packets. In high throughput scenarios where we don't end up batching much on the receive path, this can result in a lot of ACK packets being generated, which then ends up causing an excess CPU load on the sender, further reducing overall throughput.

### Proposed solution

We should have a more dynamic threshold than the currently hard-coded 2.

nibanks:
https://tools.ietf.org/html/draft-iyengar-quic-delayed-ack-01 is generally seen as the long term solution here.

PR #1305, opened 2021-03-02
Initial ACK Frequency Support

nibanks:
First part of #929. Implements https://tools.ietf.org/html/draft-iyengar-quic-delayed-ack. Supports receiving the new ACK_FREQUENCY frame but never sends it currently.

### DIFF
--- a/src/core/ack_tracker.c
+++ b/src/core/ack_tracker.c
@@ -31,7 +31,7 @@ Abstract:
     from the tracker in response to the original ACK_FRAME being
     acknowledged by the peer. Since we constantly send ACK_FRAMEs with the
     current state, most of the time having a lot of duplicate information in
-    them, we assume the data eventually gets there inone form or another. Worst
+    them, we assume the data eventually gets there in one form or another. Worst
     case, the peer has to do an additional retransmission, in an already lossy
     environment.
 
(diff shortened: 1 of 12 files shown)


nibanks:
Anyone have any idea why I'm getting this error on my PR?
```
/home/vsts/work/1/s/src/core/crypto_tls.c: In function ‘QuicCryptoTlsDecodeTransportParameters’:
/home/vsts/work/1/s/src/core/crypto_tls.c:1212:22: error: ‘ParamLength’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
 1212 |         QUIC_VAR_INT ParamLength;
      |                      ^~~~~~~~~~~
cc1: all warnings being treated as errors
```
@maximus- you have any idea? It's definitely a false positive AFAICT, but I don't know why it came up on this PR.

ThadHouse:
Thats actually not a false positive, if `BufferLength < sizeof(uint8_t) + *Offset` is true, that variable will be read uninitialized by the else if.

nibanks:
> Thats actually not a false positive, if `BufferLength < sizeof(uint8_t) + *Offset` is true, that variable will be read uninitialized by the else if.

I don't follow. What `BufferLength`? If `QuicVarIntDecode` return `FALSE` we bail.

ThadHouse:
Yeah I misread that. Hmm.

Maximus-:
Not sure exactly what "may be used uninitialized in this function" entails, but it is true that there is a possible path through that function where it does not get initialized (although in that same path, it also does not get used).

ThadHouse:
The instance of QUIC_VAR_INT right above this we do initialize to 0 even though it also can never be used uninitialized, so we can probably just do the same thing here. I want to avoid suppressing that warning. And the cost of zeroing a register is basically free.

nibanks:
> The instance of QUIC_VAR_INT right above this we do initialize to 0 even though it also can never be used uninitialized, so we can probably just do the same thing here. I want to avoid suppressing that warning. And the cost of zeroing a register is basically free.

Perhaps, but I'd really like to know why this error is happening **now**. Any ideas? What could I have changed that caused it?

ThadHouse (approved):

PR #1306, opened 2021-03-03
Enable SAL for platform component on Windows User Mode

### LINKED ISSUE #1242: SAL annotation mismatch in CxPlatTlsParamGet
The prototype is treated as an _In_ parameter, but schannel treats it as an _Inout_. These should match.

ThadHouse:
Fixes some SAL mismatches, and a few code bugs.

Closes #1242

### DIFF
--- a/src/inc/quic_platform_winuser.h
+++ b/src/inc/quic_platform_winuser.h
@@ -234,6 +234,20 @@ CxPlatLogAssert(
 #define CXPLAT_PASSIVE_CODE() CXPLAT_DBG_ASSERT(CXPLAT_IRQL() == PASSIVE_LEVEL)
 
 //
+// Wrapper functions
+//
+
+//
+// CloseHandle has an incorrect SAL annotation, so call through a wrapper.
+//
+_IRQL_requires_max_(PASSIVE_LEVEL)
+inline
+void
+CxPlatCloseHandle(_Pre_notnull_ HANDLE Handle) {
+    CloseHandle(Handle);
+}
+
+//
 // Allocation/Memory Interfaces
 //
 
@@ -522,7 +536,7 @@ typedef HANDLE CXPLAT_EVENT;
 #define CxPlatEventInitialize(Event, ManualReset, InitialState)     \
     *(Event) = CreateEvent(NULL, ManualReset, InitialState, NULL);  \
     CXPLAT_DBG_ASSERT(*Event != NULL)
-#define CxPlatEventUninitialize(Event) CloseHandle(Event)
+#define CxPlatEventUninitialize(Event) CxPlatCloseHandle(Event)
 #define CxPlatEventSet(Event) SetEvent(Event)
 #define CxPlatEventReset(Event) ResetEvent(Event)
 #define CxPlatEventWaitForever(Event) WaitForSingleObject(Event, INFINITE)
@@ -879,7 +893,7 @@ CxPlatThreadCreate(
     }
     return QUIC_STATUS_SUCCESS;
 }
-#define CxPlatThreadDelete(Thread) CloseHandle(*(Thread))
+#define CxPlatThreadDelete(Thread) CxPlatCloseHandle(*(Thread))
 #define CxPlatThreadWait(Thread) WaitForSingleObject(*(Thread), INFINITE)
 typedef uint32_t CXPLAT_THREAD_ID;
 #define CxPlatCurThreadID() GetCurrentThreadId()
@@ -911,7 +925,7 @@ typedef struct CXPLAT_RUNDOWN_REF {
     (Rundown)->RundownComplete = CreateEvent(NULL, FALSE, FALSE, NULL); \
     CXPLAT_DBG_ASSERT((Rundown)->RundownComplete != NULL)
 #define CxPlatRundownReInitialize(Rundown) (Rundown)->RefCount = 1
-#define CxPlatRundownUninitialize(Rundown) CloseHandle((Rundown)->RundownComplete)
+#define CxPlatRundownUninitialize(Rundown) CxPlatCloseHandle((Rundown)->RundownComplete)
 #define CxPlatRundownAcquire(Rundown) CxPlatRefIncrementNonZero(&(Rundown)->RefCount, 1)
 #define CxPlatRundownRelease(Rundown) \
     if (CxPlatRefDecrement(&(Rundown)->RefCount)) { \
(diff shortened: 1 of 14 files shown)


src/platform/storage_winuser.c
```diff
@@ -194,7 +194,7 @@ _IRQL_requires_max_(PASSIVE_LEVEL)
 QUIC_STATUS
 CxPlatStorageReadValue(
     _In_ CXPLAT_STORAGE* Storage,
-    _In_opt_z_ const char * Name,
+    _In_z_ const char * Name,
```
nibanks, line 197: `Name` is optional

nibanks: From MSDN: https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regqueryvalueexa

lpValueName

The name of the registry value.

If lpValueName is NULL or an empty string, "", the function retrieves the type and data for the key's unnamed or default value, if any.

src/platform/tls_openssl.c
```diff
@@ -520,10 +520,12 @@ CxPlatTlsSendAlertCallback(
     return 1;
 }
 
+_Success_(return == SSL_CLIENT_HELLO_SUCCESS)
 int
 CxPlatTlsClientHelloCallback(
     _In_ SSL *Ssl,
-    _Out_opt_ int *Alert,
+    _When_(return == SSL_CLIENT_HELLO_ERROR, _Out_)
+    int *Alert,
```
nibanks, line 528:
```suggestion
        int *Alert,
```

src/platform/tls_openssl.c
```diff
@@ -1068,11 +1075,13 @@ CxPlatTlsUninitialize(
...
 }
 
+_IRQL_requires_max_(PASSIVE_LEVEL)
 CXPLAT_TLS_RESULT_FLAGS
 CxPlatTlsProcessData(
     _In_ CXPLAT_TLS* TlsContext,
     _In_ CXPLAT_TLS_DATA_TYPE DataType,
-    _In_reads_bytes_(*BufferLength) const uint8_t* Buffer,
+    _In_reads_bytes_(*BufferLength)
+    const uint8_t* Buffer,
```
nibanks, line 1084:
```suggestion
        const uint8_t* Buffer,
```

src/platform/tls_openssl.c
```diff
@@ -1906,13 +1920,14 @@ QuicPacketKeyUpdate(
...
 
+_IRQL_requires_max_(DISPATCH_LEVEL)
 QUIC_STATUS
 CxPlatKeyCreate(
     _In_ CXPLAT_AEAD_TYPE AeadType,
     _When_(AeadType == CXPLAT_AEAD_AES_128_GCM, _In_reads_(16))
     _When_(AeadType == CXPLAT_AEAD_AES_256_GCM, _In_reads_(32))
     _When_(AeadType == CXPLAT_AEAD_CHACHA20_POLY1305, _In_reads_(32))
-        const uint8_t* const RawKey,
+    const uint8_t* const RawKey,
```
nibanks, line 1930: Undo. This is the style we use in this code.

nibanks: Same for all changes below.

ThadHouse: These were all just the editor screwing up the copy paste. I'll fix.

src/platform/tls_schannel.c
```diff
@@ -1963,7 +1963,7 @@ CxPlatTlsWriteDataToSchannel(
                 SchannelKeyReady,
                 TlsContext->Connection,
                 "Key Ready Type, %u [%hu to %hu]",
-                TrafficSecret->TrafficSecretType,
+                (unsigned)TrafficSecret->TrafficSecretType,
```
nibanks, line 1966:
```suggestion
                (uint32_t)TrafficSecret->TrafficSecretType,
```

src/inc/quic_platform_winuser.h
```diff
@@ -519,8 +519,11 @@ CxPlatRefDecrement(
 
 typedef HANDLE CXPLAT_EVENT;
 
-#define CxPlatEventInitialize(Event, ManualReset, InitialState) \
-    *(Event) = CreateEvent(NULL, ManualReset, InitialState, NULL)
+#define CxPlatEventInitialize(Event, ManualReset, InitialState)         \
+    do {                                                                \
+        *(Event) = CreateEvent(NULL, ManualReset, InitialState, NULL);  \
+        CXPLAT_FRE_ASSERT(*Event != NULL);                              \
```
nibanks, line 525: I don't want to make a functional change here. Do a DBG (or maybe TEL) assert for now please. Same for below.

src/platform/datapath_winuser.c
```diff
@@ -967,9 +967,11 @@ CxPlatDataPathInitialize(
         if (Datapath != NULL) {
             for (uint16_t i = 0; i < Datapath->ProcCount; i++) {
                 if (Datapath->Processors[i].IOCP) {
+#pragma prefast(suppress:6001, "SAL can't track processors allocation")
```
nibanks, line 970: I don't like having to do this. I don't understand why SAL is doing this. Can you add a loop after zeroing above that each IOCP is null, and wrap it in `#if defined(_PREFAST_) ... #endif`?

ThadHouse: That doesn't work, the analysis error still occurs even if I do that.

src/platform/datapath_winuser.c
```diff
@@ -731,6 +731,16 @@ CxPlatDataPathQuerySockoptSupport(
...
 }
 
+//
+// CloseHandle has an incorrect SAL annotation, so call through a wrapper.
+//
+_IRQL_requires_max_(PASSIVE_LEVEL)
+inline
+void
+CxPlatCloseHandle(_Pre_notnull_ HANDLE Handle) {
+    CloseHandle(Handle);
```
nibanks, line 741: Don't you need to suppress the SAL error? Can you also put this into the platform header instead?

ThadHouse: Nope, no need to suppress the SAL error, as just calling it through a function is enough to work around the bug. I can move it to the header though.

ThadHouse: Basically, the bad SAL annotation was thrashing ALL other handle values in the loop. Through a wrapper, it doesn't know its in a loop, so it doesn't thrash.

nibanks (approved):

PR #1309, opened 2021-03-04
Rev Build Cache

nibanks:
Forgot to do this in the previous PR.

### DIFF
--- a/.azure/templates/build-config-user.yml
+++ b/.azure/templates/build-config-user.yml
@@ -48,7 +48,7 @@ jobs:
 
   - task: Cache@2
     inputs:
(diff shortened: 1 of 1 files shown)


ThadHouse:
The reason this worked was .gitmodules was also part of the cache id, so changing that forced a rebuild. But still good to bump. We don't want to go to 0 or 1, as 3 has already been used and we'd probably forget.

ThadHouse (approved):

nibanks:
> The reason this worked was .gitmodules was also part of the cache id, so changing that forced a rebuild. But still good to bump. We don't want to go to 0 or 1, as 3 has already been used and we'd probably forget.

Ah, good to know. Thanks!

PR #1314, opened 2021-03-05
Fix Frame Type Encoding Issues

### LINKED ISSUE #1313: Frame Type Decoding Assumes Single Byte Encoding
### Describe the bug

The frame decoding logic was assuming the peer always used a single byte encoding for frame types. The ACK_FREQUENCY frame broke that assumption. Luckily, since support for new frames (that break this assumption) are negotiated, older versions of MsQuic are generally unaffected by this assumption because it never used frames that didn't fit. Only the newest code needs MUST be updated.

1. Run interop test between picoquic and msquic.
2. MsQuic aborts the connection with a FRAME_ENCODING_ERROR because of the "unknown" frame.

The connection should continue as normal and not abort.

nibanks:
Fixes #1313. The encoding is actually a variable length integer, not a simple byte. Up until the ACK_FREQUENCY frame, the assumption held, but now that we've added support for that, it no longer holds and breaks interop.

### DIFF
--- a/src/core/sent_packet_metadata.h
+++ b/src/core/sent_packet_metadata.h
@@ -74,6 +74,10 @@ typedef struct QUIC_SENT_FRAME_METADATA {
 
 } QUIC_SENT_FRAME_METADATA;
 
+CXPLAT_STATIC_ASSERT(
+    QUIC_FRAME_MAX_SUPPORTED <= (uint64_t)UINT8_MAX,
+    "Metadata 'Type' field above assumes frames types fit in 8-bits");
+
 typedef struct QUIC_SEND_PACKET_FLAGS {
 
     uint8_t KeyType                 : 2;
--- a/src/core/unittest/SpinFrame.cpp
+++ b/src/core/unittest/SpinFrame.cpp
@@ -47,7 +47,11 @@ TEST(SpinFrame, SpinFrame1000000)
     BOOLEAN InvalidFrame;
     uint8_t Buffer[255];
     uint8_t BufferLength = 0;
+
     uint8_t FrameType;
+    CXPLAT_STATIC_ASSERT(
+        QUIC_FRAME_MAX_SUPPORTED <= (uint64_t)UINT8_MAX,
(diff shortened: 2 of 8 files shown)


src/core/sent_packet_metadata.h
```diff
@@ -74,6 +74,10 @@ typedef struct QUIC_SENT_FRAME_METADATA {
 
 } QUIC_SENT_FRAME_METADATA;
 
+CXPLAT_STATIC_ASSERT(
+    QUIC_FRAME_MAX_SUPPORTED <= (uint64_t)UINT8_MAX,
+    "Metadata 'Type' field above assumes frames types fit in 8-bits");
```
anrossi, line 79: This only applies for packets that are sent, right? Not received?

src/core/unittest/SpinFrame.cpp
```diff
@@ -47,7 +47,11 @@ TEST(SpinFrame, SpinFrame1000000)
     BOOLEAN InvalidFrame;
     uint8_t Buffer[255];
     uint8_t BufferLength = 0;
+
     uint8_t FrameType;
+    CXPLAT_STATIC_ASSERT(
+        QUIC_FRAME_MAX_SUPPORTED <= (uint64_t)UINT8_MAX,
+        "Tests below assumes frames fit in 8-bits");
```
anrossi, line 54: Can you add an AckFrequency frame case to this test?

anrossi: Also, add a case in the QUIC_FRAME_TYPE ostream overload to pretty print the Ack Frequency frame

src/core/sent_packet_metadata.h
nibanks: Correct. It's "sent" metadata.

src/core/unittest/SpinFrame.cpp
nibanks: We already have a test case for the frame.

anrossi (approved):

PR #1316, opened 2021-03-08
Switch to OpenSSL 1.1 quic branch

ThadHouse:
1.1 is going to be easier to get certification for.

There still needs to be work done on macos, because of M1 issues, I'll look into that.

### DIFF
--- a/scripts/performance-helper.psm1
+++ b/scripts/performance-helper.psm1
@@ -573,7 +573,7 @@ function Get-TestResultAtIndex($FullResults, $Index) {
 }
 
 function Get-MedianTestResults($FullResults) {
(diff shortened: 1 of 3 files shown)


scripts/performance-helper.psm1
```diff
@@ -573,7 +573,7 @@ function Get-TestResultAtIndex($FullResults, $Index) {
 }
 
 function Get-MedianTestResults($FullResults) {
-    $sorted = $FullResults | Sort-Object
+    $sorted = $FullResults | Sort-Object {[int]$_}
```
nibanks, line 576: Can you push this in a separate PR too? I don't want to gate it on this PR.

nibanks (approved):

PR #1320, opened 2021-03-09
Use system libcrypto

ThadHouse:
Try to see if we can statically link openssl, but dynamically link libcrypto.

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -81,6 +81,12 @@ This script provides helpers for building msquic.
 .PARAMETER EnableTelemetryAsserts
     Enables telemetry asserts in release builds.
 
+.PARAMETER UseSystemOpenSSLCrypto
+    Use system provided OpenSSL libcrypto rather then statically linked. Only affects OpenSSL Linux builds
+
+.PARAMETER ExtraArtifactDir
+    Add an extra classifier to the artifact directory to allow publishing alternate builds of same base library
+
 .EXAMPLE
     build.ps1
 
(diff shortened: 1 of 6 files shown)


nibanks:
Since this seems to work, let's make this build-time configurable, and leave the default to completely statically link everything for now.

scripts/build.ps1
```diff
@@ -224,7 +236,12 @@ $BaseBuildDir = Join-Path $RootDir "build"
 $ArtifactsDir = Join-Path $BaseArtifactsDir "bin" $Platform
 $BuildDir = Join-Path $BaseBuildDir $Platform
 
-$ArtifactsDir = Join-Path $ArtifactsDir "$($Arch)_$($Config)_$($Tls)"
+if ("" -eq $ExtraArtifactDir) {
+    $ArtifactsDir = Join-Path $ArtifactsDir "$($Arch)_$($Config)_$($Tls)"
+} else {
+    $ArtifactsDir = Join-Path $ArtifactsDir "$($Arch)_$($Config)_$($Tls)_$($ExtraArtifactDir)"
+}
+
```
nibanks, line 244: Theoretically, a lot of tools (e.g. test.ps1) would need to be updated as well, right?

ThadHouse: If we wanted to run the builds, yes. I can do that either here or in a separate PR.

nibanks: We should add a test for the system libcrypto variant of openssl on Linux. Please add that, and only update the scripts that need to be updated for that to happen.

nibanks (approved):

PR #1321, opened 2021-03-09
Reduce batching counts for sends

ThadHouse:
Previously, we had our batch counts high. Reducing batching count caused a 200% increase in user mode Throughput download perf, and a 150% increase in kernel mode RPS pert. Both of those scenarios were lower then they should have been, and this change solves both of those issues.

### DIFF
--- a/src/core/quicdef.h
+++ b/src/core/quicdef.h
@@ -161,7 +161,7 @@ typedef struct QUIC_PATH QUIC_PATH;
 // FLUSH_SEND operation. The actual number will generally exceed this value up
 // to the limit of the current USO buffer being filled.
 //
-#define QUIC_MAX_DATAGRAMS_PER_SEND             245
+#define QUIC_MAX_DATAGRAMS_PER_SEND             40
 
 //
 // The number of packets we write for a single stream before going to the next
--- a/src/perf/lib/PerfServer.h
+++ b/src/perf/lib/PerfServer.h
@@ -120,7 +120,8 @@ private:
             .SetPeerBidiStreamCount(PERF_DEFAULT_STREAM_COUNT)
             .SetPeerUnidiStreamCount(PERF_DEFAULT_STREAM_COUNT)
             .SetDisconnectTimeoutMs(PERF_DEFAULT_DISCONNECT_TIMEOUT)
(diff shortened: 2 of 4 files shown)


nibanks (approved):

nibanks (approved):

shcalm:
Why does the WSASendMsg interface limit the WSABUF count to 1 instead of working like Linux's sendmmsg (which supports batches)? If sending batches improves throughput by reducing syscall overhead, why was this design chosen?

nibanks:
> Why does the WSASendMsg interface limit the WSABUF count to 1 instead of working like Linux's sendmmsg (which supports batches)? If sending batches improves throughput by reducing syscall overhead, why was this design chosen?

WSASendMsg supports more sends, but we didn't get any perf benefit from doing more than 1 at a time, so we reduced it in favor of lower memory usage.

shcalm:
thanks for reply . 
In my UDP test code (sending packets in a loop), WSASendMsg achieves higher throughput when dwBufferCount is set to 4+. 
Throughput was measured using the Performance tab in Task Manager. 


`        while (1) {
        WSAMSG msg;
        memset(&msg, 0, sizeof(msg));
        msg.name = (SOCKADDR*)&server_addr;
        msg.namelen = sizeof(server_addr);
        msg.lpBuffers = wsabuf;
        msg.dwBufferCount = 4;  
        DWORD bytes_sent;
        WSASendMsg(client_fd, &msg, 0, &bytes_sent, NULL, NULL);
      }`

mtfriesen:
Winsock interprets multiple buffers as scatter/gather IO, rather than distinct datagrams to be sent, so similar throughput and datagrams on the wire (N=1) should be achievable by concatenating the buffers in user mode and sending a single WSABUF down vs multiple smaller WSABUFs.

mtfriesen:
That said, if the datagram size exceeds MTU, and the socket hasn't set the don't fragment socket option, then IP will fragment the UDP datagram into multiple IP fragmented packets, producing multiple Ethernet frames. This is typically not a high performance result in end-to-end scenarios, since IP fragmentation is to be avoided.

shcalm:
Thanks for the explanation! 
Is there any API in Windows similar to Linux's sendmmsg that supports batch message sending?

PR #1324, opened 2021-03-10
Expand certificate testing

anrossi:
Add new certificate test cases which exercise both the error handling path for invalid certificates, and also test the various methods (in Schannel) for loading a certificate.


src/manifest/clog.sidecar
```diff
@@ -13785,6 +13785,10 @@
       {
         "UniquenessHash": "7368812e-a46d-0924-7643-009511288886",
         "TraceID": "InteropTestStop"
+      },
+      {
+        "UniquenessHash": "5fc489fc-c2f2-0e16-8c99-87acf91e235d",
+        "TraceID": "CertFindCertificateFriendlyName"
```
nibanks, line 13791: ??? There should be more than just this. Can you regen?

nibanks:
Lots of build errors still: https://dev.azure.com/ms/msquic/_build/results?buildId=146186&view=results

nibanks (approved):

src/test/lib/HandshakeTest.cpp
```diff
@@ -2171,3 +2171,131 @@ QuicTestValidAlpnLengths(
...
+
+        {
+            UniquePtr<TestConnection> Server;
+            ServerAcceptContext ServerAcceptCtx((TestConnection**)&Server);
+            Listener.Context = &ServerAcceptCtx;
+
+            {
+                TestConnection Client(Registration);
+                TEST_TRUE(Client.IsValid());
+                Client.SetExpectedTransportCloseStatus(QUIC_STATUS_INTERNAL_ERROR);
```
nibanks, line 2214: This makes it clear that our current error status for this scenario sucks. I think as a part of this, we need to add some more specific error codes for these scenarios so the app know (and our logs) more of what happened. Can you please look into adding some new errors around invalid certificate cases?

anrossi: I've added the framework for TLS alerts and a new define for expired certificate. We can expand this in later updates when we test more invalid certificate cases.

nibanks (approved):

ThadHouse (approved):

PR #1335, opened 2021-03-12
Fix interop reading out of range when running all tests (#1331)

ThadHouse:
Cherry-Picked from main. Required due to an AZP update.


ThadHouse:
We're just going to live with this failing forever, rather then merging a fix.

PR #1339, opened 2021-03-16
Fix issues with M1 mac build and running.

### LINKED ISSUE #1338: M1 Mac Fixes
Finally got access to an M1 mac, and there are a few issues that need to be worked out.

- [ ] CMake build by itself should build for current platform (Fix at end of issue)
- [ ] Powershell should detect current platform (Will need the first bullet to fix)
- [ ] Remove the assert in CxPlatProcCurrentNumber. On M1 running under rosetta, this flag returns 0. We want rosetta to work. We should decide if we want to do the same hardcode to 0 we do on Native M1 builds in this case too.
- [x] ~2 Failing Datagram tests. DatagramSend v4 and v6. Fails on both M1 and Rosetta. This also fails on a native Intel Mac, but not in CI for some reason.~ Not an issue. Failing because of missing hooks in release mode.

```
            # need to build with Apple's compiler
            if (CMAKE_OSX_ARCHITECTURES STREQUAL arm64)
                set(OPENSSL_CONFIG_CMD ARCHFLAGS="-arch arm64" ${CMAKE_SOURCE_DIR}/submodules/openssl/Configure darwin64-arm64-cc)
            elseif(CMAKE_OSX_ARCHITECTURES STREQUAL x86_64)
                set(OPENSSL_CONFIG_CMD ARCHFLAGS="-arch x86_64" ${CMAKE_SOURCE_DIR}/submodules/openssl/Configure darwin64-x86_64-cc)
            else()
                set(OPENSSL_CONFIG_CMD ${CMAKE_SOURCE_DIR}/submodules/openssl/config)
            endif()
```

ThadHouse:
Make the CMake build if no flags are set build for native platform. Doesn't currently affect pwsh builds.

Make rosetta builds not all throw a FRE assertion.

Closes #1338 mostly. Missing pwsh current system support, but that will need a lot more work.

### DIFF
--- a/src/platform/platform_posix.c
+++ b/src/platform/platform_posix.c
@@ -460,8 +460,13 @@ CxPlatProcCurrentNumber(
             "=c" (cpuinfo[2]),
             "=d" (cpuinfo[3])
             : "a"(1));
-    CXPLAT_FRE_ASSERT((cpuinfo[3] & (1 << 9)) != 0);
-    return (uint32_t)cpuinfo[1] >> 24;
+    //
+    // Cannot be an assert, as M1 running under Rosetta always has this flag 0.
+    //
+    if ((cpuinfo[3] & (1 << 9)) != 0) {
+        return (uint32_t)cpuinfo[1] >> 24;
+    }
+    return 0;
 #else
     //
     // arm64 macOS has no way to get the current proc, so just hardcode 0.
--- a/src/platform/platform_posix.c
+++ b/src/platform/platform_posix.c
@@ -461,7 +461,9 @@ CxPlatProcCurrentNumber(
             "=d" (cpuinfo[3])
             : "a"(1));
     //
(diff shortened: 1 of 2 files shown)


src/platform/platform_posix.c
```diff
@@ -460,8 +460,13 @@ CxPlatProcCurrentNumber(
             "=c" (cpuinfo[2]),
             "=d" (cpuinfo[3])
             : "a"(1));
-    CXPLAT_FRE_ASSERT((cpuinfo[3] & (1 << 9)) != 0);
-    return (uint32_t)cpuinfo[1] >> 24;
+    //
+    // Cannot be an assert, as M1 running under Rosetta always has this flag 0.
+    //
```
nibanks, line 465: Instead of comment why this can't be an assert (which makes no sense to anyone who hasn't watched the history of this code evolve) please update the comment to say why/what is happening below.

nibanks (approved):

PR #1347, opened 2021-03-18
Rename SendContext and Datagram from datpath to SendData and Buffer

ThadHouse:
The objects were renamed a while ago, but not the variables.


src/platform/datapath_winkernel.c
```diff
@@ -350,7 +350,7 @@ typedef struct CXPLAT_DATAPATH_PROC_CONTEXT {
     //
     // Pool of send contexts to be shared by all sockets on this core.
```
anrossi, line 351: "send data contexts" here too?

nibanks (approved):

PR #1351, opened 2021-03-19
add basic support for pkcs12 import

wfurt:
This is basic OpensSSL implementation to read certificate and keys from ASN.1 PKCS12/pfx blob. This allows in-memory handout without depending on pointers or platform knowledge. 

I can follow-up with Windows implementation for completeness. It should be possible as well. 

Also processing of certificate chains is missing right now. I'll try to follow-up separately when we get the basic functions working. 

Testing: 

I did not do any live traffic testing. 
But the existing key check is passing 
https://github.com/microsoft/msquic/blob/834fde3cddbda4a350bbeb91e35ac5621f748b6e/src/platform/tls_openssl.c#L883
and I added basic loading test. 

The test.pfx is basically cert/key generated by msquic on Windows exported via certificate manager.
I don't know if that is best way. .NET uses literal string with base64 content in some places but that is harder IMHO to maintain and test with.   
I also is _WIN32 as deciding factor right now. We can possibly wrap that in some more descriptive define.

On a side note, the adjacent code seems to depend on RSA kay. I'm not sure if quic is limited to RSA. I did some basic search but I did not see any references. Use of EVP_PKEY allows to abstract particular cipher type but I did not want to touch the existing code.


src/platform/tls_openssl.c
```diff
@@ -841,6 +852,71 @@ CxPlatTlsSecConfigCreate(
                 Status = QUIC_STATUS_TLS_ERROR;
                 goto Exit;
             }
+        } else if (CredConfig->Type == QUIC_CREDENTIAL_TYPE_CERTIFICATE_PKCS12) {
+            BIO* bio = BIO_new(BIO_s_mem());
+            PKCS12 *p12 = NULL;
+
+            if (bio) {
```
nibanks, line 859: Is it allowed for `bio` to be `NULL`? That would seem to be an error condition, right? I'd rather check for `!bio` and log that error and bail, instead of the way you're currently doing it.

src/platform/tls_openssl.c
```diff
@@ -841,6 +852,71 @@ CxPlatTlsSecConfigCreate(
...
+                Status = QUIC_STATUS_TLS_ERROR;
+                goto Exit;
+
+            }
+
+            STACK_OF(X509) *ca = NULL;
+            Ret =
+                PKCS12_parse(p12, CredConfig->CertificatePkcs12->PrivateKeyPassword, &privateKey, &X509Cert, &ca);
+            if (ca) sk_X509_pop_free(ca, X509_free); // no handling for custom certificate chains yet.
+            if (p12) PKCS12_free(p12);
```
nibanks, line 881:
```suggestion
            if (ca) {
                sk_X509_pop_free(ca, X509_free); // no handling for custom certificate chains yet.
            }
            if (p12) {
                PKCS12_free(p12);
            }
```

src/platform/unittest/TlsTest.cpp
```diff
@@ -165,7 +205,19 @@ struct TlsTest : public ::testing::TestWithParam<bool>
                 &TlsContext::TlsClientCallbacks,
                 &ClientSecConfigClientCertNoCertValidation,
                 OnSecConfigCreateComplete));
-        ASSERT_NE(nullptr, ClientSecConfigClientCertNoCertValidation );
+        ASSERT_NE(nullptr, ClientSecConfigClientCertNoCertValidation);
+#endif
+
+#ifndef _WIN32
```
nibanks, line 211: Shouldn't this work with Windows OpenSSL too? But as a bigger issue, I would prefer to add Windows CAPI support in parallel. It shouldn't be too bad. I believe we already have some code to do it elsewhere.

src/platform/unittest/TlsTest.cpp
```diff
@@ -64,6 +70,40 @@ struct TlsTest : public ::testing::TestWithParam<bool>
         ClientCertParams = (QUIC_CREDENTIAL_CONFIG*)CxPlatGetSelfSignedCert(CXPLAT_SELF_SIGN_CERT_USER, TRUE);
         ASSERT_NE(nullptr, ClientCertParams);
         ClientCertParams->Flags |= QUIC_CREDENTIAL_FLAG_NO_CERTIFICATE_VALIDATION;
+#endif
+#ifndef _WIN32
+        CertParams = (QUIC_CREDENTIAL_CONFIG*)malloc(sizeof(QUIC_CREDENTIAL_CONFIG));
+        if (CertParams) {
```
nibanks, line 76:
```suggestion
        CertParams = (QUIC_CREDENTIAL_CONFIG*)malloc(sizeof(QUIC_CREDENTIAL_CONFIG));
        ASSERT_NE(nullptr, CertParams);
```
Then you can remove the if block. Same for other allocation failures. Also, please use `CXPLAT_ALLOC_NONPAGED` for allocations here.

src/platform/unittest/TlsTest.cpp
```diff
@@ -770,6 +826,7 @@ const CXPLAT_TLS_CALLBACKS TlsTest::TlsContext::TlsClientCallbacks = {
 
 QUIC_CREDENTIAL_CONFIG* TlsTest::SelfSignedCertParams = nullptr;
 QUIC_CREDENTIAL_CONFIG* TlsTest::ClientCertParams = nullptr;
+QUIC_CREDENTIAL_CONFIG* TlsTest::CertParams = nullptr;
```
nibanks, line 829: Can you make the variable a bit more descriptive please? Also it seems everything getting allocated for it ends up being leaked, afaict.

src/platform/unittest/TlsTest.cpp
```diff
@@ -19,6 +21,8 @@ const uint32_t DefaultFragmentSize = 1200;
 
 const uint8_t Alpn[] = { 1, 'A' };
 const uint8_t MultiAlpn[] = { 1, 'C', 1, 'A', 1, 'B' };
+const char* TestPath = PfxTestPath;
+const char* PfxPass = "PLACEHOLDER";   // approved for cred scan
```
nibanks, line 25: We don't want to check in any binary files. We prefer to generate them dynamically as a part of the prepare-machine.ps1 script.

wfurt: I can do that. What directory this should go to? I assume the prepare-machine.ps1 is called once so it should persist cleaning, right?

src/platform/tls_openssl.c
wfurt: yes, BIO_new() can return null on failure. AFAIK this would happen only if you run out of memory but I can certainly unwind the error check. Since that unlikely IMHO I lump it with check for valid PKCS12 object.

src/platform/unittest/TlsTest.cpp
wfurt: Yes, it should. I did not do any testing on Windows. I can take a look when adding CAPI calls. It would be nice to do it as separate PR.

src/inc/msquic.h
```diff
@@ -241,6 +242,12 @@ typedef struct QUIC_CERTIFICATE_FILE_PROTECTED {
     const char *PrivateKeyPassword;
 } QUIC_CERTIFICATE_FILE_PROTECTED;
 
+typedef struct QUIC_CERTIFICATE_PKCS12 {
+    void *Asn1Blob;
+    int Length;
+    const char *PrivateKeyPassword;
```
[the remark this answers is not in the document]
wfurt: updated. if I leave const on Ans1BlobLength it won't let me assign value to it. So I left it out, added const for the Asn1Blob. Added note about optional password.

src/platform/tls_openssl.c
wfurt: moved error check for bio to separate block.

src/platform/unittest/TlsTest.cpp
wfurt: updated with CXPLAT_ALLOC_NONPAGED and ASSERT_NE.

src/platform/unittest/TlsTest.cpp
wfurt: That is hard one. Renamed to CertParamsFromFile. While this has self-signed certificate as SelfSignedCertParams, this may have more in future. If you have other suggestions please let me know.
Added more cleanup to TearDownTestSuite

src/inc/msquic.h
```diff
@@ -241,6 +242,12 @@ typedef struct QUIC_CERTIFICATE_FILE_PROTECTED {
     const char *PrivateKeyPassword;
 } QUIC_CERTIFICATE_FILE_PROTECTED;
 
+typedef struct QUIC_CERTIFICATE_PKCS12 {
+    const uint8_t *Asn1Blob;
+    int Asn1BlobLength;
```
nibanks, line 247: No `int` please. There is no reason for this length to be negative. `uint32_t` instead.

src/platform/tls_openssl.c
```diff
@@ -841,6 +851,84 @@ CxPlatTlsSecConfigCreate(
                 Status = QUIC_STATUS_TLS_ERROR;
                 goto Exit;
             }
+        } else if (CredConfig->Type == QUIC_CREDENTIAL_TYPE_CERTIFICATE_PKCS12) {
+            BIO* bio = BIO_new(BIO_s_mem());
```
nibanks, line 855: Can `BIO_s_mem()` fail? What happens if it does? Also, we use capitals for the first letters of our variables.

src/platform/tls_openssl.c
```diff
@@ -841,6 +851,84 @@ CxPlatTlsSecConfigCreate(
...
+
+            if (!p12) {
+                QuicTraceEvent(
+                    LibraryErrorStatus,
+                    "[ lib] ERROR, %u, %s.",
+                    ERR_get_error(),
+                    "d2i_PKCS12_bio failed");
+                Status = QUIC_STATUS_TLS_ERROR;
+                goto Exit;
+
```
nibanks, line 882: nit: remove extra newline

src/platform/unittest/TlsTest.cpp
nibanks: I don't particularly care where it goes. @anrossi or @thadhouse any suggestions?

src/platform/unittest/TlsTest.cpp
```diff
@@ -165,7 +216,19 @@ struct TlsTest : public ::testing::TestWithParam<bool>
...
+
+#ifndef _WIN32
+        VERIFY_QUIC_SUCCESS(
+            CxPlatTlsSecConfigCreate(
+                CertParamsFromFile,
+                CXPLAT_TLS_CREDENTIAL_FLAG_NONE,
+                &TlsContext::TlsClientCallbacks,
+                &Pkcs12SecConfig,
+                OnSecConfigCreateComplete));
+
```
nibanks, line 230: nit: remove newline

src/platform/unittest/TlsTest.cpp
nibanks: I'm fine with doing CAPI in another PR, but I'd like to get OpenSSL cross platform working in this one; so the `#ifndef _WIN32` won't work. You would have to add a new define (something line `QUIC_DISABLE_TICKET_KEY_TESTS`) to just disable the test code for this feature.

src/platform/unittest/TlsTest.cpp
wfurt: As I was looking at this agin, it will bring dependency on 'openssl' executable available at time when prepare-machine.ps1 runs. Other option I can possibly see is to build it into the test.ps1 and use OpenSSL we built. It feels like it should not go under build or artifacts as that could be lost during cleanup and one would need to run the  prepare-machine.ps1 again, right? 
I'll wait for  @anrossi or @ThadHouse. 
.NET has similar dislike for binaries so as I mentioned we use base64 in some places.

src/platform/tls_openssl.c
wfurt: no.
```c
const BIO_METHOD *BIO_s_mem(void)
{
    return &mem_method;
}
```
`mem_method` is `static const`.

CMakeLists.txt
```diff
@@ -259,6 +259,11 @@ if(WIN32)
         list(APPEND QUIC_COMMON_DEFINES QUIC_DISABLE_RESUMPTION)
     endif()
 
+    if(QUIC_TLS STREQUAL "schannel")
```
nibanks, line 262: Also going to need to be disabled for miTLS.

src/platform/tls_openssl.c
```diff
@@ -620,6 +622,12 @@ CxPlatTlsSecConfigCreate(
                 CredConfig->CertificateFileProtected->PrivateKeyPassword == NULL) {
                 return QUIC_STATUS_INVALID_PARAMETER;
             }
+        } else if(CredConfig->Type == QUIC_CREDENTIAL_TYPE_CERTIFICATE_PKCS12) {
+            if (CredConfig->CertificatePkcs12 == NULL ||
+                CredConfig->CertificatePkcs12->Asn1Blob == NULL ||
+                CredConfig->CertificatePkcs12->Asn1BlobLength <= 0) {
```
nibanks, line 628: Since this can't be negative any more, let's change it to `== 0` instead.

src/platform/unittest/TlsTest.cpp
```diff
@@ -64,6 +115,19 @@ struct TlsTest : public ::testing::TestWithParam<bool>
...
+        ASSERT_NE(nullptr, CertParamsFromFile);
+        CxPlatZeroMemory(CertParamsFromFile, sizeof(*CertParamsFromFile));
+        CertParamsFromFile->Type = QUIC_CREDENTIAL_TYPE_CERTIFICATE_PKCS12;
+        CertParamsFromFile->CertificatePkcs12 = (QUIC_CERTIFICATE_PKCS12*)CXPLAT_ALLOC_NONPAGED(sizeof(QUIC_CERTIFICATE_PKCS12), QUIC_POOL_TEST);
+        ASSERT_NE(nullptr, CertParamsFromFile->CertificatePkcs12);
+        CxPlatZeroMemory(CertParamsFromFile->CertificatePkcs12, sizeof(QUIC_CERTIFICATE_PKCS12));
+        CertParamsFromFile->CertificatePkcs12->Asn1Blob = (const uint8_t *) ReadFile(TestPath, &CertParamsFromFile->CertificatePkcs12->Asn1BlobLength);
+        CertParamsFromFile->CertificatePkcs12->PrivateKeyPassword = PfxPass;
+        ASSERT_NE(nullptr, CertParamsFromFile->CertificatePkcs12->Asn1Blob);
+        ASSERT_NE(0, CertParamsFromFile->CertificatePkcs12->Asn1BlobLength);
```
nibanks, line 130: Looks like the compiler is complaining about this. Try this?
```suggestion
        ASSERT_NE((uint32_t)0, CertParamsFromFile->CertificatePkcs12->Asn1BlobLength);
```

src/platform/unittest/TlsTest.cpp
```diff
@@ -64,6 +115,19 @@ struct TlsTest : public ::testing::TestWithParam<bool>
...
+#endif
+#ifndef QUIC_DISABLE_PFX_TESTS
+        CertParamsFromFile = (QUIC_CREDENTIAL_CONFIG*)CXPLAT_ALLOC_NONPAGED(sizeof(QUIC_CREDENTIAL_CONFIG), QUIC_POOL_TEST);
+        ASSERT_NE(nullptr, CertParamsFromFile);
+        CxPlatZeroMemory(CertParamsFromFile, sizeof(*CertParamsFromFile));
+        CertParamsFromFile->Type = QUIC_CREDENTIAL_TYPE_CERTIFICATE_PKCS12;
+        CertParamsFromFile->CertificatePkcs12 = (QUIC_CERTIFICATE_PKCS12*)CXPLAT_ALLOC_NONPAGED(sizeof(QUIC_CERTIFICATE_PKCS12), QUIC_POOL_TEST);
+        ASSERT_NE(nullptr, CertParamsFromFile->CertificatePkcs12);
+        CxPlatZeroMemory(CertParamsFromFile->CertificatePkcs12, sizeof(QUIC_CERTIFICATE_PKCS12));
+        CertParamsFromFile->CertificatePkcs12->Asn1Blob = (const uint8_t *) ReadFile(TestPath, &CertParamsFromFile->CertificatePkcs12->Asn1BlobLength);
```
nibanks, line 127: The cast here isn't strictly necessary, right? Let's remove it if we can.

src/platform/unittest/TlsTest.cpp
```diff
@@ -19,6 +22,8 @@ const uint32_t DefaultFragmentSize = 1200;
 
 const uint8_t Alpn[] = { 1, 'A' };
 const uint8_t MultiAlpn[] = { 1, 'C', 1, 'A', 1, 'B' };
+const char* TestPath = PfxTestPath;
```
nibanks, line 25: I prefer to use all caps for C #defined variables.

scripts/test.ps1
```diff
@@ -242,6 +242,14 @@ if ($Kernel) {
     }
 }
 
+$PfxFile = Join-Path $RootArtifactDir "test.pfx"
```
nibanks, line 245: I'd like a name a bit more descriptive than `test.pfx`. Perhaps something like `selfsignedservercert.pfx`.

src/platform/unittest/TlsTest.cpp
wfurt: I added QUIC_DISABLE_PFX_TESTS to guard the tests.

For CAPI we should be able to use. 
```
HCERTSTORE PFXImportCertStore(
  CRYPT_DATA_BLOB *pPFX,
  LPCWSTR         szPassword,
  DWORD           dwFlags
);
```
I'll need to do some more testing on life cycle for the HCERTSTORE.

src/inc/msquic.h
wfurt: updated. Sorry for the turn. I try to follow code around but the conventions I normally use are different and it is sometimes hard for me to spot the differences.

src/platform/unittest/TlsTest.cpp
anrossi: `Prepare-machine.ps1` has a case for windows and for linux.  There's example code in the windows portion for how to generate a certificate+private key and to create a PFX from that certificate which contains the private key. 
Only on linux would you need to call openssl binary to generate the pfx.

anrossi: One more thing, if you put the PFX into a well-known location like `$home` or `$env:userprofile` then it should survive any cleanup.

wfurt: Are you referring to this @anrossi ?
```
.PARAMETER TestCertificates
    Generate test certificates. Only supported for Windows test configuration.
```

There are two issues IMHO: 
How to generate the file and where should it live.

I did consider calling openssl directly but that may be add new dependency and it may be problematic on Windows IMHO. To solve this, I added new script to construct the certificate from cryptographic primitives. That gives us more control (for some parts openssl needs conjuration file and cannot be done via command line arguments) long run and works consistently on all platforms (as New-SelfSignedCertificate works only on Windows)

As far as the location,  `prepare-machine.ps1` runs before the build and it seems to modify global machine state.
Where do you suggest to put the generated pfx file?

wfurt: > One more thing, if you put the PFX into a well-known location like `$home` or `$env:userprofile` then it should survive any cleanup.

I can try that if we are ok that the file lives outside of the build tree. I don't know how much write access CI machine have.

nibanks: I would actually prefer keeping it in the tree. I think it's fine where it is for now.

src/platform/unittest/TlsTest.cpp
wfurt: updated. I also added comment with explanation.

scripts/test.ps1
wfurt: updated.

CMakeLists.txt
wfurt: updated. I briefly look but I don't know how to make it work with miTLS at the moment. Is that something users use in real or is that really just a reference implementation?

wfurt:
I hopefully addressed all the comments. Schannel support will come as follow-up PR. perf runs were failing on previous runs but it did not look like something this PR caused. Let me know if you want me to do anything else.

nibanks:
> I hopefully addressed all the comments. Schannel support will come as follow-up PR. perf runs were failing on previous runs but it did not look like something this PR caused. Let me know if you want me to do anything else.

Generally, looks good. Let's see how the CI goes, and if it's green I'll do one more pass of the change before merging.

wfurt:
It seems like the Windows/OpenSSL is failing because we take build artifacts and we move them somewhere else during CI. 
test.ps1 will generate the pfx but that is different location than what test binaries expect.
I will need to re-think how to determine correct file location.
Current code should work fine for all local builds and tests.

nibanks:
> It seems like the Windows/OpenSSL is failing because we take build artifacts and we move them somewhere else during CI.
> test.ps1 will generate the pfx but that is different location than what test binaries expect.
> I will need to re-think how to determine correct file location.
> Current code should work fine for all local builds and tests.

Ok. Makes sense. Let's just make the file path an input argument to the tool instead then (not a compile time constant).

wfurt:
I was thinking about using override environment variable @nibanks. That would allow to execute test test directly under debugger or any other tool and it would just work like it does now. In CI, we can set it either by the test.ps1 or by the pipeline code e.g. do the extra steps only in cases when code moves around.  
I briefly look at passing it as command line argument but plumbing from main to the place where we need to would be probably more complicated. 
Last option I was thinking about is finding executable location and using path relative to that. That can work on Linux and Windows but would be problematic IMHO on macOS or other generic Unixes.
Let me know what do you think.

nibanks:
I prefer to use the extra command line option. test.ps1 can abstract it all away.

nibanks (commented):
Looks good. Just a few minor things. Thanks for doing all this!

src/platform/unittest/main.cpp
```diff
@@ -24,8 +29,22 @@ class QuicCoreTestEnvironment : public ::testing::Environment {
...
+                i++;
+            }
+        }
+    }
+}
+
 int main(int argc, char** argv) {
     ::testing::AddGlobalTestEnvironment(new QuicCoreTestEnvironment);
     ::testing::InitGoogleTest(&argc, argv);
+    ProcessArguments(argc, argv);
```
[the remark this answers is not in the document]
wfurt: nice, it will make it more consistent. I missed this one as I was looking more for `getopt`. updated.

scripts/run-gtest.ps1
```diff
@@ -112,6 +112,9 @@ param (
 
     [Parameter(Mandatory = $false)]
     [String]$PfxPath = ""
```
nibanks, line 114: Needs a comma here

nibanks (approved):
LGTM! Thanks!

PR #1353, opened 2021-03-19
Implement UDP GRO On Linux

ThadHouse:
Still needs error testing added in, and both perf and a bit more functional testing with and without GRO enabled, but generally it works.

### DIFF
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -37,48 +37,82 @@ CXPLAT_STATIC_ASSERT((SIZEOF_STRUCT_MEMBER(QUIC_BUFFER, Buffer) == sizeof(void*)
 //
 #define CXPLAT_LARGE_SEND_BUFFER_SIZE         0xFFFF
 
-//
-// The maximum UDP receive coalescing payload.
-//
-#define MAX_URO_PAYLOAD_LENGTH              (UINT16_MAX - CXPLAT_UDP_HEADER_SIZE)
+// //
+// // The maximum UDP receive coalescing payload.
+// //
+// #define MAX_URO_PAYLOAD_LENGTH              (UINT16_MAX - CXPLAT_UDP_HEADER_SIZE)
+
+// //
+// // The maximum single buffer size for sending coalesced payloads.
+// //
+// #define CXPLAT_LARGE_SEND_BUFFER_SIZE         0xFFFF
 
 //
-// The maximum single buffer size for sending coalesced payloads.
+// Internal receive context.
 //
-#define CXPLAT_LARGE_SEND_BUFFER_SIZE         0xFFFF
+typedef struct CXPLAT_DATAPATH_INTERNAL_RECV_CONTEXT {
 
-//
-// A receive block to receive a UDP packet over the sockets.
-//
-typedef struct CXPLAT_DATAPATH_RECV_BLOCK {
     //
-    // The pool owning this recv block.
+    // The owning datagram pool.
     //
     CXPLAT_POOL* OwningPool;
 
     //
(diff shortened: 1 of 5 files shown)


src/platform/datapath_epoll.c
```diff
@@ -1644,6 +1535,260 @@ CxPlatSocketContextSendComplete(
...
+        CxPlatSocketAllocRecvContext(
+            ProcContext);
+
+    uint8_t* RecvPayload = ((uint8_t*)RecvContext) + SocketContext->Binding->Datapath->RecvPayloadOffset;
+
+    BOOLEAN UseURO = !!(SocketContext->Binding->Datapath->Features & CXPLAT_DATAPATH_FEATURE_RECV_COALESCING);
+    int BufferCount = UseURO ? 1 : NUMBER_OF_SEGMENTS_OR_BATCHES;
+    struct iovec RecvIov[NUMBER_OF_SEGMENTS_OR_BATCHES];
+    CXPLAT_RECV_MSG_CONTROL_BUFFER RecvMsgControl[NUMBER_OF_SEGMENTS_OR_BATCHES];
+    struct mmsghdr RecvMsgHdr[NUMBER_OF_SEGMENTS_OR_BATCHES];
```
ThadHouse, line 1621: These 3 arrays total to about 15k, so I'll probably pull them out to a separate allocation. I'll do it on a per thread level though, so it will be a static allocation per thread.

src/platform/datapath_epoll.c
```diff
@@ -30,44 +30,55 @@ CXPLAT_STATIC_ASSERT((SIZEOF_STRUCT_MEMBER(QUIC_BUFFER, Length) <= sizeof(size_t
...
-#define CXPLAT_MAX_BATCH_RECEIVE 40
 
 //
 // The maximum single buffer size for sending coalesced payloads.
 //
 #define CXPLAT_LARGE_SEND_BUFFER_SIZE         0xFFFF
 
 //
-// A receive block to receive a UDP packet over the sockets.
+// The maximum size of a receve segment or buffer if segmentation is not supported.
```
nibanks, line 40:
```suggestion
// The maximum size of a receive segment or buffer if segmentation is not supported.
```

src/platform/datapath_epoll.c
```diff
@@ -1644,10 +1499,273 @@ CxPlatSocketContextSendComplete(
...
+            CxPlatDataPathDatagramToInternalDatagramContext(Datagram);
+
+    QUIC_ADDR* LocalAddr = &InternalDatagramContext->Tuple.LocalAddress;
+    QUIC_ADDR* RemoteAddr = &InternalDatagramContext->Tuple.RemoteAddress;
+    uint8_t ECN;
+    uint16_t SegmentSize = 0;
+
+    if (!CxPlatSocketReadCmsg(&Msg->msg_hdr, SocketContext->Binding, LocalAddr, &ECN, &SegmentSize)) {
+        // TODO Error case missing ECN or Local Addr
+        CXPLAT_FRE_ASSERT(FALSE);
```
nibanks, line 1664: Should we really fre assert?

ThadHouse: No, that's why it has the giant todo on it. I wanted to make sure it worked then finish cleaning up.

ThadHouse:
It turns out that kernels before 5.12 support GRO, but not with recvmsg. There is no way to detect this case, which means any systems that say they support GRO would get a massive perf reduction, as neither segmentation or batching would occur. Because of this, there isn't much reason to keep this open. I want to keep the branch around for a bit in case we find a workaround, but for now what we have is the best we will get.

PR #1354, opened 2021-03-19
Switch to 43 packet receive batch size

ThadHouse:
For throughput tests, 43 is a magic number. Its 64,500 bytes, which is the largest UDP datagram able to be sent with a 1500 MTU. With a smaller batch size, we're not going to receive a full server send in 1 go, causing us to get behind. With a larger batch size, we can potentially grab a packet out of the next datagram send, causing send and receives to get out of sync with datagram boundaries. At 43, we pretty much run in lockstep, which results in the best performance

RPS should not be affected, we're not increasing the value enough to cause over allocations. Also, each packet is its own allocation, so underused packets are just used again without freeing and reallocating

### DIFF
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -30,7 +30,7 @@ CXPLAT_STATIC_ASSERT((SIZEOF_STRUCT_MEMBER(QUIC_BUFFER, Length) <= sizeof(size_t
 CXPLAT_STATIC_ASSERT((SIZEOF_STRUCT_MEMBER(QUIC_BUFFER, Buffer) == sizeof(void*)), "(sizeof(QUIC_BUFFER.Buffer) == sizeof(void*) must be TRUE.");
 
 #define CXPLAT_MAX_BATCH_SEND 1
-#define CXPLAT_MAX_BATCH_RECEIVE 40
+#define CXPLAT_MAX_BATCH_RECEIVE 43
 
 //
 // The maximum single buffer size for sending coalesced payloads.


nibanks (approved):

nibanks:
Logs seem to indicate this didn't help upload:
```
Running Test ThroughputUp_linux_x64_openssl_Default
Test Run Took 00:00:05.5916117
Run 1: 4126016 kbps
Test Run Took 00:00:04.9885370
Run 2: 4815666 kbps
Test Run Took 00:00:04.9509269
Run 3: 4748457 kbps
Test Run Took 00:00:04.9705483
Run 4: 4836990 kbps
Test Run Took 00:00:05.0679127
Run 5: 4774433 kbps
Median: 4774433 kbps (-.91%)
Remote: 4818085 kbps
```
But it did help download.
```
Running Test ThroughputDown_linux_x64_openssl_Default
Test Run Took 00:00:04.4039722
Run 1: 5656896 kbps
Test Run Took 00:00:04.6319755
Run 2: 5664721 kbps
Test Run Took 00:00:04.3708921
Run 3: 5909178 kbps
Test Run Took 00:00:04.3030861
Run 4: 5909211 kbps
Test Run Took 00:00:04.4335824
Run 5: 5911357 kbps
Median: 5909178 kbps (+6.16%)
Remote: 5566293 kbps
```
Is that what you expect?

ThadHouse:
I didn't see major changes locally, so I wasn't expecting so see massive differences anywhere. Just from the logging I was adding, I got a lot more consistency in send segmentation matching to receive batches. Since we were already close, the gains were pretty minimal.

nibanks:
Ok. Then I'd say go ahead and merge it. LGTM.

PR #1356, opened 2021-03-20
More Registration Documentation

nibanks:


docs/api/QUIC_REGISTRATION_CONFIG.md
```diff
@@ -0,0 +1,34 @@
...
+An optional (may be `NULL`), null-terminated string describing the application that created the registration. This field is primarily used for debugging purposes.
+
+`ExecutionProfile`
+
+Provides a hint to MsQuic on how to optimize its thread scheduling operations.
+
+**Value** | **Meaning**
+------ | ------
+**QUIC_EXECUTION_PROFILE_LOW_LATENCY**<br>0 | Indicates that scheduling should be generally optimized for reducing response latency. *The default execution profile.*
+**QUIC_EXECUTION_PROFILE_TYPE_MAX_THROUGHPUT**<br>1 | Indicates that scheduling should be optimized for maximum single connection throughput.
```
ThadHouse, line 28: We should add a note here that you should test if changing the profile is better for your use case, as in many cases LOW_LATENCY can actually result in much higher throughput then MAX_THROUGHPUT.

nibanks: I'll leave that for a future PR.

ThadHouse (approved):

PR #1359, opened 2021-03-21
quicipserver now prints out usage information when supplied with incorrect args.

sebiiV:
The quicipserver tool now prints out the usage arguments when supplied with no args.

Previously the ``PrintUsage()`` function was unused.

### old output
```bash
$ ./quicipserver 
Failed to load configuration from args!
```

### new output
```
$ ./quicipserver 

quicip runs a public IP lookup server.

Usage:
  quicipserver.exe -selfsign or -cert_hash:<...> [and -cert_store:<...> | -machine] or (-cert_file:<...> and -key_file:<...>)

```

### DIFF
--- a/src/tools/ip/server/quicipserver.cpp
+++ b/src/tools/ip/server/quicipserver.cpp
@@ -222,7 +222,7 @@ main(
         goto Error;
     }
 
-        if (argc < 2) {
+    if (argc < 2) {
         PrintUsage();
         goto Error;
     }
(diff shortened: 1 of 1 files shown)


src/tools/ip/server/quicipserver.cpp
```diff
@@ -222,6 +222,12 @@ main(
         goto Error;
     }
 
+        if (argc < 2) {
```
ThadHouse, line 225: Formatting please. This if is indented 4 spaces too many.

nibanks, line 225: Please fix spacing.

nibanks (approved):

PR #1364, opened 2021-03-22
Expose Revocation Schannel Flags

nibanks:
HTTP team indicated this flag is necessary for client side support of OCSP. No way to test currently.


src/platform/tls_schannel.c
```diff
@@ -1176,7 +1176,7 @@ CxPlatTlsSecConfigCreate(
         Credentials->dwFlags |= SCH_CRED_DEFERRED_CRED_VALIDATION;
     }
     if (IsClient) {
-        Credentials->dwFlags |= SCH_CRED_NO_DEFAULT_CREDS;
+        Credentials->dwFlags |= SCH_CRED_NO_DEFAULT_CREDS | SCH_CRED_REVOCATION_CHECK_END_CERT;
```
anrossi, line 1179: Is this a flag we would ever want to turn off?  OCSP can sometimes slow down client connection times (because it starts another connection to do the OCSP validation)

nibanks: This mimics the behavior of Windows HTTP. **If** we ever need to allow the app to disable this, they can ask for the feature and we can add a new "disable" flag for it.

ThadHouse: That's extremely confusing to have server default disabled and client default enabled. Couldn't we just use the existing flag for both server and client mode?

nibanks: Discussing over email.

anrossi (approved):

PR #1370, opened 2021-03-24
Restart perf server every perf run

ThadHouse:
We want to see if server warmup has an effect on tests


ThadHouse:
It doesn't seem like this actually changes anything. I suspect it might be the OS needing to warm up some of its kernel code.

PR #1374, opened 2021-03-25
Support multiple instances of secnetperf in kernel mode

ThadHouse:
This way loopback tests can be supported. Done by allowing user to supply driver name

### DIFF
--- a/src/perf/bin/appmain.cpp
+++ b/src/perf/bin/appmain.cpp
@@ -27,9 +27,6 @@ typedef struct {
 } QUIC_RUN_CERTIFICATE_PARAMS;
 #include "quic_driver_helpers.h"
 
-#define QUIC_DRIVER_NAME            "secnetperfdrv"
-#define QUIC_DRIVER_NAME_PRIVATE    "secnetperfdrvpriv"
-
 #endif
 
 extern "C" _IRQL_requires_max_(PASSIVE_LEVEL) void QuicTraceRundown(void) { }
@@ -177,6 +174,7 @@ QuicKernelMain(
     _In_ bool /*KeyboardWait*/,
     _In_ const QUIC_CREDENTIAL_CONFIG* SelfSignedParams,
     _In_ bool PrivateTestLibrary,
+    _In_z_ const char* DriverName,
     _In_opt_z_ const char* FileName
     )
 {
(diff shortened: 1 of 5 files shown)


src/perf/bin/drvmain.cpp
```diff
@@ -219,11 +215,45 @@ SecNetPerfDriverUnload(
...
+        ServiceNameLength++;
+    }
+    if (ServiceNameLength == 0) {
+        //LogError("[config] Empty service name!");
+        return STATUS_INVALID_PARAMETER;
+    }
+
+    UNICODE_STRING ServiceString = { ServiceNameLength, ServiceNameLength, BaseRegPath->Buffer + (BaseRegPathLength - ServiceNameLength) };
+
+    return RtlUnicodeStringCopy(ServiceName, &ServiceString);
```
nibanks, line 248: Do we really need to copy, or could we just reference the original memory?

ThadHouse: For this, yeah I could probably just reference the original memory.

src/perf/bin/appmain.cpp
```diff
@@ -295,7 +290,7 @@ QuicKernelMain(
             OutBuffer,
             OutBufferSize,
             &OutBufferWritten,
-            240000);
+            INFINITE);
```
anrossi, line 293: is this safe? Seems like a fun way to have a new hang

ThadHouse: Yes, this is safe to do. This is basically the wait for test to finish call. Which in server mode can literally be forever, as we just want it to stay running. Cancelling with ctrl c, the shutdown backdoor, and a client test actually finishing will cause this to return.

src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -139,6 +139,11 @@ QuicMainStart(
         if (QUIC_FAILED(Status)) {
             CxPlatDataPathUninitialize(Datapath);
             Datapath = nullptr;
+            //
+            // Must explicitly set binding to null, as CxPlatSocketCreateUdp
+            // can set the Binding variable even in invalid cases.
```
nibanks, line 144: Well, it should clear it then. Can you fix that with a separate PR?

nibanks (approved):

src/perf/lib/SecNetPerfMain.cpp
ThadHouse: Yes, I can do that in a separate PR.

PR #1381, opened 2021-03-26
fix: MaxWorkerQueueDelayMs -> MaxWorkerQueueDelayUs

z8674558:


nibanks:
Thanks for the PR, but this change actually breaks things. If you take a look at [settings.c#L518](https://github.com/microsoft/msquic/blob/main/src/core/settings.c#L518) you will see that we read this value from persistent storage in units of milliseconds and then convert to microseconds when we store it internally. So, if you intention is to be able to configure the delay at the microsecond granularity, then that line would have to be updated. That being said, I'm not convinced that's necessary.

z8674558:
Thank you, I got it about the conversion. Am I correct in understanding that we still need to update `quicdef.h`, because the member `MaxWorkerQueueDelayMs` does not exist? If so, the description would be `The maximum queue delay (in milliseconds) allowed for a worker thread`.

nibanks:
> MaxWorkerQueueDelayMs

`MaxWorkerQueueDelayMs` is the name of the external/persistent storage setting. `MaxWorkerQueueDelayUs` is the name of the internal QUIC_SETTING parameter. The Deployment.md is referencing the storage setting, not the parameter. So everything seems to be in order with the way things already were. Do you see any discrepancies?

z8674558:
I see. I was not familiar with storage because I haven't used one in my project. Thank you for the support.

nibanks:
> I see. I was not familiar with storage because I haven't used one in my project. Thank you for the support.

No problem! Thanks again!

PR #1382, opened 2021-03-26
Delay Stream SHUTDOWN_COMPLETE Event While Data Pending

### LINKED ISSUE #1384: SHUTDOWN_COMPLETE event happening before all received data could be processed
I am investigating failing tests in `System.Net.Quic` library that uses msquic. The tests are very basic, there is a peer sending some data to an endpoint (let me call it my app) in a unidirectional stream and then indicating a graceful shutdown. After that, as soon as my app's msquic acks this shutdown, `SHUTDOWN_COMPLETE` event appears on both sides. On my app side though, this event doesn't seem to pay attention to pending `RECEIVE` data (the one that got buffered by msquic but is yet to be read by the app). It may be important that the app reads the data asynchronously (i.e. returns `QUIC_STATUS_PENDING`, and then calls to `StreamReceiveComplete` and `StreamReceiveSetEnabled` when finished)

There are two scenarios in which I got `SHUTDOWN_COMPLETE` event in the middle of processing the data.

### 1. When the reads are complete
(peer sends e.g. 10K and I receive in a buffer of the same size)

I sometimes will get `SHUTDOWN_COMPLETE` right away after the last `RECEIVE` -- even before I'm able to call `StreamReceiveComplete`, and in this case `StreamReceiveComplete` _will fail_. Events will go as follows:

```
[PEER] data sent
[PEER] received event SEND_COMPLETE
        [APP] received event RECEIVE
[PEER] shutdown sent
        [APP] read started
[PEER] received event SEND_SHUTDOWN_COMPLETE
[PEER] received event SHUTDOWN_COMPLETE
        [APP] received event SHUTDOWN_COMPLETE
        [APP] read finished, will call StreamReceiveComplete
[PEER] connection closed
        [APP][ERROR] message: Exception occurred during connection callback: The channel has been closed.
        [APP][ERROR] System.Net.Quic.QuicException : Could not complete receive call. Error Code: INVALID_STATE
```

### 2. When reads are partial
(peer sends e.g. 10K and I read data in an 1K buffer several times)

In this case, I don't get any errors, but I _lose data_. E.g. I will read 3K and then I'll get a `SHUTDOWN_COMPLETE` event. After that there is no way to get the rest of data, even though it was already buffered inside msquic.

### Expected behavior

Is it possible for PEER_SEND_SHUTDOWN/SHUTDOWN_COMPLETE to behave in line with RECEIVE events (so I'll get it after I notify msquic that previous receive is completed)?

cc @nibanks

nibanks:
Fixes an issue where the stream SHUTDOWN_COMPLETE event is incorrectly delivered while the app is still draining received data (in the graceful shutdown scenario).

Fixes #1384.

Opened #1383 to track improving the test coverage in this area.

### DIFF
--- a/src/core/stream.c
+++ b/src/core/stream.c
@@ -502,6 +502,7 @@ QuicStreamTryCompleteShutdown(
     )
 {
     if (!Stream->Flags.ShutdownComplete &&
+        !Stream->Flags.ReceiveDataPending &&
         Stream->Flags.LocalCloseAcked &&
         Stream->Flags.RemoteCloseAcked) {
 


src/core/stream.c
```diff
@@ -502,6 +502,7 @@ QuicStreamTryCompleteShutdown(
     )
 {
     if (!Stream->Flags.ShutdownComplete &&
+        !Stream->Flags.ReceiveDataPending &&
```
ThadHouse, line 505: Does this solve both cases in #1384 ? Just want to make sure this doesn't just fix 1 of the 2 cases.

ThadHouse (approved):

src/core/stream.c
nibanks: Yes, it should. This flag is set so long as there is any unconsumed (by the app) data buffered. Data is not considered "consumed" until the receive has completed and the app has indicated the correct length/offset.

CarnaViire (approved):
I've checked it and it fixes both situations, thanks!

PR #1391, opened 2021-03-27
Only set SO_REUSEPORT on server sockets

### LINKED ISSUE #1389: Test Failure in AppData/WithSend0RttArgs1.Send0Rtt/286
## Describe the bug

From a PR run last night (`main` @ 3/26), we had the following failure (seen on Ubuntu OpenSSL run):

```
##[error][03/27/2021 06:27:56] AppData/WithSend0RttArgs1.Send0Rtt/286 failed:
Note: Google Test filter = AppData/WithSend0RttArgs1.Send0Rtt/286
[----------] 1 test from AppData/WithSend0RttArgs1
[ RUN      ] AppData/WithSend0RttArgs1.Send0Rtt/286
/home/vsts/work/1/s/src/test/lib/HandshakeTest.cpp:76: Failure
Client.GetIsConnected() not true
[  FAILED  ] AppData/WithSend0RttArgs1.Send0Rtt/286, where GetParam() = v6/2000/4/4/SendBuffer/Bidi (10003 ms)
[----------] 1 test from AppData/WithSend0RttArgs1 (10003 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (10192 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] AppData/WithSend0RttArgs1.Send0Rtt/286, where GetParam() = v6/2000/4/4/SendBuffer/Bidi
```

I think I've seen this a couple times in the past, but this is the first time I've grabbed the [quic.log](https://github.com/microsoft/msquic/files/6215939/quic.log). I see the following:

### Client
```
[1][7858.785d][06:27:46.009377][C][TX][0] LH Ver:0x16777216 DestCid:"c01c08438c9ac3c4" SrcCid:"" Type:I (Token 0 bytes) (Payload 1214 bytes)
[1][7858.785d][06:27:46.009379][C][TX][0]   CRYPTO Offset:0 Len:306
[1][7858.785d][06:27:46.009383][C][TX][0]   PADDING Len:884
[1][7858.785d][06:27:46.009393][data][0x7efd90002c20] Send 1232 bytes in 1 buffers (segment=1232) Dst=[::1]:41620, Src=[::1]:41620
```
### Server
```
[1][7858.785c][06:27:46.009443][data][0x7efd90002c20] Recv 1232 bytes (segment=1232) Src=[::1]:41620 Dst=[::1]:41620
[1][7858.785c][06:27:46.009446][bind][0x7efd90001720] DROP packet Dst=[::1]:41620 Src=[::1]:41620 Reason="Non-zero length CID on exclusive binding".
```
I'm not sure why it's getting dropped there. From `QuicBindingPreprocessDatagram`:
```C
if (Binding->Exclusive) {
    if (Packet->DestCidLen != 0) {
        QuicPacketLogDrop(Binding, Packet, "Non-zero length CID on exclusive binding");
        return FALSE;
    }
```
The listener binding definitely shouldn't be considered and `Exclusive` binding. Somehow that flag must be getting improperly set. Perhaps corruption somewhere? Once it's in this state, it's consistently happening (every packet from the client gets dropped the same way).

ThadHouse:
Seeing an issue where sometimes the client receives the servers port, which causes issues in the unit test. Not setting SO_REUSEPORT for client side might avoid this issue. Right now we don't have a great way to guarantee this, but it might help.

Might fix #1389

### DIFF
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -1278,6 +1278,12 @@ CxPlatSocketContextInitialize(
 
     if (LocalAddress && LocalAddress->Ipv4.sin_port != 0) {
         CXPLAT_DBG_ASSERT(LocalAddress->Ipv4.sin_port == Binding->LocalAddress.Ipv4.sin_port);
+    } else if (RemoteAddress && LocalAddress && LocalAddress->Ipv4.sin_port == 0) {
+        //
+        // A client socket being assigned the same port as a remote socket causes issues later
+        // in the datapath and binding paths. Check to make sure this case was not given to us.
+        //
+        CXPLAT_DBG_ASSERT(Binding->LocalAddress.Ipv4.sin_port != RemoteAddress->Ipv4.sin_port);
     }
 
     if (Binding->LocalAddress.Ipv6.sin6_family == AF_INET6) {
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -1276,6 +1276,7 @@ CxPlatSocketContextInitialize(
         goto Exit;
     }
 
+#if DEBUG
     if (LocalAddress && LocalAddress->Ipv4.sin_port != 0) {
         CXPLAT_DBG_ASSERT(LocalAddress->Ipv4.sin_port == Binding->LocalAddress.Ipv4.sin_port);
     } else if (RemoteAddress && LocalAddress && LocalAddress->Ipv4.sin_port == 0) {
@@ -1285,6 +1286,7 @@ CxPlatSocketContextInitialize(
         //
         CXPLAT_DBG_ASSERT(Binding->LocalAddress.Ipv4.sin_port != RemoteAddress->Ipv4.sin_port);
     }
+#endif
 
     if (Binding->LocalAddress.Ipv6.sin6_family == AF_INET6) {
         Binding->LocalAddress.Ipv6.sin6_family = QUIC_ADDRESS_FAMILY_INET6;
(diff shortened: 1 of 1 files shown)


src/platform/datapath_epoll.c
```diff
@@ -1272,6 +1278,12 @@ CxPlatSocketContextInitialize(
 
     if (LocalAddress && LocalAddress->Ipv4.sin_port != 0) {
         CXPLAT_DBG_ASSERT(LocalAddress->Ipv4.sin_port == Binding->LocalAddress.Ipv4.sin_port);
+    } else if (RemoteAddress && LocalAddress && LocalAddress->Ipv4.sin_port == 0) {
+        //
+        // A client socket being assigned the same port as a remote socket causes issues later
+        // in the datapath and binding paths. Check to make sure this case was not given to us.
+        //
+        CXPLAT_DBG_ASSERT(Binding->LocalAddress.Ipv4.sin_port != RemoteAddress->Ipv4.sin_port);
     }
```
nibanks, line 1287: Can we rework this so the `else if` is only evaluated in debug?

ThadHouse: The compiler should optimize this away, since it's an empty statement in release mode. I'll double check.

ThadHouse: Also the if is also only debug, so I'd just hide this whole thing behind debug. But the compiler should get rid of it all anyway.

nibanks: Personally, I'd prefer not to depend on that.

src/platform/datapath_epoll.c
```diff
@@ -1270,9 +1276,17 @@ CxPlatSocketContextInitialize(
         goto Exit;
     }
 
+#if DEBUG
```
nibanks, line 1279: This ends up making `LocalAddress` unused for release builds it seems?

nibanks (approved):

PR #1393, opened 2021-03-27
Disable Resumption on Perf Server

nibanks:
Checking the effect of disabling resumption on HPS perf.


nibanks:
Interestingly, there was little to no positive impact. Closing.

PR #1397, opened 2021-03-29
Fix targeting arm64-windows

wegylexy:
Not sure about `arm`, but Visual Studio 16.10 uses `ARM64` instead of `arm64`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -393,7 +393,7 @@ if(QUIC_TLS STREQUAL "openssl")
             message(FATAL_ERROR "UWP is not supported with OpenSSL")
         endif()
 
-        if (${CMAKE_GENERATOR_PLATFORM} STREQUAL "arm64")
+        if (${CMAKE_GENERATOR_PLATFORM} STREQUAL "ARM64")
             set(QUIC_OPENSSL_WIN_ARCH "VC-WIN64-ARM")
         elseif (${CMAKE_GENERATOR_PLATFORM} STREQUAL "arm")
             set(QUIC_OPENSSL_WIN_ARCH "VC-WIN32-ARM")


ThadHouse:
I would rather make that string comparison just not be case sensitive. I don't think our CI systems are updated to 16.10 yet, so this has the potential to break.

wegylexy:
@ThadHouse `-A arm64` is probably fine, but cannot be specified in `CMakeSettings.json` via the GUI editor nor text editor. ARM64 support in this tool is new in 16.9 Preview 3 according to https://devblogs.microsoft.com/cppblog/windows-arm64-support-for-cmake-projects-in-visual-studio/

wegylexy:
Oh, tracing the CI error message, it needs to be upper case elsewhere too.

ThadHouse:
It looks like 16.9, which is what AZP supports, fails, so this isn't something we could directly take. I'm still confused on whats not working. I build with 16.10 locally just fine.

ThadHouse:
The issue there is `STREQUAL` is case sensitive, and `CMAKE_GENERATOR_PLATFORM` is `arm64`. So the comparison needs to be `arm64` too.

wegylexy:
Alright, either Visual Studio 16.11 in the future should generate `-A arm64` or everything here needs to be upper case, to support building with CMakeSettings.json via the GUI. Or better, release pre-built arm64 openssl binaries.

ThadHouse:
@weglexy an external contributor got Ninja working in https://github.com/microsoft/msquic/pull/1415

So running from VS CMake generation should work now.

wegylexy:
Aha, the key is `${SYSTEM_PROCESSOR}`.

PR #1399, opened 2021-03-29
Run Asan on OpenSSL builds instead of Stub builds

ThadHouse:
The 3.0 branch did not pass ASAN, but 1.1.x should.


nibanks (approved):

src/platform/unittest/CryptTest.cpp
```diff
@@ -347,6 +347,8 @@ TEST_F(CryptTest, HpMaskChaCha20)
         LogTestBuffer("Calculated Mask:   ", Mask, sizeof(ExpectedMask));
         FAIL();
     }
+
+    CxPlatHpKeyFree(HpKey);
```
ThadHouse, line 351: This is only going to free if the test passes. Likely ok for our purposes, but still something to note.

nibanks: I agree, but practically that requires a C++ abstraction for all our stuff to fix properly.

nibanks (approved):

PR #1411, opened 2021-03-30
Remove stub TLS

ThadHouse:
We were only using stub TLS to be compatible with ASAN. Now that OpenSSL and Asan work together, we can remove stub TLS and reduce our TLS scope.

The only build this removes that was slightly used was the M1 mac build. OpenSSL for M1 mac will not currently build on the CI pipeline, but it will build locally. On my off time I'm going to try and get M1 openssl working, but for now at least I can test locally.


.azure/azure-pipelines.ci.yml
```diff
@@ -118,14 +118,6 @@ stages:
       config: Debug
       extraBuildArgs: -EnableTelemetryAsserts
   # Other configurations.
```
nibanks, line 120: We should remove this comment too

PR #1421, opened 2021-04-01
Prevent spuriously re-invoking make on OpenSSL targets during incremental builds

jeremyong:
*edit: It appears this problem may have existed before my last change*

My last change I believe had an unfortunate regression (? not entirely sure because I haven't worked with the project for too long).

I noticed that when compiling incrementally, "make" was running excessively on the OpenSSL target. While this isn't entirely disruptive because previously built objects were still cached by make, it meant that all OpenSSL headers and libraries would get recopied. This updates the modifiedAt timestamps which then causes downstream dependencies to relink, including `msquic` and any other downstream dependents.

This change does some minor cmake refactor. The issue was including both debug/release outputs in a single target. Because for a given config, only one variant of the lib was produced, the build system believes the target as a whole is out of date. The solution is to simply split the target into debug and release variants, and conditionally link the correct one via generator expressions for the final `OpenSSL` interface target.

I tested Visual Studio (MSVC), the build script, and Ninja on Win32 with both a clean and incremental builds for this change. Let me know if I missed something and sorry again for the regression.

### DIFF
--- a/.azure/templates/build-config-user.yml
+++ b/.azure/templates/build-config-user.yml
@@ -49,7 +49,7 @@ jobs:
 
   - task: Cache@2
     inputs:
-      key:  '"${{ parameters.platform }}_${{ parameters.arch }}_${{ parameters.tls }}_${{ parameters.extraName }}_${{ parameters.config }}_9" | .gitmodules'
+      key:  '"${{ parameters.platform }}_${{ parameters.arch }}_${{ parameters.tls }}_${{ parameters.extraName }}_${{ parameters.config }}_11" | .gitmodules'
       path: build/${{ parameters.platform }}/${{ parameters.arch }}_${{ parameters.tls }}/openssl
     displayName: Cache OpenSSL
     condition: and(succeeded(), eq('${{ parameters.tls }}', 'openssl'), not(${{ parameters.skipOpenSSLCache }}))
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -380,18 +380,6 @@ if(QUIC_TLS STREQUAL "openssl")
             message(FATAL_ERROR "UWP is not supported with OpenSSL")
         endif()
 
-        if (${SYSTEM_PROCESSOR} STREQUAL "arm64")
-            set(QUIC_OPENSSL_WIN_ARCH "VC-WIN64-ARM")
-        elseif (${SYSTEM_PROCESSOR} STREQUAL "arm")
-            set(QUIC_OPENSSL_WIN_ARCH "VC-WIN32-ARM")
-        elseif (${SYSTEM_PROCESSOR} STREQUAL "win32")
-            set(QUIC_OPENSSL_WIN_ARCH "VC-WIN32")
-        elseif (${SYSTEM_PROCESSOR} STREQUAL "x64" OR ${SYSTEM_PROCESSOR} STREQUAL "amd64")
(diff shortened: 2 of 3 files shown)


ThadHouse:
https://github.com/microsoft/msquic/blob/c9120e69393e4d6ee02fcb978c8b5343f8a31079/.azure/templates/build-config-user.yml#L52

Can you bump the 9 near the end of this line to 10. We realized that the previous pr actually broke openssl completely for platforms other then the current platform, and we didn't notice because we cache openssl for build times. So for any build changes, we need to bump that number to ensure clean builds work.

jeremyong:
Done

ThadHouse:
There's build failures in CI. Can you take a look?

jeremyong:
Yea I'm looking now, I have a name collision for a target on the release configuration for VS. I'll debug it locally, I must not have cleaned that configuration properly when testing.

jeremyong:
image

Quick update, this generator expression in the `target_link_libraries` doesn't seem to be doing what I'm expecting.

ThadHouse:
We actually use a fork of openssl to get QUIC support, so standard prebuilt binaries actually won't work, unless you're building from the fork externally anyway.

jeremyong:
Ok I'll have another go this afternoon then. My next approach will likely require more restructuring as I intend to use ExternalProject which understands how to chain configure -> make dependencies.

ThadHouse:
I think this issue was the reason I didn't actually mark the files originally as outputs, but instead just depended on the task dependencies. It worked good enough for the VS generator use case.

jeremyong:
OK I think we can try this again. To fix the oddities in the dependency graph, I had to do some research into CMake, not being a CMake expert myself. The common practice appears to be to not intermix custom build targets with the standard CMake dependency graph, preferring instead the "superbuild" idiom. To achieve this, I split out Win32 OpenSSL builds into a separate project `OpenSSLQuic` nested in the submodules folder. This is added as an in-source dependency using FetchContent. To test it, I repeatedly nuked and build the project starting from the configuration stage, then checked incremental builds for Debug and Release variants as well.

All other platform and build configurations should be untouched. As I don't have access to mac/linux at the moment, I did not convert the OpenSSL build steps for those platforms to the same idiom, although that is doable in the future. The CI builds will link the static imports directly as before.

jeremyong:
Sorry about last push. One other modification I made in this change is changing instances of `$<CONFIG:Release>` to `$<NOT:$<CONFIG:Debug>>` as a catchall for non-debug build variants to use the release build of OpenSSL (e.g. rel with debug symbols, rel with address sanitizers enabled, etc.)

ThadHouse:
Looks good! I'll try and do the same thing for linux and mac in my off time, as I like the idea. 1 formatting nit for a new line at the end of the new cmake file, and we need the license from the existing cmake file at the top of the new cmake file, but everything else looks good.

jeremyong:
Done!

jeremyong:
Last push adds the newline... VSCode was stripping it

jeremyong:
All the builds seem green at least. The schannel regressions a separate issue I assume?

nibanks:
> All the builds seem green at least. The schannel regressions a separate issue I assume?

I submitted a rerun. It should be unrelated to this change, but we like to get things green for external PRs before merging them.

ThadHouse (approved):

nibanks (approved):

PR #1424, opened 2021-04-01
Reduce the size of the receive packet queue

ThadHouse:
We had a maximum of 4k packets queued on a connection before the receive side would drop. In larger CPU limited tests, a buffer this big could actually be hit, resulting in dropped packets, and because the buffer was so big, we'd hit the max buffer size pretty consistently. By decreasing the max queue to ~4 full segments (180), we both reduce the amount of time we spend decrypting, which means if the buffer is hit, its less likely to bloat out of control.


nibanks (approved):

anrossi (approved):

nibanks (approved):
Please revert the launchsettings change, but otherwise good to go!

PR #1427, opened 2021-04-02
Always treat macOS as a single core platform

ThadHouse:
On M1 macs, there is no way to get the current CPU. So our only option is to treat the system as a single core.
On Intel macs, with certain configurations CPUID can return processor numbers that are greater then the number the syscall tells us. This results in assertions, or out of range calls in release mode. We don't currently have a workaround, so treat Intel macOS as a single core.

Since both platforms have current core issues, TL;DR just always treat macOS as a single core platform.

### DIFF
--- a/src/platform/platform_posix.c
+++ b/src/platform/platform_posix.c
@@ -420,13 +420,14 @@ CxPlatProcMaxCount(
     void
     )
 {
-#if defined(CX_PLATFORM_DARWIN) && defined(__arm64__)
+#if defined(CX_PLATFORM_DARWIN)
     //
-    // arm64 macOS has no way to get the current proc, so act like a single core.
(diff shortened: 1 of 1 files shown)


src/platform/platform_posix.c
```diff
@@ -441,7 +441,7 @@ CxPlatProcActiveCount(
     //
     return 1;
 #else
-    return (uint32_t)sysconf(_SC_NPROCESSORS_ONLN);
```
nibanks, line 444: It sounds like `_SC_NPROCESSORS_ONLN` might actually be the equivalent of "active" processors while `_SC_NPROCESSORS_CONF` is "max". Thoughts?

I also looked at the code and there are a couple of places that we call active where we should probably call max instead.

nibanks (approved):

PR #1430, opened 2021-04-03
Support Cipher Suite Allow List

### LINKED ISSUE #109: Add Option for Ciphers to QUIC_CREDENTIAL_CONFIG
Compare to [CipherSuitesPolicy](https://docs.microsoft.com/en-us/dotnet/api/system.net.security.ciphersuitespolicy?view=netcore-3.1).

The client and server need the ability in code to restrict which ciphers can be negotiated. Additionally, the server needs to be able to make this decision per connection, after receiving the Client Hello. E.g. HTTP/2 required negotiating from a restricted set of ciphers, but you wouldn't know if the connection was using HTTP/2 until the ALPN information was received.

Historically ciphers have only been configurable at the OS level and that's still true on Windows/Schannel. OpenSsl however provides the ability to specify a list of allowed ciphers in code that overrides the global settings.

This capability is useful when phasing in new ciphers or phasing out old ones. Doing so machine wide risks destabilizing too many applications at once, services want to update apps one at a time.

anrossi:
Add support to Schannel and OpenSSL to explicitly indicate which TLS1.3 cipher suites to use. Closes #109


src/inc/msquic.h
```diff
@@ -269,6 +275,8 @@ typedef struct QUIC_CREDENTIAL_CONFIG {
     const char* Principal;
     void* Reserved; // Currently unused
     QUIC_CREDENTIAL_LOAD_COMPLETE_HANDLER AsyncHandler; // Optional
+    QUIC_CIPHER_SUITE* AllowedCipherSuites;             // Optional
+    uint8_t AllowedCipherSuitesLength;
```
nibanks, line 279: `QUIC_CREDENTIAL_FLAGS Flags;` needs to control read access to this field so as to not break backwards compatibility.

nibanks: I am also leaning towards prefering this be a set of flags instead of an array of enums. It's more compact, we don't need that many config options, and it's simpler for an app to just OR some flags than build up an array (of which you have to worry about memory lifetime).

anrossi: The one thing we lose with moving to flags from an array of enums is the caller can no longer specify a preferred order of cipher suites; we determine the order when we build the string for OpenSSL.
Schannel doesn't allow for caller to specify order either; only whether it is allowed or not.
The .NET documentation for [CipherSuitesPolicy](https://docs.microsoft.com/en-us/dotnet/api/system.net.security.ciphersuitespolicy.-ctor?view=netcore-3.1#System_Net_Security_CipherSuitesPolicy__ctor_System_Collections_Generic_IEnumerable_System_Net_Security_TlsCipherSuite__)  doesn't guarantee the client-specified order is respected either, so this might be a moot point.

nibanks: Yeah... Let's just order by most secure.

src/platform/unittest/TlsTest.cpp
```diff
@@ -269,10 +290,47 @@ struct TlsTest : public ::testing::TestWithParam<bool>
...
+        VERIFY_QUIC_SUCCESS(
+            CxPlatTlsSecConfigCreate(
+                &ClientCredConfigCipherSuite,
+                CXPLAT_TLS_CREDENTIAL_FLAG_NONE,
+                &TlsContext::TlsClientCallbacks,
+                &ClientSecConfigAes128,
+                OnSecConfigCreateComplete));
+        ASSERT_NE(nullptr, ClientSecConfigAes128);
+
+        ClientCredConfigCipherSuite.AllowedCipherSuites = QUIC_ALLOWED_CIPHER_SUITE_AES_256_GCM_SHA384;
```
nibanks, line 313: Should also test negative cases too.

anrossi: Added some test cases below which cover no allowed cipher suites, and also when an unrecognized flag is passed.

nibanks (approved):

rzikm:
Hi, @anrossi, was there a specific reason to not support `TLS_AES_128_CCM_SHA256`? it is listed among the [TLS 1.3 cipher suites](https://datatracker.ietf.org/doc/html/rfc8446#appendix-B.4) and [QUIC-TLS RFC](https://datatracker.ietf.org/doc/html/rfc9001#section-5.3) only prohibits `TLS_AES_128_CCM_8_SHA256`:

> QUIC can use any of the cipher suites defined in [[TLS13](https://datatracker.ietf.org/doc/html/rfc9001#ref-TLS13)] with the
   exception of TLS_AES_128_CCM_8_SHA256.  A cipher suite MUST NOT be
   negotiated unless a header protection scheme is defined for the
   cipher suite.  This document defines a header protection scheme for
   all cipher suites defined in [[TLS13](https://datatracker.ietf.org/doc/html/rfc9001#ref-TLS13)] aside from
   TLS_AES_128_CCM_8_SHA256.  These cipher suites have a 16-byte
   authentication tag and produce an output 16 bytes larger than their
   input.

PR #1431, opened 2021-04-03
Enable ChaCha20-Poly1305 for OpenSSL

### LINKED ISSUE #996: ChaCha20-Poly1305 doesn't work in OpenSSL
OpenSSL needs more investigation into why the ChaCha20-Poly1305 cipher suite doesn't interop with mitls.
The handshake key fails to decrypt, or fails the validate the tag on the server's flight.

anrossi:
Closes #996.


src/platform/tls_openssl.c
```diff
@@ -139,7 +139,7 @@ typedef struct CXPLAT_HP_KEY {
 //
 // Default list of Cipher used.
 //
-#define CXPLAT_TLS_DEFAULT_SSL_CIPHERS    "TLS_AES_256_GCM_SHA384:TLS_AES_128_GCM_SHA256"
+#define CXPLAT_TLS_DEFAULT_SSL_CIPHERS    "TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256"
```
nibanks, line 142: If this does work, we should enable the interop test at least for server. We can't enable it on client, until we have your other PR and can explicitly configure ChaCha20 only.

nibanks:
Things left to do:

- [ ] Integrate latest `main` to get your cipher config changes.
- [ ] Use cipher config to add tests.
- [ ] Enable Interop server support (+ any QNS changes).
- [ ] Enable Interop client support (+ any QNS changes).
- [ ] Close #996.

ThadHouse:
We also need to update `CxPlatMapCipherSuite` to give results for ChaChaPoly, and add the handshake test for chachapoly.

ThadHouse (approved):

src/platform/tls_openssl.c
```diff
@@ -1811,14 +1811,11 @@ CxPlatMapCipherSuite(
...
-        // Not supporting ChaChaPoly for querying currently.
-        //
-        // case QUIC_CIPHER_SUITE_TLS_CHACHA20_POLY1305_SHA256:
-        //     HandshakeInfo->CipherAlgorithm = QUIC_ALG_CHACHA20;
-        //     HandshakeInfo->CipherStrength = 256;
-        //     HandshakeInfo->Hash = QUIC_ALG_SHA_256;
-        //     break;
+        case QUIC_CIPHER_SUITE_TLS_CHACHA20_POLY1305_SHA256:
+            HandshakeInfo->CipherAlgorithm = QUIC_CIPHER_ALGORITHM_CHACHA20;
+            HandshakeInfo->CipherStrength = 256; // TODO - Is this correct?
```
anrossi, line 1816: according to the documentation referenced, it's not. but technically, chacha20 uses a 256 bit key, so it's technically correct. The best kind of correct.

ThadHouse (approved):

nibanks (approved):

PR #1438, opened 2021-04-05
Remove Async TLS Support

nibanks:
At one point in time we considered supporting the async version of the Schannel/SSPI API, so our TLS abstraction layer was built to support it. For various reasons, we've decided not to go down that path. So this PR removes support for it.

### DIFF
--- a/src/core/operation.h
+++ b/src/core/operation.h
@@ -27,7 +27,7 @@ typedef enum QUIC_OPERATION_TYPE {
     QUIC_OPER_TYPE_UNREACHABLE,         // Process UDP unreachable event.
     QUIC_OPER_TYPE_FLUSH_STREAM_RECV,   // Indicate a stream data to the app.
     QUIC_OPER_TYPE_FLUSH_SEND,          // Frame packets and send them.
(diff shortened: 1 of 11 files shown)


src/core/operation.h
```diff
@@ -27,7 +27,7 @@ typedef enum QUIC_OPERATION_TYPE {
     QUIC_OPER_TYPE_UNREACHABLE,         // Process UDP unreachable event.
     QUIC_OPER_TYPE_FLUSH_STREAM_RECV,   // Indicate a stream data to the app.
     QUIC_OPER_TYPE_FLUSH_SEND,          // Frame packets and send them.
-    QUIC_OPER_TYPE_TLS_COMPLETE,        // A TLS process call completed.
+    QUIC_OPER_TYPE_DEPRECATED,          // No longer used.
```
ThadHouse, line 30: Since this isn't a public file, can't we just remove this?

ThadHouse: Nevermind, it would break all logs, wouldn't it.

nibanks: The problem is that these map to the ETW manifest, which is semi-public. At the very least, our Windows builds will have them for some time.

ThadHouse (approved):

Issue #1133, opened 2021-01-04
Support Static Library Build Option


okbaswc:
How can I properly do a static library build of msquic? I think this should be an option in the build scripts.

I do believe many developers could prefer to link statically to this, for all the typical reasons.

nibanks:
I'm not necessarily opposed to supporting this, but one point of complexity will be the static libraries MsQuic currently links (such as OpenSSL) that the app will have to deal with.

ThadHouse:
Another issue is we do initialization at the dll entry point, both on linux and windows. So in order to do this, a user would have to explicitly call those initialization functions in order to prepare the library for use. Its not impossible, but it would require extra calls to work in a statically linked context.

jeremyong:
I may give this a try this weekend. For the DLL entry replacement, would it be better to do lazy initialization (say, the first time the API table is retrieved) as opposed to user-driven initialization?

nibanks:
> I may give this a try this weekend. For the DLL entry replacement, would it be better to do lazy initialization (say, the first time the API table is retrieved) as opposed to user-driven initialization?

The problem with that is synchronization of two simultaneous initialization calls. That requires a lock to synchronize, but some platforms require a function call **first** to initialize that lock. So lazy initialization isn't possible there.

jeremyong:
OK it didn't occur to me that retrieving the function table needed to be threadsafe (and that this isn't C++ where we can invoke a static initializer on a static constructor in the global namespace)

nibanks:
It's theoretically possible to "roll your own" thing here that's based on interlocked operations in a loop with the first caller just initializing the lock, and the others just spinning until that's done, but I'd prefer not to do that.

jeremyong:
My initial gut is honestly just to use an atomic plus mutex (disabled in the shared lib case). Without contention, the overhead of a mutex/atomic-load is surely negligible right? That said, in my usage of the library, I cache the pointer to the function table, so I assume the concern is the pattern where people retrieve the function table frequently. One remedy is to simply say "don't do that" in the particular case of when the user opts to employ static linkage.

jeremyong:
It took me a moment to remember the pattern I was thinking of but I remembered just now: https://en.wikipedia.org/wiki/Double-checked_locking

In particular, the "problems" with the pattern are relatively easily resolved with new primitives for declaring memory order acquire/release semantics, so the implementation should be portable to Arm or other processors with weaker consistency guarantees.

nibanks:
[...]

PR #1446, opened 2021-04-07
Add initial static library support

jeremyong:
Address #1133 

This is an initial pass at this change. Setting `QUIC_BUILD_SHARED` to `OFF` and building against my test project seems to work. Two additional functions `MsQuicLoad` and `MsQuicUnload` are provided that users are expected to invoke at application launch and terminate respectively.

Caveats are that I don't have the windows insider build so I can't test schannel, I've only verified this on Windows, and I haven't added the static build configuration to the CI build matrix. Lots of caveats I know but hopefully this small change is a useful starting point nonetheless.

Cheers


src/bin/CMakeLists.txt
```diff
@@ -44,12 +49,14 @@ configure_file(msquic-config.cmake.in ${CMAKE_BINARY_DIR}/msquic-config.cmake)
...
-install(EXPORT msquic DESTINATION ${msquic_dest})
+if(BUILD_SHARED_LIBS)
+    install(EXPORT msquic DESTINATION ${msquic_dest})
+endif()
 
 if(WIN32)
     add_library(msquic.lttng INTERFACE)
 elseif(QUIC_ENABLE_LOGGING)
-    add_library(msquic.lttng SHARED $<TARGET_OBJECTS:platform.clog.provider> $<TARGET_OBJECTS:core.clog.provider>)
+    add_library(msquic.lttng $<TARGET_OBJECTS:platform.clog.provider> $<TARGET_OBJECTS:core.clog.provider>)
```
ThadHouse, line 59: This actually always needs to be a shared library, because of how this works. But this doesn't actually affect user code in the normal case.

src/bin/static/init.c
```diff
@@ -0,0 +1,29 @@
...
+}
+
+void
+MsQuicUnload(
+    void
+    )
+{
+    MsQuicLibraryUnload();
+    CxPlatSystemUnload();
+}
```
ThadHouse, line 29: New line at end of file.

ThadHouse:
Definitely a great starting point. One place I know there is going to be issues is the openssl core and platform binaries don't get put in the output directory. We probably will need some type of after build task to actually make that work, and copy the pdb's too at least on windows. But that can be tested separately easily enough. This was definitely what I thought was the best option.

jeremyong:
One option may be to use the system native archiver tool to merge the static libraries, for example as seen [here](https://stackoverflow.com/questions/60190374/how-to-bundle-multiple-static-libraries-into-single-library-in-cmake-for-windows). A corresponding per platform command would be needed (I believe this is `ar` on Linux and `libtool` on MacOS). This way, you'd end up with a single fat binary (you'd have to merge the pdbs and then copy that also).

ThadHouse:
I've done that too for some other projects, and it works pretty well. You can't easily merge the PDB's however, but you can just copy all the PDB's. I was thinking it'd probably be a good idea to merge all the static archives though. If you can integrate that easily into the CMake build that'd be awesome.

src/inc/msquic.h
```diff
@@ -1266,6 +1266,21 @@ MsQuicClose(
     _In_ _Pre_defensive_ const QUIC_API_TABLE* QuicApi
     );
 
+#ifdef QUIC_BUILD_STATIC
+//
+// If MsQuic is built as a static library, the DLL load and unload hooks are not
+// active. The user MUST invoke MsQuicLoad prior to interacting with any other
+// aspect of the API
```
nibanks, line 1273:
```suggestion
// aspect of the API.
```

src/inc/msquic.h
```diff
@@ -1266,6 +1266,21 @@ MsQuicClose(
...
+//
+void MsQuicLoad(void);
+
+//
+// If MsQuic is built as a static library, the DLL load and unload hooks are not
+// active. The user MUST invoke MsQuicUnload prior to program exit
```
nibanks, line 1279:
```suggestion
// active. The user MUST invoke MsQuicUnload prior to program exit.
```

nibanks (commented):
We need to onboard something to the CI to validate this build mode. I'll have to think how best to do that.

jeremyong:
I got this almost working last night (archiving all the static libs together) and finished it just now. It turned out to be a lot trickier than expected because I had to juggle interface dependencies, imported libraries, system libraries, and logical cmake targets. The other confounding factor was the generator expressions that didn't expand nicely in some contexts that were needed.

The latest commit solves the above and produces a single msquic.lib file (22 MB in Debug configuration) by walking the entire dependency graph and flattening a string of all linker inputs. This string contains generator expressions, so another intermediate step generates a file corresponding to each config type. At build time, this file is fed directly to `link.exe` to produce the final library, which is in turn linked directly to what is now an interface `msquic` target.

nibanks:
Great! So how do you propose validating this with the CI? Ideally, I think I'd like to build the test apps with a static msquic.lib and make sure the tests still pass. Thoughts?

jeremyong:
Ah one change I snuck in inadvertently is another optional cmake option to add a prefix to the IDE helper folders so I could more easily differentiate msquic targets from my own. Let me know if you'd rather refactor this to a separate PR or just remove it altogether.

For CI yea I think we could just double up every test by changing the `QUIC_BUILD_SHARED` option to off and things should still succeed (the final `msquic` target in this configuration literally only has the `msquic.lib` linker flag). That said, I'm not really familiar with the CI setup in this project. You're more than welcome to add that or give me pointers on what I need to change though.

jeremyong:
The release build is 35 MB btw if you were curious, I'm assuming because inlining/unrolling is far more aggressive

nibanks:
Perhaps, start by building it. Wire the new option up to the `build.ps1` script. Then add a new build job to `build_windows_debug` (for both schannel and openssl) and `build_linux_debug` (just normal openssl) that passes the `extraBuildArgs` option you add to `build.ps1`.

ThadHouse:
Is recursively enumerating through the libs also grabbing the windows SDK .lib files? I don't think we want those merged. If it is, we probably will need to explicitly list the libraries we want to join together, and only grab those from the dependencies. Those windows libraries might be why those sizes are much bigger then I'd expect.

jeremyong:
Sure, which windows libs would you expect the user to need to link? We can add those to the interface target at the end

It looks like we shouldn't archive `ws2_32`, `schannel`, `ntdll`, `bcrypt`, `ncrypt`, `crypt32`, `iphlpapi`, and `advapi32`?

I'll see if I can look through the ps1 scripts later tonight

ThadHouse:
I'm not worried about the size. Openssl is a 12 mb binary as well. I just want to make sure our lib only include binaries from us, and doesn't accidentally merge the system libraries in as well.

jeremyong:
After looking into the link options, the recent penultimate commit adds a new `EXCLUDE_LIST` local var to exclude system libraries from the archiver. For linking purposes, they are also added as a `/NODEFAULTLIB:[lib]` argument so the linker won't complain about unresolved references to symbols found in the system libs.

I have two fixes pending when using the build script. One easy (incorrect output path), the other one I need to chase down a LNK1169 error

jeremyong:
OK the LNK1169 error occurs when building any of the Quic tools. It looks like some of the tools selectively define `QuicTraceRuntime` as an empty-body function to avoid tearing down the library and IO threads. I can resolve this using the `/FORCE` linker flag but just wanted to mention that you may want a cleaner way to disable the rundown (I see that function early outs if `MsQuicLib.Loaded` isn't set for example)

ThadHouse:
I think in each of the tools we'll just have to only compile that function if building in shared mode. 

That is actually there for tracing, but for reasons the shared build of msquic doesn't expose it, so we need an empty definition in each tool. 

In a static build, the definition in the core code is the one we want, so we can easily and safely conditionally compile the copies in the tools.

jeremyong:
Oh I see I had missed your comment. I'll roll back the last commit and correct it

jeremyong:
@ThadHouse I added a pound def to exclude the empty trace rundown definitions when building the monolithic target and everything builds. I think the only thing remaining is to add the configurations to your CI and kick things off. What's the best way to do this?

.azure/azure-pipelines.ci.yml
```diff
@@ -101,6 +101,22 @@ stages:
       tls: openssl
       config: Release
       extraBuildArgs: -EnableTelemetryAsserts
+  - template: ./templates/build-config-user.yml
+    parameters:
+      image: windows-latest
+      platform: windows
+      arch: x64
+      tls: schannel
+      config: Release
```
nibanks, line 110:
```suggestion
      config: Release
      extraName: 'static'
```

.azure/azure-pipelines.ci.yml
```diff
@@ -101,6 +101,22 @@ stages:
...
+      extraBuildArgs: -EnableTelemetryAsserts -Static
+  - template: ./templates/build-config-user.yml
+    parameters:
+      image: windows-latest
+      platform: windows
+      arch: x64
+      tls: openssl
+      config: Release
```
nibanks, line 118:
```suggestion
      config: Release
      extraName: 'static'
```

.azure/azure-pipelines.ci.yml
```diff
@@ -122,6 +138,22 @@ stages:
       tls: openssl
       config: Debug
       extraBuildArgs: -EnableTelemetryAsserts
+  - template: ./templates/build-config-user.yml
+    parameters:
+      image: windows-latest
+      platform: windows
+      arch: x64
+      tls: schannel
+      config: Debug
```
nibanks, line 147:
```suggestion
      config: Debug
      extraName: 'static'
```

.azure/azure-pipelines.ci.yml
```diff
@@ -122,6 +138,22 @@ stages:
...
+      extraBuildArgs: -EnableTelemetryAsserts -Static
+  - template: ./templates/build-config-user.yml
+    parameters:
+      image: windows-latest
+      platform: windows
+      arch: x64
+      tls: openssl
+      config: Debug
```
nibanks, line 155:
```suggestion
      config: Debug
      extraName: 'static'
```

nibanks:
According to the CI, it looks like you broke normal builds' linking (just for the tools/tests?):
```
ApiTest.obj : error LNK2001: unresolved external symbol "struct QUIC_CREDENTIAL_CONFIG ServerSelfSignedCredConfig" (?ServerSelfSignedCredConfig@@3UQUIC_CREDENTIAL_CONFIG@@A) [D:\a\1\msquic\build\windows\x64_schannel\src\test\lib\testlib.vcxproj]
ApiTest.obj : error LNK2001: unresolved external symbol MsQuicClose [D:\a\1\msquic\build\windows\x64_schannel\src\test\lib\testlib.vcxproj]
ApiTest.obj : error LNK2001: unresolved external symbol MsQuicOpen [D:\a\1\msquic\build\windows\x64_schannel\src\test\lib\testlib.vcxproj]
ApiTest.obj : error LNK2001: unresolved external symbol LogTestFailure [D:\a\1\msquic\build\windows\x64_schannel\src\test\lib\testlib.vcxproj]
ApiTest.obj : error LNK2001: unresolved external symbol "class MsQuicApi const * const MsQuic" (?MsQuic@@3PEBVMsQuicApi@@EB) [D:\a\1\msquic\build\windows\x64_schannel\src\test\lib\testlib.vcxproj]
DataTest.obj : error LNK2001: unresolved external symbol CxPlatFree [D:\a\1\msquic\build\windows\x64_schannel\src\test\lib\testlib.vcxproj]
DrillDescriptor.obj : error LNK2001: unresolved external symbol CxPlatLogAssert [D:\a\1\msquic\build\windows\x64_schannel\src\test\lib\testlib.vcxproj]
HandshakeTest.obj : error LNK2001: unresolved external symbol "struct QUIC_CREDENTIAL_CONFIG ClientCertCredConfig" (?ClientCertCredConfig@@3UQUIC_CREDENTIAL_CONFIG@@A) [D:\a\1\msquic\build\windows\x64_schannel\src\test\lib\testlib.vcxproj]
HandshakeTest.obj : error LNK2001: unresolved external symbol "struct QUIC_CREDENTIAL_CONFIG ServerSelfSignedCredConfigClientAuth" (?ServerSelfSignedCredConfigClientAuth@@3UQUIC_CREDENTIAL_CONFIG@@A) [D:\a\1\msquic\build\windows\x64_schannel\src\test\lib\testlib.vcxproj]
HandshakeTest.obj : error LNK2001: unresolved external symbol CxPlatRandom [D:\a\1\msquic\build\windows\x64_schannel\src\test\lib\testlib.vcxproj]
QuicDrill.obj : error LNK2001: unresolved external symbol CxPlatSocketCreateUdp [D:\a\1\msquic\build\windows\x64_schannel\src\test\lib\testlib.vcxproj]
```

jeremyong:
Thanks I must have somehow broken it in a recent commit. I’ll take a look
after my work say

jeremyong:
Oh I just had a thought though. My guess is that certain libraries before built as static libraries but are now being built as shared libraries with the new setting. The fix may just be to fix certain libraries to be `STATIC`?

nibanks:
Progress! Now, just openssl variant fails:
```
LINK : fatal error LNK1181: cannot open input file 'OUTPUT_DIR-NOTFOUND\OpenSSL_Crypto.lib' [D:\a\1\msquic\build\windows\x64_openssl\src\bin\msquic_lib.vcxproj]
```

jeremyong:
Yep I know why too, on the CI server the OpenSSL libs are cached so there's no "archive" directory associated with them. I just need to detect the fact that this is running a CI build and substitute the appropriate path.

jeremyong:
Here goes nothing, this is probably my last and only attempt before I have to resume my day job :)

nibanks:
All the latest failures were because the static and dynamic linking build were copying over each other. I just pushed a change to fix that. There is still one code change necessary though. In static link mode, all the tools that are using msquic need to call load/unload in static mode.

jeremyong:
Yea I actually made that realization also around the time I made that lunchtime change. Thanks for addressing the CI issues, if you don't get around to it I'll give it a go after dinner today

nibanks:
Another issue I just noticed when looking at the artifacts on the CI. gtest.dll is generated as a separate binary now. We want that statically linked still please.

jeremyong:
To make gtest linked as a static library, I took the same approach as the previous OpenSSL PR and used `FetchContent` to import it as a local independent module (to which I could thread `CMAKE_ARGS` directly). Hopefully, the needed load/unload calls are happening on all the tools in a static link scenario now also.

nibanks:
The latest run seemed to have more, not less dynamic gtest libraries (now there's a mock one):

image

Also, I just tried downloading `quicinterop.exe` and running it and it failed (probably because of the load issue):

```
> quicinterop.exe -target:picoquic
MsQuicOpen failed, 0x8007139f!
```

jeremyong:
I won't be able to get to this until the evening again I'm afraid, but I don't remember seeing those libs (I believe you though, may be version differences or CMake differences). GMock in particular surprises me because that is explicitly disabled. The only thought that comes to mind is that the Googletest now believes it is a "root" project so it may behave differently although I actually read through their cmake files to verify that this wouldn't be an issue. As for the tools, I'll have to revisit them, in particular the ones I didn't really know how to use. For a number of them, they did cxplat initialization and deinitialization, so someone may want to do a pass to see if what's there interacts correctly with those routines.

nibanks:
Yeah, it's going to be a bit of a pain to make sure every tool works, but practically, every caller of `MsQuicOpen` will need the extra call when statically linking.

src/test/bin/winkernel/control.cpp
```diff
@@ -79,6 +79,10 @@ QuicTestCtlInitialize(
     WDF_IO_QUEUE_CONFIG QueueConfig;
     WDFQUEUE Queue;
 
+#ifdef QUIC_BUILD_STATIC
```
ThadHouse, line 82: This change actually isn't necessary, because we will never support kernel mode statically built. There are many more issues with doing that.

src/test/bin/winkernel/driver.cpp
```diff
@@ -68,7 +68,9 @@ void __cdecl operator delete[] (_In_opt_ void* Mem) {
     }
 }
 
+#ifndef QUIC_BUILD_STATIC
```
ThadHouse, line 71: This change actually isn't necessary, because we will never support kernel mode statically built.

src/perf/bin/drvmain.cpp
```diff
@@ -116,7 +116,9 @@ void __cdecl operator delete[] (_In_opt_ void* Mem, _In_opt_ size_t) {
     }
 }
 
+#ifndef QUIC_BUILD_STATIC
```
ThadHouse, line 119: This change actually isn't necessary, because we will never support kernel mode statically built. There are many more issues with doing that.

src/tools/interopserver/InteropServer.h
```diff
@@ -43,13 +43,24 @@ const QUIC_BUFFER QuackAckBuffer = { sizeof("quack-ack") - 1, (uint8_t*)"quack-a
 //
 // Exits if there is a failure.
 //
+#ifdef QUIC_BUILD_STATIC
 #define EXIT_ON_FAILURE(x) do { \
     auto _Status = x; \
     if (QUIC_FAILED(_Status)) { \
        printf("%s:%d %s failed!\n", __FILE__, __LINE__, #x); \
+       MsQuicUnload(); \
```
ThadHouse, line 51: Since this is not going to also cause an MsQuicClose, this is going to assert, as the library is still open. We might need to use some kind of AtExit handler here, or figure out another way to safely exit the app.

jeremyong:
Thanks for the feedback. Good news is the work week is over and I can do this more carefully now

jeremyong:
The last commit removes all kernel dep mods since that's always a dll-load scenario. The invocations of `MsQuicUnload` now happen in `atexit` handlers which seems to work locally and is a bit more robust (the handler is installed only after the load completes).

One thing I could not reproduce is the gtest/gmock dlls. Are we certain those artifacts weren't stale files left there from earlier builds?

nibanks:
We should never have stale files. Always generate fresh.

jeremyong:
> We should never have stale files. Always generate fresh.

Hmm ok, I'll keep thinking but without being able to repro this locally I'm somewhat at a loss. Having read through the entire google test cmake files as well, nothing immediately jumps out to me as where those artifacts could be coming from. I'm not seeing any gtest related libs when invoking the build script locally.

jeremyong:
I believe the gtest/gmock dlls should be gone now. The issue appears to be a CMake behavior I wasn't aware of. Apparently, the `BUILD_SHARED_LIBS` arg passed to a `FetchContent` dependency cannot shadow the value in the parent context. The solution is to force assign the value to `OFF` before reverting it to its previous value.

nibanks:
Latest CI run:

- Googletest binaries are gone!
- quicinterop.exe still fails to start

```
> quicinterop.exe -target:msquic
MsQuicOpen failed, 0x8007139f!
```

jeremyong:
Lol sorry looks like I missed on indeed. If this was a good first task I hate to see what a good second task looks like lolol

nibanks:
> Lol sorry looks like I missed one indeed. If this was a good first task I hate to see what a good second task looks like lolol

Indeed. The issue may have been a bit mislabeled. I didn't realize how much work it would entail. Thanks a lot for doing it though!

jeremyong:
Oh no problem at all it's been fun hacking around the library more generally. I'm mainly a graphics engineer by trade so just reading the code I've been learning a fair bit.

nibanks:
Good news! Interop works now with statically built openssl quicinterop.exe:
```
> quicinterop.exe -target:picoquic

      TARGET  VHDCRZSQUMBG2    VERSION   ALPN
 ============================================
    picoquic  VHDCRZSQUMBG2  0x00000001  hq-interop

Total execution time: 10.640s
```

src/bin/CMakeLists.txt
```diff
@@ -7,11 +7,199 @@ else()
...
+    target_compile_definitions(msquic_static PUBLIC QUIC_BUILD_STATIC)
+    set_property(TARGET msquic_static PROPERTY FOLDER "${QUIC_FOLDER_PREFIX}libraries")
+
+    # We need to take given library and walk all its dependencies, including
+    # transient dependencies. For each dependency, if it either compiles
+    # directly to or imports a static library, that library needs to be
+    # added to the QUIC_STATIC_LIBS list. This list is ultimately fed into
+    # the archiving utility to produce a monolithic static library as
+    # static library flags
+    set(EXCLUDE_LIST "ws2_32;schannel;ntdll;bcrypt;ncrypt;crypt32;iphlpapi;advapi32")
```
nibanks, line 25: Is there a way to get a list of all the libraries that get included, but aren't in the exclude list already?

nibanks: I guess I can compile it locally and uncomment the line below...

nibanks (approved):
I'd like to get @thadhouse to make a detailed review. He's our resident cmake expert. IMO, this adds a lot of complexity and is beyond my ability to review. I wish it had been simpler, but I don't know enough to say if there is a simpler way. I did just find that the tests aren't working with this change, because they didn't get updated for the MsQuicLoad call, but I'm fine with that for now. We can follow up in a separate PR. I might also take a stab and trying to find a "creative" solution to synchronizing first calls to MsQuicOpen (and last calls to MsQuicClose) to lazily call MsQuicLoad automatically. But I don't want to complicate this PR any further.

src/bin/CMakeLists.txt
jeremyong: Yea if you want we can emit that log always. There are a few lines I left in there to make it easier to debug.

jeremyong:
I agree that a future change that just initializes the library on first open in a threadsafe manner feels preferable to me. In hindsight it would have certainly made the surface area of this change smaller.

src/bin/CMakeLists.txt
jeremyong: One comment I can make about the implementation of this function btw is that the reason we use `GLOBAL PROPERTY` registers to track the data is because of how lexical scope rules work in CMake w.r.t. function recursion.

ThadHouse: Since these libraries are listed 3 times (once in inc, and twice in this file) I'd like to see if there is any way we can unify them. That we if we ever have to update 1, we update all 3 at the same time without missing any.

jeremyong:
Actually @nibanks should I just make this change? Honestly I don't mind trying to "do things right" if we believe that's a better direction. No sense changing the public API only to have to change it again later.

src/bin/CMakeLists.txt
jeremyong: The change I was about to make was to simply exclude `inc` which I believe might work?

ThadHouse: In theory yes that should.

nibanks:
> Actually @nibanks should I just make this change? Honestly I don't mind trying to "do things right" if we believe that's a better direction. No sense changing the public API only to have to change it again later.

Perhaps, but how were you planning on doing it. For Load, I was thinking about something like this:

```C
uint32_t Loaded = 0; // 0 = Unloaded, 1 = Loading, 2 = Loaded
void Load() {
  // ...
}
void Open() {
  uint32_t PrevLoaded;
  while ((PrevLoaded = InterlockedCompareExchange(&Loaded, 0, 1)) == 1) {
    // Someone else is in the process of loading
  }
  if (PrevLoaded == 0) {
    Load()
    InterlockedCompareExchange(&Loaded, 1, 2);
  } else {
    DBG_ASSERT(PrevLoaded == 2);
  }
  // ...
}
```

I'd have to think about how unload would factor in. Thoughts?

jeremyong:
I was planning on ref-counting the number of open/close close calls, so similar to your implementation but with a full ref count in addition to the disparate loading states. The first person to atomically increment the count from zero is responsible for loading the library. Every one else spins until the library is loaded. One minor change I would make is adding a yield instruction in that while loop to avoid overly locking that core which might be needed to initialize an affinitized worker in the thread pool. The last person to close the library (atomic decrement from 1) tears everything down.

src/bin/CMakeLists.txt
jeremyong: OK that worked. Also you can see the full final link command by looking at the contents of the file `src\bin\Debug\quicdeps.txt` for debug builds for example from the build root.

jeremyong:
One question I had, the above implementation makes some assumptions about how `MsQuicOpen` and `MsQuicClose` are used. In particular, if we anticipate races not just between separate invocations of `MsQuicOpen` and `MsQuicClose` _independently_, but also anticipate that invocations of `MsQuicOpen` and `MsQuicClose` could be concurrent, the CAS loop because significantly more complicated and I would recommend an atomic plus mutex in this case. The issue in the second scenario is that the library could be unloading while an open call is performed. I imagine it's possible to make this completely lock free with a really fancy CAS loop and packing loading/unloading bits in with the ref count (in the high bits), but this probably should be avoided unless we know it's a requirement. One other thing for consideration is that the mutex won't actually add much overhead since the majority of the time it will be uncontended (only used to guard against load/unload races).

nibanks:
The dynamic linking world we definitely have to account for all possible races for the service host scenarios. Multiple independent dlls coming and going in the same process space, that might depend on MsQuic. For statically linking, I'm not sure the same holds. If expect the code statically linking MsQuic to know about anything else also using it, but it's theoretically possible it doesn't. One case: an app uses MsQuic directly and a DNS over QUIC library that also happens to use MsQuic. Everything is statically linked. Depending on how things get initialized, it's possible there are races between the app and the DNS code. Thoughts?

jeremyong:
Here's a quick sketch of a lockfree implementation to give you an idea of what I'm thinking (I wrote after forgetting that stdatomic.h isn't available on MSVC)

```c
#ifdef QUIC_BUILD_STATIC
// This variable has a bit layout according to the following scheme:
// MSB [Unloading:1] [Loading:1] [RefCount:30] LSB
_Atomic unsigned int LoadState = 0;
#define UNLOADING_BIT (1 << 31)
#define LOADED_BIT (1 << 30)
#define REFCOUNT_MASK (LOADING_BIT - 1)
#endif // QUIC_BUILD_STATIC

QUIC_STATUS
QUIC_API
MsQuicOpen(
    _Out_ _Pre_defensive_ const QUIC_API_TABLE** QuicApi
    )
{
#ifdef QUIC_BUILD_STATIC
    unsigned int PriorState = atomic_fetch_add(&LoadState, 1);
    unsigned int PriorCount = PriorState & REFCOUNT_MASK;

    if (PriorCount == 0) {
        // We're responsible for ensuring the library is loaded

        // Spin until we aren't unloading the library
        while ((PriorState & UNLOADING_BIT) != 0) {
            thrd_yield();
            PriorState = atomic_load(&LoadState);
        }

        // NOTE: Because this open call has atomically incremented the
        // ref count already, it's impossible for the ref count to dip
        // to zero at this time

        // This expression can be FALSE if we arrived here prior to the
        // unloading process starts. In this case, we proceed without
        // further operation.
        if ((PriorState & LOADED_BIT) != 0)
        {
            CxPlatSystemLoad();
            MsQuicLibraryLoad();

            // Try and transition to the loaded state
            while (atomic_compare_exchange_strong(&LoadState, &PriorState, PriorState | LOADED_BIT)) {
                thrd_yield();
            }
        }
    } else {
        // Busy wait until we aren't in a loading state anymore
        while ((PriorState & LOADED_BIT) != 0) {
            thrd_yield();
            PriorState = atomic_load(&LoadState);
        }
    }
#endif // QUIC_BUILD_STATIC
   // Rest of MsQuicOpen code
```

And for the close

```c
void
QUIC_API
MsQuicClose(
    _In_ _Pre_defensive_ const QUIC_API_TABLE* QuicApi
    )
{
    if (QuicApi != NULL) {
        QuicTraceLogVerbose(
            LibraryMsQuicClose,
            "[ api] MsQuicClose");
        CXPLAT_FREE(QuicApi, QUIC_POOL_API);
        MsQuicRelease();
    }

#ifdef QUIC_BUILD_STATIC
    unsigned int PriorState = atomic_fetch_sub(&LoadState, 1);

    if (PriorCount == 1) {
        // We are responsible for unloading the library IFF another thread
        // doesn't open the library in the meantime

        // Attempt to transition to the unloading state. If we fail, that
        // means another thread has already incremented the ref count before
        // we arrived at this instruction. Thus, we don't bother unloading
        // the library
        if (atomic_compare_exchange_strong(&LoadState, &PriorState, UNLOADING_BIT))
        {
            // We've succeeded. Unload the library
            MsQuicLibraryUnload();
            CxPlatSystemUnoad();
        }
    }
#endif // QUIC_BUILD_STATIC
}
```

jeremyong:
Don't hate me. I pushed a change with the above (with minor corrections and ported to use Interlocked intrinsics instead). Happy to revert it if this is a bridge too far for this PR but wanted to at least post it for evaluation/review

src/core/library.c
```diff
@@ -1176,13 +1176,65 @@ QuicLibraryGetParam(
     return Status;
 }
 
+#ifdef QUIC_BUILD_STATIC
```
jeremyong, line 1179: The unquestionably tricky code is in this file. Feedback/critique appreciated

src/bin/CMakeLists.txt
jeremyong: I resolved this by removing inc (and warnings) from the intermediate target that globbed the libraries. Since there's no code in this target, we can remove warnings. The inc target meanwhile is included in the final `msquic` target. This meant I could remove the entire `EXCLUDE_LIST` variable altogether.

src/core/library.c
```diff
@@ -1756,3 +1829,4 @@ QuicLibraryEvaluateSendRetryState(
             NewSendRetryState);
     }
 }
+
```
nibanks, line 1832: nit: remove extra newline

jeremyong:
And I broke the release build. One moment

jeremyong:
@nibanks No pressure but if you want to give me permission to run the `/azp run` command, I don't mind babysitting this until everything passes (can revoke permission after).

jeremyong:
The current build break is occurring because on some reason, on the CI machine, platform.lib is being linked too late. The only thing that changed is that `msquic_static` changed from a `STATIC` target to an `INTERFACE` target so I'm going to add an empty "dummy.c" file and change it back to see if that resolves things.

jeremyong:
I'm typing this up as part of the record in an attempt to document how the lock-free init/deinit sequence works. This analysis assumes all interlocked operations are fully consistent (as opposed to release/acquire semantics which would complicate the analysis considerably).

There are 5 events that occur:
1. Ref count increment
2. Ref count decrement
3. Transition to "loaded"
4. Transition to "unloading"
5. Transition from "unloading"

A given thread can either `open` or `close` the library resulting in one of the two following timelines: increment -> loaded, decrement -> unloading -> unload unset.

Here are all the possible reorderings that can occur that are permissible even with atomic consistency:

## Case 1

From the initial state (ref count 0, not loaded)

1. Ref count increment (open called on `t_0`)
2. Ref count increment/decrement N times (open/close called 0+ times)
3. Transition to "loaded" *after* `t_0` finishes the loading sequence
4. Threads `t_n` for n > 0 are allowed to proceed

Decrementing in this scenario cannot set ref count to 1 because `t_0` holds it at 1 through step 3. This is the most straightforward case to analyze.

## Case 2

From a loaded state (ref count > 0, loaded)

1. Ref count decrement is called (close called on `t_0`)
2. Ref count increment (open called on `t_1`)
3. Transition to unloading on `t_0` fails (unloading routine never starts)
4. `t_1` proceeds with loading but notices the "loaded" bit remains set so no load is necessary

Few notes about this case. The transition to the unloading state fails because the ref count increment has changed the LSB and the compare exchange swap to the naked set unloading bit fails. This compare-exchange is a single attempt operation. Thread 0 thus doesn't even bother unloading because it knows someone else has requested the API table in the meantime.

## Case 3

From a loaded state (ref count > 0, loaded)

1. Ref count decrement is called (close called on `t_0`)
2. Transition to unloading on `t_0` succeeds (unloading starts)
3. Ref count increment (open called on `t_1`)
4. `t_1` now spins while the unloading bit is set
5. Unloading finishes on `t_0` and the unloading bit is unset
6. `t_1` resumes and because the loaded bit was unset in step 2, loading proceeds
7. Loading finalizes, the loading bit is set, and all threads that invoked open in the meantime are allowed to continue

In this timeline, an unloading thread starts the unloading procedures before another thread advertises that it would like to use the library. In this case, the unloading bit which is set atomically is used to block the incoming thread from proceeding with the load until the unload procedure is complete.

## Case 4

From a loaded state (ref count > 0, loaded)

1. Ref count decrement is called (close called on `t_0`)
2. Transition to unloading on `t_0` succeeds (unloading starts)
3. Unload finishes and unloading bit is unset

This case decays now into the same situation as case 1.

## Case 5

Ref count is 0, but in a "loaded" state.

This occurs just after a thread closes the library but before the unloading bit is set. Note that this initial condition is the same as step 2 in case 2.

## Case 6

Ref count > 0 but in an "unloading" state.

This occurs just after a thread invokes the open call and after a previous thread sets the unloading bit. This set of initial conditions is the same as step 4 in case 3.

nibanks:
@jeremyong Thanks again for all the work so far. With something this complex, we really need tests to make sure it is 100% bug free. I think a moving all the related code into separate functions that can be explicitly tested via a unit test is appropriate. Part of me wonders if we can break this PR up into smaller, more easily testable sub-PRs. Thoughts? Maybe just a PR that adds this new functionality (with the unit tests). It will be wrapped the the static #ifdef, but until the next PR (this one) that'd be unused outside of the unit test.

src/core/library.c
```diff
@@ -1176,13 +1176,64 @@ QuicLibraryGetParam(
     return Status;
 }
 
+#ifdef QUIC_BUILD_STATIC
+// This variable has a bit layout according to the following scheme:
```
nibanks, line 1180: nit: our comment syle always have prefix/postfix lines of `//`
```c
//
// This variable has a bit layout according to the following scheme:
// MSB [Unloading:1] [Loading:1] [RefCount:30] LSB
//
```

src/tools/interop/interop.cpp
```diff
@@ -121,7 +121,9 @@ std::vector<std::string> Urls;
 
 const char* SslKeyLogFileParam = nullptr;
 
+#ifndef QUIC_BUILD_STATIC
 extern "C" void QuicTraceRundown(void) { }
+#endif
```
nibanks, line 126: I've been meaning to do this, but I think I'm going to modify the tracing abstraction a bit to make these unnecessary for things that don't need a rundown.

nibanks: https://github.com/microsoft/msquic/pull/1462

jeremyong:
That's a completely reasonable view! I'll think about how best to do that but may need to work on other things for the moment. In the meantime, I will have to think about testability. The simpler option as I mentioned before may be to just use a mutex, unless we think open/close calls are perf sensitive. Then we can always check in a lockfree version later once testing is adequately implemented.

src/tools/interop/interop.cpp
jeremyong: Once #1462 is in, I can open a new PR with lock-based synchronization for init/deinit and all the CMake code as a less risky incremental checkin.

nibanks: @jeremyong it's in! Feel free to merge main.

nibanks: Just merged it.

nibanks:
I took a quick look at this, and it really reminding me of the existing `CXPLAT_RUNDOWN_REF` interface (at least on Windows). I wonder if that could be extended to support this scenario...

ThadHouse:
I'm good with the CMake changes here. Just a few changes related to Nick's comments, and then if we want to look into the CXPLAT_RUNDOWN_REF change, but I'm good otherwise.

ThadHouse (approved):

jeremyong:
Thanks for the eyes on this fellas, I'll take a look at what `CXPLAT_RUNDOWN_REF` does in the evening if nobody gets to it first

jeremyong:
FYI I plan on circling back to this soon (got busy with work). Looks like the PR needs to be rebased and updated. Has there been more thought put into how we want to implement the init/deinit procedures?

nibanks:
> FYI I plan on circling back to this soon (got busy with work). Looks like the PR needs to be rebased and updated. Has there been more thought put into how we want to implement the init/deinit procedures?

No, I haven't had much time to look into this. Like I said, I think splitting the change out and adding unit tests would be the best way forward. If you can get the lazy instantiation stuff solid, it might be best just to use it for both static and dynamic linking. If you can put that in a PR of it's own first, that'd be great.

nibanks:
@jeremyong any timeline on when you might finish this up? I've had some other interest in support for this and I need to know if we need to pick this up and finish it ourselves or not. Thanks!

jeremyong:
Sorry for being so MIA, life got far more busy on my side so I can't commit to anything immediately. :( Hopefully it isn't too far off

nibanks:
@jeremyong would you mind integrating the latest `main` branch please?

jeremyong:
Yea I actually tried a bit last night but was losing some context on what had changed so ran into a number of compile issues etc. I'll try again today

nibanks:
> Yea I actually tried a bit last night but was losing some context on what had changed so ran into a number of compile issues etc. I'll try again today

You _should_ be able to just replace most (all?) your library.c changes with what's in `main` now I think.

jeremyong:
Thanks, what if anything needs to change with the various tool and runtime files? The loading is fully lazy now so all those modifications should be reverted correct?

nibanks:
> Thanks, what if anything needs to change with the various tool and runtime files? The loading is fully lazy now so all those modifications should be reverted correct?

Yep. If you still aren't sure, feel free to push whatever you have and we can go from there.

jeremyong:
Thanks I think I've merged the changes successfully but would need an "azp run" to be sure.

jeremyong:
> msquic.lib(library.obj) : error LNK2005: MsQuicOpen already defined in perflib.lib(SecNetPerfMain.obj) [D:\a\1\msquic\build\windows\x64_openssl\src\perf\bin\secnetperf.vcxproj]

Seeing these sorts of build issues (will look after lunch)

ThadHouse:
That's going to be caused by one of our workarounds to make versioning work with backwards compatibility.. hmm let me think for a bit.

ThadHouse:
Yeah, its specifically a bad interaction between static linking and one of the shims. Can you try the following?

Replace https://github.com/microsoft/msquic/blob/main/src/inc/msquic.h#L1300 with

```
#ifdef WIN32
__forceinline
#else
```

That should have windows force inline that function, which should work around this issue.

jeremyong:
Hmm should that be `_MSC_VER` instead of `WIN32` in case we want to support clang/gcc on windows?

ThadHouse:
There's a ton that's broken in clang on windows currently. We would need to do a large pass to get things working. So for now we'll just keep it Win32

src/inc/msquic.h
```diff
@@ -1297,7 +1297,9 @@ MsQuicOpenVersion(
 #if defined(__cplusplus) || defined(WIN32)
 
 _IRQL_requires_max_(PASSIVE_LEVEL)
-#ifndef WIN32
```
nibanks, line 1300:
```c
#ifdef WIN32
__forceinline
#else
__attribute__((always_inline)) inline
#endif
QUIC_STATUS
MsQuicOpen(
    _Out_ _Pre_defensive_ const QUIC_API_TABLE** QuicApi
    )
```

nibanks:
Woot! I was just able to download `quicreach.exe` from the CI artifacts and it worked great!
```
> C:\Users\nick\Downloads\quicreach.exe -server:google.com

google.com:443:

                      smb  unreachable
               hq-interop  unreachable
                    hq-29  unreachable
                    h3-29  unreachable
  0x00000001           h3    reachable
```
No msquic.dll required!

.azure/azure-pipelines.ci.yml
```diff
@@ -99,6 +99,24 @@ stages:
       tls: openssl
       config: Release
       extraBuildArgs: -EnableTelemetryAsserts
+  - template: ./templates/build-config-user.yml
```
nibanks, line 102: I just noticed we didn't enable any static builds for Linux. Should Linux work with this PR? If so, we should enable a build for it to make sure.

nibanks: Never mind. When reviewing the files below, I see Linux is not supported yet.

jeremyong:
Regarding linux support, it shouldn't be hard to support but I don't currently have a machine to test it. The main thing I think needs to be added is the use of the archiver tool to produce the final library artifact (instead of `link.exe`, use whatever the corresponding archiver is for clang/gcc etc). I think cmake exposes a variable for this but I wasn't able to test it.

ThadHouse (approved):

PR #1447, opened 2021-04-07
Transport Close logging improvement

### LINKED ISSUE #1256: Incomplete Log Info for Rejecting Connections
### Describe the bug

When "low volume" logging is enabled on a Windows machine, and a connection is rejected (no matching ALPN for instance) there isn't a lot of information as to why.

```
[1]0000.0000::2021/02/19-09:12:32.872982300 [Microsoft-Quic][conn][0xFFFFA70A825C2460] Created, IsServer=1, CorrelationId=4439
[1]0000.0000::2021/02/19-09:12:32.872984500 [Microsoft-Quic][conn][0xFFFFA70A825C2460] Version = 16777216
[1]0000.0000::2021/02/19-09:12:32.872984800 [Microsoft-Quic][conn][0xFFFFA70A825C2460] New Local IP: 10.0.1.4:443
[1]0000.0000::2021/02/19-09:12:32.872985200 [Microsoft-Quic][conn][0xFFFFA70A825C2460] New Remote IP: **********
[1]0000.0000::2021/02/19-09:12:32.872988700 [Microsoft-Quic][conn][0xFFFFA70A825C2460] Registered with 0xFFFFA70A7B4FBF20
[1]0000.0000::2021/02/19-09:12:32.872989200 [Microsoft-Quic][conn][0xFFFFA70A825C2460] Assigned worker: 0xFFFFA70A7B5E57D0
[1]0004.106C::2021/02/19-09:12:32.873049400 [Microsoft-Quic][conn][0xFFFFA70A825C2460] Initialize complete
[1]0004.106C::2021/02/19-09:12:32.873063200 [Microsoft-Quic][conn][0xFFFFA70A825C2460] ERROR, No listener found for connection.
[1]0004.106C::2021/02/19-09:12:32.873065100 [Microsoft-Quic][conn][0xFFFFA70A825C2460] Transport Shutdown: 376 (Remote=0) (QS=0)
[1]0004.106C::2021/02/19-09:12:32.873101200 [Microsoft-Quic][conn][0xFFFFA70A825C2460] Shutdown complete, PeerFailedToAcknowledged=1.
[1]0004.106C::2021/02/19-09:12:32.873101900 [Microsoft-Quic][conn][0xFFFFA70A825C2460] Unregistered from 0xFFFFA70A7B4FBF20
[1]0004.106C::2021/02/19-09:12:32.873102300 [Microsoft-Quic][conn][0xFFFFA70A825C2460] Handle closed
[1]0004.106C::2021/02/19-09:12:32.873127000 [Microsoft-Quic][conn][0xFFFFA70A825C2460] Destroyed
```

From the transport shutdown event, we can see `376` (or `0x178`) is being returned to the peer, which (after looking at the code) maps to TLS alert `0x78` (or `120`) which is `NO_APPLICATION_PROTOCOL` which means there was a mismatch of ALPN requested and the ones currently supported.

It shouldn't require looking at the code to decode numeric values. We should provide more information in these failure cases in human readable text.

anrossi:
Adds textual mapping for transport close error codes to ETW traces.
Closes #1256


src/core/frame.h
```diff
@@ -86,8 +86,8 @@
 #define QUIC_ERROR_CRYPTO_ERROR(TlsAlertCode)   ((QUIC_VAR_INT)(0x100 | (TlsAlertCode)))
 #define IS_QUIC_CRYPTO_ERROR(QuicCryptoError)   ((QuicCryptoError & 0x100) == 0x100)
 
-#define QUIC_ERROR_CRYPTO_USER_CANCELED             QUIC_ERROR_CRYPTO_ERROR(22)  // TLS error code for 'user_canceled'
 #define QUIC_ERROR_CRYPTO_HANDSHAKE_FAILURE         QUIC_ERROR_CRYPTO_ERROR(40)  // TLS error code for 'handshake_failure'
+#define QUIC_ERROR_CRYPTO_USER_CANCELED             QUIC_ERROR_CRYPTO_ERROR(90)  // TLS error code for 'user_canceled'
```
anrossi, line 90: According to RFC8446, the TLS alert code for user_canceled is 90, not 22.  22 corresponds to record_overflow.

src/core/connection.c
```diff
@@ -1635,8 +1635,9 @@ QuicConnTryClose(
         } else {
             QuicTraceEvent(
                 ConnTransportShutdown,
-                "[conn][%p] Transport Shutdown: %llu (Remote=%hhu) (QS=%hhu)",
+                "[conn][%p] Transport Shutdown: %u (%llu) (Remote=%hhu) (QS=%hhu)",
```
anrossi, line 1638: ETW compiler disallows mapping on values larger than uint32, so in order to get textual mapping, we need to log the error code twice, once as a 32-bit unsigned integer to allow for mapping, and again as a 64-bit unsigned integer in case the error uses the high DWORD.

src/core/frame.h
nibanks: I'm at a loss as to where the `20` came from in the first place. Maybe just a typo. Can you please separate this out to a separate PR?

nibanks:
I'm sorry if things were unclear on #1256, but this isn't exactly what I wanted. I wanted new logs to be able to expand on this log:

```
0004.106C::2021/02/19-09:12:32.873063200 [Microsoft-Quic][conn][0xFFFFA70A825C2460] ERROR, No listener found for connection.
```

This currently means "Someone was listening to **an** IP address for this UDP port, and **a** ALPN, but one or both didn't match the incoming packet". We need a new event for when IP address doesn't match (I don't think we need additional details on what didn't match. We should have enough details from other events already). We also need a new event for when ALPN doesn't match. This is the most important part. We need to log what ALPNs were requested. Finally, we need to make sure to log ALPNs that were originally registered on/used by the listener.

A bit more, for this PR, we can't change the format of existing ETW messages, because it breaks backwards compatibility. I like the idea of converting things to text automatically, but I am sad ETW doesn't support uint64_t. I know we could handle this in CLOG, but that would require using CLOG to decode the traces, not inbox Windows tools (netsh.exe). You can check to see if our command line tool that converts our logs to text supports the nice version of these error codes already (I don't think it does).

src/core/listener.c
```diff
@@ -304,6 +304,11 @@ MsQuicListenerStart(
         Listener,
         Listener->Binding,
         CLOG_BYTEARRAY(sizeof(Listener->LocalAddress), &Listener->LocalAddress));
+    QuicTraceEvent(
+        ListenerAlpn,
+        "[list][%p] Accepting ALPNs: %!ALPN!",
+        Listener,
+        CLOG_BYTEARRAY(Listener->AlpnListLength, Listener->AlpnList));
```
nibanks, line 311: Can you just append this to the `ListenerStarted` event instead? Appending to existing events is safe to do.

src/core/connection.c
```diff
@@ -1635,7 +1635,7 @@ QuicConnTryClose(
         } else {
             QuicTraceEvent(
                 ConnTransportShutdown,
-                "[conn][%p] Transport Shutdown: %llu (Remote=%hhu) (QS=%hhu)",
+                "[conn][%p] Transport Shutdown: %u (Remote=%hhu) (QS=%hhu)",
```
nibanks, line 1638: I think you're going to have to undo this. The is a breaking change to the logs.

src/core/binding.c
```diff
@@ -382,6 +382,7 @@ _Success_(return != NULL)
 QUIC_LISTENER*
 QuicBindingGetListener(
     _In_ QUIC_BINDING* Binding,
+    _In_opt_ QUIC_CONNECTION* Connection,
```
nibanks, line 385: You're probably going to need to `UNREFERENCED_PARAMETER` for the no logging case.

src/manifest/MsQuicEtw.man
```diff
@@ -523,6 +523,132 @@
                 value="12"
                 />
           </valueMap>
+          <valueMap name="map_QUIC_TRANSPORT_SHUTDOWN_REASON">
```
nibanks, line 526: You can instead update quictrace log conversion to do this extra stuff.

src/core/listener.c
anrossi: It's not going to break the existing logs? It'll change the string table entry, which will make localization upset.

src/core/connection.c
anrossi: Forgot to revert this one. Got it now

src/core/listener.c
nibanks: We don't localize our strings. So it's fine.

src/core/binding.c
```diff
@@ -382,15 +382,18 @@ _Success_(return != NULL)
...
     )
 {
+    UNREFERENCED_PARAMETER(Connection);
     QUIC_LISTENER* Listener = NULL;
 
     const QUIC_ADDR* Addr = Info->LocalAddress;
     const QUIC_ADDRESS_FAMILY Family = QuicAddrGetFamily(Addr);
 
     BOOLEAN FailedAlpnMatch = FALSE;
+    BOOLEAN FailedAddrMatch = FALSE;
```
nibanks, line 396: I think this should actually default to `TRUE` and then set to `FALSE` if it passes the addr check below. The reason being is that if (somehow) no listeners are present in this loop, then that should be treated as no addr.

src/core/binding.c
```diff
@@ -426,7 +431,20 @@ QuicBindingGetListener(
...
 
+    if (FailedAddrMatch) {
+        QuicTraceEvent(
+            ConnNoListenerIp,
+            "[conn][%p] No Listener for IP address: %!ADDR!",
+            Connection,
+            CLOG_BYTEARRAY(sizeof(*Addr), Addr));
+    }
+
     if (FailedAlpnMatch) {
```
nibanks, line 442:
```suggestion
    } else if (FailedAlpnMatch) {
```
It shouldn't be possible to be both.

src/manifest/MsQuicEtw.man
```diff
@@ -827,6 +836,21 @@
...
+            <data
+                inType="win:UInt8"
+                name="AlpnLength"
+                />
+            <data
+                inType="win:Binary"
+                length="AlpnLength"
+                name="Alpn"
+                />
+          </template>
```
nibanks, line 853: No longer used?

src/manifest/MsQuicEtw.man
```diff
@@ -1250,12 +1274,6 @@
                 name="Connection"
                 />
           </template>
-          <template tid="tid_STREAM">
-            <data
-                inType="win:Pointer"
-                name="Stream"
-                />
-          </template>
```
nibanks, line 1258: Why was this removed?

nibanks: Oh, I see. It was moved down below.

src/manifest/MsQuicEtw.man
```diff
@@ -2650,6 +2689,24 @@
               template="tid_CONN"
               value="5179"
               />
+          <event
+              keywords="ut:Connection ut:LowVolume"
+              level="win:Informational"
```
nibanks, line 2694: I'm thinking error level instead. Thoughts? (Same for ALPN one below)

anrossi: I was matching the existing level for the log which says transport shutdown reason. But I think this qualifies as error level.

nibanks (approved):
LGTM

PR #1450, opened 2021-04-08
API proposal for portable certificate verification

wfurt:
This is straw man proposal for something I discussed offline with @nibanks.
While I'm not sure about the actual details the structure should make it easy for .NET or other languages to participate in certificate validation. 

Instead of of passing `void*` to `PEER_CERTIFICATE_RECEIVED` event we would pass in two separate values as:

`QUIC_CERTIFICATE* Certificate`
X509* with OpenSSL
PCERT_CONTEXT with Windows & Schannel 

and 
`QUIC_CERTIFICATE_CHAIN* Chain`
X509_STORE_CTX* with OpenSSL
HCERTSTORE with Windows & Schannel

With this, the `Certificate` really represents peer certificate on all platforms. 
the `Chain`  primarily represents `certificate_list` e.g. additional certificates received in TLS handshake https://tools.ietf.org/html/rfc8446#section-4.4.2 . They may or may not be important to building certificate chain and verifying peer certificate.

On top of this, MsQuic would provide helper functions to covert  QUIC_CERTIFICATE* or QUIC_CERTIFICATE_CHAIN* into well-known platform independent format. I'm not sure if `quic_cert.h` is right place but the prototype should sufficient.

This allows to consume the PEER_CERTIFICATE_RECEIVED  event directly with minimal overhead for anybody who is ready to directly dereference and use the types above. For anybody else, this allows to consume same data via TLV asn1 encoded buffer. While that adds allocation and extra processing, that should still be negligible comparing to cost of actual crypto and TLS Handshake.

If this looks OK, I can submit implementations for CxPlatCert* functions.

(optional) For the buffer, If Length & Buffer are provided, helper functions would try to use it. If Buffer is too small, QUIC_STATUS_BUFFER_TOO_SMALL would be returned and Length would be set to actual space needed. 
That would provide alternative mechanism to avoid buffer location by msquic.

If needed, the `QUIC_CERTIFICATE_FORMAT` can be possibly extended to support PEM, PKCS12 or other formats.


src/inc/quic_cert.h
```diff
@@ -132,3 +142,33 @@ CxPlatCertVerify(
...
+    _In_ QUIC_CERTIFICATE_CHAIN* CertificateChain,
+    _In_ QUIC_CERTIFICATE_FORMAT OutputFormat,
+    _Inout_ QUIC_BUFFER* OutputBuffer);
+
+
+//
+// This will free QUIC_BUFFER resources if allocated by MsQiuc.
+// This has undefined and possibly dangerous behavior if Buffer was not allocated by MsQuic.
+//
+void CxPlatQuicBufferFree(_In_ QUIC_BUFFER* Buffer);
```
nibanks, line 174: We only expose functions via the API table (in msquic.h). So these will need to be defined in there. I'd also prefer to return `QUIC_STATUS` instead of `BOOLEAN`.

src/inc/msquic.h
```diff
@@ -925,9 +926,10 @@ typedef struct QUIC_CONNECTION_EVENT {
             const uint8_t* ResumptionTicket;
         } RESUMPTION_TICKET_RECEIVED;
         struct {
-            void* Certificate;                  // Peer certificate (platform specific)
+            QUIC_CERTIFICATE* Certificate;      // Peer certificate (platform specific). Valid only during QUIC_CONNECTION_EVENT_PEER_CERTIFICATE_RECEIVED callback.
             uint32_t DeferredErrorFlags;        // Bit flag of errors (only valid with QUIC_CREDENTIAL_FLAG_DEFER_CERTIFICATE_VALIDATION)
             QUIC_STATUS DeferredStatus;         // Most severe error status (only valid with QUIC_CREDENTIAL_FLAG_DEFER_CERTIFICATE_VALIDATION)
+            QUIC_CERTIFICATE_CHAIN* Chain;      // Peer certificate chain (platform specific) Valid only during QUIC_CONNECTION_EVENT_PEER_CERTIFICATE_RECEIVED callback.
```
nibanks, line 932:
```suggestion
            QUIC_CERTIFICATE_CHAIN* Chain;      // Peer certificate chain (platform specific). Valid only during QUIC_CONNECTION_EVENT_PEER_CERTIFICATE_RECEIVED callback.
```
```suggestion
            QUIC_CERTIFICATE_CHAIN* Chain;      // Peer certificate chain (platform specific) Valid only during QUIC_CONNECTION_EVENT_PEER_CERTIFICATE_RECEIVED callback.
```

src/platform/tls_schannel.c
```diff
@@ -2298,8 +2298,11 @@ CxPlatTlsWriteDataToSchannel(
                         TlsContext->Connection,
 #ifdef _KERNEL_MODE
                         (void*)&PeerCert,
+                        NULL,
 #else
-                        (void*)PeerCert,
+                        (QUIC_CERTIFICATE*)PeerCert,
+                        (QUIC_CERTIFICATE_CHAIN*)(PeerCert ? PeerCert->hCertStore : NULL),
```
nibanks, line 2304: I _beleive_ `PeerCert` is guaranteed to be non-null here. Though, we should put a `CXPLAT_DBG_ASSERT` above for this.

wfurt: What value would it have on server if client did not provide certificate? 

For SslStream, the sever can request client certificate and when handshake is done we can call validation callback with no certificate. That gives application control to decide if that is ok or not. e.g. use the certificate opportunistically.

anrossi:
What about instead of adding a new API, you add a new flag to the `QUIC_CREDENTIAL_CONFIG` which specifies that the certificate returned should be in the platform-agnostic format (X509 DER binary blob)? Instead of a `void*` event parameter, the type would be a union of `QUIC_CERTIFICATE` and the platform-specific types.
This would be more consistent with the API design.

wfurt:
> What about instead of adding a new API, you add a new flag to the `QUIC_CREDENTIAL_CONFIG` which specifies that the

That would work as well. We would need to extend the event to also include length but that should be easy.
Any thought on adding flag @nibanks? The flow would be more natural IMHO as there would be no need to make extra call for the conversion and worry about  freeing memory.

nibanks:
> > What about instead of adding a new API, you add a new flag to the `QUIC_CREDENTIAL_CONFIG` which specifies that the
> 
> That would work as well. We would need to extend the event to also include length but that should be easy.
> Any thought on adding flag @nibanks? The flow would be more natural IMHO as there would be no need to make extra call for the conversion and worry about freeing memory.

Yeah, this might be a lot simpler. Just return `QUIC_BUFFER*` when that new flag is specified.

wfurt:
This looks better to me with the flag approach. I still need to do some integration testing with .NET but I wanted to put it out for feedback.

src/platform/tls_schannel.c
nibanks: I guess I'd expect `QueryContextAttributesW` to fail.

nibanks (commented):
Generally looks good. We need to add some tests, at least to TlsTest.cpp for the new flag. Also, since it's unsupported on Schannel currently, we need to fail calls that try to create a sec config with it OR add support to schannel.

src/platform/tls_schannel.c
wfurt: I can run more tests as well I'm fine with adding the assert as you suggested.

src/platform/tls_openssl.c
```diff
@@ -249,10 +253,47 @@ CxPlatTlsCertificateVerifyCallback(
         return FALSE;
     }
 
+    if (TlsContext->SecConfig->Flags & QUIC_CREDENTIAL_FLAGS_USE_PORTABLE_CERTIFICATES) {
+        if (Cert) {
+            PortableCertificate.Length = i2d_X509(Cert, &PortableCertificate.Buffer);
+            if (!PortableCertificate.Buffer) {
+                QuicTraceEvent(
```
nibanks, line 260: We should be bailing out in these error conditions.

src/platform/tls_openssl.c
```diff
@@ -249,10 +253,47 @@ CxPlatTlsCertificateVerifyCallback(
...
+                        TlsError,
+                        "[ tls][%p] ERROR, %s.",
+                        TlsContext->Connection,
+                        "Failed to serialize certificate context");
```
nibanks, line 264: nit: it _seems_ the spacing is off here. Am I wrong?

wfurt:
> Generally looks good. We need to add some tests, at least to TlsTest.cpp for the new flag. Also, since it's unsupported on Schannel currently, we need to fail calls that try to create a sec config with it OR add support to schannel.

yah. forget to push one more change to Schannel. It should fail now. I'm wondering if it would be worth to open issue to track the Schannel work - perhaps together with the pkcs12 support.

nibanks:
> > Generally looks good. We need to add some tests, at least to TlsTest.cpp for the new flag. Also, since it's unsupported on Schannel currently, we need to fail calls that try to create a sec config with it OR add support to schannel.
> 
> yah. forget to push one more change to Schannel. It should fail now. I'm wondering if it would be worth to open issue to track the Schannel work - perhaps together with the pkcs12 support.

Sounds like a reasonable approach to me.

src/platform/tls_openssl.c
wfurt: Updated.  The code will now fails if we fail to encode certificate. I left building the chain as soft failure e.g. we pass is much as we can to caller and let them to decide. .NET will try to download missing certificates if the chain is incomplete (or it may pick them up from cache)

src/platform/tls_schannel.c
wfurt: updated.

src/platform/tls_openssl.c
wfurt: updated. (assuming 4 spaces are expected)

wfurt:
I enabled the "portable" option for all non-windows builds on client. I feel that is more important than the raw pointes as the code is more complicated. I can come up with more comprehensive test when I work on client certificate support.

src/platform/unittest/TlsTest.cpp
```diff
@@ -96,6 +96,9 @@ struct TlsTest : public ::testing::TestWithParam<bool>
                 NULL,
                 QUIC_ALLOWED_CIPHER_SUITE_NONE};
             CredConfig.Flags |= CredFlags;
+#ifndef _WIN32
+            CredConfig.Flags |= QUIC_CREDENTIAL_FLAGS_USE_PORTABLE_CERTIFICATES;
+#endif
```
nibanks, line 101: I'd rather have explicit testing for both scenarios and not just have each platform test something else. Also, the code below (`OnPeerCertReceived`, which doesn't run for kernel) should be able to validate that both the cert and chain are non-null. Can you please make those updates?

wfurt: Updated. I added new test guarded by QUIC_DISABLE_PORTABLE_CERTIFICATE_TESTS and I added non-null check. 

I also look and the `SecPkgContext_Certificates` and it is actually the whole certificate chain. So I decided to pass same pointer as `Chain` instead of keeping it NULL.

nibanks, line 101: Remove this now that we have a test explicitly for it?

src/platform/tls_schannel.c
```diff
@@ -2294,12 +2298,16 @@ CxPlatTlsWriteDataToSchannel(
                         "Query peer cert");
                 }
 
+                CXPLAT_DBG_ASSERT(PeerCert != NULL);
```
nibanks, line 2301: This needs to be in the !KERNEL block.

nibanks (approved):
LGTM. Let's see what the CI says.

src/inc/msquic.h
```diff
@@ -263,7 +264,8 @@ typedef struct QUIC_CERTIFICATE_PKCS12 {
     const char *PrivateKeyPassword;     // Optional: used if provided. Ignored if NULL
 } QUIC_CERTIFICATE_PKCS12;
 
-typedef void QUIC_CERTIFICATE; // Platform specific certificate context object
+typedef void QUIC_CERTIFICATE; // Platform specific certificate object
```
anrossi, line 267: Nit: Align comments

src/platform/tls_openssl.c
```diff
@@ -231,6 +232,9 @@ CxPlatTlsCertificateVerifyCallback(
     X509_STORE_CTX *x509_ctx
     )
 {
+    int status = TRUE;
+    QUIC_BUFFER PortableCertificate = { 0, 0};
+    QUIC_BUFFER PortableChain = { 0, 0};
```
anrossi, line 237: Nit: spacing on the curly braces should be symmetrical.

src/platform/tls_openssl.c
```diff
@@ -249,10 +253,48 @@ CxPlatTlsCertificateVerifyCallback(
...
+                    TlsContext->Connection,
+                    "Failed to serialize certificate context");
+                return FALSE;
+            }
+        }
+        if (x509_ctx) {
+            int ChainCount;
+            STACK_OF(X509)* Chain = X509_STORE_CTX_get0_chain(x509_ctx);
+            if ((ChainCount = sk_X509_num(Chain)) > 0) {
+                PKCS7* p7 = PKCS7_new();
```
anrossi, line 272: Is PKCS7 more commonly used in these scenarios, and easier to consume than an array of `QUIC_BUFFER` which each represent one X509 certificate in the chain?

src/platform/tls_schannel.c
```diff
@@ -2293,13 +2297,18 @@ CxPlatTlsWriteDataToSchannel(
...
                         TlsContext->Connection,
 #ifdef _KERNEL_MODE
-                        (void*)&PeerCert,
+                        (QUIC_CERTIFICATE*)&PeerCert,
+                        (QUIC_CERTIFICATE_CHAIN*)&PeerCert,
 #else
-                        (void*)PeerCert,
+                        (QUIC_CERTIFICATE*)PeerCert,
+                        (QUIC_CERTIFICATE_CHAIN*)(PeerCert->hCertStore),
+
```
anrossi, line 2311: Nit: extra line.

anrossi:
Overall looks great! Just one question above about the use of PKCS7, versus just providing the chain certificates as raw x509 bytes (like the leaf cert).

src/platform/tls_openssl.c
wfurt: I don't have any stats. But it is certainly easier to consume. OpenSSL and Windows have already functions for import/export and for .NET this is one line with `X509Certificate2Collection.Import()`. I would hope it would be similar for other languages. 
Array of `QUIC_BUFFER` could possibly work as well but then you have probably more allocations and buffers to manage.

src/platform/unittest/TlsTest.cpp
wfurt: yes, I was meaning to but forget to do do it. updated.

src/platform/tls_openssl.c
anrossi: Ok, that sounds good to me.

anrossi (approved):

PR #1455, opened 2021-04-09
Configure IdealProcessor as part of SendData

ThadHouse:
When passing IdealProcessor down to CxPlatSocketSend directly, we got some odd performance regressions on windows kernel. We might be spilling parameters onto the stack, so adjust to set the ideal processor on the send data.


ThadHouse:
This isn't really fixing anything, was more just  a test, so closing.

PR #1459, opened 2021-04-10
Disable async fiber openssl support

ThadHouse:
Theres a possiblity this is causing issues, switch it off in all cases on Windows


anrossi (approved):

ThadHouse:
This was mainly to see if we got some of our RPS perf back from when we switched to sends on threads. We didn't, so closing, this isn't worth it otherwise.

PR #1460, opened 2021-04-10
Disable UDP Send Queuing (Temporarily)

nibanks:
Updates the UDP send logic to only queue to the worker thread if it's not actively doing something else. This is an attempt to stabilize perf. Note, it's possible this introduces reordering, which I have to think about.

UPDATE: Adds a build flag to control the UDP send queuing logic in general. Right now, disabling it because we have some other perf instability issues we need to debug before we can re-enable this.

### DIFF
--- a/src/platform/datapath_winuser.c
+++ b/src/platform/datapath_winuser.c
@@ -40,6 +40,12 @@ CxPlatFuzzerRecvMsg(
 #pragma warning(disable:4116) // unnamed type definition in parentheses
 
 //
+// If set, CXPLAT_DATAPATH_QUEUE_SENDS enables logic to queue the UDP send
+// to the datapath worker thread instead of inline execution.
+//
+//#define CXPLAT_DATAPATH_QUEUE_SENDS 1
+
+//
 // This IOCTL allows for creating per-processor sockets for the same UDP port.
 // This is used to get better parallelization to improve performance.
 //
@@ -183,6 +189,7 @@ typedef struct CXPLAT_SEND_DATA {
     //
     WSABUF ClientBuffer;
 
+#ifdef CXPLAT_DATAPATH_QUEUE_SENDS
     //
     // The local address to bind to.
     //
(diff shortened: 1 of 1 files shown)


src/platform/datapath_winuser.c
```diff
@@ -357,6 +357,11 @@ typedef struct CXPLAT_DATAPATH_PROC {
     //
     uint16_t Index;
 
+    //
+    // Flag indicates that the worker thread is currently actively processing.
+    //
+    uint16_t WorkerActive : 1;
```
ThadHouse, line 363: Is there a reason to make this a unit16_t and a bitflag? Unless we plan on making operations with this interlocked.

nibanks: Only a bit is needed. And there was space for a `uint16_t`.

nibanks:
As is, this made perf drop even more. I think we need something a bit more complicated. Like start by always queuing the packets for a given connection to the UDP worker. If that worker ever gets to be too overloaded, then fallback to doing the rest inline for the lifetime of the connection. Still needs some experimentation.

nibanks:
So, the only consistent changes I can see by disabling UDP send queuing are:

1. User mode throughput dropped back down to ~7 Gbps (from ~7.8 Gbps).
2. User mode, openssl (default) RPS went back up to ~1 Million RPS

I'm of mixed opinion for next steps. Part of me thinks that we should make this a build time flag or link it to TLS library (enable for schannel, but disable for openssl).

maolson-msft:
Remind me: why were sends queued in the first place? Were they always queued by a worker thread [that was working on some other queued operation]?

nibanks:
> Remind me: why were sends queued in the first place? Were they always queued by a worker thread [that was working on some other queued operation]?

We recently made did some experiments and made the change to queue the UDP sends to the datapath worker thread instead of running them inline on the QUIC worker thread, because of the CPU usage imbalance between the two. QUIC worker thread was maxed out at 100%, but the datapath worker was 60% to 80% usage. When we queued the UDP sends, we effectively moved 10% CPU usage from the QUIC worker to the UDP worker, achieving a better over all balance, and then more overall throughput (+10% throughput).

ThadHouse (approved):

anrossi (approved):

PR #1462, opened 2021-04-11
Refactor Trace Rundown to Remove from Tools/Tests

nibanks:
Refactors the trace rundown interface to remove the requirement for all cxplatform consumers, by making it a dynamic function pointer instead.

### DIFF
--- a/src/platform/platform_winkernel.c
+++ b/src/platform/platform_winkernel.c
@@ -59,6 +59,7 @@ typedef struct _SYSTEM_BASIC_INFORMATION {
 uint64_t CxPlatPerfFreq;
 uint64_t CxPlatTotalMemory;
 CX_PLATFORM CxPlatform = { NULL, NULL };
+QUIC_TRACE_RUNDOWN_CALLBACK* QuicTraceRundownCallback;
 
 INITCODE
 _IRQL_requires_max_(PASSIVE_LEVEL)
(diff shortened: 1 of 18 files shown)


src/platform/platform_winkernel.c
```diff
@@ -250,11 +251,15 @@ QuicEtwCallback(
...
+    if (!QuicTraceRundownCallback) {
+        return;
+    }
+
     switch(ControlCode) {
     case EVENT_CONTROL_CODE_ENABLE_PROVIDER:
     case EVENT_CONTROL_CODE_CAPTURE_STATE:
         if (CallbackContext == &MICROSOFT_MSQUIC_PROVIDER_Context) {
-            QuicTraceRundown();
+            QuicTraceRundownCallback();
```
ThadHouse, line 262: Grab a local copy of this function pointer and check that for null rather then checking the global for null then calling. To reduce the risk an outside caller sets the pointer to null after the check but before the call.

nibanks: I _could_ do that, but I don't think we need to worry about that added complexity. I'd prefer to keep the design of "this must be set (or not) before platform load and never written to again"

nibanks: Because if we did what you're suggesting that seems to encourage changing this at runtime. But there is no synchronization, so the setter might still get a callback after they clear this.

ThadHouse: Then I think instead of exposing the function pointer directly, we should expose a function that only ever allows the variable to be set once, and not be allowed to set it to null.

nibanks: I think I prefer the current model. A function then introduces the complexity around when do I call it? I'd rather just get this in so we can unblock the static linking PR. We can iterate more on this afterwards.

ThadHouse: As long as we make a Nickel bug to fix that then I'm ok with that. But its a correctness bug, so it does needs to be fixed before the next release.

ThadHouse (approved):

PR #1471, opened 2021-04-12
Add stats logging option to perf, log debug output to file

ThadHouse:
Logging connection stats can be an easy way to debug certain issues with the performance tool. Add an option to make this work.

By default, the pwsh script only prints logs in debug mode. Instead, write all output to a file as well, so this can be later reviewed as well. Otherwise all data from step 1 would just be lost.


src/perf/lib/PerfServer.h
```diff
@@ -130,6 +130,7 @@ class PerfServer : public PerfBase {
     uint16_t Port {PERF_DEFAULT_PORT};
     CXPLAT_EVENT* StopEvent {nullptr};
     QUIC_BUFFER* DataBuffer {nullptr};
+    uint8_t PrintStats{FALSE};
     QuicPoolAllocator<StreamContext> StreamContextAllocator;
```
nibanks, line 134: Nit: space before {

nibanks (approved):

src/perf/lib/ThroughputClient.cpp
```diff
@@ -23,22 +23,23 @@ PrintHelp(
...
+        "  -ip:<0/4/6>                  A hint for the resolving the hostname to an IP address. (def:0)\n"
+        "  -encrypt:<0/1>               Enables/disables encryption. (def:1)\n"
+        "  -sendbuf:<0/1>               Whether to use send buffering. (def:0)\n"
+        "  -pacing:<0/1>                Whether to use pacing. (def:1)\n"
+        "  -timed:<0/1>                 Indicates the upload/download arg time (ms). (def:0)\n"
+        "  -upload:<####>               The length of data (or time with -timed:1 arg) to send. (def:0)\n"
+        "  -download:<####>             The length of data (or time with -timed:1 arg) to request/receive. (def:0)\n"
+        "  -iosize:<####>               The size of each send request queued. (def:%u)\n"
+        "  -tcp:<0/1>                   Indicates TCP/TLS should be used instead of QUIC. (def:0)\n"
+        "  -stats:<0/1>                 Indicates connection stats should be printed at the end of the run. (def:0)\n",
```
nibanks, line 42:
```suggestion
        "  -stats:<0/1>                 Indicates connection stats should be printed at the end of the run. (def:0)\n"
```

PR #1479, opened 2021-04-14
Pass ideal processor down to CxPlatSocketCreateUdp

ThadHouse:
Currently, the socket is attached to the currently running CPU on client side. This makes it possible for multiple workers to create sockets that should be attached to their ideal core, but actually attach to other cores. This makes it so a new client socket created is always attached to its ideal processor.


src/platform/pcp.c
```diff
@@ -187,6 +187,7 @@ CxPlatPcpInitialize(
                 &GatewayAddresses[i],
                 PcpContext,
                 CXPLAT_SOCKET_FLAG_PCP,
+                0,
```
nibanks, line 190: Maybe `(uint16_t)CxPlatProcCurrentNumber(),` instead

nibanks (approved):

nibanks:
This still seems to consistently reduce perf:
```
Running Test ThroughputUp_Windows_x64_schannel_Default
Test Run Took 00:00:03.7598255
Run 1: 7479424 kbps
Test Run Took 00:00:03.7059709
Run 2: 7512255 kbps
Test Run Took 00:00:03.6440011
Run 3: 7737261 kbps
Test Run Took 00:00:03.7368126
Run 4: 7440081 kbps
Test Run Took 00:00:03.7191252
Run 5: 7494499 kbps
Median: 7494499 kbps (-3.84%)
Remote: 7794174.2 kbps
```
Compared to:

image

7.5 Gbps is definitely a drop.

thhous-msft:
Then I think we need to abandon this PR. I think fundamentally its going to be a small drop in throughput perf, and ideally for RPS things should get spread around by the scheduler anyway.

nibanks:
I agree. We can always try to revive it at a latest date if we really want to dig down into why it's behaving this way.

PR #1490, opened 2021-04-19
Rename WPA Plugin and Exe to QuicTrace

nibanks:


ThadHouse (approved):

src/plugins/trace/dll/QuicTraceLib.csproj
```diff
@@ -13,7 +13,7 @@
     <RepositoryUrl>https://github.com/microsoft/msquic.git</RepositoryUrl>
     <PackageLicenseFile>LICENSE</PackageLicenseFile>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <AssemblyName>QuicTrace</AssemblyName>
+    <AssemblyName>QuicTraceLib</AssemblyName>
```
ThadHouse, line 16: Just remove assembly name instead. Let the csproj defaults handle it, since the project is no correctly. named.

PR #1500, opened 2021-04-20
Upgrade QuicTrace to .NET 5

ThadHouse:
.NET 5 has been out for a while, and most tools should be updated. Also gives us some nice performance boosts.

### DIFF
--- a/src/plugins/trace/exe/Program.cs
+++ b/src/plugins/trace/exe/Program.cs
@@ -234,7 +234,7 @@ namespace QuicTrace
                     Console.Write("quictrace> ");
 
                     var input = Console.ReadLine();
-                    if (input == "--exit" || input == "exit" || input == "-e")
+                    if (input == null || input == "--exit" || input == "exit" || input == "-e")
                     {
                         return;
                     }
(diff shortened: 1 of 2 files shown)


src/plugins/trace/exe/Program.cs
```diff
@@ -234,7 +234,7 @@ static void Main(string[] args)
                     Console.Write("quictrace> ");
 
                     var input = Console.ReadLine();
-                    if (input == "--exit" || input == "exit" || input == "-e")
+                    if (input == null || input == "--exit" || input == "exit" || input == "-e")
```
nibanks, line 237: Does `null` mean "User hit enter" or "User hit ctrl+c"?

ThadHouse: Null means end of file. So if user passes in a file through stdin null will be returned after the last line. Always been able to be return, but 5.0 properly annotated Console.ReadLine.

nibanks: Ok. I just want to make sure "User hit enter" doesn't exit.

ThadHouse: Nope. That will return an empty string.

nibanks (approved):

PR #1501, opened 2021-04-20
Run kernel mode perf tests with a kernel mode client

ThadHouse:


nibanks (approved):

ThadHouse:
We're going to take this, even with the drop in performance it brings. Its both realistic in real world use cases, and gives us a goal to strive towards in improving kernel mode client performance.

PR #1507, opened 2021-04-22
Update GetValue function to check for full string, and validate length

ThadHouse:
Previously, -requests: would match a check for `request` as strnicmp just checks for length. This changes ensures that the character after the check is a :, and also that the string length for the argument is acceptable.

### DIFF
--- a/src/inc/msquichelper.h
+++ b/src/inc/msquichelper.h
@@ -292,6 +292,24 @@ IsValue(
     return _strnicmp(name, toTestAgainst, min(strlen(name), strlen(toTestAgainst))) == 0;
 }
 
+inline
+bool
+GetFlag(
+    _In_ int argc,
+    _In_reads_(argc) _Null_terminated_ char* argv[],
+    _In_z_ const char* name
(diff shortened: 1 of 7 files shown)


src/inc/msquichelper.h
```diff
@@ -306,7 +306,9 @@ GetValue(
 {
     const size_t nameLen = strlen(name);
     for (int i = 0; i < argc; i++) {
-        if (_strnicmp(argv[i] + 1, name, nameLen) == 0) {
+        if (_strnicmp(argv[i] + 1, name, nameLen) == 0
+            && strlen(argv[i]) > 1 + nameLen + 1
+            && *(argv[i] + 1 + nameLen) == ':') {
```
nibanks, line 311: I think there might be places where we support `-enableFlag` style args too.

ThadHouse: Ok, it looks like we have a few in the interop code. I'll investigate.

nibanks (approved):

PR #1521, opened 2021-04-26
Execute StreamShutdown Inline on Callbacks

### LINKED ISSUE #1520: Support Immediate Stream Abort in PEER_STREAM_STARTED Event

There are scenarios (HTTP/3 GOAWAY) where the app needs to immediately reject an incoming stream. Currently `StreamShutdown` queues up that work instead of executing immediately, even on the callback thread. This opens up the app to additional notifications for the stream that they don't want/need.

We need a way for an app to immediately abort the incoming stream.

### Proposed solution

The simplest way is to just execute `StreamShutdown` inline when called on the callback.

nibanks:
Fixes #1520.

### DIFF
--- a/src/core/api.c
+++ b/src/core/api.c
@@ -926,37 +926,46 @@ MsQuicStreamShutdown(
     Connection = Stream->Connection;
 
     QUIC_CONN_VERIFY(Connection, !Connection->State.Freed);
-    QUIC_CONN_VERIFY(Connection,
-        (Connection->WorkerThreadID == CxPlatCurThreadID()) ||
-        !Connection->State.HandleClosed);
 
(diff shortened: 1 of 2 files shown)


anrossi:
Did you intend to add the clog change to this PR?

anrossi:
Also, do we have tests which exercise both paths?

nibanks:
> Did you intend to add the clog change to this PR?

Nope. Already removed.

nibanks:
> Also, do we have tests which exercise both paths?

I believe so. I first pushed this to just run existing tests. I will review the test coverage to make sure.

anrossi (approved):

PR #1522, opened 2021-04-26
Bump cobalt patch version

ThadHouse:


nibanks (approved):
Gah! I wish we had an easy way to automate this. Maybe somehow we do that on release branches.

PR #1529, opened 2021-04-28
Refactor Sent Packet Metadata Stream Ref Counting

nibanks:
We've seen from past performance measurements that sometimes the stream ref counting was a non-trivial CPU chunk when building/sending QUIC packets. This PR refactors the logic here to have a separate counter for metadata and only inc/dec stream ref count when that changes to/from zero.

### DIFF
--- a/src/core/sent_packet_metadata.c
+++ b/src/core/sent_packet_metadata.c
@@ -32,19 +32,19 @@ QuicSentPacketMetadataReleaseFrames(
 #pragma warning(push)
 #pragma warning(disable:6001)
         case QUIC_FRAME_RESET_STREAM:
-            QuicStreamRelease(Metadata->Frames[i].RESET_STREAM.Stream, QUIC_STREAM_REF_SEND_PACKET);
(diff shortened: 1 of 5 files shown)


src/core/sent_packet_metadata.c
```diff
@@ -32,19 +32,19 @@ QuicSentPacketMetadataReleaseFrames(
 #pragma warning(push)
 #pragma warning(disable:6001)
```
ThadHouse, line 33: Do we still need these warning disables here?

ThadHouse: Nevermind I thought these were different. We probably still do.

nibanks: Yes, I believe so, because SAL doesn't understand the concept of ref counted deletes.

ThadHouse (approved):

src/core/sent_packet_metadata.c
ThadHouse: 6001 was actually uninitialized false positives. So definitely still needed.

PR #1533, opened 2021-04-29
Use same RNG seed for TCP vs. QUIC random loss WAN tests.

anrossi:
This should give us a better comparison between the two protocols by dropping packets in roughly the same place in the stream.

### DIFF
--- a/scripts/emulated-performance.ps1
+++ b/scripts/emulated-performance.ps1
@@ -34,6 +34,9 @@ be in the current directory.
 .PARAMETER ReorderDelayDeltaMs
     The extra delay applied to any reordered packets in the emulated network.
 
+.PARAMETER BaseRandomSeed
+    Base seed value for the DuoNic RNG.
+
 .PARAMETER DurationMs
     The duration(s) of each test run over the emulated network.
 
@@ -78,6 +81,9 @@ param (
 
     [Parameter(Mandatory = $false)]
     [Int32[]]$ReorderDelayDeltaMs = 0,
+
+    [Parameter(Mandatory = $false)]
+    [string]$BaseRandomSeed = "",
 
     [Parameter(Mandatory = $false)]
     [Int32[]]$DurationMs = 10000,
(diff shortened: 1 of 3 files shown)


scripts/emulated-performance.ps1
```diff
@@ -276,6 +282,17 @@ if ($LogProfile -ne "None") {
...
+if ($BaseRandomSeed -eq "") {
+    for ($i = 0; $i -lt 3; $i++) {
+        $BaseRandomSeed += $(Get-Random).ToString('x8')
+    }
+    $BaseRandomSeed += $(Get-Random).ToString('x8').Substring(0,6)
+    # This gives 15 bytes of random seed, and the last byte will be
+    # the iteration count.
+}
+
+Write-Debug "BaseRandomSeed: $($BaseRandomSeed)"
```
nibanks, line 294: This likely need to be a `Write-Host` because we always want to see it.

scripts/emulated-performance.ps1
```diff
@@ -388,6 +405,10 @@ foreach ($ThisReorderDelayDeltaMs in $ReorderDelayDeltaMs) {
                 }
             }
 
+            Write-Debug "Restarting NIC"
+            Restart-NetAdapter duo?
+            Start-Sleep 5 # (wait for duonic to restart)
```
nibanks, line 410: I wonder if we can reduce this time. Thoughts? It's going to be a bigger issue now that we do it after every iteration.

scripts/emulated-performance.ps1
anrossi: Agreed

scripts/emulated-performance.ps1
anrossi: Hmm, too fast and the adapter won't be ready by the time the app starts. The duonic powershell script waits 10 seconds by default, so this is already cutting it closer.  I think the speed really depends on the machine running the test. If it's a VM, 5 seconds might be as fast as possible before running into random failures.

anrossi: After some trial with 1, 2, 3, and 4 second sleeps, I don't think we can go below 5 seconds and remain reliable.

nibanks (approved):

PR #1535, opened 2021-04-30
Average the last 5 results of runs for performance comparison

ThadHouse:
Definitely gets some interesting results if we get some massive jumps, but should make failures less of an issue if we're riding the edges of the performance ranges.

### DIFF
--- a/scripts/performance-helper.psm1
+++ b/scripts/performance-helper.psm1
@@ -561,30 +561,41 @@ function Get-TestResult($Results, $Matcher) {
     }
 }
 
-function Get-LatestCommitHash([string]$Branch) {
+function Get-LatestCommitHashes([string]$Branch) {
     $Uri = "https://raw.githubusercontent.com/microsoft/msquic/performance/data/$Branch/commits.json"
     Write-LogAndDebug "Requesting: $Uri"
     try {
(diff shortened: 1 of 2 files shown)


scripts/performance-helper.psm1
```diff
@@ -561,30 +561,42 @@ function Get-TestResult($Results, $Matcher) {
...
         $AllCommits = Invoke-RestMethod -SkipHttpErrorCheck -Uri $Uri -Method 'GET' -ContentType "application/json"
         Write-LogAndDebug "Result: $AllCommits"
-        $LatestResult = ($AllCommits | Sort-Object -Property Date -Descending)[0]
-        Write-LogAndDebug "Latest Commit: $LatestResult"
-    return $LatestResult.CommitHash
+        if ($AllCommits.Count -eq 0) {
+            return ""
+        }
+        $SortedList = $AllCommits | Sort-Object -Property Date -Descending
+        if ($SortedList.Count -lt 5) {
```
nibanks, line 574: Can we make this `5` a global var or something, so it's easily adjusted (possibly) in the future? Maybe to 10 instead.

nibanks (approved):

PR #1537, opened 2021-05-01
Add support for randomly failing memory allocations to SpinQuic.

anrossi:


src/tools/spin/spinquic.cpp
```diff
@@ -640,6 +641,18 @@ CXPLAT_THREAD_CALLBACK(ServerSpin, Context)
         }
     }
 
+    if (Settings.AllocFailDenominator > 0) {
```
nibanks, line 644: This is a good start, but as it skips everything above (same for other thread) it's not giving us code coverage on those alloc failures. Please put a `TODO` comment to move this to right after we call `MsQuicOpen` instead, and then update all these things to be failure proof.

src/tools/spin/spinquic.cpp
```diff
@@ -791,8 +818,11 @@ main(int argc, char **argv)
         TryGetValue(argc, argv, "sessions", &SessionCount);
     }
 
-    uint32_t RngSeed = 6;
-    TryGetValue(argc, argv, "seed", &RngSeed);
+    uint32_t RngSeed = 0;
+    if (!TryGetValue(argc, argv, "seed", &RngSeed)) {
+        CxPlatRandom(sizeof(RngSeed), &RngSeed);
+    }
+    printf("Using seed value: 0x%d\n", RngSeed);
```
nibanks, line 825:
```suggestion
    printf("Using seed value: 0x%u\n", RngSeed);
```
Unsigned int print, not signed.

src/tools/spin/spinquic.cpp
anrossi: Yeah, that's exactly why I put them in the loop instead of after `MsQuicOpen`.  How should we handle these failures during setup? Retry? Exit?

src/core/binding.c
```diff
@@ -720,6 +720,11 @@ QuicBindingQueueStatelessOperation(
     }
 
     QUIC_WORKER* Worker = QuicLibraryGetWorker(Datagram);
+    if (Worker == NULL) {
+        QuicPacketLogDrop(Binding, CxPlatDataPathRecvDataToRecvPacket(Datagram),
+            "Stateless worker non-existent (stateless oper)");
+        return FALSE;
+    }
```
nibanks, line 727: This shouldn't be possible. If `MsQuicLib.StatelessRegistration` is non-null, `QuicLibraryGetWorker` is required to succeed. `QuicLibraryGetWorker` just returns a ref to memory, so it can never be `NULL`, though, it's theoretically possible it's pointing to uninitialized memory if there is a bug somewhere else.
```c
    return
        &MsQuicLib.StatelessRegistration->WorkerPool->Workers[
            Datagram->PartitionIndex % MsQuicLib.PartitionCount];
```
If you encountered a problem here, then I suspect the real problem is that `MsQuicLib.StatelessRegistration` should have been cleaned up if it's initialization wasn't 100% successful.

src/core/binding.c
```diff
@@ -998,7 +1003,12 @@ QuicBindingProcessStatelessOperation(
                 (uint8_t*)&Token,
                 (uint16_t)SendDatagram->Length,
                 SendDatagram->Buffer);
-        CXPLAT_DBG_ASSERT(SendDatagram->Length != 0);
+        if (SendDatagram->Length == 0) {
+            if (!CxPlatIsRandomMemoryFailureEnabled()) {
+                CXPLAT_DBG_ASSERT(FALSE);
+            }
```
nibanks, line 1009: If we're going to keep the assert (which I'm unsure about my self) replace this with the following
```suggestion
            CXPLAT_DBG_ASSERT(CxPlatIsRandomMemoryFailureEnabled());
```

src/inc/quic_platform.h
```diff
@@ -323,6 +323,29 @@ CxPlatListPopEntry(
...
+BOOLEAN
+CxPlatIsRandomMemoryFailureEnabled(
+    )
+{
+#ifdef DEBUG
+    return CxPlatGetAllocFailDenominator() != 0;
+#else
+    return FALSE;
+#endif
+}
```
nibanks, line 347: Just define `CxPlatIsRandomMemoryFailureEnabled()` as `FALSE` if not debug mode instead.

src/tools/spin/spinquic.cpp
nibanks: Just exit if it fails I think. Let's not make things too complicated. We have a retry for this at the higher layer already I think.

.azure/templates/run-spinquic.yml
```diff
@@ -7,6 +7,7 @@ parameters:
   config: 'Debug'
   arch: 'x64'
   tls: ''
+  allocFail: 1000
```
nibanks, line 10: Maybe make this 1 in a 100 instead of 1000?

src/core/library.c
```diff
@@ -778,6 +778,40 @@ QuicLibrarySetGlobalParam(
...
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+        int32_t Value = *(int32_t*)Buffer;
+        if (Value < 0) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+        CxPlatSetAllocFailDenominator(Value);
+
```
nibanks, line 793: nit: either remove this empty newline or add one before `int32_t Value = ...`. Same with below.

src/core/binding.c
anrossi: Ok, better fix incoming.

.azure/templates/run-spinquic.yml
anrossi: I'll do that in a follow up PR which enables this.

src/inc/quic_platform.h
anrossi: The compiler does some constants optimization, so either is functionally the same. But `#define CxPlatisRandomMemoryFailureEnabled() FALSE` makes life easier for the optimizer.

CMakeLists.txt
```diff
@@ -203,10 +202,6 @@ endif()
 list(APPEND QUIC_COMMON_DEFINES VER_BUILD_ID=${QUIC_VER_BUILD_ID})
 list(APPEND QUIC_COMMON_DEFINES VER_SUFFIX=${QUIC_VER_SUFFIX})
 
-if(QUIC_RANDOM_ALLOC_FAIL)
-    list(APPEND QUIC_COMMON_DEFINES QUIC_RANDOM_ALLOC_FAIL=1 QUIC_DISABLE_MEM_POOL=1)
```
nibanks, line 207: Looks like `QUIC_DISABLE_MEM_POOL` will be a good "next step" to enable at run time for spin quic.

nibanks (approved):

PR #1541, opened 2021-05-04
Fix more SpinQuic random allocation failure bugs

anrossi:
Found and fixes https://github.com/microsoft/msquic/issues/1544.

### DIFF
--- a/src/core/packet_builder.c
+++ b/src/core/packet_builder.c
@@ -180,7 +180,7 @@ QuicPacketBuilderPrepare(
         //
         // Allocate and initialize a new send buffer (UDP packet/payload).
         //
-
+        BOOLEAN SendDataAllocated = FALSE;
         if (Builder->SendData == NULL) {
             Builder->SendData =
                 CxPlatSendDataAlloc(
@@ -199,6 +199,7 @@ QuicPacketBuilderPrepare(
                     0);
                 goto Error;
             }
+            SendDataAllocated = TRUE;
         }
 
         uint16_t NewDatagramLength =
@@ -220,6 +221,10 @@ QuicPacketBuilderPrepare(
                 "Allocation of '%s' failed. (%llu bytes)",
                 "packet datagram",
                 NewDatagramLength);
+            if (SendDataAllocated) {
+                CxPlatSendDataFree(Builder->SendData);
+                Builder->SendData = NULL;
+            }
             goto Error;
         }
 
@@ -452,7 +457,7 @@ QuicPacketBuilderGetPacketTypeAndKeyForControlFrames(
         Builder->Connection,
         "Failed to get packet type for control frames, 0x%x",
         SendFlags);
-    CXPLAT_DBG_ASSERT(FALSE); // This shouldn't have been called then!
+    CXPLAT_DBG_ASSERT(CxPlatIsRandomMemoryFailureEnabled()); // This shouldn't have been called then!
 
     return FALSE;
 }
@@ -589,7 +594,7 @@ QuicPacketBuilderFinalize(
                 Builder->Datagram = NULL;
             }
         }
-        FinalQuicPacket = FlushBatchedDatagrams;
+        FinalQuicPacket = FlushBatchedDatagrams && (Builder->TotalCountDatagrams != 0);
         goto Exit;
     }
 
@@ -810,6 +815,7 @@ QuicPacketBuilderFinalize(
     //
     // Track the sent packet.
     //
+    CXPLAT_DBG_ASSERT(Builder->Metadata->FrameCount != 0);
 
     Builder->Metadata->SentTime = CxPlatTimeUs32();
     Builder->Metadata->PacketLength =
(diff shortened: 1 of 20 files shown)


src/tools/spin/spinquic.cpp
```diff
@@ -19,8 +19,14 @@
 
 #define ASSERT_ON_FAILURE(x) \
     do { \
-        QUIC_STATUS _STATUS; \
-        CXPLAT_FRE_ASSERT(QUIC_SUCCEEDED((_STATUS = x))); \
+        QUIC_STATUS _STATUS = (x); \
+        if (Settings.AllocFailDenominator != 0) { \
```
anrossi, line 23: Remove all this.

src/platform/tls_openssl.c
```diff
@@ -476,6 +476,10 @@ CxPlatTlsAddHandshakeDataCallback(
     CXPLAT_TLS_PROCESS_STATE* TlsState = TlsContext->State;
 
     QUIC_PACKET_KEY_TYPE KeyType = (QUIC_PACKET_KEY_TYPE)Level;
+    if (TlsContext->ResultFlags & CXPLAT_TLS_RESULT_ERROR) {
+        CXPLAT_DBG_ASSERT(CxPlatIsRandomMemoryFailureEnabled());
+        return -1;
+    }
```
nibanks, line 482: I don't quite understand this one. Can you explain it?

anrossi: When the server is processing the client initial crypto in `CxPlatTlsSetEncryptionSecretsCallback`, if the handshake key fails to allocate, the error flag is set and `-1` is returned to OpenSSL. However, OpenSSL still proceeds to call `CxPlatTlsAddHandshakeDataCallback` despite the failure which hits this assert `CXPLAT_DBG_ASSERT(KeyType == 0 || TlsState->WriteKeys[KeyType] != NULL);`.   The idea here is to be resilient to external library errors and not continue if an error has already occurred.

src/tools/spin/spinquic.cpp
```diff
@@ -606,37 +613,48 @@ CXPLAT_THREAD_CALLBACK(ServerSpin, Context)
...
+            CredConfig))) {
+        goto CredLoadFail;
+    }
 
-    std::vector<HQUIC> Listeners;
     for (uint32_t i = 0; i < AlpnCount; ++i) {
         for (auto &pt : Settings.Ports) {
             HQUIC Listener;
-            ASSERT_ON_FAILURE(MsQuic->ListenerOpen(Registration, SpinQuicServerHandleListenerEvent, &Connections, &Listener));
+            if (!QUIC_SUCCEEDED(
```
Maximus-, line 645: Is there a reason why we won't just re-try this until it succeeds, at least, while the failing allocator is enabled?

anrossi: The motivation here is that SpinQuic can just run another iteration when these fail.  The goal is to get coverage of these APIs failing. Retrying them until success could be a more stable solution, however.

nibanks: I'd rather fail and start over. I think it will trigger more unique code coverage. If you just keep retrying, eventually you'd always progress forward in the same way as if there wasn't failure.

nibanks (approved):

src/platform/datapath_winuser.c
```diff
@@ -1746,33 +1746,13 @@ QUIC_DISABLED_BY_FUZZER_END;
 
     if (QUIC_FAILED(Status)) {
         if (Socket != NULL) {
-            QuicTraceEvent(
-                DatapathDestroyed,
-                "[data][%p] Destroyed",
-                Socket);
             if (Socket->ProcsOutstanding != 0) {
-                for (uint16_t i = 0; i < SocketCount; i++) {
```
ThadHouse, line 1754: This all scares me. When we get here, its possible there is outstanding completion status's on any number of the initialized queues.

src/core/packet_builder.c
```diff
@@ -822,13 +828,10 @@ QuicPacketBuilderFinalize(
...
         QuicPacketTraceType(Builder->Metadata),
         Builder->Metadata->PacketLength);
-    if (QUIC_FAILED(
-        QuicLossDetectionOnPacketSent(
-            &Connection->LossDetection,
-            Builder->Path,
-            Builder->Metadata))) {
-        goto Exit;
-    }
+    QuicLossDetectionOnPacketSent(
```
anrossi, line 831: Why shouldn't we exit when this fails an allocation internally?

src/core/stream_set.c
```diff
@@ -623,18 +623,24 @@ QuicStreamSetGetStreamForPeer(
                     FrameIn0Rtt,                    // Opened0Rtt
                     &Stream);
             if (QUIC_FAILED(Status)) {
+                *ProtocolViolation = TRUE;
```
anrossi, line 626: I could rename this something more generic like "FatalError" to indicate that the connection has been killed.

src/tools/spin/spinquic.cpp
anrossi: I'm going to stick with the current design. We could change to retry-until-success at a later time.

src/core/packet_builder.c
anrossi: Opened this issue to track the bug https://github.com/microsoft/msquic/issues/1570

PR #1548, opened 2021-05-05
Properly update the start point of the current key phase to decide old or new encryption keys

### LINKED ISSUE #1552: Selecting old vs new keys during key update uses incorrect logic
The logic we currently use for deciding to use the old or the new keys during a key update is incorrect. We currently always use the old key if the receive side has started its key update but has not confirmed. However, if the send side has also initiated a key update, and that key update is considered complete, it can actually start sending with the new key, which fails to decrypt, as the old key is selected. This manifested when adding DPLPMTUD, as that caused packets to be send without ACKs, this is required to trigger this behavior. So an MsQuic client before the DPLPMTUD change cannot trigger this bug, but other clients, and newer versions of MsQuic will be able to.

ThadHouse:
A bug was found while working on a separate PR that would cause the packets to be decoded incorrect upon key update. The root cause ended up being our method for choosing old or new keys when the opposite key phase was detected was incorrect. This PR updates the logic for selecting the keys to handle an updated key being used before the local side has confirmed its keys.

Closes #1552

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -3642,7 +3642,7 @@ QuicConnRecvPrepareDecrypt(
             CXPLAT_DBG_ASSERT(Connection->Crypto.TlsState.ReadKeys[QUIC_PACKET_KEY_1_RTT_OLD] != NULL);
             CXPLAT_DBG_ASSERT(Connection->Crypto.TlsState.WriteKeys[QUIC_PACKET_KEY_1_RTT_OLD] != NULL);
             Packet->KeyType = QUIC_PACKET_KEY_1_RTT_OLD;
-        } else if (Packet->PacketNumber > PacketSpace->ReadKeyPhaseEndPacketNumber) {
+        } else {
             //
             // The packet doesn't match our key phase, and the packet number is higher
             // than the end of the current key phase, so most likely using a new key phase.
@@ -3903,7 +3903,6 @@ QuicConnRecvDecryptAndAuthenticate(
 
             QuicCryptoUpdateKeyPhase(Connection, FALSE);
             PacketSpace->ReadKeyPhaseStartPacketNumber = Packet->PacketNumber;
-            PacketSpace->ReadKeyPhaseEndPacketNumber = Packet->PacketNumber;
 
             QuicTraceLogConnVerbose(
                 UpdateReadKeyPhase,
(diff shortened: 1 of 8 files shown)


src/core/connection.c
```diff
@@ -3664,6 +3661,9 @@ QuicConnRecvPrepareDecrypt(
                 return FALSE;
             }
             Packet->KeyType = QUIC_PACKET_KEY_1_RTT_NEW;
+        } else {
+            QuicPacketLogDrop(Connection, Packet, "This packet has already been read");
```
ThadHouse, line 3665: Is this the right error to send here. Or should we let code later down the path handle this case?

src/core/connection.c
```diff
@@ -3644,12 +3642,11 @@ QuicConnRecvPrepareDecrypt(
...
             Packet->KeyType = QUIC_PACKET_KEY_1_RTT_OLD;
-        } else {
+        } else if (Packet->PacketNumber > PacketSpace->ReadKeyPhaseEndPacketNumber) {
             //
-            // The packet doesn't match our key phase, and we're not awaiting
-            // confirmation of a key phase change, or this is a newer packet
-            // number, so most likely using a new key phase. Update the keys
-            // and try it out.
+            // The packet doesn't match our key phase, and the packet number is higher
+            // then the end of the current key phase, so most likely using a new key phase.
```
anrossi, line 3648: nit: "than"

src/core/connection.c
anrossi: This only happens when the packet has the same packet number as the ReadKeyPhaseEndPacket, right? If it's a duplicate, then it should be handled elsewhere with the other duplicate packets.

ThadHouse: It happens if its anywhere in the range between start and end. From here, it would basically go to decryption, and either fail to decrypt, or decrypt and be detected as a duplicate packet.

anrossi: In the packet reordering scenario, this seems like it could cause valid packets to be dropped, I think it might be best to do nothing in this case and let it get handled elsewhere.

ThadHouse: This only happens if a packet in the opposite key phase is received that is in the range of packets from the current key phase. So in theory it should never be hit, even with reordering. If I send it through, should it be an old key or the current key?

ThadHouse: Turns out this case is actually hit in stateless reset. So we just need to let it fly.

src/core/connection.c
```diff
@@ -3644,14 +3642,12 @@ QuicConnRecvPrepareDecrypt(
             CXPLAT_DBG_ASSERT(Connection->Crypto.TlsState.ReadKeys[QUIC_PACKET_KEY_1_RTT_OLD] != NULL);
             CXPLAT_DBG_ASSERT(Connection->Crypto.TlsState.WriteKeys[QUIC_PACKET_KEY_1_RTT_OLD] != NULL);
             Packet->KeyType = QUIC_PACKET_KEY_1_RTT_OLD;
-        } else {
+        } else if (Packet->PacketNumber > PacketSpace->ReadKeyPhaseEndPacketNumber) {
```
nibanks, line 3645: I think we should do `>=` in case we receive the a duplicate packet. I don't want to mark it as a decryption failure eventually.

ThadHouse: >= is incorrect per the spec. If this is hit, we've already received that packet in the current key space. There is NO scenario that packet can be legal ever. The only cause of it would be a bad implementation or a malicious peer.

ThadHouse: The main reason is its not a fully duplicate packet. Its a duplicate packet with the incorrect key phase if its ever in range inclusive. In that case, we have 0 clue which key would even potentially decode it.

nibanks: Ok. Makes sense.

src/core/connection.c
```diff
@@ -3915,12 +3912,17 @@ QuicConnRecvDecryptAndAuthenticate(
...
+            // endpoints.
             //
-            PacketSpace->ReadKeyPhaseStartPacketNumber = Packet->PacketNumber;
+            if (Packet->PacketNumber < PacketSpace->ReadKeyPhaseStartPacketNumber) {
+                PacketSpace->ReadKeyPhaseStartPacketNumber = Packet->PacketNumber;
+            }
+            if (Packet->PacketNumber > PacketSpace->ReadKeyPhaseEndPacketNumber) {
+                PacketSpace->ReadKeyPhaseEndPacketNumber = Packet->PacketNumber;
+            }
             QuicTraceLogConnVerbose(
```
nibanks, line 3926: Only log this if `ReadKeyPhaseStartPacketNumber` is updated? Otherwise every single new packet we receive in the current key phase will trigger this, right?

nibanks (approved):
LGTM. We should add random loss to this test scenario as well.

anrossi (approved):

PR #1551, opened 2021-05-06
Add exit code debugging to kernel mode test runs

ThadHouse:
We seem to have an issue that might be related to bad exit codes. So add some logging so we can figure out why.

### DIFF
--- a/scripts/run-gtest.ps1
+++ b/scripts/run-gtest.ps1
@@ -616,8 +616,17 @@ if ($Kernel -ne "") {
     Copy-Item (Join-Path $Kernel "msquictestpriv.sys") (Split-Path $Path -Parent)
     Copy-Item (Join-Path $Kernel "msquicpriv.sys") (Split-Path $Path -Parent)
     sc.exe create "msquicpriv" type= kernel binpath= (Join-Path (Split-Path $Path -Parent) "msquicpriv.sys") start= demand | Out-Null
+    if ($LastExitCode) {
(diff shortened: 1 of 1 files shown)


scripts/run-gtest.ps1
```diff
@@ -616,8 +616,17 @@ if ($Kernel -ne "") {
...
+        Log ("sc.exe " + $LastExitCode)
+    }
     verifier.exe /volatile /adddriver afd.sys msquicpriv.sys msquictestpriv.sys netio.sys tcpip.sys /flags 0x9BB
+    if ($LastExitCode) {
+        Log ("verifier.exe " + $LastExitCode)
+    }
     net.exe start msquicpriv
+    if ($LastExitCode) {
+        Log ("net.exe " + $LastExitCode)
+    }
```
nibanks, line 629: I'm wondering if we should also add error handling itself. If we're assuming one of these fail, let's see what we might do about it. For the `sc create` we could retry. For the `verifier` I think we just warn and continue. For the `net start` we retry too. Thoughts?

ThadHouse: I think I'd want to do that later. I'd just want to see what is failing first.

anrossi (approved):

scripts/run-gtest.ps1
nibanks: Ok.

nibanks (approved):

PR #1563, opened 2021-05-10
Add DPLPMTUD support

ThadHouse:
Add support for using DPLPMTUD to search for the largest MTU a path can support. 

DPLPMTUD uses probe packets of increasing size to test a paths MTU. Send probe packets until either max MTU is reached or 3 probe packets fail. Then enter a waiting period, and try again after a period (currently a minimum of 10 minutes).

This does not directly implement RFC 8899, a few changes are made. 1st, rather then using a separate timer to detect lost probe packets, use the QUIC ack tracking to detect a lost packet, which will then allow us to retry. 2nd is instead of using a timer in the waiting stage, we just set a time entered, and only if a send occurs after the wait time do we send. This makes it so a timer baced wakeup will not be forced to occur. Finally, we only implement the searching and search complete states, as QUIC lets us skip the base case in all but black hole cases, which we don't currently support.

This does not increase our maximum MTU, although that does now become possible. Doing so does add overhead on linux however, as we always have to post multiple MTU sized buffers.


src/plugins/dbg/quictypes.h
```diff
@@ -720,7 +720,7 @@ struct Stream : Struct {
 #define QUIC_CONN_SEND_FLAG_PING                    0x00001000
 #define QUIC_CONN_SEND_FLAG_HANDSHAKE_DONE          0x00002000
 #define QUIC_CONN_SEND_FLAG_DATAGRAM                0x00004000
-#define QUIC_CONN_SEND_FLAG_PMTUD                   0x80000000
+#define QUIC_CONN_SEND_FLAG_DPLPMTUD                 0x80000000
```
nibanks, line 723: Fix indent

src/core/loss_detection.c
```diff
@@ -422,6 +423,8 @@ QuicLossDetectionOnPacketSent(
         sizeof(QUIC_SENT_FRAME_METADATA) * TempSentPacket->FrameCount);
 
     LossDetection->LargestSentPacketNumber = TempSentPacket->PacketNumber;
+    CXPLAT_DBG_ASSERT(!SentPacket->Flags.IsDPLPMTUD ||
+                      (SentPacket->Flags.IsDPLPMTUD && !SentPacket->Flags.IsAckEliciting));
```
nibanks, line 427: Why should MTU probes **not** be ACK eliciting? That doesn't make sense. We need the ACK to get confirmation of the MTU.

ThadHouse: IsAckEliciting is actually a local flag which I think is misnamed. It actually means would resend. Because its only a local flag.

src/core/loss_detection.c
```diff
@@ -603,15 +606,14 @@ QuicLossDetectionOnPacketAcknowledged(
...
-                Path->Mtu);
-            QuicDatagramOnSendStateChanged(&Connection->Datagram);
+        if (Packet->Flags.IsDPLPMTUD) {
+            if (QuicMtuDiscoveryOnAckedPacket(
+                &Connection->MtuDiscovery,
+                PacketMtu,
+                Path,
+                Connection)) {
+                QuicDatagramOnSendStateChanged(&Connection->Datagram);
+            }
```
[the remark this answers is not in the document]
ThadHouse: I'll have to think about how to fix that. We don't want anything other then DPLPMTUD packets increasing MTU after a path is initially started.

src/core/mtu_discovery.c
```diff
@@ -0,0 +1,218 @@
...
+QuicMtuDiscoveryMoveToSearchComplete(
+    _In_ QUIC_MTU_DISCOVERY* MtuDiscovery
+    )
+{
+    QUIC_CONNECTION* Connection =
+        CXPLAT_CONTAINING_RECORD(MtuDiscovery, QUIC_CONNECTION, MtuDiscovery);
+    MtuDiscovery->State = QUIC_MTU_DISCOVERY_STATE_SEARCH_COMPLETE;
+    MtuDiscovery->SearchWaitingEnterTime = CxPlatTimeUs64();
+    QuicTraceLogConnInfo(
+        MtuProbeMoveToSearchComplete,
```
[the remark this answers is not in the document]
ThadHouse: I have to regen the events anyway for settings, so its not a big deal.

src/core/loss_detection.c
nibanks: Then we should rename it `IsRetransmittable`

nibanks: But I'm still not convinced I think it should still be ack eliciting, but the code should also use the probe flag when deciding to retransmit. For instance, if we do get an ACK for this, it should affect congestion control.

ThadHouse: These packets should affect the congestion window, but should not trigger congestion control. That is what happens currently.

src/core/mtu_discovery.c
```diff
@@ -0,0 +1,218 @@
...
+    //
+
+    uint16_t Mtu = MtuDiscovery->CurrentMtu + MTU_INCREMENT;
+    if (Mtu > MtuDiscovery->MaxMtu) {
+        Mtu = MtuDiscovery->MaxMtu;
+    } else if (Mtu == 1520) {
+        //
+        // 1520 is computed by the current algorithm, but we want 1500, so force that.
+        // Changing MTU_INCREMENT requires changing this logic, as does changing the
+        // initial MTU
```
nibanks, line 67:
```suggestion
        // initial MTU.
```

src/core/mtu_discovery.h
```diff
@@ -0,0 +1,69 @@
+/*++
+
+    Copyright (c) Microsoft Corporation.
+    Licensed under the MIT License.
+
+--*/
```
nibanks, line 6: Lots of comments needed in this file please

src/core/send.c
```diff
@@ -987,6 +987,8 @@ QuicSendFlush(
     QuicConnRemoveOutFlowBlockedReason(
         Connection, QUIC_FLOW_BLOCKED_SCHEDULING | QUIC_FLOW_BLOCKED_PACING);
 
+    QuicMtuDiscoveryCheckSearchCompleteReset(&Connection->MtuDiscovery);
```
nibanks, line 990: Might be better to put this a little farther down, after all the return checks.

src/core/mtu_discovery.h
ThadHouse: Like where? I tried to explain everything that was algorithm based. The only way to comment it more is to literally comment lines of code. (I already added the abstract locally)

src/core/send.c
ThadHouse: This might actually add flags. So this has to go before the line right below it, which checks the send flags.

src/core/loss_detection.c
ThadHouse: I realized that this assert was actually incorrect anyway. So removed.

src/core/mtu_discovery.h
```diff
@@ -0,0 +1,121 @@
...
+    // MTU discovery is in the searching state.
+    //
+    QUIC_MTU_DISCOVERY_STATE_SEARCHING,
+
+    //
+    // MTU discovery is in the waiting state.
+    //
+    QUIC_MTU_DISCOVERY_STATE_SEARCH_COMPLETE,
+
+} QUIC_MTU_DISCOVERY_STATE;
```
nibanks, line 20: Looks like this can be simplified to a `IsSearching` flag instead.

src/core/mtu_discovery.c
```diff
@@ -0,0 +1,238 @@
+/*++
+
+    Copyright (c) Microsoft Corporation.
+    Licensed under the MIT License.
+
+Abstract:
+
+    This module handles the MTU discovery logic.
+
+    Upon a new path being validated, MTU discovery is started
```
nibanks, line 10: nit: on this and the header, can you please align comments to 80 characters as much as possible (but not more)?

src/core/quicdef.h
```diff
@@ -243,6 +243,11 @@ CXPLAT_STATIC_ASSERT(IS_POWER_OF_TWO(QUIC_MAX_RANGE_DECODE_ACKS), L"Must be powe
 //
 #define QUIC_DEFAULT_PATH_MTU                   1280    // TODO - Use 1200 instead
 
+//
+// The maximum IP MTU DPLPMTUD will use by default.
+//
+#define QUIC_DEFAULT_MAX_MTU                  1500
```
nibanks, line 249: Rename to `QUIC_DPLPMUTD_MAX_MTU`?

src/inc/msquic.h
```diff
@@ -494,7 +494,9 @@ typedef struct QUIC_SETTINGS {
             uint64_t ServerResumptionLevel          : 1;
             uint64_t DesiredVersionsList            : 1;
             uint64_t VersionNegotiationExtEnabled   : 1;
-            uint64_t RESERVED                       : 36;
+            uint64_t MinimumMtu                    : 1;
```
nibanks, line 497:
```suggestion
            uint64_t MinimumMtu                     : 1;
```

src/core/settings.c
```diff
@@ -391,6 +403,26 @@ QuicSettingApply(
...
         }
     }
+
+    uint16_t MinimumMtu = Destination->MinimumMtu;
+    uint16_t MaximumMtu = Destination->MaximumMtu;
+    if (Source->IsSet.MinimumMtu && (!Destination->IsSet.MinimumMtu || OverWrite)) {
+        if (Destination->MinimumMtu < QUIC_MIN_INITIAL_PACKET_LENGTH || Destination->MinimumMtu > CXPLAT_MAX_MTU) {
+            return FALSE;
+        }
+        MinimumMtu = Source->MinimumMtu;
```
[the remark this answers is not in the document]
nibanks: The problem is that the app doesn't know any of these internal limits.

src/core/settings.c
```diff
@@ -711,6 +743,31 @@ QuicSettingsLoad(
...
+            QUIC_SETTING_MAXIMUM_MTU,
+            (uint8_t*)&MaximumMtu,
+            &ValueLen);
+    }
+    if (MinimumMtu >= QUIC_MIN_INITIAL_PACKET_LENGTH &&
+        MinimumMtu <= MaximumMtu &&
+        MaximumMtu <= CXPLAT_MAX_MTU) {
+        Settings->MinimumMtu = MinimumMtu;
+        Settings->MaximumMtu = MaximumMtu;
+    }
```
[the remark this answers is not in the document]
ThadHouse: Ok. Then fixed.

src/core/mtu_discovery.h
nibanks: Let's start with simple. No black hole detection. If/when we add that, then we can add more complicated states.

src/core/quicdef.h
```diff
@@ -241,7 +241,12 @@ CXPLAT_STATIC_ASSERT(IS_POWER_OF_TWO(QUIC_MAX_RANGE_DECODE_ACKS), L"Must be powe
 // Path MTU discovery will always start with/initialize with the smallest
 // allowable MTU for QUIC (1280 bytes).
 //
-#define QUIC_DEFAULT_PATH_MTU                   1280    // TODO - Use 1200 instead
+#define QUIC_DEFAULT_MIN_MTU                   1280    // TODO - Use 1200 instead
```
nibanks, line 244: Why not `QUIC_DPLPMTUD_DEFAULT_MIN_MTU` to align with the MAX below?

src/core/quicdef.h
```diff
@@ -431,6 +436,22 @@ CXPLAT_STATIC_ASSERT(
...
+// The timeout time in microseconds for the DPLPMTUD wait time.
+//
+#define QUIC_DPLPMTUD_RAISE_TIMER_TIMEOUT 600000000
+
+//
+// The amount of bytes to increase our PLMTU each probe
+//
+#define QUIC_DPLPMTUD_INCREMENT 80
+
+
```
nibanks, line 454: nit: extra newline!

src/core/settings.c
```diff
@@ -711,6 +743,31 @@ QuicSettingsLoad(
...
+            QUIC_SETTING_MAXIMUM_MTU,
+            (uint8_t*)&MaximumMtu,
+            &ValueLen);
+    }
+    if (MaximumMtu > MinimumMtu && MaximumMtu <= CXPLAT_MAX_MTU) {
+        Settings->MaximumMtu = MaximumMtu;
+    }
+    if (MinimumMtu >= QUIC_MIN_INITIAL_PACKET_LENGTH && MinimumMtu <= MaximumMtu) {
+        Settings->MinimumMtu = MinimumMtu;
+    }
```
nibanks, line 770: What if min > max?

ThadHouse: Then I don't want to set anything, just leave the defaults. We don't have any other sane option.

nibanks: Oh, I see what you're doing. You only accept the value if it's in a valid range. I would prefer you cap it to the valid range instead. So if MaximumMtu is > CXPLAT_MAX_MTU, use CXPLAT_MAX_MTU instead.

nibanks: For instance, like `MaxBytesPerKey` above.

ThadHouse: Makes sense. Done

src/core/mtu_discovery.c
```diff
@@ -0,0 +1,237 @@
...
+{
+    //
+    // N.B. This algorithm must always be increasing. Other logic in the module
+    // depends on that behavior.
+    //
+
+    uint16_t Mtu = MtuDiscovery->CurrentMtu + QUIC_DPLPMTUD_INCREMENT;
+    if (Mtu > MtuDiscovery->MaxMtu) {
+        Mtu = MtuDiscovery->MaxMtu;
+    } else if (Mtu == 1520) {
```
nibanks, line 84: This doesn't work if the INCREMENT changes, right? How can we make this more generic? Maybe if the increment skips 1500, then go back to 1500 and increment from there?

src/test/lib/MtuTest.cpp
```diff
@@ -0,0 +1,256 @@
...
+        //
+        MsQuicSettings CurrentSettings;
+        uint32_t SettingsSize = sizeof(CurrentSettings);
+        TEST_QUIC_SUCCEEDED(
+            MsQuic->GetParam(
+                NULL,
+                QUIC_PARAM_LEVEL_GLOBAL,
+                QUIC_PARAM_GLOBAL_SETTINGS,
+                &SettingsSize,
+                &CurrentSettings));
```
nibanks, line 41: Maybe we add a `SetGlobal` and `GetGlobal` function to `MsQuicSettings` for these?

src/test/lib/MtuTest.cpp
```diff
@@ -0,0 +1,256 @@
...
+            Settings.SetMaximumMtu(0xFFFF).SetMinimumMtu(1);
+
+            TEST_QUIC_STATUS(
+                QUIC_STATUS_SUCCESS,
+                MsQuic->SetParam(
+                    ClientConfiguration,
+                    QUIC_PARAM_LEVEL_CONFIGURATION,
+                    QUIC_PARAM_CONFIGURATION_SETTINGS,
+                    sizeof(Settings),
+                    &Settings));
```
nibanks, line 104: If you have the `SetGlobal` like I mention above, you could replace this with:
```c
`TEST_QUIC_STATUS(MsQuicSettings().SetMaximumMtu(0xFFFF).SetMinimumMtu(1).SetGlobal());
```

src/test/lib/MtuTest.cpp
```diff
@@ -0,0 +1,256 @@
...
+            //
+            Settings.SetMaximumMtu(1400).SetMinimumMtu(1300);
+            TEST_QUIC_STATUS(
+                QUIC_STATUS_INVALID_PARAMETER,
+                MsQuic->SetParam(
+                    Connection.Handle,
+                    QUIC_PARAM_LEVEL_CONNECTION,
+                    QUIC_PARAM_CONN_SETTINGS,
+                    sizeof(Settings),
+                    &Settings));
```
nibanks, line 160: We should add a `GetSettings` and `SetSettings` helper on `MsQuicConnection`.

src/test/lib/MtuTest.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+            TEST_EQUAL(1450, CheckSettings.MaximumMtu);
+            TEST_EQUAL(MinimumMtu, CheckSettings.MinimumMtu);
+        }
+    }
+}
+
+static
+QUIC_STATISTICS
+GetConnStatistics(_In_ MsQuicConnection& Conn) {
```
nibanks, line 182: This should be a helper on `MsQuicConnection` instead, `GetStatistics`.

src/test/lib/MtuTest.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+    const uint16_t MinimumMtu = RaiseMinimumMtu ? 1360 : 1280;
+    const uint16_t MaximumMtu = 1500;
+
+    MsQuicAlpn Alpn("MsQuicTest");
+    MsQuicSettings Settings;
+    Settings.SetMinimumMtu(MinimumMtu).SetMaximumMtu(MaximumMtu);
+
+    MsQuicConfiguration ServerConfiguration(Registration, Alpn, Settings, ServerSelfSignedCredConfig);
+    TEST_TRUE(ServerConfiguration.IsValid());
```
nibanks, line 217: I know you copied this from my (bad) example, but it's better to do the following:
```c
TEST_QUIC_SUCCEEDED(ServerConfiguration.GetInitStatus());
```
So we get the actual error on failure.

src/inc/msquic.hpp
```diff
@@ -421,6 +448,17 @@ class MsQuicConfiguration {
                 KeyCount * sizeof(QUIC_TICKET_KEY_CONFIG),
                 KeyConfig);
     }
+    QUIC_STATUS
+    SetSettings(_In_ const MsQuicSettings& Settings) noexcept {
+        const QUIC_SETTINGS* QSettings = &Settings;
```
nibanks, line 453: Why the extra temporary variable?

src/test/bin/quic_gtest.cpp
```diff
@@ -327,6 +327,29 @@ TEST(Basic, CreateConnection) {
...
+        ASSERT_TRUE(DriverClient.Run(IOCTL_QUIC_RUN_MTU_SETTINGS));
+    } else {
+        QuicTestMtuSettings();
+    }
+}
+
+TEST_P(WithMtuArgs, MtuDiscovery) {
+    TestLoggerT<ParamType> Logger("QuicMtuDiscoveryTest", GetParam());
+    if (TestingKernelMode) {
+        //ASSERT_TRUE(DriverClient.Run(IOCTL_QUIC_RUN_BIND_CONNECTION_IMPLICIT, GetParam().Family));
```
nibanks, line 342: Don't forget to do this.

src/inc/msquic.hpp
ThadHouse: Because it lets us guarantee that sizeof that variable is the size MsQuic expects, even if we add extra variables to MsQuicSettings that increase its size.

src/test/lib/MtuTest.cpp
```diff
@@ -0,0 +1,195 @@
...
+        MsQuicSettings UpdatedSettings;
+        QUIC_STATUS GetSuccess = UpdatedSettings.GetGlobal();
+
+        CurrentSettings.IsSetFlags = 0;
+        CurrentSettings.IsSet.MaximumMtu = TRUE;
+        CurrentSettings.IsSet.MinimumMtu = TRUE;
+        TEST_QUIC_SUCCEEDED(CurrentSettings.SetGlobal());
+
+        TEST_QUIC_SUCCEEDED(SetSuccess);
+        TEST_QUIC_SUCCEEDED(GetSuccess);
```
nibanks, line 52: Why do this here? I prefer to eliminate the extra variables and just check for failure at the point of failure.

ThadHouse: Since its setting the global settings, I want to make sure we successfully revert it back.

nibanks: Ah, I see. You're trying to handle reverting any local changes.

src/test/lib/MtuTest.cpp
```diff
@@ -0,0 +1,195 @@
...
+        //
+        MsQuicSettings CurrentSettings;
+        TEST_QUIC_SUCCEEDED(CurrentSettings.GetGlobal());
+
+        MsQuicSettings NewSettings;
+        QUIC_STATUS SetSuccess =
+            NewSettings.
+            SetMinimumMtu(1400).
+            SetMaximumMtu(1400).
+            SetGlobal();
```
nibanks, line 41:
```suggestion
        QUIC_STATUS SetSuccess =
            NewSettings.
                SetMinimumMtu(1400).
                SetMaximumMtu(1400).
                SetGlobal();
```

src/test/lib/MtuTest.cpp
```diff
@@ -0,0 +1,195 @@
...
+    {
+        {
+            MsQuicCredentialConfig ClientCredConfig;
+            MsQuicConfiguration ClientConfiguration(Registration, Alpn, ClientCredConfig);
+            TEST_QUIC_SUCCEEDED(ClientConfiguration.GetInitStatus());
+
+            MsQuicSettings Settings;
+
+            //
+            // Set out of range, correct order. This should just corce our boundaries.
```
nibanks, line 70: `corce`?

src/test/lib/MtuTest.cpp
```diff
@@ -0,0 +1,195 @@
...
+            TEST_QUIC_STATUS(
+                QUIC_STATUS_INVALID_PARAMETER,
+                ClientConfiguration.SetSettings(Settings));
+        }
+
+        MsQuicSettings ServerSettings;
+        MsQuicConfiguration ServerConfiguration(Registration, Alpn, ServerSettings, ServerSelfSignedCredConfig);
+        TEST_QUIC_SUCCEEDED(ServerConfiguration.GetInitStatus());
+
+        QUIC_ADDRESS_FAMILY QuicAddrFamily = QUIC_ADDRESS_FAMILY_INET;
```
nibanks, line 90: No need for a separate variable just for this.

src/core/connection.h
```diff
@@ -1390,3 +1390,54 @@ QuicConnParamGet(
...
+//
+_IRQL_requires_max_(DISPATCH_LEVEL)
+inline
+uint16_t
+QuicConnGetMaxMtuForPath(
+    _In_ QUIC_CONNECTION* Connection,
+    _In_ QUIC_PATH* Path
+    )
+{
+    uint16_t LocalMtu = CxPlatSocketGetLocalMtu(Path->Binding->Socket);
```
nibanks, line 1405: This may be a non-trivial query in the future. Might be a good idea to cache this whole calculation.

src/core/crypto.c
```diff
@@ -1551,7 +1551,13 @@ QuicCryptoProcessTlsCompletion(
         }
         Connection->Stats.ResumptionSucceeded = Crypto->TlsState.SessionResumed;
 
-        QuicSendSetSendFlag(&Connection->Send, QUIC_CONN_SEND_FLAG_PMTUD);
+        //
+        // A handshake complete means the peer has been validated. Trigger MTU
+        // discovery on path.
+        //
+        CXPLAT_DBG_ASSERT(Connection->PathsCount == 1);
```
nibanks, line 1558: Any reason not to just assert `PathsCount >= 1`?

src/core/loss_detection.c
```diff
@@ -595,25 +596,28 @@ QuicLossDetectionOnPacketAcknowledged(
...
+        if (Packet->Flags.IsMtuProbe) {
+            CXPLAT_DBG_ASSERT(Path->IsMinMtuValidated);
+            if (QuicMtuDiscoveryOnAckedPacket(
+                    &Path->MtuDiscovery,
+                    PacketMtu,
+                    Connection)) {
+                ChangedMtu = TRUE;
+            }
+        }
+        if (ChangedMtu) {
```
nibanks, line 620: Perhaps, we just do this if `Path->IsActive` too?

ThadHouse: Internally, QuicDatagramOnSendStateChanged always computes off of the active path, and only actually fires an event if things update. So this should be fine for now.

src/core/crypto.c
ThadHouse: Is it possible to get to this point with more then 1 path? I worry that if this assert fails then some paths might not get updates. The other option is to just trigger updates on all paths. That'd probably be better actually.

src/core/connection.h
ThadHouse: If we want to do any caching, I'd want to cache it at the socket layer. I don't think caching it at the connection is a good idea.

src/core/packet_builder.c
```diff
@@ -205,7 +205,7 @@ QuicPacketBuilderPrepare(
         uint16_t NewDatagramLength =
             MaxUdpPayloadSizeForFamily(
                 QuicAddrGetFamily(&Builder->Path->RemoteAddress),
-                IsPathMtuDiscovery ? CXPLAT_MAX_MTU : DatagramSize);
+                IsPathMtuDiscovery ? Builder->Path->MtuDiscovery.ProbedSize : DatagramSize);
```
nibanks, line 208: `ProbedSize` in this context seems to indicate this size has already been probed. That's not true, right? So `ProbeSize` might be better?

ThadHouse: Its possible this is a repeat probe. Its the current probe size. So yeah ProbeSide might work better.

src/core/crypto.c
ThadHouse: Actually triggering on all paths in not safe. So we get in a weird state if we ever have more than 1 path here.

src/core/quicdef.h
```diff
@@ -367,12 +380,13 @@ CXPLAT_STATIC_ASSERT(
...
 //
 #define QUIC_ACK_DELAY_EXPONENT                 8
+
 //
 // The length of the client initial packets sent. Instead of the required min of
 // 1200 we send the full 1280 to give the server more credit to overcome
 // amplification protection.
 //
-#define QUIC_INITIAL_PACKET_LENGTH              1280
+// #define QUIC_INITIAL_PACKET_LENGTH              1280
```
nibanks, line 389: Just remove this entirely now?

src/core/quicdef.h
```diff
@@ -238,10 +238,23 @@ CXPLAT_STATIC_ASSERT(IS_POWER_OF_TWO(QUIC_MAX_RANGE_ACK_PACKETS), L"Must be powe
 CXPLAT_STATIC_ASSERT(IS_POWER_OF_TWO(QUIC_MAX_RANGE_DECODE_ACKS), L"Must be power of two");
 
 //
-// Path MTU discovery will always start with/initialize with the smallest
-// allowable MTU for QUIC (1280 bytes).
+// Minimum MTU allowed to be configured. Must be able to fit a
+// QUIC_MIN_INITIAL_PACKET_LENGTH in an IPv6 datagram.
 //
-#define QUIC_DEFAULT_PATH_MTU                   1280    // TODO - Use 1200 instead
+#define QUIC_DPLPMUTD_MIN_MTU                   QUIC_MIN_INITIAL_PACKET_LENGTH  + \
```
nibanks, line 244: Probably best to encode in parentheses.

src/core/connection.h
nibanks: The size is (eventually) per-path, not per-socket, so we can't cache at the socket. It has to be cached at the path.

src/core/loss_detection.c
nibanks: Then there isn't any reason to call the function below if it's not for active, right?

src/core/crypto.c
nibanks: I'm not saying we trigger for all paths, just that we assert that it's not 0.

ThadHouse: That doesn't work, because if somehow by this point a path that was 1 is now the active path, it might already have been triggered. Without that assert, its possible to double trigger MTU discovery. So to handle that, we'd need extra code to ensure that can't happen.

src/core/quicdef.h
```diff
@@ -431,6 +445,21 @@ CXPLAT_STATIC_ASSERT(
...
 
+//
+// The maximum number of probe packets sent before considering an MTU too large.
+//
+#define QUIC_DPLPMTUD_MAX_PROBES                    3
+
+//
+// The timeout time in microseconds for the DPLPMTUD wait time.
+//
+#define QUIC_DPLPMTUD_RAISE_TIMER_TIMEOUT           600000000
```
nibanks, line 456: Can we use the `S_TO_US(60)` instead?

src/core/send.c
```diff
@@ -996,6 +996,9 @@ QuicSendFlush(
         return TRUE;
     }
 
+    // TODO Figure out how to pass time down
+    QuicMtuDiscoveryCheckSearchCompleteTimeout(Connection, CxPlatTimeUs64());
```
nibanks, line 1000: Implement this TODO?

src/core/loss_detection.c
ThadHouse: The old code called it for all paths that changed, not just the active one. I worry if we get to that point, and add multipath, this is a place we might miss.

ThadHouse:
I just realized why macOS was failing, and that my fix is actually broken. It's possible that a path gets considered active before mtu discovery has been triggered on a new path, but the mtu discovery send flag could already be on the send flags list.

ThadHouse:
This is now fully ready to go. I've ran a lot of testing, and double reviewed the logic to make sure things are correct. All tests pass as well, as does performance testing.

src/core/packet_builder.c
```diff
@@ -314,7 +314,7 @@ QuicPacketBuilderPrepare(
             Builder->MinimumDatagramLength =
                 MaxUdpPayloadSizeForFamily(
                     QuicAddrGetFamily(&Builder->Path->RemoteAddress),
-                    QUIC_INITIAL_PACKET_LENGTH);
+                    Builder->Path->Mtu);
```
nibanks, line 317: Why change this?

src/core/packet_builder.c
```diff
@@ -828,13 +828,14 @@ QuicPacketBuilderFinalize(
...
         PacketSpace->CurrentKeyPhaseBytesSent += (PayloadLength - Builder->EncryptionOverhead);
+        uint16_t MaxMtu = QuicConnGetMaxMtuForPath(Connection, Builder->Path);
 
         //
         // Check if the next packet sent will exceed the limit of bytes per
         // key phase, and update the keys. Only for 1-RTT keys.
         //
         if (Builder->PacketType == SEND_PACKET_SHORT_HEADER_TYPE &&
-            PacketSpace->CurrentKeyPhaseBytesSent + CXPLAT_MAX_MTU >=
+            PacketSpace->CurrentKeyPhaseBytesSent + MaxMtu >=
```
nibanks, line 838: I'm thinking this change is unnecessary. We just need a rough value here. I think just using a (any) #define is best.

src/core/path.c
```diff
@@ -138,15 +138,13 @@ QuicPathSetValid(
     Path->IsPeerValidated = TRUE;
     QuicPathSetAllowance(Connection, Path, UINT32_MAX);
 
-    if (Path->IsPeerValidated && Reason == QUIC_PATH_VALID_PATH_RESPONSE) {
+    if (Reason == QUIC_PATH_VALID_PATH_RESPONSE) {
```
nibanks, line 141: We don't have any testing for different paths, right?

src/core/quicdef.h
```diff
@@ -475,3 +497,8 @@ CXPLAT_STATIC_ASSERT(
 #define QUIC_SETTING_SERVER_RESUMPTION_LEVEL        "ResumptionLevel"
 
 #define QUIC_SETTING_VERSION_NEGOTIATION_EXT_ENABLE "VersionNegotiationExtEnabled"
+
+#define QUIC_SETTING_MINIMUM_MTU                    "MinimumMtu"
+#define QUIC_SETTING_MAXIMUM_MTU                    "MaximumMtu"
+#define QUIC_SETTING_MTU_SEARCH_COMPLETE_TIMEOUT    "MtuDiscoverySearchCompleteTimeoutUs"
+#define QUIC_SETTING_MTU_MISSING_PROBE_COUNT      "MtuDiscoveryMissingProbeCount"
```
nibanks, line 504: nit: align formatting

src/inc/quic_datapath.h
```diff
@@ -107,9 +107,13 @@ PacketSizeFromUdpPayloadSize(
...
     )
 {
-    return Family == QUIC_ADDRESS_FAMILY_INET ?
+    uint32_t PayloadSize = Family == QUIC_ADDRESS_FAMILY_INET ?
         UdpPayloadSize + CXPLAT_MIN_IPV4_HEADER_SIZE + CXPLAT_UDP_HEADER_SIZE :
         UdpPayloadSize + CXPLAT_MIN_IPV6_HEADER_SIZE + CXPLAT_UDP_HEADER_SIZE;
+    if (PayloadSize > UINT16_MAX) {
+        PayloadSize = UINT16_MAX;
+    }
+    return (uint16_t)PayloadSize;
```
nibanks, line 116: Do we need to do this here? Can we handle this at the (hopefully only 1) caller that is passing in too big of a value? Probably update the annotation here too to indicate you're not supposed to pass too big of a value?

src/core/packet_builder.c
ThadHouse: This is more correct. Min MTU is something we now allow to be set by settings, which also becomes a way for us to set minimum packet size.

src/core/packet_builder.c
ThadHouse: Ok. I can just change it to CXPLAT_MAX_MTU.

src/test/lib/EventTest.cpp
```diff
@@ -299,10 +299,12 @@ QuicTestValidateConnectionEvents1(
 {
     TestScopeLogger ScopeLogger(__FUNCTION__);
 
-    MsQuicConfiguration ServerConfiguration(Registration, "MsQuicTest", ServerSelfSignedCredConfig);
+    MsQuicSettings Settings;
+    Settings.SetMinimumMtu(1280).SetMaximumMtu(1280);
```
nibanks, line 303: Why update all these tests to have a fixed MTU?

src/core/path.c
ThadHouse: We have some code that tests path changes, that do hit these cases. There were actually some failures in the new code that were hit by these cases.

The reason the extra statement was removed is it was set to TRUE 3 lines above in all cases, so it was a useless check anyway.

src/test/lib/EventTest.cpp
ThadHouse: Because otherwise we get datagram events every time we get a validated packet, which screw with the order of the events.

src/core/packet_builder.c
nibanks: Min MTU can never be less than `QUIC_INITIAL_PACKET_LENGTH` right?

src/test/lib/EventTest.cpp
nibanks: Hmmm, ok.

src/inc/quic_datapath.h
ThadHouse: Its only 1 place that calls it with potentially too large of value that we would need to truncate. However, moving the logic outside of here would increase the complication because of the different families.

src/core/packet_builder.c
ThadHouse: No, it can never be less, but it can be more. And if set to more, we want this to be that higher value on the client side to enable a higher amplification window.

src/inc/quic_datapath.h
nibanks: Hmmm. One thing we've talked about is changing all our "packet size" stuff to be in terms of UDP payload. Perhaps that'd help this? (for now, just keep as is I guess)

nibanks (approved):

src/inc/quic_datapath.h
ThadHouse: The specific thing that makes this an issue is that the TP for max peer mtu has a max of 65527. And a lot of the other implementations set to that. And thats the value that overflows. Moving to udp payload size here wouldnt actually fix the bug, as I want MTU.

PR #1569, opened 2021-05-10
Use seeded RNG for memory loss SpinQuic scenarios

anrossi:


nibanks:
Why does normal `rand()` not suffice here?

anrossi:
I'll revisit this later.

PR #1572, opened 2021-05-11
Add support receiving different API Table Versions

ThadHouse:
Because of the distribution scenario on linux, we need to handle versions. Adds a new MsQuicOpenVersion to support doing this.

In msquic.h, MsQuicOpen forwards with either a macro or inline function to MsQuicOpenVersion. Internally, MsQuicOpen which will always return V1 is exported for compatibility with old binaries.


src/inc/msquic.h
```diff
@@ -1259,9 +1259,28 @@ typedef struct QUIC_API_TABLE {
...
+    )
+{
+    return MsQuicOpenVersion(1, (const void**)QuicApi);
+}
+#else
+#define MsQuicOpen(QuicApi) MsQuicOpenVersion((const void**)QuicApi, 1)
+#endif // defined(__cplusplus) || defined(WIN32)
+#endif // QUIC_CORE_INTERNAL
+
+
```
nibanks, line 1283: 😲 Extra newlines?!

nibanks:
Instead of deleting the MsQuicOpen.md, can you keep it and just have a note about "See MsQuicOpenVersion.md" instead?

nibanks (approved):

PR #1577, opened 2021-05-12
Fix Bug in Allocation Failure Path of OnPacketSent

### LINKED ISSUE #1570: When ACK tracking data fails to allocate on send, immediately trigger OnLoss handler
If the ACK tracking data fails to allocate, then the sender has no idea if the data was ACKed or lost, so call the OnLoss handler inline. Look at `QuicLossDetectionOnPacketSent`

nibanks:
Fixes #1570.

### DIFF
--- a/src/core/loss_detection.c
+++ b/src/core/loss_detection.c
@@ -401,21 +401,30 @@ QuicLossDetectionOnPacketSent(
     _In_ QUIC_SENT_PACKET_METADATA* TempSentPacket
     )
 {
-    QUIC_SENT_PACKET_METADATA* SentPacket;
     QUIC_CONNECTION* Connection = QuicLossDetectionGetConnection(LossDetection);
-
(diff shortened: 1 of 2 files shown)


src/core/loss_detection.c
```diff
@@ -401,21 +401,30 @@ QuicLossDetectionOnPacketSent(
...
     // Allocate a copy of the packet metadata.
     //
-    SentPacket =
+    QUIC_SENT_PACKET_METADATA* SentPacket =
         QuicSentPacketPoolGetPacketMetadata(
             &Connection->Worker->SentPacketPool, TempSentPacket->FrameCount);
     if (SentPacket == NULL) {
+        //
+        // We can't allocate the memory to permanently track this packet so just
+        // go ahead and immeadiate clean up and mark the data in it as lost.
```
ThadHouse, line 416: Spelling of immediately

ThadHouse (approved):

PR #1584, opened 2021-05-13
Add test to fail every Nth allocation.

anrossi:
This test will fail every Nth allocation for N from 100 down to 2.

### DIFF
--- a/src/core/library.c
+++ b/src/core/library.c
@@ -794,7 +794,7 @@ QuicLibrarySetGlobalParam(
             Status = QUIC_STATUS_INVALID_PARAMETER;
             break;
         }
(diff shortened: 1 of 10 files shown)


src/core/library.c
```diff
@@ -794,7 +794,8 @@ QuicLibrarySetGlobalParam(
             Status = QUIC_STATUS_INVALID_PARAMETER;
             break;
         }
-        int32_t Value = *(int32_t*)Buffer;
```
nibanks, line 797: Why?

nibanks (approved):

nibanks (approved):

PR #1588, opened 2021-05-13
Send ACK Frequency Updates

### LINKED ISSUE #929: Dynamically Scale Immediate ACK Threshold Up

nibanks:
Adds supports for sending ACK frequency frames to the peer. Currently, it sends packet tolerance values based on the maximum expected send batch size, with the expectation that we'd like the peer to batch ACKs about the same as we batch sends. In the CPU/scheduling limited case, we add 1 to indicate we want the peer to wait for the second batch as well if possible.

An attempt to fix #929.


src/core/send.c
```diff
@@ -1218,6 +1247,24 @@ QuicSendFlush(
...
+            //
+            // We're scheduling limited, so we should tell the peer to use our
+            // (max) batch size + 1 as the peer tolerance as a hint that they
+            // should expect more than a single batch before needing to send an
+            // acknowledgement back.
+            //
+            QuicConnUpdatePeerPacketTolerance(Connection, Builder.TotalCountDatagrams + 1);
+        }
+
+    } else if (Builder.TotalCountDatagrams > Connection->PeerPacketTolerance) {
```
maolson-msft, line 1261: Can Builder.TotalCountDatagrams change frequently across sends?

nibanks: It's a function of congestion window, pacing and application payload size. So yes, I'd say.

maolson-msft: In that situation, you'd be sending a lot of these ackfreq frames, right?

nibanks: Well, since we only send if this increases (unless there's persistent congestion) I don't expect that many. But that being said, if we happen to send 10 or 15 is that awful?

nibanks:
Data from a local run. Top is without sending ACK frequency frames. Bottom is with them. Throughput went from 2.6 Gbps to 3.3 Gbps (with logging enabled) for this loopback test.

image

src/core/send.c
maolson-msft: Ah! I missed that this was a watermark. So then, how about issues from packet tolerance being too large?

nibanks: There is always the delayed ACK timer. This is just increasing the threshold for immediate ACKs. And the final safety net is reset on persistent congestion. But I don't have any "right" answer here. The goal is to reduce the CPU costs of ACKs in the really high throughput scenarios. That CPU usage is the last known QUIC-specific logic bottleneck. I'm just experimenting here really.

maolson-msft: Right, I was going to ask about that too: what's the connection between persistent congestion and the packet tolerance?

nibanks: The idea here is to inform the peer of the send batch size (indirectly) so they ack at a similar batch size. If we hit persistent congestion, we're going to end up resetting our CWND which will result in a reset of the send batch size. Additionally, if we're seeing that much loss, we want to make sure the peer is ACK'ing more frequently to account for loss.

maolson-msft: Got it.

maolson-msft (approved):

PR #1593, opened 2021-05-14
Run release mode tests in CI

ThadHouse:
The internal windows build uses release mode msquic and msquictest binaries, however this is not tested anywhere in CI. Its possible for tests to be added that break this, and this won't then be detectable until a windows build. Run a build in CI so this can be detected early.

### DIFF
--- a/src/test/bin/quic_gtest.cpp
+++ b/src/test/bin/quic_gtest.cpp
@@ -625,6 +625,7 @@ TEST_P(WithHandshakeArgs6, ConnectClientCertificate) {
 }
 #endif
 
+#if QUIC_TEST_FAILING_TEST_CERTIFICATES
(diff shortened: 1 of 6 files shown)


nibanks (approved):

nibanks (approved):

src/test/bin/quic_gtest.cpp
```diff
@@ -1427,11 +1433,16 @@ INSTANTIATE_TEST_SUITE_P(
...
+#ifdef QUIC_TEST_DATAPATH_HOOKS_ENABLED
+
 INSTANTIATE_TEST_SUITE_P(
     Handshake,
     WithHandshakeArgs4,
     testing::ValuesIn(HandshakeArgs4::Generate()));
 
+#endif
+
+
```
anrossi, line 1445: nit: extra new line?

PR #1608, opened 2021-05-17
Zero Length Send + QUIC_SEND_FLAG_START

nibanks:
Ensures that a zero length send with the `QUIC_SEND_FLAG_START` will force the empty initial stream frame to be sent. It's effectively the same thing as calling `StreamStart` with the `QUIC_STREAM_START_FLAG_IMMEDIATE` flag.


anrossi:
Why is this needed? Why can't a caller just wait until they send data, or use the IMMEDIATE flag?

nibanks:
> Why is this needed? Why can't a caller just wait until they send data, or use the IMMEDIATE flag?

It came up in the discussion of the StreamSend docs. It's more of a completeness feature. One difference (I'm pretty sure) is that with this you get a completion as to when it's ACK'ed by the peer, where as a StreamStart gives no such completion.

anrossi:
> > Why is this needed? Why can't a caller just wait until they send data, or use the IMMEDIATE flag?
> 
> It came up in the discussion of the StreamSend docs. It's more of a completeness feature. One difference (I'm pretty sure) is that with this you get a completion as to when it's ACK'ed by the peer, where as a StreamStart gives no such completion.

Can you add a test to ensure this completion logic behaves as expected?

anrossi (approved):

PR #1613, opened 2021-05-18
SpinQuic: Pend some receives and try resuming them.

anrossi:


src/tools/spin/spinquic.cpp
```diff
@@ -200,6 +201,11 @@ QUIC_STATUS QUIC_API SpinQuicHandleStreamEvent(HQUIC Stream, void * /* Context *
     case QUIC_STREAM_EVENT_PEER_SEND_SHUTDOWN:
         MsQuic->StreamShutdown(Stream, (QUIC_STREAM_SHUTDOWN_FLAGS)GetRandom(16), 0);
         break;
+    case QUIC_STREAM_EVENT_RECEIVE:
+        if (GetRandom(5) == 0) {
```
nibanks, line 205: Let's add another `if (GetRandom(5) == 0)` condition after this to consume only part of the data (i.e. reduce `TotalBufferLength` by some value).

nibanks: Never mind. I was confused by what you were doing below. This can wait until the next PR.

src/tools/spin/spinquic.cpp
```diff
@@ -519,6 +525,18 @@ void Spin(LockableVector<HQUIC>& Connections, std::vector<HQUIC>* Listeners = nu
...
         }
+        case SpinQuicAPICallReceiveSetEnabled: {
+            auto Connection = Connections.TryGetRandom();
+            BAIL_ON_NULL_CONNECTION(Connection);
+            auto ctx = SpinQuicConnection::Get(Connection);
+            {
+                std::lock_guard<std::mutex> Lock(ctx->Lock);
+                auto Stream = ctx->TryGetStream();
+                if (Stream == nullptr) continue;
+                MsQuic->StreamReceiveSetEnabled(Stream, TRUE);
```
nibanks, line 536: Instead of always `TRUE` why not random and sometimes do `FALSE`?

anrossi: I'll make it false 50% of the time.

nibanks (approved):

PR #1614, opened 2021-05-19
SpinQuic: Add partial receive support.

anrossi:


src/tools/spin/spinquic.cpp
```diff
@@ -537,6 +545,20 @@ void Spin(LockableVector<HQUIC>& Connections, std::vector<HQUIC>* Listeners = nu
...
+        case SpinQuicAPICallStreamReceiveComplete: {
+            auto Connection = Connections.TryGetRandom();
+            BAIL_ON_NULL_CONNECTION(Connection);
+            auto ctx = SpinQuicConnection::Get(Connection);
+            {
+                std::lock_guard<std::mutex> Lock(ctx->Lock);
+                auto Stream = ctx->TryGetStream();
+                if (Stream == nullptr) continue;
+                auto BytesRemaining = MsQuic->GetContext(Stream);
+                MsQuic->StreamReceiveComplete(Stream, (uint64_t)BytesRemaining);
```
nibanks, line 557: You should (only sometimes) complete less than the bytes remaining, which will also pause.

src/tools/spin/spinquic.cpp
```diff
@@ -554,8 +556,14 @@ void Spin(LockableVector<HQUIC>& Connections, std::vector<HQUIC>* Listeners = nu
                 auto Stream = ctx->TryGetStream();
                 if (Stream == nullptr) continue;
                 auto BytesRemaining = MsQuic->GetContext(Stream);
-                MsQuic->StreamReceiveComplete(Stream, (uint64_t)BytesRemaining);
-                MsQuic->SetContext(Stream, nullptr);
+                if (BytesRemaining != nullptr && GetRandom(10) == 0) {
+                    auto BytesConsumed = GetRandom((size_t)BytesRemaining);
```
nibanks, line 560:
```suggestion
                    auto BytesConsumed = GetRandom((uint64_t)BytesRemaining);
```
This is better. It's still uint64_t on 32-bit as well.

nibanks (approved):

nibanks (approved):

PR #1621, opened 2021-05-21
Improve Handling of Ephemeral Port Reuse Scenario

nibanks:
While looking at the code, I noticed the `QuicLibraryGetBinding` function was unnecessarily doing a lookup for an existing binding if the local port was zero. This is unnecessary because we never actually have a binding with a zero port. This is generally used to indicate to the networking stack that we wish to have an ephemeral port assigned. So simply skip this step and go straight to creation.

Edit: Added additional logic to improve the handling of the scenario when Linux reuses the ephemeral port.

### DIFF
--- a/src/core/library.c
+++ b/src/core/library.c
@@ -1464,7 +1464,7 @@ QuicLibraryGetBinding(
             // requested configuration.
             //
             QuicTraceEvent(
-                BindingErrorStatus,
+                BindingError,
                 "[bind][%p] ERROR, %s.",
                 Binding,
                 "Binding already in use");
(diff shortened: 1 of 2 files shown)


src/core/library.c
```diff
@@ -1548,8 +1556,32 @@ QuicLibraryGetBinding(
     CxPlatDispatchLockRelease(&MsQuicLib.DatapathLock);
 
     if (Binding != NULL) {
-        if (Binding->Exclusive) {
-            Status = QUIC_STATUS_INVALID_STATE;
+        if (PortUnspecified) {
+            //
+            // The datapath somehow returned us a "new" ephemeral socket that
+            // already matched one of our existing ones. We've seen this on
+            // Linux occassionally. This shouldn't happen, but it does. Log and
```
maolson-msft, line 1563:
```suggestion
            // Linux occasionally. This shouldn't happen, but it does. Log and
```

maolson-msft (approved):

PR #1625, opened 2021-05-21
Add ConnectionOpen docs.

anrossi:


docs/api/ConnectionOpen.md
```diff
@@ -47,7 +47,11 @@ The function returns a [QUIC_STATUS](QUIC_STATUS.md). The app may use `QUIC_FAIL
 
 # Remarks
 
-**TODO**
+This call only allocates the resources for the connection, it does not start the connection. For that, the application must call [ConnectionStart](ConnectionStart.md).
```
nibanks, line 50: Maybe mention that this is used only for the client side, that server side is opened via the listener callback.

nibanks (approved):

PR #1628, opened 2021-05-21
Expand ConnectionStart doc.

anrossi:


docs/api/ConnectionStart.md
```diff
@@ -53,7 +53,11 @@ The function returns a [QUIC_STATUS](QUIC_STATUS.md). The app may use `QUIC_FAIL
 
 # Remarks
 
-**TODO**
+`ConnectionStart` initiates the connection from the client application. A server application doesn't start its side of the connection until it returns `QUIC_STATUS_SUCCESS` from the `QUIC_LISTENER_EVENT_NEW_CONNECTION` event.
+
+Since 0-RTT data is opportunistically sent during the connection handshake, it must be queued for send **BEFORE** calling `ConnectionStart` otherwise it will be sent after the handshake completes.
```
nibanks, line 58:
```suggestion
Since 0-RTT data is opportunistically sent during the connection handshake, it should be queued for send **BEFORE** calling `ConnectionStart` otherwise it may be sent after the handshake.
```
It's a race condition. It's possible it might send together, but likely won't.

nibanks (approved):

nibanks (approved):

PR #1630, opened 2021-05-23
Documentation

rahuljain1311:


docs/BUILD.md
```diff
@@ -115,6 +115,14 @@ To build the code, you just need to run `build.ps1` in the `scripts` folder:
 ./scripts/build.ps1
 ```
 
+Note that in windows you will need to use Powershell 7 in administrator mode to create the builds and test.
```
anrossi, line 118: `build.ps1` shouldn't require running powershell as Administrator

rahuljain1311: Tested without Administrator and it worked.

docs/BUILD.md
```diff
@@ -115,6 +115,14 @@ To build the code, you just need to run `build.ps1` in the `scripts` folder:
 ./scripts/build.ps1
 ```
 
+Note that in windows you will need to use Powershell 7 to create the build and Powershell 7 (in administrator mode) to run the tests.
```
nibanks, line 118: It already states above that you need powershell. I don't think you need anything specific to tests in this document.

docs/TEST.md
```diff
@@ -16,6 +16,14 @@ So, for a reliable run, that also includes logs for failed tests, run:
 ./scripts/test.ps1 -LogProfile Full.Light
 ```
 
+Note that in windows you will need to use Powershell 7 to create the build and Powershell 7 (in administrator mode) to run the tests.
```
nibanks, line 19: On Windows, I believe you only need Admin to get logs, not actually run the tests. Also, since powershell was already specified as a requirement, I don't think you need to state it again here.

rahuljain1311: Just tested. Yes, we only need Admin to get the logs. Updated the statement. If it still seems redundant I can remove it.

nibanks (approved):

PR #1631, opened 2021-05-23
Refactor Crypt Code Out

nibanks:
Splits the crypt.h code (encryption, keys, hashing) from the TLS code. Updates Windows to always use bcrypt, even when using OpenSSL.


src/platform/crypt_bcrypt.c
```diff
@@ -0,0 +1,752 @@
+/*++
+
+    Copyright (c) Microsoft Corporation.
+    Licensed under the MIT License.
+
+Abstract:
+
+    SCHANNEL TLS Implementation for QUIC
```
nibanks, line 8: Update abstract

src/platform/crypt_bcrypt.c
```diff
@@ -0,0 +1,752 @@
...
+    CXPLAT_AES_ECB_ALG_HANDLE = NULL;
+    CXPLAT_AES_GCM_ALG_HANDLE = NULL;
+#endif
+    if (CXPLAT_CHACHA20_POLY1305_ALG_HANDLE != NULL) {
+        BCryptCloseAlgorithmProvider(CXPLAT_CHACHA20_POLY1305_ALG_HANDLE, 0);
+        CXPLAT_CHACHA20_POLY1305_ALG_HANDLE = NULL;
+    }
+    QuicTraceLogVerbose(
+        SchannelUninitialized,
+        "[ tls] Library uninitialized");
```
nibanks, line 307: Unnecessary

src/platform/crypt_openssl.c
```diff
@@ -0,0 +1,537 @@
+/*++
+
+    Copyright (c) Microsoft Corporation.
+    Licensed under the MIT License.
+
+Abstract:
+
+    Implements the TLS functions by calling OpenSSL.
```
nibanks, line 8: Update abstract

anrossi:
Are you sure that Bcrypt has all the features that OpenSSL offers on Windows? ChaCha20 support isn't present in Bcrypt, for example.

nibanks:
> Are you sure that Bcrypt has all the features that OpenSSL offers on Windows? ChaCha20 support isn't present in Bcrypt, for example.

Yes, I know ChaCha20 is missing. I just wanted to see how perf was like this (it didn't seem to go up much), and then I'll put it back to using OpenSSL. But I still like to have it split.

anrossi:
> > Are you sure that Bcrypt has all the features that OpenSSL offers on Windows? ChaCha20 support isn't present in Bcrypt, for example.
> 
> Yes, I know ChaCha20 is missing. I just wanted to see how perf was like this (it didn't seem to go up much), and then I'll put it back to using OpenSSL. But I still like to have it split.

Ok, sounds good. I think developers would expect setting `openssl` would use OpenSSL crypto as well. Maybe in the future we could make it a build flag to use the platform crypto library (BCrypt on Windows, and libcrypto on OpenSSL) regardless of the TLS library used. However, I'm not sure many people would want that.

anrossi (approved):

PR #1643, opened 2021-05-27
Expand listener documentation.

anrossi:


docs/api/ConnectionClose.md
```diff
@@ -27,6 +27,8 @@ The valid handle to an open connection object.
 
 A caller should shutdown an active connection via `ConnectionShutdown` before calling `ConnectionClose`; calling `ConnectionClose` without `ConnectionShutdown` will implicitly call `ConnectionShutdown` with the `QUIC_CONNECTION_SHUTDOWN_FLAG_SILENT` flag.
 
+A server application **MUST NOT** call [ConnectionClose](ConnectionClose.md) within the `QUIC_LISTENER_EVENT_NEW_CONNECTION` callback to reject a connection. This will result in a double-free in release builds, and an assert in debug builds.
```
nibanks, line 30: Well.... good thing you bring this point up, because this statement isn't exactly true. If you return PENDING or SUCCESS, you can call ConnectionClose, but not if you return failure.

anrossi: That seems complicated; can we simplify this?

nibanks: The rule is this: Return success, you own the connection (pending is success). Otherwise (failure) you don't. If you own the connection, you must eventually close it.

ThadHouse: So if you return success we don't touch the connection ever again in the listener code? Even a variable to say it was accepted?

anrossi: @ThadHouse is correct. We'd be using-after-free if the application returns success/pending but calls `ConnectionClose` in the accept callback.

nibanks: Good find. Let's open a bug.

ThadHouse: Is that a solvable bug, or something we care about? Seems like just saying don't call close in the listener callback is the better answer. It doesn't really make sense anyway to do so.

nibanks: It is a solvable bug. At first glance, it might seem reasonable to just say "Don't close in the callback" but that adds complexity to the app, because they have to know if their in the callback. We've already had at least one case (with HTTP) where the app code traverses through a number of layers and got to a point that didn't necessarily know if it was in the callback or not, and unless you were actively thinking about this special case, they'd do something we didn't support on the callback.

Bottom line, if at all possible, we should not have special case logic for "in a callback" or not.

ThadHouse: Makes sense. We just need to be really careful and have a decent amount of comments, because a quick glance at the code and that callback on our end happens like 5 functions deep. And we have to make sure at every point we note that the connection must be invalid after this point, or manipulate the ref count to make sure the connection at least stays allocated.

nibanks: It's not that bad. We just make sure we whole a ref (which we always do because of the worker) and it cannot get deleted by the app.

anrossi: I've opened a bug to track the code change to make this more resilient. and I've updated the documentation to reflect how the API should work after that code change.

nibanks (approved):

PR #1645, opened 2021-05-27
Refactor POSIX Error Codes

nibanks:
This goes through the POSIX QUIC_STATUS codes and maps them as close as possible to official POSIX error codes. This is "officially" a breaking change for POSIX platform, but as it's still not yet officially supported in any previous release, I won't be rev'ing the major version.

cc @wfurt @ManickaP @CarnaViire to make sure any .NET changes eventually are made (if necessary)


src/inc/msquic_posix.h
```diff
@@ -80,86 +80,60 @@ inline ENUMTYPE &operator ^= (ENUMTYPE &a, ENUMTYPE b) throw() { return (ENUMTYP
...
-#define QUIC_STATUS_SOCKET_ERROR            ((QUIC_STATUS)ERROR_SOCKET_ERROR)
-#define QUIC_STATUS_TLS_ERROR               ((QUIC_STATUS)ERROR_SSL_ERROR)
-#define QUIC_STATUS_USER_CANCELED           ((QUIC_STATUS)ERROR_USER_CANCELED)
-#define QUIC_STATUS_ALPN_NEG_FAILURE        ((QUIC_STATUS)ERROR_ALPN_NEG_FAILURE)
-#define QUIC_STATUS_STREAM_LIMIT_REACHED    ((QUIC_STATUS)ERROR_STREAM_LIMIT_REACHED)
+#ifndef ESTRPIPE // undefined on macOS it seems
+#define ESTRPIPE 86
+#endif // ESTRPIPE
+
+#define ERROR_BASE                          200000000                       // 0xBEBC200
```
anrossi, line 96: To be "outside the range of all possible `int` values" shouldn't this be negative?

nibanks: I'm just going to leave as is.

ThadHouse (approved):

wfurt:
the negative values are unpleasant since QUIC_STATUS is actually `unsigned int`. That means that the actual value will change based on CPU architecture. In either way it is difficult to map in reasonably portable way.

wfurt:
no, but I can think about it. It seems like we already have offset beyond base errno, right? Can we use same approach or simply cast them as 0xffXX?

PR #1646, opened 2021-05-27
Fix Windows User Mode Build Breaks from VS v16.10 Update

nibanks:
AZP just started pushing an update that includes the VS 2019 v16.10 update, causing many of our builds to fail. This PR fixes most of those issues so at least Windows user mode builds pass. Kernel mode still has a link error that I can't quite figure out.

### DIFF
--- a/src/inc/quic_platform_winkernel.h
+++ b/src/inc/quic_platform_winkernel.h
@@ -30,6 +30,7 @@ Environment:
 #pragma warning(disable:26071)
 #pragma warning(disable:28118)
 #pragma warning(disable:28196)
+#pragma warning(disable:28251)
 #pragma warning(disable:28252)
 #pragma warning(disable:28253)
 #pragma warning(disable:28309)
--- a/src/inc/quic_platform.h
+++ b/src/inc/quic_platform.h
@@ -17,6 +17,8 @@ Supported Environments:
 
 #pragma once
 
+#include <stddef.h>
+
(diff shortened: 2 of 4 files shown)


anrossi (approved):

src/inc/quic_platform_winkernel.h
```diff
@@ -30,6 +30,7 @@
 #pragma warning(disable:26071)
 #pragma warning(disable:28118)
 #pragma warning(disable:28196)
+#pragma warning(disable:28251)
```
anrossi, line 33: we probably shouldn't ignore this one, though. It tells us if our header/source annotations are in sync. Or it should.

nibanks: We're just ignoring this on public headers for the most part. We don't want to fail our build if the public SDK/WDK headers have issues (which is what is happening here).

ThadHouse:
Im not surprised stddef.h fails in kernel mode. Can you try just copying the offsetof macro locally and seeing if that works. Still a bug, but there's no way to revert the builds on azp.

nibanks:
> Im not surprised stddef.h fails in kernel mode. Can you try just copying the offsetof macro locally and seeing if that works. Still a bug, but there's no way to revert the builds on azp.

stddef.h works fine on Windows kernel. I'm just seeing the unresolved symbol error now for _vsnprintf_s.

PR #1648, opened 2021-05-27
Remove "Not Yet RFC Text"

nibanks:
It's official now!


ThadHouse (approved):

ThadHouse:
We should probably link the official RFCs right below this text though.

nibanks:
> We should probably link the official RFCs right below this text though.

I'm waiting on better links than what are available so far to be available.

Issue #1615, opened 2021-05-19
Assert in msquic!CxPlatSendDataFinalizeSendBuffer under Extreme Alloc Failure


nibanks:
### Describe the bug

When running spinquic locally with extreme alloc failure (1 in 20), I hit an assert in `CxPlatSendDataFinalizeSendBuffer`:
```
 # Child-SP          RetAddr           Call Site
00 00000035`501fef30 00007ffd`2bfd61fa msquic!CxPlatSendDataFinalizeSendBuffer+0x2bd [F:\msquic\src\platform\datapath_winuser.c @ 3437] 
01 00000035`501fef90 00007ffd`2bfce9e6 msquic!CxPlatSendDataAllocBuffer+0xca [F:\msquic\src\platform\datapath_winuser.c @ 3534] 
02 00000035`501fefd0 00007ffd`2bfcc6b4 msquic!QuicPacketBuilderPrepare+0x496 [F:\msquic\src\core\packet_builder.c @ 272] 
03 00000035`501ff0e0 00007ffd`2bfa7283 msquic!QuicPacketBuilderPrepareForStreamFrames+0xe4 [F:\msquic\src\core\packet_builder.c @ 589] 
04 00000035`501ff120 00007ffd`2bf7ee3e msquic!QuicSendFlush+0x583 [F:\msquic\src\core\send.c @ 1112] 
05 00000035`501ff760 00007ffd`2bf9a36f msquic!QuicConnDrainOperations+0x3be [F:\msquic\src\core\connection.c @ 6725] 
06 00000035`501ff880 00007ffd`2bf99018 msquic!QuicWorkerProcessConnection+0x25f [F:\msquic\src\core\worker.c @ 487] 
07 00000035`501ff9d0 00007ffd`2c116a74 msquic!QuicWorkerThread+0xd8 [F:\msquic\src\core\worker.c @ 579] 
08 00000035`501ffb00 00007ffd`a39f54e0 vfbasics+0x16a74
09 00000035`501ffb40 00007ffd`a45ff90b KERNEL32!BaseThreadInitThunk+0x10
0a 00000035`501ffb70 00000000`00000000 ntdll!RtlUserThreadStart+0x2b
```
```c
    SendData->TotalSize += SendData->ClientBuffer.len;

    if (SendData->ClientBuffer.len == SendData->SegmentSize) {
        SendData->ClientBuffer.buf += SendData->SegmentSize;
        SendData->ClientBuffer.len = 0;
    } else {
[...]

PR #1650, opened 2021-05-27
Add helper script for grabbing build config and artifact directory for local machine

ThadHouse:
I only have my M1 mac on the airplane, and the code to correctly detect the platform only existed in the build script, and not test or spin. The logic is a bit complicated, especially if we ever want to be able to build native on arm or 32 bit windows, so move to its own script that can be called from all 3 top level scripts. The new script is built out in a way where different helper functions can be added in the future.


scripts/build.ps1
```diff
@@ -179,26 +179,21 @@ param (
...
-    }
+$BuildConfig = @{
+    Platform = $Platform
+    Tls = $Tls
+    Arch = $Arch
+    ExtraArtifactDir = $ExtraArtifactDir
+    Config = $Config
 }
 
+$BuildConfig = & (Join-Path $PSScriptRoot helpers.ps1) -BuildConfig $BuildConfig
```
nibanks, line 190: Can you have a better name than `helpers.ps1`?

nibanks: Also, instead of using the pattern you have so far, I think I'd prefer to pass all the options directly into the script, use positional arguments (so you don't need to do `-Platform <..> -Tls ...` every time, and then just use `$BuildConfig.*` directly, and not create the temporary variables like `$Tls`.

scripts/helpers.ps1
```diff
@@ -0,0 +1,85 @@
...
+
+    return @{
+        Platform = $Platform
+        Tls = $Tls
+        Arch = $Arch
+        ArtifactsDir = $ArtifactsDir
+    }
+}
+
+return "FAILURE"
```
nibanks, line 85: `Write-Error` instead?

ThadHouse: Doesn't work for this use case.

scripts/build.ps1
ThadHouse: The goal was to be able to add more options to this. Something like `ScriptHelpers.ps1` better?

nibanks: Why not make this specific to build config and just make this `Get-BuildConfig.ps1` or something like that?

ThadHouse: That idea doesn't work Nick, as I need to return multiple values, which requires objects. Also, this script I also want to move other helper functions used in multiple places into.

nibanks: I didn't say to change how/what you return from the script. So I don't quite understand what doesn't work. What other helpers do you want to move? I just don't like the idea of a generic "helper" script because you have no clue what's inside it from the name.

ThadHouse:
The failures are #1615 and the scheduler perf issues. This PR is now good to go.

scripts/spin.ps1
```diff
@@ -95,24 +95,11 @@ param (
...
 
-# Default TLS based on current platform.
-if ("" -eq $Tls) {
-    if ($IsWindows) {
-        $Tls = "schannel"
-    } else {
-        $Tls = "openssl"
-    }
-}
+$BuildConfig = & (Join-Path $PSScriptRoot get-buildconfig.ps1) -Platform "" -Tls $Tls -Arch $Arch -ExtraArtifactDir $ExtraArtifactDir -Config $Config
```
nibanks, line 98: We shouldn't need to pass the `""` for `-Platform` since it already has a default, right?

scripts/build.ps1
```diff
@@ -179,25 +179,12 @@ param (
...
-        } else {
-            Write-Error "Unknown architecture"
-        }
-    } else {
-        $Arch = "x64"
-    }
-}
+$BuildConfig = & (Join-Path $PSScriptRoot get-buildconfig.ps1) -Platform $Platform -Tls $Tls -Arch $Arch -ExtraArtifactDir $ExtraArtifactDir -Config $Config
+
+$Platform = $BuildConfig.Platform
```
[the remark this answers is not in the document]
ThadHouse: Because theyre used in a LOT of places that would have to be updated. Additionally the old code modified the variables in place.

scripts/spin.ps1
nibanks: Same for other script.

nibanks (approved):

PR #1651, opened 2021-05-27
Ignore how the transports shutdown in key update random loss test

### LINKED ISSUE #1596: QuicTestKeyUpdateRandomLoss Test Failure
### Describe the bug

On the latest main CI, the QuicTestKeyUpdateRandomLoss test failed once (logs attached). There were quite a large number of packets dropped, so I wonder if the test is just being too aggressive.

[quic.zip](https://github.com/microsoft/msquic/files/6487748/quic.zip)

ThadHouse:
We don't care how the transport shuts down or if it shuts down in this case. All we care is that no key update decryptions fail.

Helps #1596

### DIFF
--- a/src/test/lib/HandshakeTest.cpp
+++ b/src/test/lib/HandshakeTest.cpp
@@ -1893,15 +1893,7 @@ QuicTestKeyUpdateRandomLoss(
                 if (!Client.WaitForShutdownComplete()) {
                     return;
                 }
-
-                TEST_FALSE(Client.GetPeerClosed());
-                TEST_FALSE(Client.GetTransportClosed());
             }
-
-#if !QUIC_SEND_FAKE_LOSS
-            TEST_TRUE(Server->GetPeerClosed());
-            TEST_EQUAL(Server->GetPeerCloseErrorCode(), QUIC_TEST_NO_ERROR);
-#endif
         }
     }
 }


anrossi:
Shouldn't you just gate the success macros on whether loss is enabled? We should care about the connection failing to shutdown when key updates occur. It's just more test coverage into scenarios we don't exercise otherwise.

ThadHouse:
This function only runs when random loss is enabled. Its another function for key update without random loss.

nibanks:
> This function only runs when random loss is enabled. Its another function for key update without random loss.

So `RandomLoss` is never zero?

anrossi:
> This function only runs when random loss is enabled. Its another function for key update without random loss.

Oh, in that case, carry on

anrossi (approved):

ThadHouse:
RandomLossPercentage is never 0. Its always 1,5, or 10. Because the main key update test enforces that a specific number of key updates occur, and with random loss thats not possible. So it either was a ton of branches or a separate function. With a separate function, all we're testing is that key updates never fail to decrypt.

nibanks (approved):

nibanks:
@thadhouse I'm going to merge this, but it's not going to fix #1596 because that was from the connection getting outright disconnected in the middle of the handshake:

```
17:09:52.308080800 [Microsoft-Quic][conn][0xFFFFA30DE017A010] LOSS_DETECTION timer expired
[0]0004.0390::2021/05/14-17:09:52.308114500 [Microsoft-Quic][conn][0xFFFFA30DE017A010] Execute: 2(TIMER.LOSS_DETECTION)
17:09:52.308123500 [Microsoft-Quic][conn][0xFFFFA30DE017A010] Indicating QUIC_CONNECTION_EVENT_SHUTDOWN_INITIATED_BY_TRANSPORT [0xc0240006]
17:09:52.308129300 [Microsoft-Quic][test] File: D:\a\1\msquic\src\test\lib\TestConnection.cpp, Function: TestConnection::HandleConnectionEvent, Line: 775
17:09:52.308130700 [Microsoft-Quic][test] FAIL: Unexpected transport Close Error, 0xc0240006
```

ThadHouse:
That's bizarre. The test doesn't start random loss until after the handshake is completed. One of the logs I looked at the failure was these calls.

nibanks:
Sorry, I mispoke/typed. It's not getting disconnected during the handshake, but afterwards, but from what I remembered it was getting disconnected during the test work (not the end). Either way, it should never get disconnected except under extreme conditions. Perhaps you have too much random packet loss if it's getting disconnected (and it's not a bug).

ThadHouse:
Yeah, which I think is generally fine for this test. That's why those checks were failing. But we have lots of random loss tests, so mainly this one is testing key updates with that. So I think the bug can be closed.

PR #1652, opened 2021-05-27
Fix Client Cert Tests (Windows Kernel)

nibanks:
Adds impersonation support to Schannel Windows kernel code to make sure we call ISC/ASC in the same caller context as we called ACH.


anrossi (approved):

.azure/azure-pipelines.ci.yml
```diff
@@ -447,7 +447,7 @@ stages:
       pool: MsQuic-Win-Latest
       platform: windows
       tls: schannel
-      logProfile: Full.Light
+      logProfile: Full.Verbose
```
nibanks, line 450:
```suggestion
      logProfile: Full.Light
```

src/platform/tls_schannel.c
```diff
@@ -1249,62 +1260,27 @@ CxPlatTlsSecConfigCreate(
...
+            &AchContext->SecConfig->CopyOnOpen,
+            &AchContext->SecConfig->EffectiveOnly,
+            &AchContext->SecConfig->ImpersonationLevel);
+
+    if (AchContext->SecConfig->ImpersonationToken == NULL) {
+        AchContext->SecConfig->ImpersonationToken =
+            PsReferencePrimaryToken(
+                PsGetCurrentProcess());
+        AchContext->SecConfig->IsPrimaryToken = TRUE;
+    }
```
anrossi, line 1276: should assert here that the impersonation token is non-null.

nibanks: The code handles NULL, so I think we're fine without. I'm not sure if there is a scenario where it's supposed to be NULL.

PR #1656, opened 2021-05-28
Add workaround for Kernel Mode builds using wrong CRT

ThadHouse:
Its still possible other functions are broken, but the unit tests should tell us. 

Annotations were changed on new and delete as well. For delete, just removed the in annotation, and just deleted the throwing new implementation entirely. Anywhere we use that is a bug in kernel mode, so we just shouldn't implement it at all.

### DIFF
--- a/src/perf/bin/drvmain.cpp
+++ b/src/perf/bin/drvmain.cpp
@@ -74,16 +74,12 @@ SecNetPerfCtlUninitialize(
         void
     );
 
-void* __cdecl operator new (size_t Size) {
-    return ExAllocatePool2(POOL_FLAG_NON_PAGED, Size, QUIC_POOL_PERF);
-}
-
 _Ret_maybenull_ _Post_writable_byte_size_(_Size)
 void* __cdecl operator new (size_t Size, const std::nothrow_t&) throw(){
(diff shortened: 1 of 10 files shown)


nibanks (approved):

src/perf/bin/drvmain.cpp
```diff
@@ -74,16 +74,12 @@ SecNetPerfCtlUninitialize(
...
-    return ExAllocatePool2(POOL_FLAG_NON_PAGED, Size, QUIC_POOL_PERF);
-}
-
 _Ret_maybenull_ _Post_writable_byte_size_(_Size)
 void* __cdecl operator new (size_t Size, const std::nothrow_t&) throw(){
     return ExAllocatePool2(POOL_FLAG_NON_PAGED, Size, QUIC_POOL_PERF);
 }
 
-void __cdecl operator delete (_In_opt_ void* Mem) {
+void __cdecl operator delete (/*_In_opt_*/ void* Mem) {
```
anrossi, line 82: why are you removing `_In_opt_`?

ThadHouse: Because `new.h` doesn't include it, and VS 16.10 is complaining that the attributes don't match. I can't change the headers, so the option is to uncomment here or suppress the warning. This is better then suppression IMO.

anrossi (approved):

PR #1661, opened 2021-05-30
fix a few non-obvious mistakes in quic where the durable ID's conflic…

chgray:
…t with macros defined elsewhere - these arent obvious because QUIC isnt using manifested ETW everywhere so it doesnt matter.  flipping event names to be clearer

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -793,7 +793,7 @@ QuicConnUpdateRtt(
     if (RttUpdated) {
         CXPLAT_DBG_ASSERT(Path->SmoothedRtt != 0);
         QuicTraceLogConnVerbose(
-            RttUpdated,
+            RttUpdatedMsg,
             Connection,
             "Updated Rtt=%u.%03u ms, Var=%u.%03u",
             Path->SmoothedRtt / 1000, Path->SmoothedRtt % 1000,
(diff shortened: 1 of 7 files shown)


src/manifest/clog.sidecar
```diff
@@ -10173,6 +10173,118 @@
         }
       ],
       "macroName": "QuicTraceLogInfo"
+    },
+    "RttUpdatedMsg": {
```
nibanks, line 10177: Can you "clean" update to remove the old ones that no longer are used?

src/core/connection.c
```diff
@@ -793,7 +793,7 @@ QuicConnUpdateRtt(
     if (RttUpdated) {
         CXPLAT_DBG_ASSERT(Path->SmoothedRtt != 0);
         QuicTraceLogConnVerbose(
-            RttUpdated,
+            RttUpdatedMsg,
```
nibanks, line 796: Instead of `Msg` what about `Evt`?

src/manifest/clog.sidecar
chgray: we shouldnt, at least not in theory - the idea there is that the list grows

src/core/connection.c
chgray: feel free to send a commit;  I like Msg better :)

nibanks (approved):
Ok. I can create a PR in the future to clean up / possible rename in the future.

PR #1662, opened 2021-05-31
Consume CLOG as a submodule

chgray:
build clog (instead of using it with nuget)


.gitmodules
```diff
@@ -5,3 +5,6 @@
 	path = submodules/openssl
 	url = https://github.com/quictls/openssl.git
 	branch = OpenSSL_1_1_1k+quic
+[submodule "submodules/clog"]
+	path = submodules/clog
+	url = https://github.com/microsoft/CLOG.git
```
nibanks, line 10: Should this be pointing to a particular `branch` like we do for openssl above or is `master` what's being used here?

CMakeLists.txt
```diff
@@ -525,12 +525,20 @@ if(QUIC_TLS STREQUAL "openssl")
...
-    execute_process(COMMAND clog --installDirectory ${QUIC_BUILD_DIR}/clog)
+    add_subdirectory(submodules/clog)
+
+    if (WIN32)
+        LIST(APPEND CMAKE_PROGRAM_PATH  "${PROJECT_BINARY_DIR}/submodules/clog/artifacts/win-x64")
+        set(CLOG_EXE  "${PROJECT_BINARY_DIR}/submodules/clog/artifacts/win-x64/clog")
+    elseif (UNIX)
+        LIST(APPEND CMAKE_PROGRAM_PATH  "${PROJECT_BINARY_DIR}/submodules/clog/artifacts/linux-x64")
+        set(CLOG_EXE  "${PROJECT_BINARY_DIR}/submodules/clog/artifacts/linux-x64/clog")
+    endif()
```
nibanks, line 536: Is it ok to always reference `x64`? I assume this means we become unbuildable on an x86 machine?

nibanks (commented):
Looks good. Just a few minor questions/thoughts.

CMakeLists.txt
ThadHouse: This also breaks macOS builds. Which is something we do need working.

ThadHouse: Fixed

.gitmodules
nibanks: If we can merge 3.0 into CLOG master first, then this should be good as is. That'd be my preference.

nibanks:
This PR is close. The main thing is we need to upload the CLOG tools to the artifacts, so they can be used by the tests to decode the LTTng logs.

PR #1666, opened 2021-06-01
Enable TLS Secrets Support for Windows Kernel

nibanks:
Sets the build flag to enable `QUIC_PARAM_CONN_TLS_SECRETS` usage for msquic.sys.


ThadHouse (approved):
Theres no security risk enabling this, right? Only if the application actually does something with it? In that case, I'm fine with it.

nibanks:
> Theres no security risk enabling this, right? Only if the application actually does something with it? In that case, I'm fine with it.

Correct, this just allows the app to pass an output buffer to write secrets to. By default, they don't so there is no buffer to write to, so nothing is written to.

PR #1673, opened 2021-06-01
Reduce minimum MTU to 1248

### LINKED ISSUE #1216: Use 1200 Byte Minimum, not 1280
### Describe the bug

The specs recommend that you use a 1200 byte minimum packet send when sending (to allow for tunneling + min IPv4 MTU) initial packets. MsQuic currently uses 1280, which doesn't allow for tunneling if the MTU is 1280. One reason we would prefer to send more is to give the server more "amplification credit" so it can send more back, to allow for larger server certificates. But the tunneling aspect might be more important.

We should investigate reducing the value used in MsQuic (btw, it's not as trivial as just decreasing the constant `QUIC_INITIAL_PACKET_LENGTH`).

ThadHouse:
For tunnelling scenarios, we want to lower the starting MTU. DPLPMTUD support made minimum MTU user configurable, so applications can enforce a higher minimum if requested

Our DPLPMTUD algorithm required 1280 at a starting point for ideal increments. This adds a special case where 1280 will always be selected as the first probe value.

closes #1216

### DIFF
--- a/src/core/mtu_discovery.c
+++ b/src/core/mtu_discovery.c
@@ -81,6 +81,15 @@ QuicGetNextProbeSize(
     // depends on that behavior.
     //
 
+    //
+    // Jump automatically to 1280 to return algorithm to ideal case. 1280 should
+    // be supported in most scenarios. With minimum being 1248, this will always
+    // be less then a full increment.
+    //
+    if (Path->Mtu < 1280) {
+        return 1280;
+    }
+
     uint16_t Mtu = Path->Mtu + QUIC_DPLPMTUD_INCREMENT;
     if (Mtu > MtuDiscovery->MaxMtu) {
         Mtu = MtuDiscovery->MaxMtu;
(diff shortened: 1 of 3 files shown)


src/core/mtu_discovery.c
```diff
@@ -81,6 +81,15 @@ QuicGetNextProbeSize(
     // depends on that behavior.
     //
 
+    //
+    // Jump automatically to 1280 to return algorithm to ideal case. 1280 should
+    // be supported in most scenarios. With minimum being 1248, this will always
+    // be less then a full increment.
+    //
+    if (Path->Mtu < 1280) {
```
anrossi, line 89: Doesn't this prevent the app from ever setting the max to less than 1280?

ThadHouse: Oh, yeah good point. Let me fix that.

nibanks (commented):
We should include a test case for this.

ThadHouse:
I just updated the existing test to start at 1248 instead of 1280. The test already tests increasing the minimum MTU from default.

anrossi (approved):

PR #1691, opened 2021-06-03
More DAL Assert Fixes

nibanks:
Attempts to fix an assert we just saw in CI.

### DIFF
--- a/src/platform/datapath_winkernel.c
+++ b/src/platform/datapath_winkernel.c
@@ -2470,6 +2470,10 @@ CxPlatSendDataCanAllocSendSegment(
     _In_ UINT16 MaxBufferLength
     )
 {
+    if (!SendData->ClientBuffer.Buffer) {
+        return FALSE;
+    }
+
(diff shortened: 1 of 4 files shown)


src/platform/datapath_winkernel.c
```diff
@@ -2655,7 +2655,7 @@ CxPlatSendDataAllocSegmentBuffer(
         return (QUIC_BUFFER*)&SendData->ClientBuffer;
     }
 
-    Buffer = CxPlatSendDataAllocDataBuffer(SendData, &ProcContext->LargeSendBufferPool);
+    UINT8* Buffer = CxPlatSendDataAllocDataBuffer(SendData, &SendData->Owner->LargeSendBufferPool);
```
anrossi, line 2658: Is the double indirection going to have a perf impact?

anrossi (approved):

src/platform/datapath_winkernel.c
nibanks: No. It already was doing this, but I just removed the temporary, local variable.

PR #1692, opened 2021-06-04
add basic packaging script

wfurt:
Is is neither perfect not complete but this scrips adds ability to create RPM package for RedHat/Centos and deb for Ubuntu/Debian distributions. 

I used 'fpm' tool recommended by .NET infrastructure people because it allows to support multiple formats including packages for macOS (not supported yet) 

I used simple shell script for now so it is easier (for me) to iterate. We can eventually fold it to some PowerShell but it also allows to package with minimal dependencies. When executed, it will produce something like 
```
furt@ubu20:~/github/wfurt-msquic$ ls -al artifacts/packages/linux/x64_Release_openssl/libmsquic*
-rw-rw-r-- 1 furt furt 1784275 Jun  4 14:57 artifacts/packages/linux/x64_Release_openssl/libmsquic-1.5.0-1.x86_64.rpm
-rw-rw-r-- 1 furt furt 1789496 Jun  4 14:57 artifacts/packages/linux/x64_Release_openssl/libmsquic_1.5.0_amd64.deb
```  

each package includes just the two libraries. We could possibly also create package(s) with other executable but that is not needed IMHO at the moment. Other file to consider may be the sidecar so one can convert lilting logs. 

This PR is mostly meant to iterate on content and locations and package details. 

The packages are _not_ signed at the moment. My current thinking is that once published as artifacts/distribution we would have follow-up process to pick them up, sign and publish on packages.microsoft.com for general use.    

Once base script is in, I will continue to work on integration with CI

### DIFF
--- a/scripts/make-packages.sh
+++ b/scripts/make-packages.sh
@@ -20,7 +20,7 @@ if [ "$OS" == 'Linux' ]; then
         LIBDIR="lib64"
     else
         ARCH=x86
-        LIBDUR="lib"
+        LIBDIR="lib"
     fi
 else
     echo Only Linux packaging is supported at the moment.
--- a/scripts/make-packages.sh
+++ b/scripts/make-packages.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+
+OS=$(uname)
+ARCH=$(uname -m)
+FPM=`which fpm` 2>/dev/null
+CONFIG=Release
+VER_MAJOR=$(cat ./src/inc/msquic.ver | grep 'define VER_MAJOR'| cut -d ' ' -f 3)
+VER_MINOR=$(cat ./src/inc/msquic.ver | grep 'define VER_MINOR'| cut -d ' ' -f 3)
+VER_PATCH=$(cat ./src/inc/msquic.ver | grep 'define VER_PATCH'| cut -d ' ' -f 3)
+
+if [ -z "$FPM" ]; then
+  echo Install 'fpm'
+  exit 1
+fi
+
+if [ "$OS" == 'Linux' ]; then
+    OS=linux
+    if [ "$ARCH" == 'x86_64' ]; then
+        ARCH='x64'
+        LIBDIR="lib64"
+    else
+        ARCH=x86
+        LIBDUR="lib"
+    fi
+else
+    echo Only Linux packaging is supported at the moment.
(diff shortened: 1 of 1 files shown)


nibanks:
Should we add this to [create-package.yml](https://github.com/microsoft/msquic/blob/main/.azure/templates/create-package.yml)?

nibanks:
@chgray thoughts on including the sidecar in the Linux MsQuic package? I think it's probably a good idea, but we'd also need CLOG to decode as well.

scripts/make-packages.sh
```diff
@@ -0,0 +1,50 @@
...
+fi
+
+ARTIFACTS="artifacts/bin/${OS}/${ARCH}_${CONFIG}_openssl"
+if [ ! -e "$ARTIFACTS/libmsquic.so" ]; then
+    echo "$ARTIFACTS/libmsquic.so" does not exist. Run build first.
+    exit 1
+fi
+OUTPUT="artifacts/packages/${OS}/${ARCH}_${CONFIG}_openssl"
+mkdir -p ${OUTPUT}
+# RedHat/Cenoos
```
nibanks, line 37: nit, personal preference, newline before `#`, and should this be `Centos` not `Cenoos`?

nibanks:
```suggestion

# RedHat/CentOS
```

nibanks (approved):

scripts/make-packages.sh
```diff
@@ -0,0 +1,51 @@
...
+fi
+
+if [ "$OS" == 'Linux' ]; then
+    OS=linux
+    if [ "$ARCH" == 'x86_64' ]; then
+        ARCH='x64'
+        LIBDIR="lib64"
+    else
+        ARCH=x86
+        LIBDUR="lib"
```
nibanks, line 23:
```suggestion
        LIBDIR="lib"
```

wfurt:
thanks @nibanks for the updates. I will need to figure out dependencies before hooking to CI.

ThadHouse (approved):

nibanks:
> thanks @nibanks for the updates. I will need to figure out dependencies before hooking to CI.

Sounds good. For now, we will go ahead and merge this. Feel free to send a new PR to integrate into CI.

PR #1695, opened 2021-06-10
create Linux packages from package-distribution.ps1

wfurt:
This is minimal hook to get unsigned x64 packages.

I did look at 1.4 release and I did not see any arm binaries. 
When we attempt to package ARM, we will  probably need to cross-target ARCH. 
That is probably not critical right now.  

It would be great to create 1.5 prerelease so we can start consuming the packages in .NET CI as well as we can proceed on signing and publishing.


nibanks:
@thadhouse we need a way to easily test the distribution CI without merging first.

ThadHouse:
Since we don't publish we could just run distribution on pr's as well. It won't make the build take longer, and since it doesn't publish anything it'd be fine.

wfurt:
I was also thinking about making it temporary change to just test this change but if there is better way it would great.

nibanks:
FYI: https://github.com/microsoft/msquic/pull/1704

nibanks:
@wfurt can you please merge/rebase `main`?

nibanks:
From the CI:
```
Install fpm
...                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                Install fpm
##[error]PowerShell exited with code '1'.
```

wfurt:
That should be installed by `prepare-machine.ps1`. However it seems like that does not run or the packaging runs on different machine. I can perhaps hook it up but that will install all prerequisites for the build. Alternatively, we can create separate script or add another option like 'packaging' to avoid installing clang when not needed. 
It also seems like the packaging runs on Ubuntu 16.04 and standard support for that ended in April.  I will need to figure out if fpm is available there - but even if it is it may be very old version. 
For .NET we support last two stable release. For Ubuntu that would be 18.04 and 20.04 and we generally build on oldest supported OS as there is generally forward compatibility but not vice versa. 

Please let me know how do you want to proceed @nibanks

nibanks:
Looks like you'd just need to update `build-distribution.yml` to run `prepare-machine.ps1` first. Should be pretty simple to copy that from one of the other yml files.

nibanks (approved):
According to the CI, the packages are now available.

wfurt:
The run now looks reasonable after fixing the `and` condition. (complaining only about insecure path)

ThadHouse (approved):

PR #1706, opened 2021-06-11
Switch uses of min and max to CXPLAT_MIN and CXPLAT_MAX

### LINKED ISSUE #1697: Remove use of min macro.
The min macro likes to interfere with things. Libraries on windows usually know how to handle this, but on linux this is not the case. A big hitter is std library, where a definition of the min macro can make them completely fail to compile. We should define our own min macro with a different name to use.

ThadHouse:
The min and max macros in Windows.h are a huge pain point, however most windows libraries, including the stl, are able to handle this. However, we define these macros on posix, which most posix libraries, including the C++ standard libraries (which have min and max functions) completely breaks. Instead of defining these on posix, define a common CXPLAT_MIN and CXPLAT_MAX macro that works on all platforms, and doesn't interfere with use in posix applications and libraries.

Closes #1697

### DIFF
--- a/src/core/send_buffer.c
+++ b/src/core/send_buffer.c
@@ -46,7 +46,7 @@ Abstract:
     streams, and its legacy APIs don't let it pass the problem further up the
     stack to the layer that really knows.
 
-    So then, we indicate min(SendBuffer.IdealBytes, Stream.SendWindow) as the
+    So then, we indicate CXPLAT_MIN(SendBuffer.IdealBytes, Stream.SendWindow) as the
     ISB to each stream. If the app steadily sends on multiple streams, this
     means more data will be buffered than needed. But usually we expect only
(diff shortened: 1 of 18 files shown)


src/core/send_buffer.c
```diff
@@ -46,7 +46,7 @@
     streams, and its legacy APIs don't let it pass the problem further up the
     stack to the layer that really knows.
 
-    So then, we indicate min(SendBuffer.IdealBytes, Stream.SendWindow) as the
+    So then, we indicate CXPLAT_MIN(SendBuffer.IdealBytes, Stream.SendWindow) as the
```
nibanks, line 49: Can you go through and either revert the comment changes or fix up the line lengths?

anrossi (approved):

nibanks (approved):

PR #1712, opened 2021-06-14
Fix kernel perf finally script

ThadHouse:
Finally was killing the wrong driver name, which is a problem


nibanks (approved):
Should we have some logic that reboots in case of failure in here?

anrossi:
instead of reboot, couldn't the script stop the driver?

nibanks:
> instead of reboot, couldn't the script stop the driver?

It is trying to stop the driver, but in case that fails (it's hung) we need to reboot.

PR #1723, opened 2021-06-16
Setting for Binding Stateless Operations

nibanks:
Adds a settings knob that can be used to tune the timeout and max count for stateless operations on Bindings.

### DIFF
--- a/docs/api/QUIC_SETTINGS.md
+++ b/docs/api/QUIC_SETTINGS.md
@@ -11,35 +11,41 @@ typedef struct QUIC_SETTINGS {
     union {
         uint64_t IsSetFlags;
         struct {
-            uint64_t MaxBytesPerKey                 : 1;
-            uint64_t HandshakeIdleTimeoutMs         : 1;
-            uint64_t IdleTimeoutMs                  : 1;
-            uint64_t TlsClientMaxSendBuffer         : 1;
-            uint64_t TlsServerMaxSendBuffer         : 1;
-            uint64_t StreamRecvWindowDefault        : 1;
-            uint64_t StreamRecvBufferDefault        : 1;
-            uint64_t ConnFlowControlWindow          : 1;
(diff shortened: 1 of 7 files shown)


docs/api/QUIC_SETTINGS.md
```diff
@@ -61,18 +67,24 @@ typedef struct QUIC_SETTINGS {
...
     uint8_t PacingEnabled                   : 1;
     uint8_t MigrationEnabled                : 1;
     uint8_t DatagramReceiveEnabled          : 1;
-    uint8_t ServerResumptionLevel           : 2;
+    uint8_t ServerResumptionLevel           : 2;    // QUIC_SERVER_RESUMPTION_LEVEL
     uint8_t VersionNegotiationExtEnabled    : 1;
     uint8_t RESERVED                        : 1;
     const uint32_t* DesiredVersionsList;
     uint32_t DesiredVersionsListLength;
+    uint16_t MinimumMtu;
```
anrossi, line 82: I think if you arrange these from largest to smallest, you'll get better packing in the struct.

nibanks: Yeah, but I don't want to change the previous settings.

ThadHouse: Its a breaking change to rearrange them.

ThadHouse: Also, these 2 are here because theres 4 bytes of padding after DesiredVersionsListLength.

anrossi: I understand that. I mean among these new settings, you can rearrange them for better packing

anrossi: oh, so it is the docs. My mistake.

ThadHouse (approved):

PR #1741, opened 2021-06-18
Update LB Test to Handle Addr In Use Failures

### LINKED ISSUE #1740: Basic/WithFamilyArgs.LoadBalanced/1 failed
### Describe the bug

Basic/WithFamilyArgs.LoadBalanced/1 failed on the latest `main` CI run:

https://dev.azure.com/ms/msquic/_build/results?buildId=181688&view=logs&j=117cea68-5cbd-59c3-b438-3dc3522134d8&t=66e7dc66-7e7c-5560-3fec-16812f94817e&l=359

```
[ RUN      ] Basic/WithFamilyArgs.LoadBalanced/1
/home/vsts/work/1/s/src/test/lib/HandshakeTest.cpp:2662: Failure
Connection.HandshakeCompleteEvent.WaitTimeout(TestWaitTimeout) not true
[  FAILED  ] Basic/WithFamilyArgs.LoadBalanced/1, where GetParam() = v6 (2043 ms)
```

Logs: [load_balanced.log](https://github.com/microsoft/msquic/files/6678383/load_balanced.log)

nibanks:
Fixes #1740.

### DIFF
--- a/src/test/lib/HandshakeTest.cpp
+++ b/src/test/lib/HandshakeTest.cpp
@@ -2660,31 +2660,40 @@ QuicTestLoadBalancedHandshake(
         TEST_QUIC_SUCCEEDED(Connection.SetLocalAddr(ConnLocalAddr)); // TODO - Put in loop in case addr is taken
         TEST_QUIC_SUCCEEDED(Connection.StartLocalhost(ClientConfiguration, Listeners.PublicAddress));
         TEST_TRUE(Connection.HandshakeCompleteEvent.WaitTimeout(TestWaitTimeout));
-        if (SchannelMode) {
+        if (!Connection.HandshakeComplete) {
             //
-            // HACK: Schannel reuses tickets, so it always resumes. Also, no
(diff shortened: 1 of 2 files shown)


ThadHouse (approved):

src/test/lib/HandshakeTest.cpp
```diff
@@ -2660,31 +2660,40 @@ QuicTestLoadBalancedHandshake(
         TEST_QUIC_SUCCEEDED(Connection.SetLocalAddr(ConnLocalAddr)); // TODO - Put in loop in case addr is taken
```
anrossi, line 2660: Why not do this instead of just ignoring address in use errors?

nibanks: Actually, I should just remove this comment. The problem is that this won't fail, because we don't actually try to bind the socket until you start the connection, which if bind fails, results in the transport shutdown error code this PR is now handling.

PR #1742, opened 2021-06-18
Expand TLS configuration testing

anrossi:
Add tests to validate only supported flags are accepted by the TLS implementation. Plumb revocation checking flags to CAPI for use with OpenSSL with CAPI on Windows.


src/platform/tls_openssl.c
```diff
@@ -810,10 +811,25 @@ CxPlatTlsSecConfigCreate(
...
+        CredConfig->Flags & QUIC_CREDENTIAL_FLAG_REVOCATION_CHECK_CHAIN_EXCLUDE_ROOT ||
+        CredConfig->Flags & QUIC_CREDENTIAL_FLAG_IGNORE_NO_REVOCATION_CHECK ||
+        CredConfig->Flags & QUIC_CREDENTIAL_FLAG_IGNORE_REVOCATION_OFFLINE) {
+#ifdef CX_PLATFORM_USES_TLS_BUILTIN_CERTIFICATE
+        return QUIC_STATUS_NOT_SUPPORTED; // Only available on windows
+#else
+        if (CredConfig->Flags & QUIC_CREDENTIAL_FLAG_USE_TLS_BUILTIN_CERTIFICATE_VALIDATION) {
+            return QUIC_STATUS_INVALID_PARAMETER;
+        }
+#endif
```
nibanks, line 830: I believe the `#ifdef` part is actually unnecessary. @thadhouse please correct me if I'm wrong.
```suggestion
        if (CredConfig->Flags & QUIC_CREDENTIAL_FLAG_USE_TLS_BUILTIN_CERTIFICATE_VALIDATION) {
            return QUIC_STATUS_INVALID_PARAMETER;
        }
```

src/platform/unittest/TlsTest.cpp
```diff
@@ -1818,4 +1865,181 @@ TEST_F(TlsTest, CipherSuiteInvalid)
...
+static void
+QUIC_API
+SchannelSecConfigCreateComplete(
+    _In_ const QUIC_CREDENTIAL_CONFIG* /* CredConfig */,
+    _In_opt_ void* Context,
+    _In_ QUIC_STATUS Status,
+    _In_opt_ CXPLAT_SEC_CONFIG* SecConfig
+    )
+{
+#if QUIC_TLS_SCHANNEL
```
nibanks, line 1878: I see why you are making these defined, but I wish there was a way to do it without these. I've tried very hard up until now to only make feature level flags to control the testing of different behavior, like `QUIC_DISABLE_0RTT_TESTS` or `QUIC_DISABLE_DEFERRED_CERT_TESTS`. Could we instead make a flag like that for this?

nibanks:
Please add a PR description too.

src/platform/tls_openssl.c
ThadHouse: Well it depends on if openssl certificate loading supports those flags at all. Or is it just openssl cert validation that supports those flags?

src/platform/unittest/TlsTest.cpp
anrossi: I can make that change, but it feels like a wasted effort since the point is that the test should pass for one TLS implementation, and fail on the other TLS implementation. The feature flag would depend on the TLS implementation that's active.

src/platform/tls_openssl.c
anrossi: Only CAPI certificate chain building consumes those flags; OpenSSL doesn't consume them (I haven't looked to see if OpenSSL has analogues to these flags which we could translate between).  The `#ifdef` provides a shortcut for implementations that will never support these flags (i.e. platforms without CAPI, so not-Windows).  The USE_TLS_BUILDING_CERTF_VALIDATION flag check is in case on Windows platforms, the application has requested OpenSSL certificate validation, not CAPI.  In that case, I think it's actually up to the app to validate the certificate; does OpenSSL on Windows have its own trusted roots?

nibanks (approved):

PR #1745, opened 2021-06-19
Always force inline the shim MsQuicOpen function

ThadHouse:
On linux, if optimization is completely turned off (can't happen with our normal build, but can with modifications that make testing easier), the linker might choose the inline definition of MsQuicOpen, which will then cause an infinite loop upon open. The solution to this is to force inline the inline function, which causes the out of line definition to never be generated.

### DIFF
--- a/src/inc/msquic.h
+++ b/src/inc/msquic.h
@@ -1307,7 +1307,7 @@ MsQuicOpen(
 
 #else
 
-#define MsQuicOpen(QuicApi) MsQuicOpenVersion((const void**)QuicApi, 1)
+#define MsQuicOpen(QuicApi) MsQuicOpenVersion(1, (const void**)QuicApi)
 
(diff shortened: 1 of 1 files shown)


nibanks (approved):

src/inc/msquic.h
```diff
@@ -1293,6 +1293,11 @@ MsQuicOpenVersion(
 #if defined(__cplusplus) || defined(WIN32)
 
 _IRQL_requires_max_(PASSIVE_LEVEL)
+#ifdef WIN32
+__forceinline
+#else
+__attribute__((always_inline))
+#endif
```
nibanks, line 1300: Let's just scope this to Posix for now:
```suggestion
#ifndef WIN32
__attribute__((always_inline))
#endif
```

nibanks (approved):

nibanks (approved):

PR #1748, opened 2021-06-19
Refactor Library Load

nibanks:
The goal is the refactor library load so that we can remove the requirement for an explicit call to `MsQuicLibraryLoad` in DllMain (or similar). This is a pre-step for static linking.


nibanks:
FYI, @jeremyong I've integrated your interlocked logic for delay loading the library into this PR. I've made it a first class citizen. This should allow you to finish your static linking PR.

nibanks:
I'm a bit confused as to why I'm getting these build errors:
```
/usr/bin/ld: CMakeFiles/quicinterop.dir/interop.cpp.o: in function `main':
interop.cpp:(.text.startup+0x2b5): undefined reference to `MsQuicOpenVersion'
/usr/bin/ld: interop.cpp:(.text.startup+0x4f5): undefined reference to `MsQuicClose'
collect2: error: ld returned 1 exit status
```

ThadHouse:
I'll look when I get home.

ThadHouse:
I don't have a system with a push key right now. Fix is in init.c replace everything there with the following.

```
#include "msquic.h"

//
// This function only exists to ensure the static libraries are loaded into the
// shared object. This function cannot be called, and is not exported from the
// library
//
void
QuicSharedCallOpen() {
    MsQuicOpenVersion(1, NULL);
    MsQuicClose(NULL);
}

```

jeremyong:
Very cool, glad you were able to incorporate that change (I've been busy, transitioning jobs etc. so again, apologies I wasn't able to keep picking it up)

src/core/library.c
```diff
@@ -30,41 +30,115 @@ QuicLibraryEvaluateSendRetryState(
...
+            CxPlatDispatchLockUninitialize(&MsQuicLib.StatelessRetryKeysLock);
+            CxPlatDispatchLockUninitialize(&MsQuicLib.DatapathLock);
+            CxPlatLockUninitialize(&MsQuicLib.Lock);
+            CxPlatSystemUnload();
+
+            //
+            // Allow possibly spinning thread that incremented ref count in
+            // the meantime to proceed.
+            //
+            InterlockedAnd((long*)&MsQuicLib.LoadState, (long)~UNLOADING_BIT);
```
nibanks, line 139: Linux build doesn't like this:
```
/home/vsts/work/1/s/src/core/library.c:139:57: error: either cast from 'unsigned int' to 'long' is ineffective, or there is loss of precision before the conversion [bugprone-misplaced-widening-cast,-warnings-as-errors]
            InterlockedAnd((long*)&MsQuicLib.LoadState, (long)~UNLOADING_BIT)
```
Ideas?

ThadHouse: Hmm, that seems like a false positive. My guess is its detecting the result would be different if you casted before the not. I'd be curious if you hardcoded 0x7FFFFFFF here if it would work. As I'm not fully sure if its the left or the right side thats complaining.

nibanks:
@jeremyong FYI, I couldn't get the interlocked code you have to work. It was hitting some weird deadlocks, and I suspect it was related to the interlocked operating treating it as a signed number. So, we talked internally and decided a simpler path would be to only support parallel calls to `MsQuicOpen` when dynamic linking. We support this by explicitly calling `MsQuicLibraryLoad` upfront to ensure there is no race for first initialization. Static linking just won't be allowed to make parallel calls, but that should be fine. We don't expect that to be needed. If the CI passes, I'll merge this, and then I'd like you to merge `main` into your PR and hopefully we can push that through too. Thanks!

anrossi (approved):

Issue #1750, opened 2021-06-19
Support Client Sharing Local to Different Remotes


nibanks:
Currently, we don't allow a client to share a local address & port (tuple) unless the remote tuple matches as well; i.e. the UDP 4-tuple must match for sharing to succeed. This is currently a work around for a Windows limitation in the port pool, where the same local port cannot be shared among sockets.

This is a painful limitation because this means an app that wants to create lots of connections and wants to be able to share will only need to track which tuples are already in use and can only share if there is a matching tuple.

We should experiment if Linux has the same restriction. If it does not, then we can remove this limitation at least for Linux.

We should also look into fixing this limitation on Windows.

timprepscius:
This does not appear to work on osx.
It seems socket is sending to the one and only remote, rather than the other.

A test of something like (pseudo code) fails:  (acceptor1 gets all of the connections)
```
			TestProduct acceptor1Product;
			auto acceptor1 = factory->newAcceptor({});
			acceptor1->open();

			auto acceptor1Delegate = newFunctionPacketDelegate(acceptor1Product);
			acceptor1->setDelegate(acceptor1Delegate);

			TestProduct acceptor2Product;
			auto acceptor2 = factory->newAcceptor({});
			acceptor2->open();

			auto acceptor2Delegate = newFunctionPacketDelegate(acceptor2Product);
			acceptor2->setDelegate(acceptor2Delegate);
			
			// delay just a bit to make sure
			// with other libraries I can flush all events
			factory->cycle();

			// all connections from a connector will originate from the same socket
			auto connector = factory->newConnector({});
			
			// make a connection to the first acceptor
			auto connection1 = connector->connectTo(acceptor1->getLocalAddress());
			factory->cycle();

			// make a connection to the second acceptor
			auto connection2 = connector->connectTo(acceptor2->getLocalAddress());
			factory->cycle();

			REQUIRE(!connection1->isClosed());
			
			// this fails
			REQUIRE(!connection2->isClosed());
			
			Packet p { 0, 1, 2, 3, 4, 5 };
			connection1->send(Protocol::Reliable, Packet(p));
			connection2->send(Protocol::Reliable, Packet(p));
			factory->cycle();
			
			REQUIRE(acceptor1Product.packetsReceived.size() == 1);
			REQUIRE(acceptor2Product.packetsReceived.size() == 1);
			
			connection1 = nullptr;
			connection2 = nullptr;
			connector = nullptr;
			acceptor1 = nullptr;
[...]

PR #1751, opened 2021-06-21
Local UDP Port Sharing for Connected Sockets (Linux only)

nibanks:
Partial fix for #1750, at least fixes for Linux. Also starts refactoring/improving some tests.

Still needs a test at the QUIC API layer for sharing the local socket.


ThadHouse (approved):
Do we know if this is going to cause any perf issues on linux?

nibanks:
> Do we know if this is going to cause any perf issues on linux?

Are you asking if there is a perf difference between two client connections (to two different servers) that share local port, vs ones that don't? If so, I don't know the answer. If the question is if this would effect any existing perf tested scenario, then the answer is, I don't expect any perf difference.

ThadHouse:
Either perf or functional issues. Isn't the reason we can't do sharing on windows a functional issue? I know we now have some tests, but I want to make sure those tests are enough.

nibanks:
There are no perf or functional issues I am aware of on Linux related to this, beyond the weird behavior we've seen when you "share the port" but don't specify an explicit local address, and the stack ends up (sometimes) giving you one that was used by someone else.

PR #1752, opened 2021-06-21
Switch sample to be built in C instead of C++

ThadHouse:
We have lots of API tests for C++, but not that actually test the API from C. Bugs have been caught in the past about this.

Because of bugs in the Windows SDK, we do have to disable 1 warning on windows.


ThadHouse:
This will fail until #1745 is merged (As that was one of the bugs this would have caught)

nibanks (approved):

PR #1758, opened 2021-06-22
Switch Param Id's to not be overlapping

### LINKED ISSUE #1708: possibly make param ids non overlapping
I wrote the code:		
Status = quic->GetParam(acceptor->handle, QUIC_PARAM_LEVEL_LISTENER, QUIC_PARAM_CONN_LOCAL_ADDRESS, &Size, &Address)

And it didn't fail.

It would be nice if it failed.  If the PARAM_CONSTANTS were non overlapping, then you could detect misuse and fail (actually I think your code already would detect misuse).

ThadHouse:
This is done by making the bottom 26 bits be the Param Id and the top 6 bits be the level. 

This is able to be done as a non breaking change by making the level Id always the enum value + 1. This means that if the level is 0, we're running against an older caller, and the level passed in is used. If the level is non 0, so a new caller, the level is checked against the level parameter passed in. If these do not match, INVALID_PARAMETER is returned. In a future breaking change, the separate level param can be removed entirely, but this is a working solution for now that will make the API less error prone.

26 bits was chosen as we had a parameter that was already using bit 24. This leaves 63 valid levels. Above 32 is reserved for private levels, so we have 31 levels for future use.

Fixes #1708


src/core/api.c
```diff
@@ -1284,6 +1284,16 @@ MsQuicSetParam(
 {
     CXPLAT_PASSIVE_CODE();
 
+    if ((Param & 0xFC000000) != 0) {
+        // Has level embedded parameter
+        QUIC_PARAM_LEVEL ParamContainedLevel = ((Param >> 26) & 0x3F) - 1;
+        if (ParamContainedLevel != Level) {
+            return QUIC_STATUS_INVALID_PARAMETER;
```
nibanks, line 1291: Add a log

src/inc/msquic.h
```diff
@@ -593,56 +593,58 @@ typedef enum QUIC_PARAM_LEVEL {
...
 //
 // Parameters for QUIC_PARAM_LEVEL_GLOBAL.
 //
-#define QUIC_PARAM_GLOBAL_RETRY_MEMORY_PERCENT          0   // uint16_t
-#define QUIC_PARAM_GLOBAL_SUPPORTED_VERSIONS            1   // uint32_t[] - network byte order
-#define QUIC_PARAM_GLOBAL_LOAD_BALACING_MODE            2   // uint16_t - QUIC_LOAD_BALANCING_MODE
-#define QUIC_PARAM_GLOBAL_PERF_COUNTERS                 3   // uint64_t[] - Array size is QUIC_PERF_COUNTER_MAX
-#define QUIC_PARAM_GLOBAL_SETTINGS                      4   // QUIC_SETTINGS
-#define QUIC_PARAM_GLOBAL_VERSION                       5   // uint32_t[4]
+#define QUIC_PARAM_GLOBAL_RETRY_MEMORY_PERCENT          QUIC_PARAM_GENERATOR(QUIC_PARAM_LEVEL_GLOBAL, 0)   // uint16_t
```
nibanks, line 601: Hard code and move the generator internal.

nibanks (commented):
As discuss on IM

src/core/api.c
```diff
@@ -1284,6 +1286,20 @@ MsQuicSetParam(
 {
     CXPLAT_PASSIVE_CODE();
 
+    if ((Param & 0xFC000000) != 0) {
+        // Has level embedded parameter
```
nibanks, line 1290: nit: comment style

nibanks (approved):

PR #1763, opened 2021-06-23
Replace #include <> for Local Headers

### LINKED ISSUE #1762: use quotes instead of angled brackets
There are changes to msquic.hpp which requires the platform files

The quic_platform.h uses < instead of where it should use "
The quic_platform_posix uses < instead of where it should use "

In general (it is my understanding that) < should only by used if referencing a header outside of your library.

nibanks:
Fixes #1762.

### DIFF
--- a/src/core/injection.c
+++ b/src/core/injection.c
@@ -14,7 +14,7 @@ Abstract:
 
 #include "precomp.h"
 
-#include <msquic_fuzz.h>
+#include "msquic_fuzz.h"
 #ifdef QUIC_CLOG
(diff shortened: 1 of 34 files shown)


ThadHouse (approved):

maolson-msft:
Just curious: does this make any difference for any known compiler, or is it purely a convention?

ThadHouse:
Yes, there is a difference. Quotes generally include the path of the current file first, whereas brackets do not. This is not standardized, but generally this is how most compiler are implemented.

PR #1765, opened 2021-06-23
Ignore Retry packets when connection is shutting down.

### LINKED ISSUE #1767: Client can fail to process part of server flight and still accept stateless retry
### Describe the bug

If a client fails partially processing frames from a properly-decrypted server handshake packet, it won't correctly block retry packets. It should set `GotFirstServerResponse` when processing any frames from the first response from the server.

1. Client sends initial packet
2. server sends response
3. Client processes crypto frame and experiences an out of memory error and starts shutting down
4. Server sends a stateless retry packet to the client
5. Client starts processing stateless retry packet and hits an assert.

The client shouldn't process a stateless retry after successfully decrypting the server's first flight.

anrossi:
If a connection hasn't completed a handshake and is shutting down, but receives a retry packet after processing some data from the server, the connection hits an assert in QuicCryptoReset. The connection has already processed some of the server's flight, so it's too late for the server to send a retry packet. The connection is already shutting down, so there's no reason to retry.
This change skips the retry logic in this case.
Fixes #1767

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -3994,6 +3994,11 @@ QuicConnRecvFrames(
     uint16_t PayloadLength = Packet->PayloadLength;
     uint64_t RecvTime = CxPlatTimeUs64();
 
+    if (!QuicConnIsServer(Connection) &&
+        !Connection->State.GotFirstServerResponse) {
+        Connection->State.GotFirstServerResponse = TRUE;
+    }
+
     uint16_t Offset = 0;
     while (Offset < PayloadLength) {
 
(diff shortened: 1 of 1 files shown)


src/core/connection.c
```diff
@@ -3153,6 +3153,14 @@ QuicConnRecvRetry(
         return;
     }
 
+    //
+    // Make sure the connection is still active
+    //
+    if (Connection->State.ClosedLocally || Connection->State.ClosedRemotely) {
+        QuicPacketLogDrop(Connection, Packet, "Connection is shutting down");
```
nibanks, line 3160:
```suggestion
        QuicPacketLogDrop(Connection, Packet, "Retry while shutting down");
```

nibanks (approved):

PR #1778, opened 2021-06-28
Stream Priority Support

nibanks:
Adds a new (optional) parameter to set a 16-bit priority on a stream that controls the relative order between the streams to send the data out.

Still need to figure out some testing.

### DIFF
--- a/scripts/update-sidecar.ps1
+++ b/scripts/update-sidecar.ps1
@@ -1,17 +1,9 @@
 <#
 
 .SYNOPSIS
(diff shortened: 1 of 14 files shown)


anrossi:
I think we should document that stream priorities take effect on the next queued send for that stream. So if a stream is queued for send at default priority, and then the application changes the priority, they understand it won't take effect until the next `StreamSend` call.

anrossi (approved):

PR #1783, opened 2021-06-30
Clear out rebound paths upon path exhaustion

ThadHouse:
Currently we have a hard limit of 4 paths for a connection. Rebinding can possibly fill up these paths, causing the connection to be dropped. As a solution, if we detect we're out of paths, we can see if any of the non active paths are on the same local address, and the same remote IP address, but just a different port, and if so we can discard the old paths, as if theyre used again they'll just show up as another rebind.

Also adds a test that rebinds 50 times.


src/test/lib/PathTest.cpp
```diff
@@ -0,0 +1,106 @@
...
+    TEST_NOT_EQUAL(nullptr, Context.Connection);
+
+    QuicAddr OrigLocalAddr;
+    TEST_QUIC_SUCCEEDED(Connection.GetLocalAddr(OrigLocalAddr));
+    QuicAddr NewLocalAddr(OrigLocalAddr);
+    
+    
+    for (int i = 0; i < 50; i++) {
+        NewLocalAddr.IncrementPort();
+        ReplaceAddressHelper AddrHelper(OrigLocalAddr.SockAddr, NewLocalAddr.SockAddr);
```
nibanks, line 77: I fear we may see issues (race conditions) during the time this is destructed and then re-constructed for the new loop. Perhaps we should created this outside the loop and then add a new function to change the addr each loop instead.

ThadHouse: I tried that, and that was actually was more problematic, and often caused the test to fail. But this issue is the reason I turn off the keep alives. Also, a race here isn't the end of the world, as eventually one of the keep alives will get the new IP address and be fine.

src/inc/msquic.hpp
```diff
@@ -802,6 +814,11 @@ struct MsQuicConnection {
                 &Value);
     }
 
+    QUIC_STATUS
+    SetKeepAlive(_In_ uint32_t Value) noexcept {
+        return SetSettings(MsQuicSettings{}.SetKeepAlive(Value));
+    }
+
```
nibanks, line 821: I'd rather not add this helper to msquic.hpp. If we follow this pattern to the logical conclusion, now you have a ton of duplication everywhere (i.e. add all settings here; then add it to the listeners, etc.). This is simple enough to put at the caller instead.

src/test/bin/quic_gtest.cpp
```diff
@@ -554,6 +554,16 @@ TEST_P(WithHandshakeArgs3, AsyncSecurityConfig) {
     }
 }
 
+TEST_P(WithFamilyArgs, LocalPathChanges) {
+    TestLoggerT<ParamType> Logger("QuicTestLocalPathChanges", GetParam());
+    if (TestingKernelMode) {
+        // TODO
+        //ASSERT_TRUE(DriverClient.Run(IOCTL_QUIC_RUN_VERSION_NEGOTIATION, GetParam().Family));
```
nibanks, line 561: Don't forget this.

src/core/path.c
```diff
@@ -196,9 +195,30 @@ QuicConnGetPathForDatagram(
...
         //
-        return NULL;
+        for (uint8_t i = Connection->PathsCount - 1; i > 0; i--) {
+            BOOLEAN IsSimilarRemoteAddress =
+                QuicAddrGetFamily(&Datagram->Tuple->RemoteAddress) == QuicAddrGetFamily(&Connection->Paths[i].RemoteAddress) &&
+                QuicAddrCompareIp(&Datagram->Tuple->RemoteAddress, &Connection->Paths[i].RemoteAddress);
+            BOOLEAN IsSameLocalAddress = QuicAddrCompare(&Datagram->Tuple->LocalAddress, &Connection->Paths[i].LocalAddress);
+            if (IsSimilarRemoteAddress
+                && IsSameLocalAddress
+                && !Connection->Paths[i].IsActive) {
```
nibanks, line 210: You can optimize this `if` without separate state variables and put the `IsActive` check first.

src/core/path.c
```diff
@@ -56,7 +56,6 @@ QuicPathRemove(
 
 #if DEBUG
     if (Path->DestCid) {
-        CXPLAT_DBG_ASSERT(Path->DestCid->AssignedPath != NULL);
```
nibanks, line 59: Can you explain why this is no longer valid?

src/test/lib/PathTest.cpp
nibanks: Can you explain why it was problematic? This test isn't producing a realistic scenario if sometimes the old path is used. I really think we should guarantee that doesn't happen.

src/test/lib/PathTest.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+    static QUIC_STATUS ConnCallback(_In_ MsQuicConnection* Conn, _In_opt_ void* Context, _Inout_ QUIC_CONNECTION_EVENT* Event) {
+        PathTestContext* Ctx = static_cast<PathTestContext*>(Context);
+        Ctx->Connection = Conn;
+        if (Event->Type == QUIC_CONNECTION_EVENT_SHUTDOWN_COMPLETE) {
+            Ctx->Connection = nullptr;
+            Ctx->ShutdownEvent.Set();
+        }
+        else if (Event->Type == QUIC_CONNECTION_EVENT_PEER_ADDRESS_CHANGED) {
+            Ctx->PeerAddrChanged = true;
```
nibanks, line 30: Have a `CxPlatEvent` here you set, so the main/test thread can just wait, instead of poll.

src/test/lib/PathTest.cpp
```diff
@@ -0,0 +1,106 @@
...
+                ServerAddressUpdated = true;
+                Context.PeerAddrChanged = false;
+                break;
+            }
+        } while (++Try <= 3);
+        TEST_TRUE(ServerAddressUpdated);
+        Connection.SetKeepAlive(0);
+    }
+
+    CxPlatSleep(1000);
```
nibanks, line 99: Replace this with a wait on an event.

src/core/path.c
ThadHouse: Because it was possible for the path to not be assigned yet by the time if was getting removed. So this check would fail if that ended up being the case.

src/test/lib/PathTest.cpp
ThadHouse: Ok, I'll try it again. It just wasn't incrementing at all for some reason.

src/test/lib/PathTest.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    static QUIC_STATUS ConnCallback(_In_ MsQuicConnection* Conn, _In_opt_ void* Context, _Inout_ QUIC_CONNECTION_EVENT* Event) {
+        PathTestContext* Ctx = static_cast<PathTestContext*>(Context);
+        Ctx->Connection = Conn;
+        if (Event->Type == QUIC_CONNECTION_EVENT_SHUTDOWN_COMPLETE) {
+            Ctx->Connection = nullptr;
+            Ctx->PeerAddrChangedEvent.Set();
+            Ctx->ShutdownEvent.Set();
+        }
+        else if (Event->Type == QUIC_CONNECTION_EVENT_PEER_ADDRESS_CHANGED) {
```
nibanks, line 30:
```suggestion
        } else if (Event->Type == QUIC_CONNECTION_EVENT_PEER_ADDRESS_CHANGED) {
```

src/test/lib/PathTest.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+        TEST_TRUE(Context.PeerAddrChangedEvent.WaitTimeout(1000))
+        Context.PeerAddrChangedEvent.Reset();
+        QuicAddr ServerRemoteAddr;
+        TEST_QUIC_SUCCEEDED(Context.Connection->GetRemoteAddr(ServerRemoteAddr));
+        TEST_TRUE(QuicAddrCompare(&AddrHelper.New, &ServerRemoteAddr.SockAddr));
+        Connection.SetSettings(MsQuicSettings{}.SetKeepAlive(0));
+    }
+
+    TEST_NOT_EQUAL(nullptr, Context.Connection);
```
nibanks, line 88: This shouldn't be necessary, right? You already check this up above.

src/test/lib/PathTest.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    Connection.Shutdown(1);
+    Context.Connection->Shutdown(1);
+
+    Context.ShutdownEvent.WaitTimeout(2000);
```
nibanks, line 93: Do we really care about waiting for shutdown? You could just make the registration pass `true` in the constructor to make sure it tears down connections on scope exit.

src/test/lib/PathTest.cpp
```diff
@@ -0,0 +1,94 @@
...
+    TEST_QUIC_SUCCEEDED(Listener.Start(Alpn, &ServerLocalAddr.SockAddr));
+    TEST_QUIC_SUCCEEDED(Listener.GetLocalAddr(ServerLocalAddr));
+
+    MsQuicConnection Connection(Registration);
+    TEST_QUIC_SUCCEEDED(Connection.GetInitStatus());
+
+    TEST_QUIC_SUCCEEDED(Connection.StartLocalhost(ClientConfiguration, ServerLocalAddr));
+    TEST_TRUE(Connection.HandshakeCompleteEvent.WaitTimeout(TestWaitTimeout));
+
+    CxPlatSleep(1000);
```
nibanks, line 68: Let's figure out a way to not require sleeps. Try to make everything event based.

src/test/lib/PathTest.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+void
+QuicTestLocalPathChanges(
+    _In_ int Family
+    )
+{
+    MsQuicRegistration Registration;
+    TEST_QUIC_SUCCEEDED(Registration.GetInitStatus());
+
+    MsQuicAlpn Alpn("MsQuicTest");
```
nibanks, line 45: To reduce total lines of code, I've been following the pattern of simply passing `"MsQuicTest"` instead of `Alpn` everywhere.

src/test/lib/PathTest.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    MsQuicConfiguration ServerConfiguration(Registration, Alpn, ServerSelfSignedCredConfig);
+    TEST_QUIC_SUCCEEDED(ServerConfiguration.GetInitStatus());
+
+    MsQuicCredentialConfig ClientCredConfig;
+    MsQuicConfiguration ClientConfiguration(Registration, Alpn, ClientCredConfig);
```
nibanks, line 51: You can just pass `MsQuicCredentialConfig()` instead of having the separate, temporary variable.

src/test/lib/PathTest.cpp
ThadHouse: This one was weird, because the handshake completion event would be triggered before the connection actually was marked as completed. And that was causing the old style of this test to fail. I'll see if I can figure out why that wasn't correct.

nibanks: There's a difference between "handshake complete" and "handshake confirmed" based on which side of the connection you are on. The tuple isn't allowed to change until the handshake is confirmed, which we don't actually expose an event for (because the app doesn't really need to know).

ThadHouse: So if an event for that isn't exposed, I need the sleep. Thats our option.

nibanks: I believe, you can wait for the server's handshake complete and that will be enough. Otherwise, I'd at least recommend a shorter sleep + poll in a loop, so you don't always wait a whole second. And on macOS, we've even had to increase our poll/wait times because things go so slow at times.

src/test/lib/PathTest.cpp
```diff
@@ -0,0 +1,86 @@
...
+    TEST_QUIC_SUCCEEDED(Connection.GetInitStatus());
+
+    TEST_QUIC_SUCCEEDED(Connection.StartLocalhost(ClientConfiguration, ServerLocalAddr));
+    TEST_TRUE(Connection.HandshakeCompleteEvent.WaitTimeout(TestWaitTimeout));
+    TEST_NOT_EQUAL(nullptr, Context.Connection);
+    TEST_TRUE(Context.Connection->HandshakeCompleteEvent.WaitTimeout(TestWaitTimeout));
+
+    QuicAddr OrigLocalAddr;
+    TEST_QUIC_SUCCEEDED(Connection.GetLocalAddr(OrigLocalAddr));
+    QuicAddr NewLocalAddr(OrigLocalAddr);
```
nibanks, line 72: `NewLocalAddr` seems unnecessary? Just use `OrigLocalAddr` directly instead?

nibanks:
The change looks good. Besides the merge conflict that needs to be resolved, it seems like some tests are failing.

nibanks (approved):

PR #1790, opened 2021-07-01
Enable renaming the output shared binary

### LINKED ISSUE #1789: Allow renaming of MsQuic output binary
.NET would like to have a different name for the output binary.

ThadHouse:
In .NET, a different library name is preferred. This PR enables that name change to be selected at build time. .NET would prefer MsQuic.Native, but the def file errors if there is a dot in the name, so MsQuicNative is chosen.

Closes #1789 

@wfurt


wfurt:
I'm not 100% sure this is needed but I appreciate the intent. The exclude patter is actually exclude="*.Native.dll" so this will not work without the extra dot. There is some logic to determine if dll is native or not and we would run ILLinker to trim code .. and that fails if we do that on msquic.dll.

ThadHouse:
Turns out it was just a warning, so we can have the `.`. I also updated the PR so any name can be passed in.

nibanks (approved):

PR #1791, opened 2021-07-01
Added packaging for pacman (Arch-based distros)

ManickaP:
For my own testing, I added capability to create Linux package for Pacman (used by Arch-based distributions).
I'm sending this change upstream just in case you're interested in having such support.

cc: @wfurt @nibanks


ThadHouse:
The azure build failed here. Also please put an empty new line at the end of the file.

ManickaP:
The fix still might not help, it needs to be run to see. 
I tried to test it on Ubuntu 18.04 and it has older version of `tar` by default. Since it's `fpm` who's calling it and not us, we would need to manually upgrade `tar`. So this might prove not to be worth the effort and I'd just close this PR.

ManickaP:
`"Process failed: tar failed (exit code 64). Full command was:[\"tar\", \"--zstd\" ...` That's what I saw as well. This is not worth the effort, closing.

PR #1802, opened 2021-07-09
Improve Event Callback Reentrancy

nibanks:
This makes a number of changes in an attempt to improve reentrancy around MsQuic event callbacks to the app:

1. The connection now tracks if it is in a reentrant API call (e.g. the app is calling a blocking function on the MsQuic thread).
2. A couple API calls were made async, even if on the callback thread (StreamShutdown and StreamStart(async)).
3. We now debug assert if we ever try to indicate an event to the app if we are in a non-*Close reentrant API call.

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -487,6 +487,7 @@ QuicConnCloseHandle(
     )
 {
     CXPLAT_TEL_ASSERT(!Connection->State.HandleClosed);
+    Connection->State.HandleClosed = TRUE;
 
     QuicConnCloseLocally(
         Connection,
@@ -498,7 +499,6 @@ QuicConnCloseHandle(
         QuicConnOnShutdownComplete(Connection);
     }
 
-    Connection->State.HandleClosed = TRUE;
     Connection->ClientCallbackHandler = NULL;
 
     if (Connection->State.Registered) {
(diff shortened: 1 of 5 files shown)


src/core/connection.h
```diff
@@ -147,18 +147,18 @@ typedef union QUIC_CONNECTION_STATE {
...
-
         //
         // When true, this indicates that reordering shouldn't elict an
         // immediate acknowledgement.
         //
         BOOLEAN IgnoreReordering : 1;
 
+        //
+        // When true, this indicates that the connection is currently executing
+        // and API call inline (from a reentrant call on a callback).
```
ThadHouse, line 158: an not and

src/core/connection.c
```diff
@@ -1402,7 +1407,7 @@ QuicConnOnShutdownComplete(
         Event.SHUTDOWN_COMPLETE.PeerAcknowledgedShutdown =
             !Connection->State.ShutdownCompleteTimedOut;
         Event.SHUTDOWN_COMPLETE.AppCloseInProgress =
-            Connection->State.AppCloseInProgress;
+            Connection->State.HandleClosed;
```
ThadHouse, line 1410: This is a slight behavior change. It looks like HandleClosed will always be set no matter if close is called from the worker thread or an app thread, which is what this variable was used to detect.

ThadHouse: Actually, double checking this might be OK.

nibanks: Yeah. I think it was actually a bug that it wasn't set when called on a callback thread before. The new behavior is the most correct, IMO.

ThadHouse (approved):

PR #1818, opened 2021-07-14
Use root level certificate verification callback in OpenSSL

### LINKED ISSUE #1810: PEER_CERTIFICATE_RECEIVED emitted multiple times on Linux client
With simple QUIC client trying to connect to server I get following events on Linux:
```
  [conn][0x7F0B14186A50] Connection received event IDEAL_PROCESSOR_CHANGED
  [conn][0x7F0B14185E30] Connection received event PEER_CERTIFICATE_RECEIVED
  [conn][0x7F0B14185E30] Connection received event PEER_CERTIFICATE_RECEIVED
  [conn][0x7F0B14185E30] Connection received event PEER_CERTIFICATE_RECEIVED
  [conn][0x7F0B14185E30] Connection received event PEER_CERTIFICATE_RECEIVED
  [conn][0x7F0B14185E30] Connection received event STREAMS_AVAILABLE
  [conn][0x7F0B14185E30] Connection received event DATAGRAM_STATE_CHANGED
  [conn][0x7F0B14185E30] Connection received event CONNECTED
  [conn][0x7F0B14186A50] Connection received event CONNECTED
  [conn][0x7F0B14186A50] Connection received event SHUTDOWN_COMPLETE
  [conn][0x7F0B14185E30] Connection received event SHUTDOWN_COMPLETE
```
0x7F0B14185E30 is the client. 
This has two issue:
1) Functionally int breaks validation in .NET for HTTP https://github.com/dotnet/runtime/issues/55192
There is code to map the event to original HttpRequest and it fails when the callback is invoked multiple times. 
2) Verifying the certificate is not cheap. Building the X509Chain and doing the crypto is not trivial. 

This looks platform specific as same code executed on Windows gives expected outcome:
```
     System.Net.Quic.Tests.MsQuicTests.ConnectWithCertificateChain [STARTING]
  [conn][0x1E4EEC86F60] Connection received event IDEAL_PROCESSOR_CHANGED
  [conn][0x1E4EEC63D40] Connection received event STREAMS_AVAILABLE
  [conn][0x1E4EEC63D40] Connection received event DATAGRAM_STATE_CHANGED
  [conn][0x1E4EEC63D40] Connection received event PEER_CERTIFICATE_RECEIVED
  [conn][0x1E4EEC63D40] Connection received event CONNECTED
  [conn][0x1E4EEC86F60] Connection received event SHUTDOWN_COMPLETE
  [conn][0x1E4EEC63D40] Connection received event SHUTDOWN_COMPLETE
      System.Net.Quic.Tests.MsQuicTests.ConnectWithCertificateChain [FINISHED] Time: 0.9379798s
    Finished:    System.Net.Quic.Functional.Tests
```

I did some quick checking and this seems to some from OpenSSL. CxPlatTlsCertificateVerifyCallback is truly invoked multiple times.  

cc: @manickap

ThadHouse:
We were previously using a callback that would be called after openssl did all of its verification internally. This resulted in multiple callbacks. Instead, we can just completely replace OpenSSL's verification. We then can decide to call openssl's verification directly or windows cert store verification depending on OS and sec config flags. This will always result in just a single callback to the app layer, but otherwise results in no actual behavior changes.

Closes #1810

### DIFF
--- a/src/platform/tls_openssl.c
+++ b/src/platform/tls_openssl.c
@@ -190,10 +190,12 @@ CxPlatTlsVerifyCertificate(
 static
 int
 CxPlatTlsCertificateVerifyCallback(
-    int preverify_ok,
-    X509_STORE_CTX *x509_ctx
+    X509_STORE_CTX *x509_ctx,
+    void* param
     )
 {
+    UNREFERENCED_PARAMETER(param);
+    int CertificateVerified = 0;
     int status = TRUE;
     QUIC_BUFFER PortableCertificate = { 0, 0 };
     QUIC_BUFFER PortableChain = { 0, 0 };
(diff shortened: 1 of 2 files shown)


nibanks (approved):

src/platform/tls_openssl.c
```diff
@@ -1129,7 +1148,8 @@ CxPlatTlsSecConfigCreate(
     }
 
     if (CredConfigFlags & QUIC_CREDENTIAL_FLAG_CLIENT) {
-        SSL_CTX_set_verify(SecurityConfig->SSLCtx, SSL_VERIFY_PEER, CxPlatTlsCertificateVerifyCallback);
+        SSL_CTX_set_cert_verify_callback(SecurityConfig->SSLCtx, CxPlatTlsCertificateVerifyCallback, NULL);
```
anrossi, line 1151: So this will only call the verify on the leaf certificate, and not for each certificate in the chain?

ThadHouse: Correct. However inside of verify it will grab the entire chain and verify it. This was being done before as well, as we would always grab the leaf certificate at the beginning of every callback. So we were just repeating the same work for every certificate in the chain.

anrossi (approved):

PR #1826, opened 2021-07-16
Version Negotiation Extension v4

anrossi:
Update to the latest version of the Version Negotiation Extension.  
* Version Information structure is used in the transport parameter for client and server
* Stricter downgrade protection on client.
* Update transport parameter ID

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -2661,6 +2661,10 @@ QuicConnProcessPeerVersionNegotiationTP(
             return QUIC_STATUS_PROTOCOL_ERROR;
         }
         if (Connection->PreviousQuicVersion != 0) {
+            //
+            // If the client has already received a version negotiation packet, do
+            // extra validation.
+            //
             if (Connection->PreviousQuicVersion == ServerVI.ChosenVersion) {
                 QuicTraceLogConnError(
                     ServerVersionInformationPreviousVersionIsChosenVersion,
@@ -2670,6 +2674,7 @@ QuicConnProcessPeerVersionNegotiationTP(
                 QuicConnTransportError(Connection, QUIC_ERROR_VERSION_NEGOTIATION_ERROR);
                 return QUIC_STATUS_PROTOCOL_ERROR;
             }
+            uint32_t ClientChosenVersion = 0;
             for (uint32_t i = 0; i < ServerVI.OtherVersionsCount; ++i) {
                 if (Connection->PreviousQuicVersion == ServerVI.OtherVersions[i]) {
                     QuicTraceLogConnError(
@@ -2679,17 +2684,22 @@ QuicConnProcessPeerVersionNegotiationTP(
                         Connection->PreviousQuicVersion);
                     QuicConnTransportError(Connection, QUIC_ERROR_VERSION_NEGOTIATION_ERROR);
                     return QUIC_STATUS_PROTOCOL_ERROR;
-
+                }
+                if (ClientChosenVersion == 0 &&
+                    QuicVersionNegotiationExtIsVersionClientSupported(Connection, ServerVI.OtherVersions[i])) {
+                    ClientChosenVersion = ServerVI.OtherVersions[i];
                 }
             }
-            //
-            // If a client has reacted to a Version Negotiation packet, it MUST
-            // validate ... that the client would have
-            // selected the same negotiated version if it had received a Version
-            // Negotiation packet whose Supported Versions field had the same
-            // contents as the server's "Other Versions" field.
-            //
-            // HOW DO??
+            if (ClientChosenVersion == 0 || ClientChosenVersion != ServerVI.ChosenVersion) {
+                QuicTraceLogConnError(
+                    ClientChosenVersionMismatchServerChosenVersion,
+                    Connection,
+                    "Client Chosen Version doesn't match Server Chosen Version: 0x%x vs. 0x%x",
+                    ClientChosenVersion,
+                    ServerVI.ChosenVersion);
+                QuicConnTransportError(Connection, QUIC_ERROR_VERSION_NEGOTIATION_ERROR);
+                return QUIC_STATUS_PROTOCOL_ERROR;
+            }
         }
     }
     return QUIC_STATUS_SUCCESS;
@@ -2731,10 +2741,11 @@ QuicConnProcessPeerTransportParameters(
             }
         }
         if (!QuicConnIsServer(Connection) && Connection->Settings.VersionNegotiationExtEnabled &&
-            Connection->PreviousQuicVersion != 0 && !(Connection->PeerTransportParams.Flags & QUIC_TP_FLAG_VERSION_NEGOTIATION)) {
+            (Connection->State.CompatibleVersionNegotiation || Connection->PreviousQuicVersion != 0) &&
+            !(Connection->PeerTransportParams.Flags & QUIC_TP_FLAG_VERSION_NEGOTIATION)) {
             //
-            // Client responded to a version negotiation packet, but server didn't send Version Info TP.
-            // Kill the connection.
+            // Client responded to a version negotiation packet, or compatible version negotiation,
+            // but server didn't send Version Info TP. Kill the connection.
             //
             QuicConnTransportError(Connection, QUIC_ERROR_VERSION_NEGOTIATION_ERROR);
             Status = QUIC_STATUS_PROTOCOL_ERROR;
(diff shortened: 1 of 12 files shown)


src/tools/etw/quicetw.h
```diff
@@ -337,7 +337,7 @@ typedef enum QUIC_EVENT_ID_CONNECTION {
     EventId_QuicConnOutFlowStreamStats,
     EventId_QuicConnPacketStats,
     EventId_QuicConnServerResumeTicket,
-    EventId_QuicConnClientCompatibleVersionList,
+    EventId_QuicConnVNEOtherVersionList,
```
nibanks, line 340: What about quictrace in plugins?

src/core/version_neg.h
```diff
@@ -65,21 +71,13 @@ QuicVersionNegotiationExtGenerateCompatibleVersionsList(
...
-
-QUIC_STATUS
-QuicVersionNegotiationExtParseServerVerNegInfo(
+QuicVersionNegotiationExtParseVersionInfo(
     _In_ QUIC_CONNECTION* Connection,
     _In_reads_bytes_(BufferLength)
         const uint8_t* const Buffer,
     _In_ uint16_t BufferLength,
-    _Out_ QUIC_SERVER_VER_NEG_INFO* ServerVNI
+    _In_ BOOLEAN IsServer,
```
nibanks, line 79: `IsServer` should be able to be inferred from `Connection`, right?

src/core/transport_params.h
```diff
@@ -180,10 +180,10 @@ typedef struct QUIC_TRANSPORT_PARAMETERS {
     uint8_t RetrySourceConnectionIDLength;
 
     //
-    // Pointer to the Version Negotiation transport parameter opaque blob.
+    // Pointer to the version_information transport parameter opaque blob.
     //
-    const uint8_t* VersionNegotiationInfo;
-    QUIC_VAR_INT VersionNegotiationInfoLength;
+    const uint8_t* VersionInfo;
+    QUIC_VAR_INT VersionInfoLength;
```
nibanks, line 186: I know this was this way before, but why does `VersionInfoLength` need to be a `QUIC_VAR_INT`? Is this because it's decoded directly from wire format?

src/core/crypto_tls.c
```diff
@@ -62,7 +62,7 @@ typedef enum eSniNameType {
 //
 #define QUIC_TP_ID_MAX_DATAGRAM_FRAME_SIZE                  32  // varint
 #define QUIC_TP_ID_DISABLE_1RTT_ENCRYPTION                  0xBAAD  // N/A
-#define QUIC_TP_ID_VERSION_NEGOTIATION_EXT                  0x73DB  // Blob
+#define QUIC_TP_ID_VERSION_NEGOTIATION_EXT                  0xFF73DB  // Blob
```
nibanks, line 65: Can you please align all the type comments at the EOL?

src/core/connection.h
```diff
@@ -159,6 +159,13 @@ typedef union QUIC_CONNECTION_STATE {
         //
         BOOLEAN InlineApiExecution : 1;
 
+        //
+        // True once a client connection has completed a compatible version
+        // negotiation, and false otherwise. Used to prevent packets with invalid
+        // version fields from being accepted.
+        //
+        BOOLEAN CompatibleVersionNegotiation : 1;
```
nibanks, line 167: Please double check this doesn't push the size to 64-bits. If it does, you need to update the `uint32_t` flags to `uint64_t`.

src/core/connection.h
```diff
@@ -581,6 +588,12 @@ typedef struct QUIC_CONNECTION {
     //
     uint32_t PreviousQuicVersion;
 
+    //
+    // Initially-attempted QUIC version.
+    // Only populated during compatible version negotiation.
+    //
+    uint32_t InitialQuicVersion;
```
nibanks, line 595: We have a lot of state that increases the size of the connection object that's only used during the handshake. In future, it might be worth having a separate (lookaside list allocated) struct just for handshake state that we can free after the handshake completes. Something for the future. Not this PR.

src/core/connection.c
```diff
@@ -2727,12 +2741,23 @@ QuicConnProcessPeerTransportParameters(
             Status = QuicConnProcessPeerVersionNegotiationTP(Connection);
             if (QUIC_FAILED(Status)) {
                 //
-                // If the Version Negotiation Info failed to parse, indicate the failure up the stack
+                // If the Version Info failed to parse, indicate the failure up the stack
                 // to perform Incompatible Version Negotiation or so the connection can be closed.
```
nibanks, line 2745: Nit: Can you fix up the comment so maximize the first line as much as possible?

src/core/version_neg.c
```diff
@@ -171,217 +171,94 @@ QuicVersionNegotiationExtGenerateCompatibleVersionsList(
...
     _In_ QUIC_CONNECTION* Connection,
     _In_reads_bytes_(BufferLength)
         const uint8_t* const Buffer,
     _In_ uint16_t BufferLength,
-    _Out_ QUIC_CLIENT_VER_NEG_INFO* ClientVNI
+    _In_ BOOLEAN IsServer,
+    _Out_ QUIC_VERSION_INFORMATION_V1* VersionInfo
     )
 {
+    const char* const Source = (IsServer ? "Server" : "Client");
```
nibanks, line 183: Since you're logging the connection pointer with all the events already, I am actually leaning towards saying this `Source` is actually unnecessary.

src/core/version_neg.c
```diff
@@ -171,217 +171,94 @@ QuicVersionNegotiationExtGenerateCompatibleVersionsList(
...
     return QUIC_STATUS_SUCCESS;
 }
 
 _IRQL_requires_max_(DISPATCH_LEVEL)
 _Success_(return != NULL)
 const uint8_t*
-QuicVersionNegotiationExtEncodeVersionNegotiationInfo(
+QuicVersionNegotiationExtEncodeVersionInfo(
     _In_ QUIC_CONNECTION* Connection,
     _Out_ uint32_t* VNInfoLength
```
nibanks, line 256: Should this be `VInfoLength` or something similar?

src/core/version_neg.c
```diff
@@ -171,217 +171,94 @@ QuicVersionNegotiationExtGenerateCompatibleVersionsList(
...
-        CASTED_CLOG_BYTEARRAY(ServerVNI->SupportedVersionCount * sizeof(uint32_t), ServerVNI->SupportedVersions));
+        Source,
+        CASTED_CLOG_BYTEARRAY(VersionInfo->OtherVersionsCount * sizeof(uint32_t), VersionInfo->OtherVersions));
 
     return QUIC_STATUS_SUCCESS;
 }
 
 _IRQL_requires_max_(DISPATCH_LEVEL)
 _Success_(return != NULL)
 const uint8_t*
```
nibanks, line 253: We should annotate that the return is of length `VNInfoLength`.

src/core/version_neg.c
```diff
@@ -396,101 +273,85 @@ QuicVersionNegotiationExtEncodeVersionNegotiationInfo(
...
+                VILen);
             return NULL;
         }
-        VNIBuf = VersionNegotiationInfo;
+        VIBuf = VersionInfo;
 
-        CxPlatCopyMemory(VNIBuf, &Connection->Stats.QuicVersion, sizeof(Connection->Stats.QuicVersion));
-        VNIBuf += sizeof(Connection->Stats.QuicVersion);
-        VNIBuf = QuicVarIntEncode(DesiredVersionsListLength, VNIBuf);
+        _Analysis_assume_(VILen >= sizeof(uint32_t));
```
nibanks, line 289: Instead of assumes, use debug asserts. They also assume, but will crash in debug mode if invalid. But, assuming you got a build error because of this, the only reason I can think of this failing is if there is wrap around because of the addition and multiplication above. It might be worth adding an assert **there** instead to ensure wrap around would be impossible.

nibanks (commented):
Generally looks fine, though I haven't really evaluated the logic itself.

src/core/version_neg.h
anrossi: The Connection is only used for logging, and I'm a little hesitant to actually use the pointer because it'll break the tests that don't use a real connection object.

src/core/transport_params.h
anrossi: It doesn't need to be a QUIC_VAR_INT; the length field is no longer sent over the wire, so I can make this any sensible type. I think uint32_t would have the least waste, but uint8_t is probably the realistic limit on the number of versions that could be included.

src/core/connection.h
anrossi: Good call! Had to upgrade to uint64_t.

src/core/connection.h
anrossi: Fair point, the tricky part is the debugging information that could be lost by doing that. We should discuss this further.

src/core/version_neg.c
anrossi: I can understand that perspective, but this makes the logs easier to read when client and server are on the same machine (testing scenarios). In production scenarios, only one side would be in the logs, probably.

src/core/version_neg.c
anrossi: Is there an example of that in our code?

src/core/connection.h
anrossi: I created this issue https://github.com/microsoft/msquic/issues/1829 to track this further.

src/core/version_neg.c
anrossi: It's not that important. I'll remove it.

src/core/version_neg.c
```diff
@@ -394,135 +265,113 @@ QuicVersionNegotiationExtEncodeVersionNegotiationInfo(
             DesiredVersionsListLength = ARRAYSIZE(DefaultSupportedVersionsList);
         }
         //
-        // Generate Server VNI.
+        // Generate Server Version Info.
         //
-        VNILen = sizeof(uint32_t) + QuicVarIntSize(DesiredVersionsListLength) +
-            (DesiredVersionsListLength * sizeof(uint32_t));
+        VILen = sizeof(uint32_t) + (DesiredVersionsListLength * sizeof(uint32_t));
+        CXPLAT_DBG_ASSERT((DesiredVersionsListLength * sizeof(uint32_t)) + sizeof(uint32_t) > DesiredVersionsListLength + sizeof(uint32_t));
```
nibanks, line 271: The code all generally looks fine, but we need to make sure we guarantee things like this through out all this code. How do you know this can't overflow? Please do a quick pass at all your calculations like these.

nibanks (approved):

PR #1832, opened 2021-07-18
Initial Rust Support

nibanks:
Initial Rust FFI that wraps the MsQuic API.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -367,3 +367,9 @@ FodyWeavers.xsd
 
 # PGO Local Test Copy
 artifacts_server/
+
+# Rust build output folder
+target/
+
+# Cargo (Rust) lock file
+Cargo.lock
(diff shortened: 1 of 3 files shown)


thhous-msft:
I wonder if we should put the ffi code in its own subfolder, either under src or under the root. Right now its only 1 file, but its likely more will be added in the future and we don't want to polute the src root that bad. For C# I was putting everything in src/cs. But maybe root ffi/rust and ffi/cs would be better.

nibanks:
> I wonder if we should put the ffi code in its own subfolder, either under src or under the root. Right now its only 1 file, but its likely more will be added in the future and we don't want to polute the src root that bad. For C# I was putting everything in src/cs. But maybe root ffi/rust and ffi/cs would be better.

From what I could tell, the file name and location was required. As far as having more files in the future, I'm not sure we would. I prefer to keep this as minimal as possible. I could possibly move the `build.rs` into the scripts folder.

PR #1836, opened 2021-07-21
Increase Send Deadlock Detection Threshold

nibanks:
Under certain circumstances (lots of streams with small payloads) the deadlock detection can hit incorrectly. Theoretically, the loop could legally run `QUIC_MAX_FRAMES_PER_PACKET` * `QUIC_MAX_DATAGRAMS_PER_SEND` which is just under 500 right now. We'll double that for good measure.

cc @wfurt

### DIFF
--- a/src/core/send.c
+++ b/src/core/send.c
@@ -1283,7 +1283,7 @@ QuicSendFlush(
         }
 
 #if DEBUG
-        CXPLAT_DBG_ASSERT(++DeadlockDetection < 100);
+        CXPLAT_DBG_ASSERT(++DeadlockDetection < 1000);
         UNREFERENCED_PARAMETER(PrevPrevSendFlags); // Used in debugging only
         PrevPrevSendFlags = PrevSendFlags;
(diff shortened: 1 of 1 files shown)


src/core/send.c
```diff
@@ -1283,7 +1283,7 @@ QuicSendFlush(
         }
 
 #if DEBUG
-        CXPLAT_DBG_ASSERT(++DeadlockDetection < 100);
+        CXPLAT_DBG_ASSERT(++DeadlockDetection < 1000);
```
ThadHouse, line 1286: Should this be `QUIC_MAX_FRAMES_PER_PACKET * QUIC_MAX_DATAGRAMS_PER_SEND * 2` rather than a magic constant?

nibanks: No. .NET folks were seeing this hit in their stress tests and I just want to fix that. I don't expect to the other things to change really any more, so I'm fine with it this way.

ThadHouse (approved):

PR #1849, opened 2021-07-23
Support ConnectionClose in NEW_CONNECTION Event

### LINKED ISSUE #1647: Be more resilient to failure cases in New_connection event.
Look at  at Peer_stream_started event for inspiration.
ExternalOwner should be moved up to ListenerAccepted.
and revert it to false on failure.

nibanks:
Fixes #1647. Correctly handles the case where the app accepts the connection (returns success) but closes the connection. Adds an additional assert to make sure the app doesn't both reject (return failure) and close the connection.

### DIFF
--- a/src/core/listener.c
+++ b/src/core/listener.c
@@ -497,6 +497,7 @@ QuicListenerClaimConnection(
     //
 
     Connection->State.ListenerAccepted = TRUE;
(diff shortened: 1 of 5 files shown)


anrossi:
No test to exercise this scenario too?

anrossi (approved):

PR #1850, opened 2021-07-23
Cancel all sends upon silent abort

thhous-msft:


nibanks (approved):

nibanks (approved):

src/platform/platform_posix.c
```diff
@@ -40,9 +40,18 @@ uint64_t CxPlatTotalMemory;
...
+    _In_ int Line,
+    _In_z_ const char* Expr
     )
 {
+    //
+    // Pass in the error info so it can be seen in the debugger.
+    //
+    UNREFERENCED_PARAMETER(File);
+    UNREFERENCED_PARAMETER(Line);
+    UNREFERENCED_PARAMETER(Expr);
```
nibanks, line 53: Why? Don't you have a callstack?

thhous-msft: In gdb, the callstack isn't seeing macros. So for the following macro

```
#define QUIC_CID_VALIDATE_NULL(Conn, Cid)                                       \
    do {                                                                        \
        CXPLAT_DBG_ASSERT(Cid->AssignedPath == NULL);                           \
        for (uint8_t PathIdx = Conn->PathsCount - 1; PathIdx > 0; PathIdx--) {  \
            CXPLAT_DBG_ASSERT(Conn->Paths[PathIdx].DestCid != Cid);             \
        }                                                                       \
    } while (0)
#else
```

I don't actually know which assert is firing. And the parameters have been slightly clobbered so I can't dig through the variables.

nibanks (approved):

PR #1862, opened 2021-07-29
Fix Openssl build with ninja on windows

thhous-msft:


nibanks (approved):

thhous-msft:
Adding tests for this is difficult, because the ninja setup in terminal requires some manual configuration, so I'll make that a future work item. But at least this will unblock the current issue. I was able to do an OpenSSL build from VS directly.

PR #1865, opened 2021-07-30
Add support for tracking process owning connection.

anrossi:
Add support, for Windows Kernel, to track the process which owns the client connection.
Tracking is not available in server processes, since kernel listeners on the same address and port will share the socket, and attribution would go to the first listener to open the socket.

### DIFF
--- a/src/core/binding.h
+++ b/src/core/binding.h
@@ -257,6 +257,9 @@ QuicBindingInitialize(
 #ifdef QUIC_COMPARTMENT_ID
     _In_ QUIC_COMPARTMENT_ID CompartmentId,
 #endif
+#ifdef QUIC_OWNING_PROCESS
+    _In_opt_ QUIC_PROCESS OwningProcess,
+#endif
     _In_ BOOLEAN ShareBinding,
     _In_ BOOLEAN ServerOwned,
     _In_opt_ const QUIC_ADDR* LocalAddress,
--- a/src/core/binding.h
+++ b/src/core/binding.h
@@ -254,17 +254,7 @@ CXPLAT_DATAPATH_UNREACHABLE_CALLBACK QuicBindingUnreachable;
 _IRQL_requires_max_(PASSIVE_LEVEL)
 QUIC_STATUS
 QuicBindingInitialize(
-#ifdef QUIC_COMPARTMENT_ID
-    _In_ QUIC_COMPARTMENT_ID CompartmentId,
-#endif
-#ifdef QUIC_OWNING_PROCESS
-    _In_opt_ QUIC_PROCESS OwningProcess,
-#endif
-    _In_ BOOLEAN ShareBinding,
-    _In_ BOOLEAN ServerOwned,
-    _In_opt_ const QUIC_ADDR* LocalAddress,
-    _In_ const uint32_t LocalInterface,
-    _In_opt_ const QUIC_ADDR* RemoteAddress,
+    _In_ const CXPLAT_UDP_CONFIG* UdpConfig,
     _Out_ QUIC_BINDING** NewBinding
     );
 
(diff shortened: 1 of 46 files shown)


src/core/connection.c
```diff
@@ -5715,6 +5718,9 @@ QuicConnParamSet(
                 QuicLibraryGetBinding(
 #ifdef QUIC_COMPARTMENT_ID
                     Connection->Configuration->CompartmentId,
+#endif
+#ifdef QUIC_OWNING_PROCESS
+                    Connection->Configuration->OwningProcess,
```
anrossi, line 5723: Should this always be `NULL`? In this case, the connection is already started, which means its attached to a binding which already has a socket that is created.

src/platform/datapath_winkernel.c
```diff
@@ -1463,7 +1463,11 @@ CxPlatSocketCreateUdp(
             WSK_FLAG_DATAGRAM_SOCKET,
             Binding,
             &Datapath->WskDispatch,
+#ifdef QUIC_OWNING_PROCESS
+            Config->OwningProcess,
```
anrossi, line 1467: Does the socket abstraction need to store this and `AddRef`/`Release` on it?  I am assuming here that WSK will `AddRef` if it needs to keep the process handle around. 
Since this call completes inline, I don't `AddRef`/`Release`, since the Configuration still owns the pointer.

src/core/binding.h
```diff
@@ -256,6 +256,9 @@ QUIC_STATUS
 QuicBindingInitialize(
```
nibanks, line 256: I think it might just be best to have this take the `UdpConfig` directly for most args now.

src/inc/quic_platform_winkernel.h
```diff
@@ -848,6 +848,16 @@ CxPlatRandom(
     _Out_writes_bytes_(BufferLen) void* Buffer
     );
 
+//
+// Process object abstraction
+//
+
+#define QUIC_PROCESS PEPROCESS
```
nibanks, line 855: Instead of putting `QUIC_OWNING_PROCESS=1` in the VS project files for kernel, just define it here directly. Then all kernel (i.e. datapath_winkernel.c) code can just assume it's set.

src/platform/datapath_winkernel.c
nibanks: You can remove the `#ifdef` check, and I don't think the socket abstraction needs to store this. Maintaining it on the configuration is enough.

src/core/listener.c
```diff
@@ -263,6 +263,9 @@ MsQuicListenerStart(
         QuicLibraryGetBinding(
 #ifdef QUIC_COMPARTMENT_ID
             QuicCompartmentIdGetCurrent(),
+#endif
+#ifdef QUIC_OWNING_PROCESS
+            NULL,           // Owning process not support for listeners.
```
nibanks, line 268:
```suggestion
            NULL,           // Owning process not supported for listeners.
```

src/platform/datapath_winkernel.c
anrossi: I have the parameter on the `UdpConfig` also wrapped in the `#ifdef` so it doesn't confuse usage when that feature is not enabled. So this should stay.

nibanks: I disagree. You only need the ifdef in the cases where that source might not have it. The winkernel.c code has it 100% of the time.

src/core/connection.c
anrossi: Since a new socket could be opened (if the connection is going to use a new local address) this will remain.

nibanks (approved):

PR #1875, opened 2021-08-03
Return ABORTED if a stream is opened or started after the connection has been shutdown

### LINKED ISSUE #1870: Improve status codes for cases when connection/stream is closed by remote peer
This is related to discussions in https://github.com/dotnet/runtime/issues/53224 and https://github.com/dotnet/runtime/issues/56133

The `INVALID_STATE` makes sense for cases when there is user error e.g. for example the connection/stream was closed locally but then we try to use it. However, there is race condition when this case be caused by remote peer so it seems like almost any call can fail with this and then what we surface may not be really helpful. 

It would be nice if we can get some different error - either existing one like `QUIC_STATUS_ABORTED` or `QUIC_STATUS_CONNECTION_REFUSED`  or have new code if that seems more appropriate. 

cc: @manickap @geoffkizer

thhous-msft:
Closes #1870

### DIFF
--- a/src/core/api.c
+++ b/src/core/api.c
@@ -602,6 +602,7 @@ MsQuicStreamOpen(
 {
     QUIC_STATUS Status;
     QUIC_CONNECTION* Connection;
+    BOOLEAN ClosedLocally;
 
     QuicTraceEvent(
         ApiEnter,
@@ -631,8 +632,11 @@ MsQuicStreamOpen(
 
     QUIC_CONN_VERIFY(Connection, !Connection->State.Freed);
 
-    if (QuicConnIsClosed(Connection)) {
-        Status = QUIC_STATUS_INVALID_STATE;
+    if (QuicConnIsClosedLocation(Connection, &ClosedLocally)) {
+        Status =
+            ClosedLocally ?
+            QUIC_STATUS_INVALID_STATE :
+            QUIC_STATUS_ABORTED;
         goto Error;
     }
 
--- a/src/core/api.c
+++ b/src/core/api.c
@@ -602,7 +602,6 @@ MsQuicStreamOpen(
 {
     QUIC_STATUS Status;
     QUIC_CONNECTION* Connection;
-    BOOLEAN ClosedLocally;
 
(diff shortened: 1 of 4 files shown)


src/core/connection.h
```diff
@@ -671,6 +671,21 @@ QuicConnIsClosed(
...
 
+//
+// Helper for checking if a connection is currently closed with a local close
+// check.
+//
+inline
+BOOLEAN
+QuicConnIsClosedLocation(
+    _In_ const QUIC_CONNECTION * const Connection,
+    _Out_ BOOLEAN ClosedLocally
```
anrossi, line 682: Pointer?

src/core/api.c
```diff
@@ -631,8 +632,11 @@ MsQuicStreamOpen(
 
     QUIC_CONN_VERIFY(Connection, !Connection->State.Freed);
 
-    if (QuicConnIsClosed(Connection)) {
-        Status = QUIC_STATUS_INVALID_STATE;
+    if (QuicConnIsClosedLocation(Connection, &ClosedLocally)) {
```
nibanks, line 635: This PR is way over complicated. `QuicConnIsClosed` is just a wrapper checking two booleans. Just check them independently and return the error accordingly. No need for extra functions.

src/core/api.c
```diff
@@ -631,8 +631,12 @@ MsQuicStreamOpen(
 
     QUIC_CONN_VERIFY(Connection, !Connection->State.Freed);
 
-    if (QuicConnIsClosed(Connection)) {
-        Status = QUIC_STATUS_INVALID_STATE;
+    BOOLEAN ClosedLocally = Connection->State.ClosedLocally;
```
anrossi, line 634: Why is `ClosedLocally` cached in a local variable, but `ClosedRemotely` isn't?

thhous-msft: Because ClosedLocally is read twice. Additionally, this function specifically is not explicitly called on the worker, so theres a definite possibility for 2 reads here to return 2 different values.

anrossi (approved):

PR #1881, opened 2021-08-06
Create nuget packages for distribution

thhous-msft:


nibanks (approved):

src/nuget/msquic-openssl.nuspec
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
+  <metadata>
+    <id>msquic-openssl</id>
+    <version>1.7.0</version>
```
nibanks, line 5: Is it not possible to include the build ID in the version?

ThadHouse: Build Id makes things see the package as prerelease. But it might be possible there's another tag to make it work. I'll investigate before merging.

thhous-msft: I got commit hash and git repo in the package. So that should work good enough.

nibanks: Perhaps. But if that's from an internal merge commit, how helpful would it be?

thhous-msft: We run into the same issue in sourcelink for debugging. I'll update the script so it uses the correct repo. Which would then be just as helpful as build numbers.

PR #1891, opened 2021-08-09
Fix QNS builds

thhous-msft:
ALIAS targets only work in 3.18. Instead of bumping the requirement, remove the ALIAS target from OpenSSL

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -138,12 +138,6 @@ set(QUIC_LIBRARY_NAME "msquic" CACHE STRING "Override the output library name")
 
 set(BUILD_SHARED_LIBS ${QUIC_BUILD_SHARED})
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -138,12 +138,6 @@ set(QUIC_LIBRARY_NAME "msquic" CACHE STRING "Override the output library name")
 
 set(BUILD_SHARED_LIBS ${QUIC_BUILD_SHARED})
 
-# FindLTTngUST does not exist before CMake 3.6, so disable logging for older cmake versions
```
thhous-msft, line 141: Removing this, as we require 3.16 at minimum higher up, so this can never be hit.

nibanks (approved):

anrossi (approved):

PR #1895, opened 2021-08-10
Remove clog dependency from 1.1 release build

thhous-msft:
Makes onebranch builds much easier.

Not a direct cherry-pick backport because the build has had some major changes. But identical to change done in main.


nibanks (approved):

scripts/build.ps1
```diff
@@ -167,6 +167,14 @@ param (
 Set-StrictMode -Version 'Latest'
 $PSDefaultParameterValues['*:ErrorAction'] = 'Stop'
 
+if ($Parallel -lt -1) {
+    if ($IsWindows) {
+        $Parallel = -1
```
anrossi, line 172: Does `$Parallel` being -1 mean that CMake will use the platform's level of parallelism?

thhous-msft: -1 means no parallelization. 0 means platform level (That happens down around line 345)

thhous-msft: And in this case, we want no parallelization on Windows (/MP to the compiler handles that) and we want platform level of parallelization on unix platforms so it will use all cores by default.

PR #1896, opened 2021-08-10
Add shim for sendmmsg when sendmmsg does not exist

### LINKED ISSUE #1884: Support epoll Without Using sendmmsg
I am trying to port msquic to Julia
The Julia packaging framework, [BinaryBuilder](https://github.com/JuliaPackaging/BinaryBuilder.jl), is based on a relatively old version of `libc` which does not support `sendmmsg` which is used in `src/platform/datapath_epoll.c`
The library versions used by BinaryBuilder, are listed in this [page](https://github.com/JuliaPackaging/Yggdrasil/blob/master/RootFS.md)

Is it possible to build msquic for Linux without the use of `sendmmsg`?
Would it impact performance?
Is there an option to add the necessary support without relying on a new version of `glibc`?

thhous-msft:
Some users have asked to build with glibc versions that do not have sendmmsg. This change checks for the existance of sendmmsg, and if its missing implements a shim version that has identical functionality.

Closes #1884

### DIFF
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -138,7 +138,7 @@ typedef struct CXPLAT_SEND_DATA {
     //
     // The number of messages of this buffer that have been sent.
     //
-    ssize_t SentMessagesCount;
+    size_t SentMessagesCount;
 
     //
     // The send segmentation size; zero if segmentation is not performed.
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -32,7 +32,7 @@ CXPLAT_STATIC_ASSERT((SIZEOF_STRUCT_MEMBER(QUIC_BUFFER, Buffer) == sizeof(void*)
 #define CXPLAT_SENDMMSG sendmmsg
 #else
 static
-ssize_t
+int
 cxplat_sendmmsg_shim(
     int fd,
     struct mmsghdr* Messages,
@@ -44,7 +44,7 @@ cxplat_sendmmsg_shim(
     while (SuccessCount < MessageLen) {
         int Result = sendmsg(fd, &Messages[SuccessCount].msg_hdr, Flags);
         if (Result < 0) {
-            return SuccessCount == 0 ? Result : (ssize_t)SuccessCount;
+            return SuccessCount == 0 ? Result : (int)SuccessCount;
         }
         Messages[SuccessCount].msg_len = Result;
         SuccessCount++;
@@ -138,7 +138,7 @@ typedef struct CXPLAT_SEND_DATA {
     //
     // The number of messages of this buffer that have been sent.
     //
-    size_t SentMessagesCount;
+    ssize_t SentMessagesCount;
 
     //
     // The send segmentation size; zero if segmentation is not performed.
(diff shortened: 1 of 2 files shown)


src/platform/datapath_epoll.c
```diff
@@ -28,6 +28,32 @@
 CXPLAT_STATIC_ASSERT((SIZEOF_STRUCT_MEMBER(QUIC_BUFFER, Length) <= sizeof(size_t)), "(sizeof(QUIC_BUFFER.Length) == sizeof(size_t) must be TRUE.");
 CXPLAT_STATIC_ASSERT((SIZEOF_STRUCT_MEMBER(QUIC_BUFFER, Buffer) == sizeof(void*)), "(sizeof(QUIC_BUFFER.Buffer) == sizeof(void*) must be TRUE.");
 
+#ifdef HAS_SENDMMSG
+#define CXPLAT_SENDMMSG sendmmsg
```
nibanks, line 32: Why even have our own define? Just implement sendmmsg directly if missing?

src/platform/datapath_epoll.c
```diff
@@ -28,6 +28,32 @@
...
+    struct mmsghdr* Messages,
+    unsigned int MessageLen,
+    int Flags
+    )
+{
+    unsigned int SuccessCount = 0;
+    while (SuccessCount < MessageLen) {
+        int Result = sendmsg(fd, &Messages[SuccessCount].msg_hdr, Flags);
+        if (Result < 0) {
+            return SuccessCount == 0 ? Result : (ssize_t)SuccessCount;
```
anrossi, line 47: Wouldn't returning `Result` incorrectly set `SuccessfullySentMessages` in line 2592?

src/platform/datapath_epoll.c
```diff
@@ -2564,7 +2590,7 @@ CxPlatSocketSendInternal(
     while (SendData->SentMessagesCount < TotalMessagesCount) {
 
         int SuccessfullySentMessages =
-            sendmmsg(
+            CXPLAT_SENDMMSG(
```
anrossi, line 2593: `cxplat_sendmmsg_shim` returns a `ssize_t` but `SuccessfullySentMessages` is an `int`.

src/platform/datapath_epoll.c
thhous-msft: No.  The contract of sendmmsg is if the first message errors that return value is returned. Otherwise it returns the number of successful sends.

src/platform/datapath_epoll.c
thhous-msft: Ah, yeah I was looking at the wrong signature. Fixed cxplat_sendmmsg_shim

src/platform/datapath_epoll.c
thhous-msft: My worry would be that if someone attempts to run a build built without sendmmsg on a build with sendmmsg, thats a potential ODR violation that would not be fun to debug. Doing it this way means we don't interfere with the symbols.

thhous-msft: And I know its static, but I've seen the linux loader do some _VERY_ weird things in the past. I'd rather just not override a libc symbol if it can be avoided.

thhous-msft:
I manually ran the test suite and enabled this flag.

hros:
Are there benchmarking data regarding the performance degradation when sendmmsg is not available?

thhous-msft:
We don't have current benchmarking numbers, because none of our systems support any off the offloading. However, we do have some older performance data. The red jump was adding sendmmsg. The black jump here is adding GSO, which with an older libc you also won't have. So I would expect performance to be similar to where things were before the red jump.

image

PR #1902, opened 2021-08-11
reintroduce clog

chgray:
add clog back as a submodule, such that build time verification/breaks are caught 

note this doesnt require using the clog NUGET - instead it builds clog as a submodule


nibanks:
We don't want CLOG as a build time dependency any more please. We just use the update script as necessary to regenerate the files and check them in.

nibanks:
Closing this PR as a newer one already merged the latest CLOG.

PR #1908, opened 2021-08-11
Actually merge integration branch into main branch

thhous-msft:


thhous-msft:
This will need to be merged with a Merge commit. Not a Squash and Merge commit

thhous-msft:
We're likely going to do this a different way, and if we do decide to do this its out of date anyway.

PR #1915, opened 2021-08-13
Update Nupkg attributes for compliance requirements

thhous-msft:


nibanks (approved):

anrossi (approved):
I think `Microsoft.Networking.Quic` is a better name, but I don't feel strongly about it.

PR #1917, opened 2021-08-13
Initial .NET Support

thhous-msft:
Similar to (https://github.com/microsoft/msquic/pull/1832), adds a base C# interop layer to MsQuic. This one is lower level then Rust, as its currently just the FFI interface, without any high level wrappers. Also includes a small test app that is ran during each build to make sure the FFI layer works.


src/cs/lib/msquic_generated.cs
```diff
@@ -0,0 +1,2243 @@
...
+        public byte* PrivateKeyPassword;
+    }
+
+    public unsafe partial struct QUIC_CREDENTIAL_CONFIG
+    {
+        public QUIC_CREDENTIAL_TYPE Type;
+
+        public QUIC_CREDENTIAL_FLAGS Flags;
+
+        [NativeTypeName("QUIC_CREDENTIAL_CONFIG::(anonymous union at C:/Users/thhous/Documents/GitHub/msquic/src/inc/msquic.h:274:5)")]
```
nibanks, line 217: Don't check in `C:/Users/thhous/Documents/GitHub`

thhous-msft: Whoops yeah. I'll fix that. Thats part of the generator, and it doesn't currently have a way to bypass that. I'll make an issue on their end.

nibanks (approved):
Things look ok to me, but we should discuss long term costs and process for this (and Rust interop).

Issue #1923, opened 2021-08-18
Allow using an app-provided thread


DemiMarie:
### Describe the feature you'd like supported

It would be nice if MsQuic allowed apps to provide their own threads, and perform event polling themselves.

### Proposed solution

See above.

### Additional context

In some environments, such as Lua and Node.js, all callbacks must eventually be run on a single thread.  This currently requires marshaling them back to the main thread, which is less efficient than if MsQuic could integrate into the built-in event loop.  Other environments, such as Rust with Tokio, already provide their own high-performance event loops, and having to use a separate thread for QUIC would require additional locking.

nibanks:
We've discussed possibly supporting this, but never came to and hard conclusion. How would you actually use this, if we added support? There is significant work involved and we wouldn't want to do this unless something would definitely use it.

DemiMarie:
> We've discussed possibly supporting this, but never came to and hard conclusion. How would you actually use this, if we added support? There is significant work involved and we wouldn't want to do this unless something would definitely use it.

I don’t have any particular plans myself, and am not in a position where I am likely to use MsQuic in the near future.  That said, I would not be surprised if some people have ruled out MsQuic as an option because of this without filing an issue.

nibanks:
> That said, I would not be surprised if some people have ruled out MsQuic as an option because of this without filing an issue.

Perhaps. There is a reason we went with this model of owning the threads in MsQuic though. There is a lot of complexity involved in implementing a performant parallelized networking layer, and by owning the threads in MsQuic we can do all the hard work internally and the apps get it for free. If we add support for this Issue, I do expect apps that use this model to have a significant performance decrease from those that do not.

DemiMarie:
> > That said, I would not be surprised if some people have ruled out MsQuic as an option because of this without filing an issue.
> 
> Perhaps. There is a reason we went with this model of owning the threads in MsQuic though. There is a lot of complexity involved in implementing a performant parallelized networking layer, and by owning the threads in MsQuic we can do all the hard work internally and the apps get it for free. If we add support for this Issue, I do expect apps that use this model to have a significant performance decrease from those that do not.

Is the current model compatible with Node.js, for example, or would that require marshalling?

nibanks:
> Is the current model compatible with Node.js, for example, or would that require marshalling?

I have no experience with Node.js so I cannot answer that.

thhous-msft:
[...]

PR #1924, opened 2021-08-18
Refactor Worker Loop

nibanks:
Related to https://github.com/microsoft/msquic/issues/1923.

Refactors the inner worker processing loop so that it may be directly exposed to the app in the future, so that the app can drive its execution.

### DIFF
--- a/src/core/worker.c
+++ b/src/core/worker.c
@@ -396,7 +396,8 @@ QuicWorkerGetNextOperation(
 _IRQL_requires_max_(PASSIVE_LEVEL)
 void
 QuicWorkerProcessTimers(
-    _In_ QUIC_WORKER* Worker
+    _In_ QUIC_WORKER* Worker,
+    _In_ CXPLAT_THREAD_ID ThreadID
     )
 {
     //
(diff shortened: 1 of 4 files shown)


src/core/worker.c
```diff
@@ -546,145 +548,197 @@ QuicWorkerProcessConnection(
...
+    QUIC_WORKER_LOOP_REASON_WAIT,
+    QUIC_WORKER_LOOP_REASON_TIMER,
+} QUIC_WORKER_LOOP_REASON;
 
+//
+// Runs one iteration of the worker loop.
+//
+_IRQL_requires_max_(PASSIVE_LEVEL)
+QUIC_WORKER_LOOP_STATE
+QuicWorkerLoop(
```
nibanks, line 569: While not a problem with the current calling pattern, eventually, if we allow an app to choose when to call this function, it may use a single thread to drive multiple workers, and in that case, would probably round robin the workers even when one of them returns `QUIC_WORKER_LOOP_CONTINUE`. I am just wondering now this might effect the `IsActive` and queue delay variables of the worker... I think it's still reasonable to keep things as they are now, but it's worth some additional thought.

thhous-msft:
This doesn't look like it regresses perf, so LGTM.

thhous-msft (approved):

PR #1925, opened 2021-08-18
Use os certificate validation on macOS and iOS

thhous-msft:
Similar to Windows, use the OS provided certificate validation functions rather than the default openssl ones. This is often a better choice, as the OS certificates (especially on mac and windows) are often more up to date.


nibanks (approved):

src/platform/darwin_openssl.c
```diff
@@ -0,0 +1,195 @@
...
+    SecCertificateRef Certificate = NULL;
+    OSStatus Status = 0;
+    CFMutableArrayRef PolicyArray = NULL;
+    SecTrustRef TrustRef = NULL;
+    CFStringRef SNIString = NULL;
+    SecPolicyRef SSLPolicy = NULL;
+    SecPolicyRef RevocationPolicy = NULL;
+    int OpenSSLCertLength = 0;
+
+    OpenSSLCertLength = i2d_X509(X509Cert, &OpenSSLCertBuffer);
```
anrossi, line 53: Is this certificate received over the wire? is `i2d` hardened against untrusted input?

thhous-msft: Yes, this is over the wire. However, since X509Cert is typed coming out of openssl, I'm assuming both its been prevalidated and i2d_X509 is hardened against bad input. We do the same thing on windows currently too.

anrossi (approved):

PR #1926, opened 2021-08-18
Fix quicip not waiting for connection shutdown before deleting context

### LINKED ISSUE #1911: quicipclient memory write error
OS:windows 10  21H1

update time : 2021-08-13
boringssl
msquic

Compile and debug tools: visual studio 2019

quicipserver -selfsign:true

quicipclient -target:xxx.xxx.xxx.xxx -local:xxx.xxx.xxx.xxx -unsecure

Error:Memory write error 

Stack:
 	quicipclient.exe!memcpy() line 237
>	quicipclient.exe!ClientStreamCallback(QUIC_HANDLE * Stream, void * _Context, QUIC_STREAM_EVENT * Event) line 60	C++
 	msquic.dll!QuicStreamIndicateEvent(QUIC_STREAM * Stream, QUIC_STREAM_EVENT * Event) line 410	C
 	msquic.dll!QuicStreamRecvFlush(QUIC_STREAM * Stream) line 776	C
 	msquic.dll!QuicConnDrainOperations(QUIC_CONNECTION * Connection) line 6865	C
 	msquic.dll!QuicWorkerProcessConnection(QUIC_WORKER * Worker, QUIC_CONNECTION * Connection) line 487	C
 	msquic.dll!QuicWorkerThread(void * Context) line 579	C

code: 
file quicip.h
function ClientStreamCallback
QUIC_IP_LOOKUP* Context = (QUIC_IP_LOOKUP*)_Context;
 switch (Event->Type) {
        if (Event->RECEIVE.AbsoluteOffset + Event->RECEIVE.TotalBufferLength <= sizeof(QUIC_ADDR)) {
            uint64_t Offset = Event->RECEIVE.AbsoluteOffset;
            for (uint32_t i = 0; i < Event->RECEIVE.BufferCount; ++i) {
                memcpy(                                                        //bug
                    ((uint8_t*)Context->PublicAddress) + Offset,
                    Event->RECEIVE.Buffers[i].Buffer,
                    Event->RECEIVE.Buffers[i].Length);
                Offset += Event->RECEIVE.Buffers[i].Length;
            }
        }
        break;

analyze: 
Context->PublicAddress pointer is modified 

The MsQuicGetPublicIPEx function in quicip.h
if (QUIC_FAILED(Context.Status = Context.MsQuic->ConfigurationOpen(Registration, &Alpn, 1, &Settings, sizeof(Settings), NULL, &Context.Configuration))) {
Set a breakpoint, after the break, break when the value of Context.PublicAddress changes 

Finally, it will be interrupted to WaitForSingleObject, its upper layer is the MsQuicRegistrationClose function of the registration.c file, CxPlatRundownReleaseAndWait(&Registration->Rundown); 

WaitForSingleObject:
00007FFC7B491940  mov         qword ptr [rsp+8],rbx  
00007FFC7B491945  mov         dword ptr [rsp+18h],r8d  
00007FFC7B49194A  push        rsi  
00007FFC7B49194B  push        rdi  
00007FFC7B49194C  push        r14  
00007FFC7B49194E  sub         rsp,80h  
00007FFC7B491955  mov         ebx,r8d  
00007FFC7B491958  mov         esi,edx  
00007FFC7B49195A  mov         rdi,rcx  
00007FFC7B49195D  mov         qword ptr [rsp+30h],48h          << modify Context->PublicAddress
00007FFC7B491966  mov         dword ptr [rsp+38h],1

thhous-msft:
Fixes #1911

### DIFF
--- a/src/tools/ip/quicip.h
+++ b/src/tools/ip/quicip.h
@@ -196,7 +196,7 @@ MsQuicGetPublicIPEx(
 
     {
         std::unique_lock Lock{Context.DoneMutex};
-        Context.DoneEvent.wait_for(Lock, std::chrono::seconds(10), [&]{return Context.IsDone;});
+        Context.DoneEvent.wait(Lock, [&]{return Context.IsDone;});
     }
 
 Error:
--- a/src/tools/ip/quicip.h
+++ b/src/tools/ip/quicip.h
@@ -115,7 +115,8 @@ ClientConnectionCallback(
         break;
     case QUIC_CONNECTION_EVENT_SHUTDOWN_COMPLETE:
         {
-            std::scoped_lock Lock{Context->DoneMutex};
+            Context->MsQuic->ConnectionClose(Context->Connection);
+            std::lock_guard Lock{Context->DoneMutex};
             Context->IsDone = true;
             Context->DoneEvent.notify_all();
         }
@@ -198,13 +199,7 @@ MsQuicGetPublicIPEx(
         Context.DoneEvent.wait_for(Lock, std::chrono::seconds(10), [&]{return Context.IsDone;});
     }
 
-    Context.MsQuic->ConnectionShutdown(Context.Connection, QUIC_CONNECTION_SHUTDOWN_FLAG_NONE, 0);
-
 Error:
-
-    if (Context.Connection) {
-        Context.MsQuic->ConnectionClose(Context.Connection);
-    }
 
     if (Context.Configuration) {
         Context.MsQuic->ConfigurationClose(Context.Configuration);
--- a/src/tools/ip/quicip.h
+++ b/src/tools/ip/quicip.h
@@ -17,7 +17,8 @@ TODO:
 #pragma once
 
 #include "msquic.h"
-#include "msquicp.h"
+#include <mutex>
+#include <condition_variable>
 
 #ifdef ENABLE_QUIC_PRINTF
 #include <stdio.h>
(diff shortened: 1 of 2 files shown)


src/tools/ip/quicip.h
```diff
@@ -135,7 +143,7 @@ MsQuicGetPublicIPEx(
     const QUIC_BUFFER Alpn = { sizeof("ip") - 1, (uint8_t*)"ip" };
     const uint16_t UdpPort = 4444;
 
-    QUIC_SETTINGS Settings{0};
+    QUIC_SETTINGS Settings = {0};
```
anrossi, line 146: I don't think these are equivalent in C++. I believe the old way will correctly zero initialize all members of the struct, while the new will (on some compilers) only zero-initialize the first member.

anrossi (approved):

src/tools/ip/quicip.h
```diff
@@ -110,7 +114,11 @@ ClientConnectionCallback(
         }
         break;
     case QUIC_CONNECTION_EVENT_SHUTDOWN_COMPLETE:
-        Context->MsQuic->ConnectionClose(Connection);
+        {
+            std::scoped_lock Lock{Context->DoneMutex};
```
nibanks, line 118: Not sure what the difference is, but in spinquic.cpp we use `std::lock_guard` instead. Perhaps we should use the same to be consistent in our code base?

nibanks: Actually, I don't quite understand the need for a lock at all. All you should need to do is have the `DoneEvent` and wait on it forever. Keep the connection clean up logic in SHUTDOWN_COMPLETE. No need to explicitly call shutdown. No need for `IsDone`. We already set the idle timeout to 2 seconds, so that will kill the connection if it gets stuck.

thhous-msft: You can't wait on a condition_variable in C++ without a mutex. Its a requirement for the API. And you need IsDone otherwise you have a race condition where the connection closes before you actually start the wait. I'll move the close back however.

nibanks (approved):

PR #1928, opened 2021-08-19
Don't allow stateless operations on a binding until fully initialized

thhous-msft:
A small race condition exists if a server binding is created and runs out of memory at a specific point in the process. If the socket and at least 1 receive buffer gets allocated, a packet can be received on that receive queue. This receive can post a stateless operation, even if one of the later receive buffers fails to allocate, and cause the socket and binding to be cleaned up early. During binding cleanup, there will be a stateless operation in the binding queue, causing an assertion to be hit. 

The solution to this is to not allow a stateless operation to queue until the binding has been fully initialized. The packet will be dropped instead, which is fine.

### DIFF
--- a/src/core/binding.c
+++ b/src/core/binding.c
@@ -166,10 +166,7 @@ QuicBindingInitialize(
 
     *NewBinding = Binding;
     Status = QUIC_STATUS_SUCCESS;
-
-    CxPlatDispatchLockAcquire(&Binding->StatelessOperLock);
-    Binding->Initialized = TRUE;
-    CxPlatDispatchLockRelease(&Binding->StatelessOperLock);
+    InterlockedIncrement16(&Binding->Initialized);
 
 Error:
 
--- a/src/core/binding.c
+++ b/src/core/binding.c
@@ -167,6 +167,10 @@ QuicBindingInitialize(
     *NewBinding = Binding;
     Status = QUIC_STATUS_SUCCESS;
 
+    CxPlatDispatchLockAcquire(&Binding->StatelessOperLock);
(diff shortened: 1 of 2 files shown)


src/core/binding.c
```diff
@@ -167,6 +167,10 @@ QuicBindingInitialize(
     *NewBinding = Binding;
     Status = QUIC_STATUS_SUCCESS;
 
+    CxPlatDispatchLockAcquire(&Binding->StatelessOperLock);
```
thhous-msft, line 170: A lock might be a bit too extreme here, but we do need a memory barrier here somehow, otherwise platforms like arm will have an issue. We don't currently have any CXPLAT code for explicit memory barriers.

anrossi: can't use interlocked exchange?

thhous-msft: Yeah I can do that and make it not a bitfield. I might just do that.

anrossi (approved):

src/core/binding.c
```diff
@@ -166,6 +166,7 @@ QuicBindingInitialize(
 
     *NewBinding = Binding;
     Status = QUIC_STATUS_SUCCESS;
+    InterlockedIncrement16(&Binding->Initialized);
```
nibanks, line 169: I believe a simpler solution would be to just use `RefCount` instead. It's not incremented until the library has accepted the binding (after initialize). So, down below just check if `RefCount > 0`.

thhous-msft: The ref count is initialized slightly later, but that should be fine to use as well.

nibanks (approved):

PR #1932, opened 2021-08-20
Change congestion control api to use "TimeNow" in microsecond

Wizmann:
We may lost the precision of the `TimeNow` timestamp if we do the `US_TO_MS` convertion in `loss_detection` layer, and this may not be compatible with other congestion control algorithm.


nibanks (approved):
Are you working on something that will build on top of this, I assume?

Wizmann:
Thank you for approve that change.

Yes, I'm working on something for congestion control algorithm.

PR #1950, opened 2021-08-30
bugfix: create new source CID in handshake retry

qzhuyan:
fix issue: microsoft#1938

During the connection handshake, server side has only one source CID in the
CID set. More CIDs will be added to CID set only when the connection is
connected (TLS handshake complete).

If the server needs to send a 'Quic Retry Packet', it runs out of CID and
cause segfault (next source CID, NULL Pointer) because RFC 9000 requires
to use a new source CID.

This commit attempts to fix the issue by creating a new CID when there is no
free CID in the CID set of the connection.
bugfix: create new source CID in handshake retry

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -2308,7 +2308,11 @@ QuicConnGenerateLocalTransportParameters(
             Connection->OrigDestCID = NULL;
 
             if (Connection->State.HandshakeUsedRetryPacket) {
-                CXPLAT_DBG_ASSERT(SourceCid->Link.Next != NULL);
+                if (SourceCid->Link.Next == NULL) {
+                    if (!QuicConnGenerateNewSourceCid(Connection, FALSE)) {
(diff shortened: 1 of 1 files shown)


nibanks:
Thanks @qzhuyan for putting this together. I have a fix internally that I am testing and hope to send it out soon. For a temporary work around for your problem, you can configure the retry limit to the max to mostly disable the feature.

src/core/connection.c
```diff
@@ -2308,7 +2308,11 @@ QuicConnGenerateLocalTransportParameters(
             Connection->OrigDestCID = NULL;
 
             if (Connection->State.HandshakeUsedRetryPacket) {
-                CXPLAT_DBG_ASSERT(SourceCid->Link.Next != NULL);
+                if (SourceCid->Link.Next == NULL) {
```
qzhuyan, line 2311: Oops, I misread the `Link.Next` which is PrevSourceCid

Issue #1938, opened 2021-08-25
Core dump NULL pointer PrevSourceCid when handshake retry


qzhuyan:
### Describe the bug

Get core dump due to PrevSourceCid NULL pointer during handshake retry

msquic: 1.6
OS: ubuntu 20.04
CMAKE_BUILD_TYPE "RelWithDebInfo"

I am testing an overload situation that application has no enough resources to handle new connections

Looks like there are some race of handling `Source Connection CID`.

### Steps to reproduce the behavior

Start 3000 concurrent connections (3000 conns/s), few of them success at the beginning than get a segfault. 

If I start with fewer concurrent connections like 1000 or reduce the rate to 100 conns/s, there is no issue.

both client and server are msquic 1.6.

Coredump backtrace and lttng trace should explain all.

gdb backtrace:
```
Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
--Type <RET> for more, q to quit, c to continue without paging--
Core was generated by `emqx -P 2097152 -Q 1048576 -e 262144 -spp true -A 4 -IOt 4 -SDio 8 -- -root /ho'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f68efb2a82c in QuicConnGenerateLocalTransportParameters (Connection=Connection@entry=0x7f688c094720,
    LocalTP=LocalTP@entry=0x7f68b6ffcbb0) at /home/ezhuwya/repo/quic/msquic/src/core/connection.c:2319
2319	                LocalTP->RetrySourceConnectionIDLength = PrevSourceCid->CID.Length;
[Current thread is 1 (Thread 0x7f68b6ffd700 (LWP 3455261))]
(gdb) bt
#0  0x00007f68efb2a82c in QuicConnGenerateLocalTransportParameters (Connection=Connection@entry=0x7f688c094720,
    LocalTP=LocalTP@entry=0x7f68b6ffcbb0) at /home/ezhuwya/repo/quic/msquic/src/core/connection.c:2319
#1  0x00007f68efb3353b in QuicConnSetConfiguration (Configuration=0x7f6938083100, Connection=0x7f688c094720)
    at /home/ezhuwya/repo/quic/msquic/src/core/connection.c:2420
#2  QuicConnSetConfiguration (Connection=0x7f688c094720, Configuration=0x7f6938083100)
    at /home/ezhuwya/repo/quic/msquic/src/core/connection.c:2347
[...]

PR #1951, opened 2021-08-30
Fix NULL Pointer Read in Stateless Retry Scenario (on Server)

nibanks:
Fixes https://github.com/microsoft/msquic/issues/1938 . Essentially, if the server takes too long to set the configuration on the connection, it's possible for MsQuic to clean up some state that it still needs later (and ends up reading a NULL pointer).

The test code was added first, to 100% reproduce the original issue. Next the core code that cleans up the state was moved to the handshake complete instead of on first usage. Verified new tests now passed.


src/core/crypto.c
```diff
@@ -1517,6 +1517,30 @@ QuicCryptoProcessTlsCompletion(
                 "Handshake confirmed (server)");
             QuicSendSetSendFlag(&Connection->Send, QUIC_CONN_SEND_FLAG_HANDSHAKE_DONE);
             QuicCryptoHandshakeConfirmed(&Connection->Crypto);
+
+            //
+            // Take this opportinuty to clean up the client chosen initial CID.
+            // It will be the second one in the list.
+            //
+            CXPLAT_DBG_ASSERT(Connection->SourceCids.Next != NULL);
```
anrossi, line 1525: should these be asserts? Or should they be runtime checks as well?

nibanks: Asserts should be enough.

anrossi (approved):

qzhuyan:
I can confirm cherry-pick d62ee62 to v1.6.0 solved issue in #1938

nibanks:
> I can confirm cherry-pick [d62ee62](https://github.com/microsoft/msquic/commit/d62ee623ded780bb311d21c17f244715d15edd30) to v1.6.0 solved issue in #1938

Great! Thanks @qzhuyan!

PR #1958, opened 2021-09-07
Add Windows Server 2022 Tests for User Mode Schannel

thhous-msft:
- Moves most things from `windows-latest` to explicitly `windows-2019` so we don't get broken when latest becomes 2022.
- Adds jobs to run schannel tests on `windows-2022` in parallel to the ones we run on our selfhosted pool.
- Fix up things to make it all work.


nibanks:
Hopefully this PR is good now. Had to disable a lot of schannel tests on 2022. For pretty much all the build jobs we pushed explicitly to 2019, we should send PRs to try them on 2022. If they work, move them. If they don't, open bugs on virtual-environments (like I did for https://github.com/microsoft/msquic/pull/2025).

PR #1959, opened 2021-09-07
Increase memory limit to physical memory limit on linux

thhous-msft:
Partial solution to #1960 in the running on hardware case.

### DIFF
--- a/src/platform/platform_posix.c
+++ b/src/platform/platform_posix.c
@@ -18,6 +18,8 @@ Environment:
 #include "quic_trace.h"
 #ifdef CX_PLATFORM_LINUX
 #include <sys/syscall.h>
+#include <sys/types.h>
+#include <sys/sysinfo.h>
 #endif
 #include <dlfcn.h>
 #include <fcntl.h>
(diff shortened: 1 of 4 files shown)


src/platform/platform_posix.c
```diff
@@ -159,14 +161,33 @@ CxPlatInitialize(
...
     RandomFd = open("/dev/urandom", O_RDONLY|O_CLOEXEC);
     if (RandomFd == -1) {
-        return (QUIC_STATUS)errno;
+        Status = (QUIC_STATUS)errno;
+        goto Exit;
+    }
+
+#ifdef CX_PLATFORM_LINUX
+    struct sysinfo MemInfo;
+    if (sysinfo(&MemInfo) != 0) {
```
nibanks, line 174: Think we should log failures for this or `RandomFd` above?

thhous-msft: Probably. I can add one.

nibanks (approved):

thhous-msft:
Replaced by #1968

PR #1968, opened 2021-09-08
Use correct physical memory limits on posix platforms, with cgroups to handle containers

### LINKED ISSUE #1960: Memory limit hardcoded or incorrect on posix platforms
Currently, we have a hardcoded limit of 1 GB of memory on posix platforms. This shouldn't be hardcoded. We have an open PR to get this working on linux hardware, but its not going to work correctly in containers, and also does not work on macos. Both of those cases need to be fixed.

thhous-msft:
Replaces #1959 and closes #1960. Will check to see if memory is being restricted, and if so return restricted memory, otherwise return physical memory. Works on all posix based platforms. Has been tested on lots of physical hardware. Need to fire up a limited container to check the cgroup code. cgroup code needs to be cleaned up. Basically a direct port from .NET


src/platform/cgroup.c
```diff
@@ -0,0 +1,479 @@
...
+
+/*++
+
+Module Name:
+
+    cgroup.cpp
+
+Abstract:
+    Read the memory limit for the current process
+--*/
```
nibanks, line 12: Fix this up?

thhous-msft: > cgroup code needs to be cleaned up

src/platform/platform_posix.c
```diff
@@ -141,32 +156,95 @@ CxPlatSystemLoad(
 
     CXPLAT_FREE(ProviderFullPath, QUIC_POOL_PLATFORM_TMP_ALLOC);
 
-#ifdef DEBUG
-    CxPlatform.AllocFailDenominator = 0;
-    CxPlatform.AllocCounter = 0;
-#endif
+    QuicTraceLogInfo(
+        PosixLoaded,
+        "[ dso] Loaded");
```
nibanks, line 161: `dso`?

thhous-msft: Dynamic shared object. On windows user mode this logs dll, and kernel logs sys, so dso makes the most sense for posix as its a common name on linux and macos.

nibanks:
```
/home/vsts/work/1/s/src/platform/cgroup.c:513:7: error: do not use 'else' after 'return' [readability-else-after-return,-warnings-as-errors]
    } else if (CGroupVersion == 1) {
```

nibanks (approved):

PR #1989, opened 2021-09-15
Some Congestion Control Cleanup

nibanks:
A bit of clean up and refactoring to (IMO) simplify the new CC abstraction for multiple algorithms. cc @wizmann


src/core/congestion_control.c
```diff
@@ -27,106 +27,15 @@ QuicCongestionControlInitialize(
...
-            QUIC_CONNECTION* Connection = QuicCongestionControlGetConnection(Cc);
-            QuicTraceLogConnWarning(
-                InvalidCongestionControlAlgorithm,
-                Connection,
-                "Unknown congestion control algorithm: %d, fallback to Cubic",
-                Settings->CongestionControlAlgorithm);
-            CubicCongestionControlInitialize(Cc, Settings);
-            break;
-        }
+    default:
```
thhous-msft, line 30: This is weird. I've never seen the default case first. Just looks really odd for me. Especially with only saving 1 line.

nibanks: It's just a case statement. I've seen this way plenty of times.

Wizmann:
I like the change. It makes everything simple and clear.
LGTM.

thhous-msft (approved):

src/core/congestion_control.c
maolson-msft: @nibanks , I don't understand this change either. What is "__fallthrough"?

thhous-msft: __fallthrough is just an attribute to make falling through a case statement not warn.

maolson-msft: Oh I see, I was confused because I didn't realize there was fallback logic and thought the code was expected to log an error trace and bail.

PR #1990, opened 2021-09-15
Add Path Change Perf Counters

### LINKED ISSUE #1819: Add Perf Counters for Path Changes
We don't currently have a lot of insights into how often path changes are happening, if they are failing, etc. The simplest way (at least on Windows) to get this information is to add a couple of performance counters. We should probably add the following:

- Path change/challenge successes
- Path change/challenge failure
- Stateless reset's sent

nibanks:
Closes #1819. Adds a few perf counters to track path changes.


src/bin/winkernel/msquicpcw.c
```diff
@@ -129,6 +129,10 @@ MsQuicPcwStartup(
         { 30, 0, QUIC_PERF_COUNTER_WORK_OPER_QUEUE_DEPTH * sizeof(int64_t), sizeof(int64_t)},
         { 31, 0, QUIC_PERF_COUNTER_WORK_OPER_QUEUED * sizeof(int64_t), sizeof(int64_t)},
         { 32, 0, QUIC_PERF_COUNTER_WORK_OPER_COMPLETED * sizeof(int64_t), sizeof(int64_t)},
+        { 32, 0, QUIC_PERF_COUNTER_PATH_VALIDATED * sizeof(int64_t), sizeof(int64_t)},
```
thhous-msft, line 132: The first numbers here need to be incrementing.

src/inc/msquic.h
```diff
@@ -464,6 +464,10 @@ typedef enum QUIC_PERFORMANCE_COUNTERS {
     QUIC_PERF_COUNTER_WORK_OPER_QUEUE_DEPTH,// Current worker operations queued.
     QUIC_PERF_COUNTER_WORK_OPER_QUEUED,     // Total worker operations queued ever.
     QUIC_PERF_COUNTER_WORK_OPER_COMPLETED,  // Total worker operations processed ever.
+    QUIC_PERF_COUNTER_PATH_VALIDATED,       // Total path challenges that succeed ever.
```
thhous-msft, line 467: In QUIC_PARAM_GLOBAL_PERF_COUNTERS, we verify the buffer for this is exactly QUIC_PERF_COUNTER_MAX. So we need to update the logic there to allow both old and new sizes, otherwise its a breaking change.

nibanks: I see the following in there:
```c
        if (*BufferLength < QUIC_PERF_COUNTER_MAX * sizeof(int64_t)) {
            //
            // Copy as many counters will fit completely in the buffer.
            //
            *BufferLength = (*BufferLength / sizeof(int64_t)) * sizeof(int64_t);
        } else {
            *BufferLength = QUIC_PERF_COUNTER_MAX * sizeof(int64_t);
        }
```
This should be fine, right?

thhous-msft: Yup I misread.

thhous-msft (approved):

PR #1999, opened 2021-09-16
Bypass TDI to make IP_PKTINFO work properly

zx2c4:
TDI/TDX only examines the cmsghdr of TDI IRPs from the
Options/OptionsLength fields if UserBuffer==(PVOID)8937731. This was
never documented and will probably never change, which in turn means
that TDI drivers that filter and reinject packets will strip the cmsghdr
unknowingly. TDI is a dead technology anyway, and bypassing it shouldn't
interfere with reasonable systems.

This mirrors https://git.zx2c4.com/wireguard-nt/commit/?id=c331ee4f18e98b7063dc74e3d8b0ab556cff3a96

### DIFF
--- a/src/platform/datapath_winkernel.c
+++ b/src/platform/datapath_winkernel.c
@@ -837,6 +837,7 @@ CxPlatDataPathInitialize(
         &NPI_WSK_INTERFACE_ID,
         WSK_EVENT_RECEIVE_FROM
     };
+    ULONG NoTdi = WSK_TDI_BEHAVIOR_BYPASS_TDI;
 
     UNREFERENCED_PARAMETER(TcpCallbacks);
     if (NewDataPath == NULL) {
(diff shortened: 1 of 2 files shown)


src/platform/datapath_winkernel.c
```diff
@@ -962,6 +963,27 @@ CxPlatDataPathInitialize(
...
+            NULL);
+    if (QUIC_FAILED(Status)) {
+        QuicTraceEvent(
+            LibraryErrorStatus,
+            "[ lib] ERROR, %u, %s.",
+            Status,
+            "WskControlClient WSK_TDI_BEHAVIOR");
+        goto Error;
+    }
+
```
nibanks, line 985: nit: please remove the extra newline :D

nibanks (approved):
LGTM. Thanks for making this change!

src/platform/datapath_winkernel.c
zx2c4: Done.

src/platform/datapath_winkernel.c
```diff
@@ -962,6 +963,26 @@ CxPlatDataPathInitialize(
...
+            NULL,
+            NULL,
+            NULL);
+    if (QUIC_FAILED(Status)) {
+        QuicTraceEvent(
+            LibraryErrorStatus,
+            "[ lib] ERROR, %u, %s.",
+            Status,
+            "WskControlClient WSK_TDI_BEHAVIOR");
+        goto Error;
```
nibanks, line 983: Your wireguard PR ignores the error here. Should we just log and continue?

zx2c4: Your call. On one hand, I kind of doubt that the MSDN mention of removing it will actually ever happen. On the other hand, if it does happen, I'd prefer the code to age well. But maybe you would be okay with just changing the code when the time comes. So that's up to you.

If you do think I should ignore the error, what's prefix would you want? `[ lib] WARN, `? Or keep it at `[ lib] ERROR`, but not `goto ERROR;` subsequently?

nibanks: You can just remove the `goto Error;` line. Keep the log as is.

zx2c4: Works for me.

nibanks (approved):

thhous-msft (approved):

PR #2000, opened 2021-09-16
Support non-RSA keys in OpenSSL-CAPI interop layer

anrossi:


nibanks:
Do we have the necessary automated tests to validate all these changes? If not, what would it take to get there? There are a lot of changes I don't have the expertise to validate by CR and it'd prefer to have tests do the rest.

PR #2002, opened 2021-09-17
Add UWP nuget package

thhous-msft:


src/inc/msquic_winuser.h
```diff
@@ -24,7 +24,7 @@
 #endif
 
 #include <windows.h>
-#include <ws2def.h>
+#include <winsock2.h>
```
thhous-msft, line 27: According to the header at the top of ws2def.h, user mode applications should include winsock2.h rather than ws2def.h. In UWP mode, Windows.h has to be included without WIN32_LEAN_AND_MEAN if you want to actually build apps, which causes ws2def include to fail. However, winsock2 does not fail. This also should not cause any user application breaks either.

nibanks (approved):

PR #2006, opened 2021-09-19
Refactor for Ack/Loss events passing to CC

Wizmann:
Refactor for Ack/Loss event to replace passing (maybe a lot in the future) parameters to CC functions: `QuicCongestionControlOnDataAcknowledged` and `QuicCongestionControlOnDataLost`.

This will help later updates for loss detection passing more information to CC algo.


nibanks:
@Wizmann what other parameters do you expect to pass in the future (I assume for BBR)? I have a few code specific comments I will put in next, but generally, I'd like to have a better idea of why we should do this.

src/core/ack_event.h
```diff
@@ -0,0 +1,41 @@
+/*++
+
+    Copyright (c) Microsoft Corporation.
+    Licensed under the MIT License.
+
+--*/
+
+typedef struct QUIC_ACK_EVENT {
```
nibanks, line 8: Instead of having separate files for ack_event and loss_event, I'd rather just define these at the top of congestion_control.h. Also, I'd rather not have the helper functions to create the events. Just declare them directly in the calling code.

src/core/loss_detection.c
```diff
@@ -1153,12 +1156,15 @@ QuicLossDetectionDiscardPackets(
...
-                LossDetection->LargestAck,
-                AckedRetransmittableBytes,
-                Path->SmoothedRtt)) {
+
+        QUIC_ACK_EVENT AckEvent = CreateQuicAckEvent();
+        AckEvent.TimeNow = TimeNow;
+        AckEvent.LargestPacketNumberAcked = LossDetection->LargestAck;
+        AckEvent.NumRetransmittableBytes = AckedRetransmittableBytes;
+        AckEvent.SmoothedRtt = Path->SmoothedRtt;
+        AckEvent.IsImplicit = TRUE;
```
nibanks, line 1165: `IsImplicit` isn't being used right now, right? Is this for some future CC? I'd rather not add things that are currently unnecessary, and instead delay adding them until they are required,

Wizmann: Agreed.

Wizmann:
Yes, it's for BBR, but not only for BBR. Actually it's for a better encapsulation for all CC algos.
I think the change makes it more convenient to add a field in `AckEvent`/`LossEvent` for one CC without changing the code of all other CCs.

nibanks:
> Yes, it's for BBR, but not only for BBR. Actually it's for a better encapsulation for all CC algos.
> I think the change makes it more convenient to add a field in `AckEvent`/`LossEvent` for one CC without changing the code of all other CCs.

Ok. I'm fine with it if you can make changes for my other comments please. Thanks!

src/core/ack_event.h
Wizmann: In my plan, there will be some optional fields in the event (something like `QUIC_SETTINGS`).
It's necessary to create a "factory function" to initialize all the field.

nibanks: But it's a bit overkill when there are only one or two callers anyways. I'd rather just have each caller initialize as necessary. If it really gets to be too much, we can refactor further to add the helper functions. Until then, the simpler (less code) the better.

Wizmann: OK.

nibanks (approved):

PR #2009, opened 2021-09-20
Add C++ CxPlat Improvements

nibanks:


src/inc/msquic.hpp
```diff
@@ -424,6 +448,13 @@ class MsQuicCredentialConfig : public QUIC_CREDENTIAL_CONFIG {
         memset(thisStruct, 0, sizeof(QUIC_CREDENTIAL_CONFIG));
         Flags = _Flags;
     }
+    MsQuicCredentialConfig(QUIC_CREDENTIAL_FLAGS _Flags, const QUIC_CERTIFICATE_HASH* _CertificateHash) {
+        QUIC_CREDENTIAL_CONFIG* thisStruct = this;
+        memset(thisStruct, 0, sizeof(QUIC_CREDENTIAL_CONFIG));
```
anrossi, line 453: shouldn't this be accomplished with an initializer?

anrossi (approved):

src/inc/msquic.hpp
thhous-msft: No, because we're inheriting a C struct here.

PR #2017, opened 2021-09-21
Grab a UDP Port Reservation

nibanks:
Explicitly grab a UDP port reservation to make sure other processes fail to use the same port in parallel.


thhous-msft:
Can interference happen between kernel and user in this same way? Is there a way to add this behavior in kernel as well?

thhous-msft (approved):

anrossi (approved):

maolson-msft:
I thought you needed to associate each socket with the port reservation with SIO_ASSOCIATE_PORT_RESERVATION before binding?

nibanks:
> I thought you needed to associate each socket with the port reservation with SIO_ASSOCIATE_PORT_RESERVATION before binding?

It doesn't look like we do? I'm not getting any bind failures on the other sockets? What should I do?

src/platform/datapath_winuser.c
```diff
@@ -1789,6 +1789,41 @@ QUIC_DISABLED_BY_FUZZER_START;
...
 
+        if (i == 0 &&
+            Config->LocalAddress &&
+            Config->LocalAddress->Ipv4.sin_port != 0) {
+            //
+            // Create a port reservation for the local port.
+            //
+            INET_PORT_RESERVATION_INSTANCE PortReservation;
+            INET_PORT_RANGE PortRange;
+            PortRange.StartPort = htons(Config->LocalAddress->Ipv4.sin_port);
```
maolson-msft, line 1800: Is Config->LocalAddress->Ipv4.sin_port already in network byte order?

maolson-msft: Also, you're passing Socket->LocalAddress (in which certainly the port is already in network order) to bind below- what's the difference between that and Config->LocalAddress?

nibanks: Yes.

nibanks: `Config->LocalAddress` is an optional (may be null) input address. If specifies, it's practically the same as `Socket->LocalAddress`.

maolson-msft: By "Yes" you mean "it's already in network byte order"? In that case, htons shouldn't be used and the reservation is being acquired for the wrong port.

nibanks: If I don't reverse it, then bind fails with `WSAEACCES`.

thhous-msft: I think thats why the 2nd binds are not failing. Based on how I read the code and the sample, you'd have to pass SIO_ASSOCIATE_PORT_RESERVATION to the first socket too.

nibanks: Still repro's if I remove the `i == 0` condition.

nibanks: It's the very first bind that fails:
```
3379 [0]7380.482C::2021/09/21-16:53:18.352520500 [Microsoft-Quic][data][0x26FDEEC1450] Created, local=0.0.0.0:9999, remote=0x
3380 [0]7380.482C::2021/09/21-16:53:18.352534200 [Microsoft-Windows-Winsock-AFD]socket: 0: Process 0xFFFFB00B68CD50C0 (0x7380), Endpoint 0xFFFFB00B87B24090, Family 23, Type 2(SOCK_DGRAM), Protocol 17, Seq 1006, Status 0x0(NT=STATUS_SUCCESS)
3383 [0]7380.482C::2021/09/21-16:53:18.352549600 [Microsoft-Windows-Winsock-AFD]socket: 1: Process 0xFFFFB00B68CD50C0 (0x7380), Endpoint 0xFFFFB00B87B24090, Family 0, Type 0(??), Protocol 0, Seq 1013, Status 0x0(NT=STATUS_SUCCESS)
3397 [0]7380.482C::2021/09/21-16:53:18.352619000 [Microsoft-Windows-Winsock-AFD]bind: 0: Process 0xFFFFB00B68CD50C0, Endpoint 0xFFFFB00B87B24090, Address [::ffff:0:0]:9999, Seq 7010, Status 0x0(NT=STATUS_SUCCESS)
3400 [0]7380.482C::2021/09/21-16:53:18.352623000 [Microsoft-Windows-Winsock-AFD]bind: 1: Process 0xFFFFB00B68CD50C0, Endpoint 0xFFFFB00B87B24090, Seq 7025, Status 0xC0000022(NT=Access Denied)
3402 [0]7380.482C::2021/09/21-16:53:18.352628200 [Microsoft-Quic][data][0x26FDEEC1450] ERROR, 10013, bind.
```

thhous-msft: That wouldn't change anything. On the first socket you need to call SIO_ACQUIRE_PORT_RESERVATION to get the reservation, and then on all sockets including 0 you need to pass the token with SIO_ASSOCIATE_PORT_RESERVATION.

thhous-msft: Yeah double checking the code SIO_ACQUIRE_PORT_RESERVATION does not assign the token to the socket, so you definitely need to assign the token to each socket, including the first one.

src/generated/linux/datapath_winuser.c.clog.h
```diff
@@ -1203,6 +1203,30 @@ tracepoint(CLOG_DATAPATH_WINUSER_C, DatapathErrorStatus , arg2, arg3, arg4);\
 
 
 
+#ifndef _clog_5_ARGS_TRACE_DatapathErrorStatus
```
maolson-msft, line 1206: I really don't want to know what this is, but FYI it looks suspect because you've added QuicTraceEvent calls with two different strings but there's only one... thing... added here.

thhous-msft: Yeah, the generated code here is REALLY weird. It generates just the first log for each type in the file and just reuses that. Since the last string is passed in as a string and not part of the log.

maolson-msft: OK cool, just so long as it isn't broken!

maolson-msft (approved):

nibanks:
Marking as blocked on a crash this seems to be hitting.

thhous-msft:
This is now fixed and good to go. The QuicInterop failure unrelated.

PR #2029, opened 2021-09-25
Work Around ASAN Deadlock in SpinQuic

### LINKED ISSUE #1933: ASAN Builds on Windows occasionally deadlock
It looks like theres currently a bug in ASAN that causes a deadlock. This manifests as a watchdog timeout in spinquic.

nibanks:
Hopefully fixes #1933 (or at least works around it).

### DIFF
--- a/scripts/spin.ps1
+++ b/scripts/spin.ps1
@@ -124,8 +124,6 @@ if ($CodeCoverage) {
 $SpinQuic = $null
 if ($IsWindows) {
     $SpinQuic = Join-Path $RootArtifactDir "spinquic.exe"
-    # Set ASAN option to prevent deadlock bug in ASAN RtlReAllocHeap.
-    $env:ASAN_OPTIONS='windows_hook_legacy_allocators=1'
 } elseif ($IsLinux -or $IsMacOS) {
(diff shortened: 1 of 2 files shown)


scripts/spin.ps1
```diff
@@ -124,6 +124,8 @@ if ($CodeCoverage) {
 $SpinQuic = $null
 if ($IsWindows) {
     $SpinQuic = Join-Path $RootArtifactDir "spinquic.exe"
+    # Set ASAN option to prevent deadlock bug in ASAN RtlReAllocHeap.
+    $env:ASAN_OPTIONS='windows_hook_legacy_allocators=1'
```
thhous-msft, line 128: This won't actually set the env for the process that gets started. In essence this does nothing.

thhous-msft: What you instead need to do is in run-executable in the Start-Executable function after you create $pinfo is call `$pinfo.Environment.Add(key, value)`

nibanks:
Thanks @thhouse-msft for fixing this! If you're good to go, let's approve/merge it!

nibanks:
Work around causing even more issues. :(

PR #2032, opened 2021-09-27
Add error code and docs for invalid address error

thhous-msft:
If an IPv4 local address and an IPv6 remote address is requested, or vise versa, this is the error that is returned. Add an explicit error code and doc to address this case.


nibanks:
Discussed offline. I'd like to see a very simple negative test case that validates these new error codes. Perhaps it can leverage an existing Validate* API test.

src/test/bin/winkernel/control.cpp
```diff
@@ -1110,6 +1111,9 @@ QuicTestCtlEvtIoDeviceControl(
         QuicTestCtlRun(QuicTestInterfaceBinding(Params->Family));
         break;
 
+    case IOCTL_QUIC_RUN_CONNECT_INVALID_ADDRESS:
+        QuicTestCtlRun(QuicTestConnectInvalidAddress());
```
nibanks, line 1115: Missing a `break`.
```suggestion
        QuicTestCtlRun(QuicTestConnectInvalidAddress());
        break;
```

nibanks (approved):
New tests seem to be passing now. LGTM.

PR #2045, opened 2021-10-05
RPS Events

nibanks:
Adds a number of new ETW events and corresponding tooling to breakdown the lifetime of a stream into different states so that we can get a better idea of the latency costs in a request/response style exchange.


src/core/stream.c
```diff
@@ -37,6 +37,12 @@ QuicStreamInitialize(
         Status = QUIC_STATUS_OUT_OF_MEMORY;
         goto Exit;
     }
+
+    QuicTraceEvent(
```
nibanks, line 41: If we're willing to break the ETW "contract" then we can move StreamOpen event here, and make a new StreamStarted event where StreamOpen is now.

src/plugins/QuicTrace.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29102.190
+# Visual Studio Version 17
+VisualStudioVersion = 17.1.31911.260
```
nibanks, line 4: I believe this automatically happened when I opened in VS 2022 (Preview). Should we check this in?

src/core/stream.c
thhous-msft: I don't think we want to break that. We have actively shipped code using the existing ETW manifest.

src/plugins/QuicTrace.sln
thhous-msft: I think its fine. I imagine most users will be moving to 2022. Its difficult to get 2019 anymore (They hide the downloads).

PR #2071, opened 2021-10-12
Implement BBR congestion control algorithm

Wizmann:
As this PR has many changes, let us set it to draft for early stage review.

Some test results (so far):

```bash
$ sudo tc qdisc add dev lo root handle 1:0 netem delay 100msec

$ ./secnetperf -TestName:Throughput -target:127.0.0.1 -port:2222 -timed:1 -download:60000 -cc:bbr  
Started!

Result: 4159858388 bytes @ 550891 kbps (60409.099 ms).
App Main returning status 0

$ ./secnetperf -TestName:Throughput -target:127.0.0.1 -port:2222 -timed:1 -download:60000 -cc:cubic
Started!

Result: 2336234876 bytes @ 310453 kbps (60201.840 ms).
App Main returning status 0
```

```bash
$ sudo tc qdisc add dev lo root handle 1:0 netem delay 100msec loss 0.01%

$ ./secnetperf -TestName:Throughput -target:127.0.0.1 -port:2222 -timed:1 -download:60000 -cc:bbr  
Started!

Result: 3694289092 bytes @ 490921 kbps (60201.728 ms).
App Main returning status 0

$ ./secnetperf -TestName:Throughput -target:127.0.0.1 -port:2222 -timed:1 -download:60000 -cc:cubic
Started!

Result: 1988873971 bytes @ 263347 kbps (60418.212 ms).
App Main returning status 0
```
```bash
$ sudo tc qdisc add dev lo root handle 1:0 netem delay 100msec loss 0.1%

$ ./secnetperf -TestName:Throughput -target:127.0.0.1 -port:2222 -timed:1 -download:60000 -cc:bbr  
Started!

Result: 3218159327 bytes @ 427651 kbps (60201.516 ms).
App Main returning status 0

$ ./secnetperf -TestName:Throughput -target:127.0.0.1 -port:2222 -timed:1 -download:60000 -cc:cubic
Started!

Result: 23285989 bytes @ 3094 kbps (60202.375 ms).
App Main returning status 0
```

```bash
$ sudo tc qdisc add dev lo root handle 1:0 netem delay 100msec loss 1%

$ ./secnetperf -TestName:Throughput -target:127.0.0.1 -port:2222 -timed:1 -download:60000 -cc:bbr  
Started!

Result: 2361863844 bytes @ 313860 kbps (60201.632 ms).
App Main returning status 0

$ ./secnetperf -TestName:Throughput -target:127.0.0.1 -port:2222 -timed:1 -download:60000 -cc:cubic
Started!

Result: 6122700 bytes @ 813 kbps (60210.360 ms).
App Main returning status 0

```


nibanks:
We need to add this to the WAN perf so that we can compare this to Cubic.

Wizmann:
[Test for 200Mbps Bandwidth, 5ms/200ms RTT, 0.25/1 BottleneckBuffer Ratio, No Loss][1]
[Test for 5Mbps Bandwidth, 5ms/200ms RTT, 0.25/1 BottleneckBuffer Ratio, No Loss][2]

[1]: https://gist.github.com/Wizmann/549dc9139512fb7728de4d86078d41f7
[2]: https://gist.github.com/Wizmann/3c144fa1751de1bf79c283386355f3f4

Wizmann:
much better this time

src/core/loss_detection.h
```diff
@@ -38,7 +38,21 @@ typedef struct QUIC_LOSS_DETECTION {
     QUIC_SENT_PACKET_METADATA* SentPackets;
     QUIC_SENT_PACKET_METADATA** SentPacketsTail;
 
-    uint32_t TimeOfLastPacketSent;
+    uint64_t TimeOfLastPacketSent;
```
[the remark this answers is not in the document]
nibanks: Yes, MsQuic may be a bit more complicated than this, but we try to squeeze every byte out of things that we can. We only represent a timestamp with a type as large as necessary. When tracking an outstanding packet, a `uint32_t` is plenty big enough.

nibanks:
@Wizmann I refactored WAN perf testing to a GitHub Action and will be removing the tests in the Azure Pipeline yml soon. Would you mind updating to the latest main, and moving your change (as necessary) to the ymls over to the GitHub Action?

Wizmann:
Sure, I'll do the update once I have time.

.azure/azure-pipelines.ci.yml
```diff
@@ -473,27 +473,28 @@ stages:
   - build_linux_release
   - build_winkernel_release
   jobs:
-  - template: ./templates/run-performance.yml
```
nibanks, line 476: Should not be necessary.

.azure/azure-pipelines.ci.yml
```diff
@@ -509,46 +510,121 @@ stages:
   - template: ./templates/run-wanperf.yml
     parameters:
       image: windows-2022
-      jobName: wanperf_5Mbps
```
nibanks, line 512: Please update `.\.github\workflows\wan-perf.yml` **instead** of this here. This is being deprecated.

.azure/azure-pipelines.ci.yml
```diff
@@ -509,46 +510,121 @@ stages:
...
+      displayName: 100 Mbps (1% Loss)
+      iterations: 3
+      durationMs: 10000
+      rateMbps: 100
+      bottleneckQueueRatio: (0.25,1.0,5.0)
+      randomLossDenominator: 100
       rttMs: (5,50,200)
       logProfile: Datapath.Light
 
-- ${{ if not(eq(variables['Build.Reason'], 'Schedule')) }}:
```
nibanks, line 543: Not necessary.

Wizmann: Changes for this file have been reverted.

Wizmann:
Seems like we missed one column "CongestionControl" in [merged results][1].

But I've added it to the result class. Please help.

[1]:https://github.com/microsoft/msquic/runs/4251456904?check_suite_focus=true#step:4:14

nibanks:
> Seems like we missed one column "CongestionControl" in [merged results](https://github.com/microsoft/msquic/runs/4251456904?check_suite_focus=true#step:4:14).
> 
> But I've added it to the result class. Please help.

You'll need to update the `FormattedResult` class in emulated-performance.ps1.

Wizmann:
> You'll need to update the `FormattedResult` class in emulated-performance.ps1.

I think it's been updated [link][1]

[1]: https://github.com/microsoft/msquic/blob/6bcd2109e23b713b0693cf07656e5eccbf0f9d88/scripts/emulated-performance.ps1#L133

nibanks:
Weird. Perhaps there is a limit on the number of columns to display by default?

nibanks:
@wizmann what's the latest on this? Would you mind merging main?

Wizmann:
> @Wizmann what's the latest on this? Would you mind merging main?

We're recently working on the integration and hopefully I can start this job later this week.

.github/workflows/wan-perf.yml
```diff
@@ -82,6 +82,7 @@ jobs:
         rtt: [5, 50, 200, 500]
         queueRatio: [0.2, 1, 5]
         loss: [0,1000,10000]
+        congestionControl: ["cubic", "bbr"]
```
nibanks, line 85: Need to pass it to `emulated-performance.ps1` below.

src/manifest/clog.sidecar
```diff
@@ -7215,6 +7215,8 @@
           "DefinationEncoding": "hhu",
           "MacroVariableName": "arg9"
         }
+      ],
+        }
```
nibanks, line 7219: ?

src/inc/msquic.h
```diff
@@ -612,6 +613,9 @@ typedef struct QUIC_SETTINGS {
     uint8_t RESERVED                        : 2;
     uint8_t MaxOperationsPerDrain;
     uint8_t MtuDiscoveryMissingProbeCount;
+    uint16_t MaxBindingStatelessOperations;
+    uint16_t StatelessOperationExpirationMs;
+    uint8_t CongestionControlAlgorithm;
```
nibanks, line 618: Should just be the new CC, and you also need it in the IsSet.

src/manifest/clog.sidecar
Wizmann: Maybe a rebase error, will be removed.

src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -86,6 +86,8 @@ PrintHelp(
         "  -bind:<addr>                A local IP address to bind to.\n"
         "  -cibir:<hex_bytes>          A CIBIR well-known idenfitier.\n"
         "\n"
+        "  -cc:<algo>                  Congestion control algorithm to use.\n"
+        "\n"
```
nibanks, line 90:
```suggestion
        "  -cibir:<hex_bytes>          A CIBIR well-known idenfitier.\n"
        "  -cc:<algo>                  Congestion control algorithm to use.\n"
        "\n"
```

.github/workflows/wan-perf.yml
```diff
@@ -73,6 +73,7 @@ jobs:
...
```
nibanks, line 76: GH Actions doesn't like this, it produces too many unique jobs. Please put this in the `env` instead.

Wizmann:
Seems like all BBR WAN perf test has no result? Crashed?
Is there any info to help debug? It works well on my side.
Thanks.

scripts/emulated-performance.ps1
```diff
@@ -382,7 +394,7 @@ if ($MergeDataFiles) {
                 }
             }
 
-            $Run = [FormattedResult]::new($_.RttMs, $_.BottleneckMbps, $_.BottleneckBufferPackets, $_.RandomLossDenominator, $_.RandomReorderDenominator, $_.ReorderDelayDeltaMs, $_.Tcp, $_.DurationMs, $_.Pacing, $_.RateKbps, $RemoteRate);
+            $Run = [FormattedResult]::new($_.RttMs, $_.BottleneckMbps, $_.BottleneckBufferPackets, $_.RandomLossDenominator, $_.RandomReorderDenominator, $_.ReorderDelayDeltaMs, $_.Tcp, $_.DurationMs, $_.Pacing, $_.RateKbps, $RemoteRate, $_.CongestionControl);
```
nibanks, line 397: Update the `Sort-Object` when dumping "All tests" below (line ~433) to add `CongestionControl` as the last thing to sort by.

src/inc/msquic.h
```diff
@@ -362,6 +362,7 @@ typedef enum QUIC_CIPHER_SUITE {
 
 typedef enum QUIC_CONGESTION_CONTROL_ALGORITHM {
     QUIC_CONGESTION_CONTROL_ALGORITHM_CUBIC,
+    QUIC_CONGESTION_CONTROL_ALGORITHM_BBR,
```
nibanks, line 365: Wrap with `#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES`
```suggestion
#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
    QUIC_CONGESTION_CONTROL_ALGORITHM_BBR,
#endif
```

src/core/congestion_control.h
```diff
@@ -25,8 +44,10 @@ typedef struct QUIC_LOSS_EVENT {
 
     uint64_t LargestPacketNumberSent;
 
-    uint32_t NumRetransmittableBytes;
+    uint32_t NumAckedRetransmittableBytes;
```
maolson-msft, line 47: Why do we need to add a "number of ACKed bytes" field to QUIC_LOSS_EVENT? It looks like BbrCongestionControlOnDataLost is adding this to the current BytesInFlight as if it expects BytesInFlight to have already been decremented by the QUIC_ACK_EVENT, but the loss detection module already passes the loss event before the ack event to the CC- see this code comment in QuicLossDetectionProcessAckBlocks:

    if (NewLargestAck) {
        //
        // Handle packet loss (and any possible congestion events) before
        // data acknowledgement so that we have an accurate bytes in flight
        // calculation for congestion events.
        //
        QuicLossDetectionDetectAndHandleLostPackets(LossDetection, AckedRetransmittableBytes, (uint32_t)TimeNow);
    }

src/core/bbr.c
```diff
@@ -0,0 +1,1327 @@
...
+    uint64_t TargetCwnd = BbrCongestionControlCalculateTargetCwnd(Cc, Bbr->CwndGain);
+    if (Bbr->BtlbwFound) {
+        TargetCwnd += WindowedFilterGetBest(&Bbr->MaxAckHeightFilter);
+    }
+    
+    uint32_t CongestionWindow = Bbr->CongestionWindow;
+
+    if (Bbr->BtlbwFound) {
+        CongestionWindow = (uint32_t)CXPLAT_MIN(TargetCwnd, CongestionWindow + AckedBytes);
+    } else if (CongestionWindow < TargetCwnd || TotalBytesAcked < Bbr->InitialCongestionWindow) {
```
maolson-msft, line 986: It seems like this line here is the only reason the TotalBytesAcked field was added to QUIC_LOSS_DETECTION. Can't we just have BbrCongestionControlCalculateTargetCwnd always return a window >= InitialCongestionWindow? That would make this extra "TotalBytesAcked < Bbr->InitialCongestionWindow" logic unnecessary.

Wizmann: I think it's different, see [Core cwnd Adjustment Mechanism](https://tools.ietf.org/id/draft-cardwell-iccrg-bbr-congestion-control-00.html#core-cwnd-adjustment-mechanism) part in the BBR doc.

I guess your focus is to remove the extra `TotalBytesAcked` in `QUIC_LOSS_DETECTION`. If it's true, we can move that field  (other fields as well) to BBR to reduce the overhead for other CC algos.

src/core/congestion_control.h
Wizmann: `AckedRetransmittableBytes` is used to adjust the recovery window when BBR enters the fast recovery mode.

See `BBRModulateCwndForRecovery` function in [Modulating cwnd in Loss Recovery](https://tools.ietf.org/id/draft-cardwell-iccrg-bbr-congestion-control-00.html#modulating-cwnd-in-loss-recovery) part in the BBR doc.

maolson-msft: I still don't understand. What variable in the spec you linked to does QUIC_CONGESTION_CONTROL_BBR::RecoveryWindow correspond to?

src/core/loss_detection.c
```diff
@@ -456,6 +463,32 @@ QuicLossDetectionOnPacketSent(
             &Connection->CongestionControl, SentPacket->PacketLength);
     }
 
+    SentPacket->Flags.IsAppLimited = QuicCongestionControlIsAppLimited(&Connection->CongestionControl);
```
maolson-msft, line 466: I see you set the IsAppLimited flag here before potentially calling QuicCongestionControlSetAppLimited. Should this line instead come afterward?

It's very hard for me to answer this question for myself because there's no comment in QUIC_SEND_PACKET_FLAGS describing what the IsAppLimited flag means exactly. In general this PR needs to add comments to the new struct fields describing what they are for.

src/core/loss_detection.c
```diff
@@ -456,6 +463,32 @@ QuicLossDetectionOnPacketSent(
...
+
+    uint64_t SendPostedBytes = Connection->SendBuffer.PostedBytes;
+    BOOLEAN LossBufferEmpty = !LossDetection->LostPackets &&
```
maolson-msft, line 469: It seems like you are checking for "!LossDetection->LostPackets" to satisfy condition 4 in "Tracking application-limited phases" here: https://datatracker.ietf.org/doc/html/draft-cheng-iccrg-delivery-rate-estimation-02#section-2.3

"4.  All the packets considered lost have been retransmitted."

But this queue can still have packets in it after we've already retransmitted them. See QuicLossDetectionDetectAndHandleLostPackets, where we move packets to the LostPackets list after calling QuicLossDetectionRetransmitFrames to retransmit them.

src/core/congestion_control.h
Wizmann: It's the `cwnd` in function `BBRModulateCwndForRecovery`.

The `RecoveryWindow` is the window size in "Fast Recovery" mode, and after we exit that mode, we'll continue use `Bbr->CongestionWindow`, which is what `BBRRestoreCwnd()` do in the spec.

src/core/loss_detection.c
Wizmann: Sure, will do.

src/core/loss_detection.c
Wizmann: Agreed, it is a mistake. Any suggestion for the fix?

maolson-msft: I think we can check if QuicStreamCanSendNow returns TRUE for any stream. If it does, then we aren't application-limited. In addition to checking for unretransmitted lost packets, QuicStreamCanSendNow also includes a check for flow-control-limited situations (where we are limited by the peer application rather than the local application). I don't see flow-control-limited situations included in the rate estimation IETF draft, but I think it's appropriate to include them. I think this calls for followup with the BBR designers if it hasn't already been discussed.

Wizmann: I just made a fix for the issue, but I don't have 100% confidence for that.
Plz have a double check.

nibanks:
```
  /home/runner/work/msquic/msquic/src/core/loss_detection.c: In function ‘QuicLossDetectionOnPacketSent’:
  /home/runner/work/msquic/msquic/src/core/loss_detection.c:476:67: error: expected ‘,’ or ‘;’ before ‘)’ token
    476 |                               !QuicStreamCanSendNow(Stream, FALSE)) &&
        |                                                                   ^
  make[2]: *** [src/core/CMakeFiles/core.dir/build.make:286: src/core/CMakeFiles/core.dir/loss_detection.c.o] Error 1
  make[1]: *** [CMakeFiles/Makefile2:214: src/core/CMakeFiles/core.dir/all] Error 2
```

src/core/cubic.c
```diff
@@ -573,8 +573,8 @@ CubicCongestionControlOnDataLost(
             LossEvent->PersistentCongestion);
     }
 
-    CXPLAT_DBG_ASSERT(Cubic->BytesInFlight >= LossEvent->NumRetransmittableBytes);
-    Cubic->BytesInFlight -= LossEvent->NumRetransmittableBytes;
+    CXPLAT_DBG_ASSERT(Cubic->BytesInFlight >= LossEvent->NumLostRetransmittableBytes);
+    Cubic->BytesInFlight -= LossEvent->NumLostRetransmittableBytes;
```
nibanks, line 577: Is this correct? You changed the `LossEvent` struct by replacing `NumRetransmittableBytes` with `NumAckedRetransmittableBytes` **and** `NumLostRetransmittableBytes`. Does that mean they add up to the original? If so, should this be using the sum of those two?

src/core/loss_detection.c
```diff
@@ -456,6 +463,39 @@ QuicLossDetectionOnPacketSent(
...
     }
 
+    SentPacket->Flags.IsAppLimited = QuicCongestionControlIsAppLimited(&Connection->CongestionControl);
+
+    uint64_t SendPostedBytes = Connection->SendBuffer.PostedBytes;
+
+    CXPLAT_LIST_ENTRY* Entry = Connection->Send.SendStreams.Flink;
+    QUIC_STREAM* Stream = (Entry != &(Connection->Send.SendStreams)) ?
+                          CXPLAT_CONTAINING_RECORD(Entry, QUIC_STREAM, SendLink) :
+                          NULL;
```
nibanks, line 473: Generally, our style for these type of multi-line expressions would be like this instead:
```suggestion
    QUIC_STREAM* Stream =
        (Entry != &(Connection->Send.SendStreams)) ?
          CXPLAT_CONTAINING_RECORD(Entry, QUIC_STREAM, SendLink) :
          NULL;
```
Same for all the multi-line expressions you've added. Thanks!

src/core/sent_packet_metadata.h
```diff
@@ -103,10 +119,17 @@ typedef struct QUIC_SENT_PACKET_METADATA {
...
+
     uint32_t SentTime; // In microseconds
+
     uint16_t PacketLength;
+
     uint8_t PathId;
 
+    uint64_t TotalBytesSentThen;
+
+    LAST_ACKED_PACKET_INFO LastAckedPacketInfo;
```
nibanks, line 131: This is a significant increase in the amount of memory required to track every single sent packet. Like 32 to 40 extra bytes multiplied by every outstanding packet. Is this absolutely necessary? It seems to really be a shame that we'd have to include this when cubic doesn't even use it.

BTW, at the very least, `TotalBytesSentThen` (which I don't even understand based on the variable name) needs to be put after `PacketNumber` for the best padding/packing. `LastAckedPacketInfo` should also be at the top if we keep it.

src/core/loss_detection.h
```diff
@@ -40,6 +40,18 @@ typedef struct QUIC_LOSS_DETECTION {
...
+
+    uint32_t LastAckedPacketSentTime;
+
+    uint32_t AdjustedLastAckedTime;
+
+    uint64_t TotalBytesSent;
+
+    uint64_t TotalBytesAcked;
+
+    uint64_t TotalBytesSentAtLastAck;
```
nibanks, line 53: I think these would be better placed right after `LargestAckEncryptLevel` above. Also, please add comments to these.

nibanks:
Windows kernel build is failing:
```
##[error]src\perf\lib\SecNetPerfMain.cpp(293,72): Error C2065: 'Cc': undeclared identifier
D:\a\1\msquic\src\perf\lib\SecNetPerfMain.cpp(293,72): error C2065: 'Cc': undeclared identifier [D:\a\1\msquic\src\perf\lib\perflib.kernel.vcxproj]
##[error]src\perf\lib\SecNetPerfMain.cpp(296,60): Error C2065: 'Cc': undeclared identifier
D:\a\1\msquic\src\perf\lib\SecNetPerfMain.cpp(296,60): error C2065: 'Cc': undeclared identifier [D:\a\1\msquic\src\perf\lib\perflib.kernel.vcxproj]
```

src/core/cubic.c
Wizmann: > Does that mean they add up to the original?
No. 
`NumLostRetransmittableBytes` is the number of lost bytes and `NumAckedRetransmittableBytes` is number of acked bytes. You can find it in function `QuicLossDetectionDetectAndHandleLostPackets` from `loss_detection.c`.
In Cubic, we only need to know the number of lost bytes. But we need to use the acked bytes of updating the recovery window in BBR. That's why for the change.

Wizmann:
The "division-by-zero" error is a potential bug, and it'll be fixed.

But I'm a little confused with these "notes" from `codecheck`:

```
/home/vsts/work/1/s/src/core/bbr.c:306:17: note: Calling 'CxPlatTimeDiff32'
                CxPlatTimeDiff32(AckedPacket->LastAckedPacketInfo.SentTime, AckedPacket->SentTime));
                ^
/home/vsts/work/1/s/src/inc/quic_platform_posix.h:642:9: note: Assuming 'T2' is > 'T1'
    if (T2 > T1) {
        ^
/home/vsts/work/1/s/src/inc/quic_platform_posix.h:642:5: note: Taking true branch
    if (T2 > T1) {
    ^
/home/vsts/work/1/s/src/inc/quic_platform_posix.h:643:9: note: Returning value
        return T2 - T1;
```

Do I need to fix all these issues? Or can I do something to stop the complaining?

scripts/emulated-performance.ps1
```diff
@@ -52,6 +52,9 @@ be in the current directory.
 .PARAMETER NoDateLogDir
     Doesn't include the Date/Time in the log directory path.
 
+.PARAMETER CongestionControl
+    The congestion control algoirthm used to test
```
maolson-msft, line 56:
```suggestion
    The congestion control algorithm used to test
```

src/core/congestion_control.h
```diff
@@ -13,10 +14,28 @@ typedef struct QUIC_ACK_EVENT {
...
+
     uint32_t NumRetransmittableBytes;
 
     uint32_t SmoothedRtt;
 
+    uint32_t MinRttSample;
+
+    uint32_t AdjustedAckTime;
+
+    uint64_t TotalBytesAcked;
```
maolson-msft, line 27: You need to comment this to make it clear what the difference is between this and the NumRetransmittableBytes field. And probably the two names should be more similar.

// Number of retransmittable bytes acked during the connection's lifetime

src/core/loss_detection.c
```diff
@@ -1149,10 +1206,19 @@ QuicLossDetectionDiscardPackets(
...
             .NumRetransmittableBytes = AckedRetransmittableBytes,
-            .SmoothedRtt = Path->SmoothedRtt
+            .SmoothedRtt = Path->SmoothedRtt,
+            .MinRttSample = 0,
+            .HasLoss = FALSE,
+            .AdjustedAckTime = 0,
+            .AckedPackets = NULL,
+            .TotalBytesAcked = 0,
+            .IsLargestAckedPacketAppLimited = FALSE,
+            .MinRttSampleValid = 0
```
maolson-msft, line 1221:
```suggestion
            .MinRttSampleValid = FALSE
```

src/core/sent_packet_metadata.h
```diff
@@ -88,12 +88,47 @@ typedef struct QUIC_SEND_PACKET_FLAGS {
...
 
 } QUIC_SEND_PACKET_FLAGS;
 
+//
+// Packet info of last acked packet
+//
+typedef struct LAST_ACKED_PACKET_INFO {
+
+    //
+    // Total bytes sent
```
maolson-msft, line 105: If the field is called "Total bytes sent", then adding a code comment that says "total bytes sent" adds zero value and is annoying to read.

I think the only fields in this struct that need comments are AdjustedAckTime and AckTime.

maolson-msft: Actually, we also need a comment explaining the difference between the TotalBytesSent field here and the one in QUIC_SENT_PACKET_METADATA.

src/core/congestion_control.h
maolson-msft: No, QUIC_CONGESTION_CONTROL_BBR::CongestionWindow is what corresponds to the "cwnd" variable in the spec. Not  QUIC_CONGESTION_CONTROL_BBR::RecoveryWindow.

I still think you should not be adding NumAckedRetransmittableBytes to BytesInFlight when you set RecoveryWindow. The logic in the spec's BBRModulateCwndForRecovery function assumes that packets_in_flight has already been decremented by packets_delivered during ACK processing (notice that BBRSetCwnd is all the way at the end of the set of operations in section 3.5.2, "per-ack steps"), but that isn't the case in the QUIC code.

src/core/bbr.h
```diff
@@ -0,0 +1,159 @@
...
+
+} BBR_BANDWIDTH_SAMPLER;
+
+typedef struct QUIC_CONGESTION_CONTROL_BBR {
+
+    BOOLEAN BtlbwFound : 1;
+
+    BOOLEAN AppLimitedSinceProbeRtt : 1;
+
+    BOOLEAN ExitingQuiescene : 1;
```
maolson-msft, line 88: Fix the spelling on this field.

src/core/bbr.c
```diff
@@ -0,0 +1,1327 @@
...
+    )
+{
+    QUIC_CONGESTION_CONTROL_BBR* Bbr = &Cc->Bbr;
+
+    if (Bbr->BtlbwFound) {
+        return;
+    }
+
+    if (IsAppLimited) {
+        return;
```
maolson-msft, line 881: Just don't call BbrCongestionControlDetectBottleneckBandwidth from BbrCongestionControlOnDataAcknowledged if LastAckedPacketAppLimited is TRUE.

In fact, please just move the code from this "helper" function into BbrCongestionControlOnDataAcknowledged.

src/core/bbr.c
```diff
@@ -0,0 +1,1327 @@
...
+    if (ShouldAdvancePacingGainCycle) {
+        Bbr->PacingCycleIndex = (Bbr->PacingCycleIndex + 1) % kNumOfCycles;
+        Bbr->CycleStart = AckTime;
+        Bbr->PacingGain = kPacingGainCycles[Bbr->PacingCycleIndex];
+    }
+}
+
+_IRQL_requires_max_(DISPATCH_LEVEL)
+BOOLEAN
+BbrCongestionControlDetectShouldExitStartup(
```
maolson-msft, line 860: Move the single line of this "helper" function to the single callsite in BbrCongestionControlOnDataAcknowledged.

src/core/loss_detection.c
maolson-msft: You're now following my advice to call QuicStreamCanSendNow, but QuicStreamCanSendNow includes a check for FC-blocked and per the discussion on the google group, FC-blocked should not be included in the set of criteria for calling the connection "app-limited". So this needs to be fixed to exclude that check.

maolson-msft: It's kind of ugly, but you could just add a check of QuicStreamAllowedByPeer(Stream) when setting the LossBufferEmpty variable.

Also, "LossBufferEmpty" is a very confusing name. You don't need a variable at all here- just add the logic to the set of checks in the "if" statement for calling QuicCongestionControlSetAppLimited.

src/core/sent_packet_metadata.h
```diff
@@ -88,12 +88,45 @@ typedef struct QUIC_SEND_PACKET_FLAGS {
...
+// 
+typedef struct LAST_ACKED_PACKET_INFO {
+
+    //
+    // Total bytes sent when the last acked packet was acked
+    //
+    uint64_t TotalBytesSent;
+
+    //
+    // Total bytes sent when the last acked packet was acked
```
maolson-msft, line 110:
```suggestion
    // Total bytes acked when the last acked packet was acked
```
```suggestion
    // Total bytes sent when the last acked packet was acked
```

src/core/sent_packet_metadata.h
```diff
@@ -102,10 +118,17 @@ typedef struct QUIC_SENT_PACKET_METADATA {
...
+
+    uint64_t SentTime; // In microseconds
+
     uint16_t PacketLength;
+
     uint8_t PathId;
 
+    uint64_t TotalBytesSentThen;
+
+    LAST_ACKED_PACKET_INFO LastAckedPacketInfo;
```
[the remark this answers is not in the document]
Wizmann: ACK.
Leave a open question here for a proper solution.

.azure/azure-pipelines.ci.yml
Wizmann: Azure pipeline has been deprecated.

src/inc/msquic.h
Wizmann: This has been checkin in another PR.

src/core/sent_packet_metadata.h
Wizmann: fixed.

src/core/loss_detection.c
Wizmann: Tried to fix in the latest commit. Need double check from @maolson-msft

src/core/congestion_control.h
Wizmann: > QUIC_CONGESTION_CONTROL_BBR::CongestionWindow is what corresponds to the "cwnd" variable in the spec. Not QUIC_CONGESTION_CONTROL_BBR::RecoveryWindow.

Let say the "good cwnd" is "the cwnd unmodulated by loss recovery or ProbeRTT", which is `QUIC_CONGESTION_CONTROL_BBR::CongestionWindow` in our code.

As it described in the spec, the "good cwnd" will be saved to `BBR.prior_cwnd` before switch to recovery mode. And `BBR.cwnd` will be assigned the size of "RecoveryWindow".

After BBR exit the recovery mode, it calls `BBRRestoreCwnd()` to restore the previous "good cwnd" to `BBR.cwnd`.

Our implementation is essentially the same.

> I still think you should not be adding NumAckedRetransmittableBytes to BytesInFlight when you set RecoveryWindow...

Fixed.

.github/workflows/wan-perf.yml
Wizmann: moved to env node

.azure/azure-pipelines.ci.yml
```diff
@@ -488,6 +488,7 @@ stages:
       durationMs: 10000
       rateMbps: 5
       bottleneckQueueRatio: (0.25,1.0)
+      randomLossDenominator: (0,0.1,1)
```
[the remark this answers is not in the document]
Wizmann: Azure pipeline has been deprecated.

src/core/congestion_control.h
maolson-msft: Now remove the NumAckedRetransmittableBytes field from QUIC_LOSS_EVENT, since it's no longer used.

src/core/loss_detection.c
maolson-msft: You neither added a comment to QUIC_SEND_PACKET_FLAGS nor did you move this IsAppLimited flag assignment after the QuicCongestionControlSetAppLimited call. Why was this thread marked resolved?

src/core/bbr.c
Wizmann: oppps...

src/core/loss_detection.c
Wizmann: my bad, my bad... I thought you were mentioning the comments in `LAST_ACKED_PACKET_INFO`.

I think now both have been fixed.

src/core/bbr.c
Wizmann: deleted now.

src/core/congestion_control.h
Wizmann: fixed

maolson-msft:
@Wizmann @nibanks OK there are still some minor issues in the code but it will be more efficient for me to fix those in a followup PR. So the only blocker at this point is the failed CI tests. Wizmann, if you fix those then we can check in.

nibanks:
I just kicked off a rerun of the failures to see if they are noise or not.

nibanks:
@wizman the following codecheck errors are real: https://dev.azure.com/ms/msquic/_build/results?buildId=326857&view=logs&j=8140d3d3-9a8b-59ca-7afc-aa501326c69d&t=88c7857a-8892-5b08-33d5-fee4cd87cdbd&l=1003. For instance:
```
/home/vsts/work/1/s/src/core/bbr.c:320:69: error: Division by zero [clang-analyzer-core.DivideZero,-warnings-as-errors]
                        AckEvent->NumTotalAckedRetransmittableBytes /
                                                                    ^
/home/vsts/work/1/s/src/core/bbr.c:958:9: note: Assuming field 'IsImplicit' is 0
    if (AckEvent->IsImplicit) {
        ^
/home/vsts/work/1/s/src/core/bbr.c:958:5: note: Taking false branch
    if (AckEvent->IsImplicit) {
    ^
/home/vsts/work/1/s/src/core/bbr.c:967:23: note: Assuming field 'BytesInFlight' is >= field 'NumRetransmittableBytes'
    CXPLAT_DBG_ASSERT(Bbr->BytesInFlight >= AckEvent->NumRetransmittableBytes);
                      ^
/home/vsts/work/1/s/src/inc/quic_platform_posix.h:242:50: note: expanded from macro 'CXPLAT_DBG_ASSERT'
#define CXPLAT_DBG_ASSERT(exp) CXPLAT_FRE_ASSERT(exp)
                                                 ^
/home/vsts/work/1/s/src/inc/quic_platform_posix.h:238:34: note: expanded from macro 'CXPLAT_FRE_ASSERT'
#define CXPLAT_FRE_ASSERT(exp) ((exp) ? (void)0 : (CxPlatLogAssert(__FILE__, __LINE__, #exp), quic_bugcheck(__FILE__, __LINE__, #exp)));
                                 ^
/home/vsts/work/1/s/src/core/bbr.c:967:5: note: '?' condition is true
    CXPLAT_DBG_ASSERT(Bbr->BytesInFlight >= AckEvent->NumRetransmittableBytes);
    ^
/home/vsts/work/1/s/src/inc/quic_platform_posix.h:242:32: note: expanded from macro 'CXPLAT_DBG_ASSERT'
#define CXPLAT_DBG_ASSERT(exp) CXPLAT_FRE_ASSERT(exp)
                               ^
/home/vsts/work/1/s/src/inc/quic_platform_posix.h:238:33: note: expanded from macro 'CXPLAT_FRE_ASSERT'
#define CXPLAT_FRE_ASSERT(exp) ((exp) ? (void)0 : (CxPlatLogAssert(__FILE__, __LINE__, #exp), quic_bugcheck(__FILE__, __LINE__, #exp)));
                                ^
/home/vsts/work/1/s/src/core/bbr.c:970:9: note: Assuming field 'MinRttSampleValid' is 0
    if (AckEvent->MinRttSampleValid) {
        ^
/home/vsts/work/1/s/src/core/bbr.c:970:5: note: Taking false branch
    if (AckEvent->MinRttSampleValid) {
    ^
/home/vsts/work/1/s/src/core/bbr.c:978:9: note: Assuming field 'AckedPackets' is not equal to NULL
        AckEvent->AckedPackets == NULL ? FALSE : AckEvent->IsLargestAckedPacketAppLimited;
        ^
...
```

nibanks:
Windows kernel build errors (need to add the file to the project):
```
core.lib(bbr.obj) : error LNK2001: unresolved external symbol SlidingWindowExtremumInitialize [D:\a\1\msquic\src\bin\winkernel\msquic.kernel.vcxproj]
core.lib(bbr.obj) : error LNK2001: unresolved external symbol SlidingWindowExtremumGet [D:\a\1\msquic\src\bin\winkernel\msquic.kernel.vcxproj]
core.lib(bbr.obj) : error LNK2001: unresolved external symbol SlidingWindowExtremumUpdateMax [D:\a\1\msquic\src\bin\winkernel\msquic.kernel.vcxproj]
core.lib(bbr.obj) : error LNK2001: unresolved external symbol SlidingWindowExtremumReset [D:\a\1\msquic\src\bin\winkernel\msquic.kernel.vcxproj]
```

nibanks:
Linux build error (need to add new file to CLOG/lttng list):
```
/home/vsts/work/1/s/src/core/unittest/SlidingWindowExtremumTest.cpp:14:10: fatal error: TestSlidingWindowExtremum.cpp.clog.h: No such file or directory
   14 | #include "TestSlidingWindowExtremum.cpp.clog.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

src/core/congestion_control.h
```diff
@@ -13,10 +14,34 @@ typedef struct QUIC_ACK_EVENT {
...
+    //
+    uint64_t NumTotalAckedRetransmittableBytes;
+
+    QUIC_SENT_PACKET_METADATA* AckedPackets;
+
     uint32_t NumRetransmittableBytes;
 
     uint32_t SmoothedRtt;
 
+    uint32_t MinRttSample;
```
nibanks, line 30: I believe we already track this somewhere else. Do we really need to duplicate that?

src/core/loss_detection.c
```diff
@@ -453,6 +460,38 @@ QuicLossDetectionOnPacketSent(
             &Connection->CongestionControl, SentPacket->PacketLength);
     }
 
+    uint64_t SendPostedBytes = Connection->SendBuffer.PostedBytes;
```
nibanks, line 463: This assumes we're using send buffering?

src/core/loss_detection.c
```diff
@@ -453,6 +460,38 @@ QuicLossDetectionOnPacketSent(
...
+    CXPLAT_LIST_ENTRY* Entry = Connection->Send.SendStreams.Flink;
+    QUIC_STREAM* Stream =
+        (Entry != &(Connection->Send.SendStreams)) ?
+          CXPLAT_CONTAINING_RECORD(Entry, QUIC_STREAM, SendLink) :
+          NULL;
+    
+    if (SendPostedBytes < Path->Mtu &&
+        QuicCongestionControlCanSend(&Connection->CongestionControl) &&
+        !QuicCryptoHasPendingCryptoFrame(&Connection->Crypto) &&
+        (Stream && QuicStreamAllowedByPeer(Stream)) && !QuicStreamCanSendNow(Stream, FALSE)) {
```
nibanks, line 474: I feel like a comment explaining why this logic means it is app limited. I'm not convinced this is correct.

src/core/loss_detection.h
```diff
@@ -22,6 +22,42 @@ typedef struct QUIC_LOSS_DETECTION {
     //
     QUIC_ENCRYPT_LEVEL LargestAckEncryptLevel;
 
+
```
nibanks, line 25: nit: should remove extra new line

src/core/loss_detection.c
```diff
@@ -1450,6 +1531,13 @@ QuicLossDetectionProcessAckBlocks(
 
     LossDetection->ProbeCount = 0;
 
+    AckedPacketsIterator = AckedPackets;
+    while (AckedPacketsIterator != NULL) {
+        QUIC_SENT_PACKET_METADATA* Packet = AckedPacketsIterator;
+        AckedPacketsIterator = AckedPacketsIterator->Next;
+        QuicSentPacketPoolReturnPacketMetadata(&Connection->Worker->SentPacketPool, Packet);
+    }
```
nibanks, line 1539: Why do we need to add a new clean up path?

src/core/sliding_window_extremum.c
```diff
@@ -0,0 +1,169 @@
+/*++
+
+    Copyright (c) Microsoft Corporation.
+    Licensed under the MIT License.
+
+Abstract:
+    The following functions implement a sliding window extremum (either maxima
```
nibanks, line 7:
```suggestion
Abstract:

    The following functions implement a sliding window extremum (either maxima
```

src/core/sliding_window_extremum.c
```diff
@@ -0,0 +1,169 @@
...
+    If there are more elements comparing to the capacity of queue, the
+    algorithm will still work but the accuracy will be compromised.
+
+--*/
+
+
+#include "precomp.h"
+
+_IRQL_requires_max_(DISPATCH_LEVEL)
+SLIDING_WINDOW_EXTREMUM
```
nibanks, line 25: Every struct we have is prefixed with `QUIC_` to avoid possible name conflicts. Please update.

src/core/sliding_window_extremum.c
```diff
@@ -0,0 +1,169 @@
...
+SlidingWindowExtremumInitialize(
+    _In_ uint64_t EntryLifetime,
+    _In_ uint32_t WindowCapacity,
+    _In_ SLIDING_WINDOW_EXTREMUM_ENTRY* Window
+    )
+{
+    SLIDING_WINDOW_EXTREMUM w;
```
nibanks, line 32: Please, no single letter variables. This looks like it should be called `Window` and the input Window variable should be called `Entry`. Should `SLIDING_WINDOW_EXTREMUM->Window` be called `SLIDING_WINDOW_EXTREMUM->Entry`?

src/core/sliding_window_extremum.h
```diff
@@ -0,0 +1,112 @@
...
+
+#if defined(__cplusplus)
+extern "C" {
+#endif
+
+typedef struct SLIDING_WINDOW_EXTREMUM_ENTRY {
+
+    uint64_t Value;
+
+    uint64_t Time;
```
nibanks, line 18: Units?

nibanks (commented):
I left a number of comments. The CI has passed. @maolson-msft, if you're ok fixing these comments up with an immediate follow up PR, I'd be ok to merge. But I leave it up to you.

src/core/sliding_window_extremum.h
Wizmann: This data structure is not designed for any specific kind of unit. You can use second, ms, us or even tick. In our use case, we use the packet number as "time".

I agree the `uint64_t Time` may not be named aptly. Any suggestions here?

nibanks: Do we use it with different units, or always the same? I agree it could be generic, but I'm not convinced we need to define it so.

src/core/sliding_window_extremum.c
Wizmann: > Please, no single letter variables.                

OK 

> Should SLIDING_WINDOW_EXTREMUM->Window be called SLIDING_WINDOW_EXTREMUM->Entry?

* `SLIDING_WINDOW_EXTREMUM->Entry` 
* `SLIDING_WINDOW_EXTREMUM->Entries`
* `SLIDING_WINDOW_EXTREMUM->Deque`
* `SLIDING_WINDOW_EXTREMUM->Extremums`

They were the candidates for the internal "window" of this data structure...

src/core/sliding_window_extremum.h
Wizmann: For a single instance of this "sliding window", the unit should be always the same.

But for different "windows", we can choose different units.

src/core/congestion_control.h
Wizmann: I'm afraid not.

This is tracking the smallest RTT we gathered from ack blocks from processing 1 ack frame.

And yes, we have tracked the smallest RTT in `QuicConnUpdateRtt`. But it for the global min/max RTT for the path, but not the local minima.

src/core/loss_detection.c
Wizmann: Because we use the acked packet info list for `QuicCongestionControlOnDataAcknowledged` function of BBR.

We have to delay the time for the clean up work until it's get done.

src/core/sliding_window_extremum.c
Wizmann: I've changed the name of internal queue to `SLIDING_WINDOW_EXTREMUM->Extremums`.
Feel free to change if you have better options.

src/core/congestion_control.h
maolson-msft: This field has been removed, so I'm sort of forced to assume you changed your mind and agree that the field wasn't needed. So I guess I'll resolve the comment?

src/core/loss_detection.c
maolson-msft: I'll add a comment later- it's the "rate estimation" spec that is weirdly separate from the main BBR spec. I'll double check the logic too.

maolson-msft (approved):

Wizmann:
image

@janksenhu

janksenhu:
linux platform above cmdline could not run, same params

PR #2106, opened 2021-11-02
Filter for remote address for connected sockets in CxPlatGetSocket

maolson-msft:
Adds a RemoteAddress parameter to CxPlatGetSocket so that only traffic matching
the socket's remote address will be delivered to a connected socket.

### DIFF
--- a/src/platform/datapath_raw_socket.c
+++ b/src/platform/datapath_raw_socket.c
@@ -56,7 +56,7 @@ CxPlatGetSocket(
     while (Entry != NULL) {
         CXPLAT_SOCKET* Temp = CONTAINING_RECORD(Entry, CXPLAT_SOCKET, Entry);
         if (QuicAddrCompareIp(&Temp->LocalAddress, LocalAddress) &&
-            (!Temp->Connected || QuicAddrCompareIp(&Temp->RemoteAddress, RemoteAddress))) {
+            (!Temp->Connected || QuicAddrCompare(&Temp->RemoteAddress, RemoteAddress))) {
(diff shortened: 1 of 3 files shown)


maolson-msft:
Right after sending this PR I noticed that QuicAddrCompareIp is being called rather than QuicAddrCompare. Is that intentional?

nibanks (approved):

nibanks:
> @nibanks Right after sending this PR I noticed that QuicAddrCompareIp is being called rather than QuicAddrCompare. Is that intentional? 

Probably not.

src/platform/datapath_raw_socket.c
```diff
@@ -54,7 +55,8 @@ CxPlatGetSocket(
     Entry = CxPlatHashtableLookup(&Pool->Sockets, LocalAddress->Ipv4.sin_port, &Context);
     while (Entry != NULL) {
         CXPLAT_SOCKET* Temp = CONTAINING_RECORD(Entry, CXPLAT_SOCKET, Entry);
-        if (QuicAddrCompareIp(&Temp->LocalAddress, LocalAddress)) {
+        if (QuicAddrCompareIp(&Temp->LocalAddress, LocalAddress) &&
+            (!Temp->Connected || QuicAddrCompareIp(&Temp->RemoteAddress, RemoteAddress))) {
```
nibanks, line 59:
```suggestion
            (!Temp->Connected || QuicAddrCompare(&Temp->RemoteAddress, RemoteAddress))) {
```
As Matt mentioned, this should compare both IP and port. The other one already has compared port because of the call to `CxPlatHashtableLookup` takes the port.

PR #2107, opened 2021-11-02
Worker Optimizations for Latency

nibanks:
Playing around with micro-optimizations and polling to try to reduce latency of the worker thread.


nibanks:
Also kicked off a PGO perf pipeline: https://dev.azure.com/ms/msquic/_build/results?buildId=238979&view=results. I will update the PGO files with the output from that.

src/core/quicdef.h
```diff
@@ -128,6 +128,12 @@ typedef struct QUIC_PATH QUIC_PATH;
 //
 #define QUIC_DEFAULT_RETRY_MEMORY_FRACTION      65 // ~0.1%
 
+//
+// If enabled, workers will poll the number of times before falling back to the
+// wait or delay state.
+//
+//#define QUIC_WORKER_POLLING                     10000
```
nibanks, line 135: Disabled for now, but shows promise in DPDK/XDP scenarios. I want to get this all in mainline before merging into dpdk branch to play around with this part some more.

PR #2111, opened 2021-11-03
Fix leak in OpenSSL PKCS12 parsing.

anrossi:
When parsing a PKCS12 that includes a certificate chain, OpenSSL wasn't freeing the certificate chain memory, and introduced a leak.
This change fixes that by freeing the certificate chain when done adding the certificates to `SSL_CTX`.
Signed-off-by: Anthony Rossi <anrossi@microsoft.com>


thhous-msft (approved):

nibanks (approved):

wfurt:
tedious but not that hard. I think we should build chain with at least 3 nodes e.g. root, intermediate and leaf.

anrossi:
I have not yet found a way to export certificates to pkcs12 cross-platform using .NET or powershell; this blocks testing this change.  I suspect I will need platform-specific code to invoke `openssl` on linux to generate the pkcs12 file.

PR #2116, opened 2021-11-04
Improve Socket Lookup

nibanks:
Updates the listener to properly handle wildcards. Consolidates the socket address comparison logic. Has been tested.

### DIFF
--- a/src/platform/datapath_raw.c
+++ b/src/platform/datapath_raw.c
@@ -207,8 +207,11 @@ CxPlatSocketCreateUdp(
                 Config->LocalAddress->Ipv4.sin_port;
         }
     } else {
-        CXPLAT_FRE_ASSERT((*NewSocket)->Connected); // Assumes connected socket
-        (*NewSocket)->LocalAddress = Datapath->ClientIP;
+        if ((*NewSocket)->Connected) {
+            (*NewSocket)->LocalAddress = Datapath->ClientIP;
+        } else {
+            (*NewSocket)->Wildcard = TRUE;
+        }
     }
 
     CXPLAT_FRE_ASSERT((*NewSocket)->Wildcard ^ (*NewSocket)->Connected); // Assumes either a pure wildcard listener or a
--- a/src/platform/datapath_raw.c
+++ b/src/platform/datapath_raw.c
@@ -199,7 +199,6 @@ CxPlatSocketCreateUdp(
             } else {
                 (*NewSocket)->Wildcard = TRUE;
             }
-            (*NewSocket)->LocalAddress = Datapath->ServerIP;
         } else {
             (*NewSocket)->LocalAddress = *Config->LocalAddress;
         }
--- a/src/platform/datapath_raw.c
+++ b/src/platform/datapath_raw.c
@@ -188,6 +188,7 @@ CxPlatSocketCreateUdp(
     (*NewSocket)->CallbackContext = Config->CallbackContext;
 
     if (Config->RemoteAddress) {
+        CXPLAT_FRE_ASSERT(!QuicAddrIsWildCard(Config->RemoteAddress));  // No wildcard remote addresses allowed.
         (*NewSocket)->Connected = TRUE;
         (*NewSocket)->RemoteAddress = *Config->RemoteAddress;
     }
--- a/src/platform/datapath_raw.c
+++ b/src/platform/datapath_raw.c
@@ -182,24 +182,38 @@ CxPlatSocketCreateUdp(
         goto Error;
     }
 
+    CxPlatZeroMemory(*NewSocket, sizeof(CXPLAT_SOCKET));
     CxPlatRundownInitialize(&(*NewSocket)->Rundown);
     (*NewSocket)->Datapath = Datapath;
     (*NewSocket)->CallbackContext = Config->CallbackContext;
+
     if (Config->RemoteAddress) {
         (*NewSocket)->Connected = TRUE;
         (*NewSocket)->RemoteAddress = *Config->RemoteAddress;
+    }
+
+    if (Config->LocalAddress) {
+        if (QuicAddrIsWildCard(Config->LocalAddress)) {
+            if ((*NewSocket)->Connected) {
+                (*NewSocket)->LocalAddress = Datapath->ClientIP;
+            } else {
+                (*NewSocket)->Wildcard = TRUE;
+            }
+            (*NewSocket)->LocalAddress = Datapath->ServerIP;
+        } else {
+            (*NewSocket)->LocalAddress = *Config->LocalAddress;
+        }
+        if (Config->LocalAddress->Ipv4.sin_port != 0) {
+            (*NewSocket)->LocalAddress.Ipv4.sin_port =
+                Config->LocalAddress->Ipv4.sin_port;
+        }
+    } else {
+        CXPLAT_FRE_ASSERT((*NewSocket)->Connected); // Assumes connected socket
         (*NewSocket)->LocalAddress = Datapath->ClientIP;
-    } else {
-        (*NewSocket)->Connected = FALSE;
-        CxPlatZeroMemory(&(*NewSocket)->RemoteAddress, sizeof(QUIC_ADDR));
-        (*NewSocket)->LocalAddress = Datapath->ServerIP;
     }
-    if (Config->LocalAddress && Config->LocalAddress->Ipv4.sin_port != 0) {
-        (*NewSocket)->LocalAddress.Ipv4.sin_port =
-            Config->LocalAddress->Ipv4.sin_port;
-    } else {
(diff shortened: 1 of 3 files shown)


src/platform/datapath_raw.c
```diff
@@ -182,25 +182,39 @@ CxPlatSocketCreateUdp(
...
-            Config->LocalAddress->Ipv4.sin_port;
+
+    if (Config->LocalAddress) {
+        if (QuicAddrIsWildCard(Config->LocalAddress)) {
+            if ((*NewSocket)->Connected) {
+                (*NewSocket)->LocalAddress = Datapath->ClientIP;
+            } else {
+                (*NewSocket)->Wildcard = TRUE;
+            }
+            (*NewSocket)->LocalAddress = Datapath->ServerIP;
```
nibanks, line 202: Oops. Remove this.

src/platform/datapath_raw.c
```diff
@@ -182,25 +182,38 @@ CxPlatSocketCreateUdp(
...
+        } else {
+            (*NewSocket)->LocalAddress = *Config->LocalAddress;
+        }
+        if (Config->LocalAddress->Ipv4.sin_port != 0) {
+            (*NewSocket)->LocalAddress.Ipv4.sin_port =
+                Config->LocalAddress->Ipv4.sin_port;
+        }
     } else {
-        (*NewSocket)->LocalAddress.Ipv4.sin_port = 0;
+        CXPLAT_FRE_ASSERT((*NewSocket)->Connected); // Assumes connected socket
```
nibanks, line 210: Let's go ahead and handle this scenario correctly.

src/platform/datapath_raw.h
```diff
@@ -155,6 +155,7 @@ typedef struct CXPLAT_SOCKET {
     void* CallbackContext;
     QUIC_ADDR LocalAddress;
     QUIC_ADDR RemoteAddress;
+    BOOLEAN Wildcard;   // Using a wildcard local address
```
maolson-msft, line 158: Shouldn't we be able to get this information from the LocalAddress field? The normal way of using sockaddrs is for (e.g. for IPv4) 0.0.0.0 to mean "wildcard".

nibanks: Yes. This is purely an optimization so you don't have to constantly read the whole address everytime.

src/platform/datapath_raw.c
```diff
@@ -182,25 +182,43 @@ CxPlatSocketCreateUdp(
...
-        (*NewSocket)->Connected = FALSE;
-        CxPlatZeroMemory(&(*NewSocket)->RemoteAddress, sizeof(QUIC_ADDR));
-        (*NewSocket)->LocalAddress = Datapath->ServerIP;
     }
-    if (Config->LocalAddress && Config->LocalAddress->Ipv4.sin_port != 0) {
-        (*NewSocket)->LocalAddress.Ipv4.sin_port =
-            Config->LocalAddress->Ipv4.sin_port;
+
+    if (Config->LocalAddress) {
+        if (QuicAddrIsWildCard(Config->LocalAddress)) {
```
maolson-msft, line 197: This code seems to support two possibilities: either a LocalAddress that's wildcard, or no LocalAddress. Are both of those codepaths currently used by the core QUIC code?

nibanks: Are you asking for just unconnected sockets or both connected and unconnected?

src/platform/datapath_raw.h
```diff
@@ -169,6 +170,32 @@ CxPlatSockPoolUninitialize(
...
+inline
+BOOL
+CxPlatSocketCompare(
+    _In_ CXPLAT_SOCKET* Socket,
+    _In_ const QUIC_ADDR* LocalAddress,
+    _In_ const QUIC_ADDR* RemoteAddress
+    )
+{
+    if (Socket->Wildcard) {
+        return TRUE; // The local port match is all that is needed.
```
maolson-msft, line 187: Let's be judicious about assuming the hashtable lookup is based on the local port, because I think that could change. In TCPIP.sys, a lot of pain has come out of making the portpool a local port based table rather than a 4tuple-based table.

nibanks: We can add a debug assert.

src/platform/datapath_raw.c
maolson-msft: I'm specifically asking here about the two copy-pasted bits where we set LocalAddress to ClientIP, so connected.

maolson-msft: Er, actually I mean both. I'm saying that Config->LocalAddress==NULL and QuicAddrIsWildCard(Config->LocalAddress) seem to have the same meaning, and wondering why there need to be two different ways to do the same thing.

src/platform/datapath_raw.h
maolson-msft: Sounds good, but my comment is more about the code architecture- I'm saying keep this possible change in mind and avoid writing code that depends deeply on the local port being the HT key.

src/platform/datapath_raw.c
nibanks: Here are the supported used cases for QUIC:

Listener:
Remote address always null
Local address is wildcard or null

Client:
Remote address always fully specified
Local address may be null, fully wildcard, only port specified or fully specified

src/platform/datapath_raw.h
maolson-msft: I'd avoid this type of microoptimization at this stage since we're going to probably change the code multiple times and have to carry this extra logic through each time; but if you insist on this, please at least add a code comment to this field that notes it's an optimization- otherwise it's confusing to see.

src/platform/datapath_raw.c
maolson-msft: So zeroing in on the listener case, why would Config->LocalAddress be null versus fully wildcard?

I'm trying to understand why both cases exist, because having both cases makes the code here more complicated.

src/platform/datapath_raw_socket.c
```diff
@@ -97,7 +96,7 @@ CxPlatTryAddSocket(
     Entry = CxPlatHashtableLookup(&Pool->Sockets, Socket->LocalAddress.Ipv4.sin_port, &Context);
     while (Entry != NULL) {
         CXPLAT_SOCKET* Temp = CONTAINING_RECORD(Entry, CXPLAT_SOCKET, Entry);
-        if (QuicAddrCompareIp(&Temp->LocalAddress, &Socket->LocalAddress)) {
+        if (!CxPlatSocketCompare(Temp, &Socket->LocalAddress, &Socket->RemoteAddress)) {
```
maolson-msft, line 99: Why is this negated? Isn't it a collision if CxPlatSocketCompare returns TRUE?

nibanks: I agree, it's a bug.
```suggestion
        if (CxPlatSocketCompare(Temp, &Socket->LocalAddress, &Socket->RemoteAddress)) {
```

src/platform/datapath_raw.h
nibanks: I agree it's an assumption. While it may change in the future, I'm not so sure it will. So I'd rather go with the assert design that would immediately break if we change the assumption.

maolson-msft: Yeah, that sounds perfectly good to me.

src/platform/datapath_raw.c
nibanks: We support cases where a listener didn't specify a port up front (mostly for testing).

maolson-msft: K

src/platform/datapath_raw.h
```diff
@@ -155,7 +155,7 @@ typedef struct CXPLAT_SOCKET {
     void* CallbackContext;
     QUIC_ADDR LocalAddress;
     QUIC_ADDR RemoteAddress;
-    BOOLEAN Wildcard;   // Using a wildcard local address
+    BOOLEAN Wildcard;   // Using a wildcard local address. Optimization to always reading LocalAddress.
```
maolson-msft, line 158: Typo

src/platform/datapath_raw.h
```diff
@@ -184,6 +184,7 @@ CxPlatSocketCompare(
     )
 {
     if (Socket->Wildcard) {
+        CXPLAT_DBG_ASSERT(QuicAddrGetPort(&Socket->LocalAddress) == QuicAddrGetPort(LocalAddress));
```
maolson-msft, line 187: This assert should be before the "if (Socket->Wildcard)".

src/platform/datapath_raw.h
nibanks: Where?

maolson-msft: "to [avoid] always reading"

maolson-msft (approved):

PR #2117, opened 2021-11-04
Support multiple XDP queues

mtfriesen:
Initial support for multiple XDP queues, i.e. multiple RSS queues.

1. QUIC core and raw abstractions do not not provide a TX spreading scheme other than implicitly via worker CPU.
2. For simplicity, use only one user mode worker thread to pump all the kernel mode XDP queues.

### DIFF
--- a/src/platform/datapath_raw_xdp.c
+++ b/src/platform/datapath_raw_xdp.c
@@ -247,7 +247,7 @@ CxPlatDpRawInitialize(
         goto Error;
     }
 
-    ZeroMemory(Xdp->Queues, Xdp->QueueCount * sizeof(*Xdp->Queues));
+    CxPlatZeroMemory(Xdp->Queues, Xdp->QueueCount * sizeof(*Xdp->Queues));
 
     for (uint32_t QueueIndex = 0; QueueIndex < Xdp->QueueCount; QueueIndex++) {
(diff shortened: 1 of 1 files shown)


src/platform/datapath_raw_xdp.c
```diff
@@ -221,232 +232,252 @@ CxPlatDpRawInitialize(
...
-            Status,
-            "XskSetSockopt(XSK_SOCKOPT_UMEM_REG)");
-        goto Error;
-    }
+        InitializeSListHead(&Queue->RxPool);
+        InitializeSListHead(&Queue->TxPool);
+        CxPlatLockInitialize(&Queue->TxLock);
+        CxPlatListInitializeHead(&Queue->TxQueue);
+        CxPlatListInitializeHead(&Queue->WorkerTxQueue);
+        CxPlatZeroMemory(&Queue->WorkerRxPool, sizeof(Queue->WorkerRxPool));
```
nibanks, line 260: Wasn't this already zero'ed with everything else above?

nibanks (approved):

src/platform/datapath_raw_xdp.c
mtfriesen: Yeah, I was debating this - previously I had been explicitly initializing the Next field to null. I'll just remove the init.

PR #2120, opened 2021-11-04
Add basic IPv6 framing and parsing support

csujedihy:
- Big-endian systems are not yet supported.
- Packets with IPv6 extensions are discarded and traced.
- ECN related bits are left zero for now.

Tested with hardcoded IPv6 address on both XDP and DPDK.


src/inc/quic_datapath.h
```diff
@@ -555,7 +555,8 @@ CXPLAT_SEND_DATA*
 CxPlatSendDataAlloc(
     _In_ CXPLAT_SOCKET* Socket,
     _In_ CXPLAT_ECN_TYPE ECN,
-    _In_ uint16_t MaxPacketSize
+    _In_ uint16_t MaxPacketSize,
+    _In_ QUIC_ADDRESS_FAMILY Family
```
nibanks, line 559: I'd prefer if we could have a design that doesn't require this (though I'm not 100% against the change). I haven't looked in detail at the change yet, but I assume this is done to allocate exactly enough backfill, based on the selected address family. What if, instead, we simply always allocated enough space for the largest we'd ever need? Or do we think this would waste too much space?

src/platform/datapath_raw_socket.c
```diff
@@ -226,8 +235,10 @@ CxPlatDpRawParseIPv4(
     Packet->Tuple->LocalAddress.Ipv4.sin_family = AF_INET;
     CxPlatCopyMemory(&Packet->Tuple->LocalAddress.Ipv4.sin_addr, IP->Destination, sizeof(IP->Destination));
 
-    if (IP->Protocol == 17) {
+    if (IP->Protocol == IPPROTO_UDP) {
         CxPlatDpRawParseUdp(Datapath, Packet, (UDP_HEADER*)IP->Data, Length);
+    } else {
+        QuicTraceEvent(LibraryError, "[ lib] ERROR, %s.", "unacceptable transport protocol");
```
nibanks, line 241: We follow a very explicit, 1 arg per line, format for all events in MsQuic. Please update these to match that pattern. Also, also please run `.\scripts\update-sidecar.ps1` (btw, it takes a long time).

src/platform/datapath_raw_socket.c
```diff
@@ -252,27 +263,10 @@ CxPlatDpRawParseIPv6(
...
-            }
-            Length -= ExtLength;
-            Data += ExtLength;
-            if (Extension->NextHeader == 17) {
-                CxPlatDpRawParseUdp(Datapath, Packet, (UDP_HEADER*)Extension->Data, Length);
-            } else if (Extension->NextHeader == 59) {
-                return;
-            }
-        } while (TRUE);
+    if (IP->NextHeader == IPPROTO_UDP) {
```
nibanks, line 266: The reason I had all this "crap" before was because I noticed real IPv6 traffic had a lot of extensions before UDP in most cases. If you want to remove it for now, I'm fine with that, but I suspect it will need to come back.

src/platform/datapath_raw_socket.c
```diff
@@ -307,10 +301,28 @@ CxPlatDpRawParseEthernet(
         return;
     }
 
-    if (Ethernet->Type == 0x0008) { // IPv4
+    uint16_t EthernetType = QuicNetByteSwapShort(Ethernet->Type);
```
nibanks, line 304: IMO, the byteswap is unnecessary. Just compare in wire format. I know it's a micro-optimization, but it's a trivial one to do. With all these PRs, we should keep perf at the forefront of our minds.

src/platform/datapath_raw_socket.c
nibanks: Also, all these "unacceptable" error cases should use `LibraryErrorStatus` and pass the value in for the "status" field, probably. It might be worth having a separate event for these too; because they might be super noisy.

src/platform/datapath_raw_socket.c
```diff
@@ -320,14 +332,90 @@ CxPlatFramingWriteHeaders(
...
+    _Inout_ QUIC_BUFFER* Buffer,
+    _In_ QUIC_ADDRESS_FAMILY Family
     )
 {
     UDP_HEADER* UDP = (UDP_HEADER*)(Buffer->Buffer - sizeof(UDP_HEADER));
-    IPV4_HEADER* IP = (IPV4_HEADER*)(((uint8_t*)UDP) - sizeof(IPV4_HEADER));
-    ETHERNET_HEADER* Ethernet = (ETHERNET_HEADER*)(((uint8_t*)IP) - sizeof(ETHERNET_HEADER));
+    ETHERNET_HEADER* Ethernet;
+    IPV4_HEADER* IPv4;
+    IPV6_HEADER* IPv6;
```
nibanks, line 342: Are these needed at the global scope? If not, push them into their specific scopes. We don't require declaring all variables at the beginning. Feel free to move them to when first used.

src/platform/datapath_raw_socket.c
```diff
@@ -320,14 +332,90 @@ CxPlatFramingWriteHeaders(
...
+        Ethernet = (ETHERNET_HEADER*)(((uint8_t*)IPv4) - sizeof(ETHERNET_HEADER));
+        IpHeaderLen = sizeof(IPV4_HEADER);
+    } else {
+        IPv6 = (IPV6_HEADER*)(((uint8_t*)UDP) - sizeof(IPV6_HEADER));
+        //
+        // IPv6 Version, Traffic Class, ECN Field and Flow Label fields in host
+        // byte order.
+        // TODO: support big-endian systems. Not quite sure if it's needed since QUIC core code
+        // also doesn't support big-endian systems due to the use of CxPlatByteSwapXxx which swaps
+        // bytes even when there's no need.
```
nibanks, line 381: No need to have this explicit TODO here. As you've pointed out, all of MsQuic doesn't support big-endian already. We haven't had a need to do so yet.

src/platform/datapath_raw_socket.c
```diff
@@ -320,14 +332,90 @@ CxPlatFramingWriteHeaders(
...
+        VersionClassEcnFlow.Version = IPV6_VERSION;
+        VersionClassEcnFlow.Class = 0;
+        VersionClassEcnFlow.EcnField = 0; // Not ECN capable currently.
+        VersionClassEcnFlow.Flow = FlowId;
+        if ((FlowId != 0) && (VersionClassEcnFlow.Flow == 0)) {
+            //
+            // If LSBs (20 bits) of Flow are all 0s, use its MSBs.
+            //
+            VersionClassEcnFlow.Flow = (FlowId >> 12);
+        }
```
nibanks, line 403: This is interesting, but is it strictly necessary?

src/platform/datapath_raw_xdp.c
```diff
@@ -683,17 +683,18 @@ CXPLAT_SEND_DATA*
 CxPlatDpRawTxAlloc(
     _In_ CXPLAT_DATAPATH* Datapath,
     _In_ CXPLAT_ECN_TYPE ECN, // unused currently
-    _In_ uint16_t MaxPacketSize
+    _In_ uint16_t MaxPacketSize,
+    _In_ QUIC_ADDRESS_FAMILY Family
```
nibanks, line 687: Please update DPDK code as well. We are maintaining both DPDK and XDP.

src/inc/quic_datapath.h
thhous-msft: We would 100% need to do perf testing on this. We've attempted changes that add extra args on the send path, and it has sometimes cost 10% of performance. I feel like a few extra bytes isn't a big deal when the alternative is this invasive of a change not normally used in any other datapath.

src/platform/datapath_raw_socket.c
maolson-msft: Is the 1-arg-per-line convention required by clog? Just curious why this convention is for only one type of macro call.

nibanks: It's not a hard requirement, just our common practice that I'd prefer to keep consistent.

src/inc/quic_datapath.h
csujedihy: Right, we need to know how much space the header needs so that the offset for the payload can be determined. Since XDP packets are all just pre-allocated in the size of MAX_ETH_FRAME_SIZE, if we always allocate say 62 (IPv6+Eth+UDP), then we will waste 20 bytes because payload is written before headers. So, I think my current change looks very natural to me.

I am surprised that an extra argument could cause 10% perf loss. In any other datapath, this parameter should be optimized out.

src/platform/datapath_raw_socket.c
csujedihy: Are you sure in most cases we will have extensions? I just took a packet capture for buffering a youtube video and I can't find any extensions in the QUIC flow.

nibanks: Maybe it was just the local traffic in the lab. We can keep this as is for now.

src/platform/datapath_raw_socket.c
csujedihy: You meant the flow field is necessary or the part that moves bits around? Flow ID is not strictly necessary. This field is for routers and middlebox to do QoS or load distribution. It's recommended (https://datatracker.ietf.org/doc/html/rfc6437) to have one per flow but leaving to zero works too.

nibanks: I just mean, in general converting the socket pointer to a flow id. It requires an extra branch in the code that has a limited perf hit, for little or no real gain.

csujedihy: Ok, I can try to skip the branch.

src/inc/quic_datapath.h
maolson-msft: I'm also surprised to hear that just adding args in a datapath function caused a 10% perf regression. Are you sure the regression wasn't actually from whatever logic was related to the added args?

src/platform/datapath_raw_xdp.c
csujedihy: DPDK updated and validated.

src/platform/datapath_raw.c
```diff
@@ -434,7 +436,8 @@ CxPlatSocketSend(
         (uint16_t)SendData->Buffer.Length,
         CASTED_CLOG_BYTEARRAY(sizeof(*RemoteAddress), RemoteAddress),
         CASTED_CLOG_BYTEARRAY(sizeof(*LocalAddress), LocalAddress));
-    CxPlatFramingWriteHeaders(Socket, LocalAddress, RemoteAddress, &SendData->Buffer);
+    CxPlatFramingWriteHeaders(
```
nibanks, line 439: There should be no need to pass family here, since the addresses are already passed in.

src/platform/datapath_raw_socket.c
```diff
@@ -307,27 +312,115 @@ CxPlatDpRawParseEthernet(
...
     _In_ const QUIC_ADDR* RemoteAddress,
-    _Inout_ QUIC_BUFFER* Buffer
+    _Inout_ QUIC_BUFFER* Buffer,
+    _In_ QUIC_ADDRESS_FAMILY Family
     )
 {
     UDP_HEADER* UDP = (UDP_HEADER*)(Buffer->Buffer - sizeof(UDP_HEADER));
-    IPV4_HEADER* IP = (IPV4_HEADER*)(((uint8_t*)UDP) - sizeof(IPV4_HEADER));
-    ETHERNET_HEADER* Ethernet = (ETHERNET_HEADER*)(((uint8_t*)IP) - sizeof(ETHERNET_HEADER));
+    ETHERNET_HEADER* Ethernet;
```
nibanks, line 356: Might there be a slight cache locality improvement to write the header from front (ethernet) to back (udp) order?

src/platform/datapath_raw.c
csujedihy: oh right, brainfart.

src/platform/datapath_raw_socket.c
csujedihy: Can you explain why? Even if we try to write eth header first, the cache doesn't necessarily start with the address of the eth header. Whether all headers are in the same line is not determined by which byte we first try to access.

nibanks: Just wondering.

src/platform/datapath_raw.c
csujedihy: Ok removed now

src/platform/datapath_raw.h
```diff
@@ -39,6 +39,7 @@ typedef struct CXPLAT_DATAPATH {
 typedef struct CXPLAT_SEND_DATA {
 
     QUIC_BUFFER Buffer;
+    QUIC_ADDRESS_FAMILY Family;
```
nibanks, line 42: Can this be removed now?

csujedihy: Sorry, I missed it. Now fixed.

csujedihy:
Need a sign-off to unblock work from Matt.

nibanks (approved):

nibanks:
Ship it, but I reserve the right to ask for more changes before it's actually merged into main.

csujedihy:
> Ship it, but I reserve the right to ask for more changes before it's actually merged into main.

Sure thing. Thanks!

PR #2123, opened 2021-11-05
Update QuicTrace app to .NET 6 and QuicTraceLib to newest SDK

### LINKED ISSUE #1491: Single File QuicTrace.exe

Right now, quictrace.exe depends on a lot of stuff. It might be better to make it a single file executable and package up all the stuff in a single binary.

thhous-msft:


thhous-msft:
New versions of WPA will not be compatible with older plugins. This PR updates the plugins to be compatible with the newer versions of WPA.

thhous-msft:
Removed single file exe as its blocked.

nibanks (approved):

PR #2125, opened 2021-11-07
GitHub Action for Repo Mirroring

nibanks:
Moves the mirroring job to a GitHub Action. Also adds tag support, and now includes mirroring for prerelease branches. No longer requires that all builds complete first. This isn't really necessary any more since these consistently pass always anyways.


thhous-msft (approved):
As long as we're sure were OK with not requiring builds to mirror I'm OK with this.

PR #2126, opened 2021-11-08
GitHub Action for WAN Perf

nibanks:
Starts the process of refactoring and moving WAN perf testing. Uses a matrix to run parallel jobs.

Future work:

- Remove AZP WAN perf and ensure this Action replaces all functionality.
- Periodically do even more cases (i.e. TCP)

### DIFF
--- a/.github/workflows/wan-perf.yml
+++ b/.github/workflows/wan-perf.yml
@@ -19,10 +19,10 @@ jobs:
       uses: actions/checkout@v2
     - name: Prepare Machine
       shell: pwsh
-      run: scripts/prepare-machine.ps1 -Configuration Build -InitSubmodules
+      run: scripts/prepare-machine.ps1 -Configuration Build -InitSubmodules -Extra "-DisableTest -DisableTools"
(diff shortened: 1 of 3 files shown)


nibanks:
Merged results available: https://github.com/microsoft/msquic/runs/4153995071?check_suite_focus=true#step:4:14

.github/workflows/wan-perf.yml
```diff
@@ -0,0 +1,110 @@
...
+    - src/cs/*
+    - src/distribution/*
+    - src/plugins/*
+    - src/tools/*
+    - src/lib.rs
+    - Cargo.toml
+    - README.md
+
+env:
+  seed: 41473a2e60b6958500ec0add7dcfb9 # TODO - Randomize?
```
anrossi, line 34: File an issue to randomize this so we don't lose track of it? WAN perf won't vary ever if this is always hard-coded.

anrossi (approved):
Looks good, modulo the TODO.

PR #2130, opened 2021-11-09
Create CXPLAT_ROUTE to replace QUIC_PATH::LocalAddress and QUIC_PATH::RemoteAddress

maolson-msft:
The signature of CxPlatSocketSend is updated so that a pointer to the QUIC_PATH's CXPLAT_ROUTE is passed instead of pointers to its LocalAddress and RemoteAddress. This enables additional state (specifically the next hop for the selected route) to be stored in QUIC_PATH and passed to the datapath layer without massive plumbing.

### DIFF
--- a/src/core/binding.c
+++ b/src/core/binding.c
@@ -1058,10 +1058,13 @@ QuicBindingProcessStatelessOperation(
         goto Exit;
     }
 
+    CXPLAT_ROUTE Route;
+    Route.LocalAddress = RecvDatagram->Tuple->LocalAddress;
+    Route.RemoteAddress = RecvDatagram->Tuple->RemoteAddress;
+
     QuicBindingSend(
         Binding,
-        &RecvDatagram->Tuple->LocalAddress,
-        &RecvDatagram->Tuple->RemoteAddress,
+        &Route,
         SendData,
         SendDatagram->Length,
         1,
@@ -1745,8 +1748,7 @@ _IRQL_requires_max_(DISPATCH_LEVEL)
 QUIC_STATUS
 QuicBindingSend(
     _In_ QUIC_BINDING* Binding,
-    _In_ const QUIC_ADDR* LocalAddress,
-    _In_ const QUIC_ADDR* RemoteAddress,
+    _In_ const CXPLAT_ROUTE* Route,
     _In_ CXPLAT_SEND_DATA* SendData,
     _In_ uint32_t BytesToSend,
     _In_ uint32_t DatagramsToSend,
@@ -1759,12 +1761,12 @@ QuicBindingSend(
     QUIC_TEST_DATAPATH_HOOKS* Hooks = MsQuicLib.TestDatapathHooks;
     if (Hooks != NULL) {
 
-        QUIC_ADDR RemoteAddressCopy = *RemoteAddress;
-        QUIC_ADDR LocalAddressCopy = *LocalAddress;
+        CXPLAT_ROUTE RouteCopy = *Route;
+
         BOOLEAN Drop =
             Hooks->Send(
-                &RemoteAddressCopy,
-                &LocalAddressCopy,
+                &RouteCopy.RemoteAddress,
+                &RouteCopy.LocalAddress,
                 SendData);
 
         if (Drop) {
@@ -1778,8 +1780,7 @@ QuicBindingSend(
             Status =
                 CxPlatSocketSend(
                     Binding->Socket,
-                    &LocalAddressCopy,
-                    &RemoteAddressCopy,
+                    &RouteCopy,
                     SendData,
                     IdealProcessor);
             if (QUIC_FAILED(Status)) {
@@ -1795,8 +1796,7 @@ QuicBindingSend(
         Status =
             CxPlatSocketSend(
                 Binding->Socket,
-                LocalAddress,
-                RemoteAddress,
+                Route,
                 SendData,
                 IdealProcessor);
         if (QUIC_FAILED(Status)) {
--- a/src/platform/unittest/DataPathTest.cpp
+++ b/src/platform/unittest/DataPathTest.cpp
@@ -242,11 +242,13 @@ protected:
                 ASSERT_NE(nullptr, ServerBuffer);
                 memcpy(ServerBuffer->Buffer, RecvData->Buffer, RecvData->BufferLength);
 
+                CXPLAT_ROUTE Route;
+                Route.LocalAddress = RecvData->Tuple->LocalAddress;
+                Route.RemoteAddress = RecvData->Tuple->RemoteAddress;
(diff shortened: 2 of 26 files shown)


csujedihy:
> The signature of CxPlatSocketSend is updated so that a pointer to the QUIC_PATH's CXPLAT_ROUTE is passed instead of pointers to its LocalAddress and RemoteAddress. This enables additional state (specifically the next hop for the selected route) to be stored in QUIC_PATH and passed to the datapath layer without massive plumbing.

Instead of passing route, should we pass the QUIC_PATH? The concept of route seems to only make sense to raw data path.  If in the future we need to plumb something specific to the path (but not the route), we will do this invasive plumbing change again.

maolson-msft:
> > The signature of CxPlatSocketSend is updated so that a pointer to the QUIC_PATH's CXPLAT_ROUTE is passed instead of pointers to its LocalAddress and RemoteAddress. This enables additional state (specifically the next hop for the selected route) to be stored in QUIC_PATH and passed to the datapath layer without massive plumbing.
> 
> Instead of passing route, should we pass the QUIC_PATH? The concept of route seems to only make sense to raw data path. If in the future we need to plumb something specific to the path (but not the route), we will do this invasive plumbing change again.

The general reason (I'm speaking here for @nibanks) for passing a substruct rather than the entire QUIC_PATH is that QUIC_PATH is a private struct to core QUIC and most of its fields are not needed in the datapath layer.

There's also a more concrete reason: in QuicBindingProcessStatelessOperation (and many places in test code/tools that directly call into the datapath layer), there's actually no QUIC_PATH to pass down, and instead a CXPLAT_ROUTE struct is created and initialized to pass down.

"If in the future we need to plumb something specific to the path (but not the route)..." - I think this is really just a disagreement on the choice of the name "route". Really a path and a route are the same thing in this context, and the "route" is just the part of the path that is used by the datapath layer. If you can think of a better name, then let me know; but without knowing what this future state is, I really think "route" is the best name for the time being.

csujedihy:
> > > The signature of CxPlatSocketSend is updated so that a pointer to the QUIC_PATH's CXPLAT_ROUTE is passed instead of pointers to its LocalAddress and RemoteAddress. This enables additional state (specifically the next hop for the selected route) to be stored in QUIC_PATH and passed to the datapath layer without massive plumbing.
> > 
> > 
> > Instead of passing route, should we pass the QUIC_PATH? The concept of route seems to only make sense to raw data path. If in the future we need to plumb something specific to the path (but not the route), we will do this invasive plumbing change again.
> 
> The general reason (I'm speaking here for @nibanks) for passing a substruct rather than the entire QUIC_PATH is that QUIC_PATH is a private struct to core QUIC and most of its fields are not needed in the datapath layer.
> 
> There's also a more concrete reason: in QuicBindingProcessStatelessOperation (and many places in test code/tools that directly call into the datapath layer), there's actually no QUIC_PATH to pass down, and instead a CXPLAT_ROUTE struct is created and initialized to pass down.
> 
> "If in the future we need to plumb something specific to the path (but not the route)..." - I think this is really just a disagreement on the choice of the name "route". Really a path and a route are the same thing in this context, and the "route" is just the part of the path that is used by the datapath layer. If you can think of a better name, then let me know; but without knowing what this future state is, I really think "route" is the best name for the time being.

Cool, I am convinced. I can't think of a better name than route.

src/core/binding.c
```diff
@@ -1058,10 +1058,13 @@ QuicBindingProcessStatelessOperation(
         goto Exit;
     }
 
+    CXPLAT_ROUTE Route;
+    Route.LocalAddress = RecvDatagram->Tuple->LocalAddress;
+    Route.RemoteAddress = RecvDatagram->Tuple->RemoteAddress;
```
nibanks, line 1063: Why aren't we replacing the `Tuple` in the `RecvDatagram` with `Route`? Then that could have simply been passed right back.

src/test/lib/QuicDrill.cpp
```diff
@@ -215,11 +212,14 @@ struct DrillSender {
         //
         memcpy(SendBuffer->Buffer, PacketBuffer->data(), DatagramLength);
 
+        CXPLAT_ROUTE Route;
+        CxPlatSocketGetLocalAddress(Binding, &Route.LocalAddress);
+        Route.RemoteAddress = ServerAddress;
```
nibanks, line 217: Why move this down? I *believe* Yi's change had to move some of these **up** so this might cause conflicts. I recommend just putting these up.

src/platform/unittest/DataPathTest.cpp
```diff
@@ -506,11 +508,13 @@ struct CxPlatSocket {
         _In_ uint16_t PartitionId = 0
         ) const noexcept
     {
+        CXPLAT_ROUTE Route;
+        Route.LocalAddress = LocalAddress;
+        Route.RemoteAddress = RemoteAddress;
```
nibanks, line 513: While it would be a larger change, I think it's more future proof to update this `Send` function to take the `Route`; otherwise you're just going to have to change it later when we start adding more to the route.

src/test/lib/QuicDrill.cpp
nibanks: Ditto with all the other ones below.

maolson-msft: The idea was to follow the convention of declaring variables as near as possible to where they're used. Will move back up here and elsewhere to reduce merge conflicts.

csujedihy: My change is already in. There should be no merge conflicts.

csujedihy: Oh, this is going to main. Nvm.

csujedihy: Actually, should this entire change go into nibanks/dpdk first and then we can bring both API changes up to main? Or are you going to first get this in main and resolve conflicts when merging to dpdk?

src/core/binding.c
maolson-msft: Yeah that's the plan, but it's a big change by itself and I'd like to put it in its own PR.

src/test/lib/QuicDrill.cpp
maolson-msft: Yeah, the plan is to merge main into nibanks/dpdk after checking this in.

csujedihy: Seems like I made a mistake not checking into main for the API changes first. Now, bringing them up will be difiicult.

nibanks (approved):

src/platform/unittest/DataPathTest.cpp
maolson-msft: Discussed offline; will leave this as-is for now.

PR #2132, opened 2021-11-10
Change CXPLAT_TUPLE field of CXPLAT_RECV_DATA to CXPLAT_ROUTE

maolson-msft:
In addition to QUIC_PATHs having a "route", now CXPLAT_RECV_DATAs (which represent received packets) also have one. This enables the route of the received packet which triggered a stateless operation to be directly passed to QuicBindingSend in QuicBindingProcessStatelessOperation.

The idea here is that LoLa can potentially skip route lookup and ARP when sending stateless replies on the dpdk or xdp datapath.

### DIFF
--- a/src/core/packet.c
+++ b/src/core/packet.c
@@ -711,8 +711,8 @@ QuicPacketLogDrop(
             ConnDropPacket,
             "[conn][%p] DROP packet Dst=%!ADDR! Src=%!ADDR! Reason=%s.",
             Owner,
-            CASTED_CLOG_BYTEARRAY(sizeof(Datagram->Tuple->LocalAddress), &Datagram->Tuple->LocalAddress),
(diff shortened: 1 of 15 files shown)


src/core/packet.c
```diff
@@ -711,17 +711,17 @@ QuicPacketLogDrop(
             ConnDropPacket,
             "[conn][%p] DROP packet Dst=%!ADDR! Src=%!ADDR! Reason=%s.",
             Owner,
-            CASTED_CLOG_BYTEARRAY(sizeof(Datagram->Tuple->LocalAddress), &Datagram->Tuple->LocalAddress),
-            CASTED_CLOG_BYTEARRAY(sizeof(Datagram->Tuple->RemoteAddress), &Datagram->Tuple->RemoteAddress),
+            CASTED_CLOG_BYTEARRAY(sizeof(Datagram->Route->LocalAddress), &Datagram->Route->LocalAddress),
```
csujedihy, line 714: run clog?

csujedihy (approved):

nibanks (approved):

PR #2146, opened 2021-11-15
Add support for building arm64ec binaries

thhous-msft:
Arm64ec is an emulation compatible arm64 calling convention. It allows x64 and arm64 code to mix in the same exectable. This is great for a program that uses plugins, as it can use exisiting x64 plugins but run most of its code as arm code.

Only supports schannel, as openssl would require changes to their assembly files, which they do not have any interest in doing.

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -229,8 +229,6 @@ if ($Arch -ne "x64" -And ($Platform -eq "gamecore_console")) {
     exit
 }
 
-
-
 if ($Arch -eq "arm64ec") {
     if (!$IsWindows) {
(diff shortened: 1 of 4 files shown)


scripts/build.ps1
```diff
@@ -229,6 +229,17 @@ if ($Arch -ne "x64" -And ($Platform -eq "gamecore_console")) {
     exit
 }
 
+
+
```
nibanks, line 233: nit: remove extra newlines

nibanks (approved):

nibanks (approved):

PR #2152, opened 2021-11-16
Stub Route Resolution and Next Hop Selection

nibanks:


src/platform/datapath_raw_xdp.c
```diff
@@ -832,3 +857,22 @@ CXPLAT_THREAD_CALLBACK(CxPlatXdpWorkerThread, Context)
...
+    if (Xdp->Affinitize) {
+        GROUP_AFFINITY Affinity = {0};
+
+        Affinity.Group = Xdp->DatapathCpuGroup;
+        Affinity.Mask = (ULONG_PTR)1 << Xdp->DatapathCpuNumber;
+        SetThreadGroupAffinity(GetCurrentThread(), &Affinity, NULL);
+    }
+
+    while (Xdp->Running) {
+        CxPlatTimeUs64();
```
csujedihy, line 865: What does this while loop do? Wouldn't this lock up the machine?

csujedihy: Oh this is just a placeholder?

nibanks: This slipped into this PR accidentally. It is a thread meant to contend with the normal polling thread, to test the effect on latency. It's just meant to run at 100%.

src/inc/quic_datapath.h
```diff
@@ -383,6 +387,31 @@ CxPlatDataPathResolveAddress(
...
+// Resolves a route.
+//
+_IRQL_requires_max_(PASSIVE_LEVEL)
+QUIC_STATUS
+CxPlatDataPathLookupRoute(
+    _In_ CXPLAT_DATAPATH* Datapath,
+    _Inout_ CXPLAT_ROUTE* Route
+    );
+
+//
```
csujedihy, line 402: resolves the next hop

src/core/connection.c
```diff
@@ -197,6 +197,10 @@ QuicConnAlloc(
             Connection,
             CASTED_CLOG_BYTEARRAY(sizeof(Path->Route.RemoteAddress), &Path->Route.RemoteAddress));
 
+        if (QUIC_FAILED(QuicPathResolveRoute(Path))) {
+            goto Error;
+        }
```
nibanks, line 202: All this logic of making sure to resolve the route at the right time is complicated and error prone (maybe we missed a spot; maybe one these is too early). @maolson-msft  what do you think of adding support for **in** the send call? We put a flag in the `CXPLAT_ROUTE` called something like `IsResolved`, and in the send, if it's not resolved it JIT does it? We'd still need to expose this for the client first packet scenario I think (maybe not?).

maolson-msft: Remember, we want to potentially re-resolve the route later on when there is retransmission, so we still need to expose the API.

Do you think you missed a spot?
-new client connection
-new server connection
-path change

What else could create a new path?

src/core/path.c
```diff
@@ -289,3 +289,25 @@ QuicPathSetActive(
     CXPLAT_DBG_ASSERT(Path->DestCid != NULL);
     CXPLAT_DBG_ASSERT(!Path->DestCid->CID.Retired);
 }
+
+#ifdef QUIC_ROUTE_EXTRAS
+_IRQL_requires_max_(PASSIVE_LEVEL)
+QUIC_STATUS
+QuicPathResolveRoute(
```
csujedihy, line 296: The function name is resolve route but it does both route lookup and next hop resolution.

nibanks:
We've decided that we don't need this. We'll do it implicitly in the send path. Closing.

PR #2153, opened 2021-11-16
Add basic RX validation

csujedihy:
This PR is adding some necessary checks so that we won't access data beyond the packet's boundary and some basic IPv4 field validations.

RX checksum validation with dpdk will be in a separate PR.

### DIFF
--- a/src/platform/datapath_raw_socket.c
+++ b/src/platform/datapath_raw_socket.c
@@ -380,7 +380,7 @@ CxPlatDpRawParseIPv4(
             Datapath,
             Length,
             "packet is too small for an IPv4 header");
-        goto Done;
+        return;
     }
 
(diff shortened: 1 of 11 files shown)


src/platform/datapath_raw_socket.c
```diff
@@ -280,8 +280,8 @@ typedef struct ETHERNET_HEADER {
 } ETHERNET_HEADER;
 
 typedef struct IPV4_HEADER {
-    uint8_t VersionAndHeaderLength;
-    uint8_t TypeOfService;
+    UINT8 VersionAndHeaderLength;
```
nibanks, line 283: Please use `uint8_t`, not Windows `UINT8` specific.

csujedihy: oh right, accidently slipped in from our tcpip code.

nibanks (approved):

PR #2155, opened 2021-11-17
Fix memory leak when indicated certificate is rejected by app

thhous-msft:
When indicate certificate is set and the certificate is rejected, we would leak the cert context. Also if the cert context failed to be grabbed we'd deref a null pointer. Both issues are fixed

### DIFF
--- a/src/platform/tls_schannel.c
+++ b/src/platform/tls_schannel.c
@@ -1596,6 +1596,8 @@ CxPlatTlsIndicateCertificateReceived(
 {
     SECURITY_STATUS SecStatus;
     CXPLAT_TLS_RESULT_FLAGS Result = 0;
+    QUIC_CERTIFICATE* Certificate = NULL;
+    QUIC_CERTIFICATE_CHAIN* CertificateChain = NULL;
 
 #ifdef _KERNEL_MODE
     SecPkgContext_Certificates PeerCert;
@@ -1620,22 +1622,20 @@ CxPlatTlsIndicateCertificateReceived(
             TlsContext->Connection,
             SecStatus,
             "Query peer cert");
-        Result |= CXPLAT_TLS_RESULT_ERROR;
-        State->AlertCode = CXPLAT_TLS_ALERT_CODE_INTERNAL_ERROR;
-        goto Exit;
-    }
-#ifndef _KERNEL_MODE
-    CXPLAT_DBG_ASSERT(PeerCert != NULL);
+    } else {
+#ifdef _KERNEL_MODE
+        Certificate = (QUIC_CERTIFICATE*)&PeerCert;
+        CertificateChain = (QUIC_CERTIFICATE_CHAIN*)&PeerCert;
+#else
+        CXPLAT_DBG_ASSERT(PeerCert != NULL);
+        Certificate = (QUIC_CERTIFICATE*)PeerCert;
+        CertificateChain = (QUIC_CERTIFICATE_CHAIN*)(PeerCert->hCertStore);
 #endif
+    }
     if (!TlsContext->SecConfig->Callbacks.CertificateReceived(
             TlsContext->Connection,
-#ifdef _KERNEL_MODE
-            (QUIC_CERTIFICATE*)&PeerCert,
-            (QUIC_CERTIFICATE_CHAIN*)&PeerCert,
-#else
-            (QUIC_CERTIFICATE*)PeerCert,
-            (QUIC_CERTIFICATE_CHAIN*)(PeerCert->hCertStore),
-#endif
+            Certificate,
+            CertificateChain,
             CertValidationResult->dwChainErrorStatus,
             (QUIC_STATUS)CertValidationResult->hrVerifyChainStatus)) {
         QuicTraceEvent(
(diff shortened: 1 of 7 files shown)


src/platform/tls_schannel.c
```diff
@@ -2091,9 +2094,11 @@ CxPlatTlsWriteDataToSchannel(
                         "Indicate certificate received failed");
                     Result |= CXPLAT_TLS_RESULT_ERROR;
                     State->AlertCode = CXPLAT_TLS_ALERT_CODE_BAD_CERTIFICATE;
-                    break;
+                    goto IndicateReceivedCleanup;
                 }
 
+IndicateReceivedCleanup:
```
anrossi, line 2100: Do we have other `goto` statements like this in the middle of a function? Usually `goto` is reserved for single-return style coding.

src/platform/tls_schannel.c
```diff
@@ -2103,6 +2108,9 @@ CxPlatTlsWriteDataToSchannel(
                     CertFreeCertificateContext(PeerCert);
                 }
 #endif
+                if ((Result & CXPLAT_TLS_RESULT_ERROR) != 0) {
```
anrossi, line 2111: I think we treat fetching the client certificate as best-effort; making it a fatal error then means that applications which don't have a hard-requirement on the client certificate will always fail their connections in this case.

src/platform/tls_schannel.c
thhous-msft: We have some goto's for retrys in a few places. But there isn't a super great way to do this other then creating a leaf function. Which might be worth it TBH because this function is 1000 lines long.

src/platform/tls_schannel.c
thhous-msft: Our option is either that, or we don't pass up to the user even though they've requested it. I'm not sure which is worse. Also OpenSSL would fail in this manner.

thhous-msft: Note that this isn't just client certificates. Server certs can be called with this too.

src/platform/tls_schannel.c
thhous-msft: I decided to just make a leaf function, its cleaner.

anrossi: This is much more readable! Thank you

src/platform/tls_schannel.c
anrossi: Maybe it should be fatal on client for customer server cert validation (if the NO_VALIDATION flag is missing), and it should be optional on server unless the REQUIRE_CLIENT_CERTS flag is present?  That sounds complex, and would require some specific tests to ensure that code path works.

thhous-msft: OHHH I get what youre getting at now. Clients would fail even though its not a requirement. Thats actually not that hard to fix.

anrossi: Also, I tried to make sure that OpenSSL and Schannel behave the same in this area. So we'd have to make sure they both still act the same for this.

thhous-msft:
We had 4 options for handling QueryContextAttributesW failures here

1. Abort the connection
2. Don't indicate to peer
3. Indicate NULL
4. Debug assert/Release null deref

The original code did 3. When portable certificate support was added, this was changed to 4, which is not good. After some discussion, it seems like 3 is the best option to keep. 1 means if there is ever an error, the certificate might have fully validated, but the connection would still be aborted. 2 ignores a contract requested by the app, and is problematic. 3 seems like the best option, as a callback still occurs and its possible to know an error getting the cert has occured, and the app can decided what to do.

src/platform/tls_schannel.c
```diff
@@ -1586,6 +1586,83 @@ CxPlatTlsUninitialize(
...
+            SecStatus,
+            "Query peer cert");
+    } else {
+#ifdef _KERNEL_MODE
+        Certificate = (QUIC_CERTIFICATE*)&PeerCert;
+        CertificateChain = (QUIC_CERTIFICATE_CHAIN*)&PeerCert;
+#else
+        CXPLAT_DBG_ASSERT(PeerCert != NULL);
+        Certificate = (QUIC_CERTIFICATE*)PeerCert;
+        CertificateChain = (QUIC_CERTIFICATE_CHAIN*)(PeerCert->hCertStore);
```
anrossi, line 1632:
```suggestion
        CertificateChain = (QUIC_CERTIFICATE_CHAIN*)(PeerCert ? PeerCert->hCertStore : NULL);
```

thhous-msft: In this case, this is ok. This only occurs when the query call returns success, and the contract there says it returns a valid cert if successful.

thhous-msft: I could theoretically remove the dbg assert too, but I'd really like to know if schannel ever breaks that contract.

anrossi: Oh, now I see what that you did there. Makes sense.  These `#ifdef`s mess up readability, IMO (but there's no alternative).

anrossi (approved):

nibanks (approved):

PR #2157, opened 2021-11-18
Update docs to make clear VS 2022 will work for building

thhous-msft:


docs/BUILD.md
```diff
@@ -98,15 +98,15 @@ For the very first time you build, it's recommend to make sure you have all the
 ./scripts/prepare-machine.ps1 -Configuration Dev
 ```
 
-Note at minimum CMake 3.16 is required. Instructions for installing the newest version on Ubuntu can be found here. https://apt.kitware.com/. The prepare-machine script will not do this for you.
+Note at minimum CMake 3.20 on windows and 3.16 on other platforms is required. Instructions for installing the newest version on Ubuntu can be found here. https://apt.kitware.com/. The prepare-machine script will not do this for you.
```
maolson-msft, line 101: At the very top of this file is another mention of CMake (which still says version 3.16). I was already going to suggest that it be removed, so this seems like a good time to do so. I think we ought to remove this entire paragraph:

"The MsQuic build system relies on [CMake](https://cmake.org/) (3.16 or better), [.NET Core](https://dotnet.microsoft.com/download/dotnet-core) (Core 3.1 or 5.0 SDK) and [Powershell](https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell) (7.0 or better) on all platforms."

nibanks: I'd like to keep that sentence, but probably remove all the version numbers.

nibanks (approved):

docs/BUILD.md
thhous-msft: I removed the cmake version, and .NET Core completely since its not needed for a raw build.

I left powershell 7.0 or higher, as otherwise confusion with built in powershell occurs. And unlike cmake which we have reasons to require new versions, that is very unlikely to ever happen with powershell.

nibanks (approved):

PR #2159, opened 2021-11-18
Pass Route into CxPlatSocketCreateUdp instead of Addresses

nibanks:
Updates `CxPlatSocketCreateUdp` to take a `CXPLAT_ROUTE` as input instead of `QUIC_ADDR`s. This is useful if we want the `CxPlatSocketCreateUdp` function itself to modify the route (route / next hop).

### DIFF
--- a/src/core/binding.c
+++ b/src/core/binding.c
@@ -58,7 +58,9 @@ QuicBindingInitialize(
     Binding->RefCount = 0; // No refs until it's added to the library's list
     Binding->Exclusive = !(UdpConfig->Flags & CXPLAT_SOCKET_FLAG_SHARE);
     Binding->ServerOwned = !!(UdpConfig->Flags & CXPLAT_SOCKET_SERVER_OWNED);
-    Binding->Connected = UdpConfig->RemoteAddress == NULL ? FALSE : TRUE;
+    Binding->Connected =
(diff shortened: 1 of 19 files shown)


src/core/binding.c
```diff
@@ -58,7 +58,9 @@ QuicBindingInitialize(
     Binding->RefCount = 0; // No refs until it's added to the library's list
     Binding->Exclusive = !(UdpConfig->Flags & CXPLAT_SOCKET_FLAG_SHARE);
     Binding->ServerOwned = !!(UdpConfig->Flags & CXPLAT_SOCKET_SERVER_OWNED);
-    Binding->Connected = UdpConfig->RemoteAddress == NULL ? FALSE : TRUE;
+    Binding->Connected =
+        (QuicAddrGetFamily(&UdpConfig->Route->RemoteAddress) == QUIC_ADDRESS_FAMILY_UNSPEC) ?
```
thhous-msft, line 62: Isn't UNSPEC generally allowed for remote address? Like if you don't care if its v4 or v6. Or is that already fixed up by the time it gets here?

maolson-msft: No, we wouldn't know what address to connect to if UNSPEC was passed

nibanks:
We've decided that we don't need this change at all. Closing.

PR #2160, opened 2021-11-18
Add portable certificate indication support to schannel

thhous-msft:
OpenSSL has portable certificate support. This makes schannel have that support, which makes it easier for users to switch between schannel and openssl, especially on windows.


src/platform/cert_capi.c
```diff
@@ -697,6 +697,177 @@ CxPlatCertParseChain(
...
+        Status = LastError;
+        QuicTraceEvent(
+            LibraryErrorStatus,
+            "[ lib] ERROR, %u, %s.",
+            LastError,
+            "CertSaveStore failed");
+        goto Exit;
+    }
+
+    Blob.pbData = CXPLAT_ALLOC_NONPAGED(Blob.cbData, QUIC_POOL_TLS_PFX);
```
anrossi, line 806: I think we should have a new pool tag here so we can distinguish between the two if a leak occurs.

thhous-msft: Its not actually possible (At least currently). The PFX code only exists in openssl, and this code only exists in schannel.

anrossi: Ah, right. This is fine then.

src/platform/cert_capi.c
```diff
@@ -697,6 +697,177 @@ CxPlatCertParseChain(
...
+    DWORD LastError;
+    CERT_CHAIN_PARA ChainPara;
+    CERT_ENHKEY_USAGE EnhKeyUsage;
+    CERT_USAGE_MATCH CertUsage;
+    PCCERT_CHAIN_CONTEXT ChainContext;
+    PCCERT_CONTEXT CertCtx = (PCCERT_CONTEXT)Certificate;
+    HCERTSTORE TempCertStore = NULL;
+    CERT_BLOB Blob = {0};
+
+    PortableCertificate->Buffer = CertCtx->pbCertEncoded;
```
anrossi, line 718: This is a little dangerous because if the `QUIC_CERTIFICATE` goes away before the caller is done with `PortableCertificate`, then this is invalid memory.
I know that's not how it's used today, but if it's used elsewhere in the future, this implementation detail will become a bug.

src/platform/cert_capi.c
```diff
@@ -697,6 +697,177 @@ CxPlatCertParseChain(
...
+        QuicTraceEvent(
+            AllocFailure,
+            "Allocation of '%s' failed. (%llu bytes)",
+            "PKCS7 data",
+            Blob.cbData);
+        Status = QUIC_STATUS_OUT_OF_MEMORY;
+        goto Exit;
+    }
+
+
```
nibanks, line 817: nit: remove extra new line

nibanks (approved):

src/platform/cert_capi.c
thhous-msft: I can use CertDuplicateCertificateContext to bump the ref count. I'll do that.

anrossi: you'll need to update `CxPlatFreePortableCertificate` to also take a `QUIC_CERTIFICATE` parameter do you can decrement the ref count, but that sounds like a good plan. The only alternative I can think of is an alloc+copy, which would waste memory.

thhous-msft: I made a wrapper struct to store the 3 items.

anrossi (approved):

PR #2165, opened 2021-11-19
Route resolution in CxPlatSendDataAlloc

maolson-msft:
Adds the local and next-hop link layer address to CXPLAT_ROUTE, and calls the necessary
iphlpapi functions to initialize these fields in the first send on a CXPLAT_ROUTE.

These additions make it unnecessary to hardcode the local and remote link layer address,
so the hardcoding logic is removed.


src/platform/datapath_raw.c
```diff
@@ -419,15 +419,122 @@ CxPlatSendDataIsFull(
     return TRUE;
 }
 
+_IRQL_requires_max_(DISPATCH_LEVEL)
+inline
+QUIC_STATUS
+CxPlatResolveRoute(
```
nibanks, line 425: Are `GetBestRoute2`, `GetIfEntry2`, and `GetIpNetEntry2` all individual syscalls to kernel? Which, if any, could theoretically block on network IO if the info isn't cached?

maolson-msft: I'm pretty sure the only call here that blocks on network IO is ResolveIpNetEntry2.

maolson-msft: And yes, each of these is a syscall.

src/platform/datapath_raw.c
```diff
@@ -419,15 +419,122 @@ CxPlatSendDataIsFull(
...
+        }
+    }
+    CXPLAT_FRE_ASSERT(IpnetRow.PhysicalAddressLength == sizeof(Route->NextHopLinkLayerAddress));
+    CxPlatCopyMemory(&Route->NextHopLinkLayerAddress, IpnetRow.PhysicalAddress, sizeof(Route->NextHopLinkLayerAddress));
+
+    Route->Resolved = TRUE;
+
+Done:
+
+    // TODO: convert NETIO_STATUS to QUIC_STATUS
```
maolson-msft, line 516: Any advice here?

thhous-msft: HRESULT_FROM_WIN32 should work. Since it looks like NETIO_STATUS is just a normal DWORD win32 status code.

src/platform/datapath_raw.c
nibanks: Let's please add comments indicating which functions may block.

src/platform/datapath_raw.c
```diff
@@ -423,7 +423,7 @@ _IRQL_requires_max_(DISPATCH_LEVEL)
 QUIC_STATUS
 CxPlatSocketSend(
     _In_ CXPLAT_SOCKET* Socket,
-    _In_ const CXPLAT_ROUTE* Route,
+    _Inout_ CXPLAT_ROUTE* Route,
```
nibanks, line 426: I've since realized that this is actually going to need to be passed in and use in `CxPlatSendDataAlloc`. Once we support both "normal" and "accelerated" sockets in parallel, we will need to use different send allocations depending on which socket we use. I'll make the change to move this logic to there instead.

nibanks (approved):
The latest works well for me in Azure VMs, Matt.

PR #2166, opened 2021-11-19
Require specific data to be received to shutdown secnetperf

thhous-msft:
Some devices broadcast on 9999, which previously would kill secnetperf

### DIFF
--- a/src/perf/lib/SecNetPerfMain.cpp
+++ b/src/perf/lib/SecNetPerfMain.cpp
@@ -272,13 +272,30 @@ QuicMainGetExtraData(
     return TestToRun->GetExtraData(Data, Length);
 }
 
+typedef struct ShutdownData {
+    unsigned long  Data1;
+    unsigned short Data2;
+    unsigned short Data3;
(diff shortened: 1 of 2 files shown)


src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -272,13 +272,30 @@ QuicMainGetExtraData(
     return TestToRun->GetExtraData(Data, Length);
 }
 
+typedef struct ShutdownData {
+    unsigned long  Data1;
+    unsigned short Data2;
+    unsigned short Data3;
+    unsigned char  Data4[ 8 ];
+} ShutdownData;
```
nibanks, line 280: You can't simple use `GUID`?

nibanks: I guess that's Windows specific...

thhous-msft: GUID doesn't exist on linux or mac.

nibanks (approved):

PR #2169, opened 2021-11-20
Remove '/msquic' from install path

ami-GS:
Previously Unix user had to do

1. set library path env variable
2. edit Makefile
3. write full path of header to include

This change solves
- 3 by default
- 1 and 2 if developer set -DCMAKE_INSTALL_PREFIX=/usr

Additionally removed useless cmake file installation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,13 +63,14 @@ elseif (UNIX)
 endif()
 message(STATUS "QUIC Platform: ${CX_PLATFORM}")
 
-if(WIN32)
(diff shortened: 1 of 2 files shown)


ThadHouse:
The cmake file isn't useless to install. That is so find_package works. Also I'd like to leave the default the way it was but make it configurable

thhous-msft:
#2175 was merged, which incorporates this and some other necessary fixes. So hopefully everything is fixed on your end. Let us know if you run into any other issues.

PR #2177, opened 2021-11-23
Fix tests on downlevel systems failing at end of test

thhous-msft:
wpr cancel was setting last exit code, which was causing the tests to return failure. Set last exit code to 0 after exiting wpr.


scripts/log.ps1
```diff
@@ -143,7 +143,9 @@ function Log-Cancel {
     if ($IsWindows) {
         try {
             wpr.exe -cancel -instancename $InstanceName 2>&1
+            $global:LASTEXITCODE = 0
```
nibanks, line 146: Why in both places, instead of just after the try/catch?

thhous-msft: Fixed.

nibanks (approved):

PR #2182, opened 2021-11-24
Finish up kernel mode app storage

### LINKED ISSUE #1416: Finish Windows Kernel Per-App Storage
The code path for per-app (i.e. custom `Path`) storage isn't fully wired up:

./src/platform/storage_winkernel.c
```c
    if (Path != NULL) {
        Status = CxPlatConvertUtf8ToUnicode(Path, &PathUnicode);
        if (QUIC_FAILED(Status)) {
            goto Exit;
        }

        InitializeObjectAttributes(
            &Attributes,
            PathUnicode,
            OBJ_CASE_INSENSITIVE | OBJ_KERNEL_HANDLE,
            NULL,
            NULL);

        Status = QUIC_STATUS_NOT_SUPPORTED; // TODO
        goto Exit;
```

This code just needs to be finished up.

thhous-msft:
Needs some testing to make sure logic is fully right, but should work.

Fixes #1416


nibanks (approved):
Have you tested it out? Easiest way would be to just set a registry key and grab the ETW logs.

nibanks (dismissed):

src/test/lib/ApiTest.cpp
```diff
@@ -2126,3 +2126,195 @@ QuicTestCredentialLoad(const QUIC_CREDENTIAL_CONFIG* Config)
...
+    }
+    RegCloseKey(Key);
+#else
+    TEST_FAILURE("Storage tests not supported on this platform");
+#endif
+
+    MsQuicSettings Settings;
+
+    //
+    // Global settings (future TODO, its too late at this point to create the Parameters key)
```
nibanks, line 2206: Let's check to see if the global key already existed, and only run these tests if it did. It will exist on WS2022+ builds.

nibanks: Oh, another option is that test.ps1 or run-gtest.ps1 can create the registry key before the test too. I think it'd be good to do both.

thhous-msft: I also need to bypass this test when registry write permission doesn't exist. But I'll work on that monday.

nibanks: Makes sense, but we currently run the tests as Admin in CI, so it's only a local problem.

src/test/bin/quic_gtest.cpp
```diff
@@ -1864,6 +1883,21 @@ INSTANTIATE_TEST_SUITE_P(
...
+    HKEY Key;
+    DWORD Result =
+        RegCreateKeyA(
+            HKEY_LOCAL_MACHINE,
+            "System\\CurrentControlSet\\Services\\MsQuic\\Parameters\\Apps\\StorageTest",
+            &Key);
+    CanRunStorageTests = Result == NO_ERROR;
+    RegCloseKey(Key);
+#endif
+
```
nibanks, line 1899: nit: remove extra ws

nibanks (approved):
Merge it if CI is green.

PR #2191, opened 2021-11-30
Compile an OpenSSL binary under VS 2022 in CI

thhous-msft:


.azure/templates/build-config-user.yml
```diff
@@ -32,6 +32,12 @@ jobs:
       version: 3.1.302
       installationPath: $(Agent.ToolsDirectory)/dotnet
 
+  - pwsh: |
+      Choco-Install -PackageName strawberryperl
+      Write-Host "##vso[task.prependpath]C:\Strawberry\perl\bin"
+    displayName: Install Perl
+    condition: and(eq('${{ parameters.image }}', 'windows-2022'), eq('${{ parameters.tls }}', 'openssl'))
```
nibanks, line 39: Can we scope this to only when necessary?

thhous-msft: It is. Thats what the condition does.

nibanks: Ah, duh. I was looking for it at the top for some reason.

nibanks (approved):

PR #2196, opened 2021-12-01
Support Multiple RSS Queues

nibanks:
Updates the XDP code to query the RSS queue count and match it, so that we don't miss received packets when multiple RSS queues are configured.

### DIFF
--- a/src/inc/quic_datapath.h
+++ b/src/inc/quic_datapath.h
@@ -152,6 +152,7 @@ typedef struct CXPLAT_ROUTE {
     uint8_t LocalLinkLayerAddress[6];
     uint8_t NextHopLinkLayerAddress[6];
     BOOLEAN Resolved; // When TRUE, LocalLinkLayerAddress and NextHopLinkLayerAddress are valid.
+    uint8_t Queue;
 
 } CXPLAT_ROUTE;
 
--- a/src/inc/quic_datapath.h
+++ b/src/inc/quic_datapath.h
@@ -152,7 +152,7 @@ typedef struct CXPLAT_ROUTE {
     uint8_t LocalLinkLayerAddress[6];
     uint8_t NextHopLinkLayerAddress[6];
     BOOLEAN Resolved; // When TRUE, LocalLinkLayerAddress and NextHopLinkLayerAddress are valid.
-    uint8_t Queue;
+    uint8_t QueueId;
 
 } CXPLAT_ROUTE;
 
--- a/src/platform/datapath_raw_xdp.c
+++ b/src/platform/datapath_raw_xdp.c
@@ -16,6 +16,7 @@ Abstract:
 #include "datapath_raw_xdp.c.clog.h"
 #endif
 
+#include <wbemidl.h>
 #include <afxdp_helper.h>
 #include <xdpapi.h>
 #include <stdio.h>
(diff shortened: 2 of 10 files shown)


src/inc/quic_datapath.h
```diff
@@ -152,6 +152,7 @@ typedef struct CXPLAT_ROUTE {
     uint8_t LocalLinkLayerAddress[6];
     uint8_t NextHopLinkLayerAddress[6];
     BOOLEAN Resolved; // When TRUE, LocalLinkLayerAddress and NextHopLinkLayerAddress are valid.
+    uint8_t Queue;
```
maolson-msft, line 155: Why does this need to be here?

maolson-msft: My main issue with this is that this struct is currently cleanly defined as "Structure to represent a network route", but this new field has nothing [modulo some philosophical arguments about the queue being part of the effective "route" of a packet] to do with that. Is there some more generic name for this struct that would be appropriate? Or, could we just plumb the queue ID as a parameter in the appropriate functions?

Also, based on names chosen elsewhere in this change, this name should be "QueueId" rather than "Queue".

src/platform/datapath_raw_xdp.c
```diff
@@ -111,6 +112,248 @@ CxPlatDataPathRecvDataToRecvPacket(
...
+    if (FAILED(hRes)) {
+        QuicTraceEvent(
+            LibraryErrorStatus,
+            "[ lib] ERROR, %u, %s.",
+            hRes,
+            "CoInitializeEx");
+        return hRes;
+    }
+
+    // Step 3: ---------------------------------------------------
```
csujedihy, line 162: Are we missing one step? Previous step is 2.

nibanks: I copied this code. 😄

csujedihy: I see. A comment just for numbering seems unnecessary.

src/platform/datapath_raw_xdp.c
```diff
@@ -111,6 +112,248 @@ CxPlatDataPathRecvDataToRecvPacket(
...
+            LibraryErrorStatus,
+            "[ lib] ERROR, %u, %s.",
+            ret,
+            "ConvertInterfaceIndexToLuid");
+        return HRESULT_FROM_WIN32(ret);
+    }
+
+    ret = ConvertInterfaceLuidToAlias(&if_luid, if_alias, RTL_NUMBER_OF(if_alias));
+    if (ret != NO_ERROR) {
+        QuicTraceEvent(
```
csujedihy, line 142: sidecar update?

nibanks: Done.

src/inc/quic_datapath.h
nibanks: I've updated the name. This is necessary so that we match RSS on the TX side.

src/platform/datapath_raw_xdp.c
nibanks: I fixed the comment.

nibanks: BTW, this code is all temporary. It will be updated once XDP RSS API is available.

csujedihy (approved):

maolson-msft (approved):

PR #2202, opened 2021-12-04
XDP automatically binds to all supported interfaces at startup

csujedihy:
1. Abstracted all interface related stuff into CXPLAT_INTERFACE, XDP_INTERFACE and DPDK_INTERFACE. Interfaces are chained by a linked list.
2. No support for dynamic NIC add/removal yet.
3. Tested with both XDP and DPDK.

For DPDK, rte_eth_dev_info_get doesn't give us a valid ifindex. So, the ifindex of the interface being bound is looked up by mac address via GetIfTable2.

### DIFF
--- a/src/inc/quic_datapath.h
+++ b/src/inc/quic_datapath.h
@@ -141,6 +141,21 @@ typedef struct CXPLAT_SEND_DATA CXPLAT_SEND_DATA;
 //
 typedef struct QUIC_BUFFER QUIC_BUFFER;
 
+typedef struct CXPLAT_INTERFACE {
+    CXPLAT_LIST_ENTRY Link;
+    uint32_t IfIndex;
+    struct {
+        struct {
+            BOOLEAN NetworkLayerXsum : 1;
+            BOOLEAN TransportLayerXsum : 1;
+        } Transmit;
+        struct {
+            BOOLEAN NetworkLayerXsum : 1;
+            BOOLEAN TransportLayerXsum : 1;
+        } Receive;
+    } OffloadStatus;
+} CXPLAT_INTERFACE;
+
 //
 // Structure to represent a network route.
 //
@@ -153,7 +168,7 @@ typedef struct CXPLAT_ROUTE {
     uint8_t NextHopLinkLayerAddress[6];
     BOOLEAN Resolved; // When TRUE, LocalLinkLayerAddress and NextHopLinkLayerAddress are valid.
     uint8_t QueueId;
-
+    CXPLAT_INTERFACE* Interface;
 } CXPLAT_ROUTE;
 
 //
--- a/src/inc/quic_datapath.h
+++ b/src/inc/quic_datapath.h
@@ -144,6 +144,7 @@ typedef struct QUIC_BUFFER QUIC_BUFFER;
 typedef struct CXPLAT_INTERFACE {
     CXPLAT_LIST_ENTRY Link;
     uint32_t IfIndex;
+    UCHAR PhysicalAddress[6];
     struct {
         struct {
             BOOLEAN NetworkLayerXsum : 1;
--- a/src/inc/quic_datapath.h
+++ b/src/inc/quic_datapath.h
@@ -141,21 +141,7 @@ typedef struct CXPLAT_SEND_DATA CXPLAT_SEND_DATA;
 //
 typedef struct QUIC_BUFFER QUIC_BUFFER;
 
-typedef struct CXPLAT_INTERFACE {
-    CXPLAT_LIST_ENTRY Link;
-    uint32_t IfIndex;
-    UCHAR PhysicalAddress[6];
-    struct {
-        struct {
-            BOOLEAN NetworkLayerXsum : 1;
-            BOOLEAN TransportLayerXsum : 1;
-        } Transmit;
-        struct {
-            BOOLEAN NetworkLayerXsum : 1;
-            BOOLEAN TransportLayerXsum : 1;
-        } Receive;
-    } OffloadStatus;
-} CXPLAT_INTERFACE;
+typedef struct CXPLAT_INTERFACE CXPLAT_INTERFACE;
 
 //
 // Structure to represent a network route.
--- a/src/platform/datapath_raw_socket.c
+++ b/src/platform/datapath_raw_socket.c
@@ -370,7 +370,7 @@ CxPlatResolveRoute(
     }
 
     if (Route->Interface == NULL) {
-        Status = ERROR_NOT_FOUND;
+        Status = QUIC_STATUS_NOT_FOUND;
         QuicTraceEvent(
             DatapathError,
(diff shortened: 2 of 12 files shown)


src/platform/datapath_raw_socket.c
```diff
@@ -356,8 +356,31 @@ CxPlatResolveRoute(
     CXPLAT_DBG_ASSERT(IpnetRow.PhysicalAddressLength == sizeof(Route->NextHopLinkLayerAddress));
     CxPlatCopyMemory(&Route->NextHopLinkLayerAddress, IpnetRow.PhysicalAddress, sizeof(Route->NextHopLinkLayerAddress));
 
-    Route->QueueId = 0; // TODO - Any way figure this out upfront?
 
+    //
+    // Find the interface that matches the route we just looked up.
+    //
+    CXPLAT_LIST_ENTRY* Entry = Socket->Datapath->Interfaces.Flink;
```
nibanks, line 363: nit: put this into the `for` loop since you don't need it outside the scope

src/platform/datapath_raw_socket.c
```diff
@@ -356,8 +356,31 @@ CxPlatResolveRoute(
...
+        if (Interface->IfIndex == IpforwardRow.InterfaceIndex) {
+            Route->Interface = Interface;
+            break;
+        }
+    }
+
+    if (Route->Interface == NULL) {
+        Status = ERROR_NOT_FOUND;
+        QuicTraceEvent(
+            DatapathErrorStatus,
```
nibanks, line 375: You can just use `DatapathError` and not log the not found. It's only necessary to log an error code when it's returned from a function you just called that failed. Here "no matching interface" is enough info.

src/platform/datapath_raw_xdp.c
```diff
@@ -428,40 +428,108 @@ CxPlatDpRawGetDapathSize(
...
 
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+CxPlatDpRawInterfaceUninitialize(
+    _Inout_ XDP_INTERFACE* Interface
+    )
+{
+
+    #pragma warning(push)
+    #pragma warning(disable:6001) // Using uninitialized memory
```
nibanks, line 439: What's this needed for?

src/platform/datapath_raw_xdp.c
```diff
@@ -428,40 +428,108 @@ CxPlatDpRawGetDapathSize(
     return sizeof(XDP_DATAPATH);
 }
 
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+CxPlatDpRawInterfaceUninitialize(
+    _Inout_ XDP_INTERFACE* Interface
+    )
+{
+
```
nibanks, line 437: nit: remove empty new line

src/platform/datapath_raw_xdp.c
```diff
@@ -428,40 +428,108 @@ CxPlatDpRawGetDapathSize(
...
+    Interface->OffloadStatus.Receive.TransportLayerXsum = Xdp->SkipXsum;
+    Interface->OffloadStatus.Transmit.NetworkLayerXsum = Xdp->SkipXsum;
+    Interface->OffloadStatus.Transmit.NetworkLayerXsum = Xdp->SkipXsum;
 
-    Xdp->Queues = CxPlatAlloc(Xdp->QueueCount * sizeof(*Xdp->Queues), QUEUE_TAG);
-    if (Xdp->Queues == NULL) {
+    Status = CxPlatGetInterfaceRssQueueCount(Interface->IfIndex, &Interface->QueueCount);
+    if (QUIC_FAILED(Status)) {
         QuicTraceEvent(
             AllocFailure,
```
nibanks, line 511: 1) It's the wrong error message type. 2) You don't need to log because the function itself already logs the point of failure.

src/platform/datapath_raw_xdp.c
```diff
@@ -690,6 +758,81 @@ CxPlatDpRawInitialize(
...
+            IfRow->OperStatus == IfOperStatusUp &&
+            IfRow->MediaType == NdisMedium802_3) {
+            XDP_INTERFACE* Interface = CxPlatAlloc(sizeof(XDP_INTERFACE), IF_TAG);
+            if (Interface == NULL) {
+                QuicTraceEvent(
+                    AllocFailure,
+                    "Allocation of '%s' failed. (%llu bytes)",
+                    "XDP interface",
+                    sizeof(*Interface));
+                continue;
```
nibanks, line 813: Probably worth making allocation failure fatal instead of continuing.

src/platform/datapath_raw_xdp.c
```diff
@@ -690,6 +758,81 @@ CxPlatDpRawInitialize(
...
+                QuicTraceEvent(
+                    LibraryErrorStatus,
+                    "[ lib] ERROR, %u, %s.",
+                    Status,
+                    "CxPlatDpRawInterfaceInitialize");
+                CxPlatFree(Interface, IF_TAG);
+                continue;
+            }
+
+            printf("Bound XDP to interface %u\n", IfRow->InterfaceIndex);
```
nibanks, line 831: For now, wrap any printfs you think we need to keep in `#if DEBUG`, otherwise remove them please.

src/inc/quic_datapath.h
```diff
@@ -141,6 +141,21 @@ typedef struct CXPLAT_SEND_DATA CXPLAT_SEND_DATA;
 //
 typedef struct QUIC_BUFFER QUIC_BUFFER;
 
+typedef struct CXPLAT_INTERFACE {
```
nibanks, line 144: Do we really need to define this for all consumers of `quic_datapath.h`? Maybe just have `typedef struct CXPLAT_INTERFACE CXPLAT_INTERFACE` and move the full definition to the raw header.

nibanks (commented):
Looks really good. Just a couple of minor comments mostly.

src/platform/datapath_raw_socket.c
csujedihy: Oh, I placed it outside so I don't need to break `for` loop into multiple lines, which seems ugly 😆

src/inc/quic_datapath.h
csujedihy: You are right. Just need a forward declaration here.

src/platform/datapath_raw_xdp.c
csujedihy: the compiler is confused and it thinks `Interface->Queues` wasn't initialized but we zero it before passing it to this function. I copied these pragma from the existing code. They were in CxPlatDpRawUninitialize.

src/platform/datapath_raw_xdp.c
csujedihy: done

nibanks (approved):
Looks good. Might need to iterate some more, in the future, but it's good as is for now.

Issue #1144, opened 2021-01-06
Properly handling connection shutdown is difficult to get right


ThadHouse:
Currently, we have some API issues around connection opening and closing, that we've ran into multiple times, and it would be nice to find a solution to. Take the following example.

```
ConnCallback(Conn, Context, Event) {
   if (Event->Type == QUIC_CONNECTION_TYPE_SHUTDOWN_COMPLETE) {
      MsQuic->ConnectionClose(Conn);
   }
}

// Somewhere not in the worker threads
MsQuic->ConnectionOpen(...) // Assume status is successful here
MsQuic->ConnectionStart(...) // Assume status is successful here
MsQuic->GetParam(Conn, ...) // Get some connection level parameter
```
The following code has a chance to crash, even though it seems fine. At a point during ConnectionStart (even before it returns) and GetParam, the shutdown on the connection can occur. This means GetParam would be called with an invalid handle. 

If we move the ConnectionClose from the event to some point not on the worker thread, we run into a synchronization issue in the other direction, where external code would be required to do cleanup of the handle. This means external code needs to keep a handle for all connections, and explicitly clean them up. We've also seen other issues related to this pop up with closing external code references. 

This is not an unsolvable problem, but is difficult and non obvious. A solution to this currently requires external code to make Close both in the event and externally work. Some code currently uses Ref Counting of the connection to solve this issue, but that is difficult to set up correctly. Others involves specifically checking on flags to see if its safe to close, but that is difficult as well.

ThadHouse:
Currently, I can think of a few solutions to this problem, but there are for sure many more.

1. Add a `QUIC_CONNECTION_TYPE_STARTED` Event that gets called when a connection is first started. This would be where users could do all their initialization code, knowing that references would not be deleted from under them. This is probably the least invasive change, as all old code would still work, and new code could just be cleaned up. As long as you only access the handle from in this event, it will not be deleted from under you.
2. Add a way for a user to increment and decrement the internal ref counts for objects. This is a lot more explicit, and we would need to decide if we want a bumped ref count coming out of ConnectionOpen, which would break all existing code. Lots of devs are used to this, as this is how COM works, but its fairly breaking. We might be able to do this with a parameter or something like that, to make it less breaking.
3. Punt on this issue in C, and in C++ wrappers add an external ref count to handle this. This is easily done, but likely not the scenario we would want to go in.

I'm sure there are more ideas as well, this is just a quick brain dump.

nibanks:
[...]

PR #2205, opened 2021-12-06
Expose external ref counting API for Connection and Stream handles

thhous-msft:
As discussed in #1144 handling connection (and by extension stream) shutdown is difficult for both simple and complex application layers. 

Many of the largest issues come during connection and stream creation. When you call Connection/StreamOpen, before the API even returns to the application, it is possible for the Shutdown Complete event for that new handle to be posted to the worker. The shutdown complete event can occur at any time after this too. At a high level, this means that Connection/Stream close cannot blindly be called during the Shutdown Complete event, and instead MUST synchronize with the application thread creating the objects. This also becomes an issue if any of the API's called from the app thread return an error, as sometimes these errors might result in Shutdown Complete and others might not.

Realistically, this means you MUST have some sort of reference count (or other synchronization method) between the worker thread and the application thread. For simple apps, this means external allocations, and also in C there isn't support in the standard library on all platforms for cross platform ref counting, which means you must either pull in an external library or use platform specific code for handling a ref count. For both simple and complex apps, you have to handle the ref count the right way. You have to know that the worker thread implicitly needs a ref count, and make sure you have an extra ref count before calling the Open functions.

We can solve these issues by exposing AddRef and Release functions from the MsQuic API. The semantics of this are pretty simple.

* You don't have to use the ref counting API if you don't want to. All old code will continue to work as is.
* If you want to use the ref counting capabilities, do not use ConnectionClose or StreamClose anymore.
* When you call ConnectionOpen or StreamOpen, the resulting handle will default to a ref count of 2. 1 of these ref counts is for the application, and 1 of them is for the worker. This means your application must call Release twice at minimum. Once when you're done using the handle in the application thread, and once when you're done in the worker thread (Likely Shutdown Complete). This follows COM rules, with the special addition that the worker thread owns 1 of the references by default.
* If you receive an error on the application thread, call shutdown before calling release. This will ensure that a shutdown complete event is sent. Without this, its possible to never get a shutdown complete event, and the connection/stream will leak. 
* QUIC_CONNECTION_EVENT_PEER_STREAM_STARTED or QUIC_LISTENER_EVENT_NEW_CONNECTION will result in the new handle having a ref count of 1, with the only owner being the worker thread. If you need to store this handle and use it on another thread, add a reference.
* When you call release, the result will tell you if the last reference was freed or not. If it was, you are safe to free any external data, as you will no longer get any events with that handle.
* AddRef and Release follow COM rules. You can read up on them for more information.

The sample and secnetperf have been updated to use this new style. secnetperf however could use more work, as its not as pragmatic as it should be.

I know this still needs API docs and lots of review, and might not even be accepted, but it is worth discussing.


src/core/api.h
```diff
@@ -253,6 +253,22 @@ MsQuicGetParam(
         void* Buffer
     );
 
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+QUIC_API
+MsQuicHandleAddRef(
```
nibanks, line 259: I'd prefer not to have `Handle` in the title. None of the other generic functions have it (`GetParam`, `SetParam`).

thhous-msft: Other option is I change the other one to MsQuicLibraryAddRef

src/perf/lib/PerfServer.cpp
```diff
@@ -134,8 +134,8 @@ PerfServer::ConnectionCallback(
             if (PrintStats) {
                 QuicPrintConnectionStatistics(MsQuic, ConnectionHandle);
             }
-            MsQuic->ConnectionClose(ConnectionHandle);
         }
+        MsQuic->Release(ConnectionHandle);
```
nibanks, line 138: Why did you move this?

thhous-msft: Because we always want to call this now. AppCloseInProgram can technically never happen now (Which means I need to move the rest out).

src/core/api.h
nibanks: Ah, I see. Ok. I'm fine with this then.

src/core/api.c
```diff
@@ -1578,6 +1578,94 @@ MsQuicGetParam(
...
+        InterlockedIncrement((volatile long*)&Connection->ExternalRefCount);
+    } else if (IS_STREAM_HANDLE(Handle)) {
+        #pragma prefast(suppress: __WARNING_25024, "Pointer cast already validated.")
+        QUIC_STREAM* Stream = (QUIC_STREAM*)Handle;
+
+        CXPLAT_TEL_ASSERT(!Stream->Flags.HandleClosed);
+        CXPLAT_TEL_ASSERT(!Stream->Flags.Freed);
+
+        InterlockedIncrement((volatile long*)&Stream->ExternalRefCount);
+    }
```
anrossi, line 1613: I think we should have a debug assert here so that callers don't use this API for handles that aren't support
```suggestion
    } else {
        CXPLAT_DBG_ASSERTMSG(FALSE, "Only supported for Connection and Stream handles!");
    }   
```

src/core/api.c
```diff
@@ -1578,6 +1578,98 @@ MsQuicGetParam(
...
+    } else if (IS_STREAM_HANDLE(Handle)) {
+        #pragma prefast(suppress: __WARNING_25024, "Pointer cast already validated.")
+        QUIC_STREAM* Stream = (QUIC_STREAM*)Handle;
+
+        CXPLAT_TEL_ASSERT(!Stream->Flags.HandleClosed);
+        CXPLAT_TEL_ASSERT(!Stream->Flags.Freed);
+
+        InterlockedIncrement((volatile long*)&Stream->ExternalRefCount);
+    } else {
+        CXPLAT_DBG_ASSERTMSG(FALSE, "Only supported for Connection and Stream handles!");
```
nibanks, line 1614: I actually think might be better as a FRE assert. Apps should not do this, or bad things might happen.

src/core/connection.c
```diff
@@ -109,6 +109,7 @@ QuicConnAlloc(
         Connection->Stats.CorrelationId);
 
     Connection->RefCount = 1;
+    Connection->ExternalRefCount = IsServer ? 1 : 2;
```
nibanks, line 112: I disagree with this model. An app that creates a new object should always get an object with a single ref. For instance, if I call `ConnectionOpen` followed by `Release` on it, that should be enough. If (and only if) I need to have additional threads, beyond the QUIC worker, then I call AddRef for each additional dependency.

thhous-msft: Then this is pointless. You HAVE to do this, otherwise you still have the race conditions described.

thhous-msft: The whole problem is the existing model basically being a ref count of 1. The APP then has to do the work to synchronize between the worker thread and itself. Thats no different then what we have currently, as the app would still have to manually handle that synchronization.

nibanks: It all depends on the application scenario/usage. If an app calls `ConnectionOpen` followed simply by `ConnectionStart` and then everything happens on the callback thread, they should be able to simply call `Release` once in the callback and be done.

If they want to continue to use the connection on the main thread after calling start, then they need to call `AddRef` before calling `ConnectionStart`.

I don't understand why that doesn't work, or is an unreasonable design.

thhous-msft: Because we've seen before you can get a ConnectionShutdown event before ConnectionOpen even returns to the app. And then if you call Release in that callback, you're getting a handle back to your app that is already freed.

thhous-msft: If you could guarantee no callbacks would ever be called until ConnectionStart has been called, yes you could start with a ref count of 1. But because you can't guarantee that, you have to start with a ref count of 2. Its just object ownership semantics. The same reason why every caller that currently calls us makes sure there is a ref count of 2 on their external ref count before calling ConnectionOpen. A ref count of 1 is fundamentally incorrect here. It doesn't matter that events are unlikely to happen before ConnectionStart. The fact that they can forces the semantics here. Its just a fact of how we designed the API and callback structure.

nibanks: We should definitely talk about this some more. Perhaps we can/should prevent the callbacks before `ConnectionStart`, as I suspect most existing code will break/crash if the connection shutdown complete before that. But that being said, that can only happen under certain (kind of invalid) app behavior of trying to shutdown the registration at the same time as creating a new connection. Let's talk some more.

Maximus-: Ideally ConnectionOpen/StreamOpen can *never* fail asynchronously; This starts to look more like ConnectionCreate/StreamCreate, since it will only create the object, and not do anything network side. The alternative is that anything that can invalidate asynchronously never gets returned from the caller, only given in a callback.

Anas198900 (approved):

Anas198900 (approved):

src/core/connection.c
ThadHouse: #2226 is solving the async case. Also open never hits the network anyway. It's basically equivalent to alloc. The only way for it to async fail was another thread shutting down the registration. Which we're solving in that other pr before this.

src/core/api.c
```diff
@@ -1615,6 +1615,102 @@ MsQuicGetParam(
     return Status;
 }
 
+_IRQL_requires_max_(PASSIVE_LEVEL)
```
nibanks, line 1618: I'm wondering if we can/should require passive. I can see instances where a dispatch thread might want to add a ref. Thoughts?

thhous-msft: AddRef we probably can call at dispatch, but not release.

src/core/api.c
```diff
@@ -1615,6 +1615,102 @@ MsQuicGetParam(
...
+    if (IS_CONN_HANDLE(Handle)) {
+        #pragma prefast(suppress: __WARNING_25024, "Pointer cast already validated.")
+        QUIC_CONNECTION* Connection = (QUIC_CONNECTION*)Handle;
+
+        QUIC_CONN_VERIFY(Connection, !Connection->State.HandleClosed);
+        QUIC_CONN_VERIFY(Connection, !Connection->State.Freed);
+
+        long RefCount = InterlockedDecrement((volatile long*)&Connection->ExternalRefCount);
+
+        if (RefCount <= 0) {
```
nibanks, line 1686: Catch double release calls. If RefCount goes negative, FRE_ASSERT

thhous-msft: We don't want to FRE_ASSERT here. We want to do the same DBG_ASSERTs changes that we did for close. The easiest way to do that is to just forward to close a 2nd time.

Maximus-: Why would we not want to FRE_ASSERT? This is essentially a double free?

thhous-msft: Its a double free that isn't actually free'd yet, so all the memory is perfectly valid. Its kind of an annoying side affect of how our shutdown model works.

https://github.com/microsoft/msquic/pull/2239

Its using the change in the above PR to make this work.

Maximus-: How do we know the object hasn't been free'd yet? I don't think there's any guarantee of that? Also calling Close twice could double-free other objects. Possibly not today, but in the future?

thhous-msft: We know the object hasn't actually been freed because by ref count rules the only way to get a reference is on the worker thread. And if you have a reference and are on the worker thread, it has not been freed. This is UB if the caller calls Release again from the non worker thread, but by then the ref count has been freed anyway, so its UB to even do the decrementing. Theres nothing to FRE_ASSERT that isn't UB in that case.

By just forwarding this to Close, we only have to ensure Close keeps the current semantics of not hard failing if the app makes a mistake. As long as we don't change Close to double free anything without checking if we're in the worker, we're good.

nibanks:
@thhous-msft I think we should just go ahead and close this PR. Thoughts?

PR #2219, opened 2021-12-09
Refactors XDP to Separate Interface Queues and Rules

nibanks:
The first step in supporting per-flow rules with XDP/DPDK layers; which will happen dynamically as a result of new sockets getting created by the QUIC layer.


src/platform/datapath_raw_xdp.c
```diff
@@ -56,10 +60,17 @@ typedef struct _XDP_QUEUE {
     CXPLAT_LIST_ENTRY TxQueue;
 } XDP_QUEUE;
 
+typedef struct XDP_FLOW {
```
csujedihy, line 63: I am little confused. What does a flow mean? Is it related to any design decisions we chose?

csujedihy: So, now a datapath can have multiple interfaces and each interface can have multiple flows and each flow can have multiple queues?

src/platform/datapath_raw_xdp.c
```diff
@@ -837,22 +846,32 @@ CxPlatDpRawInitialize(
...
-                        LibraryErrorStatus,
-                        "[ lib] ERROR, %u, %s.",
-                        Status,
-                        "CxPlatDpRawInterfaceInitialize");
                     CxPlatFree(Interface, IF_TAG);
                     continue;
                 }
+
+                // TODO - Don't hard code flows. Dynamically create them in response
+                // to sockets getting created. Will be done in follow up PR.
```
csujedihy, line 867: what type of sockets are you referring to? XSKs or CXPLAT_SOCKETs?

src/platform/datapath_raw_xdp.c
csujedihy: What is the purpose of this extra layer of abstration?

src/platform/datapath_raw_xdp.c
```diff
@@ -56,10 +60,17 @@ typedef struct _XDP_QUEUE {
     CXPLAT_LIST_ENTRY TxQueue;
 } XDP_QUEUE;
 
+typedef struct XDP_FLOW {
+    const XDP_INTERFACE* Interface;
+    CXPLAT_LIST_ENTRY Link;
+    XDP_QUIC_FLOW Descriptor;
+    XDP_QUEUE Queues[0];
```
csujedihy, line 67: what it this? It's not defined anywhere.

csujedihy: Btw, it doesn't seem to compile.

src/platform/datapath_raw_xdp.c
nibanks: Eventually, they will be equivalent, but right now, XSKs.

src/platform/datapath_raw_xdp.c
nibanks: Yes it does: https://dev.azure.com/ms/msquic/_build/results?buildId=249521&view=results

src/platform/datapath_raw_xdp.c
nibanks: Every new socket that creates will need to create a new queue + rule to receive only that traffic. See `XDP_QUIC_FLOW`.

csujedihy: There's only one hit for XDP_QUIC_FLOW, which is in the struct definition of XDP_FLOW.

Every socket means literally every socket? Like for both client and listener? CxPlatXdpWorkerThread loops over each flow, which is going to be a problem when the number of flows is huge.

src/platform/datapath_raw_xdp.c
csujedihy: I can't find the definition of XDP_QUIC_FLOW. Where is it?

nibanks: It's in the latest XDP header (program.h I beleive). You need to update your XDP to have it.

src/platform/datapath_raw_xdp.c
nibanks: For listeners, there's one socket per local UDP port used. For clients, yes, there is one per client connection.

src/platform/datapath_raw_xdp.c
csujedihy: ah I see.

src/platform/datapath_raw_xdp.c
nibanks: Yi and I discussed this a bit and decided on a slight change of direction. Please wait for the next commit.

csujedihy (approved):

PR #2224, opened 2021-12-10
Add Dynamic Rule Generation to XDP

nibanks:
Dynamically generate rules based on socket creation/deletion (XDP only).

### DIFF
--- a/src/platform/datapath_raw.c
+++ b/src/platform/datapath_raw.c
@@ -226,7 +226,7 @@ CxPlatSocketCreateUdp(
         goto Error;
     }
 
-    CxPlatDpRawOnSocketStateChange(*NewSocket, TRUE);
+    CxPlatDpRawPlumbRulesOnSocket(*NewSocket, TRUE);
 
 Error:
 
@@ -272,7 +272,7 @@ CxPlatSocketDelete(
     _In_ CXPLAT_SOCKET* Socket
     )
 {
-    CxPlatDpRawOnSocketStateChange(Socket, FALSE);
+    CxPlatDpRawPlumbRulesOnSocket(Socket, FALSE);
     CxPlatRemoveSocket(&Socket->Datapath->SocketPool, Socket);
     CxPlatRundownReleaseAndWait(&Socket->Rundown);
     CXPLAT_FREE(Socket, QUIC_POOL_SOCKET);
--- a/src/platform/datapath_raw.c
+++ b/src/platform/datapath_raw.c
@@ -226,6 +226,8 @@ CxPlatSocketCreateUdp(
         goto Error;
     }
 
+    CxPlatDpRawOnSocketStateChange(*NewSocket, TRUE);
+
 Error:
 
     if (QUIC_FAILED(Status)) {
@@ -270,6 +272,7 @@ CxPlatSocketDelete(
     _In_ CXPLAT_SOCKET* Socket
     )
 {
+    CxPlatDpRawOnSocketStateChange(Socket, FALSE);
     CxPlatRemoveSocket(&Socket->Datapath->SocketPool, Socket);
     CxPlatRundownReleaseAndWait(&Socket->Rundown);
     CXPLAT_FREE(Socket, QUIC_POOL_SOCKET);
--- a/src/platform/datapath_raw_xdp.c
+++ b/src/platform/datapath_raw_xdp.c
@@ -875,6 +875,20 @@ CxPlatDpRawInterfaceRemoveRule(
             if (Rule->Pattern.Port != Interface->Rules[i].Pattern.Port) {
                 continue;
             }
+        } else if (Rule->Match == XDP_MATCH_IPV4_UDP_TUPLE) {
+            if (Rule->Pattern.Tuple.DestinationPort != Interface->Rules[i].Pattern.Tuple.DestinationPort ||
+                Rule->Pattern.Tuple.SourcePort != Interface->Rules[i].Pattern.Tuple.SourcePort ||
+                memcmp(&Rule->Pattern.Tuple.DestinationAddress.Ipv4, &Interface->Rules[i].Pattern.Tuple.DestinationAddress.Ipv4, sizeof(IN_ADDR)) != 0 ||
+                memcmp(&Rule->Pattern.Tuple.SourceAddress.Ipv4, &Interface->Rules[i].Pattern.Tuple.SourceAddress.Ipv4, sizeof(IN_ADDR)) != 0) {
+                continue;
+            }
+        } else if (Rule->Match == XDP_MATCH_IPV6_UDP_TUPLE) {
+            if (Rule->Pattern.Tuple.DestinationPort != Interface->Rules[i].Pattern.Tuple.DestinationPort ||
+                Rule->Pattern.Tuple.SourcePort != Interface->Rules[i].Pattern.Tuple.SourcePort ||
+                memcmp(&Rule->Pattern.Tuple.DestinationAddress.Ipv6, &Interface->Rules[i].Pattern.Tuple.DestinationAddress.Ipv6, sizeof(IN6_ADDR)) != 0 ||
+                memcmp(&Rule->Pattern.Tuple.SourceAddress.Ipv6, &Interface->Rules[i].Pattern.Tuple.SourceAddress.Ipv6, sizeof(IN6_ADDR)) != 0) {
+                continue;
+            }
         } else {
             CXPLAT_FRE_ASSERT(FALSE); // Should not be possible!
         }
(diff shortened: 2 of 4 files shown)


src/platform/datapath_raw_xdp.c
```diff
@@ -957,6 +1057,61 @@ CxPlatDpRawUninitialize(
...
+            .Action = XDP_PROGRAM_ACTION_REDIRECT,
+            .Redirect.TargetType = XDP_REDIRECT_TARGET_TYPE_XSK,
+            .Redirect.Target = NULL,
+        };
+
+        CXPLAT_LIST_ENTRY* Entry;
+        for (Entry = Xdp->Interfaces.Flink; Entry != &Xdp->Interfaces; Entry = Entry->Flink) {
+            XDP_INTERFACE* Interface = CONTAINING_RECORD(Entry, XDP_INTERFACE, Link);
+            if (IsCreated) {
+                CxPlatDpRawInterfaceAddRule(Interface, &Rule);
```
maolson-msft, line 1082: Is this guaranteed to succeed?

src/platform/datapath_raw_xdp.c
```diff
@@ -957,6 +1057,61 @@ CxPlatDpRawUninitialize(
...
+    _In_ CXPLAT_SOCKET* Socket,
+    _In_ BOOLEAN IsCreated
+    )
+{
+    XDP_DATAPATH* Xdp = (XDP_DATAPATH*)Socket->Datapath;
+
+    if (Socket->Wildcard) {
+        const XDP_RULE Rule = {
+            .Match = XDP_MATCH_UDP_DST,
+            .Pattern.Port = Socket->LocalAddress.Ipv4.sin_port, // TODO - Check on on byte order
```
maolson-msft, line 1072: Looks like XDP compares this to the field in the actual packet header, i.e. expects the port in network order, so (assuming this is a normal sockaddr with port in network order) I think this is correct.

src/platform/datapath_raw_xdp.c
```diff
@@ -788,6 +803,93 @@ CxPlatDpRawInterfaceUpdateRules(
...
+    CxPlatLockRelease(&Interface->RuleLock);
+
+#pragma warning(pop)
+}
+
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+CxPlatDpRawInterfaceRemoveRule(
+    _In_ XDP_INTERFACE* Interface,
+    _In_ const XDP_RULE* NewRule
```
csujedihy, line 864: Shouldn't be called new rule

src/platform/datapath_raw_xdp.c
```diff
@@ -788,6 +803,93 @@ CxPlatDpRawInterfaceUpdateRules(
...
+            "[ lib] ERROR, %s.",
+            "No more room for rules");
+        CxPlatLockRelease(&Interface->RuleLock);
+        return;
+    }
+
+    const size_t OldSize = sizeof(XDP_RULE) * (size_t)Interface->RuleCount;
+    const size_t NewSize = sizeof(XDP_RULE) * ((size_t)Interface->RuleCount + 1);
+
+    XDP_RULE* NewRules = CxPlatAlloc(NewSize, RULE_TAG);
```
maolson-msft, line 831: Why does QUIC need to maintain a list of rules? It looks like you're only plumbing one rule per queue, so why can't you just call XdpCreateProgram and stash the returned program handle in the XDP_QUEUE, and then close that handle when the queue is deleted (or when a new rule is created)?

nibanks: It's appending the new rule to the array.

maolson-msft: I'm asking why QUIC needs to maintain this array.

src/platform/datapath_raw_xdp.c
```diff
@@ -957,6 +1057,61 @@ CxPlatDpRawUninitialize(
...
+
+        CXPLAT_LIST_ENTRY* Entry;
+        for (Entry = Xdp->Interfaces.Flink; Entry != &Xdp->Interfaces; Entry = Entry->Flink) {
+            XDP_INTERFACE* Interface = CONTAINING_RECORD(Entry, XDP_INTERFACE, Link);
+            if (IsCreated) {
+                CxPlatDpRawInterfaceAddRule(Interface, &Rule);
+            } else {
+                CxPlatDpRawInterfaceRemoveRule(Interface, &Rule);
+            }
+        }
```
csujedihy, line 1111: nit: you can move this part out of the IF ELSE block to reduce code duplication.

```
        CXPLAT_LIST_ENTRY* Entry;
        for (Entry = Xdp->Interfaces.Flink; Entry != &Xdp->Interfaces; Entry = Entry->Flink) {
            XDP_INTERFACE* Interface = CONTAINING_RECORD(Entry, XDP_INTERFACE, Link);
            if (IsCreated) {
                CxPlatDpRawInterfaceAddRule(Interface, &Rule);
            } else {
                CxPlatDpRawInterfaceRemoveRule(Interface, &Rule);
            }
        }
```

src/platform/datapath_raw.c
```diff
@@ -270,6 +272,7 @@ CxPlatSocketDelete(
     _In_ CXPLAT_SOCKET* Socket
     )
 {
+    CxPlatDpRawOnSocketStateChange(Socket, FALSE);
```
csujedihy, line 275: nit: why don't we call it CxPlatDpRawPlumbRulesOnSocket. CxPlatDpRawOnSocketStateChange seems a little weird to me. I can't easily tell what it does by just looking at the name.

src/platform/datapath_raw_xdp.c
nibanks: Because XDP doesn't support the notion of "add rule" or "remove rule". We have to plumb them all down together.

src/platform/datapath_raw_xdp.c
nibanks: It's best effort.

src/platform/datapath_raw_xdp.c
nibanks: Yes, we could, but this is temporary. It will (hopefully soon) be specific to connected sockets.

src/platform/datapath_raw.c
nibanks: Ok. I can change the name.

src/platform/datapath_raw_xdp.c
csujedihy: >>It looks like you're only plumbing one rule per queue
One rule per socket from my understanding. There could be multiple rules on a queue.

nibanks: Correct. Each socket has 1 rule, but they all apply to the same queue(s).

maolson-msft: We need to have a brownbag on the object model here.

src/platform/datapath_raw_xdp.c
```diff
@@ -788,6 +803,93 @@ CxPlatDpRawInterfaceUpdateRules(
...
+    CxPlatLockAcquire(&Interface->RuleLock);
+    // TODO - Don't always allocate a new array?
+
+    if (Interface->RuleCount + 1 == 0) {
+        QuicTraceEvent(
+            LibraryError,
+            "[ lib] ERROR, %s.",
+            "No more room for rules");
+        CxPlatLockRelease(&Interface->RuleLock);
+        return;
```
csujedihy, line 825: >return; [](http://example.com/codeflow?start=8&length=7)

 a

src/platform/datapath_raw_xdp.c
maolson-msft: But if this fails, won't that mean no connectivity? We ought to fail whatever QUIC API triggered this in that case.

src/platform/datapath_raw_xdp.c
```diff
@@ -788,6 +803,93 @@ CxPlatDpRawInterfaceUpdateRules(
...
+
+    for (uint8_t i = 0; i < Interface->RuleCount; i++) {
+        if (Interface->Rules[i].Match != Rule->Match) {
+            continue;
+        }
+
+        if (Rule->Match == XDP_MATCH_UDP_DST) {
+            if (Rule->Pattern.Port != Interface->Rules[i].Pattern.Port) {
+                continue;
+            }
```
csujedihy, line 877: This is probably not for long term but I believe we can bring down the time complexity of add/remove to O(1) or amortized O(1).

nibanks: I'm not worried too much about perf here right now.

src/platform/datapath_raw_xdp.c
nibanks: I just have to figure out clean up then, which might be really complicated. Let me look into it.

nibanks: @maolson-msft the problem comes down to `CxPlatDpRawInterfaceUpdateRules`. It applies the rules to all queues. What am I supposed to do if it fails on the 2nd+ queue (i.e. passed on at least one)?

nibanks: @maolson-msft & @csujedihy how about I just add a TODO to that function for now to figure out how to handle failure? Maybe we work with XDP to improve this so that we can update the rules for multiple queues as an atomic operation. That'd make this a lot easier.

csujedihy: Sounds good to me

csujedihy (approved):

PR #2226, opened 2021-12-13
Make RegistrationShutdown semantics easier to understand

### LINKED ISSUE #2222: Handle Registration Shutdown race with ConnectionOpen

If an app is creating a connection, while another thread is calling registration shutdown, its possible for the handle returned from ConnectionOpen to already have been closed. 

### Proposed solution

Delay the registration shutdown until the connection is started.

### Additional context

_No response_

thhous-msft:
Closes #2222 

Currently, the semantics of RegistrationShutdown are confusing. Its only affect connections that are currently created and registered. This means connections in the process of being created, or connections created after registration shutdown are not taken into consideration. Additionally, this can cause a shutdown event before a connection is returned to a user, which makes using the API much more complicated then it should be.

This change does 3 things.
1. After it is called, it doesn't allow any more connections to be created on the registration.
2. Any connections opened but not started will not receive a conn shutdown event from the registration shutdown event until they are started. This way the user knows no events on the connection will be called unless they either start the conn or explicitly shut it down or close it themselves. 
3. If an open connection is started after registration shutdown, it will immediately shut down with the error code given by registration shutdown.

Some tests are added as well to verify this behavior. More are needed, but this is a start.


src/core/connection.c
```diff
@@ -542,16 +558,28 @@ QuicConnRegister(
...
 
-    QuicTraceEvent(
-        ConnRegistered,
-        "[conn][%p] Registered with %p",
-        Connection,
-        Registration);
+    if (RegistrationShuttingDown) {
+        Connection->State.Registered = FALSE;
+        Connection->Registration = NULL;
+        CxPlatRundownRelease(&Registration->Rundown);
```
nibanks, line 573: I think we can no longer assume success for `CxPlatRundownAcquire(&Registration->Rundown)` above.

thhous-msft: True. Fixed.

src/test/bin/quic_gtest.cpp
```diff
@@ -214,6 +214,42 @@ TEST(ParameterValidation, ValidateConnection) {
     }
 }
 
+TEST(OwnershipValidation, RegistrationShutdownBeforeConnOpen) {
+    TestLogger Logger("RegistrationShutdownBeforeConnOpen");
+    if (TestingKernelMode) {
+        //ASSERT_TRUE(DriverClient.Run(IOCTL_QUIC_RUN_VALIDATE_API));
```
nibanks, line 220: You plan on doing kernel too?

thhous-msft: Yes. I'm working on a few more tests too.

src/core/registration.c
```diff
@@ -249,6 +252,10 @@ MsQuicRegistrationShutdown(
 
         CxPlatDispatchLockAcquire(&Registration->ConnectionLock);
```
nibanks, line 253: Maybe only do this if `ShuttingDown` isn't already `TRUE`?

src/test/lib/OwnershipTest.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+    HQUIC Connection = nullptr;
+    QUIC_STATUS Status =
+        MsQuic->ConnectionOpen(
+            Registration,
+            OwnershipConnCallback,
+            nullptr,
+            &Connection);
+
+    TEST_QUIC_STATUS(QUIC_STATUS_OUT_OF_MEMORY, Status);
```
nibanks, line 70: I'm not a fan that we get OUT_OF_MEMORY instead of something like INVALID_STATE. I know it would be a (slightly) bigger change, but it's probably worth returning a better error code.

src/test/lib/OwnershipTest.cpp
```diff
@@ -0,0 +1,235 @@
...
+        MsQuic->GetParam(
+            Conn.Connection,
+            QUIC_PARAM_LEVEL_CONNECTION,
+            QUIC_PARAM_CONN_STATISTICS,
+            &StatsSize,
+            &Stats);
+    TEST_QUIC_STATUS(QUIC_STATUS_SUCCESS, Status);
+    TEST_EQUAL(0, State.ShutdownCount);
+
+    Status =
```
nibanks, line 142: Shouldn't we verify this status too?

Anas198900 (approved):

src/core/api.c
```diff
@@ -61,8 +61,12 @@ MsQuicConnectionOpen(
 #pragma prefast(suppress: __WARNING_25024, "Pointer cast already validated.")
```
Anas198900, line 61: 0003000000FC3655

Anas198900: > 0003000000FC3655

Anas198900 (approved):

src/test/lib/OwnershipTest.cpp
thhous-msft: Fixed

src/test/lib/OwnershipTest.cpp
thhous-msft: I agree. I changed it up, it wasn't too bad.

src/core/registration.c
thhous-msft: Probably. I switched it. Should I dbg assert if its already true?

nibanks: No. Generally, we allow shutdown functions to be called multiple times, with them being no-op additional calls.

nibanks (approved):

PR #2235, opened 2021-12-17
Official Support for TLS Secrets

nibanks:
Remove all the compile time logic around TLS secret support and make it an official (not private) feature/parameter.


src/inc/msquicp.h
```diff
@@ -93,29 +93,8 @@ typedef struct QUIC_PRIVATE_TRANSPORT_PARAMETER {
...
-        uint8_t ServerTrafficSecret0 : 1;
-    } IsSet;
-    uint8_t ClientRandom[32];
-    uint8_t ClientEarlyTrafficSecret[CXPLAT_TLS_SECRETS_MAX_SECRET_LEN];
-    uint8_t ClientHandshakeTrafficSecret[CXPLAT_TLS_SECRETS_MAX_SECRET_LEN];
-    uint8_t ServerHandshakeTrafficSecret[CXPLAT_TLS_SECRETS_MAX_SECRET_LEN];
-    uint8_t ClientTrafficSecret0[CXPLAT_TLS_SECRETS_MAX_SECRET_LEN];
-    uint8_t ServerTrafficSecret0[CXPLAT_TLS_SECRETS_MAX_SECRET_LEN];
-} CXPLAT_TLS_SECRETS;
+#define CXPLAT_TLS_SECRETS_MAX_SECRET_LEN QUIC_TLS_SECRETS_MAX_SECRET_LEN
```
thhous-msft, line 96: Do we still need these? We're already binary breaking consumers by changing the param id, and this file isn't a public api file anyway.

nibanks: We don't strictly need them. This was just to make sure anything that was building against these continued to compile. I'm fine removing them if you think we should.

thhous-msft: I don't like doing a binary breaking change without also doing a source breaking change. Makes it much easier from a debugging standpoint to figure out why something is broken, as a program rebuild will tell you something changed. If you want to keep this, then we need the existing param id to still work at a binary level.

nibanks: Removed.

thhous-msft (approved):

PR #2239, opened 2021-12-20
Handle app calling Close from worker thread after calling from app thread

### LINKED ISSUE #2221: Handle recursive close for streams and connections

If an app layer calls close on a connection or stream, and it then calls close from shutdown complete, this results in a double close. This is something that is handlable from our end, as the actual handle won't be freed.

### Proposed solution

We should make a tel or verifier assert if this case is detected, and just not do anything.

### Additional context

_No response_

thhous-msft:
Closes #2221 

Currently, if an app calls Connection/StreamClose, and this results in a ShutdownComplete, and in that ShutdownComplete the user calls close again, the object is double freed. This is 100% an app bug, and we have an `AppCloseInProgress` flag the app can check to ensure they don't do this. However, as the handle is still alive, its valid for us to detect this case, and actually respond to it. Many apps that are written (Including many of ours) have this bug when they are first written (usually in error cases) where its not caught until much later down the line. By handling this case, we remove an entire set of bugs from the API, without compromising anything internally.

### DIFF
--- a/src/core/api.c
+++ b/src/core/api.c
@@ -111,8 +111,9 @@ MsQuicConnectionClose(
 #pragma prefast(suppress: __WARNING_25024, "Pointer cast already validated.")
     Connection = (QUIC_CONNECTION*)Handle;
 
-    BOOLEAN IsWorkerThread = Connection->WorkerThreadID == CxPlatCurThreadID();
     CXPLAT_TEL_ASSERT(!Connection->State.Freed);
+    QUIC_CONN_VERIFY(Connection, !Connection->State.HandleClosed);
+    BOOLEAN IsWorkerThread = Connection->WorkerThreadID == CxPlatCurThreadID();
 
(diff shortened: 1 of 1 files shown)


src/core/api.c
```diff
@@ -111,8 +111,20 @@ MsQuicConnectionClose(
...
+        //
+        // Close being called from worker thread after being closed by app
+        // thread. This is an app programming bug, and they should be checking
+        // the AppCloseInProgress flag, but as the handle is stil valid here
+        // we can just make this a no-op.
+        //
+        goto Error;
+    }
+
     QUIC_CONN_VERIFY(Connection, !Connection->State.HandleClosed);
```
nibanks, line 127: I at least want the CONN_VERIFY first. While we might be able to ignore this in PROD, when they have app verifier enabled, I want to tell them they did bad.

thhous-msft: Ok. Done.

nibanks (approved):

PR #2242, opened 2021-12-21
Allow 0-RTT packets not in a retry scenario to start a connection

thhous-msft:
0-RTT packets received out of order from initial packets should be considered valid packets and read. This is a behavior change from currently, where only initial packets can create a connection, and 0-RTT packets received out of order would be dropped. 

In a retry scenario, its possible for an initial to create the connection, but a 0-RTT packet to actually be the first one queued. If this occurs, its possible for the original dest cid to be overwritten. In dbg this would have asserted, but we now handle this case correctly.

### DIFF
--- a/src/core/binding.c
+++ b/src/core/binding.c
@@ -1470,6 +1470,7 @@ QuicBindingDeliverDatagrams(
 
         CXPLAT_DBG_ASSERT(QuicIsVersionSupported(Packet->Invariant->LONG_HDR.Version));
 
+        BOOLEAN IsInitial = FALSE;
         //
         // Only Initial (version specific) packets are processed from here on.
         //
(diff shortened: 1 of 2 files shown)


src/core/binding.c
```diff
@@ -1506,10 +1508,13 @@ QuicBindingDeliverDatagrams(
...
-            return
-                QuicBindingQueueStatelessOperation(
-                    Binding, QUIC_OPER_TYPE_RETRY, DatagramChain);
-
+            if (IsInitial) {
+                return
+                    QuicBindingQueueStatelessOperation(
+                        Binding, QUIC_OPER_TYPE_RETRY, DatagramChain);
+            }
+            QuicPacketLogDrop(Binding, Packet, "Reject non-initial packet during retry.");
```
nibanks, line 1516:
```suggestion
            QuicPacketLogDrop(Binding, Packet, "Non-initial packet during retry.");
```
Changed my mind on the text.

nibanks (approved):

PR #2245, opened 2021-12-22
Revert "Update to OpenSSL_1_1_1m+quic (#2229)"

thhous-msft:
This reverts commit 8894bc6f32927cbefe0b94fd96ccba673b3b1109.

This change seems to be causing a spike in test failures. PRing to verify it is this change. The failure results in no dumps.


thhous-msft:
Found the actual issue we assumed was caused by this change. Turns out to not be the case. Closing.

PR #2253, opened 2021-12-28
Kill drivers on test machines before checkout

thhous-msft:
Assuming the drivers are not hung, this should allow the checkout step to complete


thhous-msft:
Nevermind, this still happens too late.

nibanks (approved):

PR #2260, opened 2021-12-30
update msquic to clog 0.4.0

chgray:


scripts/prepare-machine.ps1
```diff
@@ -117,27 +114,10 @@ if ($InitSubmodules) {
...
-        }
-        $err = $_
-        $MessagesAtEnd.Add("$ToolName could not be installed. Building with logs will not work")
-        $MessagesAtEnd.Add($err.ToString())
-    }
-}
+
+    Write-Host "Initializing CLOG submodule"
+    git submodule init submodules/clog
+    git submodule update
```
nibanks, line 120:
```suggestion
Write-Host "Initializing CLOG submodule"
git submodule init submodules/clog
git submodule update
```

scripts/prepare-machine.ps1
```diff
@@ -381,8 +361,7 @@ if ($IsWindows) {
             Write-Host "[$(Get-Date)] Setting core dump pattern..."
             sudo sh -c "echo -n '%e.%p.%t.core' > /proc/sys/kernel/core_pattern"
             #sudo cat /proc/sys/kernel/core_pattern
-
-            Install-ClogTool "Microsoft.Logging.CLOG2Text.Lttng"
+            Install-ClogTool
```
nibanks, line 364: Where's this coming from now that you've deleted it above?

scripts/update-sidecar.ps1
```diff
@@ -19,44 +19,49 @@ $RootDir = Split-Path $PSScriptRoot -Parent
...
-Remove-Item $Sidecar -Force -ErrorAction Ignore | Out-Null
+# Build CLOG, placing results into the CLOG directory under our build directory
+dotnet publish ../submodules/clog/src/clog -o ${ClogDir} -f net5.0
 
-clog --installDirectory (Join-Path $OutputDir common)
+#
+# You may be tempted to delete the sidecar - DO NOT DO THIS - the sidecare
+#     exists for several purposes - one of those is to verify signatures of trace calls have
+#     not changed.  If you delete the sidecar, you'll miss errors that may save you broken contracts
+#     that occur when you decode
```
nibanks, line 50:
```suggestion
# You may be tempted to delete the sidecar - DO NOT DO THIS - the sidecare
#  exists for several purposes - one of those is to verify signatures of trace calls have
#  not changed.  If you delete the sidecar, you'll miss errors that may save you broken contracts
#  that occur when you decode
```
I agree with this so long as the following are not considering changes:

1. Moving an event around, relative to other events; including adding a new event before an existing one or moving between files.
2. Changing the human readable text. We tweak strings occasionally and want to replace existing event definitions; not add new ones.

src/core/packet.c
```diff
@@ -595,12 +595,12 @@ QuicPacketLogHeader(
...
             QuicTraceLogVerbose(
                 LogPacketVersionNegotiation,
-                "[%c][%cX][-] VerNeg DestCid:%s SrcCid:%s (Payload %lu bytes)",
+                "[%c][%cX][-] VerNeg DestCid:%s SrcCid:%s (Payload %u bytes)",
                 PtkConnPre(Connection),
                 (uint8_t)PktRxPre(Rx),
                 QuicCidBufToStr(DestCid, DestCidLen).Buffer,
                 QuicCidBufToStr(SourceCid, SourceCidLen).Buffer,
-                (uint16_t)(PacketLength - Offset));
+                (uint32_t)(PacketLength - Offset));
```
nibanks, line 603: Let's just replace `Payload %lu bytes` with `Payload %hu bytes` and keep it as a `uint16_t`.

src/core/packet.c
```diff
@@ -621,7 +621,7 @@ QuicPacketLogHeader(
                 (const QUIC_LONG_HEADER_V1 * const)Packet;
 
             QUIC_VAR_INT TokenLength;
-            QUIC_VAR_INT Length;
+            QUIC_VAR_INT Length = 0;
```
nibanks, line 624: Why?

src/perf/bin/drvmain.cpp
```diff
@@ -315,11 +315,11 @@ SecNetPerfCtlInitialize(
...
         goto Error;
     }
-
-    QuicTraceLogVerbose(
+/*
+    uicTraceLogVerbose(
         PerfControlInitialized,
         "[perf] Control interface initialized with %.*S", DeviceName.Length, DeviceName.Buffer);
-
+*/
```
nibanks, line 322: Why does this no longer work?

src/platform/cert_capi.c
```diff
@@ -1039,14 +1039,14 @@ CxPlatCertVerifyCertChainPolicy(
...
-    QuicTraceLogInfo(
+/*
+    uicTraceLogInfo(
         CertCapiVerifiedChain,
         "CertVerifyChain: %S 0x%x, result=0x%x",
         ServerName,
         CredFlags,
         Status);
-
+*/
```
nibanks, line 1049: No longer works?

src/tools/etw/main.c
```diff
@@ -152,7 +152,7 @@ void WINAPI EventCallback(_In_ PEVENT_RECORD ev)
     }
 
     if (TraceEvent) {
-        QuicTraceEvent(ev, ObjectId, InitialTimestamp);
+        //uicTraceEvent(ev, ObjectId, InitialTimestamp);
```
nibanks, line 155: Revert this and exclude clog from processing this file

scripts/prepare-machine.ps1
chgray: I'm debating the following 

a) change my position on checking in code;  due to tool fragmentation in Linux I'm actually starting to like checking in code :)

b) if you want to build the checked in files (clog) you install dotnet, enlist in clog, and build it  - the argument here is that dotnet will likely do a better job with installers than I will - might as well support them

scripts/update-sidecar.ps1
chgray: I like;  I'm required to disagree with #2, but that's ivory tower.  I'd argue you should create a new event if you're changing the structure of the event but at the end of the day thats your call

src/core/packet.c
chgray: I got a compiler error in a different branch actually;  you might review the code in this area?    I'm pretty confident it was linux (not mac/windows) - but I dont remember the compiler/distro

scripts/update-sidecar.ps1
nibanks: Yeah. Makes sense, but practically, I think we should just consider the binary encoding the "signature", not the string used to display that data. In other words, we should just ensure tools that decode the binary data continue to work.

scripts/clog.inputs
```diff
@@ -0,0 +1,124 @@
+../src/platform/selfsign_openssl.c
```
nibanks, line 1: I'm not a fan of requiring this file. It's yet another file we have to maintain.

scripts/prepare-machine.ps1
```diff
@@ -117,28 +114,6 @@ if ($InitSubmodules) {
     }
 }
 
-function Install-ClogTool {
```
nibanks, line 120: You'll notice the `if ($InitSubmodules)` section above. It will need to be updated to `init` the clog submodule too.

scripts/prepare-machine.ps1
nibanks: That all sounds reasonable. Checked in files for those who just want to build the code. Init the submodule and build clog for those who need to use it to update the checked in files.

scripts/update-sidecar.ps1
```diff
@@ -15,48 +15,57 @@ class SimpleStringComparer:Collections.Generic.IComparer[string] {
...
+# Change directory to the same directory as this script;  storing our original directory for later
+$OrigDir = Get-Location
+Set-Location $PSScriptRoot
 $RootDir = Split-Path $PSScriptRoot -Parent
 $SrcDir = Join-Path $RootDir "src"
 
 $OutputDir = Join-Path $RootDir "src" "generated"
+
+# Remove the linux directories - so that we delete files that have been abandoned
 if (Test-Path $OutputDir) {
```
nibanks, line 27: Is this outer `Test-Path` even necessary any more?

scripts/update-sidecar.ps1
```diff
@@ -15,48 +15,57 @@ class SimpleStringComparer:Collections.Generic.IComparer[string] {
...
+# Generate code for all different permutations we need
+Invoke-Expression "${ClogDir}/clog -p windows --scopePrefix quic.clog -s $Sidecar -c $ConfigFile --outputDirectory $TmpOutputDir --inputFiles $allFiles"
+Invoke-Expression "${ClogDir}/clog -p windows_kernel --scopePrefix quic.clog -s $Sidecar -c $ConfigFile --outputDirectory $TmpOutputDir --inputFiles $allFiles"
+Invoke-Expression "${ClogDir}/clog -p stubs --scopePrefix quic.clog -s $Sidecar -c $ConfigFile --outputDirectory $TmpOutputDir --inputFiles $allFiles"
+Invoke-Expression "${ClogDir}/clog -p linux --dynamicTracepointProvider --scopePrefix quic.clog -s $Sidecar -c $ConfigFile --outputDirectory (Join-Path $OutputDir linux) --inputFiles $allFiles"
+Invoke-Expression "${ClogDir}/clog -p macos --scopePrefix quic.clog -s $Sidecar -c $ConfigFile --outputDirectory $TmpOutputDir --inputFiles $allFiles"
+
+
+# Return to where we started
+Set-Location $OrigDir
```
nibanks, line 71:
```suggestion
Set-Location $OrigDir

```
Make sure to have trailing new lines.

src/core/packet.c
nibanks: That's weird. We have a lot of different builds onboarded to CI and they all pass without this. Since this is in logging code, I don't really care much either way, so you can just keep it.

src/generated/linux/ApiTest.cpp.clog.h
```diff
@@ -1,4 +1,6 @@
+#ifndef CLOG_DO_NOT_INCLUDE_HEADER
```
nibanks, line 1: Just curious, what's the reason for this?

scripts/build.ps1
```diff
@@ -328,6 +328,7 @@ function CMake-Generate {
     }
     $Arguments += " -DQUIC_TLS=" + $Tls
     $Arguments += " -DQUIC_OUTPUT_DIR=""$ArtifactsDir"""
+    $Arguments += " -DQUIC_LINUX_LOG_ENCODER=lttng"
```
nibanks, line 331: Should this  be in the `else` of the `if ($IsWindows) {` above instead of here?

scripts/prepare-machine.ps1
```diff
@@ -117,27 +114,9 @@ if ($InitSubmodules) {
...
-            Write-Error $_
-        }
-        $err = $_
-        $MessagesAtEnd.Add("$ToolName could not be installed. Building with logs will not work")
-        $MessagesAtEnd.Add($err.ToString())
-    }
-}
+Write-Host "Initializing clog submodule"
+git submodule init submodules/clog
+git submodule update
```
nibanks, line 119: I'm ok with this for now, but we should probably figure out how to be more selective for this in the future.

scripts/update-sidecar.ps1
nibanks: You resolved this, but the code is unchanged in the latest?

src/generated/CMakeLists.txt
```diff
@@ -10,10 +10,10 @@ if(QUIC_ENABLE_LOGGING)
         add_library(logging INTERFACE)
     else()
         target_include_directories(logging_inc INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/common)
-        target_include_directories(logging_inc INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/${QUIC_LTTNG_PLATFORM})
+        target_include_directories(logging_inc INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/${QUIC_LINUX_LOGGING_METHOD})
```
nibanks, line 13: Maybe we default `QUIC_LINUX_LOGGING_METHOD` to lttng and then the build.ps1 doesn't even need to set it?

thhous-msft (approved):

nibanks (approved):

PR #2263, opened 2022-01-03
Artificially increase ProcShifted by 1 for PacketId

thhous-msft:
The existing code has a lot of checks for Id not equal to 0. However, on core 0 with rollover this is a valid case. We can make this an always invalid case by artificially increasing the proc by 1 before shifting it.

### DIFF
--- a/src/core/binding.c
+++ b/src/core/binding.c
@@ -1563,7 +1563,7 @@ QuicBindingReceive(
     //
 
     uint32_t Proc = CxPlatProcCurrentNumber();
-    uint64_t ProcShifted = ((uint64_t)Proc) << 40;
+    uint64_t ProcShifted = ((uint64_t)Proc + 1) << 40;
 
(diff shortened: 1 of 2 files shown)


src/core/binding.c
```diff
@@ -1563,7 +1563,7 @@ QuicBindingReceive(
     //
 
     uint32_t Proc = CxPlatProcCurrentNumber();
-    uint64_t ProcShifted = ((uint64_t)Proc) << 40;
+    uint64_t ProcShifted = ((uint64_t)Proc + 1) << 40;
```
nibanks, line 1566: I believe there are more places this needs to be updated.

thhous-msft: Ah, there was 1 more. Fixed.

nibanks (approved):

anrossi (approved):

PR #2266, opened 2022-01-05
Improve Pacing Send Logic

nibanks:
Fixes an issue with large pacing send allowances, where `QUIC_MAX_DATAGRAMS_PER_SEND` actually ends up limiting the send, instead of the allowance, resulting in a complete loss of all that allowance.

The fix is to save the previous allowance, subtract what was actually sent, and then in the next send use (add) that old allowance too.

### DIFF
--- a/src/core/cubic.h
+++ b/src/core/cubic.h
@@ -59,6 +59,11 @@ typedef struct QUIC_CONGESTION_CONTROL_CUBIC {
     uint32_t BytesInFlightMax;
 
     //
+    // The leftover send allowance from a previous send. Olny used when pacing.
+    //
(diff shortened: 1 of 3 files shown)


src/core/cubic.h
```diff
@@ -58,6 +58,11 @@ typedef struct QUIC_CONGESTION_CONTROL_CUBIC {
     uint32_t BytesInFlight;
     uint32_t BytesInFlightMax;
 
+    //
+    // The leftover send allowance from a previous send. Olny used when pacing.
```
thhous-msft, line 62: *Only

anrossi (approved):

PR #2268, opened 2022-01-06
Create codetql-analysis.yml

CryptoGodfatherVA2:


nibanks:
How is this different from the [existing CodeQL](https://github.com/microsoft/msquic/blob/main/.github/workflows/codeql-analysis.yml) we have?

thhous-msft:
We already have codeql in this repo.

PR #2269, opened 2022-01-06
CUBICbis updates

maolson-msft:
Changes AIMD-friendly region algorithm to an AIMD implementation (where bytes are added to a window
value for each ACK proportional to the number of ACKed bytes) from a time-based function, to make it
easier to implement the dynamic AIMD slope that's in the new bis RFC.

### DIFF
--- a/src/core/cubic.c
+++ b/src/core/cubic.c
@@ -495,8 +495,8 @@ CubicCongestionControlOnDataAcknowledged(
             //
             // Concave or Convex region. Constrain TargetWindow within [CongestionWindow, 1.5*CongestionWindow].
             //
-            uint64_t TargetWindow = max(Cubic->CongestionWindow, min(CubicWindow, Cubic->CongestionWindow + (Cubic->CongestionWindow >> 1)));
-            Cubic->CongestionWindow += (ULONG)(((TargetWindow - Cubic->CongestionWindow) * Connection->Paths[0].Mtu) / Cubic->CongestionWindow);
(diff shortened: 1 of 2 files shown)


src/core/cubic.c
```diff
@@ -385,6 +388,10 @@ CubicCongestionControlOnDataAcknowledged(
         // Slow Start
         //
 
+        //
+        // TODO: CongestionWindow should only be allowed to grow up to SlowStartThreshold
```
maolson-msft, line 392: This is unrelated to the CUBIC bis FYI. The issue is that we intend to grow exponentially only up to SlowStartThreshold, but instead we could feasibly grow up to 2x the threshold in the slowstart code path. This is a bug that we fixed a while back in Windows TCP.

maolson-msft:
@nibanks How do I view the WAN perf test results and compare them to the baseline?

nibanks:
The results are in the GitHub action that merges the wanperf results. We don't currently have automated diff with main working.

nibanks:
@maolson-msft here are this PR's results compared to main:

image

[cubicBIS.xlsx](https://github.com/microsoft/msquic/files/7841491/cubicBIS.xlsx)

maolson-msft:
@nibanks Kind of hard to evaluate such a mass of test results, but it seems reasonable to me. A couple functional tests failed- I'm rerunning them

nibanks (approved):

PR #2270, opened 2022-01-07
Collect WAN Perf Logs and Fix Issues Found with them

nibanks:
Collect WAN Perf ETL files and fixes the following issues found with those logs:

1. When recovering from a spurious retransmit, loss detection must account for flushing a new send if previously blocked on congestion control.
2. There are cases where huge bottleneck queues can't drain in the allowed test run time, so all the sent data doesn't complete. In this case, just use the amount of data that did complete to calculate the transfer rate.
3. In some test timeout situations there was an ordering issue with the console output, which broke the PS script that parsed the output to read the final rate. Updated the code to ensure this no longer gets swapped.

### DIFF
--- a/src/core/loss_detection.c
+++ b/src/core/loss_detection.c
@@ -1305,8 +1305,13 @@ QuicLossDetectionProcessAckBlocks(
                 // All previously considered lost packets were found to be
                 // spuriously lost. Inform congestion control.
                 //
-                QuicCongestionControlOnSpuriousCongestionEvent(
-                    &Connection->CongestionControl);
+                if (QuicCongestionControlOnSpuriousCongestionEvent(
+                        &Connection->CongestionControl)) {
+                    //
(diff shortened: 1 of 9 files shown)


src/core/loss_detection.c
```diff
@@ -1305,8 +1305,13 @@ QuicLossDetectionProcessAckBlocks(
                 // All previously considered lost packets were found to be
                 // spuriously lost. Inform congestion control.
                 //
-                QuicCongestionControlOnSpuriousCongestionEvent(
-                    &Connection->CongestionControl);
+                if (QuicCongestionControlOnSpuriousCongestionEvent(
```
anrossi, line 1308: Will this result in multiple flushes happening back to back? one that actually sends data, and one that is a no-op?

nibanks: Nope

PR #2284, opened 2022-01-12
WPA Plugin Tx Delay Graph

nibanks:
image

Adds a new graph type that plots the delay (in microseconds, capped to 1 ms) between sends, over time. This can help to see how spaced out transmits are for pacing related issues.

Also just added a few bug fixes I found while using this.


nibanks:
image

Full example of latest WPA plugin output.

anrossi (approved):

PR #2288, opened 2022-01-12
Add cswitch and sampling with stacks to RPS verbose profile

mtfriesen:


src/manifest/MsQuic.wprp
```diff
@@ -26,6 +26,25 @@
...
+          <Keyword Value="Loader"/>
+          <Keyword Value="CSwitch"/>
+          <Keyword Value="ReadyThread"/>
+          <Keyword Value="SampledProfile"/>
+          <Keyword Value="DPC"/>
+          <Keyword Value="Interrupt"/>
+          <Keyword Value="IdleStates"/>
+      </Keywords>
+
+      <Stacks>
```
nibanks, line 41:
```suggestion
      </Keywords>
      <Stacks>
```

nibanks (approved):

src/manifest/MsQuic.wprp
mtfriesen: This is consistent with the block below.

mtfriesen: Otherwise I don't feel strongly one way or the other.

PR #2292, opened 2022-01-13
Limit Pacing Sends to MTU

nibanks:
Don't try to pace a send if we don't even have a full MTU to send yet.

### DIFF
--- a/src/core/cubic.c
+++ b/src/core/cubic.c
@@ -148,7 +148,7 @@ CubicCongestionControlGetSendAllowance(
         !TimeSinceLastSendValid ||
         !Connection->Settings.PacingEnabled ||
         !Connection->Paths[0].GotFirstRttSample ||
-        Connection->Paths[0].SmoothedRtt < MS_TO_US(QUIC_SEND_PACING_INTERVAL)) {
+        Connection->Paths[0].SmoothedRtt < QUIC_MIN_PACING_RTT) {
         //
         // We're not in the necessary state to pace.
         //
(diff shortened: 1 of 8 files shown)


src/core/cubic.c
```diff
@@ -189,6 +189,13 @@ CubicCongestionControlGetSendAllowance(
         }
 
         Cubic->LastSendAllowance = SendAllowance;
+
+        if (SendAllowance < Connection->Paths[0].Mtu) {
```
nibanks, line 193: I was also thinking of a slight modification to this:
> Only pace integral multiples of MTU, saving the rest for later. It would have the same result as this PR for sub-MTU allowances, but possible also have an effect at higher allowances.

cc @maolson-msft and @csujedihy for their thoughts

nibanks:
```c
        //
        // Limit send allowance to integral multiples of MTU.
        //
        uint32_t SendCount = SendAllowance / Connection->Paths[0].Mtu;
        SendAllowance = SendCount * Connection->Paths[0].Mtu;
```

nibanks: Actually:
```c
        //
        // Limit send allowance to integral multiples of MTU (converted to the
        // datagram payload length).
        //
        uint32_t DatagramLength =
            MaxUdpPayloadSizeForFamily(
                QuicAddrGetFamily(&Connection->Paths[0].Route.RemoteAddress),
                Connection->Paths[0].Mtu);
        uint32_t SendCount = SendAllowance / DatagramLength;
        SendAllowance = SendCount * DatagramLength;
```

maolson-msft: This is closely related to a conversation I was having with Yi and Praveen recently- we should be restricting the congestion window to always be multiples of MTU. Currently this is not done consistently in either windows TCP or obviously msquic.

nibanks:
Seems MTU sends regressed the perf. Will revisit in future.

PR #2296, opened 2022-01-14
General CC Improvements

nibanks:
- Update CC to use max datagram payload length, instead of MTU (at IP layer) when calculating byte limits.
- Update connection timer set API to take microseconds instead of milliseconds (it already stored as microseconds internally) to support more fine grain timers (i.e. for pacing eventually).
- Split min RTT pacing limit (below which we disable pacing) from the actual pacing interval, so we can reduce the pacing interval in the future without affecting low RTT connections.

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -1170,7 +1170,7 @@ QuicConnTimerSet(
     _In_ uint64_t Delay
     )
 {
-    const uint64_t NewExpirationTime = CxPlatTimeUs64() + MS_TO_US(Delay);
+    const uint64_t NewExpirationTime = CxPlatTimeUs64() + Delay;
 
     QuicTraceEvent(
         ConnSetTimer,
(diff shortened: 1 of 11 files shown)


src/core/connection.c
```diff
@@ -1170,7 +1170,7 @@ QuicConnTimerSet(
     _In_ uint64_t Delay
```
anrossi, line 1170: should this parameter be named `DelayUs` to indicate the expected resolution?

nibanks: It could be, but I think I'd have to update CLOG too since it's a logged parameter, and I don't want to have to do that. :D

anrossi (approved):

maolson-msft:
I thought we changed Connection->Paths[0].Mtu to be the max datagram payload size long ago? Did that never actually happen? When does QUIC ever need to know the IP MTU instead of the max datagram payload size?

nibanks:
> I thought we changed Connection->Paths[0].Mtu to be the max datagram payload size long ago? Did that never actually happen? When does QUIC ever need to know the IP MTU instead of the max datagram payload size?

No, we never did. I was just thinking it'd be worth finally doing that change...

PR #2299, opened 2022-01-16
Allow client and server to share bindings

anrossi:
Allow bindings to be shared between clients and servers. This will allow a client to create outbound connections on the same address and port as a listener on the same node.

### DIFF
--- a/src/core/binding.c
+++ b/src/core/binding.c
@@ -1441,7 +1441,7 @@ QuicBindingDeliverDatagrams(
                 Packet->DestCidLen);
     }
 
-    if (Connection == NULL) {
+    if (Connection == NULL && Binding->ServerOwned) {
 
         //
         // Because the packet chain is ordered by control packets first, we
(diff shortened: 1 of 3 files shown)


src/core/binding.h
```diff
@@ -178,13 +178,16 @@ typedef struct QUIC_BINDING {
...
-    // and servers.
+    // Indicates whether the binding has server side (i.e. listener
+    // and server connections) owners.
     //
     BOOLEAN ServerOwned : 1;
 
+    //
+    // Indicates whether this binding has client side owners or not.
+    //
+    BOOLEAN ClientOwned : 1;
```
nibanks, line 189: Let's rename this "Owned" flags to something like "Used` instead.

src/core/binding.c
```diff
@@ -1422,23 +1423,25 @@ QuicBindingDeliverDatagrams(
...
-    if (!Binding->ServerOwned || Packet->IsShortHeader) {
-        Connection =
-            QuicLookupFindConnectionByLocalCid(
-                &Binding->Lookup,
-                Packet->DestCid,
-                Packet->DestCidLen);
-    } else {
+    QUIC_CONNECTION* Connection = NULL;
+
+    if (Binding->ServerOwned) {
```
nibanks, line 1428: You don't want to do this for short header packets. Only server long header packets.

nibanks: Also to better optimize the branches, I think following the old pattern, but adding another `ByLocalCid` in the case `ByRemoteHash` fails and there are client users.

nibanks (commented):
Generally looks good! It definitely needs some corresponding test cases to make sure it can be shared. Ideally also update spinquic.

anrossi:
I think my openssl submodule is on an old commit and it's showing up in this PR. I'll fix that before merging

PR #2304, opened 2022-01-18
Don't Double CWND Reduction in Persistent Congestion

### LINKED ISSUE #2231: CUBIC double CWND reduction for CubicCongestionControlOnPersistentCongestionEvent
### Describe the bug

In the persistent congestion case, CubicCongestionControlOnPersistentCongestionEvent is called right after CubicCongestionControlOnDataLost, which results in a reduction factor of 0.7*0.7 instead of the intended 0.7.

### Affected OS

- [ ] Windows Server 2022
- [ ] Windows 11
- [ ] Ubuntu
- [ ] Debian
- [ ] Other (specify below)

### Additional OS information

_No response_

### MsQuic version

main

run connection on link with heavy loss

### Expected behavior

ssthresh is reduced by 0.7.

### Actual outcome

ssthresh is reduced by 0.49.

### Additional details

_No response_

nibanks:
Fixes #2231. Don't apply both normal and persistent congestion logic on the same "data lost" event.

### DIFF
--- a/src/core/cubic.c
+++ b/src/core/cubic.c
@@ -224,7 +224,8 @@ CubicCongestionControlUpdateBlockedState(
 _IRQL_requires_max_(DISPATCH_LEVEL)
 void
 CubicCongestionControlOnCongestionEvent(
-    _In_ QUIC_CONGESTION_CONTROL* Cc
+    _In_ QUIC_CONGESTION_CONTROL* Cc,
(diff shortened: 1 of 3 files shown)


nibanks:
@maolson-msft I'd appreciate if you could ensure I am doing things right here.

maolson-msft (approved):

PR #2309, opened 2022-01-19
Prevent exponential CongestionWindow growth beyond SlowStartThreshold

maolson-msft:
Port of a 2018 fix to the Windows TCP CUBIC implementation.

### DIFF
--- a/src/core/cubic.c
+++ b/src/core/cubic.c
@@ -367,9 +367,10 @@ CubicCongestionControlOnDataAcknowledged(
     uint64_t TimeNow = US_TO_MS(AckEvent->TimeNow);
     QUIC_CONNECTION* Connection = QuicCongestionControlGetConnection(Cc);
     BOOLEAN PreviousCanSendState = CubicCongestionControlCanSend(Cc);
+    uint32_t BytesAcked = AckEvent->NumRetransmittableBytes;
 
-    CXPLAT_DBG_ASSERT(Cubic->BytesInFlight >= AckEvent->NumRetransmittableBytes);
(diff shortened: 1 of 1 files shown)


src/core/cubic.c
```diff
@@ -397,21 +398,30 @@ CubicCongestionControlOnDataAcknowledged(
...
+        Cubic->CongestionWindow += BytesAcked;
+        BytesAcked = 0;
         if (Cubic->CongestionWindow >= Cubic->SlowStartThreshold) {
             Cubic->TimeOfCongAvoidStart = CxPlatTimeMs64();
+
+            //
+            // We only want exponential growth up to SlowStartThreshold. If
+            // CongestionWindow has increased beyond SlowStartThreshold, set it back
+            // to SlowStartThreshold and treat the spare BytesAcked as if the bytes
+            // were acknowledged during Congestion Avoidance below.
```
nibanks, line 410: How would this work, considering `TimeInCongAvoid` would be `0`?

maolson-msft: The main point here is updating the AIMD window, which is dependent on the number of bytes ACKed instead of the amount of time in CA.

maolson-msft:
@nibanks let's merge this and I'll make a PR with per-ACK increase limits for Saurav to test out.

src/core/cubic.c
nibanks: Ok. Just merge my PR and main into this one. Assuming CI doesn't fail, we can merge it.

nibanks (approved):

nibanks (approved):

PR #2323, opened 2022-01-21
Cap CWND Growth to Prevent Large Send Bursts

nibanks:
Updated the QUIC code to match TCP's logic of caping the CWND growth in slow start to no more than 8 MTUs for a single ACK.

### DIFF
--- a/src/core/cubic.c
+++ b/src/core/cubic.c
@@ -402,7 +402,7 @@ CubicCongestionControlOnDataAcknowledged(
         //
 
         const uint32_t MaxCwndGrowth =
-            (uint32_t)DatagramPayloadLength * QUIC_MAX_CWND_GROWTH_PACKETS;
+            (uint32_t)DatagramPayloadLength * QUIC_MAX_CONGESTION_WINDOW_INCREASE;
 
         if (BytesAcked > MaxCwndGrowth) {
             //
--- a/src/core/cubic.c
+++ b/src/core/cubic.c
@@ -374,8 +374,8 @@ CubicCongestionControlOnDataAcknowledged(
 
     if (BytesAcked > QUIC_MAX_CONGESTION_WINDOW_INCREASE) {
         //
-        // Cap the increase CWND growth to prevent it from resulting in large
-        // bursts into the network.
+        // Cap the CWND growth to prevent it from resulting in large bursts into
+        // the network.
         //
         BytesAcked = QUIC_MAX_CONGESTION_WINDOW_INCREASE;
     }
--- a/src/core/cubic.c
+++ b/src/core/cubic.c
@@ -369,16 +369,11 @@ CubicCongestionControlOnDataAcknowledged(
     BOOLEAN PreviousCanSendState = CubicCongestionControlCanSend(Cc);
     uint32_t BytesAcked = AckEvent->NumRetransmittableBytes;
 
+    const uint16_t DatagramPayloadLength =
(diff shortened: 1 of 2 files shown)


src/core/cubic.c
```diff
@@ -398,6 +401,17 @@ CubicCongestionControlOnDataAcknowledged(
...
 
+        const uint32_t MaxCwndGrowth =
+            (uint32_t)DatagramPayloadLength * QUIC_MAX_CONGESTION_WINDOW_INCREASE;
+
+        if (BytesAcked > MaxCwndGrowth) {
+            //
+            // Cap the CWND growth in slow start to prevent it from resulting in
+            // large bursts into the network.
+            //
+            BytesAcked = MaxCwndGrowth;
```
maolson-msft, line 412: This isn't quite right- we want to limit the SS increase, but not throw away the extra bytes in the case where we are proceeding into the CA path (since QUIC_MAX_CONGESTION_WINDOW_INCREASE doesn't apply in CA).

src/core/quicdef.h
```diff
@@ -376,6 +376,12 @@ CXPLAT_STATIC_ASSERT(
 //
 #define QUIC_SEND_PACING_INTERVAL               1000
 
+//
+// The maximum number of MTUs (datagram payload length) we will allow the
```
maolson-msft, line 380: It's more subtle than this. Might be safest to say "Congestion window increase limit L from RFC 3465".

src/core/cubic.c
nibanks: You're talking about the case where `BytesAcked + CongestionWindow > SlowStartThreshold`? How would you recommend changing this?

src/core/quicdef.h
nibanks: I'd rather have a complete explanation here instead of requiring someone to go look at an RFC. I'm fine referencing an RFC, but I don't want to require knowledge of an RFC.

src/core/cubic.c
maolson-msft: Yes. Something like
uint32_t NewCongestionWindow = Cubic->CongestionWindow + min(DatagramPayloadLength * QUIC_MAX_CONGESTION_WINDOW_INCREASE, BytesAcked);
BytesAcked -= NewCongestionWindow - Cubic->CongestionWindow;

then you'd need to change the "BytesAcked = Cubic->CongestionWindow - Cubic->SlowStartThreshold;" a few lines down to be a "+=".
CWND = NewCWND;

src/core/quicdef.h
maolson-msft: How about this then: "Slow start congestion window increase limit L from RFC 3465, to reduce burstiness". That's at least a complete description of how the threshold is used.

src/core/cubic.c
nibanks: I'm a bit confused how that logic would work. If cwnd is still < ssthresh, you want to set BytesAcked to `0`, right?

maolson-msft: Oh right. It would need to be more like

newcwnd = cwnd + min(MAX_CWND_INCREASE*MTU, BytesAcked);
if (newcmd >= ssthresh) {
    newcwnd = ssthresh;
    BytesAcked -= newcwnd - cwnd;
} else {
    BytesAcked = 0;
}
cwnd = newcwnd;

Your updated code looks like it works correctly, but consider the above if you think it might be cleaner.

maolson-msft (approved):

PR #2331, opened 2022-01-25
Switch windows-latest to windows-2022

thhous-msft:
It's going to be upgrading by default soon, so test to make sure this won't break anything


nibanks:
Is the goal to actually change these or just check that they still work?

thhous-msft:
Just a check. Cargo is currently failing, and we can decide if we want to upgrade that or not.

nibanks (approved):

PR #2336, opened 2022-01-26
Combined QUIC and Datapath Execution Context Support

nibanks:
Add a new compile-time option to have the QUIC and datapath share threads, generally for reduced latency. The latency improvements come from the removal of queuing and waking threads when performance contexts switches between the two.


src/core/worker.c
```diff
@@ -93,6 +97,14 @@ QuicWorkerInitialize(
         goto Error;
     }
 
+    Worker->ExecutionContext.Context = Worker;
+    Worker->ExecutionContext.Callback = QuicWorkerLoop;
```
maolson-msft, line 101: Is Worker->ExecutionContext.Callback used when QUIC_USE_EXECUTION_CONTEXTS is not defined? I ask because you hid most shared-EC code behind ifdefs but not this part, and because QuicWorkerThread already calls QuicWorkerLoop in the non-shared-EC version.

nibanks: Yes, I believe more _could_ be hidden behind the #ifdefs but I'm still not 100% sure we want this compile-time in the end. Thoughts on that?

maolson-msft:
What's the longer-term plan here w.r.t. support of both shared and nonshared ECs? This code is absurdly hard to understand when it does both, and I kind of feel like the benefit to small numbers of connections when using nonshared ECs is artificial (in the sense that a system only doing a couple connections at a time is unlikely to be one of those systems where CPU efficiency or absolute max throughput is critical).

src/core/worker.c
maolson-msft: Right, this is partly related to my other comment.

maolson-msft (dismissed):

src/core/worker.c
```diff
@@ -848,4 +868,4 @@ QuicWorkerPoolGetLeastLoadedWorker(
 
     WorkerPool->LastWorker = MinQueueDelayWorker;
     return MinQueueDelayWorker;
-}
+}
```
anrossi, line 871: missing newline at EOF

src/core/worker.h
```diff
@@ -196,4 +198,4 @@ void
 QuicWorkerQueueOperation(
     _In_ QUIC_WORKER* Worker,
     _In_ QUIC_OPERATION* Operation
-    );
+    );
```
anrossi, line 201: newline EOF

src/core/worker.c
anrossi: If it's that way in main, why is it showing a diff here, then?

anrossi:
No new tests to ensure the new EC path works?  Just using the old EC path with the existing tests just ensures that it's not broken, but doesn't ensure the new stuff works as expected

anrossi (approved):

PR #2337, opened 2022-01-26
Fix msquicplatformtest crashing if pfx parameter is missing

thhous-msft:


src/platform/unittest/TlsTest.cpp
```diff
@@ -777,6 +779,7 @@ TEST_F(TlsTest, Handshake)
 #ifndef QUIC_DISABLE_PFX_TESTS
 TEST_F(TlsTest, HandshakeCertFromFile)
 {
+    ASSERT_NE(nullptr, CertParamsFromFile);
```
nibanks, line 782: `SetUpTestSuite()` will still fail, causing all tests to be skipped. Perhaps we should update that to only generate `CertParamsFromFile` if `PfxPath` is present, and then keep this line here, so only this test fails?

thhous-msft: Probably a good idea. Fixed.

nibanks (approved):

anrossi (approved):

PR #2339, opened 2022-01-26
Limit Receive Flush Sizes

nibanks:
Adds code to limit the size of receive flush batches, in an effort to improve the ACK path.


src/core/quicdef.h
```diff
@@ -190,14 +190,13 @@ typedef struct QUIC_PATH QUIC_PATH;
 // The maximum number of received packets that may be queued on a single
 // connection. When this limit is reached, any additional packets are dropped.
 //
-#ifdef _KERNEL_MODE
+#define QUIC_MAX_RECEIVE_QUEUE_COUNT            4096
```
csujedihy, line 193: I bet this will solve the duonic 1G perf issue we are seeing.

src/core/connection.c
```diff
@@ -5512,24 +5512,40 @@ QuicConnRecvDatagrams(
...
-    Connection->ReceiveQueueTail = &Connection->ReceiveQueue;
+    if (Connection->ReceiveQueueCount > QUIC_MAX_RECEIVE_FLUSH_COUNT) {
+        FlushedAll = FALSE;
+        Connection->ReceiveQueueCount -= QUIC_MAX_RECEIVE_FLUSH_COUNT;
+        CXPLAT_RECV_DATA* Tail = Connection->ReceiveQueue;
+        ReceiveQueueCount = 0;
+        while (ReceiveQueueCount++ < QUIC_MAX_RECEIVE_FLUSH_COUNT) {
+            Tail = Connection->ReceiveQueue;
+            Connection->ReceiveQueue = Connection->ReceiveQueue->Next;
+        }
```
csujedihy, line 5534: I think this while loop can be optimized a bit (mainly for better readability), isn't it?

```
        while (++ReceiveQueueCount < QUIC_MAX_RECEIVE_FLUSH_COUNT) {
            Tail = Tail->Next;
        }
        Connection->ReceiveQueue = Tail->Next;
        Tail->Next = NULL;
```

csujedihy:
> // Kernel modes receive path is slightly different, so allow larger queue sizes.

not accurate anymore

---
Refers to: src/core/quicdef.h:195 in 0a9fdab. [](commit_id = 0a9fdab635a731c6088327224f35a4dde01d70c2, deletion_comment = False)

src/core/connection.c
nibanks: Yeah, I think that works.

csujedihy (approved):

PR #2346, opened 2022-01-28
Async Listener Stop

nibanks:
Another suggestion from #1709. cc @geoffkizer


src/platform/platform_posix.c
```diff
@@ -270,22 +270,23 @@ CxPlatRefIncrement(
 
 BOOLEAN
 CxPlatRefIncrementNonZero(
-    _Inout_ volatile CXPLAT_REF_COUNT* RefCount
+    _Inout_ volatile CXPLAT_REF_COUNT* RefCount,
+    _In_ uint32_t Bias
```
anrossi, line 274: For adding more than 1 ref at a time?

anrossi (changes requested):
Update docs with the new event, and add tests for calling `ListenerStop` on the worker thread, and the app thread, some API tests to validate calling Stop before Start fails, etc.

src/platform/platform_posix.c
nibanks: This particular change was just to make all platforms have the same signature.

src/core/listener.c
```diff
@@ -122,27 +155,28 @@ MsQuicListenerClose(
...
 
-    //
-    // Make sure the listener has unregistered from the binding.
-    //
-    MsQuicListenerStop(Handle);
+    QUIC_LIB_VERIFY(!Listener->AppClosed);
+    Listener->AppClosed = TRUE;
 
-    CxPlatRundownUninitialize(&Listener->Rundown);
+    if (Listener->StopCompleteThreadID == CxPlatCurThreadID()) {
```
thhous-msft, line 162: I'm not quite sure on this. What happens if the app calls ListenerClose() from a different callback?

nibanks: There's nothing we can do unless really.

docs/api/ListenerClose.md
```diff
@@ -23,7 +23,7 @@ A valid handle to an open listener object.
 
 # Remarks
 
-`ListenerClose` frees all allocated resources associated with the listener handle. If a listener has not had [ListenerStop](ListenerStop.md) called on it at the time `ListenerClose` is called, [ListenerStop](ListenerStop.md) is invoked internally. Due to this, a server application **MUST NOT** call `ListenerClose` within any callback unless it has already called [ListenerStop](ListenerStop.md) outside of any callbacks.
+`ListenerClose` frees all allocated resources associated with the listener handle. If a listener has not had [ListenerStop](ListenerStop.md) called on it at the time `ListenerClose` is called, [ListenerStop](ListenerStop.md) is invoked internally. `ListenerClose` may be called on any Listener callback.
```
thhous-msft, line 26: This needs to be changed that ListenerClose can only be called from another thread or the StopComplete event.

thhous-msft (dismissed):

docs/api/ListenerClose.md
```diff
@@ -23,7 +23,7 @@ A valid handle to an open listener object.
 
 # Remarks
 
-`ListenerClose` frees all allocated resources associated with the listener handle. If a listener has not had [ListenerStop](ListenerStop.md) called on it at the time `ListenerClose` is called, [ListenerStop](ListenerStop.md) is invoked internally. Due to this, a server application **MUST NOT** call `ListenerClose` within any callback unless it has already called [ListenerStop](ListenerStop.md) outside of any callbacks.
+`ListenerClose` frees all allocated resources associated with the listener handle. If a listener has not had [ListenerStop](ListenerStop.md) called on it at the time `ListenerClose` is called, [ListenerStop](ListenerStop.md) is invoked internally. `ListenerClose` may be called on any Listener callback. A server application **MUST NOT** call `ListenerClose` within any callback except for `QUIC_LISTENER_EVENT_STOP_COMPLETE`, or unless it has already received the `QUIC_LISTENER_EVENT_STOP_COMPLETE` event.
```
anrossi, line 26: The second part here seems to contradict the first. It says you can call `ListenerClose` without calling `ListenerStop` first. But then it says you cannot call `ListenerClose` unless you have received the `QUIC_LISTENER_EVENT_STOP_COMPLETE` event.

anrossi (approved):

docs/api/ListenerClose.md
nibanks: Fixed

thhous-msft (approved):

PR #2347, opened 2022-01-28
Updates for ACK Frequency Draft 2

nibanks:
Another change I wanted to get in before v2. Updates our ACK frequency usage to the latest draft version.

### DIFF
--- a/src/core/frame.c
+++ b/src/core/frame.c
@@ -1190,7 +1190,7 @@ typedef struct QUIC_ACK_FREQUENCY_EXTRAS {
         struct {
             uint8_t IgnoreOrder : 1;
             uint8_t IgnoreCE    : 1;
-            uint8_t FrameType   : 6;
+            uint8_t Reserved    : 6;
         };
         uint8_t Value;
     };
--- a/src/core/frame.c
+++ b/src/core/frame.c
@@ -1184,6 +1184,19 @@ QuicDatagramFrameDecode(
     return TRUE;
 }
 
+typedef struct QUIC_ACK_FREQUENCY_EXTRAS {
+
+    union {
+        struct {
+            uint8_t IgnoreOrder : 1;
+            uint8_t IgnoreCE    : 1;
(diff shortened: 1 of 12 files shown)


src/core/frame.c
```diff
@@ -1227,13 +1245,15 @@ QuicAckFrequencyFrameDecode(
...
+    QUIC_ACK_FREQUENCY_EXTRAS Extras;
     if (!QuicVarIntDecode(BufferLength, Buffer, Offset, &Frame->SequenceNumber) ||
         !QuicVarIntDecode(BufferLength, Buffer, Offset, &Frame->PacketTolerance) ||
         !QuicVarIntDecode(BufferLength, Buffer, Offset, &Frame->UpdateMaxAckDelay) ||
-        !QuicUint8tDecode(BufferLength, Buffer, Offset, &Frame->IgnoreOrder) ||
-        Frame->IgnoreOrder > 1) { // IgnoreOrder should only be 0 or 1.
+        !QuicUint8tDecode(BufferLength, Buffer, Offset, &Extras.Value)) {
         return FALSE;
     }
+    Frame->IgnoreOrder = Extras.IgnoreOrder;
```
anrossi, line 1255: Should the `FrameType` value be validated here to ensure it's the expected value?

anrossi (dismissed):
Looks good, but should have a test too

src/core/frame.c
anrossi: In `QUIC_ACK_FREQUENCY_EXTRAS`, `FrameType` is 6 bits, and I thought that was a useful value. If it's just a singular bit, I guess it doesn't need validation. Though, maybe the remaining 5 bits need to be validated as zero?

nibanks: Oops. `FrameType` should be `Reserved` and set to zero on write and ignored on read. So we have no functional issue here, but a variable name issue.

thhous-msft (approved):

anrossi (approved):

PR #2349, opened 2022-01-29
Add TLS test for the different credential types.

anrossi:
Adds a test which tests all the different credential types supported for each platform and TLS library.


src/inc/quic_platform.h
```diff
@@ -487,6 +493,15 @@ CxPlatGetTestCertificate(
...
     _When_(CredType != QUIC_CREDENTIAL_TYPE_CERTIFICATE_HASH_STORE, _Reserved_)
         QUIC_CERTIFICATE_HASH_STORE* CertHashStore,
+    _When_(CredType == QUIC_CREDENTIAL_TYPE_CERTIFICATE_FILE, _Out_)
+    _When_(CredType != QUIC_CREDENTIAL_TYPE_CERTIFICATE_FILE, _Reserved_)
+        QUIC_CERTIFICATE_FILE* CertFile,
+    _When_(CredType == QUIC_CREDENTIAL_TYPE_CERTIFICATE_FILE_PROTECTED, _Out_)
+    _When_(CredType != QUIC_CREDENTIAL_TYPE_CERTIFICATE_FILE_PROTECTED, _Reserved_)
+        QUIC_CERTIFICATE_FILE_PROTECTED* CertFileProtected,
+    _When_(CredType == QUIC_CREDENTIAL_TYPE_CERTIFICATE_PKCS12, _Out_)
+    _When_(CredType != QUIC_CREDENTIAL_TYPE_CERTIFICATE_PKCS12, _Reserved_)
```
nibanks, line 503: The number of args is getting quite large here. I am thinking a struct that contains a union of what's necessary might be better.

src/platform/selfsign_openssl.c
```diff
@@ -200,55 +256,113 @@ CxPlatTlsGenerateSelfSignedCert(
         goto Exit;
     }
 
-    Fd = fopen(PrivateKeyFileName, "wb");
+    if (!OutputPkcs12) {
+
+        Fd = fopen(PrivateKeyFileName, "wb");
+
+        if (Fd == NULL) {
```
nibanks, line 263: nit: generally we follow the pattern of `if` checks immediately follow the functions.
```suggestion
        Fd = fopen(PrivateKeyFileName, "wb");
        if (Fd == NULL) {
```

src/platform/selfsign_openssl.c
```diff
@@ -200,55 +256,113 @@ CxPlatTlsGenerateSelfSignedCert(
...
+            QuicTraceEvent(
+                LibraryError,
+                "[ lib] ERROR, %s.",
+                "fopen() failed");
```
nibanks, line 267: nit: In all these "failed" logs please remove `()`.

src/test/lib/ApiTest.cpp
```diff
@@ -1941,3 +1941,19 @@ QuicTestConnectionRejection(
...
+QuicTestCredentialLoad(const QUIC_CREDENTIAL_CONFIG* Config)
+{
+    MsQuicRegistration Registration;
+    TEST_TRUE(Registration.IsValid());
+
+    MsQuicAlpn Alpn("MsQuicTest");
+
+    MsQuicSettings Settings;
+
+    MsQuicConfiguration Configuration(Registration, Alpn, Settings);
```
nibanks, line 1955:
```suggestion
    MsQuicConfiguration Configuration(Registration, "MsQuicTest");
```

src/test/bin/quic_gtest.h
```diff
@@ -692,3 +692,76 @@ std::ostream& operator << (std::ostream& o, const RebindPaddingArgs& args) {
...
+        ::std::vector<TlsConfigArgs> List;
+        for (auto CredType : {
+#ifdef _WIN32
+            QUIC_CREDENTIAL_TYPE_CERTIFICATE_HASH,
+            QUIC_CREDENTIAL_TYPE_CERTIFICATE_HASH_STORE,
+            QUIC_CREDENTIAL_TYPE_CERTIFICATE_CONTEXT,
+#else
+            QUIC_CREDENTIAL_TYPE_CERTIFICATE_FILE,
+            QUIC_CREDENTIAL_TYPE_CERTIFICATE_FILE_PROTECTED,
+            QUIC_CREDENTIAL_TYPE_CERTIFICATE_PKCS12
```
nibanks, line 711: We don't support any of these on Windows?

src/platform/selfsign_openssl.c
anrossi: That was how it was when I moved the code from the original location.

src/inc/quic_platform.h
anrossi: I'd prefer to keep callers simpler by not requiring a separate struct. If we add another credential type, I'm open to discussing the change then.

nibanks: Ok.

src/platform/selfsign_openssl.c
nibanks: I'd still prefer it changed, since most all other code doesn't include it.

src/test/bin/quic_gtest.h
anrossi: The code path in `CxPlatTlsSecConfigCreate` is no different on Windows or POSIX if those credential types are presented.

src/platform/selfsign_openssl.c
```diff
@@ -454,32 +680,130 @@ CxPlatGetTestCertificate(
...
+            } else {
+                CertFileName = QuicTestServerPkcs12Filename;
+            }
+            CertFilePath = malloc(MAX_PATH);
+            if (CertFilePath == NULL) {
+                return FALSE;
+            }
+        }
+
+        if(!FindOrCreateTempFiles(CertFileName, KeyFileName, CertFilePath, KeyFilePath)) {
```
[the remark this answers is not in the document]
anrossi: We should just use an automated code formatter so we don't waste time on these kinds of things.

nibanks (approved):

PR #2357, opened 2022-02-01
Switch API table to v2

thhous-msft:
This makes the API change more breaking to users. This also allows us to remove the existing MsQuicOpen backcompat function. Doing this lets us add the inline helper back in. Additionally, added [[nodiscard]] to ensure the result here is not discarded

### DIFF
--- a/docs/api/MsQuicOpenVersion.md
+++ b/docs/api/MsQuicOpenVersion.md
@@ -40,7 +40,7 @@ Calls to **MsQuicOpenVersion** and [MsQuicClose](MsQuicClose.md) increment and d
 
 **MsQuicOpenVersion** may dynamically load other dependencies, so it **must not** be called from [DllMain](https://docs.microsoft.com/en-us/windows/win32/dlls/dllmain) on Windows.
 
-**MsQuicOpenVersion** Takes a version number to indicate which version of the API to use. Newer versions of the library will support older versions of the API for binary compatibility. For backwards compatiblity with old clients, the previous **MsQuicOpen** exists as an export in the binary. The **MsQuicOpen** definition in msquic.h now forwards to **MsQuicOpenVersion**
+**MsQuicOpenVersion** Takes a version number to indicate which version of the API to use. Newer versions of the library will support older versions of the API for binary compatibility. For backwards compatiblity with old clients, the previous **MsQuicOpen2** exists as an export in the binary. The **MsQuicOpen2** definition in msquic.h now forwards to **MsQuicOpenVersion**
 
 # See Also
 
--- a/docs/api/MsQuicOpenVersion.md
+++ b/docs/api/MsQuicOpenVersion.md
@@ -40,7 +40,7 @@ Calls to **MsQuicOpenVersion** and [MsQuicClose](MsQuicClose.md) increment and d
 
 **MsQuicOpenVersion** may dynamically load other dependencies, so it **must not** be called from [DllMain](https://docs.microsoft.com/en-us/windows/win32/dlls/dllmain) on Windows.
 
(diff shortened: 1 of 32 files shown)


src/tools/interop/interop.cpp
```diff
@@ -1213,7 +1213,7 @@ main(
 
     CxPlatLockInitialize(&TestResultsLock);
 
-    if (QUIC_FAILED(Status = MsQuicOpen(&MsQuic))) {
+    if (QUIC_FAILED(Status = MsQuicOpen2(&MsQuic))) {
         printf("MsQuicOpen failed, 0x%x!\n", Status);
```
nibanks, line 1217: Can we fix up these printf/comments where possible too?

src/inc/msquic.h
```diff
@@ -1334,31 +1338,31 @@ MsQuicOpenVersion(
...
+    return MsQuicOpenVersion(2, (const void**)QuicApi);
 }
 
-#else*/
-
-#define MsQuicOpen(QuicApi) MsQuicOpenVersion(1, (const void**)QuicApi)
+#else
 
-/*#endif // defined(__cplusplus) || defined(WIN32)*/
+#define MsQuicOpen2(QuicApi) MsQuicOpenVersion(2, (const void**)QuicApi)
```
maolson-msft, line 1363: Why is this alias needed?

nibanks: Preference. Not strictly required.

thhous-msft: Also, assuming youre calling from C++ you'll get the correct output type, instead of just being void**

docs/api/MsQuicOpenVersion.md
```diff
@@ -40,7 +40,7 @@ Calls to **MsQuicOpenVersion** and [MsQuicClose](MsQuicClose.md) increment and d
 
 **MsQuicOpenVersion** may dynamically load other dependencies, so it **must not** be called from [DllMain](https://docs.microsoft.com/en-us/windows/win32/dlls/dllmain) on Windows.
 
-**MsQuicOpenVersion** Takes a version number to indicate which version of the API to use. Newer versions of the library will support older versions of the API for binary compatibility. For backwards compatiblity with old clients, the previous **MsQuicOpen** exists as an export in the binary. The **MsQuicOpen** definition in msquic.h now forwards to **MsQuicOpenVersion**
+**MsQuicOpenVersion** Takes a version number to indicate which version of the API to use. Newer versions of the library will support older versions of the API for binary compatibility. For backwards compatiblity with old clients, the previous **MsQuicOpen2** exists as an export in the binary. The **MsQuicOpen2** definition in msquic.h now forwards to **MsQuicOpenVersion**
```
nibanks, line 43:
```suggestion
**MsQuicOpenVersion** Takes a version number to indicate which version of the API to use. Newer versions of the library will support older versions of the API for binary compatibility. The **MsQuicOpen2** definition in msquic.h now forwards to **MsQuicOpenVersion**
```

src/core/library.c
```diff
@@ -1296,32 +1296,38 @@ QuicLibraryGetParam(
...
+    _Out_ _Pre_defensive_ const void** QuicApi
     )
 {
     QUIC_STATUS Status;
     BOOLEAN ReleaseRefOnFailure = FALSE;
 
+    if (Version != 2) {
+        QuicTraceLogVerbose(
+            LibraryMsQuicOpenVersionUnsupported,
+            "[ api] MsQuicOpenVersion, Only version 2 supported");
```
nibanks, line 1314: I think we should promote this to `QuicTraceEvent` with `LibraryError`. It's important.

thhous-msft: Agreed. Done

nibanks (approved):

PR #2363, opened 2022-02-04
Connection ID Prefixes

nibanks:
Updates the (currently unused) CID prefix logic to be per-listener instead of per registration. This allows for scenarios where multiple independent listeners need to use different prefixes for spreading/balancing load.

Also adds a new parameter on the client side for connections to specify the CID prefix explicitly, so it can target a particular listener.

Future PRs will pass this information down to the datapath so that it can allow for sharing on the same UDP port, where possible.


src/core/connection.h
```diff
@@ -461,6 +461,12 @@ typedef struct QUIC_CONNECTION {
     //
     QUIC_CID* OrigDestCID;
 
+    //
+    // An app configured prefix for all connection IDs.
+    //
+    uint8_t CidPrefixLength;
+    uint8_t* CidPrefix;
```
thhous-msft, line 468: An inline struct here storing the length and the 6 prefix bytes would actually have a smaller memory footprint here then the separate allocation, in addition to being cleaner because of no need to maintain the separate allocation. You could actually store 15 bytes of inline data in the amount of space this takes up.

anrossi: Why does this have 15 bytes of padding instead of 7? 7 bytes of padding would make the pointer aligned on an 8-byte boundary (on 64-bit), and the following array is aligned after the pointer.

thhous-msft: It has 7 bytes of padding, but then you have 8 bytes for the pointer itself. Thats where the 15 comes from. Its 15 bytes of data, not 15 bytes of padding.

thhous-msft: On 32 bit, you have 4 bytes of data and 3 bytes of padding. So you could even fit the 7 bits of data needed here in 32 bit as well. So its a win win in all cases.

thhous-msft (approved):

PR #2371, opened 2022-02-08
Make StreamReceiveComplete Not Fail

nibanks:
The only real failure was around failure to allocate the operation. This change updates the code to preallocate the receive complete operation as a part of the stream itself.


src/core/api.c
```diff
@@ -1276,17 +1265,13 @@ MsQuicStreamReceiveComplete(
     //
     // Queue the operation but don't wait for the completion.
     //
-    QuicConnQueueOper(Connection, Oper);
-    Status = QUIC_STATUS_SUCCESS;
+    QuicConnQueueOper(Connection, &Stream->ReceiveCompleteOperation);
```
nibanks, line 1268: The ref above doesn't get released. Need to figure out a solution for that.

src/inc/quic_platform_posix.h
```diff
@@ -175,6 +175,27 @@ InterlockedCompareExchange64(
...
+}
+
+inline
+void*
+InterlockedExchangePointer(
+    _Inout_ _Interlocked_operand_ void* volatile *Target,
+    _In_ void* Value
+    )
+{
+    return __sync_fetch_and_and(Target, (size_t)Value & 0);
```
nibanks, line 196: TODO: Find real solution for this.

src/core/stream_recv.c
```diff
@@ -807,6 +807,19 @@ QuicStreamRecvFlush(
             Event.RECEIVE.Flags |= QUIC_RECEIVE_FLAG_FIN; // TODO - 0-RTT flag?
         }
 
+        if (Stream->ReceiveCompleteOperation == NULL) {
```
thhous-msft, line 810: I have a gut feeling this needs to be an interlocked operation, or at least a read that forces a memory barrier. But maybe its fine. Just worrying since its written from another thread, and the other side is an atomic.

nibanks: I'm pretty sure it's fine. This logic **always** runs before the other side. This is before we even indicate the receive event to the app, only after which may they call the receive complete; either on the same thread or after queuing work to another thread, which would have its own memory barrier in the process. Ok?

thhous-msft (dismissed):

src/core/api.c
```diff
@@ -1247,22 +1245,14 @@ MsQuicStreamReceiveComplete(
         !Connection->State.HandleClosed);
 
     if (!Stream->Flags.Started || !Stream->Flags.ReceiveCallPending) {
-        Status = QUIC_STATUS_INVALID_STATE;
```
anrossi, line 1250: Should we log here so applications which might wonder why they their stream complete call isn't working have a way to find out?

src/core/api.c
```diff
@@ -1247,22 +1245,14 @@ MsQuicStreamReceiveComplete(
...
+    Oper = InterlockedFetchAndClearPointer((void**)&Stream->ReceiveCompleteOperation);
     if (Oper == NULL) {
-        Status = QUIC_STATUS_OUT_OF_MEMORY;
-        QuicTraceEvent(
-            AllocFailure,
-            "Allocation of '%s' failed. (%llu bytes)",
-            "STRM_RECV_COMPLETE operation",
-            0);
-        goto Exit;
+        goto Exit; // Duplicate calls to receive complete
```
anrossi, line 1253: I feel like this should also be a log point. Since it's kind of bad if an application calls receive complete on a stream twice.

anrossi (approved):

PR #2376, opened 2022-02-08
Async route resolution

csujedihy:
Design:
Try to resolve route in QuicSendQueueFlush. If route+arp look up is successful inline, go ahead and queue send flush. 
If not, there are 2 cases.
Case 1: We need to actually resolve the neighbor entry. Queue an operation to route worker to resolve neighbor entry. An extra connection ref is acquired in this case.
Case 2: Route+arp look up was unsuccessful inline, e.g. no route or no interface found for the destination. Queue an operation to route worker to kill the path.

The route worker loops over route operation queue and resolves neighbor entry. Note that for each operation it calls GetIpNetEntry2 first to see if neighbor entry is already there and if not, calls ResolveIpNetEntry2 to get it. Once done, make an upcall to QUIC, which queues a route completion and sets route state to resolved and resume sends.

Tested both route inline completion and threaded completion. It's working.


src/core/connection.c
```diff
@@ -3076,6 +3076,35 @@ QuicConnQueueUnreachable(
     }
 }
 
+_IRQL_requires_max_(DISPATCH_LEVEL)
+void
+QuicConnQueueRouteCompletion(
+    _In_ QUIC_CONNECTION* Connection,
+    _When_(Succeeded == FALSE, _In_opt_)
```
nibanks, line 3083: Use `_Reserved_` when it should be NULL.

src/core/connection.c
```diff
@@ -3076,6 +3076,35 @@ QuicConnQueueUnreachable(
...
+    _When_(Succeeded == TRUE, _In_)
```
nibanks, line 3084: Do `_In_reads_(...)` to indicate how many bytes

src/core/connection.c
```diff
@@ -5626,6 +5656,34 @@ QuicConnProcessUdpUnreachable(
...
+        QUIC_PATH* Path = &Connection->Paths[0];
+        CxPlatResolveRouteComplete(&Path->Route, PhysicalAddress);
+        QuicSendQueueFlush(&Connection->Send, REASON_ROUTE_COMPLETION);
+    } else {
+        QuicTraceLogConnInfo(
+            Unreachable,
+            Connection,
+            "Route resolution failure");
+        //
+        // Close the connection since the peer is unreachable.
```
nibanks, line 5677: What if this is just one path that's unreachable?

src/core/send.c
```diff
@@ -114,16 +114,58 @@ QuicSendCanSendFlagsNow(
...
+    if (Path->Route.RouteState == RouteResolving) {
+        //
+        // Can't send now. Once route resolution completes, we will resume sending.
+        //
+        return;
+    } else {
+        //
+        // Route resolved. We are allowed to flush sends.
+        //
+    }
```
nibanks, line 164: Just remove unnecessary `else`.

src/inc/quic_datapath.h
```diff
@@ -638,6 +665,19 @@ CxPlatSocketGetParam(
...
+    _Inout_ CXPLAT_ROUTE* Route,
+    _In_ const uint8_t* PhysicalAddress
+    );
+
+QUIC_STATUS
+CxPlatResolveRoute(
+    _In_ CXPLAT_SOCKET* Socket,
+    _Inout_ CXPLAT_ROUTE* Route,
+    _In_ VOID* Context
+    );
```
nibanks, line 679: Add some comments on these.

src/core/connection.c
csujedihy: What do you mean? The else path would be reached only if route/arp look-up failed fatally.

src/core/binding.c
```diff
@@ -1715,6 +1715,24 @@ QuicBindingUnreachable(
     }
 }
 
+_IRQL_requires_max_(DISPATCH_LEVEL)
+_Function_class_(CXPLAT_DATAPATH_ROUTE_RESOLUTION_CALLBACK)
+void
+QuicBindingRoute(
+    _In_ void* Context,
```
csujedihy, line 1722: _Inout_

src/core/connection.c
```diff
@@ -3076,6 +3076,35 @@ QuicConnQueueUnreachable(
...
+        QuicConnQueueOper(Connection, ConnOper);
+    } else {
+        QuicTraceEvent(
+            AllocFailure,
+            "Allocation of '%s' failed. (%llu bytes)",
+            "Route completion operation",
+            0);
+    }
+
+    QuicConnRelease(Connection, QUIC_CONN_REF_ROUTE);
```
csujedihy, line 3105: This releases the extra ref taken in QuicSendQueueFlush if ResolveRoute returns pending.

maolson-msft:
Why was the route resolution moved from SendDataAlloc to SendQueueFlush?

src/core/binding.c
```diff
@@ -1715,6 +1715,24 @@ QuicBindingUnreachable(
     }
 }
 
+_IRQL_requires_max_(DISPATCH_LEVEL)
+_Function_class_(CXPLAT_DATAPATH_ROUTE_RESOLUTION_CALLBACK)
+void
+QuicBindingRoute(
```
nibanks, line 1721: If the context is the connection, I think it's fine we eliminate `QuicBindingRoute` entirely and just pass `QuicConnQueueRouteCompletion` as the callback pointer.

src/core/connection.c
```diff
@@ -3076,6 +3076,35 @@ QuicConnQueueUnreachable(
...
+        if (Succeeded) {
+            memcpy(ConnOper->ROUTE.PhysicalAddress, PhysicalAddress, sizeof(ConnOper->ROUTE.PhysicalAddress));
+        }
+        QuicConnQueueOper(Connection, ConnOper);
+    } else {
+        QuicTraceEvent(
+            AllocFailure,
+            "Allocation of '%s' failed. (%llu bytes)",
+            "Route completion operation",
+            0);
```
nibanks, line 3102: We're permanently stuck if we hit this, right? Nothing will "kick" the connection to make progress? What then?

src/core/connection.c
```diff
@@ -5101,7 +5130,8 @@ QuicConnRecvPostProcessing(
 
     if (Packet->HasNonProbingFrame &&
         Packet->NewLargestPacketNumber &&
-        !(*Path)->IsActive) {
+        !(*Path)->IsActive &&
+        Connection->Paths[0].Route.RouteState == RouteResolved) {
```
nibanks, line 5134: Why add this here? What's going to retrigger this path after the resolution completes?

src/core/connection.c
```diff
@@ -5626,6 +5656,34 @@ QuicConnProcessUdpUnreachable(
...
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+QuicConnProcessRouteCompletion(
+    _In_ QUIC_CONNECTION* Connection,
+    _In_ const uint8_t* PhysicalAddress,
+    _In_ BOOLEAN Succeeded
+    )
+{
+    if (Succeeded) {
+        QUIC_PATH* Path = &Connection->Paths[0];
```
nibanks, line 5668: This code is too fragile to assume just `Paths[0]`. What if there are multiple parallel paths?

src/core/send.c
```diff
@@ -114,16 +114,58 @@ QuicSendCanSendFlagsNow(
...
     _In_ QUIC_SEND_FLUSH_REASON Reason
     )
 {
+    QUIC_CONNECTION* Connection = QuicSendGetConnection(Send);
+
+#ifdef QUIC_USE_RAW_DATAPATH
+
+    QUIC_PATH* Path = &Connection->Paths[0];
+    QUIC_STATUS Status;
+    if (Path->Route.RouteState == RouteUnresolved) {
```
nibanks, line 132: I feel we shouldn't wait until the send flush to queue the route resolution. Let's do it ASAP, in the receive path, the first time we see the path.

nibanks: Also, might be a good idea to wrap a lot of this with a `QuicPathResolveRoute` function.

src/inc/quic_datapath.h
```diff
@@ -141,6 +141,15 @@ typedef struct CXPLAT_SEND_DATA CXPLAT_SEND_DATA;
 //
 typedef struct QUIC_BUFFER QUIC_BUFFER;
 
+//
+// When state is Resolved, LocalLinkLayerAddress and NextHopLinkLayerAddress of CXPLAT_ROUTE are valid.
+//
+typedef enum CXPLAT_ROUTE_STATE {
```
nibanks, line 147: I believe we plan to have something like a `RouteSuspect` state very soon (cc @maolson-mstf). I think it'd be a good idea to define it now.

src/inc/quic_datapath.h
```diff
@@ -151,7 +160,7 @@ typedef struct CXPLAT_ROUTE {
 
     uint8_t LocalLinkLayerAddress[6];
     uint8_t NextHopLinkLayerAddress[6];
-    BOOLEAN Resolved; // When TRUE, LocalLinkLayerAddress and NextHopLinkLayerAddress are valid.
+    CXPLAT_ROUTE_STATE RouteState;
```
nibanks, line 163: It's in a "Route" object already. Just call it "State"?

src/core/send.c
maolson-msft: Premature optimization IMO.

src/inc/quic_platform.h
```diff
@@ -66,80 +66,83 @@ typedef struct CXPLAT_SLIST_ENTRY {
...
+#define QUIC_POOL_DEFAULT_COMPAT_VER_LIST       '14cQ' // Qc41 - QUIC Default compatible versions list
+#define QUIC_POOL_VERSION_INFO                  '24cQ' // Qc42 - QUIC Version info
+#define QUIC_POOL_PROCESS                       '34cQ' // Qc43 - QUIC Process
+#define QUIC_POOL_TLS_TMP_TP                    '44cQ' // Qc44 - QUIC Platform TLS Temporary TP storage
+#define QUIC_POOL_PCP                           '54cQ' // Qc45 - QUIC PCP
+#define QUIC_POOL_DATAPATH_ADDRESSES            '64cQ' // Qc46 - QUIC Datapath Addresses
+#define QUIC_POOL_TLS_TICKET_KEY                '74cQ' // Qc47 - QUIC Platform TLS ticket key
+#define QUIC_POOL_TLS_CIPHER_SUITE_STRING       '84cQ' // Qc48 - QUIC TLS cipher suite string
+#define QUIC_POOL_ROUTE_RESOLUTION_WORKER       '94cQ' // Qc49 - QUIC route resolution worker
+#define QUIC_POOL_ROUTE_RESOLUTION_OPERATION    'A4cQ' // Qc4A - QUIC route resolution operation
```
nibanks, line 144: I would have shortened `_OPERATION` to `_OPER` so I didn't have to increase the spacing.

src/platform/datapath_raw.c
```diff
@@ -17,6 +17,55 @@
...
+    };
+
+    Status = CxPlatThreadCreate(&ThreadConfig, &Worker->Thread);
+    if (QUIC_FAILED(Status)) {
+        QuicTraceEvent(
+            LibraryErrorStatus,
+            "[ lib] ERROR, %u, %s.",
+            Status,
+            "CxPlatThreadCreate");
+        goto Error;
```
nibanks, line 60: Leaking `Worker`

src/platform/datapath_raw.c
```diff
@@ -71,6 +125,30 @@ CxPlatDataPathInitialize(
...
+VOID
+CxPlatDataPathRouteWorkerUninitialize(
+    _Inout_ CXPLAT_DATAPATH* DataPath
+    )
+{
+    CXPLAT_ROUTE_RESOLUTION_WORKER* Worker = DataPath->RouteResolutionWorker;
+
+    Worker->Enabled = FALSE;
+    CxPlatEventSet(Worker->Ready);
+    CxPlatEventWaitForever(Worker->Done);
```
nibanks, line 138: You don't need a separate event to wait on. You can wait on the thread itself.

src/inc/quic_datapath.h
maolson-msft: I'd rather see that state added in the later PR that actually uses it.

src/platform/datapath_raw.c
```diff
@@ -485,6 +561,75 @@ CxPlatSocketGetParam(
...
+            CXPLAT_LIST_ENTRY Operations;
+            CxPlatListInitializeHead(&Operations);
+
+            CxPlatDispatchLockAcquire(&Worker->Lock);
+            if (!CxPlatListIsEmpty(&Worker->Operations)) {
+                CxPlatListMoveItems(&Worker->Operations, &Operations);
+            }
+            CxPlatDispatchLockRelease(&Worker->Lock);
+
+            while (CxPlatListIsEmpty(&Operations)) {
```
nibanks, line 581: This lock/list is not a super hot path and doesn't need all the extra complexity here. I think going with a simpler design of lock, pop, unlock, check it cleaner here.

src/platform/datapath_raw.c
```diff
@@ -485,6 +561,75 @@ CxPlatSocketGetParam(
...
+                CXPLAT_ROUTE_RESOLUTION_OPERATION* Operation =
+                    CXPLAT_CONTAINING_RECORD(
+                        CxPlatListRemoveHead(&Operations), CXPLAT_ROUTE_RESOLUTION_OPERATION, WorkerLink);
+                NETIO_STATUS Status =
+                Status = GetIpNetEntry2(&Operation->IpnetRow);
```
nibanks, line 586: I think we should have start/end events in this thread so we can track its state in the logs.

src/platform/datapath_raw.h
```diff
@@ -15,12 +15,52 @@ typedef struct CXPLAT_SOCKET_POOL {
...
+} CXPLAT_ROUTE_RESOLUTION_WORKER;
+
+typedef struct CXPLAT_ROUTE_RESOLUTION_OPERATION {
+    //
+    // Link in the worker's operation queue.
+    // N.B. Multi-threaded access, synchronized by worker's operation lock.
+    //
+    CXPLAT_LIST_ENTRY WorkerLink;
+    MIB_IPNET_ROW2 IpnetRow;
+    VOID* Context;
```
nibanks, line 53: We use `void` instead of the Windows `VOID`.

src/core/connection.c
csujedihy: I actually don't know what to do in this case because it takes an queue operation to kill a connection. In TCP, I'd just call the connection inline.

src/platform/datapath_raw_socket.c
```diff
@@ -265,10 +265,21 @@ CxPlatRemoveSocket(
...
 
+VOID
+CxPlatResolveRouteComplete(
+    _Inout_ CXPLAT_ROUTE* Route,
+    _In_ const uint8_t* PhysicalAddress
+    )
+{
+    CxPlatCopyMemory(&Route->NextHopLinkLayerAddress, PhysicalAddress, sizeof(Route->NextHopLinkLayerAddress));
+    Route->RouteState = RouteResolved;
+}
```
nibanks, line 276: I understand the desire to separate abstraction here, but IMO this is too much. No need to have a separate function for this when the caller has all the exposure already to do it themselves.

src/platform/datapath_raw_socket.c
```diff
@@ -265,10 +265,21 @@ CxPlatRemoveSocket(
...
+
 QUIC_STATUS
 CxPlatResolveRoute(
     _In_ CXPLAT_SOCKET* Socket,
-    _Inout_ CXPLAT_ROUTE* Route
+    _Inout_ CXPLAT_ROUTE* Route,
+    _In_ VOID* Context
```
nibanks, line 282: Another `VOID` instead of `void` here.

src/platform/datapath_raw_socket.c
```diff
@@ -278,8 +289,7 @@ CxPlatResolveRoute(
     CXPLAT_DBG_ASSERT(!QuicAddrIsWildCard(&Route->RemoteAddress));
 
     //
-    // Find the best next hop IP address. If not cached, this may result in (and
-    // block on) network IO.
+    // Find the best next hop IP address.
```
nibanks, line 292: Can't `GetBestRoute2` block on IO too? If so, it needs to be on the other thread too.

src/core/send.c
csujedihy: Nick, I don't understand your suggestion. In the receive path, the route is automatically resolved and we don't need to queue anything for route resolution.

src/inc/quic_datapath.h
nibanks: Ok.

src/platform/datapath_raw_socket.c
csujedihy: It does not block on IO.

src/platform/datapath_raw_socket.c
csujedihy: This function will be called in 3 different places (2 now). Worth the abstraction?

src/core/connection.c
nibanks: QUIC supports multiple parallel paths simultaneously, possibly on different interfaces. What if one path succeeds but another fails? We shouldn't kill the connection because of one path.

src/platform/datapath_raw_socket.c
nibanks: Still debatable for a 2 line function, but I'll leave it up to you.

src/core/send.c
csujedihy: I don't remember we decided to not trust L2 info. Why can't we trust it? I think we should trust it and refresh it when we see signals indicating it can't be trusted (like persistent loss etc).

src/platform/datapath_raw.c
csujedihy: Right, some copy-paste stuff that needs clean-up.

src/core/connection.c
csujedihy: Interesting. I have not groked the parallel paths logic. I wonder how it works with the current code (without this PR) or if it's already broken with the current design (blocking arp in send alloc). Switching interface from xdp to non-xdp is definitely broken.

src/core/binding.c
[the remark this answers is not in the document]
csujedihy: The naming is kinda messed up right now. I will clean it up in the next iteration.

src/core/connection.c
csujedihy: This part I have not figured out how to deal with TBH. Any suggestions?

csujedihy:
> Why was the route resolution moved from SendDataAlloc to SendQueueFlush?

If we do it in SendDataAlloc, the design would be much more complicated. SendDataAlloc happens right before we call into datapath. If at this point, route is still in resolving state, we would need to queue up frames in datapath without mac address filled. When route is resolved, then we would need to go over each queued frame to fill the mac address.

If we do it in SendQueueFlush, we don't have this problem because we build frames once we know what to fill in L2 headers.

src/core/connection.c
nibanks: I _think_ multiple XDP interfaces would still work as is. For the multi-path stuff, ping me and we can talk.

src/core/send.c
nibanks: @maolson-msft what are your recollections? I thought we talked about "Is it Ok to just use the L2 info when we receive a packet" and it was decided that the path back might be different.

maolson-msft: Yeah, the theory was that there could be asymmetrical routes where RX packets come from one gateway but TX packets go to a different gateway. Whether this is a likely scenario is questionable, but it's definitely a POSSIBLE scenario.

I really think we ought to, for starters, just do route resolution in the TX path. That's what we need to be functional and there are no questions about correctness.

maolson-msft: > > Why was the route resolution moved from SendDataAlloc to SendQueueFlush?
> 
> If we do it in SendDataAlloc, the design would be much more complicated. SendDataAlloc happens right before we call into datapath. If at this point, route is still in resolving state, we would need to queue up frames in datapath, which requires its own synchronization, without mac address filled. When route is resolved, then we would need to go over each queued frame to fill the mac address.
> 
> If we do it in SendQueueFlush, we don't have this problem because we build frames once we know what to fill in L2 headers.

OK, makes sense.

maolson-msft:
> > Why was the route resolution moved from SendDataAlloc to SendQueueFlush?
> 
> If we do it in SendDataAlloc, the design would be much more complicated. SendDataAlloc happens right before we call into datapath. If at this point, route is still in resolving state, we would need to queue up frames in datapath, which requires its own synchronization, without mac address filled. When route is resolved, then we would need to go over each queued frame to fill the mac address.
> 
> If we do it in SendQueueFlush, we don't have this problem because we build frames once we know what to fill in L2 headers.

OK, makes sense.

src/core/send.c
csujedihy: Makes sense. I think I will need to change how the route in path is being overwritten and ignore the state variable.

src/core/connection.c
nibanks: But that's not a fatal issue. If you fail to get the ICMP error, you can still move forward. This is a completion call that fails. I have an idea on how to fix this.

src/platform/datapath_raw.c
csujedihy: Will do this in the last iteration to avoid frequent sidecar churn.

src/core/connection.c
```diff
@@ -3076,6 +3075,50 @@ QuicConnQueueUnreachable(
...
+            0);
+        if (InterlockedCompareExchange16((short*)&Connection->BackUpOperUsed, 1, 0) == 0) {
+            QUIC_OPERATION* Oper = &Connection->BackUpOper;
+            Oper->FreeAfterProcess = FALSE;
+            Oper->Type = QUIC_OPER_TYPE_API_CALL;
+            Oper->API_CALL.Context = &Connection->BackupApiContext;
+            Oper->API_CALL.Context->Type = QUIC_API_TYPE_CONN_SHUTDOWN;
+            Oper->API_CALL.Context->CONN_SHUTDOWN.Flags = QUIC_CONNECTION_SHUTDOWN_FLAG_SILENT;
+            Oper->API_CALL.Context->CONN_SHUTDOWN.ErrorCode = QUIC_ERROR_INTERNAL_ERROR;
+            Oper->API_CALL.Context->CONN_SHUTDOWN.RegistrationShutdown = FALSE;
```
nibanks, line 3114: @thhous-msft Are we actually going to need to set this flag: `RegistrationShutdown`? I think this code is fine, but just wanted to double check.

src/core/connection.c
```diff
@@ -5626,6 +5670,54 @@ QuicConnProcessUdpUnreachable(
...
+    _In_ const uint8_t* PhysicalAddress,
+    _In_ uint8_t PathId,
+    _In_ BOOLEAN Succeeded
+    )
+{
+    uint8_t PathIndex;
+    QUIC_PATH* Path = QuicConnGetPathByID(Connection, PathId, &PathIndex);
+    if (Path != NULL) {
+        if (Succeeded) {
+            if (Path->IsActive) {
```
nibanks, line 5686: We shouldn't have this `if` just for the comment below. Please reorganize things so it's not necessary.

src/core/connection.c
```diff
@@ -5626,6 +5670,54 @@ QuicConnProcessUdpUnreachable(
...
+                //
+                // The route we just resolved is not active anymore. The active path that kicked out this path
+                // should be triggering route resolution. We store the route in the path in case it will be used
+                // in the future.
+                //
+                CxPlatResolveRouteComplete(Connection, &Path->Route, PhysicalAddress);
+            }
+        } else {
+            //
+            // Kill the path that failed route resolution and make the next path active if possible.
```
nibanks, line 5698: Testing all this stuff is going to be hard...

src/core/connection.c
```diff
@@ -5626,6 +5670,54 @@ QuicConnProcessUdpUnreachable(
...
+            //
+            if (Path->IsActive && Connection->PathsCount > 1) {
+                QuicPathSetActive(Connection, &Connection->Paths[1]);
+            }
+            QuicPathRemove(Connection, PathIndex);
+        }
+    }
+
+    if (Connection->PathsCount == 0) {
```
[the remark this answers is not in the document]
nibanks: And then probably `return` after it. And do the `QuicSendQueueFlush` unconditionally at the end? Or should it only be if active path is touched...?

src/platform/datapath_raw.c
```diff
@@ -17,6 +17,56 @@
...
+            Status,
+            "CxPlatThreadCreate");
+        goto Error;
+    }
+
+    DataPath->RouteResolutionWorker = Worker;
+
+Error:
+    if (QUIC_FAILED(Status) && Worker != NULL) {
+        CXPLAT_FREE(Worker, QUIC_POOL_ROUTE_RESOLUTION_WORKER);
```
nibanks, line 65: Leaking `Worker->Ready` in case of `CxPlatThreadCreate` failure.

src/core/connection.c
csujedihy: Oh, I forgot to clean this up.

src/core/connection.c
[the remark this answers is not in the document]
csujedihy: >> This if should only be if we call QuicPathRemove above.
There is another case where Path == NULL and PathsCount == 0 too.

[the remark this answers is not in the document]
csujedihy: > And do the QuicSendQueueFlush unconditionally at the end

There are 2 cases we need to kick off the send. 1) active route has been resolved successfully and 2) when we have made a path active in this function. Calling QuicSendQueueFlush doesn't seem to have any side-effects so I put it at the end to make the code cleaner.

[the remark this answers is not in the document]
nibanks: It's not the end of the world if you unnecessarily queue, but you really shouldn't queue if you can prevent it.

src/platform/datapath_raw.c
csujedihy: Fixed

src/platform/datapath_raw.c
csujedihy: Added traces

src/core/connection.c
```diff
@@ -5115,7 +5158,8 @@ QuicConnRecvPostProcessing(
 
     if (Packet->HasNonProbingFrame &&
         Packet->NewLargestPacketNumber &&
-        !(*Path)->IsActive) {
+        !(*Path)->IsActive &&
+        Connection->Paths[0].Route.State == RouteResolved) {
```
nibanks, line 5162: I think we want to switch regardless. We just received the packet from here, we need to switch.

src/core/connection.c
```diff
@@ -5656,6 +5700,55 @@ QuicConnProcessUdpUnreachable(
...
+    _In_ BOOLEAN Succeeded
+    )
+{
+    uint8_t PathIndex;
+    QUIC_PATH* Path = QuicConnGetPathByID(Connection, PathId, &PathIndex);
+    if (Path != NULL) {
+        if (Succeeded) {
+            CxPlatResolveRouteComplete(Connection, &Path->Route, PhysicalAddress, PathId);
+            QuicSendQueueFlush(&Connection->Send, REASON_ROUTE_COMPLETION);
+            QuicTraceLogConnInfo(
```
nibanks, line 5718: Put the log at the beginning of the scope so we see it before queuing the flush.

src/core/connection.c
```diff
@@ -5656,6 +5700,55 @@ QuicConnProcessUdpUnreachable(
...
+            //
+            // Kill the path that failed route resolution and make the next path active if possible.
+            //
+            if (Path->IsActive && Connection->PathsCount > 1) {
+                QuicTraceLogConnInfo(
+                    FailedRouteResolution,
+                    Connection,
+                    "Processing failed route completion Path[%hhu]",
+                    PathId);
+                QuicPathSetActive(Connection, &Connection->Paths[1]);
```
nibanks, line 5733: `QuicPathSetActive` will change the order of the paths in the array. So it will invalidate `PathIndex` which you are using below when you call `QuicPathRemove`.

src/inc/quic_datapath.h
```diff
@@ -151,10 +160,13 @@ typedef struct CXPLAT_ROUTE {
...
     uint8_t LocalLinkLayerAddress[6];
     uint8_t NextHopLinkLayerAddress[6];
-    BOOLEAN Resolved; // When TRUE, LocalLinkLayerAddress and NextHopLinkLayerAddress are valid.
     void* Queue;
+
+    CXPLAT_ROUTE_STATE State; // Keep this as the last property in the struct.
+
 } CXPLAT_ROUTE;
 
+
```
nibanks, line 169: nit: please remove extra newline

src/core/connection.c
csujedihy: this is some leftover.

src/core/connection.c
csujedihy: It's exactly what the code expects. In that case, QuicPathRemove removes the path that just got swapped out.

src/core/connection.c
csujedihy: fixed

src/platform/datapath_raw.c
```diff
@@ -485,6 +563,74 @@ CxPlatSocketGetParam(
...
+        CxPlatListMoveItems(&Worker->Operations, &Operations);
+    }
+    CxPlatDispatchLockRelease(&Worker->Lock);
+
+    while (!CxPlatListIsEmpty(&Operations)) {
+        CXPLAT_ROUTE_RESOLUTION_OPERATION* Operation =
+            CXPLAT_CONTAINING_RECORD(
+                CxPlatListRemoveHead(&Operations), CXPLAT_ROUTE_RESOLUTION_OPERATION, WorkerLink);
+        Operation->Callback(Operation->Context, NULL, Operation->PathId, FALSE);
+        CXPLAT_FREE(Operation, QUIC_POOL_ROUTE_RESOLUTION_OPER);
```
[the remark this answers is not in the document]
csujedihy: pool allocator now

src/core/connection.h
```diff
@@ -1403,6 +1404,22 @@ QuicConnQueueUnreachable(
...
+// Queues a route completion event to a connection for processing.
+//
+_IRQL_requires_max_(DISPATCH_LEVEL)
+_Function_class_(CXPLAT_ROUTE_RESOLUTION_CALLBACK)
+void
+QuicConnQueueRouteCompletion(
+    _Inout_ QUIC_CONNECTION* Connection,
+    _When_(Succeeded == FALSE, _Reserved_)
+    _When_(Succeeded == TRUE, _In_)
+    _In_reads_bytes_(6)
```
nibanks, line 1417:
```suggestion
    _When_(Succeeded == TRUE, _In_reads_bytes_(6))
```
The SAL isn't quite right. You'll need to update this everywhere I think.

src/inc/quic_datapath.h
```diff
@@ -638,6 +649,48 @@ CxPlatSocketGetParam(
...
     );
 
+//
+// Copies L2 address into route object and sets route state to resolved.
+//
+void
+CxPlatResolveRouteComplete(
+    _In_ void* Connection,
+    _Inout_ CXPLAT_ROUTE* Route,
+    _In_ const uint8_t* PhysicalAddress,
```
nibanks, line 659: _In_reads_(6)

src/inc/quic_datapath.h
```diff
@@ -638,6 +649,48 @@ CxPlatSocketGetParam(
...
+// Function pointer type for datapath route resolution callbacks.
+//
+typedef
+_IRQL_requires_max_(DISPATCH_LEVEL)
+_Function_class_(CXPLAT_ROUTE_RESOLUTION_CALLBACK)
+void
+(CXPLAT_ROUTE_RESOLUTION_CALLBACK)(
+    _In_ void* Context,
+    _When_(Succeeded == FALSE, _Reserved_)
+    _When_(Succeeded == TRUE, _In_)
```
nibanks, line 673: _In_reads_(6)

src/platform/datapath_raw_socket.c
```diff
@@ -335,52 +383,44 @@ CxPlatResolveRoute(
...
-        goto Done;
+        Operation->IpnetRow = IpnetRow;
+        Operation->Context = Context;
+        Operation->Callback = Callback;
+        Operation->PathId = PathId;
+        CxPlatDispatchLockAcquire(&Worker->Lock);
+        CxPlatListInsertTail(&Worker->Operations, &Operation->WorkerLink);
+        CxPlatDispatchLockRelease(&Worker->Lock);
+        CxPlatEventSet(Worker->Ready);
+        Status = ERROR_IO_PENDING;
```
nibanks, line 415: We have a `QUIC_STATUS_PENDING`.

nibanks (approved):
Looks pretty good. I wish we had tests, but we don't have any for XDP stuff yet. We really need to figure out how to remedy that...

src/platform/datapath_raw_socket.c
csujedihy: Yeah, but right before we return status, we convert it to hresult. `return HRESULT_FROM_WIN32(Status);`

src/core/connection.c
thhous-msft: Its not the whole registration being shutdown, we should be fine here.

PR #2386, opened 2022-02-11
QUIC_STATISTICS_V2

nibanks:
Flattens the statistics API structs in an effort to make them less fragile to adding new ones in the future.


nibanks:
I think we're going to have to make this v2. Because HTTP.sys directly exposes our v1 struct, we cannot break it. :(

thhous-msft (approved):

PR #2389, opened 2022-02-11
QuicStreamSendFlush: Don't double start the stream

TYoungSL:
This changes the behavior of `StreamSend` when supplied with the `QUIC_SEND_FLAG_START` flag.

With this change, the stream will only be started once and not subsequently if it has already been started.

This avoids an unwanted scenario that follows when multiple stream `QUIC_STREAM_EVENT_START_COMPLETE` events are raised that risk undefined application or implementation specific behavior.

This does not change the behavior of multiple calls to `StreamStart`.

### DIFF
--- a/src/test/lib/EventTest.cpp
+++ b/src/test/lib/EventTest.cpp
@@ -1313,6 +1313,131 @@ QuicTestValidateStreamEvents7(
     } // Connections scope
 }
 
+
+void
+QuicTestValidateStreamEvents8(
+    _In_ MsQuicRegistration& Registration,
+    _In_ HQUIC Listener,
+    _In_ QuicAddr& ServerLocalAddr
+    )
+{
+    TestScopeLogger ScopeLogger(__FUNCTION__);
+
+    MsQuicSettings Settings;
+    Settings.SetPeerBidiStreamCount(1).SetMinimumMtu(1280).SetMaximumMtu(1280);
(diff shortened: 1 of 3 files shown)


nibanks (dismissed):

anrossi:
Could you add a test to test this new scenario?

nibanks (dismissed):

TYoungSL:
Updated `QuicTestValidateStreamEvents6` to show it not double starting.

Do any additional tests or scenarios need to be covered?

TYoungSL:
Added `QuicTestValidateStreamEvents8`, (a modified copy of `QuicTestValidateStreamEvents3`) that has the explicit call to `StreamStart` followed by 2 `StreamSend` calls with the start flag.

Together with the changes to `QuicTestValidateStreamEvents6` should cover the expected scenarios.

src/test/lib/EventTest.cpp
```diff
@@ -1306,6 +1313,131 @@ QuicTestValidateStreamEvents7(
...
+            0));
+
+    TEST_TRUE(Client.Complete.WaitTimeout(2000));
+    TEST_TRUE(Server.Complete.WaitTimeout(1000));
+
+    } // Stream scope
+    } // Connections scope
+}
+
 void QuicTestValidateStreamEvents(uint32_t Test)
```
nibanks, line 1441: Needs an update to `ValidateStreamEventArgs::Generate()`.

TYoungSL: 👍

nibanks (approved):

PR #2390, opened 2022-02-11
Some code comment cleanup/additions

maolson-msft:
Minor comment tweaks that I made while refreshing myself on the quic core code.

### DIFF
--- a/src/core/binding.c
+++ b/src/core/binding.c
@@ -1401,13 +1401,8 @@ QuicBindingDeliverDatagrams(
     // be used for partitioning the look up table.
     //
     // For long header packets for server owned bindings, the packet's DestCid
-    // was not necessarily generated locally, so cannot be used for routing.
-    // Instead, a hash of the tuple and source connection ID (SourceCid) is
-    // used.
-    //
-    // The exact type of lookup table associated with the binding varies on the
-    // circumstances, but it allows for quick and easy lookup based on DestCid
-    // (when used).
+    // was not necessarily generated locally, so cannot be used for lookup.
+    // Instead, a hash of the remote address/port and source CID is used.
     //
     // If the lookup fails, and if there is a listener on the local 2-Tuple,
     // then a new connection is created and inserted into the binding's lookup
--- a/src/core/loss_detection.c
+++ b/src/core/loss_detection.c
@@ -9,10 +9,16 @@ Abstract:
     have been lost or delivered to the peer.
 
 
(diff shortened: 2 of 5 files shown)


src/core/binding.c
```diff
@@ -1401,13 +1401,8 @@ QuicBindingDeliverDatagrams(
     // be used for partitioning the look up table.
     //
     // For long header packets for server owned bindings, the packet's DestCid
-    // was not necessarily generated locally, so cannot be used for routing.
-    // Instead, a hash of the tuple and source connection ID (SourceCid) is
-    // used.
-    //
-    // The exact type of lookup table associated with the binding varies on the
```
maolson-msft, line 1408: Removed this para because it seems to be redundant with the previous one; please let me know if there's some key info that was added by this para.

src/core/loss_detection.c
```diff
@@ -9,10 +9,16 @@
...
-    RACK (time-based loss detection) algorithm:
+    A combination of two similar loss detection algorithms is used:
 
-    An unacknowledged packet sent before an acknowledged packet and
-    sent more than QUIC_TIME_REORDER_THRESHOLD ago is assumed lost.
+    RACK (time-based reordering threshold) algorithm:
+        An unacknowledged packet sent before an acknowledged packet and
+        sent more than QUIC_TIME_REORDER_THRESHOLD ago is assumed lost.
+
+    FACK (packet-number-based reordering threshold) algorithm:
```
maolson-msft, line 18: The existing comment mentioned only RACK and made me wonder if we'd one RACK-only LD in here, which we didn't, so I added this mention of FACK.

nibanks:
@maolson-msft everything looks good except (I assume) you accidentally changed the submodules. Can you please fix them?

maolson-msft:
> @maolson-msft everything looks good except (I assume) you accidentally changed the submodules. Can you please fix them?

@nibanks Oh, I noticed those in git status but didn't think they would be included in the commit. How do I "update" those submodules or whatever? They keep appearing as changed when I pull or switch branches or something.

nibanks:
> > @maolson-msft everything looks good except (I assume) you accidentally changed the submodules. Can you please fix them?
> 
> @nibanks Oh, I noticed those in git status but didn't think they would be included in the commit. How do I "update" those submodules or whatever? They keep appearing as changed when I pull or switch branches or something.

@thhous-msft can you help here?

nibanks (approved):

Issue #2374, opened 2022-02-08
Allow simple configuration of supported TLS1.3 groups


baentsch:
### Describe the feature you'd like supported

TLS1.3 permits announcing algorithm groups used during key exchange via the [supported_groups extension](https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.7). 

The feature requested is to permit setting these groups for `msquic` applications to different values via some simple (command line or environment variable) manner to allow easy adaptation to different group (lists).

msquic currently fixes this list to 4 groups as per [this code](https://github.com/microsoft/msquic/blob/692be2372f94b82ff2e0567ecbbf631a4bbf0a2c/src/platform/tls_openssl.c#L1173-L1176).

### Proposed solution

An example how it may be implemented is the [openssl -groups](https://beta.openssl.org/docs/manmaster/man3/SSL_CONF_cmd.html) parameter.

### Additional context

While the chosen msquic [preference list](https://github.com/microsoft/msquic/blob/692be2372f94b82ff2e0567ecbbf631a4bbf0a2c/src/platform/tls_openssl.c#L154) is somewhat understandable from a classic cryptography point of view, some users may want to request stronger classic algorithms, e.g., by leaving away weaker algorithms like "P-256". In the context of [quantum safe cryptography](https://github.com/open-quantum-safe/openssl#key-exchange) where many different algorithms could be chosen by way of specifying different group names, the issue is even more acute.

nibanks:
This would require updating the [QUIC_CREDENTIAL_CONFIG](https://github.com/microsoft/msquic/blob/main/src/inc/msquic.h#L288) to allow the app to pass the allowable groups down. It'd be nice to support a model similar to how we do `QUIC_ALLOWED_CIPHER_SUITE_FLAGS` but I think the number of groups might be too big.

baentsch:
> It'd be nice to support a model similar to how we do QUIC_ALLOWED_CIPHER_SUITE_FLAGS but I think the number of groups might be too big.

[...]

PR #2393, opened 2022-02-12
Remove Setting OpenSSL Groups List

nibanks:
There is no need to overwrite whatever defaults OpenSSL already has. First part of #2374. cc @baentsch


anrossi:
What are the OpenSSL defaults? Does quicTLS change these in any way?

nibanks:
I don't know the default. I also don't know the default for schannel. Quictls doesn't change them.

PR #2398, opened 2022-02-14
Print cmsg headers on macos if fre assert is going to occur

thhous-msft:
I want to either figure out why we're not getting a cmsg here, or decide we just want to drop the packet rather than asserting. I have a feeling we're going to end up just dropping, but if I can figure out the problem I'd rather do that.

### DIFF
--- a/src/platform/datapath_kqueue.c
+++ b/src/platform/datapath_kqueue.c
@@ -22,6 +22,7 @@ Environment:
 #include <sys/sysctl.h>
 #include <sys/time.h>
 #include <sys/types.h>
+#include <stdio.h>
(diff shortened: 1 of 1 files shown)


nibanks (approved):

src/platform/datapath_kqueue.c
```diff
@@ -22,6 +22,7 @@
 #include <sys/sysctl.h>
 #include <sys/time.h>
 #include <sys/types.h>
+#include <stdio.h>
```
nibanks, line 25: Remove?

nibanks (approved):

nibanks (approved):

PR #2399, opened 2022-02-14
Split public settings structs

thhous-msft:
The settings struct was getting very polluted, both with settings that only mattered to specific areas, and settings that needed a slightly different API. This splits settings into 3 structs. Global only, general, and versions.

Internally, the goal is to keep everything using a unified internal struct to make this change easier. In the future, this can be split into individual components internally as well.


src/core/configuration.c
```diff
@@ -52,12 +52,6 @@ MsQuicConfigurationOpen(
         goto Error;
     }
 
-    if (Settings != NULL &&
```
nibanks, line 55: Why no size check any more?

src/core/connection.c
```diff
@@ -43,18 +43,37 @@ typedef struct QUIC_RECEIVE_PROCESSING_STATE {
...
+    _In_ const QUIC_SETTINGS_INTERNAL* NewSettings
+    );
+
 _IRQL_requires_max_(PASSIVE_LEVEL)
 BOOLEAN
 QuicConnApplyNewSettings(
     _In_ QUIC_CONNECTION* Connection,
     _In_ BOOLEAN OverWrite,
     _In_ BOOLEAN CopyExternalToInternal,
-    _In_range_(FIELD_OFFSET(QUIC_SETTINGS, DesiredVersionsList), UINT32_MAX)
```
nibanks, line 52: Why remove this?

src/inc/msquic.h
```diff
@@ -496,30 +525,27 @@ typedef struct QUIC_SETTINGS {
...
             uint64_t CongestionControlAlgorithm             : 1;
-            uint64_t RESERVED                               : 29;
+            uint64_t RESERVED                               : 33;
         } IsSet;
     };
 
     uint64_t MaxBytesPerKey;
     uint64_t HandshakeIdleTimeoutMs;
     uint64_t IdleTimeoutMs;
+    uint64_t MtuDiscoverySearchCompleteTimeoutUs;
```
nibanks, line 548: Can we keep the `IsSet` flags in the same order?

src/inc/msquic.h
```diff
@@ -639,8 +659,9 @@ void
 #define QUIC_PARAM_GLOBAL_LOAD_BALACING_MODE            0x01000002  // uint16_t - QUIC_LOAD_BALANCING_MODE
 #define QUIC_PARAM_GLOBAL_PERF_COUNTERS                 0x01000003  // uint64_t[] - Array size is QUIC_PERF_COUNTER_MAX
 #define QUIC_PARAM_GLOBAL_SETTINGS                      0x01000004  // QUIC_SETTINGS
-#define QUIC_PARAM_GLOBAL_VERSION                       0x01000005  // uint32_t[4]
-#define QUIC_PARAM_GLOBAL_DESIRED_VERSIONS              0x01000006  // uint32_t[]
+#define QUIC_PARAM_GLOBAL_ONLY_SETTINGS                 0x01000005  // QUIC_GLOBAL_SETTINGS
```
nibanks, line 662: I think I'd prefer `QUIC_PARAM_GLOBAL_GLOBAL_SETTINGS`.

src/cs/lib/msquic_generated.cs
```diff
@@ -2243,8 +2243,8 @@ public static unsafe partial class MsQuic
         [NativeTypeName("#define QUIC_PARAM_CONN_TLS_SECRETS 0x05000013")]
         public const int QUIC_PARAM_CONN_TLS_SECRETS = 0x05000013;
 
-        [NativeTypeName("#define QUIC_PARAM_CONN_DESIRED_VERSIONS 0x14000014")]
-        public const int QUIC_PARAM_CONN_DESIRED_VERSIONS = 0x14000014;
+        [NativeTypeName("#define QUIC_PARAM_CONN_VERSION_SETTINGS 0x14000014")]
+        public const int QUIC_PARAM_CONN_VERSION_SETTINGS = 0x14000014;
```
nibanks, line 2247: Will also need the define for global settings.

src/test/lib/TestHelpers.h
```diff
@@ -59,15 +59,19 @@ struct ServerAcceptContext {
 
 struct ClearGlobalVersionListScope {
     ~ClearGlobalVersionListScope() {
-        MsQuicSettings ClearVNSettings;
-        ClearVNSettings.SetDesiredVersionsList(nullptr, 0);
+        QUIC_VERSION_SETTINGS Settings = {0};
```
nibanks, line 62: Use `MsQuicVersionSettings` instead?

src/core/configuration.c
thhous-msft: I want to forward all size checks to QuicSettingsSetSettings. That way they only exist in 1 spot. Need to finish up the work doing that.

src/inc/msquic.h
```diff
@@ -473,6 +473,35 @@ typedef enum QUIC_PERFORMANCE_COUNTERS {
...
+            uint64_t VersionNegotiationExtEnabled           : 1;
+            uint64_t RESERVED                               : 62;
+        } IsSet;
+    };
+
+    const uint32_t* DesiredVersionsList;
+    uint32_t DesiredVersionsListLength;
+    uint8_t VersionNegotiationExtEnabled    : 1;
+    uint8_t RESERVED                        : 7;
+} QUIC_VERSION_SETTINGS;
```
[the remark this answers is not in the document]
thhous-msft: I was thinking about that last night, and I'll just move VN Ext enabled back to the main settings struct.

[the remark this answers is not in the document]
anrossi: With the changes for the VN draft, we probably want to make version negotiation a get/set param instead of a settings thing. Pending further discussion.

src/core/connection.c
thhous-msft: Anything using QUIC_SETTTINGS_INTERNAL does not need a size.

src/inc/msquic.h
thhous-msft: Done.

src/core/unittest/SettingsTest.cpp
```diff
@@ -14,41 +14,43 @@
...
-        ASSERT_EQ(
-            QUIC_STATUS_SUCCESS,
-            QuicSettingsGetParam(
-                &IncomingSettings,
-                &BufferSize,
-                reinterpret_cast<QUIC_SETTINGS*>(Buffer)));
-        ASSERT_EQ(sizeof(IncomingSettings), BufferSize);
-    }
-}
+// TODO Fix this test
```
nibanks, line 17: Is this planned to be fixed in this PR?

src/cs/lib/msquic_generated.cs
```diff
@@ -213,11 +215,11 @@ public unsafe partial struct QUIC_CREDENTIAL_CONFIG
 
         public QUIC_CREDENTIAL_FLAGS Flags;
 
-        [NativeTypeName("QUIC_CREDENTIAL_CONFIG::(anonymous union at src/inc/msquic.h:274:5)")]
+        [NativeTypeName("QUIC_CREDENTIAL_CONFIG::(anonymous union at src/inc/msquic.h:279:5)")]
```
nibanks, line 218: a bit annoying that it embeds line numbers.

src/inc/msquic.h
```diff
@@ -517,6 +517,38 @@ typedef enum QUIC_PERFORMANCE_COUNTERS {
...
+        } IsSet;
+    };
+
+    //
+    // Flags first so we can use offset to get versioned length of struct.
+    //
+    uint8_t VersionNegotiationExtEnabled    : 1;
+    uint8_t RESERVED                        : 7;
+    const uint32_t* DesiredVersionsList;
+    uint32_t DesiredVersionsListLength;
```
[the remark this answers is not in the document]
nibanks: I guess we could also do `const uint32_t* Versions` but I do like the idea of a contiguous buffer to make "get" simple.

scripts/generate-dotnet.ps1
```diff
@@ -19,4 +19,9 @@ $LicenseHeader = Join-Path $RootDir src cs LicenseHeader.txt
 
 ClangSharpPInvokeGenerator -f $MsQuicHeader -n Microsoft.Quic -o $MsQuicGeneratedSource -m MsQuic -l msquic `
     -c exclude-enum-operators -r _SOCKADDR_INET=QuicAddr -c generate-macro-bindings -h $LicenseHeader `
-    -e QUIC_UINT62_MAX
+    -e QUIC_UINT62_MAX -e MsQuicOpen2 -e QUIC_API_VERSION_1 -D QUIC_API_ENABLE_INSECURE_FEATURES
```
nibanks, line 22: What does the `-e` mean?

thhous-msft: Exclude. Makes it so those definitions are not included.

src/inc/msquic.h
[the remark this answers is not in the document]
thhous-msft: That makes it so this struct can never be extended. Which if we move the VN extension flag back into settings, I'd probably be ok with.

src/cs/lib/msquic_generated.cs
thhous-msft: Yeah. I plan on looking into some changes here (Because I also have to post process to remove the full path). But thats out of scope for sure.

src/inc/msquic.h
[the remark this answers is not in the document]
nibanks: Yeah, I'm not necessarily opposed to making this struct set in stone, so long as we can get it right... One option is we wrap it in some kind of `QUIC_PREVIEW_FEATURES` define to allow us to modify it in the future just in case the version negotiation draft changes before it goes RFC. Thoughts?

src/test/bin/quic_gtest.cpp
```diff
@@ -280,10 +280,11 @@ TEST_P(WithValidateStreamEventArgs, ValidateStreamEvents) {
 
 TEST(ParameterValidation, ValidateDesiredVersionSettings) {
     TestLogger Logger("QuicTestDesiredVersionSettings");
+    // TODO Fix these tests
```
nibanks, line 283: Just putting a comment here to make sure this isn't forgotten.

thhous-msft: Makes sense. This is the one I need to discuss the API scope with you and Anthony, and if VN is going to change any more. As I don't want to rewrite this, just to have it rewritten again in a week.

src/tools/reach/reach.cpp
```diff
@@ -13,7 +13,6 @@
 #include <mutex>
 #include <algorithm>
 
-#define QUIC_LEGACY_COMPILE_MODE 1 // To test legacy compiling
```
nibanks, line 16: Also remove the `#ifdef QUIC_API_VERSION_2` code please.

src/core/unittest/SettingsTest.cpp
thhous-msft: Done, added a bunch more tests too.

src/core/precomp.h
```diff
@@ -17,6 +17,7 @@
 #pragma warning(disable:28931) // Unused Assignment
 
 #define QUIC_API_ENABLE_INSECURE_FEATURES 1
+#define QUIC_PREVIEW_FEATURES 1
```
nibanks, line 20: What about matching the format above and do:
```suggestion
#define QUIC_API_ENABLE_PREVIEW_FEATURES 1
```

scripts/generate-dotnet.ps1
```diff
@@ -19,4 +19,9 @@ $LicenseHeader = Join-Path $RootDir src cs LicenseHeader.txt
 
 ClangSharpPInvokeGenerator -f $MsQuicHeader -n Microsoft.Quic -o $MsQuicGeneratedSource -m MsQuic -l msquic `
```
nibanks, line 20: Do we want to enable the "preview" features flag for this or not?

nibanks (dismissed):

anrossi (approved):

PR #2405, opened 2022-02-18
Route refresh on persistent congestion

csujedihy:
Design:
- When we hit a persistent congestion event, we mark the route state to be suspected.
- Before we flush sends, if we see route state is suspected, we will call into CxPlatResolveRoute where we will queue up an operation to refresh it later in the route worker. The current route is stored on the route operation and will be looked at later.
- In route worker, if we dequeue a refresh operation, we will unconditionally query route (new) and check if it is the same as the current route (old). 
  - If new route has different local address than the old route, we will fail this operation and kill the path (when path count goes down to zero, the connection will be killed).
  - If new route carries different MAC address, that means some other connections have already refreshed the route so we can simply just use it.
  - Otherwise, force neighbor discovery.

In this PR, we have also slightly changed the initial route resolution behavior. So now, it queries route twice- first in send flush routine and then in route worker. As long as local address doesn't change, the new route will be used.


src/platform/datapath_raw.h
```diff
@@ -348,3 +349,10 @@ CxPlatFramingWriteHeaders(
     _In_ BOOLEAN SkipNetworkLayerXsum,
     _In_ BOOLEAN SkipTransportLayerXsum
     );
+
+_IRQL_requires_max_(PASSIVE_LEVEL)
+QUIC_STATUS
+CxPlatQueryRoute(
+    _In_ const CXPLAT_SOCKET* Socket,
+    _Inout_ CXPLAT_ROUTE* Route
+    );
```
csujedihy, line 358: nl

nibanks:
I think it'd be best to rebase this on top of main after we merge dpdk branch into main.

src/platform/datapath_raw.c
```diff
@@ -592,29 +592,62 @@ CXPLAT_THREAD_CALLBACK(CxPlatRouteResolutionWorkerThread, Context)
...
-                    ResolveIpNetEntry2(&Operation->IpnetRow, NULL);
-                if (Status != 0) {
+            CXPLAT_DBG_ASSERT(
+                Operation->Route.State == RouteRefreshing || Operation->Route.State == RouteResolving);
+            CXPLAT_ROUTE NewRoute = Operation->Route;
+            QUIC_STATUS Status = CxPlatQueryRoute(Operation->Socket, &NewRoute);
+            if (Status == QUIC_STATUS_SUCCESS) {
+                if (!QuicAddrCompare(&NewRoute.LocalAddress, &Operation->Route.LocalAddress)) {
+                    //
+                    // We can't handle local address change here easily due to lack of full migration support.
```
nibanks, line 602: Probably worth a log.

src/platform/datapath_raw.c
```diff
@@ -592,29 +592,62 @@ CXPLAT_THREAD_CALLBACK(CxPlatRouteResolutionWorkerThread, Context)
...
+                    //
+                    Status = QUIC_STATUS_INVALID_STATE;
+                } else if (Operation->Route.State == RouteRefreshing &&
+                           memcmp(NewRoute.NextHopLinkLayerAddress,
+                                  Operation->Route.NextHopLinkLayerAddress,
+                                  sizeof(NewRoute.NextHopLinkLayerAddress)) == 0) {
+                    //
+                    // We are handling route refresh here. We will force neighbor discovery because net hop address
```
nibanks, line 610:
```suggestion
                    // We are handling route refresh here. We will force neighbor discovery because next hop address
```

nibanks: Also, please limit comments to 80 chars

src/platform/datapath_raw_socket.c
```diff
@@ -372,27 +369,72 @@ CxPlatResolveRoute(
     // Map the next hop IP address to a link-layer address.
     //
     MIB_IPNET_ROW2 IpnetRow = {0};
-    IpnetRow.InterfaceLuid = IpforwardRow.InterfaceLuid;
+    IpnetRow.InterfaceIndex = IpforwardRow.InterfaceIndex;
```
nibanks, line 372: Why change? I believe I used Luid because it was more efficient than index, which eventually has to be converted to Luid anyways.

src/core/operation.h
```diff
@@ -228,7 +228,7 @@ typedef struct QUIC_OPERATION {
             QUIC_STATELESS_CONTEXT* Context;
         } STATELESS; // Stateless reset, retry and VN
         struct {
-            uint8_t PhysicalAddress[6];
+            CXPLAT_ROUTE Route;
```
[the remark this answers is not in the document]
csujedihy: It does increase the size of the QUIC_OPERATION. Yeah, we don't actually need the entire route. Just MAC and interface pointer should be enough.

src/platform/datapath_raw_socket.c
csujedihy: Luid is indeed more efficient. I changed it because it would make the plumbing easier. We now store interface pointer in route object and interface object has an ifindex field. We need interface object anyway for assigning DP queue. So, we don't need to plumb Luid for route operation and CxPlatQueryRoute doesn't need to output a Luid.

csujedihy: fixed

src/platform/datapath_raw.c
```diff
@@ -589,32 +589,63 @@ CXPLAT_THREAD_CALLBACK(CxPlatRouteResolutionWorkerThread, Context)
...
+                                  sizeof(NewRoute.NextHopLinkLayerAddress)) == 0) {
+                    //
+                    // We are handling route refresh here. We will force neighbor discovery because next hop address
+                    // has not changed which implies this is the first time we are refreshing it. 
+                    //
+                    Status = QUIC_STATUS_PENDING;
+                }
+            }
+
+            if (Status == QUIC_STATUS_PENDING) {
```
maolson-msft, line 618: I was very confused about this at first because it looked like only the RouteRefreshing case would ever get into here. Now I see that you are returning PENDING from CxPlatQueryRoute when GetIpNetEntry2 doesn't return a result. Please don't use PENDING that way- PENDING is almost always used when a function kicks off some asynchronous work that you need to wait for completion of, which isn't the case here afaict.

src/platform/datapath_raw.c
```diff
@@ -589,32 +589,63 @@ CXPLAT_THREAD_CALLBACK(CxPlatRouteResolutionWorkerThread, Context)
...
-            NETIO_STATUS Status =
-            Status = GetIpNetEntry2(&Operation->IpnetRow);
-            if (Status != ERROR_SUCCESS || Operation->IpnetRow.State <= NlnsIncomplete) {
-                Status =
-                    ResolveIpNetEntry2(&Operation->IpnetRow, NULL);
-                if (Status != 0) {
+            CXPLAT_DBG_ASSERT(
+                Operation->Route.State == RouteRefreshing || Operation->Route.State == RouteResolving);
+            CXPLAT_ROUTE NewRoute = Operation->Route;
+            QUIC_STATUS Status = CxPlatQueryRoute(Operation->Socket, &NewRoute, &IpnetRow);
```
maolson-msft, line 599: Can you remind me why we now also have a call of GetBestRoute2 here in the worker thread? Why can't the call be just in CxPlatResolveRoute? I thought that was what we had discussed.

src/inc/quic_datapath.h
```diff
@@ -141,12 +141,11 @@ typedef struct CXPLAT_SEND_DATA CXPLAT_SEND_DATA;
 //
 typedef struct QUIC_BUFFER QUIC_BUFFER;
 
-//
-// When state is Resolved, LocalLinkLayerAddress and NextHopLinkLayerAddress of CXPLAT_ROUTE are valid.
-//
 typedef enum CXPLAT_ROUTE_STATE {
     RouteUnresolved,
     RouteResolving,
+    RouteSuspected,
```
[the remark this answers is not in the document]
csujedihy: The difference is for initial resolution, we don't care if next hop mac changes or not. So, they can't be the same state.

src/platform/datapath_raw_socket.c
```diff
@@ -371,28 +370,73 @@ CxPlatResolveRoute(
...
+    } else {
+        CxPlatCopyMemory(
+            &Route->NextHopLinkLayerAddress,
+            IpnetRow->PhysicalAddress,
+            sizeof(Route->NextHopLinkLayerAddress));
+    }
+
+Done:
+    if (Status > 0) {
+        return SUCCESS_HRESULT_FROM_WIN32(Status);
```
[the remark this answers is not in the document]
csujedihy: HRESULT_FROM_WIN32 doesn't work for status > 0 because it set the most significant bit to one.

src/platform/datapath_raw.c
csujedihy: It's cleaner for refresh case because we only need to call CxPlatQueryRoute once. We don't call CxPlatQueryRoute in ResolveRoute when we try to refresh the route. No more double-check GetIpNetEntry2. We could also skip calling CxPlatQueryRoute for initial resolution in the worker but resolving initial route in the worker is a rare case IMO.

src/platform/datapath_raw.c
csujedihy: makes sense. let me see if we can find a better status code.

src/platform/datapath_raw_socket.c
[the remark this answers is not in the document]
csujedihy: HRESULT_FROM_WIN32(ERROR_IO_PENDING) != QUIC_STATUS_PENDING

nibanks (commented):
You seem to be adding a lot of code I removed when I cleaned up the DPDK branch this morning.

nibanks:
> You seem to be adding a lot of code I removed when I cleaned up the DPDK branch this morning.

Actually, I'm just confused at the current diff. I think it might be best to start from scratch. 😦

PR #2407, opened 2022-02-18
Route refresh on persistent congestion

csujedihy:
Design:
- When we hit a persistent congestion event, we mark the route state to be suspected.
- Before we flush sends, if we see route state is suspected, we will call into CxPlatResolveRoute where we will try to re-resolve route inline. 
  - If GetBestRoute2 returns different local address, fail this path.
  - If GetNetIpEntry returns a different next hop address, queue an operation to the route worker to resolve route asynchronously.
  - Otherwise, take whatever re-resolving route gives.
- In route worker, simply call ResolveIpNetEntry2 to resolve it (code unchanged).

### DIFF
--- a/src/inc/quic_datapath.h
+++ b/src/inc/quic_datapath.h
@@ -145,7 +145,6 @@ typedef enum CXPLAT_ROUTE_STATE {
     RouteUnresolved,
     RouteResolving,
     RouteSuspected,
-    RouteRefreshing,
     RouteResolved,
 } CXPLAT_ROUTE_STATE;
 
@@ -659,7 +658,7 @@ void
 CxPlatResolveRouteComplete(
     _In_ void* Connection,
     _Inout_ CXPLAT_ROUTE* Route,
-    _In_ const CXPLAT_ROUTE* NewRoute,
+    _In_reads_bytes_(6) const uint8_t* PhysicalAddress,
     _In_ uint8_t PathId
     );
 
@@ -673,8 +672,8 @@ void
 (CXPLAT_ROUTE_RESOLUTION_CALLBACK)(
     _In_ void* Context,
     _When_(Succeeded == FALSE, _Reserved_)
-    _When_(Succeeded == TRUE, _In_)
-        const CXPLAT_ROUTE* Route,
+    _When_(Succeeded == TRUE, _In_reads_bytes_(6))
+        uint8_t* PhysicalAddress,
     _In_ uint8_t PathId,
     _In_ BOOLEAN Succeeded
     );
--- a/src/inc/quic_datapath.h
+++ b/src/inc/quic_datapath.h
@@ -159,7 +159,6 @@ typedef struct CXPLAT_ROUTE {
 #ifdef QUIC_USE_RAW_DATAPATH
     uint8_t LocalLinkLayerAddress[6];
     uint8_t NextHopLinkLayerAddress[6];
-    void* Interface;
     void* Queue;
 
     CXPLAT_ROUTE_STATE State; // Keep this as the last property in the struct.
--- a/src/inc/quic_datapath.h
+++ b/src/inc/quic_datapath.h
@@ -141,12 +141,11 @@ typedef struct CXPLAT_SEND_DATA CXPLAT_SEND_DATA;
 //
 typedef struct QUIC_BUFFER QUIC_BUFFER;
 
-//
-// When state is Resolved, LocalLinkLayerAddress and NextHopLinkLayerAddress of CXPLAT_ROUTE are valid.
-//
 typedef enum CXPLAT_ROUTE_STATE {
     RouteUnresolved,
     RouteResolving,
+    RouteSuspected,
+    RouteRefreshing,
     RouteResolved,
 } CXPLAT_ROUTE_STATE;
 
@@ -161,6 +160,7 @@ typedef struct CXPLAT_ROUTE {
 #ifdef QUIC_USE_RAW_DATAPATH
     uint8_t LocalLinkLayerAddress[6];
     uint8_t NextHopLinkLayerAddress[6];
+    void* Interface;
     void* Queue;
 
     CXPLAT_ROUTE_STATE State; // Keep this as the last property in the struct.
@@ -659,7 +659,7 @@ void
 CxPlatResolveRouteComplete(
     _In_ void* Connection,
     _Inout_ CXPLAT_ROUTE* Route,
-    _In_reads_bytes_(6) const uint8_t* PhysicalAddress,
+    _In_ const CXPLAT_ROUTE* NewRoute,
     _In_ uint8_t PathId
     );
 
@@ -673,8 +673,8 @@ void
 (CXPLAT_ROUTE_RESOLUTION_CALLBACK)(
     _In_ void* Context,
     _When_(Succeeded == FALSE, _Reserved_)
-    _When_(Succeeded == TRUE, _In_reads_bytes_(6))
-        uint8_t* PhysicalAddress,
+    _When_(Succeeded == TRUE, _In_)
+        const CXPLAT_ROUTE* Route,
     _In_ uint8_t PathId,
     _In_ BOOLEAN Succeeded
     );
--- a/src/platform/datapath_raw.c
+++ b/src/platform/datapath_raw.c
@@ -589,32 +589,63 @@ CXPLAT_THREAD_CALLBACK(CxPlatRouteResolutionWorkerThread, Context)
         CxPlatDispatchLockRelease(&Worker->Lock);
 
         while (!CxPlatListIsEmpty(&Operations)) {
+            MIB_IPNET_ROW2 IpnetRow = {0};
             CXPLAT_ROUTE_RESOLUTION_OPERATION* Operation =
                 CXPLAT_CONTAINING_RECORD(
                     CxPlatListRemoveHead(&Operations), CXPLAT_ROUTE_RESOLUTION_OPERATION, WorkerLink);
-            NETIO_STATUS Status =
-            Status = GetIpNetEntry2(&Operation->IpnetRow);
-            if (Status != ERROR_SUCCESS || Operation->IpnetRow.State <= NlnsIncomplete) {
-                Status =
-                    ResolveIpNetEntry2(&Operation->IpnetRow, NULL);
-                if (Status != 0) {
+            CXPLAT_DBG_ASSERT(
+                Operation->Route.State == RouteRefreshing || Operation->Route.State == RouteResolving);
+            CXPLAT_ROUTE NewRoute = Operation->Route;
+            QUIC_STATUS Status = CxPlatQueryRoute(Operation->Socket, &NewRoute, &IpnetRow);
+            if (Status == QUIC_STATUS_SUCCESS) {
+                if (!QuicAddrCompare(&NewRoute.LocalAddress, &Operation->Route.LocalAddress)) {
+                    //
+                    // We can't handle local address change here easily due to lack of full migration support.
+                    //
+                    Status = QUIC_STATUS_INVALID_STATE;
+                } else if (Operation->Route.State == RouteRefreshing &&
+                           memcmp(NewRoute.NextHopLinkLayerAddress,
+                                  Operation->Route.NextHopLinkLayerAddress,
+                                  sizeof(NewRoute.NextHopLinkLayerAddress)) == 0) {
+                    //
+                    // We are handling route refresh here. We will force neighbor discovery because next hop address
+                    // has not changed which implies this is the first time we are refreshing it. 
+                    //
+                    Status = QUIC_STATUS_PENDING;
+                }
+            }
+
+            if (Status == QUIC_STATUS_PENDING) {
+                Status = ResolveIpNetEntry2(&IpnetRow, NULL);
+                if (Status == 0) {
+                    CxPlatCopyMemory(
+                        NewRoute.NextHopLinkLayerAddress,
+                        IpnetRow.PhysicalAddress,
+                        sizeof(NewRoute.NextHopLinkLayerAddress));
+                    Status = QUIC_STATUS_SUCCESS;
+                } else {
                     QuicTraceEvent(
                         DatapathErrorStatus,
                         "[data][%p] ERROR, %u, %s.",
                         Operation,
                         Status,
                         "ResolveIpNetEntry2");
-                    Operation->Callback(
-                        Operation->Context, NULL, Operation->PathId, FALSE);
-                } else {
-                    Operation->Callback(
-                        Operation->Context, Operation->IpnetRow.PhysicalAddress, Operation->PathId, TRUE);
                 }
-                CxPlatPoolFree(&Worker->OperationPool ,Operation);
+            }
+
+            //
+            // Now that we have the result for the route resolution, we can queue
+            // a route completion event to QUIC core to continue processing.
+            //
+            if (Status == QUIC_STATUS_SUCCESS) {
+                Operation->Callback(
+                    Operation->Context, &NewRoute, Operation->PathId, TRUE);
             } else {
                 Operation->Callback(
-                    Operation->Context, Operation->IpnetRow.PhysicalAddress, Operation->PathId, TRUE);
+                    Operation->Context, NULL, Operation->PathId, FALSE); 
             }
+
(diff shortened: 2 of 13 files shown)


src/inc/quic_datapath.h
```diff
@@ -141,12 +141,11 @@ typedef struct CXPLAT_SEND_DATA CXPLAT_SEND_DATA;
...
 typedef struct QUIC_BUFFER QUIC_BUFFER;
 
-//
-// When state is Resolved, LocalLinkLayerAddress and NextHopLinkLayerAddress of CXPLAT_ROUTE are valid.
-//
 typedef enum CXPLAT_ROUTE_STATE {
     RouteUnresolved,
     RouteResolving,
+    RouteSuspected,
+    RouteRefreshing,
```
maolson-msft, line 148: Can Refreshing be combined with Resolving?

src/platform/datapath_raw_socket.c
```diff
@@ -371,28 +370,72 @@ CxPlatResolveRoute(
...
-    // First call GetIpNetEntry2 to see if there's already a cached neighbor. If there
-    // isn't one, or if the cached neighbor's state is unreachable (which, NB, can happen
-    // in the case where a route lookup resulted in a dummy neighbor entry being created
-    // in TCPIP.sys) or incomplete, then queue up a solicitation event.
+    // Call GetIpNetEntry2 to see if there's already a cached neighbor.
     //
-    Status = GetIpNetEntry2(&IpnetRow);
+    Status = GetIpNetEntry2(IpnetRow);
+    if (Status != ERROR_SUCCESS || IpnetRow->State <= NlnsIncomplete) {
+        Status = ERROR_IO_PENDING;
```
maolson-msft, line 385: This is an unidiomatic use of PENDING (because there's no actual pending operation that has been kicked off) and confused me when I was reading CxPlatRouteResolutionWorkerThread- please change

src/platform/datapath_raw_socket.c
```diff
@@ -371,28 +370,72 @@ CxPlatResolveRoute(
...
+    } else {
+        CxPlatCopyMemory(
+            &Route->NextHopLinkLayerAddress,
+            IpnetRow->PhysicalAddress,
+            sizeof(Route->NextHopLinkLayerAddress));
+    }
+
+Done:
+    if (Status > 0) {
+        return SUCCESS_HRESULT_FROM_WIN32(Status);
```
maolson-msft, line 395: I hope this can be removed when PENDING is changed to some failure status.

src/platform/datapath_raw_socket.c
```diff
@@ -371,28 +370,72 @@ CxPlatResolveRoute(
...
+        Status = QUIC_STATUS_PENDING;
+    } else {
+        Status = CxPlatQueryRoute(Socket, &RouteQueried, &IpnetRow);
+    }
+
+    if (Status == QUIC_STATUS_SUCCESS) {
+        CxPlatResolveRouteComplete(Context, Route, &RouteQueried, PathId);
+    } else if (Status == QUIC_STATUS_PENDING) {
+        //
+        // We need to queue up an route operation for either route refresh or initial route resolution.
```
maolson-msft, line 437:
```suggestion
        // We need to queue up a route operation for either route refresh or initial route resolution.
```

src/platform/datapath_raw_socket.c
csujedihy: how about QUIC_STATUS_CONTINUE (ERROR_CONTINUE)?

src/platform/datapath_raw_socket.c
```diff
@@ -371,28 +370,72 @@ CxPlatResolveRoute(
...
+        "Starting to look up neighbor on Path[%hhu] with status %d",
         PathId,
         Status);
-    if (Status != ERROR_SUCCESS || IpnetRow.State <= NlnsIncomplete) {
+
+    if (State == RouteSuspected) {
+        //
+        // We have seen hints that the route might be bad. Let the route worker take care of this.
+        //
+        Status = QUIC_STATUS_PENDING;
```
maolson-msft, line 428: Why don't you call CxPlatQueryRoute in this case? It might be that the new neighbor has already been resolved by some other connection, in which case we could do everything inline, right?

src/platform/datapath_raw.c
```diff
@@ -589,32 +589,63 @@ CXPLAT_THREAD_CALLBACK(CxPlatRouteResolutionWorkerThread, Context)
...
-            NETIO_STATUS Status =
-            Status = GetIpNetEntry2(&Operation->IpnetRow);
-            if (Status != ERROR_SUCCESS || Operation->IpnetRow.State <= NlnsIncomplete) {
-                Status =
-                    ResolveIpNetEntry2(&Operation->IpnetRow, NULL);
-                if (Status != 0) {
+            CXPLAT_DBG_ASSERT(
+                Operation->Route.State == RouteRefreshing || Operation->Route.State == RouteResolving);
+            CXPLAT_ROUTE NewRoute = Operation->Route;
+            QUIC_STATUS Status = CxPlatQueryRoute(Operation->Socket, &NewRoute, &IpnetRow);
```
maolson-msft, line 599: Can you remind me why we now also have a call of GetBestRoute2 here in the worker thread? Why can't the call be just in CxPlatResolveRoute? I thought that was what we had discussed.

maolson-msft: In general, the ResolveIpNetEntry2 call (which is really the only reason we are queuing to a work item at all) is the "rare" case- because in the common single-gateway case, only the first connection will need to call this, and then every other connection will use the cached result (and because the persistent congestion case is really also a "rare" case IMO). So I think any duplicate calls in the worker thread are not a major perf consideration.

src/platform/datapath_raw_socket.c
csujedihy: We could but we'd have this "check if it's the same logic" in two places.

src/platform/datapath_raw_socket.c
maolson-msft: That sounds better, but still not immediately obvious, so let's add a code comment about what the status indicates to the caller.

src/platform/datapath_raw_socket.c
maolson-msft: I'm not sure what you mean. Is there a typo in your response?

maolson-msft: My point here is that "refreshing" and "resolving" the route should be nearly identical codepaths.

csujedihy: => We could but we'd need to add "check if it's the same neighbor" logic in two places.

src/platform/datapath_raw.c
csujedihy: I don't really care about perf at all for both cases actually. I just want to avoid having "checking if Xxx is the same" in both ResolveRoute and worker thread. I'd rather make everything async- no inline resolution at all.

csujedihy: Not a perf consideration. I try to avoid duplicated code in both places.

src/platform/datapath_raw_socket.c
csujedihy: Nearly identical but still different. What are you suggesting here? Just do the same thing for both refresh and initial resolution?

maolson-msft: Yes, unless there's a strong reason not to. I mean think about it. We have basically the same goal in both cases: look up a new next hop MAC address. So why should the code paths be so divergent?

src/platform/datapath_raw_socket.c
csujedihy: We can't use ERROR_CONTINUE unless we want that if (Status > 0) logic because ERROR_CONTINUE is positive too like PENDING.

csujedihy: I think NOT_FOUND is better. It means next hop not found

src/platform/datapath_raw.c
csujedihy: Fixed as we discussed offline.

src/platform/datapath_raw.c
```diff
@@ -610,11 +610,12 @@ CXPLAT_THREAD_CALLBACK(CxPlatRouteResolutionWorkerThread, Context)
                     Operation->Callback(
                         Operation->Context, Operation->IpnetRow.PhysicalAddress, Operation->PathId, TRUE);
                 }
-                CxPlatPoolFree(&Worker->OperationPool ,Operation);
             } else {
                 Operation->Callback(
                     Operation->Context, Operation->IpnetRow.PhysicalAddress, Operation->PathId, TRUE);
             }
+
+            CxPlatPoolFree(&Worker->OperationPool ,Operation);
```
maolson-msft, line 618: This is a bugfix, right? It isn't mentioned in the PR description.

src/platform/datapath_raw.h
```diff
@@ -340,3 +340,12 @@ CxPlatFramingWriteHeaders(
     _In_ BOOLEAN SkipNetworkLayerXsum,
     _In_ BOOLEAN SkipTransportLayerXsum
     );
+
+_IRQL_requires_max_(PASSIVE_LEVEL)
```
maolson-msft, line 344: This declaration can be removed.

src/platform/datapath_raw_socket.c
```diff
@@ -392,7 +407,25 @@ CxPlatResolveRoute(
...
+        //
+        Status = ERROR_IO_PENDING;
+        CXPLAT_STATIC_ASSERT(
+            SUCCESS_HRESULT_FROM_WIN32(ERROR_IO_PENDING) == QUIC_STATUS_PENDING,
+            "SUCCESS_HRESULT_FROM_WIN32(ERROR_IO_PENDING) must be QUIC_STATUS_PENDING");
+    } else {
+        CxPlatResolveRouteComplete(Context, Route, IpnetRow.PhysicalAddress, PathId);
+    }
+
+    if (Status == ERROR_IO_PENDING) {
```
maolson-msft, line 428: Again an unidiomatic use of "pending". We can just put this block of code above where status is currently set to pending.

src/platform/datapath_raw.c
thhous-msft: To add on, formatting is incorrect (Comma is in the wrong spot).

src/platform/datapath_raw_socket.c
```diff
@@ -371,7 +390,6 @@ CxPlatResolveRoute(
     //
     // Map the next hop IP address to a link-layer address.
     //
-    MIB_IPNET_ROW2 IpnetRow = {0};
```
maolson-msft, line 374: This IpnetRow variable was moved to the top of the function, but the general QUIC code style is to declare variables as late as possible.

src/platform/datapath_raw_socket.c
```diff
@@ -408,21 +441,23 @@ CxPlatResolveRoute(
         Operation->Context = Context;
         Operation->Callback = Callback;
         Operation->PathId = PathId;
+        Route->State = RouteResolving;
         CxPlatDispatchLockAcquire(&Worker->Lock);
         CxPlatListInsertTail(&Worker->Operations, &Operation->WorkerLink);
```
maolson-msft, line 446: We should have a code comment here explaining why we suddenly go through this hypercomplicated code path when we discover we need to send an NS. Just something like "Neighbor solicitation requires an exchange of network packets. To keep from blocking the entire connection work queue, which is shared with other connections, queue an operation on the dedicated route resolution worker to do the solicitation."

src/platform/datapath_raw_socket.c
csujedihy: something leftover during undoing the previous change.

src/platform/datapath_raw_socket.c
csujedihy: done

src/platform/datapath_raw.c
csujedihy: yes

src/platform/datapath_raw_socket.c
```diff
@@ -392,7 +407,22 @@ CxPlatResolveRoute(
         "Starting to look up neighbor on Path[%hhu] with status %u",
         PathId,
         Status);
-    if (Status != ERROR_SUCCESS || IpnetRow.State <= NlnsIncomplete) {
+    //
+    //
```
maolson-msft, line 411:
```suggestion
```

src/inc/quic_datapath.h
```diff
@@ -161,6 +159,7 @@ typedef struct CXPLAT_ROUTE {
 #ifdef QUIC_USE_RAW_DATAPATH
     uint8_t LocalLinkLayerAddress[6];
     uint8_t NextHopLinkLayerAddress[6];
+    void* Interface;
```
maolson-msft, line 162: This is now unused, right?

src/platform/datapath_raw_socket.c
```diff
@@ -392,7 +407,21 @@ CxPlatResolveRoute(
         "Starting to look up neighbor on Path[%hhu] with status %u",
         PathId,
         Status);
-    if (Status != ERROR_SUCCESS || IpnetRow.State <= NlnsIncomplete) {
+    //
+    // We need to force neighbor solicitation (NS) if any of the following is true:
+    // 1. No neighbor entry for the given destination address.
+    // 2. The neighbor entry isn't in usuable state.
```
maolson-msft, line 413:
```suggestion
    // 2. The neighbor entry isn't in a usable state.
```

src/platform/datapath_raw_socket.c
```diff
@@ -392,7 +407,21 @@ CxPlatResolveRoute(
         "Starting to look up neighbor on Path[%hhu] with status %u",
         PathId,
         Status);
-    if (Status != ERROR_SUCCESS || IpnetRow.State <= NlnsIncomplete) {
+    //
+    // We need to force neighbor solicitation (NS) if any of the following is true:
+    // 1. No neighbor entry for the given destination address.
```
maolson-msft, line 412:
```suggestion
    // 1. No cached neighbor entry for the given destination address.
```

src/platform/datapath_raw_socket.c
```diff
@@ -392,7 +407,21 @@ CxPlatResolveRoute(
...
+    // 2. The neighbor entry isn't in usuable state.
+    // 3. In refresh scenario, the neighbor entry is the same as the existing one.
```
maolson-msft, line 414: The term "refresh" isn't well-established in the code. You've only used the word here and in a trace that says "Refreshed route has different local address", without defining the term anywhere.

src/platform/datapath_raw_socket.c
```diff
@@ -392,7 +407,21 @@ CxPlatResolveRoute(
...
+    //
+    // We queue an operation on route worker for NS because it involves network IO and
```
maolson-msft, line 416:
```suggestion
    // We queue an operation on the route worker for NS because it involves network IO and
```

src/platform/datapath_raw_socket.c
```diff
@@ -408,11 +437,15 @@ CxPlatResolveRoute(
         Operation->Context = Context;
         Operation->Callback = Callback;
         Operation->PathId = PathId;
+        Route->State = RouteResolving;
         CxPlatDispatchLockAcquire(&Worker->Lock);
         CxPlatListInsertTail(&Worker->Operations, &Operation->WorkerLink);
         CxPlatDispatchLockRelease(&Worker->Lock);
         CxPlatEventSet(Worker->Ready);
         Status = ERROR_IO_PENDING;
+        CXPLAT_STATIC_ASSERT(
```
maolson-msft, line 446: You know, above we already set the Status variable to a QUIC_STATUS_XXX value (QUIC_STATUS_INVALID_STATE).

src/platform/datapath_raw_socket.c
csujedihy: ok, not more "refreshed".

src/platform/datapath_raw_socket.c
csujedihy: I can remove it. I put it here mainly because it was a little special because you can't convert it using `HRESULT_FROM_WIN32`.

csujedihy: Now, the code works so it's not needed anymore probably.

maolson-msft (approved):

PR #2409, opened 2022-02-19
Disable Shared EC Tests

nibanks:
For some reason the shared EC tests are taking longer than normal EC tests. I need to investigate that, but for the time being I don't want CI failures, so I am disabling them.


thhous-msft (approved):

.azure/azure-pipelines.ci.yml
```diff
@@ -673,10 +673,10 @@ stages:
     parameters:
       image: windows-2022
       platform: windows
-      tls: openssl
+      tls: schannel
```
thhous-msft, line 676: Schannel EC builds are not built, so more needs to be fixed.

nibanks: I could have sworn I built both... :|

nibanks:
Seems the schannel shared EC was no better. Need to figure out a path forward.

thhous-msft (approved):

PR #2432, opened 2022-02-24
[C#] Adds Flags attribute to generated _FLAGS enums

TYoungSL:
Updates the C# code generation step to add `[Flags]` attributes to the `_FLAGS` enums.


nibanks (approved):

TYoungSL:
Needs `using System;` at the top of the generated file

thhous-msft (approved):

PR #2445, opened 2022-03-01
Support CIBIR Extension

nibanks:
## Description

Implements the [quic-cibir](https://datatracker.ietf.org/doc/html/draft-banks-quic-cibir) extension draft. Also wraps the listener and connection parameters in `QUIC_API_ENABLE_PREVIEW_FEATURES` since this stuff still might change in the future, but we'd like to get the API changes in v2 ASAP.

## Testing

Existing tests will cover the "no CIBIR encoding used" cases, but additional tests need to be added for actually using it.


docs/Settings.md
```diff
@@ -116,7 +116,7 @@ These parameters are accessed by calling [GetParam](./api/GetParam.md) or [SetPa
 |-------------------------------------------|---------------------------|-----------|-----------------------------------------------------------|
 | `QUIC_PARAM_LISTENER_LOCAL_ADDRESS`<br> 0 | QUIC_ADDR                 | Get-only  | Get the full address tuple the server is listening on.    |
 | `QUIC_PARAM_LISTENER_STATS`<br> 1         | QUIC_LISTENER_STATISTICS  | Get-only  | Get statistics specific to this Listener instance.        |
-| `QUIC_PARAM_LISTENER_CID_PREFIX`<br> 2    | uint8_t[]  | Both  | CID prefix prepended to all CIDs.        |
+| `QUIC_PARAM_LISTENER_CIBIR_ID`<br> 2    | QUIC_CIBIR_ID  | Both  | The CIBIR well-known idenfitier.        |
```
thhous-msft, line 119: We should throw this behind preview features if its going to change.

src/inc/msquic.h
```diff
@@ -639,6 +639,15 @@ typedef struct QUIC_TLS_SECRETS {
     uint8_t ServerTrafficSecret0[QUIC_TLS_SECRETS_MAX_SECRET_LEN];
 } QUIC_TLS_SECRETS;
 
+//
+// This struct abstracts the well-known identifier used for the QUIC CIBIR
+// extension.
+//
+typedef struct QUIC_CIBIR_ID {
```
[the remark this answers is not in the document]
thhous-msft: Byte array to avoid the need for a variable length struct.

docs/Settings.md
nibanks: Done: https://github.com/microsoft/msquic/pull/2447

thhous-msft (approved):

PR #2449, opened 2022-03-01
Static assert that malloc alignment is correct

thhous-msft:
We had a test run with an alignment error. Testing to see if any of our test platforms actually would have the error.

### DIFF
--- a/src/inc/quic_platform_posix.h
+++ b/src/inc/quic_platform_posix.h
@@ -311,7 +311,11 @@ CxPlatFree(
 
 typedef struct CXPLAT_LOCK {
 
-    alignas(16) pthread_mutex_t Mutex;
+    //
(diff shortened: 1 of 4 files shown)


src/inc/quic_platform_posix.h
```diff
@@ -311,7 +311,11 @@ CxPlatFree(
 
 typedef struct CXPLAT_LOCK {
 
-    alignas(16) pthread_mutex_t Mutex;
+    //
+    // Mutex and condition. The alignas is important, as the perf tanks
+    // if the event is not aligned to the size of 2 pointers.
+    //
+    alignas(sizeof(void*) * 2) pthread_mutex_t Mutex;
```
anrossi, line 318: this will be 8 bytes on 32-bit. Is that ok?

anrossi (approved):

PR #2454, opened 2022-03-02
Improve crash dump debugging in CI

thhous-msft:
Debugging a dump on linux or mac at all requires downloading the dump, and corresponding symbols. Add code to print the stacks of the failures, which often can be used in many cases to avoid needing to do this.

Additionally, from the logs folder, deducing if the log is from a sanitize run or not is difficult. Solve this by adding the extra artifacts dir to the log folder

Do not merge this in its current state, as 2 tests are purposely failing (1 with a bugcheck, 1 with a segfault) to verify everything is working correctly.

### DIFF
--- a/scripts/log.ps1
+++ b/scripts/log.ps1
@@ -196,11 +196,15 @@ function Log-Stop {
             Write-Debug "Decoding LTTng into BabelTrace format ($BableTraceFile)"
             babeltrace --names all $TempDir/* > $BableTraceFile
 
-            Write-Host "Decoding into human-readable text: $ClogOutputDecodeFile"
-            $Command = "$Clog2Text_lttng -i $BableTraceFile -s $SideCar -o $ClogOutputDecodeFile --showTimestamp --showCpuInfo"
-            Write-Host $Command
-            Invoke-Expression $Command | Write-Debug
-            Remove-Item -Path $BableTraceFile -Force | Out-Null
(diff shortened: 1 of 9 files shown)


nibanks (approved):

scripts/log.ps1
```diff
@@ -196,11 +196,15 @@ function Log-Stop {
...
-            Remove-Item -Path $BableTraceFile -Force | Out-Null
+            Write-Host "Babeltrace ran. Compile clog2text and run the following command"
+            $Command = "clog2text_lttng -i $BableTraceFile -s $SideCar -o $ClogOutputDecodeFile --showTimestamp --showCpuInfo"
+            $Command
+
+            # Write-Host "Decoding into human-readable text: $ClogOutputDecodeFile"
+            # $Command = "$Clog2Text_lttng -i $BableTraceFile -s $SideCar -o $ClogOutputDecodeFile --showTimestamp --showCpuInfo"
+            # Write-Host $Command
+            # Invoke-Expression $Command | Write-Debug
+            # Remove-Item -Path $BableTraceFile -Force | Out-Null
```
nibanks, line 207: Undo this?

thhous-msft: Nope. Clog2text_lttng was broken by the clog changes a few months ago. Theres currently no setup to build and install clog2text_lttng on the test machines. I need to file another bug to fix that, but its a bigger task.

PR #2455, opened 2022-03-02
Replace 32 bit kernel msquic with disabled library

thhous-msft:
VS 2022's WDK will not support 32 bit kernel binaries (https://docs.microsoft.com/en-us/windows-hardware/drivers/installing-preview-versions-wdk). Additionally, Windows internally does not build 32 bit media anymore. However, because of how the internal build works, 32 bit binaries are still needed for ingesting. For Ni and newer, I'm proposing this change.

I want to make sure there are no internal dependencies on 32 bit kernel mode binaries, before we end up having to mess with the build system to work around not being able to build 32 bit binaries.


src/bin/winkernel/driver32.c
```diff
@@ -0,0 +1,81 @@
...
+    Main entry point to the MsQuic.sys 32 bit driver.
+
+    This driver does not work, and will always return not supported
+
+--*/
+
+#include "quic_platform.h"
+
+INITCODE DRIVER_INITIALIZE DriverEntry;
+PAGEDX EVT_WDF_DRIVER_UNLOAD EvtDriverUnload;
```
nibanks, line 17: This is unused here, and the other file.

src/bin/winkernel/driver32.c
```diff
@@ -0,0 +1,81 @@
...
+INITCODE
+_Function_class_(DRIVER_INITIALIZE)
+_IRQL_requires_same_
+_IRQL_requires_(PASSIVE_LEVEL)
+NTSTATUS
+DriverEntry(
+    _In_ PDRIVER_OBJECT DriverObject,
+    _In_ PUNICODE_STRING RegistryPath
+    )
+/*++
```
nibanks, line 28: Go ahead and remove all these function comments since they're unnecessary.

thhous-msft:
Replaced by #2462

PR #2457, opened 2022-03-02
Add CIBIR Support to SecNetPerf

nibanks:
## Description

Adds a new option `-cibir:<bytes>` to secnetperf for specifying the CIBIR extension's well-known identifier.

## Testing

Tested by manually running over localhost.

### DIFF
--- a/src/perf/lib/HpsClient.cpp
+++ b/src/perf/lib/HpsClient.cpp
@@ -72,7 +72,8 @@ HpsClient::Init(
     }
 
     const char* target;
-    if (!TryGetValue(argc, argv, "target", &target)) {
+    if (!TryGetValue(argc, argv, "target", &target) &&
+        !TryGetValue(argc, argv, "server", &target)) {
         WriteOutput("Must specify '-target' argument!\n");
         PrintHelp();
         return QUIC_STATUS_INVALID_PARAMETER;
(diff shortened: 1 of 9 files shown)


src/perf/lib/HpsClient.cpp
```diff
@@ -72,7 +72,8 @@ HpsClient::Init(
     }
 
     const char* target;
-    if (!TryGetValue(argc, argv, "target", &target)) {
+    if (!TryGetValue(argc, argv, "target", &target) &&
+        !TryGetValue(argc, argv, "server", &target)) {
```
thhous-msft, line 76: Why this change? We don't specify "server" is allowed anywhere in the help, and this seems completely unrelated.

nibanks: This is a improvement for my personal well-being. 😄 I can never remember the arg name, and I always end up using `-server`. I don't feel the need to clutter the help, but I'd like to support it.

thhous-msft (approved):

src/perf/lib/ThroughputClient.cpp
```diff
@@ -288,6 +299,31 @@ ThroughputClient::StartQuic()
             &LocalIpAddr);
     }
 
+    if (CibirIdLength) {
+        BOOLEAN Opt = TRUE;
+        Status =
+            MsQuic->SetParam(
```
thhous-msft, line 305: Isn't the whole point of this change to not require sharing the UDP binding? I thought that with the parameter set, all short headers would get a cid, which we're trying to avoid iirc.

thhous-msft (approved):

PR #2458, opened 2022-03-02
Version Negotiation Draft5+ changes

anrossi:
Update version negotiation code to use new structs and match Draft5+. Update and re-enable tests and add new test cases.


src/core/settings.c
```diff
@@ -235,12 +235,78 @@ QuicSettingsCopy(
...
 
+_IRQL_requires_max_(PASSIVE_LEVEL)
+QUIC_VERSION_SETTINGS*
+QuicSettingsCopyVersionSettings(
+    _In_ const QUIC_VERSION_SETTINGS* const Source,
+    _In_ BOOLEAN CopyExternalToInternal
+    )
+{
+    QUIC_VERSION_SETTINGS* Destination = NULL;
+    uint32_t AllocSize =
```
nibanks, line 246: Use `size_t` instead

src/core/settings.c
```diff
@@ -235,12 +235,78 @@ QuicSettingsCopy(
...
+        CXPLAT_ALLOC_NONPAGED(
+            AllocSize,
+            QUIC_POOL_VERSION_SETTINGS);
+    if (Destination == NULL) {
+        QuicTraceEvent(
+            AllocFailure,
+            "Allocation of '%s' failed. (%llu bytes)",
+            "VersionSettings",
+            AllocSize);
+        goto Fail;
```
nibanks, line 261: Just `return` instead since there is no clean up needed.

src/core/settings.c
```diff
@@ -1028,10 +1055,18 @@ QuicSettingsDumpNew(
...
+        QuicTraceLogVerbose(SettingDumpAcceptedVersionsLength,      "[sett] FullyDeployedVerlength = %u", Settings->VersionSettings->FullyDeployedVersionsLength);
+        if (Settings->VersionSettings->AcceptableVersionsLength > 0) {
+            QuicTraceLogVerbose(SettingDumpAcceptableVersions,      "[sett] AcceptableVersions[0]  = 0x%x", Settings->VersionSettings->AcceptableVersions[0]);
+        }
+        if (Settings->VersionSettings->OfferedVersionsLength > 0) {
+            QuicTraceLogVerbose(SettingDumpOfferedVersions,         "[sett] OfferedVersions[0]     = 0x%x", Settings->VersionSettings->OfferedVersions[0]);
+        }
+        if (Settings->VersionSettings->FullyDeployedVersionsLength > 0) {
+            QuicTraceLogVerbose(SettingDumpFullyDeployedVersions,   "[sett] FullyDeployedVersion[0]= 0x%x", Settings->VersionSettings->FullyDeployedVersions[0]);
         }
```
nibanks, line 1070: Probably should dump all values, not just `[0]`.

docs/api/QUIC_SETTINGS.md
```diff
@@ -37,8 +37,6 @@ typedef struct QUIC_SETTINGS {
             uint64_t MigrationEnabled                       : 1;
             uint64_t DatagramReceiveEnabled                 : 1;
             uint64_t ServerResumptionLevel                  : 1;
-            uint64_t DesiredVersionsList                    : 1;
```
nibanks, line 40: Does this now **exactly** match msquic.h?

nibanks, line 40: Does this now **exactly** match msquic.h?

src/core/quicdef.h
```diff
@@ -453,9 +453,15 @@ CXPLAT_STATIC_ASSERT(
 #define CXPLAT_TLS_RESUMPTION_CLIENT_TICKET_VERSION      1
 
 //
-// By default the Version Negotiation Extension is disabled.
+// By default the Version Negotiation Extension is disabled
+// when preview features are disabled. Otherwise, it's enabled.
 //
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
```
nibanks, line 459: Internally we always enable preview features. I don't understand this.

nibanks, line 459: Internally we always enable preview features. I don't understand this.

anrossi: The goal here is that when that flag is present, version negotiation defaults to enabled (which it wasn't before). But if someone disables the preview features, we disable version negotiation by default (and since the API to change it is also removed when that flag is not present, there's no way to enable it). 
The goal is making it more complete when a 3rd party disables preview features

nibanks: This is an internal header and doesn't change based off the caller. It will always be enabled.

anrossi: oh, I thought it was a CMAKE build flag. My mistake. I'll just enable Version Negotiation by default then.

thhous-msft: Nope, its not a build flag at all. Preview features are always enabled for the build. Its just that they might not show up for the callers, unless they specify the define before including msquic.h

src/core/quicdef.h
```diff
@@ -453,15 +453,9 @@ CXPLAT_STATIC_ASSERT(
 #define CXPLAT_TLS_RESUMPTION_CLIENT_TICKET_VERSION      1
 
 //
-// By default the Version Negotiation Extension is disabled
-// when preview features are disabled. Otherwise, it's enabled.
+// By default the Version Negotiation Extension is enabled.
 //
-#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
 #define QUIC_DEFAULT_VERSION_NEGOTIATION_EXT_ENABLED    TRUE
```
nibanks, line 458: Needs to be off by default for now.

src/inc/msquic.hpp
```diff
@@ -598,6 +600,15 @@ class MsQuicConfiguration {
                 sizeof(*QSettings),
                 QSettings);
     }
+
+    QUIC_STATUS
+    SetVersionNegotiationExtEnabled(_In_ const BOOLEAN Value) noexcept {
```
nibanks, line 605: Should be wrapped in `#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES`.

anrossi: What's that on line 591? 😄

nibanks: 🤦🏻

nibanks (approved):

PR #2459, opened 2022-03-02
Lazily initialize datapath for core affinitization plumbing

csujedihy:
## Description

Lazily initialize datapath during registration open so that plumbing core affinitization for datapath will be easier.

## Testing

Covered by existing tests.


nibanks (dismissed):

csujedihy:
@nibanks Is this a known failure? I keep hitting this in CI and can't really tell what went wrong. I thought secnetperf crashed but I ran the commands locally in the exactly the same setup and it worked fine.

image

nibanks:
> @nibanks Is this a known failure? I keep hitting this in CI and can't really tell what went wrong. I thought secnetperf crashed but I ran the commands locally in exactly the same setup and it worked fine.
> 
> image

I've not seen it before, but I see you pushed another commit. You figure it out?

nibanks (approved):

csujedihy:
> > @nibanks Is this a known failure? I keep hitting this in CI and can't really tell what went wrong. I thought secnetperf crashed but I ran the commands locally in exactly the same setup and it worked fine.
> > image
> 
> I've not seen it before, but I see you pushed another commit. You figure it out?

I have not figured it out. The new commit is a fix for missing condition clean up in library close and some sidecar update. Let's see.

nibanks:
> > > @nibanks Is this a known failure? I keep hitting this in CI and can't really tell what went wrong. I thought secnetperf crashed but I ran the commands locally in exactly the same setup and it worked fine.
> > > image
> > 
> > 
> > I've not seen it before, but I see you pushed another commit. You figure it out?
> 
> I have not figured it out. The new commit is a fix for missing conditional datapath clean up in library close and some sidecar update. Let's see.

Hopefully the missing conditional was causing a crash, which then caused that weird issue. I've not seen it before. Perhaps @thhous-msft has.

csujedihy:
> > > > @nibanks Is this a known failure? I keep hitting this in CI and can't really tell what went wrong. I thought secnetperf crashed but I ran the commands locally in exactly the same setup and it worked fine.
> > > > image
> > > 
> > > 
> > > I've not seen it before, but I see you pushed another commit. You figure it out?
> > 
> > 
> > I have not figured it out. The new commit is a fix for missing conditional datapath clean up in library close and some sidecar update. Let's see.
> 
> Hopefully the missing conditional was causing a crash, which then caused that weird issue. I've not seen it before. Perhaps @thhous-msft has.

Unfortunately, this not the issue. I just sent another PR to trigger WAN perf to see if it's an infra issue.

csujedihy:
> > > > > @nibanks Is this a known failure? I keep hitting this in CI and can't really tell what went wrong. I thought secnetperf crashed but I ran the commands locally in exactly the same setup and it worked fine.
> > > > > image
> > > > 
> > > > 
> > > > I've not seen it before, but I see you pushed another commit. You figure it out?
> > > 
> > > 
> > > I have not figured it out. The new commit is a fix for missing conditional datapath clean up in library close and some sidecar update. Let's see.
> > 
> > 
> > Hopefully the missing conditional was causing a crash, which then caused that weird issue. I've not seen it before. Perhaps @thhous-msft has.
> 
> Unfortunately, this not the issue. I just sent another PR to trigger WAN perf to see if it's an infra issue.

It crashed with tracing on. We log datapath features right after the original call to datapath init in library init. I think I will need to move the code that sets features bits out of datapath init. Otherwise, we would be dealing with etw versioning (not sure how QUIC deals with it and there are tools consuming this feature field already) or we could also move this event to where we eventually initialize datapath?
```
    QuicTraceEvent(
        LibraryInitialized,
        "[ lib] Initialized, PartitionCount=%u DatapathFeatures=%u",
        MsQuicLib.PartitionCount,
        CxPlatDataPathGetSupportedFeatures(MsQuicLib.Datapath));
```

nibanks:
> It crashed with tracing on. We log datapath features right after the original call to datapath init in library init. I think I will need to move the code that sets feature bits out of datapath init. Otherwise, we would be dealing with etw versioning (not sure how QUIC deals with it and there are tools consuming this feature field already) or we could also move this event to where we eventually initialize datapath?

Just go ahead and split the event into two, library initialized and datapath initialized. Deprecate the old event, don't modify/delete it.

csujedihy:
> > It crashed with tracing on. We log datapath features right after the original call to datapath init in library init. I think I will need to move the code that sets feature bits out of datapath init. Otherwise, we would be dealing with etw versioning (not sure how QUIC deals with it and there are tools consuming this feature field already) or we could also move this event to where we eventually initialize datapath?
> 
> Just go ahead and split the event into two, library initialized and datapath initialized. Deprecate the old event, don't modify/delete it.

Does that mean I will need to change all the ETW consumers too?

nibanks:
Yes please. Just the C# code

csujedihy:
This is not crash, right?
```
Remove-Item : Access to the path '\\quic-server\Tests\windows\x64_Release_schannel\msquic.dll' is denied.
At C:\a\_work\1\s\scripts\performance-helper.psm1:183 char:13
+             Remove-Item -Path "$SmbDir/*" -Recurse -Force
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : PermissionDenied: (\\quic-server\Tests…schannel\msquic.dll:FileInfo) [Remove-Item], UnauthorizedAccessException
+ FullyQualifiedErrorId : RemoveFileSystemItemUnAuthorizedAccess,Microsoft.PowerShell.Commands.RemoveItemCommand
##[error]PowerShell exited with code '1'.

```

ThadHouse:
Correct. Is that from a recent run? I think something broke in secnetperf that causes it to not shut down successfully.

csujedihy:
> Correct. Is that from a recent run? I think something broke in secnetperf that causes it to not shut down successfully.

Yeah, it was triggered by the last commit.

My no-op PR also hit this so I don't think this is a bug in this PR. See https://dev.azure.com/ms/msquic/_build/results?buildId=282500&view=logs&j=e76e9942-da2a-5654-7606-edfab665a467&t=4b62f6f1-5694-59d9-6ba0-60f5ac1c4826.

thhous-msft:
Once the system gets into that state, the only way to recover it is to reboot. It was caused by the run of this PR before you fixed that last bug. I'll reboot the system and rerun.

thhous-msft:
Ok, something in the PR is still causing kernel mode to die.

thhous-msft:
Found it. QuicTraceRundown reads the Supported features in the library rundown it the library is initialized. So any time an ETW rundown occurs with the API table open, but datapath not initialized, it segfaults.

csujedihy:
> Found it. QuicTraceRundown reads the Supported features in the library rundown it the library is initialized. So any time an ETW rundown occurs with the API table open, but datapath not initialized, it segfaults.

Thanks, good catch!

src/manifest/MsQuicEtw.man
```diff
@@ -551,6 +557,12 @@
                 name="DatapathFeatures"
                 />
           </template>
+          <template tid="tid_LIBRARY_INIT_V1">
```
nibanks, line 560: Let's call it V2, since it's the second one. And please update everything else accordingly.

src/core/registration.c
```diff
@@ -46,6 +51,30 @@ MsQuicRegistrationOpen(
...
+            CxPlatLockRelease(&MsQuicLib.Lock);
+            QuicTraceEvent(
+                LibraryErrorStatus,
+                "[ lib] ERROR, %u, %s.",
+                Status,
+                "CxPlatDataPathInitialize");
+            goto Error;
+        }
+    }
+    QuicTraceEvent(
```
nibanks, line 72: Needs to go inside the `if` above.

src/core/library.c
```diff
@@ -1893,7 +1875,7 @@ QuicTraceRundown(
             LibraryRundown,
             "[ lib] Rundown, PartitionCount=%u DatapathFeatures=%u",
             MsQuicLib.PartitionCount,
-            CxPlatDataPathGetSupportedFeatures(MsQuicLib.Datapath));
+            MsQuicLib.Datapath != NULL ? CxPlatDataPathGetSupportedFeatures(MsQuicLib.Datapath) : 0);
```
nibanks, line 1878: Let's split this event. The way you have it, you can't tell from the rundown if it's initialized and supports no features or if it's not initialized.

src/plugins/trace/dll/DataModel/ETW/QuicEtwEvent.cs
```diff
@@ -115,9 +115,9 @@ private static QuicObjectType ComputeObjectType(TraceEvent evt)
                 case QuicEventId.LibraryInitialized:
                     if (QuicEvent.ParseMode != QuicEventParseMode.Full) return null;
                     return new QuicLibraryInitializedEvent(timestamp, processor, processId, threadId, pointerSize, data.ReadUInt(), data.ReadUInt());
-                case QuicEventId.LibraryInitializedV1:
+                case QuicEventId.LibraryInitializedV2:
```
csujedihy, line 118: missing rundown event

src/plugins/trace/dll/DataModel/ETW/QuicEtwEvent.cs
```diff
@@ -115,6 +115,12 @@ private static QuicObjectType ComputeObjectType(TraceEvent evt)
                 case QuicEventId.LibraryInitialized:
                     if (QuicEvent.ParseMode != QuicEventParseMode.Full) return null;
                     return new QuicLibraryInitializedEvent(timestamp, processor, processId, threadId, pointerSize, data.ReadUInt(), data.ReadUInt());
+                case QuicEventId.LibraryInitializedV2:
```
nibanks, line 118: Missing the new rundown events in these files.

csujedihy: yeah, noted in my own comment.

csujedihy:
I think sidecar update should automatically delete clog.sidecar. I have hit errors like "....not unique" when running sidecar update and delete the sidecar file resolved it.

nibanks:
> I think sidecar update should automatically delete clog.sidecar. I have hit errors like "....not unique" when running sidecar update and delete the sidecar file resolved it. Sidecar runs fast now.

No, we don't delete the sidecar so that it can warn about errors like this. Can you please take the latest sidecar from main and then run the update function? If you get errors, please share them here.

csujedihy:
I think I broke the test machines. Are there instructions to reboot them?

src/plugins/trace/dll/DataModel/QuicEvents.cs
```diff
@@ -236,6 +268,22 @@ public class QuicAllocFailureEvent : QuicEvent
...
     }
 
+    public class QuicDataPathRundownEvent : QuicEvent
+    {
+        public uint DatapathFeatures { get; }
+
+        public QuicDatapathFeatures Features => (QuicDatapathFeatures)DatapathFeatures;
+
+        public override string PayloadString =>
+            string.Format("Initialized,  DatapathFeatures=[{1}]", Features);
```
csujedihy, line 278: Rundown

src/plugins/trace/dll/DataModel/QuicEvents.cs
```diff
@@ -179,6 +195,22 @@ public class QuicLibraryInitializedEvent : QuicEvent
         }
     }
 
+    public class QuicLibraryInitializedV2Event : QuicEvent
+    {
+        public uint PartitionCount { get; }
+
+        public uint DatapathFeatures { get; }
```
csujedihy, line 202: del

csujedihy:
Some tests failed in "Initialize Job" stage:
Starting: Initialize job
Agent name: 'AZP-5'
Agent machine name: 'WIN-05CUUCOCP5E'
Current agent version: '2.198.2'
Agent running as: 'Administrator'
Prepare build directory.
##[error]One or more errors occurred. (One or more errors occurred. (Access to the path 'c:\a\w\1\s\artifacts\bin\windows\x64_Release_schannel\msquic.dll' is denied.)) (One or more errors occurred. (Access to the path 'c:\a\w\1\s\artifacts\bin\windows\x64_Release_schannel\msquictest.exe' is denied.)) (One or more errors occurred. (Access to the path 'c:\a\w\1\s\artifacts\bin\windows\x64_Release_schannel\msquicpriv.sys' is denied.)) (One or more errors occurred. (Access to the path 'c:\a\w\1\s\artifacts\bin\windows\x64_Release_schannel\msquictestpriv.sys' is denied.)) (Access to the path 'c:\a\w\1\s\artifacts\bin\windows\x64_Release_schannel\msquic.dll' is denied.) (Access to the path 'c:\a\w\1\s\artifacts\bin\windows\x64_Release_schannel\msquictest.exe' is denied.) (Access to the path 'c:\a\w\1\s\artifacts\bin\windows\x64_Release_schannel\msquicpriv.sys' is denied.) (Access to the path 'c:\a\w\1\s\artifacts\bin\windows\x64_Release_schannel\msquictestpriv.sys' is denied.)
Finishing: Initialize job

nibanks:
Deleting the sidecar deletes old, deprecated event definitions that we don't actually want to remove.

csujedihy:
> Deleting the sidecar deletes old, deprecated event definitions that we don't actually want to remove.

I see.

csujedihy:
I still have 8 tests that failed. I can't really tell if it's caused by this PR or some machines are still in a bad state.

7 of them failed at "Initialize Job".
1 (windows x64 schannel schannel Kernel) failed with this error:
```
Remove-Item : Access to the path '\\quic-server\Tests\windows\x64_Release_schannel\msquic.dll' is denied.
At C:\a\_work\1\s\scripts\performance-helper.psm1:183 char:13
+             Remove-Item -Path "$SmbDir/*" -Recurse -Force
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : PermissionDenied: (\\quic-server\Tests…schannel\msquic.dll:FileInfo) [Remove-Item], UnauthorizedAccessException
+ FullyQualifiedErrorId : RemoveFileSystemItemUnAuthorizedAccess,Microsoft.PowerShell.Commands.RemoveItemCommand
```

src/core/registration.c
```diff
@@ -46,6 +52,36 @@ MsQuicRegistrationOpen(
         QUIC_TRACE_API_REGISTRATION_OPEN,
         NULL);
 
+    if (ExternalRegistration) {
```
nibanks, line 55: First, let's move this after the invalid parameter check below. Second, internal registration is always **after** you create an external one, so the Datapath MUST have already been initialized. You should `CXPLAT_DBG_ASSERT` that and clean up how/when you do these checks. For instance, no need to check in the failure path below when releasing the lock.

src/core/registration.c
```diff
@@ -36,6 +36,12 @@ MsQuicRegistrationOpen(
     QUIC_STATUS Status;
     QUIC_REGISTRATION* Registration = NULL;
     size_t AppNameLength = 0;
+    const CXPLAT_UDP_DATAPATH_CALLBACKS DatapathCallbacks = {
+        QuicBindingReceive,
+        QuicBindingUnreachable,
+    };
+    BOOLEAN ExternalRegistration = Config->ExecutionProfile != QUIC_EXECUTION_PROFILE_TYPE_INTERNAL;
```
nibanks, line 43:
```suggestion
    const BOOLEAN ExternalRegistration =
        Config->ExecutionProfile != QUIC_EXECUTION_PROFILE_TYPE_INTERNAL;
```

src/manifest/clog.sidecar
```diff
@@ -14637,31 +14683,11 @@
...
-      },
       {
         "UniquenessHash": "8a578804-829f-fb41-f1d7-f4b5c356a328",
         "TraceID": "ConnStats",
         "EncodingString": "[conn][%p] STATS: SRtt=%u CongestionCount=%u PersistentCongestionCount=%u SendTotalBytes=%llu RecvTotalBytes=%llu CongestionWindow=%u"
       }
     ]
   }
-}
+}
```
nibanks, line 14693: Can you add a trailing new line?

nibanks:
Why is the CLOG submodule changed? Can you revert that change please?

csujedihy:
> Why is the CLOG submodule changed? Can you revert that change please?

I simply just did "git pull origin main" to merge in main.

src/manifest/clog.sidecar
csujedihy: done

src/core/registration.c
csujedihy: Addressed in the latest commit.

nibanks:
> > Why is the CLOG submodule changed? Can you revert that change please?
> 
> I simply just did "git pull origin main" to merge in main.

It looks like you reverted the update over the weekend. Try running the following to get back up to date:
```
cd submodules/clog
git checkout main
git pull
```

csujedihy:
> > > Why is the CLOG submodule changed? Can you revert that change please?
> > 
> > 
> > I simply just did "git pull origin main" to merge in main.
> 
> It looks like you reverted the update over the weekend. Try running the following to get back up to date:
> 
> ```
> cd submodules/clog
> git checkout main
> git pull
> ```

Ah, they are gone now!

nibanks (approved):

nibanks (approved):

src/core/registration.c
```diff
@@ -36,6 +36,13 @@ MsQuicRegistrationOpen(
     QUIC_STATUS Status;
     QUIC_REGISTRATION* Registration = NULL;
     size_t AppNameLength = 0;
+    const CXPLAT_UDP_DATAPATH_CALLBACKS DatapathCallbacks = {
+        QuicBindingReceive,
+        QuicBindingUnreachable,
+    };
+    const BOOLEAN ExternalRegistration =
+        Config->ExecutionProfile != QUIC_EXECUTION_PROFILE_TYPE_INTERNAL;
```
nibanks, line 44: config might be null here for external callers

nibanks:
```suggestion
        Config == NULL || Config->ExecutionProfile != QUIC_EXECUTION_PROFILE_TYPE_INTERNAL;
```

PR #2461, opened 2022-03-03
Add size support to QUIC_STATISTICS_V2

thhous-msft:
Currently, the Stats getter only checked if the minimum required size existed in the outgoing struct. This updates the logic to allow expanding stats in the future easily, along with how to get the size check right for future fields.

Also adds an example to settings how to do the same thing in the future, as we want to make sure its easy to do and check for in the future.

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -6513,8 +6513,9 @@ QuicConnGetV2Statistics(
     _Inout_ uint32_t* StatsLength,
     _Out_writes_bytes_opt_(*StatsLength)
         QUIC_STATISTICS_V2* Stats
-) {
-    uint32_t MinimumStatsSize = (uint32_t)STATISTICS_SIZE_THRU_FIELD(KeyUpdateCount);
+    )
+{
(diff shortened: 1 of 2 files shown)


src/core/connection.c
```diff
@@ -6498,6 +6498,95 @@ QuicConnParamSet(
...
+static
+QUIC_STATUS
+QuicConnGetV2Statistics(
+    _In_ const QUIC_CONNECTION* Connection,
+    _In_ BOOLEAN IsPlat,
+    _Inout_ uint32_t* StatsLength,
+    _Out_writes_bytes_opt_(*StatsLength)
+        QUIC_STATISTICS_V2* Stats
+) {
+    uint32_t MinimumStatsSize = (uint32_t)STATISTICS_SIZE_THRU_FIELD(KeyUpdateCount);
```
nibanks, line 6517:
```suggestion
    )
{
    const uint32_t MinimumStatsSize = (uint32_t)STATISTICS_SIZE_THRU_FIELD(KeyUpdateCount);
```

nibanks (approved):

PR #2462, opened 2022-03-04
Completely remove 32 bit windows kernel mode

thhous-msft:
Replaces #2455 

Windows 32 bit kernel mode is no longer supported, and the 2022 WDK will not support 32 bit kernel mode binaries. Remove the 32 bit kernel mode binaries to make this transition easier. However, windows at least for now still builds 32 bit kernel mode binaries. To make this work, in msquic.h turn the functions into macros, so no linking is required.


.azure/azure-pipelines.ci.yml
```diff
@@ -73,12 +73,6 @@ stages:
   displayName: Build Windows Drivers - Non Tested
   dependsOn: []
   jobs:
-  - template: ./templates/build-config-winkernel.yml
-    parameters:
-      arch: x86
```
nibanks, line 78: Interesting thing to note: we don't test any 32-bit builds...

nibanks (approved):

PR #2479, opened 2022-03-07
Pass CIBIR ID to XDP

nibanks:
## Description

Passes any CIBIR identifier down to XDP on the listener side.

## Testing

Manually verified on Azure VMs:

- [x] Single server with no CIBIR still works.
- [x] Single server with CIBIR works.
- [x] Multiple servers with CIBIR works.
- [x] Client using wrong CIBIR ID fails.

### DIFF
--- a/scripts/prepare-machine.ps1
+++ b/scripts/prepare-machine.ps1
@@ -66,7 +66,10 @@ param (
     [switch]$NoCodeCoverage,
 
     [Parameter(Mandatory = $false)]
-    [switch]$Xdp
+    [switch]$Xdp,
+
+    [Parameter(Mandatory = $false)]
+    [switch]$Force
 )
 
 #Requires -RunAsAdministrator
@@ -137,12 +140,13 @@ function Download-CoreNet-Deps {
 }
 
 function Download-Xdp-Kit {
+    if ($Force) { rm -Force -Recurse $ArtifactsPath }
     if (!(Test-Path $ArtifactsPath)) { mkdir $ArtifactsPath }
     $XdpPath = Join-Path $ArtifactsPath "xdp"
     if (!(Test-Path $XdpPath)) {
         Write-Host "Downloading XDP Kit"
         $ZipPath = Join-Path $ArtifactsPath "xdp.zip"
-        Invoke-WebRequest -Uri "https://lolafiles.blob.core.windows.net/nibanks/xdp.zip" -OutFile $ZipPath
+        Invoke-WebRequest -Uri "https://lolafiles.blob.core.windows.net/nibanks/xdp-latest.zip" -OutFile $ZipPath
         Expand-Archive -Path $ZipPath -DestinationPath $XdpPath -Force
         Remove-Item -Path $ZipPath
     }
--- a/src/inc/quic_datapath.h
+++ b/src/inc/quic_datapath.h
@@ -466,6 +466,12 @@ typedef struct CXPLAT_UDP_CONFIG {
 #ifdef QUIC_OWNING_PROCESS
     QUIC_PROCESS OwningProcess;         // Kernel client-only
 #endif
+#ifdef QUIC_USE_RAW_DATAPATH
+    uint8_t CibirIdLength;              // Value of 0 indicated CIBIR isn't used
+    uint8_t CibirIdOffsetSrc;
+    uint8_t CibirIdOffsetDst;
+    uint8_t CibirId[6];
+#endif
 } CXPLAT_UDP_CONFIG;
 
 //
--- a/src/inc/quic_datapath.h
+++ b/src/inc/quic_datapath.h
@@ -467,10 +467,10 @@ typedef struct CXPLAT_UDP_CONFIG {
     QUIC_PROCESS OwningProcess;         // Kernel client-only
 #endif
 #ifdef QUIC_USE_RAW_DATAPATH
-    uint8_t CibirIdLength;              // Value of 0 indicates CIBIR isn't used
-    uint8_t CibirIdOffsetSrc;
-    uint8_t CibirIdOffsetDst;
-    uint8_t CibirId[6];
+    uint8_t CibirIdLength;              // CIBIR ID length. Value of 0 indicates CIBIR isn't used
+    uint8_t CibirIdOffsetSrc;           // CIBIR ID offset in source CID
+    uint8_t CibirIdOffsetDst;           // CIBIR ID offset in destination CID
+    uint8_t CibirId[6];                 // CIBIR ID data
 #endif
 } CXPLAT_UDP_CONFIG;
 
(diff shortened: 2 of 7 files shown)


scripts/prepare-machine.ps1
```diff
@@ -137,12 +140,13 @@ function Download-CoreNet-Deps {
 }
 
 function Download-Xdp-Kit {
+    if ($Force) { rm -Force -Recurse $ArtifactsPath }
```
maolson-msft, line 143: Can we make the -Force behavior the regular behavior?

src/platform/datapath_raw.h
```diff
@@ -256,8 +256,13 @@ typedef struct CXPLAT_SOCKET {
     void* CallbackContext;
     QUIC_ADDR LocalAddress;
     QUIC_ADDR RemoteAddress;
-    BOOLEAN Wildcard;   // Using a wildcard local address. Optimization to avoid always reading LocalAddress.
-    BOOLEAN Connected;  // Bound to a remote address
+    BOOLEAN Wildcard;           // Using a wildcard local address. Optimization
+                                // to avoid always reading LocalAddress.
+    BOOLEAN Connected;          // Bound to a remote address
+    uint8_t CibirIdLength;      // Value of 0 indicated CIBIR isn't used
```
maolson-msft, line 262:
```suggestion
    uint8_t CibirIdLength;      // Value of 0 indicates CIBIR isn't used
```

scripts/prepare-machine.ps1
nibanks: I'd prefer not, since it's not that way for any/all other PS commands/scripts.

src/inc/quic_datapath.h
```diff
@@ -466,6 +466,12 @@ typedef struct CXPLAT_UDP_CONFIG {
 #ifdef QUIC_OWNING_PROCESS
     QUIC_PROCESS OwningProcess;         // Kernel client-only
 #endif
+#ifdef QUIC_USE_RAW_DATAPATH
+    uint8_t CibirIdLength;              // Value of 0 indicated CIBIR isn't used
```
maolson-msft, line 470:
```suggestion
    uint8_t CibirIdLength;              // Value of 0 indicates CIBIR isn't used
```

scripts/prepare-machine.ps1
maolson-msft: In what situation would you not want to pass -Force?

nibanks: It's just the general pattern to not overwrite things unless an explicit signal is given to do so. I've been considering rewriting this entire scripts BTW, and perhaps I can figure out a better way to do this then.

src/platform/datapath_raw.h
```diff
@@ -256,8 +256,13 @@ typedef struct CXPLAT_SOCKET {
     void* CallbackContext;
     QUIC_ADDR LocalAddress;
     QUIC_ADDR RemoteAddress;
-    BOOLEAN Wildcard;   // Using a wildcard local address. Optimization to avoid always reading LocalAddress.
-    BOOLEAN Connected;  // Bound to a remote address
+    BOOLEAN Wildcard;           // Using a wildcard local address. Optimization
+                                // to avoid always reading LocalAddress.
+    BOOLEAN Connected;          // Bound to a remote address
+    uint8_t CibirIdLength;      // Value of 0 indicates CIBIR isn't used
+    uint8_t CibirIdOffsetSrc;
```
csujedihy, line 263: >CibirIdOffsetSrc [](http://example.com/codeflow?start=12&length=16)

Src and Dst mean CIBIR ID in source CID and destination CID?

nibanks: Yes, the offsets in the source and destination CIDs.

src/platform/datapath_raw.h
```diff
@@ -256,8 +256,13 @@ typedef struct CXPLAT_SOCKET {
...
+    uint8_t CibirIdOffsetDst;
+    uint8_t CibirId[6];
```
csujedihy, line 265: >CibirId[6]; [](http://example.com/codeflow?start=12&length=11)

Would it be better to define CibirId in a struct like this?
struct {
    uint8_t Length;
    uint8_t Offset;
    uint8_t Data[4];
};

src/platform/datapath_raw_xdp.c
```diff
@@ -809,7 +810,7 @@ CxPlatDpRawInterfaceAddRule(
     CxPlatLockAcquire(&Interface->RuleLock);
     // TODO - Don't always allocate a new array?
 
-    if (Interface->RuleCount + 1 == 0) {
+    if ((uint32_t)Interface->RuleCount + (uint32_t)Count > UINT8_MAX) {
```
csujedihy, line 813: oh, we only support 255 connections for now?

nibanks: 255 rules, which is based on the number of sockets, not connections. A single socket on the server can have nearly unlimited connections. Depending on the sharing configured on the client, it can too.

src/platform/datapath_raw.h
csujedihy: maybe worth adding a line comment for them. Not immediately clear for me especially when CIBIR support for core is done in another PR.

src/platform/datapath_raw.h
nibanks: Here, the CibirId is purely the ID, no encoded values internally.

csujedihy: Hmm, that confused me because we name it CibirId in QUIC_CONNECTION.

nibanks: In the QUIC layer, we encode everything in a contiguous memory block for efficiency/packing. Here the space isn't as much an issue so it's unpacked.

src/platform/datapath_raw.h
nibanks: What would you like me to add? I personally think it's mostly self-explanatory. You still figured it out in the end without a comment. 😄

src/platform/datapath_raw.h
csujedihy: It's mainly the naming that confuses me. It's inconsistent across the entire code base. Maybe rename it to CibirIdData? or at least add a comment - "// CIBIR ID data" or maybe just using the macro QUIC_MAX_CIBIR_LENGTH would make it more clearer.

src/platform/datapath_raw.h
csujedihy: "// CIBIR ID offsets in source CID and destination CID"? Up to you. I figured it out after opening the PR for core and the draft you wrote :smile:. Maybe it's just me not knowledgeable enough in QUIC specs.

nibanks: I improved the comments. Let me know what you think.

csujedihy: Thanks! Much clearer.

maolson-msft (approved):

PR #2494, opened 2022-03-09
Embed git commit hash into binary

### LINKED ISSUE #2482: Build number is not propagated to version number in onebranch build
Makes it hard to find the sources for a specific binary.

thhous-msft:
When debugging, having a trivial way to get the commit hash would be very helpful. This adds it to MsQuicLib. Also adds a way to query it, because why not.

Also fixes a bug where kernel mode does not embed the build id in the MsQuicLib version.

Closes #2482


docs/Settings.md
```diff
@@ -88,6 +88,7 @@ These parameters are accessed by calling [GetParam](./api/GetParam.md) or [SetPa
 | `QUIC_PARAM_GLOBAL_SETTINGS`<br> 5                | QUIC_SETTINGS           | Both      | Globally change settings for all subsequent connections.                                              |
 | `QUIC_PARAM_GLOBAL_GLOBAL_SETTINGS`<br> 6         | QUIC_GLOBAL_SETTINGS    | Both      | Globally change global only settings.                                                                 |
 | `QUIC_PARAM_GLOBAL_VERSION_SETTINGS`<br> 7        | QUIC_VERSIONS_SETTINGS  | Both      | Globally change version settings for all subsequent connections.                                      |
+| `QUIC_PARAM_GLOBAL_LIBRARY_GIT_HASH`<br> 8        | char[64]                | Get-only  | Git hash used to build MsQuic (null terminated string, length on output does not include terminator)  |
```
nibanks, line 91: Either include the null in the length or don't say it's null terminated.

thhous-msft: But thats generally how strings work. strlen returns without the null terminator, but requires a buffer containing the null terminator.

src/inc/msquic.h
```diff
@@ -703,6 +703,7 @@ void
 #define QUIC_PARAM_GLOBAL_GLOBAL_SETTINGS               0x01000006  // QUIC_GLOBAL_SETTINGS
 #ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
 #define QUIC_PARAM_GLOBAL_VERSION_SETTINGS              0x01000007  // QUIC_VERSION_SETTINGS
+#define QUIC_PARAM_GLOBAL_LIBRARY_GIT_HASH              0x01000008  // char[64]
```
nibanks, line 706: Need to update C# and Rust, right?

scripts/build.rs
```diff
@@ -14,6 +14,7 @@ fn main() {
     // Builds the native MsQuic and installs it into $OUT_DIR.
     let dst = Config::new(".")
                  .define("QUIC_SOURCE_LINK", "off")
+                 .define("QUIC_EMBED_GIT_HASH" "off")
```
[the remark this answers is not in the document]
thhous-msft: External customers are generally fine here. This specific case is only if the repo is not gotten by a git clone. So submodules work fine here. I can make it a git repo check, so this either works or doesn't, and then doesn't need the option.

docs/Settings.md
nibanks: The difference is that you're not merely querying the length. You're writing memory. The write should indicate the full number of bytes written.

scripts/build.rs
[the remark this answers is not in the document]
nibanks: So `msh3` that uses MsQuic as a submodule shouldn't have to set this to `off`?

src/inc/msquic.h
thhous-msft: Rust does, but I'll need your help. I don't know how you'd read a string. Rust is also broken anyway, its param number is wrong.

scripts/build.rs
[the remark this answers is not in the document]
thhous-msft: Nope. It handles this just fine. Same reason it doesn't have to set QUIC_SOURCE_LINK to off. The only reason these have to be turned off is if they're not being built from a git repo.

src/inc/msquic.h
nibanks: Don't worry about adding a function to actually query the value, just add/fix the global params.

nibanks (approved):
LGTM. Ship it!

nibanks (approved):

nibanks (approved):

PR #2502, opened 2022-03-10
Refactor Performance Pipeline

nibanks:
## Description

This change completely refactors the performance automation so that it is only run from the "perf" pipeline. We constantly have (windows) perf machines issues which cause perf jobs to stall, which causes the main AZP CI pipeline to stall. Putting all perf in its own pipeline means it will only stall other perf pipeline runs.

The refactor also cleans up the options for running custom jobs as well.

This might be a candidate for backport to release/2.0

### DIFF
--- a/.azure/azure-pipelines.perf.yml
+++ b/.azure/azure-pipelines.perf.yml
@@ -3,8 +3,31 @@
 # This pipeline builds and runs MsQuic performance tests.
 #
 
-trigger: none
-pr: none
+trigger:
+  batch: true
+  branches:
+    include:
+    - main
+    - release/*
+  paths:
+    include:
+    - .azure/azure-pipelines.perf.yml
+    - .azure/templates/run-performance.yml
+    - src/core/*
+    - src/platform/*
+    - src/perf/*
+pr:
+  branches:
+    include:
+    - main
+    - release/*
+  paths:
+    include:
+    - .azure/azure-pipelines.perf.yml
+    - .azure/templates/run-performance.yml
+    - src/core/*
+    - src/platform/*
+    - src/perf/*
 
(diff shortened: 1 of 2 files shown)


.azure/azure-pipelines.perf.yml
```diff
@@ -3,19 +3,62 @@
...
 #
 
-trigger: none
-pr: none
+trigger:
+  batch: true
+  branches:
+    include:
+    - main
+    - release/*
```
thhous-msft, line 11: We need to double check, we might need to exclude release/1.1 and release/1.0 here. Otherwise they still might trigger this pipeline. Same thing for 2.0 until we backport it there.

thhous-msft:
I think theres a way to get these to trigger when specific steps finish, rather then waiting for the entire build pipeline to complete.

maolson-msft:
Are pipelines unable to share artifacts with each other? (I ask because it looks like the builds in the CI and perf pipelines are independent currently.)

thhous-msft:
They can share between the pipelines. But maybe the right decision is to not share, because the triggers are a bit hard to figure out there. And the builds are really quick.

maolson-msft:
> They can share between the pipelines. But maybe the right decision is to not share, because the triggers are a bit hard to figure out there. And the builds are really quick.

Yeah, you might be right. I sort of remembered the builds not being that fast while iterating on my PR yesterday, but I might have just been more impatient than usual!

.azure/azure-pipelines.perf.yml
nibanks: It doesn't matter.

thhous-msft (approved):

PR #2506, opened 2022-03-11
Refactor prepare-machine.ps1

nibanks:
## Description

Cleans up the prepare-machine.ps1 script and its usages.

## Testing

Leveraging the CI.

### DIFF
--- a/.azure/templates/run-bvt.yml
+++ b/.azure/templates/run-bvt.yml
@@ -10,7 +10,8 @@ parameters:
   kernel: false
   codeCoverage: false
   logProfile: 'Basic.Light'
-  extraArgs: ''
+  extraPrepareArgs: ''
+  extraTestArgs: ''
   extraArtifactDir: ''
 
 jobs:
@@ -41,14 +42,7 @@ jobs:
     inputs:
       pwsh: true
       filePath: scripts/prepare-machine.ps1
-      ${{ if and(ne(parameters.codeCoverage, true), ne(parameters.extraArtifactDir, '_Xdp')) }}:
-        arguments: -ForTest
-      ${{ if and(eq(parameters.codeCoverage, true), ne(parameters.extraArtifactDir, '_Xdp')) }}:
-        arguments: -ForTest -InstallCodeCoverage
-      ${{ if and(ne(parameters.codeCoverage, true), eq(parameters.extraArtifactDir, '_Xdp')) }}:
-        arguments: -ForTest -InstallDuoNic
-      ${{ if and(eq(parameters.codeCoverage, true), eq(parameters.extraArtifactDir, '_Xdp')) }}:
-        arguments: -ForTest -InstallCodeCoverage -InstallDuoNic
+      arguments: -ForTest ${{ parameters.extraPrepareArgs }}
 
   - task: PowerShell@2
     displayName: Run BVTs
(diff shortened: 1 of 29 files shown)


.azure/templates/run-bvt.yml
```diff
@@ -42,14 +41,14 @@ jobs:
     inputs:
       pwsh: true
       filePath: scripts/prepare-machine.ps1
-      ${{ if and(eq(parameters.kernel, true), eq(parameters.testCerts, true)) }}:
```
nibanks, line 45: Let's refactor to use the `extraPrepareArgs` model.

.github/workflows/check-clog.yml
```diff
@@ -30,7 +30,7 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@629c2de402a417ea7690ca6ce3f33229e27606a5
     - name: Prepare Machine
-      run: scripts/prepare-machine.ps1 -Configuration Dev
+      run: scripts/prepare-machine.ps1
```
nibanks, line 33: Do we really need build and test?

scripts/prepare-machine.ps1
```diff
@@ -4,147 +4,167 @@
...
-$RootDir = Split-Path $PSScriptRoot -Parent
-$NuGetPath = Join-Path $RootDir "nuget"
-
-
-$MessagesAtEnd = New-Object Collections.Generic.List[string]
-
 if ($PSVersionTable.PSVersion.Major -lt 7) {
-
+    # This script requires PowerShell core (mostly for xplat stuff).
     Write-Error ("`nPowerShell v7.x is needed for this script to work. " +
```
maolson-msft, line 96:
```suggestion
    Write-Error ("`nPowerShell v7.x or greater is needed for this script to work. " +
```

scripts/prepare-machine.ps1
```diff
@@ -4,147 +4,167 @@
...
         Expand-Archive -Path $ZipPath -DestinationPath $ArtifactsPath -Force
         Remove-Item -Path $ZipPath
     }
 }
 
-function Download-Xdp-Kit {
-    if ($Force) { rm -Force -Recurse $ArtifactsPath }
-    if (!(Test-Path $ArtifactsPath)) { mkdir $ArtifactsPath }
+# Downloads the latest version of XDP (for building).
+function Install-Xdp {
```
maolson-msft, line 162: But this isn't really installing XDP, it's downloading it. And we will be needing this script to actually INSTALL xdp for the xdp tests.

nibanks: It can be adapted to do that when you onboard XDP tests.

maolson-msft: I don't get it. This name is misleading. What was wrong with the original name?

nibanks: When you "install" an SDK, which consists of just headers and libraries used to build, is that not installing in your opinion?

maolson-msft: I agree that the term "install" is commonly used for SDKs when there is no potential confusion. But that isn't the case here. The XDP SDK can be installed, or the XDP driver can be installed.

It seems to me like your focus is on the name being consistent with other "Install-Xxx" functions. If so, then how about Install-Xdp-Sdk?

nibanks: Ok. I can do that.

maolson-msft (approved):

thhous-msft (approved):

PR #2508, opened 2022-03-11
Update Logs

nibanks:
Fixes a few things I noticed while looking at some logs. Also ran with dev mode so it resorted everything.

### DIFF
--- a/src/manifest/MsQuicEtw.man
+++ b/src/manifest/MsQuicEtw.man
@@ -199,40 +199,48 @@
                 value="2"
                 />
             <map
-                message="$(string.Enum.QUIC_API_TYPE.STRM_CLOSE)"
+                message="$(string.Enum.QUIC_API_TYPE.SET_CONFIGURATION)"
                 value="3"
                 />
             <map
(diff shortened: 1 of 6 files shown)


src/manifest/MsQuicEtw.man
```diff
@@ -199,41 +199,49 @@
                 value="2"
                 />
             <map
-                message="$(string.Enum.QUIC_API_TYPE.STRM_CLOSE)"
+                message="$(string.Enum.QUIC_API_TYPE.SET_CONFIGURATION)"
```
thhous-msft, line 202: This isn't going to break decoding old logs, right?

nibanks: Kind of, but we don't have a choice because the enum in the code changed. I'd rather be correct with main.

thhous-msft: Ok

thhous-msft (approved):

src/manifest/MsQuicEtw.man
thhous-msft: Then we should backport this to Ni.

nibanks: Ok.

PR #2512, opened 2022-03-11
Fix Cleanup Bug in OOM Case

nibanks:


anrossi (approved):

src/platform/platform_winuser.c
```diff
@@ -266,12 +266,18 @@ CxPlatProcessorInfoInit(
     CXPLAT_FREE(Buffer, QUIC_POOL_PLATFORM_TMP_ALLOC);
```
thhous-msft, line 266: freeing Buffer will have the same issue.

nibanks: Fixed

thhous-msft (approved):

PR #2519, opened 2022-03-12
Raw datapath processor candidates knob

csujedihy:
## Description

Allow app to specify processor(s) for datapath threads to run on.

## Testing

Unit tests added in SettingsTest. End-to-end manually tested.


src/core/settings.h
```diff
@@ -90,6 +90,8 @@ typedef struct QUIC_SETTINGS_INTERNAL {
     uint16_t MaxBindingStatelessOperations;
     uint16_t StatelessOperationExpirationMs;
     uint16_t CongestionControlAlgorithm;
+    const uint32_t* RawDataPathProcList;
+    uint32_t RawDataPathProcListLength;
```
nibanks, line 94: Store this directly on MsQuicLib, not the settings (which are duplicated per connection).

src/core/library.c
```diff
@@ -1104,6 +1145,26 @@ QuicLibraryGetGlobalParam(
...
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+
+        *BufferLength = sizeof(uint32_t) * MsQuicLib.Settings.RawDataPathProcListLength;
+
+        CxPlatCopyMemory(Buffer, MsQuicLib.Settings.RawDataPathProcList, *BufferLength);
+
+        Status = QUIC_STATUS_SUCCESS;
+        break;
```
nibanks, line 1166:
```suggestion
        *BufferLength = sizeof(uint32_t) * MsQuicLib.Settings.RawDataPathProcListLength;
        CxPlatCopyMemory(Buffer, MsQuicLib.Settings.RawDataPathProcList, *BufferLength);
        Status = QUIC_STATUS_SUCCESS;
        break;
```

src/core/library.c
```diff
@@ -889,6 +889,47 @@ QuicLibrarySetGlobalParam(
 
         break;
 
+    case QUIC_PARAM_GLOBAL_RAW_DATAPATH_PROCS:
```
nibanks, line 892: Need to wrap in `{ }` because you declare `uint32_t* RawDataPathProcList` below.

src/core/library.c
```diff
@@ -889,6 +889,47 @@ QuicLibrarySetGlobalParam(
...
+            QuicTraceEvent(
+                AllocFailure,
+                "Allocation of '%s' failed. (%llu bytes)",
+                "Raw datapath procs",
+                BufferLength);
+            Status = QUIC_STATUS_OUT_OF_MEMORY;
+            break;
+        }
+
+        CxPlatCopyMemory(RawDataPathProcList, Buffer, BufferLength);
```
nibanks, line 922: We should validate that none of the values are greater than the max CPU.

src/core/library.c
```diff
@@ -889,6 +889,47 @@ QuicLibrarySetGlobalParam(
 
         break;
 
+    case QUIC_PARAM_GLOBAL_RAW_DATAPATH_PROCS:
+        if (Buffer == NULL) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+
+        if (BufferLength < sizeof(uint32_t) || BufferLength % sizeof(uint32_t) != 0) {
```
nibanks, line 898: We should support setting null/zero-length to clear.

src/inc/quic_datapath.h
```diff
@@ -352,7 +357,8 @@ CxPlatDataPathInitialize(
     _In_ uint32_t ClientRecvContextLength,
     _In_opt_ const CXPLAT_UDP_DATAPATH_CALLBACKS* UdpCallbacks,
     _In_opt_ const CXPLAT_TCP_DATAPATH_CALLBACKS* TcpCallbacks,
-    _Out_ CXPLAT_DATAPATH** NewDatapath
+    _Out_ CXPLAT_DATAPATH** NewDatapath,
+    _In_opt_ CXPLAT_DATAPATH_CONFIG* Config
```
nibanks, line 361: Put input before the output.

src/platform/datapath_winuser.c
```diff
@@ -746,15 +746,19 @@ CxPlatDataPathInitialize(
     _In_ uint32_t ClientRecvContextLength,
     _In_opt_ const CXPLAT_UDP_DATAPATH_CALLBACKS* UdpCallbacks,
     _In_opt_ const CXPLAT_TCP_DATAPATH_CALLBACKS* TcpCallbacks,
-    _Out_ CXPLAT_DATAPATH** NewDataPath
+    _Out_ CXPLAT_DATAPATH** NewDataPath,
+    _In_opt_ CXPLAT_DATAPATH_CONFIG* Config
     )
 {
+
```
nibanks, line 753: remove extra newline

src/inc/msquic.h
```diff
@@ -704,6 +704,7 @@ void
 #ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
 #define QUIC_PARAM_GLOBAL_VERSION_SETTINGS              0x01000007  // QUIC_VERSION_SETTINGS
 #define QUIC_PARAM_GLOBAL_LIBRARY_GIT_HASH              0x01000008  // char[64]
+#define QUIC_PARAM_GLOBAL_RAW_DATAPATH_PROCS            0x01000009  // uint32_t[]
```
nibanks, line 707: I believe we use `uint16_t` everywhere internally for CPU index. I think that's plenty. Not sure we need to expose a uint32_t for this. @thhous-msft thoughts?

thhous-msft: Currently yes it is a uint16_t for everything, so we don't need to expose it as a uint32_t (Yet)

src/core/unittest/SettingsTest.cpp
```diff
@@ -417,3 +417,42 @@ TEST(SettingsTest, GlobalSettingsSizesSet)
...
+            QUIC_PARAM_GLOBAL_RAW_DATAPATH_PROCS,
+            &BufferLength,
+            GetCpus));
+    ASSERT_EQ(0, memcmp(GetCpus, SetCpus, sizeof(SetCpus)));
+}
+
+TEST(SettingsTest, GlobalRawDataPathProcsSetAfterDataPathInit)
+{
+    uint32_t SetCpus[] = {1, 2, 3, 4, 5};
+    MsQuicLib.Datapath = (CXPLAT_DATAPATH*)1; // Pretend datapath has been initialized
```
maolson-msft, line 451: Does "MsQuicLib" get reused for other testcases? Because I see you aren't setting MsQuicLib.Datapath back to NULL.

In general this is a little too hacky IMO- you are relying on the internal code not dereferencing this pointer, which could easily change in the future.

nibanks: These are unit tests all do this for the most part it's fine.

csujedihy: For these unit tests, I think, in each test case, if we read values from global variables, we should first memset them to zero in the very beginning.

csujedihy: If gtests have a nice way to do per test case level clean-up, I can do it.

csujedihy: Oh, I can zero msquiclib in test case teardown.

src/core/library.c
csujedihy: Done

src/inc/quic_datapath.h
csujedihy: Fixed

docs/Settings.md
```diff
@@ -89,6 +89,8 @@ These parameters are accessed by calling [GetParam](./api/GetParam.md) or [SetPa
 | `QUIC_PARAM_GLOBAL_GLOBAL_SETTINGS`<br> 6         | QUIC_GLOBAL_SETTINGS    | Both      | Globally change global only settings.                                                                 |
 | `QUIC_PARAM_GLOBAL_VERSION_SETTINGS`<br> 7        | QUIC_VERSIONS_SETTINGS  | Both      | Globally change version settings for all subsequent connections.                                      |
 | `QUIC_PARAM_GLOBAL_LIBRARY_GIT_HASH`<br> 8        | char[64]                | Get-only  | Git hash used to build MsQuic (null terminated string)                                                |
+| `QUIC_PARAM_GLOBAL_RAW_DATAPATH_PROCS`<br> 9      | uint32_t[]              | Both      | Globally change the list of CPUs that raw datapath can use. Must be set before opening registration.  |
```
csujedihy, line 92: >uint32_t[] [](http://example.com/codeflow?start=54&length=10)

fix

src/core/library.c
```diff
@@ -889,6 +894,68 @@ QuicLibrarySetGlobalParam(
...
+            break;
+        }
+
+        uint32_t RawDataPathProcListLength = BufferLength / sizeof(uint16_t);
+        uint16_t* Cpus = (uint16_t*)Buffer;
+        _Analysis_assume_(BufferLength >= 4);
+        for (uint32_t i = 0; i < RawDataPathProcListLength; ++i) {
+            if (Cpus[i] >= CxPlatProcActiveCount()) {
+                Status = QUIC_STATUS_INVALID_PARAMETER;
+                break;
```
csujedihy, line 927: Wrong

src/core/library.c
```diff
@@ -889,6 +894,75 @@ QuicLibrarySetGlobalParam(
 
         break;
 
+    case QUIC_PARAM_GLOBAL_RAW_DATAPATH_PROCS: {
+        if (Buffer == NULL) {
```
nibanks, line 898: I actually think BufferLength == 0 is better now that I think of it. Theoretically they might still pass a buffer pointer, ut if they are setting zero length, I think that'd mean they are clearing this out. Thoughts?

src/core/library.c
```diff
@@ -889,6 +894,75 @@ QuicLibrarySetGlobalParam(
...
+            QuicTraceLogError(
+                LibraryRawDataPathProcsSetAfterDataPathInit,
+                "[ lib] Tried to change raw datapath procs after datapath initialization");
+            Status = QUIC_STATUS_INVALID_STATE;
+            break;
+        }
+
+        uint32_t RawDataPathProcListLength = BufferLength / sizeof(uint16_t);
+        uint16_t* Cpus = (uint16_t*)Buffer;
+        _Analysis_assume_(BufferLength >= 4); // SAL incorrectly thinks 4 bytes will be accessed.
```
nibanks, line 923: Whaaaat? How?!

src/core/library.c
nibanks: You can just use `return` for all these too. Probably simpler.

src/core/library.c
```diff
@@ -889,6 +894,75 @@ QuicLibrarySetGlobalParam(
...
+        }
+
+        if (Status == QUIC_STATUS_INVALID_PARAMETER) {
+            QuicTraceLogError(
+                LibraryRawDataPathInvalidProcs,
+                "[ lib] Tried to set invalid raw datapath procs");
+            break;
+        }
+
+        if (MsQuicLib.RawDataPathProcList != NULL) {
```
nibanks, line 938: Probably better to only do this if the alloc after succeeds, so there are no unintended side effects in OOM case.

src/core/library.c
```diff
@@ -1104,6 +1178,27 @@ QuicLibraryGetGlobalParam(
         Status = QUIC_STATUS_SUCCESS;
         break;
 
+    case QUIC_PARAM_GLOBAL_RAW_DATAPATH_PROCS:
+
+        if (*BufferLength == 0 ||
+            *BufferLength < sizeof(uint16_t) * MsQuicLib.RawDataPathProcListLength) {
+            *BufferLength = sizeof(uint16_t) * MsQuicLib.RawDataPathProcListLength;
+            Status = QUIC_STATUS_BUFFER_TOO_SMALL;
```
nibanks, line 1186: What is the list is zero? Wouldn't an input of 0 be STATUS_SUCCESS instead of TOO_SMALL?

src/core/unittest/SettingsTest.cpp
```diff
@@ -417,3 +417,69 @@ TEST(SettingsTest, GlobalSettingsSizesSet)
                 &InternalSettings));
     }
 }
+
+TEST(SettingsTest, GlobalRawDataPathProcsSetAndGet)
+{
+    uint16_t SetCpus[] = {0};
```
nibanks, line 423: We always run on at least 2 (more likely min 4) CPU machines, so feel free to expand this to at least 2.

src/core/library.c
```diff
@@ -889,6 +894,75 @@ QuicLibrarySetGlobalParam(
...
+            break;
+        }
+
+        if (BufferLength < sizeof(uint16_t) || BufferLength % sizeof(uint16_t) != 0) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+
+        if (MsQuicLib.Datapath != NULL) {
+            QuicTraceLogError(
```
nibanks, line 914: Use `QuicTraceEvent(LibraryError` instead of `QuicTraceLogError`

nibanks: Same for other new error paths.

src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -175,6 +175,19 @@ QuicMainStart(
         return Status;
     }
 
+    uint32_t RawDatapathCpu;
+    if (TryGetValue(argc, argv, "cpu", &RawDatapathCpu)) {
```
nibanks, line 179: Relevant help texts should be updated.

src/platform/datapath_raw_dpdk.c
```diff
@@ -158,7 +165,8 @@ CxPlatDpRawInitialize(
     const uint32_t AdditionalBufferSize =
         sizeof(DPDK_RX_PACKET) + ClientRecvContextLength;
 
-    CxPlatDpdkReadConfig(Dpdk);
+    UNREFERENCED_PARAMETER(Config);
```
nibanks, line 168: Remove UNREF

src/core/library.c
csujedihy:
```
D:\msquic\src\core\library.c(1016,1): error C2220: the following warning is treated as an error [D:\msquic\build\windows\x64_schannel\src\core\core.vcxproj]
D:\msquic\src\core\library.c(925): warning C6385: Reading invalid data from 'Cpus':  the readable size is 'BufferLength' bytes, but '4' bytes may be read.: Lines:
 921, 736, 737, 739, 897, 898, 908, 913, 921, 922, 923, 924, 925, 924, 925 [D:\msquic\build\windows\x64_schannel\src\core\core.vcxproj]
```

I don't really know how. It's either SAL is buggy or this code is buggy but the code looks correct to me.

src/platform/datapath_raw_xdp.c
```diff
@@ -393,11 +406,7 @@ CxPlatXdpReadConfig(
             Value[strlen(Value) - 1] = '\0';
         }
 
-        if (strcmp(Line, "CpuGroup") == 0) {
-             Xdp->DatapathCpuGroup = (uint16_t)strtoul(Value, NULL, 10);
```
nibanks, line 397: Remove the DatapathCpuGroup variable too?

src/core/library.c
csujedihy: I think this all depends on documentation. If we document that setting a NULL buffer clears the list then we are good.

src/core/library.c
nibanks: Ok. That doesn't make sense. I don't see anything wrong with the code.

src/core/library.c
nibanks: The best API needs no documentation and just works as the caller expects it should. A caller might expect null + zero or ptr + zero, but in both cases, zero is supplied.

src/core/library.c
thhous-msft: Yeah, this is absolutely something wrong with SAL. Replacing the indexer with a direct pointer access also avoids the warning.

`if (*(Cpus + i) >= CxPlatProcActiveCount()) {` 

The above works... (And is the exact same code...)

csujedihy: Nice, I will work around it by doing that.

src/core/library.c
csujedihy: Agreed.

src/core/library.c
csujedihy: Now returns sucess if the list is zero.

src/generated/linux/library.c.clog.h
```diff
@@ -438,10 +454,10 @@ tracepoint(CLOG_LIBRARY_C, LibraryRelease );\
 // Decoder Ring for LibraryError
 // [ lib] ERROR, %s.
 // QuicTraceEvent(
-            LibraryError,
-            "[ lib] ERROR, %s.",
-            "Only v2 is supported in MsQuicOpenVersion");
```
nibanks, line 443: weird that clog removed this. :|

nibanks (approved):

PR #2537, opened 2022-03-17
Light up unit tests on XDP platform

maolson-msft:
Enables the unit tests on the XDP platform and fixes the code so that they run green.

### DIFF
--- a/src/inc/quic_datapath.h
+++ b/src/inc/quic_datapath.h
@@ -678,7 +678,7 @@ void
     _Inout_ void* Context,
     _When_(Succeeded == FALSE, _Reserved_)
     _When_(Succeeded == TRUE, _In_reads_bytes_(6))
-        uint8_t* PhysicalAddress,
+        const uint8_t* PhysicalAddress,
     _In_ uint8_t PathId,
     _In_ BOOLEAN Succeeded
     );
--- a/src/inc/quic_datapath.h
+++ b/src/inc/quic_datapath.h
@@ -675,7 +675,7 @@ _IRQL_requires_max_(DISPATCH_LEVEL)
 _Function_class_(CXPLAT_ROUTE_RESOLUTION_CALLBACK)
 void
 (CXPLAT_ROUTE_RESOLUTION_CALLBACK)(
-    _In_ void* Context,
+    _Inout_ void* Context,
     _When_(Succeeded == FALSE, _Reserved_)
     _When_(Succeeded == TRUE, _In_reads_bytes_(6))
         uint8_t* PhysicalAddress,
--- a/src/platform/datapath_raw.c
+++ b/src/platform/datapath_raw.c
@@ -121,6 +121,12 @@ CxPlatDataPathInitialize(
         Status = QUIC_STATUS_INVALID_PARAMETER;
         goto Exit;
     }
+    if (UdpCallbacks != NULL) {
+        if (UdpCallbacks->Receive == NULL || UdpCallbacks->Unreachable == NULL) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            goto Exit;
+        }
+    }
 
     *NewDataPath = CXPLAT_ALLOC_PAGED(DatapathSize, QUIC_POOL_DATAPATH);
     if (*NewDataPath == NULL) {
@@ -135,10 +141,6 @@ CxPlatDataPathInitialize(
     CxPlatZeroMemory(*NewDataPath, DatapathSize);
 
     if (UdpCallbacks) {
-        if (UdpCallbacks->Receive == NULL || UdpCallbacks->Unreachable == NULL) {
-            Status = QUIC_STATUS_INVALID_PARAMETER;
-            goto Error;
-        }
         (*NewDataPath)->UdpHandlers = *UdpCallbacks;
     }
 
--- a/src/platform/datapath_raw.c
+++ b/src/platform/datapath_raw.c
@@ -135,6 +135,10 @@ CxPlatDataPathInitialize(
     CxPlatZeroMemory(*NewDataPath, DatapathSize);
 
     if (UdpCallbacks) {
+        if (UdpCallbacks->Receive == NULL || UdpCallbacks->Unreachable == NULL) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            goto Error;
+        }
         (*NewDataPath)->UdpHandlers = *UdpCallbacks;
     }
 
--- a/src/platform/datapath_raw.c
+++ b/src/platform/datapath_raw.c
@@ -117,6 +117,11 @@ CxPlatDataPathInitialize(
 
     UNREFERENCED_PARAMETER(TcpCallbacks);
 
+    if (NewDataPath == NULL) {
+        Status = QUIC_STATUS_INVALID_PARAMETER;
+        goto Exit;
+    }
+
     *NewDataPath = CXPLAT_ALLOC_PAGED(DatapathSize, QUIC_POOL_DATAPATH);
     if (*NewDataPath == NULL) {
         QuicTraceEvent(
@@ -157,6 +162,8 @@ Error:
             *NewDataPath = NULL;
         }
     }
+
+Exit:
 
     return Status;
 }
--- a/src/platform/datapath_raw.c
+++ b/src/platform/datapath_raw.c
@@ -553,7 +553,7 @@ CxPlatSocketSend(
     CXPLAT_DBG_ASSERT(Route->Queue != NULL);
     const CXPLAT_INTERFACE* Interface = CxPlatDpRawGetInterfaceFromQueue(Route->Queue);
     CxPlatFramingWriteHeaders(
-        Socket, Route, &SendData->Buffer,
+        Socket, Route, &SendData->Buffer, SendData->ECN,
         Interface->OffloadStatus.Transmit.NetworkLayerXsum,
         Interface->OffloadStatus.Transmit.TransportLayerXsum);
     CxPlatDpRawTxEnqueue(SendData);
--- a/src/platform/datapath_raw_socket.c
+++ b/src/platform/datapath_raw_socket.c
@@ -499,7 +499,13 @@ typedef struct ETHERNET_HEADER {
 
 typedef struct IPV4_HEADER {
     uint8_t VersionAndHeaderLength;
-    uint8_t TypeOfServiceAndEcnField;
+    union {
+        uint8_t TypeOfServiceAndEcnField;
+        struct {
+            uint8_t EcnField : 2;
+            uint8_t TypeOfService : 2;
+        };
+    };
     uint16_t TotalLength;
     uint16_t Identification;
     uint16_t FlagsAndFragmentOffset;
@@ -625,6 +631,7 @@ CxPlatDpRawParseIPv4(
             return;
         }
 
+        Packet->TypeOfService = IP->EcnField;
         Packet->Route->RemoteAddress.Ipv4.sin_family = AF_INET;
         CxPlatCopyMemory(&Packet->Route->RemoteAddress.Ipv4.sin_addr, IP->Source, sizeof(IP->Source));
         Packet->Route->LocalAddress.Ipv4.sin_family = AF_INET;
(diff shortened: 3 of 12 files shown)


src/platform/unittest/DataPathTest.cpp
```diff
@@ -455,6 +532,21 @@ struct CxPlatSocket {
         if (QUIC_SUCCEEDED(InitStatus)) {
             CxPlatSocketGetLocalAddress(Socket, &Route.LocalAddress);
             CxPlatSocketGetRemoteAddress(Socket, &Route.RemoteAddress);
+#ifdef QUIC_USE_RAW_DATAPATH
+            if (!QuicAddrIsWildCard(&Route.RemoteAddress)) {
+                //
+                // This is a connected socket and its route must be resolved
+                // to be able to send traffic.
+                //
+                InitStatus = CxPlatResolveRoute(Socket, &Route, 0, &Route, ResolveRouteComplete);
```
maolson-msft, line 541: Any idea why my ResolveRouteComplete callback being passed to CxPlatResolveRoute doesn't compile? I've tried it both with and without wrapping the function definition in extern "C". The function signature perfectly matches, but it still claims not to be able to "convert" the argument:

E:\msquic\src\platform\unittest\DataPathTest.cpp(542,1): error C2664: 'HRESULT CxPlatResolveRoute(CXPLAT_SOCKET *,CXPLAT_ROUTE *,uint8_t,void *,CXPLAT_ROUTE_RESOLUTION_CALLBACK_HANDLER)': cannot convert argument 5 fro
m 'void (__cdecl *)(void *,const uint8_t *,uint8_t,BOOLEAN)' to 'CXPLAT_ROUTE_RESOLUTION_CALLBACK_HANDLER' [E:\msquic\build\windows\x64_schannel\src\platform\unittest\msquicplatformtest.vcxproj]

csujedihy: InitStatus = CxPlatResolveRoute(Socket, &Route, 0, &Route, ResolveRouteComplete);
=>
InitStatus = CxPlatResolveRoute(Socket, &Route, 0, &Route, (CXPLAT_ROUTE_RESOLUTION_CALLBACK_HANDLER)ResolveRouteComplete);

src/platform/datapath_raw.c
```diff
@@ -130,6 +135,10 @@ CxPlatDataPathInitialize(
     CxPlatZeroMemory(*NewDataPath, DatapathSize);
 
     if (UdpCallbacks) {
+        if (UdpCallbacks->Receive == NULL || UdpCallbacks->Unreachable == NULL) {
```
nibanks, line 138: This validation should be before allocating the memory above, otherwise you're leaking it on failure it seems.

src/platform/unittest/DataPathTest.cpp
nibanks: It might be the calling convention. We might need to specify calling convention on the declaration of all the quic_datapath.h callbacks, and use it when we define the callbacks. Please don't just cast to make the error go away.

src/platform/unittest/DataPathTest.cpp
```diff
@@ -396,6 +452,27 @@ struct CxPlatDataPath {
     uint32_t GetSupportedFeatures() const noexcept { return CxPlatDataPathGetSupportedFeatures(Datapath); }
 };
 
+#ifdef QUIC_USE_RAW_DATAPATH
+extern "C" {
```
nibanks, line 456: If it's for just a single function, you don't need the `{` after the `extern "C"`.

src/inc/quic_datapath.h
```diff
@@ -675,7 +675,7 @@ _IRQL_requires_max_(DISPATCH_LEVEL)
 _Function_class_(CXPLAT_ROUTE_RESOLUTION_CALLBACK)
 void
 (CXPLAT_ROUTE_RESOLUTION_CALLBACK)(
-    _In_ void* Context,
+    _Inout_ void* Context,
```
nibanks, line 678: Please make sure to update all implementations of this callback to use this updated SAL annotation.

src/platform/unittest/DataPathTest.cpp
thhous-msft: Solved. Was a const issue.

src/platform/unittest/DataPathTest.cpp
thhous-msft: Changed to just be static

src/inc/quic_datapath.h
maolson-msft: Yeah the only other one is QuicConnQueueRouteCompletion, and it was already marked Inout, so there are no more changes to be made.

src/platform/datapath_raw.c
```diff
@@ -117,6 +117,17 @@ CxPlatDataPathInitialize(
 
     UNREFERENCED_PARAMETER(TcpCallbacks);
 
+    if (NewDataPath == NULL) {
+        Status = QUIC_STATUS_INVALID_PARAMETER;
+        goto Exit;
```
nibanks, line 122: It's fine to use `return` instead.

src/platform/datapath_raw_xdp.c
```diff
@@ -1212,6 +1212,13 @@ CxPlatXdpRx(
             FrameBuffer,
             (uint16_t)Buffer->length);
 
+        //
+        // The route has been filled in with the packet's src/dst IP and ETH addresses, so
+        // mark it resolved. This allows stateless sends to be issued without performing
+        // a route lookup.
+        //
+        Packet->Route->State = RouteResolved;
```
csujedihy, line 1220: Is this really causing an issue? Stateless sends don't care route state. The stateless send codepath doesn't call into SendQueueFlush where the route resolution is happening.

maolson-msft: They do care. For stateless sends, the RX packet's Route is used. It's passed to CxPlatSocketSend, which then barfs because not only is the State of the route not Resolved, but also the LocalLinkLayerAddress and NextHopLinkLayerAddress are not set up. Without those, there's no way to build the stateless packet. I've fixed that by filling in the ETH addresses in CxPlatDpRawParseEthernet.

csujedihy: Local and next hop MAC addresses missing is the problem. Route state doesn't need to be changed here because it's not referenced in stateless case.

csujedihy: Stateless send doesn't check `Route->State`.

maolson-msft: It does. Look at CxPlatSocketSend.
CXPLAT_DBG_ASSERT(Route->State == RouteResolved);

csujedihy: Please remove this assert and the change to set route state to Resolved unless you just want to set the route state for this assert to pass. This is something leftover from the old code where we get route in allocating sends.

csujedihy: Nothing else in stateless send code path reference route resolution state except this assert.

src/platform/datapath_raw_socket.c
```diff
@@ -734,6 +734,9 @@ CxPlatDpRawParseEthernet(
         return;
     }
 
+    CxPlatCopyMemory(&Packet->Route->LocalLinkLayerAddress, Ethernet->Destination, sizeof(Ethernet->Destination));
+    CxPlatCopyMemory(&Packet->Route->NextHopLinkLayerAddress, Ethernet->Source, sizeof(Ethernet->Source));
```
csujedihy, line 738: Before this change, we leave LocalLinkLayerAddress and NextHopLinkLayerAddress being zeros. But then we will overwrite (excluding Route->State) route in the connection that the datagram gets indicated to? Wouldn't that just overwrite local and next hop mac addresses to zeros and end up causing the connection sending frames with zero'd MAC addresses?

csujedihy: Oh, I see. Overwriting route info only happens once per path for (stateful) connections.

maolson-msft: Resolved offline

src/platform/datapath_raw_xdp.c
maolson-msft: Resolved offline. This assert is important: we have the invariant that Route->State implies the src/dest eth/ip addresses in the Route are valid, which is required for sending a packet, so we should keep the assert here and hence keep the State update.

src/platform/datapath_raw_socket.c
```diff
@@ -625,6 +631,7 @@ CxPlatDpRawParseIPv4(
             return;
         }
 
+        Packet->TypeOfService = IP->EcnField;
```
maolson-msft, line 634: FYI, the new ECN code was verified by collecting pcap during the DataPathTest/DataPathTest.UdpDataECT0/4 and DataPathTest/DataPathTest.UdpDataECT0/6 testcases and checking that the packets have ECT0 (instead of, say, some other incorrect ECN bit) set.

src/platform/datapath_raw.c
maolson-msft: This style is following the style of the other implementations of CxPlatDataPathInitialize in winuser, winkernel, etc. I'd rather not make this style change in this PR.

src/platform/datapath_raw_socket.c
```diff
@@ -499,7 +499,13 @@ typedef struct ETHERNET_HEADER {
 
 typedef struct IPV4_HEADER {
     uint8_t VersionAndHeaderLength;
-    uint8_t TypeOfServiceAndEcnField;
+    union {
+        uint8_t TypeOfServiceAndEcnField;
+        struct {
```
nibanks, line 504: Could you complete the `struct` part, so we're not missing 4 bits?

src/platform/unittest/DataPathTest.cpp
```diff
@@ -179,15 +183,65 @@ struct DataPathTest : public ::testing::TestWithParam<int32_t>
...
         NextPort = 50000 + (CxPlatCurThreadID() % 10000) + (rand() % 5000);
 
-        LocalIPv4.Resolve(QUIC_ADDRESS_FAMILY_INET, "localhost");
-        LocalIPv6.Resolve(QUIC_ADDRESS_FAMILY_INET6, "localhost");
+        if (UseDuoNic) {
+            LocalIPv4.Resolve(QUIC_ADDRESS_FAMILY_INET, "192.168.1.11");
+            LocalIPv6.Resolve(QUIC_ADDRESS_FAMILY_INET6, "fc00::1:11");
+        } else {
+            LocalIPv4.Resolve(QUIC_ADDRESS_FAMILY_INET, "localhost");
+            LocalIPv6.Resolve(QUIC_ADDRESS_FAMILY_INET6, "localhost");
```
nibanks, line 240: I _think_ this will still not work on Linux, but I'm not sure. Because it uses a different name for v6 localhost (look in msquic_posix.h I think).

src/platform/unittest/DataPathTest.cpp
```diff
@@ -613,15 +705,18 @@ TEST_P(DataPathTest, UdpData)
...
+        unspecAddress.SockAddr.Ipv4.sin_port = GetNextPort();
+        Server.CreateUdp(Datapath, &unspecAddress.SockAddr, nullptr, &RecvContext);
     }
     VERIFY_QUIC_SUCCESS(Server.GetInitStatus());
     ASSERT_NE(nullptr, Server.Socket);
-    RecvContext.DestinationAddress = Server.GetLocalAddress();
+
+    auto serverAddress = GetNewLocalAddr();
+    RecvContext.DestinationAddress = serverAddress.SockAddr;
+    RecvContext.DestinationAddress.Ipv4.sin_port = Server.GetLocalAddress().Ipv4.sin_port;
```
nibanks, line 719: Might need to use to GetPort/SetPort helpers.

src/platform/unittest/DataPathTest.cpp
maolson-msft: The linux test passed, but I'll still switch this to use QUIC_LOCALHOST_FOR_AF.

src/platform/unittest/DataPathTest.cpp
maolson-msft: This isn't strictly needed since all platforms have a Ipv4.sin_port field, but I'll make the change anyway.

maolson-msft: Actually, after searching for "sin_port" I see this is a sweeping change and is a problem with the existing code rather than the PR, so let's leave this for its own PR.

src/platform/datapath_raw_socket.c
maolson-msft: Wow yeah, I think this was a typo- TypeOfService should be 6 bits. It just happened to still work with the typo

nibanks (approved):

PR #2538, opened 2022-03-17
Add support to secnetperf to explcitly set cipher suites

thhous-msft:
A new -cipher flag is added, which takes the decimal value of QUIC_ALLOWED_CIPHER_SUITE_FLAGS to set for the configuration. Very useful for perf testing alternate cipher suites.


src/platform/selfsign_capi.c
```diff
@@ -1032,7 +1032,8 @@ _IRQL_requires_max_(PASSIVE_LEVEL)
 const QUIC_CREDENTIAL_CONFIG*
 CxPlatGetSelfSignedCert(
     _In_ CXPLAT_SELF_SIGN_CERT_TYPE Type,
-    _In_ BOOLEAN IsClient
+    _In_ BOOLEAN IsClient,
+    _In_ uint8_t AllowedCiphers
```
nibanks, line 1036: Instead of passing this in, let's just update the function to return a non-const, and update the caller to add the flag as necessary. This model of passing things down isn't scalable.

src/perf/bin/appmain.cpp
```diff
@@ -425,6 +426,7 @@ main(
         }
     }
 
+
```
nibanks, line 429: nit: remove ws

src/perf/bin/appmain.cpp
```diff
@@ -437,6 +439,11 @@ main(
         goto Exit;
     }
 
+    if (TryGetValue(argc, argv, "cipher", &CipherSuite)) {
```
nibanks, line 442: Add to help text

nibanks (commented):
Almost ready to merge. It's still marked as draft though. Is there something else you wanted to do to this PR?

nibanks (approved):

PR #2539, opened 2022-03-17
Don't enable datapath queued sends when shared ec is enabled

thhous-msft:
Queued datapath sends were added to increase raw throughput, as the send would be processed on the datapath thread, which was often less utilized then the worker thread. However, when shared ec is enabled, these threads are the same. This means that we would queue the send, then immediately poll the iocp and do the send. This will add latency. Extra latency could be added if a receive is queued before the send, as the receive would then get processed before the send.


anrossi (approved):

CMakeLists.txt
```diff
@@ -320,8 +320,10 @@ endif()
 if(QUIC_TLS STREQUAL "schannel")
     message(STATUS "Enabling Schannel configuration tests")
     list(APPEND QUIC_COMMON_DEFINES QUIC_TEST_SCHANNEL_FLAGS=1)
-    message(STATUS "Enabling UDP Send Queuing")
-    list(APPEND QUIC_COMMON_DEFINES CXPLAT_DATAPATH_QUEUE_SENDS=1)
+    if (NOT QUIC_SHARED_EC)
```
nibanks, line 323: There still might be reasons to batch call the sends over and over. Thoughts? Let's talk about this some. Ideally, I'd like to have this decision be data-driven.

thhous-msft: I can't really think of any reasons. Because we have a maximum batch size, the send loop is going to exit as soon as it calls into send. Then the next thing that will happen will be a IOCP wakeup just to process the send that was queued. The queueing was added to take advantage of the often not full datapath thread. But with shared ec, there is no datapath thread.

nibanks (approved):

Issue #2509, opened 2022-03-11
Misc/WithKeyUpdateArgs1.KeyUpdate/0 failed


nibanks:
### Describe the bug

```
[----------] 1 test from Misc/WithKeyUpdateArgs1
[ RUN      ] Misc/WithKeyUpdateArgs1.KeyUpdate/0
D:\a\1\msquic\src\test\lib\HandshakeTest.cpp(2006): error: Server->GetLocalBidiStreamCount() not equal to (uint16_t)(101+i)
[  FAILED  ] Misc/WithKeyUpdateArgs1.KeyUpdate/0, where GetParam() = v4/0 (1492 ms)
```
When looking at [quic.txt](https://github.com/microsoft/msquic/files/8234201/quic.txt) it seems that for some reason, the packet that was built with the update MAX_STREAMS was never actually flushed. I can't tell why not.
```
2711 [0]15F4.15C8::2022/03/11-11:40:48.787017500 [Microsoft-Quic][conn][0x1ADB1C68200] Execute: 4(FLUSH_SEND)
2712 [0]15F4.15C8::2022/03/11-11:40:48.787019800 [Microsoft-Quic][conn][0x1ADB1C68200] Flushing Send. Allowance=12200 bytes
2713 [0]15F4.15C8::2022/03/11-11:40:48.787022200 [Microsoft-Quic][pack][1099511627782] Created in batch 1099511627782
2714 [0]15F4.15C8::2022/03/11-11:40:48.787028500 [Microsoft-Quic][C][TX][18] SH DestCid:007de07a097ff4a990 KP:1 SB:1 (Payload 7 bytes)
2715 [0]15F4.15C8::2022/03/11-11:40:48.787032200 [Microsoft-Quic][C][TX][18]   MAX_STREAMS[1] Count:105
2716 [0]15F4.15C8::2022/03/11-11:40:48.787033100 [Microsoft-Quic][pack][1099511627782] Encrypting
2717 [0]15F4.15C8::2022/03/11-11:40:48.787040700 [Microsoft-Quic][pack][1099511627782] Finalizing
2718 [0]15F4.15C8::2022/03/11-11:40:48.787041600 [Microsoft-Quic][conn][0x1ADB1C68200][TX][18] 5(ONE_RTT) (33 bytes)
[...]

PR #2540, opened 2022-03-17
Add logs to all cases an IOCP can exit the EC early

thhous-msft:
To make debugging #2509 easier, add logs to ensure we're not misprocessing the IOCP.


src/platform/datapath_winuser.c
```diff
@@ -4018,10 +4023,18 @@ CxPlatDataPathRunEC(
...
+            DatapathWakeupForShutdown,
+            "[data][%p] Datapath wakeup for shutdown",
+            DatapathProc->Datapath);
         return;
     }
 
     if (SocketProc == NULL || Overlapped == NULL) {
+        QuicTraceLogVerbose(
+            DatapathWakeupForECTimeout,
+            "[data][%p] Datapath wakeup for EC timeout",
```
nibanks, line 4036: This is actually both for event wake and timeout.

src/platform/datapath_winuser.c
```diff
@@ -2564,6 +2564,11 @@ CxPlatDataPathSocketContextShutdown(
 
     CxPlatRundownUninitialize(&SocketProc->UpcallRundown);
 
+    QuicTraceLogVerbose(
+        DatapathSocketContextComplete,
+        "[data][%p] Socket context shutdown",
```
nibanks, line 2569: All the events you added, I don't think you should be logging the "parent". Since there are multiple child objects, you won't be able to distinguish which one.

nibanks (approved):

PR #2546, opened 2022-03-17
Add QuicLoad Tool

nibanks:
A tool to test creating multiple H3 connections.


src/tools/load/load.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+const MsQuicApi* MsQuic;
+volatile long ConnectedCount;
+volatile long ConnectionsActive;
+
+void ResolveServerAddress(const char* ServerName, QUIC_ADDR& ServerAddress) {
+    CxPlatSystemLoad();
+    CxPlatInitialize();
+    CXPLAT_DATAPATH* Datapath = nullptr;
+    //QuicAddrSetFamily(&ServerAddress, AF_INET);
```
anrossi, line 20: not needed anymore?

src/tools/load/load.cpp
```diff
@@ -0,0 +1,90 @@
...
+    const char* ServerName = argv[1];
+    QuicAddr ServerAddress = {0};
+    ResolveServerAddress(ServerName, ServerAddress.SockAddr);
+
+    const uint32_t ConnectionCount = argc > 2 ? atoi(argv[2]) : 100;
+    const uint32_t KeepAliveMs = argc > 3 ? atoi(argv[3]) : 60 * 1000;
+    const uint32_t PollMs = argc > 4 ? atoi(argv[4]) : 10 * 1000;
+    const bool ShareUdp = argc > 5 ? (atoi(argv[5]) ? true : false) : true;
+
+    MsQuic = new(std::nothrow) MsQuicApi;
```
anrossi, line 57: Does this call `MsQuicOpen` and all that? wow!

anrossi (approved):

src/tools/load/load.cpp
nibanks: I'm leaving it there for manual usage in the future, but I don't want it in by default.

src/tools/load/load.cpp
nibanks: Yep. The power of C++!

PR #2548, opened 2022-03-18
QUIC Version 2 support

anrossi:
Add support for starting and handling connections with QUIC version 2 and update version negotiation tests to use QUIC version 2.


src/core/connection.c
```diff
@@ -4163,7 +4164,7 @@ QuicConnRecvDecryptAndAuthenticate(
 
     if (!Packet->IsShortHeader) {
         switch (Packet->LH->Type) {
```
nibanks, line 4166: This needs to have a version.

src/core/crypto.c
```diff
@@ -771,20 +771,20 @@ QuicCryptoWriteCryptoFrames(
         uint32_t EncryptLevelStart;
         uint32_t PacketTypeRight;
         switch (Builder->PacketType) {
```
nibanks, line 773: Need more version checks?

src/core/packet.c
```diff
@@ -204,7 +234,8 @@ QuicPacketValidateLongHeaderV1(
     // Validate acceptable types.
     //
     CXPLAT_DBG_ASSERT(IsServer == 0 || IsServer == 1);
-    if (QUIC_HEADER_TYPE_ALLOWED[IsServer][Packet->LH->Type] == FALSE) {
+    if (QUIC_HEADER_TYPE_ALLOWED_V1[IsServer][Packet->LH->Type] == FALSE ||
```
nibanks, line 237: You need to check the right one, not both.

src/platform/unittest/CryptTest.cpp
```diff
@@ -164,7 +164,7 @@ struct CryptTest : public ::testing::TestWithParam<int32_t>
     };
 };
 
-TEST_F(CryptTest, WellKnownClientInitial)
+TEST_F(CryptTest, WellKnownClientInitialv1)
```
nibanks, line 167: Could we write a v2 version too?

anrossi: planning on it!

src/core/connection.h
```diff
@@ -689,6 +689,30 @@ QuicConnGetNextExpirationTime(
     return Connection->Timers[0].ExpirationTime;
 }
 
+//
+// Returns the current QUIC version of the connection.
+//
+inline
+uint32_t
+QuicConnGetVersion(
```
nibanks, line 697: This helper seems a little unnecessary.

anrossi: As-is, I agree. but I was planning to make a follow-up PR where I replace the long-form with this function. The goal being to make it easier to tell where the version is being retrieved, versus where it is being set.

anrossi: As-is, I agree. but I was planning to make a follow-up PR where I replace the long-form with this function. The goal being to make it easier to tell where the version is being retrieved, versus where it is being set.

src/core/packet.h
```diff
@@ -179,7 +188,7 @@ typedef struct QUIC_LONG_HEADER_V1 {
 // order.
 //
 
-typedef struct QUIC_RETRY_V1 {
+typedef struct QUIC_RETRY_HEADER_V1 {
```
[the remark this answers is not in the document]
nibanks: Or `QUIC_RETRY_HEADER_V1V2`. I guess I don't care much either way, but we should be specific that it's used for both.

src/core/packet.h
```diff
@@ -161,7 +170,7 @@ typedef struct QUIC_LONG_HEADER_V1 {
     //uint8_t PacketNumber[PnLength];
     //uint8_t Payload[0];
 
-} QUIC_LONG_HEADER_V1;
+} QUIC_LONG_HEADER_V1, QUIC_LONG_HEADER_V2;
```
nibanks, line 173: Since the V1 and V2 structs work for both, should we just call them `QUIC_LONG_HEADER_V1V2` (and any other similar types)?

src/core/stream_send.c
```diff
@@ -1096,7 +1096,8 @@ QuicStreamSendWrite(
         uint16_t StreamFrameLength = AvailableBufferLength - Builder->DatagramLength;
         QuicStreamWriteStreamFrames(
             Stream,
-            Builder->PacketType == QUIC_INITIAL,
+            ((!QuicConnIsVersion2(Stream->Connection) && Builder->PacketType == QUIC_INITIAL_V1) ||
+            (QuicConnIsVersion2(Stream->Connection) && Builder->PacketType == QUIC_INITIAL_V2)),
```
nibanks, line 1100: Let's put this in a separate (const) variable above. As we add more versions, the logic will likely grow and I don't want to cram it all into here.

src/core/unittest/VersionNegExtTest.cpp
```diff
@@ -25,7 +29,11 @@ TEST(VersionNegExtTest, ParseVersionInfoFail)
     };
 
     QUIC_VERSION_INFORMATION_V1 ParsedVI = {0};
-    QUIC_CONNECTION* NoOpConnection = (QUIC_CONNECTION*)0x1;
+    struct { QUIC_HANDLE Handle;
+        QUIC_CONNECTION Connection;
+    } Connection {};
```
nibanks, line 34: I don't understand this struct. Is it supposed to be union?

src/tools/interopserver/InteropServer.cpp
```diff
@@ -11,9 +11,10 @@
 
 #include "InteropServer.h"
 
-#define QUIC_VERSION_1          0x01000000U     // First official version
-#define QUIC_VERSION_MS_1       0x0000cdabU     // First Microsoft version (currently same as latest draft)
-#define QUIC_VERSION_DRAFT_29   0x1d0000ffU     // IETF draft 29
+#define QUIC_VERSION_2_H        0x709a50c4U     // Second official version
```
nibanks, line 14: I wonder if we should just include quic_versions.h

src/core/stream_send.c
anrossi: Should it be a flag on the Builder?

src/core/connection.h
nibanks: All you are doing is checking if a variable == 2. While I agree it's a few more characters, I don't think it's worth the abstraction. @thhous-msft want to be the tie breaker here?

src/core/packet.h
nibanks: Probably. Thoughts?

src/core/stream_send.c
nibanks: Maybe...

nibanks:
Something important came up on Slack discussion around v2. We need to make sure the interop client/server continues to negotiate v1 for now, otherwise it shows as a break if they ever negotiate v2.

anrossi:
> for now, otherwise it shows as a break if they ever negotiate v2.

interop server currently prefers version 2, but it will work with version 1. If a client supports version 2 and version 1, and starts with version 1, the server will upgrade it to version 2 via CVN (if the enableVNE flag is passed on commandline)
Does this solve the concern?

nibanks:
> > for now, otherwise it shows as a break if they ever negotiate v2.
> 
> interop server currently prefers version 2, but it will work with version 1. If a client supports version 2 and version 1, and starts with version 1, the server will upgrade it to version 2 via CVN (if the enableVNE flag is passed on commandline) Does this solve the concern?

As far as I understand, the interop runner simply doesn't support v2, so any connection that uses it will confuse it and result in a failure.

src/core/packet.h
anrossi: It'll be consistent, but cluttered-looking, IMO. What if V3 doesn't change anything except the long header again? then everything will be `_V1V2V3`. I'm not sure what solution is best here.

src/core/connection.h
anrossi: I'll remove it and do it in another PR.

src/core/packet.h
anrossi: after offline discussion, I'll revert to `_V1` and include a comment at the definition which explains that the struct is used for versions 1 and 2.

src/core/stream_send.c
anrossi: let's see if the current changes improve the perf tests, and consider adding to the Building as a performance improvement.

src/core/unittest/VersionNegExtTest.cpp
anrossi: It's not supposed to be a union, but I see now that QUIC_CONNECTION includes the QUIC_HANDLE, so I'll remove the struct and just use a regular QUIC_CONNECTION struct on the stack.

anrossi:
> > > for now, otherwise it shows as a break if they ever negotiate v2.
> > 
> > 
> > interop server currently prefers version 2, but it will work with version 1. If a client supports version 2 and version 1, and starts with version 1, the server will upgrade it to version 2 via CVN (if the enableVNE flag is passed on commandline) Does this solve the concern?
> 
> As far as I understand, the interop runner simply doesn't support v2, so any connection that uses it will confuse it and result in a failure.

Ok, then I think we're safe; we don't enable VNE in the interop runner, so we don't run any risk of accidentally upgrading a connection to V2. and our client won't start with V2. Other clients may start with V2, but we can't control them.

src/core/packet.h
```diff
@@ -266,7 +281,7 @@ QuicPacketIsHandshake(
 _IRQL_requires_max_(DISPATCH_LEVEL)
 _Success_(return != FALSE)
 BOOLEAN
-QuicPacketValidateLongHeaderV1(
+QuicPacketValidateLongHeaderV1V2(
```
nibanks, line 284: You forgot this `V1V2`

src/core/packet.h
```diff
@@ -398,9 +413,9 @@ inline
 _IRQL_requires_max_(DISPATCH_LEVEL)
 _Success_(return != 0)
 uint16_t
-QuicPacketEncodeLongHeaderV1(
+QuicPacketEncodeLongHeaderV1V2(
```
nibanks, line 416: Another `V1V2`

src/core/packet.h
```diff
@@ -490,7 +508,7 @@ QuicPacketGenerateRetryIntegrity(
 _IRQL_requires_max_(DISPATCH_LEVEL)
 _Success_(return != 0)
 uint16_t
-QuicPacketEncodeRetryV1(
+QuicPacketEncodeRetryV1V2(
```
nibanks, line 511: Another `V1V2`

nibanks (approved):

PR #2552, opened 2022-03-18
Send MAX_DATA on Stream FIN

nibanks:
## Description

Update the connection-wide flow control logic to also send updates when a stream shuts down (drains a FIN).

## Testing

TBD

### DIFF
--- a/src/core/stream_recv.c
+++ b/src/core/stream_recv.c
@@ -1000,6 +1000,13 @@ QuicStreamReceiveComplete(
         QuicStreamTryCompleteShutdown(Stream);
 
         //
+        // Flush the connection-wide MAX_DATA.
+        //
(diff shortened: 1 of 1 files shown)


src/core/stream_recv.c
```diff
@@ -999,6 +999,13 @@ QuicStreamReceiveComplete(
         //
         QuicStreamTryCompleteShutdown(Stream);
 
+        //
+        // Flush the connection-wide MAX_DATA.
+        //
+        QuicSendSetSendFlag(
```
anrossi, line 1005: When is a `MAX_DATA` sent otherwise? when stream drains that are not `FIN` happen?

nibanks:
Going to use Matt's PR instead.

PR #2553, opened 2022-03-18
Accumulator logic for MAX_DATA updates

maolson-msft:
When many small streams are used, it's possible that we never actually send MAX_STREAM_DATA updates. The current logic ties MAX_STREAM_DATA and (connection-wide) MAX_DATA updates together, which causes the peer to be eventually connection FC blocked in this case.

This PR adds an accumulator variable for connectionwide delivered bytes. When the accumulator hits a fraction (I propose 1/4 but haven't carefully thought about it) of the conn FC window, we send a MAX_DATA frame to open the window.

### DIFF
--- a/src/core/stream_recv.c
+++ b/src/core/stream_recv.c
@@ -629,13 +629,15 @@ QuicStreamRecv(
 }
 
 //
-// Generally, every time bytes are delivered to the application we update our max
-// data (stream and connection) values and queue an update to be sent to the
-// peer. It is done every time, because nearly always an ACK frame is also
-// ready to be sent out, so we might as well take advantage of that packet to
-// send this data as well. If we don't have an ACK ready to be sent out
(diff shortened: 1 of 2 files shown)


src/core/stream_recv.c
```diff
@@ -726,9 +737,6 @@ QuicStreamOnBytesDelivered(
     Stream->MaxAllowedRecvOffset =
         Stream->RecvBuffer.BaseOffset + Stream->RecvBuffer.VirtualBufferLength;
 
-    QuicSendSetSendFlag(
-        &Stream->Connection->Send,
-        QUIC_CONN_SEND_FLAG_MAX_DATA);
```
nibanks, line 731: Let's just keep this too. It doesn't really hurt.

nibanks (approved):

Issue #2541, opened 2022-03-17
StreamReceiveSetEnabled cannot be used to disable a peer started stream upon start


ThadHouse:
StreamReceiveSetEnabled always delegates to the operation queue, even if running from the worker thread. This means if called from `QUIC_CONNECTION_EVENT_PEER_STREAM_STARTED` to disable stream receive, a receive can happen before the operation can run, making the call pointless. At minimum we need to add this quirk to the documentation, preferably I'd like to fix it.

nibanks:
The best solution without changing any MsQuic code is to simply reject the first receive (drain zero bytes).

thhous-msft:
We fixed the documentation here. I don't think fixing the underlying issue is actually feasable or worth doing.

PR #2565, opened 2022-03-21
Add note to StreamReceiveSetEnabled that disable does not disable enqueued receives

thhous-msft:
Currently, StreamReceiveSetEnabled with IsEnabled set to false won't stop all receives from coming in. It always goes into the worker queue, and there could be receives pending either in the queue or in the current processing packet, which would still be received by the app. Add a note to the docs that this is the case, and the only way to guarantee disable is to drain no bytes from a receive callback.

Noticed in #2541

### DIFF
--- a/docs/api/StreamReceiveSetEnabled.md
+++ b/docs/api/StreamReceiveSetEnabled.md
@@ -25,7 +25,7 @@ The function returns a [QUIC_STATUS](QUIC_STATUS.md). The app may use `QUIC_FAIL
 
 # Remarks
 
-**TODO**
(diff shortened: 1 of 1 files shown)


docs/api/StreamReceiveSetEnabled.md
```diff
@@ -25,7 +25,7 @@ The function returns a [QUIC_STATUS](QUIC_STATUS.md). The app may use `QUIC_FAIL
 
 # Remarks
 
-**TODO**
+This function always delegates to the worker queue, even if called from a quic worker thread. This matters if disabling receives, as there could be a receive in the queue before this call is processed, and that receive would still indicated to the app. To disable receives safely, reject a receive by draining 0 bytes in the `QUIC_STREAM_EVENT_RECEIVE` callback.
```
nibanks, line 28: I would prefer not using the word `safely` is it implies using the function is inherently unsafe. Maybe `reliably` instead.

nibanks (approved):

PR #2566, opened 2022-03-21
Only compute TailBuffer in FreeBuffer on debug builds

thhous-msft:
Newer versions of clang are detecting TailBuffer as unused in release mode, as its only ever set to, and not read from. Make it so the entire variable is gone in release mode. Will need to be backported.

### DIFF
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -2380,16 +2380,22 @@ CxPlatSendDataFreeBuffer(
     // This must be the final send buffer; intermediate buffers cannot be freed.
     //
     CXPLAT_DATAPATH_PROC_CONTEXT* DatapathProc = SendData->Owner;
+#ifdef DEBUG
     uint8_t* TailBuffer = SendData->Buffers[SendData->BufferCount - 1].Buffer;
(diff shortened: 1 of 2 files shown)


src/platform/datapath_epoll.c
```diff
@@ -2380,16 +2380,22 @@ CxPlatSendDataFreeBuffer(
     // This must be the final send buffer; intermediate buffers cannot be freed.
     //
     CXPLAT_DATAPATH_PROC_CONTEXT* DatapathProc = SendData->Owner;
+#ifdef DEBUG
     uint8_t* TailBuffer = SendData->Buffers[SendData->BufferCount - 1].Buffer;
```
nibanks, line 2384: I'm wondering how important this is to assert at all. Thoughts?

thhous-msft: I think its good to have, to know we are not passing in the incorrect buffer to free. But realistically I think there are unlikely to be bugs here.

nibanks: Ok.

nibanks (approved):

PR #2572, opened 2022-03-23
Enable many more tests for XDP

maolson-msft:
Now that the unit tests are enabled on the XDP platform, this PR enables most of the main
tests. There are still five sets of tests that have failures and will need to be fixed and enabled,
but to split up the work these tests are disabled in this PR by adding them to the filter defined
in the XDP test job in azure-pipelines.ci.yml.

### DIFF
--- a/src/inc/msquic_winuser.h
+++ b/src/inc/msquic_winuser.h
@@ -250,35 +250,10 @@ QuicAddrSetToLoopback(
     )
 {
     if (Addr->si_family == QUIC_ADDRESS_FAMILY_INET) {
-
-#ifdef DUONIC_TESTING
-        if (UseDuoNic) {
-            // 192.168.1.11
-            Addr->Ipv4.sin_addr.S_un.S_addr = 184658112;
-        } else {
-            Addr->Ipv4.sin_addr.S_un.S_un_b.s_b1 = 127;
-            Addr->Ipv4.sin_addr.S_un.S_un_b.s_b4 = 1;
-        }
-#else
         Addr->Ipv4.sin_addr.S_un.S_un_b.s_b1 = 127;
         Addr->Ipv4.sin_addr.S_un.S_un_b.s_b4 = 1;
-#endif
-
     } else {
-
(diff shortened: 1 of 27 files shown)


src/inc/msquic_winuser.h
```diff
@@ -250,10 +250,35 @@ QuicAddrSetToLoopback(
     )
 {
     if (Addr->si_family == QUIC_ADDRESS_FAMILY_INET) {
+
+#ifdef DUONIC_TESTING
```
nibanks, line 254: These are our public API headers that we ship. I'd prefer a solution that doesn't modify them for DuoNic

nibanks: Maybe define test-only wrappers (msquichelper.h?) that does what you want, but only is called by test code.

maolson-msft: I've updated the test code to override LOCALHOST_FOR_AF, and to rewrite the QuicAddrs that are created with a boolean param (AKA the loopback ones), to remove this stuff from the msquic API header.

src/test/lib/precomp.h
```diff
@@ -5,6 +5,7 @@
 
 --*/
 
+
```
nibanks, line 8: nit: remove ws

src/platform/unittest/DataPathTest.cpp
```diff
@@ -19,6 +19,12 @@
 
 extern bool UseDuoNic;
 
+//
+// Connect to the duonic address (if using duonic) or localhost (if not).
+//
+#define QUIC_TEST_LOOPBACK_FOR_AF(Af) (UseDuoNic ? ((Af == QUIC_ADDRESS_FAMILY_INET) ? "192.168.1.11" : "fc00::1:11") : QUIC_LOCALHOST_FOR_AF(Af))
+
+
```
nibanks, line 27: nit: remove the extra ws

nibanks (approved):
Assuming all the tests pass, ship it!

PR #2574, opened 2022-03-24
Implements stats v2 struct and missing stream events

Sparika:
## Description
This is a minor change to the lib.rs file to support the connection stats v2 function and struct. In addition, the PR also implements some StreamEvents types that were missing. 

## Testing
No tests implemented, porting of the C structs were manually converted to Rust.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -871,6 +871,42 @@ pub struct StreamEventSendComplete {
 
 #[repr(C)]
 #[derive(Debug, Copy, Clone)]
+pub struct StreamEventPeerSendAborted {
+    pub error_code: u64,
+}
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct StreamEventPeerReceiveAborted {
+    pub error_code: u64,
+}
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct StreamEventSendShutdownComplete {
+    pub graceful: bool,
+}
+
+
+bitfield! {
+    #[repr(C)]
+    #[derive(Clone, Copy)]
+    struct StreamEventShutdownCompleteBitfields(u8);
+    //impl Debug; // TODO Debug causing panic when fmt is called
+    // The fields default to u8
+    app_close_in_progress, _: 1, 0;
+    _reserved, _: 7, 1;
+}
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct StreamEventShutdownComplete {
+    connection_shutdown: bool,
+    flags: StreamEventShutdownCompleteBitfields
+}
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
 pub struct StreamEventIdealSendBufferSize {
     pub byte_count: u64,
 }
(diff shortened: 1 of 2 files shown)


Sparika:
The PR includes submodules changes. I don't know what caused the changes and if I should do something about it.

nibanks:
> The PR includes submodules changes. I don't know what caused the changes and if I should do something about it.

Yeah, we need to remove these changes. Working with submodules is always a pain. I think if you simple `cd` to each directory you changed and do a `git checkout` of the [commit that's on GitHub](https://github.com/microsoft/msquic/tree/main/submodules), it should fix things.

src/lib.rs
```diff
@@ -444,43 +466,116 @@ pub struct QuicStatisticsMisc {
     pub key_update_count: u32,
 }
 
+bitfield! {
+    #[repr(C)]
+    #[derive(Serialize, Deserialize, Clone, Copy)]
+    pub struct QuicStatisticsBitfields(u32);
+    //impl Debug; // TODO Debug causing panic when fmt is called
```
nibanks, line 473: What's this about? I don't understand it.

src/lib.rs
```diff
@@ -444,43 +466,116 @@ pub struct QuicStatisticsMisc {
...
+    // The fields default to u32
+    version_negotiation, _: 1, 0;
+    stateless_retry, _: 1, 1;
+    resumption_attempted, _: 1, 2;
+    resumption_succeeded, _: 1, 3;
+}
+
+impl fmt::Debug for QuicStatisticsBitfields {
```
nibanks, line 481: Would you mind putting a comment to explain this?

src/lib.rs
```diff
@@ -444,43 +466,116 @@ pub struct QuicStatisticsMisc {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_fmt(format_args!("{:#06x}", &self.0))
+    }
+}
+
+/// A helper struct for accessing connection statistics
 #[repr(C)]
 #[derive(Serialize, Deserialize, Debug, Copy, Clone)]
+pub struct QuicStatisticsV2 {
```
nibanks, line 490: Put v2 after v1?

src/lib.rs
```diff
@@ -444,43 +466,116 @@ pub struct QuicStatisticsMisc {
...
+    pub recv_total_bytes: u64,
+    /// Sum of stream payloads
+    pub recv_total_stream_bytes: u64,
+    /// Count of packet decryption failures.
+    pub recv_decryption_failures: u64,
+    /// Count of receive ACK frames.
+    pub recv_valid_ack_frames: u64,
+
+    pub key_update_count: u32,
+    // N.B. New fields must be appended to end
```
nibanks, line 547: Considering this will never be updated separate from the main API, these type of comments aren't necessary IMO.

src/lib.rs
Sparika: @nibanks I will add a more explicit comment.

Deriving Debug is useful for quicky being able to print a struct on stdout such as with `println!("{:?}", my_connection_stats);`. So `QuicStatistics` and `QuicStatisticsV2` both derive Debug. This requires that all struct contained also implement Debug.

Rust does not have a standard equivalent to C's bitfields so I had to rely on the bitfield macro to generate the equivalent bitfields within this QuicStatisticsBitfields struct. The macro allows to derive Debug for the generated struct. However, I observed a panic when printing QuicStatisticsBitfields. I considered it simpler to just manually implement Debug for the QuicStatisticsBitfields by printing the raw u32 instead of deriving pretty printing for each bit fields.

nibanks:
I am very confused as to why `cargo` action hasn't run automatically. :|

nibanks:
@sparika looks like just changing the branch isn't making git happy. Perhaps just create a new branch based on `main` and then copy/paste your Rust changes over and push that instead. Then I think everything will be good to go.

Sparika:
@nibanks I've opened a new PR based on main
https://github.com/microsoft/msquic/pull/2577

PR #2577, opened 2022-03-25
[RUST] Implements stats v2 struct and missing stream events 

Sparika:
## Description
This is a minor change to the lib.rs file to support the connection stats v2 function and struct. In addition, the PR also implements some StreamEvents types that were missing. 

## Testing
No tests implemented, porting of the C structs were manually converted to Rust.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "msquic"
-version = "2.0.1-beta"
+version = "2.1.0-beta"
 edition = "2018"
 authors = ["Microsoft"]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "msquic"
-version = "2.1.0-beta"
+version = "2.0.1-beta"
```
nibanks, line 3: Undo this please.

Sparika: I even see it when preparing the new branch but then I went too fast...

nibanks (approved):
Thanks for doing this!

PR #2591, opened 2022-03-30
Stream Shutdown on Failure in Operation Cleanup

nibanks:
## Description

While inspecting the code I noticed that we only were apply the logic for `QUIC_STREAM_START_FLAG_SHUTDOWN_ON_FAIL` in the failure path of `QuicStreamStart`, and not in the operation clean up path of `QuicOperationQueueClear` which happens when a connection is cleaned up.

## Testing

As this is a nearly impossible timing window to hit with a test, I will rely on spinquic and existing tests to hopefully exercise.

## Documentation

None. This is just fixing a bug.


nibanks:
FYI, this will need to be backported to `release/2.0`.

PR #2603, opened 2022-04-03
Add bandwidth limiter for secnetperf

Wizmann:
Add rate limiter for secnetperf to test MsQuic behavior under different sceanrio

// It's just a demo. Don't know if you like this idea or create a seperate tool like iperf.


nibanks:
I'll have to think about this change. But before we could accept it, the spec this tool follows would have to be updated: https://datatracker.ietf.org/doc/html/draft-banks-quic-performance.

PR #2606, opened 2022-04-04
Fix openssl tls initialize not verifying server vs client state checks

thhous-msft:
## Description

These states should be matched, and this is checked in schannel

## Testing

Existing tests will likely cover this, although they are going to fail. And likely fail downlevel as well.

## Documentation

No

### DIFF
--- a/src/platform/tls_openssl.c
+++ b/src/platform/tls_openssl.c
@@ -1594,6 +1594,16 @@ CxPlatTlsInitialize(
 
     CXPLAT_DBG_ASSERT(Config->HkdfLabels);
 
+    if (Config->IsServer != !(Config->SecConfig->Flags & QUIC_CREDENTIAL_FLAG_CLIENT)) {
+        QuicTraceEvent(
+            TlsError,
+            "[ tls][%p] ERROR, %s.",
+            Config->Connection,
+            "Mismatched SEC_CONFIG IsServer state");
+        Status = QUIC_STATUS_INVALID_PARAMETER;
+        goto Exit;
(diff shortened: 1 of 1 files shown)


nibanks (dismissed):

anrossi (approved):

thhous-msft:
So this bug is actually user visible. It basically means that openssl builds ignore the flag, and will happily apply a server credential to a client. I think this is a bug worth fixing, and backporting down to 2.0 for the downlevel tests to pass.

nibanks:
> So this bug is actually user visible. It basically means that openssl builds ignore the flag, and will happily apply a server credential to a client. I think this is a bug worth fixing, and backporting down to 2.0 for the downlevel tests to pass.

I agree. Let's fix this in release/2.0 first. Do we want to fully publish a new release, or "cheat" and just update the test binaries for 2.0.2?

thhous-msft:
Lets just push a release. I'll finish up the fix, and then work on a backport.

thhous-msft:
Lets just push a release. I'll finish up the fix, and then work on a backport.

thhous-msft:
@anrossi I'm going to need your help finishing this. The pfx certificate does not properly validate.

anrossi:
Talked with Thad about how to unblock him and finish this up. It's going to be a larger change to fix the tests, but worth it long-term.

thhous-msft:
This PR is nowhere near ready to go. There is a ton to do. We need to get a pcks12 writer into the C code to actually complete it.

PR #2613, opened 2022-04-06
Block Certain Source Ports

nibanks:
## Description

After discussing with the QUIC WG (on Slack) I am taking their recommendation to block certain client source ports to prevent possible DoS attacks from our servers.

## Testing

Added test case to validate.

## Documentation

N/A

### DIFF
--- a/src/core/binding.c
+++ b/src/core/binding.c
@@ -1391,7 +1391,8 @@ QuicBindingDropBlockedSourcePorts(
 
     //
     // These UDP source ports are recommended to be blocked by the QUIC WG. See
-    // draft-ietf-quic-applicability for more details.
+    // draft-ietf-quic-applicability for more details on the set of ports that
+    // may cause issues.
     //
     // N.B - This list MUST be sorted in decreasing order.
     //
@@ -1402,8 +1403,8 @@ QuicBindingDropBlockedSourcePorts(
         500,    // IKE
         389,    // CLDAP
         161,    // SNMP
-        128,    // NETBIOS Datagram Service
         137,    // NETBIOS Name Service
+        128,    // NETBIOS Datagram Service
         123,    // NTP
         111,    // Portmap
         53,     // DNS
--- a/src/core/binding.c
+++ b/src/core/binding.c
@@ -1380,6 +1380,51 @@ Exit:
     return Connection;
 }
 
+_IRQL_requires_max_(DISPATCH_LEVEL)
+BOOLEAN
+QuicBindingDropBlockedSourcePorts(
+    _In_ QUIC_BINDING* Binding,
+    _In_ const CXPLAT_RECV_DATA* const Datagram
+    )
+{
+    const uint16_t SourcePort = QuicAddrGetPort(&Datagram->Route->RemoteAddress);
+
+    //
+    // These UDP source ports are recommended to be blocked by the QUIC WG. See
+    // draft-ietf-quic-applicability for more details.
+    //
+    // N.B - This list MUST be sorted in decreasing order.
+    //
+    const uint16_t BlockedPorts[] = {
+        11211,  // memcache
+        5353,   // mDNS
+        1900,   // SSDP
+        500,    // IKE
+        389,    // CLDAP
+        161,    // SNMP
+        128,    // NETBIOS Datagram Service
+        137,    // NETBIOS Name Service
+        123,    // NTP
+        111,    // Portmap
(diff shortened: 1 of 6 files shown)


src/core/binding.c
```diff
@@ -1380,6 +1380,51 @@ QuicBindingCreateConnection(
...
+QuicBindingDropBlockedSourcePorts(
+    _In_ QUIC_BINDING* Binding,
+    _In_ const CXPLAT_RECV_DATA* const Datagram
+    )
+{
+    const uint16_t SourcePort = QuicAddrGetPort(&Datagram->Route->RemoteAddress);
+
+    //
+    // These UDP source ports are recommended to be blocked by the QUIC WG. See
+    // draft-ietf-quic-applicability for more details.
```
maolson-msft, line 1394: I looked at the latest version of draft-ietf-quic-applicability. Section 8.1 ("Source Port Selection") seems to be the relevant part. That section explicitly mentions that serverside blocking of these ports is problematic, and therefore recommends clients avoid the use of these ports for its source port.

But here you seem to be doing serverside port blocking?

Aside from that, isn't this type of thing a job for a firewall? IMO a filter like this does not belong in core QUIC.

nibanks: As I briefly mentioned in the description, I made this PR based on a Slack discussion with two of the largest deployers of QUIC worldwide: Google and Facebook. Both have taken this path of blocking things on the server side. The QUIC protocol has known issues with a couple of these other protocols that can cause infinite loops. The rest have been seen in actual DoS attacks in the wild.

> ... therefore recommends clients avoid the use of these ports for its source port.

We could also block on the client side, but that doesn't protect us when we don't own the client (an attacker isn't going to respect that). Also, it would make testing this harder (but that's not really the point).

> Aside from that, isn't this type of thing a job for a firewall? IMO a filter like this does not belong in core QUIC.

Perhaps a firewall could be responsible for blocking these ports, but until we have such protections in place (especially in Azure) QUIC is the only place we have. Additionally, since there are known protocol incompatibilities with some of these protocols, the QUIC protocol is a good place to handle these issues IMO.

maolson-msft: Have you brought this issue to Azure? I'd like to know if Google and Facebook used firewalls to block these source ports or if they put filters in their core QUIC implementations.

nibanks: > Have you brought this issue to Azure?

I have not brought this specific issue of these ports being blocked to Azure yet. But it will be discussed as a part of the larger DDoS protection effort in Azure.

> I'd like to know if Google and Facebook used firewalls to block these source ports or if they put filters in their core QUIC implementations.

They do block the ports directly in their source.

maolson-msft (dismissed):

src/core/binding.c
nibanks: I tweaked the comment. Let me know if that looks better.

maolson-msft: Yeah I think that's better

maolson-msft (approved):

src/test/lib/HandshakeTest.cpp
```diff
@@ -3032,3 +3032,41 @@ QuicTestResumptionAcrossVersions()
...
+    QuicAddr ServerLocalAddr(QuicAddrFamily);
+
+    MsQuicAutoAcceptListener Listener(Registration, ServerConfiguration, MsQuicConnection::NoOpCallback);
+    TEST_QUIC_SUCCEEDED(Listener.Start("MsQuicTest", &ServerLocalAddr.SockAddr));
+    TEST_QUIC_SUCCEEDED(Listener.GetInitStatus());
+    TEST_QUIC_SUCCEEDED(Listener.GetLocalAddr(ServerLocalAddr));
+
+    MsQuicConnection Client(Registration);
+    TEST_QUIC_SUCCEEDED(Client.GetInitStatus());
+    TEST_QUIC_SUCCEEDED(Client.SetLocalAddr(QuicAddr(QuicAddrFamily, (uint16_t)11211 /* memcache port */)));
```
anrossi, line 3063: Maybe this test should be expanded to try all the ports in the blocked list, to ensure that if the list changes, the test fails?

nibanks: One problem is that Linux doesn't let apps bind to a port <= 1024 by default. You have to run with `sudo`. Also, by doing that, all you are doing is making sure you have to update the test every time. I don't think that's necessary,

PR #2618, opened 2022-04-07
Resolve route in QuicSendFlush and enable LoLa testcases that are fixed by this change

maolson-msft:
There are cases where QuicSendFlush is called without queueing the flush with QuicSendQueueFlush, and cases where the path is switched between the route resolution in QuicSendQueueFlush and the actual flush happening. In both cases the result can be that we try to send a packet on an unresolved route. This is fixed by moving route resolution to QuicSendFlush from QuicSendQueueFlush.

QuicConnProcessRouteCompletion has also been changed to call QuicSendFlush instead of QuicSendQueueFlush. Since the main point of QuicConnProcessRouteCompletion is to finally do a send after the route resolution is finished, there isn't any point in going back to the end of the queue again.


src/core/connection.c
```diff
@@ -5820,13 +5820,8 @@ QuicConnProcessRouteCompletion(
     QUIC_PATH* Path = QuicConnGetPathByID(Connection, PathId, &PathIndex);
     if (Path != NULL) {
         if (Succeeded) {
-            QuicTraceLogConnInfo(
```
maolson-msft, line 5823: FYI, I removed this trace because it's redundant with the one logged by CxPlatResolveRouteComplete.

csujedihy: Sidecar

nibanks: @maolson-msft does the other event log the path ID, like it does here?

src/core/connection.c
```diff
@@ -5835,10 +5830,10 @@ QuicConnProcessRouteCompletion(
...
                     FailedRouteResolution,
                     Connection,
-                    "Processing failed route completion Path[%hhu]",
+                    "Route resolution failed on Path[%hhu]. Switching paths...",
                     PathId);
                 QuicPathSetActive(Connection, &Connection->Paths[1]);
-                QuicSendQueueFlush(&Connection->Send, REASON_ROUTE_COMPLETION);
+                (void)QuicSendFlush(&Connection->Send);
             }
             QuicPathRemove(Connection, PathIndex);
```
csujedihy, line 5838: This seems like another bug. If the above check `Path->IsActive && Connection->PathsCount > 1` passed, the active path we just moved up from index 1 to index 0 would be delete.

I think it should be changed to this:
```
            if (Path->IsActive && Connection->PathsCount > 1) {
                QuicTraceLogConnInfo(
                    FailedRouteResolution,
                    Connection,
                    "Processing failed route completion Path[%hhu]",
                    "Route resolution failed on Path[%hhu]. Switching paths...",
                    PathId);
                QuicPathSetActive(Connection, &Connection->Paths[1]);
                QuicSendQueueFlush(&Connection->Send, REASON_ROUTE_COMPLETION);
                (void)QuicSendFlush(&Connection->Send);
                QuicPathRemove(Connection, 1);
            } else {
                QuicPathRemove(Connection, PathIndex);
           }
```

src/core/connection.c
```diff
@@ -5820,13 +5820,8 @@ QuicConnProcessRouteCompletion(
...
     if (Path != NULL) {
         if (Succeeded) {
-            QuicTraceLogConnInfo(
-                SuccessfulRouteResolution,
-                Connection,
-                "Processing successful route completion Path[%hhu]",
-                PathId);
             CxPlatResolveRouteComplete(Connection, &Path->Route, PhysicalAddress, PathId);
-            QuicSendQueueFlush(&Connection->Send, REASON_ROUTE_COMPLETION);
+            (void)QuicSendFlush(&Connection->Send);
```
nibanks, line 5824: I'm not sure ignoring the return is a good idea. I _think_ we need to queue the send flush if it returns indicating more needs to be sent.

src/core/connection.c
maolson-msft: Yep, it does.

src/core/connection.c
maolson-msft: How about we instead have QuicSendFlush call QuicSendQueueFlush as appropriate? The only benefit that I can see of having QuicSendFlush return a bool is that QuicConnDrainOperations can reuse the same QUIC_OPERATION instead of freeing it and allocating a new QUIC_OPERATION from the pool.

(Note to self: we will need to also unset FlushOperationPending prior to calling QuicSendFlush in QuicConnDrainOperations if we make this change)

src/core/connection.c
maolson-msft: Also, QuicConnProcessRouteCompletion is called with a path ID, which is different from the path index, right? It seems like there's sort of an assumption here that the path QuicConnProcessRouteCompletion is being passed is the path at index 0; is that really a safe assumption?

maolson-msft: Oh no I see, Path->IsActive==TRUE implies that the path is at index 0.

maolson-msft: I'm trying to decide which is better: either your suggested fix above, or instead of
                QuicPathSetActive(Connection, &Connection->Paths[1]);
                QuicPathRemove(Connection, 1);

we could do
                QuicPathRemove(Connection, PathIndex);
                QuicPathSetActive(Connection, &Connection->Paths[0]);

csujedihy: The former seems slightly more readable to me.

maolson-msft: Sounds good to me.

src/core/send.h
```diff
@@ -352,7 +352,7 @@ typedef enum QUIC_SEND_FLUSH_REASON {
     REASON_STREAM_ID_FLOW_CONTROL,
     REASON_AMP_PROTECTION,
     REASON_SCHEDULING,
-    REASON_ROUTE_COMPLETION,
+    REASON_ROUTE_COMPLETION, // TODO: unused
```
nibanks, line 355: If it's not needed, remove it.

src/core/connection.c
nibanks: I think doing the flush inline is ok, but we just need to handle the case where it can't flush everything right now.

src/core/send.h
maolson-msft: I made this TODO because I wasn't sure whether I can just remove the enum value and associated valuemap entry from the tracing manifest. Will it cause issues (e.g. with trace analysis tools) if a new flush reason is added later on and given the same value as this reason I'm removing? If so then instead I should mark it deprecated but not remove it.

src/core/connection.c
maolson-msft: I'm saying instead of handling that case in the caller, we could handle it inside QuicSendFlush. So then we don't have to have the "oops, couldn't flush everything so let's queue another operation" logic in multiple places.

nibanks: I'd rather keep the operation reuse where possible.

src/core/send.h
maolson-msft: Since we are queueing on QuicSendFlush returning FALSE now, this is moot since we still need the queue reason

nibanks (commented):
LGTM. If the CI agrees, ship it.

maolson-msft:
@nibanks you didn't sign off it looks like. The only failure is that irrelevant "0-RTT wasn't used for stream data" one, so we look good here

nibanks (approved):
Oops. Meant to sign off last time.

PR #2619, opened 2022-04-07
Fix a jump anchor in detailed view

csujedihy:


csujedihy:
@nibanks I merged this in directly. Not worth reviewing.

PR #2620, opened 2022-04-08
Fix route resolution error handling

csujedihy:
## Description

The status code handling logic CxPlatResolveRoute is problematic because the status code conversion at the end of the function assumes `NETIO_STATUS Status` is `NT_STATUS`. However, in usermode, `NETIO_STATUS` = WIN32 errors.

Also, set route state to resolving before we queue a non-successful route resolution completion to prevent multiple route resolution jobs get queued up before the completion gets drained.

## Testing

Covered by existing tests.

### DIFF
--- a/src/platform/datapath_raw_socket.c
+++ b/src/platform/datapath_raw_socket.c
@@ -332,6 +332,8 @@ CxPlatResolveRoute(
 
     CXPLAT_DBG_ASSERT(!QuicAddrIsWildCard(&Route->RemoteAddress));
 
+    Route->State = RouteResolving;
+
     //
     // Find the best next hop IP address.
     //
@@ -442,7 +444,6 @@ CxPlatResolveRoute(
     // We queue an operation on the route worker for NS because it involves network IO and
     // we don't want our connection worker queue blocked.
     //
-    Route->State = RouteResolving;
     if ((Status != ERROR_SUCCESS || IpnetRow.State <= NlnsIncomplete) ||
         (State == RouteSuspected &&
          memcmp(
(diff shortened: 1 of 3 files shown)


src/platform/datapath_raw_socket.c
```diff
@@ -475,10 +475,17 @@ CxPlatResolveRoute(
 
 Done:
     if (Status != ERROR_IO_PENDING && Status != ERROR_SUCCESS) {
+        //
+        // Failed to resolve route. Queue route resolution completion with failure and
```
maolson-msft, line 479: This code here doesn't officially know that Callback will "queue" anything, so we shouldn't say anything about the queueing here. This is especially confusing because above in the ERROR_IO_PENDING path we really do queue something here.

csujedihy: Hmm. I can move `Route->State = RouteResolving;` out of the ERROR_IO_PENDING path so we do it regardless. I think it makes sense because as long as route completion has not been called, we are still resolving the route.

csujedihy: and remove the comment

maolson-msft: Good idea.

src/platform/datapath_raw_socket.c
```diff
@@ -442,6 +442,7 @@ CxPlatResolveRoute(
     // We queue an operation on the route worker for NS because it involves network IO and
     // we don't want our connection worker queue blocked.
     //
+    Route->State = RouteResolving;
```
maolson-msft, line 445: Now you can remove the "Route->State = RouteResolving;" in the below code block right?

csujedihy: Oh, I missed that.

csujedihy, line 445: I think this line should be at the top of the function.

maolson-msft (approved):

PR #2626, opened 2022-04-09
Change C# Constants to uint

### LINKED ISSUE #2625: C# generated interop: constant types do not match usages

For exaple: `QUIC_PARAM_CONN_LOCAL_ADDRESS` is defined as int: 
https://github.com/microsoft/msquic/blob/3ebe6d8b6850deed8f6d323158e96e46712c86c9/src/cs/lib/msquic_generated.cs#L2581

But `Get/SetParam` accept `uint`:
https://github.com/microsoft/msquic/blob/3ebe6d8b6850deed8f6d323158e96e46712c86c9/src/cs/lib/msquic_generated.cs#L2406

### Proposed solution

I'm not sure how far the generated interop can be tweaked, so this might not be worth fixing. But it would be nice if the types matched.

### Additional context

_No response_

nibanks:
## Description

Fixes #2625 by post processing to replace `const int` with `const uint`.

## Testing

Existing tests cover.

## Documentation

None

### DIFF
--- a/scripts/generate-dotnet.ps1
+++ b/scripts/generate-dotnet.ps1
@@ -52,6 +52,7 @@ Invoke-Expression "$ToolExe $FullArgs"
     -replace '\(anonymous struct.+\)\"', "(anonymous struct)`"" `
     -replace '\(anonymous union.+\)\"', "(anonymous union)`"" `
     -replace "public enum .*?_FLAGS","[System.Flags]`n    `$0" `
+    -replace 'const int', "const uint" `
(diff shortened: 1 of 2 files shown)


nibanks:
cc @ManickaP

ThadHouse:
LGTM. I'll ask Tanner if there is a way to make the generator generate uints, but I can't find anything in the options, so it probably doesn't currently exist.

ThadHouse (approved):

PR #2633, opened 2022-04-11
Mention C++, C# and Rust in Readme

nibanks:


maolson-msft:
Nit: In case you didn't notice, this info is already mentioned in FAQ.md. Maybe you want to move it from there to here rather than duplicating it.

nibanks:
> Nit: In case you didn't notice, this info is already mentioned in FAQ.md. Maybe you want to move it from there to here rather than duplicating it.

Sure. I will remove it from there.

maolson-msft (approved):

PR #2634, opened 2022-04-11
Add Test for Linux Package

### LINKED ISSUE #2632: Linux Package Validation

We generate RPM and DEB packages as a part of our build process, but we have no validation to ensure they actually work. We need some kind of support for validating these packages.

### Proposed solution

I'm not sure how best to validate the packages, but ideally build some minimal app that just verifies it can open a handle, and then we run it on a VM that first installs the package.

nibanks:
An attempt to fix #2632 with a new stage to run a test case using the installed package

### DIFF
--- a/.azure/templates/test-distribution.yml
+++ b/.azure/templates/test-distribution.yml
@@ -10,6 +10,7 @@ jobs:
     vmImage: ubuntu-latest
 
   steps:
+  - checkout: self
   - task: DownloadBuildArtifacts@0
     displayName: Download Test
     inputs:
--- a/.azure/templates/test-distribution.yml
+++ b/.azure/templates/test-distribution.yml
@@ -24,4 +24,4 @@ jobs:
       itemPattern: distribution/*.deb
       downloadPath: $(Build.SourcesDirectory)
   - pwsh: sudo dpkg -i $(Build.SourcesDirectory)/distribution/libmsquic_2.1.0_amd64.deb
-  - pwsh: $(Build.SourcesDirectory)/artifacts/bin/linux/x64_Release_openssl/msquictest/msquictest --gtest_filter=ParameterValidation.ValidateApi
+  - pwsh: $(Build.SourcesDirectory)/artifacts/bin/linux/x64_Release_openssl/msquictest --gtest_filter=ParameterValidation.ValidateApi
--- a/.azure/templates/test-distribution.yml
+++ b/.azure/templates/test-distribution.yml
@@ -28,8 +28,5 @@ jobs:
       sudo apt-get install -y lttng-tools
       sudo dpkg -i $(Build.SourcesDirectory)/distribution/libmsquic_2.1.0_amd64.deb
   - pwsh: |
-      echo $(Build.SourcesDirectory)/artifacts/bin/linux/x64_Release_openssl/msquictest
-      ldd $(Build.SourcesDirectory)/artifacts/bin/linux/x64_Release_openssl/msquictest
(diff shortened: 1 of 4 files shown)


.azure/templates/test-distribution.yml
```diff
@@ -0,0 +1,33 @@
...
+      downloadPath: $(Build.SourcesDirectory)
+  - pwsh: |
+      sudo apt-add-repository ppa:lttng/stable-2.12
+      sudo apt-get update
+      sudo apt-get install -y lttng-tools
+      sudo dpkg -i $(Build.SourcesDirectory)/distribution/libmsquic_2.1.0_amd64.deb
+  - pwsh: |
+      echo $(Build.SourcesDirectory)/artifacts/bin/linux/x64_Release_openssl/msquictest
+      $(Build.SourcesDirectory)/artifacts/bin/linux/x64_Release_openssl/msquictest --gtest_filter=ParameterValidation.ValidateApi
+      echo Complete
```
nibanks, line 33: @wfurt, @ManickaP do you have any idea why this test wouldn't be working? I only get the following output in Azure Pipelines:
```
========================== Starting Command Output ===========================
/usr/bin/pwsh -NoLogo -NoProfile -NonInteractive -Command . '/home/vsts/work/_temp/2d25da3b-6e8a-460e-85cc-813d6f944367.ps1'
/home/vsts/work/1/s/artifacts/bin/linux/x64_Release_openssl/msquictest
Complete
Finishing: PowerShell
```
If it can't find libmsquic, I'd expect to get an error like this:
```
error while loading shared libraries: libmsquic.so.2: cannot open shared object file: No such file or directory
```
At least that's what happened locally. I went through these steps locally and it worked just fine. Ideas?

wfurt: It is exactly same OS as you build on ? you can try to add `ldd  artifacts/bin/linux/x64_Release_openssl/msquictest` (and possibly `ls -al artifacts/bin/linux/x64_Release_openssl`)

nibanks: @wfurt
```
/home/vsts/work/1/s/artifacts/bin/linux/x64_Release_openssl/msquictest
	linux-vdso.so.1 (0x00007ffce6f8e000)
	libmsquic.so.2 => /lib/x86_64-linux-gnu/libmsquic.so.2 (0x00007f899226c000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f8992266000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f8992084000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f8991f35000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f8991f1a000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f8991ef5000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f8991d03000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f89929a5000)
total 37628
drwxr-xr-x 2 vsts docker     4096 Apr 12 15:59 .
drwxr-xr-x 3 vsts docker     4096 Apr 12 15:59 ..
-rw-r--r-- 1 vsts docker 38521656 Apr 12 15:59 msquictest
Complete
```

thhous-msft (approved):

PR #2637, opened 2022-04-12
Control socket for closing secnetperf in XDP path

csujedihy:
For XDP path, create a thread to do blocking UDP receive to control secnetperf instead of a whole new datapath.

### DIFF
--- a/src/perf/lib/SecNetPerfMain.cpp
+++ b/src/perf/lib/SecNetPerfMain.cpp
@@ -331,6 +331,7 @@ QuicMainFree(
 #ifdef QUIC_USE_RAW_DATAPATH
     if (BackdoorThreadHandle) {
         CxPlatThreadWait(&BackdoorThreadHandle);
+        CxPlatThreadDelete(&BackdoorThreadHandle);
     }
 #else
     if (Datapath) {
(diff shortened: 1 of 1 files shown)


src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -96,6 +100,49 @@ PrintHelp(
         );
 }
 
+#ifdef QUIC_USE_RAW_DATAPATH
+CXPLAT_THREAD_CALLBACK(ControlThread, Context)
+{
+    WSADATA WsaData;
+    WSAStartup(MAKEWORD(2, 2), &WsaData);
```
nibanks, line 107: This all is not cross platform. I want to keep the code cross-platform.

csujedihy: I only changed it for XDP codepath which is windows only. When XDP can coexist with any other datapath, we can simply remove this workaround.

nibanks: While XDP support is currently Windows-only, that won't be true forever, and likely not even in the long term; but I guess we might end up merging the XDP and socket datapaths before we add Linux support. But by doing it this way, you sign up for the possibility of having to do this in a cross-platform way later if we add Linux support before combining the datapaths.

nibanks (approved):

src/perf/lib/SecNetPerfMain.cpp
csujedihy: Yeah, it should be within 5 lines of code change to make this piece of code work for linux if linux support comes sooner than combining datapaths.

nibanks:
Looks like you brought down the perf machines: https://dev.azure.com/ms/msquic/_build/results?buildId=302052&view=results

csujedihy:
> Looks like you brought down the perf machines: https://dev.azure.com/ms/msquic/_build/results?buildId=302052&view=results

My first iteration causes the client to hang because it spins up the control thread for client too.

csujedihy:
> > Looks like you brought down the perf machines: https://dev.azure.com/ms/msquic/_build/results?buildId=302052&view=results
> 
> My first iteration causes the client to hang because it spins up the control thread for client too.

Hmm, this will only happen in XDP path.

PR #2638, opened 2022-04-12
Onboard XDP perf test

csujedihy:


csujedihy:
close it and play it with azure pipeline directly

PR #2646, opened 2022-04-18
Expose Listener close function

panpur:
## Description
Close listener on drop

## Testing
Manual test

## Documentation
No

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1197,7 +1197,7 @@ impl Api {
         }
         Api { table: new_table }
     }
-    
+
     pub fn close(&mut self) {
         unsafe { MsQuicClose(self.table) };
     }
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1487,6 +1491,12 @@ impl Listener {
             panic!("ListenerStart failed, {:x}!\n", status);
         }
     }
+
+    pub fn close(&self) {
```
nibanks, line 1495: Instead of this, I think you should just implement `Drop for Listener`.

src/lib.rs
```diff
@@ -1198,6 +1198,10 @@ impl Api {
         Api { table: new_table }
     }
 
+    pub fn close(&mut self) {
```
nibanks, line 1201: Why does `close` need to be explicitly exposed?

panpur: I thought to add it since Connection & Stream also have explicit close, but anyway I've removed it.

src/lib.rs
panpur: I've moved it to drop function.

nibanks (approved):

PR #2648, opened 2022-04-19
Fix only setting minimum mtu during config open failing

### LINKED ISSUE #2647: Cannot Set Min MTU without also Setting Max MTU
### Describe the bug

When an app sets a custom min MTU on a `QUIC_SETTINGS`, it fails to load.

### Affected OS

- [X] All
- [ ] Windows Server 2022
- [ ] Windows 11
- [ ] Ubuntu
- [ ] Debian
- [ ] Other (specify below)

### Additional OS information

_No response_

### MsQuic version

main

Set min MTU only

### Expected behavior

Should succeed

### Actual outcome

Fails

### Additional details

_No response_

thhous-msft:
## Description

Closes #2647. Setting a minimum mtu only during ConfigurationOpen results in a failure, because the default values are 0. Set the default values to things more sane.

Also fixes a bug where minimum can be higher then maximum, if only 1 value is set on the configuration, and the global value is not within the bounds.

## Testing

New test added to cover case

## Documentation
N/A

### DIFF
--- a/src/core/settings.c
+++ b/src/core/settings.c
@@ -229,12 +229,20 @@ QuicSettingsCopy(
                 QuicSettingsCopyVersionSettings(Source->VersionSettings, FALSE);
         }
     }
-    if (!Destination->IsSet.MinimumMtu) {
+
+    if (!Destination->IsSet.MinimumMtu && !Destination->IsSet.MaximumMtu) {
         Destination->MinimumMtu = Source->MinimumMtu;
-    }
-    if (!Destination->IsSet.MaximumMtu) {
         Destination->MaximumMtu = Source->MaximumMtu;
+    } else if (Destination->IsSet.MinimumMtu && !Destination->IsSet.MaximumMtu) {
+        if (Source->MaximumMtu > Destination->MinimumMtu) {
+            Destination->MaximumMtu = Source->MaximumMtu;
+        }
+    } else if (Destination->IsSet.MaximumMtu && !Destination->IsSet.MinimumMtu) {
+        if (Source->MinimumMtu < Destination->MaximumMtu) {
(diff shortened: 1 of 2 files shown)


src/core/settings.c
```diff
@@ -501,10 +503,8 @@ QuicSettingApply(
         if (MinimumMtu > MaximumMtu) {
             return FALSE;
         }
-        if (Destination->MinimumMtu != MinimumMtu) {
+        if (Source->IsSet.MinimumMtu || Source->IsSet.MaximumMtu) {
             Destination->IsSet.MinimumMtu = TRUE;
-        }
-        if (Destination->MaximumMtu != MaximumMtu) {
```
nibanks, line 507: I don't understand this removal

thhous-msft: Thats to solve the 2nd bug found. Basically if we set the minimum, we have to implictly set the maximum. Otherwise, when the global settings are loaded in, they can result in an out of sync state.

nibanks: If setting the min doesn't modify the max, then I don't quite understand what the problem is here.

nibanks: Generally, we need to decide priority here. Usually the order is parent < registry < app. But if the app only sets the min and the registry needs to set a max, and they don't conflict that should be supported.

thhous-msft: Well then we need to change loading from registry and global to be able to error.

nibanks: Why? And where is that error going to go? Right now, any errors in there fail silently and have fallback code.

thhous-msft: I found a way to do it silently, and added more tests.

nibanks (approved):

nibanks (approved):

PR #2650, opened 2022-04-20
Update C# enums

nibanks:
## Description

Makes the C# enums more in the spirit of the language.


thhous-msft (approved):

TYoung86:
Could *maybe* get a more generic replacement by this sort of expression;
https://regex101.com/r/EYSIzP/1
Would need to be repeated until no further changes to the document.

You'd still need special replacement for e.g. "QUIC_STREAM_OPEN_FLAG_0_RTT" => "ZERO_RTT"

PR #2653, opened 2022-04-21
Change .NET interop to internal, other type changes

thhous-msft:
## Description

Changes made by the .NET team in https://github.com/dotnet/runtime/pull/68288


nibanks (approved):

anrossi (dismissed):

src/cs/lib/msquic_generated.cs
```diff
@@ -2380,294 +2384,294 @@ public byte RESERVED
...
 
         [NativeTypeName("#define QUIC_PARAM_STREAM_PRIORITY 0x08000003")]
-        public const uint QUIC_PARAM_STREAM_PRIORITY = 0x08000003;
+        internal const uint QUIC_PARAM_STREAM_PRIORITY = 0x08000003;
 
         [NativeTypeName("#define QUIC_API_VERSION_2 2")]
-        public const uint QUIC_API_VERSION_2 = 2;
+        internal const uint QUIC_API_VERSION_2 = 2;
     }
 }
```
ManickaP, line 2677:
```suggestion
}
#pragma warning restore CS0649
```

src/cs/lib/msquic.cs
```diff
@@ -10,12 +12,22 @@
 
 namespace Microsoft.Quic
 {
-    public unsafe partial struct QUIC_BUFFER
+#if PUBLIC_MSQUIC_INTEROP_ACCESS
```
ManickaP, line 15: Is this still necessary when you have `InternalsVisibleTo` ? Or is there a reason for different approach in this file and `*_generate*.cs`?

ManickaP (approved):
Thanks!

src/cs/lib/msquic.cs
nibanks: How does `InternalsVisibleTo` work? That sounds like it might be a better solution.

ManickaP: It makes types/fields/methods marked `internal` to look like `public` for assemblies named in the `InternalsVisibleTo` attribute. Disadvantage is that your interop lib must know the consumer lib/executable name beforehand.

thhous-msft: Yeah I meant to revert these back. For our test specifically InternalsVisableTo is fine.

src/cs/lib/msquic_generated.cs
thhous-msft: Thats actually difficult to do in the generator. Is it needed to restore, even though its a generated file?

nibanks (approved):

raghavthind2005 (approved):

src/cs/lib/msquic_generated.cs
ManickaP: Not really, it should get restored at the end of the file anyway.

ManickaP (approved):
Thanks

PR #2655, opened 2022-04-22
Switch epoll and kqueue to level triggers

### LINKED ISSUE #2654: Poll based datapaths can hang sends in race condition with kernel
### Describe the bug

For sends in the datapath, we don't enable the edge triggers until after a socket write returns EAGAIN or EWOULDBLOCK. However, if sends in the kernel become enabled between calling send() and epoll_ctl (Or equivalent in other paths), the epoll event would never tirgger, and the sends would be hung forever for that socket.

### Affected OS

- [ ] All
- [ ] Windows Server 2022
- [ ] Windows 11
- [X] Ubuntu
- [ ] Debian
- [X] Other (specify below)

### Additional OS information

kqueue and epoll datapaths, in addition to PSN on windows

### MsQuic version

main

Race condition

### Expected behavior

Sends can't hang

### Actual outcome

Sends can hang

### Additional details

The solution to this will be to always keep the edge triggers enabled. Then before we send, enqueue the send into the pending queue. If the send passes, dequeue it, otherwise just continue, and the datapath thread will pick up the send later.

thhous-msft:
## Description

kqueue and epoll datapaths have a race condition upon pended sends whereas they might not get trigger notifications. By switching to level triggers, we solve this issue.

Closes #2654

## Testing

Existing tests cover the change.

## Documentation

N/A

### DIFF
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -1772,9 +1772,9 @@ CxPlatSocketContextProcessEvents(
 
     if (EPOLLIN & Events) {
         //
-        // Read up to 1 receives before moving to another event.
+        // Read up to 4 receives before moving to another event.
         //
-        for (int i = 0; i < 1; i++) {
+        for (int i = 0; i < 4; i++) {
 
             for (ssize_t i = 0; i < CXPLAT_MAX_BATCH_RECEIVE; i++) {
                 CXPLAT_DBG_ASSERT(SocketContext->CurrentRecvBlocks[i] != NULL);
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -1431,7 +1431,7 @@ CxPlatSocketContextStartReceive(
     }
 
     struct epoll_event SockFdEpEvt = {
-        .events = EPOLLIN | EPOLLLT,
+        .events = EPOLLIN,
         .data = {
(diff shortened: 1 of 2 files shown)


thhous-msft:
Don't merge this yet, I want to do more local testing. Its just easy for CI to test everything as well.

anrossi (dismissed):

nibanks:
I still have my reservations about making changes here, because we've never seen an issue with the current code. Is there any test we can write (maybe at the datapath unit test layer) to prove/disprove this change?

thhous-msft:
Well we validate that it works, and doesn't break performance. If it was going to be problematic, it would have triggered an issue with those cases. And using a TCP test, validated that the logic holds.

The issue is its basically impossible to trigger EAGAIN on write for UDP. I've tried looking if theres a way to force it, but I can't find any.

nibanks (dismissed):

nibanks:
Ok. I leave you in charge of the final decision!

thhous-msft:
The discussion created in https://github.com/microsoft/msquic/discussions/2662 is another reason for doing this change. Level triggered mode is required to make it so the entire receive buffer does not have to be read every time an event is received.

nibanks (approved):

anrossi (approved):

nibanks (approved):

PR #2670, opened 2022-04-29
Onboard XDP perf test

csujedihy:
## Description

This PR contains a few bugfixes that are needed to unblock the test. There is a known XDP bug that's causing TX to hang. The bug has been fixed in XDP upstream but we have not yet updated QUIC to consume it. So, in this PR, we work around the bug by always poking TX. Once the perf pipeline is validated, we will update QUIC to use the latest XDP drop. Right now, do not focus on the perf number as the work around has a high cost.


src/core/packet_builder.c
```diff
@@ -68,7 +68,6 @@ QuicPacketBuilderValidate(
         CXPLAT_DBG_ASSERT(Builder->Datagram->Length != 0);
         CXPLAT_DBG_ASSERT(Builder->Datagram->Length <= UINT16_MAX);
         CXPLAT_DBG_ASSERT(Builder->Datagram->Length >= Builder->MinimumDatagramLength);
-        CXPLAT_DBG_ASSERT(Builder->Datagram->Length > (uint32_t)Builder->DatagramLength);
```
nibanks, line 71: I'm still not sure about this.

scripts/performance-helper.psm1
```diff
@@ -1241,18 +1245,16 @@ class TestRunDefinition {
...
         if ($this.VariableName -eq "Default") {
             $VarVal = ""
         }
-        $Platform = $this.Local.Platform
-        if ($script:Kernel -and $this.Local.Platform -eq "Windows") {
-            $Platform = 'Winkernel'
-        }
-        $RetString = "$($this.TestName)_$($Platform)_$($script:RemoteArch)_$($script:RemoteTls)_$($this.VariableName)$VarVal"
+
+        $RetString = "$($this.TestName)_$($this.ToTestPlatformString())_$($this.VariableName)$VarVal"
```
nibanks, line 1257: @thhous-msft can you please review this?

nibanks:
We also need to configure things to use all CPUs and not just 1 (which is the default).

scripts/performance-helper.psm1
thhous-msft: I did. Its fine. Was waiting to approve until everything passed.

src/core/packet_builder.c
thhous-msft: Well its strictly worse then the check right below this. If the below check passes, this one is guaranteed to pass as well (As long as EncryptionOverhead is not somehow negative or 0 and the compiler does signed math).

csujedihy:
> We also need to configure things to use all CPUs and not just 1 (which is the default).

My plan is 1) check this in 2) make changes to the performance branch 3) fix up leftover things (proper XDP fix, using all CPU and etc).

scripts/performance-helper.psm1
csujedihy: windows user openssl perf test failed on the tcp ones. I don't think it's related to this PR.

Running Local: C:\RunningTests\secnetperf.exe Args: -test:Throughput -target:quic-server -bind:192.168.0.2:4434 -ip:4 -uni:1 -download:2000000000 -tcp:1
Running Locally: C:\RunningTests\secnetperf.exe -watchdog:115000 "--extraOutputFile:C:\RunningTests\ExtraRunFile.txt" -test:Throughput -target:quic-server -bind:192.168.0.2:4434 -ip:4 -uni:1 -download:2000000000 -tcp:1
Started!

Started!

App Main returning status 0

nibanks (approved):

csujedihy:
@nibanks  @thhous-msft the official perf run for winkernel hit system 577.

PR #2678, opened 2022-05-03
Use all processors for XDP perf test

csujedihy:
## Description

Change secnetperf to take `-cpu:-1` as all procs used by raw datapath. Also, change perf test script to pass in `-cpu:-1` for XDP.

## Testing

Covered by existing tests.

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -5453,7 +5453,7 @@ QuicConnRecvDatagrams(
 
 #ifdef QUIC_USE_RAW_DATAPATH
         if (DatagramPath->Route.State == RouteResolved && !DatagramPath->Route.QueueUpdated) {
-            Datagram->Route->Queue;
+            DatagramPath->Route.Queue = Datagram->Route->Queue;
             DatagramPath->Route.QueueUpdated = TRUE;
         }
 #endif
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -5452,10 +5452,10 @@ QuicConnRecvDatagrams(
         }
 
 #ifdef QUIC_USE_RAW_DATAPATH
-    if (!DatagramPath->Route.QueueUpdated) {
-        Datagram->Route->Queue;
-        DatagramPath->Route.QueueUpdated = TRUE;
-    }
+        if (DatagramPath->Route.State == RouteResolved && !DatagramPath->Route.QueueUpdated) {
+            Datagram->Route->Queue;
+            DatagramPath->Route.QueueUpdated = TRUE;
+        }
 #endif
 
         if (DatagramPath != CurrentPath) {
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -5451,6 +5451,13 @@ QuicConnRecvDatagrams(
             goto Drop;
         }
 
+#ifdef QUIC_USE_RAW_DATAPATH
+    if (!DatagramPath->Route.QueueUpdated) {
+        Datagram->Route->Queue;
+        DatagramPath->Route.QueueUpdated = TRUE;
+    }
+#endif
+
         if (DatagramPath != CurrentPath) {
             if (BatchCount != 0) {
                 //
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -5452,9 +5452,9 @@ QuicConnRecvDatagrams(
         }
 
 #ifdef QUIC_USE_RAW_DATAPATH
-        if (DatagramPath->Route.State == RouteResolved && !DatagramPath->Route.QueueUpdated) {
+        if (DatagramPath->Route.State == RouteResolved &&
+            DatagramPath->Route.Queue != Datagram->Route->Queue) {
             DatagramPath->Route.Queue = Datagram->Route->Queue;
-            DatagramPath->Route.QueueUpdated = TRUE;
         }
 #endif
 
--- a/src/inc/quic_datapath.h
+++ b/src/inc/quic_datapath.h
@@ -161,7 +161,15 @@ typedef struct CXPLAT_ROUTE {
     uint8_t NextHopLinkLayerAddress[6];
     void* Queue;
 
-    CXPLAT_ROUTE_STATE State; // Keep this as the last property in the struct.
+    //
+    // QuicCopyRouteInfo copies memory up to this point (not including State).
+    //
+
+    CXPLAT_ROUTE_STATE State;
+    //
+    // Indicates the queue has been updated for this route.
+    //
+    uint8_t QueueUpdated : 1;
 #endif // QUIC_USE_RAW_DATAPATH
 
 } CXPLAT_ROUTE;
@@ -703,6 +711,7 @@ CxPlatResolveRoute(
     _In_ void* Context,
     _In_ CXPLAT_ROUTE_RESOLUTION_CALLBACK_HANDLER Callback
     );
+
 #endif // QUIC_USE_RAW_DATAPATH
 
 #if defined(__cplusplus)
--- a/src/inc/quic_datapath.h
+++ b/src/inc/quic_datapath.h
@@ -166,10 +166,6 @@ typedef struct CXPLAT_ROUTE {
     //
 
     CXPLAT_ROUTE_STATE State;
-    //
-    // Indicates the queue has been updated for this route.
-    //
-    uint8_t QueueUpdated : 1;
 #endif // QUIC_USE_RAW_DATAPATH
 
 } CXPLAT_ROUTE;
--- a/src/perf/lib/SecNetPerfMain.cpp
+++ b/src/perf/lib/SecNetPerfMain.cpp
@@ -16,7 +16,9 @@ Abstract:
 #include "HpsClient.h"
 #include "Tcp.h"
 
+#ifndef _KERNEL_MODE
 #include <vector>
+#endif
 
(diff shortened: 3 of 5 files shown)


nibanks (dismissed):

src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -245,19 +249,26 @@ QuicMainStart(
 #ifndef _KERNEL_MODE
```
csujedihy, line 249: @nibanks is there a reason the code block is behind `_KERNEL_MODE` rather than `QUIC_USE_RAW_DATAPATH `?

src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -245,19 +249,26 @@ QuicMainStart(
...
     const char* CpuStr;
     if ((CpuStr = GetValue(argc, argv, "cpu")) != nullptr) {
-        uint16_t ProcList[64];
-        uint32_t ProcCount = 0;
-        do {
-            if (*CpuStr == ',') CpuStr++;
-            ProcList[ProcCount++] = (uint16_t)strtoul(CpuStr, (char**)&CpuStr, 10);
-        } while (*CpuStr && ProcCount < ARRAYSIZE(ProcList));
+        std::vector<uint16_t> ProcList;
+        if (strtoul(CpuStr, nullptr, 10) == -1) {
```
csujedihy, line 253: oops, strtoul returns an unsigned long...

src/perf/lib/SecNetPerfMain.cpp
nibanks: Because of `std::vector` isn't available in kernel.

csujedihy: I changed it from an on stack array to a vector in this PR. The code before this PR uses `_KERNEL_MODE`.

nibanks: Ah, you're right. It was because I never expected XDP to be used in kernel, but other user mode variants might.

csujedihy:
I am ready to check this in. The failure in XDP perf test is because perf regressed.

csujedihy:
OK. I thought this regression was known. Then, there's much more to do to get the metrics better.

csujedihy:
image
The 40-connection latency figure seems to indicate that shared EC isn't a good idea when there are lots of connections. I speculate a lot of connections doing core QUIC work actually starves the datapath.

src/core/connection.c
```diff
@@ -5451,6 +5451,13 @@ QuicConnRecvDatagrams(
             goto Drop;
         }
 
+#ifdef QUIC_USE_RAW_DATAPATH
+        if (DatagramPath->Route.State == RouteResolved && !DatagramPath->Route.QueueUpdated) {
+            Datagram->Route->Queue;
+            DatagramPath->Route.QueueUpdated = TRUE;
```
nibanks, line 5457: You're not setting the queue here??

csujedihy: lol, some code accidentally got eaten.

csujedihy: fixed

src/inc/quic_datapath.h
```diff
@@ -161,7 +161,15 @@ typedef struct CXPLAT_ROUTE {
...
-    CXPLAT_ROUTE_STATE State; // Keep this as the last property in the struct.
+    //
+    // QuicCopyRouteInfo copies memory up to this point (not including State).
+    //
+
+    CXPLAT_ROUTE_STATE State;
+    //
+    // Indicates the queue has been updated for this route.
+    //
+    uint8_t QueueUpdated : 1;
```
nibanks, line 172: Why do we need a flag and not just treat a null `Queue`?

csujedihy: I was trying to imitate PartitionUpdate so that queue only gets updated once after route is resolved.

maolson-msft: Do we need the queue to be non-null between the time that the route is set to RouteSuspected and the time that the new queue is assigned? If not then I also think just setting the queue to null would be better than adding a flag.

csujedihy: Are you suggesting this?
```C
if (DatagramPath->Route.State == RouteResolved && !DatagramPath->Route.Queue) {
    DatagramPath->Route.Queue = Datagram->Queue;
}
```

src/core/connection.c
```diff
@@ -5451,6 +5451,13 @@ QuicConnRecvDatagrams(
             goto Drop;
         }
 
+#ifdef QUIC_USE_RAW_DATAPATH
+        if (DatagramPath->Route.State == RouteResolved && !DatagramPath->Route.QueueUpdated) {
+            DatagramPath->Route.Queue = Datagram->Route->Queue;
```
maolson-msft, line 5456: Seems weird that elsewhere we use a helper function CxPlatDpRawAssignQueue (with different XDP and DPDK implementations) to abstract the assignment of the queue, but then here we just dig in and set it directly. Does this code break dpdk?

maolson-msft: Also, what is going on here anyway? We already assign a queue in CxPlatResolveRoute, but here we are overriding that with the packet's assigned queue, is that right? I am guessing this is done to automatically mirror the RSS configuration or something like that, but then my question is why do we assign a queue in CxPlatResolveRoute?

Also, I don't see anything in the PR description about this change.

csujedihy: DPDK has been defunct (can't compile and won't work) for a long time.

src/inc/quic_datapath.h
maolson-msft: Yes, but please first answer my question in the other thread because it looks to me like the queue would normally already be reassigned in CxPlatResolveRoute at this point.

src/core/connection.c
maolson-msft: I think my second question is the more important one here- sorry for putting it into a thread instead of making a new comment.

csujedihy: >> We already assign a queue in CxPlatResolveRoute, but here we are overriding that with the packet's assigned queue, is that right? I am guessing this is done to automatically mirror the RSS configuration or something like that, but then my question is why do we assign a queue in CxPlatResolveRoute?

The workflow here is we resolve the route and assign a default queue. When we receive the first packet after route resolution, we override the queue to match RSS. If we don't give it a default queue in CxPlatResolveRoute, sends won't go through.

csujedihy:
@nibanks It looks like we have got good results by spreading TX on the queue used by RX but using all procs seems to have made the perf machine unusable during the testing.

```
Running Test RPS_WinXDP_x64_schannel_ResponseSize_0
Test Run Took 00:00:34.4965282
Run 1: 1460253 RPS, Minimum: 42, Maximum: 836461, Percentiles: 50th: 1797.000000, 90th: 4925.000000, 99th: 37952.000000, 99.9th: 204809.000000, 99.99th: 390607.000000, 99.999th: 717418.000000, 99.9999th: 833689.000000, Standard Error: 3.666993
Test Run Took 12:56:19.6331376
WARNING: The network connection to quic-server has been interrupted. Attempting to reconnect for up to 4 minutes...
WARNING: Attempting to reconnect to quic-server ...

```

nibanks:
> @nibanks It looks like we have got good results by spreading TX on the queue used by RX but using all procs seems to have made the perf machine unusable during the testing.
> 
> ```
> Running Test RPS_WinXDP_x64_schannel_ResponseSize_0
> Test Run Took 00:00:34.4965282
> Run 1: 1460253 RPS, Minimum: 42, Maximum: 836461, Percentiles: 50th: 1797.000000, 90th: 4925.000000, 99th: 37952.000000, 99.9th: 204809.000000, 99.99th: 390607.000000, 99.999th: 717418.000000, 99.9999th: 833689.000000, Standard Error: 3.666993
> Test Run Took 12:56:19.6331376
> WARNING: The network connection to quic-server has been interrupted. Attempting to reconnect for up to 4 minutes...
> WARNING: Attempting to reconnect to quic-server ...
> 
> ```

We probably need to leave one core (0?) free.

src/core/connection.c
maolson-msft: Got it, that makes sense and I think it justifies the addition of the QueueUpdated flag to CXPLAT_ROUTE. Please add a code comment to the QueueUpdated flag to explain this behavior of choosing a "default" queue and then updating the queue based on the received packet.

But hey wait, why don't we just always compare the queue of inbound traffic to the current queue and update to match it if it's different? That way we would stay up to date with an RSS configuration change.

ThadHouse:
You’ll definitely need to leave a core free. Maybe 2. The machine still needs to be able to talk normally both to azure pipelines and the other system in the pair.

src/core/connection.c
csujedihy: That seems better. I will update the PR.

csujedihy:
> You’ll definitely need to leave a core free. Maybe 2. The machine still needs to be able to talk normally both to azure pipelines and the other system in the pair.

Makes sense. Will start with reserving proc 0.

nibanks (approved):

src/core/connection.c
nibanks: LGTM

PR #2680, opened 2022-05-03
Update listener docs for quirks regarding ListenerStart on the same port

### LINKED ISSUE #2417: Document how port sharing behaves

In TCP, when two servers start to listen on the same port, the later gets and error. In msquic, both servers will succeed and only the first one will get all the traffic. This might be hard to diagnose for Kestrel users. We should document this behavior and point runtime and ASP.NET documentation to msquic docs about this topic.

Original discussion: https://github.com/dotnet/runtime/issues/59382

### Proposed solution

To have a documentation to describe this behavior.

### Additional context

_No response_

thhous-msft:
Closes #2417

## Documentation

Its a documentation change

### DIFF
--- a/docs/api/ListenerStart.md
+++ b/docs/api/ListenerStart.md
@@ -54,7 +54,7 @@ The server application may set any combination of local address and/or port numb
 
 MsQuic listens on dual-mode wildcard sockets for each unique port number, and performs address filtering, if necessary, within the QUIC layer.
 
-Due to the use of per processor sockets for performance reasons, 2 distinct processes listening on the same port will not result in the 2nd instance failing to start. The behavior in this case is undefined, and different per platform, but will result in each app not getting the receives it expects. There is potential for a future workaround for this on Windows, but no currently known solution on Posix-based platforms.
+Due to the use of per processor sockets for performance reasons, 2 distinct processes listening on the same port will not result in the 2nd instance failing to start. The behavior in this case is undefined, and different per platform, but will result in each app not getting the receives it expects. There is potential for a future workaround for this on Windows, but no currently known solution on Posix-based platforms. This quirk does not apply if a process using UDP without MsQuic is already bound to the port, as long as that process is not using per processor sockets.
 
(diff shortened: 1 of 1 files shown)


docs/api/ListenerStart.md
```diff
@@ -52,7 +52,11 @@ ALPNs specified in `AlpnBuffers` must be less than 255 bytes in length.
...
 
-MsQuic listens on dual-mode wildcard sockets for each unique port number, and performs address filtering, if necessary, within the QUIC layer. If another application is already listening on the same UDP port as an MsQuic application, despite being a different address family, the MsQuic application will fail to use that port, and `ListenerStart` will fail.
+MsQuic listens on dual-mode wildcard sockets for each unique port number, and performs address filtering, if necessary, within the QUIC layer.
+
+Due to the use of per processor sockets for performance reasons, 2 instances of MsQuic listening on the same port will not result in the 2nd instance failing to start. The behavior in this case is undefined, and different per platform, but will result in each app not getting the receives it expects. There is potential for a future workaround for this on Windows, but no currently known solution on Posix based platforms.
```
nibanks, line 57: `2 instances` -> `2 processes`

nibanks, line 57: `Posix based` -> `Posix-based`

docs/api/ListenerStart.md
```diff
@@ -52,7 +52,11 @@ ALPNs specified in `AlpnBuffers` must be less than 255 bytes in length.
...
+
+On posix based platforms, 2 apps using wildcard port numbers can potentially receive the same port number, resulting in the above behavior. This behavior does not exist on Windows.
```
nibanks, line 59: `posix based` -> `Posix-based`

docs/api/ListenerStart.md
nibanks: The problem that in a single process, there is no problem. Maybe "isolated instances" is a little better, but for two services in the same service host, they could be considered isolated, but still there would be no problem.

thhous-msft: Ok updated it a bit.

PR #2685, opened 2022-05-06
Ensure Platform globals are freed on error

### LINKED ISSUE #2683: SpinQuic Crash due to uninitialized memory
### Describe the bug

`CxPlatNumaMasks` is not zero-initialized, which will cause debug builds to attempt to deref the tag on free and access uninitialized memory. 
Fix: zero-initialize all platform-specific structures. 

### Affected OS

- [ ] All
- [ ] Windows Server 2022
- [ ] Windows 11
- [X] Windows Insider Preview (specify affected build below)
- [ ] Ubuntu
- [ ] Debian
- [ ] Other (specify below)

### Additional OS information

_No response_

### MsQuic version

main

Run SpinQuic.
Get lucky.

### Expected behavior

No crash.

### Actual outcome

SpinQuic crash.

### Additional details

_No response_

anrossi:
## Description

Free platform globals in error path. Fixes #2683

## Testing

SpinQuic is how this was found, and the fix was validated against the SpinQuic scenario which found it.

## Documentation

N/A.


nibanks:
By default, AFAIU, the C language always auto initializes globals with zero. I don't think PR is doing anything,

anrossi:
I'm not so sure of that; in the test failure that inspired me to make this fix, these globals were just addresses and non-NULL, which caused the test to crash when trying to free them. Perhaps it was from a previous run, but the globals should be set to NULL at the end of the run, so I don't think it is that.

nibanks:
How about try writing a test that validates the initial conditions between resets?

anrossi:
I'm first adding some debug asserts to ensure we catch when these are not-NULL. I'll brainstorm some more on how to write a test for this. Seems spinquic might be exercising this scenario quite a bit, and is likely the best place to catch it.

nibanks (approved):

nibanks:
@anrossi there are a couple spinquic failures. Can you double check to see if they are hitting your new asserts?

PR #2694, opened 2022-05-09
Fix PMTU packets logging as a 0 length send in kernel mode

thhous-msft:
Packets without segmentation enabled were always resulting in SendData->TotalSize being 0. This didn't affect the sends, but make log debugging harder


nibanks (approved):
Candidate for backport?

thhous-msft:
It's not going to meet the bar IMO, since it only affects logging. We can, and just have it for the next time we ingest a backport.

nibanks:
> It's not going to meet the bar IMO, since it only affects logging. We can, and just have it for the next time we ingest a backport.

Yeah, that's what I meant. If/when we fix the leak and backport it, include this change for diagnostic purposes to verify (or something along those lines).

PR #1585, opened 2021-05-13
Disable pool allocator when memory allocation failures are enabled.

anrossi:
- Automatically disabled pool allocs when memory alloc failures are enabled.
- Fixes a few bugs in the code related to memory alloc failures.

PR #2695, opened 2022-05-09
Fix Builder->TotalDatagramsLength being incorrectly calculated

thhous-msft:
It was being incremented by a variable that had just been set to 0, which means it always stayed 0 itself


nibanks (approved):
Was this a regression?

thhous-msft:
Yes, it was, as part of https://github.com/microsoft/msquic/pull/1585

That number is exclusive to performance counters, it has no other purpose.

Issue #2573, opened 2022-03-24
Advanced client certificate support


Techwolfy:
### Describe the feature you'd like supported

MsQuic supports client certificates, but provides few options to configure validation. To bring Windows HTTP/3 client certificate support up to parity with HTTP/1.1, we will need the following features:

1. Support for anonymous client authentication on server-side
2. Support for configuring an arbitrary SecPkgCred_ClientCertPolicy
3. Support for disabling SCH_CRED_NO_SYSTEM_MAPPER
4. Support for querying the security context token post-handshake

### Proposed solution

1. SECPKG_ATTR_CERT_CHECK_RESULT_INPROC will return SEC_E_NO_CREDENTIALS, which causes MsQuic to fail the handshake with CXPLAT_TLS_RESULT_ERROR; we need this to go through the PEER_CERTIFICATE_RECEIVED callback instead. This can be resolved by updating the error check at `/src/platform/tls_schannel.c:2127` to store and ignore the error:

```
if (SecStatus == SEC_E_NO_CREDENTIALS) {
    CertValidationResult.hrVerifyChainStatus = SecStatus;
} else if (SecStatus != SEC_E_OK) {
```

The call to SECPKG_ATTR_REMOTE_CERTIFICATES may fail with SEC_E_NO_CREDENTIALS in this case as well. That error can be ignored.

2. Currently SECPKG_ATTR_CLIENT_CERT_POLICY uses a locally-defined struct supporting configuration of only 3 flags. HTTP.SYS uses every field in that structure, so it's likely more efficient to enable passing a pointer in via SetParam or the QUIC_CREDENTIAL_CONFIG than to add support for each field individually.

3. A new QUIC_CREDENTIAL_FLAG_USE_SYSTEM_MAPPER to control the behavior at `/src/platform/tls_schannel.c:1053` would be sufficient. This is dependent on the prior feature, as use of the system mapper depends on several currently-unsupported fields in SecPkgCred_ClientCertPolicy.

4. A new QUIC_PARAM_TLS_SCHANNEL_CONTEXT_TOKEN GetParam that calls QuerySecurityContextToken on the SChannel context handle would be enough here.

nibanks:
Related to https://github.com/microsoft/msquic/issues/1953.

macoxms:
Per our conversation please also switch from using SECPKG_ATTR_REMOTE_CERTIFICATES to the in-proc version given that the call is always done when the in-proc version is available.

[...]

Issue #1953, opened 2021-08-31
Allow server application to decide what to do when no client certificate is presented


anrossi:
Schannel and OpenSSL client certificate support will fail the handshake if the server requires a client certificate be presented during the handshake, and no client certificate is presented.
In some scenarios, the server would prefer to know this and decide what to do instead of the TLS layer killing the handshake.

anrossi:
A workaround exists, however: If `QUIC_CREDENTIAL_FLAG_REQUIRE_CLIENT_AUTHENTICATION` is not set on the server configuration, the server application can set `QUIC_CREDENTIAL_FLAG_INDICATE_CERTIFICATE_RECEIVED` and verify that it receives a client certificate during the handshake, and if not, it can decide what to do from there.

wfurt:
Related to https://github.com/dotnet/runtime/issues/57308
The server will _not_ ask for certificate without `QUIC_CREDENTIAL_FLAG_REQUIRE_CLIENT_AUTHENTICATION` right @anrossi? With that I think client will never send it.

anrossi:
@wfurt My understanding is not complete here; won't the client send a client certificate if it is configured with one, even if the server doesn't ask for it? Or does client wait to send the client certificate until requested by the server?

wfurt:
That depends - and it should be visible in packet capture. My experience is that it won't until it gets 
[Certificate Request](https://datatracker.ietf.org/doc/html/rfc8446#section-4.3.2).

nibanks:
Let's add a new flag that says "Indicate NULL cert if not present".

wfurt:
Here are some more references: 
Tls 1.2
https://datatracker.ietf.org/doc/html/rfc5246#section-7.4.6
```
      This is the first message the client can send after receiving a
      ServerHelloDone message.  This message is only sent if the server
      requests a certificate.  If no suitable certificate is available,
      the client MUST send a certificate message containing no
      certificates.  That is, the certificate_list structure has a
      length of zero.  If the client does not send any certificates, the
      server MAY at its discretion either continue the handshake without
      client authentication, or respond with a fatal handshake_failure
      alert.
```
This is the general behavior we looking at. I could not find quite same text in Tls 1.3 e.g. it is not clear to me if client _can_ send certificate without server asking for it. 

https://datatracker.ietf.org/doc/html/rfc8446#section-4.4.2.4
```
   If the server supplies an empty Certificate message, the client MUST
[...]

PR #2702, opened 2022-05-11
Indicate null/missing client certificate

anrossi:
## Description

When `QUIC_CREDENTIAL_FLAG_NO_CERTIFICATE_VALIDATION` or `QUIC_CREDENTIAL_FLAG_DEFER_VALIDATION` are present, and `QUIC_CREDENTIAL_FLAG_INDICATE_PEER_CERTIFICATE_RECEIVED` is specified, indicate a null certificate message to the app when no certificate was received from the peer.

## Testing

A new test was added to test this case when the client doesn't send a certificate. 

## Documentation
No documentation was added/updated.

Related items: #2573 #1953


src/inc/msquic.h
```diff
@@ -110,6 +110,7 @@ typedef enum QUIC_CREDENTIAL_FLAGS {
     QUIC_CREDENTIAL_FLAG_SET_ALLOWED_CIPHER_SUITES              = 0x00002000,
     QUIC_CREDENTIAL_FLAG_USE_PORTABLE_CERTIFICATES              = 0x00004000,
     QUIC_CREDENTIAL_FLAG_USE_SUPPLIED_CREDENTIALS               = 0x00008000, // Schannel only currently
+    QUIC_CREDENTIAL_FLAG_INDICATE_NULL_CLIENT_CERTIFICATE       = 0x00010000,
```
nibanks, line 113:
```suggestion
    QUIC_CREDENTIAL_FLAG_INDICATE_NULL_CERTIFICATE                    = 0x00010000,
```
I think we don't need to include CLIENT in the name. BTW, we need to add validation logic to fail with invalid parameter if this is passed on client.

src/platform/tls_openssl.c
```diff
@@ -102,6 +102,11 @@ typedef struct CXPLAT_TLS {
     //
     BOOLEAN IsServer;
 
+    //
+    // Indicates if the peer sent a certificate.
+    //
+    BOOLEAN PeerCertReceived;
```
nibanks, line 108: Use a bit field for this and `IsServer` above.

nibanks (commented):
Please add both positive and negative tests, for the new flag being present and missing.

src/platform/tls_openssl.c
```diff
@@ -1964,6 +1972,16 @@ CxPlatTlsProcessData(
...
+        } else if ((TlsContext->SecConfig->Flags & QUIC_CREDENTIAL_FLAG_INDICATE_NULL_CLIENT_CERTIFICATE) &&
+            !TlsContext->PeerCertReceived) {
+            if (!TlsContext->SecConfig->Callbacks.CertificateReceived(
+                TlsContext->Connection,
+                NULL,
+                NULL,
+                0,
+                0)) {
+                // TODO kill the handshake
+            }
```
nibanks, line 1984:
```suggestion
            if (!TlsContext->SecConfig->Callbacks.CertificateReceived(
                  TlsContext->Connection,
                  NULL,
                  NULL,
                  0,
                  0)) {
                TlsContext->ResultFlags |= CXPLAT_TLS_RESULT_ERROR;
                goto Exit;
            }
```
I think this is what you want.

src/inc/msquic.h
anrossi: After discussion, removing this flag and making the existing flags work.

src/platform/tls_openssl.c
```diff
@@ -1964,6 +1974,29 @@ CxPlatTlsProcessData(
...
+                TlsContext->SecConfig->Flags & QUIC_CREDENTIAL_FLAG_DEFER_CERTIFICATE_VALIDATION)) ?
+                    QUIC_STATUS_CERT_NO_CERT :
+                    QUIC_STATUS_SUCCESS;
+
+            if (!TlsContext->SecConfig->Callbacks.CertificateReceived(
+                TlsContext->Connection,
+                NULL,
+                NULL,
+                0,
+                ValidationResult)) {
```
nibanks, line 1991:
```suggestion
            if (!TlsContext->SecConfig->Callbacks.CertificateReceived(
                    TlsContext->Connection,
                    NULL,
                    NULL,
                    0,
                    ValidationResult)) {
```

src/platform/unittest/TlsTest.cpp
```diff
@@ -1777,6 +1777,26 @@ TEST_F(TlsTest, ClientCertificateDeferValidation)
     DoHandshake(ServerContext, ClientContext);
 }
 
+TEST_F(TlsTest, ClientCertificateDeferValidationNoCert)
+{
+    CxPlatClientSecConfig ClientConfig(
+        QUIC_CREDENTIAL_FLAG_NO_CERTIFICATE_VALIDATION
+#ifdef QUIC_ENABLE_ANON_CLIENT_AUTH_TESTS
+        | QUIC_CREDENTIAL_FLAG_USE_SUPPLIED_CREDENTIALS
+#endif
```
nibanks, line 1786: Why modify the behavior of a test with the `QUIC_ENABLE_ANON_CLIENT_AUTH_TESTS` and not have two separate tests?

CMakeLists.txt
```diff
@@ -338,6 +338,8 @@ if(QUIC_TLS STREQUAL "schannel")
     list(APPEND QUIC_COMMON_DEFINES QUIC_DISABLE_0RTT_TESTS)
     message(STATUS "Disabling ChaCha20 support")
     list(APPEND QUIC_COMMON_DEFINES QUIC_DISABLE_CHACHA20_TESTS)
+    message(STATUS "Enabling anonymous client auth tests")
+    list(APPEND QUIC_COMMON_DEFINES QUIC_ENABLE_ANON_CLIENT_AUTH_TESTS)
```
nibanks, line 342: Why can't se support this behavior with OpenSSL?

src/platform/unittest/TlsTest.cpp
anrossi: USE_SUPPLIED_CREDS is necessary for schannel, and is an schannel-only flag. I want this test to also cover OpenSSL.

CMakeLists.txt
anrossi: We already do. USE_SUPPLIED_CREDS is a flag that is schannel-only. OpenSSL works just fine out of the box.

PR #2717, opened 2022-05-15
Refactor XDP RX/TX Slightly

nibanks:
## Description

Try to refactor the RX and TX functions of XDP to optimize them slightly.

## Testing

Perf CI testing

## Documentation

N/A


src/platform/datapath_raw_xdp.c
```diff
@@ -1508,49 +1545,64 @@ CxPlatXdpTx(
...
-        CxPlatLockRelease(&Queue->TxLock);
-    }
 
+    //
+    // Drain anything that XDP has filled in the TX completion ring. Return the
+    // TX buffers back to the pool, and release the space back to XDP.
+    //
     uint32_t CompIndex;
-    uint32_t CompAvailable =
-        XskRingConsumerReserve(&Queue->TxCompletionRing, MAXUINT32, &CompIndex);
```
mtfriesen, line 1523: You might not want to remove the MAXUINT32 parameter but rather set it to a more precise upper bound. One benefit of a well-sized `maxCount` parameter is the helpers can (and Linux helpers _do_) locally cache the values of shared memory. This avoids the need for the consumer to acquire the shared memory cache line if the local cached value is already sufficient.

nibanks: Interesting. This PR didn't seem to make any measured change, so it will likely be abandoned. But I'm still not sure the code is currently organized in the most efficient way it could be....

PR #2723, opened 2022-05-17
Add helper service for performance testing

thhous-msft:
This service opens a handle to the certificates, which makes it so the server perf system only has to be logged into to bypass the certificate issue, rather then running a special command. This also likely makes it so another user logging in does not break performance


nibanks (approved):

thhous-msft:
Nevermind, it doesn't actually seem to work. Sad.

PR #2728, opened 2022-05-18
Add sendbuf param to RPS client

CarnaViire:
~~Add scenarios to RPS client that align with how System.Net.Quic uses MsQuic, to be able to compare performance fairly:~~
~~- Turn on/off send buffering~~
~~- Sending data after receiving stream started event~~

Add param to RPS client to turn on/off send buffering

### DIFF
--- a/src/perf/lib/RpsClient.cpp
+++ b/src/perf/lib/RpsClient.cpp
@@ -37,7 +37,6 @@ PrintHelp(
         "  -threads:<####>             The number of threads to use. Defaults and capped to number of cores\n"
         "  -affinitize:<0/1>           Affinitizes threads to a core. (def:0)\n"
         "  -sendbuf:<0/1>              Whether to use send buffering. (def:0)\n"
-        "  -waitforstart:<0/1>         Whether to wait for start complete event before sending data. (def:0)\n"
         "\n",
         RPS_DEFAULT_RUN_TIME,
         PERF_DEFAULT_PORT,
@@ -117,11 +116,6 @@ RpsClient::Init(
         Configuration.SetSettings(settings);
     }
 
-    uint32_t WaitForStart;
-    if (TryGetValue(argc, argv, "waitforstart", &WaitForStart)) {
-        WaitForStartBeforeSend = WaitForStart != 0;
-    }
-
     WorkerCount = CxPlatProcActiveCount();
     if (WorkerCount > PERF_MAX_THREAD_COUNT) {
         WorkerCount = PERF_MAX_THREAD_COUNT;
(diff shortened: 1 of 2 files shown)


src/perf/lib/RpsClient.h
```diff
@@ -22,14 +22,18 @@ struct RpsWorkerContext;
...
         _In_ RpsConnectionContext* Connection,
         _In_ uint64_t StartTime)
         : Connection{Connection}, StartTime{StartTime} { }
+    ~StreamContext() noexcept { if (Handle) { MsQuic->StreamClose(Handle); } }
     RpsConnectionContext* Connection;
     uint64_t StartTime;
+    void ExecuteStreamSend();
 #if DEBUG
-    uint8_t Padding[12];
+    uint8_t Padding[12]; // TODO: Should this change?
```
CarnaViire, line 36: @nibanks I need your guidance here 🙂

src/perf/lib/RpsClient.cpp
```diff
@@ -36,6 +36,8 @@ PrintHelp(
         "  -response:<####>            The length of request payloads. (def:%u)\n"
         "  -threads:<####>             The number of threads to use. Defaults and capped to number of cores\n"
         "  -affinitize:<0/1>           Affinitizes threads to a core. (def:0)\n"
+        "  -sendbuf:<0/1>              Whether to use send buffering. (def:0)\n"
+        "  -waitforstart:<0/1>         Whether to wait for start complete event before sending data. (def:0)\n"
```
nibanks, line 40: `sendbuf` is fine, but I thought you were just going to separate out `StreamStart` from `StreamSend`. I don't believe you should ever wait for `StreamStart` to complete. But even if we do add support for it, you shouldn't need to change so much. Just call StreamSend in start complete if `waitforstart` is set.

CarnaViire: > Just call StreamSend in start complete

But that would execute on msquic thread... And when called from System.Net.Quic it would be on user thread

nibanks: I guess, bottom line, I don't think a performance test should do that. So while .NET is currently doing that, I would argue that it should do everything it can not to do that. I don't want to encourage that behavior because it's inherently not performant.

src/perf/lib/RpsClient.h
```diff
@@ -22,14 +22,16 @@ struct RpsWorkerContext;
 class RpsClient;
 
 struct StreamContext {
+    HQUIC Handle {nullptr};
     StreamContext(
         _In_ RpsConnectionContext* Connection,
         _In_ uint64_t StartTime)
         : Connection{Connection}, StartTime{StartTime} { }
+    ~StreamContext() noexcept { if (Handle) { MsQuic->StreamClose(Handle); } }
```
CarnaViire, line 30: I can revert this too if you think it is not useful now @nibanks

src/perf/lib/RpsClient.cpp
```diff
@@ -447,7 +456,6 @@ RpsConnectionContext::StreamCallback(
         break;
     case QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE:
         Worker->Client->StreamContextAllocator.Free(StrmContext);
-        MsQuic->StreamClose(StreamHandle);
```
nibanks, line 450: Can you undo all those other changes too?

src/perf/lib/RpsClient.h
nibanks: Yeah, let's just remove everything not absolutely necessary

nibanks (approved):

PR #2729, opened 2022-05-19
Support setting SECPKG_ATTR_CLIENT_CERT_POLICY directly

anrossi:
## Description

Support callers setting Schannel credentials attributes on Configuration objects.

## Testing

SpinQuic testing on Windows.

## Documentation

No documentation changes.

Related issue: #2573


src/cs/lib/msquic_generated.cs
```diff
@@ -2580,6 +2591,9 @@ internal static unsafe partial class MsQuic
         [NativeTypeName("#define QUIC_PARAM_CONFIGURATION_VERSION_SETTINGS 0x03000002")]
         internal const uint QUIC_PARAM_CONFIGURATION_VERSION_SETTINGS = 0x03000002;
 
+        [NativeTypeName("#define QUIC_PARAM_CONFIGURATION_SCHANNEL_CREDENTIAL_ATTRIBUTE_W 0x03000003")]
+        internal const uint QUIC_PARAM_CONFIGURATION_SCHANNEL_CREDENTIAL_ATTRIBUTE_W = 0x03000003;
```
nibanks, line 2595: I just realized, both the Rust header and documentation needs to be updated too. For the recent changes as well.

src/inc/msquic.h
```diff
@@ -720,6 +720,14 @@ void
 #ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
 #define QUIC_PARAM_CONFIGURATION_VERSION_SETTINGS       0x03000002  // QUIC_VERSION_SETTINGS
 #endif
+#ifdef WIN32 // Schannel-specific Configuration parameters
```
nibanks, line 723: Since this is **part** of the configuration space, I think we should not wrap it in `#ifdef WIN32` but instead purely comment the dependency. This means spinquic can "try" it out on other platforms too, to verify it doesn't crash.

src/cs/lib/msquic_generated.cs
anrossi: Is that a manual or automated process?

anrossi: I'll do all three of them in a separate PR

nibanks: Ok.

nibanks (approved):

Techwolfy (approved):

nibanks (approved):

PR #2730, opened 2022-05-19
Support Schannel users calling QuerySecurityContextToken

anrossi:
## Description

Support Schannel users calling QuerySecurityContextToken.

## Testing

SpinQuic testing.

## Documentation

N/A

Related issue: #2573


src/platform/tls_schannel.c
```diff
@@ -2786,6 +2786,23 @@ CxPlatTlsParamGet(
...
+
+            if (Buffer == NULL) {
+                Status = QUIC_STATUS_INVALID_PARAMETER;
+            }
+
+            Status =
+                SecStatusToQuicStatus(
+                QuerySecurityContextToken(
+                    &TlsContext->SchannelContext,
+                    &Buffer));
```
Techwolfy, line 2803: `&Buffer` here should just be `Buffer`, no dereference. QuerySecurityContextToken returns a HANDLE so the current behavior doesn't crash, but the token is leaked.

Techwolfy (changes requested):

nibanks (approved):

src/platform/tls_schannel.c
anrossi: Fixed!

PR #2735, opened 2022-05-20
Allow Apps to Never Pause Stream Receive Callbacks

nibanks:
## Description

WIP change to possible allow multiple receive callbacks even if one is pending.

### DIFF
--- a/src/core/stream.c
+++ b/src/core/stream.c
@@ -606,6 +606,19 @@ QuicStreamParamSet(
         break;
     }
 
+    case QUIC_PARAM_STREAM_ALWAYS_RECEIVE: {
+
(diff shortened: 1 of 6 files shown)


thhous-msft:
I'm curious to understand what this means? Does this mean receives could come in out of order to the app?

ManickaP:
This should allow receiving additional data as they come on the wire while still having pending/outstanding read. It's an experiment to see if this could help .NET perf in any capacity. The sentiment about this is rather skeptical but since it was easy for Nick to cook up, so he made this PR for .NET to experiment with.

nibanks:
Per @ManickaP on IM, this won't be necessary, so I am closing the PR.

PR #2741, opened 2022-05-21
Add polyfill for .NET Standard support

ThadHouse:
.NET does not need to include this, but external users that want to support .NET Standard will find it useful.

.NET Standard is a .NET Target that supports old .NET Framework, UWP, and many alternate platforms as well. Fully supporting .NET Standard for all builds would result in performance compromises in mainline .NET itself, which we don't want. Instead, we use a few tricks to polyfill the necessary functions. So running on .NET standard things will use some older functionality, but everything works fine.

## Testing

Build will ensure things will still work

## Documentation

N/A

### DIFF
--- a/src/cs/LicenseHeader.txt
+++ b/src/cs/LicenseHeader.txt
@@ -6,3 +6,10 @@
 #pragma warning restore IDE0073
 
 #pragma warning disable CS0649
+
+// Polyfill for MemoryMarshall on .NET Standard
+#if NETSTANDARD && !NETSTANDARD2_1_OR_GREATER
+using MemoryMarshal = Microsoft.Quic.Polyfill.MemoryMarshal;
+#else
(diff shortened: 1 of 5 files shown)


ThadHouse:
@ManickaP Is this going to mess with anything on your end? You don't have to pull in the polyfill file (And I actually wouldn't want you to), but will the new ifdef at the top of msquic_generated.cs and msquic.cs mess with anything in the .NET repo?

nibanks (approved):

thhous-msft:
@ManickaP Just pinging again. This shouldn't affect you guys, I just want to make sure it doesn't before we merge it.

nibanks (approved):

ManickaP:
I completely missed this, sorry. It doesn't affect us.

PR #2748, opened 2022-05-23
Use port set XDP rules

csujedihy:
## Description

Use the new port set XDP rules for client-side connections.

## Testing

CI

### DIFF
--- a/src/platform/datapath_raw_xdp.c
+++ b/src/platform/datapath_raw_xdp.c
@@ -30,6 +30,7 @@ Abstract:
 #define RULE_TAG      'UpdX' // XdpU
 #define RX_BUFFER_TAG 'RpdX' // XdpR
 #define TX_BUFFER_TAG 'TpdX' // XdpT
+#define PORT_SET_TAG  'PpdX' // XdpP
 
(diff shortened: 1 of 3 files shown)


src/platform/datapath_raw_xdp.c
```diff
@@ -1138,6 +1144,26 @@ CxPlatDpRawUninitialize(
...
 }
 
+_IRQL_requires_max_(PASSIVE_LEVEL)
+VOID
+CxPlatDpRawSetPortBit(
+    _Inout_ uint8_t *BitMap,
+    _In_ uint16_t Port
+    )
+{
+    BitMap[Port >> 3] |= (1 << (Port & 0x7));
```
nibanks, line 1154: Should these be interlocked functions?

csujedihy: We are doing it under rule lock. We need to hold the lock for iterating all rules to find the right rule for the connection anyways.

nibanks: Ok. Let's go with it. If we need anything else later, we can tackle it then.

nibanks (approved):

PR #2752, opened 2022-05-24
Consume XDP from github release and stash XDP version in MsQuic for XDP nuget package

csujedihy:
## Description

- Consume XDP from github release.
- Stash XDP version in MsQuic for XDP nuget package. We need to do this until XDP is officially released (with backward compatibility) so that partner teams can find the right XDP binaries.


nibanks (approved):

csujedihy:
@nibanks need your help to override the CI gate.

PR #2756, opened 2022-05-26
Allow TCP perf to fail to complete without failing pipeline

### LINKED ISSUE #2527: Make TCP perf crashes not fail the perf pipeline
The TCP perf test has some issues. Currently if one of those issues occur, the entire pipeline files. We should make it so a tcp failure is skipped and ignored.

thhous-msft:
## Description

Make TCP connection failures not kill the perf pipeline. The test isn't that important, so skip any failures here

Closes #2527

Do not merge until I remove the tcp change that forces it to always fail.

### DIFF
--- a/src/perf/lib/ThroughputClient.cpp
+++ b/src/perf/lib/ThroughputClient.cpp
@@ -76,10 +76,6 @@ ThroughputClient::Init(
     TryGetValue(argc, argv, "download", &DownloadLength);
     TryGetValue(argc, argv, "stats", &PrintStats);
 
-    if (UseTcp) {
-        Port = 9233;
-    }
-
     if (UploadLength && DownloadLength) {
         WriteOutput("Must specify only one of '-upload' or '-download' argument!\n");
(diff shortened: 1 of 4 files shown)


src/perf/lib/ThroughputClient.cpp
```diff
@@ -614,7 +614,7 @@ ThroughputClient::OnTcpConnectionComplete(
     if (Connection) {
         Connection->Close();
     }
-    CxPlatEventSet(*StopEvent);
+    //CxPlatEventSet(*StopEvent);
```
nibanks, line 617: Why was this commented out?

thhous-msft: > Do not merge until I remove the tcp change that forces it to always fail.

Testing that it actually works.

nibanks: Suggestion: Make it a Draft PR if you don't want it merged.

nibanks (approved):

PR #2762, opened 2022-05-27
Expose Parameter for TLS Provider

### LINKED ISSUE #2761: Allow applications to query which TLS backend MsQuic uses.

Follow-up from discussion at https://github.com/dotnet/runtime/pull/69874#discussion_r883015483. .NET needs to use different mechanisms for passing certificates depending if the underlying MsQuic build uses Schannel or (statically linked?) OpenSSL.

Currently, there is no way to detect which TLS backend MsQuic is using, so we assume that Windows builds always use Schannel. In order to eventually support QUIC in .NET on older Windows versions (via MsQuic with OpenSSL), we need to condition the code based on the actual TLS backend.

### Proposed solution

Add a global GetParam option that returns an enum corresponding to the TLS backend the library uses.

### Additional context

_No response_

nibanks:
## Description

Fixes #2761. Adds a new parameter that can be called via GetParam to query the compiled in TLS provider. This allows application code to handle any TLS provider specific differences.

cc @rzikm

## Testing

Added get/set test.

## Documentation

Updated for new parameter.


thhous-msft (approved):

GF-Huang:
Is this means .NET 7 QUIC classes can work on older Win10 even if Win7? For the QUIC namespaces docs, I can't see any description about this.

rzikm:
> Is this means .NET 7 QUIC classes can work on older Win10 even if Win7? For the QUIC namespaces docs, I can't see any description about this.

We don't have test coverage for running with OpenSSL-based MsQuic on Windows, but last time I tried it it worked (after integrating the change from this PR). It is not officially supported, but we are considering official support in 8.0. If you find any issues, feel free to raise a PR in dotnet/runtime repo.

The only pitfall of OpenSSL on Windows I can remember now is that any certificates you use *must* be exportable, because we need to serialize them to memory in order to pass them to MsQuic.

Issue #1215, opened 2021-02-09
Test GetParam/SetParam for all the MsQuic objects.


anrossi:

nibanks:
Closing this as Daiki's work pretty much covered it all.

PR #2774, opened 2022-06-02
Add more SetParam tests for Global

ami-GS:
## Description

More SetParam test set for global parameter #1215

## Testing

This is test

## Documentation

N/A

### DIFF
--- a/src/test/lib/ApiTest.cpp
+++ b/src/test/lib/ApiTest.cpp
@@ -1897,6 +1897,7 @@ void QuicTestStatefulGlobalSetParam()
     //
     {
         TestScopeLogger LogScope("Set QUIC_PARAM_GLOBAL_LOAD_BALACING_MODE after connection start (MsQuicLib.InUse)");
+        GlobalSettingScope ParamScope(QUIC_PARAM_GLOBAL_LOAD_BALACING_MODE);
         MsQuicAlpn Alpn("MsQuicTest");
         MsQuicCredentialConfig ClientCredConfig;
         MsQuicConfiguration ClientConfiguration(Registration, Alpn, ClientCertCredConfig);
@@ -1912,7 +1913,7 @@ void QuicTestStatefulGlobalSetParam()
                 4433));
         CxPlatSleep(1000); // bit slow to set MsQuicLib.InUse = TRUE
 
-        uint16_t Mode = QUIC_LOAD_BALANCING_DISABLED;
+        uint16_t Mode = QUIC_LOAD_BALANCING_SERVER_ID_IP;
         TEST_QUIC_STATUS(
             QUIC_STATUS_INVALID_STATE,
             MsQuic->SetParam(
@@ -1927,6 +1928,7 @@ void QuicTestStatefulGlobalSetParam()
     //
     {
         TestScopeLogger LogScope("Set QUIC_PARAM_GLOBAL_DATAPATH_PROCESSORS when MsQuicLib.Datapath != NULL");
+        GlobalSettingScope ParamScope(QUIC_PARAM_GLOBAL_DATAPATH_PROCESSORS);
         uint16_t Data[4] = {};
         TEST_QUIC_STATUS(
             QUIC_STATUS_INVALID_STATE,
@@ -1945,6 +1947,7 @@ void QuicTestGlobalSetParam()
     //
     {
         TestScopeLogger LogScope("QUIC_PARAM_GLOBAL_RETRY_MEMORY_PERCENT");
+        GlobalSettingScope ParamScope(QUIC_PARAM_GLOBAL_RETRY_MEMORY_PERCENT);
         uint16_t Percent = 26;
         TEST_QUIC_SUCCEEDED(
             MsQuic->SetParam(
@@ -1973,6 +1976,7 @@ void QuicTestGlobalSetParam()
     //
     {
         TestScopeLogger LogScope("QUIC_PARAM_GLOBAL_LOAD_BALACING_MODE");
+        GlobalSettingScope ParamScope(QUIC_PARAM_GLOBAL_LOAD_BALACING_MODE);
         //
         // Invalid mode
         //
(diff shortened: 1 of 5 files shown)


src/test/lib/ApiTest.cpp
```diff
@@ -1813,10 +2113,151 @@ void QuicTestGlobalSetParam()
...
+                    QUIC_PARAM_GLOBAL_DATAPATH_PROCESSORS,
+                    sizeof(Data) + 1,
+                    &Data));
+        }
+
+        //
+        // one of data is bigger than the number of its' platform cpus
+        //
+        {
+            TestScopeLogger LogScope("one of data is bigger than the number of its' platform cpus");
```
nibanks, line 2145:
```suggestion
        // one of data is bigger than the number of its platform cpus
        //
        {
            TestScopeLogger LogScope("one of data is bigger than the number of its platform cpus");
```

src/test/lib/ApiTest.cpp
```diff
@@ -1813,10 +2113,151 @@ void QuicTestGlobalSetParam()
...
+            TEST_QUIC_SUCCEEDED(
+                MsQuic->SetParam(
+                    nullptr,
+                    QUIC_PARAM_GLOBAL_DATAPATH_PROCESSORS,
+                    sizeof(Data),
+                    &Data));
+        }
+    }
+
+#if QUIC_TEST_DATAPATH_HOOKS_ENABLED
```
nibanks, line 2185: I don't think you should add tests for test hooks. They already get used.

src/test/lib/ApiTest.cpp
```diff
@@ -1753,8 +1753,207 @@ class SecConfigTestContext {
...
+
 void QuicTestGlobalSetParam()
 {
+    //
+    // QUIC_PARAM_GLOBAL_RETRY_MEMORY_PERCENT
+    //
+    {
+        TestScopeLogger LogScope("QUIC_PARAM_GLOBAL_RETRY_MEMORY_PERCENT");
+        uint16_t Percent = 26;
+        TEST_QUIC_SUCCEEDED(
```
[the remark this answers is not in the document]
ami-GS: Is there any function to revert all global setting? or need to do manually?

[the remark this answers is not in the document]
nibanks: We don't really have a way to clear already set parameters, but I believe `ClearGlobalVersionListScope` does something similar to what you might want to do here. You just need to make a version that caches all settings, and then sets them back.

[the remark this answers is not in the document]
ami-GS: Thanks! I will check the function.

src/test/lib/ApiTest.cpp
ami-GS: Fixed, thanks!

src/test/lib/ApiTest.cpp
```diff
@@ -1753,8 +1753,210 @@ class SecConfigTestContext {
     }
 };
 
+void SettingApplyTests(HQUIC Handle, uint32_t Param) {
+    {
+        QUIC_SETTINGS Settings{0};
+        Settings.MaxAckDelayMs = UINT32_MAX;
```
nibanks, line 1759: Along with all these negative `UINT32_MAX` size tests, it's probably a good idea to add tests that verify that `QUIC_VAR_INT_MAX` does work.

src/test/lib/ApiTest.cpp
```diff
@@ -1753,8 +1753,210 @@ class SecConfigTestContext {
...
+        MsQuicConnection Connection(Registration);
+        TEST_QUIC_SUCCEEDED(Connection.GetInitStatus());
+        TEST_QUIC_SUCCEEDED(
+            MsQuic->ConnectionStart(
+                Connection.Handle,
+                ClientConfiguration,
+                QUIC_ADDRESS_FAMILY_INET,
+                "localhost",
+                4433));
+        CxPlatSleep(1000); // bit slow to set MsQuicLib.InUse = TRUE
```
nibanks, line 1914: I don't think this is required. Please verify and if it's not, please remove.

src/test/lib/ApiTest.cpp
```diff
@@ -1753,8 +1753,210 @@ class SecConfigTestContext {
...
+    TestScopeLogger LogScope("QuicTestStatefulGlobalSetParam");
+    MsQuicRegistration Registration(true);
+    TEST_TRUE(Registration.IsValid());
+
+    //
+    // Set QUIC_PARAM_GLOBAL_LOAD_BALACING_MODE after connection start (MsQuicLib.InUse)
+    //
+    {
+        TestScopeLogger LogScope("Set QUIC_PARAM_GLOBAL_LOAD_BALACING_MODE after connection start (MsQuicLib.InUse)");
+        GlobalSettingScope ParamScope(QUIC_PARAM_GLOBAL_LOAD_BALACING_MODE);
```
nibanks, line 1900: You should only need this helper if you plan to successfully set the parameter, right?

src/test/lib/TestHelpers.h
```diff
@@ -112,6 +112,49 @@ struct ClearGlobalVersionListScope {
...
 
+//
+// Global parameter setting might affect other tests' behavior.
+// This sets back the original value
+//
+struct GlobalSettingScope {
+    uint32_t Parameter;
+    uint32_t BufferLength;
+    void* OriginalValue;
+    GlobalSettingScope(uint32_t Parameter) : Parameter(Parameter), BufferLength(0), OriginalValue(nullptr) {
```
nibanks, line 123: My preference would be to do it this way:
```suggestion
    uint32_t Parameter;
    uint32_t BufferLength {0};
    void* OriginalValue {nullptr};
    GlobalSettingScope(uint32_t Parameter) : Parameter(Parameter) {
```

src/test/lib/ApiTest.cpp
ami-GS: Yes!

src/test/lib/ApiTest.cpp
ami-GS: Got error without this sleep...

nibanks: Try changing it to `100` ms instead then.

ami-GS: 100ms works. done!

nibanks (approved):

PR #2776, opened 2022-06-02
Add more SetParam tests for Stream

ami-GS:
## Description

SetParam test set for Stream object #1215

## Testing

This is test

## Documentation

N/A

### DIFF
--- a/src/test/lib/ApiTest.cpp
+++ b/src/test/lib/ApiTest.cpp
@@ -2103,7 +2103,6 @@ void QuicTestStreamSetParam()
 {
     MsQuicRegistration Registration;
     TEST_TRUE(Registration.IsValid());
-    TestScopeLogger LogScope("QUIC_PARAM_CONN_QUIC_VERSION is get only");
     MsQuicConnection Connection(Registration);
     TEST_QUIC_SUCCEEDED(Connection.GetInitStatus());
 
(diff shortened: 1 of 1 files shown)


src/test/lib/ApiTest.cpp
```diff
@@ -2151,6 +2153,24 @@ void QuicTestStreamSetParam()
...
     }
+
+    //
+    // QUIC_PARAM_STREAM_PRIORITY
+    //
+    {
+        TestScopeLogger LogScope("QUIC_PARAM_STREAM_PRIORITY");
+        MsQuicStream Stream(Connection, QUIC_STREAM_OPEN_FLAG_NONE);
+        Stream.Start(QUIC_STREAM_START_FLAG_IMMEDIATE); // IMMEDIATE to set Stream->SendFlags != 0
+        CxPlatSleep(100);
```
nibanks, line 2164: Why the sleep? I don't think it should be necessary. For that matter, I'm not sure Start is either.

ami-GS: At least `Stream.Start(QUIC_STREAM_START_FLAG_IMMEDIATE);` is needed to go into line 597.
This sleep is for finishing Start operation which set Flags.Started and SendFlags.
I could see the flags are set without Sleep (with debugger), but the Start operation is asynchronous.
`QuicSendUpdateStreamPriority` doesn't return status code, so I think it passes with success without Start and sleep, but the test coverage become narrow.
https://github.com/microsoft/msquic/blob/62958fe9024e2f17e66fb46a5bce07055a9820c4/src/core/stream.c#L581-L602

nibanks: The start operation is async, but when you call GetParam or SetParam immediately afterwards, the QUIC worker thread will always process the start operation first.

ami-GS: ooh, the operation is FIFO....

nibanks (approved):

PR #2778, opened 2022-06-02
Add more SetParam tests for Listener

ami-GS:
## Description

More SetParam tests for Listener #1215

## Testing

This is test

## Documentation

N/A

### DIFF
--- a/src/test/lib/ApiTest.cpp
+++ b/src/test/lib/ApiTest.cpp
@@ -1853,12 +1853,12 @@ void CibirIDTests(HQUIC Handle, uint32_t Param) {
     // buffer length test
     //
     {
-        TestScopeLogger LogScope("Buffer length test");
+        TestScopeLogger LogScope0("Buffer length test");
         //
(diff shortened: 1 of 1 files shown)


src/test/lib/ApiTest.cpp
```diff
@@ -1847,6 +1847,86 @@ void QuicTestConfigurationSetParam()
...
+                MsQuic->SetParam(
+                    Handle,
+                    Param,
+                    0,
+                    nullptr));
+        }
+    }
+
+    //
+    // Buffer starts from non-zero is not supported
```
nibanks, line 1902: I've had this thought a couple of times on other tests/PRs, but what do we want to do if/when we actually start supporting some of these negative test cases? Eventually we would like to support non-zero offsets for instance.

ami-GS: What I could do is to add comment for the time when someone try to enable the feature to remove the test without concern.

nibanks (approved):

PR #2782, opened 2022-06-03
Support multiple expected certificate validation results in tests

### LINKED ISSUE #2758: Fix downlevel certificate tests to expect new error code
This will require refactoring the `TestConnection` class to have a list of expected/allowed shutdown codes.

anrossi:
## Description

Fixes #2758, but will need to be backported to the 2.0 release in order for the tests to pass.

## Testing

CI.

## Documentation

N/A


nibanks:
> Fixes #2758, but will need to be backported to the 2.0 release in order for the tests to pass.

Can you please create another PR to fix this in 2.0? We will merge that first, create a new 2.0.* release and then update this PR to point to that release's test binaries to make sure it all works.

PR #2785, opened 2022-06-04
Add support for Schannel in-proc certificate API.

### LINKED ISSUE #2750: Improve Schannel certificate retrieval speed

Replace the existing call to `SECPKG_ATTR_REMOTE_CERTIFICATES` with the faster in-proc version. Update the portable certificates code to parse this new structure.

### Proposed solution

See above.

### Additional context

This is a breaking change since the platform-dependent data structure in the `PEER_CERTIFICATE_RECEIVED` event will change.
Splitting this out from #2573 because it's more work and the breaking change aspect is risky.

anrossi:
## Description

Add support for faster Schannel certificate API. Fixes #2750.

## Testing

New tests will need to be added to cover the new API. Tests are added to cover the new flag.

## Documentation

Documentation will need to be updated to include the new flag.


src/platform/tls_schannel.c
```diff
@@ -369,6 +374,10 @@ typedef struct _SecPkgCred_SessionTicketKeys
 #define SECPKG_ATTR_CLIENT_CERT_POLICY   0x60   // sets    SecPkgCred_ClientCertCtlPolicy
 #endif
 
+#ifndef SECPKG_ATTR_SERIALIZED_REMOTE_CERT_CONTEXT_INPROC
+#define SECPKG_ATTR_SERIALIZED_REMOTE_CERT_CONTEXT_INPROC 0x74 // returns CERT_BLOB, use only after SSPI handshake loop
+#endif
```
nibanks, line 379: This is fairly recent in Windows, right? As in, it's not supported on Win10? Assuming so, we need to document this as such.

src/platform/tls_schannel.c
```diff
@@ -1656,30 +1677,67 @@ CxPlatTlsIndicateCertificateReceived(
...
 #endif
 
+    if (PeerCertBlob->Type == QUIC_CERT_BLOB_NONE) {
+        Certificate = NULL;
+        CertificateChain = NULL;
+    } else if (PeerCertBlob->Type == QUIC_CERT_BLOB_CHAIN) {
+#ifndef _KERNEL_MODE
+        CXPLAT_DBG_ASSERT(FALSE);
+        Certificate = NULL;
+        CertificateChain = NULL;
```
nibanks, line 1687: Why do you have these, when you just set them again in the lines after this?

nibanks: Same below this.

src/platform/unittest/TlsTest.cpp
```diff
@@ -2048,7 +2048,7 @@ ValidateSecConfigStatusSchannel(
 TEST_F(TlsTest, PlatformSpecificFlagsSchannel)
 {
     for (auto TestFlag : { QUIC_CREDENTIAL_FLAG_ENABLE_OCSP, QUIC_CREDENTIAL_FLAG_USE_SUPPLIED_CREDENTIALS,
-        QUIC_CREDENTIAL_FLAG_USE_SYSTEM_MAPPER,
+        QUIC_CREDENTIAL_FLAG_USE_SYSTEM_MAPPER, QUIC_CREDENTIAL_FLAG_SCHANNEL_SERIALIZED_PEER_CERT,
```
nibanks, line 2051: We need more tests than this. We need to try calling a connection with the new flag. I expect that test to pass on WS2022, but fail on WS2019.

src/inc/msquic.h
```diff
@@ -118,6 +118,7 @@ typedef enum QUIC_CREDENTIAL_FLAGS {
     QUIC_CREDENTIAL_FLAG_USE_SYSTEM_MAPPER                      = 0x00010000, // Schannel only
     QUIC_CREDENTIAL_FLAG_CACHE_ONLY_URL_RETRIEVAL               = 0x00020000, // Windows only currently
     QUIC_CREDENTIAL_FLAG_REVOCATION_CHECK_CACHE_ONLY            = 0x00040000, // Windows only currently
+    QUIC_CREDENTIAL_FLAG_SCHANNEL_SERIALIZED_PEER_CERT          = 0x00080000, // Schanel only
```
nibanks, line 121: I don't care for having `SCHANNEL` in the name. We already have plenty of Schannel specific flags that we just comment as such. Let's generalize this a bit more. Who knows? Perhaps we add support to OpenSSL one day for some reason...

anrossi: call this `IN_PROC_PEER_CERTIFICATE`

anrossi: fix typo in comment

src/platform/tls_schannel.c
anrossi: Yes, it's so new that it's not even in the SDK.

src/platform/tls_schannel.c
anrossi: I get "uninitialized variable" compilation errors without these. annoying, right?

src/platform/unittest/TlsTest.cpp
anrossi: Added a Tls unit test which gets to the heart of the feature, and the existing unit tests caught regressions.

.azure/azure-pipelines.ci.yml
```diff
@@ -602,7 +602,7 @@ stages:
       platform: windows
       tls: schannel
       logProfile: Full.Light
-      extraTestArgs: -Filter -*ValidateConfiguration:*ValidAlpnLengths:*ResumeRejection*:*ClientCertificate*:*LoadBalanced*
+      extraTestArgs: -Filter -*ValidateConfiguration:*ValidAlpnLengths:*ResumeRejection*:*ClientCertificate*:*LoadBalanced*:TlsTest.InProc*
```
nibanks, line 605: So they don't run on WS2022? Only Win-Latest (internal CI)?

src/inc/msquic.h
```diff
@@ -118,6 +118,7 @@ typedef enum QUIC_CREDENTIAL_FLAGS {
     QUIC_CREDENTIAL_FLAG_USE_SYSTEM_MAPPER                      = 0x00010000, // Schannel only
     QUIC_CREDENTIAL_FLAG_CACHE_ONLY_URL_RETRIEVAL               = 0x00020000, // Windows only currently
     QUIC_CREDENTIAL_FLAG_REVOCATION_CHECK_CACHE_ONLY            = 0x00040000, // Windows only currently
+    QUIC_CREDENTIAL_FLAG_INPROC_PEER_CERT                       = 0x00080000, // Schannel only
```
nibanks, line 121: For _most_ names above we try to fully type out `CERTIFICATE`. I think we should do so here as well.

src/platform/cert_capi.c
```diff
@@ -700,6 +710,208 @@ CxPlatCertParseChain(
...
+        CertOpenStore(
+            CERT_STORE_PROV_SERIALIZED,
+            X509_ASN_ENCODING,
+            0,
+            CERT_STORE_DEFER_CLOSE_UNTIL_LAST_FREE_FLAG,
+            SerializedCertificateStore);
+    if (SerializedStore == NULL) {
+        LastError = GetLastError();
+        if (LastError == ERROR_OUTOFMEMORY ||
+            LastError == ERROR_NOT_ENOUGH_MEMORY) {
```
nibanks, line 740: Why either?

src/platform/cert_capi.c
```diff
@@ -700,6 +710,208 @@ CxPlatCertParseChain(
...
+    TempCertStore =
+        CertOpenStore(
+            CERT_STORE_PROV_MEMORY,
+            X509_ASN_ENCODING | PKCS_7_ASN_ENCODING,
+            0,
+            CERT_STORE_ENUM_ARCHIVED_FLAG,
+            NULL);
+    if (NULL == TempCertStore) {
+        LastError = GetLastError();
+        Status = LastError;
```
nibanks, line 761: Need to convert Win error to HRESULT.

nibanks: Same in error cases below.

.azure/azure-pipelines.ci.yml
nibanks: This makes me thing we need to onboard a run with internal machines for public CI.

nibanks (commented):
Pretty good. Just a couple of nits. I assume you've locally verified these tests on win-latest?

src/platform/cert_capi.c
anrossi: I copied this from internal, so I assume somebody found CertOpenStore could return a different error sometimes.

.azure/azure-pipelines.ci.yml
anrossi: I don't understand the second comment. I think this only works on Win-Latest. I'll remove for the next commit and see if it fails CI.

anrossi: yep it fails; only works on Win-Latest. Let's chat tomrrow about the onboarding comment.

nibanks: Ok. We can follow up with possibly onboarding an internal CI run triggered from GitHub (in the future).

nibanks (approved):

Issue #2763, opened 2022-05-27
Shared EC Hang on Shutdown


nibanks:
### Describe the bug

Found from: https://dev.azure.com/ms/msquic/_build/results?buildId=316196&view=results

Looking at [quic.log](https://github.com/microsoft/msquic/files/8787166/quic.log) from the failure, I see that the `NthAllocFail` test failed with a watchdog timeout. From the logs, I could see that we end up hanging when we call `RegistrationClose`. Looking at the dump, the main thread is in the following callstack:

```
01 000000ec`0c58edc0 00007ffa`6d4dfca0 KERNELBASE!WaitForSingleObjectEx+0x8e
02 000000ec`0c58ee60 00007ffa`6d4df402 msquic!QuicWorkerUninitialize+0x80 [D:\a\1\msquic\src\core\worker.c @ 158] 
03 000000ec`0c58eeb0 00007ffa`6d4b41de msquic!QuicWorkerPoolUninitialize+0x52 [D:\a\1\msquic\src\core\worker.c @ 824] 
04 000000ec`0c58eef0 00007ff6`66bb1fa9 msquic!MsQuicRegistrationClose+0x12e [D:\a\1\msquic\src\core\registration.c @ 249] 
05 000000ec`0c58ef40 00007ff6`66be27f5 msquictest!MsQuicRegistration::~MsQuicRegistration+0x89 [D:\a\1\msquic\src\inc\msquic.hpp @ 342] 
06 000000ec`0c58ef90 00007ff6`66a78202 msquictest!QuicTestNthAllocFail+0x8b5 [D:\a\1\msquic\src\test\lib\DataTest.cpp @ 2330] 
07 000000ec`0c58f4c0 00007ff6`66c3d5ed msquictest!Misc_NthAllocFail_Test::TestBody+0x1d2 [D:\a\1\msquic\src\test\bin\quic_gtest.cpp @ 1594] 
08 000000ec`0c58f5e0 00007ff6`66c1695d msquictest!testing::internal::HandleExceptionsInMethodIfSupported<testing::Test,void>+0xad [D:\a\1\msquic\submodules\googletest\googletest\src\gtest.cc @ 2657] 
09 000000ec`0c58f6c0 00007ff6`66c1765e msquictest!testing::Test::Run+0xcd [D:\a\1\msquic\submodules\googletest\googletest\src\gtest.cc @ 2681] 
[...]

PR #2790, opened 2022-06-04
Add synchronization for EC contexts

csujedihy:
## Description

Add synchronization for EC contexts to solve #2763. Since EC contexts are uncontended except for registration init/uninit, simply taking a CS lock to synchronize the access should be performant enough.

## Testing

CI

### DIFF
--- a/src/platform/platform_worker.c
+++ b/src/platform/platform_worker.c
@@ -224,15 +224,9 @@ CxPlatRunExecutionContexts(
 {
     Worker->ECsReady = FALSE;
     Worker->ECsReadyTime = UINT64_MAX;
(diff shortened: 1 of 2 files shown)


src/platform/platform_worker.c
```diff
@@ -188,8 +202,10 @@ CxPlatAddExecutionContext(
 {
     CXPLAT_WORKER* Worker = &CxPlatWorkers[IdealProcessor % CxPlatWorkerCount];
     Context->CxPlatContext = Worker;
+    CxPlatLockAcquire(&Worker->ECLock);
```
nibanks, line 205: Remove the TODO in the comment above.

nibanks (approved):

PR #2795, opened 2022-06-06
Try Slight Modification to BPF Code

nibanks:
## Description

Trying out a change suggested on Discord that might improve Linux throughput.

## Testing

Perf pipeline

## Documentation

N/A

### DIFF
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -914,6 +914,7 @@ CxPlatSocketConfigureRss(
 
     struct sock_filter BpfCode[] = {
         {BPF_LD | BPF_W | BPF_ABS, 0, 0, SKF_AD_OFF | SKF_AD_CPU},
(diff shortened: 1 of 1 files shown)


nibanks:
Didn't seem to really help.

rpenaran:
Now, I'm thinking I remember seeing some code where you increase the CPU by 2 to select the virtual CPU that in reality is the same one but using hyperthreading. I'm wandering if this is the case too and we are not alleviating the collision. In my side, it is +24 instead of +2.

At the end, what I discovered is that when using the throughput test downloading, the client opens a socket in CPU X, sends the requests and then it receives traffic and HW interruptions are handle in CPU Y (because of RSS hash assignation). Y and X may be the same but it is not common. On the other way, when uploading the socket and the HW interruptions are colliding in the same CPU because of the port sharing sockets and the filter (that I changed).

PR #2797, opened 2022-06-06
Changing BPF filter to avoid collisions when selecting CPU between HW interruptions and socket managing  

rpenaran:
## Description

I discovered that when using the throughput test downloading, the client opens a socket in CPU X, sends the requests and then it receives traffic and HW interruptions are handle in CPU Y (because of RSS hash assignation). Y and X may be the same but it is not common. On the other way, when uploading the socket and the HW interruptions are colliding in the same CPU because of the port sharing sockets and the filter (that I changed).

## Testing

throughput test comparing upload and download version should be enough

### DIFF
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -914,6 +914,7 @@ CxPlatSocketConfigureRss(
 
     struct sock_filter BpfCode[] = {
         {BPF_LD | BPF_W | BPF_ABS, 0, 0, SKF_AD_OFF | SKF_AD_CPU},
+        {BPF_ALU | BPF_ADD, 0, 0, 3},
         {BPF_ALU | BPF_MOD, 0, 0, SocketCount},
         {BPF_RET | BPF_A, 0, 0, 0}
     };


nibanks:
No significant perf improvements seen in automation.

rpenaran:
Well, I don't know then. What I know is that I see that working in my side, maybe it is more exacerbated because I'm working with higher speeds. I suppose I'll keep it in my machine. 
On the other hand, this also brings a discussion about latency, For latency, it would be better to have everything in the same CPU. For throughput, if the CPU is really busy, using another one always help.

nibanks:
> On the other hand, this also brings a discussion about latency, For latency, it would be better to have everything in the same CPU. For throughput, if the CPU is really busy, using another one always help.

This is exactly the type of thing we (ideally) wanted to support exposing to an application to control via `QUIC_EXECUTION_PROFILE` ([link](https://github.com/microsoft/msquic/blob/main/src/inc/msquic.h#L78)). The app could choose to use `QUIC_EXECUTION_PROFILE_TYPE_MAX_THROUGHPUT` and we'd configure to use multiple CPUs instead of all on the same.

PR #2811, opened 2022-06-09
RPS Latency Analysis Improvements

nibanks:
## Description

More improvements for understanding RPS latency

## Testing

Perf pipeline and local usage of tool.

## Documentation

N/A

### DIFF
--- a/src/plugins/trace/exe/Program.cs
+++ b/src/plugins/trace/exe/Program.cs
@@ -213,6 +213,11 @@ namespace QuicTrace
             public ulong LastStateChangeTime = 0;
 
             //
+            // The application is actively handling a receive.
+            //
+            public bool InAppRecv = false;
+
+            //
             // Triggers a state change and updates variables accordingly.
             //
             public void UpdateToState(RequestState state, ulong time, bool ignorePrevious = false)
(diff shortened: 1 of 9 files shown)


src/plugins/trace/exe/Program.cs
```diff
@@ -232,6 +238,35 @@ public void UpdateToState(RequestState state, ulong time, bool ignorePrevious =
                 State = state;
             }
 
+            public void UpdateToIdle(ulong time)
```
maolson-msft, line 241: Why a separate UpdateToIdle method when there's an UpdateToState method and "Idle" is just another RequestState value?

nibanks: To be clear this is updating to **an** idle state, but not a specific one, because it's dependent on the current state of send/recv.

nibanks: `Idle` is equivalent to "idle with no send or recv yet" but I didn't want to write that all in an enum. This code isn't just changing to that state, but one of 4:
```c
                if (FirstPacketRecv != 0 && FirstPacketSend != 0)
                {
                    State = RequestState.IdleBoth;
                }
                else if (FirstPacketRecv != 0)
                {
                    State = RequestState.IdleRecv;
                }
                else if (FirstPacketSend != 0)
                {
                    State = RequestState.IdleSent;
                }
                else
                {
                    State = RequestState.Idle;
                }
```

maolson-msft: Gotcha.

maolson-msft (approved):

PR #2821, opened 2022-06-15
Add GetParam test for Tls

ami-GS:
## Description

GetParam tests for Tls #1215

## Testing

This is test

## Documentation

N/A


nibanks (approved):

nibanks:
Test failure:
```
[ RUN      ] ParameterValidation.ValidateTlsParam
D:\a\1\msquic\src\test\lib\ApiTest.cpp(2928): error: Connection.GetParam( QUIC_PARAM_TLS_HANDSHAKE_INFO, &Length, &Info) not equal to QUIC_STATUS_INVALID_STATE, 0x80090301
[  FAILED  ] ParameterValidation.ValidateTlsParam (23 ms)
```

src/test/lib/ApiTest.cpp
```diff
@@ -2885,29 +2885,103 @@ void QuicTestTlsSetParam()
...
+                    nullptr));
+            TEST_EQUAL(Length, sizeof(QUIC_HANDSHAKE_INFO));
+
+            //
+            // Before handshake
+            //
+            {
+                TestScopeLogger LogScope2("Before handshake");
+                QUIC_HANDSHAKE_INFO Info = {};
+                TEST_QUIC_STATUS(
```
nibanks, line 2923: This is failing on schannel because of a call to `QueryContextAttributesW` (I think) not returning the expected "invalid state" error. Really, I think the product code (in TLS?) needs to explicitly check if things have been started, and if not return "invalid state" before calling into schannel or openssl. But otherwise, the test code is good. Please update the product code.

nibanks (changes requested):

src/test/lib/ApiTest.cpp
nibanks: Why do you want to change that? Just add a check before that to return invalid state?

nibanks: I _think_ you just need to check for non-NULL TLS?

ami-GS: You mean this?
https://github.com/microsoft/msquic/blob/5d076de8734ec22c4b6388789b5d70cbb2f1d505/src/core/library.c#L1370-L1377

this TLS should be initialized by just ConnectionStart

nibanks: Ah, yeah. I guess it's already validated there.

ami-GS: Should I add
`else if (Connection->Crypto.TlsState.HandshakeComplete) {Status = QUIC_STATUS_INVALID_STATE}`
here?
If so, all the parameters processed in CxPlatTlsParamGet depends on the HandshakeComplete.
Two common parameters looks good, but not sure other 3 parameters should be depends on it.
QUIC_PARAM_TLS_SCHANNEL_CONTEXT_ATTRIBUTE_W
QUIC_PARAM_TLS_SCHANNEL_CONTEXT_ATTRIBUTE_EX_W
QUIC_PARAM_TLS_SCHANNEL_SECURITY_CONTEXT_TOKEN

Can I treat `Only valid until the QUIC_CONNECTION_EVENT_CONNECTED event, or when TLS is cleaned up.` as after HandshakeComplete?

ami-GS: Schannel case doesn't work for all 5 parameters with non-HandshakeComplete state.
The 3 Schannel specific parameters have not been tested with HandshakeComplete state though
Might need to update TlsTest.cpp

.github/workflows/codeql-analysis.yml
```diff
@@ -48,7 +48,7 @@ jobs:
          cmake --build . --target OpenSSL_Target
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@27ea8f8fe5977c00f5b37e076ab846c5bd783b96
+        uses: github/codeql-action/init@41a4ada31ba866a7f1196b9602703a89edd69e22
```
nibanks, line 51: Please undo these type of changes

ami-GS: yeah, I edited windows machine which didn't have latest main.
Now fixed by rebasing to latest main

nibanks (approved):

PR #2833, opened 2022-06-21
Nibanks/streamstatetable

nibanks:
## Description

Adds a table to show stream state changes over time.

image

## Testing

N/A

## Documentation

N/A

### DIFF
--- a/src/plugins/trace/dll/DataModel/QuicStream.cs
+++ b/src/plugins/trace/dll/DataModel/QuicStream.cs
@@ -24,7 +24,7 @@ namespace QuicTrace.DataModel
 
         public uint ProcessId { get; }
 
-        public ulong StreamId { get; private set; }
+        public ulong StreamId { get; private set; } = ulong.MaxValue;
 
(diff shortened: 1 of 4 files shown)


csujedihy (approved):

src/plugins/trace/dll/DataModel/QuicStream.cs
```diff
@@ -191,6 +191,15 @@ internal void AddEvent(QuicEvent evt, QuicState state)
                                 Timings.UpdateToState(QuicStreamState.ProcessRecv, Connection.LastScheduleStateTimeStamp, true);
                             }
                             Timings.UpdateToState(QuicStreamState.Decrypt, Timings.RecvPacket.PacketDecrypt);
+
+                            if (Timings.RecvPacket.PacketDecryptComplete == Timestamp.Zero)
```
csujedihy, line 195: we need to do this for encryption as well?

nibanks: No. It's not quite the same.

PR #2835, opened 2022-06-22
Try to send inline from DpRawTxEnqueue

csujedihy:
## Description

In CxPlatDpRawTxEnqueue, if we are enqueuing TX packets onto a queue that's polled by the same thread, try to do send the packets inline.

## Testing

CI

## Documentation

N/A

### DIFF
--- a/src/platform/datapath_raw_xdp.c
+++ b/src/platform/datapath_raw_xdp.c
@@ -33,6 +33,7 @@ Abstract:
 #define PORT_SET_TAG  'PpdX' // XdpP
 
 typedef struct XDP_INTERFACE XDP_INTERFACE;
+typedef struct QUIC_CACHEALIGN XDP_WORKER XDP_WORKER;
 
 typedef struct XDP_QUEUE {
     const XDP_INTERFACE* Interface;
@@ -47,6 +48,7 @@ typedef struct XDP_QUEUE {
     XSK_RING TxRing;
     XSK_RING TxCompletionRing;
     BOOL Error;
+    XDP_WORKER* XdpWorker;
 
     CXPLAT_LIST_ENTRY WorkerTxQueue;
     CXPLAT_SLIST_ENTRY WorkerRxPool;
@@ -85,6 +87,7 @@ void XdpWorkerAddQueue(_In_ XDP_WORKER* Worker, _In_ XDP_QUEUE* Queue) {
     }
     *Tail = Queue;
     Queue->Next = NULL;
+    Queue->XdpWorker = Worker;
 }
 
 typedef struct XDP_DATAPATH {
@@ -1486,19 +1489,6 @@ CxPlatDpRawTxFree(
     InterlockedPushEntrySList(&Packet->Queue->TxPool, (PSLIST_ENTRY)Packet);
 }
 
-_IRQL_requires_max_(DISPATCH_LEVEL)
-void
-CxPlatDpRawTxEnqueue(
-    _In_ CXPLAT_SEND_DATA* SendData
-    )
-{
-    XDP_TX_PACKET* Packet = (XDP_TX_PACKET*)SendData;
-
-    CxPlatLockAcquire(&Packet->Queue->TxLock);
-    CxPlatListInsertTail(&Packet->Queue->TxQueue, &Packet->Link);
-    CxPlatLockRelease(&Packet->Queue->TxLock);
-}
-
 static
 void
 CxPlatXdpTx(
@@ -1567,6 +1557,25 @@ CxPlatXdpTx(
         XskStatus = XskGetSockopt(Queue->TxXsk, XSK_SOCKOPT_TX_ERROR, &ErrorStatus, &ErrorSize);
         printf("TX ring error: 0x%x\n", SUCCEEDED(XskStatus) ? ErrorStatus : XskStatus);
         Queue->Error = TRUE;
+    }
+}
+
+_IRQL_requires_max_(DISPATCH_LEVEL)
+void
+CxPlatDpRawTxEnqueue(
+    _In_ CXPLAT_SEND_DATA* SendData,
+    _In_ uint16_t IdealProcessor
+    )
+{
+    XDP_TX_PACKET* Packet = (XDP_TX_PACKET*)SendData;
+    CXPLAT_DBG_ASSERT(IdealProcessor == CxPlatProcCurrentNumber());
+    if (Packet->Queue->XdpWorker->ProcIndex == IdealProcessor) {
+        CxPlatListInsertTail(&Packet->Queue->WorkerTxQueue, &Packet->Link);
+        CxPlatXdpTx(Packet->Queue->XdpWorker->Xdp, Packet->Queue);
+    } else {
+        CxPlatLockAcquire(&Packet->Queue->TxLock);
+        CxPlatListInsertTail(&Packet->Queue->TxQueue, &Packet->Link);
+        CxPlatLockRelease(&Packet->Queue->TxLock);
     }
 }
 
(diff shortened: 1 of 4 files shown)


csujedihy:
On our lola VMs, it seems to improve latency by a bit.
image

csujedihy:
Not much improvement regarding latency but throughput seems to be noticeably improved.

src/platform/datapath_raw_xdp.c
```diff
@@ -47,6 +48,7 @@ typedef struct XDP_QUEUE {
     XSK_RING TxRing;
     XSK_RING TxCompletionRing;
     BOOL Error;
+    XDP_WORKER* XdpWorker;
```
nibanks, line 51: Let's put this at the top, after `Interface` and just call this `Worker`. Can it be `const` too?

src/platform/datapath_raw_xdp.c
```diff
@@ -1570,6 +1560,26 @@ CxPlatXdpTx(
...
+    XDP_QUEUE* Queue = Packet->Queue;
+    CXPLAT_DBG_ASSERT(IdealProcessor == CxPlatProcCurrentNumber());
+    if (Queue->XdpWorker->ProcIndex == IdealProcessor) {
+        CxPlatListInsertTail(&Queue->WorkerTxQueue, &Packet->Link);
+        CxPlatXdpTx(Queue->XdpWorker->Xdp, Queue);
+    } else {
+        CxPlatLockAcquire(&Packet->Queue->TxLock);
+        CxPlatListInsertTail(&Queue->TxQueue, &Packet->Link);
+        CxPlatLockRelease(&Packet->Queue->TxLock);
+    }
```
nibanks, line 1580: So far it does seem to be increasing latency in higher connection count tests. I am wondering if the overhead of calling the whole `CxPlatXdpTx` is too much. Let's try replacing this call with partial logic just to enqueue this packet to XDP and poke as necessary. Nothing else.

csujedihy: I tried that in earlier commits.

csujedihy: Yeah, [32920cf](https://github.com/microsoft/msquic/pull/2835/commits/32920cf81af6f9b93a958911fa3aca31dbeac64d)

csujedihy: In the perf pipeline, the first one for this PR is only queue and poke. The second one is call the whole thing.

src/platform/datapath_raw_xdp.c
```diff
@@ -1570,6 +1560,27 @@ CxPlatXdpTx(
...
+_IRQL_requires_max_(DISPATCH_LEVEL)
+void
+CxPlatDpRawTxEnqueue(
+    _In_ CXPLAT_SEND_DATA* SendData,
+    _In_ BOOLEAN InlineHint
+    )
+{
+    XDP_TX_PACKET* Packet = (XDP_TX_PACKET*)SendData;
+    XDP_QUEUE* Queue = Packet->Queue;
+    if (Queue->Worker->ProcIndex == CxPlatProcCurrentNumber()) {
```
nibanks, line 1572: Maybe, instead of always querying proc, just use compile-time `#ifdef QUIC_USE_EXECUTION_CONTEXTS`.

src/inc/quic_datapath.h
```diff
@@ -639,7 +639,8 @@ CxPlatSocketSend(
     _In_ CXPLAT_SOCKET* Socket,
     _In_ const CXPLAT_ROUTE* Route,
     _In_ CXPLAT_SEND_DATA* SendData,
-    _In_ uint16_t PartitionId
+    _In_ uint16_t PartitionId,
+    _In_ BOOLEAN InlineHint
```
nibanks, line 643: Doing it this way seems to be extremely disruptive. What if, instead, we defined:
```c
void
CxPlatSendDataSetInlineHint(
    _In_ CXPLAT_SEND_DATA* SendData,
    _In_ BOOLEAN SendInline
    );
```
And then, for datapaths that support this, it saves it on the `CXPLAT_SEND_DATA` internally, and uses the state when the app later calls `CxPlatSocketSend`.

csujedihy: Yeah, I thought about this but went with a more difficult way..

src/inc/quic_datapath.h
```diff
@@ -709,6 +709,26 @@ CxPlatResolveRoute(
...
+//
+// Save the processor index to be used later when queuing TX to datapath as an
+// optimization.
+//
+_IRQL_requires_max_(DISPATCH_LEVEL)
+void
+CxPlatSendDataProcIndex(
+    _In_ CXPLAT_SEND_DATA* SendData,
+    _In_ uint16_t ProcIndex
+    );
```
nibanks, line 730: 1. You're going to have to implement `CxPlatSendDataProcIndex` for the other data paths.
2. The proc index is also being passed in the send call, right? It should be one place or the other. Not both.
3. Instead of wrapping the calls to `CxPlatSendDataSetInlineHint` with `#ifdef`s, just implement it in the other data paths too, with a no-op. The compiler should be smart enough to make it efficient.

src/platform/datapath_raw_xdp.c
```diff
@@ -1570,6 +1560,26 @@ CxPlatXdpTx(
...
+_IRQL_requires_max_(DISPATCH_LEVEL)
+void
+CxPlatDpRawTxEnqueue(
+    _In_ CXPLAT_SEND_DATA* SendData
+    )
+{
+    XDP_TX_PACKET* Packet = (XDP_TX_PACKET*)SendData;
+    XDP_QUEUE* Queue = Packet->Queue;
+    if (Queue->Worker->ProcIndex == SendData->ProcIndex) {
+        CxPlatListInsertTail(&Queue->WorkerTxQueue, &Packet->Link);
```
nibanks, line 1572: You're introducing reordering by putting this into the `WorkerTxQueue` directly.

nibanks: Actually, never mind.

src/platform/datapath_raw_xdp.c
```diff
@@ -1570,6 +1553,55 @@ CxPlatXdpTx(
...
+static
+void
+CxPlatXdpFlushTxQueue(
+    _In_ XDP_QUEUE* Queue
+    )
+{
+    //
+    // Move packets from the TX queue (i.e. packets from other threads) to the
+    // worker TX queue (used to actually send packets), but only if we don't
+    // already have anything in the worker TX queue (why?).
```
nibanks, line 1565: @mtfriesen I believe you wrote the original logic (which we are just moving here), so could you answer the "why?" part? Why do we only move the TxQueue to the WorkerTxQueue if the WorkerTxQueue is currently empty?

src/inc/quic_datapath.h
nibanks: Or, maybe move the `#ifdef`s here. On non-raw datapaths, define these to be empty macros, or perhaps undefined_parameter calls.

src/inc/quic_datapath.h
```diff
@@ -709,6 +709,31 @@ CxPlatResolveRoute(
...
+
+#else
+
+#define CxPlatSendDataSetInlineHint
+#define CxPlatSendDataProcIndex
```
nibanks, line 735:
```suggestion
#define CxPlatSendDataSetInlineHint(SendData) UNREFERENCED_PARAMETER(SendData)
#define CxPlatSendDataProcIndex(SendData, ProcIndex) UNREFERENCED_PARAMETER(SendData)
```

csujedihy: hmm, I thought it compiled.

PR #2839, opened 2022-06-23
Remove Unnecessary XDP Synchronization in Shared EC Mode

nibanks:
## Description

When run in shared execution context mode, there is no need for extra synchronization around the XDP TX and RX queues, since they're always accessed by a single thread.

## Testing

Existing automation.

## Documentation

N/A


nibanks:
Hmmm. It seems even in SharedEC we're still getting some cross-EC usage.

csujedihy:
> Hmmm. It seems even in SharedEC we're still getting some cross-EC usage.

Yeah, the default TX queue is 0 for a new connection before we get any packets and migrate the connection to the receiving core.

PR #2846, opened 2022-06-27
Increase XDP RX Batch Size

nibanks:
## Description

See if increasing XDP RX batch size improves performance.

## Testing

CI

## Documentation

N/A


csujedihy:
That fixing PTO PR really busted a few unreliable datagram test cases.

PR #2850, opened 2022-06-29
Improve WPA stream state plugin to show when receive complete API is called

csujedihy:
## Description

Added a new state called ProcessAppRecv. We transition from AppRecv to ProcessAppRecv when the app calls receive complete.

image

## Testing

Tested locally with WPA.

### DIFF
--- a/src/plugins/trace/dll/DataModel/QuicStream.cs
+++ b/src/plugins/trace/dll/DataModel/QuicStream.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT License.
 //
 
+using System;
 using System.Collections.Generic;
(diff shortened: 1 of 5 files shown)


src/plugins/trace/dll/DataModel/QuicStream.cs
```diff
@@ -3,6 +3,7 @@
 // Licensed under the MIT License.
 //
 
+using System;
```
nibanks, line 6: This isn't necessary, right?

csujedihy: right

nibanks (approved):

PR #2856, opened 2022-06-30
RPS Perf Client Custom Local Addresses

nibanks:
## Description

Allows for the command line to specify a set of local addresses that are used for RPS perf testing.

## Testing

Locally tested

## Documentation

N/A

### DIFF
--- a/src/perf/lib/RpsClient.cpp
+++ b/src/perf/lib/RpsClient.cpp
@@ -24,6 +24,8 @@ PrintHelp(
         "RPS Client options:\n"
         "\n"
         "  -target:<####>              The target server to connect to.\n"
+        "  -bind:<addr>                Local IP address(es) to bind to.\n"
+        "  -addrs:<####>               The number of local addresses to use (def:%u).\n"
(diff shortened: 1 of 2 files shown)


src/perf/lib/RpsClient.cpp
```diff
@@ -24,6 +24,8 @@ PrintHelp(
         "RPS Client options:\n"
         "\n"
         "  -target:<####>              The target server to connect to.\n"
+        "  -bind:<addr>                Local IP address(es) to bind to.\n"
```
csujedihy, line 27: >Local IP address(es) to bind to [](http://example.com/codeflow?start=39&length=31)

only IP address or it could be IP:port? Is this the change that you used to make secnetperf generate consistent perf on lab machines? If so, you bound to the same set of ports, right?

nibanks: Yes, you can use a port. `1.1.1.1:8000,1.1.1.2:8001`

csujedihy: the text should be changed to reflect that.

csujedihy (approved):

PR #2865, opened 2022-07-04
Stream send & Session resumption for Conn Close test

ami-GS:
## Description

StreamOpen/Start/Send from both Server/Client
Send Session resumption ticket from server

## Testing

This is test

## Documentation

N/A

### DIFF
--- a/src/test/lib/ApiTest.cpp
+++ b/src/test/lib/ApiTest.cpp
@@ -1152,18 +1152,11 @@ DummyStreamCallback(
     return QUIC_STATUS_SUCCESS;
 }
 
-
 struct CloseFromCallbackContext {
     uint16_t CloseCount;
     volatile uint16_t CurrentCount;
-    // assuming server/client have 1 connection
-    MsQuicStream* ServerStreams[10] = {};
-    uint8_t NumServerStream {0};
-    MsQuicStream* ClientStreams[10] = {};
-    uint8_t NumClientStream {0};
-
-    uint8_t LRawBuffer[100];
-    QUIC_BUFFER LBuffer { sizeof(LRawBuffer), LRawBuffer };
+    uint8_t RawBuffer[100];
+    QUIC_BUFFER BufferToSend { sizeof(RawBuffer), RawBuffer };
 
     static QUIC_STATUS StreamCallback(_In_ MsQuicStream* Stream, _In_opt_ void*, _Inout_ QUIC_STREAM_EVENT* Event) {
 
@@ -1182,16 +1175,15 @@ struct CloseFromCallbackContext {
         }
         if (IsServer) {
             if (Event->Type == QUIC_CONNECTION_EVENT_CONNECTED) {
+                Conn->SendResumptionTicket();
+
                 QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
-
-                MsQuicStream* Stream = new MsQuicStream(*Conn, QUIC_STREAM_OPEN_FLAG_UNIDIRECTIONAL, CleanUpAutoDelete, StreamCallback, Context);
-                Ctx->ServerStreams[Ctx->NumServerStream++] = Stream; // to delete if connection is deleted earlier
-                Stream->Start(QUIC_STREAM_START_FLAG_NONE);
-                Stream->Send(&Ctx->LBuffer, 1, QUIC_SEND_FLAG_START | QUIC_SEND_FLAG_FIN);
-
-                if (QUIC_FAILED(Status = MsQuic->ConnectionSendResumptionTicket(Conn->Handle, QUIC_SEND_RESUMPTION_FLAG_NONE, 0, NULL))) {
-                    TEST_FAILURE("Failed to send ResumptionTicket %d", Status);
-                    return Status;
+                MsQuicStream* Stream = new(std::nothrow) MsQuicStream(*Conn, QUIC_STREAM_OPEN_FLAG_UNIDIRECTIONAL, CleanUpAutoDelete, StreamCallback, Context);
+                if (QUIC_FAILED(Status = Stream->Start(QUIC_STREAM_START_FLAG_SHUTDOWN_ON_FAIL))) {
+                    Stream->Close();
+                }
+                if (QUIC_FAILED(Status = Stream->Send(&Ctx->BufferToSend, 1, QUIC_SEND_FLAG_START | QUIC_SEND_FLAG_FIN))) {
+                    Stream->Close();
                 }
             }
         }
@@ -1202,11 +1194,6 @@ struct CloseFromCallbackContext {
         auto count = (uint16_t)__sync_add_and_fetch((volatile short*)&Ctx->CurrentCount, 1);
 #endif
         if (Ctx->CloseCount == count-1) {
-            MsQuicStream** Streams = IsServer ? Ctx->ServerStreams : Ctx->ClientStreams;
-            uint8_t NumStream = IsServer ? Ctx->NumServerStream : Ctx->NumClientStream;
-            for (int i = 0; i < NumStream; i++) {
-                Streams[i]->Close();
-            }
             Conn->Close();
         }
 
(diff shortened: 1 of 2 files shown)


src/test/lib/ApiTest.cpp
```diff
@@ -1152,19 +1152,44 @@ DummyStreamCallback(
     return QUIC_STATUS_SUCCESS;
 }
 
+#include "unordered_map"
```
nibanks, line 1155: You can't use these because they don't exist in kernel mode Windows.

src/test/lib/ApiTest.cpp
```diff
@@ -1152,19 +1152,48 @@ DummyStreamCallback(
     return QUIC_STATUS_SUCCESS;
 }
 
+
```
nibanks, line 1155: nit: remove added line

src/test/lib/ApiTest.cpp
```diff
@@ -1152,19 +1152,48 @@ DummyStreamCallback(
...
+            // This stream send shutdown by receiving FIN
+            new(std::nothrow) MsQuicStream(Event->PEER_STREAM_STARTED.Stream, CleanUpAutoDelete, StreamCallback, Context);
+        }
+        if (IsServer) {
+            if (Event->Type == QUIC_CONNECTION_EVENT_CONNECTED) {
+                QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
+
+                MsQuicStream* Stream = new(std::nothrow) MsQuicStream(*Conn, QUIC_STREAM_OPEN_FLAG_UNIDIRECTIONAL, CleanUpAutoDelete, StreamCallback, Context);
+                Ctx->ServerStreams[Ctx->NumServerStream++] = Stream; // to delete if connection is deleted earlier
+                Stream->Start(QUIC_STREAM_START_FLAG_NONE);
```
nibanks, line 1189: The `Start` call is unnecessary since you are passing the `QUIC_SEND_FLAG_START` in the next call.

src/test/lib/ApiTest.cpp
```diff
@@ -1152,19 +1152,48 @@ DummyStreamCallback(
...
+                Stream->Send(&Ctx->LBuffer, 1, QUIC_SEND_FLAG_START | QUIC_SEND_FLAG_FIN);
+
+                if (QUIC_FAILED(Status = MsQuic->ConnectionSendResumptionTicket(Conn->Handle, QUIC_SEND_RESUMPTION_FLAG_NONE, 0, NULL))) {
```
nibanks, line 1192: We should add a `SendResumptionTicket` function to `MsQuicConnection` so you don't have to access the `Handle` directly.

nibanks: I think it might be better to make this call "best effort" and not fail if it does (just like `Send` above).

src/test/lib/ApiTest.cpp
```diff
@@ -1152,19 +1152,48 @@ DummyStreamCallback(
...
 
+
 struct CloseFromCallbackContext {
     uint16_t CloseCount;
     volatile uint16_t CurrentCount;
+    // assuming server/client have 1 connection
+    MsQuicStream* ServerStreams[10] = {};
+    uint8_t NumServerStream {0};
+    MsQuicStream* ClientStreams[10] = {};
+    uint8_t NumClientStream {0};
```
nibanks, line 1163: I _believe_ this is all mostly unnecessary. When you use `CleanUpAutoDelete` for a stream, it will automatically `Close` when the stream is shutdown.

ami-GS: No, streams won't be deleted automatically until QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE arrives.

I saw "abort" happened at the end of the test when CloseCount is 5-7. The Server close his Connection before closing Stream.

ami-GS: hmm, code flow is something like this. This does Open/Start/Send, but I don't see any stream event against the problematic stream.
```c++
MsQuicStream* Stream = new(std::nothrow) MsQuicStream(*Conn, QUIC_STREAM_OPEN_FLAG_UNIDIRECTIONAL, CleanUpAutoDelete, StreamCallback, Context);
Stream->Send(&Ctx->LBuffer, 1, QUIC_SEND_FLAG_START | QUIC_SEND_FLAG_FIN);
Conn->Close();
```

src/test/lib/ApiTest.cpp
```diff
@@ -1155,16 +1155,37 @@ DummyStreamCallback(
...
     volatile uint16_t CurrentCount;
+    uint8_t RawBuffer[100];
+    QUIC_BUFFER BufferToSend { sizeof(RawBuffer), RawBuffer };
 
-    static QUIC_STATUS StreamCallback(_In_ MsQuicStream*, _In_opt_ void*, _Inout_ QUIC_STREAM_EVENT*) {
+    static QUIC_STATUS StreamCallback(_In_ MsQuicStream* Stream, _In_opt_ void*, _Inout_ QUIC_STREAM_EVENT* Event) {
+
+        if (Event->Type == QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE) {
+            Stream->Close(); // or abort
+        }
```
nibanks, line 1165: You don't need this for auto cleanup/delete streams. It's done for you.

src/test/lib/ApiTest.cpp
```diff
@@ -1155,16 +1155,37 @@ DummyStreamCallback(
...
+                Conn->SendResumptionTicket();
+
+                QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
+                MsQuicStream* Stream = new(std::nothrow) MsQuicStream(*Conn, QUIC_STREAM_OPEN_FLAG_UNIDIRECTIONAL, CleanUpAutoDelete, StreamCallback, Context);
+                if (QUIC_FAILED(Status = Stream->Start(QUIC_STREAM_START_FLAG_SHUTDOWN_ON_FAIL))) {
+                    Stream->Close();
+                }
+                if (QUIC_FAILED(Status = Stream->Send(&Ctx->BufferToSend, 1, QUIC_SEND_FLAG_START | QUIC_SEND_FLAG_FIN))) {
+                    Stream->Close();
+                }
```
nibanks, line 1187: You don't need the check for fail and then close. Only for Start.

src/test/lib/ApiTest.cpp
```diff
@@ -1205,16 +1233,21 @@ QuicTestConnectionCloseFromCallback() {
...
 
-        MsQuicConnection Connection(Registration,  CleanUpManual, CloseFromCallbackContext::Callback, &Context);
+        MsQuicConnection Connection(Registration,  CleanUpManual, CloseFromCallbackContext::CallbackC, &Context);
         TEST_QUIC_SUCCEEDED(Connection.GetInitStatus());
         TEST_QUIC_SUCCEEDED(Connection.Start(ClientConfiguration, ServerLocalAddr.GetFamily(), QUIC_TEST_LOOPBACK_FOR_AF(ServerLocalAddr.GetFamily()), ServerLocalAddr.GetPort()));
 
+        MsQuicStream Stream(Connection, QUIC_STREAM_OPEN_FLAG_UNIDIRECTIONAL, CleanUpManual, CloseFromCallbackContext::StreamCallback, &Context);
+        TEST_QUIC_SUCCEEDED(Stream.GetInitStatus());
+        TEST_QUIC_SUCCEEDED(Stream.Start(QUIC_STREAM_START_FLAG_SHUTDOWN_ON_FAIL));
+        TEST_QUIC_SUCCEEDED(Stream.Send(&Context.BufferToSend, 1, QUIC_SEND_FLAG_START | QUIC_SEND_FLAG_FIN));
```
nibanks, line 1249: Can we really guarantee these always succeed, even if close happens in the first connection event?

src/test/lib/ApiTest.cpp
ami-GS: This is really tricky and was magically working.

`CleanUpAutoDelete` does not only close the handle, but also `delete` the instance.
We cannot use `CleanUpAutoDelete` for client `MsQuicStream` created in main loop, because the instance is allocated on stack area.
However, the client stream get `QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE` by sending data with `QUIC_SEND_FLAG_FIN`, but it doesn't call `StreamClose()` by removing line:1164. Then `~MsQuicStream` is called at the end of loop which call `StreamClose()`. This get aborted because Connection is already closed.

nibanks (approved):

Issue #2869, opened 2022-07-05
Complexities Around Stream Start via StreamSend Racing with Connection Close


nibanks:
### Describe the bug

When you call `StreamStart` and the connection is closed before that (async) API calls is processed, then you end up with the "start" being aborted, and you get a `QUIC_STREAM_EVENT_START_COMPLETE` event with the `QUIC_STATUS_ABORTED` status. If you pass `QUIC_STREAM_START_FLAG_SHUTDOWN_ON_FAIL` to `StreamStart` then you will also get a `QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE` event.

When you pass the `QUIC_SEND_FLAG_START` flag to `StreamSend` to (async) start the stream along with the send call, but the connection is closed before that, none of the above happen.

This makes it impossible to write good clean up code for this edge condition if you want to eliminate the `StreamStart` call and just use `StreamSend`. Two thoughts on this:

1. Aborting a `StreamSend` operation with the `QUIC_SEND_FLAG_START` flag should trigger the same aborted `QUIC_STREAM_EVENT_START_COMPLETE` event, as if you called `StreamStart`.
2. I _think_ we should also treat it as if the `QUIC_STREAM_START_FLAG_SHUTDOWN_ON_FAIL` version of `StreamStart` was called and trigger the `QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE` event too. It doesn't make sense not to do this IMO.

A bigger question I have though, is should `QUIC_STREAM_START_FLAG_SHUTDOWN_ON_FAIL` really be a thing at all, and not just the default behavior?

### Affected OS

- [X] All
- [ ] Windows Server 2022
- [ ] Windows 11, version 22H2
- [ ] Windows 11, version 21H2
- [ ] Windows Insider Preview (specify affected build below)
- [ ] Ubuntu
- [ ] Debian
- [ ] Other (specify below)

### MsQuic version

main

### Steps taken to reproduce bug

See description.

### Expected behavior

See description.

### Actual outcome

See description.

PR #2870, opened 2022-07-06
Indicate Stream Start Abort Event from Send Start

nibanks:
## Description

For #2869. Updates the send abort logic to indicate a start abort if the send has a start flag.

## Testing

Added new test.

## Documentation

N/A

### DIFF
--- a/src/core/stream_send.c
+++ b/src/core/stream_send.c
@@ -378,6 +378,10 @@ QuicStreamCompleteSendRequest(
             !(Stream->SendBufferBookmark->Flags & QUIC_SEND_FLAG_BUFFERED));
     }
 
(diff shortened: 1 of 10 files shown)


src/core/stream_send.c
```diff
@@ -378,6 +378,10 @@ QuicStreamCompleteSendRequest(
             !(Stream->SendBufferBookmark->Flags & QUIC_SEND_FLAG_BUFFERED));
     }
 
+    if (SendRequest->Flags & QUIC_SEND_FLAG_START && !Stream->Flags.Started) {
```
nibanks, line 381: We probably need to de-dup multiple send calls with the `QUIC_SEND_FLAG_START` flag.

PR #2872, opened 2022-07-06
Include connection close details in stream's SHUTDOWN_COMPLETE

ManickaP:
## Description

This PR is **just a proposal of a change** that would eventually help us get rid of an explicit link from stream to connection. We'd like to adopt `SafeHandle` reference counting to keep connection alive for all the streams instead of a manual reference counting.

The event contains information whether it's happening due to connection closure, but it doesn't contain any details. As a result, the user has to manually keep reference between stream and connection if they want to use those details. This is the last thing that forces us to keep reference from stream to connection.

## Testing

Local test with System.Net.Quic and closing the connection via application and via idle timeout while stream is opened.

## Documentation

- ~~Yes. I can fill it in if this proposal gets a green light.~~
- I haven't found any reference to event details in the docs, but I've updated the log message.

### DIFF
--- a/src/core/stream.c
+++ b/src/core/stream.c
@@ -470,11 +470,17 @@ QuicStreamIndicateShutdownComplete(
             Stream->Connection->State.ClosedRemotely;
         Event.SHUTDOWN_COMPLETE.AppCloseInProgress =
             Stream->Flags.HandleClosed;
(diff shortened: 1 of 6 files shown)


nibanks:
I'm fine with this change.

nibanks:
> - I haven't found any reference to event details in the docs, but I've updated the log message.

Yeah, we still haven't added the stream event documentation.

PR #2873, opened 2022-07-07
Stream level stats API

csujedihy:
## Description

A stream level GetParam API to stats (mainly for flow blocked timings).

Flow blocked timings include both stream level and parent connection level flow blocked timings.

## Testing

CI


src/core/stream.c
```diff
@@ -713,7 +725,79 @@ QuicStreamParamGet(
 
         Status = QUIC_STATUS_SUCCESS;
         break;
+    case QUIC_PARAM_STREAM_BLOCKED_TIMINGS:
```
nibanks, line 728: nit: add a new line before this, and just go ahead and wrap this case with `{ }` because you'd declaring variables below.

src/core/stream.c
```diff
@@ -713,7 +725,79 @@ QuicStreamParamGet(
...
+            Status = QUIC_STATUS_BUFFER_TOO_SMALL;
+            break;
+        }
+
+        if (Buffer == NULL) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
 
+        uint64_t Now = CxPlatTimeUs64();
```
nibanks, line 740: I would really like if we could turn the code below into a loop somehow.

src/core/stream.c
```diff
@@ -295,11 +296,21 @@ QuicStreamStart(
             QuicConnIsServer(Stream->Connection),
             &Stream->Connection->PeerTransportParams);
     if (Stream->MaxAllowedSendOffset == 0) {
-        Stream->OutFlowBlockedReasons |= QUIC_FLOW_BLOCKED_STREAM_FLOW_CONTROL;
+        OutFlowBlockedReasons |= QUIC_FLOW_BLOCKED_STREAM_FLOW_CONTROL;
     }
     Stream->SendWindow = (uint32_t)CXPLAT_MIN(Stream->MaxAllowedSendOffset, UINT32_MAX);
 
-    if (Stream->OutFlowBlockedReasons != 0) {
+    if (OutFlowBlockedReasons != 0) {
```
nibanks, line 303: We should refactor all this to use `QuicStreamAddOutFlowBlockedReason` instead.

src/core/connection.h
```diff
@@ -853,6 +864,23 @@ QuicConnAddOutFlowBlockedReason(
     )
 {
     if (!(Connection->OutFlowBlockedReasons & Reason)) {
+        uint64_t Now = CxPlatTimeUs64();
+        if (Reason & QUIC_FLOW_BLOCKED_PACING) {
```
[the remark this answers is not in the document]
csujedihy: I don't get what you meant. Are you saying remove `if (Reason & QUIC_FLOW_BLOCKED_PACING)`? That way, we'd set start time for pacing blocked timing no matter what the reason is.

The check above checks whether Reason overlaps with OutFlowBlockedReasons. If so, we skip the `if` body. I think this check is simply wrong as you pointed out, if one does and one doesn't, we will skip the `if` body.

src/core/stream.c
csujedihy: It would be possible but I'd have to add a few more enums and make the tracking struct an array on stream and connection. Do you think it's worth it?

src/core/connection.h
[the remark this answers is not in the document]
nibanks: What happens if `Reason` has 2 flags? One that is already present and one that isn't? `if (!(Connection->OutFlowBlockedReasons & Reason))` will succeed so long as there is one flag not present. But then we could overwrite the start time of whatever was already present.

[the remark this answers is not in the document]
csujedihy: > One that is already present and one that isn't? `if (!(Connection->OutFlowBlockedReasons & Reason))` will succeed

In that case, `if (!(Connection->OutFlowBlockedReasons & Reason))` will not succeed.

`!A` => 0 if A is a positive number.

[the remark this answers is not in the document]
csujedihy: > But then we could overwrite the start time of whatever was already present.

I think the check should be like this:
```
        if (!(Connection->OutFlowBlockedReasons & QUIC_FLOW_BLOCKED_PACING) && Reason & QUIC_FLOW_BLOCKED_PACING) {
            Connection->BlockedTimings.Pacing.StartTime = Now;
        }
```

src/core/stream.c
csujedihy: I now have a strong feeling that this will make the code look way more unreadable and non-trivial after I just tried making the enum incremental and making QuicConnAddOutFlowBlockedReason adopt it.

So, I am gonna keep it this way unless you strongly insist.

src/core/connection.h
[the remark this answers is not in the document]
csujedihy: In the latest iteration, I made this function only takes one reason and added an assert.

src/core/send.h
```diff
@@ -302,6 +302,14 @@ typedef struct QUIC_SEND {
 
 } QUIC_SEND;
 
+//
+// This structure defines flow blocked timing for a stream or connection.
+//
+typedef struct QUIC_FLOW_BLOCKED_TIMING_TRACKER {
+    uint64_t CumulativeTime;
+    uint64_t StartTime;
```
maolson-msft, line 310: Should this be "LastStartTime"?

csujedihy: As you wish!

src/core/connection.h
```diff
@@ -852,7 +863,27 @@ QuicConnAddOutFlowBlockedReason(
     _In_ QUIC_FLOW_BLOCK_REASON Reason
     )
 {
+    CXPLAT_DBG_ASSERTMSG(
```
nibanks, line 866: Thanks for adding this. Makes me wonder if we could do some SAL magic to make this evaluate at build time...

src/core/stream.c
```diff
@@ -575,6 +567,7 @@ QuicStreamParamSet(
     case QUIC_PARAM_STREAM_ID:
     case QUIC_PARAM_STREAM_0RTT_LENGTH:
     case QUIC_PARAM_STREAM_IDEAL_SEND_BUFFER_SIZE:
+    case QUIC_PARAM_STREAM_BLOCKED_TIMINGS:
         Status = QUIC_STATUS_INVALID_PARAMETER;
         break;
```
nibanks, line 572: No need to make this change, but there's no point in having all these when the `default` already does this...

src/core/stream.c
```diff
@@ -713,7 +706,81 @@ QuicStreamParamGet(
...
+            Timing->Time +=
+                CxPlatTimeDiff64(Connection->BlockedTimings.CongestionControl.LastStartTime, Now);
+        }
+        ++Timing;
+        Timing->Reason = QUIC_FLOW_BLOCKED_CONN_FLOW_CONTROL;
+        Timing->Time = Connection->BlockedTimings.FlowControl.CumulativeTime;
+        if (Connection->BlockedTimings.FlowControl.LastStartTime != 0) {
+            Timing->Time +=
+                CxPlatTimeDiff64(Connection->BlockedTimings.FlowControl.LastStartTime, Now);
+        }
```
nibanks, line 779: We need to improve this. Right now this will return total timings for the lifetime of the entire connection. I _think_ the solution here would be to cache total cumulative time on the connection, when the stream is started, and then subtract that time from here when returned to the app.

src/inc/msquic.h
```diff
@@ -817,6 +835,7 @@ typedef struct QUIC_SCHANNEL_CONTEXT_ATTRIBUTE_EX_W {
 #define QUIC_PARAM_STREAM_0RTT_LENGTH                   0x08000001  // uint64_t
 #define QUIC_PARAM_STREAM_IDEAL_SEND_BUFFER_SIZE        0x08000002  // uint64_t - bytes
 #define QUIC_PARAM_STREAM_PRIORITY                      0x08000003  // uint16_t - 0 (low) to 0xFFFF (high) - 0x7FFF (default)
+#define QUIC_PARAM_STREAM_BLOCKED_TIMINGS               0X08000004  // uint64_t[]
```
nibanks, line 838:
```suggestion
#define QUIC_PARAM_STREAM_BLOCKED_TIMINGS               0X08000004  // QUIC_FLOW_BLOCKED_TIMING[QUIC_FLOW_BLOCK_REASON_COUNT]
```
It should be this, right? Should we define a `QUIC_FLOW_BLOCK_REASON_SET` (or something) that is equivalent?

nibanks: If the enum was incremental, and this set was ordered, then we could use `uint64_t[]`. BTW, Just from looking at this header, I have zero clue what the units are for `Time` above is.

src/perf/lib/ThroughputClient.cpp
```diff
@@ -579,9 +581,19 @@ ThroughputClient::StreamCallback(
         }
         MsQuic->StreamShutdown(StreamHandle, QUIC_STREAM_SHUTDOWN_FLAG_ABORT, 0);
         break;
-    case QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE:
+    case QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE: {
```
nibanks, line 584: Strictly speaking, you shouldn't need to add wrapping `{ ... }` here because everything is in the `if (PrintBlockedTimings) {` block.

src/core/stream.c
maolson-msft: That's how perf counters usually work though, isn't it? They are absolute values, and the thread that is querying and using them subtracts the values from two queries. One advantage of this idempotent design is that multiple consumers of the perf counters can operate independently. That is less of an issue for this API where only the app owning the QUIC sockets can access the counters, but

1) that could feasibly change, and
2) it seems best to stick to the Standard Model

This is all assuming we haven't already added non-absolute perf counters somewhere in QUIC!

maolson-msft: Oh wait, I think I misunderstood you. You're saying the connection-wide timings really ought to be based on the time during which the stream existed, right? If so then yeah, I agree that's not right.

csujedihy: Yeah, it's a problem.

src/core/stream.h
```diff
@@ -719,7 +727,19 @@ QuicStreamAddOutFlowBlockedReason(
     _In_ QUIC_FLOW_BLOCK_REASON Reason
     )
 {
+    CXPLAT_DBG_ASSERTMSG(
+        (Reason & (Reason - 1)) == 0,
+        "More than one reason is not allowed");
     if (!(Stream->OutFlowBlockedReasons & Reason)) {
+        uint64_t Now = CxPlatTimeUs64();
+        if (Reason & QUIC_FLOW_BLOCKED_STREAM_FLOW_CONTROL) {
```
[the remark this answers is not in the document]
csujedihy: Not really missing. We set StreamIdFlowControl outside of this function because when we set it, the stream created event has not been called.

src/core/stream.c
csujedihy: Weird. All these parameters should not have been added here then.

src/core/stream.c
nibanks: Yes, we need the connection's timings for only the lifetime of the stream.

src/core/stream.h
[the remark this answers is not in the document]
nibanks: Ok.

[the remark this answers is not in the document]
nibanks: Maybe worth a debug assert then, just to make sure no one tries to change things to call this for stream id fc.

src/core/stream.c
csujedihy: Done

src/core/stream.h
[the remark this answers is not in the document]
csujedihy: Done

src/core/stream.h
```diff
@@ -417,6 +417,19 @@ typedef struct QUIC_STREAM {
...
     QUIC_OPERATION* ReceiveCompleteOperation;
 
+    //
+    // Stream blocked timings.
+    //
+    struct {
+        QUIC_FLOW_BLOCKED_TIMING_TRACKER StreamIdFlowControl;
+        QUIC_FLOW_BLOCKED_TIMING_TRACKER FlowControl;
+        QUIC_FLOW_BLOCKED_TIMING_TRACKER App;
+        uint64_t CachedConnSchedulingUs;
```
nibanks, line 427: Did you mean to add Us to the end of all these?

csujedihy: Yeah, oops, I forgot the rest of them.

csujedihy: Done

src/core/stream.c
```diff
@@ -713,7 +731,86 @@ QuicStreamParamGet(
 
         Status = QUIC_STATUS_SUCCESS;
         break;
+    case QUIC_PARAM_STREAM_BLOCKED_TIMINGS: {
```
nibanks, line 734: nit: add a newline before this please.

src/core/stream.c
```diff
@@ -713,7 +731,86 @@ QuicStreamParamGet(
...
+            Status = QUIC_STATUS_BUFFER_TOO_SMALL;
+            break;
+        }
+
+        if (Buffer == NULL) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
 
+        uint64_t Now = CxPlatTimeUs64();
```
nibanks, line 747: We should probably return `QUIC_STATUS_INVALID_STATE` for this if not `Started`.

src/inc/msquic.h
```diff
@@ -817,6 +835,7 @@ typedef struct QUIC_SCHANNEL_CONTEXT_ATTRIBUTE_EX_W {
 #define QUIC_PARAM_STREAM_0RTT_LENGTH                   0x08000001  // uint64_t
 #define QUIC_PARAM_STREAM_IDEAL_SEND_BUFFER_SIZE        0x08000002  // uint64_t - bytes
 #define QUIC_PARAM_STREAM_PRIORITY                      0x08000003  // uint16_t - 0 (low) to 0xFFFF (high) - 0x7FFF (default)
+#define QUIC_PARAM_STREAM_BLOCKED_TIMINGS               0X08000004  // QUIC_FLOW_BLOCKED_TIMING[QUIC_FLOW_BLOCK_REASON_COUNT]
```
nibanks, line 838: Let's go ahead and make a `QUIC_STREAM_STATISTICS` struct and put the flow blocked timings in there, and have this return the stats.

csujedihy: QUIC_STREAM_STATISTICS will only have blocked timings? Should I add other stats now in this PR?

csujedihy: Done

src/perf/lib/ThroughputClient.cpp
```diff
@@ -41,6 +41,7 @@ PrintHelp(
         "  -iosize:<####>               The size of each send request queued. (def:%u)\n"
         "  -tcp:<0/1>                   Indicates TCP/TLS should be used instead of QUIC. (def:0)\n"
         "  -stats:<0/1>                 Indicates connection stats should be printed at the end of the run. (def:0)\n"
+        "  -btime:<0/1>                 Indicates connection blocked timings at the end of the run. (def:0)\n"
```
nibanks, line 44: Change to `sstats` for "stream statistics"?

src/core/stream.c
```diff
@@ -708,12 +726,83 @@ QuicStreamParamGet(
...
+        }
+        Stats->ConnBlockedByCongestionControlUs -= Stream->BlockedTimings.CachedConnCongestionControlUs;
+        Stats->ConnBlockedByFlowControlUs = Connection->BlockedTimings.FlowControl.CumulativeTimeUs;
+        if (Connection->BlockedTimings.FlowControl.LastStartTimeUs != 0) {
+            Stats->ConnBlockedByFlowControlUs +=
+                CxPlatTimeDiff64(Connection->BlockedTimings.FlowControl.LastStartTimeUs, Now);
+        }
+        Stats->ConnBlockedByFlowControlUs -= Stream->BlockedTimings.CachedConnFlowControlUs;
+
+        Status = QUIC_STATUS_SUCCESS;
```
nibanks, line 802: You need to set `*BufferLength = sizeof(QUIC_STREAM_STATISTICS)`

src/inc/msquic.h
```diff
@@ -661,6 +661,28 @@ typedef struct QUIC_TLS_SECRETS {
...
+typedef enum QUIC_FLOW_BLOCK_REASON {
+    QUIC_FLOW_BLOCKED_SCHEDULING            = 0x01,
+    QUIC_FLOW_BLOCKED_PACING                = 0x02,
+    QUIC_FLOW_BLOCKED_AMPLIFICATION_PROT    = 0x04,
+    QUIC_FLOW_BLOCKED_CONGESTION_CONTROL    = 0x08,
+    QUIC_FLOW_BLOCKED_CONN_FLOW_CONTROL     = 0x10,
+    QUIC_FLOW_BLOCKED_STREAM_ID_FLOW_CONTROL= 0x20,
+    QUIC_FLOW_BLOCKED_STREAM_FLOW_CONTROL   = 0x40,
+    QUIC_FLOW_BLOCKED_APP                   = 0x80
+} QUIC_FLOW_BLOCK_REASON;
```
nibanks, line 673: Why does this need to move to msquic.h?

src/test/lib/ApiTest.cpp
```diff
@@ -4640,6 +4640,46 @@ void QuicTestStreamParam()
...
+                QUIC_STATUS_BUFFER_TOO_SMALL,
+                MsQuic->GetParam(
+                    Stream.Handle,
+                    QUIC_PARAM_STREAM_STATISTICS,
+                    &Length,
+                    nullptr));
+            TEST_EQUAL(Length, sizeof(QUIC_STREAM_STATISTICS));
+
+            QUIC_STREAM_STATISTICS Stats = {0};
+            TEST_QUIC_SUCCEEDED(
```
nibanks, line 4675: Does this test pass, since you added the "started" check in the stream code? This stream isn't started, right?

csujedihy: It turned out I added that Started check for stream priority option.

src/inc/msquic.h
csujedihy: It doesn't have to be moved here now because we return `QUIC_STREAM_STATISTICS`. Though, if we move this back, we'd have to add duplicated enums in src/tools/etw/trace.c. Worth it?

nibanks: I'd rather not expose this publicly.

src/test/lib/ApiTest.cpp
csujedihy: Fixed it and the test case as well.

nibanks (approved):

PR #2876, opened 2022-07-07
On connection idle cid changes

### LINKED ISSUE #1800: Change CID after Idleness

We should change the CID after enough idle time so that any possible NAT rebinding would be harder to be tracked by an outside observer. The simplest way to implement this would be in the flush send code path. When first starting a flush, check how much time has elapsed since the last send. If it's greater than X (20? 30?) seconds force a CID update (inline) first.

liveans:
## Description

Closes #1800. The purpose of this pull request is to change the Destination Connection Id on the active connection when the connection is idle for an amount of time(default 20 sec).

## Testing

We have a unit test under HandshakeTests.cpp, named as Misc.IdleDestCidChange

## Documentation

"IdleChangeDestCidMs" added to necessary docs.


src/core/quicdef.h
```diff
@@ -490,6 +490,11 @@ CXPLAT_STATIC_ASSERT(
 //
 #define QUIC_CONGESTION_CONTROL_ALGORITHM_DEFAULT   QUIC_CONGESTION_CONTROL_ALGORITHM_CUBIC
 
+//
+// The default idle time to change source cid before flush
```
nibanks, line 494: How about: "The default idle timeout period after which the source CID is updated before sending again."

src/core/send.c
```diff
@@ -1098,7 +1098,6 @@ QuicSendFlush(
         return TRUE;
     }
 #endif
-
     QuicConnTimerCancel(Connection, QUIC_CONN_TIMER_PACING);
```
nibanks, line 1101: nit: Don't delete the newline above.

src/core/send.c
```diff
@@ -1120,6 +1119,17 @@ QuicSendFlush(
         return TRUE;
     }
 
+    //
+    // Connection CID changes on idle state after an amount of time
+    //
+    uint64_t TimeNow = CxPlatTimeUs64();
```
nibanks, line 1125: Move this up, and use it's value for like 1109, where it calls `QuicMtuDiscoveryCheckSearchCompleteTimeout`.

src/core/send.c
```diff
@@ -1120,6 +1119,17 @@ QuicSendFlush(
...
+    if (Connection->Settings.IdleSrcCidChangeMs != 0 &&
+        Send->LastFlushTimeValid &&
+        CxPlatTimeDiff64(Send->LastFlushTime, TimeNow) >=
+        MS_TO_US(Connection->Settings.IdleSrcCidChangeMs)) {
```
nibanks, line 1129:
```suggestion
        CxPlatTimeDiff64(Send->LastFlushTime, TimeNow) >= MS_TO_US(Connection->Settings.IdleSrcCidChangeMs)) {
```

src/core/settings.h
```diff
@@ -89,6 +90,7 @@ typedef struct QUIC_SETTINGS_INTERNAL {
     uint16_t MaxBindingStatelessOperations;
     uint16_t StatelessOperationExpirationMs;
     uint16_t CongestionControlAlgorithm;
+    uint64_t IdleSrcCidChangeMs;
```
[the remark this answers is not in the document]
liveans: Agreed, I feel like 32-bits uint is much better for this instead of converting to 16-bits with seconds.

src/core/send.c
```diff
@@ -1120,6 +1121,15 @@ QuicSendFlush(
         return TRUE;
     }
 
+    //
+    // Connection CID changes on idle state after an amount of time
+    //
+    if (Connection->Settings.IdleSrcCidChangeMs != 0 &&
+        Send->LastFlushTimeValid &&
+        CxPlatTimeDiff64(Send->LastFlushTime, TimeNow) >= MS_TO_US(Connection->Settings.IdleSrcCidChangeMs)) {
+            QuicConnGenerateNewSourceCids(Connection, TRUE);
```
nibanks, line 1130:
```suggestion
        QuicConnGenerateNewSourceCids(Connection, TRUE);
```

src/core/settings.c
```diff
@@ -1282,6 +1310,7 @@ QuicSettingsSettingsToInternal(
     SETTING_COPY_TO_INTERNAL(MigrationEnabled, Settings, InternalSettings);
     SETTING_COPY_TO_INTERNAL(DatagramReceiveEnabled, Settings, InternalSettings);
     SETTING_COPY_TO_INTERNAL(ServerResumptionLevel, Settings, InternalSettings);
+    SETTING_COPY_TO_INTERNAL(IdleSrcCidChangeMs, Settings, InternalSettings);
```
nibanks, line 1313: Please see the comment right below this. This needs to be fixed.

src/core/settings.c
```diff
@@ -1367,6 +1396,8 @@ QuicSettingsGetSettings(
     SETTING_COPY_FROM_INTERNAL(MigrationEnabled, Settings, InternalSettings);
     SETTING_COPY_FROM_INTERNAL(DatagramReceiveEnabled, Settings, InternalSettings);
     SETTING_COPY_FROM_INTERNAL(ServerResumptionLevel, Settings, InternalSettings);
+    SETTING_COPY_FROM_INTERNAL(IdleSrcCidChangeMs, Settings, InternalSettings);
```
nibanks, line 1399: Same here. Please see comment right below.

src/core/settings.h
```diff
@@ -89,6 +90,7 @@ typedef struct QUIC_SETTINGS_INTERNAL {
     uint16_t MaxBindingStatelessOperations;
     uint16_t StatelessOperationExpirationMs;
     uint16_t CongestionControlAlgorithm;
+    uint32_t IdleSrcCidChangeMs;
```
nibanks, line 93: For this internal struct, put this new `uint32_t` after all the other `uint32_t`s.

src/core/settings.c
liveans: I think you meant to use size check version of this macro instead of SETTING_COPY_TO_INTERNAL one

src/core/settings.c
```diff
@@ -1368,6 +1403,7 @@ QuicSettingsGetSettings(
     SETTING_COPY_FROM_INTERNAL(DatagramReceiveEnabled, Settings, InternalSettings);
     SETTING_COPY_FROM_INTERNAL(ServerResumptionLevel, Settings, InternalSettings);
 
+
```
nibanks, line 1406: Nit: remove extra newline

liveans:
There is a failing on a new test, I'll take a look at it.

src/core/connection.h
```diff
@@ -235,6 +235,8 @@ typedef struct QUIC_CONN_STATS {
     //
     uint32_t QuicVersion;
 
+    uint32_t ChangeSrcCidCount;         // Number of times the source CID changed.
```
nibanks, line 238: Put this down in `Misc` please

src/inc/msquic.h
```diff
@@ -399,6 +399,7 @@ typedef struct QUIC_STATISTICS {
     uint32_t Rtt;                       // In microseconds
     uint32_t MinRtt;                    // In microseconds
     uint32_t MaxRtt;                    // In microseconds
+    uint32_t ChangeSrcCidCount;         // Number of times the source CID changed.
```
nibanks, line 402: Don't change `QUIC_STATISTICS`.

liveans: Okay, then should I get statistics via v2?

src/core/connection.c
```diff
@@ -6782,6 +6783,7 @@ QuicConnParamGet(
         Stats->Rtt = Path->SmoothedRtt;
         Stats->MinRtt = Path->MinRtt;
         Stats->MaxRtt = Path->MaxRtt;
+        Stats->ChangeSrcCidCount = Connection->Stats.ChangeSrcCidCount;
```
nibanks, line 6786: Not here. Update `QuicConnGetV2Statistics` instead.

src/inc/msquic.h
nibanks: Just v2. v1 is done.

src/core/connection.c
```diff
@@ -6612,6 +6612,7 @@ QuicConnGetV2Statistics(
     Stats->RecvDecryptionFailures = Connection->Stats.Recv.DecryptionFailures;
     Stats->RecvValidAckFrames = Connection->Stats.Recv.ValidAckFrames;
     Stats->KeyUpdateCount = Connection->Stats.Misc.KeyUpdateCount;
+    Stats->ChangeSrcCidCount = Connection->Stats.Misc.ChangeSrcCidCount;
```
nibanks, line 6615: Look at the comment below about needing to be size checked. Please move this down after `SendCongestionWindow`.

src/core/connection.h
```diff
@@ -284,6 +284,7 @@ typedef struct QUIC_CONN_STATS {
 
     struct {
         uint32_t KeyUpdateCount;        // Count of key updates completed.
+        uint32_t ChangeSrcCidCount;     // Number of times the source CID changed.
```
nibanks, line 287: Maybe `SrcCidUpdateCount` instead?

src/inc/msquic.h
```diff
@@ -482,6 +482,8 @@ typedef struct QUIC_STATISTICS_V2 {
 
     uint32_t SendCongestionWindow;          // Congestion window size
 
+    uint32_t ChangeSrcCidCount;             // Number of times the source CID changed.
```
nibanks, line 485: After thinking about it, I like `SrcCidUpdate*` instead of `ChangeSrcCid*`. I'd appreciate it if you would update. Thanks!

src/test/lib/DataTest.cpp
```diff
@@ -2685,3 +2685,105 @@ QuicTestStreamAbortConnFlowControl(
...
+        {
+            UniquePtr<TestConnection> Server;
+            ServerAcceptContext ServerAcceptCtx((TestConnection**)&Server);
+            Listener.Context = &ServerAcceptCtx;
+
+            {
+                TestConnection Client(Registration);
+                TEST_TRUE(Client.IsValid());
+
+                TEST_QUIC_SUCCEEDED(Client.SetShareUdpBinding(true));
```
nibanks, line 2727: Shouldn't be necessary.

src/test/lib/DataTest.cpp
```diff
@@ -2685,3 +2685,105 @@ QuicTestStreamAbortConnFlowControl(
...
+
+                Client.SetExpectedTransportCloseStatus(QUIC_STATUS_SUCCESS);
```
nibanks, line 2729: I believe this is the default. So it should be unnecessary.

src/test/lib/DataTest.cpp
```diff
@@ -2685,3 +2685,105 @@ QuicTestStreamAbortConnFlowControl(
...
+                        ServerLocalAddr.GetPort()));
+
+                if (!Client.WaitForConnectionComplete()) {
+                    return;
+                }
+                TEST_TRUE(Client.GetIsConnected());
+
+                TEST_NOT_EQUAL(nullptr, Server);
+
+                Server->SetExpectedTransportCloseStatus(QUIC_STATUS_SUCCESS);
```
nibanks, line 2746: Unnecessary I think

src/core/settings.c
```diff
@@ -1384,6 +1419,13 @@ QuicSettingsGetSettings(
 
     *SettingsLength = CXPLAT_MIN(*SettingsLength, sizeof(QUIC_SETTINGS));
 
+    SETTING_COPY_FROM_INTERNAL_SIZED(
```
nibanks, line 1422: Can you put this before setting `SettingsLength` please?

src/core/settings.c
```diff
@@ -1123,6 +1147,10 @@ QuicSettingsDumpNew(
     if (Settings->IsSet.CongestionControlAlgorithm) {
         QuicTraceLogVerbose(SettingCongestionControlAlgorithm,      "[sett] CongestionControlAlgorithm = %hu", Settings->CongestionControlAlgorithm);
     }
+
+    if (Settings->IsSet.IdleSrcCidChangeMs) {
```
nibanks, line 1151: nit: can you remove the newline before this (and the extra ones above too)?

src/test/lib/DataTest.cpp
liveans: I've kept it as we discussed on Discord

src/inc/msquic.h
liveans: Updated!

src/core/send.c
```diff
@@ -1120,6 +1121,16 @@ QuicSendFlush(
...
     }
 
+    //
+    // Connection CID changes on idle state after an amount of time
+    //
+    if (Connection->Settings.IdleSrcCidChangeMs != 0 &&
+        Send->LastFlushTimeValid &&
+        CxPlatTimeDiff64(Send->LastFlushTime, TimeNow) >= MS_TO_US(Connection->Settings.IdleSrcCidChangeMs)) {
+        QuicConnGenerateNewSourceCids(Connection, TRUE);
+        Connection->Stats.Misc.SrcCidUpdateCount++;
```
nibanks, line 1131: This is still incrementing even if we don't use source CIDs (i.e. `!Connection->State.ShareBinding`). I wonder if we should put `Connection->Stats.Misc.SrcCidUpdateCount++` **in** `QuicConnGenerateNewSourceCids`.

src/inc/msquic.h
```diff
@@ -598,7 +600,8 @@ typedef struct QUIC_SETTINGS {
             uint64_t ServerResumptionLevel                  : 1;
             uint64_t MaxOperationsPerDrain                  : 1;
             uint64_t MtuDiscoveryMissingProbeCount          : 1;
-            uint64_t RESERVED                               : 33;
+            uint64_t IdleSrcCidChangeMs                     : 1;
```
nibanks, line 603: Following the pattern set by other times outs (such as `SendIdleTimeoutMs`) I think we should rename to `SrcCidUpdateIdleTimeoutMs`. Thoughts?

src/test/lib/TestConnection.cpp
```diff
@@ -898,3 +915,20 @@ TestConnection::HandleConnectionEvent(
 
     return QUIC_STATUS_SUCCESS;
 }
+
+uint32_t
+TestConnection::GetSrcCidChangeCountStatistic() {
```
nibanks, line 920: Still have a number of `SrcCidChange*` references that need to be updated.

src/core/send.c
liveans: We can do that, I think. Another thing here is we don't even need to check time if we're not using source CIDs, so I can simply add a condition to if `Connection->State.ShareBinding` like this

src/inc/msquic.h
liveans: I felt in the same way, renamed

src/core/send.c
liveans: Oh, on the second thought, I don't think we can do that cause we're using that function for initial generations as well. Reverting and only adding `Connection->State.ShareBinding` to if condition

liveans: I'll resolve this comment when we agreed on the same thought.

src/core/settings.h
```diff
@@ -49,7 +49,8 @@ typedef struct QUIC_SETTINGS_INTERNAL {
             uint64_t MaxBindingStatelessOperations          : 1;
             uint64_t StatelessOperationExpirationMs         : 1;
             uint64_t CongestionControlAlgorithm             : 1;
-            uint64_t RESERVED                               : 29;
+            uint64_t SrcCidUpdateIdleTimeoutMs                     : 1;
```
nibanks, line 52: nit: please align the `: 1;` with the rest.

src/inc/msquic.h
```diff
@@ -598,7 +600,8 @@ typedef struct QUIC_SETTINGS {
             uint64_t ServerResumptionLevel                  : 1;
             uint64_t MaxOperationsPerDrain                  : 1;
             uint64_t MtuDiscoveryMissingProbeCount          : 1;
-            uint64_t RESERVED                               : 33;
+            uint64_t SrcCidUpdateIdleTimeoutMs                     : 1;
```
nibanks, line 603: Please fix alignment.

src/test/lib/TestConnection.cpp
```diff
@@ -898,3 +915,20 @@ TestConnection::HandleConnectionEvent(
 
     return QUIC_STATUS_SUCCESS;
 }
+
+uint32_t
+TestConnection::GetSrcCidUpdateCountStatistic() {
```
nibanks, line 920: Let's remove `Statistic` from the name.

src/core/settings.c
```diff
@@ -951,6 +962,18 @@ QuicSettingsLoad(
             Settings->CongestionControlAlgorithm = (QUIC_CONGESTION_CONTROL_ALGORITHM)Value;
         }
     }
+    if (!Settings->IsSet.SrcCidUpdateIdleTimeoutMs) {
+        Value = QUIC_DEFAULT_IDLE_SRC_CID_CHANGE_MS;
+        ValueLen = sizeof(Value);
+        CxPlatStorageReadValue(
+            Storage,
+            QUIC_SETTING_IDLE_SRC_CID_CHANGE_MS,
```
nibanks, line 970: Need to update this too

src/core/settings.c
```diff
@@ -951,6 +962,18 @@ QuicSettingsLoad(
             Settings->CongestionControlAlgorithm = (QUIC_CONGESTION_CONTROL_ALGORITHM)Value;
         }
     }
+    if (!Settings->IsSet.SrcCidUpdateIdleTimeoutMs) {
+        Value = QUIC_DEFAULT_IDLE_SRC_CID_CHANGE_MS;
```
nibanks, line 966: This need to be updated as well.

src/core/send.c
nibanks: Let's move it to `QuicConnGenerateNewSourceCids` but only increment the counter when `ReplaceExistingCids == TRUE`, so it won't be triggered by the initial generation.

src/core/connection.c
```diff
@@ -6633,6 +6633,9 @@ QuicConnGetV2Statistics(
     if (STATISTICS_HAS_FIELD(*StatsLength, SendCongestionWindow)) {
         Stats->SendCongestionWindow = QuicCongestionControlGetCongestionWindow(&Connection->CongestionControl);
     }
+    if(STATISTICS_HAS_FIELD(*StatsLength, SrcCidUpdateCount)) {
```
nibanks, line 6636:
```suggestion
    if (STATISTICS_HAS_FIELD(*StatsLength, SrcCidUpdateCount)) {
```

src/core/send.c
liveans: Done

liveans:
As we discussed on Discord, I'm going to convert this feature to update destination cids instead of source cids.

src/core/settings.c
```diff
@@ -951,6 +962,18 @@ QuicSettingsLoad(
...
     }
+    if (!Settings->IsSet.DestCidUpdateIdleTimeoutMs) {
+        Value = QUIC_DEFAULT_DEST_CID_UPDATE_IDLE_TIMEOUT_MS;
+        ValueLen = sizeof(Value);
+        CxPlatStorageReadValue(
+            Storage,
+            QUIC_SETTING_DEST_CID_UPDATE_IDLE_TIMEOUT_MS,
+            (uint8_t*)&Value,
+            &ValueLen);
+        if (Value < UINT32_MAX) {
```
nibanks, line 973:
```suggestion
        if (Value <= UINT32_MAX) {
```
I think this is more correct.

src/core/send.c
```diff
@@ -1128,7 +1128,13 @@ QuicSendFlush(
         Send->LastFlushTimeValid &&
         CxPlatTimeDiff64(Send->LastFlushTime, TimeNow) >= MS_TO_US(Connection->Settings.DestCidUpdateIdleTimeoutMs)) {
         Connection->Stats.Misc.DestCidUpdateCount++;
-        // TODO: New implementation of destination CID update.
+        if (!QuicConnRetireCurrentDestCid(Connection, Path)) {
+            QuicTraceEvent(
+                ConnCantGetNewDestCid,
+                "[conn][%p] Connection can't get new dest CID.",
+                Connection
+            );
```
nibanks, line 1136:
```suggestion
                Connection);
```

src/manifest/MsQuicEtw.man
```diff
@@ -3356,6 +3362,15 @@
...
               />
+          <event
+              keywords="ut:Connection ut:DataFlow"
+              level="win:Verbose"
+              message="$(string.Etw.ConnCantGetNewDestCid)"
+              opcode="Connection"
+              symbol="QuicConnCantGetNewDestCid"
+              template="tid_CONN_CANT_GET_NEW_DEST_CID"
+              value="11270"
+              />
```
nibanks, line 3373: Please but this with (at the end of) the other connection events... But actually, I'm not even 100% sure this extra log is necessary in general.

liveans: I deleted that log

src/core/send.c
```diff
@@ -1120,6 +1121,16 @@ QuicSendFlush(
         return TRUE;
     }
 
+    //
+    // Connection CID changes on idle state after an amount of time
+    //
+    if (Connection->Settings.DestCidUpdateIdleTimeoutMs != 0 &&
+        Send->LastFlushTimeValid &&
+        CxPlatTimeDiff64(Send->LastFlushTime, TimeNow) >= MS_TO_US(Connection->Settings.DestCidUpdateIdleTimeoutMs)) {
+        Connection->Stats.Misc.DestCidUpdateCount++;
```
nibanks, line 1130: I know it might complicate your tests a tad, but I do think putting this in the function below would be a good idea.

src/core/send.c
```diff
@@ -1120,6 +1121,16 @@ QuicSendFlush(
...
+        QuicConnRetireCurrentDestCid(Connection, Path);
```
nibanks, line 1131:
```suggestion
        (void)QuicConnRetireCurrentDestCid(Connection, Path);
```
We use this pattern to explicitly show we are ignoring the return value.

src/manifest/clog.sidecar
```diff
@@ -15003,4 +15020,4 @@
       }
     ]
   }
-}
\ No newline at end of file
+}
```
nibanks, line 15023: Did you add the trailing newline manually? Ideally we should leave this however the auto generation does.

src/core/send.c
liveans: It shouldn't affect the test actually, since in test case it shouldn't retire all destination cids I think. I'm going to do that

src/test/bin/winkernel/control.cpp
```diff
@@ -1259,6 +1259,10 @@ QuicTestCtlEvtIoDeviceControl(
         QuicTestCtlRun(QuicTestVersionStorage());
         break;
 
+    case IOCTL_QUIC_RUN_CONNECT_AND_IDLE_FOR_DEST_CID_CHANGE:
```
nibanks, line 1262: You're missing the `0,` above in the size array now, right? Did you delete that?

liveans: No, I merged the main branch into this branch, there was a change in this file, so I forgot to add it. I'll add it now

src/manifest/clog.sidecar
liveans: Actually, I've generated this with update-sidecar script, this is strange, I'll regenerate this

src/core/send.c
```diff
@@ -1120,6 +1121,17 @@ QuicSendFlush(
...
     }
 
+    //
+    // Connection CID changes on idle state after an amount of time
+    //
+    if (Connection->Settings.DestCidUpdateIdleTimeoutMs != 0 &&
+        Send->LastFlushTimeValid &&
+        CxPlatTimeDiff64(Send->LastFlushTime, TimeNow) >= MS_TO_US(Connection->Settings.DestCidUpdateIdleTimeoutMs)) {
+        if (QuicConnRetireCurrentDestCid(Connection, Path)) {
+            Connection->Stats.Misc.DestCidUpdateCount++;
```
nibanks, line 1131: I mean put this **in** the function.

liveans: Oh, I see :D sorry

docs/api/QUIC_SETTINGS.md
```diff
@@ -79,7 +80,6 @@ typedef struct QUIC_SETTINGS {
     uint8_t RESERVED                        : 2;
     uint8_t MaxOperationsPerDrain;
     uint8_t MtuDiscoveryMissingProbeCount;
-    uint8_t CongestionControlAlgorithm;
```
nibanks, line 82: Weird! We need to figure out which of these should be deleted. cc @maolson-msft

src/lib.rs
```diff
@@ -574,6 +574,8 @@ pub struct QuicStatisticsV2 {
     pub recv_valid_ack_frames: u64,
 
     pub key_update_count: u32,
+    // Number of times the destionation CID changed.
+    pub dest_cid_update_count: u32
```
nibanks, line 578: This is actually missing `SendCongestionWindow` too, above this one.

src/lib.rs
```diff
@@ -574,6 +574,10 @@ pub struct QuicStatisticsV2 {
     pub recv_valid_ack_frames: u64,
 
     pub key_update_count: u32,
+
+    pub send_congestion_window: u32,
+    // Number of times the destionation CID changed.
```
nibanks, line 579:
```suggestion
    // Number of times the destination CID changed.
```

nibanks (approved):

PR #2877, opened 2022-07-09
Update CUBIC to match rfc8312bis alpha_cubic algorithm tweak

maolson-msft:
Updates the code to match this RFC update:
https://github.com/NTAP/rfc8312bis/pull/146/files 

During CA, the AIMD window increases with a 1/2-packet-per-ACK slope, which is
less aggressive than NewReno, then increases the slope to 1-packet-per-ACK once
CWnd hits W_max (WindowMax in the code). The committee determined that when
Fast Convergence kicks in and reduces WindowMax, this behavior becomes more
aggressive than it should be (seeing how Fast Convergence is really a "we should
slow down" indication). The solution is to maintain a second variable, prior_cwnd,
that is identical to W_max except with no Fast Convergence reduction logic, and
let this be the threshold for AIMD window slope change rather than W_max.

### DIFF
--- a/src/core/cubic.c
+++ b/src/core/cubic.c
@@ -246,6 +246,7 @@ CubicCongestionControlOnCongestionEvent(
     //
     // Save previous state, just in case this ends up being spurious.
     //
+    Cubic->PrevWindowPrior = Cubic->WindowPrior;
     Cubic->PrevWindowMax = Cubic->WindowMax;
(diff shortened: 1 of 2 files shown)


csujedihy (approved):

nibanks (approved):

maolson-msft:
@nibanks Are you concerned about any of these failing CI tests?

PR #2883, opened 2022-07-11
Fixed details of stream SHUTDOWN_COMPLETE for ConnectionShutdown

ManickaP:
## Description

Small follow up on #2872. Renamed "ByPeer" to "ByApp" since it applies to local connection close as well. And included info whether it's local or remote close.

## Testing

System.Net.Quic tests.

## Documentation

No, docs do not include event details.

### DIFF
--- a/src/inc/msquic.h
+++ b/src/inc/msquic.h
@@ -1255,8 +1255,9 @@ typedef struct QUIC_STREAM_EVENT {
         struct {
             BOOLEAN ConnectionShutdown;
             BOOLEAN AppCloseInProgress       : 1;
-            BOOLEAN ConnectionShutdownByPeer : 1;
-            BOOLEAN RESERVED                 : 6;
+            BOOLEAN ConnectionShutdownByApp  : 1;
+            BOOLEAN ConnectionClosedRemotely : 1;
+            BOOLEAN RESERVED                 : 5;
             QUIC_UINT62 ConnectionErrorCode;
         } SHUTDOWN_COMPLETE;
         struct {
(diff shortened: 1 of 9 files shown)


src/inc/msquic.h
```diff
@@ -1255,8 +1255,9 @@ typedef struct QUIC_STREAM_EVENT {
         struct {
             BOOLEAN ConnectionShutdown;
             BOOLEAN AppCloseInProgress       : 1;
-            BOOLEAN ConnectionShutdownByPeer : 1;
-            BOOLEAN RESERVED                 : 6;
+            BOOLEAN ConnectionShutdownByApp  : 1;
+            BOOLEAN ConnectionClosedRemotely : 1;
```
nibanks, line 1259: I'm fine with this change, but you definitely will want to make sure you add some tests to verify things operate as you expect. For instance, I'm not 100% positive that `Connection->State.ClosedRemotely` will always give you what you want. That flag doesn't mean the remote closed first, just that the remote side has closed (which can sometimes happen **implicitly**, such as with idle timeouts, I believe).

nibanks: Ok. Feel free to leverage existing tests to add additional logic just to verify these flags.

nibanks (dismissed):

ManickaP:
Looks like failing tests are unrelated, but please check after me, I'm not familiar well with your infra.
Otherwise, this should be good to go now. 
I pushed more test changes since the last review so it'll need another one.

nibanks (approved):

PR #2896, opened 2022-07-17
add MacOS support for prep scripts

ami-GS:
## Description

Add MacOS support

## Testing

No

## Documentation

No

### DIFF
--- a/scripts/prepare-machine.ps1
+++ b/scripts/prepare-machine.ps1
@@ -555,6 +555,8 @@ if ($IsLinux) {
 }
 
 if ($IsMacOS) {
+    brew update
+    brew install cmake
     if ($ForTest) {
         Write-Host "Setting core dump pattern"
         sudo sysctl -w kern.corefile=%N.%P.core
--- a/scripts/prepare-machine.ps1
+++ b/scripts/prepare-machine.ps1
@@ -555,8 +555,6 @@ if ($IsLinux) {
 }
 
 if ($IsMacOS) {
-    brew update
-    brew install cmake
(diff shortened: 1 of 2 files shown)


scripts/prepare-machine.ps1
```diff
@@ -555,6 +555,8 @@ if ($IsLinux) {
 }
 
 if ($IsMacOS) {
+    brew update
+    brew install cmake
```
nibanks, line 559: We don't need this for automation, so I'm hesitant to have this always run. Any downsides?

ThadHouse: Brew isn’t something built into the os, and it’s possible to grab cmake without it. So it should probably only run if brew is actually installed.

nibanks:
@ami-GS what do you plan to do with this?

scripts/prepare-machine.ps1
ami-GS: @nibanks 
I'm trying to use curl here, but how does automation download/install cmake?
I want to know which specific cmake version the automation use. If it is using latest, I want the URL because there is no latest URL as much as I checked.

nibanks: We use PowerShell to download stuff.

ami-GS: Removed the change as this is mentioned in BUILD.md

nibanks:
Mind merging main to get CI to run again?

nibanks:
@ami-GS can we just close/abandon this PR?

PR #2897, opened 2022-07-17
XDP Yield to Scheduler with No Work

nibanks:
## Description

Give up the current thread quantum to the scheduler if we do 10 iterations of the loop with no work being done.

## Testing

Existing perf tests

## Documentation

N/A

### DIFF
--- a/src/platform/platform_worker.c
+++ b/src/platform/platform_worker.c
@@ -217,7 +217,7 @@ CxPlatWakeExecutionContext(
     CxPlatWorkerWake((CXPLAT_WORKER*)Context->CxPlatContext);
 }
 
-void
+BOOLEAN // Did work?
 CxPlatRunExecutionContexts(
     _In_ CXPLAT_WORKER* Worker,
     _Inout_ uint64_t* TimeNow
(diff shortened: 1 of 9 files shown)


src/platform/platform_worker.c
```diff
@@ -300,10 +308,20 @@ CXPLAT_THREAD_CALLBACK(CxPlatWorkerThread, Context)
...
+            if (CxPlatDataPathRunEC(&Worker->DatapathEC, Worker->ThreadId, WaitTime)) {
+                NoWorkCount = 0;
+            }
         } else if (WaitTime != 0) {
             CxPlatEventWaitWithTimeout(Worker->WakeEvent, WaitTime);
+            NoWorkCount = 0;
+        }
+
+        #ifdef _WIN32
+        if (NoWorkCount > 10) {
```
csujedihy, line 320: Why don't we just yield once when the current loop iteration did not do work?

nibanks: The thought was that doing this too often would hurt perf too much.

nibanks:
It looks like this PR unexpectedly has a significant improvement on HPS:
```
Test Run Took 00:00:12.1958044
Run 1: 9754 HPS
Test Run Took 00:00:12.0586717
Run 2: 9748 HPS
Test Run Took 00:00:12.0671319
Run 3: 9758 HPS
Test Run Took 00:00:12.0931618
Run 4: 9771 HPS
Test Run Took 00:00:12.1364194
Run 5: 9748 HPS
Median: 9754 HPS (+21.94%)
Remote: 7998.8 HPS
```

csujedihy (approved):

PR #2907, opened 2022-07-23
Add registry configuration for Version Settings.

anrossi:
## Description

Enable configuration of Version Settings via the Windows Registry.

## Testing

Manual testing with `quicinteropserver` and `quicinterop`.  

- [x] Automated tests should be written to cover the storage system.

## Documentation

- [x] Documentation should be updated to reflect the new Registry value names.

### DIFF
--- a/src/core/settings.c
+++ b/src/core/settings.c
@@ -870,23 +870,23 @@ QuicSettingsLoad(
     if (!Settings->IsSet.VersionSettings) {
         uint32_t OfferedVersionsSize = 0, AcceptableVersionsSize = 0, FullyDeployedVersionsSize = 0;
         if (QUIC_SUCCEEDED(
-                CxPlatStorageReadValue(
-                    Storage,
-                    QUIC_SETTING_OFFERED_VERSIONS,
-                    (uint8_t*)NULL,
-                    &OfferedVersionsSize)) &&
+            CxPlatStorageReadValue(
+                Storage,
+                QUIC_SETTING_OFFERED_VERSIONS,
+                (uint8_t*)NULL,
+                &OfferedVersionsSize)) &&
             QUIC_SUCCEEDED(
-                CxPlatStorageReadValue(
-                    Storage,
-                    QUIC_SETTING_ACCEPTABLE_VERSIONS,
-                    (uint8_t*)NULL,
-                    &AcceptableVersionsSize)) &&
+            CxPlatStorageReadValue(
+                Storage,
+                QUIC_SETTING_ACCEPTABLE_VERSIONS,
+                (uint8_t*)NULL,
+                &AcceptableVersionsSize)) &&
             QUIC_SUCCEEDED(
-                CxPlatStorageReadValue(
-                    Storage,
-                    QUIC_SETTING_FULLY_DEPLOYED_VERSIONS,
-                    (uint8_t*)NULL,
-                    &FullyDeployedVersionsSize)) &&
+            CxPlatStorageReadValue(
+                Storage,
+                QUIC_SETTING_FULLY_DEPLOYED_VERSIONS,
+                (uint8_t*)NULL,
+                &FullyDeployedVersionsSize)) &&
             OfferedVersionsSize && FullyDeployedVersionsSize && AcceptableVersionsSize) {
             QUIC_VERSION_SETTINGS* VersionSettings = NULL;
             size_t AllocSize =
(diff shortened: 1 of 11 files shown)


src/core/settings.c
```diff
@@ -867,6 +867,106 @@ QuicSettingsLoad(
         Settings->VersionNegotiationExtEnabled = !!Value;
     }
 
+    if (!Settings->IsSet.VersionSettings) {
+        uint32_t OfferedVersionsSize = 0, AcceptableVersionsSize = 0, FullyDeployedVersionsSize = 0;
+        if (QUIC_SUCCEEDED(
```
nibanks, line 872: Why not allow for registry to overwrite just one?

anrossi: All three don't have to be updated at the same time, but all three must be present in order for version settings to work. They can all be the same list when an update is not in progress.

anrossi:
@nibanks Currently, this expects the registry format for these to be binary blobs, which is difficult to use since .reg files apparently don't support the REG_BINARY type. It's easy to read into memory, however.
Should we instead use something like REG_MULTISZ or a regular string with comma-separated values (in hex) which is easier for admins, but requires more work from us when reading in (and the risks of string parsing).
I'm open to design suggestions.

src/core/settings.c
```diff
@@ -867,6 +867,106 @@ QuicSettingsLoad(
...
+                CxPlatStorageReadValue(
+                    Storage,
+                    QUIC_SETTING_OFFERED_VERSIONS,
+                    (uint8_t*)NULL,
+                    &OfferedVersionsSize)) &&
```
nibanks, line 877: nit: We follow pattern of aligning `QUIC_SUCCEEDED(` with the function on the next line
```suggestion
        if (QUIC_SUCCEEDED(
            CxPlatStorageReadValue(
                Storage,
                QUIC_SETTING_OFFERED_VERSIONS,
                (uint8_t*)NULL,
                &OfferedVersionsSize)) &&
```
Could you please update this code and below?

nibanks:
> @nibanks Currently, this expects the registry format for these to be binary blobs, which is difficult to use since .reg files apparently don't support the REG_BINARY type. It's easy to read into memory, however. Should we instead use something like REG_MULTISZ or a regular string with comma-separated values (in hex) which is easier for admins, but requires more work from us when reading in (and the risks of string parsing). I'm open to design suggestions.

I thought `REG_BINARY` is supported everywhere. Try exporting one to a .reg file to see if it works.

nibanks:
Looks like `CLOG` failed. Please update the sidecar.

anrossi:
> > @nibanks Currently, this expects the registry format for these to be binary blobs, which is difficult to use since .reg files apparently don't support the REG_BINARY type. It's easy to read into memory, however. Should we instead use something like REG_MULTISZ or a regular string with comma-separated values (in hex) which is easier for admins, but requires more work from us when reading in (and the risks of string parsing). I'm open to design suggestions.
> I thought `REG_BINARY` is supported everywhere. Try exporting one to a .reg file to see if it works.

Ah, I tried editing the .reg file in notepad and it garbled the file encoding causing the error I was seeing; when I used VSCode to edit the .reg file, it was easy to import. I'll keep the REG_BINARY type unless you think there's a better solution.

src/test/lib/ApiTest.cpp
```diff
@@ -5321,3 +5321,371 @@ QuicTestStorage()
...
+    for (uint32_t i = 0; i < ReadSettings->AcceptableVersionsLength; i++) {
+        TEST_EQUAL(CxPlatByteSwapUint32(ReadSettings->AcceptableVersions[i]), VersionList[i]);
+    }
+    for (uint32_t i = 0; i < ReadSettings->OfferedVersionsLength; i++) {
+        TEST_EQUAL(CxPlatByteSwapUint32(ReadSettings->OfferedVersions[i]), VersionList[i]);
+    }
+    for (uint32_t i = 0; i < ReadSettings->FullyDeployedVersionsLength; i++) {
+        TEST_EQUAL(CxPlatByteSwapUint32(ReadSettings->FullyDeployedVersions[i]), VersionList[i]);
+    }
+
```
nibanks, line 5505: nit: remove extra new line

src/test/lib/ApiTest.cpp
```diff
@@ -5321,3 +5321,371 @@ QuicTestStorage()
...
+        NO_ERROR,
+        RegSetKeyValueA(
+            HKEY_LOCAL_MACHINE,
+            MSQUIC_GLOBAL_PARAMETERS_PATH,
+            QUIC_SETTING_FULLY_DEPLOYED_VERSIONS,
+            REG_BINARY,
+            &VersionList,
+            sizeof(VersionList)));
+#else
+    TEST_FAILURE("Storage tests not supported on this platform");
```
nibanks, line 5480: No need to continue doing TEST_FAILURE, since you have one above. Same for others below.

nibanks (approved):

PR #2919, opened 2022-07-29
Rename QUIC_ACK_EVENT/QUIC_LOSS_EVENT fields to match their QUIC_LOSS_DETECTION counterparts

maolson-msft:
These fields in QUIC_ACK_EVENT/QUIC_LOSS_EVENT are simply "what is the current value of
some field in the QUIC_LOSS_DETECTION". So I've renamed them to match exactly.

### DIFF
--- a/src/core/congestion_control.h
+++ b/src/core/congestion_control.h
@@ -12,9 +12,9 @@ typedef struct QUIC_ACK_EVENT {
 
     uint64_t TimeNow; // microsecond
 
-    uint64_t LargestPacketNumberAcked;
+    uint64_t LargestAck;
 
-    uint64_t LargestPacketNumberSent;
+    uint64_t LargestSentPacketNumber;
 
     //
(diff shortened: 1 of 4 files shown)


src/core/congestion_control.h
```diff
@@ -12,9 +12,9 @@ typedef struct QUIC_ACK_EVENT {
 
     uint64_t TimeNow; // microsecond
 
-    uint64_t LargestPacketNumberAcked;
+    uint64_t LargestAck;
```
nibanks, line 15: LargestAckedPacketNumber?

maolson-msft: Yeah this discrepancy occurred to me too, but we have several other names in the loss detection code that are derived from LargestAck (such as LargestAckEncryptLevel, NewLargestAckRetransmittable, NewLargestAckDifferentPath), which is why I opted here for using the original name LargestAck. Either way is kinda bad...

nibanks (approved):

PR #2922, opened 2022-07-30
BBR: remove some unnecessary indirection

maolson-msft:
-Removes several "helper" functions that make it less obvious what basic events trigger what logic.
-No more separate BBR_RTT_STATS struct with its own "constructor"- the fields are moved to QUIC_CONGESTION_CONTROL_BBR
 (except for the unnecessary Expiration field, which has just been removed).
-Some naming consistency fixes in the MinRtt bits.
-A few code comments.

### DIFF
--- a/src/core/bbr.c
+++ b/src/core/bbr.c
@@ -101,24 +101,13 @@ const uint32_t kPacingGain[GAIN_CYCLE_LENGTH] = {
 const uint32_t kProbeRttTimeInUs = 200 * 1000;
 
 //
-// RTT Stats default expiration
+// Time until a MinRtt measurement is expired.
 //
-const uint32_t kRttStatsExpirationInSecond = 10;
+const uint32_t kBbrMinRttExpirationInMicroSecs = 10 * 1000000;
 
 const uint32_t kBbrMaxBandwidthFilterLen = 10;
 
 const uint32_t kBbrMaxAckHeightFilterLen = 10;
-
-_IRQL_requires_max_(DISPATCH_LEVEL)
-void
-BbrBandwidthFilterOnAppLimited(
-    _In_ BBR_BANDWIDTH_FILTER* b,
-    _In_ uint64_t LargestSentPacketNumber
-    )
-{
-    b->AppLimited = TRUE;
-    b->AppLimitedExitTarget = LargestSentPacketNumber;
-}
 
 _IRQL_requires_max_(DISPATCH_LEVEL)
 void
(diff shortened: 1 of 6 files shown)


src/core/bbr.c
```diff
@@ -101,64 +101,14 @@ const uint32_t kPacingGain[GAIN_CYCLE_LENGTH] = {
 const uint32_t kProbeRttTimeInUs = 200 * 1000;
 
 //
-// RTT Stats default expiration
+// Time until a MinRtt measurement is expired.
 //
-const uint32_t kRttStatsExpirationInSecond = 10;
+const uint32_t kBbrMinRttExpirationInMicroSecs = 10 * 1000000;
```
nibanks, line 106:
```suggestion
const uint32_t kBbrMinRttExpirationInMicroSecs = S_TO_US(10);
```

src/core/bbr.h
```diff
@@ -91,6 +68,16 @@ typedef struct QUIC_CONGESTION_CONTROL_BBR {
     // If TRUE, ProbeRttEndTime is valid
     //
     BOOLEAN ProbeRttEndTimeValid : 1;
+
+    //
+    // If TRUE, current RTT sample is expired
+    //
+    BOOLEAN RttSampleExpired: 1;
```
nibanks, line 75: This struct could definitely do with some reordering to fix the huge amount of extra padding in here (perhaps in a follow up PR?).

src/core/loss_detection.c
```diff
@@ -1308,7 +1308,7 @@ QuicLossDetectionProcessAckBlocks(
     uint32_t AckedRetransmittableBytes = 0;
     QUIC_CONNECTION* Connection = QuicLossDetectionGetConnection(LossDetection);
     uint64_t TimeNow = CxPlatTimeUs64();
-    uint32_t SmallestRtt = (uint32_t)(-1);
+    uint32_t MinRtt = (uint32_t)(-1);
```
nibanks, line 1311: We should replace usage of `(uint32_t)(-1)` with `UINT32_MAX` in this function.

src/core/bbr.h
maolson-msft: Yeah let's do this at the end of the cleanup pass in case fields are added/removed

maolson-msft:
@nibanks there's a spinquic crash but it doesn't look related to BBR to me:
ntdll!KiUserExceptionDispatch
msquic!CxPlatSendDataComplete
msquic!CxPlatDataPathRunEC
msquic!CxPlatWorkerThread

maolson-msft:
As discussed, I've moved these commits into the new "put back BBR" PR, so I'll close this one.

PR #2929, opened 2022-08-01
Correctly handle unsupported client cert on Windows Server 2022

rzikm:
## Description

This PR adds check for failure statuses on `CxPlatTlsSetClientCertPolicy` when creating security config. Previously, attemt to require client certificates on Windows Server 2022 successfully created security config (although in debug builds it would hit an assert).

## Testing

Tested manually on Windows Server 2022.

## Documentation

No changes needed.

### DIFF
--- a/src/platform/tls_schannel.c
+++ b/src/platform/tls_schannel.c
@@ -772,6 +772,11 @@ CxPlatTlsSetClientCertPolicy(
             "SetCredentialsAttributesW(SECPKG_ATTR_CLIENT_CERT_POLICY) failed");
     }
 
+    if (SecStatus == SEC_E_UNSUPPORTED_FUNCTION)
+    {
+        return QUIC_STATUS_NOT_SUPPORTED;
+    }
+
     return SecStatusToQuicStatus(SecStatus);
 }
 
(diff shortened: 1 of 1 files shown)


src/platform/tls_schannel.c
```diff
@@ -746,6 +746,11 @@ CxPlatTlsSetClientCertPolicy(
             "SetCredentialsAttributesW(SECPKG_ATTR_CLIENT_CERT_POLICY) failed");
     }
 
+    if (SecStatus == SEC_E_UNSUPPORTED_FUNCTION)
```
rzikm, line 749: Not so about this, `SecStatusToQuicStatus` does not actually map anything, it just casts, so this would end up as an unknown status at the app layer.

src/platform/tls_schannel.c
```diff
@@ -746,6 +746,11 @@ CxPlatTlsSetClientCertPolicy(
...
+    {
+        return QUIC_STATUS_NOT_SUPPORTED;
+    }
```
nibanks, line 752:
```suggestion
    if (SecStatus == SEC_E_UNSUPPORTED_FUNCTION) {
        return QUIC_STATUS_NOT_SUPPORTED;
    }
```

src/platform/tls_schannel.c
```diff
@@ -1376,6 +1380,14 @@ CxPlatTlsSecConfigCreate(
 
     if (CredConfig->Flags & QUIC_CREDENTIAL_FLAG_REQUIRE_CLIENT_AUTHENTICATION) {
         Status = CxPlatTlsSetClientCertPolicy(AchContext->SecConfig);
+        if (QUIC_FAILED(Status)) {
+            QuicTraceEvent(
+                LibraryErrorStatus,
+                "[ lib] ERROR, %u, %s.",
+                Status,
+                "CxPlatTlsSetClientCertPolicy");
```
nibanks, line 1388: The error log is unnecessary, because `CxPlatTlsSetClientCertPolicy` already logs the first point of failure.

nibanks (approved):

PR #2936, opened 2022-08-03
Add new QUIC_STATUS_ALPN_IN_USE Status Code

nibanks:
## Description

Adds a new status code to explicitly indicate another listener is already using a specific ALPN.

## Testing

Existing tests updated.

## Documentation

Maybe?


src/core/listener.c
```diff
@@ -335,14 +335,14 @@ MsQuicListenerStart(
     CxPlatEventReset(Listener->StopEvent);
     CxPlatRefInitialize(&Listener->RefCount);
 
-    if (!QuicBindingRegisterListener(Listener->Binding, Listener)) {
+    Status = QuicBindingRegisterListener(Listener->Binding, Listener);
+    if (QUIC_FAILED(Status)) {
         QuicTraceEvent(
```
genghiskMSFT, line 340: Since QuicBindingRegisterListener now returns a Status and not a Boolean, the trace statement on line 340 can now trace out the Status value rather than just an “ERROR, %s, Register with binding” (see the trace statement above on lines 325-330 for the GetBinding trace)

nibanks: Good call! I will definitely fix it! Thanks!

anrossi (approved):

PR #2945, opened 2022-08-04
add support for arm packages

wfurt:
## Description

This adds platform recognition for arm/arm64 on Linux
## Testing

script can now make arm packages
```
root@1284925c1f1c:~/msquic# ls -al artifacts/packages/linux/arm64_Release_openssl/
total 8816
drwxr-xr-x 2 root root    4096 Aug  4 21:26 .
drwxr-xr-x 6 root root    4096 Aug  4 16:24 ..
-rw-r--r-- 1 root root 4503356 Aug  4 21:26 libmsquic-2.1.0-1.aarch64.rpm
-rw-r--r-- 1 root root 4513350 Aug  4 21:26 libmsquic_2.1.0_arm64.deb
```

## Documentation

no.

### DIFF
--- a/scripts/make-packages.sh
+++ b/scripts/make-packages.sh
@@ -61,6 +61,9 @@ while :; do
 
     lowerI="$(echo $1 | tr "[:upper:]" "[:lower:]")"
     case $lowerI in
+        -a|-arch|--arch)
+            ARCH=$1
+            ;;
         -d|-debug|--debug)
             CONFIG=Debug
             ;;
--- a/scripts/make-packages.sh
+++ b/scripts/make-packages.sh
@@ -31,8 +31,16 @@ if [ "$OS" == 'Linux' ]; then
         ARCH='x64'
         LIBDIR="lib64"
     else
-        ARCH=x86
         LIBDIR="lib"
+        if [ "$ARCH" == "aarch64" ]; then
+            ARCH=arm64
+        else
+            if [ "$ARCH" == "armv7l" ]; then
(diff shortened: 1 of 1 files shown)


nibanks (approved):

nibanks:
Break in CI: /home/vsts/work/1/s/artifacts/bin/linux/x64_Release_openssl/msquictest: error while loading shared libraries: libmsquic.so.2: cannot open shared object file: No such file or directory

scripts/make-packages.sh
```diff
@@ -31,8 +31,16 @@ if [ "$OS" == 'Linux' ]; then
         ARCH='x64'
         LIBDIR="lib64"
     else
-        ARCH=x86
         LIBDIR="lib"
```
nibanks, line 34: `LIBDIR="lib"` is correct for all of these?

wfurt: it seems like - at least on the images I look at e.g. Ubuntu.

wfurt: hmm. we may do `lib/aarch64-linux-gnu'. let me explore.

wfurt: updated platform fix-up

nibanks (approved):

nibanks:
Note, I _think_ you'll want to update our `OneBranch.Official.yml` pipeline to build and package these as well. Currently, it looks like only `x64` is built (in `build-linux.yml`) and then packaged (in `build-linux-packages.yml`).

PR #2959, opened 2022-08-09
Support for changing alpn on listener callback (#2418)

### LINKED ISSUE #2418: Support Callback Negotiation Between Matching ALPNs

When a new connection is accepted on the server side, we might want to do the final application protocol selection based on ClientHello SNI. Similar behavior is already supported via TLS over TCP for H/1.1 and H/2.

Original issue: https://github.com/dotnet/runtime/issues/49423

### Proposed solution

We would specify the complete list of supported ALPN when we start the listener and then we would narrow down the final ALPN when we set the configuration for the particular incoming connection.

### Additional context

_No response_

liveans:
## Description

Closes #2418. This pr aims to support changing alpns on ListenerCallback via the `NewNegotiatedAlpn` field.

## Testing

Alpn.ChangeAlpn test case already added, also extended that test case to check out of bounds and invalid address.

## Documentation

Yes, example usage can be found under `QUIC_LISTENER_EVENT.md`


src/inc/msquic.h
```diff
@@ -964,6 +964,10 @@ typedef struct QUIC_LISTENER_EVENT {
         struct {
             const QUIC_NEW_CONNECTION_INFO* Info;
             HQUIC Connection;
+            _Field_range_(0, QUIC_MAX_ALPN_LENGTH)
+            uint8_t NewNegotiatedAlpnLength;
+            _Field_size_bytes_(NewNegotiatedAlpnLength)
+            const uint8_t* NewNegotiatedAlpn;
```
nibanks, line 970: This must point into the `ClientAlpnList` in the `QUIC_NEW_CONNECTION_INFO`, which is in TLS format, where (I believe) it's 1-byte length prefixed. So then you don't need `NewNegotiatedAlpnLength`.

liveans: So, instead of allocating a new heap area for this, I have to find the given alpn from the list and just point it directly?

liveans: I've changed it in this way, thanks for helping on Discord!

src/core/binding.c
```diff
@@ -549,6 +549,13 @@ QuicBindingAcceptConnection(
     //
     QuicListenerAcceptConnection(Listener, Connection, Info);
 
+    //
+    // Free the ALPN buffer if it's changed by the app with NewNegotiatedAlpn.
+    //
+    if (NegotiatedAlpn != Connection->Crypto.TlsState.NegotiatedAlpn) {
+        CXPLAT_FREE(NegotiatedAlpn, QUIC_POOL_ALPN);
+    }
```
nibanks, line 557: 1. Let's move this logic into `QuicListenerClaimConnection`.
2. Do not free the memory if `Connection->Crypto.TlsState.NegotiatedAlpn == Connection->Crypto.TlsState.SmallAlpnBuffer`

src/core/listener.c
```diff
@@ -643,6 +644,33 @@ QuicListenerClaimConnection(
         return FALSE;
     }
 
+    if (Event.NEW_CONNECTION.NewNegotiatedAlpn) {
+        uint8_t* NegotiatedAlpn = NULL;
+        uint8_t NegotiatedAlpnLength = Event.NEW_CONNECTION.NewNegotiatedAlpn[0];
+        if (NegotiatedAlpnLength < TLS_SMALL_ALPN_BUFFER_SIZE) {
+            NegotiatedAlpn = Connection->Crypto.TlsState.SmallAlpnBuffer;
+            CxPlatZeroMemory(NegotiatedAlpn, TLS_SMALL_ALPN_BUFFER_SIZE);
```
nibanks, line 652: No need to zero it out.

src/core/listener.c
```diff
@@ -643,6 +644,33 @@ QuicListenerClaimConnection(
         return FALSE;
     }
 
+    if (Event.NEW_CONNECTION.NewNegotiatedAlpn) {
```
nibanks, line 647: We should verify that `NewNegotiatedAlpn` actually comes from `ClientAlpnList`. Also verify that `NewNegotiatedAlpn + NewNegotiatedAlpn[0]` does not overflow `ClientAlpnList`.

src/core/binding.c
liveans: Done

src/core/listener.c
liveans: I added an address check to verify if it comes from `ClientAlpnList` via getting it's start and end address, also I added overflow check as well

src/test/lib/TestConnection.h
```diff
@@ -92,6 +92,9 @@ class TestConnection
     uint32_t DatagramsLost;
     uint32_t DatagramsAcknowledged;
 
+    const uint8_t* NegotiatedAlpn{};
+    uint8_t NegotiatedAlpnLength{};
```
nibanks, line 96: Don't use `{}` unless you're setting a non-default value.

src/core/listener.c
```diff
@@ -642,6 +643,45 @@ QuicListenerClaimConnection(
...
     }
+    const uint8_t* NewNegotiatedAlpn = Event.NEW_CONNECTION.NewNegotiatedAlpn;
+    const uint8_t* ListStart = Info->ClientAlpnList;
+    const uint8_t* ListEnd = Info->ClientAlpnList + Info->ClientAlpnListLength;
+    if (NewNegotiatedAlpn &&
+        ListStart <= NewNegotiatedAlpn &&
+        ListEnd > NewNegotiatedAlpn &&
+        NewNegotiatedAlpn + NewNegotiatedAlpn[0] + sizeof(uint8_t) <= ListEnd
+    ) {
+        //
```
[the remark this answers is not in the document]
liveans: Is there any specific error code you want me to send, or `QUIC_ERROR_CONNECTION_REFUSED` is enough?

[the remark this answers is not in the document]
nibanks: "internal error" probably is just fine

docs/api/QUIC_LISTENER_CALLBACK.md
```diff
@@ -0,0 +1,43 @@
...
+
+# Parameters
+
+`Listener`
+
+The valid handle to the listener object this event is for.
+
+`Context`
+
+The application callback context (optionally) supplied in [ListenerOpen](ListenerOpen.md), [SetCallbackHandler](SetCallbackHandler.md) or [SetContext](SetContext.md).
```
nibanks, line 28:
```suggestion
The application callback context (optionally) supplied in [ListenerOpen](ListenerOpen.md).
```
Let's just say ListenerOpen.

docs/api/ListenerStop.md
```diff
@@ -31,3 +31,4 @@ If the server application wishes to resume receiving new connections, it may cal
 [ListenerOpen](ListenerOpen.md)<br>
 [ListenerClose](ListenerClose.md)<br>
 [ListenerStart](ListenerStart.md)<br>
+[QUIC_LISTENER_EVENT](QUIC_LISTENER_EVENT.md)<br>
```
nibanks, line 34: I think only ListenerOpen.md really needs this link.

docs/api/QUIC_LISTENER_EVENT.md
```diff
@@ -0,0 +1,89 @@
...
+
+This field indicates the [QUIC_NEW_CONNECTION_INFO](QUIC_NEW_CONNECTION_INFO.md) structure for the new connection.
+
+`Connection`
+
+This field indicates the valid handle to the new incoming connection.
+
+`NewNegotiatedAlpn`
+
+This field is optional to set. If you want to set, this field **must** indicate a valid address on the ClientAlpnList.
```
nibanks, line 47:
```suggestion
This field is optional to set. If you want to set, this field **must** indicate a valid address on the `ClientAlpnList` of the `QUIC_NEW_CONNECTION_INFO`.
```

docs/api/QUIC_LISTENER_EVENT.md
```diff
@@ -0,0 +1,89 @@
...
+If you do not set this field, server will use the first matching ALPN in the ClientAlpnList.
```
nibanks, line 48:
```suggestion
If you do not set this field, server will use `NegotiatedAlpn` in the `QUIC_NEW_CONNECTION_INFO`.
```

docs/api/QUIC_NEW_CONNECTION_INFO.md
```diff
@@ -0,0 +1,78 @@
...
+    _Field_size_bytes_opt_(ServerNameLength)
+    const char* ServerName;
+} QUIC_NEW_CONNECTION_INFO;
+```
+
+# Parameters
+
+`QuicVersion`
+
+The QUIC version of the new incoming connection.
```
nibanks, line 33:
```suggestion
The QUIC version negotiated for the new incoming connection.
```

src/core/listener.c
```diff
@@ -643,6 +644,57 @@ QuicListenerClaimConnection(
         return FALSE;
     }
 
+    const uint8_t* NewNegotiatedAlpn = Event.NEW_CONNECTION.NewNegotiatedAlpn;
+    if (NewNegotiatedAlpn) {
+        const uint8_t* ListStart = Info->ClientAlpnList;
+        const uint8_t* ListEnd = Info->ClientAlpnList + Info->ClientAlpnListLength;
+        if (!(ListStart <= NewNegotiatedAlpn &&
+            ListEnd > NewNegotiatedAlpn &&
+            NewNegotiatedAlpn + NewNegotiatedAlpn[0] + sizeof(uint8_t) <= ListEnd)
```
nibanks, line 653: I find the ~(x && y && ...) format hard to parse. Would you mind inverting it?

liveans: I'll invert it

src/core/listener.c
```diff
@@ -643,6 +644,57 @@ QuicListenerClaimConnection(
...
+        ) {
+            QuicTraceEvent(
+                ListenerError,
+                "[list][%p] ERROR, %s.",
+                Listener,
+                "'NewNegotiatedAlpn' field is out of bounds of the 'ClientAlpnList' buffer.");
+            QuicConnTransportError(
+                Connection,
+                QUIC_ERROR_INTERNAL_ERROR);
+            goto Error;
```
nibanks, line 663:
```suggestion
            return FALSE;
```

src/core/listener.c
```diff
@@ -643,6 +644,57 @@ QuicListenerClaimConnection(
...
+            if (NegotiatedAlpn == NULL) {
+                QuicTraceEvent(
+                    AllocFailure,
+                    "Allocation of '%s' failed. (%llu bytes)",
+                    "NegotiatedAlpn",
+                    NegotiatedAlpnLength);
+                QuicConnTransportError(
+                    Connection,
+                    QUIC_ERROR_INTERNAL_ERROR);
+                goto Error;
```
nibanks, line 688:
```suggestion
                return FALSE;
```

src/core/listener.c
```diff
@@ -643,6 +644,57 @@ QuicListenerClaimConnection(
...
+            }
+        }
+        NegotiatedAlpn[0] = NegotiatedAlpnLength;
+        CxPlatCopyMemory(NegotiatedAlpn + 1, Event.NEW_CONNECTION.NewNegotiatedAlpn + 1, NegotiatedAlpnLength);
+        Connection->Crypto.TlsState.NegotiatedAlpn = NegotiatedAlpn;
+    }
+
+Error:
```
nibanks, line 696:
```suggestion

```

src/core/listener.c
```diff
@@ -643,6 +644,55 @@ QuicListenerClaimConnection(
...
     }
 
+    const uint8_t* NewNegotiatedAlpn = Event.NEW_CONNECTION.NewNegotiatedAlpn;
+    if (NewNegotiatedAlpn) {
+        const uint8_t* ListStart = Info->ClientAlpnList;
+        const uint8_t* ListEnd = Info->ClientAlpnList + Info->ClientAlpnListLength;
+        if (ListStart > NewNegotiatedAlpn ||
+            ListEnd <= NewNegotiatedAlpn ||
+            NewNegotiatedAlpn + NewNegotiatedAlpn[0] + sizeof(uint8_t) > ListEnd
+        ) {
```
nibanks, line 654:
```suggestion
            NewNegotiatedAlpn + NewNegotiatedAlpn[0] + sizeof(uint8_t) > ListEnd) {
```

src/core/listener.c
```diff
@@ -643,6 +644,55 @@ QuicListenerClaimConnection(
...
+                Connection,
+                QUIC_ERROR_INTERNAL_ERROR);
+            return FALSE;
+        }
+
+        //
+        // Free current ALPN buffer if it's allocated on heap.
+        //
+        if (Connection->Crypto.TlsState.NegotiatedAlpn != Connection->Crypto.TlsState.SmallAlpnBuffer) {
+            CXPLAT_FREE(Connection->Crypto.TlsState.NegotiatedAlpn, QUIC_POOL_ALPN);
```
nibanks, line 670: You need to NULL after free, in case you fail to alloc below, you'd never clear the pointer.

liveans: I'll do it after the pipeline.

nibanks (approved):
Thanks so much for doing this!

src/cs/lib/msquic_generated.cs
```diff
@@ -1820,6 +1820,9 @@ internal unsafe partial struct _NEW_CONNECTION_e__Struct
 
                 [NativeTypeName("HQUIC")]
                 internal QUIC_HANDLE* Connection;
+
+                [NativeTypeName("const uint8_t *")]
+                internal byte* NewNegotiatedAlpn;
```
ManickaP, line 1825: What's this good for? And shouldn't it also have length specified somewhere?

ManickaP (commented):
I don't fully understand the changes here, but we don't need any API changes. What we need is to be able to narrow down ALPN list when we call `ConnectionSetConfiguration` after we got `NEW_CONNECTION`. Note that this needs to be asynchronous since we're invoking user provided async callback, so we cannot set it within `NEW_CONNECTION` event.

The configuration already supports setting ALPN, we just need msquic to handle it for inbound connections.

PR #2962, opened 2022-08-10
Make it possible to partially statically link CRT

MichalStrehovsky:
This is similar to static linking, except we don't statically link the OS-provided part. CoreCLR runtime builds in the same way. This avoids dependency on VCRuntime140.dll that doesn't ship with the OS.

On my machine, the release build of msquic with statically linked CRT is 394,240 bytes. Dynamically linked is 311,296 bytes. "Partial static" slots in between at 322,560 bytes.

Contributes to dotnet/msquic#101.


nibanks (approved):
Is there any reason not to make `QUIC_STATIC_LINK_PARTIAL_CRT` the default going forward?

PR #2967, opened 2022-08-11
[Extension]: Grease Quic Bit Support

### LINKED ISSUE #1183: Support Grease QUIC bit Extension

This extension (https://tools.ietf.org/html/draft-thomson-quic-bit-grease) allows for the negotiation of greasing the "QUIC bit" or fixed bit in the long and short QUIC packet headers. It would probably be a good idea to support this extension.

liveans:
## Description

Closes #1183. This PR aims to support [draft-ietf-quic-bit-grease-04](https://www.ietf.org/archive/id/draft-ietf-quic-bit-grease-04.html) extension.

## Testing

Necessary tests added to handshake.

## Documentation

Yes, `GreaseQuicBitEnabled` was added to docs.


src/core/quicdef.h
```diff
@@ -495,6 +495,11 @@ CXPLAT_STATIC_ASSERT(
 //
 #define QUIC_DEFAULT_DEST_CID_UPDATE_IDLE_TIMEOUT_MS 20000
 
+//
+// The default value for enabling grease quic bit extension.
+//
+#define QUIC_DEFAULT_GREASE_QUIC_BIT_ENABLED         TRUE
```
nibanks, line 501: I want to (at least for now) disable this by default.

liveans: Alright!

src/core/inline.c
```diff
@@ -517,6 +518,7 @@ QuicPacketEncodeShortHeaderV1(
     _In_ uint16_t BufferLength,
     _Out_writes_bytes_opt_(BufferLength)
         uint8_t* Buffer
+    _In_ BOOLEAN FixedBit
```
nibanks, line 521: But this after the other `_In_ BOOLEAN`

src/core/inline.c
```diff
@@ -502,7 +502,8 @@ QuicPacketEncodeLongHeaderV1(
     _Out_writes_bytes_(BufferLength)
         uint8_t* Buffer,
     _Out_ uint16_t* PayloadLengthOffset,
-    _Out_ uint8_t* PacketNumberLength
+    _Out_ uint8_t* PacketNumberLength,
+    _In_ BOOLEAN FixedBit
```
nibanks, line 506: Put this after the `PacketType`

src/core/packet.c
```diff
@@ -211,7 +211,8 @@ QuicPacketValidateLongHeaderV1(
     _Inout_ CXPLAT_RECV_PACKET* Packet,
     _Outptr_result_buffer_maybenull_(*TokenLength)
         const uint8_t** Token,
-    _Out_ uint16_t* TokenLength
+    _Out_ uint16_t* TokenLength,
+    _In_ BOOLEAN ExpectedFixedBit
```
nibanks, line 215: Not used?

src/core/packet.c
```diff
@@ -565,7 +566,8 @@ _Success_(return != FALSE)
 BOOLEAN
 QuicPacketValidateShortHeaderV1(
     _In_ const void* Owner, // Binding or Connection depending on state
-    _Inout_ CXPLAT_RECV_PACKET* Packet
+    _Inout_ CXPLAT_RECV_PACKET* Packet,
+    _In_ BOOLEAN ExpectedFixedBit
```
nibanks, line 570: Not used?

src/core/packet.c
liveans: I'll use it, just designing it for now.

src/core/packet_builder.c
```diff
@@ -197,6 +197,7 @@ QuicPacketBuilderPrepare(
         Connection->Stats.QuicVersion == QUIC_VERSION_2 ?
             QuicKeyTypeToPacketTypeV2(NewPacketKeyType) :
             QuicKeyTypeToPacketTypeV1(NewPacketKeyType);
+    BOOLEAN GreaseQuicBitTPReceived = (BOOLEAN)((Connection->PeerTransportParams.Flags & QUIC_TP_FLAG_GREASE_QUIC_BIT) > 0);
```
nibanks, line 200:
```suggestion
    BOOLEAN NoGreaseQuicBit = !(Connection->PeerTransportParams.Flags & QUIC_TP_FLAG_GREASE_QUIC_BIT);
```
And then use `NoGreaseQuicBit` below?

src/core/binding.c
```diff
@@ -955,7 +955,7 @@ QuicBindingProcessStatelessOperation(
             PacketLength - QUIC_STATELESS_RESET_TOKEN_LENGTH,
             SendDatagram->Buffer);
         ResetPacket->IsLongHeader = FALSE;
-        ResetPacket->FixedBit = 1;
+        ResetPacket->FixedBit = 1; // TODO : Find a way to check connection settings and PeerTPs.
```
nibanks, line 958: There is no connection state in this case. That's the point of stateless reset. Just always assume grease was disabled.

src/core/connection.c
```diff
@@ -3698,6 +3702,10 @@ QuicConnRecvHeader(
 
     CXPLAT_FRE_ASSERT(QuicIsVersionSupported(Connection->Stats.QuicVersion));
 
+    const BOOLEAN IgnoreFixedBit = (Connection->Settings.GreaseQuicBitEnabled &&
+        (QuicConnIsClient(Connection) ||
```
nibanks, line 3706: Why does client always do it? Put a comment here explaining this?

nibanks: And should we constantly be reevaluating this or should we store it on the connection?

liveans: I'd prefer to store this state in the connection.

liveans: If this feature is turned on in the client, it can accept FixedBit as 0 according to RFC since it will send this TP, but the Server cannot ignore this bit without receiving TP.

src/core/binding.c
liveans: Alright!

src/core/connection.c
liveans: Yeah, we should definitely store it on the connection. In this way, things will get easier to do.

src/core/connection.h
```diff
@@ -624,6 +624,10 @@ typedef struct QUIC_CONNECTION {
         QUIC_FLOW_BLOCKED_TIMING_TRACKER FlowControl;
     } BlockedTimings;
 
+    uint8_t IgnoreFixedBit : 1;
+    uint8_t FixedBit : 1;
+    uint8_t RESERVED : 6;
```
nibanks, line 629: Put these in the flags/state structure.

src/core/crypto_tls.c
```diff
@@ -1132,6 +1139,24 @@ QuicCryptoTlsEncodeTransportParameters(
...
+        QuicTraceLogConnVerbose(
+            EncodeTPGreaseQuicBit,
+            Connection,
+            "TP: Grease Quic Bit");
+
+        //
+        // An endpoint that advertises the grease_quic_bit transport parameter MUST accept packets
+        // with the QUIC Bit set to a value of 0.
+        //
+        Connection->IgnoreFixedBit = 1;
```
nibanks, line 1158: Don't actually touch the `Connection` in here. It's used purely for logging.

src/core/crypto_tls.c
```diff
@@ -1780,6 +1805,28 @@ QuicCryptoTlsDecodeTransportParameters(
...
+                    Length,
+                    "Invalid length of QUIC_TP_ID_GREASE_QUIC_BIT");
+                goto Exit;
+            }
+            TransportParams->Flags |= QUIC_TP_FLAG_GREASE_QUIC_BIT;
+            //
+            // Endpoints that receive the grease_quic_bit transport parameter from a peer SHOULD set the QUIC Bit to an unpredictable value
+            // unless another extension assigns specific meaning to the value of the bit.
+            //
+            Connection->FixedBit = 0;
```
nibanks, line 1823: Don't put this in here.

src/core/listener.c
```diff
@@ -748,6 +748,9 @@ QuicListenerAcceptConnection(
         return;
     }
 
+    Connection->FixedBit = 1;
+    Connection->IgnoreFixedBit = 0;
```
nibanks, line 752: Any non-zero defaults should be set in the connection initialization function. Better still, rework the variables so `0` is the default, so you don't have to do anything special.

liveans: I assume the Connection Initialization function is `QuicConnAlloc`?

src/core/settings.c
```diff
@@ -1453,6 +1478,7 @@ QuicSettingsSettingsToInternal(
     SETTING_COPY_TO_INTERNAL(MigrationEnabled, Settings, InternalSettings);
     SETTING_COPY_TO_INTERNAL(DatagramReceiveEnabled, Settings, InternalSettings);
     SETTING_COPY_TO_INTERNAL(ServerResumptionLevel, Settings, InternalSettings);
+    SETTING_COPY_TO_INTERNAL(GreaseQuicBitEnabled, Settings, InternalSettings); // We can't copy it via sized version due to bit field operation not allowed on it.
```
nibanks, line 1481: How does this effect down-level support?

src/core/listener.c
nibanks: Yes.

src/inc/msquic.h
```diff
@@ -635,7 +636,8 @@ typedef struct QUIC_SETTINGS {
     uint8_t MigrationEnabled                : 1;
     uint8_t DatagramReceiveEnabled          : 1;
     uint8_t ServerResumptionLevel           : 2;    // QUIC_SERVER_RESUMPTION_LEVEL
-    uint8_t RESERVED                        : 2;
+    uint8_t GreaseQuicBitEnabled            : 1;
```
nibanks, line 639: We should probably expose whether this was negotiated or not in the QUIC statistics. Also, we should update the QUIC interop tool to enable this.

liveans: Alright! I'm gonna add some stat fields about it. Also gonna take a look at QUIC Interop tools as well!

src/core/settings.c
liveans: It shouldn't affect down-level support I suppose, since we're not breaking anything here.

src/core/connection.c
```diff
@@ -1965,6 +1966,8 @@ QuicConnStart(
     Connection->RemoteServerName = ServerName;
     ServerName = NULL;
 
+    Connection->State.IgnoreFixedBit = Configuration->Settings.GreaseQuicBitEnabled;
```
nibanks, line 1969: Do we actually need the `IgnoreFixedBit` flag at all, or can we replace al usage with `Settings.GreaseQuicBitEnabled`?

src/core/binding.c
```diff
@@ -1034,6 +1034,7 @@ QuicBindingProcessStatelessOperation(
         SendDatagram->Length =
             QuicPacketEncodeRetryV1(
                 RecvPacket->LH->Version,
+                TRUE, // TODO : Find a way to check connection settings and PeerTPs.
```
nibanks, line 1037: I think when writing all stateless packets (Retry, Reset) we're going to always set the fixed bit (not grease it) because we're never going to have any connection-specific state to base this decision on. So, really, I think we might as well, undo the changes to these encode function and just have them stay fixed.

src/core/connection.h
```diff
@@ -175,6 +175,16 @@ typedef union QUIC_CONNECTION_STATE {
...
 
+        //
+        // This flag indicates that the connection should ignore the fixed bit on received packets.
+        //
+        BOOLEAN IgnoreFixedBit : 1;
+
+        //
+        // This value of the fixed bit on send packets.
+        //
+        BOOLEAN FixedBit : 1;
```
nibanks, line 186: See the comment at the top of this union. We need to keep this copy in sync with the one in quictypes.h.

src/core/binding.c
liveans: Alright! I'm going to undo it to the original signature.

src/core/connection.c
liveans: No, we don't need it. I'm gonna delete it in the next commit.

src/inc/msquic.h
liveans: I added stats, and I'll add them to the necessary places to increase values.

src/core/connection.c
```diff
@@ -6655,6 +6661,9 @@ QuicConnGetV2Statistics(
     if (STATISTICS_HAS_FIELD(*StatsLength, DestCidUpdateCount)) {
         Stats->DestCidUpdateCount = Connection->Stats.Misc.DestCidUpdateCount;
     }
+    if(STATISTICS_HAS_FIELD(*StatsLength, RecvGreaseBitTpCount)) {
```
liveans, line 6664:
```suggestion
    if (STATISTICS_HAS_FIELD(*StatsLength, RecvGreaseBitTpCount)) {
```

src/core/connection.h
```diff
@@ -280,6 +290,7 @@ typedef struct QUIC_CONN_STATS {
 
         uint64_t TotalBytes;            // Sum of UDP payloads
         uint64_t TotalStreamBytes;      // Sum of stream payloads
+        uint64_t GreaseBitTpCount;      // Count of received grease transport parameter from peer.
```
liveans, line 293: `uint64_t` might be overkill in here.

src/core/packet.h
```diff
@@ -195,7 +195,7 @@ typedef struct QUIC_RETRY_PACKET_V1 {
 
     uint8_t UNUSED          : 4;
     uint8_t Type            : 2;
-    uint8_t FixedBit        : 1;    // Must be 1.
+    uint8_t FixedBit        : 1;    // Must be 1, unless grease_quic_bit tp has been sent.
```
liveans, line 198:
```suggestion
    uint8_t FixedBit        : 1;    // Must be 1.
```

src/inc/msquic.h
```diff
@@ -485,15 +485,17 @@ typedef struct QUIC_STATISTICS_V2 {
...
     // N.B. New fields must be appended to end
 
 } QUIC_STATISTICS_V2;
 
 #define QUIC_STRUCT_SIZE_THRU_FIELD(Struct, Field) \
     (FIELD_OFFSET(Struct, Field) + sizeof(((Struct*)0)->Field))
 
-#define QUIC_STATISTICS_V2_SIZE_1   QUIC_STRUCT_SIZE_THRU_FIELD(QUIC_STATISTICS_V2, KeyUpdateCount)         // v2.0 final size
-#define QUIC_STATISTICS_V2_SIZE_2   QUIC_STRUCT_SIZE_THRU_FIELD(QUIC_STATISTICS_V2, DestCidUpdateCount)     // v2.1 final size
+#define QUIC_STATISTICS_V2_SIZE_1   QUIC_STRUCT_SIZE_THRU_FIELD(QUIC_STATISTICS_V2, KeyUpdateCount)               // v2.0 final size
```
liveans, line 497:
```suggestion
#define QUIC_STATISTICS_V2_SIZE_1   QUIC_STRUCT_SIZE_THRU_FIELD(QUIC_STATISTICS_V2, KeyUpdateCount)         // v2.0 final size
```

src/core/packet_builder.c
```diff
@@ -197,6 +197,27 @@ QuicPacketBuilderPrepare(
...
+        (Connection->PeerTransportParams.Flags & QUIC_TP_FLAG_GREASE_QUIC_BIT) > 0) {
+        //
+        // Endpoints that receive the grease_quic_bit transport parameter from a peer SHOULD set the QUIC Bit to an unpredictable value
+        // unless another extension assigns specific meaning to the value of the bit.
+        //
+        uint8_t RandomValue;
+        (void) CxPlatRandom(sizeof(RandomValue), &RandomValue);
+        Connection->State.FixedBit = (RandomValue % 2);
+        Connection->State.FixedBitNegotiated = TRUE;
+        Connection->Stats.Recv.GreaseBitTpCount++;
```
liveans, line 211: I think this is not the right place to do it, any suggestion on it @nibanks?

src/inc/msquic.h
```diff
@@ -485,6 +485,8 @@ typedef struct QUIC_STATISTICS_V2 {
 
     uint32_t DestCidUpdateCount;            // Number of times the destionation CID changed.
 
+    uint64_t RecvGreaseBitTpCount;          // Count of received grease transport parameter from the peer.
```
nibanks, line 488: No, we don't need this. I was more saying we should have a flag at the top that says `GreaseBitNegotiated`.

liveans: Oh, something like I added to stats? `FixedBitNegotiated`?

src/core/packet_builder.c
```diff
@@ -197,6 +197,26 @@ QuicPacketBuilderPrepare(
...
+        (Connection->PeerTransportParams.Flags & QUIC_TP_FLAG_GREASE_QUIC_BIT) > 0) {
+        //
+        // Endpoints that receive the grease_quic_bit transport parameter from a peer SHOULD set the QUIC Bit to an unpredictable value
+        // unless another extension assigns specific meaning to the value of the bit.
+        //
+        uint8_t RandomValue;
+        (void) CxPlatRandom(sizeof(RandomValue), &RandomValue);
+        Connection->State.FixedBit = (RandomValue % 2);
+        Connection->Stats.GreaseBitNegotiated = TRUE;
+    }
```
nibanks, line 211: This logic should go in `QuicConnProcessPeerTransportParameters` which is called when we receive the transport parameters. Also please try to keep comments to 80 character width.

src/lib.rs
```diff
@@ -459,14 +459,16 @@ pub struct QuicStatisticsRecv {
     /// Includes DuplicatePackets.
     pub dropped_packets: u64,
     pub duplicate_packets: u64,
-    /// Sum of UDP payloads
-    pub total_bytes: u64,
-    /// Sum of stream payloads
-    pub total_stream_bytes: u64,
```
nibanks, line 465: Why did you move these? From msquic.h:
```c
    struct {
        uint64_t TotalPackets;          // QUIC packets; could be coalesced into fewer UDP datagrams.
        uint64_t ReorderedPackets;      // Packets where packet number is less than highest seen.
        uint64_t DroppedPackets;        // Includes DuplicatePackets.
        uint64_t DuplicatePackets;
        uint64_t TotalBytes;            // Sum of UDP payloads
        uint64_t TotalStreamBytes;      // Sum of stream payloads
        uint64_t DecryptionFailures;    // Count of packet decryption failures.
        uint64_t ValidAckFrames;        // Count of receive ACK frames.
    } Recv;
```

src/inc/msquic.h
```diff
@@ -448,6 +448,7 @@ typedef struct QUIC_STATISTICS_V2 {
     uint32_t StatelessRetry         : 1;
     uint32_t ResumptionAttempted    : 1;
     uint32_t ResumptionSucceeded    : 1;
+    uint32_t GreaseBitNegotiated    : 1;    // Set if we negotiated the GREASE bit.
```
nibanks, line 451: Can you add a `uint32_t RESERVED : 27` too, so it's clear how much padding is here?

src/lib.rs
liveans: Oh, looks like I got the structs mixed up

src/inc/msquic.h
liveans: Yes!

src/core/packet_builder.c
liveans: I moved the logic into `QuicConnProcessPeerTransportParameters`, also I'm going to fix all comments with 80 character width!

src/core/connection.c
```diff
@@ -2884,6 +2889,19 @@ QuicConnProcessPeerTransportParameters(
         "Peer Transport Parameters Set");
     Connection->State.PeerTransportParameterValid = TRUE;
 
+    if (Connection->Stats.GreaseBitNegotiated == FALSE &&
+        (Connection->PeerTransportParams.Flags & QUIC_TP_FLAG_GREASE_QUIC_BIT) > 0) {
```
nibanks, line 2893: I don't think we should have this extra condition:
```suggestion
    if (Connection->PeerTransportParams.Flags & QUIC_TP_FLAG_GREASE_QUIC_BIT) {
```

nibanks: And actually, shouldn't we add a condition for only if the local setting is enabled?

nibanks: And, let's move this to the end of the function too please.

nibanks:
Seems like we're getting this compiler warning/error too:
```
/home/vsts/work/1/s/src/core/crypto_tls.c:1192:1: error: function 'QuicCryptoTlsDecodeTransportParameters' exceeds recommended size/complexity thresholds [google-readability-function-size,-warnings-as-errors]
QuicCryptoTlsDecodeTransportParameters(
^
/home/vsts/work/1/s/src/core/crypto_tls.c:1192:1: note: 806 statements (threshold 800)
```
Can you go ahead and simply ignore that error?

src/core/connection.c
liveans: > And actually, shouldn't we add a condition for only if the local setting is enabled?

Yes! Looks like I forgot it, I'll add it.

nibanks (dismissed):

csujedihy:
feel free to ignore [the arm docker build failure.](https://github.com/microsoft/msquic/runs/7925352774?check_suite_focus=true) for now.

nibanks (approved):

nibanks:
Thanks so much for your contributions!!!

PR #2972, opened 2022-08-11
Support ALPN List Change via ConnectionSetConfiguration function

liveans:
## Description

Related to #2418 and #2959. Correct the behavior on alpn change

## Testing

There is an existing test to test this feature.

## Documentation

Yes, removed `NewNegotiatedAlpn` from docs and added this behavior to docs.


liveans:
@nibanks I think I made a mistake while creating the branch, there are commits from the main branch. I can create the PR from scratch if you want to.

src/core/binding.c
```diff
@@ -542,6 +542,8 @@ QuicBindingAcceptConnection(
     }
     CxPlatCopyMemory(NegotiatedAlpn, Info->NegotiatedAlpn - 1, NegotiatedAlpnLength);
     Connection->Crypto.TlsState.NegotiatedAlpn = NegotiatedAlpn;
+    Connection->ClientAlpnList = Info->ClientAlpnList;
+    Connection->ClientAlpnListLength = Info->ClientAlpnListLength;
```
nibanks, line 546: Let's save this in the TlsState.

src/core/connection.c
```diff
@@ -2516,8 +2516,88 @@ QuicConnSetConfiguration(
             Status = QUIC_STATUS_INVALID_PARAMETER;
             goto Error;
         }
+        const uint8_t* AlpnList = Connection->Configuration->AlpnList;
```
nibanks, line 2519: Let's move all this logic into a help function, perhaps in crypto.c.

liveans: Done

src/core/connection.c
```diff
@@ -2516,6 +2516,14 @@ QuicConnSetConfiguration(
             Status = QUIC_STATUS_INVALID_PARAMETER;
             goto Error;
         }
+
+        Status = QuicCryptoReNegotiateAlpn(
+            Connection,
+            Connection->Configuration->AlpnListLength,
+            Connection->Configuration->AlpnList);
+        if(QUIC_FAILED(Status)) {
```
nibanks, line 2524:
```suggestion
        Status =
            QuicCryptoReNegotiateAlpn(
                Connection,
                Connection->Configuration->AlpnListLength,
                Connection->Configuration->AlpnList);
        if (QUIC_FAILED(Status)) {
```

src/core/crypto.c
```diff
@@ -2542,3 +2542,88 @@ QuicCryptoDecodeClientTicket(
...
+
+QUIC_STATUS
+QuicCryptoReNegotiateAlpn(
+    _In_opt_ QUIC_CONNECTION* Connection,
+    _In_ uint16_t AlpnListLength,
+    _In_reads_bytes_(AlpnListLength)
+        const uint8_t* AlpnList
+    )
+{
+    if(Connection == NULL) {
```
nibanks, line 2554:
```suggestion
    if (Connection == NULL) {
```

nibanks: Actually CXPLAT_DBG_ASSERT that Connection and AlpnList are non-null. And that length is not zero.

src/core/connection.c
```diff
@@ -2516,6 +2516,15 @@ QuicConnSetConfiguration(
...
         }
+
+        Status =
+            QuicCryptoReNegotiateAlpn(
+                Connection,
+                Connection->Configuration->AlpnListLength,
+                Connection->Configuration->AlpnList);
+        if (QUIC_FAILED(Status)) {
+            goto Error;
+        }
```
nibanks, line 2527:
```suggestion
        }
        Connection->Crypto.TlsState.ClientAlpnList = NULL;
        Connection->Crypto.TlsState.ClientAlpnListLength = 0;
```
Clear out the list after this, since the memory it references is going to be invalid.

src/core/crypto.c
liveans: Done

src/core/crypto.c
```diff
@@ -2542,3 +2542,83 @@ QuicCryptoDecodeClientTicket(
...
+
+    if (NewNegotiatedAlpn == NULL) {
+        QuicTraceEvent(
+            ConnError,
+            "[conn][%p] ERROR, %s.",
+            Connection,
+            "No ALPN match found");
+        QuicConnTransportError(
+                Connection,
+                QUIC_ERROR_INTERNAL_ERROR);
```
nibanks, line 2589:
```suggestion
        QuicConnTransportError(
            Connection,
            QUIC_ERROR_INTERNAL_ERROR);
```

nibanks (approved):

PR #2973, opened 2022-08-11
Fix kqueue behavior for FreeBSD

scopedog:
## Description

Fix kqueue behavior for FreeBSD and maybe MacOS.

## Testing

The fix will also affect MacOS and tests are needed.

## Documentation

It looks like
```
EV_SET(&Event, EVFILT_USER)
kevent(ProcContext->KqueueFd, &Event)
```
doesn't trigger an event at kevent() loop in CxPlatDataPathRunEC() and the loop blocks forever with FreeBSD.
Instead of adding EVFILT_USER event by kevent, I used fd created by pipe() and registered it to ProcContext->KqueueFd, then notified the loop by write(fd).

### DIFF
--- a/src/inc/msquic_posix.h
+++ b/src/inc/msquic_posix.h
@@ -425,6 +425,9 @@ QuicAddr4FromString(
         }
     }
     Addr->Ip.sa_family = QUIC_ADDRESS_FAMILY_INET;
+#if defined(__FreeBSD__)
+    Addr->Ipv4.sin_len = sizeof(struct sockaddr_in);
+#endif
     return TRUE;
 }
 
@@ -459,6 +462,9 @@ QuicAddr6FromString(
         }
     }
     Addr->Ip.sa_family = QUIC_ADDRESS_FAMILY_INET6;
+#if defined(__FreeBSD__)
+    Addr->Ipv6.sin6_len = sizeof(struct sockaddr_in6);
+#endif
     return TRUE;
 }
 
--- a/src/inc/msquic_posix.h
+++ b/src/inc/msquic_posix.h
@@ -97,6 +97,10 @@ inline ENUMTYPE &operator ^= (ENUMTYPE &a, ENUMTYPE b) throw() { return (ENUMTYP
 #define ENOKEY 126
 #endif // ENOKEY
 
+#ifndef ETIME // undefined on FreeBSD
+#define ETIME ETIMEDOUT
+#endif // ETIME
+
 #define ERROR_BASE                          200000000                       // 0xBEBC200
 #define TLS_ERROR_BASE                      256 + ERROR_BASE                // 0xBEBC300
 #define CERT_ERROR_BASE                     512 + ERROR_BASE                // 0xBEBC400
--- a/src/inc/msquic_posix.h
+++ b/src/inc/msquic_posix.h
@@ -98,7 +98,7 @@ inline ENUMTYPE &operator ^= (ENUMTYPE &a, ENUMTYPE b) throw() { return (ENUMTYP
 #endif // ENOKEY
 
 #ifndef ETIME // undefined on FreeBSD
-#define ETIME ETIMEDOUT
+#define ETIME 101 // same as macOS's: change this if 101 is added to errno.h
 #endif // ETIME
 
 #define ERROR_BASE                          200000000                       // 0xBEBC200
--- a/src/inc/quic_platform_posix.h
+++ b/src/inc/quic_platform_posix.h
@@ -23,12 +23,14 @@ Environment:
 #error "Incorrectly including Posix Platform Header from unsupported platfrom"
 #endif
 
+/*
 // For FreeBSD
 #if defined(__FreeBSD__)
 #include <sys/socket.h>
 #include <netinet/in.h>
 #define ETIME   ETIMEDOUT
 #endif
+*/
 
 #include <stdlib.h>
 #include <stdint.h>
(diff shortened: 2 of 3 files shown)


nibanks:
I'm not convinced this is necessary. There's got to be a better, less complicated way. Why doesn't `kevent(ProcContext->KqueueFd, &Event)` work?

scopedog:
I have no idea but even if kevent(ProcContext->KqueueFd, &Event) doesn't work, kevent() loop must once exit and restart.

scopedog:
I think the errors are related to clog which I don't know anything about.
Could you tell me how to fix it?

nibanks:
According to the documentation I've read, the code we have _should_ work. I am very hesitant to go forward with this change. I'm in the middle of refactoring all datapath code across the platforms into a single unified one, and this change would cause a lot of issues for that work. Setting the user event should wake up the kevent loop. Why does it not?

scopedog:
As I mentioned, the kvent() wait (CxPlatDataPathRunEC) thread and the threads that add an event (CxPlatSocketContextStartReceive, CxPlatSocketSendInternal, CxPlatDataPathWake) are different, and in my environment the wait is called first and an event is added later. In this case, the adding an event thread has to notify the wait thread that a new event was added, where I used pipe fd, then kevent() wait needs to be restarted.

If you want to do that without pipe, then you will have to register an event right after KqueueFd is created that is triggered every time there is a change in KqueueFd. I don't know if it works with
```
EV_SET(KqueueFd, EVFILT_USER, NOTE_TRIGGER);
kevent(KqueueFd);
```
but will try. If it doesn't work, then pipe will be the best solution.

scopedog:
> ```
> EV_SET(KqueueFd, EVFILT_USER, NOTE_TRIGGER);
> kevent(KqueueFd);
> ```

Unfortunately, this idea didn't work either. I'm not very sure but KqueueFd monitoring KqueueFd itself maybe doesn't work. That said, there should be a way to wake up the kevent wait with EVFILT_USER.
If you create another kqueue fd that monitors KqueueFd, you may be able to achieve what you want to do. However, I don't think you like to do that.
I'll investigate more.

scopedog:
Well, I could manage to make this work only with kevent + EVFILT_USER but need to double check if my code is correct. I will post it once it's confirmed.

src/inc/msquic_posix.h
```diff
@@ -97,6 +97,10 @@ inline ENUMTYPE &operator ^= (ENUMTYPE &a, ENUMTYPE b) throw() { return (ENUMTYP
 #define ENOKEY 126
 #endif // ENOKEY
 
+#ifndef ETIME // undefined on FreeBSD
+#define ETIME ETIMEDOUT
```
nibanks, line 101: We need a different value here so we can differentiate `QUIC_STATUS_CONNECTION_TIMEOUT` and `QUIC_STATUS_CONNECTION_IDLE` (see below).

scopedog:
As long as I see msquic_posix.h, assigning an arbitrary value to an undefined error seems OK:
```
#ifndef ENOKEY // undefined om macOS
#define ENOKEY 126
#endif // ENOKEY
```

If it is correct, we could give a number same as MacOS's like:
```
#ifndef ETIME // undefined on FreeBSD
#define ETIME 101 // same as macOS's
#endif // ETIME

```
101 is not used in FreeBSD's errno.h but if you think a larger number is better so it won't conflict with other errno in the future, we can change it.

nibanks:
@scopedog I just merged a PR that completely refactors the various network IO loops. Can you please see if the latest code fixes your problems?

scopedog:
I'm currently out of state and won't be back till 9/10. I'll check the code after that. Also, I ran test programs on FreeBSD and found some errors which seem to be due to kqueue's different behaviors from epoll's. I'll investigate it also.

scopedog:
I found some errors. Will newly PR.

nibanks:
> I have a question. In platform_worker.c around line 147, `CxPlatWorkers[i].ShutdownSqe = (CXPLAT_SQE)CxPlatWorkers[i].EventQ;` Isn't this `CxPlatWorkers[i].ShutdownSqe.Sqe = (CXPLAT_SQE)CxPlatWorkers[i].EventQ;`
> 
> Otherwise, I don't think types match.

`ShutdownSqe` is type `CXPLAT_SQE`, not `DATAPATH_SQE`.

scopedog:
OK, but in datapath_kqueue.c, ShutdownSqe is declared as:
```
    //
    // The event for the shutdown event.
    //
    DATAPATH_SQE ShutdownSqe;

```

nibanks:
> OK, but in datapath_kqueue.c, ShutdownSqe is declared as:
> 
> ```
>     //
>     // The event for the shutdown event.
>     //
>     DATAPATH_SQE ShutdownSqe;
> ```

That's a different `ShutdownSqe` from the one you referred to in platform_worker.c.

scopedog:
Oh, I see. Thanks.

scopedog:
I have resolved one issue. Changing CxPlatEventQEnqueue() as
```
    //EV_SET(&event, *sqe, EVFILT_USER, EV_ADD | EV_ONESHOT, NOTE_TRIGGER, 0, us
er_data);
    EV_SET(&event, *sqe, EVFILT_USER, 0, NOTE_TRIGGER, 0, user_data);
```
and changing all CxPlatEventQEnqueue() like:
```
CxPlatEventQEnqueue(
            &CxPlatWorkers[i].EventQ,
            &CxPlatWorkers[i].EventQ,
            //&CxPlatWorkers[i].ShutdownSqe,
            NULL);
```
```
CxPlatEventQEnqueue(
            SocketContext->DatapathProc->EventQ,
            //&SocketContext->ShutdownSqe.Sqe,
            SocketContext->DatapathProc->EventQ,
            &SocketContext->ShutdownSqe);
```
at least make the server program terminate by hitting Enter.
With kqueue, CxPlatWorkers[i].ShutdownSqe always needs to be equal to CxPlatWorkers[i].EventQ and SocketContext->ShutdownSqe.Sqe = SocketContext->DatapathProc->EventQ.
Let me know if you need to assign different values to them.

Will continue checking others.

PR #2979, opened 2022-08-13
[Posix]: sscanf with variable len format

### LINKED ISSUE #2410: Fix Security Warning from sscanf Usage in Posix Code
### Describe the bug

Code scanning warns that the Posix usage of `sscanf` in cgroup.c could be an unbounded write. Not really, but we should fix it to make the tool happy.

### Affected OS

- [ ] All
- [ ] Windows Server 2022
- [ ] Windows 11
- [X] Ubuntu
- [X] Debian
- [ ] Other (specify below)

### Additional OS information

_No response_

### MsQuic version

main

Go to https://github.com/microsoft/msquic/security/code-scanning to see the "Unbound write" warnings, such as https://github.com/microsoft/msquic/security/code-scanning/169?query=ref%3Arefs%2Fheads%2Fmain.

### Expected behavior

No warnings should be present

### Actual outcome

The code scanning fears that there might be an unbound write, but it shouldn't be the case considering how we allocate the space. But to make things a little safer, I think we should update the code to do something like the first answer here: https://stackoverflow.com/questions/5873402/is-sscanf-considered-safe-to-use.

### Additional details

_No response_

liveans:
## Description

Closes #2410. Trying to make the tool happy. 

## Testing

N/A

## Documentation

N/A

### DIFF
--- a/src/platform/cgroup.c
+++ b/src/platform/cgroup.c
@@ -201,8 +201,10 @@ FindHierarchyMount(
         //
         // See man page of proc to get format for /proc/self/mountinfo file
         //
+        char Format[40];
+        sprintf(Format, "%%%zus %%*s %%%zus", LineLen, LineLen);
         int SscanfRet = sscanf(SeparatorChar,
-                                " - %s %*s %s",
+                                Format,
                                 FilesystemType,
                                 Options);
         if (SscanfRet != 2) {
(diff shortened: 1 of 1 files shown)


liveans:
As discussed on discord with @nibanks, decided to ignore it. Looks like there is no good way to make the tool happy.

PR #2986, opened 2022-08-16
Enable minimal datagram support

mangolas:
## Description

Adds the missing (minimal) support for sending and receiving datagrams. 

## Testing

Tests do not cover datagrams.

## Documentation

Documentation in general in Rust side is very minimal.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -373,5 +373,6 @@ target/
 
 # Cargo (Rust) lock file
 Cargo.lock
-.idea
-cmake-build-debug/
+
+# IntelliJ project files
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -373,3 +373,5 @@ target/
 
 # Cargo (Rust) lock file
 Cargo.lock
+.idea
+cmake-build-debug/
--- a/scripts/build.rs
+++ b/scripts/build.rs
@@ -3,6 +3,7 @@
 
 use cmake::Config;
 use std::path::Path;
+use std::env;
 
 fn main() {
     let path_extra = "lib";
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -373,3 +373,5 @@ target/
 
 # Cargo (Rust) lock file
 Cargo.lock
+.idea
+cmake-build-debug/
```
nibanks, line 377: Do we need these?

scripts/build.rs
```diff
@@ -11,12 +12,26 @@ fn main() {
...
+
+    match target.as_str() {
+        "x86_64-apple-darwin" => config
+            .define("CMAKE_OSX_ARCHITECTURES", "x86_64")
+            .define("CMAKE_OSX_DEPLOYMENT_TARGET", "10.15"),
+        "aarch64-apple-darwin" => config
+            .define("CMAKE_OSX_ARCHITECTURES", "arm64")
+            .define("CMAKE_OSX_DEPLOYMENT_TARGET", "11.0"),
+        _ => &mut config
+    };
```
nibanks, line 32: Is this for macOS support?

src/lib.rs
```diff
@@ -1207,6 +1221,11 @@ impl Settings {
         self.idle_timeout_ms = value;
         self
     }
+    pub fn set_datagram_receive_enabled(&mut self, value: bool) -> &mut Settings {
+        self.is_set_flags |= (value as u64) << 27;
```
nibanks, line 1225: This we really be using `value` here? You are setting the receive enabled flag (to true or false) so I think this should always be set.

src/lib.rs
```diff
@@ -1310,7 +1329,7 @@ impl Drop for Registration {
 impl Configuration {
     pub fn new(
         registration: &Registration,
-        alpn: &Buffer,
+        alpn: &[Buffer],
```
nibanks, line 1332: Why the changes in this section?

.gitignore
mangolas: Not necessarily, wouldn't hurt either, but can be removed.

scripts/build.rs
mangolas: Yes, this allows to run eg. 'cargo build --target x86_64-apple-darwin` and so, in near future there is a need for Android targets as well.

.gitignore
nibanks: We can leave it. Thanks.

scripts/build.rs
nibanks: Ok. Thanks. Perhaps it's work updating the Cargo GitHub action to test macOS to validate?

src/lib.rs
mangolas: Configuration::new Rust function was missing a way to actually give more than one alpn parameter, so it had to be changed anyways, so I'm proposing here using Rust slice rather than reference and separate length. Listener `start` is another place alpn is passed so I changed that as well to be uniform.

src/lib.rs
nibanks: For instance, look at the functions above, such as `set_idle_timeout_ms`. The `is_set_flags` is always set because you are modifying the actual setting value (independent of it being true or false). So you should always be setting the appropriate bit in `is_set_flags` here. For instance:
```suggestion
        self.is_set_flags |= 1 << 27;
```

src/lib.rs
nibanks: Ok.

src/lib.rs
mangolas: Aah, finally got it, sure you are right, will change.

scripts/build.rs
nibanks: In cargo.yml try adding `macos-latest` to `os: [ubuntu-latest, windows-latest]`?

mangolas: Done.

nibanks (approved):

nibanks:
Thanks for the change!

PR #2989, opened 2022-08-17
retry pthread create if it failes with ENOKEY

wfurt:
## Description

when we try to set affinity on core that is offline pthread_create fails (on glibc) 
We should perhaps figure out better way how to do the core management but this is simple fix that unblocks .NET ARM testing. Performance may be sub-optimal but at least msquic is functional. 

## Testing

running .NET test suite on ARM64 box with some offline cores. 

## Documentation

no

### DIFF
--- a/src/platform/platform_posix.c
+++ b/src/platform/platform_posix.c
@@ -658,6 +658,12 @@ CxPlatThreadCreate(
             "[ lib] ERROR, %u, %s.",
             Status,
             "pthread_create failed");
+
+        // Try it again without attributes.
+        // pthread_create can fail with ENOKEY if we request affinity on CPU that is offline
+        if (Status == ENOKEY && pthread_create(Thread, NULL, Config->Callback, Config->Context) == 0) {
+            Status = QUIC_STATUS_SUCCESS;
+        }
     }
 
 #endif // !CXPLAT_USE_CUSTOM_THREAD_CONTEXT
--- a/src/platform/platform_posix.c
+++ b/src/platform/platform_posix.c
@@ -652,7 +652,7 @@ CxPlatThreadCreate(
 #else // CXPLAT_USE_CUSTOM_THREAD_CONTEXT
 
     //
-    // If pthread_create fails eith ENOKEY, then try again without the attribute
+    // If pthread_create fails with ENOKEY, then try again without the attribute
     // because the CPU might be offline.
     //
     if (pthread_create(Thread, &Attr, Config->Callback, Config->Context) &&
--- a/src/platform/platform_posix.c
+++ b/src/platform/platform_posix.c
@@ -651,19 +651,18 @@ CxPlatThreadCreate(
 
 #else // CXPLAT_USE_CUSTOM_THREAD_CONTEXT
 
-    if (pthread_create(Thread, &Attr, Config->Callback, Config->Context)) {
+    //
+    // If pthread_create fails eith ENOKEY, then try again without the attribute
+    // because the CPU might be offline.
+    //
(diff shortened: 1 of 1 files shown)


src/platform/platform_posix.c
```diff
@@ -658,6 +658,12 @@ CxPlatThreadCreate(
             "[ lib] ERROR, %u, %s.",
             Status,
             "pthread_create failed");
+
+        // Try it again without attributes.
+        // pthread_create can fail with ENOKEY if we request affinity on CPU that is offline
+        if (Status == ENOKEY && pthread_create(Thread, NULL, Config->Callback, Config->Context) == 0) {
+            Status = QUIC_STATUS_SUCCESS;
```
nibanks, line 665: We should still be checking `errno` just in case it is failing for some other reason.

wfurt: I can return new errno in `Status` or I can spit out another trace message. Any preference?

nibanks: How about change the code to the following?
```c
    //
    // If pthread_create fails eith ENOKEY, then try again without the attribute
    // because the CPU might be offline.
    //
    if (pthread_create(Thread, &Attr, CxPlatThreadCustomStart, CustomContext) &&
        errno != ENOKEY || pthread_create(Thread, NULL, CxPlatThreadCustomStart, CustomContext)) {
        Status = errno;
        QuicTraceEvent(
            LibraryErrorStatus,
            "[ lib] ERROR, %u, %s.",
            Status,
            "pthread_create failed");
        CXPLAT_FREE(CustomContext, QUIC_POOL_CUSTOM_THREAD);
    }
```

nibanks: (well, I copied the wrong version, but you get the idea)

nibanks:
```c
    //
    // If pthread_create fails eith ENOKEY, then try again without the attribute
    // because the CPU might be offline.
    //
    if (pthread_create(Thread, &Attr, Config->Callback, Config->Context) &&
        (errno != ENOKEY || pthread_create(Thread, NULL, Config->Callback, Config->Context))) {
        Status = errno;
        QuicTraceEvent(
            LibraryErrorStatus,
            "[ lib] ERROR, %u, %s.",
            Status,
            "pthread_create failed");
    }
```

wfurt:
```
  Discovered:  System.Net.Quic.Functional.Tests (found 101 of 107 test cases)
  Starting:    System.Net.Quic.Functional.Tests (parallel test collections = on, max threads = 4)
    System.Net.Quic.Tests.MsQuicPlatformDetectionTests.UnsupportedPlatforms_ThrowsPlatformNotSupportedException [SKIP]
      Condition(s) not met: "IsQuicUnsupported"
  Finished:    System.Net.Quic.Functional.Tests
=== TEST EXECUTION SUMMARY ===
   System.Net.Quic.Functional.Tests  Total: 320, Errors: 0, Failed: 0, Skipped: 1, Time: 100.274s
```

csujedihy (dismissed):

src/platform/platform_posix.c
```diff
@@ -651,7 +651,12 @@ CxPlatThreadCreate(
 
 #else // CXPLAT_USE_CUSTOM_THREAD_CONTEXT
 
-    if (pthread_create(Thread, &Attr, Config->Callback, Config->Context)) {
+    //
+    // If pthread_create fails eith ENOKEY, then try again without the attribute
```
nibanks, line 655:
```suggestion
    // If pthread_create fails with ENOKEY, then try again without the attribute
```

nibanks (approved):

PR #2992, opened 2022-08-18
Platform Loop Refactor

nibanks:
## Description

Fundamentally refactors the platform worker loop to own the "queue" that is used to drive the datapath.

## Testing

Automated CI

## Documentation

None yet.


scripts/performance.ps1
```diff
@@ -271,7 +271,7 @@ $RemotePlatform = Invoke-TestCommand -Session $Session -ScriptBlock {
     }
 }
 
-$OutputDir = Join-Path $RootDir "artifacts/PerfDataResults/$RemotePlatform/$($RemoteArch)_$($Config)_$($RemoteTls)"
+$OutputDir = Join-Path $RootDir "artifacts/PerfDataResults/$RemotePlatform/$($RemoteArch)_$($Config)_$($RemoteTls)$($ExtraArtifactDir)"
```
nibanks, line 274: FYI, this makes it so several PGO runs can run in parallel and not overwrite each other.

src/platform/datapath_winuser.c
```diff
@@ -4080,198 +4034,65 @@ CxPlatSocketSend(
...
+    case CXPLAT_CQE_TYPE_SOCKET_SHUTDOWN: {
+        CXPLAT_SOCKET_PROC* SocketProc =
+            CONTAINING_RECORD(CxPlatCqeUserData(Cqe), CXPLAT_SOCKET_PROC, ShutdownSqe);
+        CxPlatSocketContextUninitializeComplete(SocketProc);
+        break;
+    }
+    case CXPLAT_CQE_TYPE_SOCKET_IO: {
+        CXPLAT_SOCKET_PROC* SocketProc =
+            CONTAINING_RECORD(CxPlatCqeUserData(Cqe), CXPLAT_SOCKET_PROC, IoSqe);
+        if (Cqe->lpOverlapped == &SocketProc->IoSqe.Sqe) {
```
maolson-msft, line 4088: I don't understand this check. Should "IoSqe" have "recv" in the name or something?

nibanks: The only way to know if it's a send completion or a receive completion is by the OVERLAPPED returned. Receive completions all point to the socket processor's OVERLAPPED. Send completions are per-send. So if it's not `&SocketProc->IoSqe.Sqe` it must be send.

As to the name, the "user data" for both still returns `&SocketProc->IoSqe`, so it's general because of that.

The intricacies of the various network IO models is insane...

src/core/worker.c
```diff
@@ -621,12 +620,11 @@ QuicWorkerLoopCleanup(
...
 QuicWorkerLoop(
-    _Inout_ struct CXPLAT_EXECUTION_CONTEXT* Context,
-    _Inout_ uint64_t* TimeNow,
-    _In_ CXPLAT_THREAD_ID ThreadID
+    _Inout_ void* Context,
+    _Inout_ CXPLAT_EXECUTION_STATE* State
     )
 {
-    QUIC_WORKER* Worker = (QUIC_WORKER*)Context->Context;
+    QUIC_WORKER* Worker = (QUIC_WORKER*)Context;
```
ami-GS, line 627: This is not `CXPLAT_EXECUTION_CONTEXT*` ?

src/platform/datapath_winuser.c
maolson-msft: ......k

scripts/performance.ps1
anrossi: That is very handy! I'm gonna include that in my PR.

src/core/worker.c
nibanks: It's both.

src/platform/platform_worker.c
```diff
@@ -15,32 +15,53 @@
...
+    CXPLAT_THREAD Thread;
 
     //
-    // Thread used to drive the worker.
+    // Event queue to drive execution.
     //
-    CXPLAT_THREAD Thread;
+    CXPLAT_EVENTQ EventQ;
 
+#ifdef CXPLAT_SQE
```
maolson-msft, line 49: Above you check if CXPLAT_SQE_INIT is defined for defining InitializedShutdownSqe etc, but here you check if CXPLAT_SQE is defined for having the ShutdownSqe etc fields. But then I see usage of ShutdownSqe gated on definition of CXPLAT_SQE_INIT down in e.g. CxPlatWorkersInit.

Is there not a single symbol you can gate this stuff on?

nibanks: No. Some platforms don't require it, some require it just for working space (no init required) and some require it to be initialized.

src/platform/datapath_epoll.c
```diff
@@ -113,9 +110,9 @@ typedef struct CXPLAT_DATAPATH_RECV_BLOCK {
 
 typedef struct CXPLAT_SEND_DATA {
```
ami-GS, line 111: Just shuffling variables and adding comments?

maolson-msft:
In addition to CxPlatSchedulerYield, quic_platform_posix.h also defines a macro called YieldProcessor, which is unused. Let's delete YieldProcessor.

nibanks:
> In addition to CxPlatSchedulerYield, quic_platform_posix.h also defines a macro called YieldProcessor, which is unused. Let's delete YieldProcessor.

I'd rather not make any more unnecessary changes in this PR.

maolson-msft (approved):

PR #2998, opened 2022-08-19
retry pthread_create also on ENOENT

wfurt:
## Description

This is follow-up on #2989. I was trying to debug some more failed test runs. It seems like the behavior was same e.g. we fail to created thread when trying to set on offline CPI but the `errno` was set to `ENOENT`. I did not figure out why it would sometimes fail with different error. 
For now I added `pthread_create` retry also for other errno code. 

## Testing

running .NET test suite.

### DIFF
--- a/src/platform/platform_posix.c
+++ b/src/platform/platform_posix.c
@@ -652,11 +652,11 @@ CxPlatThreadCreate(
 #else // CXPLAT_USE_CUSTOM_THREAD_CONTEXT
 
     //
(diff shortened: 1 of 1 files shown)


csujedihy (approved):
We could also just blindly retry with no attr when we see failure the first time.

csujedihy:
Let me know if you wanna just check it in, I can hit the merge button for you.

wfurt:
yes, please. I'll give it another try build our test containers.  arm failure seems unrelated.

PR #3026, opened 2022-08-28
use cpuid to take more detail processor info

ami-GS:
## Description

- CPUID to take HyperThreading enabled flag
- Thread per core is not necessarily 2

## Documentation

Will do for libhwloc installation

### DIFF
--- a/src/inc/CMakeLists.txt
+++ b/src/inc/CMakeLists.txt
@@ -40,6 +40,7 @@ target_link_libraries(base_link INTERFACE Threads::Threads ${CMAKE_DL_LIBS})
 if (ATOMIC)
     target_link_libraries(base_link INTERFACE ${ATOMIC})
 endif()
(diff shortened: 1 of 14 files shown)


src/inc/CMakeLists.txt
```diff
@@ -40,6 +40,7 @@ target_link_libraries(base_link INTERFACE Threads::Threads ${CMAKE_DL_LIBS})
 if (ATOMIC)
     target_link_libraries(base_link INTERFACE ${ATOMIC})
 endif()
+target_link_libraries(base_link INTERFACE "/usr/lib/x86_64-linux-gnu/libhwloc.so")
```
csujedihy, line 43: this will break arm/arm64 build

ami-GS: arm?
not windows from path point of view?

PR #3034, opened 2022-08-30
App Controlled Execution Config

nibanks:
## Description

Refactors the MsQuic API to allow for the application to explicitly configure the execution of MsQuic internals. This is the next step to allow the app to fully control execution.

## Testing

Existing CI

## Documentation

to-do


src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -96,9 +92,10 @@ PrintHelp(
         "  -cc:<algo>                  Congestion control algorithm to use.\n"
         "\n"
         "Client: secnetperf -TestName:<Throughput|RPS|HPS> [options]\n"
-#ifndef _KERNEL_MODE
         "Both:\n"
-        "  -cpu:<cpu_index>            Specify the processor(s) for the datapath to use.\n"
+        "  -cpu:<cpu_index>            Specify the processor(s) to use.\n"
```
nibanks, line 96: Specify `sleepthresh`

src/core/registration.c
```diff
@@ -58,39 +54,15 @@ MsQuicRegistrationOpen(
...
-                    LibraryErrorStatus,
-                    "[ lib] ERROR, %u, %s.",
-                    Status,
-                    "CxPlatDataPathInitialize");
-                goto Error;
-            }
-            QuicTraceEvent(
-                DataPathInitialized,
-                "[data] Initialized, DatapathFeatures=%u",
-                CxPlatDataPathGetSupportedFeatures(MsQuicLib.Datapath));
```
csujedihy, line 89: This refactor seems unnecessary to me. QuicLibraryEnsureExecutionContext is essentially just initialize datapath.

src/platform/platform_worker.c
```diff
@@ -76,49 +62,74 @@ typedef struct QUIC_CACHEALIGN CXPLAT_WORKER {
...
 #pragma warning(disable:6386) // SAL is confused about the worker size
 BOOLEAN
-CxPlatWorkersInit(
-    void
+CxPlatWorkersLazyStart(
+    _In_opt_ QUIC_EXECUTION_CONFIG* Config
     )
 {
-    CxPlatWorkerCount = CxPlatProcActiveCount(); // TODO - use max instead?
+    CxPlatLockAcquire(&CxPlatWorkerLock);
```
csujedihy, line 106: >CxPlatWorkerLock [](http://example.com/codeflow?start=23&length=16)

Is this lock necessary? It's only used in CxPlatWorkersLazyStart. I can't see how CxPlatWorkersLazyStart can be called from different threads.

src/core/registration.c
nibanks: As it is now, I'd agree with you. But in an effort to break this PR up some, I haven't made the QUIC layer changes to only run on the configured CPUs. That's the next step. With that work, a lot more will change in library.c around PartitionCount, so more will be added then.

src/platform/platform_worker.c
nibanks: The platform layer is treated as a semi-public API, that could possibly be used my multiple consumers. Indeed, if you compile our tools with static linking, then the platform does end up getting shared by MsQuic and the tool itself. So just to be extra safe, we have synchronization.

scripts/RemoteTests.json
```diff
@@ -279,7 +279,7 @@
                 "Tls": ["openssl"],
                 "Arch": ["x64", "arm"],
                 "Exe": "secnetperf",
-                "Arguments": "-test:HPS -target:$RemoteAddress"
+                "Arguments": "-exec:lowlat -test:HPS -target:$RemoteAddress"
```
nibanks, line 282: Maybe `maxtput` for this one?

scripts/RemoteTests.json
```diff
@@ -172,7 +172,7 @@
                 "Tls": ["schannel", "openssl"],
                 "Arch": ["x64", "x86", "arm", "arm64"],
                 "Exe": "secnetperf",
-                "Arguments": "-test:RPS -target:$RemoteAddress"
+                "Arguments": "-exec:lowlat -test:RPS -target:$RemoteAddress"
```
nibanks, line 175: RPS tests both throughput and latency, depending on the test. So I am unsure of what to use? Both?

src/inc/msquic.h
```diff
@@ -232,6 +232,27 @@ typedef enum QUIC_DATAGRAM_SEND_STATE {
...
+
+DEFINE_ENUM_FLAG_OPERATORS(QUIC_EXECUTION_CONFIG_FLAGS)
+
+//
+// A custom configuration for thread execution in QUIC.
+//
+typedef struct QUIC_EXECUTION_CONFIG {
+
+    QUIC_EXECUTION_CONFIG_FLAGS Flags;
+    uint32_t SleepTimeoutUs;            // Number of microseconds of idleness before sleeping a datapath thread.
```
nibanks, line 247: PollingIdleTimeoutUs

Time before a polling thread, with no work to do, sleeps.

maolson-msft (approved):

src/core/worker.c
```diff
@@ -101,29 +95,36 @@ QuicWorkerInitialize(
...
-    if (QUIC_FAILED(Status)) {
-        QuicTraceEvent(
-            WorkerErrorStatus,
-            "[wrkr][%p] ERROR, %u, %s.",
-            Worker,
-            Status,
-            "CxPlatThreadCreate");
-        goto Error;
+#ifndef _KERNEL_MODE // Not supported on kernel mode
+    if (ExecProfile != QUIC_EXECUTION_PROFILE_TYPE_MAX_THROUGHPUT) {
```
ami-GS, line 99: Does max throughput related flag works without splitpartitioning?

nibanks: Yes. Max throughput just means separate threads for UDP and QUIC now.

PR #3041, opened 2022-09-05
Always creating worker threads without affinitization on error.

rzikm:
## Description

On combination Docker+Debian11+ARM64, the call to pthread_create sometimes returns EINVAL when attempting thread affinitization. This PR always retries thread creation without thread affinitization when the first call to pthread_create fails.

Related to https://github.com/dotnet/runtime/issues/74952

## Testing

Tested manually on the affected platform.

## Documentation

NO

### DIFF
--- a/src/platform/platform_posix.c
+++ b/src/platform/platform_posix.c
@@ -665,11 +665,11 @@ CxPlatThreadCreate(
 #else // CXPLAT_USE_CUSTOM_THREAD_CONTEXT
 
     //
-    // If pthread_create fails with ENOKEY or ENOENT, then try again without the attribute
+    // If pthread_create fails with an error code, then try again without the attribute
     // because the CPU might be offline.
     //
     if (pthread_create(Thread, &Attr, Config->Callback, Config->Context) &&
-        ((errno != ENOKEY && errno != ENOENT) || pthread_create(Thread, NULL, Config->Callback, Config->Context))) {
+        pthread_create(Thread, NULL, Config->Callback, Config->Context)) {
         Status = errno;
         QuicTraceEvent(
             LibraryErrorStatus,
--- a/src/platform/platform_posix.c
+++ b/src/platform/platform_posix.c
@@ -669,11 +669,9 @@ CxPlatThreadCreate(
     // because the CPU might be offline.
     //
     if (pthread_create(Thread, &Attr, Config->Callback, Config->Context)) {
-        QuicTraceEvent(
-            LibraryErrorStatus,
-            "[ lib] WARNING, %u, %s.",
-            errno,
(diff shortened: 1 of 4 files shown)


src/platform/platform_posix.c
```diff
@@ -665,11 +665,11 @@ CxPlatThreadCreate(
 #else // CXPLAT_USE_CUSTOM_THREAD_CONTEXT
 
     //
-    // If pthread_create fails with ENOKEY or ENOENT, then try again without the attribute
+    // If pthread_create fails with an error code, then try again without the attribute
     // because the CPU might be offline.
     //
     if (pthread_create(Thread, &Attr, Config->Callback, Config->Context) &&
```
nibanks, line 671: Since we're always going to try the fallback, let's log a Warning, if we did the fallback and it succeeded. And an error if both failed.

src/platform/platform_posix.c
```diff
@@ -665,17 +665,23 @@ CxPlatThreadCreate(
...
-        Status = errno;
+    if (pthread_create(Thread, &Attr, Config->Callback, Config->Context)) {
         QuicTraceEvent(
             LibraryErrorStatus,
-            "[ lib] ERROR, %u, %s.",
-            Status,
-            "pthread_create failed");
+            "[ lib] WARNING, %u, %s.",
+            errno,
+            "pthread_create failed, retrying without affinitization");
```
rzikm, line 676: I have a feeling this requires rerunning some tool to update some autogenerated files, what do I need to do? @nibanks?

nibanks:
```c
                QuicTraceLogWarning(
                    PlatformThreadCreateFailed,
                    "[ lib] pthread_create failed, retrying without affinitization");
```
Then run `.\scripts\update-sidecar.ps1`.

karelz:
@nibanks can we get this change in msquic servicing branch release/2.1?

We need to be able to determine version of msquic library (and which which backend we use - OpenSSL or Schannel) without inflicting the cost of keeping msquic threads in apps that do not use msquic. Currently all the threads are consuming some memory and they are confusing users who do deep dive into perf of their app.
We are ok with loading msquic and then unloading it after we get the info, but we need to be able to load it again if msquic is needed by the app. This fix will unblock us to load it for 2nd time.

nibanks:
> @nibanks can we get this change in msquic servicing branch release/2.1?

Yes, that should be fine.

nibanks (approved):

nibanks:
@rzikm would you mine creating a PR to cherry-pick this into the release/2.1 branch?

rzikm:
@karelz I filed https://github.com/microsoft/msquic/issues/3044

PR #3043, opened 2022-09-06
Update Patch Version Number

nibanks:
Updates various files to reflect latest patch version.


anrossi:
Is there a way to consolidate these? Or do we want each to have their own copy of the version number so they can diverge?

anrossi (approved):

nibanks:
> Is there a way to consolidate these? Or do we want each to have their own copy of the version number so they can diverge?

Unfortunately not.

PR #3064, opened 2022-09-14
Fix/suppress some new clang/cppcheck warnings

larseggert:
## Description

Fix/suppress some new clang/cppcheck warnings

## Testing

Should be covered by existing tests.

## Documentation

Nope.

### DIFF
--- a/src/core/api.c
+++ b/src/core/api.c
@@ -521,7 +521,7 @@ MsQuicConnectionSendResumptionTicket(
         goto Error;
     }
 
-    if (Flags > (QUIC_SEND_RESUMPTION_FLAG_FINAL | QUIC_SEND_RESUMPTION_FLAG_NONE)) { // cppcheck-suppress badBitmaskCheck
+    if (Flags > QUIC_SEND_RESUMPTION_FLAG_FINAL) {
         Status = QUIC_STATUS_INVALID_PARAMETER;
         goto Error;
     }
--- a/src/core/api.c
+++ b/src/core/api.c
@@ -521,7 +521,7 @@ MsQuicConnectionSendResumptionTicket(
         goto Error;
     }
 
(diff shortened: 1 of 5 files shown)


src/platform/hashtable.c
```diff
@@ -841,7 +841,7 @@ Return Value:
         Signature = CXPLAT_HASH_ALT_SIGNATURE;
     }
 
-    CXPLAT_HASHTABLE_LOOKUP_CONTEXT LocalContext;
+    CXPLAT_HASHTABLE_LOOKUP_CONTEXT LocalContext = {0};
```
nibanks, line 844: I don't think this one is truly necessary. Also, since this is in the hot path, I'm inclined to ignore this one, rather than this change. Thoughts?

src/core/api.c
```diff
@@ -521,7 +521,7 @@ MsQuicConnectionSendResumptionTicket(
         goto Error;
     }
 
-    if (Flags > (QUIC_SEND_RESUMPTION_FLAG_FINAL | QUIC_SEND_RESUMPTION_FLAG_NONE)) {
+    if (Flags > (QUIC_SEND_RESUMPTION_FLAG_FINAL | QUIC_SEND_RESUMPTION_FLAG_NONE)) { // cppcheck-suppress badBitmaskCheck
```
nibanks, line 524: Really, the `| QUIC_SEND_RESUMPTION_FLAG_NONE` should just be removed here, right? Instead of ignoring?

nibanks:
Friendly ping. Any idea on an ETA for finishing this up?

larseggert:
I updated this.

Note that it doesn't build for me at the moment on MacOS:
```
/Users/lars/Documents/Code/msquic/src/platform/datapath_kqueue.c:36:0: error: #error "No socket option specified" [preprocessorErrorDirective]
#error "No socket option specified"
^
```

nibanks (approved):

PR #3065, opened 2022-09-14
Use correct length increment for IP_PKTINFO CMsg header

rzikm:
## Description

On Arm32 devices, we encountered an issue where all attemts at connection timeout, closer investigation showed that all `sendmmsg` calls were failing with errno 22. I dug around a bit and it seems that the platform in question (Debian 10 on arm) was checking the expected length against `IP_PKTINFO` control header and failed the send. Using correct length for IPv4 and IPv6 solved the issue.

More info at https://github.com/dotnet/runtime/issues/75493#issuecomment-1246556769

## Testing

Tested on affected platform.

## Documentation

N/A

### DIFF
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -2412,7 +2412,9 @@ CxPlatSocketSendInternal(
         *(int *)CMSG_DATA(CMsg) = SendData->ECN;
 
         if (!Socket->Connected) {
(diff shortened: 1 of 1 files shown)


nibanks (approved):
Very interesting! Good find! Do you know if there is an easy way to onboard tests (in MsQuic) for this scenario. I'd like to make sure we don't regress anything over time.

wfurt:
nice. may be related to alignment...

PR #3066, opened 2022-09-15
Add vcpkg installation instructions

Cheney-W:
`MsQuic` available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for `MsQuic` and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build `MsQuic`, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/ms-quic/portfile.cmake). We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/docs/BUILD.md
+++ b/docs/BUILD.md
@@ -238,3 +238,15 @@ To run the tests:
 ```
 cargo test
 ```
+
+# Installing from vcpkg
+
+You can download and install `MsQuic` using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
+```sh
+git clone https://github.com/Microsoft/vcpkg.git
+cd vcpkg
+./bootstrap-vcpkg.sh #.\bootstrap-vcpkg.bat(for windows)
+./vcpkg integrate install
+./vcpkg install ms-quic
+```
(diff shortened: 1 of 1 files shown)


nibanks:
@Cheney-W it looks like vcpkg is way out of date (referenced v1.2.0, where the latest is v2.1.1) and it has a number of patches that modify the base? Also, how do we consume and test/validate the vcpkg?

nibanks (approved):

Cheney-W:
Thanks for opening issue in vcpkg. We will update this port ASAP.
About this issue: `how do we consume and test/validate the vcpkg? ` 
Do you want to know how to use the MsQuic generated by vcpkg? Sorry, I'm not very clear what you mean.

nibanks:
Yes, essentially, how do we use it to test it?

Cheney-W:
After you installed the port, for example `ms-quic:x64-windows`: 
Run `./vcpkg integrate install`
Then open VS and create an empty project, set the platform as x64, then you can directly include the header file of ms-quic in this project, call the function, and no longer need to manually set the path of the header file or library file.

If you want to use this port in cmake project, please refer [here](https://github.com/microsoft/vcpkg/blob/master/docs/examples/installing-and-using-packages.md#cmake-toolchain-file).

PR #3068, opened 2022-09-15
Fix QUIC_STATUS_ALPN_IN_USE value in comments and C# interop

rzikm:
## Description

This PR fixes the value of EPROTOTYPE (aliased as QUIC_STATUS_ALPN_IN_USE), on Linux, it seems to be 91 (see e.g. https://www.thegeekstuff.com/2010/10/linux-error-codes/), and on MacOS 41 (https://opensource.apple.com/source/xnu/xnu-201/bsd/sys/errno.h). This PR also fixes the C# interop stubs.

## Testing

New values tested on Linux in CI when integrating the error status in .NET.

## Documentation

N/A


nibanks (approved):

nibanks:
Will this need a backport?

PR #3076, opened 2022-09-22
Add tests for invalid-sized version_info transport parameter.

anrossi:
## Description

Add tests to validate behavior around invalid-sized version_info transport parameter.  Ensure on validation failure that the connection is closed and the correct error is returned.
Fixed some incorrect behavior discovered during the testing.

## Testing

Test change.

## Documentation

No documentation changes necessary.

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -5390,15 +5390,15 @@ QuicConnRecvDatagramBatch(
         if (!QuicConnRecvPrepareDecrypt(
                 Connection, Packet, HpMask + i * CXPLAT_HP_SAMPLE_LENGTH) ||
             !QuicConnRecvDecryptAndAuthenticate(Connection, Path, Packet)) {
-                if (Connection->State.CompatibleVerNegotiationAttempted &&
-                    !Connection->State.CompatibleVerNegotiationCompleted) {
-                    //
-                    // The packet which initiated compatible version negotation failed
-                    // decryption, so undo the version change.
-                    //
-                    Connection->Stats.QuicVersion = Connection->OriginalQuicVersion;
-                    Connection->State.CompatibleVerNegotiationAttempted = FALSE;
-                }
+            if (Connection->State.CompatibleVerNegotiationAttempted &&
+                !Connection->State.CompatibleVerNegotiationCompleted) {
+                //
+                // The packet which initiated compatible version negotation failed
+                // decryption, so undo the version change.
+                //
+                Connection->Stats.QuicVersion = Connection->OriginalQuicVersion;
+                Connection->State.CompatibleVerNegotiationAttempted = FALSE;
+            }
         } else if (QuicConnRecvFrames(Connection, Path, Packet, ECN)) {
 
             QuicConnRecvPostProcessing(Connection, &Path, Packet);
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -5387,13 +5387,19 @@ QuicConnRecvDatagramBatch(
         CXPLAT_ECN_TYPE ECN = CXPLAT_ECN_FROM_TOS(Datagrams[i]->TypeOfService);
         Packet = CxPlatDataPathRecvDataToRecvPacket(Datagrams[i]);
         CXPLAT_DBG_ASSERT(Packet->PacketId != 0);
-        if (QuicConnRecvPrepareDecrypt(
-                Connection, Packet, HpMask + i * CXPLAT_HP_SAMPLE_LENGTH) &&
-            QuicConnRecvDecryptAndAuthenticate(Connection, Path, Packet)) {
-
-            if (!QuicConnRecvFrames(Connection, Path, Packet, ECN)) {
-                continue;
-            }
+        if (!QuicConnRecvPrepareDecrypt(
+                Connection, Packet, HpMask + i * CXPLAT_HP_SAMPLE_LENGTH) ||
+            !QuicConnRecvDecryptAndAuthenticate(Connection, Path, Packet)) {
+                if (Connection->State.CompatibleVerNegotiationAttempted &&
+                    !Connection->State.CompatibleVerNegotiationCompleted) {
+                    //
+                    // The packet which initiated compatible version negotation failed
+                    // decryption, so undo the version change.
+                    //
+                    Connection->Stats.QuicVersion = Connection->OriginalQuicVersion;
+                    Connection->State.CompatibleVerNegotiationAttempted = FALSE;
+                }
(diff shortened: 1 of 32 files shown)


src/core/connection.h
```diff
@@ -630,6 +630,14 @@ typedef struct QUIC_CONNECTION {
         QUIC_FLOW_BLOCKED_TIMING_TRACKER FlowControl;
     } BlockedTimings;
 
+#ifdef DEBUG
+    //
+    // Whether to disable automatic generation of VNE transport parameter.
+    // Only used for testing, and thus only enabled for debug builds.
+    //
+    BOOLEAN DisableVneTp;
```
nibanks, line 638: Let's put this at the end of `QUIC_CONNECTION_STATE`

src/core/connection.c
```diff
@@ -5385,8 +5389,11 @@ QuicConnRecvDatagramBatch(
         CXPLAT_DBG_ASSERT(Packet->PacketId != 0);
         if (QuicConnRecvPrepareDecrypt(
                 Connection, Packet, HpMask + i * CXPLAT_HP_SAMPLE_LENGTH) &&
-            QuicConnRecvDecryptAndAuthenticate(Connection, Path, Packet) &&
-            QuicConnRecvFrames(Connection, Path, Packet, ECN)) {
+            QuicConnRecvDecryptAndAuthenticate(Connection, Path, Packet)) {
+
+            if (!QuicConnRecvFrames(Connection, Path, Packet, ECN)) {
+                continue;
+            }
```
nibanks, line 5396: I feel like there must be a better way to clean this code up. Perhaps moving the `else` code below (line 5418) somewhere else?

src/core/crypto_tls.c
```diff
@@ -1743,34 +1743,29 @@ QuicCryptoTlsDecodeTransportParameters( // NOLINT(readability-function-size, goo
...
-                TransportParams->Flags |= QUIC_TP_FLAG_VERSION_NEGOTIATION;
+            if (Length > 0) {
+                TransportParams->VersionInfo = CXPLAT_ALLOC_NONPAGED(Length, QUIC_POOL_VERSION_INFO);
+                if (TransportParams->VersionInfo == NULL) {
+                    QuicTraceEvent(
+                        AllocFailure,
+                        "Allocation of '%s' failed. (%llu bytes)",
+                        IsServerTP ?
+                            "Received Client Version Negotiation Info" :
+                            "Received Server Version Negotiation Info",
```
nibanks, line 1754: Let's just simplify this. No need to say client/server.

src/core/connection.c
anrossi: The issue is that we need to rollback compatible version negotiation on decryption failure. The `QuicConnRecvFrames` failure path here convolutes the failure path.

src/core/crypto_tls.c
```diff
@@ -1743,34 +1743,29 @@ QuicCryptoTlsDecodeTransportParameters( // NOLINT(readability-function-size, goo
             break;
 
         case QUIC_TP_ID_VERSION_NEGOTIATION_EXT:
-            if (Length < MIN_VERSION_INFO_LENGTH) {
```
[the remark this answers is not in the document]
anrossi: When the failure occurs here, the wrong error is sent back to the peer (handshake failure/crypto error). Delaying until the version_info is actually parsed to generate the error involves far less refactoring than trying to make this path return the correct error, when in all other cases, the current error is correct.

src/core/crypto_tls.c
anrossi: It does make it easier to read/search the log files for the specific issue. Otherwise, it doesn't say which side is allocating

nibanks: But this is an alloc failure. It's not something you're likely to ever see in a real log, right?

src/inc/msquicp.h
```diff
@@ -87,7 +87,7 @@ typedef struct QUIC_TEST_DATAPATH_HOOKS {
 #endif
 
 typedef struct QUIC_PRIVATE_TRANSPORT_PARAMETER {
-    uint16_t Type;
+    uint32_t Type;
```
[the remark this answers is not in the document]
anrossi: the ID for VNE is `0xFF73DB` which doesn't fit in 2 bytes. Why does it have the leading `0xFF` if it just gets truncated?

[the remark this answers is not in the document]
nibanks: Ah, you're right. For some reason I thought this was a length field. Carry on!

src/core/connection.c
anrossi: I've been thinking about this, and the options are having the code in two places (I don't like this because it could lead to bugs in the future), or having a temporary variable in the if-statement.  
Are you more concerned about readability or the performance of this code path?

nibanks: More for readability.

anrossi: how's this for readability?

anrossi:
Should `QUIC_ERROR_TRANSPORT_PARAMETER_ERROR` be mapped to `QUIC_STATUS_INTERNAL_ERROR` or should a new `QUIC_STATUS` be created?

nibanks:
> Should `QUIC_ERROR_TRANSPORT_PARAMETER_ERROR` be mapped to `QUIC_STATUS_INTERNAL_ERROR` or should a new `QUIC_STATUS` be created?

Just keep it as internal. No new error needed.

src/core/connection.c
```diff
@@ -5383,10 +5387,19 @@ QuicConnRecvDatagramBatch(
...
         Packet = CxPlatDataPathRecvDataToRecvPacket(Datagrams[i]);
         CXPLAT_DBG_ASSERT(Packet->PacketId != 0);
-        if (QuicConnRecvPrepareDecrypt(
-                Connection, Packet, HpMask + i * CXPLAT_HP_SAMPLE_LENGTH) &&
-            QuicConnRecvDecryptAndAuthenticate(Connection, Path, Packet) &&
-            QuicConnRecvFrames(Connection, Path, Packet, ECN)) {
+        if (!QuicConnRecvPrepareDecrypt(
+                Connection, Packet, HpMask + i * CXPLAT_HP_SAMPLE_LENGTH) ||
+            !QuicConnRecvDecryptAndAuthenticate(Connection, Path, Packet)) {
+                if (Connection->State.CompatibleVerNegotiationAttempted &&
```
nibanks, line 5393: Tab left please

nibanks (approved):

PR #3080, opened 2022-09-26
Remove Verifier on System Components

nibanks:
## Description

It seems the latest Windows vNext builds errors out on system verifier stuff. Try to work around that.

## Testing

Automated CI/CD.

## Documentation

N/A

PR #3085, opened 2022-09-27
Change Test TP to use 4 bytes for Type instead of 2.

anrossi:
## Description

Change the test transport parameter to use 4 bytes for `Type` instead of 2 bytes, allowing for the version negotiation extension transport parameter to be supplied.  This is needed for #3076.

## Testing

CI tests

## Documentation

N/A


nibanks (approved):
This will need to be backported to all `release/2.*` branches, including rev'ing the patch version, so that we can create new releases, for the new test binaries to publish. While we're at it #3080 should be done at the same time.

PR #3038, opened 2022-09-01
add EADDRNOTAVAIL on msquic header file

pedrobsaila:
## Description

Follow up PR for https://github.com/dotnet/runtime/issues/74053. As mentioned by @rzikm, I should add EADDRNOTAVAIL to header file

PR #3093, opened 2022-09-30
add windows QUIC_STATUS_ADDRESS_NOT_AVAILABLE

pedrobsaila:
## Description

Follow up PR for https://github.com/dotnet/runtime/issues/74053. As mentioned by @rzikm (https://github.com/microsoft/msquic/pull/3038#issuecomment-1235157740 and https://github.com/dotnet/runtime/pull/74780#discussion_r984356296), I should add a windows version of QUIC_STATUS_ADDRESS_NOT_AVAILABLE


src/cs/lib/msquic_generated_windows.cs
```diff
@@ -114,6 +114,9 @@ internal static partial class MsQuic_Windows
         [NativeTypeName("#define QUIC_STATUS_CERT_NO_CERT SEC_E_NO_CREDENTIALS")]
         public const int QUIC_STATUS_CERT_NO_CERT = unchecked((int)(0x8009030E));
 
+        [NativeTypeName("#define QUIC_STATUS_ADDRESS_NOT_AVAILABLE QUIC_STATUS_HRESULT_FROM_WIN32(WSAENETUNREACH)")]
+        public const int QUIC_STATUS_ADDRESS_NOT_AVAILABLE = unchecked((int)(10051) <= 0 ? ((int)(10051)) : ((int)(((10051) & 0x0000FFFF) | (7 << 16) | 0x80000000)));
```
nibanks, line 118: What's the point of the `<= 0` check here?

src/inc/msquic_winuser.h
```diff
@@ -116,6 +116,7 @@
 #define QUIC_STATUS_ALPN_NEG_FAILURE        ERROR_QUIC_ALPN_NEG_FAILURE                     // 0x80410007
 #define QUIC_STATUS_STREAM_LIMIT_REACHED    ERROR_QUIC_STREAM_LIMIT_REACHED                 // 0x80410008
 #define QUIC_STATUS_ALPN_IN_USE             ERROR_QUIC_ALPN_IN_USE                          // 0x80410009
+#define QUIC_STATUS_ADDRESS_NOT_AVAILABLE   HRESULT_FROM_WIN32(WSAENETUNREACH)              // 0x80072743
```
nibanks, line 119: We already have `#define QUIC_STATUS_INVALID_ADDRESS         HRESULT_FROM_WIN32(WSAEADDRNOTAVAIL)`. I'm not so sure about this change.

pedrobsaila: Is the naming QUIC_STATUS_ADDRESS_NOT_AVAILABLE that bothers you or WSAENETUNREACH as windows error code ?

nibanks: Yes, I don't like that the names don't match. I don't know which name should be preferred, because I am not sure what exactly they are being used for. I'd prefer if there was a QUIC test that explained/validated them.

pedrobsaila:
> Friendly ping. Any updates here? We really should have a test case driving this. I don't know where this error is supposed to come from (what QUIC scenario), so I don't know how to pick a good name.

Sorry for the delay. I'm still working on it. I'm having some trouble writing the test but I'll figure it out

wfurt:
Yes, that is correct @pedrobsaila. However we will still need consistent status code surfacing from MsQuic. e.g. the type of exception is irrelevant to this PR IMHO.

pedrobsaila:
> However we will still need consistent status code surfacing from MsQuic

Here are some points of view about adding the mapping for windows (check out https://www.youtube.com/watch?v=73BAn9B3ni4&t=1011s ) :

* it's okay if you know the error is induced by different conditions on different OS like that's just you know the way of the world. We should try to unify them when possible but sometimes we simply can't.
* you could just have two different errors if you know that there is an error that is caused by the address family not being available. There is no need to force it into a single value and maybe it just so happens that on Windows we will never display the address family not available error if we can't know for sure that was the underlying cause.
* it seems like we've started going down a path where basically trying to mimic a networking stack (of linux) even though we don't necessarily know why we're doing it. If the entire point is to make logging better and more detailed like exception is already rich enough
* this is doing very fuzzy mapping .. so now you need to be a PhD in networking to understand what is the difference between the two is (EAFNOSUPPORT and WSAENETUNREACH)

No one is sure what is the mapping for the POSIX error EAFNOSUPPORT in Windows, we just proved it experimentally : we might missed other cases. So merging this PR is msquic is not a good idea for the same reasons that our new enum feature was not approved.

> the type of exception is irrelevant to this PR IMHO.

I'm not sure to understand this part.

pedrobsaila:
Your proposition seems OK. Just have one question : I'm not sure of the code WSAEAFNOSUPPORT for windows kernel. Is it RPC_NT_INVALID_NAF_ID (0xC0020040) ?

wfurt:
```
C:\Program Files (x86)\Windows Kits\10\Include/10.0.18362.0\shared\winerror.h:#define WSAEAFNOSUPPORT                  10047L
```

It is network subsystem error.
`SocketError` conveniently uses same value since .NET Framework was Windows only.

pedrobsaila:
sorry I didn't describe clearly my intent : [msquic_winkernel.h](https://github.com/microsoft/msquic/blob/99a7f5a53c322ba75d8a7b233cf87773757f43c8/src/inc/msquic_winkernel.h#L4) this file contains the equivalent [NTStatus](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/596a1078-e883-4972-9bbc-49e60bebca55) error of the standard winsock errors. I was asking what is the equivalent NTStatus error for WSAEAFNOSUPPORT

wfurt:
That may be ok. Alternatively I would pick one from following two:
```
//
// MessageId: STATUS_INVALID_ADDRESS
//
// MessageText:
//
// The address handle given to the transport was invalid.
//
#define STATUS_INVALID_ADDRESS           ((NTSTATUS)0xC0000141L)

//
// MessageId: STATUS_INVALID_ADDRESS_COMPONENT
//
// MessageText:
//
// The transport rejected the network address specified as invalid.
//
#define STATUS_INVALID_ADDRESS_COMPONENT ((NTSTATUS)0xC0000207L)
```
given "INVALID" -> "not applicable" or "not available". As long as we surface something reasonably relevant and unique I think we would be OK.

src/inc/msquic_winuser.h
pedrobsaila: Sorry I didn't find a way to programmatically deactivate/reactivate IPv6. If you had any idea on how to achieve it I would be thankfull

wfurt: You can do via registry AFAIK but that works only for external targets and it requires admin access e.g. it is not good candidate IMHO for automated tests

wfurt (dismissed):
LGTM. Thanks @pedrobsaila. any more thoughts on this @nibanks? This allows us to consume new status in portable way. (in practice this will only happen on Linux AFAIK)

wfurt:
some test failures looks like infra
```
Unhandled: ENOSPC: no space left on device, write
```

nibanks:
@pedrobsaila would you mind merging main so we can rerun CI? Thanks!

PR #3106, opened 2022-10-04
Disable Preview Feature Tests on Official Builds

nibanks:
## Description

Official builds need to disable the tests for preview features since those features may change in the future. We can't publish down-level tests for things that are expected to change in the future. Those tests will break.

## Testing

Ran locally. Added a CI for build. Should I onboard a test too?

## Documentation

N/A

### DIFF
--- a/.azure/azure-pipelines.ci.yml
+++ b/.azure/azure-pipelines.ci.yml
@@ -336,6 +336,15 @@ stages:
       platform: linux
       arch: x64
       tls: openssl
+      extraBuildArgs: -OfficialRelease
+      extraName: 'official'
+
+  - template: ./templates/build-config-user.yml
+    parameters:
+      image: ubuntu-latest
+      platform: linux
+      arch: x64
+      tls: openssl
       extraBuildArgs: -Clang -ExtraArtifactDir Clang
       extraName: 'clang'
 
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -276,6 +276,7 @@ if (!$OfficialRelease) {
             $OfficialRelease = $true
         }
     } catch { }
+    $LASTEXITCODE = 0
 }
 
 # Root directory of the project.
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -75,6 +75,9 @@ This script provides helpers for building msquic.
 .PARAMETER CI
     Build is occuring from CI
 
+.PARAMETER OfficialRelease
+    Build is for an official (tag) release.
+
 .PARAMETER EnableTelemetryAsserts
     Enables telemetry asserts in release builds.
 
@@ -182,6 +185,9 @@ param (
     [switch]$CI = $false,
 
     [Parameter(Mandatory = $false)]
+    [switch]$OfficialRelease = $false,
+
+    [Parameter(Mandatory = $false)]
     [switch]$EnableTelemetryAsserts = $false,
 
     [Parameter(Mandatory = $false)]
@@ -258,6 +264,11 @@ if ($Platform -eq "ios" -and !$Static) {
     $Static = $true
     Write-Host "iOS can only be built as static"
 }
+
+try {
+#    $env:GIT_REDIRECT_STDERR = '2>&1'
+#    $CurrentBranch = git branch --show-current
+} catch { }
 
 # Root directory of the project.
 $RootDir = Split-Path $PSScriptRoot -Parent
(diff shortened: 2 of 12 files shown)


nibanks:
This will need to be backported to all releases that we run the down-level tests for (release/2.0 & release/2.1).

src/test/lib/TestConnection.cpp
```diff
@@ -295,6 +295,7 @@ TestConnection::SetQuicVersion(
     uint32_t value
     )
 {
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
```
nibanks, line 298: It's unfortunate that we broke this test, which used to work with the old param. But oh well.

.azure/azure-pipelines.ci.yml
```diff
@@ -330,6 +330,15 @@ stages:
   displayName: Build Linux - Non Tested
   dependsOn: []
   jobs:
+  - template: ./templates/build-config-user.yml
```
nibanks, line 333: I _think_ build is going to be enough. Opinions?

src/test/MsQuicTests.h
```diff
@@ -9,7 +9,9 @@
 
 --*/
 
+#ifndef QUIC_OFFICIAL_RELEASE
 #define QUIC_API_ENABLE_PREVIEW_FEATURES
+#endif
```
nibanks, line 14: Somehow on Linux, it seems `QUIC_API_ENABLE_PREVIEW_FEATURES` is not getting defined, but AFAICT `QUIC_OFFICIAL_RELEASE` is not defined. I cannot figure out why.

src/test/bin/quic_gtest.cpp
```diff
@@ -702,6 +704,7 @@ TEST_P(WithFamilyArgs, InterfaceBinding) {
     }
 }
 
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
```
anrossi, line 707: This isn't a version negotiation test, AFAICT.  It just ensures the connection still succeeds when the client uses the `QUIC_VERSION_MS_1` version instead of the official versions.

.azure/azure-pipelines.ci.yml
anrossi: build is enough for what? Releasing?

src/test/MsQuicTests.h
anrossi: Is this all `QUIC_OFFICIAL_RELEASE` is used for?  I want to make sure we don't change behavior with this flag, since that'll potentially moot testing.

scripts/build.ps1
```diff
@@ -259,6 +265,20 @@ if ($Platform -eq "ios" -and !$Static) {
     Write-Host "iOS can only be built as static"
 }
 
+if (!$OfficialRelease) {
+    try {
+        $env:GIT_REDIRECT_STDERR = '2>&1'
+        # Thanks to https://stackoverflow.com/questions/3404936/show-which-git-tag-you-are-on
+        # for this magic git command!
+        $Output = git describe --exact-match --tags $(git log -n1 --pretty='%h')
```
anrossi, line 273: I think this requires the release tag to exist before the release build has even started. Also, the tag might not exist in the build machine's git repository when this is run.

.azure/azure-pipelines.ci.yml
nibanks: No. To validate this new flag that only gets enabled for release builds otherwise.

src/test/MsQuicTests.h
nibanks: That's all it's doing now.

src/test/bin/quic_gtest.cpp
nibanks: Does it depend on the ability to set a different version though? In other words, does it depend on a preview API at all?

scripts/build.ps1
nibanks: That's ok. The release tag **triggers** the official build for it. So the tag always is created first.

src/test/bin/quic_gtest.cpp
anrossi: oh, I see. The old version SetParam was removed in favor of the preview API.

.azure/azure-pipelines.ci.yml
anrossi: As far as I can tell, yes.

anrossi (approved):

PR #3107, opened 2022-10-04
Fix iOS Build Errors

nibanks:
## Description

Fixes a recent macOS build error. Not sure why it's just now showing up.

## Testing

Automated CI

## Documentation

N/A

### DIFF
--- a/src/inc/msquic_posix.h
+++ b/src/inc/msquic_posix.h
@@ -505,7 +505,7 @@ QuicAddrToString(
             Address[0] = ']';
             Address++;
         }
-        sprintf(Address, ":%hu", ntohs(Addr->Ipv4.sin_port));
+        snprintf(Address, 64, ":%hu", ntohs(Addr->Ipv4.sin_port));
     }
     return TRUE;
 }
--- a/src/inc/msquic_posix.h
+++ b/src/inc/msquic_posix.h
@@ -209,7 +209,7 @@ extern "C" {
 //
 
 #define _strnicmp strncasecmp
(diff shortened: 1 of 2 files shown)


nibanks:
This will need to be backported to release branches that build iOS.

csujedihy (approved):

src/test/bin/quic_gtest.cpp
```diff
@@ -1116,6 +1116,7 @@ TEST_P(WithHandshakeArgs4, RandomLoss) {
             (uint8_t)GetParam().ServerStatelessRetry,
             0,  // ClientUsesOldVersion
             0,  // MultipleALPNs
+            0,  // GreaseQuicBitExtension
```
anrossi, line 1119: How did this even build with a parameter number mismatch?

src/inc/msquic_posix.h
```diff
@@ -505,7 +505,7 @@ QuicAddrToString(
             Address[0] = ']';
             Address++;
         }
-        sprintf(Address, ":%hu", ntohs(Addr->Ipv4.sin_port));
+        snprintf(Address, 64, ":%hu", ntohs(Addr->Ipv4.sin_port));
```
anrossi, line 508: No constant available for this?

src/test/bin/quic_gtest.cpp
nibanks: It just didn't initialize the last value.

src/inc/msquic_posix.h
nibanks: It's just defined right above. I don't think it's needed.

PR #3116, opened 2022-10-06
call spinquic from oss-fuzz

ami-GS:
## Description

using `FUZZING` as macro to switch to build main() or LibFuzzer's entrypoint

## Testing

This is test

## Documentation

N/A


src/tools/spin/spinquic.cpp
```diff
@@ -1054,6 +1057,89 @@ CXPLAT_THREAD_CALLBACK(RunThread, Context)
...
+            CxPlatThreadDelete(&Threads[j]);
+        }
+    }
+
+    return 0;
+}
+
+#ifdef FUZZING
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
```
nibanks, line 1121: How are you supposed to use this `data`?

ami-GS: I'm planning
- GB.Buffers for StreamSend/DatagramSend
- GB.Alpns
- ResumptionTicket
- just cast to Server/Client/CredConfiguration
- SpinQuicSetRandomXXXParam

src/tools/spin/spinquic.cpp
```diff
@@ -1117,56 +1209,7 @@ main(int argc, char **argv)
...
-            CxPlatThreadWait(&Threads[j]);
-            CxPlatThreadDelete(&Threads[j]);
-        }
-    }
-
-    return 0;
+    return start(nullptr);
 }
+
+#endif
```
nibanks, line 1215: Please add a trailing newline.

src/tools/spin/spinquic.cpp
nibanks: I'm not convinced that's going to help much. Any "payload" buffer is just a pass through and doesn't really fuzz anything. That includes any send buffers and resumption tickets. You can't just cast random buffers to structs/handles and the code will definitely just crash (which is expected).

Are there other interfaces beyond `TestOneInput`? Ideally, I think having the fuzzer drive which API, one at a time might be better. Let's definitely talk more in today's Discord meeting.

nibanks (commented):
I think we need to have some design discussion on this work.

src/tools/spin/spinquic.cpp
ami-GS: Created utility class which maintain each thread's pointer against data array from LibFuzzer.
And use it for API switching.
Fuzzer entry point is only LLVMFuzzerTestOneInput.
To use data more tricky is explained in this and its directory. Basically it is user to take care of data splitting
https://github.com/google/fuzzing/blob/master/docs/split-inputs.md

src/tools/spin/spinquic.cpp
```diff
@@ -553,7 +622,13 @@ void Spin(Gbs& Gb, LockableVector<HQUIC>& Connections, std::vector<HQUIC>* Liste
             continue; \
         }
 
-        switch (GetRandom(SpinQuicAPICallCount)) {
+        int ApiSwitch = -1;
+#ifdef FUZZING
+        Gb.FuzzData->TryGetRandom<int>(SpinQuicAPICallCount, &ApiSwitch, ThreadID);
+#else
+        ApiSwitch = GetRandom(SpinQuicAPICallCount);
```
nibanks, line 629: Instead of doing all these ifdefs, why not make `GetRandom` call the FuzzData?

src/tools/spin/spinquic.cpp
```diff
@@ -1062,6 +1147,101 @@ CXPLAT_THREAD_CALLBACK(RunThread, Context)
     CXPLAT_THREAD_RETURN(0);
 }
 
+
+int start(void* Context) {
```
nibanks, line 1151: Seems this `Context` is never passed in or used? Remove it?

nibanks: Also, you never return anything but `0`? Why have the `int` return then?

ami-GS: Removed Context and return void

src/tools/spin/spinquic.cpp
```diff
@@ -1125,58 +1336,9 @@ main(int argc, char **argv)
...
-            CxPlatThreadWait(&Threads[j]);
-            CxPlatThreadDelete(&Threads[j]);
-        }
-    }
-
-    CxPlatLockUninitialize(&RunThreadLock);
     return 0;
 }
+
+#endif
```
nibanks, line 1344: Can you put a comment indicated what #if this is ending?

src/tools/spin/spinquic.cpp
```diff
@@ -259,8 +355,10 @@ QUIC_STATUS QUIC_API SpinQuicHandleStreamEvent(HQUIC Stream, void * /* Context *
     return QUIC_STATUS_SUCCESS;
 }
 
-QUIC_STATUS QUIC_API SpinQuicHandleConnectionEvent(HQUIC Connection, void * /* Context */, QUIC_CONNECTION_EVENT *Event)
+QUIC_STATUS QUIC_API SpinQuicHandleConnectionEvent(HQUIC Connection, void *Context, QUIC_CONNECTION_EVENT *Event)
 {
+    // TODO: avoid context overwite
+    uint16_t ThreadID = 0; // *(uint16_t*)Context;
```
ami-GS, line 361: This Context becomes null by (maybe) SetContext called in spin loop.
Fuzz data based GetRandom cannot be used

src/tools/spin/spinquic.cpp
ami-GS: Done

src/tools/spin/spinquic.cpp
ami-GS: StreamCallback as well

nibanks (approved):

PR #3122, opened 2022-10-07
Fix passing pointer-pointer, skip inevitable SEGV case

ami-GS:
## Description

Fix error reported from OSS-Fuzz
1. pointer-pointer was passed
2. SEGV by accessing random address when QUIC_PARAM_GLOBAL_VERSION_SETTINGS

## Testing

this is test

## Documentation

N/A

### DIFF
--- a/src/fuzzing/fuzz.cc
+++ b/src/fuzzing/fuzz.cc
@@ -9,6 +9,7 @@ Abstract:
 
 --*/
 
+#define QUIC_API_ENABLE_PREVIEW_FEATURES 1
 #define CX_PLATFORM_LINUX 1
 #define QUIC_TEST_APIS 1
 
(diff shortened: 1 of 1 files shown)


src/fuzzing/fuzz.cc
```diff
@@ -27,11 +28,13 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
...
-			Param,
-			size,
-			&data);
+			if (Param == QUIC_PARAM_GLOBAL_VERSION_SETTINGS)
+				continue;
+			MsQuic->SetParam(
+				nullptr,
+				Param,
+				size,
+				data);
```
nibanks, line 37:
```suggestion
	if (Param != QUIC_PARAM_GLOBAL_VERSION_SETTINGS) {
	    MsQuic->SetParam(
		    nullptr,
		    Param,
		    size,
		    data);
    }
```

PR #3123, opened 2022-10-08
Add test for Chosen Version mismatch

anrossi:
## Description

Add test for scenario where the Version Info transport parameter has a Chosen Version field that doesn't match the Version field in the long header.

This is not a guarantee required by the spec, and may change in the future for future versions or extensions.

## Testing

Tests added.

## Documentation

N/A

### DIFF
--- a/src/test/lib/HandshakeTest.cpp
+++ b/src/test/lib/HandshakeTest.cpp
@@ -3281,36 +3281,15 @@ ListenerAcceptConnectionTestTP(
 }
 
 void
-QuicTestOddSizeVNTP(
+QuicTestCustomVNTP(
     _In_ bool TestServer,
-    _In_ uint16_t VNTPSize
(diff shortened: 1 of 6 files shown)


src/test/lib/HandshakeTest.cpp
```diff
@@ -3382,4 +3361,57 @@ QuicTestOddSizeVNTP(
...
+
+    QuicTestCustomVNTP(TestServer, &TestTP);
+}
+
+void
+QuicTestVNTPChosenVersionMismatch(
+    _In_ bool TestServer
+    )
+{
+    #define QUIC_TP_ID_VERSION_NEGOTIATION_EXT                  0xFF73DB
```
nibanks, line 3401: Not needed to be defined twice, right?

nibanks (approved):

PR #3132, opened 2022-10-11
Added support for ca certificate file setting in SSL_CTX.

jbevemyr:
## Description

Added support for custom ca certificate roots.

## Testing

New tests are needed but I'm not familiar with the test setup yet.

## Documentation

The new CaCertificateFile option needs to be documented.


jbevemyr:

src/platform/tls_openssl.c
```diff
@@ -1462,6 +1462,21 @@ CxPlatTlsSecConfigCreate(
         }
     }
 
+    if (CredConfig->CaCertificateFile) {
```
nibanks, line 1465: On Schannel, since this isn't supported, we should return an error if this is specified (not_supported or invalid_parameter).

src/inc/msquic.h
```diff
@@ -119,6 +119,7 @@ typedef enum QUIC_CREDENTIAL_FLAGS {
     QUIC_CREDENTIAL_FLAG_CACHE_ONLY_URL_RETRIEVAL               = 0x00020000, // Windows only currently
     QUIC_CREDENTIAL_FLAG_REVOCATION_CHECK_CACHE_ONLY            = 0x00040000, // Windows only currently
     QUIC_CREDENTIAL_FLAG_INPROC_PEER_CERTIFICATE                = 0x00080000, // Schannel only
+    QUIC_CREDENTIAL_FLAG_USE_CACERTFILE                         = 0x00100000, // OpenSSL only currently
```
nibanks, line 122: Let's call it `QUIC_CREDENTIAL_FLAG_SET_CA_CERTIFICATE_FILE` instead.

src/platform/tls_openssl.c
```diff
@@ -1462,6 +1462,30 @@ CxPlatTlsSecConfigCreate(
...
+                    ERR_get_error(),
+                    "SSL_CTX_load_verify_locations failed");
+                Status = QUIC_STATUS_TLS_ERROR;
+                goto Exit;
+            }
+        } else {
+            QuicTraceEvent(
+                LibraryErrorStatus,
+                "QUIC_CREDENTIAL_FLAG_USE_CACERTFILE is set but "
+                "CaCertificateFile is not set.");
```
nibanks, line 1483: I don't think we need to error out if the flag is set but the pointer is NULL. Let's just treat the flag as "please check the pointer". So please combine the two top-level `if` checks into one, and only do anything if both are true.

src/platform/tls_openssl.c
```diff
@@ -1462,6 +1462,30 @@ CxPlatTlsSecConfigCreate(
         }
     }
 
+    if (CredConfigFlags & QUIC_CREDENTIAL_FLAG_USE_CACERTFILE) {
+        if (CredConfig->CaCertificateFile) {
+            Ret = SSL_CTX_load_verify_locations(SecurityConfig->SSLCtx,
```
nibanks, line 1467: Would you mind updating the formatting to match more like the call to `SSL_CTX_use_certificate` above?

jbevemyr: Sorry, I've been working in another repo with wierd indentation lately :-)

src/platform/tls_schannel.c
```diff
@@ -1108,6 +1108,10 @@ CxPlatTlsSecConfigCreate(
     if (CredConfig->Flags & QUIC_CREDENTIAL_FLAG_REVOCATION_CHECK_CACHE_ONLY) {
         Credentials->dwFlags |= SCH_CRED_REVOCATION_CHECK_CACHE_ONLY;
     }
+    if (CredConfig->Flags & QUIC_CREDENTIAL_FLAG_SET_CA_CERTIFICATE_FILE) {
+        Status = QUIC_STATUS_NOT_SUPPORTED;
+        goto Error;
+    }
```
nibanks, line 1114: Please move this to the top of the function where all the other error-out conditions are?

nibanks:
Everything is looking good. We just need some tests please.

jbevemyr:
I'll try to add some tests.

src/platform/tls_openssl.c
```diff
@@ -1644,6 +1644,7 @@ CxPlatTlsInitialize(
     uint16_t ServerNameLength = 0;
     UNREFERENCED_PARAMETER(State);
 
+
```
nibanks, line 1647: nit: remove extra newline?

src/inc/msquic.h
```diff
@@ -317,7 +317,7 @@ typedef struct QUIC_CREDENTIAL_CONFIG {
     void* Reserved; // Currently unused
     QUIC_CREDENTIAL_LOAD_COMPLETE_HANDLER AsyncHandler; // Optional
     QUIC_ALLOWED_CIPHER_SUITE_FLAGS AllowedCipherSuites;// Optional
-    const char* CaCertificateFile;
+    char* CaCertificateFile;
```
nibanks, line 320: Why can't it be const? It should be.

jbevemyr: Yes, right.

src/platform/selfsign_openssl.c
```diff
@@ -85,27 +86,27 @@ static uint8_t* ReadPkcs12(const char* Name, uint32_t* Length) {
...
+
+
 QUIC_STATUS
-CxPlatTlsGenerateSelfSignedCert(
-    _In_z_ char *CertFileName,
-    _When_(OutputPkcs12 == TRUE, _Reserved_)
-    _When_(OutputPkcs12 == FALSE, _In_z_)
-        char *PrivateKeyFileName,
+CxPlatTlsGenerateSignedCert(
+    _In_opt_z_ X509 *CaX509,
```
anrossi, line 95: This is not a null-terminated byte buffer, so you'd use `_In_opt_` instead.

src/platform/selfsign_openssl.c
```diff
@@ -85,27 +86,27 @@ static uint8_t* ReadPkcs12(const char* Name, uint32_t* Length) {
...
+    _In_opt_z_ EVP_PKEY *CaPKey,
     _In_z_ char *SNI,
-    _In_opt_z_ char *Password,
-    _In_ BOOLEAN OutputPkcs12
+    _In_z_ BOOLEAN GenCaCert,
```
anrossi, line 98: This isn't null-terminated either, so `_In_` is all that's needed here.

src/platform/selfsign_openssl.c
anrossi: Also, consider renaming this to `SignerCert` or similar to state the purpose more clearly.

jbevemyr: Sorry, I was guessing what the _In_*_ annotations meant. Should have looked it up.

src/platform/selfsign_openssl.c
```diff
@@ -588,27 +808,55 @@
...
-        KeyFileName = QuicTestServerPrivateKeyFilename;
+            CertFileName = QuicTestServerCertFilenameSS;
+            KeyFileName = QuicTestServerPrivateKeyFilenameSS;
 #endif
+        }
     }
 
     CXPLAT_CREDENTIAL_CONFIG_INTERNAL* Params =
-        malloc(sizeof(CXPLAT_CREDENTIAL_CONFIG_INTERNAL) + sizeof(TEMP_DIR_TEMPLATE));
+        malloc(sizeof(CXPLAT_CREDENTIAL_CONFIG_INTERNAL) +
```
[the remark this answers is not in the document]
jbevemyr: This is a false failure, right? Or does it require me to change anything? I added a linebreak and changed the size of CXPLAT_CREDENTIAL_CONFIG_INTERNAL by adding another file path array, in line with the others in the same struct.

src/platform/selfsign_openssl.c
```diff
@@ -248,6 +294,93 @@
...
+    }
+
+    Status = CxPlatTlsGenerateSignedCert(
+        CaX509, CaPKey, SNI, FALSE, &X509, &PKey);
+
+    if (Status != QUIC_STATUS_SUCCESS)
+        goto Exit;
+
+    if (CaFileName != NULL) {
+        Fd = fopen(CaFileName, "wb");
```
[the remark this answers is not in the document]
jbevemyr: This is also a false failure, right? Or at least one that can be ignored. The file is created in the same way the other files are created. It is just test code after all. Should be ok to create with 0666 mode.

nibanks:
@anrossi could you try to review this again today? Thanks!

src/cs/lib/msquic_generated.cs
```diff
@@ -280,6 +281,9 @@ internal unsafe partial struct QUIC_CREDENTIAL_CONFIG
 
         internal QUIC_ALLOWED_CIPHER_SUITE_FLAGS AllowedCipherSuites;
 
+        [NativeTypeName("const char *")]
```
anrossi, line 284: @nibanks How does .NET handle adding a member in the middle of an object?

nibanks: It's at the end.

src/platform/unittest/TlsTest.cpp
```diff
@@ -165,6 +206,21 @@ struct TlsTest : public ::testing::TestWithParam<bool>
...
         ClientCertParams = (QUIC_CREDENTIAL_CONFIG*)CxPlatGetSelfSignedCert(CXPLAT_SELF_SIGN_CERT_USER, TRUE);
         ASSERT_NE(nullptr, ClientCertParams);
+
+        CaSelfSignedCertParams = (QUIC_CREDENTIAL_CONFIG*)CxPlatGetSelfSignedCert(CXPLAT_SELF_SIGN_CA_CERT_USER, FALSE);
+        ASSERT_NE(nullptr, CaSelfSignedCertParams);
+        CaSelfSignedCertParamsFlags = CaSelfSignedCertParams->Flags;
+        CaClientCertParams = (QUIC_CREDENTIAL_CONFIG*)CxPlatGetSelfSignedCert(CXPLAT_SELF_SIGN_CA_CERT_USER, TRUE);
+        ASSERT_NE(nullptr, ClientCertParams);
+
+        // We need to switch the ca roots around since the server must
```
anrossi, line 216: Why the swap instead of just assigning the correct certificate to the right *certparams variable?

src/cs/lib/msquic_generated.cs
anrossi: There's all the pointers to the certificate data after it, I don't think this is the end.

src/platform/selfsign_openssl.c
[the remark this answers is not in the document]
anrossi: Yeah, this failure is not unique to your change.

src/cs/lib/msquic_generated.cs
nibanks: Look at msquic.h for the actual structure. Everything below this is a property (i.e. just a helper function).

anrossi (commented):
Looking good! Just a couple things more and could you add the new flag to the `PlatformSpecificFlagsOpenSsl` test in TlsTest.cpp?
Thank you for your patience and effort

src/platform/unittest/TlsTest.cpp
jbevemyr: It is tricky for two reasons, the certparams are returned by the CxPlatGetSelfSignedCert, so in order to set them directly the ca certificate file parameter would have to be passed as an argument to CxPlatGetSelfSignedCert. However, the ca certificate file is also created by the CxPlatGetSelfSignedCert so we do not even know what to pass in until at least the "other" version returns. The only solution I see, which I've also considered for other reasons, would be to have a new function that creates both the client and server certificates in one go. That way the ca certificates can be assigned correctly from the start. Should I rework the code with this change?

anrossi: Hmm, maybe the CA certificate creation should be split out into its own function which returns the file path to the CA certificate, and then it can be passed as a parameter to `CxPlatGetSelfSignedCert` to sign the client/server certificate?  I think that would be the most flexible design. Does that sound too complicated?

jbevemyr: Sounds like a good plan, I'll give it a try and see what it looks like.

jbevemyr: I pushed a commit with this change, but it introduces quite a bit of code just to avoid the swap...

src/platform/selfsign_capi.c
```diff
@@ -1031,13 +1031,25 @@ FindOrCreateCertificate(
     return (void*) Cert;
 }
 
+_IRQL_requires_max_(PASSIVE_LEVEL)
+const char*
+CxPlatGetSelfSignedCertCaCertificateFileName(
+    _In_ BOOLEAN ClientCertificate
+    )
+{
```
anrossi, line 1039: you'll also need a `UNREFERENCED_PARAMETER(ClientCertificate)` here, otherwise the compiler will yell.

jbevemyr: Right, missed that one.

src/inc/msquic.h
```diff
@@ -319,6 +320,7 @@ typedef struct QUIC_CREDENTIAL_CONFIG {
     void* Reserved; // Currently unused
     QUIC_CREDENTIAL_LOAD_COMPLETE_HANDLER AsyncHandler; // Optional
     QUIC_ALLOWED_CIPHER_SUITE_FLAGS AllowedCipherSuites;// Optional
+    const char* CaCertificateFile;
```
nibanks, line 323: Two things:

1. This should have a comment indicating it's optional.
2. You can actually leverage the `Reserved` space instead of appending.

src/platform/tls_schannel.c
```diff
@@ -1006,6 +1006,10 @@ CxPlatTlsSecConfigCreate(
         return QUIC_STATUS_INVALID_PARAMETER;
     }
 
+    if (CredConfig->Flags & QUIC_CREDENTIAL_FLAG_SET_CA_CERTIFICATE_FILE) {
+        return  QUIC_STATUS_NOT_SUPPORTED;
```
nibanks, line 1010:
```suggestion
        return QUIC_STATUS_NOT_SUPPORTED;
```

src/platform/unittest/TlsTest.cpp
```diff
@@ -100,11 +105,49 @@ struct TlsTest : public ::testing::TestWithParam<bool>
...
+                nullptr,
+                nullptr,
+                QUIC_ALLOWED_CIPHER_SUITE_NONE,
+                CaClientCertParams->CaCertificateFile
+            };
             CredConfig.Flags |= CredFlags;
             CredConfig.AllowedCipherSuites = CipherFlags;
             Load(&CredConfig, TlsFlags);
         }
+
```
nibanks, line 150: nit: unnecessary change

src/platform/unittest/TlsTest.cpp
```diff
@@ -1326,6 +1390,37 @@ TEST_F(TlsTest, DeferredCertificateValidationAllow)
...
+        ASSERT_NE(nullptr, ServerContext.State.WriteKeys[QUIC_PACKET_KEY_1_RTT]);
+
+        Result = ClientContext.ProcessData(&ServerContext.State, DefaultFragmentSize, true);
+        ASSERT_TRUE(ClientContext.ReceivedPeerCertificate);
+        ASSERT_TRUE(Result & CXPLAT_TLS_RESULT_HANDSHAKE_COMPLETE);
+    }
+}
+
+#endif
+
```
nibanks, line 1422: nit: remove extra newline

src/platform/unittest/TlsTest.cpp
```diff
@@ -1835,6 +1930,25 @@ TEST_F(TlsTest, ClientCertificateDeferValidation)
     DoHandshake(ServerContext, ClientContext);
 }
 
+#ifndef _WIN32
```
nibanks, line 1933: Generally, we want more specific flags such as the `QUIC_ENABLE_ANON_CLIENT_AUTH_TESTS` below. Not just using `_WIN32`.

src/platform/unittest/TlsTest.cpp
```diff
@@ -1930,8 +1926,7 @@ TEST_F(TlsTest, ClientCertificateDeferValidation)
     DoHandshake(ServerContext, ClientContext);
 }
 
-#ifndef _WIN32
-
+#ifdef QUIC_ENABLE_CA_CERTIFICATE_FILE
```
nibanks, line 1929: But you still need to define this, right?

nibanks:
Can you please update `docs\api\QUIC_CREDENTIAL_CONFIG.md` too?

src/platform/unittest/TlsTest.cpp
nibanks: And the define should specifically specify `_TESTS`

jbevemyr: Yes, good points.

anrossi (approved):

jbevemyr:
Ready to merge?

nibanks:
Thanks for working on this!!!

PR #3135, opened 2022-10-11
Add -StaticCRT switch to build script and never use partial static crt linking for debug builds

csujedihy:
## Description

1. Add `-StaticCRT` to allow full static CRT linking.
2. Never apply partial static CRT linking to debug bits.

## Testing

CI

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -51,6 +51,9 @@ This script provides helpers for building msquic.
 .PARAMETER DynamicCRT
     Builds msquic with dynamic C runtime (Windows-only).
 
+.PARAMETER StaticCRT
+    Builds msquic with static C runtime (Windows-only).
+
 .PARAMETER PGO
     Builds msquic with profile guided optimization support (Windows-only).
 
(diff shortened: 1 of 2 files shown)


scripts/build.ps1
```diff
@@ -417,8 +423,13 @@ function CMake-Generate {
         }
         $Arguments += " -DCMAKE_BUILD_TYPE=" + $ConfigToBuild
     }
-    if ($DynamicCRT) {
-        $Arguments += " -DQUIC_STATIC_LINK_CRT=off -DQUIC_STATIC_LINK_PARTIAL_CRT=off"
+    if ($IsWindows) {
+        if ($DynamicCRT) {
+            $Arguments += " -DQUIC_STATIC_LINK_CRT=off -DQUIC_STATIC_LINK_PARTIAL_CRT=off"
+        }
+        if ($StaticCRT) {
```
nibanks, line 430: Why can't we statically link on Linux? Does CRT not exist?

nibanks (approved):

scripts/build.ps1
csujedihy:
```
.PARAMETER DynamicCRT
    Builds msquic with dynamic C runtime (Windows-only).
```
I think no?

csujedihy: Isn't CRT a windows thing? We link against glibc in linux?

PR #3140, opened 2022-10-12
Fix TCP Perf Test Race Condition on Shutdown

nibanks:
## Description

There is a window in which the receive callback can happen after we start deleting the connection object, but haven't yet deleted the socket (in the destructor). This updates the receive path to bail if it fails to queue in this case.

## Testing

Perf tests

## Documentation

No

### DIFF
--- a/src/perf/lib/Tcp.h
+++ b/src/perf/lib/Tcp.h
@@ -118,7 +118,7 @@ class TcpWorker {
     bool Initialize(TcpEngine* _Engine);
     void Shutdown();
     static CXPLAT_THREAD_CALLBACK(WorkerThread, Context);
-    void QueueConnection(TcpConnection* Connection);
+    bool QueueConnection(TcpConnection* Connection);
 };
 
 class TcpServer {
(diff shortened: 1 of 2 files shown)


src/perf/lib/Tcp.h
```diff
@@ -249,7 +249,7 @@ class TcpConnection {
     QUIC_BUFFER* NewSendBuffer();
     void FreeSendBuffer(QUIC_BUFFER* SendBuffer);
     bool FinalizeSendBuffer(QUIC_BUFFER* SendBuffer);
-    void AddRef() { CxPlatRefIncrement(&Ref); }
+    bool TryAddRef() { return CxPlatRefIncrementNonZero(&Ref, 1) != FALSE; }
```
csujedihy, line 252: >CxPlatRefIncrementNonZero [](http://example.com/codeflow?start=30&length=25)

Is the purpose of TryAddRef return false if ref == 0?

nibanks: Yes.

csujedihy: But if Ref drops to zero, the entire connection object has been already deleted?

nibanks: It is _being_ deleted. In the destructor, we delete the socket, which has to wait for any outstanding receives. That's the race here.

csujedihy: Oh, I got it.

PR #3142, opened 2022-10-12
Add lock to avoid MsQuicOpen abort issue for statically linked spinquic

ami-GS:
## Description

`CXPLAT_TEL_ASSERT(MsQuicLib.Loaded)` in MsQuicAddRef abort when libmsquic is statically linked with spinquic

## Testing

locally tested

## Documentation

N/A

### DIFF
--- a/src/tools/spin/spinquic.cpp
+++ b/src/tools/spin/spinquic.cpp
@@ -89,6 +89,8 @@ public:
 };
 
 static QUIC_API_TABLE MsQuic;
+// This locks MsQuicOpen2 in RunThread when statically linked with libmsquic
+CXPLAT_LOCK RunThreadLock;
 
 const uint32_t MaxBufferSizes[] = { 0, 1, 2, 32, 50, 256, 500, 1000, 1024, 1400, 5000, 10000, 64000, 10000000 };
 static const size_t BufferCount = ARRAYSIZE(MaxBufferSizes);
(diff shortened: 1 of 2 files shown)


src/tools/spin/spinquic.cpp
```diff
@@ -1074,6 +1078,7 @@ main(int argc, char **argv)
         PrintHelpText();
     }
 
+    CxPlatLockInitialize(&RunThreadLock);
```
nibanks, line 1081: Do this after initializing the platform. Also, please uninitialize it when done.

ami-GS: Thanks, fixed

src/tools/spin/spinquic.cpp
```diff
@@ -938,7 +940,9 @@ CXPLAT_THREAD_CALLBACK(RunThread, Context)
             ASSERT_ON_NOT(Gb.Buffers[j].Buffer);
         }
 
+        CxPlatLockAcquire(&RunThreadLock);
         QUIC_STATUS Status = MsQuicOpen2(&Gb.MsQuic);
+        CxPlatLockRelease(&RunThreadLock);
```
nibanks, line 945: Let's **only** do this is static linking. I want dynamic linking to stress this by calling in parallel. Can you add a #ifdef around it?

nibanks (approved):

PR #3150, opened 2022-10-14
Only Pass ECN if Not non-ECT (Windows)

nibanks:
## Description

Remove ECN ancillary data if not actually passing a non-ECT value.

## Testing

Automated tests (mostly perf)

## Documentation

N/A


src/platform/datapath_winuser.c
```diff
@@ -3877,13 +3879,15 @@ CxPlatSocketSendInline(
...
         }
 
-        WSAMhdr.Control.len += WSA_CMSG_SPACE(sizeof(INT));
-        CMsg = WSA_CMSG_NXTHDR(&WSAMhdr, CMsg);
-        CXPLAT_DBG_ASSERT(CMsg != NULL);
-        CMsg->cmsg_level = IPPROTO_IPV6;
-        CMsg->cmsg_type = IPV6_ECN;
-        CMsg->cmsg_len = WSA_CMSG_LEN(sizeof(INT));
-        *(PINT)WSA_CMSG_DATA(CMsg) = SendData->ECN;
+        if (SendData->ECN != CXPLAT_ECN_NON_ECT) {
```
csujedihy, line 3882: It would be less duplicated code and cleaner if we move it out of the PKT_INFO block.
```
        if (SendData->ECN != CXPLAT_ECN_NON_ECT) {
            WSAMhdr.Control.len += WSA_CMSG_SPACE(sizeof(INT));
            CMsg = WSA_CMSG_NXTHDR(&WSAMhdr, CMsg);
            CXPLAT_DBG_ASSERT(CMsg != NULL);
            CMsg->cmsg_level = LocalAddress->si_family == QUIC_ADDRESS_FAMILY_INET ?
                IPPROTO_IP : IPPROTO_IPV6;
            CMsg->cmsg_type = IP_ECN; // == IPV6_ECN
            CMsg->cmsg_len = WSA_CMSG_LEN(sizeof(INT));
            *(PINT)WSA_CMSG_DATA(CMsg) = SendData->ECN;
        } 
```

csujedihy:
Do we wanna do this for other datapaths?

csujedihy (approved):

src/platform/datapath_winuser.c
nibanks: I prefer to leave as is, since it's fewer branches.

nibanks:
> Do we wanna do this for other datapaths?

I don't think it's super necessary. I just did it here because there is a known perf problem with hitting the slow path on Windows. That said, perf is actually **slower** with this change. Still investigating why, but it seems that by making the send faster, we make the RX slower.

src/platform/datapath_winuser.c
csujedihy: There actually won't be any additional branches if you use IP_ECN for both v4 and v6 and use a variable to store cmsg_level for ECN and have it set in the PKT_INFO block. Up to you.

PR #3152, opened 2022-10-14
Save All Histograms from Perf

nibanks:
## Description

Depends on #3151.

- Saves all runs' histogram files from perf tests
- Moves histogram files into a separate folder
- Enables dry run perf publishing for PRs

## Testing

Perf pipeline

## Documentation

None

### DIFF
--- a/scripts/performance.ps1
+++ b/scripts/performance.ps1
@@ -478,7 +478,7 @@ function Invoke-Test {
     try {
         1..$NumIterations | ForEach-Object {
             Write-LogAndDebug "Running Local: $LocalExe Args: $LocalArguments"
-            $LocalResults = Invoke-LocalExe -Exe $LocalExe -RunArgs $LocalArguments -Timeout $Timeout -OutputDir $OutputDir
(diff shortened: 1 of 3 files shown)


scripts/performance.ps1
```diff
@@ -478,7 +478,7 @@ function Invoke-Test {
     try {
         1..$NumIterations | ForEach-Object {
             Write-LogAndDebug "Running Local: $LocalExe Args: $LocalArguments"
-            $LocalResults = Invoke-LocalExe -Exe $LocalExe -RunArgs $LocalArguments -Timeout $Timeout -OutputDir $OutputDir
+            $LocalResults = Invoke-LocalExe -Exe $LocalExe -RunArgs $LocalArguments -Timeout $Timeout -OutputDir $OutputDir -HistogramFileName "$($Test)_run$($_).txt"
```
csujedihy, line 481: Is it possible aggregate all raw input and get median instead of median of median of several input files (if I correctly understand this)?

nibanks: No, not with how the histograms are generated (they don't have all the data). We'd instead have to run one big run, instead of 5 smaller ones.

csujedihy (approved):

PR #3153, opened 2022-10-14
Make RPS Tests Single Long Run

nibanks:
## Description

1 run for 45 seconds instead of 5 for 10 seconds each.

## Testing

Perf pipeline

## Documentation

N/A


csujedihy:
I think this can significantly increase variance in some scenarios where we need to balance connections on cores like XDP. We used to have 5 iterations which somehow compensated for this problem.

nibanks:
> I think this can significantly increase variance in some scenarios where which connection lands on which core matters like XDP. Having multiple iterations somewhat compensates for this problem.

Yeah... going to abandon this for now.

PR #3164, opened 2022-10-17
Add option for some crude tracing on `stdout`

larseggert:
## Description

This adds some code to trace operation on `stdout`. Mostly useful during development on platforms where no other tracing is supported (e.g., MacOS).

This also casts values to the types indicated in the format strings, because compilers otherwise warn (because of a mismatch between the format string and the variable type.)

## Testing

No new tests needed.

## Documentation

Nope?


src/core/datagram.c
```diff
@@ -372,7 +372,7 @@ QuicDatagramQueueSend(
                 AllocFailure,
                 "Allocation of '%s' failed. (%llu bytes)",
                 "DATAGRAM_SEND operation",
-                0);
+                UINT64_C(0));
```
nibanks, line 375: Perhaps instead of all these, we use `0ULL`? That works, right?

nibanks:
Is the build going to validate all these things now? If not, how do we prevent regressions from creeping in?

src/core/datagram.c
larseggert: Yes, that would work. Will make the change.

larseggert:
In order to validate it, we would need to add a build that uses `QUIC_EVENTS_STDOUT` and `QUIC_LOGS_STDOUT`. Do you want me to add stuff to `CMakeLists.txt` to make that selectable?

larseggert:
Other idea: Are any of the current CI runs building with `QUIC_EVENTS_STUB` and `QUIC_LOGS_STUB`? Because then I cold just add ` __attribute__((format(printf, 1, 2)))` to the definition of the stub function to make the compilers check this.

nibanks:
Yeah, probably worth exposing. We also have a number of builds under the `build_windows_nontest` and `build_linux_nontest` stages that you could leverage to validate this. Or just add a new one.

src/generated/stdout/quic_trace.c
```diff
@@ -0,0 +1,102 @@
...
+                     _Out_writes_bytes_(2 * BufferLen) char * HexString);
+
+
+char * casted_clog_bytearray(const uint8_t * const data,
+                             const size_t len,
+                             struct clog_param ** head)
+{
+    struct clog_param * const param =
+        CXPLAT_ALLOC_PAGED(sizeof(*param), QUIC_POOL_TMP_ALLOC);
+    CXPLAT_FRE_ASSERT(param);
```
nibanks, line 33: You can't assert that allocations succeed.

src/generated/stdout/quic_trace.c
```diff
@@ -0,0 +1,102 @@
...
+    // return the string representation
+    return param->str;
+}
+
+
+void clog_stdout(struct clog_param * head, const char * const format, ...)
+{
+    // replace clog specifiers in the format string
+    const char * repls[] = {"!CID!", "!ADDR!", "!VNL!", "!ALPN!"};
+    char * reformat = strdup(format);
```
nibanks, line 67: Can fail to allocate

src/core/inline.c
```diff
@@ -800,3 +800,4 @@ QuicMtuDiscoveryCheckSearchCompleteTimeout(
     _In_ QUIC_CONNECTION* Connection,
     _In_ uint64_t TimeNow
     );
+
```
nibanks, line 803: nit: unnecessary change?

CMakeLists.txt
```diff
@@ -282,9 +286,14 @@ else()
...
         check_function_exists(sysctl HAS_SYSCTL)
+        if (QUIC_ENABLE_LOGGING AND QUIC_ENABLE_LOGGING_STDOUT EQUAL 0)
+            set (QUIC_ENABLE_LOGGING OFF)
+            message(STATUS "Disabling all tracing on MacOS")
+        endif()
     endif()
 
-    set(QUIC_COMMON_FLAGS "")
+    # for reproducible builds: https://reproducible-builds.org/docs/build-path/
+    set(QUIC_COMMON_FLAGS "-ffile-prefix-map=${CMAKE_SOURCE_DIR}=.")
```
nibanks, line 296: Is this necessary?

CMakeLists.txt
```diff
@@ -104,6 +104,10 @@ option(QUIC_DISABLE_POSIX_GSO "Disable GSO for systems that say they support it
 option(QUIC_OFFICIAL_RELEASE "Configured the build for an official release" OFF)
 set(QUIC_FOLDER_PREFIX "" CACHE STRING "Optional prefix for source group folders when using an IDE generator")
 set(QUIC_LIBRARY_NAME "msquic" CACHE STRING "Override the output library name")
+set(QUIC_ENABLE_LOGGING_STDOUT 0 CACHE STRING "Enables logging on stdout at given verbosity (0=off, 4=debug)")
```
nibanks, line 107: Why go this route, and not just on/off? And later, we could add a `LOGGING_LEVEL` param if we want?

CMakeLists.txt
```diff
@@ -567,16 +576,17 @@ else() #!WIN32
...
+    if (QUIC_ENABLE_SANITIZERS AND QUIC_ENABLE_LOGGING_STDOUT EQUAL 0)
         set(QUIC_ENABLE_LOGGING OFF)
         message(WARNING "LTTng logging is incompatible with sanitizers. Skipping logging")
     endif()
 
     if(QUIC_ENABLE_LOGGING)
-        if (APPLE)
-            set (QUIC_ENABLE_LOGGING OFF)
-            message(STATUS "Disabling all tracing on macos")
-            list(APPEND QUIC_COMMON_DEFINES QUIC_EVENTS_STUB QUIC_LOGS_STUB)
```
nibanks, line 579: Shouldn't we still be using stub if not stdout?

CMakeLists.txt
larseggert: No. I'll do a different PR for it.

src/core/inline.c
larseggert: Yep, good catch

CMakeLists.txt
larseggert: I revert to on/off if you prefer.

CMakeLists.txt
larseggert: The "Disabling all tracing on MacOS" stuff around line 288 sets `QUIC_ENABLE_LOGGING` to `OFF`, which means the stubs are being used.

larseggert:
One question on the stubs: the way they're currently defined (as inline functions) makes them show up in the compiled binary even when compiling for releases. Is this really what you want?

(I cold make the stubs an empty define, which would generate no code at all in that case.)

CMakeLists.txt
```diff
@@ -263,28 +267,23 @@ else()
...
+                message(WARNING "Must explicitly set QUIC_LOGGING_TYPE to \"stdout\" to enable due to performance overhead; disabling logging")
                 set(QUIC_ENABLE_LOGGING OFF)
             endif()
         endif()
-
-    elseif(CX_PLATFORM STREQUAL "darwin")
-        check_function_exists(sysctl HAS_SYSCTL)
     endif()
 
-    set(QUIC_COMMON_FLAGS "")
```
nibanks, line 287: Why remove this line?

scripts/build.ps1
```diff
@@ -24,6 +24,9 @@ This script provides helpers for building msquic.
 .PARAMETER DisableLogs
     Disables log collection.
 
+.PARAMETER StdoutLogs
+    Log to stdout (Only supported on unix)
+
```
nibanks, line 29: We should make this match CMake options. Let's update this to set the logging type. It should be optional and default to an empty string. Only if it's set to a non-empty string do we pass it down to CMake.

scripts/build.ps1
```diff
@@ -405,8 +405,8 @@ function CMake-Generate {
     if (!$DisableLogs) {
         $Arguments += " -DQUIC_ENABLE_LOGGING=on"
     }
-    if ($StdoutLogs) {
-        $Arguments += " -DQUIC_ENABLE_LOGGING=on -DQUIC_LOGGING_TYPE=stdout"
+    if ($LoggingType) {
+        $Arguments += " -DQUIC_ENABLE_LOGGING=on -DQUIC_LOGGING_TYPE=" + $LoggingType
```
nibanks, line 409: Let's put the `$LoggingType` check in the `if (!$DisableLogs)` section. Then you don't need to set `QUIC_ENABLE_LOGGING` just for the type, which makes things confusing if somehow `DisableLogs` is set too.

.azure/azure-pipelines.ci.yml
```diff
@@ -456,7 +456,7 @@ stages:
       arch: arm64
       tls: openssl
       extraName: 'stdout'
-      extraBuildArgs: -StdoutLogs
+      extraBuildArgs: -LoggingType=stdout
```
nibanks, line 459:
```suggestion
      extraBuildArgs: -LoggingType stdout
```

nibanks:
Looks like you accidentally updated the googletest submodule.

scripts/build.ps1
```diff
@@ -399,6 +405,9 @@ function CMake-Generate {
     if (!$DisableLogs) {
         $Arguments += " -DQUIC_ENABLE_LOGGING=on"
     }
+    if ($LoggingType) {
```
nibanks, line 408:
```suggestion
    if ($LoggingType -ne "") {
```

nibanks (approved):

PR #3168, opened 2022-10-18
Add ECN support: ECN validation

csujedihy:
## Description

This PR implements [13.4.2. ](https://www.rfc-editor.org/rfc/rfc9000.html#section-13.4.2)[ECN Validation](https://www.rfc-editor.org/rfc/rfc9000.html#name-ecn-validation) in RFC9000. The ECN validation algorithm implemented in this PR is very similar to the sample algorithm given in RFC9000 A.4. The details can be found at the top of path.h.

This PR also adds a new global QUIC setting, EcnEnabled and a flag, EcnCapable, in connection stats.

This PR does not implement CC ECN reaction, which will be done in a separate PR.

## Testing

CI.


src/core/connection.h
```diff
@@ -563,6 +564,18 @@ typedef struct QUIC_CONNECTION {
...
+    // ECT and CE counters for different packet number spaces.
+    //
+    uint64_t EcnEctCounters[QUIC_ENCRYPT_LEVEL_COUNT];
+    uint64_t EcnCeCounters[QUIC_ENCRYPT_LEVEL_COUNT]; // maps to ecn_ce_counters in RFC 9002.
+
+    //
+    // The total number of packets sent with each corresponding ECT codepoint in all encryption
+    // level.
+    //
+    uint64_t NumPacketsSentWithEct;
```
nibanks, line 577: Can this be moved to a sub struct? I feel like it shouldn't be on the connection itself.

nibanks: Should it be per-path?

src/core/sent_packet_metadata.h
```diff
@@ -94,6 +94,7 @@ typedef struct QUIC_SEND_PACKET_FLAGS {
     //
     BOOLEAN IsAppLimited            : 1;
     BOOLEAN HasLastAckedPacketInfo  : 1;
+    BOOLEAN EcnEctSet               : 1;
```
nibanks, line 97: This makes us use another byte per packet. I wonder if there's a way to eliminate this need.

src/core/connection.h
csujedihy: The ECN count feedback isn't per path, right? If we make this per-path, it will be unsync'd with ECN counts from the peer.

src/test/lib/TestHelpers.h
```diff
@@ -581,6 +581,26 @@ class DatapathHooks
     }
 };
 
+struct EcnModifyHelper : public DatapathHook
+{
+    CXPLAT_ECN_TYPE EcnType;
```
nibanks, line 586: Default?

src/core/sent_packet_metadata.h
csujedihy: That's not possible I think. We need to know which packets are sent with ECT bit set.

Are you sure this will use another byte? I don't see clear effort for struct packing in QUIC_SENT_PACKET_METADATA.

src/test/lib/TestHelpers.h
csujedihy: Oh right, it actually needs a IsSet flag that is default FALSE.

csujedihy: fixed

src/core/quicdef.h
```diff
@@ -506,6 +506,11 @@ CXPLAT_STATIC_ASSERT(
 //
 #define QUIC_DEFAULT_GREASE_QUIC_BIT_ENABLED         FALSE
 
+//
+// The default value for enabling sender-side ECN support.
+//
+#define QUIC_DEFAULT_ECN_ENABLED         FALSE
```
nibanks, line 512: nit: can you align the `FALSE` with the other values above?

src/core/path.h
```diff
@@ -67,6 +101,21 @@ typedef struct QUIC_PATH {
...
 
+    //
+    // ECN validation state.
+    //
+    uint8_t EcnValidationState : 2;
+
+    //
+    // Indicates whether EcnTestingEndingTime is set.
+    //
+    BOOLEAN EcnTestingEndingTimeSet : 1;
```
nibanks, line 112: Can you just consider a value of `0` for `EcnTestingEndingTime` as unset?

src/core/sent_packet_metadata.h
nibanks: I validated the size in the past. I can't claim 100% certainty now.

src/core/connection.h
```diff
@@ -194,6 +194,7 @@ typedef union QUIC_CONNECTION_STATE {
         //
         BOOLEAN DisableVneTp : 1;
 #endif
+
```
nibanks, line 197: nit: unnecessary change

src/core/loss_detection.c
```diff
@@ -1496,6 +1498,69 @@ QuicLossDetectionProcessAckBlocks(
...
+            // Per RFC 9000, we validate ECN counts from received ACK frames
+            // when the largest acked packet number increases.
+            //
+            BOOLEAN EcnValidated = TRUE;
+            int64_t EctCeDeltaSum = 0;
+            if (Ecn != NULL) {
+                EctCeDeltaSum +=
+                    Ecn->CE_Count - Connection->EcnCeCounters[EncryptLevel];
+                EctCeDeltaSum +=
+                    Ecn->ECT_0_Count - Connection->EcnEctCounters[EncryptLevel];
```
nibanks, line 1512: Can we add some debug asserts to make sure the subtraction doesn't ever underflow?

src/core/loss_detection.c
```diff
@@ -1496,6 +1498,69 @@ QuicLossDetectionProcessAckBlocks(
...
+                    EcnValidated = FALSE;
+                }
+
+                if (EcnValidated) {
+                    //
+                    // TODO: Notify CC of the ECN signal before we update the CE counts.
+                    //
+                    Connection->EcnCeCounters[EncryptLevel] = Ecn->CE_Count;
+                    Connection->EcnEctCounters[EncryptLevel] = Ecn->ECT_0_Count;
+                    if (Path->EcnValidationState <= ECN_VALIDATION_UNKNOWN) {
```
nibanks, line 1531: Is this check necessary? It's just excluding FAILED, right? We're already in a conditional that says !FAILED. Just add a debug assert instead?

src/core/loss_detection.c
```diff
@@ -1740,12 +1805,15 @@ QuicLossDetectionProcessTimerOperation(
 
     if (OldestPacket != NULL &&
         CxPlatTimeDiff32(OldestPacket->SentTime, TimeNow) >=
-            MS_TO_US(Connection->Settings.DisconnectTimeoutMs)) {
+            MS_TO_US(Connection->Settings.DisconnectTimeoutMs) &&
+        (Path->EcnValidationState != ECN_VALIDATION_TESTING ||
+            !OldestPacket->Flags.EcnEctSet)) {
```
nibanks, line 1810: So, does this extend how long we wait for handshake failures because no response from the server?

src/core/packet_builder.c
```diff
@@ -253,12 +253,36 @@ QuicPacketBuilderPrepare(
...
+                        uint32_t ThreePtosInUs =
+                            QuicLossDetectionComputeProbeTimeout(
+                                &Connection->LossDetection,
+                                &Connection->Paths[0],
+                                QUIC_CLOSE_PTO_COUNT);
+                        Builder->Path->EcnTestingEndingTime = TimeNow + ThreePtosInUs;
+                        Builder->Path->EcnTestingEndingTimeSet = TRUE;
+                    }
+                }
+            }
```
nibanks, line 278: I think this should either go (or be called from) at the beginning of `QuicSendFlush` or in `QuicPacketBuilderInitialize`.

nibanks: This is somewhat similar to `QuicMtuDiscoveryCheckSearchCompleteTimeout`.

src/core/packet_builder.c
```diff
@@ -927,7 +951,7 @@ QuicPacketBuilderFinalize(
     Builder->Metadata->SentTime = CxPlatTimeUs32();
     Builder->Metadata->PacketLength =
         Builder->HeaderLength + PayloadLength;
-
+    Builder->Metadata->Flags.EcnEctSet = CxPlatSendDataEctSet(Builder->SendData);
```
nibanks, line 954: Instead of a platform abstraction, just save the EctSet flag on the builder.

src/inc/msquic.h
```diff
@@ -473,7 +473,8 @@ typedef struct QUIC_STATISTICS_V2 {
     uint32_t ResumptionAttempted    : 1;
     uint32_t ResumptionSucceeded    : 1;
     uint32_t GreaseBitNegotiated    : 1;    // Set if we negotiated the GREASE bit.
-    uint32_t RESERVED               : 27;
+    uint32_t EcnCapable             : 1;
```
nibanks, line 476: Should we expose some info about this feature in our statistics?

src/inc/quic_datapath.h
```diff
@@ -586,6 +586,17 @@ CxPlatSendDataAlloc(
...
+// Returns whether ECT bit is set in the send context.
+//
+_IRQL_requires_max_(DISPATCH_LEVEL)
+_Success_(return != NULL)
+BOOLEAN
+CxPlatSendDataEctSet(
+    _In_ CXPLAT_SEND_DATA* SendData
+    );
+
+
```
nibanks, line 599:
```suggestion

```

src/test/bin/quic_gtest.cpp
```diff
@@ -1826,6 +1826,17 @@ TEST(Misc, StreamAbortConnFlowControl) {
     }
 }
 
+#ifdef QUIC_TEST_DATAPATH_HOOKS_ENABLED
+TEST(Misc, Ecn) {
+    TestLogger Logger("Ecn");
+    if (TestingKernelMode) {
+        ASSERT_TRUE(DriverClient.Run(IOCTL_QUIC_RUN_ECN));
+    } else {
+        QuicTestEcn();
```
nibanks, line 1835: Should we have different IPv4 and IPv6 tests?

src/test/lib/DataTest.cpp
```diff
@@ -2161,6 +2161,194 @@ QuicTestAbortReceive(
...
+    CxPlatEvent ServerStreamRecv;
+    CxPlatEvent ServerStreamShutdown;
+    MsQuicStream* ServerStream {nullptr};
+    bool ServerStreamHasShutdown {false};
+
+    static QUIC_STATUS StreamCallback(_In_ MsQuicStream* Stream, _In_opt_ void* Context, _Inout_ QUIC_STREAM_EVENT* Event) {
+        auto TestContext = (EcnTestContext*)Context;
+        if (Event->Type == QUIC_STREAM_EVENT_RECEIVE) {
+            TestContext->ServerStreamRecv.Set();
+            return QUIC_STATUS_PENDING;
```
nibanks, line 2174: Why async receives?

src/core/settings.c
```diff
@@ -1188,6 +1213,7 @@ QuicSettingsDump(
     QuicTraceLogVerbose(SettingCongestionControlAlgorithm,  "[sett] CongestionControlAlgorithm = %hu", Settings->CongestionControlAlgorithm);
     QuicTraceLogVerbose(SettingDestCidUpdateIdleTimeoutMs,  "[sett] DestCidUpdateIdleTimeoutMs = %u", Settings->DestCidUpdateIdleTimeoutMs);
     QuicTraceLogVerbose(SettingGreaseQuicBitEnabled,        "[sett] GreaseQuicBitEnabled   = %hhu", Settings->GreaseQuicBitEnabled);
+    QuicTraceLogVerbose(SettingEcnEnabled,                  "[sett] EcnEnabled   = %hhu", Settings->EcnEnabled);
```
nibanks, line 1216: nit: align the `= %hhu` part with the rest above.

src/core/connection.h
```diff
@@ -563,6 +564,18 @@ typedef struct QUIC_CONNECTION {
     //
     QUIC_LOSS_DETECTION LossDetection;
 
+    //
+    // ECT and CE counters for different packet number spaces.
+    //
+    uint64_t EcnEctCounters[QUIC_ENCRYPT_LEVEL_COUNT];
+    uint64_t EcnCeCounters[QUIC_ENCRYPT_LEVEL_COUNT]; // maps to ecn_ce_counters in RFC 9002.
```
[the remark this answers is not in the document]
csujedihy: Yes, that would be cleaner.

src/core/path.h
csujedihy: I thought about this but I don't know if this is the convention, using a possible value as sentinel, in QUIC.

src/core/packet_builder.c
csujedihy: If we move it `QuicSendFlush` or `QuicPacketBuilderInitialize`, which is essentially delaying the ECN decision, we will have to update each sent packet's meta data in `QuicSendFlush`, which will be an extra loop. It also conflicts with `CxPlatSendDataAlloc` where it implies the decision should be made when allocating the send context.

nibanks: `QuicSendFlush` or `QuicPacketBuilderInitialize` is before this code path here. I'm suggesting you take the logic you have here, calculate what the ECN info would be for the packets, cache it on the Builder, and then use that value here instead.

src/core/loss_detection.c
csujedihy: It is a real possibility. That's why we need to validate the counts.

src/core/packet_builder.c
csujedihy: Done

src/test/lib/DataTest.cpp
csujedihy: Fixed

src/core/loss_detection.c
csujedihy: This check is actually necessary. Otherwise, we would emit an ECN validation success log every time we receive a correctly marked packet.

src/core/loss_detection.c
```diff
@@ -1496,6 +1498,69 @@ QuicLossDetectionProcessAckBlocks(
     }
 
     if (NewLargestAck) {
+        if (Connection->Settings.EcnEnabled && Path->EcnValidationState != ECN_VALIDATION_FAILED) {
```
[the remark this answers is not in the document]
csujedihy: We don't.

src/core/connection.c
```diff
@@ -5940,6 +5941,16 @@ QuicConnResetIdleTimeout(
         IdleTimeoutMs = Connection->Settings.HandshakeIdleTimeoutMs;
     }
 
+    if (Path->EcnValidationState == ECN_VALIDATION_TESTING) {
+        uint64_t Now = CxPlatTimeUs64();
+        if (CxPlatTimeAtOrBefore64(Now, Path->EcnTestingEndingTime)) {
+            //
+            // Give ECN validation enough time to test ECN capability on the path.
+            //
+            IdleTimeoutMs += US_TO_MS(CxPlatTimeDiff64(Now, Path->EcnTestingEndingTime));
```
nibanks, line 5950: Why would you extend the idle time just to keep testing the network? If they want to kill the connection sooner, we should let them, right?

src/core/connection.c
```diff
@@ -7154,6 +7166,11 @@ QuicConnApplyNewSettings(
             Connection->State.FixedBit = (RandomValue % 2);
             Connection->Stats.GreaseBitNegotiated = TRUE;
         }
+
+        if (Connection->Settings.EcnEnabled) {
+            QUIC_PATH* Path = &Connection->Paths[0];
+            Path->EcnValidationState = ECN_VALIDATION_TESTING;
```
nibanks, line 7172: Should we have ETW logs for every time a path changes ECN validation state? It shouldn't happen that often, right?

src/core/connection.c
csujedihy: Because we can't tell if the connection is idle or all packets sent in ECN testing state are dropped. We have a similar idle timeout extending logic right below this block.

src/core/connection.c
csujedihy: I can add one for transiting to unknown state. We have logs for capable and failed.

nibanks: I would like a single (manifested) `PathEcnStateChange` event that we could automate processing in WPA if we want, please.

csujedihy: What's the point of automating this in WPA? In most cases, the validation succeeds right after the initial packet.

csujedihy: When there's actually something interesting happening, the state transition isn't quite useful. What really useful are the counter values in EcnValidationFailure log.

src/core/connection.c
nibanks: If we're actively sending packets, then we aren't idle. I don't think this change should be necessary.

src/core/connection.h
```diff
@@ -563,6 +563,12 @@ typedef struct QUIC_CONNECTION {
     //
     QUIC_LOSS_DETECTION LossDetection;
 
+    //
+    // The total number of packets sent with each corresponding ECT codepoint in all encryption
+    // level.
+    //
+    uint64_t NumPacketsSentWithEct;
```
nibanks, line 570: I think this kind of info should be put in `QUIC_SEND`.

src/core/loss_detection.c
nibanks: You resolve this, but didn't remove the check?

src/core/loss_detection.c
nibanks: We should discuss this in our meeting today.

src/core/send.c
```diff
@@ -1150,6 +1150,30 @@ QuicSendFlush(
     }
     _Analysis_assume_(Builder.Metadata != NULL);
 
+    if (Builder.Path->EcnValidationState == ECN_VALIDATION_CAPABLE) {
+        Builder.EcnEctSet = TRUE;
+    } else if (Builder.Path->EcnValidationState == ECN_VALIDATION_TESTING) {
+        if (Builder.Path->EcnTestingEndingTimeSet) {
+            if (!CxPlatTimeAtOrBefore64(
+                    CxPlatTimeUs64(), Builder.Path->EcnTestingEndingTime)) {
```
nibanks, line 1158:
```suggestion
                    TimeNow, Builder.Path->EcnTestingEndingTime)) {
```
We've already paid the cost to query time above. Let's use that.

src/core/path.h
nibanks: Let's just use `0`. For 64-bit time values zero is never valid.

src/core/send.c
```diff
@@ -1150,6 +1150,30 @@ QuicSendFlush(
...
+                Builder.Path->EcnValidationState = ECN_VALIDATION_UNKNOWN;
+                QuicTraceLogConnInfo(
+                    EcnValidationUnknown,
+                    Connection,
+                    "ECN unknown.");
+            }
+        } else {
+            uint32_t ThreePtosInUs =
+                QuicLossDetectionComputeProbeTimeout(
```
nibanks, line 1167: This will always use the default initial RTT for an estimate. Is that ok?

src/core/sent_packet_metadata.h
nibanks: Did you validate on Release or Debug?

csujedihy: I validated it on both. It's 80B regardless of build flavor.

src/core/path.h
csujedihy: updated

src/core/connection.c
csujedihy: removed

src/core/connection.h
csujedihy: done

src/core/send.c
csujedihy: `ThreePtosInUs` for a new connection is 3s. I can cap it if you want.

src/core/loss_detection.c
```diff
@@ -1745,7 +1806,8 @@ QuicLossDetectionProcessTimerOperation(
         // OldestPacket has been in the SentPackets list for at least
         // DisconnectTimeoutUs without an ACK for either OldestPacket or for any
         // packets sent more than the reordering threshold after it. Assume the
-        // path is dead and close the connection.
+        // path is dead and close the connection unless we are validating ECN
+        // capability on the path.
```
nibanks, line 1810: revert this comment change?

src/core/loss_detection.c
```diff
@@ -1710,7 +1772,6 @@ QuicLossDetectionProcessTimerOperation(
     )
 {
     QUIC_CONNECTION* Connection = QuicLossDetectionGetConnection(LossDetection);
-
```
nibanks, line 1713: nit: unnecessary change?

src/core/send.h
```diff
@@ -304,6 +304,12 @@ typedef struct QUIC_SEND {
     //
     uint16_t InitialTokenLength;
 
+    //
+    // The total number of packets sent with each corresponding ECT codepoint in all encryption
+    // level.
+    //
+    uint64_t NumPacketsSentWithEct;
```
nibanks, line 311: For slightly better padding, put this after `LastFlushTime`?

src/core/connection.c
nibanks: As discussed over IM, should anything be promoted to ETW?

csujedihy: Added connection rundown ETW event and ecn failure ETW event.

src/core/loss_detection.c
```diff
@@ -1496,6 +1498,66 @@ QuicLossDetectionProcessAckBlocks(
...
+                    EcnValidated = FALSE;
+                } else {
+                    //
+                    // TODO: Notify CC of the ECN signal before we update the CE counts.
+                    //
+                    Packets->EcnCeCounter = Ecn->CE_Count;
+                    Packets->EcnEctCounter = Ecn->ECT_0_Count;
+                    if (Path->EcnValidationState <= ECN_VALIDATION_UNKNOWN) {
+                        Path->EcnValidationState = ECN_VALIDATION_CAPABLE;
+                        QuicTraceLogConnInfo(
```
nibanks, line 1530: Shouldn't this be the same `ConnEcnCapable` event as in the rundown?

src/tools/etw/quicetw.h
```diff
@@ -344,6 +344,7 @@ typedef enum QUIC_EVENT_ID_CONNECTION {
     EventId_QuicConnClientReceivedVersionList,
     EventId_QuicConnServerSupportedVersionList,
     EventId_QuicConnBbr,
+    EventId_QuicConnEcnCapable,
```
nibanks, line 347: EcnFailed event too?

csujedihy: added

src/core/loss_detection.c
csujedihy: fixed

nibanks (approved):

PR #3172, opened 2022-10-20
Support Fixed Server ID Encoded Load Balancing Mode

nibanks:
## Description

Adds support for a new load balancing where a fixed 4-byte value can be encoded in the Server ID of the QUIC connection IDs generated by the server.

## Testing

Existing automation

## Documentation

Updated


src/core/settings.h
```diff
@@ -34,6 +34,7 @@ typedef struct QUIC_SETTINGS_INTERNAL {
             uint64_t PeerUnidiStreamCount                   : 1;
             uint64_t RetryMemoryLimit                       : 1;
             uint64_t LoadBalancingMode                      : 1;
+            uint64_t FixedServerID                          : 1;
```
nibanks, line 37: Should this be exposed to the public settings for apps to dynamically set, or should we just leverage the storage (i.e. registry) interface (for which we don't support Linux atm). I'm leaning towards exposing a public setting.

docs/Deployment.md
```diff
@@ -85,11 +85,24 @@ MsQuic currently supports a load balancing mode where the server encodes the loc
...
 
-`Server ID` is 4 bytes long, as encodes the complete IPv4 address OR the last 4 bytes of the IPv6 address.
+`Server ID` is 4 bytes long, and can encode various data interpretations.
 
-This encoding is **not enabled by default**. To configure it, set the `LoadBalancingMode` setting to `1`. **Note** - The server must be restarted for this setting to take effect.
+This encoding is **not enabled by default**. Please set the `LoadBalancingMode` setting accordingly to configure the desired load balancing mode. By default, <todo>.
 
-To use this load balancing model, the load balancer must support the model described above and be explicitly configured to enable it for your endpoint.
+> **Note**
+> Any time the `LoadBalancingMode` is updates, the server must be restarted for this setting to take effect.
```
maolson-msft, line 93:
```suggestion
> Any time the `LoadBalancingMode` is updated, the server must be restarted for this setting to take effect.
```

docs/Deployment.md
```diff
@@ -85,11 +85,24 @@ MsQuic currently supports a load balancing mode where the server encodes the loc
...
+
+## IP Address Encoded Server ID
+
+MsQuic supports encoding the local IPv4 address or IPv6 suffix into Server ID. With this load balancing model, the load balancer can use this encoded IP address/suffix to directly route received QUIC packets to the appropriate server.
+
+To configure this mode, set the `LoadBalancingMode` setting to `1`.
+
+## Fixed Server ID
+
+MsQuic supports encoding a fixed, 4-byte value into Server ID. With this load balancing model, the load balancer can use this value to lookup and route received QUIC packets to the appropriate server.
```
maolson-msft, line 103:
```suggestion
MsQuic supports encoding a fixed, 4-byte value into Server ID. The load balancer can use this value to look up and route received QUIC packets to the appropriate server.
```

docs/Deployment.md
```diff
@@ -85,11 +85,24 @@ MsQuic currently supports a load balancing mode where the server encodes the loc
...
-This encoding is **not enabled by default**. To configure it, set the `LoadBalancingMode` setting to `1`. **Note** - The server must be restarted for this setting to take effect.
+This encoding is **not enabled by default**. Please set the `LoadBalancingMode` setting accordingly to configure the desired load balancing mode. By default, <todo>.
 
-To use this load balancing model, the load balancer must support the model described above and be explicitly configured to enable it for your endpoint.
+> **Note**
+> Any time the `LoadBalancingMode` is updates, the server must be restarted for this setting to take effect.
+
+## IP Address Encoded Server ID
+
+MsQuic supports encoding the local IPv4 address or IPv6 suffix into Server ID. With this load balancing model, the load balancer can use this encoded IP address/suffix to directly route received QUIC packets to the appropriate server.
```
maolson-msft, line 97:
```suggestion
MsQuic supports encoding the local IPv4 address or IPv6 suffix into Server ID. The load balancer can use this encoded IP address/suffix to directly route received QUIC packets to the appropriate server.
```

src/core/connection.c
```diff
@@ -178,6 +178,12 @@ QuicConnAlloc(
                     ((uint8_t*)&Datagram->Route->LocalAddress.Ipv6.sin6_addr) + 12,
                     4);
             }
+        } else if (MsQuicLib.Settings.LoadBalancingMode == QUIC_LOAD_BALANCING_SERVER_ID_FIXED) {
+            CxPlatRandom(1, Connection->ServerID); // Randomize the first byte.
```
maolson-msft, line 182: Is "fixed" an industry term here? Because it's pretty confusing to call something "fixed" when it has a random component.

src/core/settings.c
```diff
@@ -373,12 +379,16 @@ QuicSettingApply(
         Destination->IsSet.RetryMemoryLimit = TRUE;
     }
     if (Source->IsSet.LoadBalancingMode && (!Destination->IsSet.LoadBalancingMode || OverWrite)) {
-        if (Source->LoadBalancingMode > QUIC_LOAD_BALANCING_SERVER_ID_IP) {
+        if (Source->LoadBalancingMode > QUIC_LOAD_BALANCING_SERVER_ID_FIXED) {
```
maolson-msft, line 382: Better to define a MAX value in QUIC_LOAD_BALANCING_MODE, so that this bit of code isn't dependent on the order of the modes in the enum

src/core/connection.c
nibanks: The first byte is randomized. The rest is fixed.

src/core/settings.c
nibanks: Will do.

src/core/connection.c
maolson-msft: k XD

docs/Deployment.md
```diff
@@ -85,11 +85,24 @@ MsQuic currently supports a load balancing mode where the server encodes the loc
...
+
+MsQuic supports encoding the local IPv4 address or IPv6 suffix into Server ID. The load balancer can use this encoded IP address/suffix to directly route received QUIC packets to the appropriate server.
+
+To configure this mode, set the `LoadBalancingMode` setting to `1`.
+
+## Fixed Server ID
+
+MsQuic supports encoding a fixed, 4-byte value into Server ID. The load balancer can use this value to look up and route received QUIC packets to the appropriate server.
+
+To configure this mode, set the `LoadBalancingMode` setting to `2` and the `FixedServerID` setting to your desired value.
```
anrossi, line 105: Are these regkeys then? or are apps using `SetParam`?

nibanks: It's an API setting that is also backed by a registry key.

anrossi (approved):

PR #3186, opened 2022-10-24
Async ticket validation

ami-GS:
## Description

Async ticket validation

## Testing

Handshake/WithHandshakeArgs1.Resume would cover
`./artifacts/bin/linux/x64_Debug_openssl/msquictest --gtest_filter="Handshake/WithHandshakeArgs1.Resume*/0"`

## Documentation

Not yet


setup.sh
```diff
@@ -0,0 +1,2 @@
+pwsh ./scripts/build.ps1 -DisableLog -DisablePerf -Parallel 12
```
nibanks, line 1: Please remove this file.

src/core/connection.c
```diff
@@ -2127,6 +2120,7 @@ QuicConnRecvResumptionTicket(
         const uint8_t* Ticket
     )
 {
+    fprintf(stderr, "%p receiving resumption ticket length:%d\n", Connection, TicketLength);
```
nibanks, line 2123: Please clean this stuff up.

src/core/connection.c
```diff
@@ -2178,20 +2172,26 @@ QuicConnRecvResumptionTicket(
...
-        if (ResumptionAccepted) {
+        if (Status == QUIC_STATUS_SUCCESS) {
             QuicTraceEvent(
                 ConnServerResumeTicket,
                 "[conn][%p] Server app accepted resumption ticket",
                 Connection);
+            fprintf(stderr, "%p Server app accepted resumption ticket length:%d\n", Connection, TicketLength);
+            ResumptionAccepted = TRUE;
+        } else if (Status == QUIC_STATUS_PENDING) {
+            Connection->Crypto.TicketValidationPending = TRUE;
```
nibanks, line 2185: Add a log about this being pending, similar to `CustomCertValidationPending`.

setup.sh
ami-GS: These are really draft to share current status...

src/core/crypto.c
```diff
@@ -1687,7 +1687,7 @@ QuicCryptoProcessDataComplete(
 
     QuicCryptoValidate(Crypto);
 
-    if (!Crypto->CertValidationPending) {
+    if (!Crypto->CertValidationPending && !Crypto->TicketValidationPending) {
```
nibanks, line 1690: Instead of doing it this way (for both ticket and cert validation) I think a better way is to just return at the beginning of this function `if (Crypto->CertValidationPending || Crypto->TicketValidationPending)`. Then, in the completion, just call this function (`QuicCryptoProcessDataComplete`) with the cached `RecvBufferConsumed `.

src/core/crypto.c
```diff
@@ -1719,7 +1719,53 @@ QuicCryptoCustomCertValidationComplete(
...
+        Crypto->ResultFlags = 0;
+
+        QUIC_CONNECTION* Connection = QuicCryptoGetConnection(Crypto);
+        QUIC_CONFIGURATION* Configuration = Connection->Configuration;
+        Crypto->Initialized = FALSE;
+        uint8_t* Buffer = Crypto->RecvBuffer.Buffer;
+        QuicCryptoInitialize(Crypto);
+        Crypto->RecvBuffer.WrittenRanges.UsedLength = 1; // To pass QuicRecvBufferRead
+        Crypto->RecvBuffer.Buffer = Buffer; // To pass "if" next to QuicCryptoTlsGetCompleteTlsMessageLength
+        QuicCryptoInitializeTls(Crypto, Configuration->SecurityConfig, Connection->HandshakeTP);
```
nibanks, line 1766: See my comment above. In the success case, you should just clear the pending flag and then call `QuicCryptoProcessDataComplete`. All this clearing of keys and re-initialization is unnecessary in the success path. You will have to do something along these lines for the failure path though.

src/core/crypto.c
```diff
@@ -1675,7 +1675,7 @@ QuicCryptoProcessDataComplete(
     _In_ uint32_t RecvBufferConsumed
     )
 {
-    if (RecvBufferConsumed != 0) {
+    if (RecvBufferConsumed != 0 && !Crypto->TicketValidationPending) {
```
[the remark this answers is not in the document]
ami-GS: I think it is already done.
The pending flag is disabled at the beginning of QuicCryptoCustomTicketValidationComplete, then the func goes to this QuicCryptoProcessDataComplete.

src/core/crypto.c
ami-GS: oh, maybe works😮

src/core/crypto.c
```diff
@@ -1675,6 +1675,13 @@ QuicCryptoProcessDataComplete(
     _In_ uint32_t RecvBufferConsumed
     )
 {
+    if (Crypto->CertValidationPending || Crypto->TicketValidationPending) {
+        if (Crypto->TicketValidationPending) {
+            Crypto->TicketValidationPendingBufferLength = RecvBufferConsumed;
```
nibanks, line 1680: Rename to `PendingValidationBufferLength`

src/core/crypto.h
```diff
@@ -39,6 +39,12 @@ typedef struct QUIC_CRYPTO {
     //
     BOOLEAN CertValidationPending : 1;
 
+    //
+    // Indicates Resumption ticket validation is under validation asynchronously
+    //
+    QUIC_SERVER_APP_TICKET_VALIDATION_STATUS TicketValidationStatus : 2;
```
nibanks, line 45: enums (which are signed) have issues (or at least in my experience they do) with bitfields. I think it might be better to simply have two BOOLEANS, "pending" and "previous failed" (or something like that).

src/inc/msquic.h
```diff
@@ -156,6 +156,12 @@ typedef enum QUIC_SERVER_RESUMPTION_LEVEL {
     QUIC_SERVER_RESUME_AND_ZERORTT,
 } QUIC_SERVER_RESUMPTION_LEVEL;
 
+typedef enum QUIC_SERVER_APP_TICKET_VALIDATION_STATUS {
+    QUIC_SERVER_VALIDATION_DONE,
+    QUIC_SERVER_VALIDATION_PENDING,
+    QUIC_SERVER_VALIDATION_REJECTING,
+} QUIC_SERVER_APP_TICKET_VALIDATION_STATUS;
```
nibanks, line 163: As I indicated already, we should just use two booleans instead, but even if we did keep the enum, it shouldn't be public (in msquic.h).

src/core/crypto.c
```diff
@@ -1675,6 +1675,13 @@ QuicCryptoProcessDataComplete(
     _In_ uint32_t RecvBufferConsumed
     )
 {
+    if (Crypto->CertValidationPending || Crypto->TicketValidationStatus == QUIC_SERVER_VALIDATION_PENDING) {
+        if (Crypto->TicketValidationStatus == QUIC_SERVER_VALIDATION_PENDING) {
+            Crypto->PendingValidationBufferLength = RecvBufferConsumed;
```
nibanks, line 1680: You shouldn't only cache this for ticket validation pending. Do for cert pending too.

src/core/crypto.c
```diff
@@ -1719,8 +1724,42 @@ QuicCryptoCustomCertValidationComplete(
...
+    _In_ BOOLEAN Result
+    )
+{
+    if (Crypto->TicketValidationStatus != QUIC_SERVER_VALIDATION_PENDING) {
+        return;
+    }
+
+    if (Result) {
+        Crypto->TicketValidationStatus = QUIC_SERVER_VALIDATION_DONE;
+        QuicCryptoProcessDataComplete(Crypto, Crypto->PendingValidationBufferLength);
```
nibanks, line 1744: `QuicCryptoCustomCertValidationComplete` should also be updated to call this now.

src/inc/msquic.h
```diff
@@ -846,6 +852,7 @@ typedef struct QUIC_SCHANNEL_CREDENTIAL_ATTRIBUTE_W {
 #endif
 #define QUIC_PARAM_CONN_STATISTICS_V2                   0x05000016  // QUIC_STATISTICS_V2
 #define QUIC_PARAM_CONN_STATISTICS_V2_PLAT              0x05000017  // QUIC_STATISTICS_V2
+#define QUIC_PARAM_CONN_RESUMPTION_TICKET_VALID         0x05000018  // uint8_t (BOOLEAN)
```
nibanks, line 855: I think we're actually going to want to make a new function in `QUIC_API_TABLE` for this. The problem with using a SetParam is that it is a blocking call to the app, which we don't want in this case. I think we want to create a `ConnectionResumptionTicketValidationComplete` function...

src/test/bin/quic_gtest.cpp
```diff
@@ -690,6 +719,64 @@ TEST_P(WithHandshakeArgs1, ResumeRejection) {
             0);     // RandomLossPercentage
     }
 }
+
+TEST_P(WithHandshakeArgs1, ResumeRejectionByServerApp) {
```
nibanks, line 723: We didn't already have tests for synchronous rejection by the server?

src/test/lib/HandshakeTest.cpp
```diff
@@ -264,6 +276,11 @@ QuicTestConnect(
                         QUIC_LOCALHOST_FOR_AF(QuicAddrFamily),
                         ServerLocalAddr.GetPort()));
 
+                if (AsyncTicketValidation) {
+                    CxPlatSleep(1000);
```
nibanks, line 280: You need to wait until after the server has accepted the connection before you can use `Server`. See the `if (AsyncConfiguration)` scope below.

src/test/lib/TestConnection.h
```diff
@@ -63,8 +63,10 @@ class TestConnection
     bool HasRandomLoss      : 1;
     bool AsyncCustomValidation : 1;
     bool CustomValidationResultSet : 1;
+    bool CustomTicketValidationResultSet : 1; // might not be needed
```
nibanks, line 66: Is it needed?

src/test/bin/quic_gtest.cpp
ami-GS: There is ResumeRejection test, which is caused by wrong ticket. This is rejected (validation failed) by OpenSSL functionality during SSL_do_handshake.
This is the case when using right ticket but server app reject

src/core/crypto.h
ami-GS: Yeah, actually single flag is enough to do same thing, but it makes implementation complicated and need to use `if (Pending) {}` for **rejecting** ticket by second call. which is not intuitive.

src/core/crypto.c
ami-GS: wow, kind of bug😮

src/test/lib/TestConnection.h
ami-GS: removed

src/inc/msquic.h
```diff
@@ -1482,6 +1495,7 @@ typedef struct QUIC_API_TABLE {
     QUIC_CONNECTION_SET_CONFIGURATION_FN
                                         ConnectionSetConfiguration;
     QUIC_CONNECTION_SEND_RESUMPTION_FN  ConnectionSendResumptionTicket;
+    QUIC_CONNECTION_COMP_RESUMPTION_FN  ConnectionResumptionTicketValidationComplete;
```
nibanks, line 1498: In order to not make this a breaking change, we need to put it at the end of the struct.

src/core/crypto.h
```diff
@@ -39,6 +39,13 @@ typedef struct QUIC_CRYPTO {
     //
     BOOLEAN CertValidationPending : 1;
 
+    //
+    // Indicates Resumption ticket validation is under validation asynchronously
+    //
+    BOOLEAN TicketValidationPending : 1;
+    BOOLEAN TicketValidationRejecting : 1;
+    uint32_t PendingValidationBufferLength;
```
nibanks, line 47: Let's put this in the Recv section of the struct, perhaps after `RecvTotalConsumed`.

src/core/api.c
```diff
@@ -1632,3 +1632,72 @@ MsQuicDatagramSend(
...
+        "[ api] Enter %u (%p).",
+        QUIC_TRACE_API_CONNECTION_COMPLETE_RESUMPTION_TICKET_VALIDATION,
+        Handle);
+
+    if (IS_CONN_HANDLE(Handle)) {
+        Connection = (QUIC_CONNECTION*)Handle;
+    } else {
+        Status = QUIC_STATUS_INVALID_PARAMETER;
+        goto Error;
+    }
```
nibanks, line 1659: Take a look at how we handle this kind of stuff in the other functions, such as `MsQuicConnectionStart`:
```c
    if (IS_CONN_HANDLE(Handle)) {
        Connection = (QUIC_CONNECTION*)Handle;
    } else if (IS_STREAM_HANDLE(Handle)) {
        QUIC_STREAM* Stream = (QUIC_STREAM*)Handle;
        CXPLAT_TEL_ASSERT(!Stream->Flags.HandleClosed);
        CXPLAT_TEL_ASSERT(!Stream->Flags.Freed);
        Connection = Stream->Connection;
    } else {
        goto Error;
    }
```

src/core/connection.c
```diff
@@ -7287,6 +7299,15 @@ QuicConnProcessApiOperation(
         }
         break;
 
+    case QUIC_API_TYPE_CONN_COMPLETE_RESUMPTION_TICKET_VALIDATION:
+        CXPLAT_DBG_ASSERT(QuicConnIsServer(Connection));
+
+        Status =
+            QuicCryptoCustomTicketValidationComplete(
+                &Connection->Crypto,
+                ApiCtx->CONN_COMPLETE_RESUMPTION_TICKET_VALIDATION.Result);
```
nibanks, line 7308:
```suggestion
        CXPLAT_DBG_ASSERT(QuicConnIsServer(Connection));
        Status =
            QuicCryptoCustomTicketValidationComplete(
                &Connection->Crypto,
                ApiCtx->CONN_COMPLETE_RESUMPTION_TICKET_VALIDATION.Result);
```

src/core/crypto.c
```diff
@@ -1719,10 +1722,68 @@ QuicCryptoCustomCertValidationComplete(
             "Custom cert validation failed.");
         QuicConnTransportError(
             QuicCryptoGetConnection(Crypto),
-            QUIC_ERROR_CRYPTO_ERROR(0xFF & CXPLAT_TLS_ALERT_CODE_BAD_CERTIFICATE));
+            QUIC_ERROR_CRYPTO_ERROR(0xFF & CXPLAT_TLS_ALERT_CODE_BAD_CERTIFICATE)); //
```
nibanks, line 1725:
```suggestion
            QUIC_ERROR_CRYPTO_ERROR(0xFF & CXPLAT_TLS_ALERT_CODE_BAD_CERTIFICATE));
```

src/core/connection.c
```diff
@@ -7287,6 +7299,15 @@ QuicConnProcessApiOperation(
         }
         break;
 
+    case QUIC_API_TYPE_CONN_COMPLETE_RESUMPTION_TICKET_VALIDATION:
+        CXPLAT_DBG_ASSERT(QuicConnIsServer(Connection));
+
+        Status =
+            QuicCryptoCustomTicketValidationComplete(
```
nibanks, line 7306: This is an async operation from the app. They don't wait on it, so they can't get any errors back. Therefore, there's no reason for this function to return an error code at all.

src/core/crypto.c
```diff
@@ -1719,10 +1722,68 @@ QuicCryptoCustomCertValidationComplete(
...
+        Crypto->TlsState.BufferOffset1Rtt = 0;
+        for (size_t i = QUIC_PACKET_KEY_0_RTT; i < QUIC_PACKET_KEY_COUNT; ++i) {
+            QuicPacketKeyFree(Crypto->TlsState.ReadKeys[i]);
+            Crypto->TlsState.ReadKeys[i] = NULL;
+            QuicPacketKeyFree(Crypto->TlsState.WriteKeys[i]);
+            Crypto->TlsState.WriteKeys[i] = NULL;
+        }
+        Crypto->RecvBuffer.ExternalBufferReference = FALSE;
+        QUIC_CONNECTION* Connection = QuicCryptoGetConnection(Crypto);
+        Status = QuicCryptoInitializeTls(Crypto, Connection->Configuration->SecurityConfig, Connection->HandshakeTP);
```
nibanks, line 1778: If this fails for some reason, we'll need to fatal this connection, via `QuicConnFatalError`.

docs/api/ConnectionResumptionTicketValidationComplete.md
```diff
@@ -0,0 +1,43 @@
+ConnectionResumptionTicketValidationComplete function
+======
+
+Uses the QUIC (server) handle to complete resumption ticket validation. This must be called after server app handles ticket validation and then return QUIC_STATUS_PENDING.
```
nibanks, line 4: Does it actually need to be called before the app returns from the callback (with `QUIC_STATUS_PENDING`) or can it be called in the callback?

nibanks: And to be clear, I think it should be callable in the callback.

ami-GS: Crypto status changes to TicketValidationPending==True after callback returns QUIC_STATUS_PENDING.
Calling this function in the callback could be indeterministic behavior.
do nothing, no error or complete

```
    if (!Crypto->TicketValidationPending || Crypto->TicketValidationRejecting) {
        return;
    }
```

ami-GS:
```c
        QuicTraceLogConnVerbose(
            IndicateResumed,
            Connection,
            "Indicating QUIC_CONNECTION_EVENT_RESUMED");
        Status = QuicConnIndicateEvent(Connection, &Event);

        if (Status == QUIC_STATUS_SUCCESS) {
            QuicTraceEvent(
                ConnServerResumeTicket,
                "[conn][%p] Server app accepted resumption ticket",
                Connection);
            ResumptionAccepted = TRUE;
        } else if (Status == QUIC_STATUS_PENDING) {
            QuicTraceEvent(
                ConnServerResumeTicket,
                "[conn][%p] Server app asynchronously validating resumption ticket",
                Connection);
            Connection->Crypto.TicketValidationPending = TRUE;
            ResumptionAccepted = TRUE;
        }
```

src/core/operation.h
```diff
@@ -49,6 +49,7 @@ typedef enum QUIC_API_TYPE {
     QUIC_API_TYPE_CONN_START,
     QUIC_API_TYPE_CONN_SET_CONFIGURATION,
     QUIC_API_TYPE_CONN_SEND_RESUMPTION_TICKET,
+    QUIC_API_TYPE_CONN_COMPLETE_RESUMPTION_TICKET_VALIDATION,
```
nibanks, line 52: `quictypes.h` has a mirror of this for the windbg plugin. Please update?

docs/api/ConnectionResumptionTicketValidationComplete.md
```diff
@@ -0,0 +1,40 @@
...
+
+# Return Value
+
+The function returns a [QUIC_STATUS](QUIC_STATUS.md). The app may use `QUIC_FAILED` or `QUIC_SUCCEEDED` to determine if the function failed or succeeded.
+
+# Remarks
+
+- Schannel doesn't support this feature.
+- Available from v2.2
+# See Also
```
nibanks, line 36: Thanks GitHub... this is what I meant:
```suggestion
- Available from v2.2

# See Also
```

src/core/connection.c
```diff
@@ -2190,20 +2195,27 @@ QuicConnRecvResumptionTicket(
             IndicateResumed,
             Connection,
             "Indicating QUIC_CONNECTION_EVENT_RESUMED");
-        ResumptionAccepted =
-            QUIC_SUCCEEDED(QuicConnIndicateEvent(Connection, &Event));
-
-        if (ResumptionAccepted) {
+        Status = QuicConnIndicateEvent(Connection, &Event);
+        Connection->Crypto.TicketValidationPending = Status == QUIC_STATUS_PENDING;
```
nibanks, line 2199: This flag may have already been cleared in the callback. Don't set it to pending again if `Status == QUIC_STATUS_PENDING`. Same for cert validation path.

src/test/lib/HandshakeTest.cpp
```diff
@@ -278,6 +290,11 @@ QuicTestConnect(
                     }
                 }
 
+                if (AsyncTicketValidation) {
+                    CxPlatSleep(1000);
+                    TEST_QUIC_SUCCEEDED(Server->SetCustomTicketValidationResult(SessionResumption == QUIC_TEST_RESUMPTION_ENABLED_ASYNC));
```
nibanks, line 295: You need to wait for `ServerAcceptCtx.NewConnectionReady` just like the `if (AsyncConfiguration)` scope above.

src/plugins/dbg/quictypes.h
```diff
@@ -992,6 +992,7 @@ typedef enum QUIC_API_TYPE {
     QUIC_API_TYPE_GET_PARAM,
 
     QUIC_API_TYPE_DATAGRAM_SEND,
+    QUIC_API_TYPE_CONN_COMPLETE_RESUMPTION_TICKET_VALIDATION,
 
 } QUIC_API_TYPE;
```
nibanks, line 997: This needs to be an exact copy from operation.h

src/inc/msquic.h
```diff
@@ -1493,6 +1506,8 @@ typedef struct QUIC_API_TABLE {
 
     QUIC_DATAGRAM_SEND_FN               DatagramSend;
 
+    QUIC_CONNECTION_COMP_RESUMPTION_FN  ConnectionResumptionTicketValidationComplete; // Available from v2.2
```
nibanks, line 1509: We should add support to spinquic to call this.

nibanks (approved):
Thanks for handling all my comments!

nibanks (approved):

PR #3246, opened 2022-11-22
HyStart++ for MsQuic

anrossi:
## Description

Implement HyStart++.

## Testing

WAN Perf tests in CI.

## Documentation

TODO doc changes for settings.


maolson-msft:
update abstract comment at top of cubic.c

src/core/cubic.h
```diff
@@ -32,6 +38,11 @@ typedef struct QUIC_CONGESTION_CONTROL_CUBIC {
     //
     BOOLEAN TimeOfLastAckValid : 1;
 
+    //
+    // HyStart state.
+    //
+    HYSTART_STATE HyStartState : 2;
```
maolson-msft, line 44: Why isn't this next to the other hystart fields?

src/core/cubic.h
```diff
@@ -7,6 +7,12 @@
 
 #pragma once
 
+typedef enum HYSTART_STATE {
```
nibanks, line 10: Generally, I think we follow the convention of having `QUIC_` at the beginning of all types (enums included). Let's update this one.

src/core/cubic.c
```diff
@@ -311,6 +372,7 @@ CubicCongestionControlOnCongestionEvent(
         Cubic->KCubic = S_TO_MS(Cubic->KCubic);
         Cubic->KCubic >>= 3;
 
+        CubicCongestionHyStartChangeState(Cc, HYSTART_DONE);
```
maolson-msft, line 375: Feels weird that this is excluded from the persistent congestion case, and also I feel like a fast path for when the connection is already in HYSTART_DONE might make sense here

src/core/cubic.h
nibanks: Also, be very careful with bitfields and enums. Generally enums are signed values (`int`) and if you use a bitfield to hold it, the negative representation won't match the enum value...

src/core/cubic.c
```diff
@@ -83,6 +96,49 @@ QuicConnLogCubic(
...
+        Cubic->CWndSlowStartGrowthDivisor = 1;
+        break;
+    default:
+        CXPLAT_FRE_ASSERT(FALSE);
+    }
+
+    if (Cubic->HyStartState != NewHyStartState) {
+        QuicTraceEvent(
+            ConnHyStartStateChange,
+            "[conn][%p] HyStart: State=%u CongestionWindow=%u SlowStartThreshold=%u",
```
nibanks, line 122:
```suggestion
            "[conn][%p] HyStart: State=%hhu CongestionWindow=%u SlowStartThreshold=%u",
```
Your ETW event states `State` is a `uint8_t` not `uint32_t`. Which is it?

src/core/cubic.c
```diff
@@ -767,6 +897,11 @@ CubicCongestionControlInitialize(
     Cubic->InitialWindowPackets = Settings->InitialWindowPackets;
     Cubic->CongestionWindow = DatagramPayloadLength * Cubic->InitialWindowPackets;
     Cubic->BytesInFlightMax = Cubic->CongestionWindow / 2;
+    Cubic->MinRttInCurrentRound = UINT32_MAX;
+    Cubic->MinRttInLastRound = UINT32_MAX;
+    Cubic->HyStartAckCount = 0;
+    Cubic->HyStartRoundEnd = Connection->Send.NextPacketNumber;
+    CubicCongestionHyStartResetPerRttRound(Cubic);
```
csujedihy, line 904: Is this call redundant? You are setting all that values that get reset in CubicCongestionHyStartResetPerRttRound right above the call/

csujedihy: same in CubicCongestionControlReset

src/core/cubic.c
```diff
@@ -426,6 +507,52 @@ CubicCongestionControlOnDataAcknowledged(
...
+                    //
+                    if (Cubic->MinRttInLastRound != UINT32_MAX &&
+                        Cubic->MinRttInCurrentRound != UINT32_MAX &&
+                        (Cubic->MinRttInCurrentRound >= Cubic->MinRttInLastRound + Eta)) {
+                        //
+                        // We found an exit. Now we are going to do Conservative Slow Start for
+                        // ConservativeSlowStartRounds rounds.
+                        //
+                        CubicCongestionHyStartChangeState(Cc, HYSTART_FOUND);
+                        Cubic->CWndSlowStartGrowthDivisor = ConservativeSlowStartGrowthDivisor;
```
csujedihy, line 540: >CWndSlowStartGrowthDivisor [](http://example.com/codeflow?start=31&length=26)

It doesn't look like this variable has any effect. The cwnd growth logic doesn't use it.

src/core/cubic.c
```diff
@@ -83,6 +96,49 @@ QuicConnLogCubic(
...
+CubicCongestionHyStartChangeState(
+    _In_ QUIC_CONGESTION_CONTROL* Cc,
+    _In_ HYSTART_STATE NewHyStartState
+    )
+{
+    QUIC_CONNECTION* Connection = QuicCongestionControlGetConnection(Cc);
+    QUIC_CONGESTION_CONTROL_CUBIC* Cubic = &Cc->Cubic;
+    switch (NewHyStartState) {
+    case HYSTART_FOUND:
+        Cubic->HyStartCongestionWindow = Cubic->CongestionWindow;
```
csujedihy, line 109: >Cubic->HyStartCongestionWindow = Cubic->CongestionWindow; [](http://example.com/codeflow?start=8&length=57)

You don't have to do this if you have no plan to log HyStartCongestionWindow.

csujedihy:
>     - Early slowstart exit via HyStart or similar.

not future work anymore

---
Refers to: src/core/cubic.c:12 in 058cfcb. [](commit_id = 058cfcbc5348856fc34681dc7f5b2d06474fc377, deletion_comment = False)

src/core/cubic.c
```diff
@@ -20,6 +20,19 @@ Future work:
...
 #include "cubic.h"
 
+#define HYSTART_DEFAULT_N_SAMPLING 8
+#define HYSTART_DEFAULT_MIN_ETA 4000 // 4ms
+#define HYSTART_DEFAULT_MAX_ETA 16000 // 16ms
+#define CONSERVATIVE_SLOW_START_DEFAULT_ROUNDS 5
+#define CONSERVATIVE_SLOW_START_DEFAULT_GROWTH_DIVISOR 4
+
+// TODO: Move these into the QUIC_SETTINGS struct and build regkeys for them
+uint16_t ConservativeSlowStartRounds = CONSERVATIVE_SLOW_START_DEFAULT_ROUNDS;
```
csujedihy, line 30: >uint16_t [](http://example.com/codeflow?start=0&length=8)

const uint16_t?

src/core/cubic.h
anrossi: Was trying to save some bits by stuffing it into the extra space left in the BOOLEAN above. Moving it now.

src/core/cubic.c
```diff
@@ -404,13 +446,86 @@ CubicCongestionControlOnDataAcknowledged(
         goto Exit;
     }
 
+    //
+    // Update HyStart++ RTT sample.
+    //
+    if (AckEvent->MinRttValid) {
+        uint32_t Eta;
```
nibanks, line 453: Why is `Eta` defined here and not in the inner most scope it's used?

src/core/cubic.c
```diff
@@ -404,13 +446,86 @@ CubicCongestionControlOnDataAcknowledged(
...
+            // Update Min RTT for the first N ACKs.
+            //
+            if (Cubic->HyStartAckCount < QUIC_HYSTART_DEFAULT_N_SAMPLING) {
+                Cubic->MinRttInCurrentRound =
+                    CXPLAT_MIN(
+                        Cubic->MinRttInCurrentRound,
+                        AckEvent->MinRtt);
+                Cubic->HyStartAckCount++;
+            } else {
+                Eta =
```
nibanks, line 466: This is only used inside the next if statement. And all the branches necessary to evaluate the min/max should be easily combinable with the if on line 477.

src/core/cubic.c
```diff
@@ -404,13 +446,86 @@ CubicCongestionControlOnDataAcknowledged(
...
+                    CXPLAT_MIN(
+                        QUIC_HYSTART_DEFAULT_MAX_ETA,
+                        CXPLAT_MAX(
+                            QUIC_HYSTART_DEFAULT_MIN_ETA,
+                            Cubic->MinRttInLastRound / 8));
```
nibanks, line 471: Where does the `8` come from?

csujedihy:
> CubicCongestionControlReset(

I think reset should allow us to enter hystart again because we are on a new path.

---
Refers to: src/core/cubic.c:146 in 93a4e97. [](commit_id = 93a4e97709c6f99b380392611e09fcbba2fba5b1, deletion_comment = False)

src/core/cubic.c
nibanks: Should this be `QUIC_HYSTART_DEFAULT_N_SAMPLING` right?

src/core/cubic.c
anrossi: What do you mean by "fast path"?

src/core/cubic.c
csujedihy: Nope. It's just another constant value.

src/core/cubic.c
maolson-msft: I mean don't do all that stuff in CubicCongestionHyStartChangeState if we are already in HYSTART_DONE

src/inc/msquic.h
```diff
@@ -679,6 +680,7 @@ typedef struct QUIC_SETTINGS {
     uint8_t MaxOperationsPerDrain;
     uint8_t MtuDiscoveryMissingProbeCount;
     uint32_t DestCidUpdateIdleTimeoutMs;
+    uint8_t HyStartEnabled                  : 1;
```
csujedihy: If making a new bit flags field involves a lot of work, I'd use a bigger field like uint32_t or uint64_t.

csujedihy: I am sorry EcnEnabled used the last remaining bit :)

anrossi: I think we can get away with using a `uint8_t` for now and make a larger bitfield before the next release to try to stave off this problem.

anrossi: Opened #3276 to track this further

csujedihy (approved):

PR #3267, opened 2022-12-05
Use Ubuntu 20.04 to Build OpenSSL-Systemcrypto Variant

nibanks:
## Description

Fixes https://github.com/actions/runner-images/issues/6697 by making sure to use 20.04 to build, not latest.

## Testing

Existing CI.

## Documentation

N/A

### DIFF
--- a/.azure/azure-pipelines.ci.yml
+++ b/.azure/azure-pipelines.ci.yml
@@ -387,7 +387,7 @@ stages:
       extraName: 'codecheck'
   - template: ./templates/build-config-user.yml
     parameters:
-      image: ubuntu-latest
+      image: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


mtfriesen (dismissed):

anrossi (approved):

anrossi:
Based on the build failure, even on retry, it looks like there's someplace else using the "ubuntu-latest" which should be fixed in this PR.

nibanks:
> Based on the build failure, even on retry, it looks like there's someplace else using the "ubuntu-latest" which should be fixed in this PR.

Yep. Good catch. Just pushed.

anrossi (approved):

PR #3274, opened 2022-12-08
Refactor epoll Send Path

nibanks:
## Description

Lots of refactoring for the epoll send path.

## Testing

Existing CI

## Documentation

N/A


src/test/lib/MtuTest.cpp
```diff
@@ -14,21 +14,6 @@
 #include "MtuTest.cpp.clog.h"
 #endif
 
-struct MtuTestContext {
```
nibanks, line 17: Remove this because (a) it was actually unnecessary because the `MsQuicAutoAcceptListener` already supported everything the test needed and (b) actually had a use-after-free in the test failure case (which I hit while developing).

src/platform/datapath_epoll.c
```diff
@@ -2579,27 +2211,204 @@ CxPlatSocketSend(
...
+        CASTED_CLOG_BYTEARRAY(sizeof(Route->LocalAddress), &Route->LocalAddress));
+
+    //
+    // Cache the address, mapping the remote address as necessary.
+    //
+    CxPlatConvertToMappedV6(&Route->RemoteAddress, &SendData->RemoteAddress);
+    SendData->LocalAddress = Route->LocalAddress;
+
+    //
+    // Check to see if we need to pend because there's already queue.
```
nibanks, line 2238: To align with Windows user mode, I think we should **always** queue to the datapath thread (which may or may not be different from the QUIC thread) to do the TX, otherwise I don't think we actually get much (any?) parallelization of the QUIC and UDP logic on TX.

src/platform/datapath_epoll.c
```diff
@@ -589,14 +552,17 @@ CxPlatDataPathInitialize(
...
-    QUIC_STATUS Status = CxPlatDataPathQuerySockoptSupport(Datapath);
-    if (QUIC_FAILED(Status)) {
-        CXPLAT_FREE(Datapath, QUIC_POOL_DATAPATH);
-        return Status;
+    if (Datapath->Features & CXPLAT_DATAPATH_FEATURE_SEND_SEGMENTATION) {
+        Datapath->SendDataSize = sizeof(CXPLAT_SEND_DATA);
+        Datapath->SendIoVecCount = 1;
+    } else {
+        const uint32_t SendDataSize =
+            sizeof(CXPLAT_SEND_DATA) + (CXPLAT_MAX_BATCH_SEND - 1) * sizeof(struct iovec);
```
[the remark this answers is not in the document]
nibanks: So the code I have might just allocate a bit more because of possible padding? I'm fine with that.

PR #3282, opened 2022-12-12
Simplify winuser SQE->CQE mapping and remove IO completion keys

mtfriesen:
## Description

During the RIO prototyping, the winuser SQE->CQE completion path relying on IO completion keys in addition to IO overlapped structs was both complex and inflexible. Follow the common Windows idiom and align with the cross-platform SQE->CQE user data abstraction by providing the user data submitted by each SQE to the CQE upon completion.

## Testing

CI/CD.

## Documentation

No, internal changes only.


src/platform/datapath_winuser.c
```diff
@@ -4084,12 +4114,20 @@ CxPlatDataPathProcessCqe(
...
-        CXPLAT_SOCKET_PROC* SocketProc =
-            CONTAINING_RECORD(CxPlatCqeUserData(Cqe), CXPLAT_SOCKET_PROC, IoSqe);
-        if (Cqe->lpOverlapped == &SocketProc->IoSqe.Sqe) {
-            CxPlatDataPathSocketProcessReceiveCompletion(SocketProc, Cqe);
-        } else {
-            CxPlatDataPathSocketProcessSendCompletion(SocketProc, Cqe);
+        DATAPATH_IO_SQE* Sqe =
+            CONTAINING_RECORD(CxPlatCqeUserData(Cqe), DATAPATH_IO_SQE, DatapathSqe);
+
+        switch (Sqe->IoType) {
```
mtfriesen, line 4120: This is the motivation of the PR: each IO needs to be able to provide an arbitrary completion context. Without this pattern, we were stuck with the fragile completion pattern before: one code path for when the overlapped IO matched the IO completion key, a separate code path for when it _didn't_ match the IO completion key. We then had an assortment of thread-the-needle stateful checks and magic parameter values to distinguish between completions.

src/platform/datapath_raw_xdp.c
```diff
@@ -669,6 +688,21 @@ CxPlatDpRawInterfaceInitialize(
                 &Queue->RxPool, (PSLIST_ENTRY)&Queue->RxBuffers[j * RxPacketSize]);
         }
 
+        //
+        // Disable automatic IO completions being queued if the call completes
+        // synchronously.
+        //
+        if (!SetFileCompletionNotificationModes(
```
nibanks, line 695: Oh, this wasn't already the default?

nibanks (approved):

src/platform/datapath_raw_xdp.c
mtfriesen: It was not, apparently.

PR #3284, opened 2022-12-13
Support Inline StreamSends

nibanks:
## Description

Adds support for StreamSend calls that already happen on a callback to immediately flush inline, eliminating extra queuing delays (and sometimes eve additional packets sent on the wire).

One tricky piece here is the possibilities of additional callbacks as a result. For now, I've disabled when using send buffering because that very likely will complete the send immediately if it can be buffered.

## Testing

CI/CD

## Documentation

N/A


mtfriesen (approved):

nibanks:
This seems to be crashing the release build, kernel mode tests somehow. I think they somehow generate re-entrant callbacks, and are hitting our asserts. I'm going to turn this back to draft for now, and see if I can fix the latency issue with ACK tolerance.

mtfriesen (approved):

PR #3287, opened 2022-12-13
Always Poll on XDP Perf

nibanks:
## Description

Looking at the perf data shows that the first request on an XDP connection has a higher tail latency because we were interrupt mode; idle, waiting to be woken. This adds a significant latency, and doesn't really reflect what we want to test.

## Testing

Perf automation

## Documentation

N/A


mtfriesen (approved):

nibanks:
Causing issues. Converting to draft.

PR #3294, opened 2022-12-15
WPA to support LTTng trace visualization

ami-GS:
## Description

Support LTTng  trace visualization on WPA

## Testing

## Documentation


src/plugins/trace/dll/Cookers/BaseEventCooker.cs
```diff
@@ -1,29 +1,22 @@
-﻿//
-// Copyright (c) Microsoft Corporation.
```
nibanks, line 2: Make sure all files have copyright.

src/plugins/trace/dll/DataModel/ETW/QuicEtwEvent.cs
```diff
@@ -101,209 +103,215 @@ private static QuicObjectType ComputeObjectType(TraceEvent evt)
...
         {
-            var id = (QuicEventId)evt.ID;
-            var processor = (ushort)evt.ProcessorNumber;
-            var processId = (uint)evt.ProcessID;
-            var threadId = (uint)evt.ThreadID;
-            var pointerSize = evt.PointerSize;
-            var data = new QuicEtwDataReader(evt.DataStart.ToPointer(), evt.EventDataLength, pointerSize);
-
-            switch (id)
+            switch (id.ToString())
```
nibanks, line 110: This shouldn't be necessary for ETW since we do have the ID to work with. Only LTTng event processing would need to do it by name.

src/plugins/trace/dll/DataModel/QuicEvent.cs
```diff
@@ -196,7 +196,8 @@ public enum QuicEventId : ushort
         TlsErrorStatus,
         TlsMessage,
 
-        DatapathSend = 9217,
+        Temporal = 9216, // Temporary, while there are still builds out there generating this old event
+        DatapathSend,
```
nibanks, line 200: Why?

ami-GS: There was `(QuicEventId)9216` hardcoded in QuicEtwEvent.cs, but this doesn't work with nameof() for QuicLTTngEvent.cs
I added additional enum to cover both cases.

nibanks (approved):

PR #3304, opened 2022-12-18
Linux Receive Path Improvements

nibanks:
## Description

The main improvement is the addition of GRO support. Beyond that, there are various refactorings of the receive path (to match recent send path changes).

## Testing

Existing automation

## Documentation

N/A


nibanks:
Interestingly, queuing on the send path still doesn't improve things, even with this change.

ami-GS (approved):

PR #3318, opened 2022-12-31
new API for async certificate validation

ami-GS:
## Description

Async certificate validation

PR #3320, opened 2023-01-02
Revert "new API for async certificate validation (#3318)"

ami-GS:
This reverts commit 97616f7601e6b5a021102fff09c2088cd85f9b6f.

## Description

This commit break schannel case


nibanks:
You don't need to revert. We can just wait for the fix.

PR #3331, opened 2023-01-06
Fix QUIC_SETTINGS param validation tests.

### LINKED ISSUE #3330: Downlevel tests fail on QUIC_SETTINGS size change
The issue manifests in (at least) two test failures where the size of the `QUIC_SETTINGS` struct is tested for direct equality.
The first issue is in ApiTest.c line ~2758
```c
            uint32_t Length = 0;
            TEST_QUIC_STATUS(
                QUIC_STATUS_BUFFER_TOO_SMALL,
                MsQuic->GetParam(
                    Configuration.Handle,
                    QUIC_PARAM_CONFIGURATION_SETTINGS,
                    &Length,
                    nullptr));
            TEST_EQUAL(Length, sizeof(QUIC_SETTINGS));
```
The `QUIC_SETTINGS` should be checked whether it has shrunk in size, not grown. 

The second issue is in 
```c
void SimpleGetParamTest(HQUIC Handle, uint32_t Param, size_t ExpectedLength, void* ExpectedData) {
    uint32_t Length = 0;
    TEST_QUIC_STATUS(
        QUIC_STATUS_BUFFER_TOO_SMALL,
        MsQuic->GetParam(
                Handle,
                Param,
                &Length,
                nullptr));
    if (ExpectedLength != Length) {
        TEST_FAILURE("ExpectedLength (%u) != Length (%u)", ExpectedLength, Length);
        return;
    }
```
This is a helper function used to validate various parameters so an exact may be desirable except in the case of `QUIC_SETTINGS`

anrossi:
## Description

Fixes #3330 

## Testing

Local testing and CI

## Documentation

N/A

### DIFF
--- a/src/test/lib/ApiTest.cpp
+++ b/src/test/lib/ApiTest.cpp
@@ -2702,6 +2702,9 @@ void QuicTestRegistrationParam()
     }
 }
 
+#define SETTINGS_SIZE_THRU_FIELD(SettingsType, Field) \
+    (FIELD_OFFSET(SettingsType, Field) + sizeof(((SettingsType*)0)->Field))
+
 void QuicTestConfigurationParam()
 {
     MsQuicRegistration Registration;
--- a/src/test/lib/ApiTest.cpp
+++ b/src/test/lib/ApiTest.cpp
@@ -2185,7 +2185,7 @@ void QuicTestGlobalParam()
         // in src/core/packet.h QUIC_VERSION_INFO and QuicSupportedVersionList are defined
         // but dependency issue happen when including it.
         // sizeof(QUIC_VERSION_INFO[4]) is 88 * 4
-        SimpleGetParamTest(nullptr, QUIC_PARAM_GLOBAL_SUPPORTED_VERSIONS, 88 * 4, nullptr);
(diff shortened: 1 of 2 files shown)


csujedihy (dismissed):

src/test/lib/ApiTest.cpp
```diff
@@ -3512,7 +3522,7 @@ void QuicTestConnectionParam()
             TestScopeLogger LogScope1("GetParam");
             MsQuicConnection Connection(Registration);
             TEST_QUIC_SUCCEEDED(Connection.GetInitStatus());
-            SimpleGetParamTest(Connection.Handle, QUIC_PARAM_CONN_SETTINGS, sizeof(QUIC_SETTINGS), nullptr);
+            SimpleGetParamTest(Connection.Handle, QUIC_PARAM_CONN_SETTINGS, sizeof(QUIC_SETTINGS), nullptr, true);
```
nibanks, line 3525: I believe every usage of `SimpleGetParamTest` with a `QUIC_PARAM_*_SETTINGS` needs to pass `true` now.

nibanks: Including `QUIC_PARAM_GLOBAL_SETTINGS`.

nibanks:
I updated the code to account for other parameters that might grow in future versions. I also updated the code to only validate the expected data for the given expected length.

nibanks (approved):

PR #3344, opened 2023-01-10
Adds Hashtable Restructuring

### LINKED ISSUE #3328: StreamStart performance drops significantly when too many streams are open
### Describe the bug

When more than a few hundred streams are open, the performance of the `StreamStart` function drops significantly. As the number of active streams grows beyond the default `StreamSet->StreamTable` size, the insertion performance takes a severe hit, most likely because of the high load factor of the hashtable.

### Affected OS

- [X] Windows
- [ ] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

I only tested on Windows, but I assume other platforms are affected similarly.

### MsQuic version

main

In our scenario, we had a client-server pair, with the client calling `StreamOpen` + `StreamSend(START | FIN)` at a fixed rate of $R$. Beyond a certain point, when the number of active streams is a few hundred, the performance drops, and the client grinds to a halt, spending the majority of its time in `QuicStreamSetInsertStream() -> CXPlatHashtableInsert()`.

### Expected behavior

Graceful performance degradation of `StreamStart` as the number of active streams increases.

### Actual outcome

Thrashing.

### Additional details

Flame graph of our experiment for reference:

image

nibanks:
## Description

Fixes #3328. Updates the hashtable code to expand as necessary. Steals some Windows TCP/IP code for the logic. Might not be the best, but at least is some place to start.

## Testing

Existing automation

## Documentation

N/A


src/platform/hashtable.c
```diff
@@ -1389,11 +1402,82 @@ CxPlatHashTableContract(
...
+// The maximum percentage of empty buckets in the hash table. If a hash table
+// has more empty buckets, it needs to be restructured.
+//
+#define CXPLAT_HASHTABLE_MAX_EMPTY_BUCKET_PERCENTAGE    25
+
+//
+// The maximum number of hash table restructions allowed at a time. This limits
+// the time spent expanding/contracting a hash table at dispatch.
+//
+#define CXPLAT_HASHTABLE_MAX_RESTRUCT_ATTEMPTS          8
```
nibanks, line 1424: IMHO, this number seems a bit high. Maybe 2 or 4 instead?

mtfriesen: Yeah idk - it's a magic number. Probably even `1` would suffice as long as the load factor is greater than `1` since any single failure to expand will be compensated on the next hashtable insert.

mtfriesen: It's absolutely an edge case, though. I'd just leave it alone until data indicates otherwise.

nibanks: Since we're going to call this after every single insert, why not just 1 restructure at a time?

mtfriesen: Yeah, fine by me.

src/platform/hashtable.c
```diff
@@ -81,6 +81,45 @@ CXPLAT_STATIC_ASSERT(
     CXPLAT_HASH_MIN_SIZE == BASE_HASH_TABLE_SIZE,
     "Hash table sizes should match!");
 
+//
+// The maximum number of hash table restructions allowed at a time. This limits
```
maolson-msft, line 85:
```suggestion
// The maximum number of hash table resizes allowed at a time. This limits
```

src/platform/hashtable.c
```diff
@@ -81,6 +81,45 @@ CXPLAT_STATIC_ASSERT(
...
+// the time spent expanding/contracting a hash table at dispatch.
+//
+#define CXPLAT_HASHTABLE_MAX_RESTRUCT_ATTEMPTS          1
```
maolson-msft, line 88: Change to CXPLAT_HASHTABLE_MAX_RESIZE_ATTEMPTS

src/platform/hashtable.c
```diff
@@ -81,6 +81,45 @@ CXPLAT_STATIC_ASSERT(
...
+
+//
+// The maximum average chain length in a hash table bucket. If a hash table's
+// average chain length goes above this limit, it needs to be restructured.
```
maolson-msft, line 92:
```suggestion
// average chain length goes above this limit, it needs to be expanded.
```

src/platform/hashtable.c
```diff
@@ -81,6 +81,45 @@ CXPLAT_STATIC_ASSERT(
...
+//
+#define CXPLAT_HASHTABLE_MAX_CHAIN_LENGTH               4
+
+//
+// The maximum percentage of empty buckets in the hash table. If a hash table
+// has more empty buckets, it needs to be restructured.
```
maolson-msft, line 98:
```suggestion
// has more empty buckets, it needs to be contracted.
```

src/platform/hashtable.c
```diff
@@ -81,6 +81,45 @@ CXPLAT_STATIC_ASSERT(
...
+//
+#define CXPLAT_HASHTABLE_MAX_EMPTY_BUCKET_PERCENTAGE    25
+
+#if CXPLAT_HASHTABLE_CONTRACT_SUPPORT
+uint32_t
+CxPlatHashtableGetEmptyBuckets(
```
maolson-msft, line 104: CxPlatHashtableGetEmptyBucketCount

maolson-msft: eh, nevermind

maolson-msft (approved):

PR #3343, opened 2023-01-10
gamecore_console: default to latest Windows SDK

jianye-chen:
## Description

This is the follow-up PR to [#3332](https://github.com/microsoft/msquic/pull/3332).  
We hard coded `CMAKE_SYSTEM_VERSION` to `10.0.18362.0`, which also applied to our `gamecore_console` releases. According to [Windows 10 Software Development Kit (Win10SDK)](https://learn.microsoft.com/en-us/gaming/gdk/_content/gc/getstarted/overviews/sdk-and-tools#windows-10-software-development-kit-win10sdk), `WINAPI_FAMILY_GAMES` is only usable on or after `Windows 10 SDK version 19041 or later`. 

In `10.0.18362.0`, `winapifamily.h` mistakenly categories `WINAPI_FAMILY_GAMES` under `WINAPI_PARTITION_SYSTEM`,
```cpp
/*
 * The mapping between families and partitions is summarized here.
 * An X indicates that the given partition is active for the given
 * platform/family.
 *
 *                                +---------------------------+
 *                                |      *Partition*          |
 *                                +---+---+---+---+---+---+---+
 *                                |   |   |   |   |   |   |   |
 *                                |   |   |   |   |   |   |   |
 *                                |   |   |   | P |   |   |   |
 *                                |   |   |   | H |   |   |   |
 *                                | D |   |   | O |   |   |   |
 *                                | E |   | P | N | S | S |   |
 *                                | S |   | C | E | Y | E | G |
 *                                | K |   | _ | _ | S | R | A |
 *                                | T | A | A | A | T | V | M |
 * +-------------------------+----+ O | P | P | P | E | E | E |
 * |     *Platform/Family*       \| P | P | P | P | M | R | S |
 * +------------------------------+---+---+---+---+---+---+---+
 * | WINAPI_FAMILY_DESKTOP_APP    | X | X | X |   |   |   |   |
 * +------------------------------+---+---+---+---+---+---+---+
 * |      WINAPI_FAMILY_PC_APP    |   | X | X |   |   |   |   |
 * +------------------------------+---+---+---+---+---+---+---+
 * |   WINAPI_FAMILY_PHONE_APP    |   | X |   | X |   |   |   |
 * +----------------------------- +---+---+---+---+---+---+---+
 * |      WINAPI_FAMILY_SYSTEM    |   |   |   |   | X |   |   |
 * +----------------------------- +---+---+---+---+---+---+---+
[...]

PR #3348, opened 2023-01-11
gamecore_console: fix build with 10.0.20348.0

jianye-chen:
## Description

Fix `gamecore_console` build failure on SDK `10.0.20348.0`. `WINAPI_PARTITION_GAMES` never exposed `UNICODE_STRING`.  
Bumping the minimum required SDK version to >= `10.0.20348.0` as we are using a new API (`SetThreadIdealProcessor`) exposed to `WINAPI_PARTITION_GAMES` since `10.0.20348.0`.  
See #3343.  

## Testing

Built `10.0.20348.0`, `10.0.22000.0`, `10.0.22621.0` for `gamecore_console` and verified the build passed. Binaries produced no longer import `winmm.dll`.  

## Documentation

N/A

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,8 +111,8 @@ if(NOT "${QUIC_LOGGING_TYPE}" MATCHES "^(etw|lttng|stdout|)$")
 endif()
 
 if (QUIC_GAMECORE_BUILD)
-    if(${CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION} VERSION_LESS "10.0.19041.0")
-        message(ERROR "gamecore builds require Windows 10 SDK version 19041 or later.")
+    if(${CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION} VERSION_LESS "10.0.20348.0")
+        message(ERROR "gamecore builds require Windows 10 SDK version 20348 or later.")
     endif()
 endif()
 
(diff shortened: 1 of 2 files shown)


nibanks (approved):

jianye-chen:
Looks like the internal pipeline failed again because the latest SDK available on `WindowsContainerImage2` is still `10.0.19041.0`.  
`vse2019` image does not include `10.0.20348.0` and won't work for `gamecore_console`. In #2978 we switched to a special image built in on top of `cdpxwin1809.azurecr.io/global/vse2019`. 

`vse2022` includes the required SDKs as well as `10.0.18362.0`, we might want to bump that image to `vse2022`. :/

jianye-chen:
Do we have someone experienced with [OneBranch](https://aka.ms/onebranch) ACRs?  
  
The internal pipeline failure is specific to our OneBranch builds and the ideal solution would be to switch `cdpxwin1809.azurecr.io/global/vse2019` over to `onebranch.azurecr.io/windows/ltsc2019/vse2022`. Presumably, updating the build tools shouldn't introduce new failures since we are pinning to a specific SDK version for most windows builds outside of `gamecore_console` and `uwp`.  
  
Alternatively, we would need to install `Microsoft.VisualStudio.Component.Windows10SDK.20348` through the vs installer.  
  
Without CI on OneBranch configs, it will be risky to make changes to the docker images as we would only be able to discover the failures after our changes are merged into `main`.

PR #3349, opened 2023-01-12
Upload linux perf command results

ami-GS:
## Description

Performance pipeline to publish perf command results
- Fix Linux's Remote test issue
  - `tar`'s warning to stop server
  - uploading result issue by empty SMB directory name
- Add note for problematic design where client waits server's stdout

## Documentation
src/plugins/trace/README.md


scripts/log.ps1
```diff
@@ -107,6 +137,50 @@ if ($IsLinux) {
...
+        # error
+    } else {
+        New-Item -Path $TempPerfDir -ItemType Directory -Force
+        $CommandSplit = $Command.Split(" ")
+        $OutFile = "server.perf.data"
+        if (!$Remote) {
+            $count = @(Get-ChildItem $TempPerfDir "client_*.perf.data").count
+            $OutFile = "client_$count.perf.data"
+        }
+        perf record -F 10 -a -g -o $(Join-Path $TempPerfDir $OutFile) $CommandSplit[0] $CommandSplit[1..$($CommandSplit.count-1)]
```
nibanks, line 160: The `-F 10` means only 10 times a second, right? That seems way too infrequent. Really, I think we want max.

ami-GS: Sure, made lower for test purpose (too big and too slow for post processing)

nibanks (approved):

nibanks:
It seems this still just times out the perf machines when run with logs enabled:
```
Running Test ThroughputDown_linux_x64_openssl_Default
[ perf record: Woken up 163 times to write data ]
[ perf record: Captured and wrote 44.494 MB /home/test/QUICLogs/Perf/client_21.perf.data (269320 samples) ]
Test Run Took 00:00:16.3451883
Run 1: 5644509 kbps
[ perf record: Woken up 108 times to write data ]
tar/gzip LTTng log files: /home/test/Tests/serverlogs/server.tgz
tar/gzip LTTng log files: /home/test/azpagent/_work/2/s/artifacts/PerfDataResults/linux/x64_Release_openssl/ThroughputDown_linux_x64_openssl_Default/client.tgz
dso__load_sym_internal: failed to find program header for symbol: _etext st_value: 0x23674
dso__load_sym_internal: failed to find program header for symbol: _etext st_value: 0x23674
Warning:
Processed 5110894 events and lost 10251 chunks!

Check IO/CPU overload!

##[error]The task has timed out.
```

ami-GS:
- Manual tests polluted artifact directory which overload perf's post processing -> add cleanup routine before running
- `perf -F max` generates some GBs artifact for each iteration. -> May I reduce freq? half Hz generates 500MB artifact
- running `perf script` and generating .svg takes similar long time -> May I stop generating .svg as we can visualize raw data (`perf script`ed) on WPA

nibanks (approved):

PR #3350, opened 2023-01-12
Refactor Partitioning

nibanks:
## Description

This is another step towards supporting application-driven MsQuic (no internal threads). This refactors the code base to be "per-partition" instead of explicitly "per-processor". By default, MsQuic will create per-processor partitions, but it supports the ability for apps to configure something different (though MsQuic still creates the threads to drive them).

Now with this PR, datapaths do not have the notion of processors (except for winkernel) anymore.

## Testing

Existing tests

## Documentation

N/A


src/core/binding.c
```diff
@@ -1636,7 +1636,7 @@ QuicBindingReceive(
     // connection it was delivered to.
     //
 
-    uint32_t Proc = CxPlatProcCurrentNumber();
+    uint16_t Proc = QuicLibraryGetCurrentProcessor();
```
nibanks, line 1639: Change to partition and query from the datagram itself.

src/core/library.c
```diff
@@ -99,13 +99,117 @@ MsQuicCalculatePartitionMask(
     MsQuicLib.PartitionMask = (HighBitSet << 1) - 1;
```
csujedihy, line 99: Just happened to stumble upon this function and these 2 lines seem like unnecessary.

The result is exactly the same without these 2 lines except for zero, which is an undefined value in this context.

src/core/registration.c
```diff
@@ -52,6 +52,14 @@ MsQuicRegistrationOpen(
         goto Error;
     }
 
+    //
+    // Lazy initialize partitions.
+    //
+    Status = MsQuicLibraryInitializePartitions(ExternalRegistration);
```
nibanks, line 58: Passing this flag should not be necessary because external registrations are always created first. Instead, remove this check, and put this in the `if (ExternalRegistration)` below.

csujedihy (approved):

csujedihy (approved):

PR #3358, opened 2023-01-13
Use dockers for linux builds

csujedihy:
## Description

Use onebranch linux build docker in public azure pipelines.

## Testing

CI

### DIFF
--- a/.azure/azure-pipelines.ci.yml
+++ b/.azure/azure-pipelines.ci.yml
@@ -295,6 +295,7 @@ stages:
       arch: x64
       tls: openssl
       config: Release
+      container: ghcr.io/microsoft/msquic/linux-build-xcomp:latest
 
 - stage: build_linux_debug
   displayName: Build Linux - Debug
@@ -307,6 +308,7 @@ stages:
       arch: x64
       tls: openssl
       config: Debug
+      container: ghcr.io/microsoft/msquic/linux-build-xcomp:latest
   - template: ./templates/build-config-user.yml
     parameters:
       image: ubuntu-20.04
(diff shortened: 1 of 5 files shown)


.azure/azure-pipelines.ci.yml
```diff
@@ -356,15 +358,7 @@ stages:
       tls: openssl
       extraBuildArgs: -DisableLogs
       extraName: 'android'
-
```
csujedihy, line 359: This one is redundant. Already covered by github actions.

scripts/spin.ps1
```diff
@@ -127,7 +127,11 @@ if ($CodeCoverage) {
 $SpinQuic = $null
 if ($IsWindows) {
     $SpinQuic = Join-Path $RootArtifactDir "spinquic.exe"
-} elseif ($IsLinux -or $IsMacOS) {
+} elseif ($IsLinux) {
+    $SpinQuic = Join-Path $RootArtifactDir "spinquic"
+    $SpinQuicRootDir = Split-Path $SpinQuic -Parent
+    $env:LD_LIBRARY_PATH = "${SpinQuicRootDir}:$env:LD_LIBRARY_PATH"
```
nibanks, line 133: Why is this needed only for docker builds?

csujedihy: This is still a mystery to me. When I first ran into `libmsquic.so.2: cannot open shared object file` in quicinterop and spinquic, the pipeline was run against the commit, [e079de8](https://github.com/microsoft/msquic/pull/3358/commits/e079de803036340357909911bb103978e09c9a7a), that doesn't even touch the build jobs that quicinterop and spinquic depend on. In other words, making irrelevant build tasks use container somewhat changed the behavior of quicinterop and spinquic, which I don't quite understand.

csujedihy: And btw, we don't set rpath for quicinterop and spinquic. So, they are not supposed to find shared libs from the origin folder. I don't know why they even worked before my PR.

nibanks: Maybe because everything is in the same place exactly for build and test machines before?

csujedihy: I honest don't know.

PR #3367, opened 2023-01-18
customize codespaces environment

ami-GS:
## Description

- openssh-server to ssh from GitHub CLI
- 'test' user to be more same as CI env

## Testing

N/A

## Documentation

N/A


nibanks (approved):

.devcontainer/devcontainer.json
```diff
@@ -1,9 +1,9 @@
 {
-    "image":  "ghcr.io/microsoft/msquic/linux-build-xcomp",
+    "dockerFile": "Dockerfile",
     "hostRequirements": {
         "cpus": 4,
         "memory": "2gb"
     },
-    "onCreateCommand": "pwsh -c './scripts/prepare-machine.ps1 -ForOneBranch -InitSubmodules'",
-    "runArgs": ["--privileged"]
+    "onCreateCommand": "pwsh -c './scripts/prepare-machine.ps1'",
```
csujedihy, line 7: Why remove "-ForOneBranch"?

.devcontainer/devcontainer.json
```diff
@@ -1,9 +1,9 @@
 {
-    "image":  "ghcr.io/microsoft/msquic/linux-build-xcomp",
+    "dockerFile": "Dockerfile",
```
csujedihy, line 2: Since now you are building the image on the fly, how fast is this?

.devcontainer/devcontainer.json
ami-GS: Where is the "-ForOneBranch" used?

.devcontainer/devcontainer.json
ami-GS: I don't feel such performance difference, because I don't touch the env for 5-10 min after launch...

.devcontainer/devcontainer.json
csujedihy: The reason why our devcontainer uses a docker image in the first place is we can just load up the docker image and skip the step of `./scripts/prepare-machine.ps1` and build the code right away.

.devcontainer/devcontainer.json
csujedihy: We should make this as fast as possible. You should push the image to ghcr (ideally, a github action to update the image, which I have not done for linux-build-xcomp either) and reference the image path. Then, we should skip the installation step in prepare-machine.ps1.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+RUN wget https://packages.microsoft.com/config/ubuntu/20.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb \
+    && dpkg -i packages-microsoft-prod.deb \
+    && rm packages-microsoft-prod.deb \
+    && apt-get update -y \
+    && apt-get install -y \
+    # prepare-machine.ps1 needs dotnet
+    dotnet-sdk-6.0 \
+    # prep for perf command
+    software-properties-common \
```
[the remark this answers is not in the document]
csujedihy: and this is for CPU perf measurement? We should not do any serious perf tests in devcontainer. It's a shared and maybe oversubscribed resource.

.devcontainer/devcontainer.json
ami-GS: I don't see any if-else branch in script though

csujedihy: if (!$ForOneBranch -and !$ForOneBranchPackage -and !$ForBuild -and !$ForTest -and !$InstallXdpDriver -and !$UninstallXdp) {

csujedihy: https://github.com/microsoft/msquic/blob/aaecb0fac5a3902dd244df8a05942ffc35a6729f/scripts/prepare-machine.ps1#L112

.devcontainer/devcontainer.json
ami-GS: Why need to be as fast as possible?
My understanding of this .devcontainer directory is for codespaces which we do something manually. Not test automation, right?

.devcontainer/devcontainer.json
ami-GS: oooooh it is ALL false case.....
But setting -ForOneBranch doesn't set -ForTest and -ForBuild.
If my understanding is correct, user will run prepare-machine.ps1 sooner or later, because this codespace settings is for manual operation. not test automation, right?

.devcontainer/devcontainer.json
csujedihy: This devcontainer feature is all about dev experience. Starting up codespaces faster is definitely preferable. That's why I used a docker image in the first place.

.devcontainer/devcontainer.json
csujedihy: For building the project, the user won't need to run prepare-machine.ps1 because everything is set up in the docker image already and its ready to build.

nibanks (dismissed):

nibanks (approved):

PR #3375, opened 2023-01-24
Add statement about OpenSSL 3.x

richlander:


docs/Platforms.md
```diff
@@ -18,6 +18,8 @@ Optionally, `msquic.dll` can be built with OpenSSL (see below for more details)
 
 On Linux, MsQuic relies on [OpenSSL](https://www.openssl.org/) for TLS 1.3 functionality.
 
+> **Important** MsQuic does not currently support OpenSSL 3.x. In particular, MsQuic is incompatible with Alpine 3.17+, Debian 12+, and Ubuntu 22.04+ since they exclusively offer OpenSSL 3.x. Recent versions of other distros may be similar.
```
nibanks, line 21: > MsQuic is incompatible with ...

This is not quite true. MsQuic supports statically linking all of openssl, which allows it to work on these releases. If you're using the inbox openssl, then yes, it won't work.

wfurt: One can of course also just build older OpenSSL or use binaries from older version or different distro. We did that at least in one case. So this should make it clear it is about distro provided binaries.

richlander: It is entirely possible I don't understand. Does the statically linked version of OpenSSL play nicely with the one we use for HTTP 1 and 2 TLS scenarios?

richlander: What about FIPS? I assume we should tell folks that we don't have a FIPS story for HTTP3. Fair?

nibanks: > Does the statically linked version of OpenSSL play nicely with the one we use for HTTP 1 and 2 TLS scenarios?

If MsQuic statically links OpenSSL, then it ends up using its own version/copy of OpenSSL that is different than the one uses .NET. Everything still works, but it's not the most efficient.

> What about FIPS? I assume we should tell folks that we don't have a FIPS story for HTTP3. Fair?

Statically linking breaks FIPS. The only way to get FIPS is to use the distro version.

richlander: Does .NET use the statically-linked OpenSSL? I honestly don't know. I'm not on the networking team.

It sounds like the second commit on this PR is wrong: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/777. I didn't understand that MsQuic used static linking. Correct?

wfurt: No. We depend on OS libraries to avoid servicing and responsibility for security fixes. (and get FIPS stamp if that is any value) 

I feel there may be terminology mismatch between what MsQuic _can_ do and supports vs what is shipped as available and signed binaries. (easily consumable by .NET users) 

Last dimension is testing. If you look through the Helix images, on several platforms we did various hacks to get MsQuic in. For example we have Ubuntu 22 with Quic & OpenSSL 1.1 even if the vendor does not provide one. That for example allows us to test coexistence with OpenSSL 3 & 1.1.

richlander: > get FIPS stamp if that is any value

Our customers have FIPS scenarios so definitely valuable.

samsp-msft: FIPS is a whole complex topic unto itself. IIRC, FIPS certification is not for OpenSSL as a whole, it's only done for the libssl portion that implements the crypto algorithms themselves. MsQuic relies on QuicTLS which is the QUIC capable fork of OpenSSL. QuicTLS doesn't change the libssl portion, so if a distro vendor or consumer has a FIPS certified version of libssl that can/will be loaded instead of the one included in QuicTLS.

richlander: It is easy to test, @samsp-msft . Can you try running an app on Ubuntu 22.04+, Alpine 3.17+, or Debian 12+, per our generic guidance?

samsp-msft: Sorry, I'm heads down on diagnostics work right now. @wfurt - are you able to run the tests?

docs/Platforms.md
```diff
@@ -18,6 +18,8 @@ Optionally, `msquic.dll` can be built with OpenSSL (see below for more details)
 
 On Linux, MsQuic relies on [OpenSSL](https://www.openssl.org/) for TLS 1.3 functionality.
```
samsp-msft, line 19:
```suggestion
On Linux, MsQuic relies on QuicTLS, a fork of [OpenSSL](https://www.openssl.org/) for TLS 1.3 functionality. QuicTLS is a small delta on top of OpenSSL that adds APIs to enable the QUIC handshake that establishes TLS in first packet as part of connection establishment.
```

docs/Platforms.md
richlander: I can do it if you point me to some samples.

samsp-msft: @richlander - the setup instructions and httpclient example code are all in this [doc](https://learn.microsoft.com/en-us/dotnet/core/extensions/httpclient-http3). The doc also links to a cloudflare server to test with.

As the issue is TLS support, if the client works, the server should too, but to verify do a `dotnet new web`, and then use [this](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/http3?view=aspnetcore-7.0#getting-started) modification to add an http3 endpoint on port 5001. As the client is forcing http/3, if that works for https://localhost:5001/, then you have proof both work.

wfurt: You can go to public HTTP/3 and avoid all the Kestrel setup...

https://github.com/dotnet/runtime/blob/d2a87fe8290e13bfc4d0901fc30595dfbd6bc7ac/src/libraries/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.Http3.cs#L1717-L1723

docs/Platforms.md
nibanks: We mention QuicTLS a couple lines below.

samsp-msft: I think the initial statement is partly what caused confusion as to whether the openSSL installed as part of the distro is consumed by MSQuic.

wfurt: Actually, I'm wondering if we could create additional alternative packages where OpenSSL 1.1 _is_ linked to MsQuic. So users could choose depending on their needs and base distro. For now, users requiring FIPS would just need to stay on older version but since the certification process is very long it may be required anyway.

richlander:
I ran an experiment with this code:

```csharp
using System.Net;

using var client = new HttpClient()
{
    DefaultRequestVersion =  HttpVersion.Version30,
    DefaultVersionPolicy = HttpVersionPolicy.RequestVersionExact
};

HttpResponseMessage resp = await client.GetAsync("https://www.litespeedtech.com/");
string body = await resp.Content.ReadAsStringAsync();

Console.WriteLine(
    $"status: {resp.StatusCode}, version: {resp.Version}, " +
    $"body: {body.Substring(0, Math.Min(100, body.Length))}");
```

With Debian 11:

```bash
# dotnet --version
7.0.100
# cat /etc/os-release | grep PRETTY
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
# apt list --installed libmsquic openssl
Listing... Done
libmsquic/bullseye,now 2.1.4 amd64 [installed]
openssl/stable,stable-security,now 1.1.1n-0+deb11u3 amd64 [installed,automatic]
# dotnet run
status: OK, version: 3.0, body:
<!DOCTYPE html>

<html prefix="og: http://ogp.me/ns#" xmlns="http://www.w3.org/1999/xhtml" xml:lang
```

With Ubuntu 22.04:

```bash
# dotnet --version
7.0.102
# cat /etc/os-release | grep PRETTY
PRETTY_NAME="Ubuntu 22.04.1 LTS"
# apt list --installed libmsquic openssl
Listing... Done
libmsquic/jammy,now 2.1.4 amd64 [installed]
openssl/jammy-updates,jammy-security,now 3.0.2-0ubuntu1.7 amd64 [installed,automatic]
# dotnet run
Unhandled exception. System.Net.Http.HttpRequestException: Requesting HTTP version 3.0 with version policy RequestVersionExact while unable to establish HTTP/3 connection.
   at System.Net.Http.HttpConnectionPool.ThrowGetVersionException(HttpRequestMessage request, Int32 desiredVersion, Exception inner)
   at System.Net.Http.HttpConnectionPool.SendWithVersionDetectionAndRetryAsync(HttpRequestMessage request, Boolean async, Boolean doRequestAuth, CancellationToken cancellationToken)
   at System.Net.Http.RedirectHandler.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.HttpClient.<SendAsync>g__Core|83_0(HttpRequestMessage request, HttpCompletionOption completionOption, CancellationTokenSource cts, Boolean disposeCts, CancellationTokenSource pendingRequestsCts, CancellationToken originalCancellationToken)
   at Program.<Main>$(String[] args) in /app/Program.cs:line 9
   at Program.<Main>(String[] args)
```

wfurt:
right. This is because the 2.1 release branch never picked up https://github.com/microsoft/msquic/pull/2995. With it, it would fail to install msquic because of unmet dependencies. We will make sure it is next round of published packages. 

Tracked by https://github.com/microsoft/msquic/issues/2975

richlander:
If I understand correctly, that won't do anything. Packages that depend on OpenSSL 1.1 will not install on Ubuntu 22.04 and I'm sure the same is true of the other distros. OpenSSL 1.1 is not available in modern feeds.

samsp-msft:
The goal is to fix that too, it may just take a bit longer, and will possibly require servicing 7 to adapt to other changes in msquic.

currently you get (mostly) unexplained errors at compile time, with the package dependencies being correctly specified, you'll know that there is a problem at install time.

richlander:
Makes sense. Are you intending to document that our HTTP3 support doesn't work on modern Linux so that user don't even need to get as far as installing the package?

wfurt:
I check Fedora37. While it uses OpenSSL3 it also provides compat `openssl1.1`. So everything should work just fine once we fix the package dependencies. (next 2.1 maintenance) 
There is probably solution doe the Ubuntu as well, it just won't work automatically with their package manager (for now)

richlander:
Canonical has no plans to offer OpenSSL 1.x in 22.04+ feeds.

wfurt:
I'm wondering if we can convince Canonical to build the MsQuic as part of OS packages. We can make it work even if they don't provide the compat OpenSSL  package.

richlander:
Maybe. I think we'd need to demonstrate a bit more forward-looking leadership to be taken seriously as a trusted upstream maintainer. I don't think we're doing that here.

Recall that the .NET Team added support for OpenSSL 3.0 back in .NET 6 Preview 5 in June 2021: https://devblogs.microsoft.com/dotnet/announcing-net-6-preview-5/#libraries-support-for-openssl-3. I think that was a great timeframe for doing that project. When .NET was added into the Ubuntu 22.04 feed, it meant that we didn't need to discuss OpenSSL 3.0 support.

Clearly, we're not going to talk to any distros until this problem is behind us.

nibanks:
Now that we have (at least some) support for OpenSSL 3, we should update the PR to reflect this. Otherwise, let's just close this.

nibanks:
Closing this because of no updates, and it's not as necessary anymore. Feel free to reopen with updates.

PR #3380, opened 2023-01-27
Work Around Missing NUMA Node Info

nibanks:
## Description

It seems some setups don't return NUMA node info as expected. Work around this.

## Testing

Existing CI and manual testing on repro

## Documentation

N/A

### DIFF
--- a/src/platform/platform_winuser.c
+++ b/src/platform/platform_winuser.c
@@ -189,14 +189,21 @@ CxPlatProcessorInfoInit(
         "[ dll] Processors:%u, Groups:%u, NUMA Nodes:%u",
         ActiveProcessorCount, ProcessorGroupCount, NumaNodeCount);
 
-    Offset = 0;
-    while (Offset < BufferLength) {
-        PSYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX Info =
-            (PSYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX)(Buffer + Offset);
-        if (Info->Relationship == RelationNumaNode) {
-            CxPlatNumaMasks[Info->NumaNode.NodeNumber] = (uint64_t)Info->NumaNode.GroupMask.Mask;
+    if (NumaNodeCount == 1) {
+        //
+        // All processors belong to the one NUMA node.
+        //
+        CxPlatNumaMasks[0] = (1 << ActiveProcessorCount) - 1;
+    } else {
+        Offset = 0;
+        while (Offset < BufferLength) {
+            PSYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX Info =
+                (PSYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX)(Buffer + Offset);
+            if (Info->Relationship == RelationNumaNode) {
+                CxPlatNumaMasks[Info->NumaNode.NodeNumber] = (uint64_t)Info->NumaNode.GroupMask.Mask;
+            }
+            Offset += Info->Size;
         }
-        Offset += Info->Size;
     }
 
     for (uint32_t Index = 0; Index < ActiveProcessorCount; ++Index) {
--- a/src/platform/platform_winuser.c
+++ b/src/platform/platform_winuser.c
@@ -23,7 +23,6 @@ uint64_t CxPlatPerfFreq;
 uint64_t CxPlatTotalMemory;
 CX_PLATFORM CxPlatform = { NULL };
 CXPLAT_PROCESSOR_INFO* CxPlatProcessorInfo;
-uint64_t* CxPlatNumaMasks;
 uint32_t* CxPlatProcessorGroupOffsets;
 #ifdef TIMERR_NOERROR
 TIMECAPS CxPlatTimerCapabilities;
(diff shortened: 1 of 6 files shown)


Wizmann:
BTW, return `QUIC_STATUS_OUT_OF_MEMORY` on NUMA error is misleading...

```
    if (!CxPlatProcessorInfoInit()) {
        QuicTraceEvent(
            LibraryError,
            "[ lib] ERROR, %s.",
            "CxPlatProcessorInfoInit failed");
        Status = QUIC_STATUS_OUT_OF_MEMORY; // <-----
        goto Error;
```

src/platform/platform_winuser.c
```diff
@@ -118,11 +118,7 @@ CxPlatProcessorInfoInit(
...
             (PSYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX)(Buffer + Offset);
-        if (Info->Relationship == RelationNumaNode) {
-            if (Info->NumaNode.NodeNumber + 1 > NumaNodeCount) {
-                NumaNodeCount = Info->NumaNode.NodeNumber + 1;
-            }
-        } else if (Info->Relationship == RelationGroup) {
+        if (Info->Relationship == RelationGroup) {
             if (ProcessorGroupCount == 0) {
                 CXPLAT_DBG_ASSERT(Info->Group.ActiveGroupCount != 0);
                 ProcessorGroupCount = Info->Group.ActiveGroupCount;
```
Wizmann, line 124: we can safely break the loop here

src/platform/platform_winuser.c
```diff
@@ -118,11 +118,7 @@ CxPlatProcessorInfoInit(
     while (Offset < BufferLength) {
         PSYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX Info =
             (PSYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX)(Buffer + Offset);
-        if (Info->Relationship == RelationNumaNode) {
-            if (Info->NumaNode.NodeNumber + 1 > NumaNodeCount) {
-                NumaNodeCount = Info->NumaNode.NodeNumber + 1;
-            }
-        } else if (Info->Relationship == RelationGroup) {
+        if (Info->Relationship == RelationGroup) {
```
Wizmann, line 121: We can safely break the loop after `if (ProcessorGroupCount == 0)` at L125

Wizmann (dismissed):

src/platform/platform_winuser.c
```diff
@@ -69,30 +68,30 @@ CxPlatSystemUnload(
...
     uint8_t* Buffer = NULL;
     uint32_t Offset;
 
-    uint32_t ActiveProcessorCount = CxPlatProcActiveCount();
+    const uint32_t ActiveProcessorCount = CxPlatProcActiveCount();
     uint32_t ProcessorGroupCount = 0;
     uint32_t ProcessorsPerGroup = 0;
-    uint32_t NumaNodeCount = 0;
 
+    CXPLAT_DBG_ASSERT(ActiveProcessorCount > 0);
```
Wizmann, line 93: Should return error when `ActiveProcessorCount == 0`.

See: https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getactiveprocessorcount

src/platform/platform_winuser.c
```diff
@@ -104,30 +103,29 @@ CxPlatProcessorInfoInit(
...
-            if (ProcessorGroupCount == 0) {
-                CXPLAT_DBG_ASSERT(Info->Group.ActiveGroupCount != 0);
-                ProcessorGroupCount = Info->Group.ActiveGroupCount;
-                ProcessorsPerGroup = Info->Group.GroupInfo[0].ActiveProcessorCount;
-            }
+        if (Info->Relationship == RelationGroup) {
+            CXPLAT_DBG_ASSERT(ProcessorGroupCount == 0);
+            CXPLAT_DBG_ASSERT(Info->Group.ActiveGroupCount != 0);
+            ProcessorGroupCount = Info->Group.ActiveGroupCount;
+            ProcessorsPerGroup = Info->Group.GroupInfo[0].ActiveProcessorCount;
```
Wizmann, line 127: Is it safe to make the assumption that `ProcessorsPerGroup` is identical in each group?

nibanks: I _thought_ so, but https://learn.microsoft.com/en-us/windows/win32/procthread/processor-groups makes no mention of that assumption. The closest I can find is:
> The operating system minimizes the number of groups in a system. For example, a system with 128 logical processors would have two processor groups with 64 processors in each group, not four groups with 32 logical processors in each group.

nibanks: Latest commits should handle any scenario.

Wizmann: Looks good.

mtfriesen:
Can you summarize what the unexpected NUMA info is?

nibanks:
> Can you summarize what the unexpected NUMA info is?

Sometime NUMA relationship is not returned from the Windows API, so we have to rely exclusively on Group relationships.

mtfriesen:
There are new APIs for dealing with NUMA nodes starting with Windows Server 2022: https://learn.microsoft.com/en-us/windows/win32/procthread/numa-support#behavior-starting-with-windows-10-build-20348

nibanks:
> There are new APIs for dealing with NUMA nodes starting with Windows Server 2022: https://learn.microsoft.com/en-us/windows/win32/procthread/numa-support#behavior-starting-with-windows-10-build-20348
> 
> That said, it seems like MsQuic is in fact NUMA-unaware, other than logging.

We're now Group-aware instead of NUMA. There's a lot of overlap between the two, but they're not quite the same always.

PR #3391, opened 2023-02-02
Add Support for Custom TLS Alert in Certificate Validation Failure

nibanks:
## Description

Allows the app to give more context on why certificate validation failed.

## Testing

Existing tests

## Documentation

Updated


src/cs/lib/msquic_generated.cs
```diff
@@ -44,6 +44,25 @@ internal enum QUIC_LOAD_BALANCING_MODE
         COUNT,
     }
 
+    internal enum QUIC_TLS_ALERT_CODES
+    {
+        SUCCESS = 0xFFFF,
```
anrossi, line 49: This is larger than the MAX however?

nibanks: Yeah, on purpose because while it's a valid enum, it's not a valid wire-format TLS alert.

src/tools/spin/spinquic.cpp
```diff
@@ -885,7 +885,7 @@ void Spin(Gbs& Gb, LockableVector<HQUIC>& Connections, std::vector<HQUIC>* Liste
         case SpinQuicAPICallCompleteCertificateValidation: {
             auto Connection = Connections.TryGetRandom();
             BAIL_ON_NULL_CONNECTION(Connection);
-            MsQuic.ConnectionCertificateValidationComplete(Connection, GetRandom(2) == 0);
+            MsQuic.ConnectionCertificateValidationComplete(Connection, GetRandom(2) == 0, QUIC_TLS_ALERT_CODE_BAD_CERTIFICATE);
```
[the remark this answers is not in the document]
nibanks: I don't expect that to generate any significant difference, so I decided against it for now.

src/core/api.h
```diff
@@ -275,5 +275,6 @@ QUIC_STATUS
 QUIC_API
 MsQuicConnectionCertificateValidationComplete(
     _In_ _Pre_defensive_ HQUIC Handle,
-    _In_ BOOLEAN Result
+    _In_ BOOLEAN Result,
+    _In_ QUIC_TLS_ALERT_CODES TlsAlert
```
anrossi, line 279: Could annotate this such that it's only read when Result is false.

nibanks: I put a comment in msquic.h already. I think that's the best place.

anrossi (approved):

PR #3394, opened 2023-02-03
build libssl with  no-seed

wfurt:
follow-up on https://github.com/dotnet/runtime/pull/81481#discussion_r1095979398
attempt to use system crypto fails on Alpine Linux
```
ldd /usr/lib/libmsquic.so.2
        /lib/ld-musl-x86_64.so.1 (0x7fbc0efdf000)
        libcrypto.so.1.1 => /lib/libcrypto.so.1.1 (0x7fbc0ec00000)
        libc.musl-x86_64.so.1 => /lib/ld-musl-x86_64.so.1 (0x7fbc0efdf000)
Error relocating /usr/lib/libmsquic.so.2: EVP_seed_cbc: symbol not found
```

It is because it is build with `no-seed`
https://git.alpinelinux.org/aports/tree/main/openssl/APKBUILD?h=3.15-stable#n110

I could make this Alpine specific but I feel like we should not pull in parts that we don't need. 
So I left it as general change to decrease expectation from OS.


wfurt:
cc: @manickap

nibanks (approved):

PR #3397, opened 2023-02-04
Reduce the Tests in Wan Perf

nibanks:
## Description

Reduce overhead of these tests since we don't actively look at results

## Testing

## Documentation


csujedihy (approved):
I wonder if we should make it totally optional behind a manual trigger.

nibanks:
> I wonder if we should make it totally optional behind a manual trigger.

I'd rather actually make use of it and keep the automation.

PR #3407, opened 2023-02-08
Fix unknown callgraph issue

ami-GS:
## Description

Confirmed most of unknown information has gone. then stack appearance is also changed accordingly

## Testing

N/A

## Documentation

N/A

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -606,7 +606,7 @@ else() #!WIN32
         set(MARCH -march=native)
     endif()
 
-    set(CMAKE_C_FLAGS_DEBUG "-Og")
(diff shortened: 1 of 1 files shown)


ami-GS:
hmm? setting flat to Release still doesn't work.
Which build does perf pipeline use?

nibanks:
> hmm? setting flat to Release still doesn't work.
> Which build does perf pipeline use?

What do you mean? It should use release builds.

ami-GS:
Works
https://dev.azure.com/ms/msquic/_build/results?buildId=426256&view=results

nibanks (approved):

PR #3415, opened 2023-02-10
Improve libnuma Dependency Discovery

nibanks:
## Description

Updates the cmake code to look for both the `libnuma` package and the numa.h header that comes from the `-dev` package.

## Testing

Tested locally (with it installed). Rest will test with automation.

## Documentation

N/A


nibanks:
cc @wfurt, @ManickaP

CMakeLists.txt
```diff
@@ -606,9 +612,9 @@ else() #!WIN32
         set(MARCH -march=native)
     endif()
 
-    set(CMAKE_C_FLAGS_DEBUG "-Og -fno-omit-frame-pointer")
+    set(CMAKE_C_FLAGS_DEBUG "-Og")
     set(CMAKE_C_FLAGS_MINSIZEREL "-Os -DNDEBUG")
-    set(CMAKE_C_FLAGS_RELWITHDEBINFO "-O3 -fno-omit-frame-pointer ${MARCH} -DNDEBUG")
```
ami-GS, line 611: Why removing?

nibanks: Oops. This accidentally got it. Not sure how...

ami-GS (approved):

CMakeLists.txt
```diff
@@ -427,7 +427,13 @@ if (NOT MSVC AND NOT APPLE AND NOT ANDROID)
     find_library(NUMA NAMES NUMA libnuma.so.1)
     if (NUMA)
         message(STATUS "Found libnuma: ${NUMA}")
-        list(APPEND QUIC_COMMON_DEFINES CXPLAT_NUMA_AWARE)
+        find_file(NUMA-DEV NAMES "numa.h")
```
wfurt, line 430: I think the standard way is to try to compile code fragment. But this will probably work as well on moth systems.

src/platform/platform_posix.c
```diff
@@ -32,12 +32,8 @@
 #include "platform_posix.c.clog.h"
 #endif
 
-#ifdef __aarch64__
-#undef CXPLAT_NUMA_AWARE
-#endif
-
 #ifdef CXPLAT_NUMA_AWARE
-#include <numa.h>
+#include <numa.h>               // If missing: `apt-get install -y libnuma-dev`
```
wfurt, line 36: I'm not sure the comment is appropriate. It represents subset of Linux distriburtiuons and not POSIX.

Issue #2396, opened 2022-02-14
AppData/WithSend0RttArgs1.Send0Rtt/* failed


nibanks:
### Describe the bug

I've been seeing a lot of failures with output like the following:
```
D:\a\1\msquic\src\test\lib\DataTest.cpp(149): error: 0-RTT wasn't used for stream data.
```
See: https://dev.azure.com/ms/msquic/_build/results?buildId=272959&view=logs&j=b7f0e904-2cea-585b-8e20-67062f2bef10&t=81f432f0-d5f0-5258-5253-0dcccc920e38&l=1547

It seems like a different test case every time, but has definitely picked up recently. Something has likely regressed things, perhaps some of the work in WAN perf.

### Affected OS

- [X] Windows Server 2022
- [X] Windows 11
- [X] Windows Insider Preview (specify affected build below)
- [X] Ubuntu
- [X] Debian
- [ ] Other (specify below)

### MsQuic version

main

### Steps taken to reproduce bug

Run Azure Pipelines BVTs for OpenSSL.

### Expected behavior

All tests should pass

### Actual outcome

The send 0-RTT tests sometime fail because 0-RTT wasn't used for all streams.

nibanks:
I took a look at two different logs from failures in the `release/2.0` weekly run for 3 different variables of the `Send0Rtt` tests ([here](https://dev.azure.com/ms/msquic/_build/results?buildId=303503&view=results). It looks like the packets end up getting retransmitted because they were somehow lost the first time. When they are retransmitted, they end up using the 1-RTT keys instead. The problem is that the logs are incomplete, meaning some were dropped. I wonder if the CPU usage sometimes goes too high on the VMs, which results in log drop; but could also somehow be the cause of the packet drop?

```
6379 [0]0E40.157C::‎2022‎-‎04‎-‎18 00:40:56.537 [Microsoft-Quic][strm][0x2C6610C68D0] Recovering offset 0 up to 662 
6386 [0]0E40.157C::‎2022‎-‎04‎-‎18 00:40:56.537 [Microsoft-Quic][strm][0x2C6610C68D0] Recovering offset 662 up to 1000 
6488 [2]0E40.157C::‎2022‎-‎04‎-‎18 00:40:56.538 [Microsoft-Quic][strm][0x2C6610C68D0] Built stream frame, offset=0 len=709 fin=0 
[...]

PR #3435, opened 2023-02-15
Fix Client sock to bind same port as that of server

ami-GS:
## Description

Fix of https://github.com/microsoft/msquic/issues/2396
SO_REUSEPORT is for server setting.
RemoteAddress == NULL is for server sock. The **or** condition was wrong

## Testing

Running locally.... and trying to run on CI

## Documentation

N/A

### DIFF
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -1234,7 +1234,7 @@ CxPlatSocketContextInitialize(
     // Only set SO_REUSEPORT on a server socket, otherwise the client could be
     // assigned a server port (unless it's forcing sharing).
     //
-    if (ForceShare || RemoteAddress == NULL) {
+    if (ForceShare && RemoteAddress == NULL) {
         //
         // The port is shared across processors.
         //
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -946,9 +946,7 @@ QUIC_STATUS
 CxPlatSocketContextInitialize(
     _Inout_ CXPLAT_SOCKET_CONTEXT* SocketContext,
     _In_ const QUIC_ADDR* LocalAddress,
-    _In_ const QUIC_ADDR* RemoteAddress,
-    _In_ BOOLEAN ForceShare
-    )
+    _In_ const QUIC_ADDR* RemoteAddress)
 {
     QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
     int Result = 0;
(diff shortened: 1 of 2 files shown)


ami-GS:
With this fix, any client cannot use SO_REUSEPORT at all.
The `ForceShare` is set in QuicConnStart based on Connection->State.ShareBinding, but RemoteServer param should be checked whether it is loopback?

src/platform/datapath_epoll.c
```diff
@@ -946,9 +946,7 @@ QUIC_STATUS
 CxPlatSocketContextInitialize(
     _Inout_ CXPLAT_SOCKET_CONTEXT* SocketContext,
     _In_ const QUIC_ADDR* LocalAddress,
-    _In_ const QUIC_ADDR* RemoteAddress,
-    _In_ BOOLEAN ForceShare
-    )
+    _In_ const QUIC_ADDR* RemoteAddress)
```
nibanks, line 949:
```suggestion
    _In_ const QUIC_ADDR* RemoteAddress
    )
```

nibanks (dismissed):

ami-GS:
This affects `DataPathTest/DataPathTest.UdpShareClientSocket/*` as there are two client sockets, which the second one use first one's localaddress.
Should we delete this test? because we don't expect this use case as you mentioned
https://github.com/microsoft/msquic/blob/fdd15b3d6f91e912c8212ad709edc7d5dcb03627/src/platform/unittest/DataPathTest.cpp#L917-L923

nibanks:
> This affects `DataPathTest/DataPathTest.UdpShareClientSocket/*` as there are two client sockets, which the second one use first one's localaddress. Should we delete this test? because we don't expect this use case as you mentioned
> 
> https://github.com/microsoft/msquic/blob/fdd15b3d6f91e912c8212ad709edc7d5dcb03627/src/platform/unittest/DataPathTest.cpp#L917-L923

If we don't need the scenario for QUIC, then yes, go ahead and delete it.

nibanks:
@ami-GS after looking at your most recent PR I realized why that code was there (I had forgot). We (MsQuic) support (only on Linux today) the ability to reuse the **local** address/port to connect to different remote addresses/ports for client connections. So, the platform layer was actually correct.

The real problem here is that the `QuicLibraryGetBinding` logic for `QUIC_SHARED_EPHEMERAL_WORKAROUND` isn't correctly preventing a new `PortUnspecified` client socket from matching an existing server socket. This PR should simply be abandoned (sorry!) and the correct fix made at the QUIC layer.

On a side note, I'm surprised to see we didn't have QUIC level tests that exercised the shared local ports for client sockets. I thought we did... Are you sure no QUIC layer tests were failing with this PR (I didn't check)?

nibanks:
See https://github.com/microsoft/msquic/pull/3439.

PR #3436, opened 2023-02-16
fix MAC_CTX creation with OpenSSL 3

wfurt:
Contributes to https://github.com/microsoft/msquic/issues/2039

it seems like the context was not created properly so handling of tickets was failing.
Interestingly, OpenSSL would just surface internal error so it was not obvious what was happening. 
With this, all the tests can now pass for me on Ubuntu 22.04.


nibanks (approved):
Thanks for doing this. If everything passes, ship it!

PR #3446, opened 2023-02-16
QTIP (QUIC over TCP)

csujedihy:
## Description
QTIP (QUIC over TCP): A variant of QUIC with TCP headers instead of UDP headers. This is different from a TCP tunnel in that it does not use TCP’s retransmission and congestion control logic: it simply uses TCP headers (and a few pure TCP SYN and RST packets with no QUIC payload) to masquerade as TCP traffic.

Client Side:

1. When the initial packet is passed down to raw datapath, we cache the send and send a SYN instead.
2. When SYN+ACK is received, we send an ACK and then resume the initial packet and construct a RST to be sent later (because at this moment we know what sequence and ACK numbers to use).
3. When the first data packet is received, we will know what sequence number and ACK number to use. The ACK number after handshake is always acking the peer's SYN+ACK. The sequence number is always ISN+2, where ISN+1 is reserved for RST. By doing so, all data packets are retransmissions but still in window.
4. When the QUIC connection is shut down, the cached RST is sent.

Server side:
1. Statelessly respond to all SYN/FIN received.
2. When the first data packet is received, it's identical to 3.

## Testing

Only manually tested over local VMs and azure VMs.


src/core/library.c
```diff
@@ -1322,6 +1337,26 @@ QuicLibraryGetGlobalParam(
         Status = QUIC_STATUS_SUCCESS;
         break;
 
+#ifdef QUIC_USE_RAW_DATAPATH
```
anrossi, line 1340: Should this also be wrapped in `#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES`?

anrossi:
How well does PLPMTUD play with this? Does QTIP disable the MTU probes?

src/core/library.c
csujedihy: No, I don't see any precedents. The code in this file for `QUIC_PARAM_GLOBAL_VERSION_SETTINGS` isn't wrapped in `#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES`.

csujedihy:
> How well does PLPMTUD play with this? Does QTIP disable the MTU probes?

I'd assume it's unaffected. We are effectively just increasing the UDP header to 20B.

src/inc/quic_datapath.h
```diff
@@ -74,10 +95,11 @@ typedef enum CXPLAT_ECN_TYPE {
 inline
 uint16_t
 MaxUdpPayloadSizeFromMTU(
```
nibanks, line 97: Is there a way we can achieve this without this change? i.e. Fake UDP MTU, but reduce the max MTU instead to account for the difference?

nibanks: Update `CxPlatSocketGetLocalMtu` when using TCP?

nibanks: Or, maybe just don't do anything special, and let PMTUD handle it...

csujedihy: > Or, maybe just don't do anything special, and let PMTUD handle it...
Wouldn't this overrun the packet allocated by CxPlatSendDataAlloc?

> Update CxPlatSocketGetLocalMtu when using TCP?
Seems like a good idea.

src/inc/msquic.h
```diff
@@ -812,6 +812,9 @@ void
 #define QUIC_PARAM_GLOBAL_EXECUTION_CONFIG              0x01000009  // QUIC_EXECUTION_CONFIG
 #endif
 #define QUIC_PARAM_GLOBAL_TLS_PROVIDER                  0x0100000A  // QUIC_TLS_PROVIDER
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+#define QUIC_PARAM_GLOBAL_QUIC_OVER_TCP                 0x0100000B  // uint8_t
+#endif
```
nibanks, line 817: Instead, add a flag to `QUIC_EXECUTION_CONFIG_FLAGS`:
```c
typedef enum QUIC_EXECUTION_CONFIG_FLAGS {
    QUIC_EXECUTION_CONFIG_FLAG_NONE             = 0x0000,
    QUIC_EXECUTION_CONFIG_FLAG_QTIP             = 0x0001,
#endif
} QUIC_EXECUTION_CONFIG_FLAGS;
```

src/inc/quic_datapath.h
```diff
@@ -481,6 +498,7 @@ typedef struct CXPLAT_UDP_CONFIG {
     uint8_t CibirIdOffsetSrc;           // CIBIR ID offset in source CID
     uint8_t CibirIdOffsetDst;           // CIBIR ID offset in destination CID
     uint8_t CibirId[6];                 // CIBIR ID data
+    BOOLEAN UseTcp;                     // Quic Over TCP
```
nibanks, line 501: Define a new `CXPLAT_SOCKET_*` flag instead of a new field?

src/platform/datapath_raw_socket.c
```diff
@@ -616,26 +745,29 @@ CxPlatDpRawParseIPv4(
         return;
     }
 
-    if (IP->Protocol == IPPROTO_UDP) {
-        uint16_t IPTotalLength;
-        IPTotalLength = CxPlatByteSwapUint16(IP->TotalLength);
+    uint16_t IPTotalLength;
+    IPTotalLength = CxPlatByteSwapUint16(IP->TotalLength);
```
nibanks, line 749:
```suggestion
    uint16_t IPTotalLength = CxPlatByteSwapUint16(IP->TotalLength);
```
And I _think_ it can be `const` too.

src/platform/datapath_raw_socket.c
```diff
@@ -668,40 +800,43 @@ CxPlatDpRawParseIPv6(
...
-            QuicTraceEvent(
-                DatapathErrorStatus,
-                "[data][%p] ERROR, %u, %s.",
-                Datapath,
-                IPPayloadLength,
-                "incorrect IP payload length");
-            return;
-        }
+    uint16_t IPPayloadLength;
+    IPPayloadLength = CxPlatByteSwapUint16(IP->PayloadLength);
```
nibanks, line 804:
```suggestion
    uint16_t IPPayloadLength = CxPlatByteSwapUint16(IP->PayloadLength);
```

src/inc/msquic.h
nibanks: I think by doing this, we can eliminate all (most?) of the "TCP" code in the core logic. I'd really like to have zero TCP stuff in there, and this is just a pass through to the datapath.

.azure/azure-pipelines.perf.yml
```diff
@@ -351,6 +351,34 @@ stages:
...
+        platform: windows
+        localTls: schannel
+        remoteTls: schannel
+        iterations: ${{ parameters.iterations }}
+        arch: ${{ parameters.arch }}
+        protocol: ${{ parameters.protocol }}
+        logProfile: ${{ parameters.logging }}
+        timeout: ${{ parameters.timeout }}
+        extraArtifactDir: '_QTIP'
+        extraTestArgs: -ExtraArtifactDir _QTIP -XDP -QTIP
```
nibanks, line 372: This won't work because you need to download the `_Xdp` artifacts, but if you change that, then it would upload the results over the normal XDP results.

nibanks: I _think_ for now, let's add a manual flag at the top of the file to enable QTIP and have that flag off by default. When enabled it can modify the XDP run to enable QTIP so we can get a look.

src/core/connection.c
```diff
@@ -5570,6 +5573,8 @@ QuicConnRecvDatagrams(
             DatagramPath->Route.Queue != Datagram->Route->Queue) {
             DatagramPath->Route.Queue = Datagram->Route->Queue;
         }
+
+        CxPlatUpdateRouteTcpState(&DatagramPath->Route, Datagram->Route);
```
nibanks, line 5577: Can we somehow refactor this, plus the code above (maybe), to be generic, non-TCP specific code?

src/inc/msquic.h
csujedihy: Can I just use MsQuicLib.ExecutionConfig.Flags everywhere in the code?

csujedihy: Seems like all datapath code doesn't include library.h? So I still need to do some plumbing?

.azure/azure-pipelines.perf.yml
```diff
@@ -134,6 +134,10 @@ parameters:
   type: number
   displayName: Timeout (minutes)
   default: 45
+- name: QTIP
+  type: string
```
nibanks, line 138: I think using a `boolean` would be better,

src/core/connection.c
```diff
@@ -1915,6 +1915,11 @@ QuicConnStart(
 #ifdef QUIC_OWNING_PROCESS
     UdpConfig.OwningProcess = Configuration->OwningProcess;
 #endif
+#ifdef QUIC_USE_RAW_DATAPATH
+    if (MsQuicLib.ExecutionConfig->Flags & QUIC_EXECUTION_CONFIG_FLAG_QTIP) {
+        UdpConfig.Flags |= CXPLAT_SOCKET_QTIP;
+    }
+#endif
```
nibanks, line 1922: If we're always using QTIP when this is set, the DAL already knows this flag, so how about we make it automatic at that layer, instead of having QUIC logic for it?

src/core/path.c
```diff
@@ -33,6 +33,11 @@ QuicPathInitialize(
     Path->RttVariance = Path->SmoothedRtt / 2;
     Path->EcnValidationState =
         Connection->Settings.EcnEnabled ? ECN_VALIDATION_TESTING : ECN_VALIDATION_FAILED;
+#ifdef QUIC_USE_RAW_DATAPATH
+    if (MsQuicLib.ExecutionConfig->Flags & QUIC_EXECUTION_CONFIG_FLAG_QTIP) {
+        CxPlatRandom(sizeof(Path->Route.TcpState.SequenceNumber), &Path->Route.TcpState.SequenceNumber);
+    }
+#endif
```
nibanks, line 40: Can we offload this to some datapath layer instead?

src/core/quicdef.h
```diff
@@ -516,6 +516,11 @@ CXPLAT_STATIC_ASSERT(
 //
 #define QUIC_DEFAULT_HYSTART_ENABLED                FALSE
 
+//
+// The default settings for enabling Quic over TCP.
+//
+#define QUIC_DEFAULT_QUIC_OVER_TCP_ENABLED          FALSE
+
```
nibanks, line 523: Remove?

src/cs/lib/msquic_generated.cs
```diff
@@ -2959,6 +2959,9 @@ internal static unsafe partial class MsQuic
         [NativeTypeName("#define QUIC_PARAM_GLOBAL_TLS_PROVIDER 0x0100000A")]
         internal const uint QUIC_PARAM_GLOBAL_TLS_PROVIDER = 0x0100000A;
 
+        [NativeTypeName("#define QUIC_PARAM_GLOBAL_QUIC_OVER_TCP 0x0100000B")]
+        internal const uint QUIC_PARAM_GLOBAL_QUIC_OVER_TCP = 0x0100000B;
+
```
nibanks, line 2964: Remove?

src/core/connection.c
csujedihy: MsQuicLib isn't accessible to the DAL.

nibanks: ExecutionConfig is passed to datapath initialize.

csujedihy: Done

src/core/path.c
csujedihy: Wrap it in a function called something like CxPlatInitiateDatapathRouteInfo? This is setting the initial sequence number. It has to be set when we initialize the route object.

src/inc/quic_datapath.h
```diff
@@ -693,6 +709,13 @@ CxPlatResolveRoute(
     _In_ CXPLAT_ROUTE_RESOLUTION_CALLBACK_HANDLER Callback
     );
 
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+CxPlatUpdateDataPathRoute(
```
nibanks, line 714: Maybe, following the pattern above, name it `CxPlatUpdateRoute`?

src/platform/datapath_raw_xdp.c
```diff
@@ -1308,10 +1308,12 @@ CxPlatDpRawPlumbRulesOnSocket(
     )
 {
     XDP_DATAPATH* Xdp = (XDP_DATAPATH*)Socket->Datapath;
-
+    printf("plumb rules\n");
```
csujedihy, line 1311: del

src/platform/datapath_raw.h
```diff
@@ -67,6 +69,9 @@ typedef struct CXPLAT_DATAPATH {
     BOOLEAN Uninitialized : 1;
     BOOLEAN Freed : 1;
 #endif
+#ifdef QUIC_USE_RAW_DATAPATH
+    BOOLEAN UseTcp;
+#endif
```
nibanks, line 74: We're in datapath_**raw**.h, shouldn't `QUIC_USE_RAW_DATAPATH` always be defined?
```suggestion
    BOOLEAN UseTcp;
```

src/platform/datapath_raw_socket.c
```diff
@@ -204,47 +206,188 @@ CxPlatTryAddSocket(
...
+                    Socket,
+                    WsaError,
+                    "bind (temp udp socket)");
+                CxPlatRwLockReleaseExclusive(&Pool->Lock);
+                Status = HRESULT_FROM_WIN32(WsaError);
+                goto Error;
+            }
+
+            Result =
+                connect(
```
nibanks, line 293: Should we really be calling `connect` on the TCP socket? Doesn't this start the stack's TCP handshake?

src/platform/datapath_raw_socket.c
```diff
@@ -204,47 +206,188 @@ CxPlatTryAddSocket(
             MappedAddress.Ipv6.sin6_family = AF_INET6;
         }
 #endif
+        if (Socket->UseTcp) {
+            //
+            // Create a temporary UDP socket bound to a wildcard port
+            // and connect this socket to the remote address.
+            // By doing this, the OS will select a local address for us.
```
nibanks, line 213: Whatever port the stack assigned for UDP isn't guaranteed to be available for TCP though, right?

src/platform/datapath_raw_socket.c
```diff
@@ -585,6 +783,82 @@ CxPlatDpRawParseUdp(
...
+    //
+    // Packets that don't match the rules above are discarded.
+    //
+    if (Tcp->Flags == TH_ACK && Length > 0) {
+        //
+        // Only data packets with only ACK flag set are indicated to QUIC core.
+        //
+        Packet->Reserved = L4_TYPE_TCP;
+        Packet->Route->TcpState.AckNumber = Tcp->SequenceNumber;
+        Packet->Route->TcpState.SequenceNumber = Tcp->AckNumber;
```
nibanks, line 835: It's a bit confusing that you are setting the route to the inverse of the TCP header. Perhaps a comment explaining why?

src/platform/datapath_raw_socket.c
```diff
@@ -872,27 +1317,50 @@ CxPlatFramingWriteHeaders(
...
-    UDP->Length = QuicNetByteSwapShort((uint16_t)Buffer->Length + sizeof(UDP_HEADER));
-    UDP->Checksum = 0;
+    if (Socket->UseTcp) {
+        //
+        // Fill TCP header.
+        //
+        TCP = (TCP_HEADER*)(Buffer->Buffer - sizeof(TCP_HEADER));
+        TCP->DestinationPort = Route->RemoteAddress.Ipv4.sin_port;
+        TCP->SourcePort = Route->LocalAddress.Ipv4.sin_port;
+        TCP->Window = QuicNetByteSwapShort(65535);
```
nibanks, line 1327: Can we just hard code the hex value instead of always calling the byteswap (which is still a func call I think).

src/platform/datapath_raw_socket.c
```diff
@@ -849,21 +1127,188 @@ CxPlatFramingUdpChecksum(
...
+    _In_ BOOLEAN SkipTransportLayerXsum,
+    _In_ uint32_t TcpSeqNum,
+    _In_ uint32_t TcpAckNum,
+    _In_ uint8_t TcpFlags
     )
 {
-    UDP_HEADER* UDP = (UDP_HEADER*)(Buffer->Buffer - sizeof(UDP_HEADER));
+    uint8_t *Transport;
+    uint16_t TransportLength;
+    uint8_t TransportProtocol = Socket->UseTcp ? IPPROTO_TCP : IPPROTO_UDP;
```
nibanks, line 1309: Move the set of this down into the if/else below so we don't have to do the branch twice.

src/platform/datapath_raw_socket.c
```diff
@@ -872,27 +1317,50 @@ CxPlatFramingWriteHeaders(
...
+        Transport = (uint8_t*)UDP;
+        TransportLength = sizeof(UDP_HEADER);
+    }
 
     //
     // Fill IPv4/IPv6 header.
     //
     if (Family == QUIC_ADDRESS_FAMILY_INET) {
-        IPV4_HEADER* IPv4 = (IPV4_HEADER*)(((uint8_t*)UDP) - sizeof(IPV4_HEADER));
+        IPV4_HEADER* IPv4 = (IPV4_HEADER*)(((uint8_t*)Transport) - sizeof(IPV4_HEADER));
```
nibanks, line 1355:
```suggestion
        IPV4_HEADER* IPv4 = (IPV4_HEADER*)(Transport - sizeof(IPV4_HEADER));
```

src/platform/datapath_raw_socket.c
```diff
@@ -901,13 +1369,24 @@ CxPlatFramingWriteHeaders(
...
+                    CxPlatFramingTransportChecksum(
+                        IPv4->Source, IPv4->Destination,
+                        sizeof(Route->LocalAddress.Ipv4.sin_addr),
+                        IPPROTO_UDP,
+                        (uint8_t*)UDP, sizeof(UDP_HEADER) + Buffer->Length);
+            }
         }
     } else {
-        IPV6_HEADER* IPv6 = (IPV6_HEADER*)(((uint8_t*)UDP) - sizeof(IPV6_HEADER));
+        IPV6_HEADER* IPv6 = (IPV6_HEADER*)(((uint8_t*)Transport) - sizeof(IPV6_HEADER));
```
nibanks, line 1389:
```suggestion
        IPV6_HEADER* IPv6 = (IPV6_HEADER*)(Transport - sizeof(IPV6_HEADER));
```

src/platform/datapath_raw_socket.c
csujedihy: That doesn't matter. The port chosen doesn't affect source routing. The TCP port has already been chosen before the code reaches here and that's the port will be eventually using. This temporary UDP port is only for `connect`.

src/platform/datapath_raw_socket.c
csujedihy: We are calling `connect` on a temporary **UDP** socket.

src/platform/datapath_raw_socket.c
csujedihy: Yeah, should just write 0xFFFF. Swapping makes no difference.

src/platform/datapath_raw_socket.c
csujedihy: Reversed the order here and added a comment in RouteUpdate.

nibanks (approved):

hakarim740-com-ra:
Just out of curiosity is there a standard for QTIP (QUIC over TCP)?

hakarim740-com-ra:
> Nope. It's currently meant as a stop gap measure, and not a long term feature. So, it's not something we want to fully document, nor encourage outside usage of right now.

WOW! thanks for the quick answer.

Issue #3422, opened 2023-02-11
msquic does not load with OpenSSL in FIPS mode


wfurt:
### Describe the bug

This impacts Centos 7 and Mariner 2.0 (OpenSSL 1.1.1k  FIPS) and perhaps other Linux versions.
the load fails as 
```
: symbol lookup error: /lib/libmsquic.so.2: undefined symbol: EVP_chacha20_poly1305, version OPENSSL_1_1_0
```
CHACHA is not approved algorithm and it is missing in system crypto binaries.
This seems to defeat only purpose of `UseSystemOpenSSLCrypto` option. 

### Affected OS

- [ ] Windows
- [X] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

 Centos 7 and Mariner 2.0 

### MsQuic version

release/2.1

### Steps taken to reproduce bug

Try to load published msquic library on impacted OS. 

### Expected behavior

libmsquic loads and ignores unsupported algorithms. If only CHACHA was requested via policy, API calls should return proper error. 

### Actual outcome

libmsuic fails to load because of missing dependencies. 

### Additional details

we should make `QUIC_ALLOWED_CIPHER_SUITE_CHACHA20_POLY1305_SHA256` fully optional. (and related code) 

cc: @ManickaP @CarnaViire

nibanks:
It doesn't even export `EVP_chacha20_poly1305`? Why can't it export but just return `NULL`?

wfurt:
question for OpenSSL bit it does not. I  think that is same for other ciphers you disable during configuration. If you build without them they just disappear. 

This should be easy to fix. I have change that would fix it. I'll put up PR when I clean it up and do more testing.

nibanks:
I'd rather not disable ChaCha20 at build time, but instead dynamically resolve it at runtime. Most folks that use libmsquic package likely **won't** be FIPS, and I want the algorithm to continue to work.

wfurt:
yep. since we already have the XXX_HANDLE `dlsym()` is the answer IMHO. (+little bit sanity)

wfurt:
I did more testing and https://github.com/microsoft/msquic/pull/3423 is unfortunately not sufficient. 
It seems like more creeps in from ssl/ssl_init.c
I will try to find some solution.

nibanks:
If you have something you want/need backported, please just go ahead and create the PR to the correct branch.

nibanks:
[...]

PR #3447, opened 2023-02-17
another attempt to fix quic loading when chacha is missing

wfurt:
The problem is fact that `EVP_chacha20_poly1305` is also referenced from `ossl_init_ssl_base`. That is linked  statically but the dependency may be missing. 

The fix is to override that function and return CXPLAT_CHACHA20_POLY1305_ALG_HANDLE e.g. point at the original implementation. The `ossl_init_ssl_base`  may run before the `CxPlatCryptInitialize` loads. I check and the `EVP_add_cipher` would silently ignore if `NULL` is passed in. 

contributes to #3422

### DIFF
--- a/src/platform/crypt_openssl.c
+++ b/src/platform/crypt_openssl.c
@@ -135,6 +135,16 @@ typedef struct CXPLAT_HP_KEY {
     CXPLAT_AEAD_TYPE Aead;
 } CXPLAT_HP_KEY;
 
+#if defined CXPLAT_SYSTEM_CRYPTO && !defined IS_OPENSSL_3 && !defined _WIN32
+// This is to fulfill link dependency in ssl_init.
+// If system OpenSSL has chacha support, we will redirect it to loaded handle.
+//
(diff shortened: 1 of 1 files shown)


nibanks:
How do we validate these changes in automation?

wfurt:
> How do we validate these changes in automation?

I was thinking about it quite a bit recently. I think the only good way is to add more platform coverage e.g. run BVT on more OSes (or containers). In ideal case you would support & test the oldest OS that .NET (or any other important scenario) need.  I think the static linking eliminates dependencies so variations would not give you much. Basic run with `SystemCrypto` would be sufficient. Probably even just platform subset & TLS. 

From the recent issues that would be Centos7, Alpine Linux & Ubuntu 18.04. 

cc @manickap for any more insight.

nibanks (approved):

nibanks:
Going to merge this as is, but we should look into adding at least some test coverage for this stuff.

wfurt:
thanks @nibanks. Opened https://github.com/microsoft/msquic/issues/3460 for tracking

Issue #1994, opened 2021-09-16
Basic/WithFamilyArgs.DatagramSend/1 Failure


nibanks:
### Describe the bug

Another test failure for slow macOS...

https://dev.azure.com/ms/msquic/_build/results?buildId=221987&view=logs&j=e2236ac7-b409-5bb1-0b69-b4a92122e36a&t=0d0e0a7c-7680-52dc-f392-89c1db582d99&l=313

```
2021-09-15T23:37:11.9656390Z [==========] Running 1 test from 1 test suite.
2021-09-15T23:37:11.9657860Z [----------] Global test environment set-up.
2021-09-15T23:37:11.9658500Z Initializing for User Mode tests
2021-09-15T23:37:11.9659520Z [----------] 1 test from Basic/WithFamilyArgs
2021-09-15T23:37:11.9660210Z [ RUN      ] Basic/WithFamilyArgs.DatagramSend/1
2021-09-15T23:37:11.9660880Z /Users/runner/work/1/s/src/test/lib/DatagramTest.cpp:240: Failure
2021-09-15T23:37:11.9661570Z Client.GetDatagramsSuspectLost() not equal to 1
2021-09-15T23:37:11.9662230Z [  FAILED  ] Basic/WithFamilyArgs.DatagramSend/1, where GetParam() = v6 (1218 ms)
2021-09-15T23:37:11.9664160Z [----------] 1 test from Basic/WithFamilyArgs (1218 ms total)
```

ami-GS:
Again?
I could repro on Ubuntu
https://dev.azure.com/ms/msquic/_build/results?buildId=429009&view=logs&j=3667e279-20cb-5b66-4f7d-50a166799211&t=69945233-3564-536d-0950-ae97d4372f33

ami-GS:
My assumption is race condition.

<img width="804" alt="image" src="https://user-images.githubusercontent.com/5763034/219999471-494c785f-8d3a-46e1-a850-6bcd3fa5f1b3.png">
Difference between successful and unsuccessful are the number of "Indicating DATAGRAM_SEND_STATE_CHANGED to 2" calls.
This goes callback and Callback in TestConnection.cpp increments DatagramsSuspectLost.

https://github.com/microsoft/msquic/blob/83fc6400ce7c37abfef3012e83b4bd037c5cb1c0/src/test/lib/TestConnection.cpp#L865-L874

main test code validates whether the variable is 1, but this failed because the value is not 1 (maybe 2).
[...]

PR #3458, opened 2023-02-20
Fix DatagramSend test by expecting multiple loss

ami-GS:
## Description

Fixes https://github.com/microsoft/msquic/issues/1994
Datagram get lost at unexpected point. Which causes failure by validating the number of packets get lost.

## Testing

Running some hours.....

### DIFF
--- a/src/test/lib/DatagramTest.cpp
+++ b/src/test/lib/DatagramTest.cpp
@@ -225,6 +225,7 @@ QuicTestDatagramSend(
                 TEST_EQUAL(1, Client.GetDatagramsAcknowledged());
 
 #if QUIC_TEST_DATAPATH_HOOKS_ENABLED
+                uint32_t CurrentLostCount = Client.GetDatagramsSuspectLost();
                 LossHelper.DropPackets(1);
 
                 TEST_QUIC_SUCCEEDED(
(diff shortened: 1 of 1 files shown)


src/test/lib/DatagramTest.cpp
```diff
@@ -245,7 +245,15 @@ QuicTestDatagramSend(
...
                     CxPlatSleep(100);
                 }
-                TEST_EQUAL(1, Client.GetDatagramsSuspectLost());
+                // NOTE:
+                // There are two cases of DATAGRAM_SEND_STATE_CHANGED
+                // 1. 1->4->1->2->3
+                //      This is expected state changes. 99.99...% cases
+                // 2. 1->2->5->1->2
+                //      This is unexpected. 0.0...1% cases
+                //      state changes to LOST_SUSPECT before the first datagram acked (line 225)
```
csujedihy, line 254: Why would this datagram be marked as LOST_SUSPECT in the first place? Isn't that the root cause of the failure?

nibanks: I tend to agree. Why would it be lost?

ami-GS:
Logs of both successful and unsuccessful are same up to a Server sends ack against 1st datagram.

Below is an logs, left is unsuccessful, right is successful case. Colored in red where the key difference and key part.

1. Address of Connection is same/different for "Updating ...." and "Next Expiration"
2. left shows "2 expired", this is simply timer expired for the packet including datagram. Then goes to "DATAGRAM_SEND_STATE_CHANGED to 2"
image

src/test/lib/DatagramTest.cpp
nibanks: I guess it's possible that more packets could be suspected as lost because of such a low RTT triggering the loss detection logic overly aggressively, in the case where an oversubscribed VM runner gets delayed.

Instead of all these changes, just update the code to make sure `Client.GetDatagramsSuspectLost() >= 1` to be more resilient to additional packet loss. Same for the `while` loop above. And instead of this big comment, just put a comment indicating it's possible for more than one packet to get suspected as lost if the machine is running slow (oversubscribed).

src/test/lib/DatagramTest.cpp
```diff
@@ -225,6 +225,7 @@ QuicTestDatagramSend(
                 TEST_EQUAL(1, Client.GetDatagramsAcknowledged());
 
 #if QUIC_TEST_DATAPATH_HOOKS_ENABLED
+                uint32_t CurrentLostCount = Client.GetDatagramsSuspectLost();
```
nibanks, line 228: Let's name this `InitialLostCount` (and make it `const`).

src/test/lib/DatagramTest.cpp
csujedihy: It seems to me that it's because the delay ACK timer on the peer fired too late. It was scheduled to be fire at 40.643s (40.618+0.025) but it fired at 40.655s. So, the PTO raced with it and won. If it was TCP, I'd do something to effectively disable delay ACK timeout or send 2 packets to trigger an immediate ACK.

Though, no objection to relaxing the `Client.GetDatagramsSuspectLost() >= 1` check.

nibanks: Code all looks fine. Can you please just fix up the comment above like I mentioned above?

src/test/lib/DatagramTest.cpp
```diff
@@ -241,11 +242,20 @@ QuicTestDatagramSend(
...
 
+
+                // NOTE:
+                // There are two cases of DATAGRAM_SEND_STATE_CHANGED
+                // 1. 1->4,   1->2->3
+                //      This is expected state changes. 99.99...% cases
+                // 2. 1->2-5, 1->2
+                //      This is unexpected. 0.0...1% cases
+                //      state changes to LOST_SUSPECT before the first datagram acked (line 225)
+                //      However line 225 (DatagramsAcknowledged) becomes true by treating state 5 (ACKNOWLEDGED_SPURIOUS) as 4 (ACKNOWLEDGED)
```
nibanks, line 253:
```suggestion
                //
                // Even though the test only drops a single packet it is
                // possible that an oversubscribed VM may result in additional
                // packet drops. Account for this by ensuring at least one new
                // packet is dropped by this test.
                //
```

nibanks (approved):

PR #3462, opened 2023-02-24
Ingest XDP v0.16.7 and use TCP CIBIR rules when needed

csujedihy:
## Description

Ingest XDP v0.16.7 and use TCP CIBIR rules when needed.

Changes to our tests:
1. 0-RTT tests are disabled. QTIP caches only one packet before handshake completes. So, if multiple 0-RTT packets are sent, only one of them will make it.
2. NAT rebind doesn't apply to TCP.  So, they are disabled.
3. Due to implementation limitations, sharing bindings isn't supported. Some tests are tweaked to always create a new binding per connection. For simplicity, the implementation assumes client side sockets have a 1:1 mapping to QUIC connection.

## Testing

CI

### DIFF
--- a/.azure/azure-pipelines.ci.yml
+++ b/.azure/azure-pipelines.ci.yml
@@ -661,17 +661,6 @@ stages:
       logProfile: Full.Light
   - template: ./templates/run-bvt.yml
     parameters:
-      displayName: QTIP
-      image: windows-2022
-      platform: windows
-      tls: schannel
-      logProfile: Full.Light
-      extraArtifactDir: '_Xdp'
-      extraPrepareArgs: -InstallDuoNic -InstallXdpDriver
-      # TODO: reenable these testcases: *Unreachable*:*DrillInitialPacket*
-      extraTestArgs: -DuoNic -Filter -*ValidateConfiguration:*ValidAlpnLengths:*ResumeRejection*:*ClientCertificate*:*LoadBalanced*:*Tcp*:*Unreachable*:*DrillInitialPacket*:TlsTest.InProc* -ExtraArtifactDir Xdp -UseQtip
-  - template: ./templates/run-bvt.yml
-    parameters:
       image: windows-2022
       platform: windows
       tls: schannel
--- a/.azure/azure-pipelines.ci.yml
+++ b/.azure/azure-pipelines.ci.yml
@@ -661,6 +661,17 @@ stages:
       logProfile: Full.Light
   - template: ./templates/run-bvt.yml
     parameters:
+      displayName: QTIP
+      image: windows-2022
+      platform: windows
+      tls: schannel
+      logProfile: Full.Light
+      extraArtifactDir: '_Xdp'
+      extraPrepareArgs: -InstallDuoNic -InstallXdpDriver
+      # TODO: reenable these testcases: *Unreachable*:*DrillInitialPacket*
+      extraTestArgs: -DuoNic -Filter -*ValidateConfiguration:*ValidAlpnLengths:*ResumeRejection*:*ClientCertificate*:*LoadBalanced*:*Tcp*:*Unreachable*:*DrillInitialPacket*:TlsTest.InProc* -ExtraArtifactDir Xdp -UseQtip
+  - template: ./templates/run-bvt.yml
+    parameters:
       image: windows-2022
       platform: windows
       tls: schannel
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -283,7 +283,6 @@ if (!$OfficialRelease) {
         # Thanks to https://stackoverflow.com/questions/3404936/show-which-git-tag-you-are-on
         # for this magic git command!
         $Output = git describe --exact-match --tags $(git log -n1 --pretty='%h')
-        Write-Host $Output
         if (!$Output.Contains("fatal: no tag exactly matches")) {
             Write-Host "Configuring OfficialRelease for tag build"
             $OfficialRelease = $true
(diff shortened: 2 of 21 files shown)


csujedihy:
@nibanks do you know what secnetperf commands I can use to test this end to end?

nibanks:
> @nibanks do you know what secnetperf commands I can use to test this end to end? Is it just the client and server that use the same hex passed in `-cibir:` can talk to each other?

Yeah, that should work.

csujedihy:
Oh damn, it doesn't really simply work this way. The handshake won't be completed because the handshake packets don't have "UDP" payload and won't be picked up by CIBIR rules.

nibanks:
> I am sorry I didn't foresee this. :/

No problem. 😃 None of us did.

I'll have to think about the options here. I'd prefer less changes on the XDP side, and more on the QUIC side to work around, so it doesn't possibly effect non-QTIP stuff on the XDP side.

csujedihy:
> Oh damn, it doesn't really simply work this way. The handshake won't be completed because the handshake packets don't have "UDP" payload and won't be picked up by CIBIR rules.



> > I am sorry I didn't foresee this. :/
> 
> No problem. 😃 None of us did.
> 
> I'll have to think about the options here. I'd prefer less changes on the XDP side, and more on the QUIC side to work around, so it doesn't possibly effect non-QTIP stuff on the XDP side.

Sync'd up with Michael. I am going with option 1.

nibanks:
I've tried adding the test code, but something really wonky is going on with `QUIC_API_ENABLE_PREVIEW_FEATURES` definition on Windows. I don't understand why it's not getting set... Any ideas?

csujedihy:
> > Oh damn, it doesn't really simply work this way. The handshake won't be completed because the handshake packets don't have "UDP" payload and won't be picked up by CIBIR rules.
> 
> > > I am sorry I didn't foresee this. :/
> > 
> > 
> > No problem. 😃 None of us did.
> > I'll have to think about the options here. I'd prefer less changes on the XDP side, and more on the QUIC side to work around, so it doesn't possibly effect non-QTIP stuff on the XDP side.
> 
> Sync'd up with Michael. I am going with option 1.

Pushed a new iteration with the new XDP rule. CIBIR is working now.

scripts/build.ps1
```diff
@@ -283,6 +283,7 @@ if (!$OfficialRelease) {
         # Thanks to https://stackoverflow.com/questions/3404936/show-which-git-tag-you-are-on
         # for this magic git command!
         $Output = git describe --exact-match --tags $(git log -n1 --pretty='%h')
+        Write-Host $Output
```
nibanks, line 286: Remove this now?

src/test/lib/DataTest.cpp
```diff
@@ -14,6 +14,10 @@
 #include "DataTest.cpp.clog.h"
 #endif
 
+#ifdef QUIC_USE_RAW_DATAPATH
+extern bool UseQTIP;
+#endif
```
nibanks, line 19: Would it be better to make this bool always present but only ever set to true when RAW is used? That'd simplify all these #ifdefs

src/test/lib/MtuTest.cpp
```diff
@@ -45,6 +49,12 @@ struct ResetSettings {
 void
 QuicTestMtuSettings()
 {
+#ifdef QUIC_USE_RAW_DATAPATH
+    uint16_t DefaultMaximumMtu = UseQTIP ? 1488 : 1500; // reserve 12B for TCP header
+#else
+    uint16_t DefaultMaximumMtu = 1500;
+#endif
```
nibanks, line 56: Make them `const`?

src/test/lib/MtuTest.cpp
```diff
@@ -45,6 +49,12 @@ struct ResetSettings {
 void
 QuicTestMtuSettings()
```
nibanks, line 50: This test is really fragile. It assumes a max MTU that we might change in the future...

src/test/lib/TestHelpers.h
```diff
@@ -46,6 +46,10 @@ QuicAddrSetToDuoNic(
     }
 }
 
+#ifndef QUIC_OFFICIAL_RELEASE
+#define QUIC_API_ENABLE_PREVIEW_FEATURES
+#endif
```
nibanks, line 51: We should be able to remove this now. I added it when I was trying to figure out what was going on.

csujedihy: removed

src/test/lib/MtuTest.cpp
csujedihy: done

src/test/lib/DataTest.cpp
csujedihy: That can't be shared across user and kernel mode.

.azure/azure-pipelines.ci.yml
```diff
@@ -659,6 +659,17 @@ stages:
       platform: windows
       tls: schannel
       logProfile: Full.Light
+  - template: ./templates/run-bvt.yml
```
csujedihy, line 662: How can I get this to work?

csujedihy: I am getting this error: `Stage test_bvt job bvt_windows_x64_schannel_false_false_windows_2022_Xdp appears more than once. Job names must be unique within a stage.`

nibanks: I think there's an arg like extra name you need to set to something unique

csujedihy: Ah, I see. The parameters form the job name in run-bvt.yml. However, the job name doesn't include extraName. I will remove this for name and handle it in a separate PR after getting the github actions stable.

csujedihy:
The only failing test cases are NAT rebind cases. These don't make sense for TCP.

nibanks (approved):

nibanks (approved):

Issue #2344, opened 2022-01-28
QUIC_TLS_SECRETS Not Fully Populated on Server


nibanks:
### Describe the bug

The `ClientRandom` isn't populated in the `QUIC_TLS_SECRETS` for the server side, because the code reads that from the TLS ClientHello before the connection is even indicated to the listener; meaning the app didn't have a chance to set the pointer to the struct to be written to.

### Affected OS

- [X] Windows Server 2022
- [X] Windows 11
- [X] Windows Insider Preview (specify affected build below)
- [X] Ubuntu
- [X] Debian
- [ ] Other (specify below)

### MsQuic version

main

### Steps taken to reproduce bug

Try to query the `QUIC_TLS_SECRETS`.

### Expected behavior

They are fully populated.

### Actual outcome

Everything but `ClientRandom` is populated.

PR #3481, opened 2023-03-07
get client random if setting TlsSecrets on server

wfurt:
OpenSSL only, contributes to #2344


nibanks (approved):
cc @anrossi, I'm ok with this as a temporary measure, but I'd really prefer a TLS-agnostic solution in the QUIC code, long term.

anrossi:
@nibanks Agreed. This is fine for now.

anrossi (approved):

wfurt:
related to https://github.com/dotnet/runtime/pull/83001. I work out hook in .NET (for debug build only)
I started with Linux as it was super easy to fix. 
cc: @manickap

PR #3514, opened 2023-03-16
Use the new service connection for ACR

csujedihy:
The old ACR got deleted. Update the service connection for the new ACR.

### DIFF
--- a/.azure/OneBranch.Publish.yml
+++ b/.azure/OneBranch.Publish.yml
@@ -25,7 +25,7 @@ jobs:
       displayName: Login to ACR
       inputs:
         command: login
-        containerRegistry: dockerRegSrvConn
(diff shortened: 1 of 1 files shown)


anrossi:
Looks good to me. Does this mean the URL for getting our docker container has changed?  If so, we'll have to inform current users consuming it; I'm thinking of the InteropRunner and some of our tests, are there others?

csujedihy:
> Looks good to me. Does this mean the URL for getting our docker container has changed? If so, we'll have to inform current users consuming it; I'm thinking of the InteropRunner and some of our tests, are there others?

It only affects azure container registry. I don't think we have anything that uses it except this PMC publish pipeline.

nibanks (approved):

PR #3539, opened 2023-03-27
Fix QUIC_TLS_SECRETS on Server and Client.

### LINKED ISSUE #2344: QUIC_TLS_SECRETS Not Fully Populated on Server

anrossi:
## Description

 Fixes #2344
Specifically,
* Copies the ClientRandom on server side.
* Copies the ClientEarlyTrafficSecret on server side.
* Copies the ClientRandom on client side when 0-RTT is used.

## Testing

* Manual testing with quicinteropserver to ensure the ClientRandom is logged.
* New automated tests.

## Documentation

N/A

### DIFF
--- a/src/core/crypto.c
+++ b/src/core/crypto.c
@@ -1538,7 +1538,6 @@ QuicCryptoProcessTlsCompletion(
             Crypto->TlsState.BufferLength > 0) {
 
             QuicCryptoTlsReadClientRandom(
-                Connection,
                 Crypto->TlsState.Buffer,
                 Crypto->TlsState.BufferLength,
                 Connection->TlsSecrets);
@@ -1882,7 +1881,6 @@ QuicCryptoProcessData(
                 // so now the ClientRandom can be copied.
                 //
                 QuicCryptoTlsReadClientRandom(
-                    Connection,
                     Buffer.Buffer,
                     Buffer.Length,
                     Connection->TlsSecrets);
--- a/src/core/crypto.c
+++ b/src/core/crypto.c
@@ -1874,9 +1874,9 @@ QuicCryptoProcessData(
                 Connection,
                 &Info);
 
-            if (!Connection->State.HandleClosed &&
-                Connection->State.ExternalOwner &&
-                Connection->TlsSecrets != NULL) {
+            if (Connection->TlsSecrets != NULL &&
+                !Connection->State.HandleClosed &&
+                Connection->State.ExternalOwner) {
                 //
                 // At this point, the connection was accepted by the listener,
                 // so now the ClientRandom can be copied.
--- a/src/core/crypto.c
+++ b/src/core/crypto.c
@@ -1881,10 +1881,12 @@ QuicCryptoProcessData(
                 Connection,
                 &Info);
 
-            if (!Connection->State.HandleClosed && Connection->State.ExternalOwner && Connection->TlsSecrets != NULL) {
+            if (!Connection->State.HandleClosed &&
+                Connection->State.ExternalOwner &&
+                Connection->TlsSecrets != NULL) {
                 //
-                // At this point, if the connection was accepted by the listener, and
-                // the app requests the TLS secrets, now they can be copied.
+                // At this point, the connection was accepted by the listener,
+                // and the app requests TLS secrets, so now they can be copied.
                 //
                 Status =
                     QuicCryptoTlsReadInitial(
--- a/src/core/crypto.c
+++ b/src/core/crypto.c
@@ -1536,12 +1536,11 @@ QuicCryptoProcessTlsCompletion(
             (Crypto->TlsState.WriteKey == QUIC_PACKET_KEY_INITIAL ||
                 Crypto->TlsState.WriteKey == QUIC_PACKET_KEY_0_RTT) &&
             Crypto->TlsState.BufferLength > 0) {
-            QUIC_NEW_CONNECTION_INFO Info = { 0 };
-            QuicCryptoTlsReadInitial(
+
+            QuicCryptoTlsReadClientRandom(
                 Connection,
                 Crypto->TlsState.Buffer,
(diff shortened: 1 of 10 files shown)


src/core/crypto.c
```diff
@@ -1880,6 +1880,20 @@ QuicCryptoProcessData(
                 Connection->Paths[0].Binding,
                 Connection,
                 &Info);
+
+            if (!Connection->State.HandleClosed && Connection->State.ExternalOwner && Connection->TlsSecrets != NULL) {
```
nibanks, line 1884:
```suggestion
            if (!Connection->State.HandleClosed &&
                Connection->State.ExternalOwner &&
                Connection->TlsSecrets != NULL) {
```

nibanks:
We should write a test that verifies the client and server secrets & random match.

anrossi:
> We should write a test that verifies the client and server secrets & random match.

Gonna need to think about this a bit. It would involve recreating some TLS logic, and what exactly what it be compared against? 
The test hooks might not exist for this yet.

nibanks:
> > We should write a test that verifies the client and server secrets & random match.
> 
> Gonna need to think about this a bit. It would involve recreating some TLS logic, and what exactly what it be compared against? The test hooks might not exist for this yet.

Just take a simple handshake test, update both sides to grab the secrets. After the handshake completes, compare the client and server secrets.

src/core/crypto.c
```diff
@@ -1880,6 +1880,22 @@ QuicCryptoProcessData(
                 Connection->Paths[0].Binding,
                 Connection,
                 &Info);
+
+            if (!Connection->State.HandleClosed &&
```
anrossi, line 1884: Would appreciate a second pair of eyes on this. I'm not sure if this is the best way to tell that the connection has accepted.

nibanks: I think this should be fine.

src/test/lib/DataTest.cpp
```diff
@@ -512,6 +542,42 @@ QuicTestConnectAndPing(
...
+                for (auto j = 0u; j < ConnectionCount; j++) {
+                    auto ClientSecret = &ClientSecrets[j];
+                    if (!memcmp(
+                            ServerSecret->ClientRandom,
+                            ClientSecret->ClientRandom,
+                            sizeof(ClientSecret->ClientRandom))) {
+                        if (Match) {
+                            TEST_FAILURE("Multiple clients with the same ClientRandom?!");
+                            return;
+                        }
```
nibanks, line 559: This is theoretically possible, right? Just HIGHLY unlikely. Should we fail the tests if _somehow_ it happens?

nibanks (approved):

src/core/crypto.c
```diff
@@ -1533,7 +1533,8 @@ QuicCryptoProcessTlsCompletion(
         //
         if (Connection->TlsSecrets != NULL &&
             QuicConnIsClient(Connection) &&
-            Crypto->TlsState.WriteKey == QUIC_PACKET_KEY_INITIAL &&
+            (Crypto->TlsState.WriteKey == QUIC_PACKET_KEY_INITIAL ||
+                Crypto->TlsState.WriteKey == QUIC_PACKET_KEY_0_RTT) &&
             Crypto->TlsState.BufferLength > 0) {
             QUIC_NEW_CONNECTION_INFO Info = { 0 };
             QuicCryptoTlsReadInitial(
```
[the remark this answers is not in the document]
nibanks: In the Handshake/WithHandshakeArgs2.OldVersion/* tests. Does this get called twice or something in that case?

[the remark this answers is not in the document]
anrossi: Still fails even with this line reverted, so it's something else that's leaking...

src/core/crypto_tls.c
```diff
@@ -1744,6 +1744,9 @@ QuicCryptoTlsDecodeTransportParameters( // NOLINT(readability-function-size, goo
 
         case QUIC_TP_ID_VERSION_NEGOTIATION_EXT:
             if (Length > 0) {
+                if (TransportParams->VersionInfo != NULL) {
```
anrossi, line 1747: Remove this; it's not this.

src/core/crypto_tls.c
```diff
@@ -664,6 +656,32 @@ QuicCryptoTlsReadInitial(
...
+QUIC_STATUS
+QuicCryptoTlsReadClientRandom(
+    _In_ QUIC_CONNECTION* Connection,
+    _In_reads_(BufferLength)
+        const uint8_t* Buffer,
+    _In_ uint32_t BufferLength,
+    _Inout_ QUIC_TLS_SECRETS* TlsSecrets
+    )
+{
+    if (BufferLength < TLS_MESSAGE_HEADER_LENGTH + sizeof(uint16_t) + TLS_RANDOM_LENGTH) {
```
nibanks, line 669: Can't you assert this instead? It was already validated with a previous call.

anrossi: It'll generate a SAL issue without a check

nibanks (dismissed):

nibanks (approved):

nibanks (approved):

PR #3545, opened 2023-04-01
Handle sending path challenge on paths that have not resolved route yet

csujedihy:
## Description

It's possible that when sending a path challenge, the route is not yet resolved. We need to pause and resume path challenge when that happens.

## Testing

CI

### DIFF
--- a/src/core/send.c
+++ b/src/core/send.c
@@ -1049,7 +1049,6 @@ QuicSendPathChallenges(
         // We need to set the path challenge flag back on so that when route is resolved,
         // we know we need to continue to send the challenge.
         //
-        CXPLAT_DBG_ASSERT(Path->IsActive);
         CXPLAT_DBG_ASSERT(Path->Route.State != RouteSuspected);
         if (Path->Route.State == RouteUnresolved) {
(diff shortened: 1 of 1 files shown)


csujedihy:
Will take a look at the failures when I am back on Tuesday or Wednesday

src/core/send.c
```diff
@@ -1040,6 +1042,38 @@ QuicSendPathChallenges(
             continue;
         }
 
+#ifdef QUIC_USE_RAW_DATAPATH
+        //
+        // Make sure the route is resolved before sending the path challenge.
```
nibanks, line 1047: Any ideas on how? Could we make a test to explicitly trigger this?

csujedihy: You meant how a route is not resolved before sending the path challenge? Isn't this already covered by existing address rebind test cases? When we create a new path upon seeing a different source address from the peer, we create a path with an unresolved route.

nibanks: Ok.

nibanks (approved):

nibanks:
@csujedihy wasn't there a bug that could be closed along with this?

PR #3586, opened 2023-04-26
docs: Replace NEW_STREAM with PEER_STREAM_STARTED

PeteAudinate:
## Description

There is no QUIC_CONNECTION_EVENT_NEW_STREAM event. Should this be QUIC_CONNECTION_EVENT_PEER_STREAM_STARTED  instead?

If not, it would be good to see an example of how to do this.

## Testing

N/A

## Documentation

Updated


nibanks (approved):
Good catch! Thanks for the fix.

PR #3591, opened 2023-04-27
Build and push linux build dockers in automation

csujedihy:
## Description

Build and push linux cross-compile build dockers in automation.

## Testing

GH CI


nibanks:
Instead of two separate workflows, could we have 1 workflow with different jobs/stages? It makes it a little cleaner in the Github Action UI, IMHO.

nibanks (dismissed):

nibanks (approved):

PR #3599, opened 2023-05-01
Automatic linux package publish pipeline Part 1

csujedihy:
## Description

For some reasons, the onebranch publish pipeline created on a topic branch doesn't always trigger. The pipeline is supposed to trigger just like msquic-Official-Tests pipeline. However, it only triggers if the pipeline does not have any parameters defined, which is super weird.

It does not hurt to check this in to see if the change needs to be on main for it to trigger.

## Testing

CI


.azure/obtemplates/build-linux-packages.yml
```diff
@@ -46,7 +46,6 @@ jobs:
   - ${{ if not(eq(parameters.sign, false)) }}:
     - task: onebranch.pipeline.signing@1
       displayName: Sign
-      target: linux_build_container2
```
nibanks, line 49: Why did you remove this?

csujedihy: I will use the linux_build_container be default. Now that both linux_build_container and linux_build_container2 are cross-compile build dockers, it doesn't matter which we use.

nibanks (approved):

PR #3607, opened 2023-05-03
Call UpdateQeo from Connection

ami-GS:
## Description
Call UpdateQeo from Connection
Leveraging https://github.com/microsoft/msquic/pull/3600
- QUIC_SETTINGS to have EncryptionOffloadEnabled flag
- QUIC_CONNECTION_STATE to have state to indicate whether currently offloading
- Offloading is actually enabled at the end of ConnStart


src/inc/quic_datapath.h
```diff
@@ -436,6 +436,7 @@ CxPlatDataPathUninitialize(
 #define CXPLAT_DATAPATH_FEATURE_SEND_SEGMENTATION     0x0004
 #define CXPLAT_DATAPATH_FEATURE_LOCAL_PORT_SHARING    0x0008
 #define CXPLAT_DATAPATH_FEATURE_PORT_RESERVATIONS     0x0010
+#define CXPLAT_DATAPATH_FEATURE_ENCRYPTION_OFFLOAD    0x0011
```
csujedihy, line 439: 0x0020

src/inc/msquic.h
```diff
@@ -709,6 +710,7 @@ typedef struct QUIC_SETTINGS {
         };
     };
 
+    uint8_t EncryptionOffloadEnabled        : 1;
```
nibanks, line 713: Add this above, taking one of the reserved bits.

src/core/connection.c
```diff
@@ -485,6 +485,32 @@ QuicConnUninitialize(
         QuicBindingRemoveConnection(Connection->Paths[0].Binding, Connection);
     }
 
+    if (Connection->State.EncryptionOffloading) {
+        CXPLAT_QEO_CONNECTION Offloads[] = {
```
nibanks, line 489: You will have to fill out the connection IDs at least.

src/core/connection.c
```diff
@@ -2003,6 +2029,38 @@ QuicConnStart(
         goto Exit;
     }
 
+    if (Connection->Settings.IsSet.EncryptionOffloadEnabled) {
+        CXPLAT_QEO_CONNECTION Offloads[] = {
```
nibanks, line 2033: You need to fill out the structs entirely (CIDs and key info), which means you can only fill these out when all the info is available, which is after the handshake completes and the keys are available, see `QuicCryptoProcessTlsCompletion` and maybe `QuicPathSetActive` too.

src/core/connection.c
```diff
@@ -2003,6 +2029,38 @@ QuicConnStart(
...
+                CXPLAT_QEO_CIPHER_TYPE_AEAD_AES_256_GCM,
+                0,
+                {0},
+            }
+        };
+        Status = CxPlatSocketUpdateQeo(Path->Binding->Socket, Offloads, 2);
+        if (QUIC_FAILED(Status)) {
+            // TODO: query QEO capability, set CXPLAT_DATAPATH_FEATURE_ENCRYPTION_OFFLOAD
+            goto Exit;
+        }
```
nibanks, line 2059: Failure here shouldn't be fatal. Instead, only set the offloading flag to true on success.

src/core/connection.h
```diff
@@ -194,6 +194,11 @@ typedef union QUIC_CONNECTION_STATE {
         //
         BOOLEAN DisableVneTp : 1;
 #endif
+
+        //
+        // Indicates whether this connection offloads encryption workload to HW
+        //
+        BOOLEAN EncryptionOffloading : 1;
```
nibanks, line 201: I _think_ we might need to set this per-path, not connection-wide (or at least not _just_ connection-wide) because when we set for each path, it could fail on one path and not the other depending on HW resources.

src/core/settings.c
```diff
@@ -1673,6 +1686,7 @@ QuicSettingsGetSettings(
     SETTING_COPY_FROM_INTERNAL(ServerResumptionLevel, Settings, InternalSettings);
     SETTING_COPY_FROM_INTERNAL(GreaseQuicBitEnabled, Settings, InternalSettings);
     SETTING_COPY_FROM_INTERNAL(EcnEnabled, Settings, InternalSettings);
+    SETTING_COPY_FROM_INTERNAL(EncryptionOffloadEnabled, Settings, InternalSettings);
```
nibanks, line 1689: You need to move this down after `HyStartEnabled` and do it the same way.

src/inc/msquic.h
```diff
@@ -500,7 +500,8 @@ typedef struct QUIC_STATISTICS_V2 {
     uint32_t ResumptionSucceeded    : 1;
     uint32_t GreaseBitNegotiated    : 1;    // Set if we negotiated the GREASE bit.
     uint32_t EcnCapable             : 1;
-    uint32_t RESERVED               : 26;
+    uint32_t EncryptionOffloading   : 1;
```
nibanks, line 503:
```suggestion
    uint32_t EncryptionOffloaded    : 1; // At least one path successfully offloaded encryption.
```

src/inc/quic_datapath.h
```diff
@@ -436,6 +436,7 @@ CxPlatDataPathUninitialize(
 #define CXPLAT_DATAPATH_FEATURE_SEND_SEGMENTATION     0x0004
 #define CXPLAT_DATAPATH_FEATURE_LOCAL_PORT_SHARING    0x0008
 #define CXPLAT_DATAPATH_FEATURE_PORT_RESERVATIONS     0x0010
+#define CXPLAT_DATAPATH_FEATURE_ENCRYPTION_OFFLOAD    0x0020
```
nibanks, line 439: Unnecessary, right? Let's remove for now.

src/core/settings.h
```diff
@@ -98,6 +99,7 @@ typedef struct QUIC_SETTINGS_INTERNAL {
     uint8_t EcnEnabled                      : 1;
     uint8_t HyStartEnabled                  : 1;
     uint8_t MtuDiscoveryMissingProbeCount;
+    uint8_t EncryptionOffloadEnabled        : 1;
```
nibanks, line 102:
```suggestion
    uint8_t EncryptionOffloadEnabled        : 1;
    uint8_t MtuDiscoveryMissingProbeCount;
```
The core struct can be reordered. Let's keep the bitfields together.

src/core/connection.h
```diff
@@ -243,6 +248,7 @@ typedef struct QUIC_CONN_STATS {
     uint32_t ResumptionAttempted    : 1;
     uint32_t ResumptionSucceeded    : 1;
     uint32_t GreaseBitNegotiated    : 1;
+    uint32_t EncryptionOffloading   : 1;
```
nibanks, line 251:
```suggestion
    uint32_t EncryptionOffloaded    : 1;
```

src/core/connection.c
```diff
@@ -7230,6 +7289,11 @@ QuicConnApplyNewSettings(
         }
     }
 
+    if (Connection->State.Started &&
+        (Connection->Settings.EncryptionOffloadEnabled ^ Connection->State.EncryptionOffloading)) {
+        // TODO: enable/disable after start
```
nibanks, line 7294: Put a `CXPLAT_FRE_ASSERT(FALSE)` here for now to make sure it doesn't happen.

nibanks (commented):
Looking good so far. Still a few things left to do.

src/inc/msquic.h
```diff
@@ -663,7 +664,8 @@ typedef struct QUIC_SETTINGS {
             uint64_t GreaseQuicBitEnabled                   : 1;
             uint64_t EcnEnabled                             : 1;
             uint64_t HyStartEnabled                         : 1;
-            uint64_t RESERVED                               : 29;
+            uint64_t EncryptionOffloadEnabled               : 1;
```
nibanks, line 667: Let's wrap the `QUIC_SETTINGS` changes in `#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES` checks for now. You can leave the stats one without.

nibanks, line 667: Let's call this `EncryptionOffloadAllowed` instead of enabled. It can only truely be _enabled_ if the HW & platform supports it.

src/core/connection.c
ami-GS: Which should be used? OrigDestCID? or which one of DestCIDs

nibanks: The destination IP and CID should be used, where "destination" depends on the direction. For TX, it is the peer's, but for RX, it's our local one.

src/core/quicdef.h
```diff
@@ -516,6 +516,11 @@ CXPLAT_STATIC_ASSERT(
 //
 #define QUIC_DEFAULT_HYSTART_ENABLED                FALSE
 
+//
+// The default settings for allowing QEO support.
+//
+#define QUIC_DEFAULT_ENCRYPTION_OFFLOAD_ALLOWED      TRUE
```
nibanks, line 522: Default to `FALSE` for now.

src/core/settings.c
```diff
@@ -144,6 +144,12 @@ QuicSettingsSetDefault(
     if (!Settings->IsSet.HyStartEnabled) {
         Settings->HyStartEnabled = QUIC_DEFAULT_HYSTART_ENABLED;
     }
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
```
nibanks, line 147: You don't need all these checks in the core code, just msquic.h. This is **always** defined for core.

src/core/connection.c
```diff
@@ -485,6 +485,27 @@ QuicConnUninitialize(
...
+        Offloads[0].Direction = CXPLAT_QEO_DIRECTION_TRANSMIT;
+        Offloads[0].ConnectionIdLength = Path->DestCid->CID.Length;
+        Offloads[1].Operation = CXPLAT_QEO_OPERATION_REMOVE;
+        Offloads[1].Direction = CXPLAT_QEO_DIRECTION_RECEIVE;
+        Offloads[1].ConnectionIdLength = SourceCid->CID.Length;
+        memcpy(Offloads[0].ConnectionId, Path->DestCid->CID.Data, Path->DestCid->CID.Length);
+        memcpy(Offloads[1].ConnectionId, SourceCid->CID.Data, SourceCid->CID.Length);
+        if (!QUIC_SUCCEEDED(CxPlatSocketUpdateQeo(Path->Binding->Socket, Offloads, 2))) {
+            // TODO: Log
+        }
```
nibanks, line 504:
```suggestion
        (void)CxPlatSocketUpdateQeo(Path->Binding->Socket, Offloads, 2);
```
Let's just make this stuff "best effort". I don't care much if it fails.

src/inc/quic_tls.h
```diff
@@ -198,6 +198,12 @@ typedef struct CXPLAT_TLS_CONFIG {
     //
     QUIC_TLS_SECRETS* TlsSecrets;
 
+    //
+    // Storage for TLS traffic secrets when the connection has the parameter set
+    // to allow encryption offloading
+    //
+    QUIC_TLS_OFFLOAD_SECRETS* TlsOffloadSecrets;
```
nibanks, line 205: Shouldn't this already be stored on the keys themselves?

ami-GS: You mean inside QUIC_PACKET_KEY defined as WriteKeys/ReadKeys?

src/core/crypto.c
```diff
@@ -1693,7 +1693,7 @@ QuicCryptoProcessTlsCompletion(
             memcpy(Offloads[1].PayloadIv, Path->TlsOffloadSecrets->Rx.PayloadIv, Path->TlsOffloadSecrets->Rx.PayloadIvLength);
             memcpy(Offloads[1].PayloadKey, Path->TlsOffloadSecrets->Rx.PayloadKey, Path->TlsOffloadSecrets->Rx.PayloadKeyLength);
             memcpy(Offloads[1].HeaderKey, Path->TlsOffloadSecrets->Rx.HeaderKey, Path->TlsOffloadSecrets->Rx.HeaderKeyLength);
-            if (QUIC_SUCCEEDED(CxPlatSocketUpdateQeo(Path->Binding->Socket, Offloads, 2))) {
+            if (Path->Binding != NULL && QUIC_SUCCEEDED(CxPlatSocketUpdateQeo(Path->Binding->Socket, Offloads, 2))) {
```
nibanks, line 1696: Did you see a case where there was a NULL binding? I _think_ it should always be valid here.

src/core/path.c
```diff
@@ -33,6 +33,10 @@ QuicPathInitialize(
     Path->RttVariance = Path->SmoothedRtt / 2;
     Path->EcnValidationState =
         Connection->Settings.EcnEnabled ? ECN_VALIDATION_TESTING : ECN_VALIDATION_FAILED;
+    if (Connection->Settings.EncryptionOffloadAllowed) {
+        // TODO: query QEO capability
+        Path->TlsOffloadSecrets = calloc(1, sizeof(QUIC_TLS_OFFLOAD_SECRETS));
```
ami-GS, line 38: @nibanks  What tag should be used here? (calloc doesn't work for kernel mode)

ami-GS: or allocate on stack?

src/core/crypto.c
```diff
@@ -1647,12 +1652,55 @@ QuicCryptoProcessTlsCompletion(
         }
         Connection->Stats.ResumptionSucceeded = Crypto->TlsState.SessionResumed;
 
+        CXPLAT_DBG_ASSERT(Connection->PathsCount == 1);
+        QUIC_PATH* Path = &Connection->Paths[0];
+
+        if (Path->IsActive && Connection->Settings.IsSet.EncryptionOffloadAllowed)
+        {
```
nibanks, line 1659:
```suggestion
        if (Path->IsActive && Connection->Settings.IsSet.EncryptionOffloadAllowed) {
```

src/core/crypto.c
```diff
@@ -1647,12 +1652,55 @@ QuicCryptoProcessTlsCompletion(
...
+            Offloads[1].ConnectionIdLength = SourceCid->CID.Length;
+            memcpy(&Offloads[1].Address, &Path->Route.LocalAddress, sizeof(QUIC_ADDR));
+            memcpy(Offloads[1].ConnectionId, SourceCid->CID.Data, SourceCid->CID.Length);
+            memcpy(Offloads[1].PayloadIv, Path->TlsOffloadSecrets->Rx.PayloadIv, Path->TlsOffloadSecrets->Rx.PayloadIvLength);
+            memcpy(Offloads[1].PayloadKey, Path->TlsOffloadSecrets->Rx.PayloadKey, Path->TlsOffloadSecrets->Rx.PayloadKeyLength);
+            memcpy(Offloads[1].HeaderKey, Path->TlsOffloadSecrets->Rx.HeaderKey, Path->TlsOffloadSecrets->Rx.HeaderKeyLength);
+            if (QUIC_SUCCEEDED(CxPlatSocketUpdateQeo(Path->Binding->Socket, Offloads, 2))) {
+                Connection->Stats.EncryptionOffloaded = TRUE;
+                Path->EncryptionOffloading = TRUE;
+            }
```
nibanks, line 1697: We should also log on success.

src/inc/quic_tls.h
nibanks: We discussed over IM that you make a new function that takes a packet key and then generates the key info you need to offload:
```c
QUIC_STATUS
QuicPacketKeyCreateOffload(
    _In_ QUIC_PACKET_KEY *Key,
    _In_ const QUIC_HKDF_LABELS* HkdfLabels,
    _In_z_ const char* const SecretName,
    _Out_ CXPLAT_QEO_CONNECTION* Qeo
    )
```
With this model, you don't need to save anything else.

src/core/connection.c
```diff
@@ -482,6 +482,28 @@ QuicConnUninitialize(
...
+            memcpy(Offloads[0].ConnectionId, Connection->Paths[0].DestCid->CID.Data, Connection->Paths[0].DestCid->CID.Length);
+            memcpy(Offloads[1].ConnectionId, SourceCid->CID.Data, SourceCid->CID.Length);
+            QUIC_STATUS Status = CxPlatSocketUpdateQeo(Connection->Paths[0].Binding->Socket, Offloads, 2);
+            Connection->Stats.EncryptionOffloaded = FALSE;
+            Connection->Paths[0].EncryptionOffloading = FALSE;
+            QuicTraceLogConnInfo(
+                OfflodingStop,
+                Connection,
+                "Path[%hhu] %s stop encryption offloading",
+                Connection->Paths[0].ID, QUIC_SUCCEEDED(Status) ? "Successfully" : "Failed to");
```
nibanks, line 504:
```suggestion
            QuicTraceLogConnInfo(
                PathQeoDisabled,
                Connection,
                "Path[%hhu] QEO disabled",
                Connection->Paths[0].ID);
```
Let's keep this simple. I don't care about the status.

src/core/crypto.c
```diff
@@ -1647,12 +1652,59 @@ QuicCryptoProcessTlsCompletion(
...
+            memcpy(Offloads[1].PayloadKey, Path->TlsOffloadSecrets->Rx.PayloadKey, Path->TlsOffloadSecrets->Rx.PayloadKeyLength);
+            memcpy(Offloads[1].HeaderKey, Path->TlsOffloadSecrets->Rx.HeaderKey, Path->TlsOffloadSecrets->Rx.HeaderKeyLength);
+            if (QUIC_SUCCEEDED(CxPlatSocketUpdateQeo(Path->Binding->Socket, Offloads, 2))) {
+                Connection->Stats.EncryptionOffloaded = TRUE;
+                Path->EncryptionOffloading = TRUE;
+                QuicTraceLogConnInfo(
+                    OfflodingStart,
+                    Connection,
+                    "Path[%hhu] Successfully start encryption offloading",
+                    Path->ID);
```
nibanks, line 1700:
```suggestion
                QuicTraceLogConnInfo(
                    PathQeoEnabled,
                    Connection,
                    "Path[%hhu] QEO enabled",
                    Path->ID);
```

src/core/settings.c
```diff
@@ -144,6 +144,10 @@ QuicSettingsSetDefault(
     if (!Settings->IsSet.HyStartEnabled) {
         Settings->HyStartEnabled = QUIC_DEFAULT_HYSTART_ENABLED;
     }
+    if (!Settings->IsSet.EncryptionOffloadAllowed) {
+        Settings->EncryptionOffloadAllowed = QUIC_DEFAULT_ENCRYPTION_OFFLOAD_ALLOWED;
+    }
+
```
nibanks, line 150: nit: remove extra newline

src/inc/quic_tls.h
ami-GS: As I originally wrote as in https://github.com/microsoft/msquic/blob/b1a6ecf5615f01c0dbd588af0802c101060343bc/src/platform/tls_schannel.c#L3247-L3334 
It requires at least `TlsContext->Workspace.OutSecBuffers[i]` to regenerate keys and it doesn't exist on openssl implementation. So I newly add "cache" which stores keys when it is generated during handshake.
Is there any way to retrieve keys from `QUIC_PACKET_KEY`?

ami-GS: Temporally generated keys in QuicPacketKeyDerive is used to create key objects and immediately discarded. As far as I checked there is no invert operation to retrieve raw keys from the key objects.

src/core/crypto.c
```diff
@@ -1647,12 +1647,55 @@ QuicCryptoProcessTlsCompletion(
...
+
+        if (Path->IsActive && Connection->Settings.IsSet.EncryptionOffloadAllowed) {
+            QUIC_CID_HASH_ENTRY* SourceCid =
+                CXPLAT_CONTAINING_RECORD(Connection->SourceCids.Next, QUIC_CID_HASH_ENTRY, Link);
+            CXPLAT_QEO_CONNECTION Offloads[] = {
+                {
+                    CXPLAT_QEO_OPERATION_ADD,
+                    CXPLAT_QEO_DIRECTION_TRANSMIT,
+                    CXPLAT_QEO_DECRYPT_FAILURE_ACTION_DROP,
+                    0, // KeyPhase
```
nibanks, line 1661: FYI, `Connection->Packets[QUIC_ENCRYPT_LEVEL_1_RTT].CurrentKeyPhase` is what you want to put here. Not _super_ necessary for the TLS completion callback, but for the case where new CIDs or keys are generated and used in the future, it could be more important.

src/core/crypto.c
```diff
@@ -1647,12 +1647,55 @@ QuicCryptoProcessTlsCompletion(
...
+                    Connection->Send.NextPacketNumber,
+                },
+                {
+                    CXPLAT_QEO_OPERATION_ADD,
+                    CXPLAT_QEO_DIRECTION_RECEIVE,
+                    CXPLAT_QEO_DECRYPT_FAILURE_ACTION_DROP,
+                    0, // KeyPhase
+                    0, // Reserved:0
+                    CXPLAT_QEO_CIPHER_TYPE_AEAD_AES_256_GCM,
+                    0,
```
nibanks, line 1673: `Connection->Packets[QUIC_ENCRYPT_LEVEL_1_RTT].AckTracker.LargestPacketNumberAcknowledged` should be used here.

src/core/crypto.c
```diff
@@ -1647,12 +1647,55 @@ QuicCryptoProcessTlsCompletion(
         }
         Connection->Stats.ResumptionSucceeded = Crypto->TlsState.SessionResumed;
 
+        CXPLAT_DBG_ASSERT(Connection->PathsCount == 1);
+        QUIC_PATH* Path = &Connection->Paths[0];
+
+        if (Path->IsActive && Connection->Settings.IsSet.EncryptionOffloadAllowed) {
+            QUIC_CID_HASH_ENTRY* SourceCid =
+                CXPLAT_CONTAINING_RECORD(Connection->SourceCids.Next, QUIC_CID_HASH_ENTRY, Link);
+            CXPLAT_QEO_CONNECTION Offloads[] = {
```
nibanks, line 1656: We should `CxPlatSecureZeroMemory` these structs on scope exit so the key info you generate below doesn't stay in memory.

src/core/path.h
```diff
@@ -173,6 +173,11 @@ typedef struct QUIC_PATH {
     //
     uint32_t PathValidationStartTime;
 
+    //
+    // Indicates whether this connection offloads encryption workload to HW
+    //
+    BOOLEAN EncryptionOffloading : 1;
```
nibanks, line 179: Put this up in with the other bitfields.

src/core/crypto.c
ami-GS: https://github.com/microsoft/quic-offloads/blob/main/quic-encryption-offload.md#nextpacketnumber
This says zero?

src/inc/quic_crypt.h
```diff
@@ -231,6 +231,16 @@ QuicPacketKeyDerive(
     _Out_ QUIC_PACKET_KEY **NewKey
     );
 
+#include "quic_datapath.h"
```
nibanks, line 234: Instead of including this here, just put a declaration for `CXPLAT_QEO_CONNECTION` at the top of the file.

src/core/crypto.h
```diff
@@ -439,6 +439,18 @@ QuicCryptoReNegotiateAlpn(
         const uint8_t* AlpnList
     );
 
+//
+// Fill keys for offloading
+//
+_Success_(return==TRUE)
+BOOLEAN
+QuicPacketKeyCreateOffload(
```
nibanks, line 447: This should be declared in quic_tls.h instead and probably called `QuicTlsPopulateKeyOffload` or something similar.

src/core/crypto.c
nibanks: It says _generally_, but what I have above is more accurate. Eventually, we will need to refactor this code to happen for other CIDs and keys that get created long after the handshake, and passing this in will always work.

src/inc/quic_crypt.h
ami-GS: Thanks for good catch, forgot to remove from prototyping

nibanks (approved):

PR #3618, opened 2023-05-07
Connection UDP Receive Event

nibanks:
## Description

Adds a new event to analyze how big of batches the QUIC worker thread processes on the receive path. This allows us to compare the batch size to the batch size of what UDP layer queues.

## Testing

Validated by analyzing log files from perf runs.

## Documentation

N/A


src/manifest/MsQuicEtw.man
```diff
@@ -2134,6 +2134,20 @@
...
+                name="Connection"
+                />
+            <data
+                inType="win:UInt32"
+                name="Count"
+                />
+            <data
+                inType="win:UInt32"
+                name="Count2"
+                />
```
csujedihy, line 2149: Are we intentionally being vague here?

nibanks: Not specifically, but I don't think it matters too much either way.

csujedihy (approved):

PR #3624, opened 2023-05-12
Prevent Path Removal Underflow

nibanks:
## Description

Add check for invalid path removal case. It shouldn't ever happen, but from some crashes I've seen, I suspect it might be happening.

## Testing

Existing automation

## Documentation

N/A

### DIFF
--- a/src/core/path.c
+++ b/src/core/path.c
@@ -53,7 +53,12 @@ QuicPathRemove(
     _In_ uint8_t Index
     )
 {
-    CXPLAT_DBG_ASSERT(Index < Connection->PathsCount);
+    CXPLAT_DBG_ASSERT(Connection->PathsCount > 0);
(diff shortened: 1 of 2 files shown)


csujedihy (approved):

src/core/path.c
```diff
@@ -53,8 +54,15 @@ QuicPathRemove(
     _In_ uint8_t Index
     )
 {
-    CXPLAT_DBG_ASSERT(Index < Connection->PathsCount);
+    CXPLAT_DBG_ASSERT(Connection->PathsCount > 0);
+    CXPLAT_DBG_ASSERT(Connection->PathsCount <= QUIC_MAX_PATH_COUNT);
```
csujedihy, line 58: why don't we make this a teleassert too?

nibanks: We _could_ but I really don't expect this assert to be invalid. I am not so sure about the next one, that's why I've made it a tel assert.

PR #3630, opened 2023-05-15
skip SW encryption on Tx path when QEO enabled

ami-GS:
## Description

Skip SW encryption on TX path when QEO enabled

## Testing

No actual feature yet

## Documentation

N/A

### DIFF
--- a/src/core/packet_builder.c
+++ b/src/core/packet_builder.c
@@ -789,7 +789,8 @@ QuicPacketBuilderFinalize(
             Builder->HeaderLength);
     }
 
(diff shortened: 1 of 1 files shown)


nibanks:
> `CxPlatEncrypt` is called not only from `packet_builder.c:QuicPacketBuilderFinalize`, but also from `binding.c:QuicBindingProcessStatelessOperation` and `packet.c:QuicPacketGenerateRetryIntegrity`. However not sure these are for payload of short header

QEO won't apply to stateless operations and retry packets, so you can safely ignore those.

ami-GS:
Thanks!

PR #3632, opened 2023-05-16
Plumb QEO to XDP

nibanks:
## Description

Wire up the initial plumbing to QEO in XDP.

## Testing

Must wait for real HW or official tests.

## Documentation

N/A

### DIFF
--- a/src/platform/datapath_raw_xdp.c
+++ b/src/platform/datapath_raw_xdp.c
@@ -31,6 +31,7 @@ Abstract:
 #define RX_BUFFER_TAG 'RpdX' // XdpR
 #define TX_BUFFER_TAG 'TpdX' // XdpT
 #define PORT_SET_TAG  'PpdX' // XdpP
+#define OFFLOAD_TAG   'OpdX' // XdpO
 
 typedef struct XDP_INTERFACE XDP_INTERFACE;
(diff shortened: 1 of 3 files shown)


src/platform/datapath_raw_xdp.c
```diff
@@ -1279,6 +1294,75 @@ CxPlatDpRawUninitialize(
...
+        memcpy(Connections[i].PayloadIv, Offloads[i].PayloadIv, sizeof(Connections[i].PayloadIv));
+        Connections[i].Status = 0;
+    }
+
+    //
+    // The following logic just tries all interfaces and if it's able to offload
+    // to any of them, it considers it a success. Long term though, this should
+    // only offload to the interface that the socket is bound to.
+    //
+
```
ami-GS, line 1344: If we allow enabling some of interfaces, we need to know which interfaces are offloading? to disable the offloading ones later

ami-GS: hmm, but not sure the operation is _ADD or _REMOVE...
Code could be really complex then

ami-GS: yeah, offloading functionality point of view, it should be OK for now

ami-GS (approved):

PR #3637, opened 2023-05-17
Shutdown Stream on Send/Start Failure

### LINKED ISSUE #3638: StreamSend(QUIC_SEND_FLAG_START | QUIC_SEND_FLAG_FIN) returns QUIC_SUCCEEDED but QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE never gets delivered?
### Describe the bug

My code relies on the `exactly-once` semantics of `QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE` to ensure proper releasing of critical resources associated with the stream.

The logic is as below:

```cpp
if (QUIC_FAILED(StreamSend(QUIC_SEND_FLAG_START | QUIC_SEND_FLAG_FIN)))
{
    // clean up resource immediately as stream will never be started
    return <some error code>;
}
```

```cpp
static QUIC_STATUS QUIC_API
ClientStreamCallback(
      _In_ HQUIC Stream,
      _In_opt_ void* Context,
      _Inout_ QUIC_STREAM_EVENT* Event
      )
{
case QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE: // guaranteed to fire exactly once
    // clean up resource
    break;
```

But in one instance my program had a behavior that seems only explainable if the clean up logic never occurred, which makes me wonder whether it's possible that StreamSend(QUIC_SEND_FLAG_START | QUIC_SEND_FLAG_FIN) returned QUIC_SUCCEEDED, and yet `QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE` never gets indicated to the stream callback later on, which would violate the exactly-once assumption I relied on.

### Affected OS

- [X] Windows
- [ ] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

_No response_

### MsQuic version

main

1. write a test program with above logic
2. turn on alloc failure enabled mode
3. note that it doesn't seem to happen easily

### Expected behavior

If StreamSend returns QUIC_SUCCEEDED, `QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE` must be later indicated to the stream callback, because otherwise there might not be any way to logically ensure a proper resource management.

### Actual outcome

resource clean up logic may not be executed under rare circumstances

### Additional details

_No response_

nibanks:
## Description

Fixes #3638 and #3639. In the rare case (OOM) that stream start fails when you use the `QUIC_SEND_FLAG_START` flag, it should be delivering a shutdown complete event to indicate this failure to start. Also fixes more event issues in the case of OOM in the send flush operation alloc path.

## Testing

Existing automation and manual testing from original bug opener.

## Documentation

N/A

### DIFF
--- a/src/core/api.c
+++ b/src/core/api.c
@@ -1141,7 +1141,7 @@ MsQuicStreamSend(
             Oper->API_CALL.Context = &Connection->BackupApiContext;
             Oper->API_CALL.Context->Type = QUIC_API_TYPE_CONN_SHUTDOWN;
             Oper->API_CALL.Context->CONN_SHUTDOWN.Flags = QUIC_CONNECTION_SHUTDOWN_FLAG_SILENT;
-            Oper->API_CALL.Context->CONN_SHUTDOWN.ErrorCode = QUIC_STATUS_OUT_OF_MEMORY;
+            Oper->API_CALL.Context->CONN_SHUTDOWN.ErrorCode = (QUIC_VAR_INT)QUIC_STATUS_OUT_OF_MEMORY;
             Oper->API_CALL.Context->CONN_SHUTDOWN.RegistrationShutdown = FALSE;
             Oper->API_CALL.Context->CONN_SHUTDOWN.TransportShutdown = TRUE;
             QuicConnQueueOper(Connection, Oper);
--- a/src/core/api.c
+++ b/src/core/api.c
@@ -1039,11 +1039,6 @@ MsQuicStreamSend(
         (Connection->WorkerThreadID == CxPlatCurThreadID()) ||
         !Connection->State.HandleClosed);
 
-    if (Connection->State.ClosedRemotely) {
-        Status = QUIC_STATUS_ABORTED;
-        goto Exit;
-    }
-
     TotalLength = 0;
     for (uint32_t i = 0; i < BufferCount; ++i) {
         TotalLength += Buffers[i].Length;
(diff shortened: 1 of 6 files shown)


src/core/api.c
```diff
@@ -1125,12 +1120,29 @@ MsQuicStreamSend(
...
+                    (short*)&Connection->BackUpOperUsed, 1, 0) != 0) {
+                goto Exit; // It's already started the shutdown.
+            }
+            Oper = &Connection->BackUpOper;
+            Oper->FreeAfterProcess = FALSE;
+            Oper->Type = QUIC_OPER_TYPE_API_CALL;
+            Oper->API_CALL.Context = &Connection->BackupApiContext;
+            Oper->API_CALL.Context->Type = QUIC_API_TYPE_CONN_SHUTDOWN;
+            Oper->API_CALL.Context->CONN_SHUTDOWN.Flags = QUIC_CONNECTION_SHUTDOWN_FLAG_SILENT;
+            Oper->API_CALL.Context->CONN_SHUTDOWN.ErrorCode = 0;
```
csujedihy, line 1143: Will the error code be sent to the peer? I wonder if we should fill in the OOM status code so the app knows what killed it.

nibanks: Currently, with how this is written, the app just gets a shutdown complete event, no error code. It would take a bigger change to fix that. I'll look into it tomorrow.

nibanks: Yeah, not ideal. I will have to refactor some stuff to make this work better.

nibanks: Fixed.

csujedihy (dismissed):

csujedihy:
there's a build error. Finding the error in the github action is like finding a needle in a haystack. The console coloring is all gone.

```
/home/runner/work/msquic/msquic/src/core/connection.c:1873:9: error: too few arguments to function ‘QuicConnShutdown’
 1873 |         QuicConnShutdown(Connection, ShutdownFlags, ShutdownErrorCode, FALSE);
      |         ^~~~~~~~~~~~~~~~
/home/runner/work/msquic/msquic/src/core/connection.c:437:1: note: declared here
  437 | QuicConnShutdown(
      | ^~~~~~~~~~~~~~~~
```

src/core/api.c
pongba, line 1: May need to add 3 unit test cases to cover the following:

- If StreamSend returns SUCCESS, `SEND_COMPLETE` must be indicated exactly-once
- If StreamSend returns FAIL, `SEND_COMPLETE` must not be indicated.
- If StreamSend(`QUIC_SEND_FLAG_START`) returns SUCCESS, `STREAM_SHUTDOWN_COMPLETE` must be indicated exactly-once.

nibanks:
Feel free to push a fix, otherwise I will tomorrow.

csujedihy (approved):

PR #3641, opened 2023-05-19
Build Processor Info for Max Processors

nibanks:
## Description

We're incorrectly building our processor mapping tables for only the active processor count, but then using max to look up in the table.

## Testing

Existing + manual (WIP)

## Documentation

N/A


nibanks:
[secnetperf.exe.txt](https://github.com/microsoft/msquic/files/11520160/secnetperf.exe.txt)

csujedihy (approved):

PR #3647, opened 2023-05-23
Fix transport error code for new alpn negotiation

liveans:
## Description

Currently, while we're narrowing down the ALPN list with ReNegotiation, if we can't negotiate with client on ALPN, we're closing the connection with `QUIC_ERROR_INTERNAL_ERROR`, but looks like it should be `QUIC_ERROR_CRYPTO_NO_APPLICATION_PROTOCOL`.

## Testing

I've updated the necessary unit test, but this is a breaking change since we're changing close status behavior.

## Documentation

No

### DIFF
--- a/src/core/crypto.c
+++ b/src/core/crypto.c
@@ -2697,7 +2697,7 @@ QuicCryptoReNegotiateAlpn(
             "No ALPN match found");
         QuicConnTransportError(
             Connection,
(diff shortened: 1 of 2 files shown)


ManickaP:
This should get into whatever msquic version that will be actual when .NET 8 ships. I.e.: backport to 2.2 unless another release from main will be made.

nibanks (approved):

PR #3651, opened 2023-05-25
Attaching XDP to VF

ami-GS:
## Description

If XDP runs on VM with SR-IOV, trying to attach XDP to VF

## Testing

Send/Recv is failing

## Documentation

N/A

### DIFF
--- a/src/platform/datapath_raw.h
+++ b/src/platform/datapath_raw.h
@@ -78,6 +78,7 @@ typedef struct CXPLAT_DATAPATH {
 typedef struct CXPLAT_INTERFACE {
     CXPLAT_LIST_ENTRY Link;
     uint32_t IfIndex;
+    uint32_t ActualIfIndex;
     UCHAR PhysicalAddress[ETH_MAC_ADDR_LEN];
     struct {
         struct {
--- a/src/platform/datapath_raw_socket.c
+++ b/src/platform/datapath_raw_socket.c
@@ -570,7 +570,7 @@ CxPlatResolveRoute(
     CXPLAT_LIST_ENTRY* Entry = Socket->Datapath->Interfaces.Flink;
     for (; Entry != &Socket->Datapath->Interfaces; Entry = Entry->Flink) {
         CXPLAT_INTERFACE* Interface = CONTAINING_RECORD(Entry, CXPLAT_INTERFACE, Link);
-        if (Interface->ActualIfIndex == IpforwardRow.InterfaceIndex) {
+        if (Interface->IfIndex == IpforwardRow.InterfaceIndex) {
             CXPLAT_DBG_ASSERT(sizeof(Interface->PhysicalAddress) == sizeof(Route->LocalLinkLayerAddress));
             CxPlatCopyMemory(&Route->LocalLinkLayerAddress, Interface->PhysicalAddress, sizeof(Route->LocalLinkLayerAddress));
             CxPlatDpRawAssignQueue(Interface, Route);
--- a/src/platform/datapath_raw_socket.c
+++ b/src/platform/datapath_raw_socket.c
@@ -570,7 +570,7 @@ CxPlatResolveRoute(
     CXPLAT_LIST_ENTRY* Entry = Socket->Datapath->Interfaces.Flink;
     for (; Entry != &Socket->Datapath->Interfaces; Entry = Entry->Flink) {
         CXPLAT_INTERFACE* Interface = CONTAINING_RECORD(Entry, CXPLAT_INTERFACE, Link);
-        if (Interface->IfIndex == IpforwardRow.InterfaceIndex) {
+        if (Interface->ActualIfIndex == IpforwardRow.InterfaceIndex) {
             CXPLAT_DBG_ASSERT(sizeof(Interface->PhysicalAddress) == sizeof(Route->LocalLinkLayerAddress));
             CxPlatCopyMemory(&Route->LocalLinkLayerAddress, Interface->PhysicalAddress, sizeof(Route->LocalLinkLayerAddress));
             CxPlatDpRawAssignQueue(Interface, Route);
--- a/src/platform/datapath_raw_xdp.c
+++ b/src/platform/datapath_raw_xdp.c
@@ -1104,7 +1104,7 @@ CxPlatDpRawInitialize(
             goto Error;
         }
         CxPlatZeroMemory(Interface, sizeof(*Interface));
-
+        Interface->IfIndex = Interface->ActualIfIndex = pIfRow->InterfaceIndex;
         // NOTE: O(n^2), but realistically enough small
         for (ULONG j = i+1; j < pIfTable->NumEntries; j++) {
             MIB_IF_ROW2* pIfRowNext = &pIfTable->Table[j];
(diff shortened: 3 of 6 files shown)


src/platform/datapath_raw_xdp.c
```diff
@@ -1077,88 +1077,93 @@ CxPlatDpRawInitialize(
...
+        Status = QUIC_STATUS_INTERNAL_ERROR;
+        goto Error;
+    }
+
+    struct IfMatching {
+        MIB_IF_ROW2* IfRow;
+        MIB_IF_ROW2* ActualIfRow; // for VF
+    };
+
+    struct IfMatching IfMatching[16] = {0}; // TODO: linked list
```
nibanks, line 1091: So, instead of first building a list of objects, and then going and allocating matching XDP objects for each, why not do it all together? For instance, keep the old loop we had, where we create an XDP object for each IP address/interface; but then inside that loop, go try to find a "matching" interface for the "actual" interface index. If found, you update the actual index.

src/platform/datapath_raw.h
```diff
@@ -78,6 +78,7 @@ typedef struct CXPLAT_DATAPATH {
 typedef struct CXPLAT_INTERFACE {
     CXPLAT_LIST_ENTRY Link;
     uint32_t IfIndex;
+    uint32_t ActualIfIndex;
```
mtfriesen, line 81: IMO we should do this the other way around: we should track each interface's `IfIndex` and also an optional `VfIfIndex`. We should then not have any VFs directly in the interface table, and only explicitly VF-aware code needs to use the `VfIfIndex`.

mtfriesen: And if we can't find the VF for an interface, we could just set VfIfIndex to the regular IfIndex value, or some magic value like `IFI_UNSPECIFIED`.

src/platform/datapath_raw_xdp.c
```diff
@@ -1077,88 +1077,93 @@ CxPlatDpRawInitialize(
...
+    int IfRowsCount = 0;
+    for (int i = 0; i < (int) pIfTable->NumEntries; i++) {
+        MIB_IF_ROW2* pIfRow = &pIfTable->Table[i];
+        if (pIfRow->InterfaceAndOperStatusFlags.FilterInterface ||
+            !pIfRow->InterfaceAndOperStatusFlags.HardwareInterface ||
```
mtfriesen, line 1096: Do we need the `HardwareInterface` or `ConnectorPresent` checks?

src/platform/datapath_raw_xdp.c
```diff
@@ -1077,88 +1077,93 @@ CxPlatDpRawInitialize(
...
+            !pIfRow->InterfaceAndOperStatusFlags.ConnectorPresent) {
+            continue;
+        }
+        // NOTE: O(n^2), but realistically enough small
+        for (int j = 0; j < IfRowsCount; j++) {
+            struct IfMatching* pMatching = &IfMatching[j];
+            if (pIfRow->PhysicalMediumType == NdisPhysicalMedium802_3 &&
```
mtfriesen, line 1103: Do we need to look at the physical mediums? This strikes me as being potentially very fragile.

ami-GS:
Oh, test issue was only on my dev system. something corrupted in my side

src/platform/datapath_raw_xdp.c
ami-GS: Empirically these three flags are useful for filter out useless interfaces. Not sure other corner case

src/platform/datapath_raw_xdp.c
ami-GS: This is also empirical point of view. However it should be impossible to identify which is VF if we don't use the PhysicalMediumType

src/platform/datapath_raw.h
ami-GS: Currently I'm not sure where the conventional IfIndex should be used and so use ActualIfIndex everywhere. 
If we use VfIfIndex as VF-aware code, there whould require if-else where code refers IfIndex

src/platform/datapath_raw_xdp.c
```diff
@@ -1077,88 +1077,74 @@ CxPlatDpRawInitialize(
...
-        Adapters = (IP_ADAPTER_ADDRESSES*)CxPlatAlloc(AdaptersBufferSize, ADAPTER_TAG);
-        if (Adapters == NULL) {
+    PMIB_IF_TABLE2 pIfTable;
+    if (GetIfTable2(&pIfTable) != NO_ERROR) {
+        Status = QUIC_STATUS_INTERNAL_ERROR;
+        goto Error;
+    }
+
+    CXPLAT_FRE_ASSERT(pIfTable->NumEntries <= 128);
+    BOOLEAN Initialized[128] = {0};
```
nibanks, line 1087: I'm a bit unsure, but do we _really_ need this array? I don't like the hard coded limit, but it seems more like an optimization that's not really necessary.

ami-GS: I want to use std::bitset or heap allocated array. It is possible to remove this by checking Xdp->Interfaces linked list at beginning of outer loop whether there is adapter with same Mac address is already registered.

x is same mac address (VF and NetVSC) merged as one interface, then later
```
outer  |--->x             |
inner  |      ----->x     |
```

```
outer  |      -----o-->     |
inner  |                    |
```
o is skipped because it was marked as Initialized earlier

src/platform/datapath_raw_xdp.c
```diff
@@ -1077,88 +1077,74 @@ CxPlatDpRawInitialize(
...
+    for (int i = 0; i < (int) pIfTable->NumEntries; i++) {
```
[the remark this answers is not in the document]
ami-GS: I thought it is covered by first flag check.
I was not 100% sure whether these flags is for whether having IP address or not. just empirically verified this check works  (duonic and NetVSC and VF can pass through)

```c
        if (pIfRow->InterfaceAndOperStatusFlags.FilterInterface ||
            !pIfRow->InterfaceAndOperStatusFlags.HardwareInterface ||
            !pIfRow->InterfaceAndOperStatusFlags.ConnectorPresent ||
            Initialized[i])
```

```c
            if (Adapter->IfType == IF_TYPE_ETHERNET_CSMACD &&
                Adapter->OperStatus == IfOperStatusUp &&
                Adapter->PhysicalAddressLength == ETH_MAC_ADDR_LEN)
```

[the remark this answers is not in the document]
ami-GS: You meant use GetAdaptersAddresses as outer loop?

src/platform/datapath_raw_xdp.c
nibanks: I'm still not sure why we need to keep this at all? It's not a big deal if you search the whole array N^2, since it's so small.

src/platform/datapath_raw_xdp.c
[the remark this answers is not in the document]
nibanks: Yes, with GAA as outer loop. But if you really think based on your testing this is the way to go, feel free to "won't fix" this comment.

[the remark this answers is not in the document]
ami-GS: I think the idea works. let me quickly push it.

src/platform/datapath_raw_xdp.c
ami-GS: Can be removed by using GAA for outer loop

src/platform/datapath_raw_socket.c
```diff
@@ -570,7 +570,7 @@ CxPlatResolveRoute(
     CXPLAT_LIST_ENTRY* Entry = Socket->Datapath->Interfaces.Flink;
     for (; Entry != &Socket->Datapath->Interfaces; Entry = Entry->Flink) {
         CXPLAT_INTERFACE* Interface = CONTAINING_RECORD(Entry, CXPLAT_INTERFACE, Link);
-        if (Interface->IfIndex == IpforwardRow.InterfaceIndex) {
+        if (Interface->ActualIfIndex == IpforwardRow.InterfaceIndex) {
```
nibanks, line 573: Are you sure this shouldn't be the normal IfIndex? I think it should.

ami-GS: I thought this is actual one (VF) because it is copying mac address in the if block.

ami-GS: What does the GetBestRoute2 scope? only in VM or entire system?

ami-GS: the scope seems to be in OS (VM), IfIndex should be correct

nibanks: I'm not sure of the details, but this is getting info from TCP/IP which works on top of the `IfIndex`, not the actual one, so I think this should align with TCP/IP's view.

nibanks (approved):

PR #3652, opened 2023-05-26
Use PMC v4 API for publishing packages

csujedihy:
## Description

Migrate to PMC v4 API.

## Testing

OB CI


csujedihy:
If https://mscodehub.visualstudio.com/msquic/_build/results?buildId=372436&view=logs&j=d0720056-9b4b-5135-a196-34587a82b73e&t=ecd6283f-e101-50cd-e2b5-e334fef409bb passes, we can check it in.

nibanks (approved):

nibanks:
It seems one of the operations was canceled. Was that on purpose?
```
##[error]The operation was canceled.
```

nibanks:
> It seems one of the operations was canceled. Was that on purpose?
> 
> ```
> ##[error]The operation was canceled.
> ```

Ah, it seems like the total time was > 60 minutes. Should we use a `matrix` here to parallelize these instead?

.azure/OneBranch.Publish.yml
```diff
@@ -78,17 +89,25 @@ stages:
...
+        name:  pmcv4cert
+        displayName: 'Download cert for PMC v4'
+        inputs:
+          secureFile: 'auth.pem'
       - task: Docker@2
         displayName: Login to ACR
         inputs:
           command: login
           containerRegistry: msquicdockerregistry
       - ${{ each repo in parameters.opensslrpmrepos }}:
```
nibanks, line 102: Instead of a for each, can we do a matrix strategy and parallelize?

csujedihy: In the internal repo, I am using 4 jobs to make them run in parallel.

nibanks (approved):

PR #3665, opened 2023-06-01
Support Delaying Stream ID FC Updates to StreamClose

nibanks:
## Description

Sometimes apps might want to delay the stream ID FC updates to StreamClose if they intend on doing significant work between SHUTDOWN_COMPLETE and the close call.

## Testing

Updated spinquic.

## Documentation

Updated.

### DIFF
--- a/docs/api/QUIC_CONNECTION_EVENT.md
+++ b/docs/api/QUIC_CONNECTION_EVENT.md
@@ -176,6 +176,8 @@ Value | Meaning
 **QUIC_STREAM_OPEN_FLAG_UNIDIRECTIONAL**<br>1 | A unidirectional stream.
 **QUIC_STREAM_OPEN_FLAG_0_RTT**<br>2 | The stream was received in 0-RTT.
 
+If a server wishes to use `QUIC_STREAM_OPEN_FLAG_DELAY_FC_UPDATES` for the newly started stream, it may append this flag to `Flags` before it returns from the callback.
+
 ## QUIC_CONNECTION_EVENT_STREAMS_AVAILABLE
 
 This event indicates the number of streams the peer is willing to accept has changed.
--- a/docs/api/QUIC_CONNECTION_EVENT.md
+++ b/docs/api/QUIC_CONNECTION_EVENT.md
@@ -176,7 +176,7 @@ Value | Meaning
 **QUIC_STREAM_OPEN_FLAG_UNIDIRECTIONAL**<br>1 | A unidirectional stream.
 **QUIC_STREAM_OPEN_FLAG_0_RTT**<br>2 | The stream was received in 0-RTT.
 
-If a server wishes to use `QUIC_STREAM_OPEN_FLAG_DELAY_FC_UPDATES` for the newly started stream, it may append this flag to `Flags` before it returns from the callback.
+If a server wishes to use `QUIC_STREAM_OPEN_FLAG_DELAY_ID_FC_UPDATES` for the newly started stream, it may append this flag to `Flags` before it returns from the callback.
 
 ## QUIC_CONNECTION_EVENT_STREAMS_AVAILABLE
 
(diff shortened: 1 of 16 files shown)


mtfriesen (dismissed):

docs/api/QUIC_CONNECTION_EVENT.md
```diff
@@ -176,6 +176,8 @@ Value | Meaning
 **QUIC_STREAM_OPEN_FLAG_UNIDIRECTIONAL**<br>1 | A unidirectional stream.
 **QUIC_STREAM_OPEN_FLAG_0_RTT**<br>2 | The stream was received in 0-RTT.
 
+If a server wishes to use `QUIC_STREAM_OPEN_FLAG_DELAY_FC_UPDATES` for the newly started stream, it may append this flag to `Flags` before it returns from the callback.
```
csujedihy, line 179: Is this flag for stream ID flow control or stream flow control?

csujedihy: The entire PR is for stream ID. I think it should be better renamed to QUIC_STREAM_OPEN_FLAG_DELAY_ID_FC_UPDATES

src/core/stream.c
```diff
@@ -354,7 +360,7 @@ QuicStreamClose(
 
     if (!Stream->Flags.ShutdownComplete) {
 
-        if (Stream->Flags.Started) {
+        if (Stream->Flags.Started && !Stream->Flags.HandleShutdown) {
```
csujedihy, line 363: >&& !Stream->Flags.HandleShutdown [](http://example.com/codeflow?start=34&length=32)

Is this an irrelevant fix for tracing?

nibanks: Yes.

src/manifest/clog.sidecar
```diff
@@ -963,6 +963,18 @@
       ],
       "macroName": "QuicTraceLogInfo"
     },
+    "ConfiguredForDelayedFC": {
+      "ModuleProperites": {},
+      "TraceString": "[strm][%p] Configured for delayed FC updates",
```
csujedihy, line 968: >Configured for delayed FC updates [](http://example.com/codeflow?start=33&length=33)

this is ambiguous too.

src/core/stream.c
```diff
@@ -593,7 +607,9 @@ QuicStreamTryCompleteShutdown(
         //
         // Indicate the stream is completely shut down to the connection.
         //
-        QuicStreamSetReleaseStream(&Stream->Connection->Streams, Stream);
+        if (!Stream->Flags.DelayFCUpdate || Stream->Flags.HandleClosed) {
```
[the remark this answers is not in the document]
csujedihy: If HandleClose == true, this function, QuicStreamTryCompleteShutdown, is called by QuicStreamClose, right? QuicStreamClose guarantees that QuicStreamSetReleaseStream will be called if DelayFCUpdate is set. Wouldn't we end up calling QuicStreamSetReleaseStream twice?

docs/api/QUIC_CONNECTION_EVENT.md
nibanks: Ok, I will update.

src/core/stream.c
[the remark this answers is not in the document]
nibanks: This case is tricky. If an app shuts down (abortively) a stream using the `QUIC_STREAM_SHUTDOWN_FLAG_IMMEDIATE` flag then we will immediately give them the shutdown complete event and they can close the handle, but MsQuic will continue to track the stream until we've delivered (and received the ACK for) the corresponding stream frames over the wire. In this case, the `QuicStreamClose` call will run before this code path, but we don't want to release the stream there. Later, when everything is complete, this code path will be run finally, but since the handle has already been closed, we need to release the stream now.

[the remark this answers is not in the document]
csujedihy: I see.

csujedihy (approved):

PR #3670, opened 2023-06-02
Update Linux Packages to Depend on libnuma

### LINKED ISSUE #3421: msquic depends on libnuma but packages don't state that
### Describe the bug

This is similar to  #2975 and it seems to be new in 2.2
```
$ ldd artifacts/bin/linux/x64_Debug_openssl/libmsquic.so.2.1.7
	linux-vdso.so.1 (0x00007ffd7b1c4000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fc7fa425000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fc7fa402000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fc7fa210000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fc7fa7aa000)
$ ldd artifacts/bin/linux/x64_Debug_openssl/libmsquic.so.2.2.0
	linux-vdso.so.1 (0x00007ffcb51bf000)
	libcrypto.so.1.1 => /lib/x86_64-linux-gnu/libcrypto.so.1.1 (0x00007f0d22fd8000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f0d22fd2000)
	libnuma.so.1 => /lib/x86_64-linux-gnu/libnuma.so.1 (0x00007f0d22fc5000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f0d22fa2000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f0d22db0000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f0d23423000)
```

The consequence is that it
1) can break uses on upgrade as new library needs more than the previous version
2) allows to install msquic package without warning but leave it dysfunctional.

that is unpleasant and difficult for users to diagnose. 

### Affected OS

- [ ] Windows
- [X] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

_No response_

### MsQuic version

main

easiest is to run `ldd` to show library dependencies. 

### Expected behavior

packages properly state dependencies

### Actual outcome

libnuma  is missing

### Additional details

We should either fix packages to pull in libnuma as needed or we should make it soft dependency e.g. optionally use it via function pointers. 

cc: @manickap @CarnaViire

nibanks:
## Description

Fixes #3421.

## Testing

Existing automation covers some of this.

## Documentation

N/A

### DIFF
--- a/scripts/make-packages.sh
+++ b/scripts/make-packages.sh
@@ -146,6 +146,7 @@ if [ "$OS" == "linux" ]; then
     --name ${NAME} \
     --provides ${NAME} \
     --depends "libcrypto.so.${TLSVERSION}()(${BITS})" \
+    --depends "libnuma.so.1()(${BITS})" \
     --conflicts ${CONFLICTS} \
     --version ${VER_MAJOR}.${VER_MINOR}.${VER_PATCH} \
     --description "${DESCRIPTION}" \
--- a/scripts/make-packages.sh
+++ b/scripts/make-packages.sh
@@ -183,6 +183,7 @@ if [ "$OS" == "linux" ]; then
     --provides ${NAME} \
     --conflicts ${CONFLICTS} \
     --depends "libssl${TLSVERSION}" \
+    --depends "libnuma1" \
     --version ${VER_MAJOR}.${VER_MINOR}.${VER_PATCH} \
     --description "${DESCRIPTION}" \
     --vendor "${VENDOR}" \


scripts/make-packages.sh
```diff
@@ -182,6 +183,7 @@ if [ "$OS" == "linux" ]; then
     --provides ${NAME} \
     --conflicts ${CONFLICTS} \
     --depends "libssl${TLSVERSION}" \
+    --depends "libnuma1" \
```
nibanks, line 186: @wfurt I did a combination of guessing and chatgpt to come up with these changes. Would it be possible to get your help validating?

csujedihy:
```
[root@061eeb412ce3 /]# yum provides libnuma.so.1
Last metadata expiration check: 0:04:56 ago on Fri Jun  2 22:12:46 2023.
numactl-libs-2.0.16-2.fc38.i686 : libnuma libraries
Repo        : fedora
Matched from:
Provide    : libnuma.so.1
```

```
root@f45c0dbd52df:~/msquic# apt-cache depends libnuma1
libnuma1
  Depends: libc6
```

Looks good to me at least. Both dependency names can be recognized.

csujedihy (approved):

PR #3676, opened 2023-06-03
Unify CxPlatTryAddSocket for XDP

ami-GS:
## Description

unify CxPlatTryAddSocket for both platform, win, linux

## Testing

CI

## Documentation

N/A


src/inc/quic_platform.h
```diff
@@ -45,6 +45,11 @@ Supported Environments:
 #define CXPLAT_CONTAINING_RECORD(address, type, field) \
     ((type *)((uint8_t*)(address) - offsetof(type, field)))
 
+#define CxPlatSocketError() _CxPlatSocketError()
```
nibanks, line 48: Let's put this stuff at the top of datapath_raw_socket.c instead.

src/platform/datapath_raw_socket.c
```diff
@@ -783,3 +783,328 @@ CxPlatFramingWriteHeaders(
     Buffer->Length += TransportLength + IpHeaderLen + sizeof(ETHERNET_HEADER);
     Buffer->Buffer -= TransportLength + IpHeaderLen + sizeof(ETHERNET_HEADER);
 }
+
+
+QUIC_STATUS
+CxPlatTryAddSocket(
```
nibanks, line 789: This stuff is generally fine, but eventually we're going to eliminate all this code, because we'll merge XDP support as a fast path to the normal (iocp, epoll) datapaths. So don't waste too much time/energe in getting this stuff just right because it will go away.

ami-GS: OK

PR #3693, opened 2023-06-12
Fix UDPv6 checksum

csujedihy:
## Description

UDPv6 requires checksum to be non-zero.

## Testing

CI

### DIFF
--- a/src/platform/datapath_raw_socket.c
+++ b/src/platform/datapath_raw_socket.c
@@ -1379,13 +1379,6 @@ CxPlatFramingWriteHeaders(
                         sizeof(Route->LocalAddress.Ipv4.sin_addr),
                         IPPROTO_TCP,
                         (uint8_t*)TCP, sizeof(TCP_HEADER) + Buffer->Length);
-            } else {
-                UDP->Checksum =
-                    CxPlatFramingTransportChecksum(
-                        IPv4->Source, IPv4->Destination,
-                        sizeof(Route->LocalAddress.Ipv4.sin_addr),
-                        IPPROTO_UDP,
-                        (uint8_t*)UDP, sizeof(UDP_HEADER) + Buffer->Length);
             }
         }
     } else {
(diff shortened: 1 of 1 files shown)


nibanks:
> Also, UDPv4 can be skipped.

Why?

csujedihy:
> > Also, UDPv4 can be skipped.
> 
> 
> 
> Why?
> 
> 
> 
> 

UDPv4 checksum is optional. UDP checksum is required only with IPv6.

nibanks:
> > > Also, UDPv4 can be skipped.
> > 
> > 
> > Why?
> 
> UDPv4 checksum is optional. UDP checksum is required only with IPv6.

I assume you haven't validated this in a real environment (i.e. Azure VM)? I will sign off, and leave it up to you if you're sure enough to merge without further validation.

nibanks (dismissed):

csujedihy:
> > > > Also, UDPv4 can be skipped.
> > > 
> > > 
> > > Why?
> > 
> > 
> > UDPv4 checksum is optional. UDP checksum is required only with IPv6.
> 
> I assume you haven't validated this in a real environment (i.e. Azure VM)? I will sign off, and leave it up to you if you're sure enough to merge without further validation.

Some middleboxes might not like zero UDPv4 checksum. Though, I am pretty sure it works fine over azure with linux and windows. I am gonna revert this change to be safe. We could add a knob to disable UDPv4 checksum in the future even for regular QUIC. We have UDP_NOCHECKSUM for windows UDP sockets to disable it, and UDPv4 checksum is default on.

nibanks (approved):

mtfriesen:
Does this handle the UDPv6 case that we recently fixed in XDP? That is, the UDPv6 checksum must _not_ be zero?

mtfriesen:
> Does this handle the UDPv6 case that we recently fixed in XDP? That is, the UDPv6 checksum must _not_ be zero?

Yes it does. I was looking at the wrong commit diff.

mtfriesen (approved):

PR #3694, opened 2023-06-12
Fix worker event handle leak in MsQuic!RegistrationClose

Techwolfy:
The QUIC_WORKER Worker->Ready event handle is created unconditionally in QuicWorkerInitialize, but only closed `if (!Worker->IsExternal)` in QuicWorkerUninitialize. The handle is leaked in other cases.

This leak was detected by AppVerifier in the HTTP.SYS tests.


nibanks (approved):
Thanks!

PR #3695, opened 2023-06-12
Connections Size, Not Count

nibanks:
## Description

We were passing array size instead of the total byte count.

## Testing

Existing

## Documentation

N/A


csujedihy (approved):

csujedihy:
XdpInitializeQuicConnection is never called in the code?

nibanks:
> XdpInitializeQuicConnection is never called in the code?

You're right. This API went through a couple iterations and we forgot to update this part. Just pushed a fix.

mtfriesen (approved):

PR #3702, opened 2023-06-14
Fix RPS Processor Usage

nibanks:
## Description

Recent changes to how the processor index is calculated broke RPS tests on Windows. This fixes them by taking into account that a given processor index may refer to an inactive processor (since we track maximum, not active, processors). The RPS test code wants to only use active processors, so it iterates over all processors, skipping inactive ones.

## Testing

Automation

## Documentation

N/A


mtfriesen (approved):

csujedihy:
Can we make the PR description clearer? I think I understand now why it fails. It seems that the proc info array contains items for non-active processes.

csujedihy (dismissed):

csujedihy (approved):

csujedihy (approved):

PR #3705, opened 2023-06-15
Split QuicTestConnectionParam to Reduce Stack Usage

nibanks:
## Description

See title. Relevant for arm64 testing.

## Testing

Need to ingest into Windows for arm64 testing.

## Documentation

N/A

### DIFF
--- a/src/test/lib/ApiTest.cpp
+++ b/src/test/lib/ApiTest.cpp
@@ -3482,7 +3482,7 @@ void QuicTest_QUIC_PARAM_CONN_REMOTE_ADDRESS(MsQuicRegistration& Registration, M
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/test/lib/ApiTest.cpp
```diff
@@ -3165,956 +3165,947 @@ void QuicTestListenerParam()
 #endif
 }
 
-void QuicTestConnectionParam()
```
csujedihy, line 3168: Why don't we disable compiler optimization as Ben suggested? We could disable compiler optimization for all of our unittests right?

nibanks: Because I'm pretty sure we don't use the LKG of the compiler they refer to because we're undocked. So, I don't know if it'd do anything.

csujedihy: Alright, fair enough.

csujedihy (approved):

PR #3708, opened 2023-06-16
Experiment /fsanitize=address for tests

csujedihy:
## Description

Just an experiment.

### DIFF
--- a/src/test/bin/CMakeLists.txt
+++ b/src/test/bin/CMakeLists.txt
@@ -22,7 +22,7 @@ target_link_libraries(msquictest inc gtest logging base_link)
 
 # At least /W3 must be used on all windows builds to pass compliance
 if(MSVC)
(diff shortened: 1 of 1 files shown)


src/test/bin/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ target_link_libraries(msquictest inc gtest logging base_link)
 
 # At least /W3 must be used on all windows builds to pass compliance
 if(MSVC)
-    target_compile_options(msquictest PRIVATE /W3 /bigobj)
+    target_compile_options(msquictest PRIVATE /W3 /bigobj /fsanitize=address)
```
nibanks, line 25: I didn't realize tests weren't using the same flags. FYI, there are 3 tests that might need this flag in total.

csujedihy: Build still failed.

PR #3714, opened 2023-06-17
Split build workflow for windows and unix

csujedihy:
## Description

Split build workflow into two separate workflows and try uploading artifacts if applicable.

## Testing

CI


.github/workflows/build-win.yml
```diff
@@ -0,0 +1,151 @@
...
+    - release/*
+  pull_request:
+    branches:
+    - main
+    - release/*
+
+concurrency:
+  # Cancel any workflow currently in progress for the same PR.
+  # Allow running concurrently with any other commits.
+  group: build-${{ github.event.pull_request.number || github.sha }}
```
nibanks, line 17: Need to change this and the other one to be unique, otherwise they will cancel each other.

csujedihy: Yeah, I am aware of that. Fixing a bunch of other stuff now.

csujedihy:
Any idea why the workflows have not run? 

I can't tell why `.github/workflows/bvt.yml` is conflicting either.

PR #3720, opened 2023-06-23
Resue build workflows

csujedihy:
## Description

1. Refactor the build workflow into a resusable workflow. Other workflows that need build artifacts can just call the build workflow. The reusable build workflow is essentially a function that takes various build parameters. This enables us to build and test binaries in different environments and improve how we build stuff in one place.
2. Merge stress and BVT into the same workflow so we build once for everything.

## Testing

OB CI


nibanks (commented):
This seems super complicated compared to what we already have. Is it worth it?

csujedihy:
> This seems super complicated compared to what we already have. Is it worth it?

Which part of it looks super complicated?

I think it's worth it. 

1. Now that we have all build jobs in one place, we can do a container build easily for all linux builds and reach zero flakiness at least for linux and cut about 2min per build.
2. If we add a new build flag in the future, we only need to add it in one place.
3. We can use one config to build and another config to run tests, which we still need to work around the compiler issue, right?

csujedihy:
So far, build time has been reduced to 14min vs 18-22min before. Test time remains the same. Also, the sanitizer build issue has been worked around.

I think this is a better model when we migrate perf tests to github where we would have to build somewhere else.

nibanks (approved):

PR #3728, opened 2023-06-27
Fix duonic installation hang

csujedihy:
## Description

Duonic installation hangs. Add a sleep after cert installation to see if that's the issue.

## Testing

CI


csujedihy:
turns out duonic has expired.

PR #3756, opened 2023-07-14
Minor fixes for Github workflows

csujedihy:
## Description

1. Skip installing signing certs if `ForTest` is not used with `ForKernel` in prepare-machine.ps1.
2. Only fetch the number of commits in checkout action required by the workflow.

## Testing

CI

### DIFF
--- a/.github/workflows/build-reuse-darwin-framework.yml
+++ b/.github/workflows/build-reuse-darwin-framework.yml
@@ -48,8 +48,6 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9
-      with:
-        fetch-depth: 0
     - name: Download Build Artifacts (x64)
       uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a
       with:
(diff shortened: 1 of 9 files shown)


scripts/prepare-machine.ps1
```diff
@@ -153,7 +153,13 @@ if ($ForTest) {
     # enabled for any possible test.
     $InstallTestCertificates = $true
     $InstallClog2Text = $true
-    $InstallSigningCertificates = $true; # For kernel drivers
+
+    # Since installing signing certs also checks whether test signing is enabled, which most
+    # likely will fail on a devbox, do it only when we need to test kernel drivers so that
+    # local testing setup won't be blocked by test signing not enabled.
+    if ($ForKernel) {
+        $InstallSigningCertificates = $true;
```
nibanks, line 161: Don't we need for duonic too?

.github/workflows/build-reuse-darwin-framework.yml
```diff
@@ -48,8 +48,6 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9
-      with:
-        fetch-depth: 0
```
nibanks, line 52: I thought I had a reason for this...

csujedihy: If a workflow needs to access tags, commit history or other branches than the ref branch, then it's needed.

scripts/prepare-machine.ps1
csujedihy: We have a check for duonic later on.

```
if ($InstallDuoNic) {
    $InstallSigningCertificates = $true;
}
```

nibanks (approved):

PR #3757, opened 2023-07-14
Minor Changes to Improve Listener Code Coverage

nibanks:
## Description

Noticed that a few lines were uncovered in code coverage, so I did some minimal refactoring to clean that up.

## Testing

Existing CI

## Documentation

N/A

### DIFF
--- a/src/core/listener.c
+++ b/src/core/listener.c
@@ -88,12 +88,7 @@ MsQuicListenerOpen(
 
 Error:
 
-    if (QUIC_FAILED(Status)) {
-
-        if (Listener != NULL) {
-            CXPLAT_FREE(Listener, QUIC_POOL_LISTENER);
-        }
-    }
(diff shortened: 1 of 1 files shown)


src/core/listener.c
```diff
@@ -88,12 +88,7 @@ MsQuicListenerOpen(
 
 Error:
 
-    if (QUIC_FAILED(Status)) {
-
-        if (Listener != NULL) {
-            CXPLAT_FREE(Listener, QUIC_POOL_LISTENER);
-        }
-    }
+    CXPLAT_DBG_ASSERT(QUIC_SUCCEEDED(Status) || Listener == NULL);
```
nibanks, line 91: I was able to remove this check entirely because we can never fail after allocating the `Listener`. But I put an assert here just in case someone changes that in the future.

csujedihy (approved):

PR #3758, opened 2023-07-15
Add QUIC_SETTINGS Support to SpinQuic

nibanks:
## Description

Trying to improve the code coverage numbers. Added some extra validation to the settings code to prevent obviously bad values from being set.

## Testing

Ran some locally and found/fixed a few things. Depending on the CI for the rest.

## Documentation

N/A

### DIFF
--- a/src/tools/spin/spinquic.cpp
+++ b/src/tools/spin/spinquic.cpp
@@ -633,31 +633,31 @@ void SpinQuicRandomizeSettings(QUIC_SETTINGS& Settings, uint16_t ThreadID)
         Settings.IsSet.MaximumMtu = TRUE;
         break;
     case 24:
-        Settings.SendBufferingEnabled = GetRandom(1ui8);
+        Settings.SendBufferingEnabled = GetRandom((uint8_t)1);
         Settings.IsSet.SendBufferingEnabled = TRUE;
         break;
     case 25:
-        Settings.PacingEnabled = GetRandom(1ui8);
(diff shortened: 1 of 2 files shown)


csujedihy (dismissed):

csujedihy:
Some stress tests crashed

src/tools/spin/spinquic.cpp
```diff
@@ -533,9 +533,162 @@ struct SetParamHelper {
     }
 };
 
+void SpinQuicRandomizeSettings(QUIC_SETTINGS& Settings, uint16_t ThreadID)
```
ProjectsByJackHe, line 536: I am guessing this function is added as a means to stress test (randomly assign settings to quic), Maybe this is why some of the stress tests are crashing?

nibanks: Yes, this function is about randomly assigning settings. But the "crash" was from a watchdog timer which fires if the test takes too long. So, I wonder if it's caused by some timer getting set to too big which then causes the delay. I haven't had time to investigate yet though. @ProjectsByJackHe this might be a good place for you to do your first investigation.

nibanks: I'm not sure which setting is causing the watchdog timeout. I'm going to disable most of these to try to narrow down the scope. We can follow up with other PRs to enable more.

csujedihy (approved):

PR #3769, opened 2023-07-21
Improve C++ wrapper.

sebastianpick:
## Description

This PR completes the MsQuicListener implementation in that it adds a proper C++-style MsQuicListenerCallback type. This type is in line with other callback implementations in the C++ wrapper. Furthermore, the PR improves the consistency of some function signatures and adds a few retrieval functions for certain object parameters.

## Testing

These changes are affect certain tests, which have been updated accordingly. New API additions where not added to the tests.

## Documentation

Smaller API changes and additions have been made and might require doc updates.

### DIFF
--- a/src/inc/msquic.hpp
+++ b/src/inc/msquic.hpp
@@ -160,6 +160,7 @@ struct QuicAddr {
     QUIC_ADDRESS_FAMILY GetFamily() const { return QuicAddrGetFamily(&SockAddr); }
     uint16_t GetPort() const { return QuicAddrGetPort(&SockAddr); }
     void SetPort(uint16_t Port) noexcept { QuicAddrSetPort(&SockAddr, Port); }
+    operator const QUIC_ADDR* () const noexcept { return &SockAddr; }
 };
 
 template<class T>
@@ -308,6 +309,7 @@ public:
         }
     }
     QUIC_STATUS GetInitStatus() const noexcept { return InitStatus; }
+    bool IsValid() const noexcept { return QUIC_SUCCEEDED(InitStatus); }
 };
 
 extern const MsQuicApi* MsQuic;
@@ -344,8 +346,8 @@ struct MsQuicRegistration {
     }
     QUIC_STATUS GetInitStatus() const noexcept { return InitStatus; }
     bool IsValid() const noexcept { return QUIC_SUCCEEDED(InitStatus); }
-    MsQuicRegistration(MsQuicRegistration& other) = delete;
-    MsQuicRegistration operator=(MsQuicRegistration& Other) = delete;
+    MsQuicRegistration(const MsQuicRegistration& Other) = delete;
+    MsQuicRegistration& operator=(const MsQuicRegistration& Other) = delete;
     void Shutdown(
         _In_ QUIC_CONNECTION_SHUTDOWN_FLAGS Flags,
         _In_ QUIC_UINT62 ErrorCode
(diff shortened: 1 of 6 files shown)


src/inc/msquic.hpp
```diff
@@ -765,14 +782,42 @@ struct MsQuicListener {
...
+    static
+        QUIC_STATUS
+        QUIC_API
+        NoOpCallback(
+            _In_ MsQuicListener* /* Listener */,
+            _In_opt_ void* /* Context */,
+            _Inout_ QUIC_LISTENER_EVENT* /* Event */
+            ) noexcept {
+            return QUIC_STATUS_SUCCESS;
+        }
```
nibanks, line 798: Fix formatting and we need to return an error by default, otherwise it would just accept a connection and leak it.
```suggestion
    static
    QUIC_STATUS
    QUIC_API
    NoOpCallback(
        _In_ MsQuicListener* /* Listener */,
        _In_opt_ void* /* Context */,
        _Inout_ QUIC_LISTENER_EVENT* /* Event */
        ) noexcept {
        return QUIC_STATUS_NOT_SUPPORTED;
    }
```

src/inc/msquic.hpp
```diff
@@ -670,15 +672,30 @@ struct MsQuicConfiguration {
 #endif
 };
 
+enum MsQuicCleanUpMode {
+    CleanUpManual,
+    CleanUpAutoDelete,
+};
+
+typedef QUIC_STATUS MsQuicListenerCallback(
+    _In_ struct MsQuicListener* listener,
```
nibanks, line 681:
```suggestion
    _In_ struct MsQuicListener* Listener,
```

src/inc/msquic.hpp
nibanks: In fact, I'm not sure it ever makes sense to have a no-op callback for a listener. Maybe we just remove it and always require a callback.

src/perf/lib/PerfServer.cpp
```diff
@@ -129,7 +129,7 @@ PerfServer::GetExtraData(
 
 QUIC_STATUS
 PerfServer::ListenerCallback(
-    _In_ HQUIC /*ListenerHandle*/,
+    _In_ MsQuicListener* /*ListenerHandle*/,
```
nibanks, line 132:
```suggestion
    _In_ MsQuicListener* /*Listener*/,
```

src/perf/lib/PerfServer.h
```diff
@@ -90,14 +90,14 @@ class PerfServer : public PerfBase {
 
     QUIC_STATUS
     ListenerCallback(
-        _In_ HQUIC ListenerHandle,
+        _In_ MsQuicListener* ListenerHandle,
```
nibanks, line 93: Let's update `ListenerHandle` to `Listener` here and the others below.

src/test/lib/ApiTest.cpp
```diff
@@ -335,23 +335,26 @@ void QuicTestValidateConfiguration()
...
-        StopCompleteEvent->Set();
-        return QUIC_STATUS_SUCCESS;
+    _Function_class_(QUIC_LISTENER_CALLBACK)
+    template<typename T>
+    QUIC_STATUS
+        QUIC_API
+        DummyListenerCallback(
+            T,
+            void* Context,
+            QUIC_LISTENER_EVENT* Event
```
nibanks, line 347:
```suggestion
    QUIC_API
    DummyListenerCallback(
        T,
        void* Context,
        QUIC_LISTENER_EVENT* Event
```

src/test/lib/ApiTest.cpp
```diff
@@ -335,23 +335,26 @@ void QuicTestValidateConfiguration()
...
-    )
+namespace
 {
-    CxPlatEvent* StopCompleteEvent = (CxPlatEvent*)Context;
-    if (StopCompleteEvent &&
-        Event->Type == QUIC_LISTENER_EVENT_STOP_COMPLETE) {
-        StopCompleteEvent->Set();
-        return QUIC_STATUS_SUCCESS;
+    _Function_class_(QUIC_LISTENER_CALLBACK)
+    template<typename T>
```
nibanks, line 341: Why is this templated? It should always be the type that matches MsQuicListener now, right?

nibanks: I see what you're doing below.... I guess it's ok for now.

src/inc/msquic.hpp
sebastianpick: Makes sense. I will remove this an also update the default parameters in the MsQuicListener c'tor.

nibanks (approved):

PR #3770, opened 2023-07-21
Dynamically load/unload xdp dll

mtfriesen:
## Description

Use `XdpLoadApi` which uses run-time linking instead of `XdpOpenApi` which uses load-time linking.

## Testing

CI/CD

## Documentation

No.

### DIFF
--- a/src/platform/CMakeLists.txt
+++ b/src/platform/CMakeLists.txt
@@ -55,8 +55,7 @@ if ("${CX_PLATFORM}" STREQUAL "windows" AND QUIC_USE_XDP)
         platform
         PUBLIC
         inc
-        wbemuuid
-        ${PROJECT_SOURCE_DIR}/artifacts/xdp/lib/xdpapi.lib)
+        wbemuuid)
 else()
     target_link_libraries(platform PUBLIC inc)
 endif()
--- a/src/platform/datapath_raw_xdp_win.c
+++ b/src/platform/datapath_raw_xdp_win.c
@@ -1018,8 +1018,7 @@ CxPlatDpRawInitialize(
     const uint16_t* ProcessorList;
 
     CxPlatListInitializeHead(&Xdp->Interfaces);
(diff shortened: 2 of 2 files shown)


csujedihy (approved):

ProjectsByJackHe:
Out of curiosity, why is run-time linking better than load time linking in our context? From some light research, with load time linking you can bundle everything into a self-contained executable and performance wise load time linking is a bit better (correct me if I am wrong). Happy to learn something new!

mtfriesen:
> Out of curiosity, why is run-time linking better than load time linking in our context? From some light research, with load time linking you can bundle everything into a self-contained executable and performance wise load time linking is a bit better (correct me if I am wrong). Happy to learn something new!

Yup, absolutely, load time linking can perform more performance optimizations and some of them are quite valuable, especially with Spectre mitigations. The reason we want run-time linking is to (eventually) allow MsQuic to use XDP if it happens to be installed and available on the system, and otherwise use regular sockets. If the XDP DLL isn't present on the system, then load-time linking will prevent the MsQuic DLL from loading, which we do not want.

ProjectsByJackHe:
> > Out of curiosity, why is run-time linking better than load time linking in our context? From some light research, with load time linking you can bundle everything into a self-contained executable and performance wise load time linking is a bit better (correct me if I am wrong). Happy to learn something new!
> 
> Yup, absolutely, load time linking can perform more performance optimizations and some of them are quite valuable, especially with Spectre mitigations. The reason we want run-time linking is to (eventually) allow MsQuic to use XDP if it happens to be installed and available on the system, and otherwise use regular sockets. If the XDP DLL isn't present on the system, then load-time linking will prevent the MsQuic DLL from loading, which we do not want.

Ahh I see, thanks! So in the past, we just always assumed XDP was always available, otherwise MsQuic.dll wouldn't have been able to load.

nibanks (approved):

PR #3777, opened 2023-07-26
Use UDP Length for XDP RX

### LINKED ISSUE #3776: UDP parsing logic does not honor UDP length field
### Describe the bug

UDP parsing logic does not honor UDP length field. It infers the payload length from IP layer.

### Affected OS

- [X] Windows
- [ ] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

_No response_

### MsQuic version

main

1. Receive a UDP packet whose IP payload is greater than UDP payload.

### Expected behavior

Data after UDP payload should not be accepted

### Actual outcome

Data after UDP payload will be indicated up

### Additional details

_No response_

nibanks:
## Description

Uses the UDP length field instead of the IP length for receiving the UDP payload. Fixes #3776.

## Testing

CI

## Documentation

N/A

### DIFF
--- a/src/platform/datapath_raw_socket.c
+++ b/src/platform/datapath_raw_socket.c
@@ -128,7 +128,6 @@ CxPlatDpRawParseUdp(
         return;
     }
 
-    Length -= sizeof(UDP_HEADER);
     Packet->Reserved = L4_TYPE_UDP;
 
     Packet->Route->RemoteAddress.Ipv4.sin_port = Udp->SourcePort;
--- a/src/platform/datapath_raw_socket.c
+++ b/src/platform/datapath_raw_socket.c
@@ -118,6 +118,16 @@ CxPlatDpRawParseUdp(
         return;
     }
 
+    if (Length < QuicNetByteSwapShort(Udp->Length)) {
+        QuicTraceEvent(
+            DatapathErrorStatus,
+            "[data][%p] ERROR, %u, %s.",
+            Datapath,
+            Length,
+            "UDP Length larger than IP length");
+        return;
+    }
+
     Length -= sizeof(UDP_HEADER);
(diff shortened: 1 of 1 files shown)


src/platform/datapath_raw_socket.c
```diff
@@ -118,14 +118,24 @@ CxPlatDpRawParseUdp(
...
+        QuicTraceEvent(
+            DatapathErrorStatus,
+            "[data][%p] ERROR, %u, %s.",
+            Datapath,
+            Length,
+            "UDP Length larger than IP length");
+        return;
+    }
+
     Length -= sizeof(UDP_HEADER);
```
csujedihy, line 131: We should remove this. It's not used.

src/platform/datapath_raw_socket.c
```diff
@@ -118,14 +118,24 @@ CxPlatDpRawParseUdp(
         return;
     }
 
+    if (Length < QuicNetByteSwapShort(Udp->Length)) {
```
csujedihy, line 121: nit: maybe we should store the UDP length from UDP header in a local var so we don't have to do the swap twice. Though, that swap might be completely optimized out.

csujedihy:
Solving #3776

src/platform/datapath_raw_socket.c
nibanks: I'll leave it as is for now. Let's see if the compiler can deal with it.

csujedihy (approved):

csujedihy (commented):
Is sidecar needed to be updated?

nibanks:
> Is sidecar needed to be updated?

No, because it's using an existing event with new payload.

PR #3778, opened 2023-07-26
Reliable Reset Stream Support - Phase 1 - Feature Negotiation

ProjectsByJackHe:
## Description

This PR is a **start** to trying to solve issue#3775. 
It's implementation is based off https://datatracker.ietf.org/doc/draft-ietf-quic-reliable-stream-reset/ 

As you can imagine, the changes will be HUUUUUUUUUUGE, so I wanted to split it into multiple PRs, to ease the review process.

In terms of scope for this PR,

we want to build all the necessary logic for proper feature negotiation between client and server.
This includes adding a new Transport parameter, a new settings parameter (internal and external), and logic to handle
negotiation of the settings. Additionally, a new Connection event is necessary to communicate to the app the outcome of a feature negotiation, so the app knows and can respond accordingly. 

## Testing

New unit tests are added to validate the behavior of the various components. Including new Frame tests, Settings tests, and Transport parameter tests.

End 2 end testing is also in scope for this PR, to validate the behavior of the TLS negotiation.

## Documentation

Not in this PR, but for future, there should be docs outlining how to use this new experimental feature.


docs/Settings.md
```diff
@@ -169,7 +169,8 @@ These parameters are accessed by calling [GetParam](./api/GetParam.md) or [SetPa
 | `QUIC_PARAM_CONN_STATISTICS_V2`<br> 22            | QUIC_STATISTICS_V2            | Get-only  | Connection-level statistics, version 2.                                                   |
 | `QUIC_PARAM_CONN_STATISTICS_V2_PLAT`<br> 23       | QUIC_STATISTICS_V2            | Get-only  | Connection-level statistics with platform-specific time format, version 2.                |
 | `QUIC_PARAM_CONN_ORIG_DEST_CID` <br> 24           | uint8_t[]                     | Get-only  | The original destination connection ID used by the client to connect to the server.       |
-
+| `QUIC_PARAM_CONN_THIS_RELIABLE_RESET_ENABLED` 25  | uint8_t (BOOLEAN)             | Both      | Indicate/query support for QUIC CLOSE_STREAM frames. Must be set before start.            |
+| `QUIC_PARAM_CONN_PEER_RELIABLE_RESET_ENABLED` 26  | uint8_t (BOOLEAN)             | Get-only  | Indicates peer advertised support for QUIC CLOSE_STREAM frames.                           |
```
nibanks, line 173: Let's only have 1 option. Do you enable this feature or not?

ProjectsByJackHe: Do we not care about the peer?

src/core/connection.c
```diff
@@ -6602,6 +6602,30 @@ QuicConnParamSet(
...
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+
+        if (QUIC_CONN_BAD_START_STATE(Connection)) {
+            Status = QUIC_STATUS_INVALID_STATE;
+            break;
+        }
+
+        Connection->Settings.ThisReliableResetFrameEnabled = *(BOOLEAN*)Buffer;
```
nibanks, line 6617: No need to include "Frame" in any of these variable names. Name it after the feature, which is support for ReliableStreamReset.

docs/Settings.md
nibanks: You can start with it enabled. After the handshake it reflects the negotiated value.

src/core/frame.h
```diff
@@ -145,7 +145,9 @@ typedef enum QUIC_FRAME_TYPE {
     QUIC_FRAME_CONNECTION_CLOSE     = 0x1cULL, // to 0x1d
     QUIC_FRAME_CONNECTION_CLOSE_1   = 0x1dULL,
     QUIC_FRAME_HANDSHAKE_DONE       = 0x1eULL,
-    /* 0x1f to 0x2f are unused currently */
+    /* 0x1f to 0x20 are unused currently */ 
+    QUIC_FRAME_RELIABLE_RESET_STREAM= 0x21ULL, // intentionally ignore type 0x20 of QUIC_RESET_STREAM cause it's likely to be removed from RFC.
```
ami-GS, line 149: space before '='

src/core/settings.h
```diff
@@ -40,6 +40,7 @@ typedef struct QUIC_SETTINGS_INTERNAL {
             uint64_t PacingEnabled                          : 1;
             uint64_t MigrationEnabled                       : 1;
             uint64_t DatagramReceiveEnabled                 : 1;
+            uint64_t ThisReliableResetFrameEnabled          : 1;
```
ami-GS, line 43: RESERVED to be 22

ami-GS, line 43: And, what does "This" prefix mean?

src/core/frame.h
```diff
@@ -253,6 +255,39 @@ QuicResetStreamFrameDecode(
...
+// QUIC_FRAME_RELIABLE_RESET_STREAM Encoding/Decoding
+//
+typedef struct QUIC_RELIABLE_RESET_STREAM_EX {
+
+    QUIC_VAR_INT StreamID;
+    QUIC_VAR_INT ErrorCode;
+    QUIC_VAR_INT FinalSize;
+    QUIC_VAR_INT ReliableSize;
+    
+} QUIC_RELIABLE_RESET_STREAM_EX;
```
nibanks, line 268: Let's add a test to FrameTest.cpp for this new frame.

src/inc/msquic.h
```diff
@@ -890,6 +890,8 @@ typedef struct QUIC_SCHANNEL_CREDENTIAL_ATTRIBUTE_W {
 #define QUIC_PARAM_CONN_STATISTICS_V2                   0x05000016  // QUIC_STATISTICS_V2
 #define QUIC_PARAM_CONN_STATISTICS_V2_PLAT              0x05000017  // QUIC_STATISTICS_V2
 #define QUIC_PARAM_CONN_ORIG_DEST_CID                   0x05000018  // uint8_t[]
+#define QUIC_PARAM_CONN_THIS_RELIABLE_RESET_ENABLED     0x05000019  // uint8_t (BOOLEAN)
```
ami-GS, line 893: I think "THIS" is "LOCAL" if you really need these two settings

src/core/connection.c
```diff
@@ -7206,6 +7230,44 @@ QuicConnParamGet(
...
+            break;
+        }
+
+        if (Buffer == NULL) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+
+        *BufferLength = sizeof(BOOLEAN);
+        *(BOOLEAN*)Buffer = Connection->Datagram.PeerReliableResetEnabled;
```
nibanks, line 7266: Copy/paste issue? :D

src/core/datagram.h
```diff
@@ -38,6 +38,11 @@ typedef struct QUIC_DATAGRAM {
     //
     BOOLEAN SendEnabled : 1;
 
+    //
+    // Indicates that the peer accepts RELIABLE_RESET kind of frames, in addition to RESET_STREAM frames.
+    //
+    BOOLEAN PeerReliableResetEnabled: 1;
```
nibanks, line 44: This shouldn't go in the Datagram, but the connection.

docs/Settings.md
ProjectsByJackHe: Ahh I see, so start the flag to be off. If the app turns it on, then connects to a server, if the server says their flag is turned off (passed through connection), then we turn ours off (overriding the app's decision)?

src/core/frame.h
ProjectsByJackHe: Added! Will update.

src/core/datagram.h
ProjectsByJackHe: Updated! Will push later.

src/core/settings.h
ProjectsByJackHe: Sorry, I don't understand. Are you saying IsSet->RESERVED should be 22 instead of 23?

nibanks: As you add new bits, you effectively take away reserved bits, so the count of reserved should reduce.

docs/Settings.md
nibanks: Essentially.

src/core/settings.h
ProjectsByJackHe: That makes a lot of sense. I was wondering what RESERVED meant lol. Will update.

src/inc/msquic.h
ProjectsByJackHe: got rid of 'this' lol

src/core/datagram.h
```diff
@@ -37,7 +37,6 @@ typedef struct QUIC_DATAGRAM {
     // send.
     //
     BOOLEAN SendEnabled : 1;
-
```
nibanks, line 40: nit: undo this change

src/core/frame.h
nibanks, line 149: `QUIC_FRAME_IS_KNOWN` needs to be updated to include this.

src/inc/msquic.h
```diff
@@ -890,6 +890,7 @@ typedef struct QUIC_SCHANNEL_CREDENTIAL_ATTRIBUTE_W {
 #define QUIC_PARAM_CONN_STATISTICS_V2                   0x05000016  // QUIC_STATISTICS_V2
 #define QUIC_PARAM_CONN_STATISTICS_V2_PLAT              0x05000017  // QUIC_STATISTICS_V2
 #define QUIC_PARAM_CONN_ORIG_DEST_CID                   0x05000018  // uint8_t[]
+#define QUIC_PARAM_CONN_RELIABLE_RESET_ENABLED          0x05000019  // uint8_t (BOOLEAN)
```
nibanks, line 893: I was thinking about this more... and I think putting this in the QUIC_SETTINGs is a better route. You already added it to the internal settings, but add it to the external one (at the end so you don't break back compat), and update settings.c accordingly. So then you don't have this.

You will also need to add a new Connection event to indicate if the feature was negotiated; and it's only delivered for apps that try to enable it.

I'd be happy to talk more of the details next week.

src/core/connection.h
```diff
@@ -630,6 +630,11 @@ typedef struct QUIC_CONNECTION {
     //
     uint16_t KeepAlivePadding;
 
+    //
+    // Indicates that the peer accepts RELIABLE_RESET kind of frames, in addition to RESET_STREAM frames.
+    //
+    uint8_t PeerReliableResetEnabled;
```
nibanks, line 636: Rename this to something like `ReliableResetStreamNegotiated` and only set it to TRUE if the local app enables it and the remote/peer does too.

src/core/frame.h
ProjectsByJackHe: Done.

src/inc/msquic.h
ProjectsByJackHe: Good point, that makes a lot of sense. I added `ReliableResetEnabled` in QUIC_SETTINGS in msquic.h. 

In terms of scope for this PR, I think including all the code for the variable initializations and the logic for the setup and teardown could be a good stopping point. I'd love to discuss this more with you more if you have the time!

src/core/connection.h
ProjectsByJackHe: I am guessing once we add the new connection event, that's where we will set this flag?

src/inc/msquic.h
```diff
@@ -1212,6 +1215,12 @@ typedef struct QUIC_CONNECTION_EVENT {
             QUIC_STATUS DeferredStatus;         // Most severe error status (only valid with QUIC_CREDENTIAL_FLAG_DEFER_CERTIFICATE_VALIDATION)
             QUIC_CERTIFICATE_CHAIN* Chain;      // Peer certificate chain (platform specific). Valid only during QUIC_CONNECTION_EVENT_PEER_CERTIFICATE_RECEIVED callback.
         } PEER_CERTIFICATE_RECEIVED;
+        
+        #ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+        struct {
+            uint8_t ReliableResetEnabledNegotitatedValue;
+        } RELIABLE_RESET_NEGOTIATED;
```
nibanks, line 1222: You also need to add a new type to `QUIC_CONNECTION_EVENT_TYPE`.

src/inc/msquic.h
```diff
@@ -1212,6 +1215,12 @@ typedef struct QUIC_CONNECTION_EVENT {
             QUIC_STATUS DeferredStatus;         // Most severe error status (only valid with QUIC_CREDENTIAL_FLAG_DEFER_CERTIFICATE_VALIDATION)
             QUIC_CERTIFICATE_CHAIN* Chain;      // Peer certificate chain (platform specific). Valid only during QUIC_CONNECTION_EVENT_PEER_CERTIFICATE_RECEIVED callback.
         } PEER_CERTIFICATE_RECEIVED;
+        
+        #ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
```
nibanks, line 1219: nit: Our style generally has #ifdef's all the way to the left.

src/inc/msquic.h
```diff
@@ -1212,6 +1215,12 @@ typedef struct QUIC_CONNECTION_EVENT {
...
+        struct {
+            uint8_t ReliableResetEnabledNegotitatedValue;
```
nibanks, line 1221:
```suggestion
            BOOLEAN IsNegotiated;
```

src/core/settings.c
```diff
@@ -1280,6 +1296,7 @@ QuicSettingsDump(
     QuicTraceLogVerbose(SettingEcnEnabled,                  "[sett] EcnEnabled             = %hhu", Settings->EcnEnabled);
     QuicTraceLogVerbose(SettingHyStartEnabled,              "[sett] HyStartEnabled         = %hhu", Settings->HyStartEnabled);
     QuicTraceLogVerbose(SettingEncryptionOffloadAllowed,    "[sett] EncryptionOffloadAllowed = %hhu", Settings->EncryptionOffloadAllowed);
+    QuicTraceLogVerbose(SettingReliableResetEnabled,        "[sett] ReliableResetEnabled   = %hhu", Settings->ReliableResetEnabled);
```
nibanks, line 1299: You will need to run the script to regenerate the sidecar.

src/core/datagram.h
nibanks: Looks like you added some trailing spaces. 😄 Please delete those too. You can configure most code editors to do this automatically.

src/core/frame.h
```diff
@@ -253,6 +256,39 @@ QuicResetStreamFrameDecode(
...
+BOOLEAN
+QuicReliableResetFrameDecode(
+    _In_ uint16_t BufferLength,
+    _In_reads_bytes_(BufferLength)
+        const uint8_t * const Buffer,
+    _Inout_ uint16_t* Offset,
+    _Out_ QUIC_RELIABLE_RESET_STREAM_EX* Frame
+    );
+
+
```
nibanks, line 291: nit: extra new line here

src/core/datagram.h
ProjectsByJackHe: LOL I'll install a vscode extension for that or smth

src/core/connection.h
```diff
@@ -630,6 +630,11 @@ typedef struct QUIC_CONNECTION {
     //
     uint16_t KeepAlivePadding;
 
+    //
+    // Indicates that the peer accepts RELIABLE_RESET kind of frames, in addition to RESET_STREAM frames.
+    //
+    uint8_t ReliableResetStreamNegotiated;
```
nibanks, line 636: Let's put this in `QUIC_CONNECTION_STATE` instead of here.

src/core/frame.h
```diff
@@ -163,7 +165,8 @@ CXPLAT_STATIC_ASSERT(
 #define QUIC_FRAME_IS_KNOWN(X) \
```
nibanks, line 165: Also update `QuicConnRecvFrames` to process the frames. Even it's a no-op for now.

src/core/connection.h
ProjectsByJackHe: Added, will update, and I can change its type to BOOLEAN.

src/core/frame.h
ProjectsByJackHe: Added case.

src/core/connection.c
```diff
@@ -6602,10 +6606,6 @@ QuicConnParamSet(
         return QUIC_STATUS_SUCCESS;
     }
 
-    //
```
nibanks, line 6605: Undo these changes here

src/core/frame.h
ProjectsByJackHe: Missed this, done!

src/core/connection.c
ProjectsByJackHe: Done

src/core/connection.c
```diff
@@ -3059,6 +3059,17 @@ QuicConnProcessPeerTransportParameters(
...
         }
 
+        if (Connection->Settings.ReliableResetEnabled &&
+            (Connection->PeerTransportParams.Flags & QUIC_TP_FLAG_RELIABLE_RESET_ENABLED) > 0) {
+            //
+            // Reliable Reset is enabled on both sides.
+            //
+            Connection->State.ReliableResetStreamNegotiated = TRUE;
+        } else {
+            Connection->State.ReliableResetStreamNegotiated = TRUE;
```
nibanks: Remove this line.

nibanks: Actually, I think you should remove the whole else; just like Grease bit above.

src/core/crypto_tls.c
```diff
@@ -1164,6 +1171,18 @@ QuicCryptoTlsEncodeTransportParameters(
...
+        TPBuf =
+            TlsWriteTransportParam(
+                QUIC_TP_ID_RELIABLE_RESET_ENABLED,
+                0,
+                NULL,
+                TPBuf);
+        QuicTraceLogConnVerbose(
+            EncodeTPReliableResetEnabled,
+            Connection,
+            "TP: Encode Reliable Reset Enabled");
```
nibanks, line 1184:
```suggestion
        QuicTraceLogConnVerbose(
            EncodeTPReliableReset,
            Connection,
            "TP: Encode Reliable Reset");
```

src/core/crypto_tls.c
```diff
@@ -1822,6 +1841,23 @@ QuicCryptoTlsDecodeTransportParameters( // NOLINT(readability-function-size, goo
...
+                    Connection,
+                    Length,
+                    "Invalid length of QUIC_TP_ID_RELIABLE_RESET_ENABLED");
+                goto Exit;
+            }
+            TransportParams->Flags |= QUIC_TP_FLAG_RELIABLE_RESET_ENABLED;
+            QuicTraceLogConnVerbose(
+                DecodeTPReliableResetEnabled,
+                Connection,
+                "TP: Decode Reliable Reset Enabled");
```
nibanks, line 1858:
```suggestion
            QuicTraceLogConnVerbose(
                DecodeTPReliableReset,
                Connection,
                "TP: Decode Reliable Reset");
```

src/core/connection.c
ProjectsByJackHe: Shouldn't we also want to update the app's internal settings to be false?

nibanks: No. That flag reflects what the app wants. The fact that the feature wasn't negotiated doesn't change the fact that the app wanted it.

src/core/settings.c
```diff
@@ -1423,6 +1440,9 @@ QuicSettingsDumpNew(
     if (Settings->IsSet.EncryptionOffloadAllowed) {
         QuicTraceLogVerbose(SettingEncryptionOffloadAllowed,        "[sett] EncryptionOffloadAllowed   = %hhu", Settings->EncryptionOffloadAllowed);
     }
+    if (Settings->IsSet.ReliableResetEnabled) {
+        QuicTraceLogVerbose(SettingReliableResetEnabled,            "[sett] ReliableResetEnabled   = %hhu", Settings->ReliableResetEnabled);
```
nibanks, line 1444: Please make sure the spacing before the `=` aligns with all the others.

src/core/settings.h
```diff
@@ -92,13 +93,14 @@ typedef struct QUIC_SETTINGS_INTERNAL {
     uint8_t SendBufferingEnabled            : 1;
     uint8_t PacingEnabled                   : 1;
     uint8_t MigrationEnabled                : 1;
-    uint8_t DatagramReceiveEnabled          : 1;
+    uint8_t DatagramReceiveEnabled          : 1;
```
nibanks, line 96: Did you add a space here? 😄 Please remove

src/core/connection.c
```diff
@@ -3059,6 +3059,14 @@ QuicConnProcessPeerTransportParameters(
...
         }
 
+        if (Connection->Settings.ReliableResetEnabled &&
+            (Connection->PeerTransportParams.Flags & QUIC_TP_FLAG_RELIABLE_RESET_ENABLED) > 0) {
+            //
+            // Reliable Reset is enabled on both sides.
+            //
+            Connection->State.ReliableResetStreamNegotiated = TRUE;
+        }
+
```
nibanks, line 3069: Right here, if `Connection->Settings.ReliableResetEnabled` is set, you should then deliver the new event to the app passing `Connection->State.ReliableResetStreamNegotiated` to them.

src/core/settings.h
ProjectsByJackHe: image

src/core/connection.c
ProjectsByJackHe: Added call to callback.

src/core/settings.c
ProjectsByJackHe: Added a few spaces align =

ProjectsByJackHe:
@nibanks Build tests are passing finally! For actually testing this new extension, I was wondering if the settings and transport parameter tests I added are sufficient? It looks like Greasebit and Hystart both have similar tests.

src/core/connection.c
nibanks: You put the callback above, not on this line. We need to give this callback regardless of if the feature was negotiated. It needs to be delivered if the app set `Connection->Settings.ReliableResetEnabled` to `TRUE`.

src/core/connection.c
```diff
@@ -5276,6 +5293,15 @@ QuicConnRecvFrames(
             AckImmediately = TRUE;
             break;
 
+        case QUIC_FRAME_RELIABLE_RESET_STREAM:
+            QuicTraceLogConnInfo(
+                ReliableResetFrameReceived,
+                Connection,
+                "Received a reliable reset frame. Local reliable reset extension preferences: %hhu",
+                Connection->Settings.ReliableResetEnabled);
```
nibanks, line 5301: Please remove this log. We have a different way of logging per-frame info (in packet.c).

src/core/connection.c
ProjectsByJackHe: Thanks for the clarification!
We always deliver callback if the app set local settings to TRUE. 
I can move callback out of the negotiation branch.

src/core/connection.c
```diff
@@ -3065,15 +3065,16 @@ QuicConnProcessPeerTransportParameters(
...
-                // Send event to app to indicate completion of negotiation.
-                //
-                QUIC_CONNECTION_EVENT Event;
-                Event.Type = QUIC_CONNECTION_EVENT_RELIABLE_RESET_NEGOTIATED;
-                Event.RELIABLE_RESET_NEGOTIATED.IsNegotiated = Connection->State.ReliableResetStreamNegotiated;
-                QuicConnIndicateEvent(Connection, &Event);
-            }
+        }
+
+        if (Connection->Settings.IsSet.ReliableResetEnabled) {
```
nibanks, line 3070:
```suggestion
        if (Connection->Settings.ReliableResetEnabled) {
```

src/core/connection.c
ProjectsByJackHe: Figured it out, just remove the break so the branch falls through to get rid of the warning lol

src/core/connection.c
```diff
@@ -2426,6 +2426,16 @@ QuicConnGenerateLocalTransportParameters(
         LocalTP->Flags |= QUIC_TP_FLAG_GREASE_QUIC_BIT;
     }
 
+    if (Connection->Settings.ReliableResetEnabled) {\
```
nibanks, line 2429:
```suggestion
    if (Connection->Settings.ReliableResetEnabled) {
```

src/core/connection.c
```diff
@@ -2426,6 +2426,16 @@ QuicConnGenerateLocalTransportParameters(
...
+        QuicTraceLogConnInfo(
+        SetReliableResetTPparameter,
+        Connection,
+        "Set Reliable Reset TP, is client = %hhu",
+        QuicConnIsClient(Connection));
```
nibanks, line 2434: Remove before checking in.

src/core/settings.c
```diff
@@ -1615,6 +1620,7 @@ QuicSettingsSettingsToInternal(
     SETTING_COPY_TO_INTERNAL(DatagramReceiveEnabled, Settings, InternalSettings);
     SETTING_COPY_TO_INTERNAL(ServerResumptionLevel, Settings, InternalSettings);
     SETTING_COPY_TO_INTERNAL(GreaseQuicBitEnabled, Settings, InternalSettings);
+    SETTING_COPY_TO_INTERNAL(ReliableResetEnabled, Settings, InternalSettings);
```
nibanks, line 1623: You had it right before, undo this change.

src/core/connection.c
ProjectsByJackHe: Sounds good

src/core/settings.c
ProjectsByJackHe: done and pushed

src/test/bin/winkernel/control.cpp
```diff
@@ -517,7 +517,8 @@ typedef union {
     QUIC_RUN_VN_TP_ODD_SIZE_PARAMS OddSizeVnTpParams;
     UINT8 TestServerVNTP;
     BOOLEAN Bidirectional;
-
+    BOOLEAN ServerSupport;
+    BOOLEAN ClientSupport;
```
nibanks, line 521: You need to put this in a struct. You're putting this in a `union` which means they all share the same memory. So these are always the same value.

src/test/bin/winkernel/control.cpp
```diff
@@ -1338,7 +1339,14 @@ QuicTestCtlEvtIoDeviceControl(
                 Params->CustomCertValidationParams.AcceptCert,
                 Params->CustomCertValidationParams.AsyncValidation));
         break;
-
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+    case IOCTL_QUIC_RELIABLE_RESET_NEGOTIATION:
+        CXPLAT_FRE_ASSERT(Params != nullptr);
+        QuicTestCtlRun(
+            QuicTestReliableResetNegotiation(Params->Family, Params->ServerSupport, Params->ClientSupport)
+        );
```
nibanks, line 1347:
```suggestion
        QuicTestCtlRun(
            QuicTestReliableResetNegotiation(
                Params->Family,
                Params->ServerSupport,
                Params->ClientSupport));
```

src/test/bin/quic_gtest.cpp
```diff
@@ -2255,6 +2270,11 @@ INSTANTIATE_TEST_SUITE_P(
     Handshake,
     WithHandshakeArgs7,
     testing::ValuesIn(HandshakeArgs7::Generate()));
+
+INSTANTIATE_TEST_SUITE_P(
+    Handshake,
+    WithHandshakeArgs10,
```
nibanks, line 2276: Call the args something like `WithReliableResetArgs`

src/test/MsQuicTests.h
```diff
@@ -1177,3 +1190,7 @@ typedef struct {
     // QUIC_RUN_CUSTOM_CERT_VALIDATION
 
 #define QUIC_MAX_IOCTL_FUNC_CODE 110
+
+#define IOCTL_QUIC_RELIABLE_RESET_NEGOTIATION \
+    QUIC_CTL_CODE(111, METHOD_BUFFERED, FILE_WRITE_DATA)
+    // QUIC_RUN_RELIABLE_RESET_NEGOTIATION
```
nibanks, line 1196: Add this before `QUIC_MAX_IOCTL_FUNC_CODE` and make sure to increment `QUIC_MAX_IOCTL_FUNC_CODE` to `111`. Also the comment that follows is the argument type (i.e. struct) for the IOCTL.

nibanks: You will also have to update `QUIC_IOCTL_BUFFER_SIZES` in control.cpp with the size.

src/manifest/clog.sidecar
nibanks, line 1: Please make sure no extra logs go left behind from your test events you added during development.

src/inc/msquic.h
```diff
@@ -1212,6 +1217,12 @@ typedef struct QUIC_CONNECTION_EVENT {
             QUIC_STATUS DeferredStatus;         // Most severe error status (only valid with QUIC_CREDENTIAL_FLAG_DEFER_CERTIFICATE_VALIDATION)
             QUIC_CERTIFICATE_CHAIN* Chain;      // Peer certificate chain (platform specific). Valid only during QUIC_CONNECTION_EVENT_PEER_CERTIFICATE_RECEIVED callback.
         } PEER_CERTIFICATE_RECEIVED;
+
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
```
nibanks, line 1221:
```suggestion
#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
```

src/inc/msquic.h
```diff
@@ -1140,6 +1142,9 @@ typedef enum QUIC_CONNECTION_EVENT_TYPE {
     QUIC_CONNECTION_EVENT_RESUMED                           = 13,   // Server-only; provides resumption data, if any.
     QUIC_CONNECTION_EVENT_RESUMPTION_TICKET_RECEIVED        = 14,   // Client-only; provides ticket to persist, if any.
     QUIC_CONNECTION_EVENT_PEER_CERTIFICATE_RECEIVED         = 15,   // Only with QUIC_CREDENTIAL_FLAG_INDICATE_CERTIFICATE_RECEIVED set
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+    QUIC_CONNECTION_EVENT_RELIABLE_RESET_NEGOTIATED         = 16,
```
nibanks, line 1146: Add a comment to indicate this is only indicated if `ReliableResetEnabled` is true.

src/inc/msquic.h
```diff
@@ -716,7 +717,8 @@ typedef struct QUIC_SETTINGS {
             uint64_t HyStartEnabled            : 1;
 #ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
             uint64_t EncryptionOffloadAllowed  : 1;
-            uint64_t ReservedFlags             : 62;
+            uint8_t ReliableResetEnabled       : 1;
```
nibanks, line 720:
```suggestion
            uint64_t ReliableResetEnabled      : 1;
```
match the types of everything else in the bitfield

src/core/settings.c
```diff
@@ -1711,6 +1744,7 @@ QuicSettingsGetSettings(
     SETTING_COPY_FROM_INTERNAL(DatagramReceiveEnabled, Settings, InternalSettings);
     SETTING_COPY_FROM_INTERNAL(ServerResumptionLevel, Settings, InternalSettings);
     SETTING_COPY_FROM_INTERNAL(GreaseQuicBitEnabled, Settings, InternalSettings);
+    SETTING_COPY_FROM_INTERNAL(ReliableResetEnabled, Settings, InternalSettings);
```
nibanks, line 1747: Remove this line. You already have the copy below.

src/core/settings.c
```diff
@@ -602,6 +608,11 @@ QuicSettingApply(
         Destination->IsSet.GreaseQuicBitEnabled = TRUE;
     }
 
+    if (Source->IsSet.ReliableResetEnabled && (!Destination->IsSet.ReliableResetEnabled || OverWrite)) {
```
nibanks, line 611: Move this down, after `EncryptionOffloadAllowed`.

src/core/connection.c
```diff
@@ -6479,6 +6511,7 @@ QuicConnParamSet(
         //
         // Must be set before the client connection is started.
         //
+
```
nibanks, line 6514: nit: remove this new line and add back the one you removed below.

src/manifest/clog.sidecar
ProjectsByJackHe: Sounds good. The only logs I added were the "this app ... [state]", and "reliable reset negotiated!" when the conditions were met, I removed them.

src/inc/msquic.h
ProjectsByJackHe: Sounds good.

src/test/bin/winkernel/control.cpp
ProjectsByJackHe: Done. will push

src/test/MsQuicTests.h
ProjectsByJackHe: Sounds good, interesting how we implement a hash table using an array.

src/test/bin/quic_gtest.cpp
ProjectsByJackHe: Done

src/core/settings.c
ProjectsByJackHe: good catch! I can remove

src/core/settings.c
ProjectsByJackHe: Done

src/core/connection.c
ProjectsByJackHe: Done

nibanks (approved):

PR #3779, opened 2023-07-27
Consume XDP v0.17.0

mtfriesen:
## Description

Upgrade to XDP v0.17.0 This contains a number of editorial API changes and an MSI installer. The QEO function is also now an optional function in XDP.

## Testing

Builds locally. CI/CD.

## Documentation

I don't think so.

### DIFF
--- a/scripts/prepare-machine.ps1
+++ b/scripts/prepare-machine.ps1
@@ -222,11 +222,6 @@ function Install-Xdp-Sdk {
     if (!$IsWindows) { return } # Windows only
     $XdpPath = Join-Path $ArtifactsPath "xdp"
     if ($Force) {
-        try {
-            # Make sure an old driver isn't installed.
-            netcfg.exe -u ms_xdp
-            pnputil.exe /delete-driver "$XdpPath\bin\xdp.inf"
-        } catch {}
         rm -Force -Recurse $XdpPath -ErrorAction Ignore | Out-Null
     }
     if (!(Test-Path $XdpPath)) {
@@ -251,7 +246,7 @@ function Install-Xdp-Driver {
     }
 
     Write-Host "Installing XDP driver"
-    netcfg.exe -l "$XdpPath\bin\xdp.inf" -c s -i ms_xdp
+    msiexec.exe /i $XdpPath\bin\xdp-for-windows.msi /quiet
 }
 
 # Completely removes the XDP driver and SDK.
@@ -261,8 +256,7 @@ function Uninstall-Xdp {
     if (!(Test-Path $XdpPath)) { return; }
 
     Write-Host "Uninstalling XDP"
-    try { netcfg.exe -u ms_xdp } catch {}
-    try { pnputil.exe /delete-driver "$XdpPath\bin\xdp.inf" } catch {}
+    try { msiexec.exe /x $XdpPath\bin\xdp-for-windows.msi /quiet } catch {}
     rm -Force -Recurse $XdpPath -ErrorAction Ignore | Out-Null
 }
 
--- a/scripts/prepare-machine.ps1
+++ b/scripts/prepare-machine.ps1
@@ -246,7 +246,7 @@ function Install-Xdp-Driver {
     }
 
     Write-Host "Installing XDP driver"
-    msiexec.exe /i $XdpPath\bin\xdp-for-windows.msi /quiet
+    msiexec.exe /i $XdpPath\bin\xdp-for-windows.msi /quiet | Out-Null
 }
 
 # Completely removes the XDP driver and SDK.
@@ -256,7 +256,7 @@ function Uninstall-Xdp {
     if (!(Test-Path $XdpPath)) { return; }
 
     Write-Host "Uninstalling XDP"
-    try { msiexec.exe /x $XdpPath\bin\xdp-for-windows.msi /quiet } catch {}
+    try { msiexec.exe /x $XdpPath\bin\xdp-for-windows.msi /quiet | Out-Null } catch {}
     rm -Force -Recurse $XdpPath -ErrorAction Ignore | Out-Null
 }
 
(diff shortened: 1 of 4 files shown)


scripts/xdp-install.ps1
```diff
@@ -31,16 +31,15 @@ $ProgressPreference = 'SilentlyContinue'
 $NetworkPath = (irm "https://raw.githubusercontent.com/microsoft/msquic/main/scripts/xdp-devkit.json").path
 $ZipPath = Join-Path $Destination "xdp.zip"
 $XdpPath = Join-Path $Destination "xdp"
-$XdpInstalled = Test-Path "$XdpPath\bin\xdp.inf"
+$InstallId = (Get-CimInstance Win32_Product -Filter "Name = 'XDP for Windows'").IdentifyingNumber
```
csujedihy, line 34: Wouldn't this fail the entire script if XDP is not installed?

csujedihy: Oh, `$ProgressPreference = 'SilentlyContinue'`, nvm.

csujedihy (dismissed):

scripts/prepare-machine.ps1
```diff
@@ -251,7 +246,7 @@ function Install-Xdp-Driver {
     }
 
     Write-Host "Installing XDP driver"
-    netcfg.exe -l "$XdpPath\bin\xdp.inf" -c s -i ms_xdp
+    msiexec.exe /i $XdpPath\bin\xdp-for-windows.msi /quiet
```
nibanks, line 249: So this automatically uninstalls anything previously installed? Is that why you removed the delete above?

scripts/xdp-devkit.json
```diff
@@ -1,3 +1,3 @@
 {
-    "path" : "https://github.com/microsoft/xdp-for-windows/releases/download/v0.16.8/xdp-devkit-x64-0.16.8-prerelease.zip"
+    "path" : "https://github.com/microsoft/xdp-for-windows/releases/download/v0.17.0/xdp-devkit-x64-0.17.0.zip"
```
nibanks, line 2: Not a comment on this PR, but should the release still be named prerelease so long as there are no signed binaries, or have we just moved away from that naming convention in XDP?

src/platform/datapath_raw_xdp_win.c
```diff
@@ -476,7 +478,8 @@ CxPlatDpRawInterfaceUninitialize(
     if (Interface->Rules != NULL) {
         for (uint8_t i = 0; i < Interface->RuleCount; ++i) {
             if (Interface->Rules[i].Pattern.IpPortSet.PortSet.PortSet) {
-                CxPlatFree(Interface->Rules[i].Pattern.IpPortSet.PortSet.PortSet, PORT_SET_TAG);
+                CxPlatFree(
+                    (uint8_t*)Interface->Rules[i].Pattern.IpPortSet.PortSet.PortSet, PORT_SET_TAG);
```
nibanks, line 482: Why was this cast necessary?

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1023,6 +1026,8 @@ CxPlatDpRawInitialize(
         goto Error;
     }
 
+    Xdp->XdpQeoSet = (XDP_QEO_SET_FN *)Xdp->XdpApi->XdpGetRoutine(XDP_QEO_SET_FN_NAME);
```
nibanks, line 1029: Should we check the result here? Or maybe just a debug assert?

nibanks: Ah, I see. You check for NUL below. I guess that's ok too, but it's probably worth a log to says "Hey, you're using an XDP that no longer has QEO"

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1346,11 +1351,15 @@ CxPlatSocketUpdateQeo(
...
     BOOLEAN AtLeastOneSucceeded = FALSE;
     for (CXPLAT_LIST_ENTRY* Entry = Xdp->Interfaces.Flink; Entry != &Xdp->Interfaces; Entry = Entry->Flink) {
-        Status =
-            Xdp->XdpApi->XdpQeoSet(
-                CONTAINING_RECORD(Entry, XDP_INTERFACE, Link)->XdpHandle,
-                Connections,
-                sizeof(Connections));
+        if (Xdp->XdpQeoSet != NULL) {
+            Status =
+                Xdp->XdpQeoSet(
```
nibanks, line 1356: Does `XdpQeoSet` ever return `E_NOINTERFACE` itself? I'm just wondering if we can differentiate between not available or the call failed.

scripts/prepare-machine.ps1
mtfriesen: It uninstalls anything that has been installed using this package. It would be an error for something else to leave XDP installed.

scripts/xdp-devkit.json
mtfriesen: Right, we're moving away from that notion. The -prerelease tag was never meant to be in the release titles; the 0 major series is implicitly prerelease. The -prerelease tag was meant for builds in e.g. main.

src/platform/datapath_raw_xdp_win.c
mtfriesen: It looks like MsQuic is using the rule to hold a pointer to the PortSet. This makes the MsQuic code clunky because XDP rules now take a `const` pointer to the port set.

src/platform/datapath_raw_xdp_win.c
mtfriesen: The error is logged in the QEO function any time QEO is attempted to be set.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1480,7 +1489,8 @@ CxPlatDpRawPlumbRulesOnSocket(
             if (IsCreated) {
                 if (Rule) {
                     CxPlatDpRawSetPortBit(
-                        Rule->Pattern.IpPortSet.PortSet.PortSet, Socket->LocalAddress.Ipv4.sin_port);
+                        (uint8_t*)Rule->Pattern.IpPortSet.PortSet.PortSet,
```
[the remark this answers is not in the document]
mtfriesen: Resolving as duplicate.

nibanks (approved):

PR #3781, opened 2023-07-29
Remove QUIC_USE_RAW_DATAPATH macro

ami-GS:
## Description
Remove build time functional switch of QUIC_USE_RAW_DATAPATH

## Testing

Adding.....

## Documentation

Adding....


src/platform/datapath_raw_socket.c
```diff
@@ -43,6 +43,57 @@ CxPlatGetSocket(
     return Socket;
 }
 
+BOOLEAN
+CxPlatIsRouteReady(
+    _In_ QUIC_CONNECTION *Connection,
+    _In_ CXPLAT_ROUTE_RESOLUTION_CALLBACK_HANDLER Callback,
+    _In_ BOOLEAN PathChallenge
+) {
+    QUIC_PATH* Path = &Connection->Paths[0];
```
nibanks, line 52: This breaks the abstraction boundaries. Platform can't use, or know about, any QUIC layer structures.

nibanks (changes requested):

src/platform/datapath_raw_socket.c
ami-GS: then how core code distinguish underlying datapath is raw or normal?

src/platform/datapath_raw.h
```diff
@@ -18,6 +18,7 @@ typedef struct CXPLAT_SOCKET_POOL {
 } CXPLAT_SOCKET_POOL;
 
 typedef struct CXPLAT_DATAPATH CXPLAT_DATAPATH;
+typedef struct QUIC_PATH QUIC_PATH;
```
nibanks, line 21: Shouldn't be necessary. Please remove now.

src/core/path.c
```diff
@@ -34,12 +34,13 @@ QuicPathInitialize(
     Path->RttVariance = Path->SmoothedRtt / 2;
     Path->EcnValidationState =
         Connection->Settings.EcnEnabled ? ECN_VALIDATION_TESTING : ECN_VALIDATION_FAILED;
-#ifdef QUIC_USE_RAW_DATAPATH
+
+    // for RAW datapath
```
nibanks, line 38: Comment doesn't follow our style. I think you can just remove it.

src/core/cubic.c
```diff
@@ -312,9 +312,9 @@ CubicCongestionControlOnCongestionEvent(
             "[conn][%p] Persistent congestion event",
             Connection);
         Connection->Stats.Send.PersistentCongestionCount++;
-#ifdef QUIC_USE_RAW_DATAPATH
-        Connection->Paths[0].Route.State = RouteSuspected;
-#endif
+
+        Connection->Paths[0].Route.State = RouteSuspected; // used only for RAW datapath
```
nibanks, line 316: @csujedihy Should we have a similar line in bbr.c?

src/core/send.c
```diff
@@ -1018,6 +1018,45 @@ QuicSendGetNextStream(
...
+    //
+    // Make sure the route is resolved before sending packets.
+    //
+    //
+    // We need to set the path challenge flag back on so that when route is resolved,
+    // we know we need to continue to send the challenge.
+    //
+    CXPLAT_DBG_ASSERT((!PathChallenge && Path->IsActive) ||
+                      (PathChallenge && Path->Route.State != RouteSuspected));
+    if ((!PathChallenge && (Path->Route.State == RouteUnresolved || Path->Route.State == RouteSuspected)) ||
```
nibanks, line 1036: @csujedihy & @maolson-msft do we really need special logic for "Path Challenge" scenarios where only `RouteUnresolved` will trigger route resolution, instead of also `RouteSuspected`?

src/core/send.c
```diff
@@ -1042,37 +1081,10 @@ QuicSendPathChallenges(
...
-                //
-                CXPLAT_DBG_ASSERT(Status == QUIC_STATUS_PENDING || QUIC_FAILED(Status));
-                Send->SendFlags |= QUIC_CONN_SEND_FLAG_PATH_CHALLENGE;
-                continue;
-            }
-        } else if (Path->Route.State == RouteResolving) {
-            //
-            // Can't send now. Once route resolution completes, we will resume sending.
-            //
+        if (CxPlatIsRawDatapath() && !CxPlatIsRouteReady(Connection, TRUE)) {
```
nibanks, line 1084: We can actually do a tiny bit of refactoring and simplify this logic:

1. For socket-based datapaths, make `CxPlatResolveRoute` just set the route to the `RouteResolved` state inline, and then return success. There's nothing it needs to do because routes are practically already resolved always.
2. Remove the `CxPlatIsRawDatapath` function and just always call `CxPlatIsRouteReady`, which will call `CxPlatResolveRoute` if not already resolved.

src/test/bin/quic_gtest.cpp
```diff
@@ -13,9 +13,7 @@
 bool TestingKernelMode = false;
 bool PrivateTestLibrary = false;
 bool UseDuoNic = false;
-#if defined(QUIC_USE_RAW_DATAPATH) && defined(QUIC_API_ENABLE_PREVIEW_FEATURES)
```
nibanks, line 16: Keep the `#if defined(QUIC_API_ENABLE_PREVIEW_FEATURES)`?

src/test/bin/quic_gtest.cpp
```diff
@@ -2375,12 +2381,14 @@ int main(int argc, char** argv) {
         } else if (strcmp("--duoNic", argv[i]) == 0) {
             UseDuoNic = true;
         } else if (strcmp("--useQTIP", argv[i]) == 0) {
-#if defined(QUIC_USE_RAW_DATAPATH) && defined(QUIC_API_ENABLE_PREVIEW_FEATURES)
-            UseQTIP = true;
-#else
-            printf("QTIP is not supported in this build.\n");
-            return -1;
+            UseQTIP = CxPlatIsRawDatapath();
+#if !defined(QUIC_API_ENABLE_PREVIEW_FEATURES)
```
nibanks, line 2385: Maybe we expose a feature flag for QTIP support too.

src/test/bin/quic_gtest.cpp
ami-GS: This ifdef just wraps UseQTIP definition. No need to conditionally build

src/test/bin/quic_gtest.cpp
ami-GS: The flag is put in Datapath->Features, but at this point the flag (Datapath) is not available

src/core/send.c
ami-GS: CxPlatIsRawDatapath can be removed from send.c, but it is needed for tests

src/core/send.c
csujedihy: I think I added that path challenge route resolution logic. I think the assumption I had is when sending path challenge, it's impossible that the route on the path is in suspected state because suspected state is only valid if we already have the route resolved. But, path challenge can only happen on a fresh path that either has not resolved the route or is in middle of resolving it.

csujedihy: And that assert has not been hit so far. So, I think my assumption is still correct.

nibanks: In the future, we might send a challenge on a previously valid, but now suspect path. We don't today, but shouldn't preclude the possibility. Do you see any reason not to have the same logic for normal sends and path challenges?

src/core/cubic.c
csujedihy: yes

src/core/send.c
csujedihy: No, it's probably fine to remove that suspected check. At least we can try.

src/platform/datapath_winkernel.c
```diff
@@ -796,6 +796,10 @@ CxPlatDataPathQuerySockoptSupport(
 
     } while (FALSE);
 
+#if defined(QUIC_API_ENABLE_PREVIEW_FEATURES)
+    Datapath->Features |= CXPLAT_DATAPATH_FEATURE_TCP;
```
csujedihy, line 800: Why setting the TCP flag?

ami-GS: oh, which datapath support TCP (QTIP)? only XDP?

csujedihy: only winxdp supports QTIP.

src/core/send.c
```diff
@@ -1018,6 +1018,46 @@ QuicSendGetNextStream(
...
+    if (Path->Route.State == RouteResolved) {
+        return TRUE;
+    }
+
+    //
+    // We need to set the path challenge flag back on so that when route is resolved,
+    // we know we need to continue to send the challenge.
+    //
+    CXPLAT_DBG_ASSERT(Path->IsActive);
+    if (Path->Route.State == RouteUnresolved) {
```
nibanks, line 1038:
```suggestion
    if (Path->Route.State == RouteUnresolved || Path->Route.State == RouteSuspected) {
```
We need to copy the logic originally in the normal send path, which includes the `RouteSuspected` case.

src/platform/datapath_winuser.c
```diff
@@ -1138,6 +1138,7 @@ CxPlatDataPathQuerySockoptSupport(
     } else {
         Datapath->Features |= CXPLAT_DATAPATH_FEATURE_RECV_COALESCING;
     }
+
```
nibanks, line 1141: nit: please remove extra newline

src/platform/unittest/DataPathTest.cpp
```diff
@@ -535,8 +533,8 @@ struct CxPlatSocket {
         if (QUIC_SUCCEEDED(InitStatus)) {
             CxPlatSocketGetLocalAddress(Socket, &Route.LocalAddress);
             CxPlatSocketGetRemoteAddress(Socket, &Route.RemoteAddress);
-#ifdef QUIC_USE_RAW_DATAPATH
-            if (!QuicAddrIsWildCard(&Route.RemoteAddress)) {
+            if (!!(CxPlatDataPathGetSupportedFeatures(Datapath) & CXPLAT_DATAPATH_FEATURE_RAW_SOCKET) &&
```
nibanks, line 536: To make this a bit cleaner (IMO), let's add a new helper function to `CxPlatDataPath` called `IsSupported(flag)` so you can change this line to:
```suggestion
            if (Datapath.IsSupported(CXPLAT_DATAPATH_FEATURE_RAW_SOCKET) &&
```
And then use it below too.

src/platform/unittest/DataPathTest.cpp
```diff
@@ -1007,10 +1004,12 @@ TEST_P(DataPathTest, MultiBindListener) {
 }
 
 #ifdef _WIN32
-#ifndef QUIC_USE_RAW_DATAPATH
 TEST_F(DataPathTest, TcpListener)
 {
     CxPlatDataPath Datapath(nullptr, &EmptyTcpCallbacks);
+    if (!(CxPlatDataPathGetSupportedFeatures(Datapath) & CXPLAT_DATAPATH_FEATURE_TCP)) {
```
nibanks, line 1010: Now that we have this flag, you can remove the `#ifdef _WIN32` above too.

src/test/bin/quic_gtest.cpp
```diff
@@ -1399,8 +1395,18 @@ TEST_P(WithHandshakeArgs4, RandomLossResumeRejection) {
 #endif // QUIC_DISABLE_RESUMPTION
 #endif // QUIC_TEST_DATAPATH_HOOKS_ENABLED
 
-#ifndef QUIC_USE_RAW_DATAPATH
 TEST_P(WithFamilyArgs, Unreachable) {
+    uint32_t Length = sizeof(uint32_t);
+    uint32_t Features = 0;
+    GTEST_ASSERT_EQ(QUIC_STATUS_SUCCESS,
```
nibanks, line 1401: Can you use `ASSERT_TRUE(QUIC_SUCCEEDED` instead?

src/test/bin/quic_gtest.cpp
nibanks: It was also protecting usage of `QUIC_EXECUTION_CONFIG_FLAG_QTIP` below. The goal is to eventually remove QTIP, so I'd rather have it isolated as much as possible.

src/inc/msquic.h
```diff
@@ -824,10 +824,9 @@ void
 #define QUIC_PARAM_GLOBAL_VERSION_SETTINGS              0x01000007  // QUIC_VERSION_SETTINGS
 #endif
 #define QUIC_PARAM_GLOBAL_LIBRARY_GIT_HASH              0x01000008  // char[64]
-#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
 #define QUIC_PARAM_GLOBAL_EXECUTION_CONFIG              0x01000009  // QUIC_EXECUTION_CONFIG
-#endif
 #define QUIC_PARAM_GLOBAL_TLS_PROVIDER                  0x0100000A  // QUIC_TLS_PROVIDER
+#define QUIC_PARAM_GLOBAL_DATAPATH_FEATURES             0x0100000B  // uint32_t
```
nibanks, line 829: Let's put this is msquicp.h and have this private for now. I don't want a public API that depends on a type only in quic_datapath.h, which isn't public.

src/test/bin/quic_gtest.h
```diff
@@ -697,11 +694,19 @@ struct ValidateConnectionEventArgs {
...
+
+#if !defined(QUIC_DISABLE_0RTT_TESTS) // TODO: Fix openssl/XDP bug and enable this back
+        uint32_t Length = sizeof(uint32_t);
+        uint32_t Features = 0;
+        MsQuic->GetParam(
+            nullptr,
+            QUIC_PARAM_GLOBAL_DATAPATH_FEATURES,
+            &Length,
+            &Features);
+        TestCount = !!(Features & CXPLAT_DATAPATH_FEATURE_RAW_SOCKET) ? 2 : 3;
```
nibanks, line 707: You have to do this in several places. How about we make a helper function to query either if a feature is supported, or specifically just if RawSocket is supported?

src/inc/quic_datapath.h
```diff
@@ -446,6 +444,8 @@ CxPlatDataPathUpdateConfig(
 #define CXPLAT_DATAPATH_FEATURE_SEND_SEGMENTATION     0x0004
 #define CXPLAT_DATAPATH_FEATURE_LOCAL_PORT_SHARING    0x0008
 #define CXPLAT_DATAPATH_FEATURE_PORT_RESERVATIONS     0x0010
+#define CXPLAT_DATAPATH_FEATURE_TCP                   0x0020
+#define CXPLAT_DATAPATH_FEATURE_RAW_SOCKET            0x0040
```
nibanks, line 448: I'm actually inclined to call this simply `CXPLAT_DATAPATH_FEATURE_RAW`.

src/platform/datapath_winuser.c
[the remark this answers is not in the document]
csujedihy: I think Daiki meant to use that flag for QTIP.

src/core/path.c
```diff
@@ -35,10 +35,12 @@ QuicPathInitialize(
     Path->EcnValidationState =
         Connection->Settings.EcnEnabled ? ECN_VALIDATION_TESTING : ECN_VALIDATION_FAILED;
 
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
```
nibanks, line 38: This check should be unnecessary in core code. It's always set in core.

src/core/send.c
```diff
@@ -1018,6 +1018,46 @@ QuicSendGetNextStream(
     return NULL;
 }
 
+BOOLEAN
+CxPlatIsRouteReady(
+    _In_ QUIC_CONNECTION *Connection
+) {
```
nibanks, line 1024:
```suggestion
    )
{
```

src/core/send.c
```diff
@@ -1018,6 +1018,47 @@ QuicSendGetNextStream(
...
+            CXPLAT_DBG_ASSERT(Status == QUIC_STATUS_PENDING || QUIC_FAILED(Status));
+            return FALSE;
+        }
+    } else if (Path->Route.State == RouteResolving) {
+        //
+        // Can't send now. Once route resolution completes, we will resume sending.
+        //
+        return FALSE;
+    }
+    return TRUE;
```
nibanks, line 1059: How is this case any different than the return at the beginning. I'd prefer not to have lines that are impossible to hit.

src/inc/msquic.h
```diff
@@ -824,9 +824,7 @@ void
 #define QUIC_PARAM_GLOBAL_VERSION_SETTINGS              0x01000007  // QUIC_VERSION_SETTINGS
 #endif
 #define QUIC_PARAM_GLOBAL_LIBRARY_GIT_HASH              0x01000008  // char[64]
-#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
```
nibanks, line 827: Please add this back. This is still preview.

src/core/send.c
ami-GS: It is arrived if code goes line 1045. route resolution succeeded

src/core/send.c
```diff
@@ -1018,6 +1018,46 @@ QuicSendGetNextStream(
...
+                Path->Binding->Socket, &Path->Route, Path->ID, (void*)Connection, QuicConnQueueRouteCompletion);
+        if (Status == QUIC_STATUS_SUCCESS) {
+            QuicConnRelease(Connection, QUIC_CONN_REF_ROUTE);
+        } else {
+            //
+            // Route resolution failed or pended. We need to pause sending.
+            //
+            CXPLAT_DBG_ASSERT(Status == QUIC_STATUS_PENDING || QUIC_FAILED(Status));
+        }
+        return Status == QUIC_STATUS_SUCCESS;
```
nibanks, line 1052:
```suggestion
        if (Status == QUIC_STATUS_SUCCESS) {
            QuicConnRelease(Connection, QUIC_CONN_REF_ROUTE);
            return TRUE;
        }
        //
        // Route resolution failed or pended. We need to pause sending.
        //
        CXPLAT_DBG_ASSERT(Status == QUIC_STATUS_PENDING || QUIC_FAILED(Status));
```

src/test/bin/quic_gtest.cpp
```diff
@@ -1705,11 +1708,10 @@ TEST_P(WithSendArgs3, SendIntermittently) {
...
 
 TEST_P(WithSend0RttArgs1, Send0Rtt) {
-#if defined(QUIC_USE_RAW_DATAPATH) && defined(QUIC_API_ENABLE_PREVIEW_FEATURES)
-    if (UseQTIP) {
+#if defined(QUIC_API_ENABLE_PREVIEW_FEATURES)
+    if (QuitTestIsFeatureSupported(CXPLAT_DATAPATH_FEATURE_RAW) && UseQTIP) {
         //
-        // QTIP doesn't work with 0-RTT. QTIP only pauses and caches 1 packet during
-        // TCP handshake.
+        // NAT rebind doesn't make sense for TCP and QTIP.
```
nibanks, line 1714: Please revert the changes in comment for these 0-RTT tests.

src/core/send.c
```diff
@@ -1018,6 +1018,45 @@ QuicSendGetNextStream(
     return NULL;
 }
 
+BOOLEAN
+CxPlatIsRouteReady(
+    _In_ QUIC_CONNECTION *Connection
+    )
+{
+    QUIC_PATH* Path = &Connection->Paths[0];
```
nibanks, line 1026: I just realized after looking at the path challenge code, that this function must take `Path` as an argument, because the callers are operating on different paths. Not just `Paths[0]`.

ami-GS: great catch, thanks

nibanks (approved):

nibanks:
@ami-GS please make sure failing tests aren't because of this PR before merging.

ami-GS:
`CxPlatIsRouteReady` seems to affect functionality. maybe due to unexpected CXPLAT_ROUTE.State value

nibanks (approved):

src/test/lib/DataTest.cpp
```diff
@@ -14,7 +14,11 @@
 #include "DataTest.cpp.clog.h"
 #endif
 
-#if defined(QUIC_USE_RAW_DATAPATH) && defined(QUIC_API_ENABLE_PREVIEW_FEATURES)
+#if defined(_KERNEL_MODE)
+bool UseQTIP = false;
+// currently x is only CXPLAT_DATAPATH_FEATURE_RAW
```
csujedihy, line 19: x?

ami-GS: `QuitTestIsFeatureSupported(x)` this x

csujedihy: Ah I see

src/test/bin/quic_gtest.cpp
```diff
@@ -81,9 +81,9 @@ class QuicTestEnvironment : public ::testing::Environment {
             printf("Initializing for User Mode tests\n");
             MsQuic = new(std::nothrow) MsQuicApi();
             ASSERT_TRUE(QUIC_SUCCEEDED(MsQuic->GetInitStatus()));
-#if defined(QUIC_USE_RAW_DATAPATH) && defined(QUIC_API_ENABLE_PREVIEW_FEATURES)
-            if (UseQTIP) {
-                QUIC_EXECUTION_CONFIG Config = {QUIC_EXECUTION_CONFIG_FLAG_QTIP, 10000, 0};
+#if defined(QUIC_API_ENABLE_PREVIEW_FEATURES)
+            if (QuitTestIsFeatureSupported(CXPLAT_DATAPATH_FEATURE_RAW) && UseQTIP) {
```
ami-GS, line 85: @nibanks Datapath is not initialized at this point.

ami-GS: all other call in quic_gtest.cpp neither.

src/test/lib/precomp.h
```diff
@@ -35,6 +35,11 @@
 #define WIN_ASSERT CXPLAT_FRE_ASSERT
 #endif
 #include "karray.h"
+
+#ifndef GLOBAL_FOR_KERNEL
+#define GLOBAL_FOR_KERNEL
+bool UseQTIP = false;
+#endif
```
nibanks, line 42: Is this necessary any more?

nibanks (approved):

src/test/lib/precomp.h
ami-GS: Each test has `extern bool UseQTIP` which seems to be visible by user mode for kernel test. However kernel mode cannot link quic_gtest.cpp which has `bool UseQTIP = false`.

csujedihy: Just an idea: I think the ideal solution is passing this flag down from the gtest program along with those test ioctls. It should be a trivial change.

PR #3787, opened 2023-08-02
Fix CFS warnings

csujedihy:
## Description

Add NuGet.config and cargo config file to use Azure artifacts so we can avoid CFS warnings in OneBranch pipeline. We don't actually use them in the pipeline but their presence makes the warnings go away.

## Testing

OB CI


nibanks:
> We don't actually use them in the pipeline but their presence makes the warnings go away.

Looks like _something_ is trying to use them, especially the Nuget config. We have an exception for the Nuget already I believe. Maybe just try to fix/ignore the Cargo one for now?

csujedihy:
> > We don't actually use them in the pipeline but their presence makes the warnings go away.
> 
> Looks like _something_ is trying to use them, especially the Nuget config. We have an exception for the Nuget already I believe. Maybe just try to fix/ignore the Cargo one for now?

Wow, I added 2 new files and existing tests broke. I didn't expect that. These configs only make sense for OneBranch pipelines. I wonder if I can make the github workflow not pick them up.

nibanks:
> Another proper way to fix it would be actually authenticate with the sources defined in the configs.

What about the exception route? That worked for Nuget.

csujedihy:
> > Another proper way to fix it would be actually authenticate with the sources defined in the configs.
> 
> What about the exception route? That worked for Nuget.

I took another look at the warnings. They were emitted due to Cargo config. Though, the pipeline output still flagged nuget config but they didn't end up being actual warnings.

So, I guess we only need to deal with cargo config.

.github/workflows/cargo.yml
```diff
@@ -40,6 +40,12 @@ jobs:
     - name: Install Cargo
       if: runner.os == 'Linux'
       run: curl https://sh.rustup.rs -sSf | sh -s -- -y
+    # The cargo config file is required for OneBranch CFS.
+    # Remove it for this workflow. Otherwise, cargo will try to reach the internal feed
```
nibanks, line 44: There's got to be a better way...

csujedihy: I have seen some proposals here https://internals.rust-lang.org/t/problems-of-cargo-config-files-and-possible-solutions/12987/8 but I have not seen an elegant way to ignore .cargo/config.toml. We could also move the cargo file into a child folder, where cargo would try to locate the config file there.e

nibanks: I'm worried while this fixed **our** repo, other consumers of our crate will be broken by this. Is there no way to tell OneBranch to ignore this?

csujedihy: What do you mean by other consumers of our crate will be broken by this?

csujedihy: We delete this config file so our crate behaves the same as it originally did. What will be broken?

csujedihy: There is a way to tell CFS to ignore everything at repo level by submitting a form.

nibanks: Let's go that route.

Issue #3640, opened 2023-05-18
Publish MsQuic packages for all Linux distros supported by .NET


CarnaViire:
### Describe the bug

Officiall list of .NET supported distros: https://github.com/dotnet/core/blob/main/release-notes/8.0/supported-os.md#linux

The following versions don't have msquic packages:
_(*) distro has OpenSSL 3 by default_

- **Fedora**
  - [x] 38 (*) (https://packages.microsoft.com/fedora/38/prod/Packages/)
- **Mariner**
  - [x] 2.0 (https://packages.microsoft.com/cbl-mariner/2.0/prod/Microsoft/x86_64/Packages/l/)
    - libmsquic 2.2.0 is present, but 2.2.1 is missing
- **RHEL**
  - [x] 8 (https://packages.microsoft.com/rhel/8/prod/Packages/)
  - [x] 9 (*) (https://packages.microsoft.com/rhel/9/prod/Packages/)
- **Ubuntu**
  - [x] 22.10 (*) (https://packages.microsoft.com/ubuntu/22.10/prod/pool/main/)
  - [x] 23.04 (*) (https://packages.microsoft.com/ubuntu/23.04/prod/pool/main/)

Upcoming releases (not sure if we want to track them separately or we can publish packages in advance):

- Debian 12 (2023-06-10)
- Fedora 39 (~2023-11)
- openSUSE 16.0 (Fall 2023)
- Ubuntu 23.10 (2023-10-12)

The following distros don't even have a dedicated directory in packages.microsoft.com (I don't know whether you can create one, so I'm not sure whether something can be done)

- RHEL family
  -  CentOS Stream 8&9
  - AlmaLinux 8&9
  - Rocky Linux 8&9

cc @wfurt @ManickaP

### Affected OS

- [ ] Windows
- [X] Linux
- [ ] macOS
- [ ] Other (specify below)

### MsQuic version

v2.2

### Steps taken to reproduce bug

Checking packages.microsoft.com for each of the distros

### Expected behavior

Package exists in packages.microsoft.com and can be installed

### Actual outcome

Package doesn't exist in packages.microsoft.com 

### Additional details

<details>
<summary>List of all supported distros with EOL dates</summary>

```
UPD: 2023-05-18

alpine

cycle    eol
-----    ---
3.15     2023-11-01 (!)
3.16     2024-05-23
3.17     2024-11-22
3.18 (*) 2025-05-09


centos

cycle eol
----- ---
7     2024-06-30


[...]

PR #3789, opened 2023-08-04
Add Debian 12 package & fix Fedora 37 package

CarnaViire:
## Description

Adds Debian 12 (bookworm) to the list of distros to publish to. Debian 12 was released on [June 10](https://www.debian.org/releases/bookworm/). The distro already exists on [packages.microsoft.com](https://packages.microsoft.com/debian/12/prod/). LMK if I've missed something @nibanks @csujedihy

The second change is an attempt to fix Fedora 37 usecase, which has strange problems with OpenSSL 3 package (https://github.com/microsoft/msquic/issues/3640#issuecomment-1579439966). However, the distro has a compat OpenSSL 1.1 package, so I propose to fallback to that, given that the distro will be supported only until [2023-11-14](https://fedorapeople.org/groups/schedule/f-39/f-39-key-tasks.html).

## Testing

I've verified that Ubuntu 22.04 package (OpenSSL 3) works on Debian 12, and that CentOS 8 package (OpenSSL 1.1) works on Fedora 37 ([verification script](https://github.com/CarnaViire/QuicDockerTest/blob/main/quic-docker.ps1)). I don't have access to verify actual publish scripts.

## Documentation

N/A

cc @wfurt @ManickaP

### DIFF
--- a/.azure/OneBranch.Publish.yml
+++ b/.azure/OneBranch.Publish.yml
@@ -52,6 +52,7 @@ parameters:
   - microsoft-ubuntu-jammy-prod-apt
   - microsoft-ubuntu-kinetic-prod-apt
   - microsoft-ubuntu-lunar-prod-apt
+  - microsoft-debian-bookworm-prod-apt
 - name: openssl3rpmrepos
   type: object
   default:
--- a/.azure/OneBranch.Publish.yml
+++ b/.azure/OneBranch.Publish.yml
@@ -32,6 +32,7 @@ parameters:
   - microsoft-fedora32-prod-yum
   - microsoft-fedora33-prod-yum
   - microsoft-fedora34-prod-yum
+  - microsoft-fedora37-prod-yum
   - microsoft-rhel7.3-prod-yum
   - microsoft-rhel8.0-prod-yum
   - microsoft-rhel8.1-prod-yum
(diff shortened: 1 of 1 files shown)


CarnaViire:
The change literally only touches a publishing script 🤷‍♀️ I have no idea how this can affect code coverage, especially mysterious are those "indirect changes" 😅

nibanks (approved):
LGTM, but I want @csujedihy to look on Monday.

csujedihy (approved):

wfurt:
should this be ported to release branch(es)?

nibanks:
> should this be ported to release branch(es)?

@csujedihy do we need to backport these package changes to run them for a release branch? Also, can we rerun the publish for a release that was already previously published or do we need to make a new release?

csujedihy:
> do we need to backport these package changes to run them for a release branch?

We do at least for now. The branch being tagged is the release branch. The publishing automation is triggered by tags.

> Also, can we rerun the publish for a release that was already previously published or do we need to make a new release?

We could just rerun the pipeline but I'd rather not try to publish the same version more than once. Last time I did this, I somewhat broke PMC. I was politely warned by PMC folks not to do it again. Though, they fixed all the bugs that I found.

CarnaViire:
@nibanks @csujedihy I see that the packages are still unavailable, is there anything else that should be done? Or are we just waiting for the next release? If so, do you have any ETA on when that will happen?

csujedihy:
The packages for the new distros should be published with the next release. Republishing an existing version is possible but I want to avoid doing that. I think mid-October we will have to release a new version? @nibanks

nibanks:
Talked with @carnaviire on IM. Looks like we missed a backport PR. Sorry! We will backport it real quick and try to create a quick v2.2.4 just for this.

PR #3796, opened 2023-08-08
Update XDP Usage to v1.0

nibanks:
## Description

Updates MsQuic to consume the new v1.0 release of XDP for Windows.

## Testing

Automated tests.

## Documentation

N/A


mtfriesen (dismissed):

mtfriesen (dismissed):

scripts/xdp-install.ps1
```diff
@@ -1,71 +0,0 @@
...
-    }
-
-    # Download and extract the latest version of the XDP kit (overwriting as
-    # necessary).
-    Write-Host "Downloading and extracting $NetworkPath"
-    Invoke-WebRequest -Uri $NetworkPath -OutFile $ZipPath
-    Expand-Archive -Path $ZipPath -DestinationPath $XdpPath -Force
-    Remove-Item -Path $ZipPath
-
-    if (!$SkipCerts) {
```
mtfriesen, line 53: For prerelease (not officially signed) XDP builds, do we still need to install the certs somewhere?

scripts/package-nuget.ps1
```diff
@@ -145,7 +145,7 @@ $NugetSourceFolder = Join-Path $RootDir "src/distribution"
 if ($UWP) {
     $PackageName = "Microsoft.Native.Quic.MsQuic.UWP.$Tls"
 } elseif ($XDP) {
-    Copy-Item -Path (Join-Path $PSScriptRoot xdp-devkit.json) -Destination (Join-Path $PackagingDir xdp-devkit-temp.json)
+    Copy-Item -Path (Join-Path $PSScriptRoot xdp.json) -Destination (Join-Path $PackagingDir xdp-temp.json)
```
csujedihy, line 148: Why we would still call it -temp?

nibanks: Why was it called temp to begin with?

csujedihy: I remember because you said this file was temporary until we have an official XDP release so we can remove it or something.

csujedihy: Why do we still need this file with XDP 1.0? I assume now that XDP is official, XDP will guarantee backcompat and we will never need to know which XDP version MsQuic is built on?

mtfriesen: That's a fair question. I think more formalized dependency/package managers tend to check in the precise version they were built against (lock files) even if the logical dependency is less strict, to ensure repeatable builds. Otherwise yes I agree, QUIC should be able to depend on 1.x.

PR #3802, opened 2023-08-10
Configure Codecov to reduce noise

csujedihy:
## Description

Add a config file for codecov so that the workflow won't just fail for every PR when code coverage moves a little bit due to stress test.

Feel free to push a change to modify the threshold and target.

Currently, for the whole project, it compares with the parent commit and fails if the coverage is down more than 2% (threshold).

For the diff in a PR/commit, I currently set the target to 80% with a 2% threshold.

## Testing

CI

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,9 @@
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 2%
+    patch:
+      target: 80%
+      threshold: 2%


nibanks:
Didn't seem to work? Still see codecov as failing.

csujedihy:
> Didn't seem to work? Still see codecov as failing.

I think it needs to be merged to take effect.

nibanks (approved):

PR #3815, opened 2023-08-12
Enable logs to get the code coverage

nibanks:
## Description

There are a lot of lines not executed in packet.c and frame.c that are related to logging. Enable those logs during code coverage.

## Testing

Automation

## Documentation

N/A


ami-GS (approved):
awesome catch

PR #3817, opened 2023-08-16
Phase 2 of enabling new Stream Frame: Reliable Reset

ProjectsByJackHe:
## Description

Reliable Reset is a new QUIC extension in the works with a draft RFC: https://datatracker.ietf.org/doc/draft-ietf-quic-reliable-stream-reset/. 

This PR implements the current spec as outlined by the IETF. This includes implementing the new CLOSE_STREAM frame, along with the necessary metadata changes, and updates to the SEND and RECV paths to ensure sender / receiver complies with the protocol extension.

The API changes involves adding a new SetParam the user can configure, called ReliableOffsetSend, which determines the minimum amount of bytes that must be delivered before a stream can be aborted. 

## Testing

A new Data Test was added to check the behavior of the API. 
A new Events Test was added to validate the receipt of events upon a reliable reset. 
A new API test was added to check the behavior of the Get/Set Parameters.

## Documentation

Documentation in Settings.md and StreamShutdown.md was added to outline usage of the preview feature.


src/inc/msquic.h
```diff
@@ -1473,6 +1473,20 @@ QUIC_STATUS
...
 
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+//
+// Shuts the stream down reliably using error code 0x21,
+// which will guarantee the transmission of "ReliableSize" bytes before aborting the stream.
+//
+typedef
+_IRQL_requires_max_(DISPATCH_LEVEL)
+QUIC_STATUS
+(QUIC_API * QUIC_STREAM_SHUTDOWN_RELIABLE_FN)(
```
nibanks, line 1484: Instead of a new API, let's use a SetParam parameter to configure a reliable reset offset. Then you just use the normal shutdown function.

ProjectsByJackHe: sounds good

src/core/stream.c
```diff
@@ -70,6 +70,8 @@ QuicStreamInitialize(
     Stream->RefCount = 1;
     Stream->SendRequestsTail = &Stream->SendRequests;
     Stream->SendPriority = QUIC_STREAM_PRIORITY_DEFAULT;
+    Stream->ReliableOffsetRecv = 0;
+    Stream->ReliableOffsetSend = 0;
```
nibanks, line 74: Unnecessary because of the `CxPlatZeroMemory(Stream, sizeof(QUIC_STREAM));` line above.

nibanks: Notice we only initialize non-zero variables above.

src/core/stream.c
```diff
@@ -659,6 +661,38 @@ QuicStreamParamSet(
         break;
     }
 
+   case QUIC_PARAM_STREAM_RELIABLE_OFFSET_SEND:
+
+        if (BufferLength != sizeof(uint64_t) || Buffer == NULL) {
```
nibanks, line 666: I just noticed that `QUIC_PARAM_STREAM_PRIORITY` above doesn't check for NULL. Can you fix that please?

src/core/stream.c
```diff
@@ -659,6 +661,38 @@ QuicStreamParamSet(
         break;
     }
 
+   case QUIC_PARAM_STREAM_RELIABLE_OFFSET_SEND:
```
nibanks, line 664: Since the peer won't be setting the receive, only send, let's remove `_SEND` from the name here.

src/core/stream.h
```diff
@@ -363,10 +363,20 @@ typedef struct QUIC_STREAM {
     //
     uint16_t SendPriority;
 
+    //
+    // If > 0, bytes up to offset must be re-transmitted and ACK'd from peer before we can abort this stream.
+    //
+    uint64_t ReliableOffsetSend;
```
nibanks, line 369: For padding (extra space in the struct) reasons you want to put this `uint64_t` after one of the other `uint64_t` above.

src/inc/msquic.h
```diff
@@ -922,6 +922,8 @@ typedef struct QUIC_SCHANNEL_CONTEXT_ATTRIBUTE_EX_W {
 #define QUIC_PARAM_STREAM_IDEAL_SEND_BUFFER_SIZE        0x08000002  // uint64_t - bytes
 #define QUIC_PARAM_STREAM_PRIORITY                      0x08000003  // uint16_t - 0 (low) to 0xFFFF (high) - 0x7FFF (default)
 #define QUIC_PARAM_STREAM_STATISTICS                    0X08000004  // QUIC_STREAM_STATISTICS
+#define QUIC_PARAM_STREAM_RELIABLE_OFFSET_SEND          0x08000005  // uint64_t
```
nibanks, line 925: Wrap in preview #define

src/test/bin/winkernel/control.cpp
```diff
@@ -1167,6 +1167,10 @@ QuicTestCtlEvtIoDeviceControl(
         QuicTestCtlRun(QuicTestStreamAbortConnFlowControl());
         break;
 
+    case IOCTL_QUIC_RUN_STREAM_RELIABLE_RESET:
```
nibanks, line 1170: Always put the new case statement at the end of the others.

nibanks: Also, you'll need to update the big array before this function.

src/core/stream.c
ProjectsByJackHe: For sure!

src/test/bin/winkernel/control.cpp
ProjectsByJackHe: done. will remember to always put new cases at the end.

src/core/stream.c
```diff
@@ -659,6 +661,38 @@ QuicStreamParamSet(
...
+        if (!Stream->Connection->State.ReliableResetStreamNegotiated) {
+            Status = QUIC_STATUS_INVALID_STATE;
+            break;
+        }
+
+        Stream->ReliableOffsetSend = *(uint64_t*)Buffer;
+        Status = QUIC_STATUS_SUCCESS;
+        break;
+
+    case QUIC_PARAM_STREAM_RELIABLE_OFFSET_RECV:
```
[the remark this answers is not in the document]
ProjectsByJackHe: Ahh, that makes more sense. I was wondering how MsQuic could synchronize the receive / send reliableOffsets from App and Peer. I am guessing we set the Receive offset somewhere in the receive path.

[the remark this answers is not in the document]
ProjectsByJackHe: I can update.

src/core/stream.c
ProjectsByJackHe: Done.

src/core/stream.h
ProjectsByJackHe: Can update.

src/core/stream.c
```diff
@@ -659,6 +659,22 @@ QuicStreamParamSet(
...
     }
 
+   case QUIC_PARAM_STREAM_RELIABLE_OFFSET:
+
+        if (BufferLength != sizeof(uint64_t) || Buffer == NULL) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+
+        if (!Stream->Connection->State.ReliableResetStreamNegotiated || *(uint64_t*)Buffer > Stream->QueuedSendOffset) {
```
nibanks, line 669: Generally, we try to keep lines under 80 characters long.
```suggestion
        if (!Stream->Connection->State.ReliableResetStreamNegotiated ||
            *(uint64_t*)Buffer > Stream->QueuedSendOffset) {
```

src/test/lib/DataTest.cpp
```diff
@@ -2960,6 +2960,116 @@ QuicTestStreamAbortConnFlowControl(
...
+
+    */
+    MsQuicRegistration Registration(true);
+    TEST_TRUE(Registration.IsValid());
+
+    MsQuicSettings ServerSettings;
+    MsQuicSettings ClientSettings;
+    ServerSettings.SetReliableResetEnabled(true);
+    ClientSettings.SetReliableResetEnabled(true);
+    ServerSettings.SetPeerBidiStreamCount(1);
```
nibanks, line 3027: To simplify the test, just use the same settings for both.

src/test/lib/DataTest.cpp
```diff
@@ -2960,6 +2960,116 @@ QuicTestStreamAbortConnFlowControl(
...
+    TEST_QUIC_SUCCEEDED(ServerConfiguration.GetInitStatus());
+
+    MsQuicConfiguration ClientConfiguration(Registration, "MsQuicTest", ClientSettings, MsQuicCredentialConfig());
+    TEST_QUIC_SUCCEEDED(ClientConfiguration.GetInitStatus());
+
+    StreamReliableReset Context;
+    #define BUFFER_SIZE 10000
+    #define RELIABLE_SIZE 1000
+    uint8_t SendDataBuffer[BUFFER_SIZE];
+    uint8_t ReceiveDataBuffer[BUFFER_SIZE];
```
nibanks, line 3039: You don't need `ReceiveDataBuffer` any more, right?

src/test/lib/DataTest.cpp
```diff
@@ -2960,6 +2960,116 @@ QuicTestStreamAbortConnFlowControl(
...
+    TEST_QUIC_SUCCEEDED(Listener.GetInitStatus());
+    TEST_QUIC_SUCCEEDED(Listener.Start("MsQuicTest"));
+    QuicAddr ServerLocalAddr;
+    TEST_QUIC_SUCCEEDED(Listener.GetLocalAddr(ServerLocalAddr));
+
+    MsQuicConnection Connection(Registration);
+    TEST_QUIC_SUCCEEDED(Connection.GetInitStatus());
+
+    MsQuicStream Stream(Connection, QUIC_STREAM_OPEN_FLAG_NONE, CleanUpManual, StreamReliableReset::ClientStreamCallback, &Context);
+    TEST_QUIC_SUCCEEDED(Stream.GetInitStatus());
```
nibanks, line 3055: No need to open here. Just move down to where you use the stream, so we can keep all the stream code together.

src/test/lib/DataTest.cpp
```diff
@@ -2960,6 +2960,116 @@ QuicTestStreamAbortConnFlowControl(
...
+
+    TEST_QUIC_SUCCEEDED(Connection.Start(ClientConfiguration, ServerLocalAddr.GetFamily(), QUIC_TEST_LOOPBACK_FOR_AF(ServerLocalAddr.GetFamily()), ServerLocalAddr.GetPort()));
+    TEST_TRUE(Connection.HandshakeCompleteEvent.WaitTimeout(TestWaitTimeout));
+    TEST_TRUE(Connection.HandshakeComplete);
+    TEST_TRUE(Listener.LastConnection->HandshakeCompleteEvent.WaitTimeout(TestWaitTimeout));
+    TEST_TRUE(Listener.LastConnection->HandshakeComplete);
+    CxPlatSleep(50);
```
nibanks, line 3062:
```suggestion
    CxPlatSleep(50); // Wait for things to idle out
```

src/test/lib/DataTest.cpp
```diff
@@ -2960,6 +2960,116 @@ QuicTestStreamAbortConnFlowControl(
     TEST_TRUE(Context.ClientStreamShutdownComplete.WaitTimeout(TestWaitTimeout));
 }
 
+struct StreamReliableReset {
```
nibanks, line 2963: Go ahead and put this inside the function below.

src/test/lib/DataTest.cpp
```diff
@@ -2960,6 +2960,116 @@ QuicTestStreamAbortConnFlowControl(
...
+    static QUIC_STATUS ConnCallback(_In_ MsQuicConnection*, _In_opt_ void* Context, _Inout_ QUIC_CONNECTION_EVENT* Event) {
+        if (Event->Type == QUIC_CONNECTION_EVENT_PEER_STREAM_STARTED) {
+            new(std::nothrow) MsQuicStream(Event->PEER_STREAM_STARTED.Stream, CleanUpAutoDelete, ServerStreamCallback, Context);
+        }
+        return QUIC_STATUS_SUCCESS;
+    }
+};
+
+void
+QuicTestStreamReliableReset(
```
nibanks, line 2996: Generally, try to put new tests at the end of the file.

src/test/lib/DataTest.cpp
```diff
@@ -2960,6 +2960,116 @@ QuicTestStreamAbortConnFlowControl(
...
+     - artificially delay receiver callback
+     - Peer Event Started
+     - Theoretically you can block receive event...
+
+     Quic Send Flag: QUIC_SEND_FLAG_DELAY_FLAG (not always happens)
+      - Here's a send, you don't have to send it yet.
+      - wait for handshake, sleep(50)
+      - start, send (with delay), (set reliable offset), shutdown (do everything after handshake completes)
+
+    */
```
nibanks, line 3019: When you're done, I think you can remove most of this comment.

src/test/lib/DataTest.cpp
ProjectsByJackHe: Good catch! Removed.

src/test/lib/DataTest.cpp
ProjectsByJackHe: Sounds good. Moved code around.

src/test/lib/DataTest.cpp
ProjectsByJackHe: Sounds good.

src/test/lib/DataTest.cpp
ProjectsByJackHe: Sounds good.

src/test/lib/DataTest.cpp
ProjectsByJackHe: Sounds good.

src/core/connection.c
```diff
@@ -4667,6 +4667,7 @@ QuicConnRecvFrames(
         }
 
         case QUIC_FRAME_RESET_STREAM:
+        case QUIC_FRAME_RELIABLE_RESET_STREAM:
```
nibanks, line 4670: nit: Generally, try to keep `case` statements in order as much as possible. So please put this one after `QUIC_FRAME_STREAM_DATA_BLOCKED`.

src/core/stream.c
```diff
@@ -79,6 +79,8 @@ QuicStreamInitialize(
     Stream->RefTypeCount[QUIC_STREAM_REF_APP] = 1;
 #endif
 
+    Stream->Flags.ShutdownReliableSend = FALSE;
+
```
nibanks, line 83: Don't need because struct is zero initialized.

src/core/stream.c
```diff
@@ -659,6 +661,23 @@ QuicStreamParamSet(
...
+            break;
+        }
+
+        if (!Stream->Connection->State.ReliableResetStreamNegotiated ||
+            *(uint64_t*)Buffer > Stream->QueuedSendOffset) {
+            Status = QUIC_STATUS_INVALID_STATE;
+            break;
+        }
+
+        Stream->ReliableOffsetSend = *(uint64_t*)Buffer;
```
nibanks, line 677: We should add a log too. Something like:
```c
QuicTraceLogStreamInfo(
    ReliableSendOffsetSet,
    Stream,
    "Reliable send offset set to %llu",
    *(uint64_t*)Buffer);
```

src/core/stream_recv.c
```diff
@@ -162,6 +162,39 @@ QuicStreamRecvQueueFlush(
...
+    _In_ QUIC_VAR_INT ReliableOffset
+    )
+{
+    if (Stream->ReliableOffsetRecv == 0 || ReliableOffset < Stream->ReliableOffsetRecv) {
+        //
+        // As outlined in the spec, if we receive multiple CLOSE_STREAM frames, we only accept strictly
+        // decreasing offsets.
+        //
+        Stream->ReliableOffsetRecv = ReliableOffset; // TODO: Remove ReliableOffsetRecv once everything works.
+        Stream->RecvMaxLength = ReliableOffset;
```
nibanks, line 183: We should add a log too. Something like:
```c
QuicTraceLogStreamInfo(
    ReliableRecvOffsetSet,
    Stream,
    "Reliable recv offset set to %llu",
    ReliableOffset);
```

src/core/stream.h
```diff
@@ -152,6 +152,8 @@ typedef union QUIC_STREAM_FLAGS {
 
         BOOLEAN InStreamTable           : 1;    // The stream is currently in the connection's table.
         BOOLEAN DelayIdFcUpdate         : 1;    // Delay stream ID FC updates to StreamClose.
+
+        BOOLEAN ShutdownReliableSend    : 1;    // Indicates that we should shutdown the send path once we sent/ACK'd ReliableOffsetSend bytes.
```
nibanks, line 156: Put this right after `LocalCloseReset` and call it (for now) `LocalCloseResetReliable`.

src/core/stream_recv.c
```diff
@@ -162,6 +162,39 @@ QuicStreamRecvQueueFlush(
...
+    }
+
+
```
nibanks, line 186: nit: only one newline space between scopes

src/core/stream_recv.c
```diff
@@ -162,6 +162,39 @@ QuicStreamRecvQueueFlush(
...
+    if (Stream->RecvBuffer.BaseOffset >= Stream->RecvMaxLength) {
+        //
+        // All data delivered. Queue flush.
+        //
+        Stream->Flags.ReceiveDataPending = TRUE;
+        QuicStreamRecvQueueFlush(
```
nibanks, line 192: You don't need to queue a recv flush. Instead just probably deliver the shutdown callback.

src/core/stream_send.c
```diff
@@ -159,7 +159,7 @@ QuicStreamSendShutdown(
             QUIC_STREAM_SEND_FLAG_FIN,
             DelaySend);
 
-    } else {
+    } else if (Stream->ReliableOffsetSend == 0) {
 
         //
```
nibanks, line 164: If `Stream->Flags.LocalCloseReset` is already set, then we've already sent a (non-reliable) reset, and there is no need to do anything more.

src/core/stream.c
```diff
@@ -659,6 +661,23 @@ QuicStreamParamSet(
...
     }
 
+   case QUIC_PARAM_STREAM_RELIABLE_OFFSET:
+
+        if (BufferLength != sizeof(uint64_t) || Buffer == NULL) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+
+        if (!Stream->Connection->State.ReliableResetStreamNegotiated ||
```
nibanks, line 671: Let's also fail if the `Stream->Flags.LocalCloseReset` is already set, indicating that the peer aborted the send before setting the offset.

src/core/stream_send.c
```diff
@@ -234,6 +234,26 @@ QuicStreamSendShutdown(
                 QUIC_STREAM_SEND_FLAG_OPEN |
                 QUIC_STREAM_SEND_FLAG_FIN);
         }
+    } else {
+        uint64_t CurrOffset = Stream->UnAckedOffset;
```
nibanks, line 238: If `Stream->Flags.LocalCloseReset` is already set, then we've already sent a (non-reliable) reset, and there is no need to do anything more.

src/inc/msquicp.h
```diff
@@ -134,6 +134,10 @@ typedef struct QUIC_PRIVATE_TRANSPORT_PARAMETER {
 #define QUIC_PARAM_CONN_KEEP_ALIVE_PADDING              0x85000003  // uint16_t
 #define QUIC_PARAM_CONN_DISABLE_VNE_TP_GENERATION       0x85000004  // BOOLEAN
 
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+#define QUIC_PARAM_STREAM_RELIABLE_OFFSET_RECV          0x08000006  // uint64_t
```
nibanks, line 138:
```suggestion
#define QUIC_PARAM_STREAM_RELIABLE_OFFSET_RECV          0x88000000  // uint64_t
```
Private ones have the high bit set (i.e. the `8` at the beginning). And it's a different number space from non-private, so the first one here starts with 0 at the low bits.

src/test/lib/DataTest.cpp
```diff
@@ -8,12 +8,13 @@
     Tests various features related to the data path.
 
 --*/
-
+// remove
+#include <iostream>
+// remove.
```
nibanks, line 13: Just adding a note to remember to delete this.

src/core/connection.c
ProjectsByJackHe: Sounds good, I'll remember to override my intuition of putting reliable_reset and reset together haha

src/core/stream.c
ProjectsByJackHe: Sounds good

src/core/stream.c
ProjectsByJackHe: Added

src/core/stream_recv.c
ProjectsByJackHe: Added

src/core/stream.h
ProjectsByJackHe: Done

src/core/stream.h
```diff
@@ -410,6 +418,11 @@ typedef struct QUIC_STREAM {
     //
     uint64_t RecvInlineCompletionLength;
 
+    //
+    // If > 0, bytes up to offset must be received from peer before we can abort this stream.
+    //
+    uint64_t ReliableOffsetRecv;
```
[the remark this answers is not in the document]
ProjectsByJackHe: Sounds good. Updated.

src/core/stream_recv.c
ProjectsByJackHe: Added, and I also added an event indication.

src/core/stream.c
ProjectsByJackHe: Done.

src/core/stream_send.c
```diff
@@ -234,6 +234,26 @@ QuicStreamSendShutdown(
...
+        while (ApiSendRequests != NULL && CurrOffset < Stream->ReliableOffsetSend) {
+            //
+            // Deliver the necessary send requests prior to aborting.
+            //
+            QUIC_SEND_REQUEST* SendRequest = ApiSendRequests;
+            ApiSendRequests = ApiSendRequests->Next;
+            QuicStreamCompleteSendRequest(Stream, SendRequest, TRUE, FALSE);
```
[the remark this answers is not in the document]
ProjectsByJackHe: Yes, good catch! I wanted to re-transmit according to graceful path, but for now I'll just enqueue an abortive frame and let the existing framework handle re-transmits ;)

src/core/stream_send.c
ProjectsByJackHe: Added a check! Also checked for graceful closes as well.

src/core/frame.c
```diff
@@ -1943,6 +1943,11 @@ QuicFrameLog(
         break;
     }
 
+    case QUIC_FRAME_RELIABLE_RESET_STREAM: {
+        // TODO: Add log.
```
nibanks, line 1947: Leaving a note here so you remember to do this.

src/core/stream.c
```diff
@@ -567,7 +567,7 @@ QuicStreamShutdown(
     }
 
     if (!!(Flags & QUIC_STREAM_SHUTDOWN_FLAG_IMMEDIATE) &&
-        !Stream->Flags.ShutdownComplete) {
+        !Stream->Flags.ShutdownComplete && !Stream->Flags.RemoteCloseResetReliable) {
```
nibanks, line 570: Should we return an error instead of just ignoring the IMMEDIATE flag in this case?

src/core/stream.c
```diff
@@ -838,6 +867,39 @@ QuicStreamParamGet(
...
+        }
+        if (Buffer == NULL) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+        if (!Stream->Flags.RemoteCloseResetReliable) {
+            *(uint64_t*) Buffer = 0;
+        } else {
+            *(uint64_t*) Buffer = Stream->RecvMaxLength;
+        }
```
nibanks, line 899:
```suggestion
        if (!Stream->Flags.RemoteCloseResetReliable) {
            *(uint64_t*)Buffer = 0;
        } else {
            *(uint64_t*)Buffer = Stream->RecvMaxLength;
        }
```

src/core/stream_recv.c
```diff
@@ -162,6 +162,76 @@ QuicStreamRecvQueueFlush(
...
+//
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+QuicStreamProcessReliableResetFrame(
+    _In_ QUIC_STREAM* Stream,
+    _In_ QUIC_VAR_INT ErrorCode,
+    _In_ QUIC_VAR_INT ReliableOffset
+    )
+{
+    if (Stream->RecvMaxLength == 0 || ReliableOffset < Stream->RecvMaxLength) {
```
nibanks, line 176: Make sure to error out if we receive the frame, but haven't negotiated the extension.

src/core/stream_recv.c
```diff
@@ -162,6 +162,76 @@ QuicStreamRecvQueueFlush(
...
+
+        QuicTraceLogStreamInfo(
+            ReliableRecvOffsetSet,
+            Stream,
+            "Reliable recv offset set to %llu",
+            ReliableOffset);
+    }
+
+    if (Stream->RecvBuffer.BaseOffset >= Stream->RecvMaxLength) {
+        // TODO: Update this to be the same as ReceiveComplete.
```
nibanks, line 192: Putting a note here to remember this todo.

src/core/stream_recv.c
```diff
@@ -380,14 +450,23 @@ QuicStreamProcessStreamFrame(
...
+    if (EndOffset > Stream->RecvMaxLength && !Stream->Flags.RemoteCloseResetReliable) {
         //
-        // Frame goes past the FIN.
+        // Frame goes past the FIN, and the stream is not reset reliably.
         //
         Status = QUIC_STATUS_INVALID_PARAMETER;
         goto Error;
     }
 
+      if (Stream->RecvBuffer.BaseOffset >= Stream->RecvMaxLength && Stream->Flags.RemoteCloseResetReliable) {
```
nibanks, line 461: I think we might be able to merge this and the previous if with sub-if scopes to reduce the number of checks that have to be made.

src/core/stream_recv.c
```diff
@@ -1062,6 +1155,44 @@ QuicStreamReceiveComplete(
             QUIC_STREAM_SEND_FLAG_MAX_DATA | QUIC_STREAM_SEND_FLAG_RECV_ABORT);
     }
 
+     if (Stream->Flags.RemoteCloseResetReliable && Stream->RecvBuffer.BaseOffset >= Stream->RecvMaxLength && !Stream->Flags.ReceiveClosedReliable) {
+
+        Stream->Flags.ReceiveClosedReliable = TRUE;
+        Stream->Flags.LocalCloseAcked = TRUE;
+        Stream->Flags.RemoteCloseAcked = TRUE;
+
```
nibanks, line 1163: nit: too many new lines

src/core/stream.c
ProjectsByJackHe: Just curious, why don't we throw an error if `Stream->Flags.ShutdownComplete = TRUE`?

src/core/stream_recv.c
ProjectsByJackHe: Noted.

src/core/stream_recv.c
ProjectsByJackHe: Yeah, that makes sense, for performance :)

src/core/stream_recv.c
ProjectsByJackHe: Done.

src/core/stream_recv.c
ProjectsByJackHe: Updated.

src/core/stream_recv.c
ProjectsByJackHe: Updated!

src/core/frame.c
ProjectsByJackHe: Updated.

scripts/iter_command.ps1
```diff
@@ -0,0 +1,34 @@
+
+# For local unit testing, often certain bugs only appear due to a specific ordering of events in the network.
```
nibanks, line 2: Can we instead integrate this into `test.ps1` directly?

src/core/stream_recv.c
```diff
@@ -162,6 +162,79 @@ QuicStreamRecvQueueFlush(
...
+            Stream,
+            "Reliable recv offset set to %llu",
+            ReliableOffset);
+    }
+
+    if (Stream->RecvBuffer.BaseOffset >= Stream->RecvMaxLength) {
+        Stream->Flags.ReceiveClosedReliable = TRUE;
+        Stream->Flags.LocalCloseAcked = TRUE;
+        Stream->Flags.RemoteCloseAcked = TRUE;
+        QuicTraceLogStreamVerbose(
```
nibanks, line 208: First, you don't need to prefix the name with `Stream`. Second, you already have a log about delivering the event below and one above about setting the offset, so I don't think this event is necessary.

src/core/stream_recv.c
```diff
@@ -162,6 +162,79 @@ QuicStreamRecvQueueFlush(
...
+            Stream,
+            "Shutting down stream from Process Reliable Reset Frame. Recv Side.");
+        QUIC_STREAM_EVENT Event;
+        Event.Type = QUIC_STREAM_EVENT_PEER_SEND_SHUTDOWN;
+        QuicTraceLogStreamVerbose(
+            IndicatePeerSendShutdown,
+            Stream,
+            "Indicating QUIC_STREAM_EVENT_PEER_SEND_SHUTDOWN");
+        (void)QuicStreamIndicateEvent(Stream, &Event);
+        QuicStreamRecvShutdown(Stream, TRUE, QUIC_ERROR_NO_ERROR);
```
nibanks, line 219: Why are you passing `QUIC_ERROR_NO_ERROR` here when there is an error being passed in? Also, why are you setting `TRUE` for silent?

src/core/stream_recv.c
ProjectsByJackHe: Sounds good, I can simplify the logic and naming.

src/core/stream_recv.c
ProjectsByJackHe: The prior codepath did exactly the same thing as calling RecvShutdown silently. I can update the error code, though.

scripts/iter_command.ps1
ProjectsByJackHe: I'll look into it!

src/core/stream.c
ProjectsByJackHe: Sounds good, so if an app tries immediately shutting down after calling shutdown reliably, that's an error.

ProjectsByJackHe: MsQuic will return a transport error if the app tries immediately shutting down but it previously initiated or in the process of a reliable reset.

scripts/test.ps1
```diff
@@ -172,7 +177,10 @@ param (
     [switch]$UseQtip = $false,
 
     [Parameter(Mandatory = $false)]
-    [string]$OsRunner = ""
+    [string]$OsRunner = "",
+
+    [Parameter(Mandatory = $false)]
+    [int]$NumIterations = 0
```
nibanks, line 183: Instead of defaulting this to `0`, default to `1` and then always do the loop below, put only print out iteration to console if `NumIterations > 0` to simplify things a bit.

src/test/lib/DataTest.cpp
```diff
@@ -3178,3 +3176,196 @@ QuicTestConnectAndIdleForDestCidChange(
...
+            QUIC_STATUS_INVALID_STATE,
+            Stream.SetReliableOffset(1));
+    }
+}
+
+
+
+void
+QuicTestStreamReliableResetMultipleSends(
+) {
```
nibanks, line 3287:
```suggestion
    )
{
```

src/test/lib/DataTest.cpp
```diff
@@ -3178,3 +3176,196 @@ QuicTestConnectAndIdleForDestCidChange(
...
+        return QUIC_STATUS_SUCCESS;
+    }
+
+    static QUIC_STATUS ConnCallback(_In_ MsQuicConnection*, _In_opt_ void* Context, _Inout_ QUIC_CONNECTION_EVENT* Event) {
+        if (Event->Type == QUIC_CONNECTION_EVENT_PEER_STREAM_STARTED) {
+            new(std::nothrow) MsQuicStream(Event->PEER_STREAM_STARTED.Stream, CleanUpAutoDelete, ServerStreamCallback, Context);
+        }
+        return QUIC_STATUS_SUCCESS;
+    }
+};
```
nibanks, line 3221:
```suggestion
    };
```

src/test/lib/DataTest.cpp
```diff
@@ -3178,3 +3176,196 @@ QuicTestConnectAndIdleForDestCidChange(
...
 }
+
+void
+QuicTestStreamReliableReset(
+    )
+{
+    #define BUFFER_SIZE 10000
+    #define RELIABLE_SIZE 5000
+
+     struct StreamReliableReset {
```
nibanks, line 3187:
```suggestion
    struct StreamReliableReset {
```

src/test/lib/DataTest.cpp
```diff
@@ -3178,3 +3176,196 @@ QuicTestConnectAndIdleForDestCidChange(
...
+
+        // We shouldn't be able to change ReliableSize now that the stream has already been reset.
+        TEST_QUIC_STATUS(
+            QUIC_STATUS_INVALID_STATE,
+            Stream.SetReliableOffset(1));
+    }
+}
+
+
+
```
nibanks, line 3284:
```suggestion

```

scripts/test.ps1
ProjectsByJackHe: Sounds good

src/core/stream_send.c
```diff
@@ -1069,17 +1089,28 @@ QuicStreamSendWrite(
     }
 
     if (Stream->SendFlags & QUIC_STREAM_SEND_FLAG_SEND_ABORT) {
```
nibanks, line 1091: Since the logic really is completely different below, should we have a separate send flag type now?

src/core/stream_send.c
```diff
@@ -1177,6 +1208,13 @@ QuicStreamOnLoss(
         return FALSE;
     }
 
+    if (Stream->Flags.LocalCloseResetReliableAcked && Stream->UnAckedOffset >= Stream->ReliableOffsetSend) {
+        //
+        // Ignore any STREAM frame packet loss if we aborted reliably and
+        // received acks for enough data.
+        return FALSE;
```
nibanks, line 1215:
```suggestion
        //
        // Ignore any STREAM frame packet loss if we aborted reliably and
        // received acks for enough data.
        //
        return FALSE;
```

src/core/stream_send.c
```diff
@@ -1436,6 +1474,10 @@ QuicStreamOnAck(
                     "[strm][%p] Send State: %hhu",
                     Stream,
                     QuicStreamSendGetState(Stream));
+                QuicTraceLogStreamVerbose(
+                    GracefulShutdown,
+                    Stream,
+                    "Shutting down [gracefully] from OnStreamAck. Send Side.");
```
nibanks, line 1480: Is this really needed?

src/core/stream_send.c
```diff
@@ -1495,6 +1554,32 @@ QuicStreamOnAck(
     QuicStreamValidateRecoveryState(Stream);
 }
 
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+QuicStreamCleanupReliableReset(
+    _In_ QUIC_STREAM* Stream
+) {
```
nibanks, line 1561:
```suggestion
    )
{
```

src/core/stream_send.c
```diff
@@ -1508,11 +1593,40 @@ QuicStreamOnResetAck(
...
         QuicStreamIndicateSendShutdownComplete(Stream, FALSE);
         QuicStreamTryCompleteShutdown(Stream);
     }
 }
 
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+QuicStreamOnResetReliableAck(
+    _In_ QUIC_STREAM* Stream
+) {
```
nibanks, line 1609:
```suggestion
    )
{
```

src/core/stream_send.c
```diff
@@ -1508,11 +1593,40 @@ QuicStreamOnResetAck(
             "[strm][%p] Send State: %hhu",
             Stream,
             QuicStreamSendGetState(Stream));
+        QuicTraceLogStreamVerbose(
+            OnResetAck,
+            Stream,
+            "Shutting down stream [abortively] from OnResetAck. Send Side.");
```
nibanks, line 1599: Necessary?

src/core/stream_send.c
```diff
@@ -1495,6 +1554,32 @@ QuicStreamOnAck(
...
+        // Ignore send requests that should have been ACK'd, i.e., < ReliableOffsetSend
+        //
+        if (Req->StreamOffset + Req->TotalLength < Stream->ReliableOffsetSend) {
+            continue;
+        }
+        Stream->SendRequests = Req->Next;
+        if (Stream->SendRequests == NULL) {
+            Stream->SendRequestsTail = &Stream->SendRequests;
+        }
+        QuicStreamCompleteSendRequest(Stream, Req, FALSE, TRUE);
```
nibanks, line 1579: You must complete sends in the order they were queued. This seems to skip over requests still in the queue that are valid, and then cancel the ones after them. We can't do that. You have to wait to cancel these until the others have completed.

Ideally, we should have a test case specifically for this scenario; that requests are completed in the right order.

src/core/stream_send.c
```diff
@@ -1508,11 +1593,40 @@ QuicStreamOnResetAck(
...
+) {
+    Stream->Flags.LocalCloseResetReliableAcked = TRUE;
+    QuicTraceLogStreamVerbose(
+            ResetReliableAck,
+            Stream,
+            "Reset Reliable ACKed in OnResetReliableAck. Send side. UnAckedOffset=%llu, ReliableOffsetSend=%llu",
+            Stream->UnAckedOffset, Stream->ReliableOffsetSend);
+    if (Stream->UnAckedOffset >= Stream->ReliableOffsetSend && !Stream->Flags.LocalCloseAcked) {
+        Stream->Flags.LocalCloseAcked = TRUE;
+        Stream->Flags.RemoteCloseAcked = TRUE;
```
nibanks, line 1618: We must log `StreamSendState` log any time that state changes. Tooling automation uses this.

src/core/stream_send.c
ProjectsByJackHe: Adding the SEND flag I think will make things more explicit.

src/core/stream_send.c
ProjectsByJackHe: I found it useful while debugging with TextAnalysis.

src/core/stream_send.c
ProjectsByJackHe: Noted. I can add StreamSendState.

src/core/stream_send.c
ProjectsByJackHe: Sorry! This bit of code was added while I was figuring out the design of my system. Notice how the code coverage never hits it, as this code only runs when UnAckedOffset >= Reliable, so it is guaranteed to complete send requests in order. I can get rid of the useless code that skips over the always false case.

src/core/stream_send.c
ProjectsByJackHe: Added

src/core/stream_send.c
ProjectsByJackHe: Helped me while debugging with text analysis

src/core/stream.c
```diff
@@ -574,6 +574,14 @@ QuicStreamShutdown(
         // and shutdown complete events now, if they haven't already been
         // delivered.
         //
+        if (Stream->Flags.RemoteCloseResetReliable || Stream->Flags.LocalCloseResetReliable) {
+             QuicTraceLogStreamWarning(
+                BadShutdownImmediate,
+                Stream,
+                "App tried ditching this stream but it still has stuff to do.");
```
nibanks, line 581:
```suggestion
                "Invalid immediate shutdown request (pending reliable reset).");
```

src/core/stream.c
```diff
@@ -574,6 +574,14 @@ QuicStreamShutdown(
...
+            QuicConnTransportError(Stream->Connection, QUIC_ERROR_STREAM_STATE_ERROR);
```
nibanks, line 582: Don't kill the whole connection because of this. Just log the warning and that's it.

src/core/stream.c
```diff
@@ -838,6 +898,39 @@ QuicStreamParamGet(
...
+        if (*BufferLength < sizeof(uint64_t)) {
+            *BufferLength = sizeof(uint64_t);
+            Status = QUIC_STATUS_BUFFER_TOO_SMALL;
+            break;
+        }
+        if (Buffer == NULL) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+        *(uint64_t*) Buffer = Stream->ReliableOffsetSend;
```
nibanks, line 912: Should we return "invalid state" error for this and recv offset if they haven't been set? While zero is understood internally to mean "not set" I think externally it should be treated as an invalid value.

src/core/stream_recv.c
```diff
@@ -162,6 +162,77 @@ QuicStreamRecvQueueFlush(
...
+{
+    if (!Stream->Connection->State.ReliableResetStreamNegotiated) {
+        //
+        // The peer tried to use an exprimental feature without
+        // negotiating first. Kill the connection.
+        //
+        QuicTraceLogStreamWarning(
+            ReliableResetNotNegotiated,
+            Stream,
+            "Tried to use ReliableReset without negotiation.");
```
nibanks, line 184:
```suggestion
            "Received ReliableReset without negotiation.");
```

src/core/stream_recv.c
```diff
@@ -162,6 +162,77 @@ QuicStreamRecvQueueFlush(
...
+        QuicConnTransportError(Stream->Connection, QUIC_ERROR_STREAM_STATE_ERROR);
```
nibanks, line 185: Is this the error the spec says to use?

src/core/stream_recv.c
```diff
@@ -162,6 +162,77 @@ QuicStreamRecvQueueFlush(
...
+    }
+
+    if (Stream->RecvBuffer.BaseOffset >= Stream->RecvMaxLength) {
+        Stream->Flags.LocalCloseAcked = TRUE;
+        Stream->Flags.RemoteCloseAcked = TRUE;
+        QuicTraceEvent(
+            StreamRecvState,
+            "[strm][%p] Recv State: %hhu",
+            Stream,
+            QuicStreamRecvGetState(Stream));
```
nibanks, line 211: I wonder if we should define new `QUIC_STREAM_SEND_STATE` and `QUIC_STREAM_RECV_STATE` for this feature.

src/core/stream_recv.c
```diff
@@ -162,6 +162,77 @@ QuicStreamRecvQueueFlush(
...
+        QuicTraceLogStreamVerbose(
```
nibanks, line 212: I think if we update the send/recv state stuff, these extra events, while helpful for debugging won't be necessary any more.

src/core/stream_recv.c
```diff
@@ -162,6 +162,77 @@ QuicStreamRecvQueueFlush(
...
+        //
+        QUIC_STREAM_EVENT Event;
+        Event.Type = QUIC_STREAM_EVENT_PEER_SEND_ABORTED;
+        Event.PEER_SEND_ABORTED.ErrorCode = ErrorCode;
+        QuicTraceLogStreamVerbose(
+            IndicatePeerSendAbort,
+            Stream,
+            "Indicating QUIC_STREAM_EVENT_PEER_SEND_ABORTED (0x%llX)",
+            ErrorCode);
+        (void)QuicStreamIndicateEvent(Stream, &Event);
```
nibanks, line 232: There are so many changes with events around this, I am thinking we probably need a new test case in `QuicTestValidateStreamEvents` for what events happen, and in what order, when using this.

src/core/stream_recv.c
```diff
@@ -257,6 +328,10 @@ QuicStreamProcessResetFrame(
             (void)QuicStreamIndicateEvent(Stream, &Event);
         }
 
+        QuicTraceLogStreamVerbose(
```
nibanks, line 331: This is already handled by the StreamRecvState event.

src/core/stream_recv.c
```diff
@@ -1046,6 +1144,10 @@ QuicStreamReceiveComplete(
             "Indicating QUIC_STREAM_EVENT_PEER_SEND_SHUTDOWN");
         (void)QuicStreamIndicateEvent(Stream, &Event);
 
+        QuicTraceLogStreamVerbose(
```
nibanks, line 1147: `StreamRecvState` handles this already.

src/core/stream_recv.c
```diff
@@ -1060,6 +1162,20 @@ QuicStreamReceiveComplete(
...
             QUIC_STREAM_SEND_FLAG_MAX_DATA | QUIC_STREAM_SEND_FLAG_RECV_ABORT);
+    } else if (Stream->Flags.RemoteCloseResetReliable && Stream->RecvBuffer.BaseOffset >= Stream->RecvMaxLength) {
+        Stream->Flags.LocalCloseAcked = TRUE;
+        Stream->Flags.RemoteCloseAcked = TRUE;
+        QuicTraceEvent(
+            StreamRecvState,
+            "[strm][%p] Recv State: %hhu",
+            Stream,
+            QuicStreamRecvGetState(Stream));
+        QuicTraceLogStreamVerbose(
```
nibanks, line 1173: `StreamRecvState` handles this already

src/test/lib/DataTest.cpp
```diff
@@ -3178,3 +3176,195 @@ QuicTestConnectAndIdleForDestCidChange(
         }
     }
 }
+
+void
+QuicTestStreamReliableReset(
+    )
+{
+    #define BUFFER_SIZE 10000
+    #define RELIABLE_SIZE 5000
```
nibanks, line 3185: Since these are used beyond the scope of just this function, let's move them out. Also, we should test even bigger values (larger than 64k), which means we need to dynamically allocate the buffers (not on the stack) because kernel will crash if the stack usage is that big.

src/test/lib/DataTest.cpp
```diff
@@ -3178,3 +3176,195 @@ QuicTestConnectAndIdleForDestCidChange(
...
+        TEST_QUIC_STATUS(QUIC_STATUS_INVALID_STATE, Stream.SetReliableOffset(RELIABLE_SIZE));
+        // Should behave similar to QUIC_STREAM_SHUTDOWN_FLAG_GRACEFUL, with some restrictions.
+        TEST_TRUE(Context.ClientStreamShutdownComplete.WaitTimeout(TestWaitTimeout));
+        TEST_TRUE(Context.ServerStreamShutdownComplete.WaitTimeout(TestWaitTimeout));
+        TEST_TRUE(Context.ReceivedBufferSize >= RELIABLE_SIZE);
+
+        // We shouldn't be able to change ReliableSize now that the stream has already been reset.
+        TEST_QUIC_STATUS(
+            QUIC_STATUS_INVALID_STATE,
+            Stream.SetReliableOffset(1));
```
nibanks, line 3280:
```suggestion
        TEST_QUIC_STATUS(QUIC_STATUS_INVALID_STATE, Stream.SetReliableOffset(1));
```

src/test/lib/DataTest.cpp
```diff
@@ -3178,3 +3176,195 @@ QuicTestConnectAndIdleForDestCidChange(
...
+    TEST_TRUE(Listener.LastConnection->HandshakeCompleteEvent.WaitTimeout(TestWaitTimeout));
+    TEST_TRUE(Listener.LastConnection->HandshakeComplete);
+    CxPlatSleep(50); // Wait for things to idle out
+
+    MsQuicStream Stream(Connection, QUIC_STREAM_OPEN_FLAG_NONE, CleanUpManual, StreamReliableReset::ClientStreamCallback, &Context);
+    TEST_QUIC_SUCCEEDED(Stream.GetInitStatus());
+    TEST_QUIC_SUCCEEDED(Stream.Start());
+    TEST_QUIC_SUCCEEDED(Stream.Send(&SendBuffer, 1, QUIC_SEND_FLAG_DELAY_SEND, &Context));
+    TEST_QUIC_SUCCEEDED(Stream.Send(&SendBuffer, 1, QUIC_SEND_FLAG_DELAY_SEND, &Context));
+    TEST_QUIC_SUCCEEDED(Stream.Send(&SendBuffer, 1, QUIC_SEND_FLAG_DELAY_SEND, &Context));
```
nibanks, line 3363: You should explicitly verify (1) order of completion and (2) if a send is completed successfully or canceled.

ProjectsByJackHe: Implemented proposal as discussed. Leaving this open, but treating as resolved for now.

src/core/stream.c
ProjectsByJackHe: Sounds good.

src/core/stream_recv.c
ProjectsByJackHe: I see, I can remove!

src/core/stream_recv.c
ProjectsByJackHe: I can remove!

src/core/stream_recv.c
ProjectsByJackHe: I can remove!

src/core/stream_recv.c
ProjectsByJackHe: I can remove!

src/core/stream_recv.c
ProjectsByJackHe: But I agree, QUIC_ERROR_TRANSPORT_ERROR is a better error code in our list of available error codes.

src/core/stream_recv.c
ProjectsByJackHe: That would be a great idea, actually. Thanks for pointing that out!

ProjectsByJackHe: Added.

src/core/stream.c
ProjectsByJackHe: Correct if I am wrong, we agreed adding the GetParam for (send / receive) here is for internal testing purposes only (Added specifically for logging). Externally, apps don't really care about these internal details. Also, since I added GetParam tests in ApiTest, adding this requirement would complicate the logic a lot.

nibanks: This is a public, not private interface. While it's mainly for set, it is available for query, so we have to decide what we want the official behavior to be. What should the result of querying this offset be if the app never set it? Does it make sense to indicate `0` to the app?

src/core/stream.h
```diff
@@ -471,6 +489,8 @@ QuicStreamRecvGetState(
 {
     if (Stream->Flags.RemoteNotAllowed) {
         return QUIC_STREAM_RECV_DISABLED;
+    } else if (Stream->Flags.RemoteCloseResetReliable) {
+        return QUIC_STREAM_RECV_RELIABLE_RESET;
     } else if (Stream->Flags.RemoteCloseReset) {
```
nibanks, line 494: Does receiving a normal reset override the reliable reset state? If so, then it should be first in the `if`. And we need to make sure our logic elsewhere agrees.

src/core/stream.h
```diff
@@ -448,6 +460,12 @@ QuicStreamSendGetState(
 {
     if (Stream->Flags.LocalNotAllowed) {
         return QUIC_STREAM_SEND_DISABLED;
+    } else if (Stream->Flags.LocalCloseResetReliable) {
+        if (Stream->Flags.LocalCloseResetReliableAcked) {
+            return QUIC_STREAM_SEND_RELIABLE_RESET_ACKED;
+        } else {
+            return QUIC_STREAM_SEND_RELIABLE_RESET;
```
nibanks, line 467: Why is this line never hit in tests? Is this state actually valid?

src/core/stream_send.c
```diff
@@ -1477,6 +1523,23 @@ QuicStreamOnAck(
...
+    //
+    BOOLEAN ReliableResetShutdown = !Stream->Flags.LocalCloseAcked && Stream->Flags.LocalCloseResetReliableAcked && Stream->UnAckedOffset >= Stream->ReliableOffsetSend;
+    if (ReliableResetShutdown) {
+        Stream->Flags.LocalCloseAcked = TRUE;
+        Stream->Flags.RemoteCloseAcked = TRUE;
+        QuicTraceEvent(
+                StreamSendState,
+                "[strm][%p] Send State: %hhu",
+                Stream,
+                QuicStreamSendGetState(Stream));
```
nibanks, line 1537:
```suggestion
        QuicTraceEvent(
            StreamSendState,
            "[strm][%p] Send State: %hhu",
            Stream,
            QuicStreamSendGetState(Stream));
```
Too many spaces

src/core/stream_send.c
```diff
@@ -1477,6 +1523,23 @@ QuicStreamOnAck(
         }
     }
 
+    //
+    // If this stream has been reset reliably, we only close if we have received enough bytes.
+    //
+    BOOLEAN ReliableResetShutdown = !Stream->Flags.LocalCloseAcked && Stream->Flags.LocalCloseResetReliableAcked && Stream->UnAckedOffset >= Stream->ReliableOffsetSend;
```
nibanks, line 1529: Line is too long.
```suggestion
    const BOOLEAN ReliableResetShutdown =
        !Stream->Flags.LocalCloseAcked &&
        Stream->Flags.LocalCloseResetReliableAcked &&
        Stream->UnAckedOffset >= Stream->ReliableOffsetSend;
```

src/core/stream_send.c
```diff
@@ -1513,6 +1597,32 @@ QuicStreamOnResetAck(
...
+QuicStreamOnResetReliableAck(
+    _In_ QUIC_STREAM* Stream
+    )
+{
+    Stream->Flags.LocalCloseResetReliableAcked = TRUE;
+    QuicTraceLogStreamVerbose(
+            ResetReliableAck,
+            Stream,
+            "Reset Reliable ACKed in OnResetReliableAck. Send side. UnAckedOffset=%llu, ReliableOffsetSend=%llu",
+            Stream->UnAckedOffset, Stream->ReliableOffsetSend);
```
nibanks, line 1611:
```suggestion
    QuicTraceLogStreamVerbose(
        ResetReliableAck,
        Stream,
        "Reset Reliable ACKed in OnResetReliableAck. Send side. UnAckedOffset=%llu, ReliableOffsetSend=%llu",
        Stream->UnAckedOffset, Stream->ReliableOffsetSend);
```

src/core/stream_send.c
```diff
@@ -1513,6 +1597,32 @@ QuicStreamOnResetAck(
...
+    if (Stream->UnAckedOffset >= Stream->ReliableOffsetSend && !Stream->Flags.LocalCloseAcked) {
+        Stream->Flags.LocalCloseAcked = TRUE;
+        Stream->Flags.RemoteCloseAcked = TRUE;
+        QuicTraceEvent(
+                StreamSendState,
+                "[strm][%p] Send State: %hhu",
+                Stream,
+                QuicStreamSendGetState(Stream));
```
nibanks, line 1619:
```suggestion
        QuicTraceEvent(
            StreamSendState,
            "[strm][%p] Send State: %hhu",
            Stream,
            QuicStreamSendGetState(Stream));
```

src/core/stream.h
ProjectsByJackHe: This state is when we have set reliableOffset and called shutdown, but awaiting the receiver to acknowledge that we are reliably shutting down. Kind of like an intermediary state.

src/core/stream.h
ProjectsByJackHe: I thought initially we couldn't reset, but it looks like reset frames according to the protocol are allowed, I can update.

src/core/stream_send.c
```diff
@@ -1531,16 +1531,14 @@ QuicStreamOnAck(
...
-        Stream->Flags.RemoteCloseAcked = TRUE;
         QuicTraceEvent(
             StreamSendState,
             "[strm][%p] Send State: %hhu",
             Stream,
             QuicStreamSendGetState(Stream));
         QuicStreamCleanupReliableReset(Stream);
         Stream->Flags.LocalCloseReset = TRUE;
-        QuicStreamSendShutdown(Stream, FALSE, TRUE, FALSE, 0x21);
+        QuicStreamSendShutdown(Stream, FALSE, TRUE, FALSE, Stream->SendShutdownErrorCode);
```
nibanks, line 1541: You should update the tests to use different error codes for different scenarios and ensure the proper ones are actually received by the peer.

src/test/lib/DataTest.cpp
```diff
@@ -8,12 +8,10 @@
     Tests various features related to the data path.
 
 --*/
-
```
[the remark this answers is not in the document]
ProjectsByJackHe: Whoops

src/core/stream_send.c
ProjectsByJackHe: Added tests AND a new Event to alert the app that the stream has been aborted reliably for extra clarity.

src/core/stream.c
ProjectsByJackHe: Invalid if not set sounds good

src/inc/msquic.h
```diff
@@ -1371,6 +1374,9 @@ typedef enum QUIC_STREAM_EVENT_TYPE {
     QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE         = 7,
     QUIC_STREAM_EVENT_IDEAL_SEND_BUFFER_SIZE    = 8,
     QUIC_STREAM_EVENT_PEER_ACCEPTED             = 9,
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+    QUIC_STREAM_EVENT_PEER_RELIABLE_ABORT_SEND  = 10,
+#endif
```
nibanks, line 1379: You don't actually need to create a new event. Just use the normal abort after you've reliably delivered the necessary bytes.

docs/api/StreamShutdown.md
```diff
@@ -54,6 +54,11 @@ The stream can also be gracefully shutdown via the `QUIC_SEND_FLAG_FIN` flag. Se
 
 Any stream (even one that hasn't been started) may be called to shutdown. If the stream has not been started yet, then the shutdown is effectively queued. If the app never calls [StreamStart](StreamStart.md) then the shutdown will never been sent out on the wire.
 
+# Reliable Reset
+
+If an app decides to enable Preview Features, the shutdown path can be configured with the QUIC_PARAM_STREAM_RELIABLE_OFFSET Stream
+Parameter, which determines the amount of bytes a sender must deliver before it can shutdown their SEND path.
```
nibanks, line 60:
```suggestion
If an app decides to enable Preview Features, the shutdown path can be configured with the QUIC_PARAM_STREAM_RELIABLE_OFFSET Stream parameter, which determines the number of bytes a sender must deliver before it can shut down their SEND path.
```

src/core/stream.h
```diff
@@ -122,6 +122,9 @@ typedef union QUIC_STREAM_FLAGS {
         BOOLEAN LocalNotAllowed         : 1;    // Peer's unidirectional stream.
         BOOLEAN LocalCloseFin           : 1;    // Locally closed (graceful).
         BOOLEAN LocalCloseReset         : 1;    // Locally closed (locally aborted).
+        BOOLEAN LocalCloseResetReliable : 1;    // Indicates that we should shutdown the send path once we sent/ACK'd ReliableOffsetSend bytes.
+        BOOLEAN LocalCloseResetReliableAcked : 1; // Indicates the peer has acknowledged we will stop sending once we sent/ACK'd ReliableOffsetSend bytes.
```
nibanks, line 126: Can this flag be replaced by reusing `LocalCloseAcked`? `LocalCloseAcked && LocalCloseAcked` essentially equals this, right?

src/core/stream_recv.c
```diff
@@ -162,6 +162,83 @@ QuicStreamRecvQueueFlush(
...
+void
+QuicStreamAlertAppPeerSendAborted(
+    _In_ QUIC_STREAM* Stream,
+    _In_ QUIC_VAR_INT ErrorCode
+    )
+{
+    QuicTraceLogStreamInfo(
+                RemoteCloseReset,
+                Stream,
+                "Closed remotely (reset)");
```
nibanks, line 178:
```suggestion
    QuicTraceLogStreamInfo(
        RemoteCloseReset,
        Stream,
        "Closed remotely (reset)");
```

src/core/stream_recv.c
```diff
@@ -1062,6 +1149,19 @@ QuicStreamReceiveComplete(
             QUIC_STREAM_SEND_FLAG_MAX_DATA | QUIC_STREAM_SEND_FLAG_RECV_ABORT);
     }
 
+    else if (Stream->Flags.RemoteCloseResetReliable && Stream->RecvBuffer.BaseOffset >= Stream->RecvMaxLength) {
+        //
+        // ReliableReset
```
nibanks, line 1154: Maybe a little better comment?

src/core/stream_send.c
```diff
@@ -1513,6 +1598,40 @@ QuicStreamOnResetAck(
...
+            QuicStreamSendGetState(Stream));
+        QuicStreamCleanupReliableReset(Stream);
+        Stream->Flags.LocalCloseReset = TRUE;
+        QuicStreamSendShutdown(Stream, FALSE, TRUE, FALSE, Stream->SendShutdownErrorCode);
+    } else {
+        QuicTraceEvent(
+        StreamSendState,
+        "[strm][%p] Send State: %hhu",
+        Stream,
+        QuicStreamSendGetState(Stream));
```
nibanks, line 1630:
```suggestion
        QuicTraceEvent(
            StreamSendState,
            "[strm][%p] Send State: %hhu",
            Stream,
            QuicStreamSendGetState(Stream));
```

src/test/lib/ApiTest.cpp
```diff
@@ -4971,6 +4971,76 @@ void QuicTestStreamParam()
             TEST_EQUAL(Length, sizeof(QUIC_STREAM_STATISTICS));
         }
     }
+
+
```
nibanks, line 4975:
```suggestion

```

src/test/lib/DataTest.cpp
```diff
@@ -3178,3 +3177,209 @@ QuicTestConnectAndIdleForDestCidChange(
         }
     }
 }
+
+#define BUFFER_SIZE 10000
+#define RELIABLE_SIZE 5000
+#define BUFFER_SIZE_MULTI_SENDS 10000
+#define RELIABLE_SIZE_MULTI_SENDS 20000
+
+
```
nibanks, line 3186:
```suggestion

```

src/test/lib/DataTest.cpp
```diff
@@ -3178,3 +3177,209 @@ QuicTestConnectAndIdleForDestCidChange(
...
+};
+struct StreamReliableReset {
+
+    CxPlatEvent ClientStreamShutdownComplete;
+    CxPlatEvent ServerStreamShutdownComplete;
+    uint64_t ReceivedBufferSize;
+    uint64_t SequenceNum;
+    QUIC_UINT62 ShutdownErrorCode;
+
+
```
nibanks, line 3204:
```suggestion

```

src/test/lib/DataTest.cpp
```diff
@@ -3178,3 +3177,209 @@ QuicTestConnectAndIdleForDestCidChange(
...
+    static QUIC_STATUS ClientStreamCallback(_In_ MsQuicStream*, _In_opt_ void* ClientContext, _Inout_ QUIC_STREAM_EVENT* Event) {
+        auto TestContext = (StreamReliableReset*)ClientContext;
+        if (Event->Type == QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE) {
+            TestContext->ClientStreamShutdownComplete.Set();
+        }
+            // Get the send context of the Event
+        if (Event->Type == QUIC_STREAM_EVENT_SEND_COMPLETE) {
```
nibanks, line 3211:
```suggestion
        // Get the send context of the Event
        if (Event->Type == QUIC_STREAM_EVENT_SEND_COMPLETE) {
```
But is this comment really necessary?

src/test/lib/DataTest.cpp
```diff
@@ -3178,3 +3177,209 @@ QuicTestConnectAndIdleForDestCidChange(
...
+        if (Event->Type == QUIC_STREAM_EVENT_PEER_SEND_ABORTED) {
+            TestContext->ShutdownErrorCode = Event->PEER_SEND_ABORTED.ErrorCode;
+        }
+        if (Event->Type == QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE) {
+            TestContext->ServerStreamShutdownComplete.Set();
+        }
+        return QUIC_STATUS_SUCCESS;
+    }
+
+    static QUIC_STATUS ConnCallback(_In_ MsQuicConnection*, _In_opt_ void* Context, _Inout_ QUIC_CONNECTION_EVENT* Event) {
```
nibanks, line 3233: I _think_ we might have an auto-accept callback already that supports this.

src/core/stream.h
ProjectsByJackHe: Not sure, but even if we could I would much prefer the separation, because LocalCloseAcked is used in soooo many other different places we could potentially risk regression if we overextend its usage to this preview feature. The flags here are all internally used, no? An app is not concerned with this stuff.

nibanks: Just because a variable is used in a lot of places doesn't mean is usage is overextended. `LocalCloseAcked` is meant to indicate whatever close mechanism has been used has now been acknowledged. This is a fairly clean design because it effectively means the send direction is 100% done now. Having multiple variables that mean effectively the same thing means you complicated the "are we done yet" logic by requiring more variables to be checked.

src/test/lib/DataTest.cpp
ProjectsByJackHe: To me 1 month ago it was, now it's not necessary.

src/core/stream_recv.c
```diff
@@ -162,6 +162,83 @@ QuicStreamRecvQueueFlush(
...
+    if (Stream->RecvBuffer.BaseOffset >= Stream->RecvMaxLength) {
+        QuicTraceEvent(
+            StreamRecvState,
+            "[strm][%p] Recv State: %hhu",
+            Stream,
+            QuicStreamRecvGetState(Stream));
+        QuicStreamAlertAppPeerSendAborted(Stream, ErrorCode);
+        QuicStreamRecvShutdown(Stream, TRUE, ErrorCode);
+    } else {
+        Stream->RecvShutdownErrorCode = ErrorCode;
```
nibanks, line 238: This is worth a comment along the lines of "we still have data to deliver to the app, just cache the error code for later".

ProjectsByJackHe: Added

src/core/stream_send.c
```diff
@@ -1513,6 +1597,37 @@ QuicStreamOnResetAck(
...
+    _In_ QUIC_STREAM* Stream
+    )
+{
+    QuicTraceLogStreamVerbose(
+        ResetReliableAck,
+        Stream,
+        "Reset Reliable ACKed in OnResetReliableAck. Send side. UnAckedOffset=%llu, ReliableOffsetSend=%llu",
+        Stream->UnAckedOffset, Stream->ReliableOffsetSend);
+
+    if (Stream->UnAckedOffset >= Stream->ReliableOffsetSend && !Stream->Flags.LocalCloseAcked) {
```
nibanks, line 1612: Can we add an assert that the stream hasn't been somehow reset already too? I'm just wondering if this check for `LocalCloseAcked` isn't enough. What if the peer reset their receive before we could get an ACK for reliable reset on send?

ProjectsByJackHe: Resolved from discussions

src/test/MsQuicTests.h
```diff
@@ -544,6 +544,14 @@ void
 QuicTestStreamAbortConnFlowControl(
     );
 
+void
+QuicTestStreamReliableReset(
+    );
+
+void
+QuicTestStreamReliableResetMultipleSends(
+);
```
nibanks, line 553:
```suggestion
void
QuicTestStreamReliableResetMultipleSends(
    );
```

src/test/MsQuicTests.h
```diff
@@ -1216,7 +1224,14 @@ typedef struct {
     QUIC_CTL_CODE(112, METHOD_BUFFERED, FILE_WRITE_DATA)
     // QUIC_RUN_FEATURE_NEGOTIATION
 
+
 #define IOCTL_QUIC_RUN_STATELESS_RESET_KEY \
```
nibanks, line 1228:
```suggestion
#define IOCTL_QUIC_RUN_STATELESS_RESET_KEY \
```

src/test/MsQuicTests.h
```diff
@@ -1216,7 +1224,14 @@ typedef struct {
...
     QUIC_CTL_CODE(113, METHOD_BUFFERED, FILE_WRITE_DATA)
 
-#define QUIC_MAX_IOCTL_FUNC_CODE 113
+  #define IOCTL_QUIC_RUN_STREAM_RELIABLE_RESET \
```
nibanks, line 1231:
```suggestion
#define IOCTL_QUIC_RUN_STREAM_RELIABLE_RESET \
```

src/test/bin/winkernel/control.cpp
```diff
@@ -485,6 +485,8 @@ size_t QUIC_IOCTL_BUFFER_SIZES[] =
     sizeof(QUIC_RUN_FEATURE_NEGOTIATION),
     sizeof(QUIC_RUN_FEATURE_NEGOTIATION),
     0,
+    0,
+    0
```
nibanks, line 489:
```suggestion
    0,
```
Always leave a trailing `,` to make the diff nicer when you add a new line (don't have to modify the old one)

nibanks (commented):
Just a few nits, but otherwise good to go!

nibanks (approved):

nibanks:
Going to merge this, but I did notice one test failure:
```
[ RUN      ] Misc/WithAbortiveArgs.AbortiveShutdown/502
D:\a\msquic\msquic\src\test\lib\DataTest.cpp(1197): error: Client failed to get connected before timeout!

[  FAILED  ] Misc/WithAbortiveArgs.AbortiveShutdown/502, where GetParam() = v6/0/1/0/0/1/2/1/1/0 (3274 ms)
[----------] 1 test from Misc/WithAbortiveArgs (3275 ms total)
```
This test is (rarely) flaky, so hopefully it's just that, but we need to keep an eye out on this going forward. @projectsbyjackhe can you please open a bug and attach the logs for this failure to track?

PR #3819, opened 2023-08-18
Unify datapath

ami-GS:
## Description

Unifying windows normal socket and raw (xdp) socket

## Testing

adjusting tests

## Documentation

Architecture.md and TEST.md


src/test/lib/MtuTest.cpp
```diff
@@ -49,12 +49,13 @@ struct ResetSettings {
 void
 QuicTestMtuSettings()
 {
+    MsQuicRegistration Registration(true);
```
nibanks, line 52: There are a number of changes in the test code that are unnecessary, and I assume accidental. Please remove them if you can.

src/inc/quic_datapath.h
```diff
@@ -281,7 +321,8 @@ typedef struct CXPLAT_QEO_CONNECTION {
 //
 CXPLAT_RECV_DATA*
 CxPlatDataPathRecvPacketToRecvData(
-    _In_ const CXPLAT_RECV_PACKET* const RecvPacket
+    _In_ const CXPLAT_RECV_PACKET* const RecvPacket,
+    _In_ uint16_t BufferFrom
     );
```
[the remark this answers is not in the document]
ami-GS: Found problem, normal socket have additional hidden member, `CXPLAT_DATAPATH_INTERNAL_RECV_BUFFER_CONTEXT`. which makes accessing the flag difficult

src/core/binding.c
```diff
@@ -1150,6 +1150,7 @@ QuicBindingPreprocessDatagram(
     CxPlatZeroMemory(&Packet->PacketNumber, sizeof(QUIC_RX_PACKET) - sizeof(uint64_t));
     Packet->AvailBuffer = Datagram->Buffer;
     Packet->AvailBufferLength = Datagram->BufferLength;
+    Packet->BufferFrom = Datagram->BufferFrom;
```
nibanks, line 1153: These changes are no longer necessary, right?

src/core/binding.h
```diff
@@ -160,6 +160,13 @@ typedef struct QUIC_RX_PACKET {
     // Flag indicating the packet contained a non-probing frame.
     //
     BOOLEAN HasNonProbingFrame : 1;
+
+    // //
```
nibanks, line 164: Also unnecessary?

src/core/path.h
```diff
@@ -302,13 +302,6 @@ QuicConnGetPathForDatagram(
     _In_ const CXPLAT_RECV_DATA* Datagram
     );
 
-_IRQL_requires_max_(PASSIVE_LEVEL)
-void
-QuicCopyRouteInfo(
```
nibanks, line 307: Why was this deleted?

src/inc/quic_datapath.h
```diff
@@ -126,11 +126,31 @@ PacketSizeFromUdpPayloadSize(
...
 typedef struct CXPLAT_SOCKET CXPLAT_SOCKET;
 
+typedef struct CXPLAT_UDP_CONFIG CXPLAT_UDP_CONFIG;
+
+//
+// Can be defined to whatever the client needs.
+//
+typedef struct CXPLAT_RECV_PACKET CXPLAT_RECV_PACKET;
+
+typedef enum CXPLAT_BUFFER_FROM {
```
nibanks, line 146: Let's name this more along the lines of `CXPLAT_DATAPATH_TYPE`.

src/inc/quic_datapath.h
```diff
@@ -126,11 +126,31 @@ PacketSizeFromUdpPayloadSize(
...
 //
+typedef struct CXPLAT_SOCKET_RAW CXPLAT_SOCKET_RAW;
 typedef struct CXPLAT_SOCKET CXPLAT_SOCKET;
 
+typedef struct CXPLAT_UDP_CONFIG CXPLAT_UDP_CONFIG;
+
+//
+// Can be defined to whatever the client needs.
+//
+typedef struct CXPLAT_RECV_PACKET CXPLAT_RECV_PACKET;
```
nibanks, line 144: Remove. No longer needed.

src/platform/CMakeLists.txt
```diff
@@ -12,12 +12,7 @@ endif()
 set(SOURCES crypt.c hashtable.c pcp.c platform_worker.c toeplitz.c)
 
 if("${CX_PLATFORM}" STREQUAL "windows")
-    set(SOURCES ${SOURCES} platform_winuser.c storage_winuser.c)
-    if(QUIC_USE_XDP)
-        set(SOURCES ${SOURCES} datapath_raw_win.c datapath_raw_socket.c datapath_raw_socket_win.c datapath_raw_xdp_win.c)
-    else()
-        set(SOURCES ${SOURCES} datapath_winuser.c)
-    endif()
+    set(SOURCES ${SOURCES} platform_winuser.c storage_winuser.c datapath_win.c datapath_winuser.c datapath_raw_win.c datapath_raw_socket.c datapath_raw_socket_win.c datapath_raw_xdp_win.c)
```
nibanks, line 15: This line is so long, let's put one file per line now.

src/test/lib/DataTest.cpp
```diff
@@ -19,6 +19,7 @@ static bool UseQTIP = false;
 #elif defined(QUIC_API_ENABLE_PREVIEW_FEATURES)
 extern bool UseQTIP;
 #endif
+extern uint64_t LARGE_SEND_SIZE;
```
nibanks, line 22: Unnecessary?

src/test/lib/MtuTest.cpp
```diff
@@ -56,7 +56,6 @@ QuicTestMtuSettings()
 #else
     const uint16_t DefaultMaximumMtu = 1500;
 #endif
-
```
nibanks, line 59: Unnecessary?

src/test/bin/quic_gtest.h
```diff
@@ -8,6 +8,7 @@
 #define QUIC_TEST_APIS 1
 
 #include "quic_platform.h"
+#include "quic_datapath.h"
```
nibanks, line 11: Actually necessary?

src/platform/platform_worker.c
```diff
@@ -212,7 +212,6 @@ CxPlatWorkersLazyStart(
     CxPlatRundownInitialize(&CxPlatWorkerRundown);
 
     CxPlatLockRelease(&CxPlatWorkerLock);
-
```
nibanks, line 215: Unnecessary

src/platform/platform_worker.c
```diff
@@ -469,7 +468,6 @@ CXPLAT_THREAD_CALLBACK(CxPlatWorkerThread, Context)
 {
     CXPLAT_WORKER* Worker = (CXPLAT_WORKER*)Context;
     CXPLAT_DBG_ASSERT(Worker != NULL);
-
```
nibanks, line 472: Unnecessary

src/test/lib/HandshakeTest.cpp
```diff
@@ -3207,7 +3207,7 @@ QuicTestInterfaceBinding(
     MsQuicConnection Connection1(Registration);
     TEST_QUIC_SUCCEEDED(Connection1.GetInitStatus());
     TEST_QUIC_SUCCEEDED(Connection1.SetLocalInterface(LoopbackInterfaceIndex));
-    TEST_QUIC_SUCCEEDED(Connection1.Start(ClientConfiguration, ServerLocalAddr.GetFamily(), QUIC_TEST_LOOPBACK_FOR_AF(ServerLocalAddr.GetFamily()), ServerLocalAddr.GetPort()));
+    TEST_QUIC_SUCCEEDED(Connection1.Start(ClientConfiguration, ServerLocalAddr.GetFamily(), QUIC_LOCALHOST_FOR_AF(ServerLocalAddr.GetFamily()), ServerLocalAddr.GetPort()));
```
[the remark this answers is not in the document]
ami-GS: Previously this fixed issue of hybrid datapath case (forgot what).
However original implementation (sock datapath with --duoNic or xdp datapath) has been skipping or failing. I will add GTEST_SKIP if duonic enabled

ami-GS:
`-Platform uwp` has build issue....
I tried to include header file (netioapi.h) which has type affecting build doesn't solve anything.🤷‍♂️

src/platform/datapath_win.c
```diff
@@ -0,0 +1,615 @@
...
+    if (Datapath->RawDataPath) {
+        return DataPathGetSupportedFeatures(Datapath) |
+               RawDataPathGetSupportedFeatures(Datapath->RawDataPath);
+    }
+    return DataPathGetSupportedFeatures(Datapath);
+}
+
+_IRQL_requires_max_(DISPATCH_LEVEL)
+BOOLEAN
+CxPlatDataPathIsPaddingPreferred(
```
ami-GS, line 120: which should be taken?

src/platform/datapath_win.c
```diff
@@ -0,0 +1,615 @@
...
+}
+
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+QuicCopyRouteInfo(
+    _Inout_ CXPLAT_ROUTE* DstRoute,
+    _In_ CXPLAT_ROUTE* SrcRoute
+    )
+{
+    if (!IS_LOOPBACK(SrcRoute->RemoteAddress)) {
```
csujedihy, line 562: >!IS_LOOPBACK(SrcRoute->RemoteAddress) [](http://example.com/codeflow?start=8&length=37)

How is the route info copy logic coupled to loopback?

src/platform/platform_internal.h
```diff
@@ -104,6 +160,695 @@
...
+
+//
+// Type of IO for XDP.
+//
+typedef enum DATAPATH_XDP_IO_TYPE {
+    DATAPATH_XDP_IO_SIGNATURE         = 'XDPD',
+    DATAPATH_XDP_IO_RECV              = DATAPATH_XDP_IO_SIGNATURE + 1,
+    DATAPATH_XDP_IO_SEND              = DATAPATH_XDP_IO_SIGNATURE + 2
+} DATAPATH_XDP_IO_TYPE;
+
```
csujedihy, line 195: why is this not in _xdp.h?

src/platform/datapath_raw_dummy.c
```diff
@@ -0,0 +1,252 @@
+/*++
```
csujedihy, line 1: What is the purpose of this file?

src/inc/quic_datapath.h
```diff
@@ -589,6 +599,19 @@ CxPlatSocketDelete(
     _In_ CXPLAT_SOCKET* Socket
     );
 
+
+_IRQL_requires_max_(PASSIVE_LEVEL)
+BOOLEAN
+CxPlatRawDataPathAvailable(
```
csujedihy, line 605: >CxPlatRawDataPathAvailable [](http://example.com/codeflow?start=0&length=26)

Is this function used? I can't find the definition.

csujedihy: same as the function below

csujedihy:
>     if (closesocket(Socket->AuxSocket) == SOCKET_ERROR) {

Where's all the logic for the aux socket? Like how we create an aux socket for normal XDP and QTIP? Those seem to have been deleted?

---
Refers to: src/platform/datapath_raw_socket_win.c:65 in 08220d3. [](commit_id = 08220d38279b54e2a38b2cd63e2c646fa9fe0ca2, deletion_comment = False)

src/platform/datapath_raw_dummy.c
csujedihy: Looks like you are trying to get UWP build to work. Can you not use `#ifdef QUIC_UWP_BUILD` to achieve the same goal?

csujedihy:
> `-Platform uwp` has build issue.... I tried to include header file (netioapi.h) which has type affecting build doesn't solve anything.🤷‍♂️

What do you need from netioapi.h for non-kernel datapaths?

src/platform/platform_internal.h
ami-GS: These constants need to be referred in datapath_win.c to switch receive path for normal and xdp

src/platform/datapath_win.c
ami-GS: loopback is not enough to determine the logic. adding one more parameter

src/platform/datapath_raw_dummy.c
ami-GS: You mean all datapath API in `_win.c` to have `ifdef` for UWP?

csujedihy: Not all datapath API usage. I think you only need to use `#ifdef QUIC_UWP_BUILD` (sorry I wrote `ifdef` earlier) at a few places. The most obvious place is the raw datapath initialization. You might still need to put some more pieces of code behind the macro that do not compile under UWP like where we use IP MIB structs or some other functions.

Looking at where we currently use `QUIC_UWP_BUILD`, IIUC, it seems like some functions are exported in onecore.lib (which is what UWP build links against) but the function declaration is missing/not exposed in the UWP kit. (If you search for `#ifdef QUIC_UWP_BUILD`, you will know what I am talking about). I wonder if you can do the same for the failures you were seeing for UWP build.

ami-GS: As far as I saw before, there are a lot of type not found error in case of uwp.
Adding ifdef for such type would varying many places and it end up with building successfully, but the functionality would be gone for uwp even though we don't use raw for uwp. I'm not sure such change makes sense.

csujedihy: I took a quick look at the compile errors in the failed build. Is it just CxPlatGetInterfaceRssQueueCount and RawResolveRoute that cannot compile for UWP? If so, I think just we can just whack them with `ifdef` like this:

```
#ifdef QUIC_UWP_BUILD
QUIC_STATUS
RawResolveRoute(
    _In_ CXPLAT_SOCKET* Sock,
    _Inout_ CXPLAT_ROUTE* Route,
    _In_ uint8_t PathId,
    _In_ void* Context,
    )
{
    return QUIC_SUCCESS;
}
#else
// actual definition
#endif
```

If there are more instances, then maybe a dummy file is better but we need to rename it and make it clear it's for UWP.

csujedihy: "it end up with building successfully, but the functionality would be gone for uwp even though we don't use raw for uwp."
I don't quite understand this sentence. Did you mean we would end up having unreachable code in the binary?

src/core/connection.c
```diff
@@ -182,7 +182,7 @@ QuicConnAlloc(
 
         Connection->Stats.QuicVersion = Packet->Invariant->LONG_HDR.Version;
         QuicConnOnQuicVersionSet(Connection);
-        QuicCopyRouteInfo(&Path->Route, Packet->Route);
+        QuicCopyRouteInfo(&Path->Route, Packet->Route, ((CXPLAT_RECV_DATA*)Packet)->DatapathType);
```
nibanks, line 185: Why don't you save the type in the `Route` if necessary? We should do everything we can to simplify things for the upper layer. It shouldn't have to care about this stuff as much as possible.

src/inc/quic_datapath.h
```diff
@@ -125,12 +125,21 @@ PacketSizeFromUdpPayloadSize(
...
 //
 typedef struct CXPLAT_DATAPATH CXPLAT_DATAPATH;
+typedef struct CXPLAT_DATAPATH_RAW CXPLAT_DATAPATH_RAW;
 
 //
 // Represents a UDP or TCP abstraction.
 //
 typedef struct CXPLAT_SOCKET CXPLAT_SOCKET;
 
+typedef enum CXPLAT_DATAPATH_TYPE {
```
nibanks, line 135: Does the QUIC layer _need_ to know the type? Again, with the goal of hiding all extra complexity from the upper layers, I think this should be defined and used internally only if possible.

src/inc/quic_datapath.h
```diff
@@ -589,6 +600,7 @@ CxPlatSocketDelete(
     _In_ CXPLAT_SOCKET* Socket
     );
 
+
```
nibanks, line 603: nit: please remove extra newline

src/platform/datapath_raw.h
```diff
@@ -273,7 +256,9 @@ typedef struct CXPLAT_SOCKET {
 
     CXPLAT_SEND_DATA* PausedTcpSend; // Paused TCP send data *before* framing
     CXPLAT_SEND_DATA* CachedRstSend; // Cached TCP RST send data *after* framing
-} CXPLAT_SOCKET;
+
+    CXPLAT_SOCKET;
```
nibanks, line 260: Generally, it's best to put these up front, because then it makes casting much easier.

src/platform/datapath_raw_dummy.c
nibanks: I also would prefer and #ifdef approach instead of dummy file.

src/platform/datapath_raw_socket.c
```diff
@@ -43,19 +59,8 @@ CxPlatGetSocket(
...
-    _In_ CXPLAT_ROUTE* SrcRoute
-    )
-{
-    CxPlatCopyMemory(DstRoute, SrcRoute, (uint8_t*)&SrcRoute->State - (uint8_t*)SrcRoute);
-    CxPlatUpdateRoute(DstRoute, SrcRoute);
-}
-
 void
-CxPlatResolveRouteComplete(
+RawResolveRouteComplete(
```
nibanks, line 63: Generally, I'd prefer to have either `CxPlat` prefix all functions (usually only for "public" functions to upper layer) or something like `Datapath` or the object type.

src/platform/datapath_raw_socket.c
```diff
@@ -492,15 +499,15 @@ CxPlatFramingTransportChecksum(
...
+    _In_ CXPLAT_SOCKET_RAW* Socket,
     _In_ CXPLAT_RECV_DATA* Packet
     )
 {
     CXPLAT_DBG_ASSERT(Socket->UseTcp);
 
     CXPLAT_ROUTE* Route = Packet->Route;
     CXPLAT_SEND_CONFIG SendConfig = { Route, 0, CXPLAT_ECN_NON_ECT, 0 };
-    CXPLAT_SEND_DATA *SendData = CxPlatSendDataAlloc(Socket, &SendConfig);
+    CXPLAT_SEND_DATA *SendData = (CXPLAT_SEND_DATA*)CxPlatSendDataAlloc((CXPLAT_SOCKET*)Socket, &SendConfig);
```
nibanks, line 510: What does `CxPlatSendDataAlloc` return? Usually C doesn't make you cast things if it's a `void*`.

src/platform/datapath_raw_win.c
```diff
@@ -102,38 +102,26 @@ CxPlatDataPathRouteWorkerInitialize(
     return Status;
 }
 
+
```
nibanks, line 105: nit: please remove extra newline

src/platform/datapath_raw_win.c
```diff
@@ -280,229 +265,113 @@ CxPlatDataPathGetGatewayAddresses(
...
-    _In_ CXPLAT_DATAPATH* Datapath,
+RawSocketCreateUdp(
+    _In_  CXPLAT_DATAPATH_RAW* Raw,
     _In_ const CXPLAT_UDP_CONFIG* Config,
-    _Out_ CXPLAT_SOCKET** NewSocket
+    _Out_ CXPLAT_SOCKET_RAW* Socket
     )
 {
+    #pragma warning(push)
+    #pragma warning(disable:6001) // Using uninitialized memory
```
nibanks, line 277: Fix this by making `Socket` and `_Inout_`.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -993,17 +993,17 @@ CxPlatDpRawGetDatapathSize(
     return sizeof(XDP_DATAPATH) + (PartitionCount * sizeof(XDP_PARTITION));
 }
 
+
```
nibanks, line 996: nit: please remove extra newline

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1284,14 +1284,18 @@ CxPlatDpRawUpdateConfig(
...
-    _In_ CXPLAT_SOCKET* Socket,
+RawSocketUpdateQeo(
+    _In_ CXPLAT_SOCKET_RAW* Socket,
     _In_reads_(OffloadCount)
         const CXPLAT_QEO_CONNECTION* Offloads,
     _In_ uint32_t OffloadCount
     )
 {
-    XDP_DATAPATH* Xdp = (XDP_DATAPATH*)Socket->Datapath;
+    if (!Socket->RawDatapath) {
```
nibanks, line 1294: Is this an expected path that can/would be taken? Or can this be a debug assert instead?

src/platform/datapath_win.c
```diff
@@ -0,0 +1,628 @@
...
+        Status = QUIC_STATUS_INVALID_PARAMETER;
+        goto Error;
+    }
+
+    Status = DataPathInitialize(
+        ClientRecvContextLength,
+        UdpCallbacks,
+        TcpCallbacks,
+        Config,
+        NewDataPath);
```
nibanks, line 52: We use the following style:
```suggestion
    Status =
        DataPathInitialize(
            ClientRecvContextLength,
            UdpCallbacks,
            TcpCallbacks,
            Config,
            NewDataPath);
```

src/inc/quic_datapath.h
```diff
@@ -446,7 +456,8 @@ CxPlatDataPathGetSupportedFeatures(
 _IRQL_requires_max_(DISPATCH_LEVEL)
 BOOLEAN
 CxPlatDataPathIsPaddingPreferred(
-    _In_ CXPLAT_DATAPATH* Datapath
+    _In_ CXPLAT_DATAPATH* Datapath,
```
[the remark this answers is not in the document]
ami-GS: I don't see Datapath pointer in CXPLAT_SEND_DATA. Does it have implicit members like recv data?

src/platform/datapath_raw.h
ami-GS: Sadly impossible as CXPLAT_SOCKET has variable length member `CXPLAT_SOCKET_PROC PerProcSockets[0]`.
Move the member to `CXPLAT_SOCKET_RAW` from `CXPLAT_SOCKET`?

src/platform/datapath_raw_socket.c
ami-GS: wow, this looks like simply bug....

src/platform/datapath_raw_xdp_win.c
ami-GS: This is not needed anymore... I expected this was kind of debug assert, but covered by wrapper in datapath_win.c

src/platform/datapath_raw_dummy.c
ami-GS: - Not only function using MIB, but also `afxdp_helper.h` depending data (like XSK_RING which is in `XDP_RX_PACKET{ XDP_QUEUE{ XSK_RING } }`) get affected
- Not non unreachable code, but non functional code.

ami-GS: image

ami-GS:
`Binding->Socket` becomes NULL when running .\scripts\spin.ps1
https://github.com/microsoft/msquic/blob/f91ccc2f975225a278bc68238555ed0923e676ba/src/core/binding.c#L1609

src/platform/datapath_raw_socket_win.c
```diff
@@ -61,17 +61,6 @@ CxPlatRemoveSocket(
 {
     CxPlatRwLockAcquireExclusive(&Pool->Lock);
     CxPlatHashtableRemove(&Pool->Sockets, &Socket->Entry, NULL);
-
```
csujedihy, line 64: I don't understand how we can completely remove the use of aux socket. For QTIP, we need a bound TCP socket so that the OS won't bind to the same local address.

ami-GS: I think the AuxSocket is covered by normal socket

csujedihy: How does normal socket cover that? Do we bind a TCP socket at all?

src/inc/quic_datapath.h
```diff
@@ -180,6 +181,8 @@ typedef struct CXPLAT_ROUTE {
     CXPLAT_ROUTE_STATE State;
     CXPLAT_RAW_TCP_STATE TcpState;
 
+    uint16_t DatapathType; // CXPLAT_DATAPATH_TYPE
```
nibanks, line 184: Why is this a `uint16_t` and not simply `CXPLAT_DATAPATH_TYPE`? And should it be _before_ the "copies memory up to this point" section?

src/inc/quic_datapath.h
```diff
@@ -224,7 +227,8 @@ typedef struct CXPLAT_RECV_DATA {
     //
     uint16_t Allocated : 1;          // Used for debugging. Set to FALSE on free.
     uint16_t QueuedOnConnection : 1; // Used for debugging.
-    uint16_t Reserved : 6;
+    uint16_t DatapathType : 2;       // CXPLAT_DATAPATH_TYPE
+    uint16_t Reserved : 4;
     uint16_t ReservedEx : 8;
```
nibanks, line 232: I'm not sure why we have separate Reserved fields. Let's just combine these into one.

src/platform/datapath_raw.h
```diff
@@ -253,27 +238,25 @@ CxPlatDpRawTxEnqueue(
...
 //
 
-typedef struct CXPLAT_SOCKET {
+typedef struct CXPLAT_SOCKET_RAW {
 
     CXPLAT_HASHTABLE_ENTRY Entry;
     CXPLAT_RUNDOWN_REF Rundown;
-    CXPLAT_DATAPATH* Datapath;
+    CXPLAT_DATAPATH_RAW* RawDatapath;
     SOCKET AuxSocket;
```
nibanks, line 246: Put a comment that this is only used for QUIC over TCP? (it is, right?)

src/platform/datapath_raw_socket.c
```diff
@@ -17,20 +17,36 @@
 #pragma warning(disable:4116) // unnamed type definition in parentheses
 #pragma warning(disable:4100) // unreferenced formal parameter
 
+
+uint32_t
+CxPlatGetRawSocketSize () {
```
nibanks, line 22:
```suggestion
CxPlatGetRawSocketSize(void) {
```
nit: We don't add spaces, and for `c` code we also add `void` instead empty functions.

src/platform/datapath_raw_socket_win.c
```diff
@@ -262,50 +264,31 @@ CxPlatTryAddSocket(
...
-            DatapathErrorStatus,
-            "[data][%p] ERROR, %u, %s.",
-            Socket,
-            WsaError,
-            "Set IPV6_V6ONLY");
-        Status = HRESULT_FROM_WIN32(WsaError);
-        goto Error;
-    }
+    if (Socket->UseTcp)
+    {
```
nibanks, line 268:
```suggestion
    if (Socket->UseTcp) {
```

src/platform/datapath_raw_socket_win.c
```diff
@@ -315,49 +298,80 @@ CxPlatTryAddSocket(
...
-            Socket->AuxSocket,
-            (struct sockaddr*)&MappedAddress,
-            sizeof(MappedAddress));
-    if (Result == SOCKET_ERROR) {
-        int WsaError = SocketError();
-        QuicTraceEvent(
-            DatapathErrorStatus,
-            "[data][%p] ERROR, %u, %s.",
+    if (Socket->UseTcp)
+    {
```
nibanks, line 339:
```suggestion
    if (Socket->UseTcp) {
```

src/platform/datapath_raw_socket_win.c
```diff
@@ -315,49 +298,80 @@ CxPlatTryAddSocket(
...
+            CxPlatRwLockReleaseExclusive(&Pool->Lock);
+            Status = HRESULT_FROM_WIN32(WsaError);
+            goto Error;
+        }
 
-    if (Socket->Connected) {
-        CxPlatZeroMemory(&MappedAddress, sizeof(MappedAddress));
-        CxPlatConvertToMappedV6(&Socket->RemoteAddress, &MappedAddress);
+        if (Socket->Connected)
+        {
```
nibanks, line 366:
```suggestion
        if (Socket->Connected) {
```

src/platform/datapath_raw_socket_win.c
```diff
@@ -315,49 +298,80 @@ CxPlatTryAddSocket(
...
             Socket,
-            WsaError,
-            "bind");
-        CxPlatRwLockReleaseExclusive(&Pool->Lock);
-        Status = HRESULT_FROM_WIN32(WsaError);
-        goto Error;
-    }
+            LocalAddressString.Address);
+        Result =
+            bind(
```
nibanks, line 348: Are we going to always try to bind to the same addresess as the normal UDP socket?

src/platform/datapath_raw_xdp_win.c
```diff
@@ -24,8 +24,9 @@
 #endif
 
 typedef struct XDP_DATAPATH {
-    CXPLAT_DATAPATH;
     DECLSPEC_CACHEALIGN
+    CXPLAT_DATAPATH_RAW;
```
nibanks, line 28: Why did you move this to after the alignment?

src/platform/datapath_win.c
```diff
@@ -0,0 +1,636 @@
...
+    )
+{
+    if (DatapathType(SendData) == CXPLAT_DATAPATH_TYPE_USER) {
+        return DataPathIsPaddingPreferred(Datapath);
+    } else if (DatapathType(SendData) == CXPLAT_DATAPATH_TYPE_XDP) {
+        return RawDataPathIsPaddingPreferred(Datapath);
+    } else {
+        CXPLAT_DBG_ASSERT(FALSE);
+    }
+    return FALSE;
```
nibanks, line 132:
```suggestion
    CXPLAT_DBG_ASSERT(
        DatapathType(SendData) == CXPLAT_DATAPATH_TYPE_USER ||
        DatapathType(SendData) == CXPLAT_DATAPATH_TYPE_XDP);
    return
        DatapathType(SendData) == CXPLAT_DATAPATH_TYPE_USER ?
            DataPathIsPaddingPreferred(Datapath) : RawDataPathIsPaddingPreferred(Datapath);
```

src/platform/datapath_win.c
```diff
@@ -0,0 +1,636 @@
...
+_Success_(QUIC_SUCCEEDED(return))
+QUIC_STATUS
+CxPlatDataPathGetLocalAddresses(
+    _In_ CXPLAT_DATAPATH* Datapath,
+    _Outptr_ _At_(*Addresses, __drv_allocatesMem(Mem))
+        CXPLAT_ADAPTER_ADDRESS** Addresses,
+    _Out_ uint32_t* AddressesCount
+    )
+{
+    // TODO: XDP doesn't support, could be inlined here
```
nibanks, line 145: Let's go ahead and inline these two functions please.

src/platform/datapath_win.c
```diff
@@ -0,0 +1,636 @@
...
+    _Inout_ CXPLAT_SEND_CONFIG* Config
+    )
+{
+    CXPLAT_SEND_DATA* SendData = NULL;
+    // TODO: fallback?
+    if (Socket->UseTcp || Config->Route->DatapathType == CXPLAT_DATAPATH_TYPE_XDP ||
+        (Config->Route->DatapathType == CXPLAT_DATAPATH_TYPE_UNKNOWN &&
+        Socket->RawSocketAvailable && !IS_LOOPBACK(Config->Route->RemoteAddress))) {
+        SendData = RawSendDataAlloc(SocketToRaw(Socket), Config);
+        if (SendData) {
```
nibanks, line 451: Instead of setting the Type fields here, update the two SendDataAlloc functions to set the type.

src/platform/datapath_win.c
```diff
@@ -0,0 +1,636 @@
...
+CxPlatRecvDataReturn(
+    _In_opt_ CXPLAT_RECV_DATA* RecvDataChain
+    )
+{
+    if (RecvDataChain == NULL) {
+        return;
+    }
+
+    if (RecvDataChain->DatapathType == CXPLAT_DATAPATH_TYPE_USER) {
+        RecvDataReturn(RecvDataChain);
```
nibanks, line 429: Since you always check for NULL above, make sure these two functions don't also check, and update the SAL annotations to indicate no longer `_opt_`.

src/platform/datapath_win.c
```diff
@@ -0,0 +1,636 @@
...
+    }
+    return QUIC_STATUS_NOT_SUPPORTED;
+}
+
+void
+CxPlatDataPathProcessCqe(
+    _In_ CXPLAT_CQE* Cqe
+    )
+{
+    // TODO: refactoring
```
nibanks, line 551: What is left to do here?

src/platform/datapath_win.c
```diff
@@ -0,0 +1,636 @@
...
+void
+CxPlatResolveRouteComplete(
+    _In_ void* Connection,
+    _Inout_ CXPLAT_ROUTE* Route,
+    _In_reads_bytes_(6) const uint8_t* PhysicalAddress,
+    _In_ uint8_t PathId
+    )
+{
+    if (Route->State != RouteResolved) {
+        RawResolveRouteComplete(Connection, Route, PhysicalAddress, PathId);
```
nibanks, line 597: Debug assert that the type is RAW/XDP?

src/platform/platform_internal.h
```diff
@@ -39,6 +39,58 @@
 
 #endif
 
+// TODO: create header files for each datapath
```
nibanks, line 42: Either do this or remove the TODO please

src/platform/platform_internal.h
```diff
@@ -39,6 +39,58 @@
...
+
+typedef struct DATAPATH_SQE {
+    uint32_t CqeType;
+#ifdef CXPLAT_SQE
+    CXPLAT_SQE Sqe;
+#endif
+} DATAPATH_SQE;
+
+// not needed?
```
nibanks, line 51: Is it needed or not? Please remove either comment or struct.

src/platform/platform_internal.h
```diff
@@ -39,6 +39,58 @@
...
+typedef struct CXPLAT_SEND_DATA_COMMON {
+    uint16_t DatapathType; // CXPLAT_DATAPATH_TYPE
+
+    //
+    // The type of ECN markings needed for send.
+    //
+    uint8_t ECN; // CXPLAT_ECN_TYPE
+} CXPLAT_SEND_DATA_COMMON;
+
+
```
nibanks, line 85: nit: only one newline please

src/platform/platform_internal.h
```diff
@@ -104,6 +156,713 @@ typedef struct CX_PLATFORM {
...
+    //
+    // Per-processor socket contexts.
+    //
+    CXPLAT_SOCKET_PROC PerProcSockets[0];
+
+} CXPLAT_SOCKET;
+
+typedef struct CXPLAT_SOCKET_RAW CXPLAT_SOCKET_RAW;
+
+// TODO: move to common area
```
nibanks, line 529: There are a number of TODOs here and below that should be addressed.

src/platform/platform_internal.h
```diff
@@ -104,6 +156,713 @@ typedef struct CX_PLATFORM {
...
+// TODO: rename
+CXPLAT_SOCKET*
+RawToSocket(CXPLAT_SOCKET_RAW* Socket);
+
+// TODO: rename
+CXPLAT_SOCKET_RAW*
+SocketToRaw(CXPLAT_SOCKET* Socket);
+
+uint32_t
+CxPlatGetRawSocketSize ();
```
nibanks, line 722:
```suggestion
CxPlatGetRawSocketSize();
```

src/platform/datapath_winuser.c
```diff
@@ -269,11 +236,6 @@ typedef struct CXPLAT_SEND_DATA {
     //
     uint16_t SegmentSize;
 
-    //
-    // The type of ECN markings needed for send.
-    //
-    uint8_t ECN; // CXPLAT_ECN_TYPE
```
nibanks, line 275: Does ECN really need to be in the common?

src/platform/datapath_winuser.c
```diff
@@ -1654,109 +1292,19 @@ CxPlatDataPathGetGatewayAddresses(
...
-                    IN6_GET_ADDR_V4MAPPED(&SockAddr6->sin6_addr);
-            SockAddr4->sin_port = SockAddr6->sin6_port;
-            return;
-        }
-    }
-
-    CxPlatCopyMemory(Address, Ai->ai_addr, Ai->ai_addrlen);
-}
-
+// TODO: DELETE
```
nibanks, line 1295: Delete this function?

src/platform/datapath_winuser.c
```diff
@@ -3085,19 +2661,21 @@ CxPlatSocketUpdateQeo(
     return QUIC_STATUS_NOT_SUPPORTED;
 }
 
+// TODO: remove?
```
nibanks, line 2664: Remove?

src/platform/datapath_winuser.c
```diff
@@ -3085,19 +2661,21 @@ CxPlatSocketUpdateQeo(
...
-CxPlatSocketGetLocalMtu(
+SocketGetLocalMtu(
     _In_ CXPLAT_SOCKET* Socket
     )
 {
     CXPLAT_DBG_ASSERT(Socket != NULL);
     return Socket->Mtu;
 }
 
+// TODO: remove
```
nibanks, line 2675: Remove?

src/platform/datapath_winuser.c
```diff
@@ -3106,9 +2684,10 @@ CxPlatSocketGetLocalAddress(
     *Address = Socket->LocalAddress;
 }
 
+// TODO: remove
```
nibanks, line 2687: Remove?

src/platform/datapath_winuser.c
```diff
@@ -3214,12 +2793,12 @@ CxPlatSocketStartAccept(
     if (ListenerSocketProc->AcceptSocket == NULL) {
         Status =
             CxPlatSocketCreateTcpInternal(
-                Datapath,
+                ListenerSocketProc->Parent->Datapath,
```
nibanks, line 2796: Undo this change?

src/platform/datapath_winuser.c
```diff
@@ -3980,7 +3560,6 @@ CxPlatDataPathSocketProcessRioCompletion(
 {
     UNREFERENCED_PARAMETER(Cqe);
     CXPLAT_SOCKET_PROC* SocketProc = CONTAINING_RECORD(Sqe, CXPLAT_SOCKET_PROC, RioSqe);
-    CXPLAT_DATAPATH* Datapath = SocketProc->DatapathProc->Datapath;
```
nibanks, line 3983: Why?

src/platform/datapath_winuser.c
```diff
@@ -4310,19 +3890,18 @@ RioSendDataFree(
     )
 {
     CXPLAT_SEND_DATA* SendData = Entry;
-    CXPLAT_DATAPATH* Datapath = SendData->Owner->Datapath;
```
nibanks, line 4313: Instead of deleting these temp variables, let's keep them.

src/platform/datapath_winuser.c
```diff
@@ -5090,18 +4668,8 @@ CxPlatDataPathProcessCqe(
...
-    _Inout_ CXPLAT_ROUTE* DstRoute,
-    _In_ CXPLAT_ROUTE* SrcRoute
-    )
-{
-    *DstRoute = *SrcRoute;
-}
-
 void
-CxPlatResolveRouteComplete(
+ResolveRouteComplete(
```
nibanks, line 4672: Are these Route functions ever called from anywhere? If not, let's delete them.

nibanks:
Is there any reason to keep the `QUIC_USE_XDP` build flag now? AFAICT, we're always building with XDP now, right?

src/inc/quic_datapath.h
ami-GS: oops, I made the field?

src/platform/datapath_raw.h
ami-GS: No, This can be any raw socket. XDP, DPDK and QUIC over TCP (XDP)

src/platform/datapath_raw_socket_win.c
ami-GS: I changed it only when QTIP enabled

src/platform/datapath_raw_xdp_win.c
ami-GS: not needed..... I changed experimentally....

src/inc/quic_datapath.h
ami-GS: Reserved is used in `CxPlatDpRawRxEthernet` for packet parsing. Let me reduce ReservedEx size

ami-GS: ReservedEx is also used. Reserved requires at least 3 bits, I can leave this as it is.

src/platform/datapath_win.c
ami-GS: I noticed that this API is not used at all. inline normal datapath should not affect anything.

src/platform/datapath_win.c
ami-GS: Allowing to call from any datapath, but functionality is basically for RAW

src/inc/quic_datapath.h
nibanks: Why can't we combine `Reserved` and `ReservedEx` into one `uint16_t Reserved : 12;`?

src/inc/quic_datapath.h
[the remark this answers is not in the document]
nibanks: Last time I checked, SendData had a pointer to the socket context, which has a pointer to the datapath.

src/platform/datapath_raw_dummy.c
nibanks: We can leave the dummy file.

src/inc/quic_datapath.h
csujedihy: `Reserved` and `ReservedEx` are used for two different things. If we combine them, it would be hard to consume them.

`Reserved` is socket type and `ReservedEx` is TCP header length. `ReservedEx` can be reduced to 6-bit but I don't think we need to touch them at all.

src/platform/datapath_winuser.c
ami-GS: ECN seems to be used in both normal socket and xdp.

src/platform/datapath_winuser.c
ami-GS: Maybe the "Datapath" is used only once, but let me  change back

ami-GS:
`QUIC_USE_XDP` is still used for Linux XDP build (not functional)

nibanks:
> `QUIC_USE_XDP` is still used for Linux XDP build (not functional)

Since it's still a no-op, maybe you just integrate it too? I want to get rid of any XDP-specific build flags.

.azure/obtemplates/build-winuser-xdp.yml
```diff
@@ -1,36 +0,0 @@
...
-    ob_sdl_binskim_break: true # https://aka.ms/obpipelines/sdl
-    ob_sdl_codeSignValidation_excludes: -|**\*.exe # Disable signing requirements for test executables
-  steps:
-  - task: PowerShell@2
-    displayName: Prepare Build Machine
-    target: windows_build_container2
-    inputs:
-      pwsh: true
-      filePath: scripts/prepare-machine.ps1
-      arguments: -InstallXdpSdk -ForContainerBuild
```
nibanks, line 22: Don't we need to update `prepare-machine.ps1` to always install the XDP SDK (at least on Linux) now?

.azure/azure-pipelines.perf.yml
```diff
@@ -194,9 +194,9 @@ stages:
         extraName: 'xdp'
         extraPrepareArgs: -DisableTest -InstallXdpSdk
         ${{ if eq(parameters.pgo_mode, false) }}:
-          extraBuildArgs: -DisableTest -DisableTools -UseXdp -ExtraArtifactDir Xdp
+          extraBuildArgs: -DisableTest -DisableTools -ExtraArtifactDir Xdp
         ${{ if eq(parameters.pgo_mode, true) }}:
-          extraBuildArgs: -DisableTest -DisableTools -UseXdp -ExtraArtifactDir Xdp -PGO
+          extraBuildArgs: -DisableTest -DisableTools -ExtraArtifactDir Xdp -PGO
```
nibanks, line 199: Remove this XDP specific build all together, and have the XDP run use the normal build (but with XDP installed).

.azure/OneBranch.Official.yml
```diff
@@ -204,7 +188,6 @@ extends:
       dependsOn:
       - build_windows
       - build_uwp
-      - build_windows_xdp
       jobs:
       - template: .azure/obtemplates/build-nuget.yml@self
```
nibanks, line 192: I think this yaml and associated scripts will need to be updated to not do anything XDP specific any more.

.azure/obtemplates/build-winuser.yml
```diff
@@ -13,6 +13,13 @@ jobs:
     ob_sdl_binskim_break: true # https://aka.ms/obpipelines/sdl
     ob_sdl_codeSignValidation_excludes: -|**\*.exe # Disable signing requirements for test executables
   steps:
+  - task: PowerShell@2
+    displayName: Prepare Build Machine
+    target: windows_build_container2
+    inputs:
+      pwsh: true
+      filePath: scripts/prepare-machine.ps1
+      arguments: -InstallXdpSdk -ForContainerBuild
```
nibanks, line 22: Why do we need the `-ForContainerBuild` option now, when it wasn't needed before?

.azure/templates/build-config-user.yml
```diff
@@ -43,7 +43,7 @@ jobs:
     inputs:
       pwsh: true
       filePath: scripts/prepare-machine.ps1
-      arguments: -Tls ${{ parameters.tls }} -ForBuild ${{ parameters.extraPrepareArgs }}
+      arguments: -Tls ${{ parameters.tls }} -ForBuild ${{ parameters.extraPrepareArgs }} -UseXdp
```
nibanks, line 46: Like I said in a separate comment, `prepare-machine.ps1` should just always install the XDP SDK when building Windows now.

.github/workflows/build-reuse-win.yml
```diff
@@ -85,15 +85,15 @@ jobs:
       uses: ilammy/setup-nasm@321e6ed62a1fc77024a3bd853deb33645e8b22c4
     - name: Prepare Machine
       shell: pwsh
-      run: scripts/prepare-machine.ps1 -ForBuild -Tls ${{ inputs.tls }}
+      run: scripts/prepare-machine.ps1 -ForBuild -Tls ${{ inputs.tls }} ${{ inputs.xdp }}
```
nibanks, line 88: We shouldn't need any build-specific stuff for XDP now. No special jobs. No special flags. Windows user just always builds in XDP.

CMakeLists.txt
```diff
@@ -555,11 +555,6 @@ if(WIN32)
     set(CMAKE_EXE_LINKER_FLAGS_RELEASE "${CMAKE_EXE_LINKER_FLAGS_RELEASE} /LTCG /IGNORE:4075 /DEBUG /OPT:REF /OPT:ICF")
 
     # Find the right PGO file.
-    if(QUIC_USE_XDP)
-        set(QUIC_PGO_NAME "msquic.${QUIC_TLS}.xdp.pgd")
-        set(QUIC_PGO_FILE "${CMAKE_CURRENT_SOURCE_DIR}/src/bin/winuser/pgo_${SYSTEM_PROCESSOR}/${QUIC_PGO_NAME}")
-    endif()
```
nibanks, line 561: This will make things a bit complicated, because we'll need to run both XDP and non-XDP and combine their PGO files.

scripts/prepare-machine.ps1
```diff
@@ -109,17 +109,6 @@ if ($PSVersionTable.PSVersion.Major -lt 7) {
                  "Please visit https://github.com/microsoft/msquic/blob/main/docs/BUILD.md#powershell-usage")
 }
 
-if ($UseXdp) {
-    # Helper for XDP usage
-    if ($ForBuild) {
-        $InstallXdpSdk = $true;
```
ami-GS, line 115: Why skipping installing XdpSdk for build?

nibanks:
Not sure it's related to this PR or not, but it seems stress hit a crash:
https://github.com/microsoft/msquic/actions/runs/6252310697/job/16975704088?pr=3819
```
==4800==ERROR: AddressSanitizer: access-violation on unknown address 0x01000000013f (pc 0x7ffb4c74f5e3 bp 0x000000000000 sp 0x00a724efde00 T187)
==4800==The signal is caused by a READ memory access.
    #0 0x7ffb4c74f5e2 in CxPlatPoolAlloc D:\a\msquic\msquic\src\inc\quic_platform_winuser.h:430
    #1 0x7ffb4c884db1 in SendDataAlloc D:\a\msquic\msquic\src\platform\datapath_winuser.c:3635
    #2 0x7ffb4c860800 in CxPlatSendDataAlloc D:\a\msquic\msquic\src\platform\datapath_win.c:634
    #3 0x7ffb4c845270 in QuicPacketBuilderPrepare D:\a\msquic\msquic\src\core\packet_builder.c:269
    #4 0x7ffb4c840d18 in QuicPacketBuilderPrepareForControlFrames D:\a\msquic\msquic\src\core\packet_builder.c:572
    #5 0x7ffb4c7f47cb in QuicSendFlush D:\a\msquic\msquic\src\core\send.c:1327
    #6 0x7ffb4c776a7d in QuicConnDrainOperations D:\a\msquic\msquic\src\core\connection.c:7691
    #7 0x7ffb4c7be1ed in QuicWorkerProcessConnection D:\a\msquic\msquic\src\core\worker.c:510
    #8 0x7ffb4c7baf30 in QuicWorkerLoop D:\a\msquic\msquic\src\core\worker.c:668
    #9 0x7ffb4c87e46e in CxPlatRunExecutionContexts D:\a\msquic\msquic\src\platform\platform_worker.c:395
    #10 0x7ffb4c87da45 in CxPlatWorkerThread D:\a\msquic\msquic\src\platform\platform_worker.c:492
    #11 0x7ffb49d9da49  (D:\a\msquic\msquic\artifacts\bin\windows\x64_Debug_schannel\clang_rt.asan_dynamic-x86_64.dll+0x18005da49)
    #12 0x7ffb83ff4ddf  (C:\Windows\System32\KERNEL32.DLL+0x180014ddf)
    #13 0x7ffb8449ec4a  (C:\Windows\SYSTEM32\ntdll.dll+0x18007ec4a)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: access-violation D:\a\msquic\msquic\src\inc\quic_platform_winuser.h:430 in CxPlatPoolAlloc
Thread T187 created by T185 here:
    #0 0x7ffb49d9f118  (D:\a\msquic\msquic\artifacts\bin\windows\x64_Debug_schannel\clang_rt.asan_dynamic-x86_64.dll+0x18005f118)
    #1 0x7ffb4c7bee54 in CxPlatThreadCreate D:\a\msquic\msquic\src\inc\quic_platform_winuser.h:1135
    #2 0x7ffb4c87cf1a in CxPlatWorkersLazyStart D:\a\msquic\msquic\src\platform\platform_worker.c:205
    #3 0x7ffb4c88388a in DataPathInitialize D:\a\msquic\msquic\src\platform\datapath_winuser.c:793
    #4 0x7ffb4c85ca8d in CxPlatDataPathInitialize D:\a\msquic\msquic\src\platform\datapath_win.c:47
    #5 0x7ffb4c722402 in QuicLibraryLazyInitialize D:\a\msquic\msquic\src\core\library.c:678
    #6 0x7ffb4c751868 in MsQuicRegistrationOpen D:\a\msquic\msquic\src\core\registration.c:59
    #7 0x7ff6eaa575ff in RunThread(void *) D:\a\msquic\msquic\src\tools\spin\spinquic.cpp:1331
    #8 0x7ffb49d9da49  (D:\a\msquic\msquic\artifacts\bin\windows\x64_Debug_schannel\clang_rt.asan_dynamic-x86_64.dll+0x18005da49)
    #9 0x7ffb83ff4ddf  (C:\Windows\System32\KERNEL32.DLL+0x180014ddf)
    #10 0x7ffb8449ec4a  (C:\Windows\SYSTEM32\ntdll.dll+0x18007ec4a)

Thread T185 created by T0 here:
    #0 0x7ffb49d9f118  (D:\a\msquic\msquic\artifacts\bin\windows\x64_Debug_schannel\clang_rt.asan_dynamic-x86_64.dll+0x18005f118)
    #1 0x7ff6eaa690a4 in CxPlatThreadCreate D:\a\msquic\msquic\src\inc\quic_platform_winuser.h:1135
    #2 0x7ff6eaa58b46 in start(void) D:\a\msquic\msquic\src\tools\spin\spinquic.cpp:1497
    #3 0x7ff6eaa59810 in main D:\a\msquic\msquic\src\tools\spin\spinquic.cpp:1603
    #4 0x7ff6eaac496b in invoke_main D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:78
    #5 0x7ff6eaac496b in __scrt_common_main_seh D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:288
    #6 0x7ffb83ff4ddf  (C:\Windows\System32\KERNEL32.DLL+0x180014ddf)
    #7 0x7ffb8449ec4a  (C:\Windows\SYSTEM32\ntdll.dll+0x18007ec4a)
```

.azure/obtemplates/build-winuser.yml
ami-GS: Because XDP dependencies become must items for any winuser build?

CMakeLists.txt
ami-GS: I'm not familiar with PGO, How to combine these two?

nibanks: We will need to combine the perf testing scenarios to test both normal and XDP. We can do that in a follow up PR.

nibanks: Opened https://github.com/microsoft/msquic/issues/3866 to track.

scripts/prepare-machine.ps1
nibanks: `if ($ForBuild) {` below already has `$InstallXdpSdk = $true`.

nibanks (approved):
If all the tests pass, let's get this merged. Thanks Daiki for all the hard work!

ami-GS:
Thanks for review!!

PR #3823, opened 2023-08-22
feat: close listeners in registration rundown

### LINKED ISSUE #3818: Support Listener stop via RegistrationShutdown API
### Describe the bug

deadlock if Registration is closed without first closing the listener.

### Affected OS

- [ ] Windows
- [X] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

regular ubuntu 20.04 image.
other OS untested.

### MsQuic version

v2.1

1.  Open Registration 
2. Open & Start listener with Registration
3. Shutdown and Close Registration
4. no return, hangs the caller

### Expected behavior

in 3) it should shutdown the listener as one of the side effects.
    4) should return

### Actual outcome

in 4) no return, hangs the caller. for the single thread app, it is a deadlock.

The only way to get out is to shut down the listener in another thread.

### Additional details

RegistrationShutdown shutdowns the connections of the configurations.
So ideally RegistrationShutdown should also shut down the listeners.

from code, Registration keeps track of the  Configurations it opens, it should also track all the listener it opens so that it could shut down the listeners in rundown callbacks.

qzhuyan:
## Description
fix #3818

## Testing

## Documentation

### DIFF
--- a/src/core/listener.c
+++ b/src/core/listener.c
@@ -80,12 +80,12 @@ MsQuicListenerOpen(
     BOOLEAN Result = CxPlatRundownAcquire(&Registration->Rundown);
     CXPLAT_DBG_ASSERT(Result); UNREFERENCED_PARAMETER(Result);
 
-    CxPlatDispatchLockAcquire(&Registration->ListenerLock);
+    CxPlatLockAcquire(&Registration->ListenerLock);
     RegistrationShuttingDown = Registration->ShuttingDown;
     if (!RegistrationShuttingDown) {
         CxPlatListInsertTail(&Registration->Listeners, &Listener->RegistrationLink);
     }
-    CxPlatDispatchLockRelease(&Registration->ListenerLock);
+    CxPlatLockRelease(&Registration->ListenerLock);
 
     if (RegistrationShuttingDown) {
         CxPlatRundownRelease(&Registration->Rundown);
@@ -137,9 +137,9 @@ QuicListenerFree(
 
     if (!Listener->Registration->ShuttingDown)
     {
-        CxPlatDispatchLockAcquire(&Listener->Registration->ListenerLock);
+        CxPlatLockAcquire(&Listener->Registration->ListenerLock);
         CxPlatListEntryRemove(&Listener->RegistrationLink);
-        CxPlatDispatchLockRelease(&Listener->Registration->ListenerLock);
+        CxPlatLockRelease(&Listener->Registration->ListenerLock);
     }
 
     CxPlatRefUninitialize(&Listener->RefCount);
--- a/src/core/listener.c
+++ b/src/core/listener.c
@@ -75,6 +75,10 @@ MsQuicListenerOpen(
     QuicSiloAddRef(Listener->Silo);
 #endif
 
+    CxPlatDispatchLockAcquire(&Registration->ListenerLock);
+    CxPlatListInsertTail(&Registration->Listeners, &Listener->RegistrationLink);
+    CxPlatDispatchLockRelease(&Registration->ListenerLock);
+
     BOOLEAN Result = CxPlatRundownAcquire(&Registration->Rundown);
     CXPLAT_DBG_ASSERT(Result); UNREFERENCED_PARAMETER(Result);
 
@@ -116,6 +120,10 @@ QuicListenerFree(
 #ifdef QUIC_SILO
     QuicSiloRelease(Listener->Silo);
 #endif
+
+    CxPlatDispatchLockAcquire(&Listener->Registration->ListenerLock);
+    CxPlatListEntryRemove(&Listener->RegistrationLink);
+    CxPlatDispatchLockRelease(&Listener->Registration->ListenerLock);
 
     CxPlatRefUninitialize(&Listener->RefCount);
     CxPlatEventUninitialize(Listener->StopEvent);
--- a/src/core/listener.c
+++ b/src/core/listener.c
@@ -135,8 +135,7 @@ QuicListenerFree(
 #endif
 
     CxPlatDispatchLockAcquire(&Listener->Registration->ConnectionLock);
-    if (!Listener->Registration->ShuttingDown)
-    {
+    if (!Listener->Registration->ShuttingDown) {
         CxPlatListEntryRemove(&Listener->RegistrationLink);
     }
     CxPlatDispatchLockRelease(&Listener->Registration->ConnectionLock);
--- a/src/core/listener.c
+++ b/src/core/listener.c
@@ -75,20 +75,34 @@ MsQuicListenerOpen(
     QuicSiloAddRef(Listener->Silo);
 #endif
 
-    CxPlatDispatchLockAcquire(&Registration->ListenerLock);
-    CxPlatListInsertTail(&Registration->Listeners, &Listener->RegistrationLink);
-    CxPlatDispatchLockRelease(&Registration->ListenerLock);
+    BOOLEAN RegistrationShuttingDown;
 
     BOOLEAN Result = CxPlatRundownAcquire(&Registration->Rundown);
     CXPLAT_DBG_ASSERT(Result); UNREFERENCED_PARAMETER(Result);
 
-    QuicTraceEvent(
-        ListenerCreated,
-        "[list][%p] Created, Registration=%p",
-        Listener,
-        Listener->Registration);
-    *NewListener = (HQUIC)Listener;
-    Status = QUIC_STATUS_SUCCESS;
+    CxPlatDispatchLockAcquire(&Registration->ListenerLock);
+    RegistrationShuttingDown = Registration->ShuttingDown;
+    if (!RegistrationShuttingDown) {
+        CxPlatListInsertTail(&Registration->Listeners, &Listener->RegistrationLink);
+    }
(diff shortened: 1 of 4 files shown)


src/core/registration.c
```diff
@@ -247,6 +249,18 @@ MsQuicRegistrationShutdown(
...
         }
 
+        CxPlatDispatchLockAcquire(&Registration->ListenerLock);
+        Entry = Registration->Listeners.Flink;
+        while (Entry != &Registration->Listeners) {
+
+            QUIC_LISTENER* Listener =
+                CXPLAT_CONTAINING_RECORD(Entry, QUIC_LISTENER, RegistrationLink);
+
+            MsQuicListenerClose((void *)Listener);
```
nibanks, line 259: You should call `MsQuicListenerStop` instead. The app should **always** be the one to explicitly call Close.

src/core/registration.c
```diff
@@ -247,6 +249,18 @@ MsQuicRegistrationShutdown(
             Entry = Entry->Flink;
         }
 
+        CxPlatDispatchLockAcquire(&Registration->ListenerLock);
```
nibanks, line 252: You can't do all this under lock. Once you call ListenerStop instead of Close, it's likely the app may call ListenerClose in the Stop completion callback. As it stands now, in ListenerClose you are grabbing the registration's lock (this lock) again to remove it from the list. There are several issues with that:

1. Locks are not reentrant, at least in kernel mode. So it will bug check.
2. `Entry` here is part of the `Listener` you much assume `Listener` has been freed after this Stop/Close call below. So you cannot reference `Entry` after that call.

qzhuyan: thanks for the comments!

I think I have to grab the lock to protect the concurrent access of  Registration->Listeners.

what about adding a flag that marks the Registration of the Listener is currently in rundown process where lock&update the `Registration ->Listeners` can be skipped? 

For use after free as in 2), we could swap L260 and L 259.

I try to avoid making `DISPATCH_LEVEL` API for the listeners that looks like will be huge changes.

src/core/registration.c
qzhuyan: You are right. But for already stopped listener, calling ListenerStop will not trigger the stop complete callback as it already stopped and that registaion close will still hang because their is no way to trigger the Listener close.

nibanks: That is the same thing for Connections. If the app has gotten a completion event for an object, but not closed/freed it, then it is an app issue; not our issue.

src/core/listener.c
```diff
@@ -75,6 +75,10 @@ MsQuicListenerOpen(
     QuicSiloAddRef(Listener->Silo);
 #endif
 
+    CxPlatDispatchLockAcquire(&Registration->ListenerLock);
+    CxPlatListInsertTail(&Registration->Listeners, &Listener->RegistrationLink);
```
qzhuyan, line 79: also should check if the target Registration is under 'rundown'

src/core/listener.c
```diff
@@ -117,6 +135,13 @@ QuicListenerFree(
     QuicSiloRelease(Listener->Silo);
 #endif
 
+    if (!Listener->Registration->ShuttingDown)
+    {
+        CxPlatDispatchLockAcquire(&Listener->Registration->ListenerLock);
+        CxPlatListEntryRemove(&Listener->RegistrationLink);
+        CxPlatDispatchLockRelease(&Listener->Registration->ListenerLock);
```
qzhuyan, line 142: oh, shouldn't be dispatch lock

src/core/registration.c
```diff
@@ -291,6 +305,17 @@ QuicRegistrationTraceRundown(
...
+    CxPlatLockAcquire(&Registration->ListenerLock);
+
+    for (CXPLAT_LIST_ENTRY* Link = Registration->Listeners.Flink;
+        Link != &Registration->Listeners;
+        Link = Link->Flink) {
+        QuicListenerTraceRundown(
+            CXPLAT_CONTAINING_RECORD(Link, QUIC_LISTENER, RegistrationLink));
+    }
+
+    CxPlatLockRelease(&Registration->ListenerLock);
```
nibanks, line 318: This doesn't need to be called from here, because it is already called from the Binding (which is called from the library).

src/core/listener.c
```diff
@@ -75,16 +75,34 @@ MsQuicListenerOpen(
...
+        goto Error;
+    } else {
+        QuicTraceEvent(
+            ListenerCreated,
+            "[list][%p] Created, Registration=%p",
+            Listener,
+            Listener->Registration);
+        *NewListener = (HQUIC)Listener;
+        Status = QUIC_STATUS_SUCCESS;
+    }
```
nibanks, line 105:
```suggestion
    }

    QuicTraceEvent(
        ListenerCreated,
        "[list][%p] Created, Registration=%p",
        Listener,
        Listener->Registration);
    *NewListener = (HQUIC)Listener;
    Status = QUIC_STATUS_SUCCESS;
```
No need to have the `else`, because the `if` has that `goto`.

src/core/registration.c
```diff
@@ -247,6 +249,18 @@ MsQuicRegistrationShutdown(
...
 
+        CxPlatLockAcquire(&Registration->ListenerLock);
+        Entry = Registration->Listeners.Flink;
+        while (Entry != &Registration->Listeners) {
+
+            QUIC_LISTENER* Listener =
+                CXPLAT_CONTAINING_RECORD(Entry, QUIC_LISTENER, RegistrationLink);
+
+            Entry = Entry->Flink;
+            MsQuicListenerStop((HQUIC)Listener);
```
nibanks, line 260: We can't do this under lock. Instead, move `Registration->Listeners` into a local list while under lock. Then, outside the lock, do this work.

src/core/registration.c
qzhuyan: So it is unnecessary to trace the listener removed from registration.

ok.

src/core/registration.h
```diff
@@ -90,6 +90,16 @@ typedef struct QUIC_REGISTRATION {
     //
     CXPLAT_LIST_ENTRY Connections;
 
+    //
+    // Protects access to the Listener list.
+    //
+    CXPLAT_LOCK ListenerLock;
```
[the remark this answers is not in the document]
qzhuyan: wait,  I have a question, The connection lock is ` CXPLAT_DISPATCH_LOCK `, in Linux it is the same as `CXPLAT_LOCK `but they are different in the window. does that matter?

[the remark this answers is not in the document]
qzhuyan: There is another  ConfigLock for `Configurations`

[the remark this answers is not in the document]
nibanks: There is a difference between the locks in kernel mode. If the code is running at DISPATCH_LEVEL it can only use a dispatch lock.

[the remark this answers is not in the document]
nibanks: Let's leave the `ConfigLock` for now. Just reuse the `ConnectionLock` for connections and listeners. Please add a comment to that effect too.

src/core/registration.c
qzhuyan: ok. I will do the work after releasing the lock.

We don't need a local list because after  `Registration->ShuttingDown = TRUE;` , neither `MsQuicListenerOpen` nor `QuicListenerFree` can access/update this list since I added a check there.

do you agree?

nibanks: Yeah, I think that's fine. Thanks!

src/core/listener.c
```diff
@@ -117,6 +134,13 @@ QuicListenerFree(
     QuicSiloRelease(Listener->Silo);
 #endif
 
+    CxPlatDispatchLockAcquire(&Listener->Registration->ConnectionLock);
+    if (!Listener->Registration->ShuttingDown)
+    {
+        CxPlatListEntryRemove(&Listener->RegistrationLink);
+    }
```
nibanks, line 141:
```suggestion
    if (!Listener->Registration->ShuttingDown) {
        CxPlatListEntryRemove(&Listener->RegistrationLink);
    }
```

qzhuyan:
two TCs are failing at local run, will check later...
```
Handshake_WithHandshakeArgs10.HandshakeSpecificLossPatterns_2
Handshake_WithHandshakeArgs10.HandshakeSpecificLossPatterns_3
```

qzhuyan:
> two TCs are failing at local run, will check later...
> 
> ```
> Handshake_WithHandshakeArgs10.HandshakeSpecificLossPatterns_2
> Handshake_WithHandshakeArgs10.HandshakeSpecificLossPatterns_3
> ```
Are these two flaky test cases? Now I cannot reproduce them.

nibanks:
> > @qzhuyan do you want to mark this "Ready for review"?
> 
> I am waiting for CI jobs to finish to see if there are any regressions. I see windows kernel mode perf regression -25%, is it something we should worry about?
> 
> ```
> Performance regression in ThroughputDown_Winkernel_x64_schannel_Default. -22.07% < -10
> ```

No, the results are inconsistent. You can ignore it.

nibanks:
> I am waiting for CI jobs to finish to see if there are any regressions.

Generally, everything looks fine.

nibanks (approved):

PR #3826, opened 2023-08-22
Refactor the RX Datapath Object Abstraction Model

nibanks:
## Description

Cleans up a few things:

1. Creates a clear abstraction between QUIC and platform layer.
2. Renames QUIC layer objects to remove CXPLAT name.
3. Refactors internals of datapaths to have similar structs/names to make it easier to understand (and maybe converge eventually).

## Testing

Existing automation

## Documentation

N/A

### DIFF
--- a/src/inc/quic_datapath.h
+++ b/src/inc/quic_datapath.h
@@ -231,7 +231,7 @@ typedef struct CXPLAT_RECV_DATA {
     // Variable length data of size `ClientRecvContextLength` passed into
     // CxPlatDataPathInitialize.
     //
-    uint8_t ClientData[0];
+    __attribute__((aligned(16))) uint8_t ClientData[0];
 
 } CXPLAT_RECV_DATA;
 
--- a/src/inc/quic_datapath.h
+++ b/src/inc/quic_datapath.h
@@ -132,11 +132,6 @@ typedef struct CXPLAT_DATAPATH CXPLAT_DATAPATH;
 typedef struct CXPLAT_SOCKET CXPLAT_SOCKET;
 
 //
-// Can be defined to whatever the client needs.
-//
-typedef struct CXPLAT_RECV_PACKET CXPLAT_RECV_PACKET;
-
-//
 // Structure that maintains the 'per send' context.
 //
 typedef struct CXPLAT_SEND_DATA CXPLAT_SEND_DATA;
@@ -232,7 +227,16 @@ typedef struct CXPLAT_RECV_DATA {
     uint16_t Reserved : 6;
     uint16_t ReservedEx : 8;
 
+    //
+    // Variable length data of size `ClientRecvContextLength` passed into
+    // CxPlatDataPathInitialize.
+    //
+    uint8_t ClientData[0];
+
 } CXPLAT_RECV_DATA;
+
+#define CxPlatRecvDataSize(ClientRecvContextLength) \
+    (sizeof(CXPLAT_RECV_DATA) + (ClientRecvContextLength))
 
 //
 // QUIC Encryption Offload (QEO) interfaces
@@ -275,22 +279,6 @@ typedef struct CXPLAT_QEO_CONNECTION {
     uint8_t HeaderKey[32];    // Length determined by CipherType
     uint8_t PayloadIv[12];
 } CXPLAT_QEO_CONNECTION;
-
-//
-// Gets the corresponding receive data from its context pointer.
-//
-CXPLAT_RECV_DATA*
-CxPlatDataPathRecvPacketToRecvData(
-    _In_ const CXPLAT_RECV_PACKET* const RecvPacket
-    );
-
-//
-// Gets the corresponding client context from its receive data pointer.
-//
-CXPLAT_RECV_PACKET*
-CxPlatDataPathRecvDataToRecvPacket(
-    _In_ const CXPLAT_RECV_DATA* const RecvData
-    );
 
 //
 // Function pointer type for datapath TCP accept callbacks.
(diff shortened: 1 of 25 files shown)


src/platform/datapath_raw_xdp_win.c
```diff
@@ -91,9 +91,9 @@ typedef struct XDP_QUEUE {
 } XDP_QUEUE;
 
 typedef struct DECLSPEC_ALIGN(MEMORY_ALLOCATION_ALIGNMENT) XDP_RX_PACKET {
-    CXPLAT_RECV_DATA;
-    CXPLAT_ROUTE RouteStorage;
     XDP_QUEUE* Queue;
+    CXPLAT_ROUTE RouteStorage;
+    CXPLAT_RECV_DATA RecvData;
```
csujedihy, line 96: >CXPLAT_RECV_DATA RecvData; [](http://example.com/codeflow?start=4&length=26)

Have you made sure RecvData is still on a 16-byte alignment boundary?

csujedihy: The first pointer sized member of `RecvData` is used as SLIST_ENTRY in `CxPlatDpRawRxFree`, which requires 16B alignment. Otherwise, the last 4 bits of that field isn't guaranteed to be zero.

csujedihy: That's probably `RecvData` was the first member in the parent struct and the parent struct has an alignment requirement.

src/inc/quic_datapath.h
```diff
@@ -232,6 +227,12 @@ typedef struct CXPLAT_RECV_DATA {
     uint16_t Reserved : 6;
     uint16_t ReservedEx : 8;
 
+    //
+    // Variable length data of size `ClientRecvContextLength` passed into
+    // CxPlatDataPathInitialize.
+    //
+    uint8_t ClientData[0];
```
ami-GS, line 234: `uint8_t ClientData[0] __attribute__((__aligned__(8)));`?

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1589,12 +1574,11 @@ CxPlatXdpRx(
...
 
-        if (Packet->Buffer) {
-            Packet->Allocated = TRUE;
-            Packet->Queue = Queue;
-            Buffers[PacketCount++] = (CXPLAT_RECV_DATA*)Packet;
+        if (Packet->RecvData.Buffer) {
+            Packet->RecvData.Allocated = TRUE;
+            Buffers[PacketCount++] = &Packet->RecvData;
         } else {
             CxPlatListPushEntry(&Queue->PartitionRxPool, (CXPLAT_SLIST_ENTRY*)Packet);
```
csujedihy, line 1583: >CxPlatListPushEntry(&Queue->PartitionRxPool, (CXPLAT_SLIST_ENTRY*)Packet); [](http://example.com/codeflow?start=12&length=74)

This is wrong now.

csujedihy:
>         Tail = &((SLIST_ENTRY*)Packet)->Next;

This line still assumes the first member of XDP_RX_PACKET is a slist_entry.

---
Refers to: src/platform/datapath_raw_xdp_win.c:1661 in cb21e8e. [](commit_id = cb21e8edc1a1b4ecfb5f70eb86bf821bce122a5a, deletion_comment = False)

src/platform/datapath_raw_xdp_win.c
nibanks: Talked over IM. We are reinterpreting the packet memory as a list entry, so it should be fine so long as we don't use both at the same time.

mtfriesen (approved):

src/platform/datapath_winuser.c
```diff
@@ -189,19 +190,20 @@ typedef struct CXPLAT_DATAPATH_INTERNAL_RECV_CONTEXT {
...
         ];
 
-} CXPLAT_DATAPATH_INTERNAL_RECV_CONTEXT;
+} DATAPATH_RX_IO_BLOCK;
 
-//
-// Internal receive context.
-//
-typedef struct CXPLAT_DATAPATH_INTERNAL_RECV_BUFFER_CONTEXT {
+typedef struct DECLSPEC_ALIGN(MEMORY_ALLOCATION_ALIGNMENT) DATAPATH_RX_PACKET {
```
csujedihy, line 195: >DECLSPEC_ALIGN(MEMORY_ALLOCATION_ALIGNMENT) [](http://example.com/codeflow?start=15&length=43)

Why do we need this alignment requirement here?

csujedihy: If it's actually needed, we then need a comment similar to what we have in the XDP datapath.

nibanks: I'm not sure if it's strictly necessary. It was copy/paste.

csujedihy: It doesn't look like it's necessary. There is no SLIST operations in this file.

I will leave it up to you.

csujedihy (approved):

src/platform/datapath_winuser.c
```diff
@@ -189,19 +190,20 @@ typedef struct CXPLAT_DATAPATH_INTERNAL_RECV_CONTEXT {
...
 
     //
-    // The owning allocation.
+    // Publicly visible receive data.
     //
-    CXPLAT_DATAPATH_INTERNAL_RECV_CONTEXT* RecvContext;
+    CXPLAT_RECV_DATA Data;
 
-} CXPLAT_DATAPATH_INTERNAL_RECV_BUFFER_CONTEXT;
+} DATAPATH_RX_PACKET;
```
csujedihy, line 206: I'd add a comment saying there's a client data following the last field.

nibanks: There is already a comment at the end of `CXPLAT_RECV_DATA` to that effect. I'm going to leave that as is for now.

PR #3827, opened 2023-08-23
More Refactoring of the QUIC Layer RX Packets

nibanks:
## Description

Updates most everything to use the QUIC_RX_PACKET instead of some using the cxplat struct.

## Testing

Existing CI

## Documentation

N/A

### DIFF
--- a/src/core/binding.c
+++ b/src/core/binding.c
@@ -625,11 +625,11 @@ QUIC_STATELESS_CONTEXT*
 QuicBindingCreateStatelessOperation(
     _In_ QUIC_BINDING* Binding,
     _In_ QUIC_WORKER* Worker,
-    _In_ CXPLAT_RECV_DATA* Datagram
+    _In_ QUIC_RX_PACKET* Packet
     )
 {
     uint32_t TimeMs = CxPlatTimeMs32();
-    const QUIC_ADDR* RemoteAddress = &Datagram->Route->RemoteAddress;
+    const QUIC_ADDR* RemoteAddress = &Packet->Route->RemoteAddress;
     uint32_t Hash = QuicAddrHash(RemoteAddress);
     QUIC_STATELESS_CONTEXT* StatelessCtx = NULL;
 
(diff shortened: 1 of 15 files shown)


src/core/binding.c
```diff
@@ -1430,15 +1411,14 @@ QuicBindingDropBlockedSourcePorts(
 _IRQL_requires_max_(DISPATCH_LEVEL)
```
csujedihy, line 1411:
```
//
// Looks up or creates a connection to handle a chain of datagrams.
// Returns TRUE if the datagrams were delivered, and FALSE if they should be
// dropped.
//
```

Should the comment above use the term packets as well?

src/core/binding.c
```diff
@@ -1694,7 +1674,7 @@ QuicBindingReceive(
         // Perform initial validation.
         //
         BOOLEAN ReleaseDatagram;
-        if (!QuicBindingPreprocessDatagram(Binding, Datagram, &ReleaseDatagram)) {
+        if (!QuicBindingPreprocessDatagram(Binding, (QUIC_RX_PACKET*)Datagram, &ReleaseDatagram)) {
```
csujedihy, line 1677: Should this function be renamed to QuicBindingPreprocessPacket?

src/core/binding.c
nibanks: Yes

src/core/binding.c
nibanks: Yes

src/core/binding.c
```diff
@@ -1652,7 +1632,7 @@ QuicBindingReceive(
         DatagramChain = Datagram->Next;
         Datagram->Next = NULL;
 
-        QUIC_RX_PACKET* Packet = GetQuicRxPacket(Datagram);
+        QUIC_RX_PACKET* Packet = (QUIC_RX_PACKET*)Datagram;
```
csujedihy, line 1635: It looks like we can completely get rid of the notion of datagram in this function now.

nibanks: We could, but I think it would add a lot of casts back and forth for all the chains that are built. The is the top level function from the platform layer, so it has to receive datagrams. For now, I was just going to leave this as is.

csujedihy (approved):

PR #3833, opened 2023-08-31
Add DocFx Integration

ProjectsByJackHe:
## Description

While working on supporting a new experimental feature (RELIABLE_RESET) https://github.com/microsoft/msquic/pull/3817 

It became apparent how painful it is digging through docs on markdown. 

Additionally, ChatGPT has little context and starts hallucinating and giving downright wrong answers due to a mixture of poor indexing and incomplete documentation in its training set.

To spearhead the effort to improve GPT integration, search engine indexing, and just having a nice searching feature (with built-in indexing), while minimizing the amount of work, DocFx fits our requirements perfectly.

A preview of DocFx is live here: https://projectsbyjackhe.github.io/MsQuicDocFx/docs/API.html 
(Try out the search feature on the top right bar 💯)

## Testing

N/A

## Documentation

Yes! In fact, you can check out a preview here: https://projectsbyjackhe.github.io/MsQuicDocFx/docs/API.html


scripts/update-doc-yml.py
```diff
@@ -0,0 +1,38 @@
+"""
+Author: Jack He (jackhe@microsoft.com)
```
csujedihy, line 2: I'd hide my corp email unless you wanna get spammed. There's no need to add a line for authorship.

ProjectsByJackHe: Oof

scripts/update-doc-yml.py
```diff
@@ -0,0 +1,38 @@
...
+
+This utility script is useful for creating a toc.yml file specifically for use with DocFx.
+
+DocFx is a dotnet tool for generating documentation from markdown files.
+It requires a toc.yml file to link markdowns to generate the static site.
+"""
+
```
csujedihy, line 9: IMO, we should always use the already established scripting language, which is powershell.

docfx.json
```diff
@@ -0,0 +1,46 @@
...
+      "memberLayout": "samePage",
+      "EnumSortOrder": "alphabetic",
+      "allowCompilationErrors": false
+    }
+  ],
+  "build": {
+    "content": [
+      {
+        "files": [ "**/*.{md,yml}" ],
+        "exclude": [ "_site/**", "obj/**" ]
```
csujedihy, line 26: `_site`? Do we have this folder?

scripts/update-doc-yml.py
```diff
@@ -0,0 +1,38 @@
+"""
+Author: Jack He (ProjectsByJackHe)
+
+This utility script is useful for creating a toc.yml file specifically for use with DocFx.
+
+DocFx is a dotnet tool for generating documentation from markdown files.
+It requires a toc.yml file to link markdowns to generate the static site.
+"""
+
+import os
```
csujedihy, line 10: IMO, generating toc.yml should be part of the github workflow. Is it really necessary to store toc.yml files in the repo (vs generating them and publishing them to GitHub pages)?

docfx.json
ProjectsByJackHe: It will be auto-generated by the Github runner and copied to the gh-Pages branch. Then, Github will automatically deploy whatever is in the gh-pages branch. That's what I have setup in the .github/workflow.

scripts/update-doc-yml.py
ProjectsByJackHe: That's a good point! 

According to the official DocFx Github (https://github.com/dotnet/docfx/tree/main/docs) we're actually not supposed to auto-generate toc.yml as that is our manifest for generating the site, and it's meant to be manually updated, but that's fine since it's isn't often we add / remove markdowns. 

With that said, I didn't want to type out manually the gigantic amount of markdown files we have, so I wrote a Python script for it. But in the future, we can certainly setup something in the build process to call my python / powershell script to dynamically create the toc.yml file, and have it ready just in time to build the full static site.

scripts/update-doc-yml.py
ProjectsByJackHe: Sounds good. I can update.

.github/workflows/publish-docfx.yml
```diff
@@ -6,6 +6,9 @@ on:
   push:
     branches: ["main"]
 
+  pull_request:
```
csujedihy, line 9: why do we need to trigger on PR events?

I think the only event that should trigger the deployment of the doc is push on main with a file filter set to the docs folder.

ProjectsByJackHe: Agreed, my initial reasoning was having this workflow run on PR would help reviewers see UI changes, but we could just include them in the description. I can remove this.

scripts/update-doc-yml.py
ProjectsByJackHe: Ok, I updated the powershell script to recursively generate a toc.yml for any subdirectories, so we could run that in the github workflow to create the toc.yml just in time in the github runner.

submodules/googletest
nibanks, line 1: Please revert the submodule changes.

toc.yml
```diff
@@ -0,0 +1,5 @@
+- name: MsQuic
```
nibanks, line 1: Can these new files be put in the `doc` folder instead of the root?

docfx.json
```diff
@@ -0,0 +1,46 @@
...
+    "fileMetadataFiles": [],
+    "template": [
+      "default",
+      "modern"
+    ],
+    "postProcessors": ["ExtractSearchIndex"],
+    "keepFileLink": false,
+    "disableGitFeatures": false
+  }
+}
```
nibanks, line 46: Generally, please leave trailing newlines at the end of all files.

scripts/generate-docfx-yml.ps1
```diff
@@ -0,0 +1,61 @@
+
+# Usage: ./generate-docfx-yml.ps1 <directoryPath>
+#
+# Recursively iterates directoryPath and creates a toc.yml in each subdirectory,
+# generating a table of contents for each markdown file. This is useful for docfx
+# as docfx requires a manifest for just in time static site compilation.
+#
+
+param (
+    [string]$directoryPath
```
nibanks, line 10: Any reason why this shouldn't be hardcoded or at least defaulted to the `docs` folder?

ProjectsByJackHe: Oh I wanted to keep it as generic as possible so other folks interested in docfx can reuse it and follow our workflow of dynamically generating toc.yml if they don't want to manually maintain their own yamls.

.gitignore
```diff
@@ -377,3 +377,6 @@ Cargo.lock
 
 # IntelliJ project files
 .idea/
+
+# docfx cache folder
+_site
```
csujedihy, line 382: nit: newline after

toc.yml
ProjectsByJackHe: I don't think so, because the docfx engine uses directory structure to build the site. Each subfolder has a toc.yml to define its manifest. This toc.yml file is the root manifest.

nibanks: I very strongly don't want any more files at the root of the repo. We need to keep things as clean as possible. Most people will never need to look at these files in the repo iteself (instead use the doc website).

I assume we should be able to have an extra step when generating the docfx directory structure that JIT moves things where we want them.

.github/workflows/publish-docfx.yml
```diff
@@ -0,0 +1,39 @@
...
+
+concurrency:
+  group: "pages"
+  cancel-in-progress: false
+jobs:
+  publish-docs:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
```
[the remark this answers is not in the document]
nibanks: Please update all these pinned dependencies (copy from other ymls or use the tool recommended).

.github/workflows/publish-docfx.yml
```diff
@@ -0,0 +1,39 @@
+
+name: Deploy DocFx
+
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["main"]
```
nibanks, line 7: Can we limit this to run only when documentation files are actually changed? We don't want to unnecessarily push a new commit for every non-docs change.

.github/workflows/publish-docfx.yml
```diff
@@ -0,0 +1,39 @@
+
+name: Deploy DocFx
+
+on:
+  # Runs on pushes targeting the default branch
+  push:
```
nibanks, line 6: Can we run this for PRs too, but just not commit the results?

.github/workflows/publish-docfx.yml
```diff
@@ -0,0 +1,37 @@
+
```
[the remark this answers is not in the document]
nibanks: You need to explicitly set `read` top-level permissions.

.github/workflows/publish-docfx.yml
```diff
@@ -0,0 +1,37 @@
...
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["main"]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+concurrency:
```
nibanks, line 12: See how we do concurrency in other yamls. If you add support to run (but not commit) this on PRs, then this will need to be updated.

.github/workflows/publish-docfx.yml
ProjectsByJackHe: Sure! Added.

.github/workflows/publish-docfx.yml
ProjectsByJackHe: Added paths/ check.

.github/workflows/publish-docfx.yml
ProjectsByJackHe: Updated.

toc.yml
ProjectsByJackHe: Ahh I see, I understand the concern. JIT sounds good!

.github/workflows/publish-docfx.yml
```diff
@@ -0,0 +1,52 @@
...
+on:
+  workflow_dispatch:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["main"]
+    paths: ["docs/**"]
+  
+  pull_request:
+    branches: ["main"]
+    paths: ["docs/**"]
```
nibanks, line 13: Include this yaml file

ProjectsByJackHe:
Note: Publishing DocFx Workflow will use `actions/setup-dotnet@3447fd6a9f9e57506b15f895c5b76d3b197dc7c2` Which is part of release V3.2.0: https://github.com/actions/setup-dotnet/releases/tag/v3.2.0

.github/workflows/publish-docfx.yml
ProjectsByJackHe: Done

nibanks (approved):

PR #3848, opened 2023-09-11
Migrate Send Logic to 64-bit Time

nibanks:
## Description

Updates the send logic to use 64-bit time instead of 32-bit. Makes calculating one-way delay time easier. Also required for larger RTT (i.e. space) scenarios.

## Testing

Existing automation

## Documentation

N/A

### DIFF
--- a/docs/Architecture.md
+++ b/docs/Architecture.md
@@ -28,6 +28,57 @@ It uses asynchronous callbacks driven by the current execution model (see [below
 
 Currently, it also has preview support of XDP on Windows.
 
+```mermaid
+flowchart TB
+    Datapath-.->p1[Partition]
+    Datapath-.->Socket
+    Datapath-.->p2[Partition]
+    subgraph p1[Partition]
+    rx1["RX IO
+         Pool"]
+    tx1["TX IO
+         Pool"]
+    end
+    subgraph p2[Partition]
+    rx2["RX IO
+         Pool"]
+    tx2["TX IO
+         Pool"]
+    end
+    Socket-.->q1
+    Socket-.->q2
+    subgraph q1[Queue]
+    plat_sock1[Platform Socket]
+    end
+    q1-.->p1
+    subgraph q2[Queue]
+    plat_sock2[Platform Socket]
+    end
+    q2-.->p2
+    subgraph rxio[RX IO]
+    rx_meta[Internal Data]
+    rx_packet[IP/Client Data]
+    rx_buffer[Packet Payload]
+    end
+    rx1-.->rxio
+    rx2-.->rxio
+    subgraph txio[TX IO]
+    tx_meta[Internal Data]
+    tx_packet[IP/Client Data]
+    tx_buffer[Packet Payload]
+    end
+    tx1-.->txio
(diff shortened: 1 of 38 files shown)


mtfriesen (approved):

src/core/loss_detection.c
```diff
@@ -334,23 +334,34 @@ QuicLossDetectionUpdateTimer(
...
+        // is an outstanding packet.
+        //
         const uint64_t DisconnectTime =
             OldestPacket->SentTime + MS_TO_US(Connection->Settings.DisconnectTimeoutMs);
-        if (TimeNow >= DisconnectTime) {
+        if (CxPlatTimeAtOrBefore64(DisconnectTime, TimeNow)) {
             Delay = 0;
         } else {
             const uint64_t MaxDelay = CxPlatTimeDiff64(TimeNow, DisconnectTime);
             if (Delay > MaxDelay) {
```
csujedihy, line 355: `Delay` is uninitialized here?

src/core/loss_detection.c
```diff
@@ -341,30 +341,28 @@ QuicLossDetectionUpdateTimer(
...
         Delay = CxPlatTimeDiff64(TimeNow, TimeFires);
+
+        if (OldestPacket != NULL) {
+            //
+            // Limit the timeout to the remainder of the disconnect timeout if there
+            // is an outstanding packet.
+            //
+            const uint64_t DisconnectTime =
+                OldestPacket->SentTime + MS_TO_US(Connection->Settings.DisconnectTimeoutMs);
+            if (CxPlatTimeAtOrBefore64(DisconnectTime, TimeNow)) {
```
csujedihy, line 354: Was there really a bug in the code? All 64-bit time related funtions (CxPlatTimeAtOrBefore64 and CxPlatTimeDiff64) assume no wrap around.

nibanks: The bug was that I was doing a diff before checking if the timer expired.

docs/Architecture.md
nibanks, line 1: accidental include

csujedihy (approved):

src/plugins/dbg/quictypes.h
```diff
@@ -931,8 +931,8 @@ struct SentPacketMetadata : Struct {
         return ReadType<UINT64>("PacketNumber");
     }
 
-    UINT32 SentTime() {
-        return ReadType<UINT32>("SentTime"); // Microseconds
+    UINT64 SentTime() {
+        return ReadType<UINT64>("SentTime"); // Microseconds
```
anrossi, line 935: What if a new debug plugin is used to debug old, 32-bit time structures? Will this handle it gracefully?

nibanks: No. We never gracefully handle this in our debugger plugin

anrossi (approved):

src/plugins/dbg/quictypes.h
anrossi: We need to use the debugger plugin from the same MsQuic build. good to know. Should document that

PR #3855, opened 2023-09-15
Set default toolchain file only for OneBranch/container builds

csujedihy:
## Description

.NET builds msquic using the same script (script/build.ps1) that we use to build the project, which does not work well with their cross-compile method where a toolchain file is not used as reported in https://github.com/dotnet/msquic/pull/149.

This PR allows empty toolchain file for cross-compile. So, we only use a default toolchain file for our own OneBranch/container builds.

## Testing

Needs help from @wfurt to test it.


nibanks (approved):

wfurt (approved):
LGTM. Thanks for quick fix @csujedihy

PR #3861, opened 2023-09-18
Filter Build Artifacts

nibanks:
## Description

Unnecessary stuff was being included in the build artifacts uploaded to GitHub. Clean them up.

## Testing

CI/CD

## Documentation

N/A

### DIFF
--- a/.github/workflows/build-reuse-win.yml
+++ b/.github/workflows/build-reuse-win.yml
@@ -87,6 +87,12 @@ jobs:
       if: inputs.test == ''
       shell: pwsh
       run: scripts/build.ps1 -Config ${{ inputs.config }} -Platform ${{ inputs.plat }} -Arch ${{ inputs.arch }} -Tls ${{ inputs.tls }} ${{ inputs.xdp }} ${{ inputs.sanitize }} ${{ inputs.static }}
+    - name: Filter Build Artifacts
+      shell: pwsh
+      run: |
+        Remove-Item artifacts/bin/xdp -Recurse -Force -ErrorAction Ignore
+        Remove-Item artifacts/corenet-ci-main -Recurse -Force -ErrorAction Ignore
+        Remove-Item artifacts/xdp -Recurse -Force -ErrorAction Ignore
     - name: Upload build artifacts
       uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32
       with:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-reuse-win.yml
```diff
@@ -87,6 +87,12 @@ jobs:
       if: inputs.test == ''
       shell: pwsh
       run: scripts/build.ps1 -Config ${{ inputs.config }} -Platform ${{ inputs.plat }} -Arch ${{ inputs.arch }} -Tls ${{ inputs.tls }} ${{ inputs.xdp }} ${{ inputs.sanitize }} ${{ inputs.static }}
+    - name: Filter Build Artifacts
+      shell: pwsh
+      run: |
+        Remove-Item artifacts/bin/xdp -Recurse -Force -ErrorAction Ignore
+        Remove-Item artifacts/corenet-ci-main -Recurse -Force -ErrorAction Ignore
+        Remove-Item artifacts/xdp -Recurse -Force -ErrorAction Ignore
```
csujedihy, line 95: Are you seeing an issue with these files? I remember it's sort of intentional, so the test workflow won't need to download them again.

nibanks: I'm trying to reduce the size to optimize download/copy times. For one, there are two copies of the XDP binaries. Do we really need both? Are you sure the other stuff is used?

csujedihy: I don't remember why we have 2 copies of xdp. corenet-ci-main is uploaded so the tests for duonic/kernel do not have to download stuff externally.

How much space have you saved though? I remember I tried to trim the artifacts and I then realized the gain was only like tens of MBs or even less.

csujedihy: We probably don't need xdp binaries in tests anymore as we now install the msi when preparing the environment?

nibanks: Here's a before and after comparison:
```
Release-windows-windows-2022-x64-schannel     204 MB	
Release-winkernel-windows-2022-x64-schannel   104 MB

Release-windows-windows-2022-x64-schannel     121 MB
Release-winkernel-windows-2022-x64-schannel    21 MB
```

nibanks: So far, tests seem to be passing, so I don't think we need any of the other stuff.

csujedihy: They are for speeding up the tests and reducing flakiness caused by downloading extra stuff because prepare-machine.ps1 will download them if not present.

I will leave it up to you. Not a big deal to me.

csujedihy (approved):

.github/workflows/build-reuse-win.yml
csujedihy: (Removing them saves a lot of space for release schannel but not so much for openssl)

PR #3867, opened 2023-09-23
Disable Segmentation Support on EIO

### LINKED ISSUE #3865: [msquic2.2] linux sendmmsg failed via GSO disabled interface
### Describe the bug

msquic failed to send out udp on the interface which GSO is disabled.

log shows
```
[data][0xfffee0003570] ERROR, 5, sendmmsg failed. [DatapathErrorStatus:/home/ubuntu/repo/quic/msquic/src/platform/datapath_epoll.c:2468]
```

### Affected OS

- [ ] Windows
- [X] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

mostly with VM because the virtual network interface does not support GSO.

### MsQuic version

v2.2

1. have loopback interface GSO enabled
2. have external interface GSO disabled.
3. start a connection via the external interface
4. The call return success 

### Expected behavior

msquic should emit some udp out from the external interface for handshake.

### Actual outcome

silent.
after a while (timeout) get a connection callback for transport down due to idle. 

### Additional details

# Causes
The GSO support is a runtime decision during init done by CxPlatDataPathCalculateFeatureSupport

The CxPlatDataPathCalculateFeatureSupport judges/test the GSO support with remote addr `127.0.0.1` which is always from a loopback interface. 

```
# external interface
ethtool -k enp0s1
Features for enp0s1:
...
generic-segmentation-offload: off [requested on]
...
```
```
#loopback
ethtool -k lo
Features for lo:
...
generic-segmentation-offload: on
...
```
The bug is triggered In an env that loopback has GSO on but the external interface has GSO disabled so traffic via the external interface will fail.

# Workaround
a. Turn GSO on on the external interface if you can
b. turnoff GSO on loopback device.
```
ethtool -K lo gso off
```

# Solution

As talked to Nick, we could intro `QUIC_EXECUTION_CONFIG_FLAGS` to enable/disable the feature rather than using auto-detect.

nibanks:
## Description

Update Linux epoll datapath to fallback to non-GSO sends if the `EIO` error is returned, which most likely indicates the NIC doesn't support it. This seems to be the case on at least some Android devices.

Fixed #3865.

## Testing

CI/CD for regression testing. Not sure how to actually test this fallback atm.

## Documentation

N/A

### DIFF
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -2450,12 +2450,34 @@ CxPlatSendDataSend(
             Status = QUIC_STATUS_PENDING;
         } else {
             Status = errno;
+#ifdef UDP_SEGMENT
+            QuicTraceEvent(
+                DatapathErrorStatus,
+                "[data][%p] ERROR, %u, %s.",
+                SocketContext->Binding,
+                Status,
+                "sendmsg (GSO) failed");
+#else
             QuicTraceEvent(
                 DatapathErrorStatus,
                 "[data][%p] ERROR, %u, %s.",
                 SocketContext->Binding,
                 Status,
                 "sendmmsg failed");
+#endif
+
+            if (Status == EIO) {
+                //
+                // EIO generally indicates the GSO isn't supported by the NIC,
+                // so disable segmentation on the datapath globally.
+                //
+                QuicTraceEvent(
+                    LibraryError,
+                    "[ lib] ERROR, %s.",
+                    "Disabling segmentation support globally");
+                SendData->SocketContext->Binding->Datapath->Features &=
+                    ~CXPLAT_DATAPATH_FEATURE_SEND_SEGMENTATION;
+            }
 
             //
             // Unreachable events can sometimes come synchronously.
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -2466,7 +2466,8 @@ CxPlatSendDataSend(
                 "sendmmsg failed");
 #endif
 
-            if (Status == EIO) {
+            if (Status == EIO &&
+                SocketContext->Binding->Datapath->Features & CXPLAT_DATAPATH_FEATURE_SEND_SEGMENTATION) {
                 //
                 // EIO generally indicates the GSO isn't supported by the NIC,
                 // so disable segmentation on the datapath globally.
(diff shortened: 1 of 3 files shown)


src/platform/datapath_epoll.c
```diff
@@ -2450,12 +2450,34 @@ CxPlatSendDataSend(
...
+#else
             QuicTraceEvent(
                 DatapathErrorStatus,
                 "[data][%p] ERROR, %u, %s.",
                 SocketContext->Binding,
                 Status,
                 "sendmmsg failed");
+#endif
+
+            if (Status == EIO) {
```
jwhited, line 2469: It may be worth also checking that the feature was used on the syscall and/or is still enabled. Unlikely to see EIO for other reasons, but will be confusing if so.

src/platform/datapath_epoll.c
```diff
@@ -2450,12 +2450,34 @@ CxPlatSendDataSend(
...
+                //
+                // EIO generally indicates the GSO isn't supported by the NIC,
+                // so disable segmentation on the datapath globally.
+                //
+                QuicTraceEvent(
+                    LibraryError,
+                    "[ lib] ERROR, %s.",
+                    "Disabling segmentation support globally");
+                SendData->SocketContext->Binding->Datapath->Features &=
```
jwhited, line 2478: Is it safe to assume the transport will retry by nature in the common case?

It may be worth segmenting and retrying to reduce cost of transport retrying. See also the comment above about being sure UDP_SEGMENT was involved, as we wouldn't want to get stuck retrying forever.

nibanks: Yes, the transport will retry. Perhaps in the future we can consider segmenting and trying again, but it's not strictly required right now.

mtfriesen:
Interesting. Does Linux GSO not perform software fallback?

src/platform/datapath_epoll.c
```diff
@@ -2450,12 +2450,35 @@ CxPlatSendDataSend(
...
+                SocketContext->Binding->Datapath->Features & CXPLAT_DATAPATH_FEATURE_SEND_SEGMENTATION) {
+                //
+                // EIO generally indicates the GSO isn't supported by the NIC,
+                // so disable segmentation on the datapath globally.
+                //
+                QuicTraceEvent(
+                    LibraryError,
+                    "[ lib] ERROR, %s.",
+                    "Disabling segmentation support globally");
+                SocketContext->Binding->Datapath->Features &=
```
mtfriesen, line 2479: Ideally this would a per-interface disable, right?

mtfriesen (approved):

src/platform/datapath_epoll.c
nibanks: Probably, that'd be the ideal solution. But practically, this is enough for now.

jwhited:
> Interesting. Does Linux GSO not perform software fallback?

In the case of software UDP GSO, Linux still requires the egress netdev to offload checksumming: https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/commit/man7/udp.7?id=806eabd74910447f21005160e90957bde4db0183

PR #3879, opened 2023-09-25
Expose Interface for App to Set Stateless Reset Key

### LINKED ISSUE #1719: Allow stateless reset token key to be set
In certain deployment scenarios, all machines in a fleet need to have the same stateless reset token key in order to send stateless reset tokens. Currently, each instance of MsQuic uses a different stateless reset token key.   Adding a global param and/or a setting, and a registry value, to set the stateless reset token key is needed in these environments.

ukingliu:
## Description
At present, each MsQuic server utilizes a distinct stateless reset key. When the MsQuic client and server successfully complete the handshake process, they establish communication through an encrypted channel, utilizing an encryption key generated from the respective stateless reset key for their interactions. However, in the event of a server crash followed by a restart, the client remains unaware of this disruption, leaving it unable to communicate with the server due to the unavailability of the encryption key. To resolve this issue, we have introduced an interface that allows the App to set the same stateless reset key on the MsQuic server side.

Fixes #1719.

## Testing
We have introduced two new test cases. The first one is designed to validate the length of the stateless reset key to ensure its validity. The second test case focuses on validating the behavior when the stateless reset key is changed and the server is subsequently shut down silently, leading to a client timeout instead of abort scenario.

## Documentation
No.


src/test/MsQuicTests.h
```diff
@@ -791,6 +796,9 @@ typedef struct {
     QUIC_CTL_CODE(27, METHOD_BUFFERED, FILE_WRITE_DATA)
     // int - Family
 
+#define IOCTL_QUIC_RUN_STATELESS_RESETKEY \
+    QUIC_CTL_CODE(28, METHOD_BUFFERED, FILE_WRITE_DATA)
```
nibanks, line 800: `28` is already used. You need to add this to the end of the list/file.

src/inc/msquic.h
```diff
@@ -70,6 +70,11 @@ typedef _In_range_(0, QUIC_UINT62_MAX) uint64_t QUIC_UINT62;
 //
 #define QUIC_MAX_RESUMPTION_APP_DATA_LENGTH     1000
 
+//
+// The number of bytes of stateless reset key
```
nibanks, line 74:
```suggestion
// The number of bytes of stateless reset key.
```

src/test/lib/DataTest.cpp
```diff
@@ -691,7 +691,7 @@ ListenerAcceptConnectionAndStreams(
 void
 QuicTestClientDisconnect(
     bool StopListenerFirst
-    )
+)
```
nibanks, line 694: Please revert

src/test/lib/DataTest.cpp
```diff
@@ -788,6 +788,118 @@ QuicTestClientDisconnect(
     }
 }
 
+void
+QuicTestStatelessResetKey(
+    bool ChangeStatelessResetKey
+)
```
nibanks, line 794:
```suggestion
void
QuicTestStatelessResetKey(
    bool ChangeStatelessResetKey
    )
```

src/test/lib/DataTest.cpp
```diff
@@ -788,6 +788,118 @@ QuicTestClientDisconnect(
...
+                    MsQuic->SetParam(
+                        nullptr,
+                        QUIC_PARAM_GLOBAL_STATELESS_RESETKEY,
+                        sizeof(statelessResetKey),
+                        statelessResetKey))) {
+                    TEST_FAILURE("Failed to set stateless reset key %d\n", Status);
+                }
+            }
+
+            CxPlatSleep(15); // Sleep for just a bit.
```
nibanks, line 892: I think you'll want to sleep _before_ changing the key.

src/test/lib/DataTest.cpp
```diff
@@ -788,6 +788,118 @@ QuicTestClientDisconnect(
...
+            TEST_TRUE(Client->GetIsConnected());
+
+            TEST_NOT_EQUAL(nullptr, Server);
+            if (!Server->WaitForConnectionComplete()) {
+                return;
+            }
+            TEST_TRUE(Server->GetIsConnected());
+
+            if (ChangeStatelessResetKey)
+            {
```
nibanks, line 877:
```suggestion
            if (ChangeStatelessResetKey) {
```

src/test/lib/DataTest.cpp
```diff
@@ -788,6 +788,118 @@ QuicTestClientDisconnect(
...
+
+            Client->SetExpectedTransportCloseStatus(ClientStats.ExpectedCloseStatus);
+            TEST_QUIC_SUCCEEDED(Client->SetDisconnectTimeout(1000)); // ms
+
+            if (!SendPingBurst(
+                Client,
+                ClientStats.StreamCount,
+                ClientStats.PayloadLength)) {
+                return;
+            }
```
nibanks, line 856:
```suggestion
            if (!SendPingBurst(
                    Client,
                    ClientStats.StreamCount,
                    ClientStats.PayloadLength)) {
                return;
            }
```

src/core/library.c
```diff
@@ -1112,6 +1112,30 @@ QuicLibrarySetGlobalParam(
...
+            CxPlatLockAcquire(&PerProc->ResetTokenLock);
+            CxPlatHashFree(PerProc->ResetTokenHash);
+            Status =
+                CxPlatHashCreate(
+                    CXPLAT_HASH_SHA256,
+                    (uint8_t*)Buffer,
+                    QUIC_STATELESS_RESETKEY_LENGTH * sizeof(uint8_t),
+                    &PerProc->ResetTokenHash);
+            if (QUIC_FAILED(Status)) {
+                return Status;
```
nibanks, line 1132: Need to set the pointer to NULL on failure, so it's not tried to free again later. 
```suggestion
                PerProc->ResetTokenHash = NULL;
                return Status;
```
But actually, the more I think about it, if we fail and leave one of these NULL, we're in bad shape. We should probably:

1. create the new hash first
2. if it succeeds:
  a. lock
  b. swap the pointers
  c. unlock
  d. free the old hash

src/inc/msquic.h
```diff
@@ -832,7 +837,7 @@ void
 #define QUIC_PARAM_GLOBAL_EXECUTION_CONFIG              0x01000009  // QUIC_EXECUTION_CONFIG
 #endif
 #define QUIC_PARAM_GLOBAL_TLS_PROVIDER                  0x0100000A  // QUIC_TLS_PROVIDER
-
+#define QUIC_PARAM_GLOBAL_STATELESS_RESETKEY            0x0100000B  // uint8_t[] Array size is QUIC_STATELESS_RESETKEY_LENGTH
```
nibanks, line 840: Slight name change suggestion:
```suggestion
#define QUIC_PARAM_GLOBAL_STATELESS_RESET_KEY           0x0100000B  // uint8_t[] Array size is QUIC_STATELESS_RESETKEY_LENGTH
```

src/test/bin/winkernel/control.cpp
```diff
@@ -768,6 +768,11 @@ QuicTestCtlEvtIoDeviceControl(
         QuicTestCtlRun(QuicTestClientDisconnect(Params->StopListenerFirst));
         break;
 
+    case IOCTL_QUIC_RUN_STATELESS_RESETKEY:
```
nibanks, line 771: Once you make this define a _new_ code, you should move this to the end, and update `QUIC_IOCTL_BUFFER_SIZES` with the new size.

src/test/bin/quic_gtest.cpp
```diff
@@ -1878,6 +1878,17 @@ TEST(Misc, ClientDisconnect) {
     }
 }
 
+TEST(Misc, StatelessResetKey) {
+    TestLogger Logger("QuicTestStatelessResetKey");
+    if (TestingKernelMode) {
+        uint8_t Param = 1;
+        ASSERT_TRUE(DriverClient.Run(IOCTL_QUIC_RUN_STATELESS_RESETKEY, Param));
+    }
+    else {
```
nibanks, line 1887:
```suggestion
    } else {
```

src/test/bin/quic_gtest.cpp
```diff
@@ -1878,6 +1878,17 @@ TEST(Misc, ClientDisconnect) {
...
+        QuicTestStatelessResetKey(true);
```
nibanks, line 1888: If we're always going to set this to `true` then there isn't much reason to pass a parameter.

src/manifest/clog.sidecar
nibanks, line 1: This diff looks a LOT bigger than normal...

src/core/library.c
ukingliu: How about we do the same way as random reset key function, we call MsQuicLibraryFreePartitions() to reset PerProc->ResetTokenHash at failure case?

src/test/lib/DataTest.cpp
ukingliu: My bad, will revert this change.

src/test/lib/DataTest.cpp
ukingliu: I'll do the change in coming commit

src/test/lib/DataTest.cpp
ukingliu: I'll address this change in comming commit.

src/core/library.c
ukingliu: Make sense. One more thing, should we "free the old hash" before "swap the pointers"?

src/core/library.c
```diff
@@ -1112,6 +1112,31 @@ QuicLibrarySetGlobalParam(
...
+            break;
+        }
+
+        CXPLAT_HASH* tokenHash = NULL;
+        Status =
+            CxPlatHashCreate(
+                CXPLAT_HASH_SHA256,
+                (uint8_t*)Buffer,
+                QUIC_STATELESS_RESET_KEY_LENGTH * sizeof(uint8_t),
+                &tokenHash);
```
nibanks, line 1127: You still need to do this in the loop, once per proc.

src/inc/msquic.h
```diff
@@ -832,7 +837,7 @@ void
 #define QUIC_PARAM_GLOBAL_EXECUTION_CONFIG              0x01000009  // QUIC_EXECUTION_CONFIG
 #endif
 #define QUIC_PARAM_GLOBAL_TLS_PROVIDER                  0x0100000A  // QUIC_TLS_PROVIDER
-
+#define QUIC_PARAM_GLOBAL_STATELESS_RESET_KEY            0x0100000B  // uint8_t[] Array size is QUIC_STATELESS_RESET_KEY_LENGTH
```
nibanks, line 840:
```suggestion
#define QUIC_PARAM_GLOBAL_STATELESS_RESET_KEY           0x0100000B  // uint8_t[] Array size is QUIC_STATELESS_RESET_KEY_LENGTH
```

src/test/bin/winkernel/control.cpp
```diff
@@ -768,6 +769,11 @@ QuicTestCtlEvtIoDeviceControl(
         QuicTestCtlRun(QuicTestClientDisconnect(Params->StopListenerFirst));
         break;
 
+    case IOCTL_QUIC_RUN_STATELESS_RESET_KEY:
+        CXPLAT_FRE_ASSERT(Params != nullptr);
+        QuicTestCtlRun(QuicTestStatelessResetKey(Params->StatelessResetKey));
+        break;
```
nibanks, line 775: Put this at the end of the switch statement please.

src/test/lib/ApiTest.cpp
```diff
@@ -2654,6 +2654,26 @@ void QuicTestGlobalParam()
...
 
+    //
+    // QUIC_PARAM_GLOBAL_STATELESS_RESET_KEY
+    //
+    {
+        uint8_t statelessResetkey[QUIC_STATELESS_RESET_KEY_LENGTH - 1];
+        CxPlatRandom(sizeof(statelessResetkey), statelessResetkey);
+        TestScopeLogger LogScope0("QUIC_PARAM_GLOBAL_STATELESS_RESET_KEY");
+        {
+            // should the global setting hard-coded?
```
nibanks, line 2665: What does this comment mean? And is it necessary?

src/test/lib/DataTest.cpp
```diff
@@ -788,6 +788,117 @@ QuicTestClientDisconnect(
     }
 }
 
+void
+QuicTestStatelessResetKey(
+    bool ChangeStatelessResetKey
+    )
```
nibanks, line 794:
```suggestion
void
QuicTestStatelessResetKey(
    )
```
Just don't pass the parameter and always change the key.

src/core/library.c
ukingliu: In this case, if CxPlatHashCreate failed in the middle of the loop, should we free all proc's ResetTokenHash?

ukingliu: Code like below looks right?

    case QUIC_PARAM_GLOBAL_STATELESS_RESET_KEY:
        if (BufferLength != QUIC_STATELESS_RESET_KEY_LENGTH * sizeof (uint8_t)) {
            break;
        }

		Status = QUIC_STATUS_SUCCESS;
        for (uint16_t i = 0; i < MsQuicLib.ProcessorCount; ++i) {
            CXPLAT_HASH* tokenHash = NULL;
            Status =
                CxPlatHashCreate(
                    CXPLAT_HASH_SHA256,
                    (uint8_t*)Buffer,
                    QUIC_STATELESS_RESET_KEY_LENGTH * sizeof(uint8_t),
                    &tokenHash);
            if (QUIC_SUCCEEDED(Status))
            {
                QUIC_LIBRARY_PP* PerProc = &MsQuicLib.PerProc[i];
                CxPlatLockAcquire(&PerProc->ResetTokenLock);
                CxPlatHashFree(PerProc->ResetTokenHash);
                PerProc->ResetTokenHash = tokenHash;
                CxPlatLockRelease(&PerProc->ResetTokenLock);
            } else {				
				for (uint16_t i = 0; i < MsQuicLib.ProcessorCount; ++i) {
					// should we free all tokenHash?
					QUIC_LIBRARY_PP* PerProc = &MsQuicLib.PerProc[i];
					CxPlatLockAcquire(&PerProc->ResetTokenLock);
					CxPlatHashFree(PerProc->ResetTokenHash);
					PerProc->ResetTokenHash = NULL;
					CxPlatLockRelease(&PerProc->ResetTokenLock);					
				}
				break;
            }
        }       
        break;

src/manifest/clog.sidecar
ukingliu, line 1: I replaced the file from main branch, but still show changes on this file.

src/core/library.c
nibanks: No need to have the other loop that cleans up. If you fail to create a new hash, you just leave the old/other ones and bail.

src/test/lib/ApiTest.cpp
ukingliu: will remove it.

src/test/MsQuicTests.h
```diff
@@ -1210,4 +1215,8 @@ typedef struct {
     QUIC_CTL_CODE(112, METHOD_BUFFERED, FILE_WRITE_DATA)
     // QUIC_RUN_FEATURE_NEGOTIATION
 
-#define QUIC_MAX_IOCTL_FUNC_CODE 112
+#define IOCTL_QUIC_RUN_STATELESS_RESET_KEY \
+    QUIC_CTL_CODE(113, METHOD_BUFFERED, FILE_WRITE_DATA)
+    // uint8_t - StatelessResetKey
```
nibanks, line 1220:
```suggestion
    QUIC_CTL_CODE(113, METHOD_BUFFERED, FILE_WRITE_DATA)
```

src/inc/msquic.h
```diff
@@ -832,7 +837,7 @@ void
 #define QUIC_PARAM_GLOBAL_EXECUTION_CONFIG              0x01000009  // QUIC_EXECUTION_CONFIG
 #endif
 #define QUIC_PARAM_GLOBAL_TLS_PROVIDER                  0x0100000A  // QUIC_TLS_PROVIDER
-
+#define QUIC_PARAM_GLOBAL_STATELESS_RESET_KEY           0x0100000B  // uint8_t[] - Array size is QUIC_STATELESS_RESET_KEY_LENGTH
```
nibanks, line 840: We should also update spinquic.cpp to support setting this.

src/core/library.c
```diff
@@ -1112,6 +1112,33 @@ QuicLibrarySetGlobalParam(
...
+                    &tokenHash);
+            if (QUIC_SUCCEEDED(Status)) {
+                QUIC_LIBRARY_PP* PerProc = &MsQuicLib.PerProc[i];
+                CxPlatLockAcquire(&PerProc->ResetTokenLock);
+                CxPlatHashFree(PerProc->ResetTokenHash);
+                PerProc->ResetTokenHash = tokenHash;
+                CxPlatLockRelease(&PerProc->ResetTokenLock);
+            } else {
+                break;
+            }
```
nibanks, line 1138:
```suggestion
        for (uint16_t i = 0; i < MsQuicLib.ProcessorCount; ++i) {
            CXPLAT_HASH* TokenHash = NULL;
            Status =
                CxPlatHashCreate(
                    CXPLAT_HASH_SHA256,
                    (uint8_t*)Buffer,
                    QUIC_STATELESS_RESET_KEY_LENGTH * sizeof(uint8_t),
                    &TokenHash);
            if (QUIC_FAILED(Status)) {
                break;
            }

            QUIC_LIBRARY_PP* PerProc = &MsQuicLib.PerProc[i];
            CxPlatLockAcquire(&PerProc->ResetTokenLock);
            CxPlatHashFree(PerProc->ResetTokenHash);
            PerProc->ResetTokenHash = TokenHash;
            CxPlatLockRelease(&PerProc->ResetTokenLock);
```
Minor style suggestions. Also, we always capitalize variables.

src/test/lib/DataTest.cpp
```diff
@@ -788,6 +788,113 @@ QuicTestClientDisconnect(
...
+            QUIC_STATUS Status;
+            if (QUIC_FAILED(
+                Status =
+                MsQuic->SetParam(
+                    nullptr,
+                    QUIC_PARAM_GLOBAL_STATELESS_RESET_KEY,
+                    sizeof(statelessResetKey),
+                    statelessResetKey))) {
+                TEST_FAILURE("Failed to set stateless reset key %d\n", Status);
+            }
```
nibanks, line 887:
```suggestion
            TEST_QUIC_SUCCEEDED(
                MsQuic->SetParam(
                    nullptr,
                    QUIC_PARAM_GLOBAL_STATELESS_RESET_KEY,
                    sizeof(statelessResetKey),
                    statelessResetKey));
```

src/test/lib/DataTest.cpp
```diff
@@ -788,6 +788,113 @@ QuicTestClientDisconnect(
...
+void
+QuicTestStatelessResetKey(
+    )
+{
+    //
+    // If the listener is stopped at the same time the server side of the
+    // connection is silently closed, then the UDP binding will also be cleaned
+    // up. This means the endpoint will no longer send Stateless Reset packets
+    // back to the client as it continues to receive the client's UDP packets.
+    //
```
nibanks, line 800: This comment should be updated to reflect what exactly is being tested here; essentially, that by changing the stateless reset key, the stateless reset packets the client receives after the server side is shut down no longer match, eventually resulting in a timeout on the client instead of an abort.

nibanks (commented):
Great job! Just few small things left and this should be good to merge.

nibanks:
Looks like the [CheckDotnet](https://github.com/microsoft/msquic/actions/runs/6305251397/job/17118300214?pr=3879) test failed because you need to update the generated file again. Please run `.\scripts\generate-dotnet.ps1`.

src/tools/spin/spinquic.cpp
```diff
@@ -1377,6 +1377,18 @@ CXPLAT_THREAD_CALLBACK(RunThread, Context)
...
+        if (0 == GetRandom(4)) {
+            uint8_t StatelessResetKey[QUIC_STATELESS_RESET_KEY_LENGTH];
+            CxPlatRandom(sizeof(StatelessResetKey), StatelessResetKey);
+            if (!QUIC_SUCCEEDED(MsQuic.SetParam(
+                nullptr,
+                QUIC_PARAM_GLOBAL_STATELESS_RESET_KEY,
+                sizeof(StatelessResetKey),
+                StatelessResetKey))) {
+                break;
+            }
```
nibanks, line 1389:
```suggestion
            if (!QUIC_SUCCEEDED(MsQuic.SetParam(
                    nullptr,
                    QUIC_PARAM_GLOBAL_STATELESS_RESET_KEY,
                    sizeof(StatelessResetKey),
                    StatelessResetKey))) {
                break;
            }
```
Can you also put this up after the other global params, on line 1327? Thanks!

nibanks:
**CheckDotnet** is still failing. Please fix that up.

**CIFuzz** is failing because spinquic is setting the global paramter before a registration is created (good catch):
```
==40==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000[110](https://github.com/microsoft/msquic/actions/runs/6321366569/job/17165278108?pr=3879#step:5:111) (pc 0x7f686229bfc4 bp 0x7f685c3f4cd0 sp 0x7f685c3f4c08 T10)
==40==The signal is caused by a READ memory access.
==40==Hint: address points to the zero page.
SCARINESS: 10 (null-deref)
    #0 0x7f686229bfc4 in __pthread_mutex_lock (/lib/x86_64-linux-gnu/libpthread.so.0+0xafc4) (BuildId: 7b4536f41cdaa5888408e82d0836e33dcf436466)
    #1 0x4eb114 in __interceptor_pthread_mutex_lock /src/llvm-project/compiler-rt/lib/asan/../sanitizer_common/sanitizer_common_interceptors.inc
    #2 0x584379 in QuicLibrarySetGlobalParam /src/msquic/src/core/library.c:1135:13
    #3 0x86fb89 in MsQuicSetParam /src/msquic/src/core/api.c:1383:18
    #4 0x57a997 in RunThread(void*) /src/./msquic/src/tools/spin/spinquic.cpp:1330:18
    #5 0x7f6862299608 in start_thread (/lib/x86_64-linux-gnu/libpthread.so.0+0x8608) (BuildId: 7b4536f41cdaa5888408e82d0836e33dcf436466)
    #6 0x7f6862193132 in __clone (/lib/x86_64-linux-gnu/libc.so.6+0x11f132) (BuildId: 1878e6b475720c7c51969e69ab2d276fae6d1dee)
```
The code in library.c needs to handle this case (just fail with invalid state if not initialized yet).

nibanks:
Oh, and we do have preexisting docs for parameters in settings.md. Please update those docs for this new setting.

src/tools/spin/spinquic.cpp
```diff
@@ -1325,6 +1325,9 @@ CXPLAT_THREAD_CALLBACK(RunThread, Context)
         }
 
         if (0 == GetRandom(4)) {
+            if (!MsQuicLib.LazyInitComplete) {
+                break;
+            }
```
nibanks, line 1330: This needs to be put in the `library.c` code, not in spinquic.

ukingliu: Thanks, will do that

src/test/lib/ApiTest.cpp
```diff
@@ -2663,13 +2663,12 @@ void QuicTestGlobalParam()
...
-                    nullptr,
-                    QUIC_PARAM_GLOBAL_STATELESS_RESET_KEY,
-                    sizeof(StatelessResetkey),
-                    StatelessResetkey));
+            QUIC_STATUS Status = MsQuic->SetParam(
+                nullptr,
+                QUIC_PARAM_GLOBAL_STATELESS_RESET_KEY,
+                sizeof(StatelessResetkey),
+                StatelessResetkey);
+            TEST_TRUE((Status == QUIC_STATUS_INVALID_STATE) || (Status == QUIC_STATUS_INVALID_PARAMETER));
```
nibanks, line 2671: Can we split these out into two separate tests, and explicitly trigger each scenario separately?

docs/Settings.md
```diff
@@ -107,8 +107,9 @@ These parameters are accessed by calling [GetParam](./api/GetParam.md) or [SetPa
...
 | `QUIC_PARAM_GLOBAL_VERSION_SETTINGS`<br> 7        | QUIC_VERSIONS_SETTINGS  | Both      | Globally change version settings for all subsequent connections.                                      |
 | `QUIC_PARAM_GLOBAL_LIBRARY_GIT_HASH`<br> 8        | char[64]                | Get-only  | Git hash used to build MsQuic (null terminated string)                                                |
-| `QUIC_PARAM_GLOBAL_EXECUTION_CONFIG`<br> 9        | QUIC_EXECUTION_CONFIG   | Both      | Globally configure the execution model used for QUIC. Must be set before opening registration.      |
+| `QUIC_PARAM_GLOBAL_EXECUTION_CONFIG`<br> 9        | QUIC_EXECUTION_CONFIG   | Both      | Globally configure the execution model used for QUIC. Must be set before opening registration.        |
 | `QUIC_PARAM_GLOBAL_TLS_PROVIDER`<br> 10           | QUIC_TLS_PROVIDER       | Get-Only  | The TLS provider being used by MsQuic for the TLS handshake.                                          |
+| `QUIC_PARAM_GLOBAL_STATELESS_RESET_KEY`<br> 11    | uint8_t[]               | Both      | Globally change the stateless reset key for all subsequent connections.                               |
```
nibanks, line 112: It's not `Both`, right? You only added `Set` support. Let's change this to `Set-Only`.

nibanks (approved):
Great job! Thanks for doing this!

src/test/lib/DataTest.cpp
```diff
@@ -788,6 +788,108 @@ QuicTestClientDisconnect(
...
+            uint8_t StatelessResetKey[QUIC_STATELESS_RESET_KEY_LENGTH];
+            CxPlatRandom(sizeof(StatelessResetKey), StatelessResetKey);
+            TEST_QUIC_SUCCEEDED(
+                MsQuic->SetParam(
+                    nullptr,
+                    QUIC_PARAM_GLOBAL_STATELESS_RESET_KEY,
+                    sizeof(StatelessResetKey),
+                    StatelessResetKey));
+
+            Server->Shutdown(QUIC_CONNECTION_SHUTDOWN_FLAG_SILENT, 0);
```
anrossi, line 884: The client doesn't try to send anything to the server once it's gone, are you sure the reset token gets used?

nibanks: `SendPingBurst` above should be doing that. This largely just copies from a previous test.

anrossi: Nevermind, I see it now.

PR #3882, opened 2023-09-26
fix: leakage of NumaNodeMasks

qzhuyan:
## Description

fix a NumaNodeMasks leakage.

## Testing

## Documentation

N/A

### DIFF
--- a/src/platform/platform_posix.c
+++ b/src/platform/platform_posix.c
@@ -210,6 +210,9 @@ CxPlatSystemUnload(
     void
     )
 {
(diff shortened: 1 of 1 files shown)


nibanks (approved):
Thanks for the fix! I don't know how I missed this when I added the code originally!

qzhuyan:
No rush.

I am checking if it is ok to free since it is used by  `pthread_setaffinity_np`

> Thanks for the fix! I don't know how I missed this when I added the code originally!

PR #3893, opened 2023-10-02
Remove unnecessary logging, be more careful which SEND flags we clear.

### LINKED ISSUE #3891: Misc.StreamReliableReset Test Failed
### Describe the bug

During a unrelated PR run, this [test failed once](https://github.com/microsoft/msquic/actions/runs/6371572667/job/17296060306?pr=3890) (in kernel mode) and 2 other kernel builds crashed ([[1]](https://github.com/microsoft/msquic/actions/runs/6371572667/job/17296060376?pr=3890) [[2]](https://github.com/microsoft/msquic/actions/runs/6371572667/job/17296060445?pr=3890)), which may be related (we don't know for sure).

From [quic.log](https://github.com/microsoft/msquic/files/12782313/quic.log), I see:
```
[1]18E8.055C::2023/10/01-19:31:55.303198100 [Microsoft-Quic][test] File: D:\a\msquic\msquic\src\test\lib\DataTest.cpp, Function: QuicTestStreamReliableReset, Line: 3398
[1]18E8.055C::2023/10/01-19:31:55.303200300 [Microsoft-Quic][test] FAIL: Context.ClientStreamShutdownComplete.WaitTimeout(TestWaitTimeout) not true
[1]18E8.055C::2023/10/01-19:31:55.303271700 [Microsoft-Quic][test]<--- Try Reliably Shutting Down Stream 1
```

I noticed a few things from the logs:

1. We shouldn't be trying to run any drop-based logic on Release mode (since it's not supported). I forgot about this when I did the review.
2. In the first loop of the test, even though the app indicated a reliable offset of 5000, the send logic still sends all 10000 bytes. It should stop sending at the reliable offset to save space.
3. `QUIC_TRACE_API_TYPE` is not in sync with `map_QUIC_TRACE_API_TYPE` in the ETW manifest.
4. In the second loop, which eventually, times out, it looks to be because the STOP_SENDING frame was never sent from the client, even though it called the API to abort the receive. This is the real bug that caused the test failures.
5. This event, ` [Microsoft-Quic][strm][0xFFFF9F021563E5E0] Reset Reliable ACKed in OnResetReliableAck. Send side. UnAckedOffset=7077, ReliableOffsetSend=5000`, isn't necessary because (1) you log the reliable offset when it's originally set,a nd (2) we already log the unacked offset: `[Microsoft-Quic][strm][0xFFFF9F021563E5E0] SF:0 FC:72613 QS:10000 MAX:9890 UNA:7077 NXT:9890 RECOV:0-0`. Perhaps we should just update that second log to include reliable offset (though I'm not sure).

### Affected OS

- [X] Windows
- [ ] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

Windows kernel mode

### MsQuic version

main

Run automation.

### Expected behavior

All tests pass

### Actual outcome

One test doesn't and we also see crashes.

### Additional details

_No response_

ProjectsByJackHe:
## Description

Fixes #3891, we can remove the piece of logging that gets triggered when a RESET RELIABLE frame gets ACK'd (to boost performance by having less logs). 

In addition, the reason we are sometimes not sending a STOP_SENDING frame from the client is because send-silent shutdown **clears ALL send flags**, but we still need to process RECV_ABORT flag first, which leads to a race condition.

The most likely reason this wasn't caught earlier was because the send Buffer size was too small in the tests, which meant all the data would get delivered in 1 RECEIVE_COMPLETE event, which perhaps indirectly resolved this race condition.

One interesting idea to explore in the future is what if we get rid of all the sender/receiver reliable abort logic, and simply enqueue a RESET frame once the sender has sent / ack'd enough data. That way, the sender doesn't have to wait until the reliable reset gets acknowledged, in fact, there is no need at all for the CLOSE_STREAM / RESET_RELIABLE frame, it would just be some internal logic that the sender keeps track of, and once enough data is sent / acked, then it will go down the normal abortive path. I started an issue on the QUIC reliable reset repo here: https://github.com/quicwg/reliable-stream-reset/issues/25 

## Testing

Modified the BUFFER_SIZE to be larger to trigger more code paths.

## Documentation

N/A


src/core/stream_send.c
```diff
@@ -1537,8 +1532,21 @@ QuicStreamOnAck(
...
+        //
+        Stream->Flags.LocalCloseAcked = TRUE;
+        QuicSendClearStreamSendFlag(
+            &Stream->Connection->Send,
+            Stream,
+            QUIC_STREAM_SEND_FLAG_DATA_BLOCKED |
+            QUIC_STREAM_SEND_FLAG_RELIABLE_ABORT |
+            QUIC_STREAM_SEND_FLAG_DATA |
+            QUIC_STREAM_SEND_FLAG_OPEN |
+            QUIC_STREAM_SEND_FLAG_FIN);
```
nibanks, line 1546: Do we already have a #define for these? If not, perhaps we should define something that represents this logic grouping.

src/test/lib/DataTest.cpp
```diff
@@ -3376,6 +3376,9 @@ QuicTestStreamReliableReset(
     TEST_TRUE(Listener.LastConnection->HandshakeComplete);
     CxPlatSleep(50); // Wait for things to idle out
 
+    //
+    // Note: we shouldn't be testing dropped packets during Release builds.
+    //
```
nibanks, line 3381: Let's go ahead and fix this part up. The easiest way is to just do the for-loop and `BitmapLossHelper` in `DEBUG` mode only. Otherwise, just do one loop without loss in Release mode.

src/test/lib/DataTest.cpp
```diff
@@ -3280,7 +3280,7 @@ QuicTestConnectAndIdleForDestCidChange(
     }
 }
 
-#define BUFFER_SIZE 10000
+#define BUFFER_SIZE 30000
```
nibanks, line 3283: This buffer needs to by dynamically allocated on the heap, not the stack because Windows kernel mode can bugcheck.

ProjectsByJackHe: Done! 91f4021db5f2d340a613e505675611cb83c29e18

src/test/lib/DataTest.cpp
ProjectsByJackHe: Done! 91f4021db5f2d340a613e505675611cb83c29e18

src/core/stream_send.c
ProjectsByJackHe: Done! 91f4021db5f2d340a613e505675611cb83c29e18

src/core/stream.h
```diff
@@ -932,11 +932,11 @@ QuicStreamOnResetReliableAck(
     );
 
 //
-// Cleanups up state once we finish processing a RELIABLE_RESET frame.
+// Cancels any queued send requests. Usually right before we abort.
 //
 _IRQL_requires_max_(PASSIVE_LEVEL)
 void
-QuicStreamCleanupReliableReset(
+QuicStreamCleanupQueuedRequests(
```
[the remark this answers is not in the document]
ProjectsByJackHe: Done! 91f4021db5f2d340a613e505675611cb83c29e18

src/test/lib/DataTest.cpp
```diff
@@ -3355,9 +3355,9 @@ QuicTestStreamReliableReset(
     TEST_QUIC_SUCCEEDED(ClientConfiguration.GetInitStatus());
 
     StreamReliableReset Context;
-    uint8_t SendDataBuffer[BUFFER_SIZE];
+    uint8_t *SendDataBuffer = new(std::nothrow) uint8_t[BUFFER_SIZE];
```
nibanks, line 3358: Need to make sure this gets freed at the end. I think we have a `UniquePtr` helper already.

nibanks: Need a `UniquePtr` here too (just like below).

src/test/lib/DataTest.cpp
```diff
@@ -3376,7 +3376,12 @@ QuicTestStreamReliableReset(
...
-    for (uint64_t Bitmap = 0; Bitmap < 8; ++Bitmap) { // Try dropping the first 3 packets
+    uint64_t Loss = 1;
+    #if DEBUG
+    Loss = 8;
+    #endif
+
+    for (uint64_t Bitmap = 0; Bitmap < Loss; ++Bitmap) {
         char Name[64]; sprintf_s(Name, sizeof(Name), "Try Reliably Shutting Down Stream %llu", (unsigned long long)Bitmap);
         TestScopeLogger logScope(Name);
         BitmapLossHelper LossHelper(Bitmap);
```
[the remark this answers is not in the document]
ProjectsByJackHe: Smart

src/test/lib/DataTest.cpp
ProjectsByJackHe: silly me

src/core/stream_send.c
```diff
@@ -249,6 +240,10 @@ QuicStreamSendShutdown(
         }
         Stream->Flags.LocalCloseResetReliable = TRUE;
         Stream->SendShutdownErrorCode = ErrorCode;
+
+        // Enqueue all the stuff in ApiSendRequests to SendRequests
+        // It is guaranteed enough sends were queued prior to calling Reliable Shutdown.
```
nibanks, line 245: This is still a "todo" right?

ProjectsByJackHe: Correct, there is a lot of logic in SendFlush that may or may not be relevant in this case. It may be worthwhile to leave that for a future PR, and get this PR merged first to address the CLEAR_ALL_SEND_FLAG and stack/heap issue.

ProjectsByJackHe: https://github.com/microsoft/msquic/pull/3900

PR #3900, opened 2023-10-06
Handle bad behavior by app in reliable shutdown path.

ProjectsByJackHe:
## Description

This PR proposes a solution to handle bad behavior by the app: the application exhibits undesirable behavior by attempting to send data after or during a send shutdown operation in the reliable reset path.

This PR also refactors the sendFlush method, and attempts to address the ETW trace misconfiguration in issue #3891 

## Testing

Added an extra send in Datateset after calling shutdown.

## Documentation

n/a

### DIFF
--- a/src/core/stream.h
+++ b/src/core/stream.h
@@ -861,16 +861,6 @@ QuicStreamIndicateSendShutdownComplete(
     );
 
 //
-// Enqueues a SendRequest from the temporary queue to the actual queue.
-//
-_IRQL_requires_max_(PASSIVE_LEVEL)
-void
-QuicStreamEnqueueSendRequest(
-    _In_ QUIC_STREAM* Stream,
-    _Inout_ QUIC_SEND_REQUEST* SendRequest
-);
-
-//
 // Indicates data has been queued up to be sent out on the stream.
 //
 _IRQL_requires_max_(PASSIVE_LEVEL)
--- a/src/core/stream.h
+++ b/src/core/stream.h
@@ -861,6 +861,16 @@ QuicStreamIndicateSendShutdownComplete(
     );
 
 //
+// Enqueues a SendRequest from the temporary queue to the actual queue.
+//
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+QuicStreamEnqueueSendRequest(
+    _In_ QUIC_STREAM* Stream,
+    _Inout_ QUIC_SEND_REQUEST* SendRequest
+);
+
+//
 // Indicates data has been queued up to be sent out on the stream.
 //
 _IRQL_requires_max_(PASSIVE_LEVEL)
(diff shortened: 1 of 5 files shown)


src/core/stream.h
```diff
@@ -860,6 +860,16 @@ QuicStreamIndicateSendShutdownComplete(
...
 
+//
+// Enqueues a SendRequest from the temporary queue to the actual queue.
+//
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+QuicStreamEnqueueSendRequest(
+    _In_ QUIC_STREAM* Stream,
+    _Inout_ QUIC_SEND_REQUEST* SendRequest
+);
```
nibanks, line 871:
```suggestion
_IRQL_requires_max_(PASSIVE_LEVEL)
void
QuicStreamEnqueueSendRequest(
    _In_ QUIC_STREAM* Stream,
    _Inout_ QUIC_SEND_REQUEST* SendRequest
    );
```

nibanks: For functions only called in one file, the declaration usually is just put at the top of that file, instead of the header. The header is so other files can call it.

src/core/stream_send.c
```diff
@@ -249,6 +249,34 @@ QuicStreamSendShutdown(
...
+            QuicStreamEnqueueSendRequest(Stream, SendRequest);
+
+            //
+            // Update flags.
+            //
+            QuicSendSetStreamSendFlag(
+                &Stream->Connection->Send,
+                Stream,
+                QUIC_STREAM_SEND_FLAG_DATA,
+                !!(SendRequest->Flags & QUIC_SEND_FLAG_DELAY_SEND));
```
nibanks, line 270: I don't think we should need to set the flags in this case, because we know we don't need to send it (we're aborting).

src/core/stream_send.c
```diff
@@ -514,6 +542,62 @@ QuicStreamSendBufferRequest(
     return QUIC_STATUS_SUCCESS;
 }
 
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+QuicStreamEnqueueSendRequest(
+    _In_ QUIC_STREAM* Stream,
+    _Inout_ QUIC_SEND_REQUEST* SendRequest
+) {
```
nibanks, line 550:
```suggestion
    )
{
```

src/core/stream_send.c
ProjectsByJackHe: Removed! 2d8415f6bf02af962e461fc41b8ef9b23b9ab669

src/core/stream_send.c
ProjectsByJackHe: Commited! 2d8415f6bf02af962e461fc41b8ef9b23b9ab669

src/core/stream.h
ProjectsByJackHe: Moved the declaration! 2d8415f6bf02af962e461fc41b8ef9b23b9ab669

nibanks (approved):
LGTM. If all the tests pass, merge it!

nibanks (approved):

PR #3909, opened 2023-10-10
Add JIT compilation to replace relative path ../src with full URL to prepare for docfx

ProjectsByJackHe:
## Description

In original docs, links were written with relative paths under the assumption that the user was browsing on github. Now that we have DocFx for our site, we need the absolute paths. This PR implements a JIT script to replace all the relative paths with their absolute paths.

## Testing

Using CI.

## Documentation

N/A

### DIFF
--- a/.github/workflows/publish-docfx.yml
+++ b/.github/workflows/publish-docfx.yml
@@ -44,6 +44,10 @@ jobs:
     - run: chmod +x ./scripts/generate-docfx-yml.ps1
     - run: ./scripts/generate-docfx-yml.ps1 ./docs
       shell: pwsh
+    - run: chmod +x ./scripts/docfx-replace-url.ps1
+      shell: pwsh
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-docfx.yml
```diff
@@ -30,6 +30,8 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608
+      with:
+        ref: main
```
nibanks, line 34: Why is this necessary? You already state this only runs on main anyways.

ProjectsByJackHe: I did this mainly so that the "Edit this page" button for docfx would point to the main branch, instead of the temporary PR branch blob that Actions/Checkout creates. This isn't a great solution, though, so I instead set an ENV variable here 5ac064479483e4279994905b7b4f436547799275

PR #3931, opened 2023-10-19
Streams Hold References on Connections

nibanks:
## Description

Make sure to keep the connection alive as long as the stream.

## Testing

CI/CD

## Documentation

N/A

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -503,6 +503,8 @@ QuicConnUninitialize(
     QuicCryptoUninitialize(&Connection->Crypto);
     QuicTimerWheelRemoveConnection(&Connection->Worker->TimerWheel, Connection);
     QuicOperationQueueClear(Connection->Worker, &Connection->OperQ);
+    QuicLossDetectionUninitialize(&Connection->LossDetection);
+    QuicSendUninitialize(&Connection->Send);
 
     if (Connection->CloseReasonPhrase != NULL) {
         CXPLAT_FREE(Connection->CloseReasonPhrase, QUIC_POOL_CLOSE_REASON);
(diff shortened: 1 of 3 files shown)


ManickaP:
Does this mean that we don't have to keep connection alive anymore until all streams are closed?

Because we have stream ref counting which holds back `ConnectionClose` call until the count is 0.

nibanks:
> Does this mean that we don't have to keep connection alive anymore until all streams are closed?
> 
> Because we have stream ref counting which holds back `ConnectionClose` call until the count is 0.

Not necessarily. At least not with just this PR. We _might_ end up going that direction (especially if you really want/need it), but right now, we're working on solving a very small race edge case where even if you close a stream and then the connection (all on the QUIC callback) while you were calling send on your thread, it's possible for MsQuic to use-after-free. We've only seen this with SMB in extremely rare cases; and we're trying to fix it.

FYI, this PR as it stands right now though, is causing deadlocks. 😦 We need to fix this first, but I _think_ once it is fixed, it'd be easy enough to support closing in different orders.

ManickaP:
Oh, I see, thanks for the explanation. We don't need it as we do have a working solution. I just got excited, because it would simplified our code a little.

src/core/connection.c
```diff
@@ -503,6 +503,8 @@ QuicConnUninitialize(
     QuicCryptoUninitialize(&Connection->Crypto);
     QuicTimerWheelRemoveConnection(&Connection->Worker->TimerWheel, Connection);
     QuicOperationQueueClear(Connection->Worker, &Connection->OperQ);
+    QuicLossDetectionUninitialize(&Connection->LossDetection);
+    QuicSendUninitialize(&Connection->Send);
```
ProjectsByJackHe, line 507: Interesting that the code was working perfectly fine before without un-initializing loss detection and send, and its only when we added a Ref from Stream ---> Connection that deadlocks happen

ProjectsByJackHe (approved):
All tests pass except for a handshake test, but its testing using some randomness and this PR doesn't directly impact handshakes, so LGTM!

src/core/connection.c
csujedihy: Yeah, was it a bug that got discovered from this PR?

nibanks: Not a bug previously. This already happens on free, but because it holds onto streams, we need to clean up here.

csujedihy (approved):

PR #3938, opened 2023-10-25
Support Using Streams after Connection Closure

nibanks:
## Description

Allows apps to not worry about close order between streams and connections.

## Testing

New test added. Also leverages existing tests (which found some other issues).

## Documentation

TODO?

### DIFF
--- a/src/test/lib/OwnershipTest.cpp
+++ b/src/test/lib/OwnershipTest.cpp
@@ -230,3 +230,44 @@ void QuicTestRegistrationShutdownAfterConnOpenAndStart()
     TEST_TRUE(State.StateEvent.WaitTimeout(2000));
     TEST_EQUAL(1, State.ShutdownCount);
 }
+
+void QuicTestConnectionCloseBeforeStreamClose()
+{
+    MsQuicRegistration Registration;
+    TEST_TRUE(Registration.IsValid());
+
+    MsQuicRegistration ServerRegistration{true};
+    TEST_TRUE(ServerRegistration.IsValid());
+
+    MsQuicAlpn Alpn("MsQuicTest");
+    MsQuicCredentialConfig ClientCredConfig;
+    MsQuicConfiguration ClientConfiguration(Registration, Alpn, ClientCredConfig);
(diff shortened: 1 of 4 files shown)


nibanks:
> So currently I can't use stream after its connection closure? I thought the api would reliably return some kind of error code.

Well, once I finish this PR (ETA: this week) then it won't matter, but the problem before was that the connection would be deleted when you close the stream, so any access would be a use-after-free, without adding more state to the stream itself.

src/test/lib/OwnershipTest.cpp
```diff
@@ -230,3 +230,44 @@ void QuicTestRegistrationShutdownAfterConnOpenAndStart()
...
+    TEST_QUIC_SUCCEEDED(Listener.Start(Alpn, &ServerLocalAddr.SockAddr));
+    TEST_QUIC_SUCCEEDED(Listener.GetLocalAddr(ServerLocalAddr));
+
+    {
+        MsQuicConnection Connection(Registration);
+        TEST_QUIC_SUCCEEDED(Connection.GetInitStatus());
+        MsQuicStream Stream(Connection, QUIC_STREAM_OPEN_FLAG_NONE);
+        TEST_QUIC_SUCCEEDED(Stream.GetInitStatus());
+        Connection.Close();
+    }
```
ProjectsByJackHe, line 263: Just a quick question regarding the testing methodology. The way I understand this change is, we want apps to still be able to make API calls to Streams after connection closure without getting a seg-fault. Shouldn't `Connection.close()` here be before we call `Stream.GetInitStatus()`?

nibanks: This makes it so the stream is closed after the connection. We could also add additional calls on the stream after this connection close call, for instance to query the stream ID. I can add that for extra coverage.

ProjectsByJackHe: Ahh, Streams implicitly auto-close. Before, apps had to explicitly close all streams before connection closure. Yeah sounds good, adding additional queries to Streams after connection closure ensures no seg-faults occur.

ProjectsByJackHe (approved):
LGTM!

PR #3940, opened 2023-10-25
Fix epoll TCP implementation

csujedihy:
## Description

There are a few bugs in the epoll TCP implementation.
1. TX: busy spinning on non-blocking socket regardless of whether eagain/ewouldblock is encountered.
2. RX: not draining all of the ready data upon EPOLLIN signal.
3. Wrong buffer length passed to read, which causes debug build to crash consistently.
4. Unnecessary RSS config for TCP server sockets.

Local perf test in WSL showed 2x perf improvement with loopback but the numbers from the perf machines didn't seem to move at all.

## Testing
CI

### DIFF
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -2487,14 +2487,20 @@ CxPlatSendDataSend(
 
     QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
     CXPLAT_SOCKET_CONTEXT* SocketContext = SendData->SocketContext;
-    BOOLEAN Success =
-        SocketType == CXPLAT_SOCKET_UDP ?
+    BOOLEAN Success;
+
+    if (SocketType == CXPLAT_SOCKET_UDP) {
+        Success =
 #ifdef UDP_SEGMENT
             SendData->SegmentationSupported ?
-                CxPlatSendDataSendSegmented(SendData) :
+                CxPlatSendDataSendSegmented(SendData) : CxPlatSendDataSendMessages(SendData);
+#else
+            CxPlatSendDataSendMessages(SendData);
 #endif
-                CxPlatSendDataSendMessages(SendData) :
-            CxPlatSendDataSendTcp(SendData);
+    } else {
+        Success = CxPlatSendDataSendTcp(SendData);
+    }
+
     if (!Success) {
         if (errno == EAGAIN || errno == EWOULDBLOCK) {
             Status = QUIC_STATUS_PENDING;
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -1367,16 +1367,6 @@ CxPlatSocketCreateTcpInternal(
         goto Exit;
     }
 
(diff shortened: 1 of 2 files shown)


nibanks (approved):

nibanks:
> Local perf test in WSL showed 2x perf improvement with loopback but the numbers from perf machine didn't seem to move at all.

I assume this means our baremetal tests weren't ever hitting this partial sends/recvs (by chance)?

csujedihy:
> > Local perf test in WSL showed 2x perf improvement with loopback but the numbers from perf machine didn't seem to move at all.
> 
> I assume this means our baremetal tests weren't ever hitting this partial sends/recvs (by chance)?

I believe they do hit the partial sends/recvs most of the time. If not, it means we are not posting data faster than the socket buffers get drained by the kernel.

The baremetal loopback tests without **encryption** also show very slow numbers. Much slower than what I got on my WSL VM actually. I remember I was able to hit 90Gbps with netput on my Z240 WSL VM with 1 connection while on the host I was able to get only 30-50Gbps inconsistently. There's definitely something not quite right.

nibanks:
> > > Local perf test in WSL showed 2x perf improvement with loopback but the numbers from perf machine didn't seem to move at all.
> > 
> > 
> > I assume this means our baremetal tests weren't ever hitting this partial sends/recvs (by chance)?
> 
> I believe they do hit the partial sends/recvs most of the time. If not, it means we are not posting data faster than the socket buffers get drained by the kernel. It's also why the continue was added in the first place in the original PR because otherwise, the sends won't fully complete.
> 
> The baremetal loopback tests without **encryption** also show very slow numbers. Much slower than what I got on my WSL VM actually. I remember I was able to hit 90Gbps with netput on my Z240 WSL VM with 1 connection while on the host I was able to get only 30-50Gbps inconsistently. There's definitely something not quite right.

The difference here is that we're doing encrypted TCP. This is much different than simple netput. Crypto is a huge bottleneck.

csujedihy:
> > > > Local perf test in WSL showed 2x perf improvement with loopback but the numbers from perf machine didn't seem to move at all.
> > > 
> > > 
> > > I assume this means our baremetal tests weren't ever hitting this partial sends/recvs (by chance)?
> > 
> > 
> > I believe they do hit the partial sends/recvs most of the time. If not, it means we are not posting data faster than the socket buffers get drained by the kernel. It's also why the continue was added in the first place in the original PR because otherwise, the sends won't fully complete.
> > The baremetal loopback tests without **encryption** also show very slow numbers. Much slower than what I got on my WSL VM actually. I remember I was able to hit 90Gbps with netput on my Z240 WSL VM with 1 connection while on the host I was able to get only 30-50Gbps inconsistently. There's definitely something not quite right.
> 
> The difference here is that we're doing encrypted TCP. This is much different than simple netput. Crypto is a huge bottleneck.

wait, "encrypt:0" not honored for TCP?

nibanks:
> wait, "encrypt:0" not honored for TCP?

No, that feature isn't supported for TCP.

PR #3941, opened 2023-10-26
Don't Affinitize Platform Worker

nibanks:
## Description

We've found concrete examples where CPU affinitization hurts latency significantly, and initial tests seem to indicate minimal negative impact of just setting ideal processor.

## Testing

CI/CD

## Documentation

N/A

### DIFF
--- a/src/platform/platform_worker.c
+++ b/src/platform/platform_worker.c
@@ -150,7 +150,7 @@ CxPlatWorkersLazyStart(
     }
 
     CXPLAT_THREAD_CONFIG ThreadConfig = {
-        CXPLAT_THREAD_FLAG_SET_AFFINITIZE,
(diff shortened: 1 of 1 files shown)


mtfriesen (approved):

mtfriesen:
Bonus points: have you tried CPU sets or setting a broader affinity mask along with ideal processor? For example, set the hard affinity to the NUMA node or even the physical processor, and the ideal processor to the logical RSS processor?

nibanks:
> Bonus points: have you tried CPU sets or setting a broader affinity mask along with ideal processor? For example, set the hard affinity to the NUMA node or even the physical processor, and the ideal processor to the logical RSS processor?

Yes, this flag sets ideal processor to (hopefully) the RSS processor, and it sets the hard affinity to matching NUMA node.

PR #3948, opened 2023-10-30
Support setting flow control limits for individual stream types

### LINKED ISSUE #3897: Setting initial flow-control limits for individual stream types

We are considering exposing more QUIC configuration options from .NET side, and I notice that MsQuic exposes only one option for setting initial stream max_data (which applies to all 3 types). Was there a specific reason behind this choice?

We would like to expose limits for all three types (unidirectional, remotely/locally initiated bidirectional) limits to be in line with other QUIC implementations.

### Proposed solution

Since there already is `QUIC_SETTINGS.StreamRecvWindowDefault` which covers all three stream types, the only proposal I came up with is adding the new values as separate options and have a precedence-fallback relationship between them. E.g. if `QUIC_SETTINGS.UnidiStreamRecvWindowDefault` is set, than it takes precedence over the potential value set in `QUIC_SETTINGS.StreamRecvWindowDefault`.

### Additional context

If approved, I am willing to contribute the change. Just tell me upfront if you have preferences about option naming.

rzikm:
Closes #3897.

## Description

This PR adds 3 new QUIC_SETTINGS members:

- StreamRecvWindowBidiLocalDefault
- StreamRecvWindowBidiRemoteDefault
- StreamRecvWindowUnidiDefault

With the semantic that if those are set, then they overwrite the StreamRecvWindowDefault setting.

This should preserve backwards compatibility since StreamRecvWindowDefault setting still has effect, but new apps can use the new settings to specify more granular limits.

## Testing

Tests in SettingsTests.cpp were updated. I will also verify the change once I managed to write a test app against it.

## Documentation

Documentation files were updated.


src/core/settings.c
```diff
@@ -496,6 +514,32 @@ QuicSettingApply(
         }
         Destination->StreamRecvWindowDefault = Source->StreamRecvWindowDefault;
         Destination->IsSet.StreamRecvWindowDefault = TRUE;
+
+        // set also window size for individual stream types, they will be overwritten by a more specific settings if set
```
nibanks, line 518:
```suggestion
        //
        // Also set the window size for individual stream types, they will be overwritten by a more specific settings if set.
        ///
```

src/core/stream.c
```diff
@@ -114,11 +114,17 @@ QuicStreamInitialize(
         }
     }
 
+    int32_t FlowControlWindowSize = Stream->Flags.Unidirectional
```
nibanks, line 117:
```suggestion
    const uint32_t FlowControlWindowSize = Stream->Flags.Unidirectional
```

nibanks (commented):
We need to add a new test case on top of MsQuic API to make sure things work as expected.

nibanks (approved):

PR #3957, opened 2023-11-13
Improve Documentation on Connection Upcalls

nibanks:
## Description

Improves the documentation around execution model for upcalls for a connection and its streams. This is a common question and should be well documented.

## Testing

N/A

## Documentation

Yes.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -69,7 +69,7 @@ The **PATCH** version **only changes** when a servicing fix is made to an existi
 
 In general, MsQuic uses a callback model for all asynchronous events up to the app. This includes things like connection state changes, new streams being created, stream data being received, and stream sends completing. All these events are indicated to the app via the callback handler, on a thread owned by MsQuic.
 
-Generally, MsQuic creates multiple threads to parallelize work, and therefore will make parallel/overlapping upcalls to the application, but not for the same connection. All upcalls to the app for a single connection and all child streams area always delivered serially. This is not to say, though, it will always be on the same thread. MsQuic does support the ability to shuffle connections around to better balance the load.
+Generally, MsQuic creates multiple threads to parallelize work, and therefore will make parallel/overlapping upcalls to the application, but not for the same connection. All upcalls to the app for a single connection and all child streams are always delivered serially. This is not to say, though, it will always be on the same thread. MsQuic does support the ability to shuffle connections around to better balance the load.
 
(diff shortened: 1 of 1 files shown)


ProjectsByJackHe (approved):
LGTM

docs/API.md
```diff
@@ -69,6 +69,8 @@ The **PATCH** version **only changes** when a servicing fix is made to an existi
 
 In general, MsQuic uses a callback model for all asynchronous events up to the app. This includes things like connection state changes, new streams being created, stream data being received, and stream sends completing. All these events are indicated to the app via the callback handler, on a thread owned by MsQuic.
 
+Generally, MsQuic creates multiple threads to parallelize work, and therefore will make overlapping upcalls to the application, but not for the same connection. All upcalls to the app for a single connection and all child streams area always delivered synchronously. This is not to say, though, it will always be on the same thread. MsQuic does support the ability to shuffle connections around to better balance the load.
```
anrossi, line 72: s/overlapping/parallel/

csujedihy (dismissed):

docs/API.md
nibanks: I used overlapping because I just said "parallelize" earlier in the sentence and I felt like it was too much "parallel" to say it again 😄

anrossi: Hmm, I think that could be re-worded to clearly say parallel upcalls are always for different connections, and for a given connection, all upcalls are sequential/single-threaded.

nibanks: Made a slight change. How's it now?

docs/API.md
```diff
@@ -69,6 +69,8 @@ The **PATCH** version **only changes** when a servicing fix is made to an existi
 
 In general, MsQuic uses a callback model for all asynchronous events up to the app. This includes things like connection state changes, new streams being created, stream data being received, and stream sends completing. All these events are indicated to the app via the callback handler, on a thread owned by MsQuic.
 
+Generally, MsQuic creates multiple threads to parallelize work, and therefore will make parallel/overlapping upcalls to the application, but not for the same connection. All upcalls to the app for a single connection and all child streams area always delivered serially. This is not to say, though, it will always be on the same thread. MsQuic does support the ability to shuffle connections around to better balance the load.
```
ProjectsByJackHe, line 72: "area" --> "are"

ProjectsByJackHe, line 72:
```suggestion
Generally, MsQuic creates multiple threads to parallelize work, and therefore will make parallel/overlapping upcalls to the application, but not for the same connection. All upcalls to the app for a single connection and all child streams are always delivered serially. This is not to say, though, it will always be on the same thread. MsQuic does support the ability to shuffle connections around to better balance the load.
```

PR #3961, opened 2023-11-17
Basic NMR integration for MsQuic

csujedihy:
## Description

The PR adds an NMR interface (NMR provider) for kernel mode msquic. The MsQuic provide only exposes 2 interfaces, MsQuicOpenVersion and MsQuicClose, which are the same as what are exposed in msquic.sys. This PR does not remove the exports.

Currently, only QUIC tests are hooked up with NMR.

## Testing

CI


src/inc/msquic.h
```diff
@@ -1646,6 +1646,41 @@ MsQuicClose(
...
+    _In_ _Pre_defensive_ const void* QuicApi
+    );
+
+#ifdef _KERNEL_MODE
+
+static const GUID MSQUIC_NPI_ID = {
+    0xC43138E3, 0xCD13, 0x4CB1, { 0x9C, 0xAE, 0xE0, 0x05, 0xC8, 0x55, 0x7A, 0xBA }
+}; // C43138E3-CD13-4CB1-9CAE-E005C8557ABA
+
+static const GUID MSQUIC_MODULE_ID = {
```
csujedihy, line 1671: Seems like `static` creates a copy of this struct in all obj files, which isn't ideal. How to properly share a struct in a header file?

src/test/bin/winkernel/control.cpp
```diff
@@ -81,7 +170,20 @@ QuicTestCtlInitialize(
...
+    if (!NT_SUCCESS(Status)) {
+        QuicTraceEvent(
+            LibraryErrorStatus,
+            "[ lib] ERROR, %u, %s.",
+            Status,
+            "QuicTestRegisterNmrClient failed");
+        goto Error;
+    }
+
+    MsQuic =
```
mtfriesen, line 183: Verify the ProviderDispatch version first? Though I suppose since its version is currently set to zero, I think your check would be impossible to fail 😄

src/inc/msquic.h
mtfriesen: DECLSPEC_SELECTANY is the right macro on Windows

src/test/bin/winkernel/control.cpp
csujedihy: I am not exactly sure how to use all the version fields NMR. There's also a version field in the NPI_REGISTRATION_INSTANCE passed in the attach callback. MsQuic's entry function also has an argument for version.

Would it make sense to just defer the version related stuff to MsQuicOpenVersion and assume there are always and only MsQuicOpenVersion and MsQuicClose in the provider dispatch?

src/bin/winkernel/nmrprovider.c
```diff
@@ -0,0 +1,109 @@
+#include "quic_platform.h"
```
nibanks, line 1: Need copyright at top of the file.

src/test/bin/winkernel/control.cpp
csujedihy: (the NMR doc just says for network modules set version to 0??)

src/bin/winkernel/nmrprovider.c
```diff
@@ -0,0 +1,109 @@
...
+#include <wdm.h>
+#include "msquic.h"
+#include "quic_trace.h"
```
nibanks, line 4: Should we add some logs to this file to understand when things connect/disconnect?

src/inc/msquic_winkernel.h
```diff
@@ -24,6 +24,7 @@
 #include <minwindef.h>
 #include <ntstatus.h>
 #include <basetsd.h>
+#include <netioddk.h>
```
nibanks, line 27: This is the "public" header that external clients consume. Is there any risk of this breaking their builds?

src/inc/msquic.h
```diff
@@ -1646,6 +1646,41 @@ MsQuicClose(
...
+    0x698F7C72, 0xC2E6, 0x49CD, { 0x8C, 0x39, 0x98, 0x85, 0x1D, 0x50, 0x19, 0x01 }
+}; // 698F7C72-C2E6-49CD-8C39-98851D501901
+
+typedef struct MSQUIC_NMR_DISPATCH {
+    USHORT Version;
+    USHORT Reserved;
```
nibanks, line 1677:
```suggestion
    uint16_t Version;
    uint16_t Reserved;
```

src/test/bin/winkernel/control.cpp
```diff
@@ -65,6 +65,95 @@ WDFDEVICE QuicTestCtlDevice = nullptr;
...
+    _In_ CONST NPI_REGISTRATION_INSTANCE *ProviderRegistrationInstance
+    )
+{
+    UNREFERENCED_PARAMETER(ProviderRegistrationInstance);
+
+    NTSTATUS Status;
+    QuicTestNmrClient* Client = (QuicTestNmrClient*)ClientContext;
+    PVOID ProviderContext;
+
+    #pragma warning(suppress:6387) // _Param_(2) could be '0' - by design.
```
nibanks, line 93: Does the NMR definition need to be fixed?

src/test/bin/winkernel/control.cpp
```diff
@@ -65,6 +65,95 @@ WDFDEVICE QuicTestCtlDevice = nullptr;
...
+    BOOLEAN Deleting;
+} QuicTestNmrClient;
+
+static QuicTestNmrClient NmrClient;
+
+static
+NTSTATUS
+QuicTestClientAttachProvider(
+    _In_ HANDLE NmrBindingHandle,
+    _In_ VOID *ClientContext,
```
nibanks, line 83: Let's try to use basic C types (`void`, `const` `void*`) here.

src/test/bin/winkernel/control.cpp
```diff
@@ -65,6 +65,95 @@ WDFDEVICE QuicTestCtlDevice = nullptr;
...
+
+    #pragma warning(suppress:6387) // _Param_(2) could be '0' - by design.
+    Status =
+        NmrClientAttachProvider(
+            NmrBindingHandle,
+            Client,
+            NULL,
+            &ProviderContext,
+            (const void**)&Client->ProviderDispatch);
+    return Status;
```
nibanks, line 101:
```suggestion
    return
        NmrClientAttachProvider(
            NmrBindingHandle,
            Client,
            NULL,
            &ProviderContext,
            (const void**)&Client->ProviderDispatch);
```

src/inc/msquic_winkernel.h
csujedihy: It might if their code conflicts with the definitions or something in the header but netioddk.h is a public header file published in DDK. I don't know what we can do to reduce the risk. Are there really external KM consumers?

src/test/bin/winkernel/control.cpp
```diff
@@ -65,6 +65,95 @@ WDFDEVICE QuicTestCtlDevice = nullptr;
...
+
+NTSTATUS
+QuicTestRegisterNmrClient(
+    void
+    )
+{
+    NPI_REGISTRATION_INSTANCE *ClientRegistrationInstance;
+    NTSTATUS Status = STATUS_SUCCESS;
+
+    KeInitializeEvent(&NmrClient.RegistrationCompleteEvent, NotificationEvent, FALSE);
```
nibanks, line 126: Perhaps we can use the CxPlat abstraction layer for events and waiting on them instead?

src/test/bin/winkernel/control.cpp
```diff
@@ -65,6 +65,95 @@ WDFDEVICE QuicTestCtlDevice = nullptr;
...
+    ClientRegistrationInstance = &NmrClient.NpiClientCharacteristics.ClientRegistrationInstance;
+    ClientRegistrationInstance->Size = sizeof(*ClientRegistrationInstance);
+    ClientRegistrationInstance->Version = 0;
+    ClientRegistrationInstance->NpiId = &MSQUIC_NPI_ID;
+    ClientRegistrationInstance->ModuleId = &NmrClient.ModuleId;
+
+    Status =
+        NmrRegisterClient(
+            &NmrClient.NpiClientCharacteristics, &NmrClient, &NmrClient.NmrClientHandle);
+    if (!NT_SUCCESS(Status)) {
```
nibanks, line 144: I think we should be logging on all failures in the new code here.

src/bin/winkernel/nmrprovider.c
csujedihy: Added

src/test/bin/winkernel/control.cpp
csujedihy: IMO, they should be.

src/bin/winkernel/nmrprovider.c
```diff
@@ -0,0 +1,133 @@
...
+
+void
+MsQuicDeregisterNmrProvider(
+    void
+    )
+{
+    if (NmrProvider.NmrProviderHandle != NULL) {
+        NTSTATUS Status = NmrDeregisterProvider(NmrProvider.NmrProviderHandle);
+        CXPLAT_FRE_ASSERTMSG(Status == STATUS_PENDING, "deregistration failed");
+        NmrWaitForProviderDeregisterComplete(NmrProvider.NmrProviderHandle);
```
nibanks, line 130: I assume one of these functions is a blocking call, waiting on clients to detach completely too? Should we explicitly try to kill any active connections if we want to detach (i.e. service shutdown) in the middle of things?

csujedihy: `NmrWaitForProviderDeregisterComplete` is blocking and returns when the client's detach callback completes inline or asynchronously.

When the provider initiates the deregistration, the client gets a detach callback. The client should kill active connections and any other clean-up work if needed and wait for in-progress operations to complete. The provider should not need to do anything except signaling the client, which is done by NMR firing the client detach callback.

src/bin/winkernel/nmrprovider.c
```diff
@@ -0,0 +1,133 @@
...
+NTSTATUS
+MsQuicNmrProviderAttachClient(
+    _In_ HANDLE NmrBindingHandle,
+    _In_ void *ProviderContext,
+    _In_ const NPI_REGISTRATION_INSTANCE *ClientRegistrationInstance,
+    _In_ void *ClientBindingContext,
+    _In_ const void *ClientDispatch,
+    _Out_ void **ProviderBindingContext,
+    _Out_ const void **ProviderDispatch
+)
```
nibanks, line 44:
```suggestion
    )
```

src/bin/winkernel/nmrprovider.c
nibanks: Ok. Sounds reasonable.

src/inc/msquic.h
```diff
@@ -1646,6 +1646,41 @@ MsQuicClose(
...
+    0xC43138E3, 0xCD13, 0x4CB1, { 0x9C, 0xAE, 0xE0, 0x05, 0xC8, 0x55, 0x7A, 0xBA }
+}; // C43138E3-CD13-4CB1-9CAE-E005C8557ABA
+
+DECLSPEC_SELECTANY GUID MSQUIC_MODULE_ID = {
+    0x698F7C72, 0xC2E6, 0x49CD, { 0x8C, 0x39, 0x98, 0x85, 0x1D, 0x50, 0x19, 0x01 }
+}; // 698F7C72-C2E6-49CD-8C39-98851D501901
+
+typedef struct MSQUIC_NMR_DISPATCH {
+    uint16_t  Version;
+    uint16_t  Reserved;
```
nibanks, line 1677: Why the `Reserved` field? And for `Version` should it not match the same format as the binary version?

src/inc/msquic.hpp
```diff
@@ -296,18 +296,24 @@ class UniquePtrArray {
...
 public:
-    MsQuicApi() noexcept {
-        if (QUIC_SUCCEEDED(InitStatus = MsQuicOpen2(&ApiTable))) {
+    MsQuicApi(
+        MsQuicOpenVersionFn _OpenFn = MsQuicOpenVersion,
+        MsQuicCloseFn _CloseFn = MsQuicClose) noexcept {
+
+        CloseFn = _CloseFn;
+
+        if (QUIC_SUCCEEDED(InitStatus = _OpenFn(QUIC_API_VERSION_2, &ApiTable))) {
```
nibanks, line 309:
```suggestion
    MsQuicApi(
        MsQuicOpenVersionFn _OpenFn = MsQuicOpenVersion,
        MsQuicCloseFn _CloseFn = MsQuicClose) noexcept : CloseFn(_CloseFn ) {
        if (QUIC_SUCCEEDED(InitStatus = _OpenFn(QUIC_API_VERSION_2, &ApiTable))) {
```

src/inc/msquic.hpp
```diff
@@ -296,18 +296,24 @@ class UniquePtrArray {
 };
 
 class MsQuicApi : public QUIC_API_TABLE {
-    const QUIC_API_TABLE* ApiTable {nullptr};
-    QUIC_STATUS InitStatus;
+    const void* ApiTable {nullptr};
+    QUIC_STATUS InitStatus {QUIC_STATUS_INVALID_STATE};
+    MsQuicCloseFn CloseFn {nullptr};
```
nibanks, line 301:
```suggestion
    const MsQuicCloseFn CloseFn {nullptr};
```

src/test/bin/winkernel/control.cpp
```diff
@@ -65,6 +65,113 @@ WDFDEVICE QuicTestCtlDevice = nullptr;
...
+        QuicTraceEvent(
+            LibraryErrorStatus,
+            "[ lib] ERROR, %u, %s.",
+            Status,
+            "NmrRegisterClient failed");
+        goto Exit;
+    }
+
+    if (!CxPlatEventWaitWithTimeout(NmrClient.RegistrationCompleteEvent, 1000)) {
```
nibanks, line 160: It's unclear to me where/who sets this.

src/test/bin/winkernel/control.cpp
```diff
@@ -222,6 +342,26 @@ QuicTestCtlUninitialize(
...
+        //
+        // We are already in the middle of detaching the client.
+        // Complete it now.
+        //
+        NmrClientDetachProviderComplete(NmrClient.NmrClientHandle);
+    }
+
+    if (NmrClient.NmrClientHandle) {
+        NTSTATUS Status = NmrDeregisterClient(NmrClient.NmrClientHandle);
+        CXPLAT_FRE_ASSERTMSG(Status == STATUS_PENDING, "client deregistration failed");
```
nibanks, line 355: Should we really bugcheck if this fails? Can OOM trigger this failure? What then?

src/inc/msquic_winkernel.h
nibanks: We don't have external KM customers. I'm just worried about an ingest PR possibly breaking Windows build.

src/inc/msquic.h
csujedihy: The first 2 fields are required by NMR.

src/test/bin/winkernel/control.cpp
csujedihy: It's supposed to be set in at the end of clientAttachProvider callback. I am surprised the code still works.

nibanks (approved):
LGTM. How can/should we test this more. For instance, any negative test cases?

src/test/bin/winkernel/control.cpp
csujedihy: NmrDeregisterClient strictly only returns STATUS_PENDING. This is sort of a dummy check.

csujedihy:
> LGTM. How can/should we test this more. For instance, any negative test cases?

Test coverage for NMR itself should be added in os repo and is being added (nmrtest). The more interesting test coverage will be added when we add msquic specific loading logic to NMR. Right now, msquic as an NMR provider is too simple to have more test cases. It literally just returns SUCCESS and sets the dispatch table for the client.

nibanks:
> > LGTM. How can/should we test this more. For instance, any negative test cases?
> 
> Test coverage for NMR itself should be added in os repo and is being added (nmrtest). The more interesting test coverage will be added when we add msquic specific loading logic to NMR. Right now, msquic as an NMR provider is too simple to have more test cases. It literally just returns SUCCESS and sets the dispatch table for the client. 
> 

Sounds good. Ship it.

PR #3994, opened 2023-12-08
Simple QUIC Forwarding Proxy

nibanks:
## Description

Supports forwarding QUIC connections and stream (no datagram support).

## Testing

Tested using secnetperf over loopback.

## Documentation

A bit in the abstract.

### DIFF
--- a/src/tools/forwarder/forwarder.cpp
+++ b/src/tools/forwarder/forwarder.cpp
@@ -1,0 +1,209 @@
+/*++
+
+    Copyright (c) Microsoft Corporation.
+    Licensed under the MIT License.
+
+Abstract:
+
+    This tool creates a terminating QUIC proxy to forward all incoming traffic
+    to a specified target.
+
+--*/
+
+#include "msquichelper.h"
+#include "msquic.hpp"
+
+const char* Alpn;
+uint16_t FrontEndPort;
+const char* BackEndTarget;
+uint16_t BackEndPort;
+QUIC_CERTIFICATE_HASH Cert;
+
+const MsQuicApi* MsQuic;
+MsQuicRegistration* Registration;
+MsQuicConfiguration* FrontEndConfiguration;
+MsQuicConfiguration* BackEndConfiguration;
+
+bool ParseArgs(char **argv) {
+    Alpn = argv[1];
+    FrontEndPort = (uint16_t)atoi(argv[2]);
+    BackEndTarget = argv[3];
+    char* port = strchr(argv[3], ':');
+    if (!port) {
+        printf("Invalid target specified (no port).\n");
+        return false;
+    }
+    *port = '\0'; ++port;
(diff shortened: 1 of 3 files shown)


ProjectsByJackHe (commented):
Looks like a lot of Windows builds are failing due to a warning treated as an error.

```
Building Custom Rule D:/a/msquic/msquic/src/tools/forwarder/CMakeLists.txt
  forwarder.cpp
D:\a\msquic\msquic\src\tools\forwarder\forwarder.cpp(60,77): error C2220: the following warning is treated as an error [D:\a\msquic\msquic\build\windows\x86_schannel\src\tools\forwarder\quicforward.vcxproj]
D:\a\msquic\msquic\src\tools\forwarder\forwarder.cpp(60,77): warning C4244: 'argument': conversion from 'uint64_t' to 'size_t', possible loss of data [D:\a\msquic\msquic\build\windows\x86_schannel\src\tools\forwarder\quicforward.vcxproj]
```

src/tools/forwarder/forwarder.cpp
```diff
@@ -0,0 +1,213 @@
...
+QUIC_CERTIFICATE_HASH Cert;
+bool BufferedMode = true;
+
+const MsQuicApi* MsQuic;
+MsQuicRegistration* Registration;
+MsQuicConfiguration* FrontEndConfiguration;
+MsQuicConfiguration* BackEndConfiguration;
+
+#define USAGE \
+    "Usage: quicforward <alpn> <local-port> <target-name/ip>:<target-port> <thumbprint> [0/1-buffered-mode]\n"
```
csujedihy, line 29: Is it possible to make a totally transparent QUIC proxy? i.e. without specifying alpn and cert stuff on the proxy side?

src/tools/forwarder/forwarder.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+Abstract:
+
+    This tool creates a terminating QUIC proxy to forward all incoming traffic
+    to a specified target.
+
+--*/
+
+#include "msquichelper.h"
+#include "msquic.hpp"
```
csujedihy, line 14: If this serves as a sample code, do we really want to use cpp where the wrapper isn't officially supported?

csujedihy (approved):

src/tools/forwarder/forwarder.cpp
nibanks: the ALPN must be the same on both side since you're proxying the same protocol. Certificate is also something you only specify on the front end (where you're the server). You can't "proxy" the backend's cert.

src/tools/forwarder/forwarder.cpp
nibanks: It's not necessarily a sample, but might actually grow to a real tool. Also, I have plans of eventually officially supporting the C++ wrapper.

PR #4015, opened 2023-12-17
Support Optional Units on secnetperf Arguments

nibanks:
## Description

Now instead of doing:
```
secnetperf -target:localhost -up:10000 -timed:1
```
You do:
```
secnetperf -target:localhost -up:10s
```
It supports both length/size units (gb, ms, kb, b) and time units (m, s, ms, us).

## Testing

Tested locally. CI/CD

## Documentation

Updated help text.


maolson-msft:
I suggest "GB" rather than "gb" since both bytes and bits are used in our field

nibanks:
> I suggest "GB" rather than "gb" since both bytes and bits are used in our field

It's generally all case insensitive.

src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -76,14 +76,13 @@ PrintHelp(
...
-        "  -download:<####>         The length of bytes to receive on each stream. (def:0)\n"
+        "  -upload:<####>[unit]     The length of bytes to send on each stream, with an optional (time or length) unit. (def:0)\n"
+        "  -download:<####>[unit]   The length of bytes to receive on each stream, with an optional (time or length) unit. (def:0)\n"
         "  -iosize:<####>           The size of each send request queued.\n"
-        "  -timed:<0/1>             Indicates the upload/download args are times (in ms). (def:0)\n"
         //"  -inline:<0/1>            Create new streams on callbacks. (def:0)\n"
         "  -rconn:<0/1>             Repeat the scenario at the connection level. (def:0)\n"
         "  -rstream:<0/1>           Repeat the scenario at the stream level. (def:0)\n"
-        "  -runtime:<####>          The total runtime (in ms). Only relevant for repeat scenarios. (def:0)\n"
+        "  -runtime:<####>[unit]    The total runtime, with an optional unit (def unit is us). Only relevant for repeat scenarios. (def:0)\n"
```
anrossi, line 85: default is us, really? that seems... overly precise.

src/perf/lib/PerfClient.cpp
```diff
@@ -15,6 +15,58 @@
...
+    while (*names && (value = (char*)GetValue(argc, argv, *names)) == nullptr) {
+        names++;
+    }
+    if (!value) { return false; }
+
+    // Search to see if the value has a time unit specified at the end.
+    for (uint32_t i = 0; i < ARRAYSIZE(TimeUnits); ++i) {
+        size_t len = strlen(TimeUnits[i]);
+        if (len < strlen(value) &&
+            _strnicmp(value + strlen(value) - len, TimeUnits[i], len) == 0) {
```
csujedihy, line 46: wouldn't "ms" always get recognized as "s" as "s" get checked first?

anrossi (approved):

src/perf/lib/PerfClient.cpp
nibanks: Yeah, good catch. I blame Copilot. 😆 I will reorder `s` last.

csujedihy (approved):

PR #4031, opened 2023-12-28
fix: apply lb settings after lb mode change

qzhuyan:
## Description

Without this fix, the CID prefix len: `MsQuicLib.CidServerIdLength` and source CID length (`MsQuicLib.CidTotalLength`) are not updated after changing the lb mode from disabled (default) to `QUIC_LOAD_BALANCING_SERVER_ID_IP` which causes the Server source CIDs are not prefixed with src IP. 

## Testing

## Documentation

### DIFF
--- a/src/core/library.c
+++ b/src/core/library.c
@@ -839,6 +839,8 @@ QuicLibrarySetGlobalParam(
             "[ lib] Updated load balancing mode = %hu",
             MsQuicLib.Settings.LoadBalancingMode);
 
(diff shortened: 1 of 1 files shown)


nibanks:
1. We need to apply changes to `main` first and then we can backport (if decided) second. So please make this PR on `main`.
2. Can you please add a test case to validate this change,

qzhuyan:
replaced with https://github.com/microsoft/msquic/pull/4036

PR #4032, opened 2023-12-29
Simplify timer operations

csujedihy:
## Description

The existing timer operations rely on a sorted timer entry array, where each entry has a type and expiration time, to maintain the earliest expiration time. However, it's not necessary to maintain a sorted array for this purpose. The earliest expiration time can be calculated on the fly when setting/canceling/expiring timers with the same complexity (it doesn't really matter as we only have a few timer types) and more readable code.

## Testing

CI

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -1120,6 +1120,20 @@ QuicConnReplaceRetiredCids(
 }
 
 _IRQL_requires_max_(PASSIVE_LEVEL)
+uint64_t
+QuicGetEarliestExpirationTime(
+    _In_ const QUIC_CONNECTION* Connection
+    )
+{
+    uint64_t EarliestExpirationTime = UINT64_MAX;
+    for (QUIC_CONN_TIMER_TYPE Type = 0; Type < QUIC_CONN_TIMER_COUNT; ++Type) {
+        EarliestExpirationTime =
+            CXPLAT_MIN(EarliestExpirationTime, Connection->ExpirationTimes[Type]);
+    }
+    return EarliestExpirationTime;
+}
+
+_IRQL_requires_max_(PASSIVE_LEVEL)
 void
 QuicConnTimerSetEx(
     _Inout_ QUIC_CONNECTION* Connection,
@@ -1129,6 +1143,7 @@ QuicConnTimerSetEx(
     )
 {
     const uint64_t NewExpirationTime = TimeNow + Delay;
+    uint64_t NewEarliestExpirationTime = UINT64_MAX;
 
     QuicTraceEvent(
         ConnSetTimer,
@@ -1138,7 +1153,8 @@ QuicConnTimerSetEx(
         Delay);
 
     Connection->ExpirationTimes[Type] = NewExpirationTime;
-    if (Connection->EarliestExpirationTime > NewExpirationTime) {
+    NewEarliestExpirationTime  = QuicGetEarliestExpirationTime(Connection);
+    if (NewEarliestExpirationTime != Connection->EarliestExpirationTime) {
         Connection->EarliestExpirationTime = NewExpirationTime;
         QuicTimerWheelUpdateConnection(&Connection->Worker->TimerWheel, Connection);
     }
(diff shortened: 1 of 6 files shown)


src/core/connection.c
```diff
@@ -1137,56 +1137,9 @@ QuicConnTimerSetEx(
...
-        Connection->Timers[CurIndex].ExpirationTime = NewExpirationTime;
-        NewIndex = CurIndex;
-    }
-
-    if (NewIndex == 0 || CurIndex == 0) {
-        //
-        // The first timer was updated, so make sure the timer wheel is updated.
-        //
+    Connection->ExpirationTimes[Type] = NewExpirationTime;
+    if (Connection->EarliestExpirationTime > NewExpirationTime) {
```
nibanks, line 1141: What if you changed the previous earliest time? You need to update that here too then.

src/core/connection.h
```diff
@@ -508,9 +491,14 @@ typedef struct QUIC_CONNECTION {
     uint8_t CibirId[2 + QUIC_MAX_CIBIR_LENGTH];
 
     //
-    // Sorted array of all timers for the connection.
+    // Expiration time (absolute time in us) for each timer type. We use UINT64_MAX as a sentinel
+    // to indicate that the timer is not set.
+    //
+    uint64_t ExpirationTimes[QUIC_CONN_TIMER_COUNT];
+    //
```
nibanks, line 498: nit: add a newline before this

src/core/connection.c
csujedihy: You meant when the timer type being set happens to hold the earliest expiration time and we are setting it to a later time (setting it to an earlier time is covered by the if check)?

That case is handled in a lazy manner. Since the earliest exp time is not updated, it's possible that we end up calling QuicConnTimerExpired for that stale earliest exp time. However, since the value's guaranteed to be earlier than any scheduled timers on this connection, there will be no side-effect, and QuicConnTimerExpired will then update the timer wheel with up-to-date earliest exp time. I guess we don't need to have tricky stuff like this.

src/core/connection.h
```diff
@@ -736,7 +725,7 @@ QuicConnGetNextExpirationTime(
     _In_ const QUIC_CONNECTION * const Connection
     )
 {
-    return Connection->Timers[0].ExpirationTime;
+    return Connection->EarliestExpirationTime;
```
nibanks, line 728: Probably worth just refactoring out this function now.

src/core/connection.c
```diff
@@ -1119,6 +1119,20 @@ QuicConnReplaceRetiredCids(
     return TRUE;
 }
 
+_IRQL_requires_max_(PASSIVE_LEVEL)
```
nibanks, line 1122:
```suggestion
_IRQL_requires_max_(DISPATCH_LEVEL)
```
Not that it really matters, but there is no reason to require passive.

src/core/connection.c
```diff
@@ -1119,6 +1119,20 @@ QuicConnReplaceRetiredCids(
...
+uint64_t
+QuicGetEarliestExpirationTime(
+    _In_ const QUIC_CONNECTION* Connection
+    )
+{
+    uint64_t EarliestExpirationTime = UINT64_MAX;
+    for (QUIC_CONN_TIMER_TYPE Type = 0; Type < QUIC_CONN_TIMER_COUNT; ++Type) {
+        EarliestExpirationTime =
+            CXPLAT_MIN(EarliestExpirationTime, Connection->ExpirationTimes[Type]);
+    }
```
nibanks, line 1132:
```suggestion
    uint64_t EarliestExpirationTime = Connection->ExpirationTimes[0];
    for (QUIC_CONN_TIMER_TYPE Type = 1; Type < QUIC_CONN_TIMER_COUNT; ++Type) {
        if (Connection->ExpirationTimes[Type] < EarliestExpirationTime) {
            EarliestExpirationTime = Connection->ExpirationTimes[Type];
        }
    }
```
Very minor improvements this way.

src/core/connection.c
```diff
@@ -1129,6 +1143,7 @@ QuicConnTimerSetEx(
     )
 {
     const uint64_t NewExpirationTime = TimeNow + Delay;
+    uint64_t NewEarliestExpirationTime;
```
nibanks, line 1146: Since there doesn't seem to be a `goto` between here and when the variable is written, just remove this line and add the `uint64_t` below, where it is set.

src/core/connection.c
```diff
@@ -1198,66 +1167,34 @@ QuicConnTimerCancel(
...
-                    //
-                    CxPlatMoveMemory(
-                        Connection->Timers + i,
-                        Connection->Timers + i + 1,
-                        sizeof(QUIC_CONN_TIMER_ENTRY) * (j - i - 1));
-                    Connection->Timers[j - 1].Type = Type;
-                    Connection->Timers[j - 1].ExpirationTime = UINT64_MAX;
-                }
+    if (Connection->ExpirationTimes[Type] == Connection->EarliestExpirationTime) {
+        uint64_t NewEarliestExpirationTime = UINT64_MAX;
```
nibanks, line 1180: Remove this line and move type to first real write below.

src/core/connection.c
```diff
@@ -1268,66 +1205,57 @@ QuicConnTimerExpired(
...
+                        AllocFailure,
+                        "Allocation of '%s' failed. (%llu bytes)",
+                        "expired timer operation",
+                        0);
+                }
             }
+        } else {
+            Connection->EarliestExpirationTime =
+                CXPLAT_MIN(Connection->EarliestExpirationTime, Connection->ExpirationTimes[Type]);
         }
```
nibanks, line 1259:
```suggestion
        } else if (Connection->ExpirationTimes[Type] < Connection->EarliestExpirationTime) {
            Connection->EarliestExpirationTime = Connection->ExpirationTimes[Type];
        }
```

src/core/connection.c
csujedihy: all the functions in timer_wheel.c require passive level. I guess that's incorrect too?

nibanks: Anything that is just running logic can be at dispatch. If it calls into other layers (i.e. the rest of the connection), then it might need to be at passive. But don't worry about going and changing all that right now.

src/core/connection.c
```diff
@@ -1198,66 +1167,32 @@ QuicConnTimerCancel(
     _In_ QUIC_CONN_TIMER_TYPE Type
     )
 {
-    for (uint32_t i = 0;
-        i < ARRAYSIZE(Connection->Timers) &&
-            Connection->Timers[i].ExpirationTime != UINT64_MAX;
-        ++i) {
+    CXPLAT_DBG_ASSERT(Connection->EarliestExpirationTime <= Connection->ExpirationTimes[Type]);
 
+    if (Connection->EarliestExpirationTime == UINT64_MAX) {
```
nibanks, line 1172: I'm wondering how necessary this check is... It might be simpler just to remove it.

nibanks (approved):

PR #4036, opened 2024-01-03
fix: apply lb settings after lb mode change

qzhuyan:
## Description

Fixing an issue where the CID prefix
lengths (`MsQuicLib.CidServerIdLength` and `MsQuicLib.CidTotalLength`) weren't being updated properly when transitioning the lb mode from its default state (disabled) to `QUIC_LOAD_BALANCING_SERVER_ID_IP`. This resulted in the server source CIDs not being prefixed with the source IP

## Testing

New unittest:
```TEST(SettingsTest, GlobalLoadBalancingServerIDSet)```

## Documentation

NO

### DIFF
--- a/src/core/library.c
+++ b/src/core/library.c
@@ -856,6 +856,9 @@ QuicLibrarySetGlobalParam(
 
         MsQuicLib.Settings.LoadBalancingMode = *(uint16_t*)Buffer;
         MsQuicLib.Settings.IsSet.LoadBalancingMode = TRUE;
+
+        QuicLibApplyLoadBalancingSetting();
+
         QuicTraceLogInfo(
             LibraryLoadBalancingModeSet,
             "[ lib] Updated load balancing mode = %hu",
--- a/src/core/unittest/SettingsTest.cpp
+++ b/src/core/unittest/SettingsTest.cpp
@@ -525,6 +525,31 @@ TEST(SettingsTest, GlobalSettingsSizesSet)
     }
 }
 
+TEST(SettingsTest, GlobalLoadBalancingServerIDSet)
+{
+    uint16_t Mode = QUIC_LOAD_BALANCING_SERVER_ID_IP;
+    uint16_t OldMode = MsQuicLib.Settings.LoadBalancingMode;
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/core/library.c [hunk 1]
```diff
@@ -856,6 +856,9 @@ QuicLibrarySetGlobalParam(
 
         MsQuicLib.Settings.LoadBalancingMode = *(uint16_t*)Buffer;
         MsQuicLib.Settings.IsSet.LoadBalancingMode = TRUE;
+
+        QuicLibApplyLoadBalancingSetting();
```
nibanks, line 860: I think this function will need some synchronization if this function is called while in the middle of operation (i.e. connections exist).

qzhuyan: I read L848 that this set op will fail early if `MsQuicLib.InUse` is true, so I assume there is no ongoing traffic (No active registration) at the time of this applying?

qzhuyan: or do you mean we need atomic changes in `QuicLibApplyLoadBalancingSetting`?

nibanks: I think the simplest solution is to restrict this SetParam to only work when `!MsQuicLib.InUse`, and then there should be no race condition related to calling `QuicLibApplyLoadBalancingSetting`. Does that work for you?

qzhuyan: I am ok with this, I understand I cannot change LB mode once you have traffic running that means `MsQuicLib.InUse` is `false` and there is no registration. 
My understanding is we have it already, just few lines above.

https://github.com/microsoft/msquic/blob/8faef0db35c9d8d6758f0bf976de22bfc6b05951/src/core/library.c#L848-L854

`QuicLibApplyLoadBalancingSetting` will not be called if `MsQuicLib.InUse` is true.

So what I should update now is to protects these lines with mutex to avoid race of calling `QuicLibApplyLoadBalancingSetting` ?

nibanks: Ah, you're right. I missed that somehow. So, yeah, I think this PR is good to go then. Thanks for baring with me!

nibanks (approved):

PR #4042, opened 2024-01-07
Receive Buffer Refactoring

nibanks:
## Description

Refactors the receive buffer interface to support 3 modes of operation:

1. Single buffer mode - This is used by the crypto stream, and ensures reads are always one continuous block (because TLS doesn't support gather semantics).
2. Circular buffer mode - This is used by the streams, and allows for the read calls to return 2 pointers/lengths that map to different parts of the circular buffers stored internally.
3. Multi-read buffer mode - This will be used (opt-in) by the streams in a future PR. It allows for multiple outstanding reads from a single receive buffer.

The first two "modes" already practically existed, but were defined with slightly different flags. Now the distinction is very clear. There is also a lot of other logic internally required to manage external references in a more generic method.

## Testing

Lots of new unit tests added. Also leverages existing CI/CD for any e2e issues.

## Documentation

Heavily commented code, but no external docs.


nibanks:
> Based on the high-level description, CIRCULAR sounds like just a N=1 case of MULTIPLE. So why have a separate mode?

They're related, but not exactly that simple. Only the first buffer is used in a circular fashion. This comes does to the usage pattern, and how draining from the front happens. Only the first buffer is ever drained from, so only it may have a case where partial data is read from, and it makes sense (to eliminate the copy) to just move where we consider the HEAD.

src/core/unittest/RecvBufferTest.cpp
```diff
@@ -19,362 +19,614 @@
...
     ASSERT_EQ(WriteLength+WriteOffset, RecvBuf.GetTotalLength()); // Total length should be offset plus write length
     uint64_t ReadOffset;
-    QUIC_BUFFER ReadBuffer;
-    ASSERT_EQ(
-        WriteOffset == 0, // If we didn't write to the front, then it's not ready to read yet
-        RecvBuf.Read(&ReadOffset, &ReadBuffer));
-    if (WriteOffset != 0) return; // Nothing else to validate if we didn't write to the front
+    QUIC_BUFFER ReadBuffers[3];
+    uint32_t BufferCount = ARRAYSIZE(ReadBuffers);
+    if (!NewDataReady) return; // Can't read if it's not ready
```
ProjectsByJackHe, line 168: Not needed? You did an assert earlier.

nibanks: No, because different modes have different behaviors.

nibanks: The assert `ASSERT_EQ(WriteOffset == 0, NewDataReady != FALSE);` indicates that it's possible for `NewDataReady` to be `FALSE`. So on this line, we bail in that scenario. Because if we were to try to read when not ready, that would assert in the product code (you're not supposed to do that).

ami-GS (approved):

ProjectsByJackHe (approved):

Issue #4044, opened 2024-01-09
Stream blocked timings stats incorrectly initialized


jarnovanderlinden:
### Describe the bug

The BlockedTimings for a stream are incorrectly initialized in QuicStreamStart(). For example:
```c++
    Stream->BlockedTimings.CachedConnSchedulingUs =
        Stream->Connection->BlockedTimings.Scheduling.CumulativeTimeUs +
        Stream->Connection->BlockedTimings.Scheduling.LastStartTimeUs != 0 ?
            CxPlatTimeDiff64(Stream->Connection->BlockedTimings.Scheduling.LastStartTimeUs, Now) : 0;
```
should probably be:
```c++
    Stream->BlockedTimings.CachedConnSchedulingUs =
        Stream->Connection->BlockedTimings.Scheduling.CumulativeTimeUs +
        ( Stream->Connection->BlockedTimings.Scheduling.LastStartTimeUs != 0 ?
            CxPlatTimeDiff64(Stream->Connection->BlockedTimings.Scheduling.LastStartTimeUs, Now) : 0 );
```
and so on for the other BlockedTimings initializations.

### Affected OS

- [ ] Windows
- [ ] Linux
- [ ] macOS
- [ ] Other (specify below)

### MsQuic version

main

### Steps taken to reproduce bug

By code inspection after noticing unexpectedly large values from getting QUIC_PARAM_STREAM_STATISTICS.

### Expected behavior

Expect to see small values.

### Actual outcome

See very large values.

nibanks:
Thanks for the bug. I didn't realize `?:` operator was so low in the precedence order. We will fix, or if you'd like, feel free to make a PR.

PR #4046, opened 2024-01-10
Fix stream blockedtimings stats initialization

jarnovanderlinden:
## Description

Fix for stream blockedtimings stats intialization https://github.com/microsoft/msquic/issues/4044
Just a matter of ensuring the correct parts of the expression are used for the ternary operator condition.


jarnovanderlinden:
@microsoft-github-policy-service agree company="Vizrt, Inc."

nibanks (approved):
Thanks for making these changes!

PR #4050, opened 2024-01-12
Restart Server on Each NetPerf Test Run

nibanks:
## Description

Improvements to the NetPerf automation to restart the server for each test, but also includes several other fixes found along the way:

- Refactoring the helper to have a "run" of a test to run the server only for the time it takes to do the couple of tries for that test.
- Fix log collections (client and server).
- Collect crash dumps (Windows only for now) on client and server.
- Fix a bug in the perf server code that wasn't passing the context pointer in for a callback.
- Improvements (but not a complete fix yet) of the TCP cleanup path to handle some race conditions.

## Testing

NetPerf Automation

## Documentation

N/A

### DIFF
--- a/scripts/secnetperf-helpers.psm1
+++ b/scripts/secnetperf-helpers.psm1
@@ -106,8 +106,10 @@ function Invoke-Secnetperf {
 
     if ($LogProfile -ne "" -and $LogProfile -ne "NULL") {
         Invoke-Command -Session $Session -ScriptBlock {
+            try { & "$Using:RemoteDir/scripts/log.ps1" -Cancel } catch {} # Cancel any previous logging
             & "$Using:RemoteDir/scripts/log.ps1" -Start -Profile $Using:LogProfile -ProfileInScriptDirectory
         }
+        try { .\scripts\log.ps1 -Cancel } catch {} # Cancel any previous logging
         .\scripts\log.ps1 -Start -Profile $LogProfile
     }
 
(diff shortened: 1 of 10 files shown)


scripts/secnetperf-helpers.psm1
```diff
@@ -3,14 +3,78 @@
...
 
-# Write a GitHub warning message to the console.
-function Write-GHWarning($msg) {
-    Write-Host "::warning::$msg"
+# Configured the remote machine to collect dumps on crash.
+function Configure-DumpCollection {
+    param ($Session)
+    if ($isWindows) {
+        Invoke-Command -Session $Session -ScriptBlock {
+            $DumpDir = "C:/_work/quic/artifacts/crashdumps"
```
ami-GS, line 22: Is this common to create on top directory? how about using working directory? or C:/_work/quic is already commonly used for any purpose by msquic perf test?

scripts/secnetperf-helpers.psm1
```diff
@@ -3,14 +3,78 @@
...
+    }
+    return $false
+}
+
+# Collect any crash dumps that were generated on the remote machine.
+function Collect-RemoteDumps {
+    param ($Session, $OutputDir)
+    if ($isWindows) {
+        $DumpFiles = Invoke-Command -Session $Session -ScriptBlock {
+            Get-ChildItem "C:/_work/quic/artifacts/crashdumps" | Where-Object { $_.Extension -eq ".dmp" }
```
ami-GS, line 62: Use globally defined const name?

scripts/secnetperf-helpers.psm1
```diff
@@ -63,117 +127,133 @@ function Stop-RemoteServer {
...
+
+    $values = @(@(), @())
+    $hasFailures = $false
+
+    # TODO: This logic is pretty fragile. Needs improvement.
+    $metric = "throughput-download"
+    if ($exeArgs.Contains("plat:1")) {
+        $metric = "latency"
+    } elseif ($exeArgs.Contains("prate:1")) {
+        $metric = "hps"
```
ami-GS, line 162: How are these "plat" and "prate" associated with each metric name? Looks really high context naming

src/inc/msquic.hpp
```diff
@@ -207,20 +207,30 @@ class CxPlatThread {
...
     CxPlatEvent ShutdownEvent {true};
     CxPlatThread WatchdogThread;
     uint32_t TimeoutMs;
+    bool WriteToConsole;
     static CXPLAT_THREAD_CALLBACK(WatchdogThreadCallback, Context) {
         auto This = (CxPlatWatchdog*)Context;
         if (!This->ShutdownEvent.WaitTimeout(This->TimeoutMs)) {
+#ifndef _KERNEL_MODE // Not supported in kernel mode
+            if (This->WriteToConsole) {
+                printf("Error: Watchdog timeout fired!\n");
```
ami-GS, line 224: If error, how about `fprintf(stderr, )` ?

scripts/secnetperf-helpers.psm1
nibanks: Yes, for our perf machines, this is a fixed path.

scripts/secnetperf-helpers.psm1
nibanks: As mentioned in the comment above, this stuff is fragile and needs to be improved in future PRs.

src/inc/msquic.hpp
nibanks: Printing to error stream causes other complexity at the PowerShell level if you decide you want to ignore certain errors (temporarily) like we are doing currently. So I prefer to keep it this way.

scripts/secnetperf-helpers.psm1
nibanks: This is the "common" working directory for our tests. It's used throughout.

ami-GS (approved):

PR #4051, opened 2024-01-12
Update the Reusable Build Workflows to Support Perf Builds

nibanks:
## Description

Updates to allow NetPerf to build just the perf tools to improve build time, as well as copy/download artifact time.

Each reusable workflow renames the `test` parameter to a more general `build` parameter that does much the same thing, but also allows for `-Perf` to be passed in, and it will disable test and tools from being built.

## Testing

CI/CD

## Documentation

N/A

### DIFF
--- a/.github/workflows/package-reuse-linux.yml
+++ b/.github/workflows/package-reuse-linux.yml
@@ -42,9 +42,13 @@ on:
         required: false
         default: ''
         type: string
-      test:
+      sanitize:
         required: false
         default: ''
+        type: string
+      build:
+        required: false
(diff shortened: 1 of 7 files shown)


.github/workflows/package-reuse-linux.yml
```diff
@@ -42,10 +42,14 @@ on:
         required: false
         default: ''
         type: string
-      test:
+      sanitize:
```
nibanks, line 45: This seemed like a bug that it was missing. @csujedihy can you confirm?

mtfriesen (approved):

PR #4052, opened 2024-01-12
Allow NMR helpers to not wait at all

csujedihy:
## Description

Passing zero as timeout will cause KeWaitForSingleObject to not wait at all. We will lose the ability to wait forever but that's not useful and can still be somewhat achieved by passing a ULONG_MAX.

## Testing

CI

### DIFF
--- a/src/inc/msquic.h
+++ b/src/inc/msquic.h
@@ -1722,7 +1722,11 @@ __MsQuicClientAttachProvider(
                 NULL,
                 &ProviderContext,
                 (const void**)&Client->ProviderDispatch);
-        KeSetEvent(&Client->RegistrationCompleteEvent, IO_NO_INCREMENT, FALSE);
+        if (NT_SUCCESS(Status)) {
+            KeSetEvent(&Client->RegistrationCompleteEvent, IO_NO_INCREMENT, FALSE);
+        } else {
+            InterlockedDecrement(&Client->BindingCount);
+        }
     } else {
         Status = STATUS_NOINTERFACE;
     }
--- a/src/inc/msquic.h
+++ b/src/inc/msquic.h
@@ -1689,6 +1689,7 @@ typedef struct MSQUIC_NMR_DISPATCH {
 //
 typedef struct __MSQUIC_NMR_CLIENT {
     NPI_CLIENT_CHARACTERISTICS NpiClientCharacteristics;
+    LONG BindingCount;
     HANDLE NmrClientHandle;
     NPI_MODULEID ModuleId;
     KEVENT RegistrationCompleteEvent;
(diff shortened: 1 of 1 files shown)


mtfriesen (dismissed):

src/inc/msquic.h
```diff
@@ -1712,15 +1713,20 @@ __MsQuicClientAttachProvider(
...
-    #pragma warning(suppress:6387) // _Param_(2) could be '0' - by design.
-    Status =
-        NmrClientAttachProvider(
-            NmrBindingHandle,
-            Client,
-            NULL,
-            &ProviderContext,
-            (const void**)&Client->ProviderDispatch);
-    KeSetEvent(&Client->RegistrationCompleteEvent, IO_NO_INCREMENT, FALSE);
+    if (InterlockedIncrement(&Client->BindingCount) == 1) {
```
mtfriesen, line 1716: I'm not sure we want to do this in this PR - the rest of the code looks sketchy if more that one NMR binding occurs.

nibanks (dismissed):

src/inc/msquic.h
csujedihy: We do this to allow exactly one binding. Did I do it wrong?

mtfriesen: Ah, no, it will prevent more than one client, you're right. It should decrement the count if the attach fails below, but that's not what I was concerned about.

I was mainly concerned that the provider dispatch could get overwritten, but that's within the if block.

csujedihy:
@nibanks "84 test(s) failed, 84 test(s) crashed." 84 test cases failed in winkernel on prerelease. Are you aware of this?

src/inc/msquic.h
csujedihy: oh, right. NmrClientAttachProvider could theorectically fail.

nibanks:
> @nibanks "84 test(s) failed, 84 test(s) crashed." 84 test cases failed in winkernel on prerelease. Are you aware of this?

No? What happened? Just your PR?

csujedihy:
> > @nibanks "84 test(s) failed, 84 test(s) crashed." 84 test cases failed in winkernel on prerelease. Are you aware of this?
> 
> No? What happened? Just your PR?

I just saw it in this PR. Not sure if it happened elsewhere. It passed on retail releases but not on prerelease.

src/inc/msquic.h
csujedihy: Hmm, this is still not bulletproof. We need to hold an exclusive lock while doing the unique binding check and calling NmrClientAttachProvider.

mtfriesen: doesn't the interlocked increment/decrement provide an implicit mutex?

csujedihy: There is a case where we increment the binding count to 1 but then NmrClientAttachProvider fails and on another thread clientAttachProvider is being called before the decrement happens. In this case, no bindings will be formed.

mtfriesen: Oh I see, I assumed that was by design since the first provider could very possibly have succeeded. IMO it's fine to kick that can down the road until we have dynamic load/unload with multiple providers fully specified and tested.

mtfriesen (approved):

src/inc/msquic.h
csujedihy: Agreed. We can push that back until the provider actually can fail to attach.

nibanks:
> > > @nibanks "84 test(s) failed, 84 test(s) crashed." 84 test cases failed in winkernel on prerelease. Are you aware of this?
> > 
> > 
> > No? What happened? Just your PR?
> 
> I just saw it in this PR. Not sure if it happened elsewhere. It passed on retail releases but not on prerelease.

I just checked the last main run from 5 hours ago (https://github.com/microsoft/msquic/actions/runs/7505154403) and it passed. So, either your build just picked up a new prerelease that broke things OR your PR _somehow_ breaks those tests.

PR #4055, opened 2024-01-13
Fix tests not targeting private msquic.sys

csujedihy:
## Description

The inbox msquic and the driver built by the CI/CD pipeline share the same NMR NPI. The tests might be testing the inbox version. This PR makes msquicpriv and tests use a private NPI that's different from the public msquic NMR NPI.

## Testing

CI

### DIFF
--- a/src/test/bin/winkernel/control.cpp
+++ b/src/test/bin/winkernel/control.cpp
@@ -84,7 +84,9 @@ QuicTestCtlInitialize(
     WDF_IO_QUEUE_CONFIG QueueConfig;
     WDFQUEUE Queue;
 
+#ifdef QUIC_TEST_NMR_PROVIDER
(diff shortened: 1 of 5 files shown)


src/test/bin/winkernel/control.cpp
```diff
@@ -82,6 +84,8 @@ QuicTestCtlInitialize(
     WDF_IO_QUEUE_CONFIG QueueConfig;
     WDFQUEUE Queue;
 
+    QUIC_ENABLE_PRIVATE_NMR_PROVIDER();
```
nibanks, line 87: You have to have two versions. Normal tests still need to use the normal MMR interface, because that's what's tested inbox.

nibanks (approved):

PR #4059, opened 2024-01-16
Fix logic to persist data to JSON

ProjectsByJackHe:
## Description

With the merge of https://github.com/microsoft/msquic/commit/b0edc581f4ec2e230fe99fa5e57c0b198e3da6fe, it looks like the JSON produced is missing the entire array, but including only a single value like this: 

image

This PR will fix that. 

UPDATE: https://github.com/microsoft/netperf/actions/runs/7550005766

## Testing

CI

## Documentation

N/A

### DIFF
--- a/scripts/secnetperf-helpers.psm1
+++ b/scripts/secnetperf-helpers.psm1
@@ -199,7 +199,11 @@ function Invoke-Secnetperf {
             Write-Host $rawOutput
             if ($metric -eq "latency") {
                 $latency_percentiles = '(?<=\d{1,3}(?:\.\d{1,2})?th: )\d+'
-                $values[$tcp] += [regex]::Matches($rawOutput, $latency_percentiles) | ForEach-Object {$_.Value}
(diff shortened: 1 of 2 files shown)


scripts/secnetperf-helpers.psm1
```diff
@@ -199,7 +199,11 @@ function Invoke-Secnetperf {
             Write-Host $rawOutput
             if ($metric -eq "latency") {
                 $latency_percentiles = '(?<=\d{1,3}(?:\.\d{1,2})?th: )\d+'
-                $values[$tcp] += [regex]::Matches($rawOutput, $latency_percentiles) | ForEach-Object {$_.Value}
+                $perc = [regex]::Matches($rawOutput, $latency_percentiles) | ForEach-Object {$_.Value}
+                $values[$tcp] += $perc
```
nibanks, line 203: Why would splitting across two different lines make a difference?

ProjectsByJackHe: Oh I am using this to debug, to check if powershell is doing something weird where it truncates certain values. Once a bug has been found/fixed I can convert this PR from draft mode.

PR #4062, opened 2024-01-18
Run secnetperf in Separate Process to Wait on (with Timeout)

nibanks:
## Description

We've been seeing hangs on shutdown, so we are going to try running the test in a separate process so we can wait with a timeout. Also updated the yaml to trigger when necessary. Finally, should enable core dump collection for Linux.

## Testing

netperf automation

## Documentation

N/A

### DIFF
--- a/.github/workflows/netperf.yml
+++ b/.github/workflows/netperf.yml
@@ -5,9 +5,35 @@ on:
   push:
     branches:
     - main
+    paths:
+    - .github/workflows/netperf.yml
+    - .gitmodules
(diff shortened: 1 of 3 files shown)


ProjectsByJackHe:
Do you have a successful / semi-successful run you can paste here? I see 

https://github.com/microsoft/netperf/actions/runs/7574015320/job/20627588004

which is running into an issue

ProjectsByJackHe (approved):
Run looks good, and artifacts produced looks good as well.

PR #4064, opened 2024-01-19
Support Testing Multiple IO Models with netperf

nibanks:
## Description

This updates netperf so that it may test things like RIO and XDP.

## Testing

Verified with a couple private runs.

## Documentation

N/A


ProjectsByJackHe (approved):
LGTM

PR #4082, opened 2024-01-25
Get Kernel Mode netperf Running

nibanks:
## Description

Working around the cert issues by:

1. Only load certs on server, since they're unnecessary on client.
2. Always use user certs in the MY store, even for kernel.

Also made a few other minor fixes I found necessary while debugging.

## Testing

Manual testing.

## Documentation

N/A


csujedihy (approved):

mtfriesen (dismissed):

ProjectsByJackHe (approved):
HPS / kernel mode TODO looks good.

PR #4083, opened 2024-01-26
Update Schannel Logic to Handler Larger Output Buffers

nibanks:
## Description

A user reported an issue with the schannel variant of MsQuic sometimes not working, and it was an issue with too large of output buffers. This change updates the code to dynamically increase the buffer size as necessary.

**Note** - We might want to backport this.

## Testing

Updated a test to exercise the new code paths.

## Documentation

N/A

### DIFF
--- a/src/platform/tls_schannel.c
+++ b/src/platform/tls_schannel.c
@@ -2132,7 +2132,16 @@ CxPlatTlsWriteDataToSchannel(
             SchannelOutBufferTooSmall,
             TlsContext->Connection,
             "Increasing TLS output buffer size");
-        uint16_t NewBufferLength = State->BufferAllocLength << 2; // TODO - Where to get the new length from?
+        uint16_t NewBufferLength = State->BufferAllocLength << 1;
+        if (NewBufferLength < State->BufferAllocLength) { // Integer overflow.
+            QuicTraceEvent(
+                TlsError,
+                "[ tls][%p] ERROR, %s.",
+                TlsContext->Connection,
+                "TLS buffer too large");
+            Result |= CXPLAT_TLS_RESULT_ERROR;
+            break;
+        }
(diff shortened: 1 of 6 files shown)


src/test/lib/TestHelpers.h
```diff
@@ -319,14 +319,15 @@ struct StatelessRetryHelper
...
 
 struct PrivateTransportHelper : QUIC_PRIVATE_TRANSPORT_PARAMETER
 {
-    PrivateTransportHelper(bool Enabled) {
+    PrivateTransportHelper(bool Enabled, bool Resumption = false) {
         if (Enabled) {
             Type = PRIVATE_TP_TYPE;
-            Length = PRIVATE_TP_LENGTH;
-            Buffer = new(std::nothrow) uint8_t[PRIVATE_TP_LENGTH];
+            Length = Resumption ? PRIVATE_TP_LENGTH : PRIVATE_TP_LENGTH_HUGE;
```
nibanks, line 329: Must do this because increasing the size increases the chance that 0-RTT doesn't work in the face of random packet loss.

src/platform/tls_schannel.c
```diff
@@ -2123,6 +2123,45 @@ CxPlatTlsWriteDataToSchannel(
...
+    case SEC_E_BUFFER_TOO_SMALL: {
+        //
+        // The output buffer for the TLS response is too small. We need to grow
+        // the buffer and try again.
+        //
+        QuicTraceLogConnInfo(
+            SchannelOutBufferTooSmall,
+            TlsContext->Connection,
+            "Increasing TLS output buffer size");
+        uint16_t NewBufferLength = State->BufferAllocLength << 1;
```
anrossi, line 2135: Doesn't the assignment need a cast to prevent assigning a 4-byte int to a 2-byte int?

anrossi: well, it built, so I guess not.

anrossi (approved):

Issue #4092, opened 2024-01-28
libmsquic package is missing for Fedora 39


sillyotter:
### Describe the bug

Referencing Issue #3610 and #3640, it seems that libmsquic is not being published for Fedora 39.  39 is listed as a supported distro.

https://packages.microsoft.com/fedora/39/prod/Packages/ does not have an l folder in it with the msquic packages in it.

### Affected OS

- [ ] Windows
- [X] Linux
- [ ] macOS
- [ ] Other (specify below)

### MsQuic version

v2.2

### Steps taken to reproduce bug

dnf install libmsquic

### Expected behavior

should be there, as I understand the earlier issues, and install

### Actual outcome

does not install.

### Additional details

I'm guessing this will happen again when fedora 40 comes out, if some other fix isn't put in place to keep the versions built and published in sync with the other projects..

sillyotter:
See PR #4095

I can provide such, but I have several questions.  The list of things published seems to be a bit disorganized and mostly a list of what people remembered to put in but its hard to see order here

For instance, in the opensslrpmrepo section, I see sles, centos, suse, rhel, and fedora listed.  I see fedora 32, 33, 34, and 37.  Why no 35 and 36?  And why no 38?  I assume this is for old ssl 2.X packages, and maybe 37 and below are all that use that, but that doesnt explain the missing 35 and 36,

In the openssl3rpmrepo section, I see fedora 36, and 38.  But no 37.  That ruins my guess that the distribution of packages was based on ssl version, since 37 is in the old ssl section, but 36 is in the new section.

In any event, I believe the change I made will deal with the explicit issue I raised above, no support for 39, but I dont see much rhyme or reason to what gets up where, why some are where they are, nor how one picks if it is ssl or ssl3?

nibanks:
As you have see, we haven't done a great job of keeping that up to date. It's mostly been adhoc as .NET asks us to add a new one. Much of their asks come from others' complaints/asks. We don't have a ton of expertise in what Linux distros are out there and should be supported. It'd be great to get some community support in this area.

nibanks:
Oh, on the topic of "ssl or ssl3" we need to leverage whatever the inbox version of OpenSSL is (1.1 or 3.*), so we have to build for the right version depending on the distro/version.

alexrp:
[...]

PR #4095, opened 2024-01-29
Add Fedora 39 to OneBranch.Publish.yml

sillyotter:
## Description

Add Fedora 39 to the list of rpm packages that get published

## Testing

I do not believe I have the ability to force a publish, so I doubt I can test this.

## Documentation

Most documentation already says dotnet core and msquic support fedora 37+, so I do not think anything needs updated.

### DIFF
--- a/.azure/OneBranch.Publish.yml
+++ b/.azure/OneBranch.Publish.yml
@@ -60,6 +60,7 @@ parameters:
   default:
   - microsoft-fedora36-prod-yum
   - microsoft-fedora38-prod-yum
+  - microsoft-fedora39-prod-yum
   - microsoft-rhel9.0-prod-yum
 - name: debug # debug mode will not actually upload and publish packages
   type: boolean


nibanks:
@csujedihy can you please verify this? We'll also want to take this for release/2.3 most likely.

sillyotter:
This is in reference to Issue #4092 raised earlier.  There were questions raised there about the wider question of what is in this file and how things get chosen to be built with each parameter group, why some historical things are where they are, etc.  I think this PR deals with the explicit issue I raised, but I am wondering about other issues about what gets published and when.  Perhaps they can be discussed in the issue or a new conversation can be started.

Thanks

csujedihy (approved):

csujedihy:
> @csujedihy can you please verify this? We'll also want to take this for release/2.3 most likely.

LGTM. The package for fedora 39 will be published when 2.3.1 is released.

PR #4115, opened 2024-02-06
Fix Possible DISPATCH_SEND Inline Issue

nibanks:
## Description

We've seen indication of a possible issue when sending at DISPATCH_LEVEL, so we're fixing the behavior to always queue in these cases (no inline execution).

## Testing

Updated some tests to send at DISPATCH_LEVEL in kernel.

## Documentation

N/A

### DIFF
--- a/src/core/api.c
+++ b/src/core/api.c
@@ -1082,10 +1082,13 @@ MsQuicStreamSend(
     SendRequest->TotalLength = TotalLength;
     SendRequest->ClientContext = ClientSendContext;
 
+#pragma warning(push)
+#pragma warning(disable:6240) // CXPLAT_AT_DISPATCH only really does anything for kernel mode
     SendInline =
         !Connection->Settings.SendBufferingEnabled &&
         !CXPLAT_AT_DISPATCH() && // Never run inline if at DISPATCH
         Connection->WorkerThreadID == CxPlatCurThreadID();
+#pragma warning(pop)
 
     CxPlatDispatchLockAcquire(&Stream->ApiSendRequestLock);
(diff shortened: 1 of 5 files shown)


src/core/api.c
```diff
@@ -1083,8 +1083,11 @@ MsQuicStreamSend(
     SendRequest->ClientContext = ClientSendContext;
 
     SendInline =
-        !Connection->Settings.SendBufferingEnabled &&
-        Connection->WorkerThreadID == CxPlatCurThreadID();
+        !Connection->Settings.SendBufferingEnabled
+#ifdef CXPLAT_AT_DISPATCH
```
csujedihy, line 1087: why not have a dummy macro for other platforms?

nibanks: I can, but this required fewer files to be changed. That's why I did this instead.

csujedihy: I think only 2 headers needed to be modified. We should be consistent with what we did for CXPLAT_PASSIVE_CODE and CXPLAT_IRQL.

csujedihy:
We have the same inline execution for other APIs like StreamClose and ConnectionClose. Should we change them as well?

src/core/api.c
nibanks: Fixed

nibanks:
> We have the same inline execution for other APIs like StreamClose and ConnectionClose. Should we change them as well?

No, because *Close functions are required to be called at PASSIVE_LEVEL.

csujedihy (approved):

csujedihy (approved):

PR #4118, opened 2024-02-07
use cred to make remote pwsh session

ProjectsByJackHe:
## Description

WS2025 exhibits weird behavior when starting remote PWSH sessions. This is a workaround.

## Testing

CI

## Documentation

N/A

### DIFF
--- a/scripts/secnetperf.ps1
+++ b/scripts/secnetperf.ps1
@@ -98,7 +98,14 @@ $useXDP = ($io -eq "xdp" -or $io -eq "qtip")
 # Set up the connection to the peer over remote powershell.
 Write-Host "Connecting to $RemoteName"
 if ($isWindows) {
-    $Session = New-PSSession -ComputerName $RemoteName -ConfigurationName PowerShell.7
+    if ($os -eq "windows-2025") {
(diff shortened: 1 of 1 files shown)


scripts/secnetperf.ps1
```diff
@@ -98,7 +98,14 @@ $useXDP = ($io -eq "xdp" -or $io -eq "qtip")
 # Set up the connection to the peer over remote powershell.
 Write-Host "Connecting to $RemoteName"
 if ($isWindows) {
-    $Session = New-PSSession -ComputerName $RemoteName -ConfigurationName PowerShell.7
+    if ($os -eq "windows-2025") {
```
nibanks, line 101: Should we simply just _always_ do this? If it works for 2022, I think I'd rather have the simpler script.

ProjectsByJackHe: lets try and see if this works for WS 2022

ProjectsByJackHe: it looks like it works. Lab failures are unrelated, currently doing some tweaking to diagnose poor performance.

nibanks (approved):

PR #4120, opened 2024-02-08
Add QUIC_PERF_COUNTER_CONN_LOAD_REJECT counter.

rzikm:
## Description

This PR adds a counter which reports the number of connections which were automatically rejected on MsQuic layer because the the worker is already overloaded and is barely able to service current connections.

## Testing

Tested by running .NET System.Net.Quic tests suite in parallel which exhibited this problem in the past.

## Documentation

Documentation was added.


nibanks:
I'm not opposed to this, but I'm curious why the Listener statistics isn't enough, since this is all in-proc for .NET.

nibanks:
Looks like we're getting a compile error:
```
D:\a\msquic\msquic\src\core\library.c(268,1): warning C4310: cast truncates constant value [D:\a\msquic\msquic\src\core\core.kernel.vcxproj]
```
This is coming from:
```c
    QuicTraceEvent(
        PerfCountersRundown,
        "[ lib] Perf counters Rundown, Counters=%!CID!",
        CASTED_CLOG_BYTEARRAY(sizeof(PerfCounterSamples), PerfCounterSamples));
```
Which uses:
```c
#define CLOG_BYTEARRAY(Len, Data) (uint8_t)(Len), (uint8_t*)(Data)
#define CASTED_CLOG_BYTEARRAY(Len, Data) CLOG_BYTEARRAY((unsigned char)(Len), (const unsigned char*)(Data))
```
The issue is that you just pushed past 256 bytes. I need to look at what needs to be changed to address this.

nibanks:
Ok, luckily the ETW event actually takes a 16-bit in, not 8-bit, so it's just a few macros we need to update I believe. Let's define
```c
#define CLOG_BYTEARRAY16(Len, Data) (uint16_t)(Len), (uint8_t*)(Data)
#define CASTED_CLOG_BYTEARRAY16(Len, Data) CLOG_BYTEARRAY16((unsigned short)(Len), (const unsigned char*)(Data))
```
And use that in the two places you're getting errors.

**Note** - You will also need to define the version for all the other platforms. I think you can just copy/paste and make a similar update.

nibanks:
Looks like you need to update the CLOG files too now.

src/inc/quic_trace.h
```diff
@@ -142,6 +142,7 @@ extern QUIC_TRACE_RUNDOWN_CALLBACK* QuicTraceRundownCallback;
 #endif
 
 #define CASTED_CLOG_BYTEARRAY(Len, Data) CLOG_BYTEARRAY((unsigned char)(Len), (const unsigned char*)(Data))
+#define CASTED_CLOG_BYTEARRAY16(Len, Data) CLOG_BYTEARRAY16((unsigned short)(Len), (const unsigned char*)(Data))
```
nibanks, line 145: CLOG is still throwing some build errors. Try this instead?
```suggestion
#define CASTED_CLOG_BYTEARRAY16(Len, Data) CLOG_BYTEARRAY((unsigned short)(Len), (const unsigned char*)(Data))
```

nibanks (approved):

PR #4121, opened 2024-02-08
Fix Cancel On Loss Test's Race Condition

nibanks:
## Description

Fixes a race condition in the test code where it set a completion event before the return code, so the main thread that waits on the event and validates the code some times fails.

Also fixed a number of style nits.

## Testing

Locally. CI/CD

## Documentation

N/A

### DIFF
--- a/src/test/lib/DataTest.cpp
+++ b/src/test/lib/DataTest.cpp
@@ -1398,7 +1398,6 @@ struct CancelOnLossContext
     CxPlatEvent SendPhaseEndedEvent = {};
 };
 
-
 _Function_class_(MsQuicStreamCallback)
 QUIC_STATUS
 QuicCancelOnLossStreamHandler(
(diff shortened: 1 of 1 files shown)


src/test/lib/DataTest.cpp
```diff
@@ -1407,27 +1406,18 @@ QuicCancelOnLossStreamHandler(
...
     auto TestContext = reinterpret_cast<CancelOnLossContext*>(Context);
-
-    QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
-
     switch (Event->Type) {
     case QUIC_STREAM_EVENT_RECEIVE:
         if (TestContext->IsServer) { // only server receives
-            TestContext->SendPhaseEndedEvent.Set();
             TestContext->ExitCode = CancelOnLossContext::SuccessExitCode;
+            TestContext->SendPhaseEndedEvent.Set();
```
nibanks, line 1414: This is the real fix.

csujedihy (approved):

PR #4129, opened 2024-02-12
Fix overflow in winkernel CxPlatEventWaitWithTimeout

nigriMSFT:
## Description
When UINT32_MAX is passed as a timeout value to CxPlatEventWaitWithTimeout on the winkernel platform, an arithmetic overflow happens resulting in a positive timeout value and the wait timing out immediately. The winuser and posix platform implementations of CxPlatEventWaitWithTimeout appear to handle this case gracefully, even though callers should really be using CxPlatEventWaitForever for the infinite wait case. Fixing winkernel to handle this gracefully to eliminate this developer trap.

Also adding asserts to catch cases where the developer is calling the CxPlatEventWaitWithTimeout API incorrectly.

## Testing
Local testing w/ debugger confirmation of overflow and the fix.

## Documentation
N/A

### DIFF
--- a/src/inc/quic_platform_winuser.h
+++ b/src/inc/quic_platform_winuser.h
@@ -673,7 +673,7 @@ typedef HANDLE CXPLAT_EVENT;
 #define CxPlatEventWaitForever(Event) WaitForSingleObject(Event, INFINITE)
 inline
 BOOLEAN
-_CxPlatEventWaitWithTimeout(
+CxPlatEventWaitWithTimeout(
     _In_ CXPLAT_EVENT Event,
     _In_ uint32_t TimeoutMs
     )
@@ -681,7 +681,6 @@ _CxPlatEventWaitWithTimeout(
     CXPLAT_DBG_ASSERT(TimeoutMs != UINT32_MAX);
     return WAIT_OBJECT_0 == WaitForSingleObject(Event, TimeoutMs);
 }
-#define CxPlatEventWaitWithTimeout(Event, TimeoutMs) _CxPlatEventWaitWithTimeout(Event, TimeoutMs)
 
 //
 // Event Queue Interfaces
--- a/src/inc/quic_platform_winuser.h
+++ b/src/inc/quic_platform_winuser.h
@@ -671,8 +671,17 @@ typedef HANDLE CXPLAT_EVENT;
 #define CxPlatEventSet(Event) SetEvent(Event)
 #define CxPlatEventReset(Event) ResetEvent(Event)
 #define CxPlatEventWaitForever(Event) WaitForSingleObject(Event, INFINITE)
-#define CxPlatEventWaitWithTimeout(Event, timeoutMs) \
-    (WAIT_OBJECT_0 == WaitForSingleObject(Event, timeoutMs))
+inline
+BOOLEAN
+_CxPlatEventWaitWithTimeout(
+    _In_ CXPLAT_EVENT Event,
+    _In_ uint32_t TimeoutMs
+    )
+{
+    CXPLAT_DBG_ASSERT(TimeoutMs != UINT32_MAX);
+    return WAIT_OBJECT_0 == WaitForSingleObject(Event, TimeoutMs);
+}
+#define CxPlatEventWaitWithTimeout(Event, TimeoutMs) _CxPlatEventWaitWithTimeout(Event, TimeoutMs)
 
 //
 // Event Queue Interfaces
--- a/src/inc/quic_platform_posix.h
+++ b/src/inc/quic_platform_posix.h
@@ -870,6 +870,8 @@ CxPlatInternalEventWaitWithTimeout(
     struct timespec Ts = {0, 0};
     int Result;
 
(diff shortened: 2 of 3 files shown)


nibanks (approved):

mtfriesen (dismissed):

nigriMSFT:
> Should we assert the timeout != INFINITE? It'll take ~46 days, but all of a sudden waits will be satisfied without their wait condition being true.
> 
> Of course, 0xFFFFFFFF may also be a valid timeout, depending on how MsQuic defines this routine.

It seems like 0xFFFFFFFF should be a valid timeout and not special based on the given APIs (CxPlatEventWaitWithTimeout and CxPlatEventWaitForever) and lack of documentation for CxPlatEvent*. I'd be happy to iron out the contract here as the behavior does not seem consistent across platforms. So I guess I'd propose not adding an assert and instead add code to winuser CxPlatEventWaitWithTimeout:
```
inline
BOOLEAN
_CxPlatEventWaitWithTimeout(
    _In_ CXPLAT_EVENT* Event,
    _In_ uint32_t TimeoutMs
    )
{
    if (TimeoutMs == INFINITE) {
        --Timeout;
    }
    return WAIT_OBJECT_0 == WaitForSingleObject(Event, TimeoutMs);
}
#define CxPlatEventWaitWithTimeout(Event, TimeoutMs) _CxPlatEventWaitWithTimeout(&Event, TimeoutMs)
```

mtfriesen:
> It seems like 0xFFFFFFFF should be a valid timeout and not special based on the given APIs (CxPlatEventWaitWithTimeout and CxPlatEventWaitForever) and lack of documentation for CxPlatEvent*. I'd be happy to iron out the contract here as the behavior does not seem consistent across platforms.

Yeah, exactly, the API can say 0xFFFFFFFF means 46 days or it means forever, but it shouldn't be platform dependent. I'd suggest 0xFFFFFFFF means 46 days given the separate "forever" routine, but maybe I'd "accelerate" the 46 day wait in stress tests somehow.

nibanks:
I believe all code up until now differentiates waiting forever vs a specific timeout by calling separate functions. I would prefer to keep it that way unless there is good reason. If we do go that way, then we should add the assert.

nibanks:
For instance, some code I just copied from worker.c:
```c
        BOOLEAN Ready = InterlockedFetchAndClearBoolean(&EC->Ready);
        if (!Ready) {
            if (EC->NextTimeUs == UINT64_MAX) {
                CxPlatEventWaitForever(Worker->Ready);

            } else if (EC->NextTimeUs > State.TimeNow) {
                uint64_t Delay = US_TO_MS(EC->NextTimeUs - State.TimeNow) + 1;
                if (Delay >= (uint64_t)UINT32_MAX) {
                    Delay = UINT32_MAX - 1; // Max has special meaning for most platforms.
                }
                CxPlatEventWaitWithTimeout(Worker->Ready, (uint32_t)Delay);
            }
        }
```

nigriMSFT:
> I believe all code up until now differentiates waiting forever vs a specific timeout by calling separate functions. I would prefer to keep it that way unless there is good reason. If we do go that way, then we should add the assert.

From experience, it is easy to end up calling the CxPlatWaitWithTimeout(INFINITE) either accidentally or on purpose.

I don't mind which contract we choose, but I want to improve the developer experience so others don't spend time debugging inconsistent behavior between platforms.

nibanks:
I prefer to keep the split. That's why we have two separate functions to begin with.

nigriMSFT:
Ok, if I am interpreting correctly, calling WaitWithTimeout(0xFFFFFFFF) should not happen. I can add an assert to all WaitWithTimeout implementations to help devs catch this

src/inc/quic_platform_winuser.h
```diff
@@ -671,8 +671,17 @@ typedef HANDLE CXPLAT_EVENT;
...
+BOOLEAN
+_CxPlatEventWaitWithTimeout(
+    _In_ CXPLAT_EVENT Event,
+    _In_ uint32_t TimeoutMs
+    )
+{
+    CXPLAT_DBG_ASSERT(TimeoutMs != UINT32_MAX);
+    return WAIT_OBJECT_0 == WaitForSingleObject(Event, TimeoutMs);
+}
+#define CxPlatEventWaitWithTimeout(Event, TimeoutMs) _CxPlatEventWaitWithTimeout(Event, TimeoutMs)
```
nibanks, line 684: Why have this macro at all? Just use an inline function?

nigriMSFT: Was following the CxPlatEvent patterns in quic_platform_winkernel.h and quic_platform_winuser.h?

nigriMSFT: Ok, will update. Also it appears we are hitting this new assert in functional tests for both winkernel and winuser platforms. I might end up reverting the assert addition (and maybe open an issue to investigate this incorrect API usage?)

nibanks (approved):

PR #4135, opened 2024-02-14
Add a Threat Model doc

nibanks:
## Description

Documentation of the threat model for MsQuic.

## Testing

N/A

## Documentation

This is new documentation.

### DIFF
--- a/docs/ThreatModel.md
+++ b/docs/ThreatModel.md
@@ -1,0 +1,32 @@
+# Threat Model
+
+## Overview
+
+MsQuic is a library that provides a set of APIs for applications to use for networking based on the QUIC protocol. It operates in the same process as the application and uses UDP sockets and TLS for communication.
+
+## Threat Surface
+
+### Bottom Edge (UDP sockets and TLS)
+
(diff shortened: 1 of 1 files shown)


ProjectsByJackHe (approved):
looks good.

qzhuyan:
maybe the dynamic linked libraries should also be mentioned in this doc?

qzhuyan:
> > maybe the dynamic linked libraries should also be mentioned in this doc?
> 
> Are you asking if we should specifically list out the libraries msquic.dll / libmsquic.so depends on?

Not sure if it is in this scope,
Maybe the reader need to be aware of which library the msquic links to, just like a reminder.

nibanks:
> > > maybe the dynamic linked libraries should also be mentioned in this doc?
> > 
> > 
> > Are you asking if we should specifically list out the libraries msquic.dll / libmsquic.so depends on?
> 
> Not sure if it is in this scope, Maybe the reader need to be aware of which library the msquic links to, just like a reminder.

I'm not opposed to listing out dependencies (though probably in a follow up PR). I just wanted to confirm I understood your comment.

nibanks:
We can follow up with a dependency list.

PR #4145, opened 2024-02-19
Don't send client CONNECTION_CLOSE on 1-RTT before handshake completion.

### LINKED ISSUE #4132: Certificate rejection via ConnectionCertificateValidationComplete leads to connection timeout.
### Describe the bug

I am offloading computation-heavy certificate validation to a non-MsQuic thread in System.Net.Quic, and a unit test where the cert is rejected keeps failing.

Wireshark capture of when the `ConnectionCertificateValidationComplete` function is invoked from inside the `PEER_CERTIFICATE_RECEIVED` handler:

image

Notably, the Connection Close frame is sent at the Handshake protection level. This works fine.

When the invocation is moved to be done later from another thread:

image

Note that this time, the Connection Close is sent at 1RTT level, which seems to get ignored by the server. The connection attempt times out and get's closed by the server (the last line). The capture is attached below, together with sslkeylogfile to decrypt it.

[quic-cert-validation.zip](https://github.com/microsoft/msquic/files/14270310/quic-cert-validation.zip)

### Affected OS

- [x] Windows
- [ ] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

Reproduced on Windows, didn't try other OSes

### MsQuic version

main

Use `ConnectionCertificateValidationComplete` to reject certificate from another thread.

### Expected behavior

The server-side connection receives TLS alert which was sent by the client.

### Actual outcome

The connection attempt times out

### Additional details

_No response_

rzikm:
## Description

Fixes #4132.

## Testing

Fixed tests so that they would detect the bug before the fix.

## Documentation

No.

### DIFF
--- a/src/test/lib/HandshakeTest.cpp
+++ b/src/test/lib/HandshakeTest.cpp
@@ -820,6 +820,14 @@ QuicTestConnectAndIdle(
     }
 }
 
+void hexdump(FILE* f, void *ptr, int buflen) {
+  unsigned char *buf = (unsigned char*)ptr;
+  int i;
+  for (i=0; i<buflen; i+=1) {
+    fprintf(f, "%02x", buf[i]);
+  }
+}
+
 void
 QuicTestCustomServerCertificateValidation(
     _In_ bool AcceptCert,
@@ -857,6 +865,7 @@ QuicTestCustomServerCertificateValidation(
             }
             Listener.Context = &ServerAcceptCtx;
 
+            QUIC_TLS_SECRETS TlsSecrets{};
             {
                 TestConnection Client(Registration);
                 TEST_TRUE(Client.IsValid());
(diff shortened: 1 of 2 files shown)


src/test/lib/HandshakeTest.cpp
```diff
@@ -885,13 +894,11 @@ QuicTestCustomServerCertificateValidation(
                 }
                 TEST_EQUAL(AcceptCert, Client.GetIsConnected());
 
-                if (AcceptCert) { // Server will be deleted on reject case, so can't validate.
```
rzikm, line 888: I don't see how the server would get deleted. The new connection instance is held by the unique_ptr above and the AutoDelete bit is not set, so the connection should be valid until the unique_ptr goes out of scope.

nibanks: I believe this was meant to handle a case where the listener could reject the connection, and then this pointer would never be set. That being said, I cannot see a path where that would happen in this test case (or the one below). And since the tests seem to pass just fine, I'm ok with these changes.

nibanks (approved):

rzikm:
> @rzikim would it be possible to add (or modify) a test to validate #4132 is actually fixed?

That is handled by the tests modified in this PR, without the fix in packet_builder.c, one of the async/reject case would've failed. This was not detected before because the test did not wait for server shutdown notification. (which would otherwise timeout)

nibanks:
> > @rzikim would it be possible to add (or modify) a test to validate #4132 is actually fixed?
> 
> That is handled by the tests modified in this PR, without the fix in packet_builder.c, the async/reject case would've failed. This was not detected before because the test did not wait for server shutdown notification which would've timed out.

Ah! I see. Ok, then let's ship it!

PR #4148, opened 2024-02-21
Unify data storage

ProjectsByJackHe:
## Description

This PR attempts to unify everything data-storage related.

It will accommodate the next stage in netperf development where we have dynamically changing test environments.

In addition, this PR will persist the full latency txt curves, and netperf will handle parsing/condensing/saving it. 

## Testing

CI

## Documentation

N/A

### DIFF
--- a/scripts/secnetperf.ps1
+++ b/scripts/secnetperf.ps1
@@ -80,7 +80,7 @@ $RemoteDir = "C:/_work/quic"
 $localDir = "C:/actions-runner/_work/netperf/netperf"
 if (!$isWindows) {
     $RemoteDir = "/home/secnetperf/_work/quic"
-    $localDir = "/home/secnetperf/_work/netperf/netperf"
+    $localDir = "/home/secnetperf/actions-runner/_work/netperf/netperf"
 }
 $SecNetPerfDir = "artifacts/bin/$plat/$($arch)_Release_$tls"
 $SecNetPerfPath = "$SecNetPerfDir/secnetperf"
--- a/scripts/secnetperf.ps1
+++ b/scripts/secnetperf.ps1
@@ -80,7 +80,7 @@ $RemoteDir = "C:/_work/quic"
 $localDir = "C:/actions-runner/_work/netperf/netperf"
 if (!$isWindows) {
     $RemoteDir = "/home/secnetperf/_work/quic"
-    if ($env -eq "lab") {
+    if ($environment -eq "lab") {
         $localDir = "/home/secnetperf/_work/netperf/netperf"
     } else {
         $localDir = "/home/secnetperf/actions-runner/_work/netperf/netperf"
--- a/scripts/secnetperf.ps1
+++ b/scripts/secnetperf.ps1
@@ -77,8 +77,10 @@ $PSDefaultParameterValues["*:ErrorAction"] = "Stop"
 
 # Set up some important paths.
 $RemoteDir = "C:/_work/quic"
+$localDir = "C:/actions-runner/_work/netperf/netperf"
 if (!$isWindows) {
     $RemoteDir = "/home/secnetperf/_work/quic"
+    $localDir = "/home/secnetperf/_work/netperf/netperf"
 }
 $SecNetPerfDir = "artifacts/bin/$plat/$($arch)_Release_$tls"
 $SecNetPerfPath = "$SecNetPerfDir/secnetperf"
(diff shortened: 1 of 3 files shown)


scripts/secnetperf.ps1
```diff
@@ -184,13 +184,27 @@ VALUES ("$MsQuicCommit", "$(Get-Date -Format "yyyy-MM-dd HH:mm:ss")", 1, "TODO")
...
 
+# Create file to persist full latency curve data:
+New-Item -ItemType File -Name "full_latency_curve.txt"
+
 # > All tests:
 $allTests["tput-up"] = "-exec:maxtput -up:12s -ptput:1"
 $allTests["tput-down"] = "-exec:maxtput -down:12s -ptput:1"
 $allTests["hps-conns-100"] = "-exec:maxtput -rconn:1 -share:1 -conns:100 -run:12s -prate:1"
-$allTests["rps-up-512-down-4000"] = "-exec:lowlat -rstream:1 -up:512 -down:4000 -run:20s -plat:1"
+$allTests["rps-up-512-down-4000"] = "-exec:lowlat -rstream:1 -up:512 -down:4000 -run:20s -plat:1 -extraOutputFile:full_latency_curve.txt"
```
nibanks, line 207: I think you need to specify a file path into the logs directory, so this gets uploaded?

scripts/secnetperf.ps1
```diff
@@ -184,13 +184,27 @@ VALUES ("$MsQuicCommit", "$(Get-Date -Format "yyyy-MM-dd HH:mm:ss")", 1, "TODO")
...
 
+# Create file to persist full latency curve data:
+New-Item -ItemType File -Name "full-latency-curve.txt"
+
 # > All tests:
 $allTests["tput-up"] = "-exec:maxtput -up:12s -ptput:1"
 $allTests["tput-down"] = "-exec:maxtput -down:12s -ptput:1"
 $allTests["hps-conns-100"] = "-exec:maxtput -rconn:1 -share:1 -conns:100 -run:12s -prate:1"
-$allTests["rps-up-512-down-4000"] = "-exec:lowlat -rstream:1 -up:512 -down:4000 -run:20s -plat:1"
+$allTests["rps-up-512-down-4000"] = "-exec:lowlat -rstream:1 -up:512 -down:4000 -run:20s -plat:1 -extraOutputFile:full-latency-curve.txt"
```
nibanks, line 207: 1. I think we should fully specify path.
2. Let's not put this here, but instead, whenever we execute an RPS test (in the helper script) always append this argument to the client (not server).
3. Let's name this file '<test-name>-latency.txt`.
4. We are going to want to post-process this file to a nicer format for automation to use.  We can either do it here, or later. IMHO, doing it here, so we only ever upload the nicer format is the best option.

ProjectsByJackHe: For post-processing to a more condensed format, my thoughts were to merge all the latency.txt files produced from all the separate environment runs into 1 file, and commit that to some orphan branch with the commit name. This step would happen after all the runs for the different environments have completed. Thoughts?

nibanks: I'm fine with merging into a single file, but I don't want to go creating tons or branches. I'd rather a folder with the commit name in a single branch.

ProjectsByJackHe:
Whoops! accidently added some untracked files. Will revert that.

ProjectsByJackHe:
https://github.com/microsoft/netperf/actions/runs/8026065439/job/21929654888 most recent run, where all the latency curves are successfully produced / uploaded.

src/perf/bin/appmain.cpp
```diff
@@ -74,6 +74,7 @@ QuicHandleExtraData(
 #endif
         if (FileErr) {
             printf("Failed to open file '%s' for write\n", FileName);
+            printf("Error code: %d\n", FileErr);
```
nibanks, line 77:
```suggestion
            printf("Failed to open file '%s' for write, error: %d\n", FileName, FileErr);
```

scripts/secnetperf.ps1
```diff
@@ -114,6 +114,9 @@ if ($null -eq $Session) {
 # Make sure nothing is running from a previous run.
 Cleanup-State $Session $RemoteDir
 
+# Create intermediary files.
+New-Item -ItemType File -Name "latency.txt"
```
nibanks, line 118: I think this is unnecessary now, but we can remove in a future PR.

nibanks (approved):

PR #4164, opened 2024-02-29
[release/2.3] Don't send client CONNECTION_CLOSE on 1-RTT before handshake completion. (#4145)

rzikm:
Ports #4145 to release/2.3


rzikm:
We should probably hold on this before merging, the fix I did may not be enough in some cases.

rzikm:
False alarm, we hit a slightly different bug: https://github.com/microsoft/msquic/issues/4166 which can be fixed separately.

nibanks (approved):

PR #4169, opened 2024-03-01
Convert application close frame to transport close frame during handshake.

### LINKED ISSUE #4166: ConnectionShutdown with pending (async) CertificateValidation sends Application CONNECTION_CLOSE
### Describe the bug

When `QUIC_STATUS_PENDING` is returned when handling the QUIC_CONNECTION_EVENT_PEER_CERTIFICATE_RECEIVED event and application layer decides to shutdown the connection, MsQuic will end up sending the application-level CONNECTION_CLOSE frame. However, since the handshake is not complete at this point, the peer may not process this frame and may not reply to it, which makes server wait until the handshake timeouts.

Note that the above *may* happen even if application calls `ConnectionCertificateValidationComplete` before `ConnectionShutdown`, because `ConnectionShutdown` is scheduled to the front of the operation queue and will be processed first.

### Affected OS

- [X] Windows
- [X] Linux
- [X] macOS
- [ ] Other (specify below)

### Additional OS information

_No response_

### MsQuic version

main

See description.

Quick repro can be achieved in tests by following changes:

https://github.com/microsoft/msquic/blob/3ef19310a3690604ceb62162aacf523b433b2bde/src/test/lib/HandshakeTest.cpp#L887-L894

```cpp
                if (AsyncValidation) {
                    CxPlatSleep(1000);
                    // TEST_QUIC_SUCCEEDED(Client.SetCustomValidationResult(AcceptCert));
                    Client.Shutdown(QUIC_CONNECTION_SHUTDOWN_FLAG_NONE, QUIC_TEST_NO_ERROR);
                }

                // if (!Client.WaitForConnectionComplete()) {
                if (!Client.WaitForShutdownComplete()) {
                    return;
                }
```

and running the `Handshake/WithHandshakeArgs5.CustomServerCertificateValidation/1` test case.

### Expected behavior

Ideally, if `ConnectionCertificateValidationComplete` was called first (with Result=FALSE), I would expect the CONNECTION_CLOSE sent by MsQuic to reflect that.

At the very least, MsQuic should not attempt sending application CONNECTION_CLOSE before the handshake ends.

### Actual outcome

Server hangs until configured handshake timeout.

### Additional details

Possible fix lies in tweaking the condition in

https://github.com/microsoft/msquic/blob/f62a363f74079cbc5a875c2580b8052e0beab908/src/core/connection.c#L1443-L1462

Similarly to https://github.com/microsoft/msquic/issues/4132, the condition `Connection->Crypto.TlsState.WriteKey < QUIC_PACKET_KEY_1_RTT` may be false while still in handshake and server is not allowed to process such 1RTT packets.

However, this suggestion does not fix the "race" when `ConnectionCertificateValidationComplete` and `ConnectionShutdown` are called in quick succession.

rzikm:
Fixes #4166.

## Description

This PR ensures we convert application CONNECTION_CLOSE to transport CONNECTION_CLOSE when the handshake is not confirmed yet (and the peer may not be able to process the frame yet).

## Testing

New tests added.

## Documentation

No change needed.

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -5136,7 +5136,7 @@ QuicConnRecvFrames(
                     HandshakeConfirmedFrame,
                     Connection,
                     "Handshake confirmed (frame)");
-                QuicCryptoHandshakeConfirmed(&Connection->Crypto);
+                QuicCryptoHandshakeConfirmed(&Connection->Crypto, TRUE);
             }
 
             AckEliciting = TRUE;
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -1451,13 +1451,15 @@ QuicConnTryClose(
     }
 
     if (!ClosedRemotely) {
-
         if ((Flags & QUIC_CLOSE_APPLICATION) &&
-            Connection->Crypto.TlsState.WriteKey < QUIC_PACKET_KEY_1_RTT) {
+            !Connection->State.HandshakeConfirmed) {
             //
             // Application close can only happen if we are using 1-RTT keys.
-            // Otherwise we have to send "user_canceled" TLS error code as a
-            // connection close. Overwrite all application provided parameters.
+            // But until the handshake is confirmed, we can't be sure if the
+            // peer is able to process the close frame.
+            //
+            // Send "user_canceled" TLS error code as a connection close instead.
+            // Overwrite all application provided parameters.
             //
             Flags &= ~QUIC_CLOSE_APPLICATION;
             ErrorCode = QUIC_ERROR_CRYPTO_USER_CANCELED;
@@ -5628,9 +5630,7 @@ QuicConnRecvDatagrams(
                     Packet->BufferLength - (uint16_t)(Packet->AvailBuffer - Packet->Buffer);
             }
 
-            if (Connection->Crypto.CertValidationPending ||
-                Connection->Crypto.TicketValidationPending ||
-                !QuicConnRecvHeader(
+            if (!QuicConnRecvHeader(
                     Connection,
                     Packet,
                     Cipher + BatchCount * CXPLAT_HP_SAMPLE_LENGTH)) {
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -1452,12 +1452,12 @@ QuicConnTryClose(
 
     if (!ClosedRemotely) {
         if ((Flags & QUIC_CLOSE_APPLICATION) &&
-            QuicConnIsClient(Connection) &&
-            !Connection->State.HandshakeConfirmed) {
+            (Connection->Crypto.TlsState.WriteKey < QUIC_PACKET_KEY_1_RTT ||
+            (QuicConnIsClient(Connection) && !Connection->State.HandshakeConfirmed))) {
             //
             // Application close can only happen if we are using 1-RTT keys.
-            // But until the handshake is confirmed, we can't be sure if the
-            // peer is able to process the close frame.
+            // In case of Clients, we may have 1-RTT write keys earlier but
+            // server may not process 1-RTT packets until the handshake is confirmed.
             //
             // Send "user_canceled" TLS error code as a connection close instead.
             // Overwrite all application provided parameters.
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -3697,6 +3697,15 @@ QuicConnGetKeyOrDeferDatagram(
         return FALSE;
     }
 
+    if (QuicConnIsServer(Connection) && !Connection->State.HandshakeConfirmed &&
+        Packet->KeyType == QUIC_PACKET_KEY_1_RTT) {
+        //
+        // A server MUST NOT process incoming 1-RTT protected packets before the TLS
+        // handshake is complete.
+        //
+        return FALSE;
+    }
+
     _Analysis_assume_(Packet->KeyType >= 0 && Packet->KeyType < QUIC_PACKET_KEY_COUNT);
     if (Connection->Crypto.TlsState.ReadKeys[Packet->KeyType] == NULL) {
         //
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -1452,6 +1452,7 @@ QuicConnTryClose(
 
     if (!ClosedRemotely) {
         if ((Flags & QUIC_CLOSE_APPLICATION) &&
+            QuicConnIsClient(Connection) &&
             !Connection->State.HandshakeConfirmed) {
             //
             // Application close can only happen if we are using 1-RTT keys.
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -1390,6 +1390,7 @@ QuicErrorCodeToStatus(
     case QUIC_ERROR_NO_ERROR:                       return QUIC_STATUS_SUCCESS;
     case QUIC_ERROR_CONNECTION_REFUSED:             return QUIC_STATUS_CONNECTION_REFUSED;
     case QUIC_ERROR_PROTOCOL_VIOLATION:             return QUIC_STATUS_PROTOCOL_ERROR;
+    case QUIC_ERROR_APPLICATION_ERROR:           
     case QUIC_ERROR_CRYPTO_USER_CANCELED:           return QUIC_STATUS_USER_CANCELED;
     case QUIC_ERROR_CRYPTO_HANDSHAKE_FAILURE:       return QUIC_STATUS_HANDSHAKE_FAILURE;
     case QUIC_ERROR_CRYPTO_NO_APPLICATION_PROTOCOL: return QUIC_STATUS_ALPN_NEG_FAILURE;
@@ -1450,29 +1451,29 @@ QuicConnTryClose(
         }
     }
 
-    if (!ClosedRemotely) {
-        if ((Flags & QUIC_CLOSE_APPLICATION) &&
-            (Connection->Crypto.TlsState.WriteKey < QUIC_PACKET_KEY_1_RTT ||
-            (QuicConnIsClient(Connection) && !Connection->Crypto.TlsState.HandshakeComplete))) {
-            //
-            // Application close can only happen if we are using 1-RTT keys.
-            // In case of Clients, we may have 1-RTT write keys earlier but
-            // server may not process 1-RTT packets until the handshake is complete.
-            //
-            // Send "user_canceled" TLS error code as a connection close instead.
-            // Overwrite all application provided parameters.
-            //
-            Flags &= ~QUIC_CLOSE_APPLICATION;
-            ErrorCode = QUIC_ERROR_CRYPTO_USER_CANCELED;
-            RemoteReasonPhrase = NULL;
-            RemoteReasonPhraseLength = 0;
+    // if (!ClosedRemotely) {
+    //     if ((Flags & QUIC_CLOSE_APPLICATION) &&
+    //         (Connection->Crypto.TlsState.WriteKey < QUIC_PACKET_KEY_1_RTT ||
+    //         (QuicConnIsClient(Connection) && !Connection->Crypto.TlsState.HandshakeComplete))) {
+    //         //
+    //         // Application close can only happen if we are using 1-RTT keys.
+    //         // In case of Clients, we may have 1-RTT write keys earlier but
+    //         // server may not process 1-RTT packets until the handshake is complete.
+    //         //
+    //         // Send "user_canceled" TLS error code as a connection close instead.
+    //         // Overwrite all application provided parameters.
+    //         //
+    //         Flags &= ~QUIC_CLOSE_APPLICATION;
+    //         ErrorCode = QUIC_ERROR_CRYPTO_USER_CANCELED;
+    //         RemoteReasonPhrase = NULL;
+    //         RemoteReasonPhraseLength = 0;
 
-            QuicTraceLogConnInfo(
-                CloseUserCanceled,
-                Connection,
-                "Connection close using user canceled error");
-        }
-    }
+    //         QuicTraceLogConnInfo(
+    //             CloseUserCanceled,
+    //             Connection,
+    //             "Connection close using user canceled error");
+    //     }
+    // }
 
     BOOLEAN ResultQuicStatus = !!(Flags & QUIC_CLOSE_QUIC_STATUS);
 
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -1451,30 +1451,6 @@ QuicConnTryClose(
         }
     }
 
-    // if (!ClosedRemotely) {
-    //     if ((Flags & QUIC_CLOSE_APPLICATION) &&
-    //         (Connection->Crypto.TlsState.WriteKey < QUIC_PACKET_KEY_1_RTT ||
-    //         (QuicConnIsClient(Connection) && !Connection->Crypto.TlsState.HandshakeComplete))) {
-    //         //
-    //         // Application close can only happen if we are using 1-RTT keys.
-    //         // In case of Clients, we may have 1-RTT write keys earlier but
-    //         // server may not process 1-RTT packets until the handshake is complete.
-    //         //
-    //         // Send "user_canceled" TLS error code as a connection close instead.
-    //         // Overwrite all application provided parameters.
-    //         //
-    //         Flags &= ~QUIC_CLOSE_APPLICATION;
-    //         ErrorCode = QUIC_ERROR_CRYPTO_USER_CANCELED;
-    //         RemoteReasonPhrase = NULL;
-    //         RemoteReasonPhraseLength = 0;
-
-    //         QuicTraceLogConnInfo(
-    //             CloseUserCanceled,
-    //             Connection,
-    //             "Connection close using user canceled error");
-    //     }
-    // }
-
     BOOLEAN ResultQuicStatus = !!(Flags & QUIC_CLOSE_QUIC_STATUS);
 
     BOOLEAN IsFirstCloseForConnection = TRUE;
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -1453,11 +1453,11 @@ QuicConnTryClose(
     if (!ClosedRemotely) {
         if ((Flags & QUIC_CLOSE_APPLICATION) &&
             (Connection->Crypto.TlsState.WriteKey < QUIC_PACKET_KEY_1_RTT ||
-            (QuicConnIsClient(Connection) && !Connection->State.HandshakeConfirmed))) {
+            (QuicConnIsClient(Connection) && !Connection->Crypto.TlsState.HandshakeComplete))) {
             //
             // Application close can only happen if we are using 1-RTT keys.
             // In case of Clients, we may have 1-RTT write keys earlier but
-            // server may not process 1-RTT packets until the handshake is confirmed.
+            // server may not process 1-RTT packets until the handshake is complete.
             //
             // Send "user_canceled" TLS error code as a connection close instead.
             // Overwrite all application provided parameters.
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -1390,7 +1390,7 @@ QuicErrorCodeToStatus(
     case QUIC_ERROR_NO_ERROR:                       return QUIC_STATUS_SUCCESS;
     case QUIC_ERROR_CONNECTION_REFUSED:             return QUIC_STATUS_CONNECTION_REFUSED;
     case QUIC_ERROR_PROTOCOL_VIOLATION:             return QUIC_STATUS_PROTOCOL_ERROR;
-    case QUIC_ERROR_APPLICATION_ERROR:           
+    case QUIC_ERROR_APPLICATION_ERROR:
     case QUIC_ERROR_CRYPTO_USER_CANCELED:           return QUIC_STATUS_USER_CANCELED;
     case QUIC_ERROR_CRYPTO_HANDSHAKE_FAILURE:       return QUIC_STATUS_HANDSHAKE_FAILURE;
     case QUIC_ERROR_CRYPTO_NO_APPLICATION_PROTOCOL: return QUIC_STATUS_ALPN_NEG_FAILURE;
@@ -5079,12 +5079,33 @@ QuicConnRecvFrames(
             if (Frame.ApplicationClosed) {
                 Flags |= QUIC_CLOSE_APPLICATION;
             }
-            QuicConnTryClose(
-                Connection,
-                Flags,
-                Frame.ErrorCode,
-                Frame.ReasonPhrase,
-                (uint16_t)Frame.ReasonPhraseLength);
+
+            if (!Frame.ApplicationClosed && Frame.ErrorCode == QUIC_ERROR_APPLICATION_ERROR) {
+                //
+                // The APPLICATION_ERROR transport error should be sent only
+                // when closing the connection before the handshake is
+                // confirmed. In such case, we can also expect peer to send the
+                // application CONNECTION_CLOSE frame in a 1-RTT packet
+                // (presumably also in the same UDP datagram).
+                //
+                // We want to prioritize reporting the application-layer error
+                // code to the application, so we postpone the call to
+                // QuicConnTryClose and check again after processing incoming
+                // datagrams in case it does not arrive.
+                //
+                QuicTraceEvent(
+                    ConnDelayCloseApplicationError,
+                    "[conn][%p] Received APPLICATION_ERROR error, delaying close in expectation of a 1-RTT CONNECTION_CLOSE frame.",
+                    Connection);
+                Connection->State.DelayedApplicationError = TRUE;
+            } else {
+                QuicConnTryClose(
+                    Connection,
+                    Flags,
+                    Frame.ErrorCode,
+                    Frame.ReasonPhrase,
+                    (uint16_t)Frame.ReasonPhraseLength);
+            }
 
             AckEliciting = TRUE;
             Packet->HasNonProbingFrame = TRUE;
@@ -5722,6 +5743,20 @@ QuicConnRecvDatagrams(
             Cipher,
             &RecvState);
         BatchCount = 0; // cppcheck-suppress unreadVariable; NOLINT
+    }
+
+    if (Connection->State.DelayedApplicationError && Connection->CloseStatus == 0) {
+        //
+        // We received transport APPLICATION_ERROR, but didn't receive the expected
+        // CONNECTION_ERROR frame, so close the connection with originally postponed
+        // APPLICATION_ERROR.
+        //
+        QuicConnTryClose(
+            Connection,
+            QUIC_CLOSE_REMOTE | QUIC_CLOSE_SEND_NOTIFICATION,
+            QUIC_ERROR_APPLICATION_ERROR,
+            NULL,
+            (uint16_t)0);
     }
 
     if (RecvState.ResetIdleTimeout) {
--- a/src/core/crypto.c
+++ b/src/core/crypto.c
@@ -1721,8 +1721,10 @@ QuicCryptoCustomCertValidationComplete(
             "Custom cert validation succeeded");
         QuicCryptoProcessDataComplete(Crypto, Crypto->PendingValidationBufferLength);
 
-        // more data may have been received while waiting for user to perform the validation
-        QuicCryptoProcessData(Crypto, FALSE);
+        if (QuicRecvBufferHasUnreadData(&Crypto->RecvBuffer)) {
+            // more data was received while waiting for user to perform the validation
+            QuicCryptoProcessData(Crypto, FALSE);
+        }
     } else {
         QuicTraceEvent(
             ConnError,
@@ -1761,8 +1763,10 @@ QuicCryptoCustomTicketValidationComplete(
         Crypto->TicketValidationPending = FALSE;
         QuicCryptoProcessDataComplete(Crypto, Crypto->PendingValidationBufferLength);
 
-        // more data may have been received while waiting for user to perform the validation
-        QuicCryptoProcessData(Crypto, FALSE);
+        if (QuicRecvBufferHasUnreadData(&Crypto->RecvBuffer)) {
+            // more data was received while waiting for user to perform the validation
+            QuicCryptoProcessData(Crypto, FALSE);
+        }
     } else {
         //
         // Need to rollback status before processing client's initial packet, because outgoing buffer and
@@ -1821,16 +1825,12 @@ QuicCryptoProcessData(
     } else {
         uint64_t BufferOffset;
 
-        BOOLEAN DataAvailable =
-            QuicRecvBufferRead(
-                &Crypto->RecvBuffer,
-                &BufferOffset,
-                &BufferCount,
-                &Buffer);
+        QuicRecvBufferRead(
+            &Crypto->RecvBuffer,
+            &BufferOffset,
+            &BufferCount,
+            &Buffer);
 
-        if (!DataAvailable) {
-            return Status;
-        }
         CXPLAT_DBG_ASSERT(BufferCount == 1);
 
         QUIC_CONNECTION* Connection = QuicCryptoGetConnection(Crypto);
--- a/src/core/crypto.c
+++ b/src/core/crypto.c
@@ -1722,7 +1722,7 @@ QuicCryptoCustomCertValidationComplete(
         QuicCryptoProcessDataComplete(Crypto, Crypto->PendingValidationBufferLength);
 
         if (QuicRecvBufferHasUnreadData(&Crypto->RecvBuffer)) {
-            // more data was received while waiting for user to perform the validation
+            // More data was received while waiting for user to perform the validation.
             QuicCryptoProcessData(Crypto, FALSE);
         }
     } else {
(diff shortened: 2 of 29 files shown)


src/core/connection.c
```diff
@@ -5618,9 +5620,7 @@ QuicConnRecvDatagrams(
                     Packet->BufferLength - (uint16_t)(Packet->AvailBuffer - Packet->Buffer);
             }
 
-            if (Connection->Crypto.CertValidationPending ||
-                Connection->Crypto.TicketValidationPending ||
-                !QuicConnRecvHeader(
```
rzikm, line 5623: I am not 100% sure about this change, the original code seems to cause all incoming packets to be dropped while we have some pending validation, and thus we end up ignoring incoming CONNECTION_CLOSE. Was this intended?

src/core/connection.c
```diff
@@ -1441,13 +1441,15 @@ QuicConnTryClose(
...
             //
             // Application close can only happen if we are using 1-RTT keys.
-            // Otherwise we have to send "user_canceled" TLS error code as a
-            // connection close. Overwrite all application provided parameters.
+            // But until the handshake is confirmed, we can't be sure if the
+            // peer is able to process the close frame.
+            //
+            // Send "user_canceled" TLS error code as a connection close instead.
+            // Overwrite all application provided parameters.
             //
```
rzikm, line 1453: Open question: This change makes the behavior symmetrical, it is possible for the server to send Application CONNECTION_CLOSE because client has the keys to read it.

rzikm: The condition was originally introduced by https://github.com/microsoft/msquic/pull/3186

src/core/connection.c
rzikm: The condition was originally introduced by https://github.com/microsoft/msquic/pull/3186

src/core/connection.c
nibanks: I would prefer to minimize behavior change, and limit this to client side only.

src/core/connection.c
nibanks: I'm not so sure either. I will follow up.

src/core/connection.c
```diff
@@ -1441,13 +1441,16 @@ QuicConnTryClose(
     }
 
     if (!ClosedRemotely) {
-
         if ((Flags & QUIC_CLOSE_APPLICATION) &&
-            Connection->Crypto.TlsState.WriteKey < QUIC_PACKET_KEY_1_RTT) {
```
nibanks, line 1446: It's possible for the server to close the connection before we have 1-RTT keys I believe (though very rare). I feel like we should keep the old logic for server too.

src/core/connection.c
rzikm: The failing tests seem to give a clue that this change breaks something.

When we do async validation, we pretend that the cert is valid at the TLS layer and up the ReadKey on server side, so we start receiving and processing incoming 1RTT packets

```
Error: [03/01/2024 14:50:45] Handshake/WithHandshakeArgs5.CustomClientCertificateValidation/1 failed (in 2.1450049 sec):
Note: Google Test filter = Handshake/WithHandshakeArgs5.CustomClientCertificateValidation/1
[----------] 1 test from Handshake/WithHandshakeArgs5
[ RUN      ] Handshake/WithHandshakeArgs5.CustomClientCertificateValidation/1
D:\a\msquic\msquic\src\test\lib\HandshakeTest.cpp(916): error: Unexpected new Stream received

[  FAILED  ] Handshake/WithHandshakeArgs5.CustomClientCertificateValidation/1, where GetParam() = Reject/Async (2084 ms)
[----------] 1 test from Handshake/WithHandshakeArgs5 (2084 ms total)
```

src/platform/tls_schannel.c
```diff
@@ -2505,11 +2505,19 @@ CxPlatTlsWriteDataToSchannel(
                         Result |= CXPLAT_TLS_RESULT_ERROR;
                         break;
                     }
-                    State->ReadKey = QUIC_PACKET_KEY_1_RTT;
-                    QuicTraceLogConnInfo(
-                        SchannelRead1RttStart,
-                        TlsContext->Connection,
-                        "Reading 1-RTT data starts now");
+                    //
```
nibanks, line 2508: I'm not sure about these TLS layer changes... Why can't this be abstracted away at the QUIC layer, like we did do before?

rzikm: Not sure where would be the best place for that, I thought it better to reflect the condition in the ReadKey value itself and not add additional conditions on multiple places. Elsewhere.

Do you have any suggestions?

nibanks: We should defer packet processing at the QUIC layer until we're ready.

rzikm: I see, so you think it would be better somewhere like `QuicConnGetKeyOrDeferDatagram` and keep the TLS layer intact?

nibanks: Yeah. I think so.

rzikm:
I had some more time to spend on this, and it turns out to be a slightly bigger rabbit hole than I originally thought:

The way TLS abstraction on top of the CRYPTO stream works, is that when there is a pending validation, we keep "checked out" part of the buffer, and if we attempt to process future CRYPTO frame before the async validation is completed, an assert segfaults, which is a good thing I guess. However, this means that we have to pause processing CRYPTO data until the user sets the result for the async operation.

This is inconvenient especially on client, as current MsQuic implementation sends the CONNECTION_CLOSE on 1-RTT level, because it has the keys, but client derives the keys only once the (asynchronous) cert validation completes (on Linux at least). So the client will defer the CONNECTION_CLOSE until the async validation completes. We should implement sending CC on "Current and previous levels" as RFC suggests to get the desired behavior, but I would prefer to do that as a separate change.

src/core/connection.c
rzikm: Removing the check also caused asserts on Linux because MsQuic attempted to process CRYPTO frames while we have pending async validation.

src/core/crypto.c
```diff
@@ -1813,8 +1827,9 @@ QuicCryptoProcessData(
                 &BufferCount,
                 &Buffer);
 
-        UNREFERENCED_PARAMETER(DataAvailable);
-        CXPLAT_TEL_ASSERT(DataAvailable);
+        if (!DataAvailable) {
+            return Status;
+        }
```
rzikm, line 1832: This is needed for when the function is called from *ValidationComplete functions, if we want to preserve the assert then it should be possible by adding another parameter to the function saying "There should be data available"

rzikm:
There seem to be some test failures in the kernel mode runs, @nibanks Do you have any suggestions how to debug those?

For the others, there seems to be some nondeterminism in the added tests. And the best way to fix it seemed to me to bite the bullet and implement the sending of CC on both current and lower encryption levels, but that is going to break down-level tests because it introduces a behavioral change (the client is going to process CC in Handshake packets earlier than CC in 1-RTT packets, where previously we did not send CC in Handshake packets). What do you think?

src/core/connection.c
```diff
@@ -3694,6 +3698,15 @@ QuicConnGetKeyOrDeferDatagram(
...
     }
 
+    if (QuicConnIsServer(Connection) && !Connection->State.HandshakeConfirmed &&
+        Packet->KeyType == QUIC_PACKET_KEY_1_RTT) {
+        //
+        // A server MUST NOT process incoming 1-RTT protected packets before the TLS
+        // handshake is complete.
+        //
+        return FALSE;
+    }
```
nibanks, line 3708: This is practically a different bug, right? What exactly is the impact of this? Should this be backported?

src/core/connection.c
```diff
@@ -1450,26 +1451,29 @@ QuicConnTryClose(
...
-            RemoteReasonPhrase = NULL;
-            RemoteReasonPhraseLength = 0;
-
-            QuicTraceLogConnInfo(
-                CloseUserCanceled,
-                Connection,
-                "Connection close using user canceled error");
-        }
-    }
+    // if (!ClosedRemotely) {
```
nibanks, line 1454: Go ahead and delete this.

src/core/packet_builder.c
```diff
@@ -481,6 +481,37 @@ QuicPacketBuilderGetPacketTypeAndKeyForControlFrames(
...
+                // first packet, use lower key
+                *PacketKeyType = PreviousKeyType;
+            } else {
+                *PacketKeyType = MaxKeyType;
+            }
+
+            return TRUE;
+        }
+
+        if (Connection->Crypto.TlsState.WriteKey == QUIC_PACKET_KEY_0_RTT) {
```
nibanks, line 507: Can you add a comment explaining why this is necessary?

src/core/packet_builder.c
```diff
@@ -481,6 +481,37 @@ QuicPacketBuilderGetPacketTypeAndKeyForControlFrames(
...
+            *PacketKeyType = QUIC_PACKET_KEY_INITIAL;
+        } else {
+            *PacketKeyType = MaxKeyType;
+        }
+        return TRUE;
```
nibanks, line 512:
```suggestion

        return TRUE;
```

src/core/send.c
```diff
@@ -529,12 +528,29 @@ QuicSendWriteFrames(
...
+        if (IsApplicationClose && ! Is1RttEncryptionLevel) {
+            // 
+            // A CONNECTION_CLOSE of type 0x1d MUST be replaced by a
+            // CONNECTION_CLOSE of type 0x1c when sending the frame in Initial or
+            // Handshake packets. Otherwise, information about the application
+            // state might be revealed. Endpoints MUST clear the value of the
+            // Reason Phrase field and SHOULD use the APPLICATION_ERROR code when
+            // converting to a CONNECTION_CLOSE of type 0x1c.
+            //
+            CloseErrorCode = QUIC_ERROR_APPLICATION_ERROR;
```
nibanks, line 543: Why aren't we using the `QUIC_ERROR_CRYPTO_USER_CANCELED` error? I thought that's what is supposed to be used.

src/test/lib/HandshakeTest.cpp
```diff
@@ -1010,6 +1010,92 @@ QuicTestCustomClientCertificateValidation(
...
+
+                CxPlatSleep(1000);
+
+                //
+                // By now, the handshake is waiting for custom certificate validation.
+                //
+                if (ClientShutdown) {
+                    Client.Shutdown(QUIC_CONNECTION_SHUTDOWN_FLAG_NONE, QUIC_TEST_NO_ERROR);
+
+                    if (!Client.WaitForShutdownComplete()) {
```
nibanks, line 1071: Should these be `TEST_` checks instead of just returns? We should verify they happen correctly.

src/core/crypto.c
```diff
@@ -1720,6 +1720,11 @@ QuicCryptoCustomCertValidationComplete(
             QuicCryptoGetConnection(Crypto),
             "Custom cert validation succeeded");
         QuicCryptoProcessDataComplete(Crypto, Crypto->PendingValidationBufferLength);
+
+        if (QuicRecvBufferHasUnreadData(&Crypto->RecvBuffer)) {
+            // more data was received while waiting for user to perform the validation
+            QuicCryptoProcessData(Crypto, FALSE);
+        }
```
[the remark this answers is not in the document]
rzikm: It feels more appropriate to keep it separate and not have the logic mutually recursive (QuicCryptoProcessData would call QuicCryptoProcessDataComplete and vice versa).

The additional data are potentially present only when async validations are performed, for synchronous processing, there is never any further additional data since we process everything in single call.

src/core/send.c
rzikm: See [RFC](https://datatracker.ietf.org/doc/html/rfc9000#section-10.2.3-3)

> Sending a CONNECTION_CLOSE of type 0x1d in an Initial or Handshake packet could expose application state or be used to alter application state. A CONNECTION_CLOSE of type 0x1d MUST be replaced by a CONNECTION_CLOSE of type 0x1c when sending the frame in Initial or Handshake packets. Otherwise, information about the application state might be revealed. Endpoints MUST clear the value of the Reason Phrase field and SHOULD use the APPLICATION_ERROR code when converting to a CONNECTION_CLOSE of type 0x1c.

Follow up question: should we introduce also equivalent QUIC_STATUS code?

src/test/lib/HandshakeTest.cpp
rzikm: Those functions assert that given event is raised within some time (2s). The early returns are there to avoid continuing further and crashing due to unmet assumptions of the testing code.

src/test/lib/EventTest.cpp
```diff
@@ -395,7 +395,7 @@ QuicTestValidateConnectionEvents1(
     ConnValidator Server(
         new(std::nothrow) ConnEventValidator* [6] {
             new(std::nothrow) ConnEventValidator(QUIC_CONNECTION_EVENT_CONNECTED, 0, true),
-            new(std::nothrow) ConnEventValidator(QUIC_CONNECTION_EVENT_SHUTDOWN_INITIATED_BY_PEER),
+            new(std::nothrow) ConnEventValidator(QUIC_CONNECTION_EVENT_SHUTDOWN_INITIATED_BY_TRANSPORT),
```
rzikm, line 398: These and other changes are most likely inevitable. The affected scenario is clients closing the connections from inside `QUIC_CONNECTION_EVENT_CONNECTED` callback. (doing so from server still sends app close).

The reason is that RFC suggests that before handshake is **confirmed** (which for client means receiving HANDSHAKE_COMPLETE frame from server), endpoints send CONNECTION CLOSE also on lower protection levels. However, the `QUIC_CONNECTION_EVENT_CONNECTED` is signaled immediately after client-side handshake *completes* (meaning TLS layer finishes processing remote Finished and produces local Finished TLS messages).

To keep the old behavior, we would likely have to either
- delay the event until handshake is confirmed which would be more disrupting change as it would add latency to connection establishment, or
- somehow send the app close only, and only after some time send the CONNECTION_CLOSE on lower levels in case peer is "stuck" and does not process the 1-RTT frame with app close. This sounds rather difficult given the code architecture and feels rather brittle.

src/core/send.c
nibanks: This must have changed after we wrote the USER_CANCELED code, and then we just never updated to match.
> Follow up question: should we introduce also equivalent QUIC_STATUS code?
No. Let's leave it as is for now.

src/core/crypto.c
```diff
@@ -1720,6 +1720,11 @@ QuicCryptoCustomCertValidationComplete(
             QuicCryptoGetConnection(Crypto),
             "Custom cert validation succeeded");
         QuicCryptoProcessDataComplete(Crypto, Crypto->PendingValidationBufferLength);
+
+        if (QuicRecvBufferHasUnreadData(&Crypto->RecvBuffer)) {
+            // More data was received while waiting for user to perform the validation.
```
nibanks, line 1725: Please update to the style we use for comments.
```suggestion
            //
            // More data was received while waiting for user to perform the validation.
            //
```
Same for the function below.

src/core/packet_builder.c
```diff
@@ -540,16 +558,12 @@ QuicPacketBuilderGetPacketTypeAndKeyForControlFrames(
         }
     }
 
-    if (SendFlags & (QUIC_CONN_SEND_FLAG_CONNECTION_CLOSE | QUIC_CONN_SEND_FLAG_PING)) {
+    if (SendFlags & QUIC_CONN_SEND_FLAG_PING) {
         //
-        // CLOSE or PING is ready to be sent. This is always sent with the
+        // PING is ready to be sent. This is always sent with the
         // current write key.
```
nibanks, line 564:
```suggestion
        // PING is ready to be sent. This is always sent with the current write key.
```

src/core/send.c
```diff
@@ -543,7 +558,10 @@ QuicSendWriteFrames(
                 AvailableBufferLength,
                 Builder->Datagram->Buffer)) {
 
-            Send->SendFlags &= ~(QUIC_CONN_SEND_FLAG_CONNECTION_CLOSE | QUIC_CONN_SEND_FLAG_APPLICATION_CLOSE);
+            if (Builder->Key->Type == Connection->Crypto.TlsState.WriteKey) {
+                Send->SendFlags &= ~(QUIC_CONN_SEND_FLAG_CONNECTION_CLOSE | QUIC_CONN_SEND_FLAG_APPLICATION_CLOSE);
```
nibanks, line 562: Put a comment here to say why, please.

src/test/lib/EventTest.cpp
nibanks: Let's just take (and backport) this change. For the backport, we will likely need to update this to allow for either event (making them both optional).

rzikm:
Looking at CI errors, there are many failures on different cases of `Handshake/WithHandshakeArgs1.Connect`

```
 Handshake/WithHandshakeArgs1.Connect/9
D:\a\msquic\msquic\src\test\lib\HandshakeTest.cpp(427): error: Server->GetPeerClosed() not true
```

I think this PR creates a race condition. The `Handshake.Connect` test initiates a connection, asserts some properties and then closes the connection at the client side. However, based on the timing (whether client received HANDSHAKE_COMPLETED frame from Server), it would either send App close, or transport close followed by App close.

An idea occurred to me, if we encounter transport close with APPLICATION_ERROR, we could delay the closing of the connection and check if there is a following 1-RTT packet + App close frame and report that close error instead. This would keep the same behavior as before (at least when running against other MsQuic servers of the same version). What do you think, @nibanks?

nibanks:
> An idea occurred to me, if we encounter transport close with APPLICATION_ERROR, we could delay the closing of the connection and check if there is a 1-RTT packet in the same datagram with an App close frame and report that close error instead. This would keep the same behavior as before (at least when running against other MsQuic servers of the same version). What do you think, @nibanks?

I like the idea. I wonder if instead of delay, we treat it more like we allow the 1-RTT app close to overwrite the transport close. The change likely will just be that we need to keep processing the packets (and frames).

rzikm:
> I like the idea. I wonder if instead of delay, we treat it more like we allow the 1-RTT app close to overwrite the transport close. The change likely will just be that we need to keep processing the packets (and frames).

Right, I did not mean time delay, just postponing indicating the event to user.

rzikm:
@nibanks I think the PR is ready for another round of review. there are some failing CI legs, but I don't see how those failures would be related.

FYI, one of the test failures seem to be from not receiving any CONN_CLOSE at all. And the logs show some weird things, see excerpt with my comments.

```
[1]07C8.14F8::2024/03/27-12:48:40.926143200 [Microsoft-Quic][conn][0x19E86CD9690] App Shutdown: 0 (Remote=0)
[1]07C8.14F8::2024/03/27-12:48:40.926144700 [Microsoft-Quic][conn][0x19E86CD9690] Execute: 4(FLUSH_SEND)
[1]07C8.14F8::2024/03/27-12:48:40.926145800 [Microsoft-Quic][conn][0x19E86CD9690] Flushing Send. Allowance=11902 bytes
[2]07C8.167C::2024/03/27-12:48:40.926147000 [Microsoft-Quic][ xdp][0000019EFDA36700] XDP async IO start (TX)
[1]07C8.14F8::2024/03/27-12:48:40.926148100 [Microsoft-Quic][pack][4398046511105] Created in batch 4398046511105
[2]07C8.167C::2024/03/27-12:48:40.926149300 [Microsoft-Quic][ xdp][0000019EFDA36A30] XDP async IO start (RX)
[1]07C8.14F8::2024/03/27-12:48:40.926153000 [Microsoft-Quic][C][TX][17] LH Ver:0x1000000 DestCid:b83026dd404b26455f SrcCid: Type:HS (Payload 24 bytes)
[1]07C8.14F8::2024/03/27-12:48:40.926155200 [Microsoft-Quic][C][TX][17]   CONN_CLOSE ErrorCode:0xB FrameType:0
[1]07C8.14F8::2024/03/27-12:48:40.926155700 [Microsoft-Quic][pack][4398046511105] Encrypting
[1]07C8.14F8::2024/03/27-12:48:40.926160000 [Microsoft-Quic][pack][4398046511105] Finalizing
[1]07C8.14F8::2024/03/27-12:48:40.926160900 [Microsoft-Quic][conn][0x19E86CD9690][TX][17] 3(HANDSHAKE) (42 bytes)
[1]07C8.14F8::2024/03/27-12:48:40.926162200 [Microsoft-Quic][pack][4398046511106] Created in batch 4398046511105
[1]07C8.14F8::2024/03/27-12:48:40.926165000 [Microsoft-Quic][C][TX][18] SH DestCid:b83026dd404b26455f KP:0 SB:0 (Payload 7 bytes)
[1]07C8.14F8::2024/03/27-12:48:40.926166400 [Microsoft-Quic][C][TX][18]   CONN_CLOSE (App) ErrorCode:0x0
[1]07C8.14F8::2024/03/27-12:48:40.926166900 [Microsoft-Quic][pack][4398046511106] Encrypting
[1]07C8.14F8::2024/03/27-12:48:40.926168100 [Microsoft-Quic][pack][4398046511106] Finalizing
[1]07C8.14F8::2024/03/27-12:48:40.926168500 [Microsoft-Quic][conn][0x19E86CD9690][TX][18] 5(ONE_RTT) (33 bytes)
[1]07C8.14F8::2024/03/27-12:48:40.926169900 [Microsoft-Quic][conn][0x19E86CD9690] Sending batch. 1 datagrams

// packet with CONN_CLOSE frames
[1]07C8.14F8::2024/03/27-12:48:40.926170400 [Microsoft-Quic][data][0x19E86CDAEA0] Send 75 bytes in 1 buffers (segment=75) Dst=[fc00::1:11]:63581 Src=[fc00::1:12]:63582

[1]07C8.14F8::2024/03/27-12:48:40.926184800 [Microsoft-Quic][pack][4398046511105] Batch sent
[1]07C8.14F8::2024/03/27-12:48:40.926186200 [Microsoft-Quic][conn][0x19E86CD9690] Flush complete flags=0x0
[1]07C8.14F8::2024/03/27-12:48:40.926187100 [Microsoft-Quic][conn][0x19E86CD9690] Execute: 1(FLUSH_RECV)
[1]07C8.14F8::2024/03/27-12:48:40.926188000 [Microsoft-Quic][conn][0x19E86CD9690] Recv 3 UDP datagrams, 2762 bytes
[1]07C8.14F8::2024/03/27-12:48:40.926189200 [Microsoft-Quic][conn][0x19E86CD9690] IN: BytesRecv=18040
[1]07C8.14F8::2024/03/27-12:48:40.926190700 [Microsoft-Quic][conn][0x19E86CD9690] IN: BytesRecv=19272
[1]07C8.14F8::2024/03/27-12:48:40.926191500 [Microsoft-Quic][conn][0x19E86CD9690] IN: BytesRecv=19602
[1]07C8.14F8::2024/03/27-12:48:40.926192700 [Microsoft-Quic][conn][0x19E86CD9690] Batch Recv 3 UDP datagrams
[1]07C8.14F8::2024/03/27-12:48:40.926194600 [Microsoft-Quic][pack][1099511627814] Decrypting
[1]07C8.14F8::2024/03/27-12:48:40.926200900 [Microsoft-Quic][C][RX][15] SH DestCid: KP:0 SB:0 (Payload 1183 bytes)
[1]07C8.14F8::2024/03/27-12:48:40.926202700 [Microsoft-Quic][C][RX][15]   HANDSHAKE_DONE
[1]07C8.14F8::2024/03/27-12:48:40.926206600 [Microsoft-Quic][C][RX][15]   NEW_CONN_ID Seq:1 RPT:0 CID:b8300606d00256dcf4 Token:74551f4b8b6a5e7c5747e12ef384dcc7
[1]07C8.14F8::2024/03/27-12:48:40.926210000 [Microsoft-Quic][C][RX][15]   NEW_CONN_ID Seq:2 RPT:0 CID:b830f3bf2ebc9478a7 Token:ec333970054bb94e6577c425d9519681
[1]07C8.14F8::2024/03/27-12:48:40.926213300 [Microsoft-Quic][C][RX][15]   NEW_CONN_ID Seq:3 RPT:0 CID:b830b6c254b146e629 Token:2e3cf1e36df323260ec6b5a482f04b43
[1]07C8.14F8::2024/03/27-12:48:40.926218500 [Microsoft-Quic][C][RX][15]   PADDING Len:1091
[1]07C8.14F8::2024/03/27-12:48:40.926219300 [Microsoft-Quic][conn][0x19E86CD9690][RX][15] 5(ONE_RTT) (1184 bytes)
[1]07C8.14F8::2024/03/27-12:48:40.926220900 [Microsoft-Quic][conn][0x19E86CD9690] Handshake confirmed (frame)
[1]07C8.14F8::2024/03/27-12:48:40.926232300 [Microsoft-Quic][conn][0x19E86CD9690] Discarding key type = 2
[1]07C8.14F8::2024/03/27-12:48:40.926238400 [Microsoft-Quic][C][TX][15] ACKed (implicit)
[1]07C8.14F8::2024/03/27-12:48:40.926239000 [Microsoft-Quic][conn][0x19E86CD9690][TX][15] 3(HANDSHAKE) ACKed
[1]07C8.14F8::2024/03/27-12:48:40.926241300 [Microsoft-Quic][conn][0x19E86CD9690] Received ack for 52 crypto bytes, offset=2972
[1]07C8.14F8::2024/03/27-12:48:40.926243400 [Microsoft-Quic][conn][0x19E86CD9690] QS:3024 MAX:3024 UNA:3024 NXT:3024 RECOV:0-0
[1]07C8.14F8::2024/03/27-12:48:40.926245400 [Microsoft-Quic][C][TX][17] ACKed (implicit)
[1]07C8.14F8::2024/03/27-12:48:40.926246000 [Microsoft-Quic][conn][0x19E86CD9690][TX][17] 3(HANDSHAKE) ACKed
[1]07C8.14F8::2024/03/27-12:48:40.926247500 [Microsoft-Quic][conn][0x19E86CD9690] OUT: BytesSent=19405 InFlight=0 CWnd=12000 ConnFC=16777216 ISB=131072 PostedBytes=0 SRtt=18158 1Way=530784605917672855
[1]07C8.14F8::2024/03/27-12:48:40.926248300 [Microsoft-Quic][conn][0x19E86CD9690] OUT: StreamFC=0 StreamSendWindow=0
[1]07C8.14F8::2024/03/27-12:48:40.926254700 [Microsoft-Quic][C][RX][15] Marked for ACK (ECN=0)
[1]07C8.14F8::2024/03/27-12:48:40.926257200 [Microsoft-Quic][pack][1099511627815] Decrypting
[2]07C8.167C::2024/03/27-12:48:40.926258200 [Microsoft-Quic][ xdp][0000019EFDA36700] XDP async IO complete (RX)
[2]07C8.167C::2024/03/27-12:48:40.926259300 [Microsoft-Quic][ xdp][0000019EFDA36A30] XDP async IO complete (TX)

// datagram received, notice weird port numbers
[2]07C8.167C::2024/03/27-12:48:40.926262000 [Microsoft-Quic][data][0x19EFDA3C0D0] Recv 75 bytes (segment=75) Src=[fc00::1:11]:63581 Dst=[fc00::1:12]:63582

[2]07C8.167C::2024/03/27-12:48:40.926262600 [Microsoft-Quic][pack][1099511627817] Received
[1]07C8.14F8::2024/03/27-12:48:40.926263200 [Microsoft-Quic][C][RX][16] SH DestCid: KP:0 SB:0 (Payload 1215 bytes)
[1]07C8.14F8::2024/03/27-12:48:40.926265000 [Microsoft-Quic][C][RX][16]   PING
[1]07C8.14F8::2024/03/27-12:48:40.926270600 [Microsoft-Quic][C][RX][16]   PADDING Len:1210
[2]07C8.167C::2024/03/27-12:48:40.926271400 [Unknown]Unknown( 32): GUID=bbccf6c1-6cd1-48c4-80ff-839482e37671 (No Format Information found).

// looks like the packet is dropped for some reason
[2]07C8.167C::2024/03/27-12:48:40.926271400 [Microsoft-Quic][bind][0x19E86CFCFB0] DROP packet Dst=[fc00::1:11]:63581 Src=[fc00::1:12]:63582 Reason=Non-initial packet not matched with a connection.

[1]07C8.14F8::2024/03/27-12:48:40.926271600 [Microsoft-Quic][conn][0x19E86CD9690][RX][16] 5(ONE_RTT) (1216 bytes)
[2]07C8.167C::2024/03/27-12:48:40.926275600 [Microsoft-Quic][ xdp][0000019EFDA36700] XDP async IO start (RX)
[1]07C8.14F8::2024/03/27-12:48:40.926277200 [Microsoft-Quic][C][RX][16] Marked for ACK (ECN=0)
[2]07C8.167C::2024/03/27-12:48:40.926277900 [Microsoft-Quic][ xdp][0000019EFDA36A30] XDP async IO start (TX)
```

There seem to be no attempts to retransmit CONN_CLOSE even if the endpoint is receiving other packets from the other side. (RFC says something along the lines that endpoints that are closing should reply to all incoming packets with a repeated CONN_CLOSE frame).

nibanks (commented):
PR looks good. I agree the test failures are weird. Can you merge main once more so the tests can run again to see if they are consistently failing?

rzikm:
@nibanks There are some test failures, but there are no logs attached in the downloadable artifacts, they might be the same as in https://github.com/microsoft/msquic/pull/4169#issuecomment-2023207603 but I can't say for sure.

rzikm:
The comment above the mentioned code suggests that at server-side, Destination CIDs are not under server control. I needed to refresh my memory by reading the [RFC](https://datatracker.ietf.org/doc/html/rfc9000#section-7.2-8)

> A client MUST change the Destination Connection ID it uses for sending packets in response to only the first received Initial or Retry packet. A server MUST set the Destination Connection ID it uses for sending packets based on the first received Initial packet. Any further changes to the Destination Connection ID are only permitted if the values are taken from NEW_CONNECTION_ID frames; if subsequent Initial packets include a different Source Connection ID, they MUST be discarded. This avoids unpredictable outcomes that might otherwise result from stateless processing of multiple Initial packets with different Source Connection IDs.

So it seems to me that RFC-conforming client can only choose the DCID of the first Initial packet, so it seems safe to lookup Handshake packets by DCID as well. what do you think?

rzikm:
Following quick and dirty attempt fixes the issue for my local repro

```diff
-   if (!Binding->ServerOwned || Packets->IsShortHeader) {
+   if (!Binding->ServerOwned || Packets->IsShortHeader || Packets->LH->Type == QUIC_HANDSHAKE_V1) {
        Connection =
            QuicLookupFindConnectionByLocalCid(
                &Binding->Lookup,
                Packets->DestCid,
                Packets->DestCidLen);
    } else {
```

Proper fix would have to take QUIC Version-dependent header Type values to account ofc.

What do you think, @nibanks?

nibanks:
> So it seems to me that RFC-conforming client can only choose the DCID of the first Initial packet, so it seems safe to lookup Handshake packets by DCID as well. what do you think?

Client can (and do I believe) continue to use the initial DCID beyond the first round trip. AFAIK, there is nothing that prevents that. So, while it might work with MsQuic as a client, it doesn't necessarily work with all clients.

You also need to consider if there is any (probably not) attack implication by allowing these types of packets to be routed.

> What do you think, @nibanks?

You also need to make sure those fields have been validated already. Another thought I had though, is should we delay removing the DCID from the lookup table on client, until we have an ACK for the handshake confirmed frame? Would that fix this too?

rzikm:
> Client can (and do I believe) continue to use the initial DCID beyond the first round trip.

I thought that is against the very first sentence of the linked paragraph.

> A client MUST change the Destination Connection ID it uses for sending packets in response to only the first received Initial or Retry packet.

As for possible attacks, I don't see any, third parties should not have access to necessary handshake keys, and packets failing decryption/integrity check shouldn't affect the connection.

> You also need to make sure those fields have been validated already.

The fields are luckily already validated at this point (assert at the beginning of the containing function).

>  Another thought I had though, is should we delay removing the DCID from the lookup table on client, until we have an ACK for the handshake confirmed frame? Would that fix this too?

That would work as well, but the fix would be slightly more complicated. I can change it if you think this would be a better fix.

rzikm:
> Another thought I had though, is should we delay removing the DCID from the lookup table on client, until we have an ACK for the handshake confirmed frame? Would that fix this too?

My fix broke 0RTT tests. Looks like your idea works better, so let's go with that :D

src/core/crypto.c
```diff
@@ -485,15 +485,18 @@ QuicCryptoOnVersionChange(
...
-    QUIC_PATH* Path = &Connection->Paths[0];
-    CXPLAT_DBG_ASSERT(Path->Binding != NULL);
-    QuicBindingOnConnectionHandshakeConfirmed(Path->Binding, Connection);
+    if (SignalBinding) {
+        QUIC_PATH* Path = &Connection->Paths[0];
+        CXPLAT_DBG_ASSERT(Path->Binding != NULL);
+        QuicBindingOnConnectionHandshakeConfirmed(Path->Binding, Connection);
+    }
 
     QuicCryptoDiscardKeys(Crypto, QUIC_PACKET_KEY_HANDSHAKE);
```
nibanks, line 501: Does it make sense to throw away the handshake key still? Or should this whole function just be delayed?

rzikm: I don't think it makes sense to keep them, when TLS handshake finishes at server side, client already must have 1-RTT keys. The issue this fixes is only about getting the coalesced datagram routed to the connection.

nibanks (approved):

PR #4177, opened 2024-03-06
Support Multiple Pending Receives

nibanks:
## Description

Some internal refactoring in preparation to support multiple parallel receives to the app. No actual external behavior change yet. Just making sure no existing scenarios are broken.

## Testing

CI/CD

## Documentation

N/A

### DIFF
--- a/src/core/stream.c
+++ b/src/core/stream.c
@@ -75,6 +75,12 @@ QuicStreamInitialize(
     QuicRangeInitialize(
         QUIC_MAX_RANGE_ALLOC_SIZE,
         &Stream->SparseAckRanges);
+    Stream->ReceiveCompleteOperation = &Stream->ReceiveCompleteOperationStorage;
+    Stream->ReceiveCompleteOperationStorage.API_CALL.Context = &Stream->ReceiveCompleteApiCtxStorage;
+    Stream->ReceiveCompleteOperation->Type = QUIC_OPER_TYPE_API_CALL;
+    Stream->ReceiveCompleteOperation->FreeAfterProcess = FALSE;
+    Stream->ReceiveCompleteOperation->API_CALL.Context->Type = QUIC_API_TYPE_STRM_RECV_COMPLETE;
+    Stream->ReceiveCompleteOperation->API_CALL.Context->STRM_RECV_COMPLETE.Stream = Stream;
 #if DEBUG
     Stream->RefTypeCount[QUIC_STREAM_REF_APP] = 1;
 #endif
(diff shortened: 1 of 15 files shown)


src/core/stream_recv.c
```diff
@@ -1049,18 +1044,17 @@ QuicStreamReceiveComplete(
     _In_ uint64_t BufferLength
     )
 {
-    if (!Stream->Flags.ReceiveCallPending) {
+    if (!Stream->RecvCallPendingCount) {
         return FALSE;
     }
+    Stream->RecvCallPendingCount--;
```
nibanks, line 1050: This isn't going to work. With multiple receives, we want to support completing multiple with a single call too.

src/core/stream_recv.c
```diff
@@ -1049,46 +1021,50 @@ QuicStreamReceiveComplete(
...
 
+    CXPLAT_TEL_ASSERTMSG(
+        BufferLength <= Stream->RecvPendingLength,
+        "App overflowed read buffer!");
+
     //
     // Reclaim any buffer space comsumed by the app.
     //
-    if (Stream->RecvPendingLength == 0 ||
+    if (!Stream->RecvPendingLength ||
```
csujedihy, line 1044: Why changing it to `!Stream->RecvPendingLength`? The implicit conversion from an integer to a boolean makes the code less readable.

src/core/stream.c
```diff
@@ -195,10 +201,6 @@ QuicStreamFree(
     CxPlatDispatchLockUninitialize(&Stream->ApiSendRequestLock);
     CxPlatRefUninitialize(&Stream->RefCount);
 
-    if (Stream->ReceiveCompleteOperation) {
```
csujedihy, line 198: where do we free this now?

src/core/stream_recv.c
ProjectsByJackHe: Maybe `Stream->RecvPendingLength == 0`?

src/core/stream.c
nibanks: It's allocated with the Stream itself, so when the stream is freed.

src/core/stream_recv.c
ProjectsByJackHe: +1 on this

nibanks: I changed all the usages to this new model. What do folks prefer, the old version or the new?

csujedihy (approved):

PR #4208, opened 2024-03-20
add memory alignment for ARM32

### LINKED ISSUE #3958: msquic crashing on ARM32 machines
### Describe the bug

It seems like regression from https://github.com/microsoft/msquic/pull/3826 @nibanks.
I did binary search on changes and previous commit (cd49fbb) works fine. 
This is likely root cause for https://github.com/dotnet/runtime/issues/91757  causing .NET CI failures.
While reported on Alpine, I can reproduce on my Raspberry PI with Debian 10.

### Affected OS

- [ ] Windows
- [X] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

I tried main and commit 972e677
Comment cd49fbb seems ok. 

### MsQuic version

main

I can reproduce it running .NET Quic test suite. 

### Expected behavior

no crash and everything runs as before

### Actual outcome

```
Thread 14 ".NET Long Runni" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x678da440 (LWP 10187)]
0x683dc210 in InterlockedDecrement (Addend=0x60) at /home/furt/github/msquic/src/inc/quic_platform_posix.h:115
115	    return __sync_sub_and_fetch(Addend, (long)1);
(gdb) bt
#0  0x683dc210 in InterlockedDecrement (Addend=0x60) at /home/furt/github/msquic/src/inc/quic_platform_posix.h:115
#1  RecvDataReturn (RecvDataChain=0x0) at /home/furt/github/msquic/src/platform/datapath_epoll.c:2067
#2  0x683ce240 in CxPlatRecvDataReturn (RecvDataChain=<optimized out>) at /home/furt/github/msquic/src/platform/datapath_linux.c:338
#3  0x68381814 in QuicConnRecvDatagrams (Connection=Connection@entry=0x73398d58, Packets=0x0, Packets@entry=0x6ef34450, PacketChainCount=PacketChainCount@entry=2, PacketChainByteCount=PacketChainByteCount@entry=2324, IsDeferred=<optimized out>,
    IsDeferred@entry=0 '\000') at /home/furt/github/msquic/src/core/connection.c:5848
#4  0x68381d0c in QuicConnFlushRecv (Connection=Connection@entry=0x73398d58) at /home/furt/github/msquic/src/core/connection.c:5929
#5  0x68385844 in QuicConnDrainOperations (Connection=Connection@entry=0x73398d58) at /home/furt/github/msquic/src/core/connection.c:7676
#6  0x68359900 in QuicWorkerProcessConnection (Worker=Worker@entry=0x72921c10, Connection=0x73398d58, ThreadID=<optimized out>, TimeNow=TimeNow@entry=0x678d9e10) at /home/furt/github/msquic/src/core/worker.c:510
#7  0x6835a668 in QuicWorkerLoop (Context=0x72921c10, State=0x678d9e10) at /home/furt/github/msquic/src/core/worker.c:668
#8  0x683c8654 in CxPlatRunExecutionContexts (Worker=Worker@entry=0x7290aad0, State=State@entry=0x678d9e10) at /home/furt/github/msquic/src/platform/platform_worker.c:395
#9  0x683c8a74 in CxPlatWorkerThread (Context=0x7290aad0) at /home/furt/github/msquic/src/platform/platform_worker.c:492
#10 0x76f7c310 in start_thread (arg=0x678da440) at pthread_create.c:477
#11 0x76cc6598 in ?? () at ../sysdeps/unix/sysv/linux/arm/clone.S:73 from /lib/arm-linux-gnueabihf/libc.so.6
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```

### Additional details

I don't tank if this is because of speed tf the machine or if this is because ARM has weak memory model. (or something completely else)

ami-GS:
## Description

Fixes #3958 
Support arm32 memory alignment

## Testing

Tested and passed on raspberry pi.
It had been failed at processing 3rd or 4th packets consistently (during handshake). now it has gone with the fix.

## Documentation

### DIFF
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -76,7 +76,7 @@ typedef struct DATAPATH_RX_IO_BLOCK {
 
 } DATAPATH_RX_IO_BLOCK;
 
-typedef struct DATAPATH_RX_PACKET {
+typedef struct __attribute__((aligned(16))) DATAPATH_RX_PACKET {
     //
     // The IO block that owns the packet.
     //


src/platform/datapath_epoll.c
```diff
@@ -76,7 +76,7 @@ typedef struct DATAPATH_RX_IO_BLOCK {
 
 } DATAPATH_RX_IO_BLOCK;
 
-typedef struct DATAPATH_RX_PACKET {
+typedef struct __attribute__((aligned(16))) DATAPATH_RX_PACKET {
```
nibanks, line 79: Can you add more info in your PR description as to why this helps and is necessary?

ami-GS: Not 100% sure yet, so trying to run automation first

nibanks: Are you sure yet? 😄

wfurt: I did play with this little bit more and it is also sufficient (as far as I can tell) to increase `RecvBlockStride`

https://github.com/microsoft/msquic/blob/fbef931d274779bb02e635f8f4957a813126b911/src/platform/datapath_epoll.c#L333-L334

I originally added `+4` but it also works for me with +1 or +2. I sort of expected corrupted pointer if something overrides the location.

It seems like we are casting bunch of packets to single big buffer, right @nibanks? Are we positive the structures always land on aligned boundary? It seems like the issue is triggered only when we pack multiple packets into single buffer. In the test we look at with @ami-GS, there is single packet and then we have few together. As @ami-GS noticed it was always crashing on processing third one, e.g. first that is packed second in the buffer.

nibanks: @ami-GS can we shoot to get this merged this week? Please address my comment above, and associate this PR with the bug that's tracking the issue.

nibanks (approved):

PR #4215, opened 2024-03-25
Don't do full drain on chunks containing future buffered data

rzikm:
## Description

This PR fixes bug where in certain situations, MsQuic would drop/corrupt data when received out of order + specific read timing from application side.

## Testing

Failing test added to detect future regressions.

## Documentation

N/A.

### DIFF
--- a/src/core/recv_buffer.c
+++ b/src/core/recv_buffer.c
@@ -917,7 +917,7 @@ QuicRecvBufferDrain(
             //
             QuicRangeSize(&RecvBuffer->WrittenRanges) > 1) {
             QuicRecvBufferPartialDrain(RecvBuffer, DrainLength);
-            return PartialDrain;
+            return !PartialDrain;
         }
 
         DrainLength = QuicRecvBufferFullDrain(RecvBuffer, DrainLength);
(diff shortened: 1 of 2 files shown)


rzikm:
We still have some failures on the stress tests even with this fix, (https://github.com/dotnet/runtime/issues/100064#issuecomment-2018431845), let's not merge just yet, I will see if there is more to be done in this PR.

src/core/recv_buffer.c
```diff
@@ -907,7 +907,12 @@ QuicRecvBufferDrain(
     }
 
     do {
-        if ((uint64_t)RecvBuffer->ReadLength > DrainLength) {
+        if ((uint64_t)RecvBuffer->ReadLength > DrainLength ||
```
nibanks, line 910: This isn't quite the right fix. It should still call `QuicRecvBufferFullDrain` to indicate it drained everything that was read, but `QuicRecvBufferFullDrain` should be handling the fact that just because it read everything doesn't mean there isn't more after a gap to read.

nibanks: I am looking into what the right fix should be.

nibanks: We talked over IM and maybe using the partial drain is the right way to go. We'd just kind of redefine 'partial drain' here to indicate it's not partial of the read length, but of all (possibly discontiguous) data in the buffer.

nibanks (approved):

PR #4218, opened 2024-03-29
Support Client Migration (Client Side)

### LINKED ISSUE #1946: Support Client Migration (Client Side)

Support for QUIC clients to change their/add/remove local addresses (IP or port) should be added to MsQuic. More info at the protocol level can be found at https://datatracker.ietf.org/doc/html/rfc9000#section-9.

This feature allows the client to seamlessly switch between different networks, such as Wi-Fi and cellular, and keep the existing connection alive.

### Proposed solution

Starting at the API, probably the best way to tackle this would be to add a way for the app layer to add and remove addresses. New addresses won't be used as the primary by default. But QUIC can start probing them immediately to see if they work. The app will have to "set" the current local address to make it a primary. The app shouldn't need to add it and then set it; it can just add it if it wants to immediately switch it a new one.

Internally, these different addresses would map to separate Paths. All the client side logic for probing and switching paths would need to be completed.

### Additional thoughts

- Should MsQuic automatically forget addresses ever?
- Eventually, if the client doesn't explicitly set (bind) any local address, MsQuic should support automatically switching when the current address doesn't work. This would require a number of additional platform abstractions: querying addresses, getting notification of address changes, etc.
  - This is not in scope for this task though. That would be future work.
- The work should include as full test coverage as possible, using the existing datapath hooks to spoof anything necessary. Most of the work can be tested just by changing the local port of the client though.

masa-koz:
## Description

This PR adds the support of Client Migration (Client SIde).  Fix #1946.

## Testing

I've added two test.
- Basic/WithProbePathArgs.ProbePath/*
- Basic/WithMigrationArgs.Migration/*

## Documentation

In this PR, the following three Connection level Parameters will be added.
- QUIC_PARAM_CONN_ADD_LOCAL_ADDRESS
- QUIC_PARAM_CONN_REMOVE_LOCAL_ADDRESS
- QUIC_PARAM_CONN_LOCAL_UNUSED_DEST_CID_COUNT


masa-koz:

src/inc/msquic.h
```diff
@@ -909,6 +909,9 @@ typedef struct QUIC_SCHANNEL_CREDENTIAL_ATTRIBUTE_W {
 #define QUIC_PARAM_CONN_STATISTICS_V2                   0x05000016  // QUIC_STATISTICS_V2
 #define QUIC_PARAM_CONN_STATISTICS_V2_PLAT              0x05000017  // QUIC_STATISTICS_V2
 #define QUIC_PARAM_CONN_ORIG_DEST_CID                   0x05000018  // uint8_t[]
+#define QUIC_PARAM_CONN_ADD_LOCAL_ADDRESS               0x05000019  // QUIC_ADDR
+#define QUIC_PARAM_CONN_REMOVE_LOCAL_ADDRESS            0x0500001A  // QUIC_ADDR
+#define QUIC_PARAM_CONN_LOCAL_UNUSED_DEST_CID_COUNT     0x0500001B  // uint16_t
```
nibanks, line 914: Is this for test only? Do real clients really need this? If it's test only, please put in `msquicp.h`.

src/inc/quic_platform.h
```diff
@@ -80,7 +80,7 @@ typedef struct CXPLAT_SLIST_ENTRY {
 #define QUIC_POOL_WORKER                    'A0cQ' // Qc0A - QUIC Worker
 #define QUIC_POOL_LISTENER                  'B0cQ' // Qc0B - QUIC Listener
 #define QUIC_POOL_CID                       'C0cQ' // Qc0C - QUIC CID
-#define QUIC_POOL_CIDHASH                   'D0cQ' // Qc0D - QUIC CID Hash
+#define QUIC_POOL_CIDSLIST                  'D0cQ' // Qc0D - QUIC CID Hash
```
nibanks, line 83: Instead of replacing this and adding it down at the bottom, can you leave this as is, and put the new one and the end?

src/inc/quic_platform.h
```diff
@@ -144,6 +144,8 @@ typedef struct CXPLAT_SLIST_ENTRY {
 #define QUIC_POOL_ROUTE_RESOLUTION_WORKER   'A4cQ' // Qc4A - QUIC route resolution worker
 #define QUIC_POOL_ROUTE_RESOLUTION_OPER     'B4cQ' // Qc4B - QUIC route resolution operation
 #define QUIC_POOL_EXECUTION_CONFIG          'C4cQ' // Qc4C - QUIC execution config
+#define QUIC_POOL_CIDHASH                   'D4cQ' // Qc4D - QUIC CID Hash
+
```
nibanks, line 148: nit: unnecessary new line

src/platform/datapath_winkernel.c
```diff
@@ -2523,7 +2523,6 @@ CxPlatRecvDataReturn(
         DATAPATH_RX_IO_BLOCK* IoBlock =
             CXPLAT_CONTAINING_RECORD(Datagram, DATAPATH_RX_PACKET, Data)->IoBlock;
 
-        CXPLAT_DBG_ASSERT(Binding == NULL || Binding == IoBlock->Binding);
```
nibanks, line 2526: Why?

src/test/lib/PathTest.cpp
```diff
@@ -119,3 +119,181 @@ QuicTestLocalPathChanges(
...
+    TEST_TRUE(ClientConfiguration.IsValid());
+
+    MsQuicAutoAcceptListener Listener(Registration, ServerConfiguration, PathTestContext::ConnCallback, &Context);
+    TEST_QUIC_SUCCEEDED(Listener.GetInitStatus());
+    QUIC_ADDRESS_FAMILY QuicAddrFamily = (Family == 4) ? QUIC_ADDRESS_FAMILY_INET : QUIC_ADDRESS_FAMILY_INET6;
+    QuicAddr ServerLocalAddr(QuicAddrFamily);
+    TEST_QUIC_SUCCEEDED(Listener.Start("MsQuicTest", &ServerLocalAddr.SockAddr));
+    TEST_QUIC_SUCCEEDED(Listener.GetLocalAddr(ServerLocalAddr));
+
+    MsQuicConnection Connection(Registration, MsQuicCleanUpMode::CleanUpManual, ClientCallback, &PeerStreamsChanged);
```
nibanks, line 149:
```suggestion
    MsQuicConnection Connection(Registration, CleanUpManual, ClientCallback, &PeerStreamsChanged);
```
nit: should be unnecessary

src/test/lib/PathTest.cpp
```diff
@@ -119,3 +119,181 @@ QuicTestLocalPathChanges(
...
+                QUIC_PARAM_CONN_LOCAL_UNUSED_DEST_CID_COUNT,
+                &Size,
+                &Count);
+        if (QUIC_FAILED(Status)) {
+            break;
+        }
+    } while (Count == 0 && ++Try <= 3);
+    TEST_NOT_EQUAL(Count, 0);
+
+    QuicAddr SecondLocalAddr;
```
nibanks, line 180: nit: Rename to `ClientLocalAddr`?

src/test/lib/PathTest.cpp
```diff
@@ -119,3 +119,181 @@ QuicTestLocalPathChanges(
...
+    TEST_TRUE(ProbeHelper.ServerReceiveProbeEvent.WaitTimeout(TestWaitTimeout * 10));
+    TEST_TRUE(ProbeHelper.ClientReceiveProbeEvent.WaitTimeout(TestWaitTimeout * 10));
+    QUIC_STATISTICS_V2 Stats;
+    uint32_t Size = sizeof(Stats);
+    TEST_QUIC_SUCCEEDED(
+        Connection.GetParam(
+            QUIC_PARAM_CONN_STATISTICS_V2_PLAT,
+            &Size,
+            &Stats));
+    TEST_EQUAL(Stats.RecvDroppedPackets, 0);
```
nibanks, line 201: How does this line help? Why would the client drop a packet? Also, are there other test conditions we can add here to determine if the new path worked?

src/test/lib/PathTest.cpp
```diff
@@ -119,3 +119,181 @@ QuicTestLocalPathChanges(
...
+        TEST_EQUAL(Stats.RecvDroppedPackets, 0);
+    }
+
+    TEST_QUIC_SUCCEEDED(
+        Connection.SetParam(
+            QUIC_PARAM_CONN_LOCAL_ADDRESS,
+            sizeof(SecondLocalAddr.SockAddr),
+            &SecondLocalAddr.SockAddr));
+
+    Connection.SetSettings(MsQuicSettings{}.SetKeepAlive(25));
```
nibanks, line 291: Why set this here and not at the beginning?

src/core/binding.h
```diff
@@ -516,3 +521,16 @@ QuicRetryTokenDecrypt(
...
+//
+inline
+_Ret_notnull_
+QUIC_BINDING*
+QuicLookupGetBinding(
+    _In_ QUIC_LOOKUP* Lookup
+    )
+{
+    return CXPLAT_CONTAINING_RECORD(Lookup, QUIC_BINDING, Lookup);
+}
```
nibanks, line 536: nit: please keep trailing empty newline

src/core/lookup.c
```diff
@@ -746,8 +768,24 @@ QuicLookupAddLocalCid(
             Hash);
 
     if (ExistingConnection == NULL) {
-        Result =
-            QuicLookupInsertLocalCid(Lookup, Hash, SourceCid, TRUE);
+        QUIC_CID_HASH_ENTRY* CID =
```
nibanks, line 771: Just a thought, but we might need to start using pools for allocating and freeing all these entries so they don't hurt perf.

src/inc/msquic.h
masa-koz: Real clients also need this.

src/core/connection.c
```diff
@@ -836,12 +837,37 @@ QuicConnGenerateNewSourceCid(
...
             return NULL;
         }
-        if (!QuicBindingAddSourceConnectionID(Connection->Paths[0].Binding, SourceCid)) {
-            CXPLAT_FREE(SourceCid, QUIC_POOL_CIDHASH);
+
+        BOOLEAN Collision = FALSE;
+
+        CxPlatDispatchLockAcquire(&MsQuicLib.DatapathLock);
+
+        // Check whether a sourceCid collides or not around all the connections.
```
nibanks, line 849: Can you clarify this comment. I don't quite understand. Also, can you update this to match the existing style for comments, so it has ``//` before and after too?

src/inc/msquic.h
nibanks: I'd really prefer to be able to abstract away the complexities of connection IDs from clients. Why exactly do clients need this information? To know if they can add a new path?

src/core/connection.c
```diff
@@ -5065,7 +5145,7 @@ QuicConnRecvFrames(
                     !memcmp(Frame.Data, TempPath->Challenge, sizeof(Frame.Data))) {
                     QuicPerfCounterIncrement(QUIC_PERF_COUNTER_PATH_VALIDATED);
                     QuicPathSetValid(Connection, TempPath, QUIC_PATH_VALID_PATH_RESPONSE);
-                    break;
+                   break;
```
nibanks, line 5148: Accident? Please undo?

src/core/connection.c
```diff
@@ -6567,6 +6660,217 @@ QuicConnParamSet(
...
+
+        if (QuicConnIsServer(Connection)) {
+            Status = QUIC_STATUS_INVALID_STATE;
+            break;
+        }
+
+        if (!Connection->State.Started) {
+            Status = QUIC_STATUS_INVALID_STATE;
+            break;
+        }
```
nibanks, line 6678: Why not? It seems kind of intuitive to me that a client could create a connection, queue all paths it wants to use and then start it. No?

src/core/connection.c
```diff
@@ -6567,6 +6660,217 @@ QuicConnParamSet(
...
 
+    case QUIC_PARAM_CONN_ADD_LOCAL_ADDRESS: {
+
+        if (BufferLength != sizeof(QUIC_ADDR)) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+
+        if (QuicConnIsServer(Connection)) {
+            Status = QUIC_STATUS_INVALID_STATE;
```
nibanks, line 6671: I'm thinking of 'not supported' instead of 'invalid state'.

src/core/connection.c
```diff
@@ -6567,6 +6660,217 @@ QuicConnParamSet(
...
+        if (!QuicAddrIsValid(LocalAddress)) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+
+        BOOLEAN AddrInUse = FALSE;
+        for (uint8_t i = 0; i < Connection->PathsCount; ++i) {
+            if (QuicAddrCompare(
+                &Connection->Paths[i].Route.LocalAddress,
+                LocalAddress)) {
```
nibanks, line 6697:
```suggestion
            if (QuicAddrCompare(
                    &Connection->Paths[i].Route.LocalAddress,
                    LocalAddress)) {
```

src/core/connection.c
```diff
@@ -6567,6 +6660,217 @@ QuicConnParamSet(
...
+                break;
+            }
+        }
+
+        if (AddrInUse) {
+            Status = QUIC_STATUS_ADDRESS_IN_USE;
+            break;
+        }
+
+        if (Connection->PathsCount == QUIC_MAX_PATH_COUNT) {
```
nibanks, line 6708: Put this check up before creating a new destination CID. Also, return 'invalid state' instead?

src/core/connection.c
```diff
@@ -6567,6 +6660,217 @@ QuicConnParamSet(
...
+        }
+
+        if (!Connection->State.Started) {
+            Status = QUIC_STATUS_INVALID_STATE;
+            break;
+        }
+
+        QUIC_CID_LIST_ENTRY* NewDestCid = QuicConnGetUnusedDestCid(Connection);
+        if (NewDestCid == NULL) {
+            Status = QUIC_STATUS_INVALID_STATE;
```
nibanks, line 6682: I wonder if we can support a model where we always queue the path (somehow) even if we don't have a CID to use for it, _yet_...

src/core/connection.c
```diff
@@ -6567,6 +6660,217 @@ QuicConnParamSet(
...
+            break;
+        }
+
+        if (!Connection->State.Started) {
+            Status = QUIC_STATUS_INVALID_STATE;
+            break;
+        }
+
+        QUIC_CID_LIST_ENTRY* NewDestCid = QuicConnGetUnusedDestCid(Connection);
```
nibanks, line 6680: Does this need to be cleaned up at all the failure cases below?

src/core/connection.c
```diff
@@ -6567,6 +6660,217 @@ QuicConnParamSet(
         return QUIC_STATUS_SUCCESS;
     }
 
+    case QUIC_PARAM_CONN_ADD_LOCAL_ADDRESS: {
```
nibanks, line 6663: The code for handling all this is quite complex, and I think it deserves to be put in a separate function.

src/core/connection.c
```diff
@@ -6567,6 +6660,217 @@ QuicConnParamSet(
...
+            Connection,
+            CASTED_CLOG_BYTEARRAY(sizeof(Path->Route.LocalAddress), &Path->Route.LocalAddress));
+
+        QuicSendSetSendFlag(&Connection->Send, QUIC_CONN_SEND_FLAG_PATH_CHALLENGE);
+
+        Status = QUIC_STATUS_SUCCESS;
+        break;
+    }
+
+    case QUIC_PARAM_CONN_REMOVE_LOCAL_ADDRESS: {
```
nibanks, line 6817: Let's put this in a separate function too.

src/core/connection.c
```diff
@@ -6567,6 +6660,217 @@ QuicConnParamSet(
...
+        if (!QuicAddrIsValid(LocalAddress)) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+
+        uint8_t PathIndex = Connection->PathsCount;
+        for (uint8_t i = 0; i < Connection->PathsCount; ++i) {
+            if (QuicAddrCompare(
+                &Connection->Paths[i].Route.LocalAddress,
+                LocalAddress)) {
```
nibanks, line 6845:
```suggestion
            if (QuicAddrCompare(
                    &Connection->Paths[i].Route.LocalAddress,
                    LocalAddress)) {
```

src/platform/datapath_winkernel.c
masa-koz: When a connection binds more than one bindings (i.e. after adding a new local address on the client side), we cannot assure that the Datagrams in RecvDataChain come from the same binding.

src/core/connection.c
masa-koz: //
        // Check whether a new sourceCid collides with any sourceCid which belong to all
        // the bindings including the ones which this connection doesn't bound.
        //

Does this make sense?

src/core/connection.c
masa-koz: I've implemented this.
https://github.com/masa-koz/msquic/pull/3/files

src/inc/msquic.h
masa-koz: Yes. This information is needed to know when clients can add a new local address. But this is no longer needed if we adopt a new way in which clients defer sending a path challenge until they receive a new ConID.

src/core/connection.c
masa-koz: I agree with you. So, I've also implemented this. https://github.com/masa-koz/msquic/pull/4

src/core/connection.c
```diff
@@ -836,12 +837,37 @@ QuicConnGenerateNewSourceCid(
...
+                sizeof(QUIC_CID_SLIST_ENTRY) + MsQuicLib.CidTotalLength);
             QuicConnFatalError(Connection, QUIC_STATUS_INTERNAL_ERROR, NULL);
             return NULL;
         }
-        if (!QuicBindingAddSourceConnectionID(Connection->Paths[0].Binding, SourceCid)) {
-            CXPLAT_FREE(SourceCid, QUIC_POOL_CIDHASH);
+
+        BOOLEAN Collision = FALSE;
+
+        CxPlatDispatchLockAcquire(&MsQuicLib.DatapathLock);
```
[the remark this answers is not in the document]
masa-koz: I agree with you. I've tried to remove a global lock in https://github.com/masa-koz/msquic/pull/5.
In this change, we will replace all the existing source connection IDs if we find the collision in adding a source connection ID to a newly opened bindings.

src/test/lib/PathTest.cpp
masa-koz: The intention is that I would like to check whether the client succeeded in adding the source connection IDs to a newly opened bindings (if it failed, it will drop a packet whose destination is a newly added local address).
And I think that an event notification about the validation of a path is helpful for both testing and real application.

src/core/connection.c
masa-koz: I agree with you. I will change this.

src/core/connection.c
masa-koz: I suppose that cleaning up is not needed because UsedLocally is not set yet. Isn't it?

src/core/binding.h
```diff
@@ -383,30 +394,24 @@ _IRQL_requires_max_(DISPATCH_LEVEL)
...
-// Moves all the connections source CIDs from the one binding's lookup table to
-// another.
-//
 _IRQL_requires_max_(DISPATCH_LEVEL)
 void
-QuicBindingMoveSourceConnectionIDs(
-    _In_ QUIC_BINDING* BindingSrc,
-    _In_ QUIC_BINDING* BindingDest,
-    _In_ QUIC_CONNECTION* Connection
+QuicBindingRemoveRemoteHash(
```
anrossi, line 412: Does this mean there's no way to move source CIDs between bindings anymore?

src/core/connection.c
```diff
@@ -836,12 +837,40 @@ QuicConnGenerateNewSourceCid(
...
+        // Check whether a new sourceCid collides with any sourceCid which belong to all
+        // the bindings including the ones which this connection doesn't bound.
+        //
+        for (CXPLAT_LIST_ENTRY* Link = MsQuicLib.Bindings.Flink;
+            Link != &MsQuicLib.Bindings;
+            Link = Link->Flink) {
+
+            QUIC_BINDING* Binding =
+                CXPLAT_CONTAINING_RECORD(Link, QUIC_BINDING, Link);
+            QUIC_CONNECTION* Connection1 =
```
anrossi, line 859: Can we use a more descriptive name instead of `Connection1`? Like `SearchConnection` or `CollisionConnection`

src/core/connection.c
```diff
@@ -6099,6 +6186,201 @@ QuicConnUpdatePeerPacketTolerance(
...
+
+    QuicSendSetSendFlag(&Connection->Send, QUIC_CONN_SEND_FLAG_PATH_CHALLENGE);
+
+    return QUIC_STATUS_SUCCESS;
+}
+
+_IRQL_requires_max_(PASSIVE_LEVEL)
+static
+QUIC_STATUS
+QuicConnRemoveLocalAddress(
```
anrossi, line 6337: There doesn't appear to be any test coverage for the Remove address function here. Could you add a test for that?

src/core/path.c
```diff
@@ -221,6 +243,7 @@ QuicConnGetPathForPacket(
                 && QuicAddrGetFamily(&Packet->Route->RemoteAddress) == QuicAddrGetFamily(&Connection->Paths[i].Route.RemoteAddress)
                 && QuicAddrCompareIp(&Packet->Route->RemoteAddress, &Connection->Paths[i].Route.RemoteAddress)
                 && QuicAddrCompare(&Packet->Route->LocalAddress, &Connection->Paths[i].Route.LocalAddress)) {
+                QuicLibraryReleaseBinding(Connection->Paths[i].Binding);
```
anrossi, line 246: Where is this reference added?

anrossi (changes requested):
First of all, your changes look great, thanks for doing this!
I'd like to see some more tests. Since you're adding new APIs, could you add some negative test cases for `SetParam(QUIC_CONN_ADD_LOCAL_ADDRESS)`/`SetParam(QUIC_CONN_REMOVE_LOCAL_ADDRESS)`. For example, tests that add/remove an address that is not valid, add/remove an address before the connection is started, add the same address twice/remove the same address twice, a test that adds more addresses than the maximum path count, and one that adds an address that isn't on the system, but is valid, e.g. 100.64.10.20. 

In the Path tests, I'd appreciate a test that adds an address and removes the old one and verifies no traffic drops. If you have time, also a test that removes all addresses from a connection.

Thanks again!

src/core/path.c
masa-koz: In the case of server, QuicLibraryTryAddRefBinding() was called in QuicConnGetPathForPacket(). In the case of client, QuicBindingGetLocalAddress() was called in QuicConnAddLocalAddress().
By the way, the non-null check should be added.

src/core/binding.h
masa-koz: No. We can move source CIDs between old binding and new binding by setting QUIC_PARAM_CONN_LOCAL_ADDRESS. In this case, we will call QuicBindingAddAllSourceConnectionIDs() for new binding and QuicBindingRemoveAllSourceConnectionIDs() for old binding.

src/core/connection.c
masa-koz: I agree. I will change this.

anrossi:
@masa-koz Are you going to have time to add the requested extra tests? If not, do you mind if we add them later?

src/core/connection.c
masa-koz: Now, we defer sending a path challange if there is no unused ConnID. So, I've removed `QUIC_PARAM_CONN_LOCAL_UNUSED_DEST_CID_COUNT`.
And I've modified `Basic/WithProbePathArgs.ProbePath` for testing wheter a path challenge will be sent when the peer sends a new ConnID.

https://github.com/microsoft/msquic/pull/4218/commits/c6f7e3d7ecbb21b742095104834889d5937effc9

src/core/connection.c
masa-koz: Now, we can add local address before connection start. And I've added tests for adding/removing address.
https://github.com/microsoft/msquic/pull/4218/commits/2b3c6fd450f8a93746e3bfd8a03d00d7c34aeafc

src/core/connection.c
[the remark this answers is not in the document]
masa-koz: I`ve committed it in https://github.com/microsoft/msquic/pull/4218/commits/5da50d7f3b71670b503a8f62ce4dda3e6bc25a63

masa-koz:
@anrossi I'm sorry for my TOO late response.
I've added some tests for `SetParam(QUIC_CONN_ADD_LOCAL_ADDRESS)`/`SetParam(QUIC_CONN_REMOVE_LOCAL_ADDRESS)` in
https://github.com/microsoft/msquic/pull/4218/commits/2b3c6fd450f8a93746e3bfd8a03d00d7c34aeafc. Currently, not all the cases are covered. I will add them later.

src/test/bin/quic_gtest.h
```diff
@@ -929,3 +929,54 @@ std::ostream& operator << (std::ostream& o, const TlsConfigArgs& args) {
...
+
+std::ostream& operator << (std::ostream& o, const MigrationArgs& args) {
+    return o << (args.Family == 4 ? "v4" : "v6") << "/"
+        << (args.ShareBinding ? "ShareBinding" : "not ShareBinding") << "/"
+        << (args.Smooth ? "Smooth" : "not Smooth");
+}
+
+class WithMigrationArgs : public testing::Test,
+    public testing::WithParamInterface<MigrationArgs> {
+};
```
nibanks, line 982: nit: please leave trailing empty newlines

src/test/lib/ApiTest.cpp
```diff
@@ -4521,6 +4521,188 @@ void QuicTest_QUIC_PARAM_CONN_ORIG_DEST_CID(MsQuicRegistration& Registration, Ms
...
+                    Connection.Handle,
+                    ClientConfiguration,
+                    QUIC_ADDRESS_FAMILY_INET,
+                    "localhost",
+                    4433));
+
+            QuicAddr ClientLocalAddr;
+            TEST_QUIC_SUCCEEDED(Connection.GetLocalAddr(ClientLocalAddr));
+            TEST_QUIC_STATUS(
+                QUIC_STATUS_INVALID_STATE,
```
nibanks, line 4697: Can/should we really block removal of current active path? In the future, removal can/should be in response to the address being removed from the system itself (i.e. Wi-Fi gets disconnected). You can't say 'no, you're not allowed because I'm using it'. So, really, I think we need to allow removal, but if that's the last one, the connection just dies.

src/test/lib/PathTest.cpp
```diff
@@ -119,3 +119,241 @@ QuicTestLocalPathChanges(
...
+    }
+
+    TEST_QUIC_SUCCEEDED(Connection.Start(ClientConfiguration, ServerLocalAddr.GetFamily(), QUIC_TEST_LOOPBACK_FOR_AF(ServerLocalAddr.GetFamily()), ServerLocalAddr.GetPort()));
+    TEST_TRUE(Connection.HandshakeCompleteEvent.WaitTimeout(TestWaitTimeout));
+    TEST_TRUE(Context.HandshakeCompleteEvent.WaitTimeout(TestWaitTimeout));
+    TEST_NOT_EQUAL(nullptr, Context.Connection);
+
+    QuicAddr SecondLocalAddr;
+    TEST_QUIC_SUCCEEDED(Connection.GetLocalAddr(SecondLocalAddr));
+    SecondLocalAddr.IncrementPort();
```
nibanks, line 173: It's rare, but it's possible this port is used by something else in the system already. We need to handle this somehow.

src/test/lib/PathTest.cpp
```diff
@@ -119,3 +119,241 @@ QuicTestLocalPathChanges(
...
+    MsQuicConnection Connection(Registration, CleanUpManual, ClientCallback, &PeerStreamsChanged);
+    TEST_QUIC_SUCCEEDED(Connection.GetInitStatus());
+
+    if (ShareBinding) {
+        Connection.SetShareUdpBinding();
+    }
+
+    QuicAddr ClientLocalAddrs[4] = {QuicAddrFamily, QuicAddrFamily, QuicAddrFamily, QuicAddrFamily};
+    for (uint8_t i = 0; i < 4; i++) {
+        ClientLocalAddrs[i].SetPort(44433 + i);
```
nibanks, line 331: What happens if something is already using this?

src/inc/msquic.h
```diff
@@ -922,6 +922,8 @@ typedef struct QUIC_SCHANNEL_CREDENTIAL_ATTRIBUTE_W {
 #define QUIC_PARAM_CONN_STATISTICS_V2                   0x05000016  // QUIC_STATISTICS_V2
 #define QUIC_PARAM_CONN_STATISTICS_V2_PLAT              0x05000017  // QUIC_STATISTICS_V2
 #define QUIC_PARAM_CONN_ORIG_DEST_CID                   0x05000018  // uint8_t[]
+#define QUIC_PARAM_CONN_ADD_LOCAL_ADDRESS               0x05000019  // QUIC_ADDR
```
nibanks, line 925: Should we make these preview features?

nibanks (commented):
A few comments on the test code

src/test/lib/ApiTest.cpp
masa-koz: I agree with you. How should we handle the challenge? Separate PR?

nibanks:
This PR is big, so would it be possible to add more detailed description of what exactly is changed and why?

masa-koz:
@nibanks  I've updated the description. If any missing, please notify me.

src/core/binding.c
```diff
@@ -560,46 +560,100 @@ _IRQL_requires_max_(DISPATCH_LEVEL)
...
+        while (*Link1 != NULL) {
+            QUIC_CID_HASH_ENTRY* Entry1 = 
+                CXPLAT_CONTAINING_RECORD(
+                    *Link1,
+                    QUIC_CID_HASH_ENTRY,
+                    Link);
+            if (Entry1->Binding == Binding) {
+                QuicBindingRemoveSourceConnectionID(Binding, Entry1);
+                *Link1 = (*Link1)->Next;
+                CxPlatListPushEntry(&EntriesToFree, &Entry1->Link);
```
nibanks, line 632: Why have a separate list that you have to iterate again later to free? Just free here directly.

src/core/connection.c
```diff
@@ -6149,6 +6296,274 @@ QuicConnUpdatePeerPacketTolerance(
...
+    }
+
+    QuicTraceEvent(
+        ConnLocalAddrAdded,
+        "[conn][%p] New Local IP: %!ADDR!",
+        Connection,
+        CASTED_CLOG_BYTEARRAY(sizeof(Path->Route.LocalAddress), &Path->Route.LocalAddress));
+
+
+    QUIC_CID_LIST_ENTRY* NewDestCid = QuicConnGetUnusedDestCid(Connection);
```
nibanks, line 6370:
```suggestion
    QUIC_CID_LIST_ENTRY* NewDestCid = QuicConnGetUnusedDestCid(Connection);
```

src/core/connection.c
```diff
@@ -6617,6 +7055,28 @@ QuicConnParamSet(
         return QUIC_STATUS_SUCCESS;
     }
 
+    case QUIC_PARAM_CONN_ADD_LOCAL_ADDRESS: {
+
+        if (BufferLength != sizeof(QUIC_ADDR)) {
```
nibanks, line 7060: Let's move `QuicAddrIsValid(LocalAddress)` to here, instead of in `QuicConnAddLocalAddress`.

src/core/connection.c
```diff
@@ -6617,6 +7055,28 @@ QuicConnParamSet(
...
+            break;
+        }
+
+        Status = QuicConnAddLocalAddress(Connection, (QUIC_ADDR*)Buffer);
+        break;
+    }
+
+    case QUIC_PARAM_CONN_REMOVE_LOCAL_ADDRESS: {
+
+        if (BufferLength != sizeof(QUIC_ADDR)) {
```
nibanks, line 7071: Let's move `QuicAddrIsValid(LocalAddress)` to here, instead of in `QuicConnRemoveLocalAddress`.

src/core/lookup.c
```diff
@@ -179,16 +179,27 @@ QuicLookupRebalance(
...
                             Entry,
-                            QUIC_CID_HASH_ENTRY,
+                            QUIC_CID_SLIST_ENTRY,
                             Link);
-                    (void)QuicLookupInsertLocalCid(
-                        Lookup,
-                        CxPlatHashSimple(CID->CID.Length, CID->CID.Data),
-                        CID,
-                        FALSE);
+                    CXPLAT_SLIST_ENTRY* Entry1 = CID->HashEntries.Next;
```
nibanks, line 187:
```suggestion
                    CXPLAT_SLIST_ENTRY* HashEntry = CID->HashEntries.Next;
```

src/core/lookup.c
```diff
@@ -216,14 +227,14 @@ QuicLookupRebalance(
...
-                    QUIC_CID_HASH_ENTRY *CID =
+                    QUIC_CID_HASH_ENTRY* CID =
                         CXPLAT_CONTAINING_RECORD(
                             Entry,
                             QUIC_CID_HASH_ENTRY,
                             Entry);
                     (void)QuicLookupInsertLocalCid(
                         Lookup,
-                        CxPlatHashSimple(CID->CID.Length, CID->CID.Data),
+                        CxPlatHashSimple(CID->CID->CID.Length, CID->CID->CID.Data),
```
nibanks, line 237: `CID->CID->CID` is a bit ridiculous. 😄 We should figure out how to get some more descriptive names.

src/core/lookup.c
```diff
@@ -286,12 +298,24 @@ QuicCidMatchConnection(
...
-        const QUIC_CID_HASH_ENTRY* const Entry =
-            CXPLAT_CONTAINING_RECORD(Link, const QUIC_CID_HASH_ENTRY, Link);
+        const QUIC_CID_SLIST_ENTRY* const Entry =
+            CXPLAT_CONTAINING_RECORD(Link, const QUIC_CID_SLIST_ENTRY, Link);
+
+        for (CXPLAT_SLIST_ENTRY* Link1 = Entry->HashEntries.Next;
+            Link1 != NULL;
+            Link1 = Link1->Next) {
+
+            const QUIC_CID_HASH_ENTRY* const Entry1 =
```
nibanks, line 308:
```suggestion
            const QUIC_CID_HASH_ENTRY* const HashEntry =
```

src/test/lib/ApiTest.cpp
nibanks: Ideally I think we need to have it in this PR.

src/core/cid.h
```diff
@@ -163,12 +163,22 @@ typedef struct QUIC_CID_LIST_ENTRY {
 #define QUIC_CID_VALIDATE_NULL(Conn, Cid) UNREFERENCED_PARAMETER(Cid)
 #endif
 
+typedef struct QUIC_CID_SLIST_ENTRY {
+
+    CXPLAT_SLIST_ENTRY Link;
+    QUIC_CONNECTION* Connection;
```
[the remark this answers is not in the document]
masa-koz: No. The only use of this pointer was located in `QuicLookupAddLocalCid()`. Now, we directly pass the Connection pointer in calling `QuicLookupAddLocalCid()`.

src/core/lookup.c
masa-koz: I agree with you. Now, we use HashEntry instead of CID. And we renamed the pointer of `QUIC_CID_SLIST_ENTRY` in `QUIC_CID_HASH_ENTRY` from `CID` to `Parent`.

src/test/lib/ApiTest.cpp
masa-koz: OK. I'll try it.

src/core/connection.c
```diff
@@ -7062,7 +7053,14 @@ QuicConnParamSet(
             break;
         }
 
-        Status = QuicConnAddLocalAddress(Connection, (QUIC_ADDR*)Buffer);
+        QUIC_ADDR* LocalAddress = (QUIC_ADDR*)Buffer;
+
+        if (!QuicAddrIsValid(LocalAddress)) {
```
nibanks, line 7058: Personally, I'd have combined this with the other invalid parameter check (on the same line) as the one above. Looks a bit cleaner IMO.

src/test/lib/ApiTest.cpp
masa-koz: Now, we can remove a local address that belongs to the active path.

src/core/connection.c
masa-koz: Now, all the checks are done on the same line.

masa-koz:
@nibanks ping

nibanks:
> @nibanks ping

I will try to review this week. Thanks for your patience.

masa-koz:
@nibanks Could you review the latest change?

masa-koz:
@anrossi Could you review my change?

guhetier:
@masa-koz I'll try to review this in the coming days, but it might take me a bit. Thanks for your patience.

masa-koz:
@guhetier Any progress?

badeggg:
@masa-koz @guhetier Hi, before this pr, there is only one local address, `QUIC_PARAM_CONN_LOCAL_ADDRESS` will shift local address to a new one(I am not sure, please correct me if I miss something), after this pr, there will be multiple local addresses, how will `QUIC_PARAM_CONN_LOCAL_ADDRESS` work? Will it act like 'add a new address' and 'switch to the new address immediately'?

Seems `QUIC_PARAM_CONN_LOCAL_ADDRESS` and the new introduced `QUIC_PARAM_CONN_ADD_LOCAL_ADDRESS ` `QUIC_PARAM_CONN_REMOVE_LOCAL_ADDRESS` have overlap functions. After introducing the two new parameters, will `QUIC_PARAM_CONN_LOCAL_ADDRESS` be deprecated gradually?

If `QUIC_PARAM_CONN_LOCAL_ADDRESS` will be deprecated, how will the app do 'add a new address and switch to the new address immediately'? Since [the original post](https://github.com/microsoft/msquic/issues/1946#issue-981307126) mentions that 'The app shouldn't need to add it and then set it; it can just add it if it wants to immediately switch it a new one.'

I'm just an enthusiast of quic and msquic.

PR #4222, opened 2024-04-02
Add logic to remove any logging residue. 

ProjectsByJackHe:
## Description

Thus far for secnetperf runs, we have relied on the Github runner agent to cleanup any managed directories pertaining to github actions. 

However, MsQuic modifies directories outside of the ones managed by Github Actions. 

Specifically, for logs, we know the /artifacts/logs gets cleaned up after every run, but /AppData/Local/Temp remains. 

So we must clean them up too. 

## Testing

CI 

## Documentation

N/A

### DIFF
--- a/scripts/secnetperf-helpers.psm1
+++ b/scripts/secnetperf-helpers.psm1
@@ -224,10 +224,10 @@ function Cleanup-State {
     }
 
     # Clean up any ETL residue.
-    try { .\scripts\log.ps1 -Stop -OutputPath "$artifactDir/client" -RawLogOnly }
+    try { .\scripts\log.ps1 -Cancel }
     catch { Write-Host "Failed to stop logging on client!" }
     Invoke-Command -Session $Session -ScriptBlock {
-        try { & "$Using:RemoteDir/scripts/log.ps1" -Stop -OutputPath "$Using:remoteArtifactDir/server" -RawLogOnly }
+        try { & "$Using:RemoteDir/scripts/log.ps1" -Cancel }
         catch { Write-Host "Failed to stop logging on server!" }
     }
 }
--- a/scripts/secnetperf-helpers.psm1
+++ b/scripts/secnetperf-helpers.psm1
@@ -224,14 +224,12 @@ function Cleanup-State {
     }
 
     # Clean up any ETL residue.
-    try {
-        if ($isWindows) {
-            $homeDirectory = $env:USERPROFILE
-        } else {
-            $homeDirectory = $env:HOME
-        }
-        Remove-Item $homeDirectory/AppData/Local/Temp -Recurse -Force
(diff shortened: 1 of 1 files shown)


scripts/secnetperf-helpers.psm1
```diff
@@ -222,6 +222,14 @@ function Cleanup-State {
             if ($null -ne (Get-Service msquicpriv -ErrorAction Ignore)) { throw "msquicpriv still running remotely!" }
         }
     }
+
+    # Clean up any ETL residue.
+    try { .\scripts\log.ps1 -Cancel }
```
nibanks, line 227: I'm wondering if your first commit is still a good idea. Maybe the `-Cancel` should also clean up old files too in that directory (in log.ps1)?

scripts/secnetperf-helpers.psm1
```diff
@@ -222,6 +222,14 @@ function Cleanup-State {
             if ($null -ne (Get-Service msquicpriv -ErrorAction Ignore)) { throw "msquicpriv still running remotely!" }
         }
     }
+
+    # Clean up any ETL residue.
```
nibanks, line 226: Another thought: Should we do this **before** running tests in addition? In case a previous test crashed...

scripts/secnetperf-helpers.psm1
ProjectsByJackHe: Right, I agree we should contain the cleanup logic in 1 place. I can open a separate PR for log.ps1.

scripts/secnetperf-helpers.psm1
ProjectsByJackHe: I don't think we need to go that far. This only became a problem after accumulating ETLs for months, and with this PR we would always cleanup every run. Making it every test I don't think is necessary.

scripts/secnetperf-helpers.psm1
ProjectsByJackHe: https://github.com/microsoft/msquic/pull/4223

scripts/secnetperf-helpers.psm1
nibanks: We only clean up if previous doesn't crash. If you run a test and it crashes, you never cleanup.

ProjectsByJackHe: That's true, I meant I feel like its ok to let a little residual buildup. If you run a test, it crashes, you don't cleanup, but the buildup is minimal (this only became a problem after accumulating ETLs for a long time). Then on the next workflow run, it cleans up whatever is left from the last. I don't want this to be a blocker, though. I can add that additional cleanup as well.

nibanks: We can start with this and go from there.

nibanks (approved):

PR #4243, opened 2024-04-15
Bug Fix 

cpestano-m:
## Description

Bug fix, when running on some windows systems file "prepare-machine.ps1" Invoke-WebRequest will fail if output directory "scripts/artifacts" does not exist, fixed by adding explicit mkdir for artifacts directory.

## Testing

No tests needed.

## Documentation

No documentation change.

### DIFF
--- a/scripts/prepare-machine.ps1
+++ b/scripts/prepare-machine.ps1
@@ -257,6 +257,7 @@ function Install-NASM {
     if (!(Test-Path $NasmExe) -and $env:GITHUB_PATH -eq $null) {
         Write-Host "Downloading NASM"
         $NasmArch = "win64"
+        mkdir artifacts
         if (![System.Environment]::Is64BitOperatingSystem) { $NasmArch = "win32" }
         try {
             Invoke-WebRequest -Uri "https://www.nasm.us/pub/nasm/releasebuilds/$NasmVersion/win64/nasm-$NasmVersion-$NasmArch.zip" -OutFile "artifacts\nasm.zip"


scripts/prepare-machine.ps1
```diff
@@ -257,6 +257,7 @@ function Install-NASM {
     if (!(Test-Path $NasmExe) -and $env:GITHUB_PATH -eq $null) {
         Write-Host "Downloading NASM"
         $NasmArch = "win64"
+        mkdir artifacts
```
nibanks, line 260: This should already be happening on line 165 above, correct?
```PowerShell
if (!(Test-Path $ArtifactsPath)) { mkdir $ArtifactsPath | Out-Null }
```

cpestano-m: I would assume so but when on first run that line didn't fire and gave me this error.

Invoke-WebRequest: C:\Users\cpestano\Desktop\msquic\scripts\prepare-machine.ps1:265
Line |
 265 |              Invoke-WebRequest -Uri "https://fossies.org/windows/misc/ …
     |              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Could not find a part of the path 'C:\Users\cpestano\Desktop\msquic\scripts\artifacts\nasm.zip'.

cpestano-m: Figured out the issue, the artifacts dir is created in the parent msquic directory by line 165. But if user is cd'ed into the scripts directory to run prepare-machine.ps1, artifacts will be created in msquic but Invoke-WebRequest expects the artifacts dir to be in scripts since the . dir that the user is cd'ed into is scripts. Do you think this warrants a change or no?

nibanks: I think it's fine to fix bugs around where you run the script from. Most of the time we do run scripts from the root directory, but the scripts should also work from the scripts path. This likely means there is a problem with relative paths being used where they shouldn't be.

PR #4256, opened 2024-04-23
Fix Order of TCP Accept and Receive Operations

nibanks:
## Description

Fixes a bug in the CxPlat abstraction around the ordering of TCP accept and receive callbacks.

## Testing

Ran locally, testing with CI/CD for the rest.

## Documentation

N/A

### DIFF
--- a/src/platform/datapath_winuser.c
+++ b/src/platform/datapath_winuser.c
@@ -2646,15 +2646,15 @@ CxPlatDataPathSocketProcessAcceptCompletion(
             goto Error;
         }
 
-        CxPlatDataPathStartReceiveAsync(AcceptSocketProc);
-
-        AcceptSocketProc->IoStarted = TRUE;
         Datapath->TcpHandlers.Accept(
             ListenerSocketProc->Parent,
             ListenerSocketProc->Parent->ClientContext,
             ListenerSocketProc->AcceptSocket,
             &ListenerSocketProc->AcceptSocket->ClientContext);
         ListenerSocketProc->AcceptSocket = NULL;
+
+        CxPlatDataPathStartReceiveAsync(AcceptSocketProc);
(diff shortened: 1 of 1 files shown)


csujedihy:
What was the outcome of this bug?

nibanks:
> What was the outcome of this bug?

Receive callbacks were happening before Accept had been fully executed, effectively meaning the connection was used before fully initialized.

csujedihy (dismissed):

ProjectsByJackHe:
Lots of BVT crashes are happening due to an Assert failure in `CxPlatSocketEnqueueSqe` 
image

nibanks:
> Lots of BVT crashes are happening due to an Assert failure in `CxPlatSocketEnqueueSqe` image

Can you share the full stack?

ProjectsByJackHe:
> > Lots of BVT crashes are happening due to an Assert failure in `CxPlatSocketEnqueueSqe` image
> 
> Can you share the full stack?

image



Here is the crash dump I was looking at: https://github.com/microsoft/msquic/actions/runs/8805206873/artifacts/1440546807

ProjectsByJackHe (approved):
LGTM

ami-GS:
I hit this error by TcpConnect test.
when accept callback takes long, but main thread delete the socket (goes to `CxPlatSocketContextUninitialize`), The socket is deleted with error by `if (!SocketProc->IoStarted)` which set `SocketProc->Uninitialized = TRUE;`
This should be same erorr as Jack mentioned.

Might need to do below?
```
lock {
Accept callback;
IoStarted = TRUE;
}

------

lock {
    if (!SocketProc->IoStarted) {
   ....
    }
}
```

nibanks:
Yeah, maybe.

PR #4262, opened 2024-04-25
Enable secnetperf with xdp on azure VM

ami-GS:
## Description

Few fix to use secnetperf with xdp on two physically separated machines.

## Testing

Passed with command
`sudo pwsh ./scripts/secnetperf.ps1 -MsQuicCommit main -environment azure -plat linux -os ubuntu-22.04 -arch x64 -tls openssl3 -UserName root -io xdp`

## Documentation
N/A


scripts/secnetperf.ps1
```diff
@@ -225,19 +232,29 @@ if ($isWindows) {
...
+if ($useXDP -and $isWindows) { Install-XDP $Session $RemoteDir }
 if ($io -eq "wsk") { Install-Kernel $Session $RemoteDir $SecNetPerfDir }
 
 if (!$isWindows) {
     # Make sure the secnetperf binary is executable.
     Write-Host "Updating secnetperf permissions"
+    $GRO = "on"
+    $ENABLE_XDP=0
+    if ($io -eq "xdp") {
+        $GRO = "off"
```
nibanks, line 244: Why do you need to disable GRO to use XDP?

scripts/secnetperf.ps1
```diff
@@ -225,19 +232,29 @@ if ($isWindows) {
...
         $env:LD_LIBRARY_PATH = "${env:LD_LIBRARY_PATH}:$Using:RemoteDir/$Using:SecNetPerfDir"
         chmod +x "$Using:RemoteDir/$Using:SecNetPerfPath"
+        sudo sh -c "ethtool -K eth0 generic-receive-offload $Using:GRO"
+        $env:MSQUIC_ENABLE_XDP=$Using:ENABLE_XDP
     }
     $fullPath = Repo-Path $SecNetPerfDir
     $env:LD_LIBRARY_PATH = "${env:LD_LIBRARY_PATH}:$fullPath"
     chmod +x "./$SecNetPerfPath"
+    sudo sh -c "ethtool -K eth0 generic-receive-offload $GRO"
+    $env:MSQUIC_ENABLE_XDP=$ENABLE_XDP
```
nibanks, line 257: Reset this in cleanup?

src/platform/datapath_raw_xdp_linux.c
```diff
@@ -357,7 +357,7 @@ AttachXdpProgram(struct xdp_program *Prog, XDP_INTERFACE *Interface, struct xsk_
         unsigned int xdp_flag;
     } AttachTypePairs[]  = {
         // { XDP_MODE_HW, XDP_FLAGS_HW_MODE },
-        { XDP_MODE_NATIVE, XDP_FLAGS_DRV_MODE },
+        // { XDP_MODE_NATIVE, XDP_FLAGS_DRV_MODE },
```
nibanks, line 360: Why?

ami-GS: DRV mode is not fully functional on Azure VM, but attachable to NIC without error. Normally there should be error if kenrel/hw doesn't support some features, then we can fallback to lower grade mode.
I'm asking accelerated networking team how to programmatically identify the functionality

scripts/secnetperf.ps1
ami-GS: cleanup phase (beginning of secnetperf.ps1) cannot know the default state of GRO.
So  I specify the feature off/on for xdp/others.

scripts/secnetperf.ps1
ami-GS: Decryption in connection.c start failing with XDP & GRO.
Maybe need additional implementation to split the merged chunks?

ami-GS: GRO might not work with generic XDP? (SKB_MODE)
image
https://conferences.sigcomm.org/sigcomm/2023/files/workshop-ebpf/1-CloudNative.pdf

and LRO (HW_GRO) does not work with XDP?
https://www.infradead.org/~mchehab/kernel_docs/networking/device_drivers/microsoft/netvsc.html

ami-GS: might be supported? need investigation. kernel version, library version, host/guest driver
https://lpc.events/event/11/contributions/939/attachments/771/1551/xdp-multi-buff.pdf

nibanks: I didn't see you disable GRO for all the other tests. Also, can't you programmatically disable it for your XDP socket instead of doing it system-wide?

scripts/secnetperf.ps1
nibanks: Is there no way to query the state at the beginning and save it? But really, as I said in the comment above, we should figure out a programmatic solution for just the XDP socket, instead of a system-wide setting.

src/platform/datapath_raw_xdp_linux.c
nibanks: Ok.

scripts/secnetperf.ps1
ami-GS: I realized that GRO in duonic is default off. And duonic support DRV mode attach.
DRV (driver) mode can reside with GRO because packet redirection by XDP program is "before" GRO.
DRV mode XDP doesn't GRO actually.
I tried 4 combination with duonic, SKB_MODE/DRV_MODE and GRO on/off. However the maximum ethernet frame size is always 1514. I don't actually see GROed packet. I saw larger packets on eth0 on Azure VM with GRO on then decryption failure.

As far as I checked ethtool code and document, it opens another socket by `socket(AF_NETLINK, SOCK_RAW, NETLINK_GENERIC);`. Then send payload, which is not well documented though, to kernel to set/unset flags. Document says that netlink feature is kind of ioctl so that netlink operation should be for each device (eth0 on Azure VM).
This socket (AF_NETLINK) is not associated with XDP socket. I believe it is difficult to disable/enable for each socket level. It is technically possible to do programmatically for system-wide, but it is kind of re-inventing the wheel of ethtool in msquic

scripts/secnetperf-helpers.psm1
```diff
@@ -228,6 +228,12 @@ function Cleanup-State {
             try { & "$Using:RemoteDir/scripts/log.ps1" -Cancel }
             catch { Write-Host "Failed to stop logging on server!" }
         }
+    } else {
+        # iterate all interface and "ip link set ${iface} xdp off"
+        $ifaces = ip link show | grep -oP '^\d+: \K[\w@]+' | cut -d'@' -f1
+        foreach ($iface in $ifaces) {
+            sudo ip link set $iface xdp off
```
nibanks, line 235: Can these commands fail? Should they be in a `try`/`catch` so one failure doesn't stop us from doing the rest?

ami-GS: Doesn't fail. "off" always succeeds

scripts/secnetperf.ps1
ami-GS: Possible to just query, but need external file to save the state for the case if the workload is killed during the run.

scripts/secnetperf-helpers.psm1
```diff
@@ -511,7 +529,7 @@ function Invoke-Secnetperf {
 
     # Start the server running.
     "> secnetperf $serverArgs" | Add-Content $serverOut
-    $job = Start-RemoteServer $Session "$RemoteDir/$SecNetPerfPath $serverArgs"
+    $job = Start-RemoteServer $Session "$RemoteDir/$SecNetPerfPath $serverArgs" (!$IsWindows -and $io -eq "xdp")
```
nibanks, line 532: 1. Since this condition never changes, let's make a `$sudo` variable above for this, equal to "" or "sudo MSQUIC_ENABLE_XDP=1".
2. Instead of changing the functions above, just inject `$sudo` here and below.
```suggestion
    $job = Start-RemoteServer $Session "$sudo $RemoteDir/$SecNetPerfPath $serverArgs"
```
Example change below in other comment.

scripts/secnetperf-helpers.psm1
```diff
@@ -521,7 +539,7 @@ function Invoke-Secnetperf {
         Write-Host "==============================`nRUN $($try+1):"
         "> secnetperf $clientArgs" | Add-Content $clientOut
         try {
-            $process = Start-LocalTest $clientPath $clientArgs $artifactDir
+            $process = Start-LocalTest $clientPath $clientArgs $artifactDir (!$IsWindows -and $io -eq "xdp")
```
nibanks, line 542:
```suggestion
            $process = Start-LocalTest "$sudo$clientPath" $clientArgs $artifactDir (!$IsWindows -and $io -eq "xdp")
```

scripts/secnetperf-helpers.psm1
```diff
@@ -291,6 +297,7 @@ function Start-LocalTest {
     } else {
         # We use bash to execute the test so we can collect core dumps.
         $pinfo.FileName = "bash"
+        $sudo = ""
```
nibanks, line 300: Unnecessary?

scripts/secnetperf-helpers.psm1
```diff
@@ -483,6 +490,11 @@ function Invoke-Secnetperf {
         continue
     }
 
+    $sudo=""
```
nibanks, line 493:
```suggestion
    $sudo = ""
```

scripts/secnetperf-helpers.psm1
```diff
@@ -483,6 +490,11 @@ function Invoke-Secnetperf {
...
+    if ((!$IsWindows -and $io -eq "xdp")) {
```
nibanks, line 494:
```suggestion
    if (!$IsWindows -and $io -eq "xdp") {
```

scripts/secnetperf-helpers.psm1
```diff
@@ -521,11 +533,14 @@ function Invoke-Secnetperf {
...
         "> secnetperf $clientArgs" | Add-Content $clientOut
         try {
-            $process = Start-LocalTest $clientPath $clientArgs $artifactDir
+            $process = Start-LocalTest "$sudo$clientPath" $clientArgs $artifactDir
             $rawOutput = Wait-LocalTest $process $artifactDir ($io -eq "wsk") 30000
             Write-Host $rawOutput
             $values[$tcp] += Get-TestOutput $rawOutput $metric
             if ($extraOutput) {
+                if ($sudo -ne "") {
+                    sudo chown $UserName $extraOutput
```
nibanks, line 542: Why is this necessary?

ami-GS: The extraOutput (latency.txt) is owned by root when running secnetperf with sudo. deleting the file from normal user get failed. Existence of latency.txt make next run fail

scripts/secnetperf-helpers.psm1
```diff
@@ -483,6 +489,9 @@ function Invoke-Secnetperf {
         continue
     }
 
+     # Linux XDP requires env var and sudo for now
+    $sudo = (!$IsWindows -and $io -eq "xdp") ? "sudo MSQUIC_ENABLE_XDP=1 " : ""
```
nibanks, line 493: Daiki, you had `sudo -E` here, but I replaced it with the above because I thought that was actually what you needed, right? Please confirm before merging.

nibanks: Oh, I see, you do it in the other script with `$env:`. I'll change back to `-E`

scripts/secnetperf-helpers.psm1
```diff
@@ -483,6 +489,9 @@ function Invoke-Secnetperf {
         continue
     }
 
+     # Linux XDP requires sudo for now
+    $sudo = (!$IsWindows -and $io -eq "xdp") ? "sudo -E " : ""
```
nibanks, line 493: Can you explain why the `-E` needs to be here? As I understand it, it gives edit permissions? Is that for the output artifacts?

scripts/secnetperf.ps1
```diff
@@ -235,19 +242,29 @@ if ($isWindows) {
 Configure-DumpCollection $Session
 
 # Install any dependent drivers.
-if ($useXDP) { Install-XDP $Session $RemoteDir }
+if ($useXDP -and $isWindows) { Install-XDP $Session $RemoteDir }
```
nibanks, line 245: This just made me realize, we won't be able to differentiate between 'Windows QUIC + XDP' and 'Linux QUIC + XDP' because we name them 'QUIC+XDP' in the data file (I think). @ProjectsByJackHe do we need to make the `$io` names unique?

nibanks (approved):

scripts/secnetperf.ps1
```diff
@@ -235,19 +242,29 @@ if ($isWindows) {
...
         $env:LD_LIBRARY_PATH = "${env:LD_LIBRARY_PATH}:$Using:RemoteDir/$Using:SecNetPerfDir"
+        $env:MSQUIC_ENABLE_XDP=$Using:ENABLE_XDP
         chmod +x "$Using:RemoteDir/$Using:SecNetPerfPath"
+        sudo sh -c "ethtool -K eth0 generic-receive-offload $Using:GRO"
     }
     $fullPath = Repo-Path $SecNetPerfDir
     $env:LD_LIBRARY_PATH = "${env:LD_LIBRARY_PATH}:$fullPath"
+    $env:MSQUIC_ENABLE_XDP=$ENABLE_XDP
     chmod +x "./$SecNetPerfPath"
+    sudo sh -c "ethtool -K eth0 generic-receive-offload $GRO"
```
nibanks, line 267: This seems to be failing: https://github.com/microsoft/netperf/actions/runs/8867963327/job/24347253651
```
Error: ethtool: bad command line argument(s)

OriginInfo            : netperf-peer
Exception             : System.Management.Automation.RemoteException: ethtool: 
                        bad command line argument(s)
TargetObject          : ethtool: bad command line argument(s)
CategoryInfo          : NotSpecified: (ethtool: bad command line argument(s):St
                        ring) [], RemoteException
FullyQualifiedErrorId : NativeCommandError
ErrorDetails          : 
InvocationInfo        : System.Management.Automation.InvocationInfo
ScriptStackTrace      : at <ScriptBlock>, <No file>: line 6
PipelineIterationInfo : {}

ethtool: bad command line argument(s)
```

scripts/secnetperf-helpers.psm1
ami-GS: sudo reset environment variables.
If my understanding is correct, LD_LIBRARY_PATH is not propagated to sudo execution.

nibanks:
I see this on the 20.04 lab machine run of netperf ([link](https://github.com/microsoft/netperf/actions/runs/8873638329/job/24361315839)):
```
sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
```

ami-GS:
netperf repo is monitoring each msquic PRs?

nibanks:
> netperf repo is monitoring each msquic PRs?

Yep. Here is the latest run for your PR: https://github.com/microsoft/netperf/actions/runs/8882454145

src/platform/datapath_xplat.c
```diff
@@ -49,6 +49,7 @@ CxPlatDataPathInitialize(
     char* EnableXdp = getenv("MSQUIC_ENABLE_XDP");
     if (EnableXdp != NULL && EnableXdp[0] == '1') {
 #endif
+    if (Config && Config->Flags & QUIC_EXECUTION_CONFIG_FLAG_XDP) {
```
nibanks, line 52: Please update Windows XDP to use this flag too.

src/platform/datapath_xplat.c
```diff
@@ -49,6 +49,7 @@ CxPlatDataPathInitialize(
     char* EnableXdp = getenv("MSQUIC_ENABLE_XDP");
```
nibanks, line 49: Please remove env variable usage, and make sure Windows and Linux both use the flag below.

src/inc/msquic.h
```diff
@@ -268,6 +268,7 @@ typedef enum QUIC_EXECUTION_CONFIG_FLAGS {
     QUIC_EXECUTION_CONFIG_FLAG_QTIP             = 0x0001,
     QUIC_EXECUTION_CONFIG_FLAG_RIO              = 0x0002,
 #endif
+    QUIC_EXECUTION_CONFIG_FLAG_XDP              = 0x0004,
```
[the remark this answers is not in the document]
ami-GS: DataPathTest.cpp cannot see `QUIC_API_ENABLE_PREVIEW_FEATURES`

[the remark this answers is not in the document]
nibanks: It needs to set `QUIC_API_ENABLE_PREVIEW_FEATURES` then. This needs to be hidden for now.

src/tools/spin/spinquic.cpp
```diff
@@ -1538,6 +1538,20 @@ void start() {
...
+            QUIC_EXECUTION_CONFIG ExecutionConfig { QUIC_EXECUTION_CONFIG_FLAG_XDP, 10000, 0, {0} };
+            if (QUIC_FAILED(
+            Status =
+            MsQuic->SetParam(
+                nullptr,
+                QUIC_PARAM_GLOBAL_EXECUTION_CONFIG,
+                sizeof(ExecutionConfig),
+                &ExecutionConfig))) {
+                printf("Failed to set execution config %d\n", Status);
+            }
```
nibanks, line 1552:
```suggestion
            if (QUIC_FAILED(
                Status =
                MsQuic->SetParam(
                    nullptr,
                    QUIC_PARAM_GLOBAL_EXECUTION_CONFIG,
                    sizeof(ExecutionConfig),
                    &ExecutionConfig))) {
                printf("Failed to set execution config %d\n", Status);
            }
```

src/platform/unittest/DataPathTest.cpp
```diff
@@ -18,6 +18,7 @@
 #endif
 
 extern bool UseDuoNic;
+#define QUIC_EXECUTION_CONFIG_FLAG_XDP (QUIC_EXECUTION_CONFIG_FLAGS)0x0004
```
nibanks, line 21: Instead of this, simply define `QUIC_API_ENABLE_PREVIEW_FEATURES` above, before all the includes.

nibanks (dismissed):

nibanks (approved):

src/platform/unittest/DataPathTest.cpp
```diff
@@ -427,7 +427,9 @@ QuicAddr DataPathTest::UnspecIPv4;
 QuicAddr DataPathTest::UnspecIPv6;
 
 struct CxPlatDataPath {
-    QUIC_EXECUTION_CONFIG DefaultExecutionConfig { QUIC_EXECUTION_CONFIG_FLAG_NONE, 10000, 0, {0} };
+    static uint8_t RawConfig[QUIC_EXECUTION_CONFIG_MIN_SIZE + 256 * sizeof(uint16_t)] {0};
+    static QUIC_EXECUTION_CONFIG* DefaultExecutionConfig;
```
nibanks, line 431: I don't understand why you made this change. The old stuff should work. You definitely don't need to fully specify the list of processors.

ami-GS: `PollingIdleTimeoutUs == 10000` is for QTIP only?

ami-GS: OK this value is the root cause

nibanks (approved):

PR #4264, opened 2024-04-26
Fix broken build on Alpine

ManickaP:
lsb-release is not present on Alpine, whereas os-release is on most distributions.


ManickaP:
This is still not enough. Since the build kind of assumes Ubuntu as the only distro and enables XDP everywhere except Ubuntu 20.04: 
https://github.com/microsoft/msquic/blob/5052f346237bd53907b227aa7a10cd638be9dcf3/src/platform/CMakeLists.txt#L38

Which leads to:
```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
BPF_LIB
    linked by target "platform" in directory /tmp/msquic/src/platform
NL_LIB
    linked by target "platform" in directory /tmp/msquic/src/platform
NL_ROUTE_LIB
    linked by target "platform" in directory /tmp/msquic/src/platform
XDP_LIB
    linked by target "platform" in directory /tmp/msquic/src/platform
```
on Alpine.

src/platform/CMakeLists.txt
```diff
@@ -27,16 +27,11 @@ else()
...
         set(SOURCES ${SOURCES} datapath_linux.c datapath_epoll.c)
-        if ((${UBUNTU_VERSION} STREQUAL "20.04") OR
-            (${CMAKE_TARGET_ARCHITECTURE} STREQUAL "arm64") OR
-            (${CMAKE_TARGET_ARCHITECTURE} STREQUAL "arm") OR
-            (${CMAKE_SYSTEM_PROCESSOR} STREQUAL "arm") OR
-            (ANDROID) OR
-            (QUIC_CARGO_BUILD))
-            set(SOURCES ${SOURCES} datapath_xplat.c datapath_raw_dummy.c)
-        else()
+        if ("${UBUNTU_VERSION}" STREQUAL "22.04")
```
ManickaP, line 30: I turned the condition around so that XDP will light up only on Ubuntu 22.04.
But I think it would be better to check for XDP support in a distribution agnostic way.
cc @wfurt

ami-GS:
Thanks for the change. As far as I remember, our docker environment doesn't have /etc/os-release. let me double check

ami-GS:
Exists!😵‍💫

src/platform/CMakeLists.txt
ami-GS: Could you add below? (not sure right syntax). only allowing Ubuntu22.04 and x64 (amd64) for now.
`( ${CMAKE_TARGET_ARCHITECTURE} STREQUAL "x64") OR ${CMAKE_SYSTEM_PROCESSOR} STREQUAL "x64") ) AND !(ANDROID) AND !(QUIC_CARGO_BUILD)`

wfurt: I generally feel putting version numbers here is bad idea. Versions do change and how about Debian and other distributions? 

My suggestion would be to parametrize the build ... something like `--enable-xdp` and put the logic into pipeline part e.g. it would work in CI but the make logic would not have all the glory details. 

And it would be much more friendly to any other distribution IMHO.

ManickaP: I like @wfurt idea, @ami-GS are you ok with me adding a new argument to the build?

nibanks: I'm fine with that model.

scripts/build.ps1
```diff
@@ -166,6 +169,9 @@ param (
     [Parameter(Mandatory = $false)]
     [switch]$PGO = $false,
 
+    [Parameter(Mandatory = $false)]
+    [switch]$EnableXDP = $false,
```
nibanks, line 173: XDP is **always** supported on Windows. This flag is specific to Linux. Let's calls this `EnableLinuxXDP`. Similarly, let's call the cmake flag `QUIC_LINUX_XDP_ENABLED`.

nibanks: ...and, this eliminates the need for the separate `QUIC_CARGO_BUILD` flag. Please remove that and fix up the existing usage.

Finally, we need to actually use the `EnableLinuxXDP` flag in our CI jobs.

ManickaP:
I renamed the parameter, updated the pipeline (hopefully in the right place) and removed `QUIC_CARGO_BUILD`, PTAL.

nibanks:
Everything looks good to me, but @ami-GS can you please take a look? I noticed [netperf](https://github.com/microsoft/netperf/actions/runs/9061600291/job/24893830011) failed, but I wonder if this is because of something else being broken? I know you're working on that in a parallel PR.

nibanks (dismissed):

nibanks:
What should we do for the Official, OneBranch builds? Enabled or disabled?

ami-GS:
rerunning BVT xdp test...

wfurt (dismissed):
LGTM

nibanks (dismissed):

ManickaP:
BTW, I don't have rights to merge this, so unless there is more feedback to address (let me know if there still something is), feel free to merge on my behalf. And thanks all for help with this!

docs/BUILD.md
```diff
@@ -215,22 +215,22 @@ sudo ./artifacts/bin/linux/x64_Debug_openssl3/msquictest --duoNic
 ```
 
 **Q&A**
-- Q: Is this workload really running on XDP?
```
ami-GS, line 218: This as well

docs/BUILD.md
```diff
@@ -215,22 +215,22 @@ sudo ./artifacts/bin/linux/x64_Debug_openssl3/msquictest --duoNic
...
 docker0                <No XDP program loaded!>
-duo2                   xdp_dispatcher    native   608211 4d7e87c0d30db711 
+duo2                   xdp_dispatcher    native   608211 4d7e87c0d30db711
  =>              50     xdp_main                  608220 c8fcabdd9e3895f3  XDP_PASS
-duo1                   xdp_dispatcher    native   608225 4d7e87c0d30db711 
+duo1                   xdp_dispatcher    native   608225 4d7e87c0d30db711
  =>              50     xdp_main                  608228 c8fcabdd9e3895f3  XDP_PASS
 ```
 
-- Q: Is Ubuntu 20.04LTS supported?
```
ami-GS, line 233: this

docs/BUILD.md
```diff
@@ -182,11 +182,11 @@ sudo dnf install libatomic
 ```
 
 #### Linux XDP
-Linux XDP is experimentally supported on amd64 && Ubuntu 22.04LTS.
```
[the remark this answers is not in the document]
ManickaP: Sorry, that's my editor that removes white spaces from all EOLs. Reverted all WS changes.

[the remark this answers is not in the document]
nibanks: I was talking to Daiki on IM. IMO, we should use extra newlines instead of trailing spaces. Editors should remove trailing spaces.

ami-GS (approved):

PR #4271, opened 2024-04-29
pkg_config setting for arm

ami-GS:
## Description

Some install deps using prepare-machines.ps1, some are not.

## Testing

see automation

## Documentation

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -401,7 +401,6 @@ function CMake-Generate {
             switch ($Arch) {
                 "arm"   { $env:PKG_CONFIG_PATH="$SysRoot/usr/lib/arm-linux-gnueabihf/pkgconfig" }
                 "arm64" { $env:PKG_CONFIG_PATH="$SysRoot/usr/lib/aarch64-linux-gnu/pkgconfig" }
-                default { $env:PKG_CONFIG_PATH="$SysRoot/usr/lib/x86_64-linux-gnu/pkgconfig" }
             }
        }
     }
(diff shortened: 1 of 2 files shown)


nibanks:
This PR didn't fix the problem. I don't think it's about XDP dependencies. The issue is more about architecture stuff with the openssl build.

ami-GS:
Maybe this is using newer docker image, but using older prepare-machine.ps1/build.ps1 which causes pkg_config side effect

ami-GS:
Ready to go👌

scripts/build.ps1
```diff
@@ -396,6 +396,12 @@ function CMake-Generate {
                 "arm64" { $Arguments += " -DCMAKE_CXX_COMPILER_TARGET=aarch64-linux-gnu -DCMAKE_C_COMPILER_TARGET=aarch64-linux-gnu -DCMAKE_TARGET_ARCHITECTURE=arm64" }
                 "arm" { $Arguments += " -DCMAKE_CXX_COMPILER_TARGET=arm-linux-gnueabihf  -DCMAKE_C_COMPILER_TARGET=arm-linux-gnueabihf -DCMAKE_TARGET_ARCHITECTURE=arm" }
             }
+
+            # to build with pkg-config
+            switch ($Arch) {
```
nibanks, line 401: Shouldn't this just be part of the switch right above this?

nibanks:
Can you please update the title/description to match the change now? Thanks!

scripts/build.ps1
ami-GS: It becomes diverging the change between main branch. is it okey?

PR #4276, opened 2024-05-02
Fail on Stream Blocked Immediate

nibanks:
## Description

A prototype change to update the `StreamStart` API to fail inline if the app passes `QUIC_STREAM_START_FLAG_FAIL_BLOCKED`.

## Testing

CI/CD

## Documentation

TODO

### DIFF
--- a/src/core/stream_set.c
+++ b/src/core/stream_set.c
@@ -49,6 +49,7 @@ QuicStreamSetInitialize(
     )
 {
     CxPlatListInitializeHead(&StreamSet->ClosedStreams);
+    CxPlatDispatchLockInitialize(&StreamSet->TypesLock);
 #if DEBUG
     CxPlatListInitializeHead(&StreamSet->AllStreams);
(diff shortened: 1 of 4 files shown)


src/core/stream_set.c
```diff
@@ -191,8 +193,10 @@ QuicStreamSetReleaseStream(
     uint8_t Flags = (uint8_t)(Stream->ID & STREAM_ID_MASK);
     QUIC_STREAM_TYPE_INFO* Info = &StreamSet->Types[Flags];
 
+    CxPlatDispatchLockAcquire(&StreamSet->TypesLock);
```
nibanks, line 196: I _think_ this is the only other synchronization we'd need (besides in `QuicStreamSetNewLocalStreamID` below), but it's a bit tricky.

nibanks:
FYI, since this is a behavior change for 'fail on block' stream start calls, the down-level tests are failing. Might have to create a new flag to the effect of `FAIL_ON_BLOCKED_INLINE`.

PR #4277, opened 2024-05-02
Always set execution config for 'no sleep'

ami-GS:
## Description

sleeping with normal socket, but no sleep with special socket (feature).

## Testing

see netperf?

## Documentation

N/A


nibanks:
> sleeping with normal socket, but no sleep with special socket (feature).

Why?

ami-GS:
I just surprised that xdp mode uses 100% CPU, but epoll is 0%. Is this expected?

nibanks:
XDP is set up to use polling, yes. But not sockets.

ami-GS:
How about epoll? it is named poll as well.

nibanks:
> How about epoll? it is named poll as well.

No. Still no polling for it. Only XDP right now.

PR #4279, opened 2024-05-03
Support Priority Work on Connections

nibanks:
## Description

Sometimes an application would like to prefer new work over existing work already queued on the connection. This PR adds the notion of priority to the connection's operation queue, and exposes flags through the API to allow the app to chose to leverage this.

## Testing

SpinQuic updated.
Need to add some explicit BVTs.

## Documentation

TODO


mtfriesen (dismissed):

src/test/lib/DataTest.cpp
```diff
@@ -3362,6 +3362,90 @@ QuicTestStreamAbortConnFlowControl(
...
+    MsQuicStream* Streams[ConnectionPriorityTestContext::NumSend-1] = {0};
+    for (uint8_t i = 0; i < ConnectionPriorityTestContext::NumSend-1; ++i) {
+        Streams[i] = new MsQuicStream(Connection, QUIC_STREAM_OPEN_FLAG_UNIDIRECTIONAL);
+        TEST_QUIC_SUCCEEDED(Streams[i]->GetInitStatus());
+    }
+    for (uint8_t i = 0; i < ConnectionPriorityTestContext::NumSend-1; ++i) {
+        TEST_QUIC_SUCCEEDED(Streams[i]->Send(&Buffer, 1, QUIC_SEND_FLAG_START | QUIC_SEND_FLAG_FIN));
+    }
+    MsQuicStream Stream(Connection, QUIC_STREAM_OPEN_FLAG_UNIDIRECTIONAL);
+    TEST_QUIC_SUCCEEDED(Stream.Send(&PriorityBuffer, 1, QUIC_SEND_FLAG_START | QUIC_SEND_FLAG_FIN | (QUIC_SEND_FLAGS)(QUIC_STREAM_START_FLAG_PRIORITY_WORK)));
```
ami-GS, line 3433: This Send arrives at the end....

src/test/lib/DataTest.cpp
```diff
@@ -3362,6 +3362,102 @@ QuicTestStreamAbortConnFlowControl(
...
+    for (uint8_t i = 0; i < ConnectionPriorityTestContext::NumSend; ++i) {
+        Streams[i] = new MsQuicStream(Connection, QUIC_STREAM_OPEN_FLAG_UNIDIRECTIONAL, CleanUpManual, ConnectionPriorityTestContext::ClientStreamCallback, &Context);
+        TEST_QUIC_SUCCEEDED(Streams[i]->GetInitStatus());
+        // Queueing 100 StreamSendFlush operations during the Connection thread is blocked
+        TEST_QUIC_SUCCEEDED(Streams[i]->Send(&Buffer, 1, QUIC_SEND_FLAG_START | QUIC_SEND_FLAG_FIN));
+    }
+
+    // This GetParam operation should be somewhere earlier than the 100th StreamSendFlush operation
+    QUIC_STATISTICS_V2 Stat = {0};
+    uint32_t StatSize = sizeof(Stat);
```
ami-GS, line 3446: As far as I checked in `QuicConnProcessApiOperation`, `QUIC_API_TYPE_GET_PARAM` operation is always called at the end of queues. Not in the middle of Sends

src/core/operation.c
```diff
@@ -162,6 +185,9 @@ QuicOperationEnqueueFront(
     CXPLAT_DBG_ASSERT(Oper->Link.Flink == NULL);
 #endif
     StartProcessing = CxPlatListIsEmpty(&OperQ->List) && !OperQ->ActivelyProcessing;
+    if (OperQ->PriorityTail == &OperQ->List) {
+        OperQ->PriorityTail = &Oper->Link;
+    }
```
ami-GS, line 189: @nibanks  What does this for?

ami-GS:
What's this build error?
image

nibanks:
> What's this build error? image

Means you're not calling `new` correctly. You need to use the std::nothrow version.

src/core/operation.c
nibanks: If you enqueue something at the very front, it is considered a priority item. This is saying that if the priority tail currently is the HEAD, then we need to point the tail after this new item instead.

ami-GS:
CIFulzz is failling in `QuicConnRelease`, by `CXPLAT_TEL_ASSERT(Connection->RefTypeCount[QUIC_CONN_REF_STREAM] > 0)`.
I believe that `QuicOperationFree` is not called as expected? and Stream might not be released.

ami-GS:
Cannot repro using oss-fuzz....

src/core/operation.c
```diff
@@ -149,6 +151,27 @@
...
+    )
+{
+    BOOLEAN StartProcessing;
+    CxPlatDispatchLockAcquire(&OperQ->Lock);
+#if DEBUG
+    CXPLAT_DBG_ASSERT(Oper->Link.Flink == NULL);
+#endif
+    StartProcessing = CxPlatListIsEmpty(&OperQ->List) && !OperQ->ActivelyProcessing;
+    CxPlatListInsertTail(*OperQ->PriorityTail, &Oper->Link);
+    OperQ->PriorityTail = &Oper->Link.Flink;
```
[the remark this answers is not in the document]
ami-GS: @nibanks  I don't think this is accessing address in stack. How can ignore?

ami-GS (approved):

PR #4281, opened 2024-05-03
Debugger Extension Fixes & Improvements

nibanks:
## Description

Fixes some broken stuff and adds improved library output on the workers' state for each registration. This helps you easily see which are currently active.

## Testing

Used locally on a dump

## Documentation

N/A

### DIFF
--- a/src/plugins/dbg/quictypes.h
+++ b/src/plugins/dbg/quictypes.h
@@ -1303,6 +1303,28 @@ struct Listener : Struct {
     }
 };
 
+struct CxPlatWorker : Struct {
+
+    CxPlatWorker(ULONG64 Addr) : Struct("msquic!CXPLAT_WORKER", Addr) { }
+
+    ULONG64 Thread() {
+        return ReadPointer("Thread");
+    }
+};
+
+struct CxPlatExecutionContext : Struct {
+
+    CxPlatExecutionContext(ULONG64 Addr) : Struct("msquic!CXPLAT_EXECUTION_CONTEXT", Addr) { }
+
+    ULONG64 CxPlatContext() {
+        return ReadPointer("CxPlatContext");
+    }
+
+    CxPlatWorker GetCxPlatWorker() {
+        return CxPlatWorker(CxPlatContext());
+    }
+};
+
 struct Worker : Struct {
 
     Worker(ULONG64 Addr) : Struct("msquic!QUIC_WORKER", Addr) { }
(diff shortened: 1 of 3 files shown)


src/plugins/dbg/quictypes.h
```diff
@@ -1429,6 +1458,21 @@ struct Registration : Struct {
...
+    PSTR GetWorkersState() {
+        auto Workers = GetWorkerPool();
+        UCHAR WorkerCount = Workers.WorkerCount();
+        bool HasQueuedWorker = false;
+        for (UCHAR i = 0; i < WorkerCount; i++) {
+            if (Workers.GetWorker(i).IsActive()) {
+                return " ACTIVE";
+            }
+            if (Workers.GetWorker(i).HasWorkQueue()) {
+                HasQueuedWorker = true;
```
ami-GS, line 1471: why not `return "ACTIVE"; `?

src/plugins/dbg/quictypes.h
```diff
@@ -1429,6 +1458,21 @@ struct Registration : Struct {
...
+            }
+        }
+        return HasQueuedWorker ? "QUEUED" : "  IDLE";
```
ami-GS, line 1474: then `return "IDLE";`

src/plugins/dbg/quictypes.h
nibanks: Because the thread isn't actually running yet. It's just queued. But if any thread is running, I want to mark the whole registration as active.

src/plugins/dbg/quictypes.h
nibanks: I want to differentiate between queued and idle, so that even if it's not running yet, we can easily find a worker that will be running soon.

src/plugins/dbg/quictypes.h
```diff
@@ -1429,6 +1458,21 @@ struct Registration : Struct {
...
+        bool HasQueuedWorker = false;
+        for (UCHAR i = 0; i < WorkerCount; i++) {
+            if (Workers.GetWorker(i).IsActive()) {
+                return "ACTIVE";
+            }
+            if (Workers.GetWorker(i).HasWorkQueue()) {
+                HasQueuedWorker = true;
+            }
+        }
+        return HasQueuedWorker ? "QUEUED" : "  IDLE";
```
ami-GS, line 1474: Space? "  IDLE"

nibanks: I have the spacing so all the strings are the same length, to make the formatting in the output nicer.

ami-GS (approved):

PR #4283, opened 2024-05-03
Fix azure pipeline to newer image

ami-GS:
## Description

Azure pipeline have been using really old image. need update.

## Testing

N/A

## Documentation

N/A

### DIFF
--- a/.azure/obtemplates/build-linux.yml
+++ b/.azure/obtemplates/build-linux.yml
@@ -2,10 +2,6 @@ parameters:
   config: ''
   tls: 'openssl'
   platform: 'linux'
-  image1: ''
-  image2: ''
-  image3: ''
-  image4: ''
 
 jobs:
 - job: build_${{ parameters.platform }}_${{ parameters.tls }}_${{ parameters.config }}
@@ -20,9 +16,9 @@ jobs:
   - task: PowerShell@2
     displayName: Prepare Build Machine
     ${{ if eq(parameters.tls, 'openssl') }}:
-      target: ${{ parameters.image1 }}
+      target: ubuntu_2004_x86_64
     ${{ else }}:
-      target: ${{ parameters.image3 }}
+      target: ubuntu_2204_x86_64
     inputs:
       pwsh: true
       filePath: scripts/prepare-machine.ps1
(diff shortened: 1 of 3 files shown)


nibanks (approved):

.azure/obtemplates/build-linux.yml
```diff
@@ -16,39 +16,39 @@ jobs:
   - task: PowerShell@2
     displayName: Prepare Build Machine
     ${{ if eq(parameters.tls, 'openssl') }}:
-      target: linux_build_container2
+      target: linux_build_container2-x86_64
     ${{ else }}:
-      target: linux_build_container
+      target: linux_build_container-x86_64
```
csujedihy, line 21: We can only use linux_build_container and linux_build_container2, which map to LinuxContainerImage and LinuxContainerImage respectively.

ami-GS: where it is defined and how can we update it?

csujedihy: We can't update or change it. It's part of 1ES template that we don't have control over.

csujedihy: Perhaps you can override the value of LinuxContainerImage and LinuxContainerImage2 in the `variables` property of the job that the tasks belong to.

csujedihy: but you still only have 2 linux container image you can use so if you wanna try that you need to make an extra job for the other build flavors.

.azure/obtemplates/build-linux.yml
```diff
@@ -16,39 +19,39 @@ jobs:
   - task: PowerShell@2
     displayName: Prepare Build Machine
     ${{ if eq(parameters.tls, 'openssl') }}:
-      target: linux_build_container2
+      target:  ${{ variables.image }}-20.04-x86_64
```
nibanks, line 22: I don't think this will work in OneBranch. Please try it out via a PR (don't merge) [internally](https://mscodehub.visualstudio.com/_git/msquic). You can use git to add a remote to your current enlistment to push your branch there to try to make a PR. Please work with @csujedihy to get this fixed ASAP. Our internal pipeline is blocked until we do.

nibanks (approved):

PR #4298, opened 2024-05-13
Fix HPS test for Linux xdp

ami-GS:
## Description

When using sudo for Linux XDP, `ulimit -n` under sudo env becomes 1024 which easily run out of opening file quota.
Set same number as the one of normal user (1048576).

## Testing

HPS test locally work .

## Documentation

N/A


ProjectsByJackHe:
Looks like the changes in this PR are causing crashes for our other test configurations: 
image

nibanks:
Should this be set in prepare-machine.ps1 instead?

ami-GS:
As far as I researched, Any limits written in /etc/security/limits.conf are reflected for a user who need to login which means requiring reboot. In other words a daemon process which doesn't need to login cannot refer the value.

So,
for daemon process
- Use startup script with ulimit commands
- Use systemd setupfile
- Use setrlimit() syscall

for manual run
- Use /etc/security/limits.conf then reboot
- Use ulimit from test scripts

ami-GS:
This should be ready to go

nibanks (approved):

Issue #4299, opened 2024-05-13
Linux XDP instructions break the machine


larseggert:
### Describe the bug

This step in `docs/BUILD.md`
```
sudo apt-add-repository "deb http://mirrors.kernel.org/ubuntu noble main" -y
```
hoses a 22.04 machine, because it unconditionally adds debs from 24.04.

### Affected OS

- [ ] Windows
- [X] Linux
- [ ] macOS
- [ ] Other (specify below)

### MsQuic version

main

### Steps taken to reproduce bug

Followed XDP steps in `docs/BUILD.md`

### Expected behavior

Expected to not have a broken machine after :-)

### Actual outcome

Base 22.04 debs get overwritten with 24.04 ones.

nibanks:
@ami-GS please take a look.

ami-GS:
@nibanks  This is expected, or we need to build from source

larseggert:
The it would be good to call out in big, red letters that following the steps will render a 22.04 machine un-updatable.

ami-GS:
@larseggert PR was merged

PR #4302, opened 2024-05-15
show warning prompt for installing Linux XDP deps

ami-GS:
## Description

Fix of https://github.com/microsoft/msquic/issues/4299
Currently Linux XDP setup is for Ubuntu 22.04, but installing dependencies from Ubuntu 24.04 which should break the environment.
This change shows warning to user before actually execute prepare-machine.ps1

## Testing

Locally works this "YES" prompt

## Documentation

wrote.

### DIFF
--- a/scripts/prepare-machine.ps1
+++ b/scripts/prepare-machine.ps1
@@ -71,6 +71,9 @@ param (
     [switch]$UseXdp,
 
     [Parameter(Mandatory = $false)]
+    [switch]$InstallLinuxXdpByForce,
+
+    [Parameter(Mandatory = $false)]
     [switch]$InstallArm64Toolchain,
 
     [Parameter(Mandatory = $false)]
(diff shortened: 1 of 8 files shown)


scripts/prepare-machine.ps1
```diff
@@ -70,6 +70,9 @@ param (
     [Parameter(Mandatory = $false)]
     [switch]$UseXdp,
 
+    [Parameter(Mandatory = $false)]
+    [switch]$InstallLinuxXdpByForce,
```
nibanks, line 74: Maybe `ForceXdpInstall`?

nibanks: Maybe comment/document the parameter too.

nibanks (dismissed):

nibanks (dismissed):

nibanks (approved):

Issue #4309, opened 2024-05-17
ParameterValidation/WithValidateConnectionEventArgs.ValidateConnectionEvents/2 fail on Linux with duoNic


ami-GS:
### Describe the bug

ParameterValidation/WithValidateConnectionEventArgs.ValidateConnectionEvents/2 fail with duoNic

There are two causes of this issue.
- Linux XDP initialization fails for some reason
- Normal socket with duoNic fail by QuicAddrSetToDuoNic with Addr family of 0

### Affected OS

- [ ] Windows
- [X] Linux
- [ ] macOS
- [ ] Other (specify below)

### MsQuic version

main

### Steps taken to reproduce bug

1. build with -EnableLinuxXDP and run the test with `--duoNic` until it hits issue (never tried).
2. build without `-EnableLinuxXDP` and run the test with `--duoNic`

./artifacts/bin/linux/x64_Debug_openssl3/msquictest --gtest_filter="ParameterValidation/WithValidateConnectionEventArgs.ValidateConnectionEvents/2" --duoNic 

### Expected behavior

pass

### Actual outcome

[ RUN      ] ParameterValidation/WithValidateConnectionEventArgs.ValidateConnectionEvents/2
[1]    4190107 IOT instruction (core dumped)  ./artifacts/bin/linux/x64_Debug_openssl3/msquictest  --gtest_repeat=1 --duoNi

PR #4310, opened 2024-05-17
QuicTestValidateConnectionEvents3 fix to use duoNic with normal socket

ami-GS:
## Description

One fix of https://github.com/microsoft/msquic/issues/4309

## Testing
Build without "-EnableLinuxXDP"
`./artifacts/bin/linux/x64_Debug_openssl3/msquictest --gtest_filter="ParameterValidation/WithValidateConnectionEventArgs.ValidateConnectionEvents/2" --duoNic`
Locally passed

## Documentation

N/A

### DIFF
--- a/src/test/lib/EventTest.cpp
+++ b/src/test/lib/EventTest.cpp
@@ -499,7 +499,7 @@ QuicTestValidateConnectionEvents3(
 
     QUIC_BUFFER* ResumptionTicket = nullptr;
     QuicTestPrimeResumption(
-        QuicAddrGetFamily(&ServerLocalAddr.SockAddr),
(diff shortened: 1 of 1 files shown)


src/test/lib/EventTest.cpp
```diff
@@ -499,7 +499,7 @@ QuicTestValidateConnectionEvents3(
 
     QUIC_BUFFER* ResumptionTicket = nullptr;
     QuicTestPrimeResumption(
-        QuicAddrGetFamily(&ServerLocalAddr.SockAddr),
+        QuicAddrGetFamily(&ServerLocalAddr.SockAddr) == QUIC_ADDRESS_FAMILY_INET ? QUIC_ADDRESS_FAMILY_INET : QUIC_ADDRESS_FAMILY_INET6,
```
nibanks, line 502: Why?

ami-GS: This is always 0, which does not work appropriately with duonic as it assert check the family whether 4 or 6.
Normal socket case just uses ipv6 localhost if it is 0. This change is for duonic to work with 0.

nibanks (approved):

PR #4316, opened 2024-05-20
Uninitialize LossDetection earlier to cleanup Datagram frame

ami-GS:
## Description

App allocated buffer for Datagram Frame has been leaking due to
- missing callback handler when waiting Ack for the Datagram frame.
- allocation failure for packet metadata

## Testing

Ran local test. No leak detected. No leak found from my own memory tracker either.
`./artifacts/bin/linux/x64_Debug_openssl3/spinquic both -timeout:600000 -repeat_count:20 -alloc_fail:100`
`./artifacts/bin/linux/x64_Debug_openssl/spinquic both -timeout:600000 -repeat_count:20 -alloc_fail:100`

Let's see automation results

## Documentation

N/A

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -1322,6 +1322,9 @@ QuicConnOnShutdownComplete(
     Connection->State.ShutdownComplete = TRUE;
     Connection->State.UpdateWorker = FALSE;
 
+    // call earlier then others for frames which requires ClientCallbackHandler to freeing app memory (Datagram)
+    QuicLossDetectionUninitialize(&Connection->LossDetection);
+
     QuicTraceEvent(
         ConnShutdownComplete,
         "[conn][%p] Shutdown complete, PeerFailedToAcknowledged=%hhu.",
@@ -1368,7 +1371,6 @@ QuicConnOnShutdownComplete(
     // Clean up the rest of the internal state.
     //
     QuicTimerWheelRemoveConnection(&Connection->Worker->TimerWheel, Connection);
-    QuicLossDetectionUninitialize(&Connection->LossDetection);
     QuicSendUninitialize(&Connection->Send);
 
     if (!Connection->State.ExternalOwner) {
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -1322,34 +1322,11 @@ QuicConnOnShutdownComplete(
     Connection->State.ShutdownComplete = TRUE;
     Connection->State.UpdateWorker = FALSE;
 
-    // call earlier then others for frames which requires ClientCallbackHandler to freeing app memory (Datagram)
-    QuicLossDetectionUninitialize(&Connection->LossDetection);
-
     QuicTraceEvent(
         ConnShutdownComplete,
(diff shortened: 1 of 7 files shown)


src/core/connection.c
```diff
@@ -1322,6 +1322,9 @@ QuicConnOnShutdownComplete(
     Connection->State.ShutdownComplete = TRUE;
     Connection->State.UpdateWorker = FALSE;
 
+    // call earlier then others for frames which requires ClientCallbackHandler to freeing app memory (Datagram)
+    QuicLossDetectionUninitialize(&Connection->LossDetection);
```
nibanks, line 1326: Instead of moving this up, let's try moving the shutdown complete event down, after send uninitialize?

nibanks:
Since there are still leaks, even with this change, I feel like `QuicDatagramSendShutdown` isn't getting called before shutdown complete somehow. Can you add some debug asserts for:

1. Before the shutdown complete event, assert that Datagram has been shutdown/disabled.
2. Before the shutdown complete event, assert that zero outstanding datagrams are pending.

nibanks:
I wonder if we need to add a `QuicDatagramSendShutdown` after `QuicSendUninitialize` in this same function.

ami-GS:
DatagramSend is still called after QuicDatagramSendShutdown which touches released `Datagram->ApiQueueLock` then crash

ami-GS:
Another leak
image
```c
    QUIC_SENT_PACKET_METADATA* SentPacket =
        QuicSentPacketPoolGetPacketMetadata(
            &Connection->Worker->SentPacketPool, TempSentPacket->FrameCount);
```

ami-GS:
The `QUIC_SENT_PACKET_METADATA` should be released by `QuicLossDetectionUninitialize`....

ami-GS:
There is still case of `    QUIC_DATAGRAM_SEND_SENT` -> `QUIC_DATAGRAM_SEND_LOST_SUSPECT` -> leak

ami-GS:
- `LossDetection->SentPackets`
- `LossDetection->LostPackets`

are cleaned by `QuicLossDetectionUninitialize` in `QuicConnOnShutdownComplete` though

ami-GS:
Fixed!

nibanks (dismissed):
LGTM. Thanks for making these changes!

nibanks:
I see there are a number of BVT failures. I"m rerunning them to see if they are real regressions or not.

ami-GS:
Linux XDP related is fixed by different PR

src/core/loss_detection.c
```diff
@@ -408,6 +409,7 @@ QuicLossDetectionOnPacketSent(
             "Sent packet metadata",
             SIZEOF_QUIC_SENT_PACKET_METADATA(TempSentPacket->FrameCount));
         QuicLossDetectionRetransmitFrames(LossDetection, TempSentPacket, FALSE);
+        QuicLossDetectionOnPacketDiscarded(LossDetection, TempSentPacket, FALSE, FALSE);
```
nibanks, line 412: We talked in person about a slightly different way to do this. Please update. Thanks!

nibanks (approved):

PR #4320, opened 2024-05-23
Possibility for datagrams to be dropped when not able to send

iiztp:
## Description

When adding a Token-Bucket filter and sending more data than we are allowed to, I found out that there was a lot of delay (the more we wait the more we have delay) between sending and receiving.
After some discussion, it was established that msquic is buffering datagrams when it can't send them immediately.
With this PR, it is now possible to drop them when you can't send them immediately.
I added a flag to the QUIC_SEND_FLAGS : "QUIC_SEND_FLAG_DGRAM_CANCEL_ON_BLOCKED"
When this flag is set, on sending it will check if there is queued messages to send, if there is, it drops (cancels) them.

## Testing

_Do any existing tests cover this change?_

I don't think so.

 _Are new tests needed?_

Maybe...

## Documentation

Just the new flag : "QUIC_SEND_FLAG_DGRAM_CANCEL_ON_BLOCKED"

## Some more infos

This is the one-way delay of some datagrams with a TBF after sending more than we are allowed to, see images.
Without flag (before patch):
image

With flag:
image

The values are like this : "{packet number}, {one-way delay in ms}"
The values with the flag are around the same values as I have with UDP (which normally drops packets when it can't send them).


iiztp:

src/core/datagram.c
```diff
@@ -421,7 +421,13 @@ QuicDatagramSendFlush(
         CXPLAT_DBG_ASSERT(!(SendRequest->Flags & QUIC_SEND_FLAG_BUFFERED));
         CXPLAT_TEL_ASSERT(Datagram->SendEnabled);
 
-        if (SendRequest->TotalLength > (uint64_t)Datagram->MaxSendLength || QuicConnIsClosed(Connection)) {
+        // Cancels the sending of a Datagram if:
+        // The length we want to send > the length we can send
+        // The connection is closed
+        // The sending mode requires to send immediately (no datagram in queue)
```
nibanks, line 427: Just because there is another datagram in the queue doesn't mean we should cancel this one, IMO. This logic should be moved to `QuicSendFlush`, to cancel any datagrams that can't be sent the next time we send out packets.

nibanks:
> Are new tests needed?
>
> Maybe...

Yeah, we will need to add some tests, as well as make the minor edits to the docs.

nibanks:
You will also need to run `.\scripts\generate-dotnet.ps1` to update the .NET files.

src/core/datagram.c
iiztp: Actually, I printed out the values we had in the Datagram object to check what changes when we do or do not have congestion, and that was the only thing that changed (from NULL to a value) that is why I've done that check

iiztp: I think I understand what you meant sorry, you want to extend it to Quic packets and not only Quic Datagrams ?

src/core/datagram.h
```diff
@@ -103,3 +103,9 @@ QuicDatagramProcessFrame(
         const uint8_t * const Buffer,
     _Inout_ uint16_t* Offset
     );
+
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+QuicDatagramCancelPending(
+    _In_ QUIC_CONNECTION* Connection
+    );
```
nibanks, line 111: nit: please leave a trailing empty newline

src/core/datagram.c
```diff
@@ -583,3 +575,19 @@ QuicDatagramProcessFrame(
...
+{
+    QUIC_DATAGRAM* Datagram = &Connection->Datagram;
+    while (Datagram->SendQueue != NULL) {
+        QUIC_SEND_REQUEST* SendRequest = Datagram->SendQueue;
+        Datagram->SendQueue = SendRequest->Next;
+        QuicDatagramCancelSend(Connection, SendRequest);
+    }
+    Datagram->PrioritySendQueueTail = &Datagram->SendQueue;
+    Datagram->SendQueueTail = &Datagram->SendQueue;
+}
```
nibanks, line 593: nit: please leave a trailing empty newline

src/core/datagram.c
```diff
@@ -583,3 +575,19 @@ QuicDatagramProcessFrame(
...
+void
+QuicDatagramCancelPending(
+    _In_ QUIC_CONNECTION* Connection
+    )
+{
+    QUIC_DATAGRAM* Datagram = &Connection->Datagram;
+    while (Datagram->SendQueue != NULL) {
+        QUIC_SEND_REQUEST* SendRequest = Datagram->SendQueue;
+        Datagram->SendQueue = SendRequest->Next;
+        QuicDatagramCancelSend(Connection, SendRequest);
```
nibanks, line 589: Only cancel sends that have the appropriate flag. It's possible some do and some don't.

src/core/datagram.c
```diff
@@ -178,13 +178,7 @@ QuicDatagramSendShutdown(
     //
     // Cancel all outstanding send requests.
     //
-    while (Datagram->SendQueue != NULL) {
```
nibanks, line 181: Let's leave the code in this function unchanged.

src/core/send.c
```diff
@@ -1498,6 +1498,13 @@ QuicSendFlush(
         //QuicConnUpdatePeerPacketTolerance(Connection, Builder.TotalCountDatagrams);
     }
 
+    //
+    // Cancel all outstanding send requests if not able to send.
+    //
+    if(!(Send->SendFlags & QUIC_SEND_FLAG_DGRAM_CANCEL_ON_BLOCKED)) {
```
nibanks, line 1504: These are different kinds of send flags, so you need to just always call this function (no `if` check).

src/core/datagram.h
```diff
@@ -103,3 +103,9 @@ QuicDatagramProcessFrame(
         const uint8_t * const Buffer,
     _Inout_ uint16_t* Offset
     );
+
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+QuicDatagramCancelPending(
```
nibanks, line 109: Now that I think about it, I think we need a better name. I want this function to only cancel the datagram sends that want to be cancelled (i.e., they have the flag).

src/inc/msquic.h
```diff
@@ -241,6 +241,7 @@ typedef enum QUIC_SEND_FLAGS {
     QUIC_SEND_FLAG_DGRAM_PRIORITY           = 0x0008,   // Indicates the datagram is higher priority than others.
     QUIC_SEND_FLAG_DELAY_SEND               = 0x0010,   // Indicates the send should be delayed because more will be queued soon.
     QUIC_SEND_FLAG_CANCEL_ON_LOSS           = 0x0020,   // Indicates that a stream is to be cancelled when packet loss is detected.
+    QUIC_SEND_FLAG_DGRAM_CANCEL_ON_BLOCKED  = 0x0030,   // Indicates that a datagram should be dropped when it can't be sent immediately.
```
nibanks, line 244: I do wonder if a similar flag would be useful for streams (in the future), so should we remove the `DGRAM` from the name...

nibanks, line 244: We are going to need to add some tests for this, as well as update any relevant docs.

nibanks (commented):
Looking pretty good so far! Thanks!

src/inc/msquic.h
iiztp: Shall we change the function name / position to another file then such as "send.c", it would leave us with "QuicSendCancelOnBlocked" smth like that

nibanks: No. These are meant to be for the send requests, which is not the same thing as what send.c deals with. The location here is good.

iiztp: Well, what do you think of "QuicDatagramCancelOnBlocked" then, but we'll have to change that in the future if you want to include the same for Streams ?

nibanks: Sounds good; maybe just `QuicDatagramCancelBlocked`. In the future, we can add a `QuicConnectionCancelBlockedStreams`.

src/core/datagram.c
iiztp: I might need some review on this one lol

src/core/datagram.c
```diff
@@ -426,7 +426,6 @@ QuicDatagramSendFlush(
             continue;
         }
         TotalBytesSent += SendRequest->TotalLength;
-
```
nibanks, line 429: nit: please revert these changes

src/core/datagram.c
```diff
@@ -583,3 +581,37 @@ QuicDatagramProcessFrame(
...
+void
+QuicDatagramCancelBlocked(
+    _In_ QUIC_CONNECTION* Connection
+    )
+{
+    QUIC_DATAGRAM* Datagram = &Connection->Datagram;
+    QUIC_SEND_REQUEST* SendQueue = Datagram->SendQueue;
+    QUIC_SEND_REQUEST** PrioritySendQueueTail = Datagram->PrioritySendQueueTail;
+    Datagram->SendQueue = NULL;
+    bool priority = true;
```
nibanks, line 595:
```suggestion
    BOOLEAN Priority = TRUE;
```

src/core/datagram.c
```diff
@@ -583,3 +581,37 @@ QuicDatagramProcessFrame(
...
+
+    while (SendQueue != NULL) {
+        if(SendQueue->Flags & QUIC_SEND_FLAG_CANCEL_ON_BLOCKED) {
```
nibanks, line 598:
```suggestion
        if (SendQueue->Flags & QUIC_SEND_FLAG_CANCEL_ON_BLOCKED) {
```

src/core/datagram.c
```diff
@@ -583,3 +581,37 @@ QuicDatagramProcessFrame(
...
+            QuicDatagramCancelSend(Connection, SendQueue);
+        } else {
+            if(Datagram->SendQueue == NULL) {
```
nibanks, line 601:
```suggestion
            if (Datagram->SendQueue == NULL) {
```

src/core/datagram.c
```diff
@@ -583,3 +581,37 @@ QuicDatagramProcessFrame(
...
+                Datagram->SendQueueTail = &SendQueue;
+            } else {
+                if(priority) {
+                    (*(Datagram->PrioritySendQueueTail))->Next = SendQueue;
+                }
+                (*(Datagram->SendQueueTail))->Next = SendQueue;
+                Datagram->SendQueueTail = &SendQueue;
+            }
+        }
+        if(&SendQueue == PrioritySendQueueTail) {
```
nibanks, line 612:
```suggestion
        if (&SendQueue == PrioritySendQueueTail) {
```

src/core/datagram.c
```diff
@@ -583,3 +581,32 @@ QuicDatagramProcessFrame(
...
+                Datagram->PrioritySendQueueTail = PreviousAdded;
+            }
+            if(PreviousAdded != NULL) {
+                (*PreviousAdded)->Next = SendRequest->Next;
+            }
+            *SendQueue = SendRequest->Next;
+            QuicDatagramCancelSend(Connection, SendRequest);
+        } else {
+            PreviousAdded = SendQueue;
+            SendQueue = &((*SendQueue)->Next);
```
nibanks, line 608: You need this in the `if` above too.

nibanks: nvm

src/core/datagram.c
```diff
@@ -583,3 +581,32 @@ QuicDatagramProcessFrame(
...
+    QUIC_SEND_REQUEST** SendQueue = &Datagram->SendQueue;
+    QUIC_SEND_REQUEST** PreviousAdded = NULL;
+
+    while (*SendQueue != NULL) {
+        if ((*SendQueue)->Flags & QUIC_SEND_FLAG_CANCEL_ON_BLOCKED) {
+            QUIC_SEND_REQUEST* SendRequest = *SendQueue;
+            if(SendQueue == Datagram->PrioritySendQueueTail) {
+                Datagram->PrioritySendQueueTail = PreviousAdded;
+            }
+            if(PreviousAdded != NULL) {
```
nibanks, line 601:
```suggestion
            if (SendQueue == Datagram->PrioritySendQueueTail) {
                Datagram->PrioritySendQueueTail = PreviousAdded;
            }
            if (PreviousAdded != NULL) {
```

iiztp: Sorry I'm not used to it ahah

nibanks:
The changes looks good, but how do should we add a test to verify it? One minimum option is to update spinquic.cpp to ensure it uses this new flag (should just require changing one number). But ideally, it'd be nice to have a functional test case.

iiztp:
> The changes looks good, but how do should we add a test to verify it? One minimum option is to update spinquic.cpp to ensure it uses this new flag (should just require changing one number). But ideally, it'd be nice to have a functional test case.

Maybe we should just add more than we can send i.e a lot of DatagramSend stuck in a loop until we reach out the MTU which is 1500 if we are over ethernet or wifi (so maybe 2 datagrams of a thousand bytes then multiple datagrams that we can drop) or afaik ~65000 for localhost (2 datagrams of 45 000 bytes and multiple datagrams that we can drop) ?

I don't know if this would work, otherwise we could just create a new CI and run a QUIC server on it with special traffic control rules (i.e bandwidth limit) and check if some packets that we allowed to be dropped are dropped and the others are not but the outputs would be different for each run and wouldn't give us a general idea.

submodules/googletest
nibanks, line 1: Please revert this (I assume accidental) change.

src/test/lib/DatagramTest.cpp
```diff
@@ -156,7 +156,10 @@ QuicTestDatagramSend(
     TEST_TRUE(ServerConfiguration.IsValid());
 
     uint8_t RawBuffer[] = "datagram";
+    uint8_t RawBigBuffer[1100];
+    memset(RawBigBuffer, 0x00, 1100);
```
nibanks, line 160:
```suggestion
    uint8_t RawBigBuffer[1100] = {0};
```

iiztp:
Quick question on this PR, even though some tests have failed (because they were too long to run or because they had their connection that was interrupted (?) ) do I have something more to do? Like, do I have to fix the failed tests?

nibanks:
I'm not sure what happened, but it looks like you pushed a lot of unexpected new files.

iiztp:
> I'm not sure what happened, but it looks like you pushed a lot of unexpected new files.

Other than merging, I did nothing else :/

nibanks:
I think the problem is the msquicdocs folder shouldn't be in the main branch. Can you please remove that?

src/test/lib/DatagramTest.cpp
```diff
@@ -255,6 +257,22 @@ QuicTestDatagramSend(
                 TEST_TRUE(Client.GetDatagramsSuspectLost() > InitialLostCount);
                 CxPlatSleep(100);
 #endif
+                for (int i = 0; i < 20; i++) {
```
nibanks, line 260: Can we actually create a new test for this stuff, instead of putting it in this existing one? Also, It's possible, though unlikely, that you don't get any cancelled datagrams if they all happen to fit in the congestion control window. So, I recommend the new test queues (sends) all these datagrams on the connection **before** you call `Client.Start` to ensure you've queued up more datagrams than can be sent in the initial congestion window.

nibanks: @iiztp do you plan to finish up these last comments so we can merge?

iiztp: > @iiztp do you plan to finish up these last comments so we can merge?

Hi, I'll try to do that this week if you're okay with that.
Been a bit busy these last weeks

src/test/MsQuicTests.h
```diff
@@ -1332,3 +1337,7 @@ typedef struct {
     // BOOLEAN - EnableResumption
 
 #define QUIC_MAX_IOCTL_FUNC_CODE 125
+
+#define IOCTL_QUIC_RUN_DATAGRAM_DROP \
```
nibanks, line 1341: Put before Max definition and increment max.

src/test/bin/winkernel/control.cpp
```diff
@@ -955,6 +955,13 @@ QuicTestCtlEvtIoDeviceControl(
                 Params->Family));
         break;
 
+    case IOCTL_QUIC_RUN_DATAGRAM_DROP:
```
nibanks, line 958: There is a big array above that needs to be updated too.

src/core/datagram.c
```diff
@@ -588,3 +588,34 @@ QuicDatagramProcessFrame(
...
     return TRUE;
 }
+
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+QuicDatagramCancelBlocked(
+    _In_ QUIC_CONNECTION* Connection
+    )
+{
+    QUIC_DATAGRAM* Datagram = &Connection->Datagram;
```
nibanks, line 598: Should we add a `if (*SendQueue != NULL) return;` at the beginning here, so that we don't have to do everything below every single send flush loop if there are no datagrams?

src/core/datagram.c
```diff
@@ -588,3 +588,34 @@ QuicDatagramProcessFrame(
...
+    QUIC_SEND_REQUEST** SendQueue = &Datagram->SendQueue;
+    while (*SendQueue != NULL) {
```
nibanks, line 600:
```suggestion
    if(*SendQueue == NULL) {
        return;
    }

        while (*SendQueue != NULL) {
```

src/core/datagram.c
iiztp: Hi Nick,

I don't think it's an interesting addition (my opinion).
Although if it seems clearer for you let's add it, though, assembly-speaking if the compiler doesn't do the optimisation itself, it will just add up some instructions that aren't useful as it is already verified by the while loop.

nibanks: The part that really matters is the end of the function IMO:
```c
    Datagram->SendQueueTail = SendQueue;

    if (Datagram->SendQueue != NULL) {
        QuicSendSetSendFlag(&Connection->Send, QUIC_CONN_SEND_FLAG_DATAGRAM);
    } else {
        QuicSendClearSendFlag(&Connection->Send, QUIC_CONN_SEND_FLAG_DATAGRAM);
    }

    QuicDatagramValidate(Datagram);
```
Having to call `QuicSendClearSendFlag` every loop for no reason for instance.

BTW, If we do add this `if` check, let's change the `while` loop to a `do {} while` loop.

iiztp: Fair

nibanks (approved):

nibanks (approved):

iiztp:
Sorry for the spam, I think I needed to add the bindings to the win file also (some tests were failing)

nibanks (approved):

PR #4332, opened 2024-05-31
Retry flaky test

ami-GS:
## Description

Several tests sometimes fail due to randomness and slower test environment on Github Action.
This change executes retry once if first trial fails. Then failure twice in a row is considered as actual failure.
(e.g. take over 10sec to start processing received packets which causes IDLE.TIMEOUT then shutdown its connection.)

## Testing

See automation....

## Documentation

N/A

### DIFF
--- a/scripts/run-gtest.ps1
+++ b/scripts/run-gtest.ps1
@@ -299,7 +299,7 @@ function Add-XmlResults($TestCase) {
         $TestCaseNode = $Node.testcase | Where-Object { $_.Name -eq $TestCaseName }
         if ($TestCaseNode) {
             $IsRetry = $true
-            $Node.retries = ($Node.retries -as [Int]) + 1
+            $Node.retried = ($Node.retried -as [Int]) + 1
         }
         # Already has a matching test suite. Add the test case to it.
         $Node.tests = ($Node.tests -as [Int]) + 1
--- a/scripts/run-gtest.ps1
+++ b/scripts/run-gtest.ps1
@@ -924,12 +924,17 @@ try {
     Log "$($TestCount) test(s) run."
     if ($KeepOutputOnSuccess -or ($TestsFailed -ne 0) -or ($global:CrashedProcessCount -ne 0)) {
         Log "Output can be found in $($LogDir)"
+        $ConsistentFailure = $TestsFailed - $TestsRetried
+        $PassedWithRetry = $TestsRetried - $ConsistentFailure
         if ($ErrorsAsWarnings -or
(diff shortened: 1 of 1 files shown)


scripts/run-gtest.ps1
```diff
@@ -295,6 +296,11 @@ function Add-XmlResults($TestCase) {
         $Node = $XmlResults.testsuites.testsuite | Where-Object { $_.Name -eq $TestSuiteName }
     }
     if ($null -ne $Node) {
+        $TestCaseNode = $Node.testcase | Where-Object { $_.Name -eq $TestCaseName }
+        if ($TestCaseNode) {
+            $IsRetry = $true
+            $Node.retried = ($Node.retried -as [Int]) + 1
```
ami-GS, line 302: powershell failed at here.
$Node doesn't have `retried` property.
How can I add?

ami-GS:
Error annotation is shown, but not treated as failure by retry

image

scripts/run-gtest.ps1
```diff
@@ -913,11 +924,12 @@ try {
     Log "$($TestCount) test(s) run."
     if ($KeepOutputOnSuccess -or ($TestsFailed -ne 0) -or ($global:CrashedProcessCount -ne 0)) {
         Log "Output can be found in $($LogDir)"
-        if ($ErrorsAsWarnings) {
+        if ($ErrorsAsWarnings -or
+            (($IsolationMode -eq "Isolated") -and ($TestsFailed -ne 0) -and ($TestsFailed -eq $TestsRetried))) {
             Write-Warning "$($TestsFailed) test(s) failed."
-            Write-Warning "$($TestsFailed) test(s) failed, $($global:CrashedProcessCount) test(s) crashed."
+            Write-Warning "$($TestsFailed) test(s) failed, $($TestsRetried) test(s) retried, $($global:CrashedProcessCount) test(s) crashed."
```
nibanks, line 930: What do you think about simplifying this message?
```suggestion
            Write-Warning "$($TestsFailed) test(s) passed with retry."
```

scripts/run-gtest.ps1
```diff
@@ -913,11 +924,12 @@ try {
...
         } else {
-            Write-Error "$($TestsFailed) test(s) failed, $($global:CrashedProcessCount) test(s) crashed."
+            Write-Error "$($TestsFailed) test(s) failed, $($TestsRetried) test(s) retried, $($global:CrashedProcessCount) test(s) crashed."
```
nibanks, line 932: Can we identify how many tests passed on retry and how many failed on retry? We want to highlight the consistent failures.

nibanks (approved):

PR #4348, opened 2024-06-14
Install Perl on Windows via WinGet

anrossi:
## Description

Updates the prepare-machine script to install Perl on windows for OpenSSL builds.

## Testing

Ran the script locally and it worked fine.

## Documentation

N/A

### DIFF
--- a/scripts/prepare-machine.ps1
+++ b/scripts/prepare-machine.ps1
@@ -140,7 +140,6 @@ if ($ForBuild) {
     # enabled for any possible build.
     $InstallNasm = $true
     $InstallJom = $true
-    $InstallPerl = $false
     $InstallCoreNetCiDeps = $true; # For kernel signing certs
 }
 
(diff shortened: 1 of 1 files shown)


nibanks (approved):

scripts/prepare-machine.ps1
```diff
@@ -137,6 +140,7 @@ if ($ForBuild) {
     # enabled for any possible build.
     $InstallNasm = $true
     $InstallJom = $true
+    $InstallPerl = $true
```
nibanks, line 143: Don't default this on. winget isn't available everywhere.
```suggestion
    $InstallPerl = $false
```

scripts/prepare-machine.ps1
```diff
@@ -137,6 +140,7 @@ if ($ForBuild) {
     # enabled for any possible build.
     $InstallNasm = $true
     $InstallJom = $true
+    $InstallPerl = $false
```
nibanks, line 143:
```suggestion
    $InstallJom = $true
```

mtfriesen (approved):

anrossi:
> Now we don't need to install anything manually for windows anymore?

I hope this is the last one needed. The developer will need to add the switch on the prepare-machine command line (because the build system uses the script, so it can't be automatic), but this should make it easier.

PR #4349, opened 2024-06-14
Log QUIC_TLS_SECRETS on all TestConnections and write to disk on cleanup

anrossi:
## Description

Logs QUIC_TLS_SECRETs on all TestConnections. If a filepath is provided, then it will write the QUIC_TLS_SECRETS in an sslkeylogfile format to the provided file path when the TestConnection is destructed.

## Testing

CI

## Documentation

N/A


src/test/lib/TestConnection.h
```diff
@@ -17,6 +17,8 @@ enum NEW_STREAM_START_TYPE {
     NEW_STREAM_START_ASYNC      // Start asynchronously
 };
 
+#define DEFAULT_SSLKEYLOGFILE_PATH "sslkeylogfile.txt"
```
ami-GS, line 20: PATH? not FILENAME?

anrossi: I don't know what word to use to describe that it's both path and filename without writing that out and making the name longer

anrossi:
DO NOT COMPLETE THIS PR. Trying out using a different `SetParam` in the client TestConnection constructor to see if it's a SetParam issue, or a TLS_SECRETS issue.

anrossi:
I can't repro the test failure locally.

nibanks (approved):
LGTM!

PR #4351, opened 2024-06-14
Spinquic to use operation priority

ami-GS:
## Description

Spinquic to use StreamSend, StreamStart and Get/SetParm op priorities.

## Testing

See automation

## Documentation

N/A

### DIFF
--- a/src/tools/spin/spinquic.cpp
+++ b/src/tools/spin/spinquic.cpp
@@ -582,6 +582,9 @@ struct SetParamHelper {
     void SetUint64(uint32_t _Type, uint64_t Value) {
         Type = _Type; Param.u64 = Value; Size = sizeof(Value);
     }
+    void SetPriority() {
+        Type |= QUIC_PARAM_HIGH_PRIORITY;
+    }
     void Apply(HQUIC Handle) {
         if (Type != -1) {
             MsQuic.SetParam(Handle, Type, Size, IsPtr ? Param.ptr : &Param);
@@ -828,6 +831,10 @@ void SpinQuicSetRandomConnectionParam(HQUIC Connection, uint16_t ThreadID)
         break;
     }
 
+    if (GetRandom(2)) {
+        Helper.SetPriority();
+    }
+
     Helper.Apply(Connection);
 }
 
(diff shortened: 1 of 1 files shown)


ami-GS:
What! CIFuzz passes!!

src/core/connection.h
```diff
@@ -1067,6 +1067,10 @@ QuicConnRelease(
     QuicConnValidate(Connection);
 
 #if DEBUG
+    if (Connection->RefTypeCount[Ref] <= 0) {
+        fprintf(stderr, "RefTypeCount[%d] is %d\n", Ref, Connection->RefTypeCount[Ref]);
```
ami-GS, line 1071: `RefTypeCount[6] is -22829`

RefTypeCount is `short` type (from -32,768 to 32,767). So not just off-by-one issue

ami-GS: OSS-Fuzz issue dashboard has been listing same issue and the cause seems to be from other PR?

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=68176&q=msquic
https://oss-fuzz.com/testcase-detail/4533150459756544
https://github.com/microsoft/msquic/compare/d65bb1af4cc96f996d0312f445ce648f114b380d...102a491ff1394a1bad4d98350126ec2681814700

ami-GS: @nibanks
Reverting https://github.com/microsoft/msquic/pull/3931 solves CIFuzz issue which has also been reported from OSS-Fuzz before integrating priority feature.
The reason only CIFuzz and OSS-Fuzz find should be due to its random algorithm based on genetic algorithm.
Do you have some idea the change has potential timing issues?

nibanks: My guess is that the combination of the that change and priority exposes some bug, most likely in spinquic, for object lifetime. But I have no idea specifically what could be causing it.

nibanks:
@ami-GS do we still need this PR?

ami-GS:
This needs to be merged. Current spinquic doesn't fully utilize priority operation feature

nibanks:
Can you please merge `main` into this so we can get the CI to run again?

src/core/api.c
```diff
@@ -796,6 +796,8 @@ MsQuicStreamStart(
     QUIC_STATUS Status;
     QUIC_STREAM* Stream;
     QUIC_CONNECTION* Connection;
+    BOOLEAN IsPriority = !!(Flags & QUIC_STREAM_START_FLAG_PRIORITY_WORK);
+    Flags &= ~QUIC_STREAM_START_FLAG_PRIORITY_WORK;
```
nibanks, line 800: Why are all these changes necessary?

ami-GS: I found this is written for other api (StreamSend, SetParam etc.), so I added. Originally to check whether this fix stall issue

PR #4363, opened 2024-06-18
install dotnet in prepare-machine.ps1

ami-GS:
## Description

Installing dotnet always hit known issue
https://learn.microsoft.com/en-us/dotnet/core/install/linux-package-mixup?pivots=os-linux-ubuntu

The cause should be because we already set repository for powershell which conflicts with the dotnet components.

## Testing

Locally done

## Documentation

N/A

### DIFF
--- a/scripts/log_wrapper.sh
+++ b/scripts/log_wrapper.sh
@@ -15,5 +15,5 @@ if [ $sessionCreated -eq 0 ]; then
     lttng destroy msquic
 
     babeltrace --names all $dirname/data > $dirname/quic.babel.txt
-    ./submodules/clog/src/clog2text/clog2text_lttng/bin/Release/net6.0/publish/clog2text_lttng -i $dirname/quic.babel.txt -s ./src/manifest/clog.sidecar -o $dirname/quic.log --showTimestamp --showCpuInfo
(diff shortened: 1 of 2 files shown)


scripts/log_wrapper.sh
```diff
@@ -15,5 +15,5 @@ if [ $sessionCreated -eq 0 ]; then
     lttng destroy msquic
 
     babeltrace --names all $dirname/data > $dirname/quic.babel.txt
-    ./submodules/clog/src/clog2text/clog2text_lttng/bin/Release/net6.0/publish/clog2text_lttng -i $dirname/quic.babel.txt -s ./src/manifest/clog.sidecar -o $dirname/quic.log --showTimestamp --showCpuInfo
+    ./submodules/clog/src/clog2text/clog2text_lttng/bin/Debug/net6.0/publish/clog2text_lttng -i $dirname/quic.babel.txt -s ./src/manifest/clog.sidecar -o $dirname/quic.log --showTimestamp --showCpuInfo
```
nibanks, line 18: We shouldn't be using debug

ami-GS: `pwsh ./scripts/prepare-machine.ps1` generates this Debug build automatically though

Maybe this

```pwsh
    dotnet build (Join-Path $RootDir submodules clog)
```

PR #4366, opened 2024-06-20
Fixed lttng calls to not to cause compilation errors.

### LINKED ISSUE #4365: Cannot build MsQuic due to lttng compilation error
### Describe the bug

Build of main MsQuic started recently failing for me with many errors like this:
```
In file included from /usr/include/lttng/tracepoint-event.h:69,
                 from /home/manicka/repositories/msquic/src/generated/linux/Tcp.cpp.clog.h:16,
                 from /home/manicka/repositories/msquic/src/generated/linux/quic.clog_Tcp.cpp.clog.h.c:7:
/home/manicka/repositories/msquic/src/generated/linux/Tcp.cpp.clog.h.lttng.h: In function ‘lttng_ust__event_probe__CLOG_TCP_CPP___PerfTcpCreateClient’:
/home/manicka/repositories/msquic/src/generated/linux/Tcp.cpp.clog.h.lttng.h:17:9: error: initialization of ‘uint64_t’ {aka ‘long unsigned int’} from ‘const void *’ makes integer from pointer without a cast [-Wint-conversion]
   17 |         ctf_integer_hex(uint64_t, arg2, arg2)
      |         ^~~~~~~~~~~~~~~
```
I'm building with lttng logging on `-DQUIC_ENABLE_LOGGING=true`, 

### Affected OS

- [ ] Windows
- [X] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

cmake version 3.29.5
clang version 17.0.6
lttng-ust 2.13.8

### MsQuic version

main

```
configuration=Release
options="-DQUIC_BUILD_TOOLS=off -DQUIC_BUILD_TEST=off -DQUIC_BUILD_PERF=off -DQUIC_ENABLE_LOGGING=true -DQUIC_USE_SYSTEM_LIBCRYPTO=true -DQUIC_TLS=openssl3"

cmake -B build -DCMAKE_BUILD_TYPE=${configuration} ${options}
cd build
cmake --build . --config ${configuration}
```

### Expected behavior

It builds.

### Actual outcome

```
In file included from /usr/include/lttng/tracepoint-event.h:69,
                 from /home/manicka/repositories/msquic/src/generated/linux/Tcp.cpp.clog.h:16,
                 from /home/manicka/repositories/msquic/src/generated/linux/quic.clog_Tcp.cpp.clog.h.c:7:
/home/manicka/repositories/msquic/src/generated/linux/Tcp.cpp.clog.h.lttng.h: In function ‘lttng_ust__event_probe__CLOG_TCP_CPP___PerfTcpCreateClient’:
/home/manicka/repositories/msquic/src/generated/linux/Tcp.cpp.clog.h.lttng.h:17:9: error: initialization of ‘uint64_t’ {aka ‘long unsigned int’} from ‘const void *’ makes integer from pointer without a cast [-Wint-conversion]
   17 |         ctf_integer_hex(uint64_t, arg2, arg2)
      |         ^~~~~~~~~~~~~~~
```

### Additional details

_No response_

ManickaP:
Fixes #4365

Not sure if this a right way to fix this, but it's the only one that worked for me. (I tried adding warning exceptions into CMake, but I don't understand it well enough to be able to make it work)

### DIFF
--- a/src/generated/linux/Tcp.cpp.clog.h.lttng.h
+++ b/src/generated/linux/Tcp.cpp.clog.h.lttng.h
@@ -12,9 +12,9 @@
 ----------------------------------------------------------*/
 TRACEPOINT_EVENT(CLOG_TCP_CPP, PerfTcpCreateClient,
     TP_ARGS(
-        const void *, arg2), 
+        const void *, arg2),
     TP_FIELDS(
-        ctf_integer_hex(uint64_t, arg2, arg2)
+        ctf_integer_hex(uint64_t, arg2, (uint64_t)arg2)
     )
 )
 
@@ -31,9 +31,9 @@ TRACEPOINT_EVENT(CLOG_TCP_CPP, PerfTcpCreateClient,
 ----------------------------------------------------------*/
 TRACEPOINT_EVENT(CLOG_TCP_CPP, PerfTcpCreateServer,
     TP_ARGS(
-        const void *, arg2), 
+        const void *, arg2),
     TP_FIELDS(
-        ctf_integer_hex(uint64_t, arg2, arg2)
+        ctf_integer_hex(uint64_t, arg2, (uint64_t)arg2)
     )
 )
 
@@ -50,9 +50,9 @@ TRACEPOINT_EVENT(CLOG_TCP_CPP, PerfTcpCreateServer,
 ----------------------------------------------------------*/
 TRACEPOINT_EVENT(CLOG_TCP_CPP, PerfTcpDestroyed,
     TP_ARGS(
-        const void *, arg2), 
+        const void *, arg2),
     TP_FIELDS(
-        ctf_integer_hex(uint64_t, arg2, arg2)
+        ctf_integer_hex(uint64_t, arg2, (uint64_t)arg2)
     )
 )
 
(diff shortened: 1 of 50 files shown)


nibanks:
As you can see from the CLOG / Validate failure, we cannot change these manually. They must be fixed upstream in microsoft/CLOG. I am curious why you are seeing these build failures, but our automation does not. The bug you opened doesn't seem to indicate what the difference might be.

ManickaP:
I did a bit more digging and it might be GCC, I'm on 14 and the docs for 13 are slightly different for this error:
- https://gcc.gnu.org/onlinedocs/gcc-13.3.0/gcc/Warning-Options.html#index-Wno-int-conversion
- https://gcc.gnu.org/onlinedocs/gcc-14.1.0/gcc/Warning-Options.html#index-Wno-int-conversion

And Ubuntu 24.04 is on gcc 13: https://packages.ubuntu.com/noble/gcc

As usual, I'm not on a "supported" distro (Arch), but most of the time I'm just ahead with the packages and they get eventually propagated to other distros as well.

Please let me know if you're willing to accept this change or not. I'll put up PR into CLOG, but I don't want to spend time on this if this would get rejected.

nibanks:
I agree with you, @ManickaP, that while we might not be hitting this yet, it's likely to come in the future, so it's good to get ahead of and fix it in CLOG. The project is mostly dotnet, so you (@ManickaP) probably can fix it faster than I can. 😄 Also, @chgray may have some time to help you with the CLOG stuff.

ManickaP:
Closing, replaced by PR in CLOG: https://github.com/microsoft/CLOG/pull/112

PR #4376, opened 2024-06-27
Add increments into STREAMS_AVAILABLE event

ManickaP:
## Description

In case the current count of available streams is bellow 0, STREAMS_AVAILABLE event will indicate 0, making it impossible to know the actual update that came over the wire.

## Alternatives

- Report the number from the frame / initial setting directly (`MaxStreams` value) instead of the increments.
- Use different naming for the properties.

## Testing

No existing tests cover the pre-existing values in this event, but this was tested with .NET runtime and System.Net.Quic that plans to take advantage of this feature for it's own stream counting (https://github.com/dotnet/runtime/commit/36210649c502c006e0bdd1439f9ba4c8a976df28)

## Documentation

Updated.

cc @MihaZupan @rzikm @CarnaViire @liveans

### DIFF
--- a/docs/api/QUIC_CONNECTION_EVENT.md
+++ b/docs/api/QUIC_CONNECTION_EVENT.md
@@ -190,6 +190,14 @@ The number of bidirectional streams the peer is willing to accept.
 
 The number of unidirectional streams the peer is willing to accept.
 
+`BidirectionalIncrement`
+
+The increment of how many additional bidirectional streams the peer is willing to accept.
+
+`UnidirectionalIncrement`
+
+The increment of how many additional of unidirectional streams the peer is willing to accept.
+
 ## QUIC_CONNECTION_EVENT_PEER_NEEDS_STREAMS
 
 This event indicates the peer is currently blocked on the number of parallel streams the app has configured it is willing to accept.
--- a/src/core/stream_set.c
+++ b/src/core/stream_set.c
@@ -235,7 +235,9 @@ QuicStreamSetDrainClosedStreams(
 _IRQL_requires_max_(PASSIVE_LEVEL)
 void
 QuicStreamSetIndicateStreamsAvailable(
-    _Inout_ QUIC_STREAM_SET* StreamSet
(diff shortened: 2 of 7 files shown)


src/core/stream_set.c
```diff
@@ -423,9 +429,10 @@ QuicStreamSetUpdateMaxStreams(
             CxPlatHashtableEnumerateEnd(StreamSet->StreamTable, &Enumerator);
         }
 
+        uint64_t increment = MaxStreams - Info->MaxTotalStreamCount;
```
nibanks, line 432:
```suggestion
        uint64_t Increment = MaxStreams - Info->MaxTotalStreamCount;
```

docs/api/QUIC_CONNECTION_EVENT.md
```diff
@@ -190,6 +190,14 @@ The number of bidirectional streams the peer is willing to accept.
 
 The number of unidirectional streams the peer is willing to accept.
 
+`BidirectionalIncrement`
+
+The increment of how many additional bidirectional streams the peer is willing to accept.
```
nibanks, line 195:
```suggestion
The count of new bidirectional streams the peer is willing to accept.
```

nibanks (commented):
Generally, the change looks fine, but I'd really like to understand more why you want this. How will this help you?

ManickaP:
> Generally, the change looks fine, but I'd really like to understand more why you want this. How will this help you?

We're exposing a callback similar to STREAMS_AVAILABLE event in System.Net.Quic (https://github.com/dotnet/runtime/issues/101534). The main use is to support opening of new connection when the limit is reached on an existing one. But we also want to know when we can start re-using the original one. 

The goal of this change is to report correct numbers, which is not possible due to https://github.com/microsoft/msquic/blob/f5bec530c8ef0824df7294f11d7b3516919899b1/src/core/stream_set.c#L482-L484

I'm also not happy about the uint_16.max cut off for the max value, so I'm starting to get more inclined to expose the raw number from MAX_STREAMS frame in here.

PR #4182, opened 2024-03-08
Multiple Receive API Support

nibanks:
## Description

Adds a settings for apps to configure streams to support multiple parallel receives.

## Testing

TODO

## Documentation

TODO

PR #4383, opened 2024-07-01
Stream Multi Receive implementation

ami-GS:
## Description

Stream Multi Receive internal feature.
Dispatched from https://github.com/microsoft/msquic/pull/4182

## Testing

RecvBufferTest.cpp

## Documentation

N/A

### DIFF
--- a/src/core/recv_buffer.c
+++ b/src/core/recv_buffer.c
@@ -94,6 +94,7 @@ QuicRecvBufferInitialize(
     RecvBuffer->HasDataOnLeft = FALSE;
     RecvBuffer->LockFirstChunk = FALSE;
     RecvBuffer->Shrunk1stChunkLength = 0;
+    RecvBuffer->LockedOffset = UINT32_MAX;
     RecvBuffer->VirtualBufferLength = VirtualBufferLength;
     RecvBuffer->RecvMode = RecvMode;
     Status = QUIC_STATUS_SUCCESS;
@@ -282,6 +283,7 @@ QuicRecvBufferResize(
     //
 
     if (RecvBuffer->RecvMode == QUIC_RECV_BUF_MODE_MULTIPLE) {
+        RecvBuffer->LockedOffset = RecvBuffer->BaseOffset;
         return TRUE;
     }
 
@@ -361,8 +363,6 @@ QuicRecvBufferGetTotalAllocLength(
     // the first chunk has already been drained, so we don't use the allocated
     // length, but ReadLength instead when calculating total available space.
     //
-    // uint32_t AllocLength = RecvBuffer->ReadLength;
-    // uint32_t AllocLength = Chunk->AllocLength;
     uint32_t AllocLength = RecvBuffer->ReadLength;
     while (Chunk->Link.Flink != &RecvBuffer->Chunks) {
         Chunk =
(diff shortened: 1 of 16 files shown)


nibanks:
Just include the recv buffer and test changes? Not everything else

ami-GS:
settings.c and stream.c etc. are for public api?

src/core/recv_buffer.h
```diff
@@ -20,7 +20,7 @@ typedef enum QUIC_RECV_BUF_MODE {
 //
 typedef struct QUIC_RECV_CHUNK {
     CXPLAT_LIST_ENTRY Link;         // Link in the list of chunks.
-    uint32_t AllocLength : 31;      // Allocation size of Buffer
+    uint32_t AllocLength : 31;      // Allocation size of Buffer,
```
nibanks, line 23: nit: revert

src/core/recv_buffer.h
```diff
@@ -68,6 +68,12 @@ typedef struct QUIC_RECV_BUFFER {
     //
     uint32_t VirtualBufferLength;
 
+    //
+    // Basically same as Chunk->AllocLength of first chunk,
+    // but start shrinking by drain operation after next chunk is allocated.
```
nibanks, line 73: nit: please fill up first line (to 80 column) before wrap

src/core/unittest/RecvBufferTest.cpp
```diff
@@ -59,6 +59,7 @@ struct RecvBuffer {
         _Out_ BOOLEAN* NewDataReady
         ) {
         auto BufferToWrite = new (std::nothrow) uint8_t[WriteLength];
+        memset(BufferToWrite, 0, WriteLength); // Zero out the buffer (for debugging purposes
```
nibanks, line 62:
```suggestion
        memset(BufferToWrite, 0, WriteLength); // Zero out the buffer (for debugging purposes)
```

src/core/unittest/RecvBufferTest.cpp
```diff
@@ -560,8 +563,14 @@ TEST_P(WithMode, ReadPendingMultiWrite)
     RecvBuf.Read(&ReadOffset, &BufferCount, ReadBuffers);
     ASSERT_FALSE(RecvBuf.HasUnreadData());
     ASSERT_EQ(32ull, ReadOffset);
-    ASSERT_EQ(1u, BufferCount);
-    ASSERT_EQ(48u, ReadBuffers[0].Length);
+    if (Mode == QUIC_RECV_BUF_MODE_MULTIPLE) {
```
nibanks, line 566: Why is this needed?

src/core/unittest/RecvBufferTest.cpp
```diff
@@ -693,6 +702,767 @@ TEST_P(WithMode, DrainFrontChunkWithPendingGap)
     ASSERT_TRUE(RecvBuf.Drain(1));
 }
 
+void Check(
```
nibanks, line 705: Move these functions into RecvBuffer

src/core/unittest/RecvBufferTest.cpp
ami-GS: Default AllocLength is 64. Then

Write 0,32
Read
Write 32,48 (append next chunk)
Drain 32
Read -> 32 from first chunk, 16 from second chunk.

src/core/recv_buffer.c
```diff
@@ -419,58 +421,100 @@ QuicRecvBufferCopyIntoChunks(
                 RecvBuffer->Chunks.Flink, // First chunk
                 QUIC_RECV_CHUNK,
                 Link);
-        uint32_t ChunkLength;
+
+        uint64_t ChunkLength;
```
nibanks, line 425: Does this really need to be 64 bits? We only use 31 bits for the allocation length of a chunk. So it should be impossible for ChunkLength to be bigger, right?

ami-GS:
@nibanks This PR is ready to go. there is stall issue, but it is with API implementation

nibanks (approved):

PR #4385, opened 2024-07-03
Handle error from RawDataPathInitialize

ami-GS:
## Description

Error from RawDataPathInitialize has been ignored because it is optional.
Handle the error as the raw datapath is now user opt-in.

## Testing

See automation

## Documentation

N/A

### DIFF
--- a/src/platform/unittest/DataPathTest.cpp
+++ b/src/platform/unittest/DataPathTest.cpp
@@ -553,6 +553,7 @@ struct CxPlatSocket {
                 // wait for an event set by ResolveRouteComplete.
                 //
                 EXPECT_EQ(InitStatus, QUIC_STATUS_SUCCESS);
+                EXPECT_TRUE(CxPlatSocketRawSocketAvailable(Socket));
             }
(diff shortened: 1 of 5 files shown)


nibanks (commented):
This change seems ok, but for our test code, it explicitly needs to test the XDP datapaths. So it needs to know if this failed. What do we do for that? I think this PR should update the tests to validate XDP is used when expected.

src/platform/unittest/DataPathTest.cpp
```diff
@@ -668,11 +672,13 @@ TEST_F(DataPathTest, Initialize)
         VERIFY_QUIC_SUCCESS(Datapath.GetInitStatus());
         ASSERT_NE(nullptr, Datapath.Datapath);
     }
+    if (UseDuoNic)
     {
```
nibanks, line 676:
```suggestion
    if (UseDuoNic) {
```

nibanks (dismissed):

nibanks (approved):

PR #4387, opened 2024-07-05
QTIP case only delete raw socket

ami-GS:
## Description

QTIP case has been failing recently.
segfault by SocketProc->RefCount become negative value, but SocketProc under normal socket is not touched at all for QTIP case.

## Testing

See automation.

## Documentation

N/A

### DIFF
--- a/src/core/quicdef.h
+++ b/src/core/quicdef.h
@@ -348,7 +348,7 @@ CXPLAT_STATIC_ASSERT(
 // The minimum number of bytes of send allowance we must have before we will
 // send another packet.
 //
-#define QUIC_MIN_SEND_ALLOWANCE                 76
+#define QUIC_MIN_SEND_ALLOWANCE                 76  // Magic number to indicate a threshold of 'enough' allowance to send another packet.
 
 //
 // The minimum buffer space that we require before we will pack another
--- a/src/core/quicdef.h
+++ b/src/core/quicdef.h
@@ -348,7 +348,7 @@ CXPLAT_STATIC_ASSERT(
 // The minimum number of bytes of send allowance we must have before we will
 // send another packet.
 //
-#define QUIC_MIN_SEND_ALLOWANCE                 75
+#define QUIC_MIN_SEND_ALLOWANCE                 76
 
 //
 // The minimum buffer space that we require before we will pack another
--- a/src/platform/datapath_winuser.c
+++ b/src/platform/datapath_winuser.c
@@ -1167,7 +1167,7 @@ SocketCreateUdp(
     if (Config->Flags & CXPLAT_SOCKET_FLAG_PCP) {
         Socket->PcpBinding = TRUE;
     }
-    CxPlatRefInitializeEx(&Socket->RefCount, SocketCount);
+    CxPlatRefInitializeEx(&Socket->RefCount, Socket->UseTcp ? 1 : SocketCount);
 
     if (Datapath->UseTcp) {
         //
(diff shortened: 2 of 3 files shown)


ami-GS:
Datagram->Length become too small. only server side. looks like Path->Allowance logic relates.
`(27 >= 14 + 16)`
https://github.com/microsoft/msquic/blob/db66ab98e3ed32bdacd8ce7d9dd226cc25a0def6/src/core/packet_builder.c#L71

ami-GS:
@nibanks 

quic_datapath.h does only care about UDP case (smaller header size). Is this expected?
https://github.com/microsoft/msquic/blob/41c7eb5ac893677f468d084bd1b3b1c9281cf6ac/src/inc/quic_datapath.h#L92-L102

nibanks:
> @nibanks
> 
> quic_datapath.h does only care about UDP case (smaller header size). Is this expected?
> 
> https://github.com/microsoft/msquic/blob/41c7eb5ac893677f468d084bd1b3b1c9281cf6ac/src/inc/quic_datapath.h#L92-L102

I don't understand the question. It's returning UDP payload size.

ami-GS:
This second parameter is 75, `NewDatagramLength` get assigned as 27. The difference is 48 which is IPv6 + UDP header size.
The 27 violates size requirements in `QuicPacketBuilderValidate`.
TCP header is bigger than UDP, but there is no place using CXPLAT_TCP_HEADER_SIZE anywherein code. I'm suspecting some kind of accumulated small gap emerges now

https://github.com/microsoft/msquic/blob/41c7eb5ac893677f468d084bd1b3b1c9281cf6ac/src/core/packet_builder.c#L283-L286

ami-GS:
hmm, maybe the difference is adjusted by MTU size?.

Another difference is that when QTIP calls `QuicPacketBuilderPrepareForStreamFrames`, normal XDP calls `QuicPacketBuilderPrepareForControlFrames`.
SendFlags seems to be messed up?

https://github.com/microsoft/msquic/blob/41c7eb5ac893677f468d084bd1b3b1c9281cf6ac/src/core/send.c#L1347-L1380

ami-GS:
> hmm, maybe the difference is adjusted by MTU size?.
> 
> Another difference is that when QTIP calls `QuicPacketBuilderPrepareForStreamFrames`, normal XDP calls `QuicPacketBuilderPrepareForControlFrames`. SendFlags seems to be messed up?
> 
> https://github.com/microsoft/msquic/blob/41c7eb5ac893677f468d084bd1b3b1c9281cf6ac/src/core/send.c#L1347-L1380

Not related. timing issue

ami-GS:
QTIP SendFlag is 0x0, XDP SendFlag is 0x080000103

0x1                QUIC_CONN_SEND_FLAG_ACK
0x2                QUIC_CONN_SEND_FLAG_CRYPTO
0x100            QUIC_CONN_SEND_FLAG_NEW_CONNECTION_ID
0x80000000  QUIC_CONN_SEND_FLAG_DPLPMTUD

ami-GS:
HERE,
Path->Allowance is 74 using XDP, 75 for QTIP. 1 byte difference.

https://github.com/microsoft/msquic/blob/41c7eb5ac893677f468d084bd1b3b1c9281cf6ac/src/core/send.c#L1287-L1294

ami-GS:
Looking for 1 byte diff.....

ami-GS:
Suspicious difference is here at very beginning

image

or here
image

nibanks:
Do we have a problem with MTU calculation for QTIP? Because we're calculating with the UDP header? But if so, why is this just now a problem?

ami-GS:
MTU calculation should be fine. This is really rare corner case. I totally have no idea why this start happening now.

The gap seems to start client TP. The difference seems to be only MTU. this makes 1 byte diff?

image

ami-GS:
set QUIC_MIN_SEND_ALLOWANCE to 74 makes normal XDP test to fail as well.

ami-GS:
Normal socket also uses this condition for exit flush loop.
@nibanks Do you remember how is this `QUIC_MIN_SEND_ALLOWANCE` decided?

https://github.com/microsoft/msquic/blob/41c7eb5ac893677f468d084bd1b3b1c9281cf6ac/src/core/send.c#L1287-L1294

ami-GS:
> Do we have a problem with MTU calculation for QTIP? Because we're calculating with the UDP header? But if so, why is this just now a problem?

This PR might be the reason...? Looks like we had not run QTIP test cases appropriately. I saw the issue started from mid of Jun.
https://github.com/microsoft/msquic/pull/4346

src/core/quicdef.h
```diff
@@ -348,7 +348,7 @@ CXPLAT_STATIC_ASSERT(
 // The minimum number of bytes of send allowance we must have before we will
 // send another packet.
 //
-#define QUIC_MIN_SEND_ALLOWANCE                 75
+#define QUIC_MIN_SEND_ALLOWANCE                 76
```
nibanks, line 351:
```suggestion
#define QUIC_MIN_SEND_ALLOWANCE                 76  // Magic number to indicate a threshold of 'enough' allowance to send another packet.
```

nibanks (approved):

PR #4401, opened 2024-07-16
[release/2.3] Add memory alignment for ARM32

liveans:
## Description

Backport of #4208 

The fix was for an alignment problem -> arm is sensitive in places where  x64 is not.

## Testing

Manually tested

## Documentation

No

/cc @ManickaP


ProjectsByJackHe (approved):
It would be great to fill in some details missing from the original fix about why this works, but otherwise LGTM!

nibanks (approved):

PR #4434, opened 2024-08-06
Fix package distribution. Specify OS explicitly

ami-GS:
## Description

Checking artifact directory name doesn't work for identifying Ubuntu version for XDP build.

## Testing

locally done and used internal repo

## Documentation

N/A

### DIFF
--- a/scripts/make-packages.sh
+++ b/scripts/make-packages.sh
@@ -14,7 +14,7 @@ CONFIG=Release
 NAME=libmsquic
 TLS=openssl
 TLSVERSION=1.1
-UBUNTU=
+XDP="False"
 CONFLICTS=
 DESCRIPTION="Microsoft implementation of the IETF QUIC protocol"
 VENDOR="Microsoft"
@@ -86,9 +86,9 @@ while :; do
             shift
             OUTPUT=$1
             ;;
-        -u|-ubuntu|--ubuntu)
+        -x|-xdp|--xdp)
             shift
-            UBUNTU=$1
+            XDP=$1
             ;;
         -t|-tls|--tls)
             shift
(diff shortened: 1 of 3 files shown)


csujedihy (approved):

scripts/make-packages.sh
```diff
@@ -184,6 +184,12 @@ if [ "$OS" == "linux" ]; then
      FILES="${FILES} ${ARTIFACTS}/libmsquic.lttng.${LIBEXT}.${VER_MAJOR}.${VER_MINOR}.${VER_PATCH}=/usr/${LIBDIR}/libmsquic.lttng.${LIBEXT}.${VER_MAJOR}.${VER_MINOR}.${VER_PATCH}"
   fi
 
+  # XDP means Ubuntu 24.04 which installs libssl3t64 instead of libssl3
```
csujedihy, line 187: Is there a case where we build a package for ubuntu 24.04 without depending on XDP?

scripts/make-packages.sh
```diff
@@ -183,7 +184,14 @@ if [ "$OS" == "linux" ]; then
      FILES="${FILES} ${ARTIFACTS}/libmsquic.lttng.${LIBEXT}.${VER_MAJOR}.${VER_MINOR}.${VER_PATCH}=/usr/${LIBDIR}/libmsquic.lttng.${LIBEXT}.${VER_MAJOR}.${VER_MINOR}.${VER_PATCH}"
   fi
 
-  if [ "$UBUNTU" == '2404' ]; then
+  # XDP means Ubuntu 24.04 which installs libssl3t64 instead of libssl3
```
liveans, line 187: I would call this time64Distro or something like that actually, because there might be some cases in the future where you might not want to use XDP and get rid of it as dependency on different distros but might still require you to get this t64 version if libssl3

scripts/package-distribution.ps1
```diff
@@ -167,17 +180,13 @@ foreach ($Build in $AllBuilds) {
...
+
         if ($BuildBaseName -like "*arm64_*") {
-            & $RootDir/scripts/make-packages.sh --output $DistDir --arch arm64 --tls $Tls --ubuntu $UbuntuVersion
+            & $RootDir/scripts/make-packages.sh --output $DistDir --arch arm64 --tls $Tls --xdp $UseXdp --time64 $Time64Distro
         } elseif ($BuildBaseName -like "*arm_*") {
-            & $RootDir/scripts/make-packages.sh --output $DistDir --arch arm --tls $Tls --ubuntu $UbuntuVersion
+            & $RootDir/scripts/make-packages.sh --output $DistDir --arch arm --tls $Tls --xdp $UseXdp --time64 $Time64Distro
         } else {
-            & $RootDir/scripts/make-packages.sh --output $DistDir --tls $Tls --ubuntu $UbuntuVersion # x64
+            & $RootDir/scripts/make-packages.sh --output $DistDir --tls $Tls --xdp $UseXdp --time64 $Time64Distro # x64
```
liveans, line 189: like I said, this is only requirement for armhf for now, but I believe it's fine to keep it as-is, since all architectures of Ubuntu24.04 is already installing `libssl3t64` if you try to install `libssl3`

liveans (approved):

scripts/package-distribution.ps1
csujedihy, line 189: Honestly, I don't see a fundamental difference between UbuntuVersion when it's ubuntu and UseXdp and Time64Distro. They are basically the same thing and make-packages.sh actually assumes that XDP is ubuntu 24.04.

They knobs would make sense if you pass the correct value to them from here.

For example, for arm, we don't pass `--xdp` but we pass `--time64` because we don't build xdp for arm but if `$Time64Distro` is true, we use the t64 version of libssl3.

The current code basically creates 2 aliases for ubuntu24.04.

PR #4435, opened 2024-08-06
Fix package distribution. Specify OS explicitly (#4434)

ami-GS:
* Pass OS information explicitly to package-distribution

* libssl3 variants for ubuntu24.04

* detail flag for time64

## Description

## Testing

## Documentation


ami-GS:
The issue happens before actually start spinquic function for some reason.
The dump is not uploaded due to permission issue, so there is no way to debug right now.

image

image

nibanks (approved):

PR #4446, opened 2024-08-08
Lock OperQ before checking content

ami-GS:
## Description

There is stall issue which maybe related to multi recv feature.
Try adding lock to OperQ.

## Testing

see automation. might need to run may times.

## Documentation

### DIFF
--- a/src/core/worker.c
+++ b/src/core/worker.c
@@ -564,7 +564,10 @@ QuicWorkerProcessConnection(
     if (!Connection->State.UpdateWorker) {
         if (Connection->HasQueuedWork) {
             Connection->Stats.Schedule.LastQueueTime = CxPlatTimeUs32();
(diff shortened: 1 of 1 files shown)


src/core/worker.c
```diff
@@ -564,7 +564,10 @@ QuicWorkerProcessConnection(
     if (!Connection->State.UpdateWorker) {
         if (Connection->HasQueuedWork) {
             Connection->Stats.Schedule.LastQueueTime = CxPlatTimeUs32();
-            if (&Connection->OperQ.List.Flink != Connection->OperQ.PriorityTail) {
```
nibanks, line 567: You should make a helper function in the operation interface.

nibanks:
I also found that `QuicWorkerMoveConnection` isn't correctly taking the connection's priority into account.

nibanks (approved):

PR #4447, opened 2024-08-08
Key Update Crash on Allocation Failure

nibanks:
## Description

While running spinquic locally I hit a crash in the keyupdate scenario where write key failed to allocate and was later tried to be read. This accounts for that edge case now.

## Testing

Covered by spinquic. Also added a QuicDrill test.

## Documentation

N/A


src/test/bin/quic_gtest.cpp
```diff
@@ -2298,6 +2298,15 @@ TEST_P(WithDrillInitialPacketTokenArgs, QuicDrillTestServerVNPacket) {
     }
 }
 
+TEST_P(WithDrillInitialPacketTokenArgs, QuicDrillTestKeyUpdateDuringHandshake) {
+    TestLoggerT<ParamType> Logger("QuicDrillTestKeyUpdateDuringHandshake", GetParam());
+    if (TestingKernelMode) {
+        //ASSERT_TRUE(DriverClient.Run(IOCTL_QUIC_RUN_DRILL_VN_PACKET_TOKEN, GetParam().Family));
```
nibanks, line 2304: Need to add kernel support.

anrossi (approved):

src/test/lib/DrillDescriptor.cpp
```diff
@@ -228,12 +230,123 @@ DrillInitialPacketDescriptor::write(
...
+    ) const
+{
+    DrillBuffer PacketBuffer;
+    QuicShortHeader Header = { 0 };
+    Header.PacketNumLen = 3;
+    Header.KeyPhase = KeyPhase;
+
+    PacketBuffer.push_back(Header.HeaderByte);
+    PacketBuffer.insert(PacketBuffer.end(), DestCid.begin(), DestCid.end());
+    PacketBuffer.push_back((uint8_t) (PacketNumber >> 24));// TODO - different packet number sizes
```
guhetier, line 345: nit: Should the todo be addressed now?

guhetier (approved):

PR #4454, opened 2024-08-12
Add ref option to workflow dispatch

ProjectsByJackHe:
## Description

While triaging a kernel test failure in MsCodeHub mirror from the latest ingestion, 839a2fbde9a2cecdcf0c41649037fa091cbb0e06,
it became apparent that reproducing the failure (kernel mode test) would be challenging using the current CI setup. 

This PR makes it so a developer can target a specific version of MsQuic while running the BVT test suite. 

## Testing

CI

## Documentation

N/A

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,6 +2,11 @@ name: BVT
 
 on:
   workflow_dispatch:
+    inputs:
+      ref:
+        required: false
+        default: ''
+        type: string
   push:
     branches:
     - main
@@ -37,6 +42,7 @@ jobs:
       arch: ${{ matrix.vec.arch }}
       tls: ${{ matrix.vec.tls }}
       build: ${{ matrix.vec.build }}
+      ref: ${{ inputs.ref || '' }}
 
   build-windows:
     name: Build WinUser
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -2,6 +2,11 @@ name: BVT
 
 on:
   workflow_dispatch:
+    inputs:
+      ref:
```
nibanks, line 6: Do you really need this ref? Shouldn't it work automatically?

ProjectsByJackHe: What I am trying to do is target a specific version to repro any failures. Most recently, version MsQuic 2.4.3 seems to be failing the Kernel mode test `Tests::RunKernelMode: Basic/WithFamilyArgs.ClientBlockedSourcePort/1 failed `. 

What I am also considering is perhaps adding options to target specific tests, on specific platforms, and running it N times.

ProjectsByJackHe:
Overall, having the option to configure and target which version of MsQuic (by commit hash) to run BVT on in a workflow dispatch I believe to be a useful utility. 

Thoughts?

nibanks (approved):

PR #4455, opened 2024-08-12
Publish package to more distros and refactoring

ami-GS:
## Description

- Add more distros. fedora39, 49, azure linux 3.0
- Correct openssl dependencies (e.g. fedora 37, debian 12)
- Refactoring
- Stop using ubuntu_2X04 keyword for differentiate builds. Use kernel version instead.

## Testing

Use internal pipeline

## Documentation

N/A

### DIFF
--- a/.azure/obtemplates/build-linux-packages.yml
+++ b/.azure/obtemplates/build-linux-packages.yml
@@ -46,15 +46,9 @@ jobs:
       find -name "*.tar" -exec tar -xvf '{}' \;
   - task: PowerShell@2
     displayName: Distribution
-    ${{ if not(eq(parameters.xdp, '')) }}:
-      inputs:
-        pwsh: false
-        filePath: scripts/package-distribution.ps1
-        arguments: -UseXdp
-    ${{ else }}:
-      inputs:
-        pwsh: false
-        filePath: scripts/package-distribution.ps1
+    inputs:
+      pwsh: false
+      filePath: scripts/package-distribution.ps1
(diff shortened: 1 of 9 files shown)


.azure/obtemplates/build-linux-packages.yml
```diff
@@ -7,19 +7,16 @@ jobs:
...
       ubuntu_2404:
-        os: ubuntu_2404
-        tls: openssl3
+        kernel: "kernel6_8"
   pool:
     type: linux
   variables:
     ob_outputDirectory: $(Build.SourcesDirectory)/artifacts/dist
-    ob_artifactSuffix: _$(os)_$(tls)
+    ob_artifactSuffix: _$(kernel) # drop_package_linux_${{ jobname }}_${{ kernel }}
```
nibanks, line 19: Where does `${{ jobname }}` come from?

ami-GS: line 5.

nibanks (approved):

scripts/package-distribution.ps1
```diff
@@ -141,6 +115,17 @@ foreach ($Build in $AllBuilds) {
         $Libraries += Join-Path $ArtifactsDir "msquic.lib"
     }
 
+    # if datapath_raw_xdp_kern.o exists under $ArtifactsDir, $UseXdp to be true
```
liveans, line 118: Yes, this is correct but we shouldn't tie $Time64Distro to this, because for example: Ubuntu 24.04 is still Time64Distro but if we're compiling it against ARM, dependency will get converted to libssl3 from libssl3t64

PR #4458, opened 2024-08-13
Add option to configure TCP execution context

ProjectsByJackHe:
## Description

As of today, MsQuic has several execution modes, namely Max Throughput, and Low Latency.

We use the Secnetperf tool to measure this performance impact between the different modes.
However, with our Secnetperf TCP implementation, there is no such distinction between modes.

Meaning, we are artificially giving QUIC an advantage in performance when we run it in Lowlat mode to optimize thread / context switching, but we are not doing the same for TCP.

This PR starts the process to give TCP that same advantage. 

## Testing

We will leverage Netperf CI on Azure / Lab + dashboard to see if there is any significant differences in perf when executing in the different modes.

## Documentation

N/A


src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -22,6 +22,7 @@ PerfClient* Client;
 
 uint32_t MaxRuntime = 0;
 QUIC_EXECUTION_PROFILE PerfDefaultExecutionProfile = QUIC_EXECUTION_PROFILE_LOW_LATENCY;
+TCP_EXECUTION_PROFILE TcpDefaultExecutionProfile = TCP_EXECUTION_PROFILE_MAX_THROUGHPUT;
```
nibanks, line 25: They should default to the same thing.

src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -193,14 +194,18 @@ QuicMainStart(
     if (ExecStr != nullptr) {
         if (IsValue(ExecStr, "lowlat")) {
             PerfDefaultExecutionProfile = QUIC_EXECUTION_PROFILE_LOW_LATENCY;
+            TcpDefaultExecutionProfile = TCP_EXECUTION_PROFILE_LOW_LATENCY;
         } else if (IsValue(ExecStr, "maxtput")) {
             PerfDefaultExecutionProfile = QUIC_EXECUTION_PROFILE_TYPE_MAX_THROUGHPUT;
+            TcpDefaultExecutionProfile = TCP_EXECUTION_PROFILE_MAX_THROUGHPUT;
         } else if (IsValue(ExecStr, "scavenger")) {
             PerfDefaultExecutionProfile = QUIC_EXECUTION_PROFILE_TYPE_SCAVENGER;
+            // TODO: Should we add a scavenger profile for TCP?
```
nibanks, line 203: No need atm. Just default to lowlat.

src/perf/lib/SecNetPerfMain.cpp
ProjectsByJackHe: Done

src/perf/lib/SecNetPerfMain.cpp
ProjectsByJackHe: Noted

src/perf/lib/PerfClient.cpp
```diff
@@ -229,7 +229,8 @@ PerfClient::Init(
                 nullptr,
                 PerfClientConnection::TcpConnectCallback,
                 PerfClientConnection::TcpReceiveCallback,
-                PerfClientConnection::TcpSendCompleteCallback));
+                PerfClientConnection::TcpSendCompleteCallback,
+                TCP_EXECUTION_PROFILE_MAX_THROUGHPUT)); // TODO: Client should always be max tput?
```
nibanks, line 233: Client should default to lowlat but be configurable by command line too

ProjectsByJackHe: ACK, updated.

src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -193,14 +194,16 @@ QuicMainStart(
...
         } else if (IsValue(ExecStr, "maxtput")) {
             PerfDefaultExecutionProfile = QUIC_EXECUTION_PROFILE_TYPE_MAX_THROUGHPUT;
+            TcpDefaultExecutionProfile = TCP_EXECUTION_PROFILE_MAX_THROUGHPUT;
         } else if (IsValue(ExecStr, "scavenger")) {
             PerfDefaultExecutionProfile = QUIC_EXECUTION_PROFILE_TYPE_SCAVENGER;
         } else if (IsValue(ExecStr, "realtime")) {
             PerfDefaultExecutionProfile = QUIC_EXECUTION_PROFILE_TYPE_REAL_TIME;
         } else {
-            WriteOutput("Failed to parse execution profile[%s], use lowlat as default\n", ExecStr);
+            WriteOutput("Failed to parse execution profile[%s], use lowlat as default for QUIC, maxtput as default for TCP.\n", ExecStr);
```
nibanks, line 206: This can be reverted now, right?

ProjectsByJackHe: Good catch! Updated error message.

nibanks (approved):

Issue #4450, opened 2024-08-09
libmsquic missing from Azure Linux 3.0


richlander:
### Describe the feature you'd like supported

```bash
root [ / ]# cat /etc/os-release | grep PRETTY_
PRETTY_NAME="Microsoft Azure Linux 3.0"
root [ / ]# tdnf install libmsquic
Loaded plugin: tdnfrepogpgcheck
libmsquic package not found or not installed
Error(1011) : No matching packages
```

### Proposed solution

Make the package available.

ami-GS:
@richlander 
Do you have docker image?
I found this `mcr.microsoft.com/cbl-mariner/base/core:2.0`, but cannot find 3.0

richlander:
Here is the Dockerfile we just merged.

https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/main/src/azurelinux/3.0/helix/Dockerfile

richlander:
FYI: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1170

ami-GS:
Generally How can we find repository name?
 
Azure pipeline cannot find `cbl-mariner-3.0-prod-Microsoft-x86_64-yum`.
I just incremented the name from `cbl-mariner-2.0-prod-Microsoft-x86_64-yum` which has been working

jslobodzian:
In 3.0 there were two important changes:
1) Rebranding as "azure-linux"
2) repository name for "Microsoft" changed names to "ms-oss" (Microsoft Open Source Software) to more clearly delineate it from "ms-non-oss" (which was previously known as Extras).

The repositories are here: https://packages.microsoft.com/azurelinux/3.0/prod/ms-oss/

ami-GS:
Thanks, but what is the associated repository name, previously named as `cbl-mariner-2.0-prod-Microsoft-x86_64-yum`?

ami-GS:
maybe combining path by `-`?
https://packages.microsoft.com/cbl-mariner/2.0/prod/Microsoft/x86_64/ -> `cbl-mariner-2.0-prod-Microsoft-x86_64-yum`
https://packages.microsoft.com/azurelinux/3.0/prod/ms-oss/x86_64/ -> `azurelinux-3.0-prod-ms-oss-x86_64-yum`
?

ami-GS:
assumption is correct, `azurelinux-3.0-prod-ms-oss-x86_64-yum` works

[...]

PR #4459, opened 2024-08-14
Make rpm package for x86_64

ami-GS:
## Description

rpm package generation is skipped when `(!xdp && x86_64)`. remove the check.
https://github.com/microsoft/msquic/issues/4450

## Testing

running internal CI

## Documentation

N/A

### DIFF
--- a/.azure/OneBranch.Publish.yml
+++ b/.azure/OneBranch.Publish.yml
@@ -76,6 +76,7 @@ parameters:
   type: object
   default:
   - azurelinux-3.0-prod-ms-oss-x86_64-yum # 3.0 6.6
+  - azurelinux-3.0-prod-ms-oss-aarch64-yum # 3.0 6.6
 - name: kernel6_8rpmrepos
   type: object
   default:
--- a/.azure/OneBranch.Publish.yml
+++ b/.azure/OneBranch.Publish.yml
@@ -15,6 +15,10 @@ variables:
   DisableDockerDetector: true
 
 parameters:
+- name: debug # debug mode will not actually upload and publish packages
+  type: boolean
+  default: false
+
   # build on ubuntu 20.04 openssl 1.1
 - name: kernel5_4rpmcblrepos
   type: object
--- a/scripts/make-packages.sh
+++ b/scripts/make-packages.sh
@@ -160,8 +160,8 @@ if [ "$OS" == "linux" ]; then
       --provides ${NAME} \
       --depends "libcrypto.so.${TLSVERSION}()(${BITS})" \
       --depends "libnuma.so.1()(${BITS})" \
-      --depends "libxdp.so.1.4.0" \
-      --depends "libnl-route-3.so.200" \
+      --depends "libxdp >= 1.4.0" \
+      --depends "libnl3 >= 3.0" \
       --conflicts ${CONFLICTS} \
       --version ${VER_MAJOR}.${VER_MINOR}.${VER_PATCH} \
       --description "${DESCRIPTION}" \
--- a/scripts/make-packages.sh
+++ b/scripts/make-packages.sh
@@ -139,62 +139,60 @@ echo "ARCH=$ARCH PKGARCH=$PKGARCH ARTIFACTS=$ARTIFACTS"
 mkdir -p ${OUTPUT}
 
 if [ "$OS" == "linux" ]; then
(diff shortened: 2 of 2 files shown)


scripts/make-packages.sh
```diff
@@ -139,62 +139,60 @@ echo "ARCH=$ARCH PKGARCH=$PKGARCH ARTIFACTS=$ARTIFACTS"
...
+    fpm \
+      --force \
+      --input-type dir \
+      --output-type rpm \
+      --architecture ${PKGARCH} \
+      --name ${NAME} \
+      --provides ${NAME} \
+      --depends "libcrypto.so.${TLSVERSION}()(${BITS})" \
+      --depends "libnuma.so.1()(${BITS})" \
+      --depends "libxdp.so.1.4.0" \
```
liveans, line 163: Why there is a hard dependency on 1.4 version of libxdp?

ami-GS: I don't have any other environment to check the libraries. at least x64 depends on it

liveans: Yes, but I believe this should depend on 'libxdp.so.1', right? It will be symbolic link to latest version of libxdp anyway

ami-GS: I can finally check on azure linux 3.0 with this change. wait a sec

.azure/OneBranch.Publish.yml
```diff
@@ -15,6 +15,10 @@ variables:
   DisableDockerDetector: true
 
 parameters:
+- name: debug # debug mode will not actually upload and publish packages
```
nibanks, line 18: Is this automatically used somewhere?

ami-GS: Line 119 uses it. I accidentally deleted it

scripts/make-packages.sh
```diff
@@ -139,62 +139,60 @@ echo "ARCH=$ARCH PKGARCH=$PKGARCH ARTIFACTS=$ARTIFACTS"
...
+      --force \
+      --input-type dir \
+      --output-type rpm \
+      --architecture ${PKGARCH} \
+      --name ${NAME} \
+      --provides ${NAME} \
+      --depends "libcrypto.so.${TLSVERSION}()(${BITS})" \
+      --depends "libnuma.so.1()(${BITS})" \
+      --depends "libxdp >= 1.4.0" \
+      --depends "libnl3 >= 3.0" \
```
wfurt, line 164: do we need libnl or is that dependency libxdp will drag in? I would put here as little as really needed.

ami-GS: libnl is needed when using xdp (raw socket). to resolve route to next hop.

nibanks: Yeah, MsQuic itself resolves the route when using XDP. It's not an XDP dependency.

nibanks (dismissed):
If it works, ship it!

scripts/make-packages.sh
```diff
@@ -139,62 +139,60 @@ echo "ARCH=$ARCH PKGARCH=$PKGARCH ARTIFACTS=$ARTIFACTS"
...
-        --description "${DESCRIPTION}" \
-        --vendor "${VENDOR}" \
-        --maintainer "${MAINTAINER}" \
-        --package "${OUTPUT}" \
-        --license MIT \
-        --url https://github.com/microsoft/msquic \
-        --log error \
-        ${FILES}
-    fi
+  if [ "$XDP" == "True" ] && [[ "$ARCH" == x* ]]; then
```
wfurt, line 152: do we need the arch check? XDP is off by default so I would think we should package it when we are told to. e.g. the distro/arch choices should not be here IMHO.

ami-GS: arm build doesn't enable xdp at build time. so this if else is needed

ami-GS:
I just realized that Publish pipeline for kernel 6.8 for rpm package only publish arm packages

kernel 6.8
image

kernel 5.15
image

ami-GS:
> I just realized that Publish pipeline for kernel 6.8 for rpm package only publish arm packages
> 
> kernel 6.8 image
> 
> kernel 5.15 image

oops, this PR itself resolve this problem🤦‍♂️

nibanks (approved):

PR #4461, opened 2024-08-15
Fix the Enums for the WPA Plugin

nibanks:
## Description

They got out of sync with the code.

## Testing

Will install new plugin

## Documentation

N/A

### DIFF
--- a/src/plugins/trace/dll/DataModel/QuicEvents.cs
+++ b/src/plugins/trace/dll/DataModel/QuicEvents.cs
@@ -32,7 +32,6 @@ namespace QuicTrace.DataModel
         ConnectionStart,
         ConnectionSetConfiguration,
         ConnectionSendResumptionTicket,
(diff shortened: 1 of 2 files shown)


src/plugins/trace/dll/DataModel/QuicEvents.cs
```diff
@@ -32,15 +32,16 @@ public enum QuicApiType
         ConnectionStart,
         ConnectionSetConfiguration,
         ConnectionSendResumptionTicket,
-        ConnectionCompleteResumptionTicketValidation,
```
ami-GS, line 35: Does the order matter?

ami-GS (approved):

src/plugins/trace/dll/DataModel/QuicEvents.cs
nibanks: The order in here must match `QUIC_TRACE_API_TYPE`

PR #4463, opened 2024-08-15
Fix Time64Distro Packaging on GH Actions

liveans:
## Description

Fixes the packaging of Time64Distros

## Testing

CI

## Documentation

No

### DIFF
--- a/.azure/OneBranch.PullRequest.yml
+++ b/.azure/OneBranch.PullRequest.yml
@@ -201,6 +201,8 @@ extends:
       - build_linux
       jobs:
       - template: .azure/obtemplates/build-linux-packages.yml@self
+        parameters:
+          kernel: "kernel5_4"
       - template: .azure/obtemplates/build-linux-packages.yml@self
         parameters:
           kernel: "kernel5_15"
--- a/.github/workflows/package-linux.yml
+++ b/.github/workflows/package-linux.yml
@@ -44,6 +44,7 @@ jobs:
       arch: ${{ matrix.vec.arch }}
       tls: ${{ matrix.vec.tls }}
       xdp: ${{ matrix.vec.xdp }}
+      time64: ${{ matrix.vec.time64 }}
 
   test-packages:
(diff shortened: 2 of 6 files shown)


.github/workflows/package-linux.yml
```diff
@@ -33,9 +33,9 @@ jobs:
           { config: "Release", os: "ubuntu-22.04", arch: "arm", tls: "openssl3" },
           { config: "Release", os: "ubuntu-22.04", arch: "arm64", tls: "openssl3" },
           { config: "Release", os: "ubuntu-22.04", arch: "x64", tls: "openssl3" },
-          { config: "Release", os: "ubuntu-24.04", arch: "arm", tls: "openssl3" },
```
nibanks, line 36: Do we need to update the .azure pipelines too?

liveans: Yes we should, I'll grab packages from GH actions first, verify the fix on docker image and will do changes on azure pipelines if it works.

scripts/package-distribution.ps1
```diff
@@ -6,6 +6,16 @@
 
 #>
 
+param (
+    [Parameter(Mandatory = $false)]
+    [switch]$Time64Distro
+)
+
+$Time64 = $false
```
ami-GS, line 14: I think $Time64Distro is already boolean. why do you need to add one more variable?

ami-GS:
made PR against this forked repo
https://github.com/liveans/msquic/pull/1/files

.azure/OneBranch.PullRequest.yml
```diff
@@ -201,6 +201,12 @@ extends:
       - build_linux
       jobs:
       - template: .azure/obtemplates/build-linux-packages.yml@self
```
nibanks, line 203: Let's be explicit about the `kernel` parameter being passed in here, so the reader doesn't need to go look at the template to figure out the default. Same for Official yml.

PR #4465, opened 2024-08-15
Implement LowLat mode in TCP

ProjectsByJackHe:
## Description

As of today, MsQuic has several execution modes, namely Max Throughput, and Low Latency.

We use the Secnetperf tool to measure this performance impact between the different modes.
However, with our Secnetperf TCP implementation, there is no such distinction between modes.

Meaning, we are artificially giving QUIC an advantage in performance when we run it in Lowlat mode to optimize thread / context switching, but we are not doing the same for TCP.

This PR implements the actual TCP lowlat execution context part.

## Testing

## Documentation

### DIFF
--- a/src/perf/lib/Tcp.cpp
+++ b/src/perf/lib/Tcp.cpp
@@ -217,7 +217,7 @@ bool TcpWorker::Initialize(TcpEngine* _Engine, uint16_t AssignedCPU)
     }
     #endif
 
-    CXPLAT_THREAD_CONFIG Config = { 0, 0, "TcpPerfWorker", WorkerThread, this };
+    CXPLAT_THREAD_CONFIG Config = { 0, AssignedCPU, "TcpPerfWorker", WorkerThread, this };
     if (QUIC_FAILED(
         CxPlatThreadCreate(
             &Config,
@@ -234,6 +234,10 @@ void TcpWorker::Shutdown()
     if (Initialized && !IsExternal) {
         CxPlatEventSet(WakeEvent);
         CxPlatThreadWait(&Thread);
+    }
+    if (IsExternal) {
+       // TODO: Do we need to wake up external thread?
+       // TODO: Do we need to wait for external thread to finish?
     }
 }
 
@@ -289,6 +293,15 @@ CXPLAT_THREAD_CALLBACK(TcpWorker::WorkerThread, Context)
     CXPLAT_THREAD_RETURN(0);
 }
 
+void TcpWorker::WakeWorkerThread() {
+    ExecutionContext.Ready = TRUE;
+    if (IsExternal) {
+        CxPlatWakeExecutionContext(&ExecutionContext);
+    } else {
+        CxPlatEventSet(WakeEvent);
+    }
+}
+
 bool TcpWorker::QueueConnection(TcpConnection* Connection)
 {
     bool Result = true;
@@ -302,7 +315,7 @@ bool TcpWorker::QueueConnection(TcpConnection* Connection)
             Connection->QueuedOnWorker = true;
             *ConnectionsTail = Connection;
             ConnectionsTail = &Connection->Next;
-            CxPlatEventSet(WakeEvent);
+            WakeWorkerThread();
         } else {
             Result = false;
         }
--- a/src/perf/lib/Tcp.cpp
+++ b/src/perf/lib/Tcp.cpp
@@ -199,7 +199,7 @@ TcpWorker::~TcpWorker()
     CxPlatEventUninitialize(WakeEvent);
 }
 
-bool TcpWorker::Initialize(TcpEngine* _Engine, uint16_t AssignedCPU)
+bool TcpWorker::Initialize(TcpEngine* _Engine, uint16_t)
 {
     Engine = _Engine;
     CXPLAT_THREAD_CONFIG Config = { 0, 0, "TcpPerfWorker", WorkerThread, this };
@@ -208,10 +208,7 @@ bool TcpWorker::Initialize(TcpEngine* _Engine, uint16_t AssignedCPU)
     this->ExecutionContext.Context = this;
     this->ExecutionContext.Ready = TRUE;
 
-    if (Engine->TcpExecutionProfile == TCP_EXECUTION_PROFILE_LOW_LATENCY) {
-        // The CxPlat worker thread should also handle any 'Perf' protocol work, instead of creating a new thread for it.
-        CxPlatAddExecutionContext(&this->ExecutionContext, AssignedCPU);
-    } else if (QUIC_FAILED(
+    if (QUIC_FAILED(
         CxPlatThreadCreate(
             &Config,
             &Thread))) {
--- a/src/perf/lib/Tcp.cpp
+++ b/src/perf/lib/Tcp.cpp
@@ -206,10 +206,12 @@ bool TcpWorker::Initialize(TcpEngine* _Engine, uint16_t AssignedCPU)
     ExecutionContext.Context = this;
     ExecutionContext.Ready = TRUE;
     ExecutionContext.NextTimeUs = 0; // TODO: Figure out why adding this line makes lowlat tcp work?
+    PartitionIndex = AssignedCPU;
 
     if (Engine->TcpExecutionProfile == TCP_EXECUTION_PROFILE_LOW_LATENCY) {
         CxPlatAddExecutionContext(&ExecutionContext, AssignedCPU);
         Initialized = true;
+        IsExternal = true;
         return true;
     }
 
--- a/src/perf/lib/Tcp.cpp
+++ b/src/perf/lib/Tcp.cpp
@@ -252,7 +252,7 @@ TcpWorker::DoWork(
 {
     TcpWorker* This = (TcpWorker*)Context;
     TcpConnection* Connection;
-    if (This->Engine->Shutdown) {
+    if (This == NULL || This->Engine->Shutdown) {
         return FALSE;
     }
     CxPlatDispatchLockAcquire(&This->Lock);
--- a/src/perf/lib/Tcp.cpp
+++ b/src/perf/lib/Tcp.cpp
@@ -114,7 +114,7 @@ TcpEngine::TcpEngine(
 {
     CxPlatListInitializeHead(&Connections);
     for (uint16_t i = 0; i < ProcCount; ++i) {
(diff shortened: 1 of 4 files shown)


nibanks (dismissed):

nibanks:
Looks like netperf broke from this?

src/perf/lib/Tcp.cpp
```diff
@@ -222,31 +222,54 @@ void TcpWorker::Shutdown()
...
-            if (This->ConnectionsTail == &Connection->Next) {
-                This->ConnectionsTail = &This->Connections;
-            }
-            Connection->QueuedOnWorker = false;
-            Connection->Next = NULL;
+    while (true) {
+        BOOLEAN hasMoreWork = WorkerLoop(This, nullptr);
+        if (!hasMoreWork) {
+            break;
         }
```
nibanks, line 268: I still prefer:
```suggestion
    while (!WorkerLoop(This, nullptr)) { // Loop until worker loop is ready to shutdown
```

src/perf/lib/Tcp.h
```diff
@@ -21,6 +21,7 @@ class TcpServer;
 class TcpConnection;
 struct TcpFrame;
 
+
```
nibanks, line 24: nit: revert?

ProjectsByJackHe: Done

ProjectsByJackHe:
> Looks like netperf broke from this?

Yeah, fixed a small bug. I was looking at `This->Shuttingdown` when it should have been `This->Shutdown`

src/perf/lib/Tcp.cpp
ProjectsByJackHe: Refactored it to remove the break clause.

src/perf/lib/Tcp.cpp
```diff
@@ -199,10 +199,15 @@ TcpWorker::~TcpWorker()
...
 
-bool TcpWorker::Initialize(TcpEngine* _Engine)
+bool TcpWorker::Initialize(TcpEngine* _Engine, uint16_t)
 {
     Engine = _Engine;
     CXPLAT_THREAD_CONFIG Config = { 0, 0, "TcpPerfWorker", WorkerThread, this };
+
+    this->ExecutionContext.Callback = DoWork;
+    this->ExecutionContext.Context = this;
+    this->ExecutionContext.Ready = TRUE;
```
nibanks, line 209: Not need to use `this->` unless there is a name conflict.
```suggestion
    ExecutionContext.Callback = DoWork;
    ExecutionContext.Context = this;
    ExecutionContext.Ready = TRUE;
```

src/perf/lib/Tcp.cpp
```diff
@@ -222,29 +227,48 @@ void TcpWorker::Shutdown()
...
+        Connection->QueuedOnWorker = false;
+        Connection->Next = NULL;
+    }
+    CxPlatDispatchLockRelease(&This->Lock);
+    if (Connection) {
+        Connection->Process();
+        Connection->Release();
+        This->ExecutionContext.Ready = TRUE; // We just did work, let's keep this thread hot.
+    } else {
+        This->ExecutionContext.Ready = FALSE;
```
nibanks, line 262: Don't have this `else` clause. It's unnecessary and all extra instructions are pure overhead. Since this is a perf sensitive code path, we want to remove any fluff.

ProjectsByJackHe: I removed the `if` check in the loop. Does having an `else` clause add more instructions in assembly than an extra `if` check?

nibanks: This function must assume `Ready == FALSE` on entry. And the WorkerThread should enforce that. That's how the CxPlat execution context works.

ProjectsByJackHe: done

src/perf/lib/Tcp.cpp
```diff
@@ -199,9 +199,19 @@ TcpWorker::~TcpWorker()
...
-bool TcpWorker::Initialize(TcpEngine* _Engine)
+bool TcpWorker::Initialize(TcpEngine* _Engine, uint16_t AssignedCPU)
 {
     Engine = _Engine;
+    ExecutionContext.Callback = DoWork;
+    ExecutionContext.Context = this;
+    ExecutionContext.Ready = TRUE;
+
+    if (Engine->TcpExecutionProfile == TCP_EXECUTION_PROFILE_LOW_LATENCY && AssignedCPU != UINT16_MAX) {
+        // CxPlatAddExecutionContext(&ExecutionContext, AssignedCPU); TODO - This code does not work.
```
ProjectsByJackHe, line 210: After setting the exec context, the CxPlat workers will execute the doWork() function once then seem to start to hang.

ProjectsByJackHe: This works now, but it's unclear why modifying `noWorkCount` and `nextTimeUs = 0` makes this all work.

src/perf/lib/Tcp.cpp
```diff
@@ -199,9 +199,22 @@ TcpWorker::~TcpWorker()
...
 }
 
-bool TcpWorker::Initialize(TcpEngine* _Engine)
+bool TcpWorker::Initialize(TcpEngine* _Engine, uint16_t AssignedCPU)
 {
     Engine = _Engine;
+    ExecutionContext.Callback = DoWork;
+    ExecutionContext.Context = this;
+    ExecutionContext.Ready = TRUE;
+    ExecutionContext.NextTimeUs = 0; // TODO: Figure out why adding this line makes lowlat tcp work?
```
nibanks, line 208: You need to default this to zero to indicate you don't have a timer you're waiting on.

nibanks (commented):
Looking really good!

src/perf/lib/Tcp.h
```diff
@@ -112,7 +112,10 @@ class TcpEngine {
 class TcpWorker {
     friend class TcpEngine;
     friend class TcpConnection;
+    CXPLAT_EXECUTION_CONTEXT ExecutionContext;
     bool Initialized{false};
+    bool IsExternal{false};
+    uint16_t PartitionIndex{0};
```
nibanks, line 118: Why is saving this necessary?

ProjectsByJackHe: WinKernel builds fail because since `CxPlatAddExecutionContext` is not defined in that environment, the function parameter `uint6_t AssignedCPU` is no longer referenced. This leads to a warning treated as an error. 

This code gives me an excuse to use `AssignedCPU` somewhere, which we may leverage later down the line.

nibanks: Copy how worker.c does this. It disabled calls to CxPlat for this on kernel mode. That's what you need to do. I don't think you'll need the variable saved at all.

ProjectsByJackHe: Ok, after some debugging with `valgrind`, looks like the reason we are running into SIGABRT (exit code 134) on Linux is because in the destructor of `Tcp.cpp`, we are attempting to free / wait on a non-existent thread. Currently, we use the `TcpWorker::Initialized` variable to determine whether or not to free/wait the thread. We should need the `TcpWorker::IsExternal` flag to avoid freeing a NULL pointer.

src/perf/lib/Tcp.cpp
```diff
@@ -192,16 +192,29 @@ TcpWorker::TcpWorker()
...
+    PartitionIndex = AssignedCPU;
+
+    if (Engine->TcpExecutionProfile == TCP_EXECUTION_PROFILE_LOW_LATENCY) {
+        CxPlatAddExecutionContext(&ExecutionContext, AssignedCPU);
+        Initialized = true;
+        IsExternal = true;
+        return true;
+    }
+
     CXPLAT_THREAD_CONFIG Config = { 0, 0, "TcpPerfWorker", WorkerThread, this };
```
ProjectsByJackHe, line 218: Also BTW, in the `maxtput` case, why do set the ideal processor to always be processor 0 here?

nibanks: Probably a bug 😄

src/perf/lib/Tcp.cpp
nibanks, line 1: You also need to fix the 'wake' code path, similar to [QuicWorkerThreadWake](https://github.com/microsoft/msquic/blob/main/src/core/worker.c#L39).

src/perf/lib/Tcp.cpp
ProjectsByJackHe: Set it to AssignedCPU

src/perf/lib/Tcp.cpp
ProjectsByJackHe: Added a `TcpWorker::WakeWorkerThread()` helper. What's interesting is we're currently not doing anything in the `TcpWorker::Shutdown()` function in the external case, whereas if we created a TcpWorker thread, we would wake and wait for that thread in the shutdown path. I wonder if this is not necessary in the external case?

nibanks: Maybe. If it works, let's merge it and fix any issues that arrise.

nibanks (dismissed):

src/perf/lib/Tcp.cpp
ProjectsByJackHe: Sounds good. What I am currently looking into is why Ubuntu 20.04 (and only Ubuntu 20.04) is still running into exit 134 every 2 out of 3 runs. Inspecting the crash dumps confirms that my initial fix to not delete/wait on a non-existant TCPworker thread is not the cause (anymore). The problem now seems to be in the TcpConnection, we try to acquire a lock that seems to be already acquired by the same thread... (at least that's my understanding from reading the backtrace)

nibanks: In the destructor, you need to add something like what we have in worker.c:
```c
    Worker->Enabled = FALSE;
    if (Worker->ExecutionContext.Context) {
        QuicWorkerThreadWake(Worker);
        CxPlatEventWaitForever(Worker->Done);
    }
    CxPlatEventUninitialize(Worker->Done);
```

Then in the shutdown path of the DoWork, call
```c
CxPlatEventSet(Worker->Done);
```

nibanks (dismissed):

ProjectsByJackHe:
Looks like a bunch of our BVT tests are failing now because of a heap use-after free. Investigating.

nibanks (approved):

PR #4483, opened 2024-08-22
Socket Errors are Asynchronous

nibanks:
## Description

The send path error processing logic assumed errors could/should be processed synchronously, and since we never see them for UDP, it wasn't a big deal. For TCP, though, we see them more often, and it results in a use-after-free because we incorrectly release the last ref inline, and later the completion fires after the object had been freed.

P.S. I wish there was a `FILE_SKIP_COMPLETION_PORT_ON_ERROR` so that only PENDING results completed async. Errors should be processed inline too.

## Testing

Locally via secnetperf

## Documentation

N/A

### DIFF
--- a/src/platform/datapath_winuser.c
+++ b/src/platform/datapath_winuser.c
@@ -4086,7 +4086,6 @@ CxPlatSocketSendInline(
         return QUIC_STATUS_PENDING;
     }
 
-    QUIC_STATUS Status;
     int Result;
     DWORD BytesSent;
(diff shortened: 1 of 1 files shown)


mtfriesen:
> P.S. I wish there was a FILE_SKIP_COMPLETION_PORT_ON_ERROR so that only PENDING results completed async. Errors should be processed inline too.

Yeah, that is unfortunate.

mtfriesen (approved):

PR #4484, opened 2024-08-22
Align XDP Queues with RSS Better

nibanks:
## Description

Updates the XDP datapath layer (on Windows) to try to query the RSS processor before fully creating the queues, so that the real queue may be correctly aligned with RSS.

## Testing

CI/CD

## Documentation

N/A


src/platform/datapath_raw_xdp_win.c
```diff
@@ -107,246 +107,137 @@ CxPlatXdpExecute(
...
-            LibraryErrorStatus,
-            "[ lib] ERROR, %u, %s.",
-            ret,
-            "ConvertInterfaceIndexToLuid");
-        return HRESULT_FROM_WIN32(ret);
+    ETHERNET_HEADER* Ethernet = (ETHERNET_HEADER*)Packet->FrameBuffer;
+    IPV4_HEADER* IPv4 = (IPV4_HEADER*)(Ethernet + 1);
+    UDP_HEADER* UDP = (UDP_HEADER*)(IPv4 + 1);
+
+    // Set Ethernet header
```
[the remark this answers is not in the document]
mtfriesen: Our API contract says yes. In practice with VF indirection tables differing from the vNIC above it, results may be incorrect. The packet contents will not make a difference.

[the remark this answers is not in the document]
mtfriesen: BTW, I am planning to change our RSS APIs in XDP to allow query/set of VF RSS settings via the upper vNIC binding.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -532,6 +431,23 @@ CxPlatDpRawInterfaceInitialize(
...
+        uint32_t QueueIndex = UINT32_MAX;
+        for (uint32_t j = 0; j < Interface->QueueCount; j++) {
+            if (Processors[j] == i) {
+                QueueIndex = j;
+                break;
+            }
+        }
+        if (QueueIndex == UINT32_MAX) {
+            Status = QUIC_STATUS_INVALID_STATE;
+            QuicTraceEvent(
```
nibanks, line 443: @mtfriesen I'm hitting this error in test (definitely over duonic, but also netperf is failing, but I don't have logs). I'm assuming that means we're trying to bind a queue to a processor that doesn't have an RSS queue? I'm trying to decide how to handle this. Thoughts?

src/platform/datapath_raw_xdp_win.c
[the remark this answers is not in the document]
mtfriesen: Nah - that's another can of worms.

src/platform/datapath_raw_xdp_win.c
mtfriesen: More queues than processors in the indirection table is completely legal. I'd just leave any leftover queues unaffinitized and hope for the best.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -778,10 +679,20 @@ CxPlatDpRawInterfaceInitialize(
...
+        BOOLEAN Found = FALSE;
+        for (uint16_t j = 0; j < Xdp->PartitionCount; j++) {
+            if (Xdp->Partitions[j].Processor == Interface->Queues[i].RssProcessor) {
+                XdpWorkerAddQueue(&Xdp->Partitions[j], &Interface->Queues[i]);
+                Found = TRUE;
+                break;
+            }
+        }
+        if (!Found) {
+            CXPLAT_FRE_ASSERT(FALSE); // TODO - What do we do if there is no partition for this processor?
```
nibanks, line 694: @mtfriesen thoughts here? We allow an app to specify the CPUs they want us to run on. What do we do if there is an RSS queue that doesn't have a CPU being run on?

src/platform/datapath_raw_xdp_win.c
mtfriesen: Hmm, ok, I'm not that familiar with this QUIC code. NICs just indicate packets on the processors in the indirection table - the rest are idle.

src/platform/datapath_raw_xdp_win.c
mtfriesen: I think the question you are asking here is the question I answered in the incorrect location above - a queue without an assigned processor is effectively not a queue.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -107,246 +108,137 @@ CxPlatXdpExecute(
...
-    IWbemClassObject *pclsObj = NULL;
-    ULONG uReturn = 0;
-    while (pEnum) {
-        HRESULT hr = pEnum->lpVtbl->Next(pEnum, WBEM_INFINITE, 1,
-            &pclsObj, &uReturn);
-
-        if (0 == uReturn) {
+        uint32_t Flags = XSK_BIND_FLAG_TX;
+        Status = Xdp->XdpApi->XskBind(TxXsk, InterfaceIndex, i, Flags);
+        if (QUIC_FAILED(Status)) { // No more queues. Break out.
```
mtfriesen, line 192: there could be other errors. probably worth logging explicitly.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -107,246 +108,137 @@ CxPlatXdpExecute(
...
+        XSK_BUFFER_DESCRIPTOR* Buffer = XskRingGetElement(&TxRing, TxIndex++);
+        Buffer->Address.BaseAddress = 0;
+        Buffer->Address.Offset = FIELD_OFFSET(XDP_TX_PACKET, FrameBuffer);
+        Buffer->Length = TxPacket.Buffer.Length;
+        XskRingProducerSubmit(&TxRing, 1);
 
-            free(rssTable);
-        }
+        XSK_NOTIFY_RESULT_FLAGS OutFlags;
+        Status = Xdp->XdpApi->XskNotifySocket(TxXsk, XSK_NOTIFY_FLAG_POKE_TX|XSK_NOTIFY_FLAG_WAIT_TX, 1000, &OutFlags);
```
mtfriesen, line 220: 1000 magic number - should be `#define` ?

nibanks: Will do.

src/platform/datapath_raw_xdp_win.c
nibanks: You want me to just log this one?

src/platform/datapath_raw_xdp_win.c
```diff
@@ -107,246 +108,137 @@ CxPlatXdpExecute(
...
-    //AF_XDP_LOG(INFO, "counted %u active queues on %s\n", cnt, if_name);
-    *Count = cnt;
+        PROCESSOR_NUMBER ProcNumber;
+        uint32_t ProcNumberSize = sizeof(PROCESSOR_NUMBER);
+        Status = Xdp->XdpApi->XskGetSockopt(TxXsk, XSK_SOCKOPT_TX_PROCESSOR_AFFINITY, &ProcNumber, &ProcNumberSize);
+        if (QUIC_FAILED(Status)) { CloseHandle(TxXsk); return Status; }
 
-Cleanup:
+        const CXPLAT_PROCESSOR_GROUP_INFO* Group = &CxPlatProcessorGroupInfo[ProcNumber.Group];
+        Queues[i] = Group->Offset + (ProcNumber.Number % Group->Count);
```
mtfriesen, line 236: why is the modulo necessary? i probably just need to go read more of the QUIC code, but i would have expected all ProcNumber.Numbers to be less than Group->Count.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -480,6 +372,9 @@ CxPlatDpRawInterfaceInitialize(
     Interface->OffloadStatus.Transmit.NetworkLayerXsum = Xdp->SkipXsum;
     Interface->Xdp = Xdp;
 
+    uint32_t Processors[256]; // TODO - Use max processor count
```
mtfriesen, line 375: unfinished TODO

src/platform/datapath_raw_xdp_win.c
```diff
@@ -521,6 +427,7 @@ CxPlatDpRawInterfaceInitialize(
     for (uint8_t i = 0; i < Interface->QueueCount; i++) {
         XDP_QUEUE* Queue = &Interface->Queues[i];
 
+        Queue->RssProcessor = (uint16_t)Processors[i]; // TODO - Should memory be aligned with this?
```
mtfriesen, line 430: as in, should all allocations for this queue be aligned to that processor? ideally yes, for sure.

nibanks: Exactly, this is a larger issue too. For all memory allocation that is meant to be used on a particular process, should we temporarily affinitize the thread, allocate the memory and then revert the affinitization?

src/platform/datapath_raw_xdp_win.c
mtfriesen: this and the "activate" are the only two that might fail in nontrivial and unpredictable ways. it's probably worth logging them all if you have time.

src/platform/datapath_raw_xdp_win.c
mtfriesen: in user mode, there are fancy APIs to allocate from a specific node, even if you're not currently running on that node. in kernel mode, it's more difficult to allocate pool from an arbitrary node, though i think newer OSes have better APIs. one sec.

that said, XDP itself treats the calling thread as the "ideal" thread when it performs its own allocations for queues, so temporarily setting the affinity would probably solve the problem the most universally and with the least fuss.

src/platform/datapath_raw_xdp_win.c
nibanks: I _think_ it shouldn't be necessary.

src/platform/datapath_raw_xdp_win.c
mtfriesen: yeah, PoolExtendedParameterNumaNode is supported for ExAllocatePool**3**. i'd still just set the thread and be done.

nibanks: So, if we're all on a single NUMA, how much would this actually matter?

src/platform/datapath_raw_xdp_win.c
mtfriesen: is it for hot add? where QUIC has a fixed set of CPUs per group that might see more CPUs added to the group later?

nibanks: No, we fix this list at process start.

src/platform/datapath_raw_xdp_win.c
nibanks: For VMs that are currently affixed to just one of the NUMA nodes, are we sure there's no per-proc memory perf involved? Also, can we be guaranteed that the VM actually knows the NUMA of the HW?

mtfriesen: AFAIK the VM can't do anything to optimize that scenario. It's all hidden and controlled by the host/hypervisor.

src/platform/datapath_raw_xdp_win.c
mtfriesen: right, so if the group is fixed at size 1 during process start, and the OS later adds proc 2 to the group, then that's why you have the modulo?

src/platform/datapath_raw_xdp_win.c
nibanks: Leaving it as a future TODO

mtfriesen (approved):

PR #4488, opened 2024-08-23
Do not exit gracefully in error cases

ProjectsByJackHe:
## Description

In the Secnet tool, 

when something is wrong or supposed to crash, exiting gracefully leads to quiet errors and a massive pain while debugging.

This PR makes it so it will actually crash and a core dump gets generated.

## Testing

Netperf CI

## Documentation

N/A

### DIFF
--- a/scripts/secnetperf.ps1
+++ b/scripts/secnetperf.ps1
@@ -257,7 +257,7 @@ $allTests = [System.Collections.Specialized.OrderedDictionary]::new()
 $allTests["tput-up"] = "-exec:maxtput -up:12s -ptput:1"
 $allTests["tput-down"] = "-exec:maxtput -down:12s -ptput:1"
 $allTests["hps-conns-100"] = "-exec:maxtput -rconn:1 -share:1 -conns:100 -run:12s -prate:1"
-$allTests["rps-up-512-down-4000"] = "-exec:lowlat -rstream:1 -up:512 -down:4000 -run:20s -plat:1"
+$allTests["rps-up-512-down-4000"] = "-exec:lowlat -rstream:1 -up:512 -down:4000 -run:20s -plat:1 -abortOnFailure"
 
 $hasFailures = $false
 $json["run_args"] = $allTests
--- a/scripts/secnetperf.ps1
+++ b/scripts/secnetperf.ps1
@@ -257,7 +257,7 @@ $allTests = [System.Collections.Specialized.OrderedDictionary]::new()
 $allTests["tput-up"] = "-exec:maxtput -up:12s -ptput:1"
 $allTests["tput-down"] = "-exec:maxtput -down:12s -ptput:1"
 $allTests["hps-conns-100"] = "-exec:maxtput -rconn:1 -share:1 -conns:100 -run:12s -prate:1"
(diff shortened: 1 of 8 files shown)


nibanks (approved):

nibanks:
> This PR makes it so it will actually crash and a core dump gets generated.

Why do you say this will make it crash? It's just returning an error, not crashing.

src/perf/bin/appmain.cpp
```diff
@@ -125,6 +126,10 @@ QuicUserMain(
     if (!SimpleOutput) {
         printf("App Main returning status %d\n", Status);
     }
+    if (!QUIC_SUCCEEDED(Status) && AbortOnFailure) {
+        printf("AbortOnFailure: non zero exit code: %d, Aborting to generate core dump. \n", Status);
+        abort();
```
nibanks, line 131: Use `CXPLAT_FRE_ASSERTMSG` instead?

ProjectsByJackHe:
> > This PR makes it so it will actually crash and a core dump gets generated.
> 
> Why do you say this will make it crash? It's just returning an error, not crashing.

Great point! I was debating on how to enable this feature. I've settled on a flag that you pass in if you want secnet to crash and generate a core dump upon a non-zero exit code.

In cases where we have a flaky test, we'd still want to run the other tests.

In cases where we have a consistent failing test, enabling the abortonfailure flag just on that test will generate the core dump we want.

nibanks:
But why do you think a dump will help? From my experience, logs are the answer here, because you need to know what happened in the past. The current state of the app doesn't help there.

ProjectsByJackHe:
> But why do you think a dump will help? From my experience, logs are the answer here, because you need to know what happened in the past. The current state of the app doesn't help there.

For my specific use case, I was attaching GDB to the dumps and inspecting the various variables further down the stack trace.
Though, this didn't yield any useful.

I still think that the program shouldn't terminate with an exit code of 0 if we hit "No Connections" error case though.

nibanks:
> I still think that the program shouldn't terminate with an exit code of 0 if we hit "No Connections" error case though.

I'm fine with that change.

scripts/secnetperf.ps1
```diff
@@ -257,7 +257,7 @@ $allTests = [System.Collections.Specialized.OrderedDictionary]::new()
 $allTests["tput-up"] = "-exec:maxtput -up:12s -ptput:1"
 $allTests["tput-down"] = "-exec:maxtput -down:12s -ptput:1"
 $allTests["hps-conns-100"] = "-exec:maxtput -rconn:1 -share:1 -conns:100 -run:12s -prate:1"
-$allTests["rps-up-512-down-4000"] = "-exec:lowlat -rstream:1 -up:512 -down:4000 -run:20s -plat:1"
+$allTests["rps-up-512-down-4000"] = "-exec:lowlat -rstream:1 -up:512 -down:4000 -run:20s -plat:1 -abortOnFailure"
```
nibanks, line 260: Do we really need to abort on failure by default in netperf?

ProjectsByJackHe: I have this as a temporary change to collect the dumps and fiddle with GDB. Will revert this.

nibanks (approved):

PR #4510, opened 2024-08-28
Use Rundown Protection for TCP Send Completions

nibanks:
## Description

Finally fixes the real reason we were seeing secnetperf crash.

## Testing

Local. CI/CD

## Documentation

N/A


mtfriesen (approved):

ProjectsByJackHe (approved):

ProjectsByJackHe:
Thank you! This fixed the UDP kill packet problem. Looks like there is something weird still about Ubuntu TCP Lowlat mode, but that's unrelated. Approved!

PR #4511, opened 2024-08-29
Add flag to not set ideal proc on worker threads

ThadHouse:
## Description

For resource reasons, I'm having to create a datapath with just a single threaded worker pool. I wouldn't want that thread affinitized to any specific thread. Add a flag to skip setting ideal proc on the worker thread.
 
## Testing

Covered by existing tests

## Documentation

Execution configs is currently an undocumented feature.


src/inc/msquic.h
```diff
@@ -271,6 +271,7 @@ typedef enum QUIC_EXECUTION_CONFIG_FLAGS {
     QUIC_EXECUTION_CONFIG_FLAG_RIO              = 0x0002,
     QUIC_EXECUTION_CONFIG_FLAG_XDP              = 0x0004,
 #endif
+    QUIC_EXECUTION_CONFIG_FLAG_NO_IDEAL_PROC    = 0x0008,
```
nibanks, line 274: For now, let's keep all these in PREVIEW_FEATURES

src/platform/platform_worker.c
```diff
@@ -149,8 +149,12 @@ CxPlatWorkersLazyStart(
         goto Error;
     }
 
+    BOOLEAN SetIdealProc = !(Config->Flags & QUIC_EXECUTION_CONFIG_FLAG_NO_IDEAL_PROC);
```
nibanks, line 152: `Config` may be `NULL`.
```suggestion
    BOOLEAN SetIdealProc = !Config || !(Config->Flags & QUIC_EXECUTION_CONFIG_FLAG_NO_IDEAL_PROC);
```

src/inc/msquic.h
nibanks, line 274:
```suggestion
    QUIC_EXECUTION_CONFIG_FLAG_NO_IDEAL_PROC    = 0x0008,
#endif
```

nibanks:
.NET needs to be updated too please.

nibanks (approved):

PR #4517, opened 2024-08-29
Prunable Pool Allocator

nibanks:
## Description

Updates the datapath RX buffers (on Windows) to periodically prune items to keep from growing and holding on to too much memory.

## Testing

CI/CD

## Documentation

N/A


src/platform/datapath_winuser.c
```diff
@@ -2465,7 +2469,7 @@ CxPlatSocketAllocRxIoBlock(
     if (SocketProc->Parent->UseRio) {
         OwningPool = &DatapathProc->RioRecvPool;
     } else {
-        OwningPool = &DatapathProc->RecvDatagramPool;
+        OwningPool = (CXPLAT_POOL*)&DatapathProc->RecvDatagramPool;
```
mtfriesen, line 2472: It might be better to name the CXPLAT_POOL field of CXPLAT_POOL_EX in C, too, and then be able to use &Pool.Foo instead of a typecast and assumption about the field layouts.

src/platform/platform_worker.c
```diff
@@ -300,6 +307,65 @@ CxPlatWorkersUninit(
...
+CxPlatProcessDynamicPoolAllocators(
+    _In_ CXPLAT_WORKER* Worker
+    )
+{
+    QuicTraceLogVerbose(
+        PlatformWorkerProcessPools,
+        "[ lib][%p] Processing pools",
+        Worker);
+
+    CxPlatLockAcquire(&Worker->ECLock);
```
mtfriesen, line 359: Is this lock heavily contended? If the worker has to scan 1000+ CPUs and free 8 items per list per CPU, that could add some jitter.

mtfriesen: Ah, this is a per-worker lock, so you're only looking at 8 items per list. Seems pretty lightweight.

src/platform/platform_worker.c
```diff
@@ -300,6 +307,65 @@ CxPlatWorkersUninit(
     CxPlatLockUninitialize(&CxPlatWorkerLock);
 }
 
+#define DYNAMIC_POOL_PROCESSING_TIME 1000000 // 1 second
```
maolson-msft, line 310: "TIME" -> "PERIOD" ("TIME" sounds like amount of time spent processing, and is also confusing next to LastPoolProcessTime)

src/platform/platform_worker.c
```diff
@@ -504,6 +570,11 @@ CXPLAT_THREAD_CALLBACK(CxPlatWorkerThread, Context)
             CxPlatSchedulerYield();
             State.NoWorkCount = 0;
         }
+
+        if (State.TimeNow - State.LastPoolProcessTime > DYNAMIC_POOL_PROCESSING_TIME) {
```
maolson-msft, line 574: Minor: I see TimeNow is updated inside CxPlatRunExecutionContexts. I feel like up here in CxPlatWorkerThread, at the beginning of the while loop, would be a more natural place for that.

For example: I'm not sure if it's a bug, but I see there's a codepath where CxPlatRunExecutionContexts fails to update TimeNow. It's not great to have to think about ECs when thinking through the correctness of this new code that uses the TimeNow variable.

src/platform/unittest/DataPathTest.cpp
```diff
@@ -696,12 +696,6 @@ TEST_F(DataPathTest, InitializeInvalid)
         ASSERT_EQ(QUIC_STATUS_INVALID_PARAMETER, Datapath.GetInitStatus());
         ASSERT_EQ(nullptr, Datapath.Datapath);
     }
-    if (!UseDuoNic) {
```
nibanks, line 699: This test case is invalid if you have XDP installed locally (which I do).

mtfriesen: WOW

mtfriesen (approved):

Issue #4457, opened 2024-08-13
QUIC_PARAM_TLS_HANDSHAKE_INFO is failing on server size connection


wfurt:
### Describe the bug

I but to this when working on support for https://github.com/dotnet/runtime/issues/70184
The test is pretty simple: we establish Quic connection and we query the property on both sides and we expect them to be the same. On server, this fails with `QUIC_STATUS_INVALID_PARAMETER`. That alone is suboptimal IMHO as all the parameters _are_ valid. From quic debug it seems like the `Connection->Crypto.TLS` is NULL.

### Affected OS

- [ ] Windows
- [X] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

I only tested this on Ubuntu 22.04 but I doubt this is specific only to my version. 

### MsQuic version

main

### Steps taken to reproduce bug

1. establish Quic connection and trey to get TLS info on both sides.

### Expected behavior

The query should succeed and `QUIC_HANDSHAKE_INFO` should be filled with correct details. 

### Actual outcome

the query fails with `QUIC_STATUS_INVALID_PARAMETER`

### Additional details

https://github.com/dotnet/runtime/pull/106368 is my draft PR to expose `NegotiatedCipherSuite` in .NET.

nibanks:
cc @anrossi can you take a look?

anrossi:
@wfurt when are you querying this? In the `QUIC_CONNECTION_EVENT_CONNECTED` event callback? Or after?  When resumption/0-RTT is not enabled on server, it cleans up the TLS state as soon as the handshake is complete (immediately after the CONNECTED callback returns). This is a memory-saving optimization.

wfurt:
since this is now exposed on `QuicStream` the original code would simply pass that to msquic. @rzikm put up different implementation where we query it and cache it regardless of the user here: https://github.com/dotnet/runtime/pull/106391

If anything I wish there was some more meaningful error coming back.

anrossi:
Would an "invalid state" error be more helpful?

wfurt:
yes, perhaps as this is related to the inner state and not passed parameters. Also it did work on client but not on server in exactly same scenario.

anrossi:
Right, because the client keeps the TLS state around for the lifetime of the connection. This is because the server could send a resumption/0-RTT ticket at any time, and the client needs TLS still active to process that ticket.  In this case, the server and client are not perfectly symmetrical.

anrossi:
[...]

PR #4518, opened 2024-08-30
Make GetParam return INVALID_STATE error when TLS has been cleaned up

anrossi:
## Description

One the server-side, QUIC cleans up the TLS state once the connection handshake is complete. This means that the last time `QUIC_PARAM_TLS_HANDSHAKE_INFO` can be queried on a server is in the `QUIC_CONNECTION_EVENT_CONNECTED` event.
Once cleaned up, attempting to query this state returned `QUIC_STATUS_INVALID_PARAMETER`, as pointed out in #4457, this is confusing. This change updates it so that `QUIC_STATUS_INVALID_STATE` is returned once the TLS state has been cleaned up to more clearly indicate that internal state has changed.

## Testing

Tested by CI

## Documentation

This behavior isn't documented, so no new documentation is needed. However, maybe the limitation of this GetParam needs to be documented better for the difference between client and server behavior.

### DIFF
--- a/src/core/library.c
+++ b/src/core/library.c
@@ -1667,8 +1667,10 @@ QuicLibraryGetParam(
 
     case QUIC_PARAM_PREFIX_TLS:
     case QUIC_PARAM_PREFIX_TLS_SCHANNEL:
-        if (Connection == NULL || Connection->Crypto.TLS == NULL) {
+        if (Connection == NULL) {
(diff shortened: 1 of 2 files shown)


nibanks (dismissed):

anrossi:
The code coverage results inspired me to add some testing to cover this

nibanks (approved):

PR #4519, opened 2024-08-30
Fix test failures from #4518

anrossi:
## Description

#4518 Was completed with some tests still failing, so this PR attempts to fix the test failures introduced in that PR.

## Testing

Fix the ApiTests for TLS parameters to run on XDP and not run test cases unsupported by kernel mode.

## Documentation

N/A

### DIFF
--- a/src/test/lib/ApiTest.cpp
+++ b/src/test/lib/ApiTest.cpp
@@ -4533,6 +4533,13 @@ void QuicTestTlsParam()
         //
         {
             TestScopeLogger LogScope1("GetParam");
+            TEST_QUIC_STATUS(
+                QUIC_STATUS_INVALID_PARAMETER,
+                Connection.GetParam(
+                    QUIC_PARAM_TLS_HANDSHAKE_INFO,
+                    nullptr,
+                    nullptr));
+
             uint32_t Length = 0;
             TEST_QUIC_STATUS(
                 QUIC_STATUS_BUFFER_TOO_SMALL,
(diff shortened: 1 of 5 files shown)


nibanks (approved):

src/test/lib/ApiTest.cpp
```diff
@@ -4603,26 +4602,34 @@ void QuicTestTlsParam()
                     ServerConfiguration);
                 TEST_QUIC_SUCCEEDED(Listener.IsValid());
```
nibanks, line 4603: Linux test failures claiming this is `FALSE`.
```
[ RUN      ] ParameterValidation.ValidateTlsParam
/__w/msquic/msquic/src/test/lib/ApiTest.cpp:4603: Failure
Listener.IsValid() failed, 0x1
[  FAILED  ] ParameterValidation.ValidateTlsParam (4 ms)
```

src/test/lib/ApiTest.cpp
```diff
@@ -4590,6 +4588,7 @@ void QuicTestTlsParam()
                 ), QUIC_STATUS_SUCCESS);
             }
 
+#ifndef _KERNEL_MODE
```
nibanks, line 4591: What makes this not work in kernel mode, where it did before?

anrossi: Are you sure it worked in kernel before? There's no kernel coverage for this. In a previous iteration, the kernel tests would return `STATUS_NOT_SUPPORTED`.

nibanks (dismissed):

nibanks (approved):

PR #4527, opened 2024-09-03
Fix kqueue CSPLAT_SQE Type

ThadHouse:
## Description

`kevent.ident` is a `uintptr_t`, not an `int`. C didn't really care, but if C++ included quic_platform.h, an error would occur as narrowing from signed to unsigned in a designated initializer is an error in C++. Could be solved either by adding an explicit cast, or changing the type to be correct. Changing the type seemed like the better option.

## Testing

Existing tests will cover it.

## Documentation

N/A

### DIFF
--- a/src/platform/platform_posix.c
+++ b/src/platform/platform_posix.c
@@ -245,6 +245,8 @@ CxPlatInitialize(
         return Status;
     }
 
+    CxPlatWorkersInit();
+
     CxPlatTotalMemory = CGroupGetMemoryLimit();
 
     QuicTraceLogInfo(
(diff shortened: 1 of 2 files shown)


src/platform/platform_posix.c
```diff
@@ -245,8 +245,6 @@ CxPlatInitialize(
         return Status;
     }
 
-    CxPlatWorkersInit();
```
nibanks, line 248: These changes unintentional?

ThadHouse: Oh. Yup. I copied these from a tree on my other PR. Whoops.

nibanks (approved):

PR #4531, opened 2024-09-05
Improve Thread Creation Logic

nibanks:
## Description

This PR improves various parts of thread creation in MsQuic. Most notably it adds the ability for apps to configure usage of high priority threads (not enabled by default). It also fixes/improves setting the ideal processor on systems with more than 64 procs in user mode. Also, it improves the default flags used in various places. Finally, it adds logs for various failures in thread creation.

## Testing

Manual secnetperf over localhost.
CI/CD

## Documentation

N/A


ami-GS (approved):

src/platform/platform_winuser.c
```diff
@@ -176,10 +176,11 @@ CxPlatProcessorInfoInit(
             if (Proc >= CxPlatProcessorGroupInfo[Group].Offset &&
                 Proc < CxPlatProcessorGroupInfo[Group].Offset + Info->Group.GroupInfo[Group].ActiveProcessorCount) {
                 CxPlatProcessorInfo[Proc].Group = Group;
-                CxPlatProcessorInfo[Proc].Index = (Proc - CxPlatProcessorGroupInfo[Group].Offset);
+                CXPLAT_DBG_ASSERT(Proc - CxPlatProcessorGroupInfo[Group].Offset < UINT8_MAX);
```
ThadHouse, line 179: I know its very unlikely to have a group with 256 threads, but shouldn't this check be <= UINT8_MAX?

nibanks: Good catch. I will update. Though, it is a debug assert, so it's SUPER unlikely to hit....

ProjectsByJackHe:
Looks really good!

Just 1 question.

Where did we get the confidence that there won't be any breaking changes in the "Scavenger" or "Real time" execution profile mode? 

At least according to CodeCov bot those paths weren't exercised.

nibanks:
> Looks really good!
> 
> Just 1 question.
> 
> Where did we get the confidence that there won't be any breaking changes in the "Scavenger" or "Real time" execution profile modes? 
> 
> At least according to CodeCov bot those paths weren't exercised. 

AFAIK no one uses those execution modes, so we should be fine.

PR #4492, opened 2024-08-25
Treat CxPlatWorker as a full object

ThadHouse:
## Description

Currently, if you have an app that is creating a datapath outside of msquic, that datapath always shares the same worker context as msquic. There are some cases where this is not the desired behavior, and instead a completely separate worker context should be used, such as if you need to guarantee there is always a free worker, and you have limited the msquic execution context with QUIC_PARAM_GLOBAL_EXECUTION_CONFIG.

This PR changes CxPlatWorker into a full object, with a default instance used. All the datapath init functions take a Worker Manager, and if that manager is null, the default one is used in its place.

## Testing

Alll existing tests will implicitly hit this path, and cover any necessary testing.

## Documentation

No documentation changes, as using CxPlat directly generally isn't a supported scenario.

PR #4535, opened 2024-09-07
Add global internal param to get default datapath worker pool

ThadHouse:
## Description

Before #4492, it was possible for apps statically linking to MsQuic and CxPLat to get the default worker pool used by Quic. After that PR, the default worker pool is no longer accessible. Add a new global param getter to make that accessable.

## Testing

Added test to cover the new parameter.

## Documentation

Internal, so not documented.

### DIFF
--- a/src/inc/msquicp.h
+++ b/src/inc/msquicp.h
@@ -116,6 +116,7 @@ typedef struct QUIC_PRIVATE_TRANSPORT_PARAMETER {
 #endif
 #define QUIC_PARAM_GLOBAL_IN_USE                        0x81000004  // BOOLEAN
 #define QUIC_PARAM_GLOBAL_DATAPATH_FEATURES             0x81000005  // uint32_t
+#define QUIC_PARAM_GLOBAL_GET_DATAPATH_WORKER_POOL      0x81000006  // CXPLAT_WORKER_POOL**
 
 //
 // The different private parameters for Configuration.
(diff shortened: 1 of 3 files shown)


src/inc/msquicp.h
```diff
@@ -116,6 +116,7 @@ typedef struct QUIC_PRIVATE_TRANSPORT_PARAMETER {
 #endif
 #define QUIC_PARAM_GLOBAL_IN_USE                        0x81000004  // BOOLEAN
 #define QUIC_PARAM_GLOBAL_DATAPATH_FEATURES             0x81000005  // uint32_t
+#define QUIC_PARAM_GLOBAL_GET_DATAPATH_WORKER_POOL      0x81000006  // CXPLAT_WORKER_POOL**
```
nibanks, line 119: Don't put `GET` in the name. We just won't allow SET, which is what we do for several parameters.

nibanks: And let's call it `_PLATFORM_WORKER_POOL`.

nibanks: Why return a `**` instead of `*`? Is the problem that it's not necessarily initialized yet? Perhaps trigger lazy instantiation if called?

ThadHouse: Yeah, it's not a **, its just a *. Doesn't actually change how the code works (A ** has sizeof() equal to a *).

I'll fix the other things.

src/test/lib/ApiTest.cpp
```diff
@@ -2681,6 +2681,50 @@ void QuicTestGlobalParam()
         }
     }
 
+    //
+    // QUIC_PARAM_GLOBAL_PLATFORM_WORKER_POOL
```
nibanks, line 2685: We need to disable this test for release build because it's not guaranteed to not break with future changes. The reason being is that future changes to expose application controlled execution context likely will result in more changes in this area.

submodules/googletest
nibanks, line 1: Accidental? Please revert.

nibanks (dismissed):

nibanks (approved):

PR #4536, opened 2024-09-08
Add a method to reject a TCP connection

ThadHouse:
## Description

Currently, it is impossible to reject a TCP connection. Calling CxPlatSocketDelete() from the Accept callback results in a use after free, and there isn't any way to detect that, other then to return a value from Accept()

## Testing

New test added to cover this case.

## Documentation

No, there is no documentation for TCP.

### DIFF
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -901,7 +901,7 @@ CxPlatSocketContextInitialize(
         // Only set SO_REUSEPORT on a server socket, otherwise the client could be
         // assigned a server port (unless it's forcing sharing).
         //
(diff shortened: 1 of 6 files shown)


nibanks (approved):

src/platform/datapath_epoll.c
```diff
@@ -901,7 +901,7 @@ CxPlatSocketContextInitialize(
         // Only set SO_REUSEPORT on a server socket, otherwise the client could be
         // assigned a server port (unless it's forcing sharing).
         //
-        if ((Config->Flags & CXPLAT_SOCKET_FLAG_SHARE || Config->RemoteAddress == NULL) && 
+        if ((Config->Flags & CXPLAT_SOCKET_FLAG_SHARE || Config->RemoteAddress == NULL) &&
```
nibanks, line 904: Unrelated change?

PR #4542, opened 2024-09-09
Don't set EPOLLOUT when creating TCP sockets

ThadHouse:
## Description

Epollout shouldn't be set until a pending send is available.

## Testing

Not easy to test because there isn't really a failure case, but the UDP side explicitly adds epollout only when needed. 

## Documentation

N/A


nibanks:
Lots of BVT failures

ThadHouse:
I guess you actually do need epollout set for connect() to work.

PR #4544, opened 2024-09-10
Set SO_REUSEADDR on epoll tcp listener sockets

ThadHouse:
## Description

Unix is a bit more strict about TIME_WAIT state, and actually puts any sockets that have had a valid accept() called on them into the TIME_WAIT state.

This makes writing a listener app difficult, as if that ever crashes the bind() will fail for the next few minutes.

Pretty much all other TCP libraries set SO_REUSEADDR (Including libuv, which is what our app has used before). Libuv sets it on all TCP sockets, but its generally less required on client sockets, as they rarely actually specify a local port.

## Testing

Not really possible to test with our current setup, as a crash is the only way to get a listener socket into TIME_WAIT. A properly shutdown socket doesn't stay in TIME_WAIT.

## Documentation

N/A

### DIFF
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -926,6 +926,29 @@ CxPlatSocketContextInitialize(
                 goto Exit;
             }
         }
+    } else if (SocketType == CXPLAT_SOCKET_TCP_LISTENER) {
+        //
+        // Set SO_REUSEADDR on listener sockets to avoid
+        // TIME_WAIT state on shutdown.
+        //
+        Option = TRUE;
+        Result =
+            setsockopt(
+                SocketContext->SocketFd,
+                SOL_SOCKET,
+                SO_REUSEADDR,
+                (const void*)&Option,
+                sizeof(Option));
+        if (Result == SOCKET_ERROR) {
+            Status = errno;
+            QuicTraceEvent(
+                DatapathErrorStatus,
+                "[data][%p] ERROR, %u, %s.",
+                Binding,
+                Status,
+                "setsockopt(SO_REUSEPORT) failed");
+            goto Exit;
+        }
     }
 
     CxPlatCopyMemory(&MappedAddress, &Binding->LocalAddress, sizeof(MappedAddress));


nibanks (approved):

src/platform/datapath_epoll.c
```diff
@@ -926,6 +926,29 @@ CxPlatSocketContextInitialize(
                 goto Exit;
             }
         }
+    } else if (SocketType == CXPLAT_SOCKET_TCP_LISTENER) {
+        //
+        // Set SO_REUSEADDR on listener sockets to avoid
+        // TIME_WAIT state on shutdown.
```
csujedihy, line 932: I am a bit confused. The listener socket itself can go from listen state to timewait state? I thought a socket had to go through finwait1 and finwait2 to enter timewait state (which implicitly means only sockets who *initiate* graceful shutdown).

ThadHouse: Yes, on linux a listener socket itself can go into timewait. Its a bit weird. But it was extremely easy to reproduce.

csujedihy: Mind sharing how it could be repro'd?

Just create a listener socket and have a connection accepted and kill both the accepted connection and the listener socket itself? Then bind will fail?

ThadHouse: Yup. Have the listener socket accept a connection, and then kill the listener side. The listener side then will not be able to restart due to a bind failure.

csujedihy: I am really surprised linux has this weird behavior as it violates RFC793.

image

ThadHouse: Wouldn't the listener socket follow the left side of that tree down to TIME_WAIT?

csujedihy: The socket that's accepted from the listener socket would follow the left side but the listener socket itself, in my understanding, would not. When you call accept() on a listener socket, you get a new socket and the listener itself should stay in listening state.

ThadHouse: It was a bit confusing to me too. libuv actually sets SO_REUSEADDR for both Client initiated sockets and listener sockets unconditionally, but at least for my use case I didn't need it set on the client side, and thats a little more invasive too.

PR #4551, opened 2024-09-20
[WIP] Linux Connection ID based RSS

ami-GS:
## Description

Conventional NIC RSS does not work well with QUIC as it can continue its connection with the Connection ID after changing IP/port.
This PR makes XDP to route packets to appropriate core using Connection ID.

- XSKMAP (AF_XDP) based redirection doesn't work at least on my environment (netvsc). Working internally about inter-queue redirection feature.
- CPUMAP based redirection works. This is hybrid implementation with datapath_epoll.c

## Testing

There is no functional test yet, but secnetperf based test (sometimes) works.
```sh
# server side
$ sudo MSQUIC_XDP_RSS_TYPE=1 MSQUIC_XDP_NORMAL_DATAPATH=1 ./artifacts/bin/linux/x64_Debug_openssl3/secnetperf -exec:maxtput -io:xdp -pollidle:1000

# client side
# multi-connection has bug. (e.g. -conns:10)
$ sudo MSQUIC_XDP_DROP_PATH_CHALLENGE=0  MSQUIC_FAKE_NAT_REBINDING=1 ./artifacts/bin/linux/x64_Debug_openssl3/secnetperf -target:<SERVER_IP> -exec:maxtput -down:10s -ptput:1 -tcp:0 -trimout -watchdog:25000 -pconn:1 -pstream:1 -io:xdp -conns:1
```

## Documentation

Not yet

Server side option
MSQUIC_XDP_RSS_TYPE: 0x01 -> using Connection ID which is available after finishing handshake
MSQUIC_XDP_RSS_TYPE: 0x08 -> using Hash function to Connection ID. (not implemented yet)
MSQUIC_XDP_NORMAL_DATAPATH:1 -> this uses normal socket with xdp (not AF_XDP socket).

Client side option
MSQUIC_FAKE_NAT_REBINDING:0/1 -> client changes (changes back) port to 55555 as if NAT rebinding happens
MSQUIC_XDP_DROP_PATH_CHALLENGE:0/1 ->  whether to ignore PATH_CHALLENGE frame after NAT rebinding

### DIFF
--- a/src/platform/datapath_raw_xdp_linux.c
+++ b/src/platform/datapath_raw_xdp_linux.c
@@ -955,6 +955,15 @@ CxPlatDpRawPlumbRulesOnSocket(
                         XdpSetPortFails,
                         "[ xdp] Failed to set port %d on %s", port, Interface->IfName);
                 }
+                char* FakeNatRebinding = getenv("MSQUIC_FAKE_NAT_REBINDING");
+                if (FakeNatRebinding && FakeNatRebinding[0] == '1') {
+                    port = htons(55555);
+                    if (bpf_map_update_elem(bpf_map__fd(port_map), &port, &exist, BPF_ANY)) {
+                        QuicTraceLogVerbose(
+                            XdpSetPortFails,
+                            "[ xdp] Failed to set port %d on %s", port, Interface->IfName);
+                    }
+                }
             } else {
                 if (bpf_map_delete_elem(bpf_map__fd(port_map), &port)) {
                     QuicTraceLogVerbose(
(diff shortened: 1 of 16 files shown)


src/platform/datapath_raw_xdp_linux.c
```diff
@@ -1016,13 +1127,19 @@ CxPlatDpRawRxFree(
...
         XskInfo = Packet->Queue->XskInfo;
 
         CxPlatLockAcquire(&XskInfo->UmemLock);
         while (PacketChain) {
             Packet =
                 CXPLAT_CONTAINING_RECORD(PacketChain, XDP_RX_PACKET, RecvData);
             PacketChain = PacketChain->Next;
+            CXPLAT_DBG_ASSERT(Packet != NULL);
+            CXPLAT_DBG_ASSERT(Packet->Queue != NULL);
+            CXPLAT_DBG_ASSERT(Packet->Queue->XskInfo != NULL);
```
ami-GS, line 1142: XskInfo is sometimes broken when running with multi connection, like -conns:100

ami-GS:
When disabling the RSS.
Using different Partition for NAT rebinded packets
image

When enabling the RSS.
Always use same Partition (once connection is established)
image

PR #4556, opened 2024-09-23
Change the CXPLAT LARGE IO BUFFER SIZE to 0xFFE3 on unix platforms.

noah-hwang:
## Description
Sendmsg always returns its length parameter, indicating success. With messages greater than 65507 (0xFFE3) bytes, it returns a Message too long error,  corresponding error code EOVERFLOW.
Payload size calculate: 65535 - 8 (UDP header) - 20 (IP header) = 65507 bytes.
Therefore the size of the IO Buffer should be less than or equal to 65507 bytes.

Online discussion of related issues
https://stackoverflow.com/questions/11809727/sendto-returns-values-greater-than-mtu-on-sock-dgram-udp-socket

## Testing

## Documentation


nibanks:
Thanks so much for finding/fixing this bug!

src/platform/datapath_epoll.c
```diff
@@ -33,8 +33,9 @@ CXPLAT_STATIC_ASSERT((SIZEOF_STRUCT_MEMBER(QUIC_BUFFER, Buffer) == sizeof(void*)
 
 //
 // The maximum single buffer size for coalesced IO payloads.
+// Payload size: 65535 - 8 (UDP header) - 20 (IP header) = 65507 bytes.
```
nibanks, line 36: What happens for IPv6? Its header size is bigger. We can use this for the pool allocation size, but should we be returning a different max size to actually _use_ when doing IPv6?

noah-hwang: Refer to this article
[(https://blog.cloudflare.com/virtual-networking-101-understanding-tap/)]
Payload size: 65507 bytes for IPv4(without options and padding field) and 65527 for IPv6(Payload length field no include P header)
Perhaps a more rigorous approach is needed here

nibanks (approved):

PR #4561, opened 2024-09-25
Don't Access Volatile Variables Directly

nibanks:
## Description

We were getting arm64 build errors when we accessed volatile variables directly in C++ code. Remove those.

## Testing

CI/CD

## Documentation

N/A

### DIFF
--- a/src/perf/lib/Tcp.cpp
+++ b/src/perf/lib/Tcp.cpp
@@ -199,7 +199,7 @@ bool TcpWorker::Initialize(TcpEngine* _Engine, uint16_t PartitionIndex)
     Engine = _Engine;
     ExecutionContext.Callback = DoWork;
     ExecutionContext.Context = this;
-    InterlockedFetchAndSetBoolean(&ExecutionContext.Ready);
+    WriteBooleanNoFence(&ExecutionContext.Ready, TRUE);
     ExecutionContext.NextTimeUs = UINT64_MAX;
 
     #ifndef _KERNEL_MODE // Not supported on kernel mode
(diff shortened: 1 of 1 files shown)


src/perf/lib/Tcp.cpp
```diff
@@ -199,7 +199,7 @@ bool TcpWorker::Initialize(TcpEngine* _Engine, uint16_t PartitionIndex)
     Engine = _Engine;
     ExecutionContext.Callback = DoWork;
     ExecutionContext.Context = this;
-    ExecutionContext.Ready = TRUE;
+    InterlockedFetchAndSetBoolean(&ExecutionContext.Ready);
```
mtfriesen, line 202: are you just trying to WriteBooleanNoFence?

nibanks: Perhaps. I'm not sure if the fence is needed or not. We already had this helper so I was just using it.

mtfriesen: On Windows, at least, WriteBooleanNoFence/WriteBooleanRelease are the recommended ways to set a volatile boolean - they should suppress any ISO volatile warnings you're seeing.

nibanks: Updated

src/perf/lib/Tcp.cpp
```diff
@@ -199,7 +199,7 @@ bool TcpWorker::Initialize(TcpEngine* _Engine, uint16_t PartitionIndex)
     Engine = _Engine;
     ExecutionContext.Callback = DoWork;
     ExecutionContext.Context = this;
-    ExecutionContext.Ready = TRUE;
+    WriteBooleanNoFence(&ExecutionContext.Ready, TRUE);
```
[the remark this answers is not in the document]
nibanks: I've reverted back to `InterlockedFetchAndSetBoolean` for now since it's cross platform. We can look at a future solution for cross-platform `WriteBooleanNoFence`.

mtfriesen (approved):

Issue #4409, opened 2024-07-22
Register fails on system with many cores


ddavila0:
### Describe the bug

Trying to execute the sample binary fails when calling RegistrationOpen (https://github.com/microsoft/msquic/blob/main/src/tools/sample/sample.c#L1015) with the following error:
```
[ lib][0xa5aa20] Worker start [PlatformWorkerThreadStart:/home/users/ddavila/projects/tcp/quic/git/msquic/src/platform/platform_worker.c:477]
[ lib][0xa5aab0] Worker start [PlatformWorkerThreadStart:/home/users/ddavila/projects/tcp/quic/git/msquic/src/platform/platform_worker.c:477]
[ lib] ERROR, CxPlatEventQInitialize. [LibraryError:/home/users/ddavila/projects/tcp/quic/git/msquic/src/platform/platform_worker.c:173]
```
This happens to thread number 254 out of 255 (might be a hint). If I reduce the number of threads to start to 16, by modifying the code (adding CxPlatWorkerCount=16; here: https://github.com/microsoft/msquic/blob/main/src/platform/platform_worker.c#L137) (couldn't find how to configure this on Linux), the registration works (it fails somewhere else down the road as I'd expect).

### Affected OS

- [ ] Windows
- [X] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

AlmaLinux release 8.10 (Cerulean Leopard)
Linux uaf-3.t2.ucsd.edu 6.3.8-1.el8.elrepo.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Jun 14 16:16:07 EDT 2023 x86_64 x86_64 x86_64 GNU/Linux

### MsQuic version

main

### Steps taken to reproduce bug

build with -DCMAKE_BUILD_TYPE=Debug -DQUIC_BUILD_TOOLS=True
./sample

### Expected behavior

I'll get to the part where it complains about missing arguments.

### Actual outcome

Fails with: 
RegistrationOpen failed, 0xc!

### Additional details

In a very similar system (with 16 cores) it works just fine

nibanks:
The logs indicate this is failing:
```c
inline
BOOLEAN
CxPlatEventQInitialize(
    _Out_ CXPLAT_EVENTQ* queue
    )
{
    return (*queue = epoll_create1(EPOLL_CLOEXEC)) != -1;
}
```
Do you not have enough handles or something configured? (I'm not a Linux expert)

nibanks:
[...]

PR #4562, opened 2024-09-26
print errno message for too many open files error

ddavila0:
## Description

Print a significant error message when "too many open files" limit is reached

## Testing

No tests needed, not a new feature

## Documentation

I don't think so


ddavila0:
this has to do with the issue discussed here:
https://github.com/microsoft/msquic/issues/4409

nibanks:
We don't want to add printf statements to the core code.

nibanks:
Please either convert this to a trace event or simply update the documentation on how to diagnose this.

PR #4602, opened 2024-10-07
Expose TTL (on handshake) to applications

ProjectsByJackHe:
## Description

There was an external ask for security purposes to expose packet TTL values in QUIC.

This PR bubbles up the TTL value from the network layer up to the transport layer.

TODO: We will need to rinse and repeat these changes for ALL platforms. 

## Testing

Modified the handshake test to do a HandshakeHopLimitTTL > 0 check.
Modified the Datapath test to check packets have a HopLimit > 0.

## Documentation

N/A


src/platform/datapath_winuser.c
```diff
@@ -3126,6 +3167,10 @@ CxPlatDataPathUdpRecvComplete(
                 } else if (CMsg->cmsg_type == IPV6_ECN) {
                     ECN = *(PINT)WSA_CMSG_DATA(CMsg);
                     CXPLAT_DBG_ASSERT(ECN < UINT8_MAX);
+                } else if (CMsg->cmsg_type == IPV6_HOPLIMIT) {
+                    HopLimitTTL = *(PINT)WSA_CMSG_DATA(CMsg);
+                    CXPLAT_DBG_ASSERT(HopLimitTTL <= 256);
+                    // printf("HopLimitTTL: %d \n", HopLimitTTL);
```
ProjectsByJackHe, line 3173: I will remove these once all platforms are supported. There is still a pending TODO in `Connection.c`

src/core/connection.c
```diff
@@ -5511,6 +5511,9 @@ QuicConnRecvDatagramBatch(
         CXPLAT_DBG_ASSERT(Packets[i]->Allocated);
         CXPLAT_ECN_TYPE ECN = CXPLAT_ECN_FROM_TOS(Packets[i]->TypeOfService);
         Packet = Packets[i];
+
+        Connection->Stats.Handshake.HopLimitTTL = Packet->HopLimitTTL; // TODO: Right now, this will update on EVERY packet. How can we filter to just the handshake packets?
```
nibanks, line 5515: The simplest way to only set it once is only set it if it's zero. Also, instead of being put here, I'd put this on line 5639, right after
```c
Connection->Stats.Recv.TotalBytes += Packet->BufferLength;
```

src/inc/msquic.h
```diff
@@ -555,6 +555,8 @@ typedef struct QUIC_STATISTICS_V2 {
 
     uint32_t SendEcnCongestionCount;        // Number of congestion events caused by ECN.
 
+    uint8_t  HopLimitTTL;                   // The TTL value in the initial packet of the handshake.
```
nibanks, line 558: Let's call it `HandshakeHopLimitTTL`.

src/inc/quic_datapath.h
```diff
@@ -222,6 +222,11 @@ typedef struct CXPLAT_RECV_DATA {
     //
     uint8_t TypeOfService;
 
+    //
+    // TTL Hoplimit field of the IP header of the received packet on handshake.
+    //
+    uint8_t HopLimitTTL;
```
nibanks, line 228: We should add a test in `DataPathTest.cpp` verifying packets received have non-zero values of this.

src/inc/quic_datapath.h
```diff
@@ -225,7 +225,7 @@ typedef struct CXPLAT_RECV_DATA {
     //
     // TTL Hoplimit field of the IP header of the received packet on handshake.
     //
-    uint8_t HopLimitTTL;
+    uint8_t HandshakeHopLimitTTL;
```
nibanks, line 228: This one should stay `HopLimitTTL` as this is per-packet.

src/inc/quic_datapath.h
ProjectsByJackHe: https://github.com/microsoft/msquic/pull/4602/commits/4ee3e03b29c83cc6418970566758d86632ae053b#diff-548ffedbc89d78947ec587388ebe4a695dbaaf6aa8d9a5032f355eec08b98800

src/core/connection.c
```diff
@@ -5511,6 +5511,7 @@ QuicConnRecvDatagramBatch(
         CXPLAT_DBG_ASSERT(Packets[i]->Allocated);
         CXPLAT_ECN_TYPE ECN = CXPLAT_ECN_FROM_TOS(Packets[i]->TypeOfService);
         Packet = Packets[i];
+
```
nibanks, line 5514: nit: revert newline

src/core/connection.h
```diff
@@ -275,6 +275,7 @@ typedef struct QUIC_CONN_STATS {
         uint32_t ClientFlight1Bytes;    // Sum of TLS payloads
         uint32_t ServerFlight1Bytes;    // Sum of TLS payloads
         uint32_t ClientFlight2Bytes;    // Sum of TLS payloads
+        uint8_t HandshakeHopLimitTTL;            // TTL value in the initial packet of the handshake.
```
nibanks, line 278:
```suggestion
        uint8_t HandshakeHopLimitTTL;   // TTL value in the initial packet of the handshake.
```

src/inc/msquic.h
```diff
@@ -555,6 +555,8 @@ typedef struct QUIC_STATISTICS_V2 {
 
     uint32_t SendEcnCongestionCount;        // Number of congestion events caused by ECN.
 
+    uint8_t  HandshakeHopLimitTTL;                   // The TTL value in the initial packet of the handshake.
```
nibanks, line 558:
```suggestion
    uint8_t  HandshakeHopLimitTTL;          // The TTL value in the initial packet of the handshake.
```

src/platform/datapath_epoll.c
```diff
@@ -1816,6 +1864,11 @@ CxPlatSocketContextRecvComplete(
                     CXPLAT_DBG_ASSERT_CMSG(CMsg, uint8_t);
                     TOS = *(uint8_t*)CMSG_DATA(CMsg);
                     FoundTOS = TRUE;
+                } else if (CMsg->cmsg_type == IP_TTL) {
+                    CXPLAT_DBG_ASSERT_CMSG(CMsg, uint8_t);
```
csujedihy, line 1868: The length of the IP_TTL ancillary data payload seems sizeof(int) as you use `CMSG_SPACE(sizeof(int))` to reserve the control buffer space.

Why are we not asserting it's a uint8?

src/core/connection.c
```diff
@@ -5637,6 +5637,9 @@ QuicConnRecvDatagrams(
 
         if (!IsDeferred) {
             Connection->Stats.Recv.TotalBytes += Packet->BufferLength;
+            if (Connection->Stats.Handshake.HandshakeHopLimitTTL == 0) {
+                Connection->Stats.Handshake.HandshakeHopLimitTTL = Packet->HopLimitTTL;
```
csujedihy, line 5641: You are using zero as a sentinel value. So, a valid TTL value should never be zero? (it would make sense for zero TTL packets to be dropped anyways but I am just double checking here).

src/platform/datapath_epoll.c
ProjectsByJackHe: good catch! I wasn't sure reserving X space maps to getting X sized payload back. I will remove this assert and keep the cast to uint8_t

src/platform/datapath_epoll.c
```diff
@@ -1816,6 +1864,11 @@ CxPlatSocketContextRecvComplete(
...
+                    HopLimitTTL = *(uint8_t*)CMSG_DATA(CMsg);
+                    CXPLAT_DBG_ASSERT(HopLimitTTL < 256);
+                    CXPLAT_DBG_ASSERT(HopLimitTTL > 0);
```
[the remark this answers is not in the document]
ProjectsByJackHe: nice catch! Will add with next commit.

src/platform/datapath_kqueue.c
```diff
@@ -1127,6 +1127,7 @@ CxPlatSocketContextRecvComplete(
 
     RecvPacket->Route->Queue = SocketContext;
     RecvPacket->TypeOfService = 0;
+    RecvPacket->HopLimitTTL = 0; // TODO: Check to see if it's even possible to get this information on macOS and FreeBSD. If so, implement it. Otherwise, document that it's not supported.
```
csujedihy, line 1130: https://man.freebsd.org/cgi/man.cgi?query=ip&sektion=4&manpath=FreeBSD+9.0-RELEASE

See "IP_RECVTTL" in this page. I think it's supported.

src/core/connection.c
ProjectsByJackHe: Good points about whether using 0 as a sentinel value is correct. 
I think we can do it this way, but @nibanks can chime in if not.

In terms of where to set this value, setting it here is the simplest. It all depends on the perf impact.

nibanks: If you move it into the handshake code then you'll need to pass along packet info. I don't care much either way.

csujedihy: No strong opinion. So feel free to ignore.

src/platform/datapath_kqueue.c
ProjectsByJackHe: Added a feature flag for IP_TTL support or not depending on the platform. Decided to hold off on MacOs enablement.

csujedihy (approved):

src/platform/datapath_epoll.c
```diff
@@ -199,8 +199,9 @@ typedef struct CXPLAT_SEND_DATA {
 } CXPLAT_SEND_DATA;
 
 typedef struct CXPLAT_RECV_MSG_CONTROL_BUFFER {
-    char Data[CMSG_SPACE(sizeof(struct in6_pktinfo)) +
-              2 * CMSG_SPACE(sizeof(int))];
+    char Data[CMSG_SPACE(sizeof(struct in6_pktinfo)) + // IP_PKTINFO
+              2 * CMSG_SPACE(sizeof(int)) // TOS
+              + CMSG_SPACE(sizeof(int))]; // IP_TTL
```
nibanks, line 204: nit, this should have just been `3 * CMSG_SPACE(sizeof(int)`

src/platform/datapath_epoll.c
```diff
@@ -1833,6 +1895,10 @@ CxPlatSocketContextRecvComplete(
 
         CXPLAT_FRE_ASSERT(FoundLocalAddr);
         CXPLAT_FRE_ASSERT(FoundTOS);
+        //
+        // TTL should always be available/enabled on Linux.
```
nibanks, line 1899: comment not really needed

src/platform/datapath_kqueue.c
```diff
@@ -577,6 +577,9 @@ CxPlatDataPathGetSupportedFeatures(
     _In_ CXPLAT_DATAPATH* Datapath
     )
 {
+    //
+    // Intentionally not enabling Feature_TTL on MacOS for now.
```
nibanks, line 581: comment not needed

src/platform/datapath_raw.c
```diff
@@ -150,7 +150,10 @@ RawDataPathGetSupportedFeatures(
     )
 {
     UNREFERENCED_PARAMETER(Datapath);
-    return CXPLAT_DATAPATH_FEATURE_RAW;
+    //
+    // TTL should always be available / enabled for XDP.
```
nibanks, line 154: comment not really needed

src/platform/datapath_winkernel.c
```diff
@@ -765,6 +766,25 @@ CxPlatDataPathQuerySockoptSupport(
 
     } while (FALSE);
 
+    do {
```
nibanks, line 769: IMO, a do/while loop here is overkill. You could have just had the if statement, and only added the feature flag if the status was success and build not equal to 20348.

src/platform/datapath_epoll.c
```diff
@@ -344,6 +345,10 @@ CxPlatDataPathCalculateFeatureSupport(
     }
 
     Datapath->Features |= CXPLAT_DATAPATH_FEATURE_TCP;
+    //
+    // TTL should always be available / enabled on Linux.
```
nibanks, line 349: Unnecessary

src/platform/datapath_winuser.c
```diff
@@ -734,6 +741,28 @@ CxPlatDataPathQuerySockoptSupport(
         Datapath->Features |= CXPLAT_DATAPATH_FEATURE_RECV_COALESCING;
     }
 }
+    //
+    // TODO: This "TTL_FEATURE check" code works, and mirrors the approach for Kernel mode.
+    //       However, it is considered a "hack" and we should determine whether or not
+    //       the current release story fits this current workaround.
+    //
+    HMODULE NtDllHandle = LoadLibraryA("ntdll.dll");
```
nibanks, line 749: We shouldn't have this logic in two places. Perhaps, instead we update `CX_PLATFORM` to store the `BuildNumber` (on Windows) and put this logic in platform_winuser.c (and might as well update _winkernel.c).

PR #4023, opened 2023-12-22
draft km xdp proto

nigriMSFT:
## Description

_Describe the purpose of and changes within this Pull Request._

## Testing

_Do any existing tests cover this change? Are new tests needed?_

## Documentation

_Is there any documentation impact for this change?_

PR #4605, opened 2024-10-09
Kernel Platform Worker Support

nibanks:
## Description

Extracts/refactors some of the code in #4023 to support platform worker threads, which is a prereq for kernel xdp support.

## Testing

CI/CD

## Documentation

N/A

### DIFF
--- a/src/inc/quic_platform_winkernel.h
+++ b/src/inc/quic_platform_winkernel.h
@@ -648,7 +648,7 @@ CxPlatEventQCleanup(
 _IRQL_requires_max_(DISPATCH_LEVEL)
 QUIC_INLINE
 BOOLEAN
-CxPlatEventQEnQueue(
+CxPlatEventQEnqueue(
     _In_ CXPLAT_EVENTQ* Queue,
     _In_ CXPLAT_SQE* Sqe
     )
@@ -670,7 +670,7 @@ CxPlatEventQEnQueue(
 _IRQL_requires_max_(PASSIVE_LEVEL)
 QUIC_INLINE
 uint32_t
-CxPlatEventQDeQueue(
+CxPlatEventQDequeue(
     _In_ CXPLAT_EVENTQ* Queue,
     _Out_writes_to_(Count, return) CXPLAT_CQE* Events,
     _In_range_(0, CXPLAT_EVENTQ_DEQUEUE_MAX)


nibanks:
Looks like the kernel BVTs are crashing. I will try to investigate later.

src/inc/quic_platform_winkernel.h
```diff
@@ -476,56 +477,122 @@ _CxPlatEventWaitWithTimeout(
...
+typedef struct CXPLAT_CQE {
+    void* UserData;
+} CXPLAT_CQE;
+
+#define CXPLAT_SQE CXPLAT_SQE
+#define CXPLAT_SQE_DEFAULT {0}
+typedef struct CXPLAT_SQE {
+    LIST_ENTRY Link;
+    void* UserData;
+    BOOLEAN IsQueued; // Prevent double queueing.
```
nibanks, line 495: Instead of a whole extra variable, can we use `Link->Flink` being `NULL` or not?

mtfriesen: If you have the appropriate synchronization, and you set Flink to NULL whenever it's not in the list, then sure. It's also not unusual to set an empty (non-inserted) link to be an empty list, i.e., Flink points to itself.

nigriMSFT: Agreed. IsQueued was originally copied from the winuser CXPLAT_SQE

nigriMSFT (dismissed):

src/platform/platform_internal.h
```diff
@@ -662,10 +663,14 @@ CxPlatWorkerPoolGetEventQ(
     _In_ uint16_t Index // Into the config processor array
     );
 
+#ifdef _KERNEL_MODE // Not supported on kernel mode (yet)
+#define CxPlatDataPathProcessCqe(X) CXPLAT_FRE_ASSERT(FALSE)
```
ami-GS, line 667: This means kernel mode doesn't process event (especially IO event)?
We will leave kernel mode datapath to be Wsk callback based?

nibanks: Correct. WSK stays as a callback.

ami-GS (approved):

nibanks:
Something seems broken in netperf runs. I can't tell if it's because of this or not. Needs to be investigated before merging.

ami-GS:
Is the netperf result reliable? wsk related failure is really few. others are more

nibanks:
Converted this back to draft, because we need to switch to using `NtCreateIoCompletion`, `NtSetIoCompletion`, and `NtRemoveIoCompletionEx`.

mtfriesen:
> Converted this back to draft, because we need to switch to using `NtCreateIoCompletion`, `NtSetIoCompletion`, and `NtRemoveIoCompletionEx`.

I suggest using the Zw variants for KM, unless QUIC is calling on behalf of the user mode thread calling QUIC, and all in the same thread context.

src/platform/datapath_winkernel.c
```diff
@@ -847,6 +846,11 @@ CxPlatDataPathInitialize(
         }
     }
 
+    if (!CxPlatWorkerPoolLazyStart(WorkerPool, Config)) {
```
ami-GS, line 849: Launching threads here can be waste of resources? because pure winkernel doesn't use them.
Should be put in raw datapath initialization? but no need to fix in this PR I think.

src/inc/quic_platform_winkernel.h
```diff
@@ -498,10 +497,13 @@ CxPlatEventQInitialize(
...
-    InitializeListHead(&queue->Events);
-    CxPlatEventInitialize(&queue->EventsAvailable, TRUE, FALSE);
-    return TRUE;
+    NTSTATUS status = NtCreateIoCompletion(
+                 queue,
+                 IO_COMPLETION_ALL_ACCESS,
+                 NULL,
+                 0
+                 );
+    return NT_SUCCESS(status);
```
nibanks, line 506:
```suggestion
    return NT_SUCCESS(NtCreateIoCompletion(queue, IO_COMPLETION_ALL_ACCESS, NULL, 0));
```

src/inc/quic_platform_winkernel.h
```diff
@@ -598,12 +579,46 @@ CxPlatEventQReturn(
...
+}
+
+inline
+CXPLAT_SQE*
+CxPlatCqeGetSqe(
     _In_ const CXPLAT_CQE* cqe
     )
 {
-    return cqe->UserData;
+    return CONTAINING_RECORD(cqe->ApcContext, CXPLAT_SQE, Cqe);
```
nibanks, line 621:
```suggestion
    return (CXPLAT_SQE)cqe->ApcContext;
```
It's not a containing record. You pass the SQE into `NtSetIoCompletion` as the `ApcContext`. So, you just need to cast it back to the right type.

src/inc/quic_platform_winkernel.h
```diff
@@ -452,6 +452,7 @@ typedef KEVENT CXPLAT_EVENT;
 #define CxPlatEventUninitialize(Event) UNREFERENCED_PARAMETER(Event)
 #define CxPlatEventSet(Event) KeSetEvent(&(Event), IO_NO_INCREMENT, FALSE)
 #define CxPlatEventReset(Event) KeResetEvent(&(Event))
+#define CxPlatEventClear(Event) KeClearEvent(&(Event))
```
nibanks, line 455: Unnecessary now. Please remove.

src/inc/quic_platform_winkernel.h
```diff
@@ -469,58 +469,155 @@ _CxPlatEventWaitWithTimeout(
...
+#define IO_COMPLETION_MODIFY_STATE  0x0002  // winnt
+#define IO_COMPLETION_ALL_ACCESS (STANDARD_RIGHTS_REQUIRED|SYNCHRONIZE|0x3) // winnt
+
+typedef struct _FILE_IO_COMPLETION_INFORMATION {
+    PVOID               KeyContext;
+    PVOID               ApcContext;
+    IO_STATUS_BLOCK     IoStatusBlock;
+} FILE_IO_COMPLETION_INFORMATION, *PFILE_IO_COMPLETION_INFORMATION;
+
+inline
```
nibanks, line 487: You shouldn't be defining these at all. Just declare them, so you can link to the kernel.

ami-GS: hmm? I did but failed to link.
I think these APIs are not published with SDK, right?

src/inc/quic_platform_winkernel.h
```diff
@@ -469,58 +469,131 @@ _CxPlatEventWaitWithTimeout(
...
+        count,
+        &entriesRemoved,
+        (wait_time == UINT32_MAX) ? NULL : &timeout,
+        FALSE
+    );
+
+    if (!NT_SUCCESS(status)) {
+        return 0;
+    }
+    return entriesRemoved;
```
nibanks, line 593:
```suggestion
    NTSTATUS status;
    ULONG entriesRemoved;
    if (wait_time == UINT32_MAX) {
        status =
            NtRemoveIoCompletionEx(
                *queue,
                events,
                count,
                &entriesRemoved,
                NULL,
                FALSE);
    } else {
        LARGE_INTEGER timeout;
        timeout.QuadPart = -10000LL * wait_time;
        status =
            NtRemoveIoCompletionEx(
                *queue,
                events,
                count,
                &entriesRemoved,
                &timeout,
                FALSE);
    }
    if (!NT_SUCCESS(status)) {
        return 0;
    }
    return entriesRemoved;
```
If we do it this way, then we don't have to calculate the `timeout` if there is no `wait_time`.

ami-GS:
Because platform_winkernel.c copies from zwapi.h, quic_platform_winkernel.h does same.

ami-GS:
https://github.com/microsoft/msquic/blob/101f6b92eedd569e1022edb4175f6fe8b17d91a9/src/core/worker.c#L174
RegistrationClose doesn't return by here

ami-GS:
Randomly hangs by WaitForSingleObject under Registration destructor

image

src/inc/quic_platform_winkernel.h
```diff
@@ -469,58 +469,235 @@ _CxPlatEventWaitWithTimeout(
...
+    LIST_ENTRY ListEntry;
+    ULONG PacketType;
+    PVOID KeyContext;
+    PVOID ApcContext;
+    NTSTATUS IoStatus;
+    ULONG_PTR IoStatusInformation;
+    PIO_MINI_PACKET_CALLBACK_ROUTINE MiniPacketCallback;
+    PVOID Context;
+    BOOLEAN Allocated;
+};
```
nibanks, line 542: I think you can delete this and leave it undefined in here, since you only deal in pointers and never write to it.

src/inc/quic_platform_winkernel.h
```diff
@@ -469,58 +469,235 @@ _CxPlatEventWaitWithTimeout(
...
+    OUT         PHANDLE IoCompletionHandle,
+    IN          ACCESS_MASK DesiredAccess,
+    IN OPTIONAL POBJECT_ATTRIBUTES ObjectAttributes,
+    IN OPTIONAL ULONG Count
+    );
+
+NTSYSAPI
+NTSTATUS
+NTAPI
+ZwSetIoCompletion (
```
nibanks, line 501: Unused now

src/inc/quic_platform_winkernel.h
```diff
@@ -469,58 +469,235 @@ _CxPlatEventWaitWithTimeout(
...
+        return FALSE;
+    }
+
+    queue->miniPacket = IoAllocateMiniCompletionPacket(IoMiniPacketCallbackRoutinePlaceholder, NULL);
+    if (queue->miniPacket == NULL) {
+        NtClose(queue->Handle);
+        return FALSE;
+    }
+
+    return NT_SUCCESS(Status);
```
nibanks, line 639:
```suggestion
    OBJECT_ATTRIBUTES KernelObjectAttributes;
    InitializeObjectAttributes(&KernelObjectAttributes,
                               /* ObjectName */ NULL,
                               /* AttributesFlag */ OBJ_KERNEL_HANDLE,
                               /* RootDirectory */ NULL,
                               /* SecurityAttributes */ NULL);

    NTSTATUS Status =
        ZwCreateIoCompletion(
            &queue->Handle,
            IO_COMPLETION_ALL_ACCESS,
            &KernelObjectAttributes,
            0);
    if (!NT_SUCCESS(Status)) {
        queue->Handle = NULL;
        return FALSE;
    }

    Status = ObReferenceObjectByHandle(queue->Handle,
                                       IO_COMPLETION_ALL_ACCESS,
                                       NULL,
                                       KernelMode,
                                       (PVOID*)&queue->IoCompletion,
                                       NULL);
    if (!NT_SUCCESS(Status)) {
        NtClose(queue->Handle);
        return FALSE;
    }

    queue->miniPacket = IoAllocateMiniCompletionPacket(IoMiniPacketCallbackRoutinePlaceholder, NULL);
    if (queue->miniPacket == NULL) {
        NtClose(queue->Handle);
        return FALSE;
    }

    return NT_SUCCESS(Status);
```

src/inc/quic_platform_winkernel.h
```diff
@@ -469,58 +469,235 @@ _CxPlatEventWaitWithTimeout(
...
+    PVOID ApcContext;
+    NTSTATUS IoStatus;
+    ULONG_PTR IoStatusInformation;
+    PIO_MINI_PACKET_CALLBACK_ROUTINE MiniPacketCallback;
+    PVOID Context;
+    BOOLEAN Allocated;
+};
+
+NTSTATUS
+IoSetIoCompletionEx (
```
nibanks, line 545: We don't need any `NTSYSAPI` or `NTAPI` on this or `IoFreeMiniCompletionPacket`? It still links/runs on all platforms?

ami-GS: hmm I don't see such place declaring with these annotations. and I don't see any link/run issue other than the hang

src/inc/quic_platform_winkernel.h
```diff
@@ -469,58 +469,241 @@ _CxPlatEventWaitWithTimeout(
...
+CxPlatEventQEnqueue(
+    _In_ CXPLAT_EVENTQ* queue,
+    _In_ CXPLAT_SQE* sqe
     )
 {
-    KeSetEvent(queue, IO_NO_INCREMENT, FALSE);
-    return TRUE;
+    return
+        NT_SUCCESS(
+        IoSetIoCompletionEx(
```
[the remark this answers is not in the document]
mtfriesen: It depends on what gets published in the API contract. We shouldn't depend on implementation details. I see no harm in returning the actual status here, unless the new kernel API clearly states the minipacket guarantees success.

src/inc/quic_platform_winkernel.h
```diff
@@ -469,58 +469,229 @@ _CxPlatEventWaitWithTimeout(
...
+    Status =
+        ObReferenceObjectByHandle(
+            Handle,
+            IO_COMPLETION_ALL_ACCESS,
+            NULL,
+            KernelMode,
+            (PVOID*)&queue->IoCompletion,
+            NULL);
+
+    NtClose(Handle);
```
mtfriesen, line 616: ideally we NT_VERIFY(NT_SUCCESS()) the result of closing handles.

nibanks: Why?

mtfriesen: It doesn't cost anything on release builds, and the assert on debug builds will catch otherwise silent failures.

src/core/library.c
```diff
@@ -2482,8 +2467,6 @@ QuicLibraryGenerateStatelessResetToken(
     return Status;
 }
 
-#ifndef _KERNEL_MODE
```
guhetier, line 2485: Loosely related comment:
`MsQuicExecutionDelete` as it is right now seems very error prone: it ignores the execution configs and delete the worker pool. What if only a subset of the execution configs are provided at once to the API?

src/inc/quic_platform_winkernel.h
```diff
@@ -494,58 +494,222 @@ _CxPlatEventWaitWithTimeout(
...
+        uint32_t count,
     _In_ uint32_t wait_time // milliseconds
     )
 {
-    UNREFERENCED_PARAMETER(count);
-    *events = NULL;
-    return STATUS_SUCCESS == _CxPlatEventWaitWithTimeout(queue, wait_time) ? 1 : 0;
+    NTSTATUS status;
+    ULONG entriesRemoved;
+    PLIST_ENTRY EntryPtrArray[CXPLAT_EVENTQ_DEQUEUE_MAX];
```
guhetier, line 683: nit: A debug assert for `count <= CXPLAT_EVENTQ_DEQUEUE_MAX` could help.

src/inc/quic_platform_winkernel.h
```diff
@@ -494,58 +494,222 @@ _CxPlatEventWaitWithTimeout(
 #define CxPlatEventWaitWithTimeout(Event, TimeoutMs) \
     (STATUS_SUCCESS == _CxPlatEventWaitWithTimeout(&Event, TimeoutMs))
 
+#ifndef IO_COMPLETION_ALL_ACCESS // TODO: Remove onces APIs are published
```
guhetier, line 497: Isn't all this already included from `msquic_winkernel` (which is included in this file?)

src/inc/quic_platform_winkernel.h
```diff
@@ -494,58 +494,222 @@ _CxPlatEventWaitWithTimeout(
...
 
+_IRQL_requires_max_(PASSIVE_LEVEL)
 QUIC_INLINE
 void
 CxPlatEventQCleanup(
     _In_ CXPLAT_EVENTQ* queue
     )
 {
-    UNREFERENCED_PARAMETER(queue);
+    ObDereferenceObject(*queue);
```
guhetier: nit: All this new code doesn't follow the variable naming convention

src/platform/platform_worker.c
```diff
@@ -657,6 +676,11 @@ CxPlatProcessDynamicPoolAllocators(
     _In_ CXPLAT_WORKER* Worker
     )
 {
+    if (Worker->State.TimeNow - Worker->State.LastPoolProcessTime < DYNAMIC_POOL_PROCESSING_PERIOD) {
+        return ; // Don't process pools too frequently
```
guhetier:
```suggestion
        return; // Don't process pools too frequently
```

src/core/library.c
nibanks: That's not really specific to this PR. But based on one of my other WIP PRs, I'm actually considering removing that function completely and cleaning up purely on MsQuicClose. Regardless, I'm not going to touch that in this PR.

src/inc/quic_platform_winkernel.h
nibanks: Not exactly. Not all include both I believe.

guhetier (approved):

PR #4607, opened 2024-10-10
Standard SecNetPerf Scenarios

nibanks:
## Description

Adds a new secnetperf cmd line arg `-scenario:` which standardizes a few common test cases. This then simplifies the secnetperf scripts for netperf to just use these scenarios.

## Testing

CI/CD

## Documentation

Updated help text

### DIFF
--- a/scripts/secnetperf-helpers.psm1
+++ b/scripts/secnetperf-helpers.psm1
@@ -524,7 +524,7 @@ function Get-LatencyOutput {
 
 # Invokes secnetperf with the given arguments for both TCP and QUIC.
 function Invoke-Secnetperf {
-    param ($Session, $RemoteName, $RemoteDir, $UserName, $SecNetPerfPath, $LogProfile, $TestId, $ExeArgs, $io, $Filter, $Environment, $RunId, $SyncerSecret)
+    param ($Session, $RemoteName, $RemoteDir, $UserName, $SecNetPerfPath, $LogProfile, $Scenario, $io, $Filter, $Environment, $RunId, $SyncerSecret)
 
     $values = @(@(), @())
     $latency = $null
@@ -536,26 +536,25 @@ function Invoke-Secnetperf {
         $tcpSupported = 0
     }
     $metric = "throughput"
-    if ($exeArgs.Contains("conns:16cpu")) { # TODO: figure out a better way to detect max RPS tests
+    if ($Scenario.Contains("rps")) {
         $metric = "rps"
-    } elseif ($exeArgs.Contains("plat:1")) {
+    } elseif ($Scenario.Contains("latency")) {
         $metric = "latency"
         $latency = @(@(), @())
         $extraOutput = Repo-Path "latency.txt"
         if (!$isWindows) {
             chmod +rw "$extraOutput"
         }
-    } elseif ($exeArgs.Contains("prate:1")) {
+    } elseif ($Scenario.Contains("hps")) {
         $metric = "hps"
     }
 
     for ($tcp = 0; $tcp -le $tcpSupported; $tcp++) {
 
     # Set up all the parameters and paths for running the test.
-    $execMode = $ExeArgs.Substring(0, $ExeArgs.IndexOf(" ")) # First arg is the exec mode
     $clientPath = Repo-Path $SecNetPerfPath
-    $serverArgs = "$execMode -io:$io"
-    $clientArgs = "-target:$RemoteName $ExeArgs -tcp:$tcp -trimout -watchdog:25000"
+    $serverArgs = "-scenario:$Scenario -io:$io"
+    $clientArgs = "-target:$RemoteName -scenario:$Scenario -io:$io -tcp:$tcp -trimout -watchdog:25000"
     if ($io -eq "xdp" -or $io -eq "qtip") {
         $serverArgs += " -pollidle:10000"
         $clientArgs += " -pollidle:10000"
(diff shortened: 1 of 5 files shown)


scripts/secnetperf-helpers.psm1
```diff
@@ -524,7 +524,7 @@ function Get-LatencyOutput {
 
 # Invokes secnetperf with the given arguments for both TCP and QUIC.
 function Invoke-Secnetperf {
-    param ($Session, $RemoteName, $RemoteDir, $UserName, $SecNetPerfPath, $LogProfile, $TestId, $ExeArgs, $io, $Filter, $Environment, $RunId, $SyncerSecret)
+    param ($Session, $RemoteName, $RemoteDir, $UserName, $SecNetPerfPath, $LogProfile, $Scenario, $io, $Filter, $Environment, $RunId, $SyncerSecret)
```
mtfriesen, line 527: nit: `$io` is randomly all lower-case.

nibanks: I noticed that when looking here, but decided not to make the unrelated change.

scripts/secnetperf.ps1
```diff
@@ -254,11 +254,7 @@ if ($isWindows) {
 $allTests = [System.Collections.Specialized.OrderedDictionary]::new()
 
 # > All tests:
-$allTests["tput-up"] = "-exec:maxtput -up:12s -ptput:1"
-$allTests["tput-down"] = "-exec:maxtput -down:12s -ptput:1"
-$allTests["hps-conns-100"] = "-exec:maxtput -rconn:1 -share:1 -conns:100 -run:12s -prate:1"
-$allTests["rps-up-512-down-4000"] = "-exec:lowlat -rstream:1 -up:512 -down:4000 -run:20s -plat:1"
-$allTests["max-rps-up-512-down-4000"] = "-exec:lowlat -conns:16cpu -streams:10 -rstream:1 -up:512 -down:4000 -run:20s -plat:1"
+$allTests = @("upload", "download", "hps", "rps-single", "rps-multi", "latency")
```
nibanks, line 257: I think this is effectively a breaking change, from an identifier `tput-up` to `upload` (and so on), and I might need to keep the old ID and map it to the scenario. @ProjectsByJackHe please let me know what you think.

src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -77,6 +77,8 @@ PrintHelp(
         "  -platency<0/1>           Print latency statistics. (def:0)\n"
         "\n"
         "  Scenario options:\n"
+        "  -scenario:<profile>      Scenario profile to use.\n"
+        "                            - {upload, download, hps, rps-single, rps-multi, latency}.\n"
```
nibanks, line 81: Bikeshed: Should we standardize on naming to be {scenario}, {scenario}-multi to indicate single connection vs multi connection versions, or {scenario}-single, {scenario}-multi? Because I was thinking we should have multi connection throughput and possible latency tests too....

I think I like {scenario}, {scenario}-multi

nibanks: In other words, thoughts on:
```
{upload, upload-multi, download, download-multi, hps, rps, rps-multi, latency, latency-multi}
```

scripts/secnetperf.ps1
nibanks: @ProjectsByJackHe as we discussed in our meeting yesterday, we need to ensure we're Ok with this change, and then update the dashboard and other netperf files as necessary. Please take a look. Thanks!

ProjectsByJackHe: Tracking: https://github.com/microsoft/netperf/issues/401

nibanks:
@ProjectsByJackHe where are we on this? Is netperf ready to merge this?

ProjectsByJackHe:
> @ProjectsByJackHe where are we on this? Is netperf ready to merge this?

Doing some local dry runs to ensure these changes do not break compatibility.

Tracking: https://github.com/microsoft/netperf/issues/401

scripts/secnetperf.ps1
```diff
@@ -251,17 +251,11 @@ if ($isWindows) {
...
-$allTests["tput-up"] = "-exec:maxtput -up:12s -ptput:1"
-$allTests["tput-down"] = "-exec:maxtput -down:12s -ptput:1"
-$allTests["hps-conns-100"] = "-exec:maxtput -rconn:1 -share:1 -conns:100 -run:12s -prate:1"
-$allTests["rps-up-512-down-4000"] = "-exec:lowlat -rstream:1 -up:512 -down:4000 -run:20s -plat:1"
-$allTests["max-rps-up-512-down-4000"] = "-exec:lowlat -conns:16cpu -streams:10 -rstream:1 -up:512 -down:4000 -run:20s -plat:1"
+# Test all supported scenarios.
+$allScenarios = @("upload", "download", "hps", "rps", "rps-multi", "latency")
 
 $hasFailures = $false
-$json["run_args"] = $allTests
```
ProjectsByJackHe, line 264: Hmm, if netperf will always use the default options (stream count, connection count...) for each scenario, then this is fine. But otherwise, we might still need this.

ProjectsByJackHe:
> > @ProjectsByJackHe where are we on this? Is netperf ready to merge this?
> 
> Doing some local dry runs to ensure these changes do not break compatibility.
> 
> Tracking: [microsoft/netperf#401](https://github.com/microsoft/netperf/issues/401)

Almost ready for merge! Once https://github.com/microsoft/netperf/pull/414 gets merged first, validation passes, then this PR is ready for merge.

ProjectsByJackHe (approved):
LGTM; I tested and ran a dry run. Should be good.

src/inc/msquichelper.h
```diff
@@ -304,7 +304,7 @@ IsValue(
     _In_z_ const char* toTestAgainst
     )
 {
-    return _strnicmp(name, toTestAgainst, CXPLAT_MIN(strlen(name), strlen(toTestAgainst))) == 0;
+    return _strnicmp(name, toTestAgainst, strlen(toTestAgainst)) == 0;
```
ProjectsByJackHe, line 307: Why was this change made?

nibanks: We wanted to do an exact match, not just on the shortest part. This is because we're using some of the same prefix and we don't want incorrect matches.

PR #4614, opened 2024-10-15
Support Thread Affinity for Execution Config

nibanks:
## Description

Adds another execution config knob for setting affinitizing threads to their CPU.

## Testing

CI/CD

## Documentation

N/A


src/inc/msquic.h
```diff
@@ -272,6 +272,7 @@ typedef enum QUIC_EXECUTION_CONFIG_FLAGS {
     QUIC_EXECUTION_CONFIG_FLAG_XDP              = 0x0004,
     QUIC_EXECUTION_CONFIG_FLAG_NO_IDEAL_PROC    = 0x0008,
     QUIC_EXECUTION_CONFIG_FLAG_HIGH_PRIORITY    = 0x0010,
+    QUIC_EXECUTION_CONFIG_FLAG_AFFINITIZE       = 0x0020,
```
mtfriesen, line 275: IMO, this should indicate what the threads are being affinitized to - NUMA node, single RSS CPU, something else?

nibanks: I'd prefer to just leverage documentation for what, which is currently being deferred since these are all prototype features. I just want to get this in so our partners can experiment. If it doesn't actually help, it might ultimately be removed.

mtfriesen (approved):

src/inc/msquic.h
mtfriesen: Generally speaking, I think even experimental features should be designed as if they will be retained, because it is difficult to remove/rename existing dependencies. That said, I'm ok with unblocking folks here.

src/core/worker.c
```diff
@@ -118,8 +118,13 @@ QuicWorkerInitialize(
...
         }
 
-        if (MsQuicLib.ExecutionConfig && MsQuicLib.ExecutionConfig->Flags & QUIC_EXECUTION_CONFIG_FLAG_HIGH_PRIORITY) {
-            ThreadFlags |= CXPLAT_THREAD_FLAG_HIGH_PRIORITY;
+        if (MsQuicLib.ExecutionConfig) {
+            if (MsQuicLib.ExecutionConfig->Flags & QUIC_EXECUTION_CONFIG_FLAG_HIGH_PRIORITY) {
+                ThreadFlags |= CXPLAT_THREAD_FLAG_HIGH_PRIORITY;
+            }
+            if (MsQuicLib.ExecutionConfig->Flags & QUIC_EXECUTION_CONFIG_FLAG_AFFINITIZE) {
+                ThreadFlags |= CXPLAT_THREAD_FLAG_SET_AFFINITIZE;
```
ami-GS, line 126: Looks like just adding flag? where is the actual logic?

nibanks: The logic was already implemented elsewhere. This is just a new config flag.

PR #4616, opened 2024-10-16
Initial External Execution Interface

nibanks:
## Description

Provides an interface for the applicaiton layer to completely control execution of all MsQuic work. In other words, MsQuic would create no threads internally.

## Testing

TODO

## Documentation

TODO

### DIFF
--- a/src/inc/msquic.h
+++ b/src/inc/msquic.h
@@ -264,15 +264,15 @@ typedef enum QUIC_DATAGRAM_SEND_STATE {
 #define QUIC_DATAGRAM_SEND_STATE_IS_FINAL(State) \
     ((State) >= QUIC_DATAGRAM_SEND_LOST_DISCARDED)
 
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+
 typedef enum QUIC_EXECUTION_CONFIG_FLAGS {
     QUIC_EXECUTION_CONFIG_FLAG_NONE             = 0x0000,
-#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
     QUIC_EXECUTION_CONFIG_FLAG_QTIP             = 0x0001,
     QUIC_EXECUTION_CONFIG_FLAG_RIO              = 0x0002,
     QUIC_EXECUTION_CONFIG_FLAG_XDP              = 0x0004,
     QUIC_EXECUTION_CONFIG_FLAG_NO_IDEAL_PROC    = 0x0008,
     QUIC_EXECUTION_CONFIG_FLAG_HIGH_PRIORITY    = 0x0010,
-#endif
 } QUIC_EXECUTION_CONFIG_FLAGS;
 
 DEFINE_ENUM_FLAG_OPERATORS(QUIC_EXECUTION_CONFIG_FLAGS)
@@ -292,6 +292,82 @@ typedef struct QUIC_EXECUTION_CONFIG {
 
 #define QUIC_EXECUTION_CONFIG_MIN_SIZE \
     (uint32_t)FIELD_OFFSET(QUIC_EXECUTION_CONFIG, ProcessorList)
+
+//
+// Execution Context abstraction, which allows the application layer to
+// completely control execution of all MsQuic work.
+//
+
+typedef struct QUIC_EXECUTION_CONTEXT_CONFIG {
+    uint32_t IdealProcessor;
+    uint32_t PollingIdleTimeoutUs;
+    QUIC_EVENTQ* EventQ;
+} QUIC_EXECUTION_CONTEXT_CONFIG;
+
+typedef struct QUIC_EXECUTION_CONTEXT QUIC_EXECUTION_CONTEXT;
+
+//
+// This is called create the execution contexts.
+//
+typedef
+_IRQL_requires_max_(PASSIVE_LEVEL)
+QUIC_STATUS
+(QUIC_API * QUIC_EXECUTION_CREATE_FN)(
+    _In_ QUIC_EXECUTION_CONFIG_FLAGS Flags, // Used for datapath type
+    _In_ uint32_t Count,
+    _In_reads_(Count) QUIC_EXECUTION_CONTEXT_CONFIG* Configs,
+    _Out_writes_(Count) QUIC_EXECUTION_CONTEXT** ExecutionContexts
+    );
+
+//
+// This is called to delete the execution contexts.
+//
+typedef
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+(QUIC_API * QUIC_EXECUTION_DELETE_FN)(
+    _In_ uint32_t Count,
+    _In_reads_(Count) QUIC_EXECUTION_CONTEXT** ExecutionContexts
+    );
+
+//
+// This is called to allow MsQuic to process any polling work. It returns the
+// number of milliseconds until the next scheduled timer expiration.
+//
+// TODO: Should it return an indication for if we should yield?
+//
+typedef
+_IRQL_requires_max_(PASSIVE_LEVEL)
+uint32_t
+(QUIC_API * QUIC_EXECUTION_POLL_FN)(
+    _In_ QUIC_EXECUTION_CONTEXT* ExecutionContext
+    );
+
+//
+// This is called to allow MsQuic to process any completions that have occurred.
+//
+typedef
+_IRQL_requires_max_(PASSIVE_LEVEL)
+uint32_t
+(QUIC_API * QUIC_EXECUTION_PROCESS_CQE_FN)(
+    _In_ QUIC_EXECUTION_CONTEXT* ExecutionContext,
+    _In_reads_(CqeCount) QUIC_CQE* Cqes,
+    _In_ uint32_t CqeCount
+    );
+
+//
+// The table of execution functions.
+//
+typedef struct QUIC_EXECUTION_TABLE {
+
+    QUIC_EXECUTION_CREATE_FN ExecutionCreate;
+    QUIC_EXECUTION_DELETE_FN ExecutionDelete;
+    QUIC_EXECUTION_POLL_FN Poll;
+    QUIC_EXECUTION_PROCESS_CQE_FN ProcessCqe;
+
+} QUIC_EXECUTION_TABLE;
+
+#endif
 
 typedef struct QUIC_REGISTRATION_CONFIG { // All fields may be NULL/zero.
     const char* AppName;
@@ -857,6 +933,10 @@ void
 #endif
 #define QUIC_PARAM_GLOBAL_TLS_PROVIDER                  0x0100000A  // QUIC_TLS_PROVIDER
 #define QUIC_PARAM_GLOBAL_STATELESS_RESET_KEY           0x0100000B  // uint8_t[] - Array size is QUIC_STATELESS_RESET_KEY_LENGTH
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+#define QUIC_PARAM_GLOBAL_EXECUTION_TABLE               0x0100000C  // QUIC_EXECUTION_TABLE
+#endif
+
 //
 // Parameters for Registration.
 //
--- a/src/inc/msquic_posix.h
+++ b/src/inc/msquic_posix.h
@@ -515,6 +515,26 @@ QuicAddrToString(
     return TRUE;
 }
 
+//
+// Event Queue Abstraction
+//
+
+#if __linux__ // epoll
+
+typedef int QUIC_EVENTQ;
+typedef struct epoll_event QUIC_CQE;
+
+#elif __APPLE__ || __FreeBSD__ // kqueue
+
+typedef int QUIC_EVENTQ;
+typedef struct kevent QUIC_CQE;
+
+#else
+
+#error Unsupported Platform
+
+#endif
+
 #if defined(__cplusplus)
 }
 #endif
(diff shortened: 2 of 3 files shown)


src/inc/msquic.h
```diff
@@ -293,6 +293,82 @@ typedef struct QUIC_EXECUTION_CONFIG {
...
+
+//
+// This is called to allow MsQuic to process any completions that have occurred.
+//
+typedef
+_IRQL_requires_max_(PASSIVE_LEVEL)
+uint32_t
+(QUIC_API * QUIC_EXECUTION_PROCESS_CQE_FN)(
+    _In_ QUIC_EXECUTION_CONTEXT* ExecutionContext,
+    _In_reads_(CqeCount) QUIC_CQE* Cqes,
```
nibanks, line 354: Need a function/way for the app to determine a CQE is for QUIC and not themselves.

nibanks: I added a function above, but it still has these assumptions.

src/inc/msquic.h
```diff
@@ -293,6 +293,100 @@ typedef struct QUIC_EXECUTION_CONFIG {
...
+    _In_ uint32_t CqeCount
+    );
+
+//
+// The table of execution functions.
+//
+typedef struct QUIC_EXECUTION_TABLE {
```
nibanks, line 376: Should this be a separate table, or do we simply append to the QUIC API table?

src/inc/msquic.h
mtfriesen: This needs to scale to work with arbitrary numbers of layers and unrelated libraries sharing an event queue. I don't think integer/bitfields will be sufficient, unless we want to operate a bitfield registry or perform dynamic bit allocations.

mtfriesen: With IOCPs on Windows, the standard way to solve this is to define an encapsulating struct containing an OVERLAPPED and whatever other data is required to provide app context per IO, e.g. an arbitrary function pointer and function pointer context. IDK if that will also be viable with epoll-style models, but I'd assume so.

src/tools/execution/execution_windows.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+        while (true) {
+            uint32_t WaitTime = MsQuicExec.Poll(ExecContext);
+
+            OVERLAPPED_ENTRY Cqes[8];
+            ULONG CqeCount = 0;
+            if (GetQueuedCompletionStatusEx(IOCP, Cqes, ARRAYSIZE(Cqes), &CqeCount, WaitTime, FALSE)) {
+                for (ULONG i = 0; i < CqeCount; ++i) {
+                    if (MsQuicExec.CheckCqe(Cqes+i)) {
+                        MsQuicExec.ProcessCqe(ExecContext, Cqes+i, 1);
```
nibanks, line 89: I wonder if any app usage would ever be able to pass more than 1 in, considering completion might be interleaved in the queue. Should the function always just take one?

src/inc/msquic.h
mtfriesen: I don't think the struct itself needs to be consistent across platforms - you just need to define CxPlat wrappers for each operation. The suggestion above (function pointer, function pointer context, per IO) is sufficiently abstract and unambiguous, and should, I think, be viable on any platform.

mtfriesen: You also don't _really_ need to store a function pointer context itself, if you allow this CXPLAT_IO struct to be further encapsulated by callers, and simply pass the CXPLAT_IO during completion. Then they can unpack their own function pointer context, if they wish.

mtfriesen: Windows:

```
typedef struct CXPLAT_IO {
    OVERLAPPED Ov;
    VOID (*Completion)(struct CXPLAT_IO *Io);
};
```

mtfriesen: Linux:

struct CXPLAT_IO {
    CXPLAT_LISTENTRY Link; // IDK ?
    VOID (*Completion)(struct CXPLAT_IO *Io);
};

mtfriesen: And I suppose none of this should technically be restricted to "IO" and instead consider all of the above to mean "Work".

src/tools/execution/execution_windows.cpp
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        while (!AllDone) {
+            uint32_t WaitTime = MsQuic->ExecutionPoll(ExecContext);
+
+            OVERLAPPED_ENTRY Overlapped[8];
+            ULONG OverlappedCount = 0;
+            if (GetQueuedCompletionStatusEx(IOCP, Overlapped, ARRAYSIZE(Overlapped), &OverlappedCount, WaitTime, FALSE)) {
+                for (ULONG i = 0; i < OverlappedCount; ++i) {
+                    QUIC_CQE* Cqe = CONTAINING_RECORD(Overlapped[i].lpOverlapped, QUIC_CQE, Overlapped);
```
mtfriesen, line 79: is anything creating QUIC_CQEs in this sample program?

src/tools/execution/execution_windows.cpp
```diff
@@ -0,0 +1,91 @@
...
+{
+    MsQuicApi _MsQuic;
+    if (!_MsQuic.IsValid()) { return 1; }
+    MsQuic = &_MsQuic;
+
+    QUIC_STATUS Status;
+    HANDLE IOCP = CreateIoCompletionPort(INVALID_HANDLE_VALUE, nullptr, 0, 1);
+    QUIC_EXECUTION_CONTEXT_CONFIG ExecConfig = { 0, 0, &IOCP };
+    QUIC_EXECUTION_CONTEXT* ExecContext = nullptr;
+    if (QUIC_FAILED(Status = MsQuic->ExecutionCreate(QUIC_EXECUTION_CONFIG_FLAG_NONE, 1, &ExecConfig, &ExecContext))) {
```
mtfriesen, line 44: how may different binaries are expected to use this interface? i'm wondering about how layering will work, if, say, an app is built on top of an http library on top of msquic.dll.

should the app also open a handle directly to the msquic execution library, or should the http library create a thin wrapper around this, e.g. by typedefing a struct equivalent to the QUIC_CQE struct?

mtfriesen: high-level point: if the intent of this interface is to allow arbitrary code within a process to efficiently share an execution context, IO completions, timers, etc., then how are component/binaries boundaries supposed to interact - with each other, or directly with the lowest-level execution scheduler?

mtfriesen: > high-level point: if the intent of this interface is to allow arbitrary code within a process to efficiently share an execution context, IO completions, timers, etc., then how are component/binaries boundaries supposed to interact - with each other, or directly with the lowest-level execution scheduler?
> 
> I don't understand the question.

You've already answered the question here, which is whether or not all binaries/layer should directly register with this EC library, or register with the layer directly below them. I generally agree they should register with the layer below, not directly with this library.

src/tools/execution/execution_windows.cpp
nibanks: No, not yet. I will add one.

docs/Execution.md
```diff
@@ -109,3 +109,75 @@ graph TD
...
+
+```c++
+HANDLE IOCP = CreateIoCompletionPort(INVALID_HANDLE_VALUE, nullptr, 0, 1);
+QUIC_EXECUTION_CONTEXT_CONFIG ExecConfig = { 0, &IOCP };
+
+QUIC_EXECUTION_CONTEXT* ExecContext = nullptr;
+QUIC_STATUS Status = MsQuic->ExecutionCreate(QUIC_EXECUTION_CONFIG_FLAG_NONE, 0, 1, &ExecConfig, &ExecContext);
+```
+
+The above code createa a new IOCP (for Windows), sets up an execution config, indicating an ideal processor of 0 and the pointer to the IOCP, and then calls MsQuic to create 1 execution context.
```
mtfriesen, line 158: A processor of `0` has no clear, unambiguous meaning on Win32. A [`PROCESSOR_NUMBER`](https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-processor_number) is the only unambiguous processor identifier in Windows user mode, unless MsQuic provides its own canonical representation to apps.

mtfriesen: One could argue that `0` should always map to `Group 0, Number 0` in any sensible CPU numbering scheme, but certainly once you reach the integer `1`, the point stands.

docs/Execution.md
```diff
@@ -109,3 +109,75 @@ graph TD
...
+
+## Custom Execution
+
+MsQuic also supports scenarios where the application layer creates the threads that MsQuic uses to execute on.
+In this mode, the application creates one or more execution contexts that MsQuic will use to run all its internal logic.
+The application is responsible for calling down into MsQuic to allow these execution contexts to run.
+
+To create an execution context, the app much first create an event queue object, which is a platform specific type:
+
+- Windows: IOCP
```
csujedihy, line 121: why do we expose a platform specific queue object rather than having a more abstracted interface?

nibanks: Such as what? The reason we use the platform queue object is because that is also a requirement for other (i.e. storage) IO. So, it's not expected to be a _new_ requirement. Also, this is all opt-in. They can continue to use MsQuic without doing this, and it will simply create the background threads still.

docs/Execution.md
nibanks: We can look around to see how other platform abstraction layers (std, boost) solve this problem.

docs/Execution.md
nibanks: No, they shouldn't create anything specific for MsQuic. If they have an epoll queue (on Linux) for their storage IO, they should reuse that for MsQuic.

docs/Execution.md
mtfriesen: According to stackoverflow, boost sidesteps the problem by allowing you to get a native thread handle, and then go off and call platform-specific affinity routines yourself. Personally I think an `int` should be a reasonable way to identify CPUs, so if QUIC just exposes ConvertProcessorNumberToProcessorIndex and ConvertProcessorIndexToProcessorNumber helper routines on Windows, you'd be set.

mtfriesen: Alternatively, you could `typedef int QUIC_CPU_ID` everywhere else and `typedef PROCESSOR_NUMBER QUIC_CPU_ID` on Windows and let apps deal with the difference if they also compile cross-plat.

src/platform/platform_worker.c
```diff
@@ -471,22 +563,39 @@ CxPlatRunExecutionContexts(
...
     }
 }
 
+uint32_t
+CxPlatWorkerPoolWorkerPoll(
+    _In_ QUIC_EXECUTION* Execution
+    )
+{
+    CXPLAT_WORKER* Worker = (CXPLAT_WORKER*)Execution;
+    Worker->State.ThreadID = CxPlatCurThreadID();
```
nibanks, line 588: TODO: Set this in all callbacks?

nibanks: Shouldn't be necessary, as we're not using this in api.c anymore, because we're a custom execution context.

docs/Execution.md
nibanks: Let's open a bug to track this for possible future changes. For now, it's orthogonal to this PR.

mtfriesen: IMO it is a blocker for this PR, because it takes an existing internal definition (which we could fix at any point down the road, and was internally consistent within MsQuic and makes it into an API/ABI surface that we cannot break and lacks consistency.

nibanks: This is a "preview" API right now, so we can still change it. We can look at this issue very soon after this PR.

src/rs/ffi/mod.rs
```diff
@@ -8,6 +8,50 @@
 
 pub type QUIC_ADDR = std::ffi::c_void;
 
+#[cfg(target_os = "windows")]
+pub type HANDLE = std::os::windows::raw::HANDLE;
+
+#[cfg(target_os = "windows")]
+#[repr(C)]
+pub struct OVERLAPPED {
```
youyuanwu, line 16: U can allowlist the bindgen types by add them here:
https://github.com/microsoft/msquic/blob/58fd56c6ba8b8190042660ca83d7748b7a1ca558/scripts/build.rs#L210C1-L210C61

This saves the manual work here but can get a lot more generated code.
I originally blocklisted QUIC_ADDR  because it creates too much code for manipulating socket_addr. 
Maybe we should allowlist all types and get rid of the c_types dependencies.

docs/Execution.md
```diff
@@ -116,6 +116,78 @@ graph TD
...
+bool AllDone = false;
+while (!AllDone) {
+    uint32_t WaitTime = MsQuic->ExecutionPoll(ExecContext);
+
+    ULONG OverlappedCount = 0;
+    OVERLAPPED_ENTRY Overlapped[8];
+    if (GetQueuedCompletionStatusEx(IOCP, Overlapped, ARRAYSIZE(Overlapped), &OverlappedCount, WaitTime, FALSE)) {
+        for (ULONG i = 0; i < OverlappedCount; ++i) {
+            QUIC_SQE* Sqe = CONTAINING_RECORD(Overlapped[i].lpOverlapped, QUIC_SQE, Overlapped);
+            Sqe->Completion(&Overlapped[i]);
```
youyuanwu, line 179: Can user quic callback event be invoked on this user thread here?
(I am thinking about if it is possible to write a single thread application, and potential deadlocks.)

nibanks: Yes, this app is single threaded. It works and I have the logs to prove it.

mtfriesen (approved):

Issue #4144, opened 2024-02-19
secnetperf not working for big payload sizes


PratikKumarSaha:
### Describe the bug

The QUIC version (till 7th Oct 2023) is working fine on my machine for payload size of 1.5GB, however, it is not working for the commits after 7th October 2023.

We have assigned 2 CPUs each for SERVER and CLIENT. For small payload size (eg : 150000 ) it works fine but not for bigger payload sizes. Following is the output to distinguish the results.

```
[root@9030bf879c51 x86_Debug_openssl]# ./secnetperf -exec:maxtput -test:tput -target:192.168.96.3 -up:150000 -ptput:1
Started!

Result: Upload 150000 bytes @ 236173 kbps (5.081 ms) .

[root@9030bf879c51 x86_Debug_openssl]# ./secnetperf -exec:maxtput -test:tput -target:192.168.96.3 -up:1500000000 -ptput:1
Started!

Aborted (core dumped) 
```

### Affected OS

- [ ] Windows
- [X] Linux
- [ ] macOS
- [X] Other (specify below)

### Additional OS information

- NAME = "Red Hat Enterprise Linux" 
- VERSION="8.8 (Ootpa)"
- ID = " rhel"
- Architecture: x86_64
- CPU op-mode(s): 32-bit, 64-bit
- Byte Order: Little Endian
- CPU(s) : 40
- On-line CPUS(s) list: 0-39
- Thread(s) per core: 2
- Core(s) per socket: 10
- Socket(s) : 2
- NUMA mode: 2
- Vendor ID: GenuineIntel
- CPU family: 6
- Model: 85
- Model Name: Intel(R) Xeon(R) Silver 4210 CPU @ 2.20GHz

### MsQuic version

main

### Steps taken to reproduce bug

1. /secnetperf -exec:maxtput  **(Server)**
2.  ./secnetperf -exec:maxtput -test:tput -target:192.168.96.3 -up:1500000000 -ptput:1 **(Client)**

### Expected behavior

It should not have thrown core dumped since it was working properly for the commits before and om 7th October 2023.

### Actual outcome

```
[root@9030bf879c51 x86_Debug_openssl]# ./secnetperf -exec:maxtput -test:tput -target:192.168.96.3 -up:150000 -ptput:1
Started!

Result: Upload 150000 bytes @ 236173 kbps (5.081 ms) .

[...]

PR #4621, opened 2024-10-21
Fix crash issue on Samsung phone A217M when send big payload size

URNOTCharlotte:
## Description

My phone: Samsung A217M (Android 12)
After I upgraded MsQuic to v2.3.6, crash occurs when sending big payload size. I have tried the releases both v.2.3.4 and v.2.3.5, also failed on my phone.
With MsQuic v2.2.4, it's ok. 
Please see the discussion here:
https://github.com/microsoft/msquic/issues/4144

## Testing

No need to add new test case. The gtest "WithSendArgs2.SendLarge" in guic_gtest.cpp already covers this issue.

### DIFF
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -2257,6 +2257,7 @@ CxPlatSendDataFinalizeSendBuffer(
         IoVec->iov_base = SendData->ClientBuffer.Buffer;
         IoVec->iov_len = SendData->ClientBuffer.Length;
         if (SendData->TotalSize + SendData->SegmentSize > sizeof(SendData->Buffer) ||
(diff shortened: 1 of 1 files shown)


ProjectsByJackHe (commented):
Looks good to me! 

If it's practical, have you tried taking an msquic build from this PR and verifying your Samsung doesn't crash with large payload sizes?

Our CI automated tests all use the default github actions runners, which does not cover the issue you brought up I believe, correct me if I am wrong.

URNOTCharlotte:
> Looks good to me!
> 
> If it's practical, have you tried taking an msquic build from this PR and verifying your Samsung doesn't crash with large payload sizes?
> 
> Our CI automated tests all use the default github actions runners, which does not cover the issue you brought up I believe, correct me if I am wrong.

Yes, I have verified it on my phone.

nibanks (approved):

PR #4623, opened 2024-10-23
Locate `mc.exe` at configuration time

redbaron:
The `find_program()` command is now executed during the configuration phase to locate `mc.exe` in predefined locations. This approach allows the CMake toolchain to influence the search without modifying the `%PATH%`.

If `mc.exe` is not found, a clear error message is reported at configuration time, improving the ease of diagnosing build issues compared to discovering the failure in build logs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -449,11 +449,21 @@ if(WIN32)
     # Generate the MsQuicEtw header file.
     file(MAKE_DIRECTORY ${QUIC_BUILD_DIR}/inc)
 
+    if(NOT DEFINED CMAKE_FIND_USE_SYSTEM_ENVIRONMENT_PATH OR CMAKE_FIND_USE_SYSTEM_ENVIRONMENT_PATH)
+        find_program(MC_EXE NAMES mc.exe)
+        if(MC_EXE MATCHES "NOTFOUND")
+            # If not found, then VS project generator will set %PATH% env var to WinSDK bin directory
+            set(MC_EXE "mc.exe" CACHE STRING "mc.exe from %PATH%" FORCE)
+        endif()
+    else()
+        find_program(MC_EXE NAMES mc.exe REQUIRED)
+    endif()
+
     add_custom_command(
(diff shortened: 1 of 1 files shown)


redbaron:
Hm according to `find_program()` [docs](https://cmake.org/cmake/help/latest/command/find_program.html):

> 5.  Search the standard system environment variables. This can be skipped if NO_SYSTEM_ENVIRONMENT_PATH is passed or by setting the [CMAKE_FIND_USE_SYSTEM_ENVIRONMENT_PATH](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_USE_SYSTEM_ENVIRONMENT_PATH.html#variable:CMAKE_FIND_USE_SYSTEM_ENVIRONMENT_PATH) to FALSE.
>
> The directories in PATH itself.

Similar to  `add_custom_command` [behaviour](https://cmake.org/cmake/help/latest/command/add_custom_command.html):

> If neither of the above conditions are met, it is assumed that the command name is a program to be found on the PATH at build time.

Do you set `CMAKE_FIND_USE_SYSTEM_ENVIRONMENT_PATH` explicitly to `FALSE` somewhere ?

redbaron:
OK, it seems that VS Generator sets `%PATH%` at build time to point to  implicitly selected WinSDK bin dir, a.k.a. magic. Something which CMake can't see and do at configure time. 
 
Interestingly I made this fix to resolve opposite problem:  Ninja generator was failing to call `mc.exe` because although my toolchain sets [CMAKE_SYSTEM_PROGRAM_PATH](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROGRAM_PATH.html#variable:CMAKE_SYSTEM_PROGRAM_PATH) it wasn't respected by msquic build scripts.

I'll update patch to make `find_program()` required if `CMAKE_FIND_USE_SYSTEM_ENVIRONMENT_PATH=FALSE`  to respect the intent of that var, otherwise `find_program()` is optional. If optional and not found, it will try calling `mc.exe` as before, expecting VS generator do it's magic.

redbaron:
Linux test failure is unrelated

PR #4625, opened 2024-10-23
Delegate build environment configuration with QUIC_EXTERNAL_TOOLCHAIN

redbaron:
Well managed CMake build environments make use of CMAKE_TOOLCHAIN_FILE to configure build environment such as include/lib search paths, system libraries, build flags, etc.

This change introduces QUIC_EXTERNAL_TOOLCHAIN build option, which when enabled stops MSQuic CMake scripts from attempting to do CMake toolchain job.

QUIC_EXTERNAL_TOOLCHAIN defaults to OFF to preserve original behaviour.


src/platform/CMakeLists.txt
```diff
@@ -132,5 +132,8 @@ if(QUIC_TLS STREQUAL "openssl" OR QUIC_TLS STREQUAL "openssl3")
         target_link_libraries(platform PUBLIC "-framework CoreFoundation" "-framework Security")
     endif()
 elseif(QUIC_TLS STREQUAL "schannel")
-    target_link_libraries(platform PUBLIC secur32 onecore)
+    target_link_libraries(platform PUBLIC secur32)
```
nibanks, line 135: I think this change broke normal Windows builds:
```
platform.lib(tls_schannel.obj) : error LNK2019: unresolved external symbol _QueryContextAttributesExW@16 referenced in function _CxPlatTlsParamGet [D:\a\msquic\msquic\build\windows\x86_schannel\src\bin\msquic.vcxproj]
```

redbaron: reverted this part

redbaron: `onecore` must not be linked in GAMECORE builds because linking 2 umbrella libs (xgameplatform and onecore) is not supported. Changed it to special case gamecore builds

nibanks (approved):

PR #4626, opened 2024-10-24
Add quic_callback as a netperf dep

ProjectsByJackHe:
## Description

Changing `scripts/quic_callback` should trigger netperf

## Testing

CI

## Documentation

N/A

### DIFF
--- a/.github/workflows/netperf.yml
+++ b/.github/workflows/netperf.yml
@@ -27,6 +27,7 @@ on:
     - .gitmodules
     - scripts/secnetperf.ps1
     - scripts/secnetperf-helpers.psm1
+    - scripts/quic_callback.ps1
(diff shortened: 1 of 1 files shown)


.github/workflows/netperf.yml
```diff
@@ -27,6 +27,7 @@ on:
     - .gitmodules
     - scripts/secnetperf.ps1
     - scripts/secnetperf-helpers.psm1
+    - scripts/quic_callback.ps1
```
nibanks, line 30: Also needs to be updated above for `push`

ProjectsByJackHe: ACK, fixed in the next iteration.

nibanks (approved):

PR #4631, opened 2024-10-28
Kernel Assert Callbacks at PASSIVE

nibanks:
## Description

Assert that callbacks are happening at the correct IRQL level.

## Testing

CI/CD

## Documentation

N/A

### DIFF
--- a/src/core/listener.c
+++ b/src/core/listener.c
@@ -406,6 +406,7 @@ QuicListenerIndicateEvent(
     _Inout_ QUIC_LISTENER_EVENT* Event
     )
 {
+    CXPLAT_PASSIVE_CODE();
     CXPLAT_FRE_ASSERT(Listener->ClientCallbackHandler);
     return
         Listener->ClientCallbackHandler(
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -664,6 +664,7 @@ QuicConnIndicateEvent(
     _Inout_ QUIC_CONNECTION_EVENT* Event
     )
 {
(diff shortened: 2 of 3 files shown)


src/core/listener.c
```diff
@@ -406,6 +406,7 @@ QuicListenerIndicateEvent(
     _Inout_ QUIC_LISTENER_EVENT* Event
     )
 {
+    CXPLAT_PASSIVE_CODE();
```
mtfriesen, line 409: If you want to save a few bytes of memory, you could make each of these purely `CXPLAT_PASSIVE_CODE` functions pageable, so the linker puts them into a different code segment.

mtfriesen (approved):

src/core/listener.c
nibanks: What exactly does that get us? What does it cost? Does it mean the function can never raise the irql?

mtfriesen: Right, if the function ever executes at dispatch or higher, it cannot be pageable. You can also lock and unlock paged sections at runtime, e.g. if you go from zero to one connections.

csujedihy: Marking the function pageable won't reliably detect irql violations right? It would AV only when the pages are paged out?

mtfriesen: Right, driver verifier will occasionally page out pageable pages, but otherwise you'd have to rely on PAGED_CODE or PASSIVE_CODE asserts to detect IRQL violations. The benefit of making it pageable is primarily to be friendly to the memory manager.

PR #4632, opened 2024-10-28
Don't Share Previous IRQL for Dispatch RW Locks

nibanks:
## Description

Fixes an issue with kernel mode RW locks where they were sharing the previous IRQL state across calls, which could result in incorrect state on release.

## Testing

CI/CD

## Documentation

N/A

### DIFF
--- a/src/inc/quic_platform_winkernel.h
+++ b/src/inc/quic_platform_winkernel.h
@@ -314,17 +314,14 @@ typedef EX_PUSH_LOCK CXPLAT_RW_LOCK;
 #define CxPlatRwLockReleaseShared(Lock) ExReleasePushLockShared(Lock); KeLeaveCriticalRegion()
 #define CxPlatRwLockReleaseExclusive(Lock) ExReleasePushLockExclusive(Lock); KeLeaveCriticalRegion()
 
-typedef struct CXPLAT_DISPATCH_RW_LOCK {
-    EX_SPIN_LOCK SpinLock;
-    KIRQL PrevIrql;
-} CXPLAT_DISPATCH_RW_LOCK;
+typedef EX_SPIN_LOCK CXPLAT_DISPATCH_RW_LOCK;
 
-#define CxPlatDispatchRwLockInitialize(Lock) (Lock)->SpinLock = 0
+#define CxPlatDispatchRwLockInitialize(Lock) Lock = 0
 #define CxPlatDispatchRwLockUninitialize(Lock)
-#define CxPlatDispatchRwLockAcquireShared(Lock) (Lock)->PrevIrql = ExAcquireSpinLockShared(&(Lock)->SpinLock)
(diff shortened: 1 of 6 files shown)


src/inc/quic_platform_winkernel.h
```diff
@@ -314,17 +314,14 @@ typedef EX_PUSH_LOCK CXPLAT_RW_LOCK;
...
+typedef EX_SPIN_LOCK CXPLAT_DISPATCH_RW_LOCK;
 
-#define CxPlatDispatchRwLockInitialize(Lock) (Lock)->SpinLock = 0
+#define CxPlatDispatchRwLockInitialize(Lock) *(Lock) = 0
 #define CxPlatDispatchRwLockUninitialize(Lock)
-#define CxPlatDispatchRwLockAcquireShared(Lock) (Lock)->PrevIrql = ExAcquireSpinLockShared(&(Lock)->SpinLock)
-#define CxPlatDispatchRwLockAcquireExclusive(Lock) (Lock)->PrevIrql = ExAcquireSpinLockExclusive(&(Lock)->SpinLock)
-#define CxPlatDispatchRwLockReleaseShared(Lock) ExReleaseSpinLockShared(&(Lock)->SpinLock, (Lock)->PrevIrql)
-#define CxPlatDispatchRwLockReleaseExclusive(Lock) ExReleaseSpinLockExclusive(&(Lock)->SpinLock, (Lock)->PrevIrql)
+#define CxPlatDispatchRwLockAcquireShared(Lock) KIRQL RwLockPrevIrql = ExAcquireSpinLockShared(Lock)
```
csujedihy, line 321: This requires CxPlatDispatchRwLockAcquireShared to be called in the same scope as CxPlatDispatchRwLockReleaseShared? Is that good?

csujedihy: Can we change the code to not acquire the shared locked if the caller already holds the exclusive lock? Like adding a param to QuicLookupFindConnectionByLocalCidInternal?

mtfriesen: I feel like these macros with hidden variables are intrinsically dangerous. Why not just make up a cross-platform lock handle?

mtfriesen: E.g. `CxPlatDispatchRwLockAcquireShared(CXPLAT_LOCK *Lock, CXPLAT_LOCK_HANDLE *LockHandle);`

nibanks: I am trying to not add extra stuff that isn't needed on any other platform.

mtfriesen: IMO, let the compiler or preprocessor optimize away unused variables, rather than hide them in macros. As a programmer, I would not have expected CxPlatDispatchRwLockAcquireShared to declare a hidden variable.

csujedihy (dismissed):

csujedihy (approved):

PR #4639, opened 2024-10-30
Check lttng log

ami-GS:
DO NOT MERGE

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -157,6 +157,9 @@ jobs:
         $ManifestPath = ".\src\manifest\MsQuicEtw.man"
         wevtutil.exe um $ManifestPath
         wevtutil.exe im $ManifestPath /rf:$($MsQuicDll) /mf:$($MsQuicDll)
+    - name: Stop lttng-sessiond
+      if: matrix.vec.plat == 'linux'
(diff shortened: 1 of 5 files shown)


ami-GS:
lttng seems to be running, but data is not collected.

image

ami-GS:
lttng process is running.
image

ami-GS:
lttng is internally active. write permission?

image

ami-GS:
Moving log tmp directory from $HOME to artifacts doesn't change

image

ami-GS:
UST application registration is not executed appropriately

left is github action. right is local VM

image

nibanks:
So, what are the next steps here?

ami-GS:
I realized that ubuntu 24.04 + xdp workload successfully generate the logs. This means sudo related config can generate file. (I investigated file permissions htough)
I will investigate the difference.

ami-GS:
hmm, Ubuntu 24.04 seems taking another trace.
image

ami-GS:
Interesting point.

Ubuntu 20.04 has dotnet 6,7,8, but Ubuntu 24.04 has only dotnet 8. (both uses 8 though)
Ubuntu 24.04 can generate lttng log, but failed to call clog2text_lttng

Ubuntu 20.04
image
Ubuntu 24.04
image

PR #4640, opened 2024-10-31
Update Execution Docs

nibanks:
## Description

This revamps our documentation on MsQuic's execution mode, including how we manage callbacks and threads.

## Testing

N/A

## Documentation

Updated

### DIFF
--- a/docs/Execution.md
+++ b/docs/Execution.md
@@ -32,16 +32,16 @@ QUIC_STATUS
 Above is an example of a callback delivered to the listener interface.
 The application must register a callback handler to manage all events MsQuic may indicate, returning a status to show if it was successfully handled or not.
 
-This approach differs significantly from BSD sockets, where the application must make a call (e.g., `send` or `recv`) to determine if something happened.
+This approach differs significantly from sockets and most networking libraries, where the application must make a call (e.g., `send` or `recv`) to determine if something happened.
 This design choice was made for several reasons:
 
-- The MsQuic API **runs in-process**, eliminating the need for a kernel to user mode boundary switch to notify the application layer. This makes the callback-based design more practical compared to BSD sockets.
+- The MsQuic API **runs in-process**, eliminating the need for a kernel to user mode boundary switch to notify the application layer. This makes the callback-based design more practical compared to sockets.
 
 - MsQuic, due to the QUIC protocol, has numerous event types. Applications may have hundreds of objects with potential state changes. The callback model allows the application to avoid managing pending calls on each object.
 
-- Writing correct, performant code on top of the BSD-style interface has proven challenging. Callbacks, executed at the correct time and on the correct thread/processor, enable MsQuic to abstract much complexity from applications, making things "just work" out of the box.
+- Writing correct, scalable  code on top of the socket interfaces has proven challenging. Callbacks, executed at the correct time and on the correct thread/processor, enable MsQuic to abstract much complexity from applications, making things "just work" out of the box.
 
-- It simplifies MsQuic's logic by eliminating the need for a queue or cached state to indicate to the application. In the BSD model, the networking stack must wait for a top-down call from the application before indicating completion, increasing code size, complexity, and memory usage.
+- It simplifies MsQuic's logic by eliminating the need for a queue or cached state to indicate to the application. In the socket model, the networking stack must wait for a top-down call from the application before indicating completion, increasing code size, complexity, and memory usage.
 
 ### Writing Event Handlers
 
--- a/docs/Execution.md
+++ b/docs/Execution.md
@@ -6,7 +6,7 @@ The sections below detail the designs MsQuic uses and the reasons behind these c
 
 ## Event Model
 
-In the MsQuic API, all state changes and notifications are indicated directly to the application via a callback.
+In the MsQuic API, all asynchronous state changes and notifications are indicated directly to the application via a callback.
 This includes connection state changes, new streams being created, stream data being received, and stream sends completing.
 
 ```c
--- a/docs/Execution.md
+++ b/docs/Execution.md
@@ -1,7 +1,7 @@
 Execution
 ======
 
-The MsQuic API takes a very difference stance when it comes to its execution model compared to BSD-style sockets (and most other networking libraries built on top of them).
+The MsQuic API takes a very different stance when it comes to its execution model compared to BSD-style sockets (and most other networking libraries built on top of them).
 The sections below detail the designs MsQuic uses, with some of the details as to why these design choices were made.
 
 ## Event Model
@@ -30,14 +30,14 @@ QUIC_STATUS
 ```
 
 Above is an example of the type of callback delivered to the listener interface.
-The application is requires to register a callback handler that should handle all the events MsQuic may indicate, returning a status for if it was successfully handled or not.
+The application is required to register a callback handler that should handle all the events MsQuic may indicate, returning a status for if it was successfully handled or not.
 
 This is very different from BSD sockets which required the application to make a call (e.g., `send` or `recv`) in order to determine something happened.
 This difference was made for several reasons:
 
 - The MsQuic API **runs in-proc**, and therefore doesn't require a kernel to user mode boundary switch to indicate something to the application layer. This allows for the callback-based design which is not as practical for BSD sockets.
 
-- MsQuic, by virtue of the QUIC protocol itself, has a lot of different types of events. Just considering streams, the app maybe have hundreds of objects at once which may have some state change. By leveraging the callback model, the application doesn't have to manage having pending calls on each object.
+- MsQuic, by virtue of the QUIC protocol itself, has a lot of different types of events. Just considering streams, the app may have hundreds of objects at once which may have some state change. By leveraging the callback model, the application doesn't have to manage having pending calls on each object.
 
 - Experience has shown it to be very difficult to write correct, performant code on top of the BSD-style interface. By leveraging callbacks (that happen at the correct time, on the correct thread/processor), it allows MsQuic to abstract a lot of complexity away from applications and make things "just work" out of the box.
 
@@ -49,7 +49,7 @@ Event handlers are **required** for all objects (that have them), because much o
 Additionally, important events, such as "shutdown complete" events provide crucial information to the application to function properly.
 Without these events, the application cannot not know when it is safe to clean up objects.
 
-Applicationss are expected to keep any execution time in the callbacks **to a minimum**.
+Applications are expected to keep any execution time in the callbacks **to a minimum**.
 MsQuic does not use separate threads for the protocol execution and upcalls to the application.
 Therefore, any significant delays on the callback **will delay the protocol**.
 Any significant time or work needed to be completed by the application must happen on its own thread.
@@ -58,16 +58,16 @@ This doesn't mean the application isn't allowed to do any work in the callback h
 In fact, many things are expressly designed to be most efficient when the application does them on the callback.
 For instance, closing a handle to a connection or stream is ideally implemented in the "shutdown complete" indications.
 
-One important aspect of this design is that all blocking API (down) calls invoked on a callback always happen inline (to prevent deadlocks), and will supercede any calls in progress or queued from a separate thread.
+One important aspect of this design is that all blocking API (down) calls invoked on a callback always happen inline (to prevent deadlocks), and will supersede any calls in progress or queued from a separate thread.
 
 ## Threading
 
-By default, MsQuic creates its own threads to manage execution its logic.
+By default, MsQuic creates its own threads to manage the execution of its logic.
 The nature of the number and configuration of these threads depends on the configuration the apps passes to [RegistrationOpen](api/RegistrationOpen.md) or `QUIC_PARAM_GLOBAL_EXECUTION_CONFIG`.
 
 The default behavior is to create dedicated, per-processor threads that are hard affinitized to a given NUMA-node, and soft-affinitized (set 'ideal processor') to a given processor.
 These threads are then used to drive both the datapath (i.e. UDP) and QUIC layers.
-Great care it taken to (try to) align the MsQuic processing logic with the rest of the networking stack (including hardware RSS) so that all processing stays at least on the same NUMA node, but ideally the same processor.
+Great care is taken to (try to) align the MsQuic processing logic with the rest of the networking stack (including hardware RSS) so that all processing stays at least on the same NUMA node, but ideally the same processor.
 
 The complexity required to achieve alignment in processing across various threads and processors is why MsQuic takes the stance of managing all its own threading by default.
 The goal is to abstract all this complexity from the many applications that build on top, so every app doesn't have to build the necessary logic to do this itself.
--- a/docs/Execution.md
+++ b/docs/Execution.md
@@ -82,31 +82,26 @@ For listeners, the application callback will be called in parallel for new conne
 
 ```mermaid
 graph TD
-    subgraph NIC
-        RSS1
-        RSS2
-        RSS3
+    subgraph Kernel
+        NIC-Queue1[NIC Queue]
+        NIC-Queue2[NIC Queue]
+        NIC-Queue1 -->|RSS Receive| UDP1[IP/UDP]
+        NIC-Queue2 -->|RSS Receive| UDP2[IP/UDP]
     end
-    RSS1 -->|Receive| Processor1
-    RSS2 -->|Receive| Processor2
-    RSS3 -->|Receive| Processor3
-    subgraph Processor1
-        Thread1
-        Thread1 -->|Manages| Connection1
-        Thread1 -->|Manages| Connection2
-        Connection1 -->|Delivers Event| ApplicationCallback1
-        Connection2 -->|Delivers Event| ApplicationCallback2
-    end
-    subgraph Processor2
-        Thread2
-        Thread2 -->|Manages| Connection3
-        Connection3 -->|Delivers Event| ApplicationCallback3
-    end
-    subgraph Processor3
-        Thread3
-        Thread3 -->|Manages| Connection4
-        Thread3 -->|Manages| Connection5
-        Connection4 -->|Delivers Event| ApplicationCallback4
-        Connection5 -->|Delivers Event| ApplicationCallback5
+    subgraph MsQuic Process
+        UDP1 -.-> Processor1
+        UDP2 -.-> Processor2
+        subgraph Processor1[Processor 0]
+            Thread1[Thread]
+            Thread1 -->|Manages| Connection1[Connection 1]
+            Thread1 -->|Manages| Connection2[Connection 2]
+            Connection1 -->|Delivers Event| ApplicationCallback1[App Callback]
+            Connection2 -->|Delivers Event| ApplicationCallback2[App Callback]
+        end
+        subgraph Processor2[Processor 1]
+            Thread2[Thread]
+            Thread2 -->|Manages| Connection3[Connection 3]
+            Connection3 -->|Delivers Event| ApplicationCallback3[App Callback]
+        end
     end
 ```
--- a/docs/Execution.md
+++ b/docs/Execution.md
@@ -29,55 +29,53 @@ QUIC_STATUS
     );
 ```
 
-Above is an example of the type of callback delivered to the listener interface.
-The application is required to register a callback handler that should handle all the events MsQuic may indicate, returning a status for if it was successfully handled or not.
+Above is an example of a callback delivered to the listener interface.
+The application must register a callback handler to manage all events MsQuic may indicate, returning a status to show if it was successfully handled or not.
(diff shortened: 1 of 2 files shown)


docs/Execution.md
```diff
@@ -0,0 +1,107 @@
...
+    _In_ HQUIC Listener,
+    _In_opt_ void* Context,
+    _Inout_ QUIC_LISTENER_EVENT* Event
+    );
+```
+
+Above is an example of a callback delivered to the listener interface.
+The application must register a callback handler to manage all events MsQuic may indicate, returning a status to show if it was successfully handled or not.
+
+This approach differs significantly from BSD sockets, where the application must make a call (e.g., `send` or `recv`) to determine if something happened.
```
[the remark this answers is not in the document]
keith-horton: it would also be useful to compare this to Winsock OVERLAPPED I/O APIs. since this seems like it's a combination of "OVERLAPPED I/O semantics + select/event semantics". e.g., when I connect my OVERLAPPED I/O request to an NT Threadpool, or drain completions manually from an IO completion port, things start to look very similar.

It *sounds* so far like the difference is that QUIC gives these callback invocations on state changes as well - not just posting IO -- which to me sounds much more like the select/event model, when the app is notified when the underlying socket has changed some internal state ... (e.g. "read to send data now" state, or "ready to receive data now" state, or "your connect request completed" state, or "you have new inbound connections ready to accept" state, or "your socket is in a failed state which we RST" state)

docs/Execution.md
```diff
@@ -0,0 +1,107 @@
...
+This design choice was made for several reasons:
+
+- The MsQuic API **runs in-process**, eliminating the need for a kernel to user mode boundary switch to notify the application layer. This makes the callback-based design more practical compared to BSD sockets.
+
+- MsQuic, due to the QUIC protocol, has numerous event types. Applications may have hundreds of objects with potential state changes. The callback model allows the application to avoid managing pending calls on each object.
+
+- Writing correct, performant code on top of the BSD-style interface has proven challenging. Callbacks, executed at the correct time and on the correct thread/processor, enable MsQuic to abstract much complexity from applications, making things "just work" out of the box.
```
keith-horton, line 42: I'm not a big fan of this statement. one can write correct code using select, and on non-Windows operating systems, select performs well enough for many (most?) apps in the past. it's just Windows did not optimize the non-blocking IO path like *nix OS's did.

This statement also sounds non-convincing:  'making things "just work" out of the box". When I read through this, I don't see it "just work" - I just see my business logic moving from select() loops to callbacks.

That change *can* be a good thing - it avoids burning threads when one must poll with select calls. so I could see the argument made that these callbacks *scale* much more effectively because one doesn't need to manually poll against X number of sockets on Y number of threads -- scaling *out* to large numbers of sockets can indeed be problematic.

This is where OVERLAPPED I/O helps scale better than non-blocking I/O - with very large #s of sockets.

It would be useful to differentiate:
how does this callback model change from the non-blocking model for the benefit of the developer?
IMHO, it helps by (a) scaling out naturally without the developer creating their own threadpool and eventing workflow (e.g., avoiding thundering herd problems).

How does this callback model change from the OVERLAPPED I/O model for the benefit of the developer?
IMHO, it helps by not requiring posting an I/O to just receive state change notifications; it helps by not requiring the developer to post I/O up-front, which has a scale-out problem (if I have 100k connections, I don't want to have to pend 100K buffer allocations - I just want to know when a connection is ready, so I can post a send/recv with my buffer when the socket is ready to handle the IO)

docs/Execution.md
```diff
@@ -0,0 +1,107 @@
...
+Applications should keep the execution time within callbacks **to a minimum**.
+MsQuic does not use separate threads for protocol execution and upcalls to the application.
+Therefore, any significant delays in the callback **will delay the protocol**.
+Any substantial work required by the application should be performed on its own thread.
+
+This does not mean the application cannot perform any work in the callback handler.
+In fact, many operations are designed to be most efficient when executed within the callback.
+For example, closing a handle to a connection or stream is ideally done during the "shutdown complete" indication.
+
+A crucial aspect of this design is that all blocking API (down) calls invoked within a callback always occur inline (to prevent deadlocks) and will take precedence over any calls in progress or queued from a separate thread.
```
keith-horton, line 61: so an application must be very careful if they are doing some work inline in their callback, and some work deferred in their own separate threads? that sounds difficult to accomplish without locks involved to guarantee predictable order-of-execution. and it sounds like acquiring locks in the callback is going to be a bad idea - as that could cause effectively random latency jumps on a connection :(

So is the model more often, "Pick one: handle inline, or defer work to your own threads; if you do both, you'll need to provide your own synchronization since any calls inline will be preferred over whatever your own threads are doing"?

keith-horton: even synchronizing won't help the app track exactly what is happening in what order, since even if I synchronize "I'm calling QUIC function X from my own thread; now my lock unblocks the callback thread and I compare state and make a new call from my callback -- oops, my call from my callback just took effect before my prior call from my own thread!"

docs/Execution.md
```diff
@@ -0,0 +1,107 @@
...
+
+Typically, MsQuic creates dedicated threads for each processor, which are hard-affinitized to a specific NUMA node and soft-affinitized (set as 'ideal processor') to a specific processor.
+These threads handle both the datapath (i.e., UDP) and QUIC layers.
+MsQuic aligns its processing logic with the rest of the networking stack (including hardware RSS) to ensure that all processing stays on the same NUMA node, and ideally, the same processor.
+
+The complexity of aligning processing across various threads and processors is why MsQuic manages its own threading by default.
+This abstraction simplifies the development process for applications built on top of MsQuic, ensuring that things "just work" out of the box.
+
+Each thread manages the execution of one or more connections.
+Connections are distributed across threads based on their RSS alignment, which should evenly distribute traffic based on different UDP tuples.
```
keith-horton, line 76: except RSS hashing isn't always evenly distributed :(
so maybe it's more correct to remove "which should be evenly distributed"

docs/Execution.md
```diff
@@ -0,0 +1,107 @@
...
+Therefore, any significant delays in the callback **will delay the protocol**.
+Any substantial work required by the application should be performed on its own thread.
+
+This does not mean the application cannot perform any work in the callback handler.
+In fact, many operations are designed to be most efficient when executed within the callback.
+For example, closing a handle to a connection or stream is ideally done during the "shutdown complete" indication.
+
+A crucial aspect of this design is that all blocking API (down) calls invoked within a callback always occur inline (to prevent deadlocks) and will take precedence over any calls in progress or queued from a separate thread.
+
+## Threading
```
keith-horton, line 63: this section is very well written.

docs/Execution.md
nibanks: Well, we already say "which **should** evenly distribute traffic". 😄 In the future, we plan to add QUIC-based RSS which will allow us explicit control over this, so we will be able to guarantee it.

docs/Execution.md
```diff
@@ -0,0 +1,107 @@
...
+This design choice was made for several reasons:
+
+- The MsQuic API **runs in-process**, eliminating the need for a kernel to user mode boundary switch to notify the application layer. This makes the callback-based design more practical compared to BSD sockets.
+
+- MsQuic, due to the QUIC protocol, has numerous event types. Applications may have hundreds of objects with potential state changes. The callback model allows the application to avoid managing pending calls on each object.
+
+- Writing correct, performant code on top of the BSD-style interface has proven challenging. Callbacks, executed at the correct time and on the correct thread/processor, enable MsQuic to abstract much complexity from applications, making things "just work" out of the box.
+
+- It simplifies MsQuic's logic by eliminating the need for a queue or cached state to indicate to the application. In the BSD model, the networking stack must wait for a top-down call from the application before indicating completion, increasing code size, complexity, and memory usage.
```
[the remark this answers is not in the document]
nibanks: An event happens directly inline to our processing. For instance, if we parse a packet and it indicates a new QUIC stream has been created, before we process any further in that packet, we make that upcall to the app telling them about the new stream. So, there is no extra queuing logic required.

docs/Execution.md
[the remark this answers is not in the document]
nibanks: QUIC is fundamentally different. You set up your object, be it a listener, connection or stream, up front with a callback and your context pointer **once**. Then all events happen for that object for the rest of its lifetime.

docs/Execution.md
nibanks: Perhaps we can agree to disagree? 😆 

I have been working on our socket abstraction for all the platforms for over half a decade and I'd argue it's still not perfect (on any platform). So, IMHO, it is not possible to get sockets working out of the box. Even with the years of dedicated networking experience, and the fact that we **own** the Windows layers, we still struggle to get it right. We also get near constant internal partner questions/investigations that lead me towards the conclusion that others have similar problems.

That being said, I'm totally fine with making text changes here.

docs/Execution.md
nibanks: Locks by themselves are not a bad thing. I believe most production code on top of MsQuic uses locks just fine in the callbacks today. Locks that you hold or contend a lot are bad. But that's true in the general sense; not just in our callbacks.

The only really bad 'gotcha' we have is if you try to make a blocking call on Connection B in the callback of Connection A. This can results in a deadlock and is not super obvious to the developer at first glance.

So, generally, my recommendation is to run code in the callback if it's purely computational (i.e. no I/O). But anything that might block on some other component or take a really long time, you should queue a worker.

docs/Execution.md
```diff
@@ -0,0 +1,107 @@
+Execution
+======
+
+The MsQuic API uses a different execution model compared to BSD-style sockets and most other networking libraries built on them.
+The sections below detail the designs MsQuic uses and the reasons behind these choices.
+
+## Event Model
+
+In the MsQuic API, all state changes and notifications are indicated directly to the application via a callback.
```
mtfriesen, line 9: all "asynchronous" state changes? presumably some APIs are synchronous, so QuicBind() could return QUIC_SUCCESS and implicitly transition state?

docs/Execution.md
```diff
@@ -0,0 +1,107 @@
...
+QUIC_STATUS
+(QUIC_API QUIC_LISTENER_CALLBACK)(
+    _In_ HQUIC Listener,
+    _In_opt_ void* Context,
+    _Inout_ QUIC_LISTENER_EVENT* Event
+    );
+```
+
+Above is an example of a callback delivered to the listener interface.
+The application must register a callback handler to manage all events MsQuic may indicate, returning a status to show if it was successfully handled or not.
```
mtfriesen, line 33: nit: 'must register a callback handler for each type of event MsQuic may indicate" since this seems to suggest a single handler for all events.

docs/Execution.md
nibanks: I cannot think of any blocking APIs beyond getting/setting parameters (i.e. socket options). Everything is async. Your notion of a 'bind' is simply setting a 'local address' parameter before calling the `ConnectionStart` async function, which then will attempt to use that local address.

docs/Execution.md
```diff
@@ -0,0 +1,107 @@
...
+
+This approach differs significantly from sockets and most networking libraries, where the application must make a call (e.g., `send` or `recv`) to determine if something happened.
+This design choice was made for several reasons:
+
+- The MsQuic API **runs in-process**, eliminating the need for a kernel to user mode boundary switch to notify the application layer. This makes the callback-based design more practical compared to sockets.
```
mtfriesen, line 38: i don't know if you want to mention traditional Windows user APCs here, which do provide user mode callbacks for arbitrary events, including kernel IO completion. they did come at huge costs, though, which is why nobody uses them.

docs/Execution.md
```diff
@@ -0,0 +1,107 @@
...
+
+In the MsQuic API, all state changes and notifications are indicated directly to the application via a callback.
+This includes connection state changes, new streams being created, stream data being received, and stream sends completing.
+
+```c
+typedef struct QUIC_LISTENER_EVENT {
+    QUIC_LISTENER_EVENT_TYPE Type;
+    union {
+        struct { ... } NEW_CONNECTION;
+        struct { ... } STOP_COMPLETE;
```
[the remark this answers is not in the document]
nibanks: I don't care either way, but today, there are only two structs in there. Connection and Stream have a lot more.

docs/Execution.md
nibanks: I don't know them, so I'm definitely not qualified to write documentation about them. Unless you have something you think I should add, I'd lean towards not saying anything.

docs/Execution.md
```diff
@@ -0,0 +1,107 @@
...
+The application must register a callback handler to manage all events MsQuic may indicate, returning a status to show if it was successfully handled or not.
+
+This approach differs significantly from sockets and most networking libraries, where the application must make a call (e.g., `send` or `recv`) to determine if something happened.
+This design choice was made for several reasons:
+
+- The MsQuic API **runs in-process**, eliminating the need for a kernel to user mode boundary switch to notify the application layer. This makes the callback-based design more practical compared to sockets.
+
+- MsQuic, due to the QUIC protocol, has numerous event types. Applications may have hundreds of objects with potential state changes. The callback model allows the application to avoid managing pending calls on each object.
+
+- Writing correct, scalable  code on top of the socket interfaces has proven challenging. Callbacks, executed at the correct time and on the correct thread/processor, enable MsQuic to abstract much complexity from applications, making things "just work" out of the box.
```
mtfriesen, line 42: "executed at the correct time and on the correct thread/processor" is probably too strong a statement, since "correct" from the application's perspective is subjective, not objective. i'd reword to say that this offloads some of the problem, and allows the MsQuic library to take responsibility for optimizing scheduling.

docs/Execution.md
```diff
@@ -0,0 +1,107 @@
...
+Applications should keep the execution time within callbacks **to a minimum**.
+MsQuic does not use separate threads for protocol execution and upcalls to the application.
+Therefore, any significant delays in the callback **will delay the protocol**.
+Any substantial work required by the application should be performed on its own thread.
+
+This does not mean the application cannot perform any work in the callback handler.
+In fact, many operations are designed to be most efficient when executed within the callback.
+For example, closing a handle to a connection or stream is ideally done during the "shutdown complete" indication.
+
+A crucial aspect of this design is that all blocking API (down) calls invoked within a callback always occur inline (to prevent deadlocks) and will take precedence over any calls in progress or queued from a separate thread.
```
mtfriesen, line 61: does QUIC make any promises about recursive callbacks? e.g. if an app calls QuicDoFoo() from within the first MsQuic callback, could MsQuic invoke a FooComplete() callback? if not, does this imply MsQuic in fact builds a queue of notifications, contrary to the point above?

nibanks: MsQuic can recursively invoke callbacks in some limited cases, but only if the app opts in. But most of the time, there is no need.

docs/Execution.md
nibanks: Made a small change

docs/Execution.md
```diff
@@ -0,0 +1,107 @@
...
+
+By default, MsQuic creates its own threads to manage the execution of its logic.
+The number and configuration of these threads depend on the settings passed to [RegistrationOpen](api/RegistrationOpen.md) or `QUIC_PARAM_GLOBAL_EXECUTION_CONFIG`.
+
+Typically, MsQuic creates dedicated threads for each processor, which are hard-affinitized to a specific NUMA node and soft-affinitized (set as 'ideal processor') to a specific processor.
+These threads handle both the datapath (i.e., UDP) and QUIC layers.
+MsQuic aligns its processing logic with the rest of the networking stack (including hardware RSS) to ensure that all processing stays on the same NUMA node, and ideally, the same processor.
+
+The complexity of aligning processing across various threads and processors is why MsQuic manages its own threading by default.
+This abstraction simplifies the development process for applications built on top of MsQuic, ensuring that things "just work" out of the box.
```
mtfriesen, line 73: i think we need to call out somewhere that QUIC is optimizing these threads for network IO. for apps where QUIC is a minority of CPU time, this probably doesn't "just work" out of the box.

docs/Execution.md
```diff
@@ -0,0 +1,107 @@
...
+
+A crucial aspect of this design is that all blocking API (down) calls invoked within a callback always occur inline (to prevent deadlocks) and will take precedence over any calls in progress or queued from a separate thread.
+
+## Threading
+
+By default, MsQuic creates its own threads to manage the execution of its logic.
+The number and configuration of these threads depend on the settings passed to [RegistrationOpen](api/RegistrationOpen.md) or `QUIC_PARAM_GLOBAL_EXECUTION_CONFIG`.
+
+Typically, MsQuic creates dedicated threads for each processor, which are hard-affinitized to a specific NUMA node and soft-affinitized (set as 'ideal processor') to a specific processor.
+These threads handle both the datapath (i.e., UDP) and QUIC layers.
```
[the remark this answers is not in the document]
nibanks: The app can configure QUIC to execute in that model, but it's not the default.

docs/Execution.md
mtfriesen: should that option (and default behavior) be described here? this is a common question for any code interacting with callbacks, because it can significantly affect implementations, e.g., can i acquire a spinlock within a callback?

docs/Execution.md
mtfriesen: i think they're interesting from a [mis]design perspective, but i agree they _should_ be irrelevant in practice.

docs/Execution.md
mtfriesen: >must register a callback handler for each type of event MsQuic may indicate

☝️  this was my wording suggestion

docs/Execution.md
nibanks: Sure. I will add a line about recursive callback behavior.

docs/Execution.md
mtfriesen: is this saying that there is a per-connection callback, and a per-stream callback, and a per-listener callback, etc., and then all events for that object type go through that callback? if so, i'd just write that.

docs/Execution.md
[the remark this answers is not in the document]
mtfriesen: ok, i think anyone can figure that out.

docs/Execution.md
nibanks: There is a per-object callback. Each callback handles all the types of events for that object. I will try to improve the text to make that clearer.

docs/Execution.md
mtfriesen: ok, maybe i'm splitting hairs here, but i still suggest that "asynchronous" be added here, because calling `ConnectionStart` implicitly changes the state of a connection from "doesn't exist" or "not started" etc., to "starting" from the perspective of an application. just because an explicit notification wasn't created for that state change doesn't make it less of a conceptual state change, imo.

docs/Execution.md
mtfriesen: i'd add two words, so that things "just work" **for QUIC** out of the box.

docs/Execution.md
nibanks: Added.

docs/Execution.md
nibanks: Will do

docs/Execution.md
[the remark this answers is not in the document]
mtfriesen: OK. the word "typically" is doing a lot of work here, and IMO that execution model option is in scope here.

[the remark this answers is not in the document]
nibanks: To be fair, it says 'thread**s** for each processor' and then goes on to say these handle both layer's events. Nothing is explicitly saying **one** thread for one processor and that thread always processes both layers.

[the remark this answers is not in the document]
mtfriesen: Yeah, I agree, though I think this is more difficult to grasp / less clear than ideal. QUIC has gone to tremendous effort to be able to unify data path + QUIC in some configurations, e.g. for polling, and IMO the option to both unify and de-unify the threads is a significant feature.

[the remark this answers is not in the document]
nibanks: Added

docs/Execution.md
```diff
@@ -0,0 +1,109 @@
...
+QUIC_STATUS
+(QUIC_API QUIC_LISTENER_CALLBACK)(
+    _In_ HQUIC Listener,
+    _In_opt_ void* Context,
+    _Inout_ QUIC_LISTENER_EVENT* Event
+    );
+```
+
+Above is an example of a callback delivered to the listener interface.
+The application must register a per-object callback handler to manage all events MsQuic may indicate for that object, returning a status to show if it was successfully handled or not.
```
ami-GS, line 33: How about emphasizing "must" like **MUST** used in RFC? **MUST MAY REQUIRD** etc.

ami-GS: OK!

mtfriesen (approved):

PR #4646, opened 2024-11-05
Improve version deployment documentation

anrossi:
## Description

Improve the version deployment documentation.

### DIFF
--- a/docs/Settings.md
+++ b/docs/Settings.md
@@ -79,7 +79,7 @@ The following settings are available via registry as well as via [QUIC_VERSION_S
 | Acceptable Versions List          | uint32_t[] | AcceptableVersions           | Unset             | Sets the list of versions that a given server instance will use if a client sends a first flight using them. |
 | Offered Versions List             | uint32_t[] | OfferedVersions              | Unset             | Sets the list of versions that a given server instance will send in a Version Negotiation packet if it receives a first flight from an unknown version. This list will most often be equal to the Acceptable Versions list. |
 | Fully-Deployed Versions List      | uint32_t[] | FullyDeployedVersions        | Unset             | Sets the list of QUIC versions that is supported and negotiated by every single QUIC server instance in this deployment. Used to generate the AvailableVersions list in the Version Negotiation Extension Transport Parameter. |
-| Version Negotiation Ext. Enabled  | uint32_t   | VersionNegotiationExtEnabled | 0 (FALSE)         | Enables the Version Negotiation Extension                                                                                     |
+| Version Negotiation Ext. Enabled  | uint32_t   | VersionNegotiationExtEnabled | 0 (FALSE)         | Enables the Version Negotiation Extension. |
 
 The `uint32_t[]` type is a `REG_BINARY` blob of the versions list, with each version in little-endian format.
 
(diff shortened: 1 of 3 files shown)


docs/Settings.md
```diff
@@ -74,11 +74,12 @@ While `REG_DWORD` can hold values larger than `uint16_t`, the administrator shou
...
-| Acceptable Versions List          | uint32_t[] | AcceptableVersions       | Unset             | Sets the list of versions that a given server instance will use if a client sends a first flight using them. |
-| Offered Versions List             | uint32_t[] | OfferedVersions          | Unset             | Sets the list of versions that a given server instance will send in a Version Negotiation packet if it receives a first flight from an unknown version. This list will most often be equal to the Acceptable Versions list. |
-| Fully-Deployed Versions List      | uint32_t[] | FullyDeployedVersions    | Unset             | Sets the list of QUIC versions that is supported and negotiated by every single QUIC server instance in this deployment. Used to generate the AvailableVersions list in the Version Negotiation Extension Transport Parameter. |
+| Setting                           | Type       | Registry Name                | Default           | Description                                                                                                                   |
```
nibanks, line 77: nit: Don't change it now, but in the future, there's no need to make all the `|` line up, as all it really does is make the diff bigger.

docs/Versions.md
```diff
@@ -92,9 +92,30 @@ MsQuic->SetParam(
     &Settings);
 ```
 
+### Configuring Versions via Windows Registry
+MsQuic supports setting the Acceptable Versions, Offered Versions, and Fully-Deployed Versions lists via the Windows registry. These settings are global for all servers and clients on the machine.
+The registry settings are overridden by settings specified in the code.
+The registry values must be created under the HKLM\System\CurrentControlSet\Services\MsQuic\Parameters key.
```
nibanks, line 98:
```suggestion
The registry values must be created under the `HKLM\System\CurrentControlSet\Services\MsQuic\Parameters` key.
```

docs/Versions.md
```diff
@@ -92,9 +92,30 @@ MsQuic->SetParam(
...
+Each list is stored in the registry as a `REG_BINARY` type, with the version numbers in little-endian (host) order.
+The registry value for Acceptable Versions must be named "AcceptableVersions".
+The registry value for Offered Versions must be named "OfferedVersions".
+The registry value for Fully-Deployed Versions must be named "FullyDeployedVersions".
```
nibanks, line 102:
```suggestion
The registry value for Acceptable Versions must be named `AcceptableVersions`.
The registry value for Offered Versions must be named `OfferedVersions`.
The registry value for Fully-Deployed Versions must be named `FullyDeployedVersions`.
```

docs/Versions.md
```diff
@@ -92,9 +92,30 @@ MsQuic->SetParam(
...
+
+Here's a sample .reg file that creates all three lists with QUIC version 2 first and QUIC version 1 after, in little endian order, and enables version negotiation:
+```reg
+Windows Registry Editor Version 5.00
+
+[HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\MsQuic\Parameters]
+"AcceptableVersions"=hex:cf,43,33,6b,01,00,00,00
```
nibanks, line 109: I know it doesn't work this way today, but I think it's worth updating the settings code (in a future PR) to handle a more simpler configuration model, where only one of these _has_ to be implemented; including the `Enabled` one. Thoughts?

docs/Settings.md
anrossi: I agree, so I fixed the new lines at least. we can do a PR later to change the rest if desired.

docs/Versions.md
anrossi: I think the risk of a single-list is that a deployment is made using the single-list model, and then we add QUIC 3 support, and the deployment either does the version rollout wrong, or the infrastructure isn't built to do the rollout, so they have to spend more time updating infrastructure and testing it to even do a rollout.

anrossi: I do think that setting any version list via registry should enable the version negotiation extension without the extra registry value.

nibanks (approved):

PR #4654, opened 2024-11-13
Windows Kernel XDP Support

ami-GS:
## Description

built with placeholders

## Testing

N/A

## Documentation

N/A


ProjectsByJackHe:
Before merging this PR, let's make sure Netperf doesn't break.

ami-GS:
Tests succeed, but trace says there are many `XDP async IO start`, but there is no completion at all.

ami-GS:
Maybe `CxPlatXdpExecute` should call `XskNotifyAsync`, not `ZwSetInformationFile`. However doing it gets error as 

image

ami-GS:
Sometimes passes with few `XDP async IO start` calls

ami-GS:
> Maybe `CxPlatXdpExecute` should call `XskNotifyAsync`, not `ZwSetInformationFile`. However doing it gets error as
> 
> image

This was due to handling kernel mode error as user mode error. both are for PENDING

PR #4657, opened 2024-11-15
Remove hardcoded endpoint

ami-GS:
## Description

Removed hardcoded endpoint ("google.com")
Reported by https://vnext.s360.msftcloudes.com/blades/quality/reliability?blade=AssignedTo:All~KPI:510f681d-1be2-4568-a686-85d1fc0e5006~SLA:0~Forums:All~KPI%20Ranking:All~waves:All~Tab:Summary~_loc:Reliability&peopleBasedNodes=nibanks_team&global=4:b7b51c2f-738e-43e4-8f76-05d9a8e5f6d7

## Testing

CI

## Documentation

N/A


src/cs/tool/Program.cs
```diff
@@ -32,6 +32,7 @@ static unsafe void Main(string[] args)
             }
 
             var ApiTable = MsQuic.Open();
+            const string DomainName = "google.com";
```
nibanks, line 35: But you still hard code it here...? How does this actually fix it? My understanding of the problem is that you any URL in here.

ami-GS: What does this "tool" want to do?
Connection establishment over internet using C#?
Then any endpoint is needed, isn't it?

nibanks: Yes, just a sanity check that it can use C# + msquic to make a connection.

ami-GS: What is the next step?
- Report ignoring?
- Remove this tool?
- Connect localhost?
- Is there any acceptable endpoint?

nibanks: Let's just remove the tool.

nibanks (dismissed):

src/cs/tool/Program.cs
nibanks: Crap. Looks like the tests were using this. Let's add back the tool, make the URL a command line arg, and then update the test code to pass in google.com instead.

src/cs/tool/Program.cs
```diff
@@ -18,12 +18,32 @@ static unsafe void Main(string[] args)
...
+
+            string DomainName = args[0];
+
+            if (string.IsNullOrWhiteSpace(DomainName))
+            {
+                Console.WriteLine("DomainName cannot be empty.");
+                return;
+            }
+
+            if (DomainName.Length > 253)
```
guhetier, line 35: Is there a reason for "253" here?
Below, when we convert it to bytes, we have a buffer of 50 bytes. Wont it be too short for any input of more than 25 characters accepted here?

nibanks (approved):

PR #4669, opened 2024-11-26
Enable Raw datapath from kernel mode test

ami-GS:
## Description

Kernel mode test to call SetParam for setting execution context.

## Testing

see CI

## Documentation

N/A

### DIFF
--- a/src/perf/lib/SecNetPerfMain.cpp
+++ b/src/perf/lib/SecNetPerfMain.cpp
@@ -33,6 +33,36 @@ uint8_t PerfDefaultAffinitizeThreads = false;
 #ifdef _KERNEL_MODE
 volatile int BufferCurrent;
 char Buffer[BufferLength];
+static inline LONG _strtol(const CHAR* nptr, CHAR** endptr, int base) {
+    UNREFERENCED_PARAMETER(base);
+    ULONG temp;
+    RtlCharToInteger(nptr, base, &temp);
+    if (endptr != NULL) {
+        const CHAR* ptr = nptr;
+        while (*ptr >= '0' && *ptr <= '9') {
+            ptr++;
+        }
+        *endptr = (CHAR*)ptr;
+    }
+    return (LONG)temp;
+}
+
+static inline ULONG _strtoul(const CHAR* nptr, CHAR** endptr, int base) {
+    UNREFERENCED_PARAMETER(base);
+    ULONG temp;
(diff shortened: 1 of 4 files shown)


src/test/bin/winkernel/control.cpp
```diff
@@ -690,6 +690,19 @@ QuicTestCtlEvtIoDeviceControl(
...
+#if defined(QUIC_API_ENABLE_PREVIEW_FEATURES)
+        if (UseDuoNic) {
+            QUIC_EXECUTION_CONFIG Config = Params->TestConfigurationParams.Config;
+            Status =
+                MsQuic->SetParam(
+                    nullptr,
+                    QUIC_PARAM_GLOBAL_EXECUTION_CONFIG,
+                    sizeof(Config),
+                    &Config);
+        }
```
nibanks, line 703: Change this to be closer to what user mode does?
```suggestion
        QUIC_EXECUTION_CONFIG Config = Params->TestConfigurationParams.Config;
        if (Config->Flags != QUIC_EXECUTION_CONFIG_FLAG_NONE) {
            Status =
                MsQuic->SetParam(
                    nullptr,
                    QUIC_PARAM_GLOBAL_EXECUTION_CONFIG,
                    sizeof(Config),
                    &Config);
        }
```

ami-GS: LGTM!

src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -142,10 +142,9 @@ QuicMainStart(
     QUIC_EXECUTION_CONFIG* Config = (QUIC_EXECUTION_CONFIG*)RawConfig;
     Config->PollingIdleTimeoutUs = 0; // Default to no polling.
     bool SetConfig = false;
-
-#ifndef _KERNEL_MODE
     const char* IoMode = GetValue(argc, argv, "io");
 
+#ifndef _KERNEL_MODE
```
nibanks, line 147: Actually, I _think_ the only thing that wouldn't be supported in kernel mode after all this is the `rio` stuff. Everything else can be used for kernel.

nibanks: You can exclude it for now, but depending on the final implementation, that might be common code and supported as well.

nibanks (approved):

nibanks (approved):

PR #4670, opened 2024-11-28
Updated ACK_FREQUENCY according to draft-ietf-quic-ack-frequency-10

gaurav2699:
## Description
This PR updates the ACK_FREQUENCY frame to the latest draft. A new reordering threshold is added which indicates the maximum number of packets that can be out of order before an immediate acknowledgment (ACK) is triggered.

## Testing
New Unit test cases have been added to test the reordering threshold logic.

## Documentation

No


src/core/connection.c
```diff
@@ -5245,6 +5247,11 @@ QuicConnRecvFrames(
             } else {
                 Connection->PacketTolerance = UINT8_MAX; // Cap to 0xFF for space savings.
             }
+            if(Frame.ReorderingThreshold < UINT8_MAX) {
```
nibanks, line 5250:
```suggestion
            if (Frame.ReorderingThreshold < UINT8_MAX) {
```

src/core/connection.h
```diff
@@ -463,6 +463,21 @@ typedef struct QUIC_CONNECTION {
...
     uint64_t NextRecvAckFreqSeqNum;
 
+    //
+    // The maximum number of packets that can be out of order before an immediate
+    // acknowledgment (ACK) is triggered. If no specific instructions (ACK_FREQUENCY
+    // frames) are received from the peer, the receiver will immediately acknowledge
+    // any out-of-order packets, which means the default value is 1. A value of 0
+    // means out-of-order packets do not trigger an immediate ACK.
+    //
+    uint8_t ReorderingThreshold;
```
nibanks, line 473: Put these up after the other `uint8_t` (`SendAckFreqSeqNum`) so there isn't so much extra padding

src/core/connection.h
```diff
@@ -463,6 +463,21 @@ typedef struct QUIC_CONNECTION {
...
+
+    //
+    // The maximum number of packets that the peer can be out of order before an immediate
+    // acknowledgment (ACK) is triggered.
+    //
+    uint8_t PeerReorderingThreshold;
```
nibanks, line 479: Is this meant to be what we report to the peer, or what they report to us? Because the code you have seems to be doing both. You're initializing this to our own default, setting it when you receive a frame from the peer and using it when you send frames to the peer. It shouldn't be both. Likely, it should just be what the peer sends to us. For the frames we send, we can hard code something for now.

gaurav2699: When we receive packets, I am setting ReorderingThreshold not PeerReorderingThreshold. PeerReorderingThreshold is supposed to be meant for what we report to the peer while ReorderingThreshold is what they report to us.

nibanks: Ah, you're right. My fault. Sorry!

nibanks:
Looks like CLOG needs to be updated again.

src/core/connection.h
gaurav2699: Didnt get you. I only see uint64_t SendAckFreqSeqNum;

nibanks: Put these before `SendAckFreqSeqNum`, next to the other `uint8_t` fields.

src/core/unittest/FrameTest.cpp
```diff
@@ -257,7 +257,7 @@ struct ResetStreamFrameParams {
                 uint8_t TestValue = (j & 1) ? 64 : 255;
 
                 if (i & 1) {
-                    Temp.Buffer[1] = TestValue;
+                    Temp.Buffer[1] = TestValue;
```
nibanks, line 260: Did you add a tab or remove it here? It's hard to tell.

src/core/ack_tracker.c
```diff
@@ -196,11 +196,12 @@ QuicAckTrackerAckPacket(
         Connection->Settings.MaxAckDelayMs == 0 ||
         (Tracker->AckElicitingPacketsToAcknowledge >= (uint16_t)Connection->PacketTolerance) ||
         (!Connection->State.IgnoreReordering &&
+         Connection->ReorderingThreshold > 0 &&
          (NewLargestPacketNumber &&
           QuicRangeSize(&Tracker->PacketNumbersToAck) > 1 && // There are more than two ranges, i.e. a gap somewhere.
             QuicRangeGet(
             &Tracker->PacketNumbersToAck,
-          QuicRangeSize(&Tracker->PacketNumbersToAck) - 1)->Count == 1))) { // The gap is right before the last packet number.
+          QuicRangeSize(&Tracker->PacketNumbersToAck) - 1)->Count == Connection->ReorderingThreshold))) {
```
nibanks, line 204: I'm thinking the logic is not actually enough (at least in some edge cases). Currently, it's saying 'if there is more than one ACK range, once the last ACK range count equals the reordering threshold, immediately send an ACK'. But if we have sporadic loss, and each ACK range never hits the reordering threshold, we won't send an ACK.

For example, if the reordering threshold is 3 packets, but we never get more than 2 in a row, then this fails. We would only look at the last range and see count is 2, which is less than 3.

I think we need to make a new helper function, `QuicAckTrackerDidHitReorderingThreshold`, which handles the case where there are multiple gaps. I _think_ the function just needs to calculate the difference between the last number in the first range, and the last number in the last range. I still would like to put this in a helper function because this logic is so complex here.

src/core/frame.h
```diff
@@ -846,6 +846,7 @@ typedef struct QUIC_ACK_FREQUENCY_EX {
     QUIC_VAR_INT SequenceNumber;
     QUIC_VAR_INT PacketTolerance;
     QUIC_VAR_INT UpdateMaxAckDelay; // In microseconds (us)
+    QUIC_VAR_INT ReorderingThreshold;
     BOOLEAN IgnoreOrder;
     BOOLEAN IgnoreCE;
```
nibanks, line 851: These fields are no longer in the spec AFAICT:
```
ACK_FREQUENCY Frame {
  Type (i) = 0xaf,
  Sequence Number (i),
  Ack-Eliciting Threshold (i),
  Requested Max Ack Delay (i),
  Reordering Threshold (i),
}
```
Please remove them and update.

gaurav2699: Yes was just doing that, QUIC_ACK_FREQUENCY_EXTRAS also goes right cause we dont use it anyways

nibanks:
`QUIC_TP_ID_MIN_ACK_DELAY` needs to be updated to match the latest in the spec: 
```
min_ack_delay (0xff04de1b):
A variable-length integer representing the minimum amount of time, in microseconds, that the endpoint sending this value is willing to delay an acknowledgment. This limit could be based on the data receiver's clock or timer granularity. min_ack_delay is used by the data sender to avoid requesting too small a value in the Requested Max Ack Delay field of the ACK_FREQUENCY frame.
```

src/core/ack_tracker.h
```diff
@@ -153,3 +155,7 @@ QuicAckTrackerHasPacketsToAck(
         !Tracker->AlreadyWrittenAckFrame &&
         QuicRangeSize(&Tracker->PacketNumbersToAck) != 0;
 }
+
+#if defined(__cplusplus)
+}
+#endif
```
nibanks, line 161: nit: please have a trailing new line

src/core/unittest/FrameTest.cpp
```diff
@@ -243,6 +243,43 @@ TEST(FrameTest, ReliableResetStreamFrameEncodeDecode)
     ASSERT_EQ(Frame.ReliableSize, DecodedFrame.ReliableSize);
 }
 
+TEST(FrameTest, TestQuicAckTrackerDidHitReorderingThreshold)
```
nibanks, line 246: https://datatracker.ietf.org/doc/html/draft-ietf-quic-ack-frequency#section-6.2.1 actually has a lot of examples. Could we put all these into a unit test?

src/core/unittest/FrameTest.cpp
```diff
@@ -243,6 +243,91 @@ TEST(FrameTest, ReliableResetStreamFrameEncodeDecode)
...
+
+    QuicRangeAddValue(&Tracker.PacketNumbersToAck, 3);
+    ASSERT_FALSE(QuicAckTrackerDidHitReorderingThreshold(&Tracker, ReorderingThreshold, 3));
+
+    QuicRangeAddValue(&Tracker.PacketNumbersToAck, 4);
+    ASSERT_FALSE(QuicAckTrackerDidHitReorderingThreshold(&Tracker, ReorderingThreshold, 4));
+
+    QuicRangeAddValue(&Tracker.PacketNumbersToAck, 5);
+    ASSERT_TRUE(QuicAckTrackerDidHitReorderingThreshold(&Tracker, ReorderingThreshold, 5));
+    QuicRangeAddValue(&Tracker.PacketNumbersToAck, 2); // 2 is reported missing
```
nibanks, line 286: Why are you adding these numbers to `PacketNumbersToAck`. In the spec examples, they're never reported as received.

gaurav2699: I need a way to declare that we have reported it missing, I am trying to figure out a way else it will keep acknowledging for each subsequent packets as the gap will always be there. Any suggestions?

src/core/ack_tracker.c
```diff
@@ -99,21 +99,38 @@ QuicAckTrackerAddPacketNumber(
...
+    int result = QuicRangeSearch(&Tracker->PacketNumbersToAck, &Key);
+    if (result >= 0 && result < (int)QuicRangeSize(&Tracker->PacketNumbersToAck) - 1) {
+        SmallestUnreportedMissingPacketNumber = 
+            QuicRangeGetHigh(QuicRangeGet(&Tracker->PacketNumbersToAck, result)) + 1;
+    }
+    else if (result < 0) {
+        SmallestUnreportedMissingPacketNumber = LargestReported;
+    }
+    else {
+        return FALSE;
```
nibanks, line 131:
```suggestion
    } else if (result < 0) {
        SmallestUnreportedMissingPacketNumber = LargestReported;
    } else {
        return FALSE;
```

src/core/ack_tracker.c
```diff
@@ -96,6 +96,42 @@ QuicAckTrackerAddPacketNumber(
...
+
+    uint64_t LargestUnackedPacketNumber = QuicRangeGetMax(&Tracker->PacketNumbersToAck);
+    uint64_t LargestReported = 0; // The largest packet number that could be declared lost 
+    uint64_t SmallestUnreportedMissingPacketNumber; 
+
+    if (Tracker->LargestPacketNumberAcknowledged >= ReorderingThreshold) {
+        LargestReported = Tracker->LargestPacketNumberAcknowledged - ReorderingThreshold + 1;
+    }
+
+    if (LargestReported > LargestUnackedPacketNumber) {
```
nibanks, line 117: I think this should be put in the `if` statement above. It's impossible to be hit if we didn't update `LargestReported` to something other than zero.

src/core/ack_tracker.c
```diff
@@ -182,10 +218,10 @@ QuicAckTrackerAckPacket(
     //   1. The packet included an IMMEDIATE_ACK frame.
     //   2. ACK delay is disabled (MaxAckDelayMs == 0).
     //   3. We have received 'PacketTolerance' ACK eliciting packets.
-    //   4. We received an ACK eliciting packet that doesn't directly follow the
-    //      previously received packet number. So we assume there might have
-    //      been loss and should indicate this info to the peer. This logic is
-    //      disabled if 'IgnoreReordering' is TRUE.
+    //   4. We had received an ACK eliciting packet that is out of order and the
```
nibanks, line 221:
```suggestion
    //   4. We have received an ACK eliciting packet that is out of order and the
```
Let's match the wording/tense above.

src/core/ack_tracker.c
```diff
@@ -96,6 +96,42 @@ QuicAckTrackerAddPacketNumber(
...
+    QUIC_RANGE_SEARCH_KEY Key = { LargestReported, LargestReported };
+    int result = QuicRangeSearch(&Tracker->PacketNumbersToAck, &Key);
+    if (result >= 0 && result < (int)QuicRangeSize(&Tracker->PacketNumbersToAck) - 1) {
+        SmallestUnreportedMissingPacketNumber = 
+            QuicRangeGetHigh(QuicRangeGet(&Tracker->PacketNumbersToAck, result)) + 1;
+    } else if (result < 0) {
+        SmallestUnreportedMissingPacketNumber = LargestReported;
+    } else {
+        return FALSE;
+    }
```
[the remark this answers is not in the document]
gaurav2699: Umm it actually means result is not the last index, as the smallestreportedmissingpacket cannot be LargestUnackedPacketNumber+1.
Let me give you an example:

LargestAck = 18
PacketsToAck = [7 - 10], [12 - 20]
LargestUnacked = 20
LargestReported = 16
SmallestUnreportedMissing = N/A
Send? = N/A

We can do a check later if SmallestUnreportedMissing is bigger than largetunackedpacket too, if that is easier to understand

src/core/ack_tracker.c
```diff
@@ -96,6 +96,39 @@ QuicAckTrackerAddPacketNumber(
         !RangeUpdated;
 }
 
+BOOLEAN
+QuicAckTrackerDidHitReorderingThreshold(
```
nibanks, line 100: Let's put a comment above this that indicated this implements the logic defined in https://datatracker.ietf.org/doc/html/draft-ietf-quic-ack-frequency#name-response-to-out-of-order-pa.

src/core/ack_tracker.c
```diff
@@ -96,6 +96,39 @@ QuicAckTrackerAddPacketNumber(
...
+    _In_ QUIC_ACK_TRACKER* Tracker,
+    _In_ uint8_t ReorderingThreshold
+    )
+{
+    if (ReorderingThreshold == 0) {
+        return FALSE;
+    }
+
+    uint64_t LargestUnackedPacketNumber = QuicRangeGetMax(&Tracker->PacketNumbersToAck);
```
nibanks, line 109: Let's name these exactly what the spec calls them.
```suggestion
    uint64_t LargestUnacked = QuicRangeGetMax(&Tracker->PacketNumbersToAck);
```

nibanks, line 109: Let's put a comment before all these variables, exactly copying the definition from the spec.

src/core/ack_tracker.c
```diff
@@ -96,6 +96,39 @@ QuicAckTrackerAddPacketNumber(
...
+    uint64_t LargestReported = 0; // The largest packet number that could be declared lost 
+    uint64_t SmallestUnreportedMissingPacketNumber;
```
nibanks, line 111:
```suggestion
    uint64_t SmallestUnreportedMissing; 
```

src/core/ack_tracker.c
nibanks, line 100: I've been thinking about this logic and I think the following is a better way to do this:
```c++
bool DidHitReorderingThreshold() {

  if (ReorderingThreshold == 0) return false

  LargestUnacked = AckRanges[AckRanges.Count - 1].Max

  for (Index = AckRanges.Count - 1; Index > 0 Index--) {
    SmallestMissing = AckRanges[Index-1].Max + 1
    if (LargestAcked > SmallestMissing) return false
    if (LargestUnacked - SmallestMissing > ReorderingThreshold) return true
  }

  return false
}
```

src/core/ack_tracker.c
```diff
@@ -96,6 +96,39 @@ QuicAckTrackerAddPacketNumber(
...
+    if (result < 0) {
+        SmallestUnreportedMissingPacketNumber = LargestReported;
+    } else if (result < (int)QuicRangeSize(&Tracker->PacketNumbersToAck) - 1) {
+        SmallestUnreportedMissingPacketNumber = 
+            QuicRangeGetHigh(QuicRangeGet(&Tracker->PacketNumbersToAck, result)) + 1;
+    } else {
+        return FALSE;
+    }
+
+    return LargestUnackedPacketNumber - SmallestUnreportedMissingPacketNumber >= ReorderingThreshold;
```
nibanks, line 129:
```suggestion
    uint64_t LargestUnacked = QuicRangeGetMax(&Tracker->PacketNumbersToAck);

    for (uint32_t Index = QuicRangeSize(&Tracker->PacketNumbersToAck) - 1; Index > 0; --Index) {
        SmallestMissing = QuicRangeGetHigh(QuicRangeGet(&Tracker->PacketNumbersToAck, Index)) + 1;
        if (Tracker->LargestPacketNumberAcknowledged > SmallestMissing) {
            return FALSE;
        }
        if (LargestUnacked - SmallestMissing > ReorderingThreshold) {
            return TRUE;
        }
    }

    return FALSE;
```

src/core/ack_tracker.c
```diff
@@ -96,6 +96,58 @@ QuicAckTrackerAddPacketNumber(
...
+{
+    if (ReorderingThreshold == 0) {
+        return FALSE;
+    }
+
+    const uint64_t LargestUnacked = QuicRangeGetMax(&Tracker->PacketNumbersToAck);
+    uint64_t LargestReported = 0; // The largest packet number that could be declared lost 
+
+    if (Tracker->LargestPacketNumberAcknowledged >= ReorderingThreshold &&
+        Tracker->LargestPacketNumberAcknowledged - ReorderingThreshold + 1 <= LargestUnacked) {
```
nibanks, line 117: If this line (117) is false, does it make sense to do anything? It means we haven't received enough packets between largest ACK and largest Unacked to trigger the threshold, right? Why is this simply not a sub-if block in this to return false?

src/core/ack_tracker.c
```diff
@@ -96,6 +96,58 @@ QuicAckTrackerAddPacketNumber(
...
+    // packet number that is after the largest reported packet number. If the difference
+    // between that missing number and the largest unack'ed number is more than the
+    // reordering threshold, then the condition has been met to send an immediate
+    // acknowledgement.
+    //
+
+    for (uint32_t Index = QuicRangeSize(&Tracker->PacketNumbersToAck) - 1; Index > 0; --Index) {
+        uint64_t SmallestMissing = 0; // Smallest missing after this range
+        uint64_t HighestMissing = QuicRangeGet(&Tracker->PacketNumbersToAck, Index)->Low; // Highest missing in this range
+        if(Index != 0) {
```
nibanks, line 132:
```suggestion
        if (Index != 0) {
```

src/core/ack_tracker.c
```diff
@@ -96,6 +96,58 @@ QuicAckTrackerAddPacketNumber(
...
+            SmallestMissing = QuicRangeGetHigh(QuicRangeGet(&Tracker->PacketNumbersToAck, Index - 1)) + 1;
+        }
+          
+        if (LargestReported > SmallestMissing) {
+            if (HighestMissing > LargestReported) {
+                SmallestMissing = LargestReported;
+            }
+            else {
```
nibanks, line 140:
```suggestion
            } else {
```

src/core/ack_tracker.c
gaurav2699: Yes you are right

src/core/ack_tracker.c
```diff
@@ -96,6 +96,64 @@ QuicAckTrackerAddPacketNumber(
...
+{
+    if (ReorderingThreshold == 0) {
+        return FALSE;
+    }
+
+    const uint64_t LargestUnacked = QuicRangeGetMax(&Tracker->PacketNumbersToAck);
+    uint64_t LargestReported = 0; // The largest packet number that could be declared lost 
+
+    if (Tracker->LargestPacketNumberAcknowledged != 0 &&
+        Tracker->LargestPacketNumberAcknowledged - ReorderingThreshold + 1 > LargestUnacked) {
```
nibanks, line 117: This condition is impossible. `LargestUnacked` is **always** >= `Tracker->LargestPacketNumberAcknowledged`.

src/core/ack_tracker.c
```diff
@@ -96,6 +96,64 @@ QuicAckTrackerAddPacketNumber(
...
+    // Loop through all previous ACK ranges (before last) to find the smallest missing
+    // packet number that is after the largest reported packet number. If the difference
+    // between that missing number and the largest unack'ed number is more than the
+    // reordering threshold, then the condition has been met to send an immediate
+    // acknowledgement.
+    //
+
+    for (int Index = QuicRangeSize(&Tracker->PacketNumbersToAck) - 1; Index >= 0; --Index) {
+        uint64_t SmallestMissing = 0; // Smallest missing in the previous gap
+        uint64_t HighestMissing = QuicRangeGet(&Tracker->PacketNumbersToAck, Index)->Low; // Highest missing in the prevous gap
```
nibanks, line 135: This is not referring to a 'missing' packet number. This is simply the lowest number in a range that we have received. Did you mean to subtract one or something?

gaurav2699: Oh yeah subtracting one will be truer to the variable name. Just wanted to find the range of the gap.

src/core/ack_tracker.c
```diff
@@ -283,11 +336,12 @@ QuicAckTrackerOnAckFrameAcked(
...
     //
     // Drop all packet numbers less than or equal to the largest acknowledged
-    // packet number.
+    // packet number - Reordering Threshold + 1. This is so that we dont lose 
+    // memory of packets that are missing and are yet to be reported. 
     //
     QuicRangeSetMin(
         &Tracker->PacketNumbersToAck,
-        LargestAckedPacketNumber + 1);
+        LargestAckedPacketNumber - Connection->ReorderingThreshold + 2);
```
nibanks, line 344: I thought we decided this wouldn't help, without actually tracking the first gap before.

src/core/ack_tracker.c
```diff
@@ -96,6 +96,64 @@ QuicAckTrackerAddPacketNumber(
...
+        return FALSE;
+    }
+
+    const uint64_t LargestUnacked = QuicRangeGetMax(&Tracker->PacketNumbersToAck);
+    uint64_t LargestReported;  // The largest packet number that could be declared lost 
+
+    //
+    // Largest Reported is equal to the largest packet number acknowledged minus the Reordering Threshold.
+    // If the difference between the largest packet number acknowledged and the Reordering Threshold is smaller than the
+    // smallest packet in the ack tracker, then the largest reported is the smallest packet in the ack tracker.
```
nibanks, line 119: Please limit block comment line to 80 characters.

src/core/ack_tracker.c
```diff
@@ -96,6 +96,64 @@ QuicAckTrackerAddPacketNumber(
...
+    //
+
+    if (Tracker->LargestPacketNumberAcknowledged >= QuicRangeGet(&Tracker->PacketNumbersToAck, 0)->Low + ReorderingThreshold) {
+        LargestReported = Tracker->LargestPacketNumberAcknowledged - ReorderingThreshold + 1;
+    } else {
+        LargestReported = QuicRangeGet(&Tracker->PacketNumbersToAck, 0)->Low;
+    }
```
nibanks, line 126: Extract `QuicRangeGet(&Tracker->PacketNumbersToAck, 0)->Low` out into a new variable `SmallestTracked`.

src/core/ack_tracker.c
```diff
@@ -96,6 +96,64 @@ QuicAckTrackerAddPacketNumber(
...
+    // reordering threshold, then the condition has been met to send an immediate
+    // acknowledgement.
+    //
+
+    for (uint32_t Index = QuicRangeSize(&Tracker->PacketNumbersToAck) - 1; Index > 0; --Index) {
+        uint64_t SmallestMissing = QuicRangeGetHigh(QuicRangeGet(&Tracker->PacketNumbersToAck, Index - 1)) + 1; // Smallest missing in the previous gap
+        uint64_t RangeStart = QuicRangeGet(&Tracker->PacketNumbersToAck, Index)->Low; // Lowest Packet number in the subrange
+          
+        // 
+        // Check if largest reported packet is missing. In that case, the smalles missing packet becomes the largest reported packet.
```
nibanks, line 141:
```suggestion
        // Check if largest reported packet is missing. In that case, the smallest missing packet becomes the largest reported packet.
```
Also, please limit line to 80 characters.

src/core/ack_tracker.c
```diff
@@ -96,6 +96,64 @@ QuicAckTrackerAddPacketNumber(
...
+        //
+        if (LargestReported > SmallestMissing) {
+            if (RangeStart > LargestReported) {
+                SmallestMissing = LargestReported;
+            } else {
+                return FALSE;
+            }
```
nibanks, line 148:
```suggestion
            if (LargestReported >= RangeStart) {
                //
                // TODO - Put comment explaining why you return.
                //
                return FALSE;
            }
            SmallestMissing = LargestReported;
```
Personal preference is to swap the order of the if statements so we can put the return first. Then we don't need an explicit `else` after the return.

src/core/ack_tracker.c
```diff
@@ -96,6 +96,64 @@ QuicAckTrackerAddPacketNumber(
...
+    // Loop through all previous ACK ranges (before last) to find the smallest missing
+    // packet number that is after the largest reported packet number. If the difference
+    // between that missing number and the largest unack'ed number is more than the
+    // reordering threshold, then the condition has been met to send an immediate
+    // acknowledgement.
+    //
+
+    for (uint32_t Index = QuicRangeSize(&Tracker->PacketNumbersToAck) - 1; Index > 0; --Index) {
+        uint64_t SmallestMissing = QuicRangeGetHigh(QuicRangeGet(&Tracker->PacketNumbersToAck, Index - 1)) + 1; // Smallest missing in the previous gap
+        uint64_t RangeStart = QuicRangeGet(&Tracker->PacketNumbersToAck, Index)->Low; // Lowest Packet number in the subrange
```
nibanks, line 138: My personal preference is to swap the order of the definitions of each of these variables. Also, I think the comments on each are probably unnecessary. Finally, let's make them `const`. While mostly unnecessary, it's nice to declare variables `const` if you know they will never be written to later.

nibanks: Also, what do you think of calling `SmallestMissing` `PreviousSmallestMissing`, since it's for the gap before the current range? Is there a better word than 'previous'?

src/core/unittest/FrameTest.cpp
```diff
@@ -243,6 +243,52 @@ TEST(FrameTest, ReliableResetStreamFrameEncodeDecode)
     ASSERT_EQ(Frame.ReliableSize, DecodedFrame.ReliableSize);
 }
 
+BOOLEAN DidHitReorderingThreshold(
```
nibanks, line 246:
```suggestion
bool DidHitReorderingThreshold(
```

src/core/unittest/FrameTest.cpp
```diff
@@ -243,6 +243,52 @@ TEST(FrameTest, ReliableResetStreamFrameEncodeDecode)
...
+    uint8_t ReorderingThreshold,
+    uint64_t LargestPacketNumberAcknowledged, 
+    const std::vector<std::vector<int>>& AckTrackerLists) 
+{
+    QUIC_ACK_TRACKER Tracker;
+    QuicAckTrackerInitialize(&Tracker);;
```
nibanks, line 252:
```suggestion
    QuicAckTrackerInitialize(&Tracker);
```

src/core/unittest/FrameTest.cpp
```diff
@@ -243,6 +243,52 @@ TEST(FrameTest, ReliableResetStreamFrameEncodeDecode)
...
+    Tracker.LargestPacketNumberAcknowledged = LargestPacketNumberAcknowledged;
+    BOOLEAN Result = QuicAckTrackerDidHitReorderingThreshold(&Tracker, ReorderingThreshold);
+    QuicAckTrackerUninitialize(&Tracker);
+    return Result;
+}
+
+TEST(FrameTest, TestQuicAckTrackerDidHitReorderingThreshold)
+{
+
+    ASSERT_FALSE(DidHitReorderingThreshold(0, 0, {{100}}));
```
nibanks, line 267:
```suggestion
    ASSERT_FALSE(DidHitReorderingThreshold(0, 0, {{100}}));
```

src/core/unittest/FrameTest.cpp
```diff
@@ -243,6 +243,52 @@ TEST(FrameTest, ReliableResetStreamFrameEncodeDecode)
...
+
+    // Case 1 
+
```
nibanks, line 270:
```suggestion
    // Case 1
```

src/core/unittest/FrameTest.cpp
```diff
@@ -243,6 +243,52 @@ TEST(FrameTest, ReliableResetStreamFrameEncodeDecode)
...
+    ASSERT_FALSE(DidHitReorderingThreshold(5, 0, {{0}}));
+    ASSERT_FALSE(DidHitReorderingThreshold(5, 0, {{0, 1}}));
+    ASSERT_FALSE(DidHitReorderingThreshold(5, 0, {{0, 1}, {3}}));
+    ASSERT_FALSE(DidHitReorderingThreshold(5, 0, {{0, 1}, {3}, {5}}));
+    ASSERT_FALSE(DidHitReorderingThreshold(5, 0, {{0, 1}, {3}, {5, 6}}));
+    ASSERT_TRUE(DidHitReorderingThreshold(5, 0, {{0, 1}, {3}, {5, 6, 7}}));
+    ASSERT_FALSE(DidHitReorderingThreshold(5, 7, {{0, 1}, {3}, {5, 6, 7, 8}}));
+    ASSERT_TRUE(DidHitReorderingThreshold(5, 7, {{0, 1}, {3}, {5, 6, 7, 8, 9}}));
+
+}
```
nibanks, line 290:
```suggestion
}
```

src/core/unittest/FrameTest.cpp
```diff
@@ -243,6 +243,52 @@ TEST(FrameTest, ReliableResetStreamFrameEncodeDecode)
...
+            QuicRangeAddValue(&Tracker.PacketNumbersToAck, i);
+        }
+    }
+    Tracker.LargestPacketNumberAcknowledged = LargestPacketNumberAcknowledged;
+    BOOLEAN Result = QuicAckTrackerDidHitReorderingThreshold(&Tracker, ReorderingThreshold);
+    QuicAckTrackerUninitialize(&Tracker);
+    return Result;
+}
+
+TEST(FrameTest, TestQuicAckTrackerDidHitReorderingThreshold)
```
nibanks, line 264: Can you please add more of the cases we discussed on our call?

src/core/unittest/FrameTest.cpp
nibanks, line 246: Thoughts on renaming this to simply `TestReorderingThreshold`? Also, let's add a comment to explain it some.

src/core/ack_tracker.c
gaurav2699: Oh yeah, it does help us in few edge cases but not always.

src/core/ack_tracker.c
gaurav2699: The smallestmissing can be changed later so we cant do const for that

nibanks (commented):
Really close!

src/core/frame.c
```diff
@@ -1259,27 +1246,20 @@ QuicAckFrequencyFrameEncode(
...
-
     Buffer = Buffer + *Offset;
     Buffer = QuicVarIntEncode(QUIC_FRAME_ACK_FREQUENCY, Buffer);
     Buffer = QuicVarIntEncode(Frame->SequenceNumber, Buffer);
-    Buffer = QuicVarIntEncode(Frame->PacketTolerance, Buffer);
-    Buffer = QuicVarIntEncode(Frame->UpdateMaxAckDelay, Buffer);
-    QuicUint8Encode(Extras.Value, Buffer);
+    Buffer = QuicVarIntEncode(Frame->AckElicitingThreshold, Buffer);
+    Buffer = QuicVarIntEncode(Frame->RequestedMaxAckDelay, Buffer);
+    Buffer = QuicVarIntEncode(Frame->ReorderingThreshold, Buffer);
```
nibanks, line 1262:
```suggestion
    QuicVarIntEncode(Frame->ReorderingThreshold, Buffer);
```

src/core/unittest/FrameTest.cpp
```diff
@@ -243,6 +243,60 @@ TEST(FrameTest, ReliableResetStreamFrameEncodeDecode)
...
+
+    // Case 1
+    ASSERT_FALSE(TestReorderingThreshold(3, 0, {{0}}));
+    ASSERT_FALSE(TestReorderingThreshold(3, 0, {{0, 1}}));    
+    ASSERT_FALSE(TestReorderingThreshold(3, 0, {{0, 1}, {3}}));
+    ASSERT_FALSE(TestReorderingThreshold(3, 0, {{0, 1}, {3, 4}}));
+    ASSERT_TRUE(TestReorderingThreshold(3, 0, {{0, 1}, {3, 4, 5}}));
+    ASSERT_FALSE(TestReorderingThreshold(3, 5, {{0, 1}, {3, 4, 5}, {8}}));
+    ASSERT_TRUE(TestReorderingThreshold(3, 5, {{0, 1}, {3, 4, 5}, {8, 9}}));
+    ASSERT_TRUE(TestReorderingThreshold(3, 9, {{0, 1}, {3, 4, 5}, {8, 9, 10}}));
```
nibanks, line 283: A thought: Having separate arrays for each subrange doesn't seem super necessary, unless each sub array is actually min max. So, we should use one of the following instead:
```c
ASSERT_TRUE(TestReorderingThreshold(3, 9, {{0, 1}, {3, 5}, {8, 10}}));
```
or
```c
ASSERT_TRUE(TestReorderingThreshold(3, 9, {0, 1, 3, 4, 5, 8, 9, 10}));
```
I think the first is easier to read.

nibanks: Feel free to say 'Won't Fix' though.

src/core/ack_tracker.c
```diff
@@ -96,6 +96,73 @@ QuicAckTrackerAddPacketNumber(
...
+// This function implements the logic defined in Section 6.2 [draft-ietf-quic-frequence-10]
+// to determine if the reordering threshold has been hit.
+//
+BOOLEAN
+QuicAckTrackerDidHitReorderingThreshold(
+    _In_ QUIC_ACK_TRACKER* Tracker,
+    _In_ uint8_t ReorderingThreshold
+    )
+{
+    if (ReorderingThreshold == 0) {
```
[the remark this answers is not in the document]
gaurav2699: Actually since we are assuming the smallest tracked packet isnt out of order, we would need atleast 2 subranges for our logic. So it should be  QuicRangeSize(&Tracker->PacketNumbersToAck) < 2

nibanks (dismissed):

nibanks (approved):

PR #4677, opened 2024-12-04
reduce overhead of rps-multi

ProjectsByJackHe:
## Description

Recently we've been seeing azure runners in iocp "crashing" (losing connection to the Github Backend).

One possible cause is that we are starving the CPU and that fails some heartbeat check on Github's side.

Let's try reducing this constraint.
## Testing

CI

## Documentation

N/A


nibanks:
Closing as this seems unnecessary now.

Issue #4682, opened 2024-12-04
Fedora 41 packages not being built or released


sillyotter:
### Describe the feature you'd like supported

A year ago I opened an issue (#4092) and submitted PR (#4095) to get Fedora 39 packages of msquic released.  I have looked but haven't seen any incoming PR's to add Fedora 41 to the list, so I thought I might submit my own again.  Apologies if someone else is already fixing this.  I'll submit the PR just as soon as I submit the issue.

### Proposed solution

Modify .azure/OneBranch.Publish.yml to include Fedora 41

sillyotter:
PR #4683 has the fix

sillyotter:
We got some really bad miscommunications in the original PR. I ended up commiting a bunch of incorrect changes so I deleted it and started over.  The new fix is in #4686 

Initially, someone pointed out there was a typo in the change I submitted. There was, but there was also a typo in the suggested change.   "destionations" was wrong, but the correction, at least what I saw, was to make it a singular "destination".  I didn't even seen the o was wrong, just that the suggested fix was to make it singular.  I pointed out that the plural was used everywhere else and was suggested I could fix them all, so I did.   But, once we all got on the same page, I fixed it correctly.  One additional fedora 41 line and a fix to the destionations typo in the containing block.

nibanks:
Thank you so much for working through this with us!

PR #4683, opened 2024-12-04
Add publishing support for fedora 41

sillyotter:
## Description

Modify the `OneBranch.publish` file to include fedora 41

References issue #4682 

## Testing

I can't invoke your publish process, which is what this changed, so I can't really test it.

## Documentation

Not that I can find.

### DIFF
--- a/.azure/OneBranch.Publish.yml
+++ b/.azure/OneBranch.Publish.yml
@@ -27,7 +27,7 @@ parameters:
     type: object
     default:
     - name: RPM
-      destinations:
+      destination:
         - microsoft-sles12-prod-yum     # 12 3.12
         - microsoft-sles15-prod-yum     # 15 4.12
         - microsoft-centos7-prod-yum    # 7 3.10
@@ -41,7 +41,7 @@ parameters:
         - microsoft-rhel8.0-prod-yum    # 8.0 4.18
         - microsoft-rhel8.1-prod-yum    # 8.1 4.18
     - name: DEB
-      destinations:
+      destination:
         - microsoft-ubuntu-xenial-prod-apt   # 16.04 4.4
         - microsoft-ubuntu-bionic-prod-apt   # 18.04 4.15
         - microsoft-ubuntu-focal-prod-apt    # 20.04 5.4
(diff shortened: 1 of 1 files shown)


nibanks:
@ami-GS can you please review this?

.azure/OneBranch.Publish.yml
```diff
@@ -80,6 +80,7 @@ parameters:
     - name: RPM
       destionations:
```
guhetier, line 81: Looks like there is a typo here `destionations` -> `destination`?

nibanks: I agree

sillyotter: If there is, it's been there for a while. Fedora 40 has been published, so i assumed what's there works.  I didn't touch that line. Just duplicated the 40 line and tweaked it a bit to say 41.

sillyotter: Also, the DEB section uses the same wording just a few lines away

nibanks: Mind fixing them?

sillyotter:

.azure/OneBranch.Publish.yml
sillyotter: PR updated with the requested fix, though as before, I can't prove it works

.azure/OneBranch.Publish.yml
```diff
@@ -27,7 +27,7 @@ parameters:
     type: object
     default:
     - name: RPM
-      destinations:
+      destination:
```
nibanks, line 30: These need to be `destinations` with the `s`. that's what is used down below.

sillyotter: I'm afraid we're not communicating very well.  The instructions above were to turn it from destionation**s** to destination,  There is a typo in what was there earlier, the extra o someone left in there, but the fix suggested was to strip off the s at the end.  I didn't even see the o, just the removal of the s.  My comment that they were all like that (plural) coupled with your 'can you fix them all' comment made me think you wanted it to just say destination (singular) everywhere.  In any event, now that I've caught up, proper fix inbound.

PR #4690, opened 2024-12-06
Replacing panics with `Result` returns, and adds platform-specific status codes

masa-koz:
## Description

This pull request to `src/lib.rs` includes changes to improve error handling by replacing panics with `Result` returns, and adds platform-specific status codes to inspect Err's content.

## Testing

cargo test

## Documentation

No


src/lib.rs
```diff
@@ -1698,8 +1830,9 @@ impl Stream {
             )
         };
         if Status::failed(status) {
-            panic!("StreamSend failure 0x{:x}", status);
+            return Err(status);
         }
+        Ok(())
```
nibanks, line 1835: 1. Does Rust allow for non-zero `Ok` status codes? For instance, we have the 'pending' status.
2. Should we return `Ok(status)` here?

src/lib.rs
```diff
@@ -1782,8 +1915,8 @@ extern "C" fn test_stream_callback(
 
 #[test]
 fn test_module() {
-    let api = Api::new();
-    let registration = Registration::new(&api, ptr::null());
+    let api = Api::new().unwrap();
```
nibanks, line 1918: If this returns an `Err` will `unwrap` trigger the panic now?

masa-koz: I've refactored error handling in tests in the latest commit.

nibanks (approved):

PR #4695, opened 2024-12-08
Refactoring ApiTable and Handle

masa-koz:
## Description

This pull request includes significant changes to the `src/lib.rs` file aimed at improving safety and concurrency by introducing `Arc` for shared ownership and marking certain structs as `Sync` and `Send`. Additionally, the some structs such as `Registration`, are  refactored to hold `Arc` of ApiTable in order to make sure that MsQuicClose will be alled after all these struct are dropped.

## Testing

cargo test

## Documentation

No

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -11,9 +11,11 @@ use libc::c_void;
 use serde::{Deserialize, Serialize};
 use std::convert::TryInto;
 use std::fmt;
+use std::ops::Deref;
 use std::option::Option;
 use std::ptr;
 use std::result::Result;
+use std::sync::Arc;
 #[macro_use]
 extern crate bitfield;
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1229,50 +1239,47 @@ struct ApiTable {
...
 pub struct Listener {
-    table: *const ApiTable,
+    table: Arc<ApiTable>,
     handle: Handle,
 }
 
 /// A single QUIC stream on a parent connection.
 pub struct Stream {
-    table: *const ApiTable,
+    table: Arc<ApiTable>,
```
youyuanwu, line 1282: Arc on every handle might have perf impact especially if sever uses many streams. Maybe we should follow the cpp wrapper to have a global API table: https://github.com/microsoft/msquic/blob/21e5b41e4e9f88d809de4c78e461b57641bfca8e/src/inc/msquic.hpp#L434

masa-koz: The global API table may be better in the performance perspective. But it is difficult to free it in rust.

youyuanwu: Maybe it is a right use case for ctor: https://github.com/mmastrac/rust-ctor

masa-koz: I've opend #4708.

PR #4696, opened 2024-12-09
Auto generate rust bindings

youyuanwu:
## Description
The FFI definitions for rust should be auto generated instead of manually written.
This PR adds a new rust crate with auto generated bindings aiming to replace or remove all the manual bindings in the existing crate. Existing code is not changed. In future, the safe rust wrappers should be written on top of this generated bindings.

The generated code is from winmd and windows-bindgen tool chain and is cross platform. It does not have any function that requires linking, and I plan to leave those functions to be dynamically loaded.

## Testing
NA

## Documentation
TBD


src/lib.rs
```diff
@@ -143,8 +143,10 @@ mod status {
 #[cfg(target_os = "linux")]
```
youyuanwu, line 143: I must have run cargo fmt. This might need to be reverted.

nibanks: Yeah, let's revert that, and maybe bring it in a separate PR if you think it's really needed.

youyuanwu:
Superseded by #4723

PR #4692, opened 2024-12-07
Refactor: Add namespace, enable sanitize builds

talregev:
- Add namespace
- Enable sanitize builds
- Refactor platform unittest external cmake

PR #4701, opened 2024-12-10
Release 2.4.8

talregev:
Release 2.4.8

Refactor:
- Add namespace
- Enable sanitize builds
- Refactor platform unittest external cmake
- Add alias

### DIFF
--- a/src/bin/CMakeLists.txt
+++ b/src/bin/CMakeLists.txt
@@ -272,7 +272,7 @@ configure_file(msquic-config.cmake.in ${CMAKE_BINARY_DIR}/msquic-config.cmake)
 install(FILES ${CMAKE_BINARY_DIR}/msquic-config.cmake DESTINATION share/msquic)
 
 if(BUILD_SHARED_LIBS)
-    install(EXPORT msquic DESTINATION share/msquic)
+    install(EXPORT msquic NAMESPACE msquic:: DESTINATION share/msquic)
 endif()
 
 if (MSVC AND NOT QUIC_ENABLE_SANITIZERS AND BUILD_SHARED_LIBS)
--- a/.azure/OneBranch.Package.yml
+++ b/.azure/OneBranch.Package.yml
@@ -95,7 +95,7 @@ extends:
           ob_createvpack_owneralias: quicdev
           ob_createvpack_description: msquic.$(Build.SourceBranchName)
           ob_createvpack_versionAs: string
-          ob_createvpack_version: 2.4.7-$(Build.BuildId)
+          ob_createvpack_version: 2.4.8-$(Build.BuildId)
         steps:
         - task: DownloadPipelineArtifact@2
           inputs:
--- a/.github/workflows/build-reuse-win.yml
+++ b/.github/workflows/build-reuse-win.yml
@@ -87,11 +87,11 @@ jobs:
       shell: pwsh
       run: scripts/build.ps1 -Config ${{ inputs.config }} -Platform ${{ inputs.plat }} -Arch ${{ inputs.arch }} -Tls ${{ inputs.tls }} -DisablePerf -DynamicCRT ${{ inputs.sanitize }}
     - name: Build External Platform Test
-      if: inputs.build == '-Test' && inputs.sanitize != '-Sanitize'
+      if: inputs.build == '-Test'
       shell: pwsh
       run: |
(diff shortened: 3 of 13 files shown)


src/bin/CMakeLists.txt
```diff
@@ -272,7 +272,7 @@ configure_file(msquic-config.cmake.in ${CMAKE_BINARY_DIR}/msquic-config.cmake)
 install(FILES ${CMAKE_BINARY_DIR}/msquic-config.cmake DESTINATION share/msquic)
 
 if(BUILD_SHARED_LIBS)
-    install(EXPORT msquic DESTINATION share/msquic)
+    install(EXPORT msquic NAMESPACE msquic:: DESTINATION share/msquic)
```
dg0yt, line 275: I'm happy seeing CMake namespaces being added to exported targets. But it should be clear that this is a breaking change if not complemented by providing an alias with the old name. I wonder if a breaking change is desired in a patch release.

talregev: I already submit it to upstream and it accepted. and maybe it will be release.
https://github.com/microsoft/msquic/pull/4692
https://github.com/microsoft/msquic/pull/4701

dg0yt: This repo is upstream. That's why I raise the point:
Adding the namespace without add polyfill breaks users.

nibanks: How do we add the alias? I assume this is only a build time breaking change? Regardless, I would like to prevent those, if possible.

dg0yt: The namespace is a useful service for consumers only: It signals to CMake that the identifier is a CMake target, not plain input to the linker. This improves detecting configuration errors.

Alias could be setup in `msquic-config.cmake`, after including the generated export files. Sth. like
~~~cmake
foreach(_t IN ITEMS msquic msquic_platform ...)
    if(TARGET msquic::${_t} AND NOT TARGET ${_t})
        add_library(${_t} ALIAS msquic::${_t})
    endif()
endforeach()
~~~

dg0yt: This could be removed in 2.5 or 3, after proper deprecation.

talregev: @dg0yt How you can add this to `msquic-config.cmake`? this file generate automatic. I will glad for your help.

talregev:
@nibanks @dg0yt 
I don't have the knowledge how to add alias. I am close this PR.

dg0yt:
The suggested change was applied almost literally. (I didn't fill in all targets.)
The change is to be inserted after this line:
https://github.com/microsoft/msquic/blob/21e5b41e4e9f88d809de4c78e461b57641bfca8e/src/bin/msquic-config.cmake.in#L4

Are there any other obstacles?

talregev:
This should be done in separate PR to the main.

talregev:
@nibanks 
I added the alias

talregev:
Better wait to @dg0yt changes in vcpkg:
https://github.com/microsoft/vcpkg/pull/42788

talregev:
@nibanks At the moment, vcpkg port was merge without apply the patch inside.
Do you want to move forward with these changes that will match more closer msquic head to release?

Let me know what do you think.

dg0yt:
IMO one PR shouldn't mix these independent topics: 
- Bumping version number
- Adding CMake export namespace and aliases
- Adding a test for external CMake usage.

The PR which adds the test could also add commits with necessary fixes. That's how to get to a good CMake usage pattern. Similar to what was done in vcpkg with vcpkg-ci-msh3.

Dealing with export namespace and aliases would follow the integration of the test.

dg0yt:
> Dealing with export namespace and aliases would follow the integration of the test.

Probably it is enough to add an alias for `msquic` because this is the only target name which was exported in past releases IIUC.

talregev:
> > Dealing with export namespace and aliases would follow the integration of the test.
> 
> Probably it is enough to add an alias for `msquic` because this is the only target name which was exported in past releases IIUC.

This commits are cherry picks. As upstream rules, you just bring previous commit from head without change them. If I understood it correctly. 

I have one commit that bring the namespace and refactor platform unittest external cmake. (There is no adding test, it old text).
And second commit is the alias.
and the third is bump the version number (without cherry picks).

nibanks:
Can we close this?

talregev:
> Can we close this?

Yes.

PR #4706, opened 2024-12-12
Restrict MaxOperationsPerDrain to Greater Than Zero

### LINKED ISSUE #4705: set MaxOperationsPerDrain to 0 causes worker thread busy spinning
### Describe the bug

set connection Settings.MaxOperationsPerDrain to 0 causes worker busy spinning and connection hang (no callback.) 

### Affected OS

- [ ] Windows
- [x] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

_No response_

### MsQuic version

v2.3

1. In Listener callback handles `QUIC_LISTENER_EVENT_NEW_CONNECTION`
2. in another thread (none worker thread), set the new connection settings `MaxOperationsPerDrain` to zero with 
```
MsQuic->SetParam(Connection,
                                   QUIC_PARAM_CONN_SETTINGS,
                                   sizeof(QUIC_SETTINGS),
                                    &Settings)))
```
and it returns success. 

3. In the same thread same as above, set conn configuration with `MsQuic->ConnectionSetConfiguration`
and it returns success.

4. One of the worker thread busy spin (100% CPU usage)

### Expected behavior

Worker thread should not busy spin. 

OR

Disallow set `MaxOperationsPerDrain` to `0` and  step 3) should return error.

AND
Document the usecase of `MaxOperationsPerDrain=0` if it is by design.

### Actual outcome

4. One of the worker thread busy spin (100% CPU usage)

### Additional details

This is found during prop based fuzzy test of [quicer](github.com/emqx/quicer)

nibanks:
## Description

Fixes #4705 by disallowing the app to set a value of zero.

## Testing

Added

## Documentation

N/A

### DIFF
--- a/src/core/settings.c
+++ b/src/core/settings.c
@@ -431,6 +431,9 @@ QuicSettingApply(
         Destination->IsSet.DatagramReceiveEnabled = TRUE;
     }
     if (Source->IsSet.MaxOperationsPerDrain && (!Destination->IsSet.MaxOperationsPerDrain || OverWrite)) {
+        if (Source->MaxOperationsPerDrain == 0) {
+            return FALSE;
+        }
         Destination->MaxOperationsPerDrain = Source->MaxOperationsPerDrain;
         Destination->IsSet.MaxOperationsPerDrain = TRUE;
     }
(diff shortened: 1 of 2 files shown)


qzhuyan (approved):

guhetier (approved):
Should the documentation be updated to mention that zero is not a valid value?

nibanks:
> Should the documentation be updated to mention that zero is not a valid value?

Nah, I don't think it's really necessary. Zero is fundamentally a bad value. That's essentially saying the max work you're allowed to do is nothing.

PR #4708, opened 2024-12-13
Use global API table instead of each handle's having table ptr

masa-koz:
## Description
This pull request includes several changes to the `src/lib.rs` file and an update to the `Cargo.toml` file. The primary focus is on removing the `table` pointer from various structs and centralizing the API table management using a static `API` instance. Additionally, the `ctor` crate is introduced for constructor and destructor functions to manage the API lifecycle.

## Testing
cargo test

## Documentation
No


src/lib.rs
```diff
@@ -1450,27 +1445,45 @@ impl Drop for Api {
...
 }
 
+pub static mut API: Api = Api { table: ptr::null() };
+
+#[ctor::ctor]
+fn open_msquic() {
+    unsafe {
+        API = match Api::new() {
+            Ok(api) => api,
+            Err(status) => panic!("Failed to open MsQujsonic: {}", status),
```
nibanks, line 1455:
```suggestion
            Err(status) => panic!("Failed to open MsQuic: {}", status),
```

src/lib.rs
```diff
@@ -1450,27 +1445,45 @@ impl Drop for Api {
     }
 }
 
+pub static mut API: Api = Api { table: ptr::null() };
+
+#[ctor::ctor]
+fn open_msquic() {
```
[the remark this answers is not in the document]
youyuanwu: It runs when executable starts or dll load. The other option is to use global OnceCell, but that has the limitation that its destructor and Drop impl will not be executed (rust limitation), and memory and handle cleanup of the API table would be left to the os at program exit time.

youyuanwu (approved):

src/lib.rs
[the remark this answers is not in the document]
youyuanwu: It might be possible to do OnceCell to do lazy init, and then use dtor to close the api table.

masa-koz:
Now, we use lazy init for MsQuic API instead of a global init.

src/lib.rs
```diff
@@ -1446,31 +1442,46 @@ impl Api {
...
 impl Drop for Api {
     fn drop(&mut self) {
-        unsafe { MsQuicClose(self.table) };
+        if self.table != ptr::null() {
+            unsafe { MsQuicClose(self.table) };
+        }
+    }
+}
+
+pub static mut API: LazyLock<Api> = LazyLock::new(|| {
```
nibanks, line 1451: I assume this means it actually uses a Lock every time the API is accessed now?

masa-koz: No. I suppose that we can access API in paralell from more than one threads after initiliazation. I assume LazyLock acquires "lock" only when it executes the closure for initilization. 

https://github.com/rust-lang/rust/blob/6d9f6ae36ae1299d6126ba40c15191f7aa3b79d8/library/std/src/sync/lazy_lock.rs#L207-L224

https://github.com/rust-lang/rust/blob/6d9f6ae36ae1299d6126ba40c15191f7aa3b79d8/library/std/src/sync/once.rs#L152-L158

nibanks: But usually, you need a lock to check for initialization too.

masa-koz: Yes. But I suppose that the cost is cheap because an `AtomicPtr` is used to check for initilization.
https://github.com/rust-lang/rust/blob/6d9f6ae36ae1299d6126ba40c15191f7aa3b79d8/library/std/src/sys/sync/once/queue.rs#L125C1-L129C62

nibanks: But it imposes a cost on every single usage of the variable, which might be happening on a LOT of parallel threads. Imagine a machine with a 100+ CPUs, all grabbing the lock every time they want to check if the variable is initialized.

I think something a bit better is necessary. Only when you are initializing a registration _might_ you not have an initialized library after; but once you have a registration, you **know** you have initialized the library. So I'm wondering if we have (1) an unprotected global variable for the library and (2) a lazy 'run once' call when RegistrationOpen is called to make sure the global variable is properly set.

youyuanwu: The first time init is expensive and requires lock. After that is should be cheap. But if it is just an atomic var comparison, it   is still expensive. (Does it messup the cpu cache? with atomic acquire Relaxed?)

This is similar to cpp11 thread save static var init: https://en.cppreference.com/w/cpp/language/storage_duration#Static_block_variables
It can do none-atomic bool check.

nibanks: You still need to have a volatile read, where the only real place you need synchronization is the initialization point of all the callers (i.e. where they call RegistrationOpen). I want to make sure we don't introduce overhead here.

masa-koz: We initialize APITABLE when we first initialize Api or Registration using `std::sync::Once`. There is no longer overhead when accessing APITABLE!

src/lib.rs
```diff
@@ -1393,28 +1392,34 @@ impl CredentialConfig {
...
 impl Api {
-    pub fn new() -> Result<Api, u32> {
-        let new_table: *const ApiTable = ptr::null();
-        let status = unsafe { MsQuicOpenVersion(2, &new_table) };
-        if Status::failed(status) {
-            return Err(status);
+    pub fn new() -> Self {
+        // We initialize APITABLE at only once.
+        unsafe {
+            START_MSQUIC.call_once(|| {
```
nibanks, line 1398: I'm confused why this is both in the Api and Registration interface. Aren't they separate 'call once' things?

masa-koz: Api may be initialized and its method may be called before registration initilized. But Api may not initialized. So, I suppose that we should call `MsQuicOpenVersion()` when Api OR Registration is initialized.
And both `Api::new()` and `Registration::new()` will access the same `START_MSQUIC`. If Api is initialized before Registration is initialized, the closure which is passed to `START_MSQUIC.call_once` in `Registration::new` will not be executed because `START_MSQUIC.call_once` is already executed in `Api::new()`. If Registration is first initialized, its closure will be executed.

masa-koz: @nibanks Let me know what you think about this.

nibanks (approved):

PR #4712, opened 2024-12-16
Run fmt and clippy for rust code

youyuanwu:
## Description
Rust code should be formatted and linted to ensure consistency and quality.
Added fmt and clippy lint to cargo pipeline.

## Testing
NA

## Documentation
NA

### DIFF
--- a/.github/workflows/cargo.yml
+++ b/.github/workflows/cargo.yml
@@ -40,13 +40,10 @@ jobs:
     - name: Install Cargo
       if: runner.os == 'Linux'
       run: curl https://sh.rustup.rs -sSf | sh -s -- -y
-    
     - name: Cargo fmt
       run: cargo fmt --all -- --check
-
     - name: Cargo clippy
       run: cargo clippy --all-targets -- -D warnings
-
     - name: Cargo build
       run: cargo build --all
     - name: Cargo test
--- a/.github/workflows/cargo.yml
+++ b/.github/workflows/cargo.yml
@@ -40,6 +40,13 @@ jobs:
     - name: Install Cargo
       if: runner.os == 'Linux'
       run: curl https://sh.rustup.rs -sSf | sh -s -- -y
+    
+    - name: Cargo fmt
+      run: cargo fmt --all -- --check
+
+    - name: Cargo clippy
+      run: cargo clippy --all-targets -- -D warnings
+
     - name: Cargo build
       run: cargo build --all
     - name: Cargo test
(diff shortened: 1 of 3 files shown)


.github/workflows/cargo.yml
```diff
@@ -40,6 +40,13 @@ jobs:
     - name: Install Cargo
       if: runner.os == 'Linux'
       run: curl https://sh.rustup.rs -sSf | sh -s -- -y
+
```
nibanks, line 43: nit: please remove the extra spaces, since they're not used anywhere else.

src/lib.rs
```diff
@@ -33,9 +33,9 @@ pub type BOOLEAN = ::std::os::raw::c_uchar;
 
 /// Family of an IP address.
 pub type AddressFamily = u16;
-pub const ADDRESS_FAMILY_UNSPEC: AddressFamily = c_types::AF_UNSPEC as u16;
-pub const ADDRESS_FAMILY_INET: AddressFamily = c_types::AF_INET as u16;
-pub const ADDRESS_FAMILY_INET6: AddressFamily = c_types::AF_INET6 as u16;
+pub const ADDRESS_FAMILY_UNSPEC: AddressFamily = c_types::AF_UNSPEC;
+pub const ADDRESS_FAMILY_INET: AddressFamily = c_types::AF_INET;
+pub const ADDRESS_FAMILY_INET6: AddressFamily = c_types::AF_INET6;
```
nibanks, line 38: Looks like this breaks on other platforms?

youyuanwu: Fixed by keeping the cast and suppressing the linting.

src/lib.rs
```diff
@@ -143,8 +146,10 @@ mod status {
 #[cfg(target_os = "linux")]
 mod status {
     pub const QUIC_STATUS_SUCCESS: u32 = 0;
-    pub const QUIC_STATUS_PENDING: u32 = 0xFFFFFFFE; /// -2
-    pub const QUIC_STATUS_CONTINUE: u32 = 0xFFFFFFFF; /// -1
+    pub const QUIC_STATUS_PENDING: u32 = 0xFFFFFFFE;
+    /// -2
+    pub const QUIC_STATUS_CONTINUE: u32 = 0xFFFFFFFF;
+    /// -1
```
nibanks, line 152: Why isn't it allowing comments on the same line? I'm not a fan of this change.

nibanks (approved):
Thanks for making this change!

PR #4717, opened 2024-12-19
fix(epoll): buffer overflow when GSO off

qzhuyan:
## Description

Fix buffer overflow that SendData buffer size could not be checked correctly in `CxPlatSendDataFinalizeSendBuffer` when `SendData->SegmentSize` is 0.

This issue was found in a env where GSO is off, mtu size 1400 and sending large payload (size: 780038)  + Linux epoll.
It is hard to reproduce in other env. 

this is a quick fix and hurt performance,  I think for long term the `CxPlatSendDataFinalizeSendBuffer` should be aware of the allocating buffer size. 

## Testing
_

## Documentation

_


qzhuyan:
More info of ASAN check (without the fix) here:

https://github.com/emqx/quic/issues/334

qzhuyan:
I found a better fix here which get merged already
 https://github.com/microsoft/msquic/pull/4621/changes

PR #4723, opened 2024-12-31
Auto generate rust bindings using rust bindgen

youyuanwu:
## Description
The FFI definitions for rust should be auto generated instead of manually written.
This PR adds a new rust crate with auto generated bindings aiming to replace or remove all the manual bindings in the existing crate. Existing code is not changed. In future, the safe rust wrappers should be written on top of this generated bindings.

supersede #4696 . This PR uses rust bindgen instead of windows bindgen that has better cross platform support.
Generated code is checked in, and an "overwrite" feature can be used to regenerate the code.
MacOs currently uses Linux bindings, we need to generate for MacOs eventually (I don't have a Mac).

## Testing
NA. Generated code compiles.

## Documentation
NA


src/lib.rs
```diff
@@ -23,6 +23,7 @@ use std::result::Result;
 use std::sync::Once;
 #[macro_use]
 extern crate bitfield;
+pub mod ffi;
```
nibanks, line 26: So the bindings you generate aren't actually used yet though?

src/ffi/wrapper.hpp
```diff
@@ -0,0 +1,44 @@
+// header wrapper to feed into bindgen
+
+#include "msquic.h"
+
+// bindgen or clang has problem with c marcro functions
```
nibanks, line 5: But this requires people to know to update this file if a new QUIC_STATUS_* is added. Any way to automate validation for that?

scripts/build.rs
```diff
@@ -63,4 +63,46 @@ fn main() {
...
         println!("cargo:rustc-link-lib=static=msquic");
     }
+
+    #[cfg(feature = "overwrite")]
+    overwrite_bindgen();
+}
+
+/// Read the c header and generate rust bindings.
+#[cfg(feature = "overwrite")]
+fn overwrite_bindgen() {
```
nibanks, line 73: Can we update the GitHub action to ensure the bindings are up to date?

src/lib.rs
youyuanwu: Currently they are not used, to avoid merge conflicts. It will replace manually wrote bindings in future PRs.

nibanks (approved):

PR #4726, opened 2024-12-31
CMake: Force openssl QUIC_TLS on non-Windows

dg0yt:
## Description

On Windows, `QUIC_TLS` can be "schannel" or "openssl". For non-Windows targets, `openssl` is the only valid value. By marking the cache variable as `INTERNAL`, this value is [enforced](https://cmake.org/cmake/help/v3.20/command/set.html#set-cache-entry) (regardless of user input), and a false impression of choice is avoided.

## Testing

No change.

## Documentation

No change.  
Related: https://github.com/microsoft/msquic/blob/main/docs/Platforms.md.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,7 +76,7 @@ enable_testing()
 if (WIN32)
     set(QUIC_TLS "schannel" CACHE STRING "TLS Library to use")
 else()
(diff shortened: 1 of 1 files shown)


dg0yt:
Igore, there is "openssl", and there is "openssl3".

PR #4737, opened 2025-01-05
Provide namespaced CMake targets

dg0yt:
## Description

Projects (e.g. msh3) that use CMake targets from dependencies (here: msquic) should use namespaced target names from those dependencies. This enables CMake to recognize the intention to use a target, not a library name. CMake can report a missing target definition early (at generation time), instead of creating linker instruction which may fail at build time.

msquic already added namespace `msquic::` to the exported CMake config. However, this should be complemented with `msquic::` `ALIAS` targets inside the build, for reverse dependencies which integrate msquic via `add_subdirectory`. This PR adds the missing targets.

In addition the PR shortens the exported name of `msquic_platform` to `msquic::platform`.

The `msquic::base_link` alias is added for use in msh3. (This nicely illustrate how the namespace clarifies the origin of `base_link`.)

## Testing

Covered by src/platform/unittest/external/CMakeLists.txt.

~The msquic/msh3 integration with namespaced targets is tested in https://github.com/microsoft/vcpkg/pull/43018/files, with a DLL build of msquic passed to msh3 via CMake config.~(Blocked.)

## Documentation

No change due to lack of specific CMake usage documentation.
(vcpkg prints an generated hint which will need to be reduced to linking `msquic::msquic`.)

### DIFF
--- a/src/bin/CMakeLists.txt
+++ b/src/bin/CMakeLists.txt
@@ -11,6 +11,7 @@ endif()
 
 if(BUILD_SHARED_LIBS)
     add_library(msquic SHARED ${SOURCES})
+    add_library(msquic::msquic ALIAS msquic)
     target_include_directories(msquic PUBLIC $<INSTALL_INTERFACE:include>)
     target_link_libraries(msquic PRIVATE core msquic_platform inc warnings logging base_link main_binary_link_args)
     set_target_properties(msquic PROPERTIES OUTPUT_NAME ${QUIC_LIBRARY_NAME})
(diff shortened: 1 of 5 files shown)


nibanks (approved):

dg0yt:
This PR should be good, but the integration in vcpkg cannot complete, https://github.com/nibanks/msh3/issues/235#issuecomment-2571663870.

nibanks:
CI is failing

nibanks (approved):

PR #4740, opened 2025-01-08
[Expose TTL Follow up PR] move version logic to platform_xyz instead of datapath

ProjectsByJackHe:
## Description

There was some additional feedback from https://github.com/microsoft/msquic/pull/4602

The biggest one being that we have the version fetching logic in a bunch of different places, let's instead standardize it in the CX_PLATFORM global object so subroutines can just reference it.

## Testing

CI

## Documentation

No.

### DIFF
--- a/src/platform/datapath_winkernel.c
+++ b/src/platform/datapath_winkernel.c
@@ -600,24 +600,9 @@ CxPlatDataPathQuerySockoptSupport(
 
     } while (FALSE);
 
-    do {
-        RTL_OSVERSIONINFOW osInfo;
-        RtlZeroMemory(&osInfo, sizeof(osInfo));
-        osInfo.dwOSVersionInfoSize = sizeof(osInfo);
-        NTSTATUS status = RtlGetVersion(&osInfo);
-        if (NT_SUCCESS(status)) {
-            DWORD BuildNumber = osInfo.dwBuildNumber;
-            //
-            // Some USO/URO bug blocks TTL feature support on Windows Server 2022.
-            //
-            if (BuildNumber == 20348) {
-                break;
-            }
-        } else {
-            break;
-        }
+    if (CxPlatform.dwBuildNumber != 20348)
         Datapath->Features |= CXPLAT_DATAPATH_FEATURE_TTL;
-    } while (FALSE);
+    }
 
 Error:
 
--- a/src/platform/datapath_winkernel.c
+++ b/src/platform/datapath_winkernel.c
@@ -600,6 +600,9 @@ CxPlatDataPathQuerySockoptSupport(
 
     } while (FALSE);
 
+    //
+    // Some USO/URO bug blocks TTL feature support on Windows Server 2022.
+    //
     if (CxPlatform.dwBuildNumber != 20348) {
         Datapath->Features |= CXPLAT_DATAPATH_FEATURE_TTL;
     }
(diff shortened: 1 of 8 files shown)


src/platform/datapath_winuser.c
```diff
@@ -706,27 +700,8 @@ CxPlatDataPathQuerySockoptSupport(
...
-    //
-    HMODULE NtDllHandle = LoadLibraryA("ntdll.dll");
-    if (NtDllHandle) {
-        FuncRtlGetVersion VersionFunc = (FuncRtlGetVersion)GetProcAddress(NtDllHandle, "RtlGetVersion");
-        if (VersionFunc) {
-            RTL_OSVERSIONINFOW VersionInfo = {0};
-            VersionInfo.dwOSVersionInfoSize = sizeof(VersionInfo);
-            if ((*VersionFunc)(&VersionInfo) == 0) {
-                //
-                // Some USO/URO bug blocks TTL feature support on Windows Server 2022.
```
guhetier, line 722: I think this comment (or an improved version of it) should be kept.
Otherwise, 20348 is just a magic number and soon, nobody will know why TTL is disabled on that build.

src/platform/datapath_winkernel.c
```diff
@@ -600,24 +600,9 @@ CxPlatDataPathQuerySockoptSupport(
...
 
-    do {
-        RTL_OSVERSIONINFOW osInfo;
-        RtlZeroMemory(&osInfo, sizeof(osInfo));
-        osInfo.dwOSVersionInfoSize = sizeof(osInfo);
-        NTSTATUS status = RtlGetVersion(&osInfo);
-        if (NT_SUCCESS(status)) {
-            DWORD BuildNumber = osInfo.dwBuildNumber;
-            //
-            // Some USO/URO bug blocks TTL feature support on Windows Server 2022.
```
guhetier, line 611: Same remark here for the comment.

src/platform/datapath_winuser.c
ProjectsByJackHe: Agreed, I will add back those comments in the next iteration.

src/platform/datapath_winkernel.c
ProjectsByJackHe: Agreed, I will add back those comments in the next iteration.

src/platform/platform_winkernel.c
```diff
@@ -120,6 +120,17 @@ CxPlatInitialize(
...
 
+    RTL_OSVERSIONINFOW osInfo;
+    RtlZeroMemory(&osInfo, sizeof(osInfo));
+    osInfo.dwOSVersionInfoSize = sizeof(osInfo);
+    NTSTATUS status = RtlGetVersion(&osInfo);
+    if (NT_SUCCESS(status)) {
+        DWORD BuildNumber = osInfo.dwBuildNumber;
+        CxPlatform.dwBuildNumber = BuildNumber;
+    } else {
+        CXPLAT_DBG_ASSERT(FALSE); // TODO: Is the assert here enough or is there an appropritae QUIC_STATUS we return?
```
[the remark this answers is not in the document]
ProjectsByJackHe: Yeah, this is inside the Init routine, and so if these steps fail, I am not sure if we should fail the entire routine, or silently tell the caller through QUIC_STATUS that we weren't able to fetch the version info. 

My bias is if we couldn't fetch from NTDLL for some reason, then something else is completely broken and we should not be silent about it.

nibanks (approved):

PR #4743, opened 2025-01-13
Implement Error and ErrorCode for Rust

youyuanwu:
## Description
Rust Api typically returns Result<T,Error> where Error contains the error details of the api failure.
This PR implements ErrorCode enum type which can be easily converted from FFI error numbers, and Error type which produces decent error message when printed or formated. For example:
```rs
        let err = Error::new(ErrorCode::QUIC_ERROR_ABORTED);
        assert_eq!(format!("{err}"), "QUIC_ERROR_ABORTED (0x80004004)");
```
Changes all Apis to return Result<T,Error> instead of Result<T,u32> and removed all redundant code.

## Testing
Added unit tests for error code conversion and formating.

## Documentation
NA

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,0 +1,210 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use crate::ffi::QUIC_ERROR;
+
+/// Defines quic error code enum and its conversion
+/// to raw type using macro.
+macro_rules! define_quic_error_code{
+  ($( $code1:ident ),*) =>{
+    /// Enum of quic status codes.
+    #[allow(non_camel_case_types)]
+    #[derive(Debug, Clone, PartialEq)]
+    #[repr(u32)]
+    pub enum ErrorCode {
+      $(
+        $code1 = crate::ffi::$code1 as u32,
+      )*
+    }
+
+    /// Convert from ffi type to enum.
+    /// Conversion failes when the status value is not a quic status.
+    impl std::convert::TryFrom<crate::ffi::QUIC_ERROR> for ErrorCode {
+      type Error = &'static str;
+      fn try_from(value: crate::ffi::QUIC_ERROR) -> Result<Self, Self::Error> {
+          match value {
+              $(
+                crate::ffi::$code1 => Ok(Self::$code1),
+              )*
+              _ => Err("Unknown QUIC_ERROR")
+          }
+      }
+    }
+  }
+}
+
+// defines all quic error codes.
+define_quic_error_code!(
+    QUIC_ERROR_SUCCESS,
+    QUIC_ERROR_PENDING,
+    QUIC_ERROR_CONTINUE,
+    QUIC_ERROR_OUT_OF_MEMORY,
+    QUIC_ERROR_INVALID_PARAMETER,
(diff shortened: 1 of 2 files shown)


src/error.rs
```diff
@@ -0,0 +1,210 @@
...
+}
+
+/// The display message is in the same format as error in windows crate.
+impl core::fmt::Display for Error {
+    fn fmt(&self, fmt: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        let str_code = match self.try_as_error_code() {
+            Ok(c) => Some(c),
+            Err(_) => None,
+        };
+        match str_code {
```
nibanks, line 174: I prefer to print the numeric values as hex please.

youyuanwu: Done

nibanks (approved):

src/error.rs
```diff
@@ -0,0 +1,210 @@
...
+    /// Temp api to use in manually written ffi code which is going to be
+    /// removed and replaced by auto generated ffi code.
+    /// This api will be replaced by from_raw.
+    pub fn from_u32(ec: u32) -> Self {
+        Self(ec as QUIC_ERROR)
+    }
+
+    /// Return Err if the error is considered a failure.
+    /// Ok includes both "no error" and "pending" status codes.
+    pub fn ok(self) -> Result<(), Self> {
```
guhetier, line 118: Just a question (I'm not that familiar with idiomatic error handling, so feel free to ignore)

Is it usual to use `ok` for converting an error type to a `Result`?
I am familiar with it to convert a `Result` to an `Option`, which basically means "it is ok to fail here". This is somewhat similar to what happens here ("this error is actually not an error").

But it leads to some strange patterns (`if let Err(e) = err.ok() { panic!(...)`). I am wondering if a different name could help, either for the method or for the type (`ResultCode` instead of `Error` since it is not always an error?).

youyuanwu: I need to add another struct type QuicError(pub QUIC_ERROR) and implement ok() on QuicError which returns the Error{} type. This maybe more confusing since we will have ErrorCode enum, Error struct, and QuicError struct. and it is hard to tell which to use.
I cannot add ok() on ErrorCode because it is not directly convertable from u32 and some u32 are not valid enums. I cannot add ok() to QUIC_ERROR because it is a type alias to u32/i32. 
U can create a windows_core::Error with S_OK. U might also able to create std::io::Error from os errno 0 which is success (not very sure).
So there are 2 options: 1. add QuicError struct 2. keep like this or make ok() pub(crate) or private.

guhetier: I agree with you, adding another type could be confusing.
I was talking at most about potentially renaming to avoid `ok`, which seems confusing to me. But I don't have a good name to offer, so please ignore this if you this the current version is good.

youyuanwu: I think in future PR I can change the ok() api to ok_from_raw(ec: QUIC_ERROR) -> Result<(),Error>) for dedicated conversion from ffi status returns. But that requires migrating to use ffi bindings that is still in my other branch.
I would like to merge the code as is and change it later. @nibanks

nibanks (approved):

PR #4756, opened 2025-01-20
Rust: Migrate to use ffi QUIC_API_TABLE

youyuanwu:
## Description
This is the beginning of a series of PRs to migrate from manually written code to use auto generated Rust ffi bindings, and make Rust APIs safe in default, and handles ffi internally in the crate.

This PR changes the following:
* Migrate to use ffi QUIC_API_TABLE from manually written ApiTable.
* Change Error::from_u32 and Error::ok to Error::ok_from_raw for dedicated error conversion from ffi function returns.
* Reorganize apis like get_param, set_param and simplify other apis built on top of them.
* Refactor the static api table initialization.
* Add unified raw handle functions for all wrapper handle types. All wrappers support from_raw(HQUIC), into_raw()->HQUIC, as_raw()->HQUIC, close() and Drop. This allows user to use unsafe ffi apis directly.
* Refactored the test code into test mod under test cfg section.
* Removed types that are no longer needed or replaced by ffi mod.

The migration process will break the compatibility of the APIs but this is ok as msquic rust crate is experimental.

## Testing
Existing rust test.
More tests will be added once we establish the Rust safe APIs.

## Documentation
NA

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1532,7 +1532,7 @@ impl Connection {
         }
     }
 
-    pub fn get_stats(&self) -> QuicStatistics {
+    pub fn get_stats(&self) -> Result<QuicStatistics, Error> {
         let mut stat_buffer: [u8; std::mem::size_of::<QuicStatistics>()] =
             [0; std::mem::size_of::<QuicStatistics>()];
         let stat_size_mut = std::mem::size_of::<QuicStatistics>();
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -1571,14 +1537,14 @@ impl Connection {
...
                 self.handle,
                 PARAM_CONN_STATISTICS,
                 (&stat_size_mut) as *const usize as *const u32 as *mut u32,
-                stat_buffer.as_mut_ptr() as *const c_void,
+                stat_buffer.as_mut_ptr() as *mut c_void,
             )
-        };
-
+        }
+        .expect("fail to get stats");
```
guhetier, line 1547: For `get_perf`, you chose to return an error, while you assert here.
Is there a reason for the difference? Or should we be consistent?

guhetier:
Thanks for following up on the error type! :)

src/lib.rs
youyuanwu: The original API does not return error so I kept the same. It maybe infallible? Or I can change it.

src/lib.rs
```diff
@@ -1822,172 +1815,175 @@ impl Stream {
...
+
+    extern "C" fn test_stream_callback(
+        stream: HQUIC,
+        _context: *mut c_void,
+        event: &StreamEvent,
+    ) -> u32 {
+        match event.event_type {
+            crate::STREAM_EVENT_START_COMPLETE => {
+                println!("Stream start complete 0x{:x}", unsafe {
+                    event.payload.start_complete.status
```
nibanks, line 1905: Not specific to this PR, but why is accessing `event.payload.start_complete.status` considered _unsafe_?

src/lib.rs
youyuanwu: Made all of them return Result.

src/lib.rs
youyuanwu: payload is a union type. Rust union is for ffi only (without metadata), and the compiler has no way to check if the access is safe. However, Rust enum has metadata embedded in it so enum is safe. I have another branch that does the union to enum conversion (in unsafe code).

nibanks (approved):

youyuanwu:
Need to reopen PR to retry CI.

PR #4757, opened 2025-01-21
Add support for setting TypeOfService/TrafficClass on connections

### LINKED ISSUE #4767: Support Specifying DSCP for QUIC Connections

For certain scenarios, it's advantageous for the application to configure the DCSP to indicate to the network that this connection should either be prioritized or deprioritized.

### Proposed solution

We need to expose a parameter to set the DSCP value for a connection.

### Additional context

_No response_

anrossi:
## Description

Resolves #4767. Add support for applications to set the Type of Service/Traffic Class field all traffic sent from a connection.

## Testing

Updated

## Documentation

Updated the settings document with the new get/set param.


src/platform/datapath_winkernel.c
```diff
@@ -1552,6 +1600,47 @@ SocketCreateUdp(
         }
     }
 
+    if ((Datapath->Features & CXPLAT_DATAPATH_FEATURE_TYPE_OF_SERVICE) &&
+        Config->TypeOfService != 0) {
+        Option = Config->TypeOfService;
+        Status =
+            CxPlatDataPathSetControlSocket(
```
nibanks, line 1607: We're going to have to use ancillary data instead, because of the server side. Server side shares the socket and some datagrams will be different traffic class.

src/core/connection.c
```diff
@@ -7207,27 +7228,55 @@ QuicConnParamGet(
...
+
         //
         // Tell app how much buffer we copied.
         //
         *BufferLength = Connection->OrigDestCID->Length;
+
+        Status = QUIC_STATUS_SUCCESS;
+        break;
+
+     case QUIC_PARAM_CONN_DSCP:
```
anrossi, line 7261: Should GetParam return the last-set DSCP value, or the most recently received DSCP value?
Or should getting the received DSCP value be a separate, get-only param?

src/platform/datapath_winkernel.c
```diff
@@ -2923,6 +2973,18 @@ SocketSend(
         *(PINT)WSA_CMSG_DATA(CMsg) = SendData->ECN;
     }
 
+    if (Binding->Datapath->Features & CXPLAT_DATAPATH_FEATURE_DSCP) {
+        CMsg = (PWSACMSGHDR)&CMsgBuffer[CMsgLen];
+        CMsgLen += WSA_CMSG_SPACE(sizeof(INT));
+        CMsg->cmsg_level =
+            Route->LocalAddress.si_family == QUIC_ADDRESS_FAMILY_INET ?
```
anrossi, line 2980: LocalAddress might be the wrong address to use; if trying to send to an IPv4 address using a dual-mode socket, you need to set the IPv4 IP_TOS, instead of the IPV6_TCLASS

src/core/binding.c
```diff
@@ -803,7 +803,7 @@ QuicBindingProcessStatelessOperation(
         Binding,
         OperationType);
 
-    CXPLAT_SEND_CONFIG SendConfig = { RecvPacket->Route, 0, CXPLAT_ECN_NON_ECT, 0 };
+    CXPLAT_SEND_CONFIG SendConfig = { RecvPacket->Route, 0, CXPLAT_ECN_NON_ECT, 0, 0 };
```
nibanks, line 806: Should we have a `#define` for zero here instead of using a magic number?

src/core/connection.c
```diff
@@ -6623,6 +6623,27 @@ QuicConnParamSet(
         return QUIC_STATUS_SUCCESS;
     }
 
+    case QUIC_PARAM_CONN_DSCP:
+    {
```
nibanks, line 6627:
```suggestion
    case QUIC_PARAM_CONN_DSCP: {

```

src/core/connection.c
```diff
@@ -6623,6 +6623,27 @@ QuicConnParamSet(
...
+        if (BufferLength != sizeof(uint8_t) || Buffer == NULL) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+
+        uint8_t DSCP = 0;
+        CxPlatCopyMemory(&DSCP, Buffer, BufferLength);
```
nibanks, line 6634: Simplify?
```suggestion
        uint8_t DSCP = *Buffer;
```

src/core/connection.c
nibanks: It's probably better to support querying the last value set by the app. So, then we would need two parameters, one for send and one (get-only) for receive. But let's save receive for a future PR?

src/inc/msquic.h
```diff
@@ -923,6 +923,7 @@ typedef struct QUIC_SCHANNEL_CREDENTIAL_ATTRIBUTE_W {
 #define QUIC_PARAM_CONN_STATISTICS_V2                   0x05000016  // QUIC_STATISTICS_V2
 #define QUIC_PARAM_CONN_STATISTICS_V2_PLAT              0x05000017  // QUIC_STATISTICS_V2
 #define QUIC_PARAM_CONN_ORIG_DEST_CID                   0x05000018  // uint8_t[]
+#define QUIC_PARAM_CONN_DSCP                            0x50000019  // uint8_t
```
nibanks, line 926: Maybe call it `_SEND_DSCP`?
```suggestion
#define QUIC_PARAM_CONN_SEND_DSCP                       0x50000019  // uint8_t
```

src/inc/quic_datapath.h
```diff
@@ -444,6 +455,7 @@ CxPlatDataPathUpdateConfig(
 #define CXPLAT_DATAPATH_FEATURE_TCP                   0x0020
 #define CXPLAT_DATAPATH_FEATURE_RAW                   0x0040
 #define CXPLAT_DATAPATH_FEATURE_TTL                   0x0080
+#define CXPLAT_DATAPATH_FEATURE_DSCP                  0x0100
```
nibanks, line 458: Will we need separate features for send/receive support?

src/core/connection.c
anrossi: Won't the simplification cause a crash if the buffer is unaligned on ARM?

src/inc/msquic.h
anrossi: I was thinking of the same pattern that socket options use where the send one has no prefix, but the receive one has a prefix. So when we do the last-received DSCP value, it would be `QUIC_PARAM_CONN_RECV_DSCP`

src/inc/quic_datapath.h
anrossi: I *think* the receive support already exists in all stacks, it's just the send support that's missing on some.

src/platform/datapath_winuser.c
```diff
@@ -1672,7 +1708,7 @@ SocketCreateUdp(
             setsockopt(
                 SocketProc->Socket,
                 IPPROTO_IP,
-                IP_ECN,
+                IP_RECVTOS,
```
anrossi, line 1711: I think this might be causing test failures because packets aren't getting received in some tests since adding this change.

src/platform/datapath_winuser.c
```diff
@@ -3464,9 +3500,9 @@ CxPlatDataPathUdpRecvComplete(
                     LocalAddr->Ipv4.sin_port = SocketProc->Parent->LocalAddress.Ipv6.sin6_port;
                     LocalAddr->Ipv6.sin6_scope_id = PktInfo->ipi_ifindex;
                     FoundLocalAddr = TRUE;
-                } else if (CMsg->cmsg_type == IP_ECN) {
-                    ECN = *(PINT)WSA_CMSG_DATA(CMsg);
-                    CXPLAT_DBG_ASSERT(ECN < UINT8_MAX);
+                } else if (CMsg->cmsg_type == IP_TOS) {
```
anrossi, line 3503: Same here; I think this change might be causing receives to fail.

src/core/connection.c
anrossi: Sure, receive can be a future PR

src/inc/quic_datapath.h
anrossi: Given that receive support seems to not be working on Windows in CI, I am going to make separate features for send and receive.

src/core/connection.c
```diff
@@ -7207,27 +7228,55 @@ QuicConnParamGet(
     }
 
     case QUIC_PARAM_CONN_ORIG_DEST_CID:
+
```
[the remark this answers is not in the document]
anrossi: I just felt the newlines needed to be added for consistency and they wouldn't be added any other time

src/platform/datapath_raw_socket.c
```diff
@@ -366,7 +366,7 @@ CxPlatDpRawParseIPv6(
     } VersionClassEcnFlow;
     VersionClassEcnFlow.Value = CxPlatByteSwapUint32(IP->VersionClassEcnFlow);
 
-    Packet->TypeOfService = (uint8_t)VersionClassEcnFlow.EcnField;
+    Packet->TypeOfService = ((uint8_t)VersionClassEcnFlow.EcnField) | (uint8_t)(VersionClassEcnFlow.Class << 2);
```
nibanks, line 369: I wonder if it makes a difference (in terms of instructions) to just cast the result of the OR to a `uint8_t` instead of casting both first...

src/platform/datapath_winkernel.c
```diff
@@ -471,7 +471,7 @@ CxPlatDataPathQuerySockoptSupport(
         Datapath->WskProviderNpi.Dispatch->
         WskSocket(
             Datapath->WskProviderNpi.Client,
-            AF_INET,
+            AF_INET6,
```
nibanks, line 474: Why did you make this change?

src/platform/datapath_winkernel.c
```diff
@@ -2923,6 +2973,20 @@ SocketSend(
...
 
+    if (Binding->Datapath->Features & CXPLAT_DATAPATH_FEATURE_SEND_DSCP) {
+        CMsg = (PWSACMSGHDR)&CMsgBuffer[CMsgLen];
+        CMsgLen += WSA_CMSG_SPACE(sizeof(INT));
+        CMsg->cmsg_level =
+            Route->LocalAddress.si_family == QUIC_ADDRESS_FAMILY_INET ?
+                IPPROTO_IP : IPPROTO_IPV6;
+        CMsg->cmsg_type =
+            Route->LocalAddress.si_family == QUIC_ADDRESS_FAMILY_INET ?
+                IP_TOS : IPV6_TCLASS;
```
nibanks, line 2984: Can we do the `if` check only once here?
```suggestion
        if (Route->LocalAddress.si_family == QUIC_ADDRESS_FAMILY_INET) {
            CMsg->cmsg_level = IPPROTO_IP;
            CMsg->cmsg_type = IP_TOS;
        } else { // INET6
            CMsg->cmsg_level = IPPROTO_IPV6;
            CMsg->cmsg_type = IPV6_TCLASS;
        }
```

src/platform/datapath_winkernel.c
```diff
@@ -2923,6 +2973,20 @@ SocketSend(
         *(PINT)WSA_CMSG_DATA(CMsg) = SendData->ECN;
     }
 
+    if (Binding->Datapath->Features & CXPLAT_DATAPATH_FEATURE_SEND_DSCP) {
```
nibanks, line 2976: The lines above are already trying to set `IP_ECN`. Should we be setting both that and this?

src/platform/datapath_winuser.c
```diff
@@ -4491,6 +4528,16 @@ CxPlatSocketSendInline(
...
 
+        if (Socket->Datapath->Features & CXPLAT_DATAPATH_FEATURE_SEND_DSCP) {
+            WSAMhdr.Control.len += WSA_CMSG_SPACE(sizeof(INT));
+            CMsg = WSA_CMSG_NXTHDR(&WSAMhdr, CMsg);
+            CXPLAT_DBG_ASSERT(CMsg != NULL);
+            CMsg->cmsg_level = IPPROTO_IP;
+            CMsg->cmsg_type = IP_TOS;
+            CMsg->cmsg_len = WSA_CMSG_LEN(sizeof(INT));
+            *(PINT)WSA_CMSG_DATA(CMsg) = SendData->ECN | (SendData->DSCP << 2);
+        }
```
nibanks, line 4539: This is something that came up before for ECN, but I think it might be worth doing now: Don't actually set the ancillary data at all, if the final result is zero, which would just match what the stack already puts in the packet. It just results in extra work for no net change in behavior....

src/core/connection.c
```diff
@@ -7207,27 +7232,55 @@ QuicConnParamGet(
...
+     case QUIC_PARAM_CONN_SEND_DSCP:
+
+        if (*BufferLength < sizeof(uint8_t)) {
+            Status = QUIC_STATUS_BUFFER_TOO_SMALL;
+            *BufferLength = sizeof(uint8_t);
+            break;
+        }
+
+        if (Buffer == NULL) {
```
mtfriesen, line 7273: I see you're just following the pattern in this function, but IMO if a [trusted] BufferLength parameter is > 0, it should imply the buffer pointer is not null.

src/platform/datapath_winkernel.c
anrossi: IPv4 silently fails on previous versions of Windows, while IPv6 fails immediately. 
I could open a separate IPv6 socket just for the DSCP feature test, but it's a lot of extra boilerplate code

src/platform/datapath_winkernel.c
anrossi: I was wondering about that. I was thinking about only setting IP_ECN when `CXPLAT_DATAPATH_FEATURE_SEND_DSCP` is not supported and only setting DSCP+ECN when DSCP is supported

src/core/connection.c
anrossi: Right, but nothing has validated prior to this function that the buffer is not NULL (discovered via testing). So the validation must happen here.

src/platform/datapath_raw_socket.c
anrossi: GCC outputs identical code for both versions; MSVC outputs the same number of instructions (but slightly different ordering); Clang outputs different instructions and the single cast has 2 fewer instructions
see here for my experiment https://godbolt.org/z/YW5o75cv3

src/platform/datapath_winkernel.c
anrossi: I'll clean it up for readability, but I'll mention the compiler is smart enough to optimize this to one comparison.

src/core/connection.c
mtfriesen: From an SDL [security] perspective it might be worthwhile relocating all these scattered checks to the top of the function: `if (BufferLength > 0 && Buffer == NULL) return ERROR_INVALID_PARAM;` and make it more difficult to have errors of omission in each case statement.

anrossi: I agree, and was expecting to find that already to be the case. However, the API contract allows calling the API with a NULL buffer and the buffer length field set to 0 to obtain the amount of buffer needed to receive the parameter. I don't think that is compatible with an entry-point check of `if (BufferLength > 0 && Buffer == NULL)`

mtfriesen: Ahhh, this is where so many APIs take subtly different solutions. This is nearly orthogonal to your PR, so I'll move on.

nibanks (approved):

nibanks (approved):
If the CI passes, ship it!

Issue #4747, opened 2025-01-17
Support Scatter Gather Buffers on the client to copy data directly to application provided buffers


srsubra:
### Describe the feature you'd like supported

Currently, MsQuic with XDP provides the application level callbacks using internal buffers. This requires a mem copy from the MsQuic Internal buffers to the target application buffer. For large blob, allocating this memory and performing a memory copy 

### Proposed solution

Provide an option where the user can specify the Target buffer address where the data needs to be copied to and MsQuic core layer can automatically copy the data to the target buffer.

guhetier:
Hi!

Here is how we see how the implementation would look like:

- Probably a new flag "QUIC_STREAM_OPEN_FLAG_USE_EXTERNAL_RECEIVE_BUFFERS" to opt-in the feature, when opening the stream
- A new API function along the lines of "StreamProvideReceiveBuffers(stream, nb_buff, QUIC_BUFFERS* buff)"
    - The user can call this on an open stream to provide a list of buffers
    - The user would call it before starting the stream and may call it again later (on a receive callback, etc...) if it needs to provide more buffers

MsQuic would write the received data in the buffers in the order they are provided. A buffer would never be internally allocated or re-used: if MsQuic runs out of provided buffer space for receiving data, the stream would be failed.

The user would need to ensure the buffer remain valid until:
- either the stream is closed, and no more data can be received
- or a "receive" notification indicates all the bytes of the buffer
    - the user needs to keep track of the total number of bytes to know this
    - a "receive" notification could indicate only part of a buffer, multiple buffers, etc... depending on how the data is received

Does it sound like this would work for you?

nibanks:
> Probably a new flag "QUIC_STREAM_OPEN_FLAG_USE_EXTERNAL_RECEIVE_BUFFERS" to opt-in the feature, when opening the stream

I'm not convinced we need this for a couple reasons:

1. This doesn't work for peer started streams. You never _open_ them locally.
2. I think it's enough to say calling `StreamProvideReceiveBuffers` opts into the feature. You don't need two signals for the same thing.

So, I'm leaning towards only the new API.

guhetier:
Good point for the peer started stream.

I was considering the flag because we currently initialize buffer management during the stream initialization - we may have to move the recvBuffer initialization later or change its mode on the fly.

For peer started stream, we may have to look at it closer: I think the stream is started before the user is notified, so data could be buffered before external buffers are provided (which we would then need to copy)

[...]

PR #4758, opened 2025-01-22
Add a mode to the receive buffer to handle external buffers

guhetier:
## Description

As part of #4747, the receive buffer needs a mode to operate on buffers provided by the application.
This PR defines a new `QUIC_RECV_BUF_MODE_EXTERNAL` mode.

- `QUIC_RECV_CHUNK`s now have a pointer to the actual buffer instead of an end-of-struct array
    - Except for the EXTERNAL mode, the buffer is still part of the same allocation as the chunk
    - For EXTERNAL mode, buffers are provided by the application and chunks point to them
- The EXTERNAL mode writes only once to the provided buffers and discard them when they are drained. Buffers are never re-used.
- Reads and writes can span through multiple chunks. They are never treated as circular buffers.
- The EXTERNAL mode never allocates / re-allocate buffers
    - The virtual buffer size is set to the sum of the provided buffer sizes to avoid receiving more data than we can buffer
    - The virtual buffer size can now decrease when data is drained - but the maximum byte offset that can be received still only increases.

## Testing

Unit tests have been added.
System tests have been added.

## Documentation

Documentation added as code comments.
Out of line documentation will be added in a follow up PR.


src/core/recv_buffer.h
```diff
@@ -22,9 +25,19 @@ typedef struct QUIC_RECV_CHUNK {
     CXPLAT_LIST_ENTRY Link;         // Link in the list of chunks.
     uint32_t AllocLength : 31;      // Allocation size of Buffer
     uint32_t ExternalReference : 1; // Indicates the buffer is being used externally.
-    uint8_t Buffer[0];
+    uint8_t *Buffer;                // Pointer to the buffer itself. Doesn't need to be freed independently:
```
nibanks, line 28: Somewhat independent of this PR, I am wondering if we could/should annotate this pointer:
```suggestion
    _Field_size_bytes_(AllocLength)
    uint8_t *Buffer;                // Pointer to the buffer itself. Doesn't need to be freed independently:
```

src/core/recv_buffer.c
```diff
@@ -170,10 +182,74 @@ QuicRecvBufferIncreaseVirtualBufferLength(
...
+    _Inout_ CXPLAT_LIST_ENTRY* /* QUIC_RECV_CHUNKS */ Chunks
+    )
+{
+    CXPLAT_DBG_ASSERT(!CxPlatListIsEmpty(Chunks));
+
+    //
+    // External chunks can be provided only if already in external mode, or if
+    // nothing has been written to the buffer yet.
+    //
+    uint64_t rangeMax;
```
nibanks, line 204: FYI, our style capitalizes all variables and fields.
```suggestion
    uint64_t RangeMax;
```
Please update for everything else too. Thanks!

src/core/recv_buffer.c
```diff
@@ -170,10 +182,74 @@ QuicRecvBufferIncreaseVirtualBufferLength(
...
+        QUIC_RECV_CHUNK* firstChunk = CXPLAT_CONTAINING_RECORD(Chunks->Flink, QUIC_RECV_CHUNK, Link);
+        RecvBuffer->Capacity = firstChunk->AllocLength;
+    }
+
+    uint32_t totalAdditionalLength = 0;
+    for (CXPLAT_LIST_ENTRY* Link = Chunks->Flink;
+         Link != Chunks;
+         Link = Link->Flink) {
+        QUIC_RECV_CHUNK* Chunk = CXPLAT_CONTAINING_RECORD(Link, QUIC_RECV_CHUNK, Link);
+        // TODO guhetier: Must deal with overflow?
```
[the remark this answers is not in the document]
nibanks: FYI, you can overflow both here and on the line below: `RecvBuffer->VirtualBufferLength += totalAdditionalLength;`. Both need to be checked.

src/core/recv_buffer.c
```diff
@@ -186,6 +262,7 @@ QuicRecvBufferResize(
     _In_ uint32_t TargetBufferLength
     )
 {
+    CXPLAT_DBG_ASSERT(RecvBuffer->RecvMode != QUIC_RECV_BUF_MODE_EXTERNAL); // Should never resize in external mode
```
nibanks, line 265: FYI, we also have `CXPLAT_DBG_ASSERTMSG` which could be used here instead of the comment:
```suggestion
    CXPLAT_DBG_ASSERTMSG(RecvBuffer->RecvMode != QUIC_RECV_BUF_MODE_EXTERNAL, "Should never resize in external mode");
```
But you don't have to use it. We don't use it that much.

src/core/recv_buffer.c
```diff
@@ -573,24 +660,28 @@ QuicRecvBufferWrite(
     // N.B. We do this before updating the written ranges below so we don't have
     // to support rolling back those changes on the possible allocation failure
     // here.
+    // This is skiped in external mode since the entire virtual length is
```
nibanks, line 663:
```suggestion
    // This is skipped in external mode since the entire virtual length is
```

src/core/recv_buffer.c
```diff
@@ -800,6 +901,36 @@ QuicRecvBufferRead(
...
+                RecvBuffer->Chunks.Flink,
+                QUIC_RECV_CHUNK,
+                Link);
+        Buffers[0].Buffer = Chunk->Buffer + RecvBuffer->ReadStart;
+        Buffers[0].Length = RecvBuffer->ReadLength;
+        remainingDataToRead -= RecvBuffer->ReadLength;
+
+        // Continue reading from the next chunks until we run out of buffers or data
+        for (uint32_t i = 0; i < *BufferCount && remainingDataToRead > 0; ++i)
+        {
```
nibanks, line 919: style nit:
```suggestion
        for (uint32_t i = 0; i < *BufferCount && remainingDataToRead > 0; ++i) {
```

src/core/recv_buffer.c
```diff
@@ -977,13 +1124,16 @@ QuicRecvBufferDrain(
     CXPLAT_DBG_ASSERT(FirstRange->Low == 0);
     do {
         BOOLEAN PartialDrain = (uint64_t)RecvBuffer->ReadLength > DrainLength;
-        if (PartialDrain ||
-            (QuicRangeSize(&RecvBuffer->WrittenRanges) > 1 &&
-             RecvBuffer->BaseOffset + RecvBuffer->ReadLength == FirstRange->Count)) {
+        // TODO guhetier: Seems wrong, if the end of the chunk aligns with the end of the first range, we still need a full drain
```
nibanks, line 1127: For stuff like this, can you update the unittest to verify?

src/core/recv_buffer.c
```diff
@@ -47,6 +47,19 @@
 #include "recv_buffer.c.clog.h"
 #endif
 
+_IRQL_requires_max_(DISPATCH_LEVEL)
+void
+QuicRecvChunkInitialize(
+    _Inout_ QUIC_RECV_CHUNK* Chunk,
+    _In_ uint32_t AllocLength,
+    _In_ uint8_t* Buffer
```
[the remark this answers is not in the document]
guhetier: Yes. I also need to add the corresponding annotations on the chunk struct itself.

src/core/recv_buffer.c
guhetier: That's the plan!

src/core/recv_buffer.c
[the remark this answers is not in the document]
guhetier: Added overflow cheks.

src/core/recv_buffer.c
guhetier: The fix was not trivial here - hopefully I got it right, at least, old and new tests are passing.
Multiple variables are being used in subtly different ways depending on the mode (not sure if it is intentional or not), making those few lines much trickier than expected.
In particular, this code works for Single/Circular mode only because `ReadLength` is actually the readable size on the last chunk for these modes, not the first as documented.

src/core/recv_buffer.c
```diff
@@ -164,14 +181,66 @@ QuicRecvBufferHasUnreadData(
 }
 
 _IRQL_requires_max_(DISPATCH_LEVEL)
-void
-QuicRecvBufferIncreaseVirtualBufferLength(
+BOOLEAN
+QuicRecvBufferTryIncreaseVirtualBufferLength(
```
nibanks, line 185: Missed one caller of the old name:
```
D:\a\msquic\msquic\src\core\stream_recv.c(805,17): warning C4013: 'QuicRecvBufferIncreaseVirtualBufferLength' undefined; assuming extern returning int [D:\a\msquic\msquic\build\windows\x86_openssl\src\core\core.vcxproj]
```

src/core/recv_buffer.h
```diff
@@ -122,12 +140,24 @@ QuicRecvBufferHasUnreadData(
 // Changes the buffer's virtual buffer length.
 //
 _IRQL_requires_max_(DISPATCH_LEVEL)
-void
-QuicRecvBufferIncreaseVirtualBufferLength(
+BOOLEAN
```
nibanks, line 143: Do we really need to return the BOOLEAN? If no one uses it, let's just drop it.

src/core/recv_buffer.c
```diff
@@ -164,14 +181,66 @@ QuicRecvBufferHasUnreadData(
...
+         Link != Chunks;
+         Link = Link->Flink) {
+        QUIC_RECV_CHUNK* Chunk = CXPLAT_CONTAINING_RECORD(Link, QUIC_RECV_CHUNK, Link);
+        NewBufferLength += Chunk->AllocLength;
+        if (NewBufferLength < Chunk->AllocLength) {
+            //
+            // Overflow: we can't handle that much buffer space.
+            //
+            return QUIC_STATUS_INVALID_PARAMETER;
+        }
```
nibanks, line 220: Why are you limiting the chunks to less than the virtual length, instead of just increasing the length? Also, if we keep it, I'd just optimize this to a single check after summing everything instead of doing a check for each chunk.

src/core/recv_buffer.c
```diff
@@ -186,6 +255,7 @@ QuicRecvBufferResize(
     _In_ uint32_t TargetBufferLength
     )
 {
+    CXPLAT_DBG_ASSERTMSG(RecvBuffer->RecvMode != QUIC_RECV_BUF_MODE_EXTERNAL, "Should never resize in External mode");
```
nibanks, line 258: Generally, we try to keep code lines to <= 80 characters. Comments can go to 120.
```suggestion
    CXPLAT_DBG_ASSERTMSG(
        RecvBuffer->RecvMode != QUIC_RECV_BUF_MODE_EXTERNAL,
        "Should never resize in External mode");
```

src/core/recv_buffer.c
```diff
@@ -644,12 +733,20 @@ QuicRecvBufferRead(
 {
     CXPLAT_DBG_ASSERT(QuicRangeGetSafe(&RecvBuffer->WrittenRanges, 0) != NULL); // Only fail if you call read before write indicates read ready.
     CXPLAT_DBG_ASSERT(!CxPlatListIsEmpty(&RecvBuffer->Chunks)); // Should always have at least one chunk
+    //
+    // Only MULTIPLE mode allows concurrent reads
```
nibanks, line 737: nit: I'm not sure we use all caps to describe MULTIPLE or EXTERNAL anywhere else in comments. Let's be consistent.

src/core/recv_buffer.h
```diff
@@ -12,19 +12,31 @@ extern "C" {
 typedef enum QUIC_RECV_BUF_MODE {
     QUIC_RECV_BUF_MODE_SINGLE,      // Only one receive with a single contiguous buffer at a time.
     QUIC_RECV_BUF_MODE_CIRCULAR,    // Only one receive that may indicate two contiguous buffers at a time.
-    QUIC_RECV_BUF_MODE_MULTIPLE     // Multiple independent receives that may indicate up to two contiguous buffers at a time.
+    QUIC_RECV_BUF_MODE_MULTIPLE,    // Multiple independent receives that may indicate up to two contiguous buffers at a time.
+    QUIC_RECV_BUF_MODE_EXTERNAL     // Uses memory buffer provided by the app. Only one receive at a time,
```
nibanks, line 16: Do you think the fact that we use the word 'external' to mean the app currently has a reference to a chunk via a receive call might confuse things? One option is to all this `_APP`.

src/core/recv_buffer.h
```diff
@@ -12,19 +12,31 @@ extern "C" {
...
 //
 typedef struct QUIC_RECV_CHUNK {
-    CXPLAT_LIST_ENTRY Link;         // Link in the list of chunks.
-    uint32_t AllocLength : 31;      // Allocation size of Buffer
-    uint32_t ExternalReference : 1; // Indicates the buffer is being used externally.
-    uint8_t Buffer[0];
+    CXPLAT_LIST_ENTRY Link;                     // Link in the list of chunks.
+    uint32_t AllocLength : 31;                  // Allocation size of Buffer
+    uint32_t ExternalReference : 1;             // Indicates the buffer is being used externally.
+    _Field_size_(AllocLength) uint8_t *Buffer;  // Pointer to the buffer itself. Doesn't need to be freed independently:
```
nibanks, line 27: nit: instead of pushing the comments all out farther, just put the annotation on it's own line?

src/core/recv_buffer.h
```diff
@@ -122,12 +140,24 @@ QuicRecvBufferHasUnreadData(
...
+//
+// Provide externally allocated chunks to the buffer.
+// At least one chunk must be provided.
+// Only valid for QUIC_RECV_BUF_MODE_EXTERNAL mode.
+//
+_IRQL_requires_max_(DISPATCH_LEVEL)
+QUIC_STATUS
+QuicRecvBufferProvideChunks(
+    _Inout_ QUIC_RECV_BUFFER* RecvBuffer,
+    _Inout_ CXPLAT_LIST_ENTRY* /* QUIC_RECV_CHUNKS */ Chunks
```
nibanks, line 158: This assumes the caller allocates space and initializes a `QUIC_RECV_CHUNK` struct, instead of an array of `QUIC_BUFFER`s? While that simplifies this interface, I'm not sure that is reasonable.

src/core/unittest/RecvBufferTest.cpp
```diff
@@ -39,7 +46,38 @@ struct RecvBuffer {
                     QUIC_POOL_TEST);
         }
         printf("Initializing: [mode=%u,vlen=%u,alen=%u]\n", RecvMode, VirtualBufferLength, AllocBufferLength);
-        auto Result = QuicRecvBufferInitialize(&RecvBuf, AllocBufferLength, VirtualBufferLength, RecvMode, PreallocChunk);
+
+        auto Result = QUIC_STATUS_SUCCESS;
+        Result = QuicRecvBufferInitialize(&RecvBuf, AllocBufferLength, VirtualBufferLength, RecvMode, PreallocChunk);
```
nibanks, line 51: nit: Why did you split this into two lines?

src/core/unittest/RecvBufferTest.cpp
```diff
@@ -39,7 +46,38 @@ struct RecvBuffer {
...
+            CxPlatListInsertHead(&ChunkList, &Chunk->Link);
+            if (VirtualBufferLength > AllocBufferLength) {
+                auto Chunk2 = (QUIC_RECV_CHUNK *)CXPLAT_ALLOC_NONPAGED(sizeof(QUIC_RECV_CHUNK), QUIC_POOL_RECVBUF);
+                QuicRecvChunkInitialize(Chunk2, VirtualBufferLength - AllocBufferLength, ExternalBuffer + AllocBufferLength);
+                CxPlatListInsertTail(&ChunkList, &Chunk2->Link);
+            }
+            Result = QuicRecvBufferProvideChunks(&RecvBuf, &ChunkList);
+        }
+        else
+        {
```
nibanks, line 77:
```suggestion
        } else {
```

src/core/unittest/RecvBufferTest.cpp
```diff
@@ -100,21 +143,14 @@ struct RecvBuffer {
...
-        int ChunkCount = 1;
-        QUIC_RECV_CHUNK* Chunk =
-            CXPLAT_CONTAINING_RECORD(
-                RecvBuf.Chunks.Flink,
-                QUIC_RECV_CHUNK,
-                Link);
-        ASSERT_EQ(Chunk->ExternalReference, ExternalReferences[0]);
-        while (Chunk->Link.Flink != &RecvBuf.Chunks) {
+
+        int ChunkCount = 0;
```
nibanks, line 147: Just to confirm my understanding, now it's possible for there to be zero chunks, because the app hasn't provided any? And that's why you made this change?

src/core/recv_buffer.c
guhetier: I missed a file when pushing that iteration, when filtering out the changes for the API part. Fixed it.

src/core/recv_buffer.h
guhetier: The return value will be used by the caller - mostly for diagnostic (avoiding confusing logs about increasing a buffer length when it actually stays the same).

src/core/recv_buffer.c
guhetier: This was checking for overflow of the addition above. But thinking about it, since we run in the app process and can trust it to respect the API contract, we don't need that check, overflowing the uint64 would mean allocating more memory than reasonably possible. I'll remove this check.

src/core/unittest/RecvBufferTest.cpp
guhetier: Yes. There are two scenarios where we can have zero chunks:
- the app hasn't provided any (yet).
- all provided chunks have been used and relinquished.

src/core/recv_buffer.h
guhetier: The idea is to expose an API that takes an array of QUIC_BUFFERs as a parameter, convert it inline to a list of QUIC_RECV_CHUNKS (which takes memory allocation and could fail), then call `QuicRecvBufferProvideChunks` in the workqueue.
This way, we can keep `QuicRecvBufferProvideChunks` mostly failure free (the overflow is still possible, but it would be reasonable to kill the connection in that case I think).

I was trying to keep the API and recvbuffer parts separate to simplify the review, but it might be starting to backfire. I might push everything at once so you have the full context.

src/core/api.c
```diff
@@ -658,6 +658,15 @@ MsQuicStreamOpen(
         goto Error;
     }
 
+    if (!!(Flags & QUIC_STREAM_OPEN_FLAG_EXTERNAL_BUFFERS) &&
+        Connection->Settings.StreamMultiReceiveEnabled) {
```
nibanks, line 662: What if they want to use multi-receive for streams that don't use external buffers? I think we can't block that.

src/core/api.c
```diff
@@ -1351,6 +1360,148 @@ MsQuicStreamReceiveComplete(
...
+            goto Error;
+        }
+    }
+
+    //
+    // Allocate a chunk for each buffer, linking them together.
+    // The allocation is done here to make the worker thread task failure free.
+    //
+    for (uint32_t i = 0; i < BufferCount; ++i) {
+        QUIC_RECV_CHUNK* Chunk = CXPLAT_ALLOC_NONPAGED(sizeof(QUIC_RECV_CHUNK), QUIC_POOL_RECVBUF);
```
nibanks, line 1435: Should use a pool allocator (to be stored in the worker, like the others) since this is (a) fixed size and (b) on the datapath.

src/core/api.c
```diff
@@ -1351,6 +1360,148 @@ MsQuicStreamReceiveComplete(
...
+        //
+        // Queue the operation but don't wait for the completion.
+        //
+        QuicConnQueueOper(Connection, Oper);
+    }
+
+    Status = QUIC_STATUS_SUCCESS;
+
+Error:
+    // Cleanup allocated chunks if the operation failed.
```
nibanks, line 1491: nit: comment style

src/core/api.c
```diff
@@ -1351,6 +1360,148 @@ MsQuicStreamReceiveComplete(
...
+    while (!CxPlatListIsEmpty(&ChunkList)) {
+        QUIC_RECV_CHUNK* Chunk = CXPLAT_CONTAINING_RECORD(CxPlatListRemoveHead(&ChunkList), QUIC_RECV_CHUNK, Link);
+        CXPLAT_FREE(Chunk, QUIC_POOL_RECVBUF);
```
nibanks, line 1494:
```suggestion
        CXPLAT_FREE(
            CXPLAT_CONTAINING_RECORD(
                CxPlatListRemoveHead(&ChunkList),
                QUIC_RECV_CHUNK,
                Link),
            QUIC_POOL_RECVBUF);
```

src/core/operation.c
```diff
@@ -119,6 +119,8 @@ QuicOperationFree(
             }
         } else if (ApiCtx->Type == QUIC_API_TYPE_STRM_RECV_SET_ENABLED) {
             QuicStreamRelease(ApiCtx->STRM_RECV_SET_ENABLED.Stream, QUIC_STREAM_REF_OPERATION);
+        } else if (ApiCtx->Type == QUIC_API_TYPE_STRM_PROVIDE_RECV_BUFFERS) {
+            QuicStreamRelease(ApiCtx->STRM_PROVIDE_RECV_BUFFERS.Stream, QUIC_STREAM_REF_OPERATION);
```
nibanks, line 123: What about `ApiCtx->STRM_PROVIDE_RECV_BUFFERS.Chunks`? We need to clean those up too, right?

nibanks: Operations can be canceled before they're ever processed. So, this needs to handle that case.

src/core/stream.h
```diff
@@ -1005,3 +1014,23 @@ QuicStreamRecvSetEnabledState(
...
+//
+// Provide new chunks for the stream receive buffer.
+// Terminate the connection on failure.
+//
+_IRQL_requires_max_(DISPATCH_LEVEL)
+QUIC_STATUS
+QuicStreamProvideRecvBuffers(
+    _In_ QUIC_STREAM* Stream,
+    _Inout_ CXPLAT_LIST_ENTRY* /* QUIC_RECV_CHUNK */ Chunks
+    );
```
nibanks, line 1036: nit: always have a trailing empty newline

src/inc/msquic.h
```diff
@@ -199,6 +199,8 @@ typedef enum QUIC_STREAM_OPEN_FLAGS {
     QUIC_STREAM_OPEN_FLAG_0_RTT             = 0x0002,   // The stream was opened via a 0-RTT packet.
     QUIC_STREAM_OPEN_FLAG_DELAY_ID_FC_UPDATES = 0x0004, // Indicates stream ID flow control limit updates for the
                                                         // connection should be delayed to StreamClose.
+    QUIC_STREAM_OPEN_FLAG_EXTERNAL_BUFFERS = 0x0008,    // No buffer will be allocated for the stream, external buffers
+                                                        // must be provided (see StreamProvideReceiveBuffers)
```
nibanks, line 203:
```suggestion
    QUIC_STREAM_OPEN_FLAG_EXTERNAL_BUFFERS  = 0x0008,   // No buffer will be allocated for the stream, external buffers
                                                        // must be provided (see StreamProvideReceiveBuffers)
```

src/inc/msquic.h
```diff
@@ -1631,6 +1647,8 @@ typedef struct QUIC_API_TABLE {
     QUIC_CONNECTION_COMP_RESUMPTION_FN  ConnectionResumptionTicketValidationComplete; // Available from v2.2
     QUIC_CONNECTION_COMP_CERT_FN        ConnectionCertificateValidationComplete;      // Available from v2.2
 
+    QUIC_STREAM_PROVIDE_RECEIVE_BUFFERS_FN
+                                        StreamProvideReceiveBuffers; // Available from v2.2
```
nibanks, line 1651: Next API version will be 2.5
```suggestion
                                        StreamProvideReceiveBuffers; // Available from v2.5
```

src/inc/quic_trace.h
```diff
@@ -103,6 +103,7 @@ typedef enum QUIC_TRACE_API_TYPE {
     QUIC_TRACE_API_DATAGRAM_SEND,
     QUIC_TRACE_API_CONNECTION_COMPLETE_RESUMPTION_TICKET_VALIDATION,
     QUIC_TRACE_API_CONNECTION_COMPLETE_CERTIFICATE_VALIDATION,
+    QUIC_TRACE_API_STREAM_PROVIDE_RECEIVE_BUFFERS,
```
nibanks, line 106: Our ETW plugin will need to be updated too.

src/test/lib/DataTest.cpp
```diff
@@ -4509,4 +4509,199 @@ QuicTestStreamMultiReceive(
...
+        // Send data
+        QUIC_BUFFER Buffer{BufferSize, SendDataBuffer};
+        TEST_QUIC_SUCCEEDED(SenderStream->Send(&Buffer, 1));
+
+        ReceiveContext.StreamClosed.WaitTimeout(TestWaitTimeout);
+        TEST_EQUAL(ReceiveContext.ReceivedBytes, BufferSize);
+        TEST_EQUAL(0, memcmp(SendDataBuffer, ReceiveDataBuffer, BufferSize));
+    }
+}
 #endif // QUIC_API_ENABLE_PREVIEW_FEATURES
```
nibanks, line 4707: Why does this need to be in the preview features section?

guhetier: I haven't done it yet, but I assumed the entire new API should go initially in the preview feature, until we treat it stable.

src/core/api.c
guhetier: You would allow an "app buffer" stream when multi-receive is enabled?
The problem is that this stream would not support the multi-receive behavior (at least for now, we could make it happen), which could be confusing for the app.

And if app takes dependency on the fact that "app buffer" stream ignore multi-receive, then we can't add support without an API break or new flags.

src/core/api.c
guhetier:
```suggestion
    //
    // Cleanup allocated chunks if the operation failed.
    //
```

src/core/api.c
guhetier: This means each chunk will need a flag to track whether they are allocated from the pool or not then. The size of the chunk seems to be rather optimized (AllocLength restricted to 31bits to avoid a few extra bytes), is it ok to simply add a "flag" byte?

Which makes me realize I need to check the size of the app provided buffers won't override that bit.

src/inc/quic_trace.h
guhetier: Will update these (and SpinQuic) in separate PRs

src/core/api.c
```diff
@@ -1351,6 +1360,163 @@ MsQuicStreamReceiveComplete(
...
+        Handle);
+
+    if (!IS_STREAM_HANDLE(Handle) || Buffers == NULL || BufferCount == 0) {
+        Status = QUIC_STATUS_INVALID_PARAMETER;
+        goto Error;
+    }
+
+    for (uint32_t i = 0; i < BufferCount; ++i) {
+        if (Buffers[i].Length == 0 ||
+            (Buffers[i].Length & ~(UINT32_MAX >> 2)) != 0) {
```
nibanks, line 1391: I thought you were going to change the design in receive buffer to not pack the bits so tightly?

src/core/recv_buffer.h
```diff
@@ -12,33 +12,56 @@ extern "C" {
...
 //
 typedef struct QUIC_RECV_CHUNK {
-    CXPLAT_LIST_ENTRY Link;         // Link in the list of chunks.
-    uint32_t AllocLength : 31;      // Allocation size of Buffer
-    uint32_t ExternalReference : 1; // Indicates the buffer is being used externally.
-    uint8_t Buffer[0];
+    CXPLAT_LIST_ENTRY Link;          // Link in the list of chunks.
+    uint32_t AllocLength       : 31; // Allocation size of Buffer
+    uint32_t ExternalReference : 1;  // Indicates the buffer is being used externally.
+    uint8_t AppOwnedBuffer     : 1;  // Indicates the buffer is managed by the app.
```
nibanks, line 27:
```suggestion
    uint32_t AllocLength;            // Allocation size of Buffer
    uint8_t ExternalReference  : 1;  // Indicates the buffer is being used externally.
    uint8_t AppOwnedBuffer     : 1;  // Indicates the buffer is managed by the app.
```
Let's leave the 32-bits all for the length, and then put the bits together.

src/core/recv_buffer.h
```diff
@@ -12,33 +12,56 @@ extern "C" {
...
     //
     // A list of chunks that make up the buffer.
     //
     CXPLAT_LIST_ENTRY Chunks;
 
+    //
+    // Pool for the chunks managing app provided buffers.
+    // See QUIC_RECV_CHUNK::AppOwnedBuffer
+    //
+    CXPLAT_POOL *AppBufferChunkPool;
```
nibanks, line 53:
```suggestion
    CXPLAT_POOL* AppBufferChunkPool;
```
nit: the `*` is always put with the type.

src/core/recv_buffer.h
```diff
@@ -146,6 +191,16 @@ QuicRecvBufferWrite(
     _Out_ BOOLEAN* NewDataReady
     );
 
+//
+// Returns how many QUIC_BUFFERs should be passed to `QuicRecvBufferRead` to
+// read all the available data in the buffer.
+//
+_IRQL_requires_max_(DISPATCH_LEVEL)
+uint32_t
+QuicRecvBufferReadBufferNeededCount(
```
nibanks, line 200: Add unit test coverage?

src/core/stream.c
```diff
@@ -957,3 +971,56 @@ QuicStreamParamGet(
 
     return Status;
 }
+
+_IRQL_requires_max_(DISPATCH_LEVEL)
+void
+QuicStreamSwitchToAppOwnedBuffers(
+    _In_ QUIC_STREAM* Stream
+)
```
nibanks, line 979:
```suggestion
    )
```

src/core/stream.c
```diff
@@ -957,3 +971,56 @@ QuicStreamParamGet(
...
+    QuicRecvBufferUninitialize(&Stream->RecvBuffer);
+    if (Stream->RecvBuffer.PreallocatedChunk) {
+        CxPlatPoolFree(
+            &Worker->DefaultReceiveBufferPool,
+            Stream->RecvBuffer.PreallocatedChunk);
+        Stream->RecvBuffer.PreallocatedChunk = NULL;
+    }
+
+    //
+    // Rq: Can't fail when initializing in app-owned mode.
```
nibanks, line 994:
```suggestion
    // Can't fail when initializing in app-owned mode.
```

src/core/stream.c
```diff
@@ -957,3 +971,56 @@ QuicStreamParamGet(
...
+        NULL);
+    Stream->Flags.UseAppOwnedRecvBuffers = TRUE;
+}
+
+_IRQL_requires_max_(DISPATCH_LEVEL)
+QUIC_STATUS
+QuicStreamProvideRecvBuffers(
+    _In_ QUIC_STREAM* Stream,
+    _Inout_ CXPLAT_LIST_ENTRY* /* QUIC_RECV_CHUNK */ Chunks
+)
```
nibanks, line 1011:
```suggestion
    )
```

src/core/stream.c
```diff
@@ -957,3 +971,56 @@ QuicStreamParamGet(
...
+{
+    QUIC_STATUS Status = QuicRecvBufferProvideChunks(&Stream->RecvBuffer, Chunks);
+    if (Status == QUIC_STATUS_SUCCESS) {
+        // Update the maximum allowed received size to take into account the new
+        // capacity.
```
nibanks, line 1016:
```suggestion
        //
        // Update the maximum allowed received size to take into account the new
        // capacity.
        //
```

src/core/stream.c
```diff
@@ -957,3 +971,56 @@ QuicStreamParamGet(
...
+        Stream->MaxAllowedRecvOffset =
+            Stream->RecvBuffer.BaseOffset + Stream->RecvBuffer.VirtualBufferLength;
+        QuicSendSetStreamSendFlag(
+            &Stream->Connection->Send,
+            Stream,
+            QUIC_STREAM_SEND_FLAG_MAX_DATA,
+            FALSE);
+    }
+    return Status;
+}
```
nibanks, line 1026: nit: please leave training newline

src/core/api.c
```diff
@@ -658,6 +658,15 @@ MsQuicStreamOpen(
         goto Error;
     }
 
+    if (!!(Flags & QUIC_STREAM_OPEN_FLAG_APP_OWNED_BUFFERS) &&
```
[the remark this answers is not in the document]
guhetier: I can do the change, but I don't like it. We will have to document it very clearly, since it breaks user expectations. It would be very natural to assume that both the connection and stream level settings are being applied, and completely mis-manage buffers.

src/core/stream.h
```diff
@@ -112,6 +112,7 @@ typedef union QUIC_STREAM_FLAGS {
         BOOLEAN Initialized             : 1;    // Initialized successfully. Used for Debugging.
         BOOLEAN Started                 : 1;    // The app has started the stream.
         BOOLEAN StartedIndicated        : 1;    // The app received a start complete event.
+        BOOLEAN PeerStreamStartEventActive : 1; // The app is processing QUIC_CONNECTION_EVENT_PEER_STREAM_STARTED
```
nibanks, line 115: As the comment above this struct indicates, please keep the copy in quictypes.h up to date.

src/core/stream_recv.c
```diff
@@ -793,18 +793,19 @@ QuicStreamOnBytesDelivered(
...
-                    IncreaseRxBuffer,
-                    Stream,
-                    "Increasing max RX buffer size to %u (MinRtt=%llu; TimeNow=%llu; LastUpdate=%llu)",
-                    Stream->RecvBuffer.VirtualBufferLength * 2,
-                    Stream->Connection->Paths[0].MinRtt,
-                    TimeNow,
-                    Stream->RecvWindowLastUpdate);
-
-                QuicRecvBufferIncreaseVirtualBufferLength(
+                if (QuicRecvBufferTryIncreaseVirtualBufferLength(
```
nibanks, line 796: style nit: tab over the subsequent lines with the input args.

src/core/stream_recv.c
```diff
@@ -869,28 +870,55 @@ QuicStreamRecvFlush(
...
+        Event.RECEIVE.BufferCount = ARRAYSIZE(StackRecvBuffers);
+        Event.RECEIVE.Buffers = StackRecvBuffers;
 
         //
         // Try to read the next available buffers.
         //
         BOOLEAN DataAvailable = QuicRecvBufferHasUnreadData(&Stream->RecvBuffer);
         if (DataAvailable) {
+            uint32_t NumBuffersNeeded = QuicRecvBufferReadBufferNeededCount(&Stream->RecvBuffer);
+            if (NumBuffersNeeded > RecvBufferCount) {
```
nibanks, line 895: This is annoying to have to do.... I wonder if there's a way we can eliminate this. We could only indicate at most 3 at a time...

src/core/unittest/RecvBufferTest.cpp
```diff
@@ -49,8 +91,33 @@ struct RecvBuffer {
...
+        Status = ProvideChunks(ChunkList);
+        if (Status != QUIC_STATUS_SUCCESS) {
+            FreeChunkList(ChunkList);
+        }
+        return Status;
+    }
     void IncreaseVirtualBufferLength(uint32_t Length) {
-        QuicRecvBufferIncreaseVirtualBufferLength(&RecvBuf, Length);
+        QuicRecvBufferTryIncreaseVirtualBufferLength(&RecvBuf, Length);
     }
```
nibanks, line 121: Should this helper also return the boolean? Should we have unittest to verify it's behavior?

src/inc/msquic.h
```diff
@@ -199,6 +199,10 @@ typedef enum QUIC_STREAM_OPEN_FLAGS {
     QUIC_STREAM_OPEN_FLAG_0_RTT             = 0x0002,   // The stream was opened via a 0-RTT packet.
     QUIC_STREAM_OPEN_FLAG_DELAY_ID_FC_UPDATES = 0x0004, // Indicates stream ID flow control limit updates for the
                                                         // connection should be delayed to StreamClose.
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+    QUIC_STREAM_OPEN_FLAG_APP_OWNED_BUFFERS  = 0x0008,  // No buffer will be allocated for the stream, the app must
```
nibanks, line 203:
```suggestion
    QUIC_STREAM_OPEN_FLAG_APP_OWNED_BUFFERS = 0x0008,   // No buffer will be allocated for the stream, the app must
```

sslkeylogfile.txt
nibanks, line 1: Delete this file

src/core/unittest/RecvBufferTest.cpp
guhetier: I guess a unit test won't hurt - not for the helper itself, it is pretty trivial, but we have not validation for the "increase the virtual length then write" scenario.

src/core/stream.c
nibanks: I'm not opposed to it, but I don't think it will be very easy to put together.

src/core/recv_buffer.h
```diff
@@ -12,33 +12,56 @@ extern "C" {
...
 //
 typedef struct QUIC_RECV_CHUNK {
-    CXPLAT_LIST_ENTRY Link;         // Link in the list of chunks.
-    uint32_t AllocLength : 31;      // Allocation size of Buffer
-    uint32_t ExternalReference : 1; // Indicates the buffer is being used externally.
-    uint8_t Buffer[0];
+    CXPLAT_LIST_ENTRY Link;          // Link in the list of chunks.
+    uint32_t AllocLength;            // Allocation size of Buffer
+    uint8_t ExternalReference : 1;   // Indicates the buffer is being used externally.
+    uint8_t AppOwnedBuffer     : 1;  // Indicates the buffer is managed by the app.
```
nibanks, line 27:
```suggestion
    uint8_t ExternalReference  : 1;  // Indicates the buffer is being used externally.
    uint8_t AppOwnedBuffer     : 1;  // Indicates the buffer is managed by the app.
```

src/core/recv_buffer.c
```diff
@@ -186,16 +267,21 @@ QuicRecvBufferResize(
...
     )
 {
+    CXPLAT_DBG_ASSERTMSG(
+        RecvBuffer->RecvMode != QUIC_RECV_BUF_MODE_APP_OWNED,
+        "Should never resize in App-owned mode");
     CXPLAT_DBG_ASSERT(
         TargetBufferLength != 0 &&
         (TargetBufferLength & (TargetBufferLength - 1)) == 0); // Power of 2
-    CXPLAT_DBG_ASSERT(!CxPlatListIsEmpty(&RecvBuffer->Chunks)); // Should always have at least one chunk
+    // Should always have at least one chunk
```
nibanks, line 276: style nit: either keep this on the same line like it was (my preference) or you need to use the format:
```c
//
// Should always have at least one chunk
//
```

src/core/recv_buffer.c
```diff
@@ -670,7 +843,6 @@ QuicRecvBufferRead(
                 Link);
         CXPLAT_DBG_ASSERT(!Chunk->ExternalReference);
         CXPLAT_DBG_ASSERT(RecvBuffer->ReadStart == 0);
-        CXPLAT_DBG_ASSERT(*BufferCount >= 1);
```
nibanks, line 673: Instead of deleting this, should be simply add a `|| RecvBuffer->RecvMode == APP_OWNED`?

src/core/recv_buffer.c
```diff
@@ -818,28 +1033,25 @@ QuicRecvBufferPartialDrain(
             RecvBuffer->Chunks.Flink,
             QUIC_RECV_CHUNK,
             Link);
-    CXPLAT_DBG_ASSERT(Chunk->ExternalReference);
```
nibanks, line 821: Why is this remove here and below?

src/core/recv_buffer.c
```diff
@@ -876,20 +1091,23 @@ QuicRecvBufferPartialDrain(
         RecvBuffer->ReadLength -= (uint32_t)DrainLength;
     }
 
-    if (RecvBuffer->RecvMode != QUIC_RECV_BUF_MODE_MULTIPLE) {
-        //
-        // Unless we are in multiple mode, a partial drain means the app isn't
-        // referencing any chunks anymore.
-        //
-        Chunk->ExternalReference = FALSE;
```
nibanks, line 884: We lost this line in the case of single/circular mode AFAICT. Isn't that a bug?

src/core/recv_buffer.c
```diff
@@ -876,20 +1091,23 @@ QuicRecvBufferPartialDrain(
...
         //
-        // If all ReadPending data is drained, then we can release the external reference
+        // If all ReadPending data is drained, then we can release the external reference.
         //
         Chunk->ExternalReference = RecvBuffer->ReadPendingLength != DrainLength;
         CXPLAT_DBG_ASSERT(DrainLength <= RecvBuffer->ReadPendingLength);
         RecvBuffer->ReadPendingLength -= DrainLength;
     }
+
+    if (RecvBuffer->RecvMode == QUIC_RECV_BUF_MODE_APP_OWNED) {
```
nibanks, line 1103:
```suggestion
    
    } else if (RecvBuffer->RecvMode == QUIC_RECV_BUF_MODE_APP_OWNED) {
```

src/core/recv_buffer.c
```diff
@@ -905,57 +1123,72 @@ QuicRecvBufferFullDrain(
...
-        if (Chunk->AllocLength < RecvBuffer->ReadLength) {
-            RecvBuffer->ReadLength = Chunk->AllocLength;
-        }
+    //
+    // The rest of the contiguous data might not fit in just the next chunk
+    // so we need to update the ReadLength of the first chunk to be no more
+    // than the next chunk's allocation length.
+    // Capacity is also updated to reflect the new first chunk's allocation length.
+    //
+    // Update the ReadLength and Capacity to match the new first chunk.
```
nibanks, line 1183:
```suggestion
    // Update the ReadLength and Capacity to match the new first chunk.
    //
```

src/core/recv_buffer.c
```diff
@@ -969,26 +1202,71 @@ QuicRecvBufferDrain(
...
+    // Mark chunks as no longer externally referenced and reset the read-pending data length.
+    // For Multiple mode, this is done when each chunk is drained.
+    //
     if (RecvBuffer->RecvMode != QUIC_RECV_BUF_MODE_MULTIPLE) {
+        for (CXPLAT_LIST_ENTRY* Link = RecvBuffer->Chunks.Flink;
+            Link != &RecvBuffer->Chunks;
+            Link = Link->Flink) {
+            QUIC_RECV_CHUNK* Chunk =
+                CXPLAT_CONTAINING_RECORD(Link, QUIC_RECV_CHUNK, Link);
+            Chunk->ExternalReference = FALSE;
```
nibanks, line 1216: I don't understand why you've changed how/when we set/clear `ExternalReference`.

src/core/recv_buffer.c
```diff
@@ -969,26 +1202,71 @@ QuicRecvBufferDrain(
...
+            //
+            PartialDrain = (uint64_t)RecvBuffer->Capacity > DrainLength;
+        }
+
+        if (PartialDrain) {
+            //
+            // In single/circular mode, a full drain must be done only if all the data
+            // written to the buffer was read.
+            // A partial drain is done if not all the readily readable data was read
+            // or if the read is limited by a gap in the data.
```
nibanks, line 1260: This comment is completely duplicated from above.

src/core/stream_recv.c
```diff
@@ -767,8 +767,8 @@ QuicStreamOnBytesDelivered(
         //
         // Limit stream FC window growth by the connection FC window size.
         //
-        if (Stream->RecvBuffer.VirtualBufferLength <
-            Stream->Connection->Settings.ConnFlowControlWindow) {
+        if (Stream->RecvBuffer.VirtualBufferLength != 0 &&
```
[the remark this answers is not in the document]
nibanks: I'm inclined to try to skip as much unnecessary logic as possible. It's fine for stream logic to understand the buffer internals a bit.

src/core/recv_buffer.c
guhetier: In app-owned mode, on a drain, I need to dereference all chunks that have been referenced, even if not enough data is drained to reach them in the draining loop below.

For Single / Circular mode, what needs to happen is similar: a drain dereference everything. Except that for these modes, the limitation on the number of chunks (1 active and potentially 1 to delete) meant the logic in `QuicRecvBufferPartialDrain` would work.

The drain logic is already the most complex part of recv_buffer, so I am trying to consolidate the dereferencing here for the mode that behave the same in that regard.

src/core/recv_buffer.c
```diff
@@ -48,56 +48,93 @@
...
+{
+    if (Chunk == RecvBuffer->PreallocatedChunk) {
+        return;
+    }
+
+    if (Chunk->AppOwnedBuffer) {
+       CxPlatPoolFree(RecvBuffer->AppBufferChunkPool, Chunk);
+    } else {
+        CXPLAT_FREE(Chunk, QUIC_POOL_RECVBUF);
+    }
```
nibanks, line 80:
```suggestion
    if (Chunk->AppOwnedBuffer) {
        CxPlatPoolFree(RecvBuffer->AppBufferChunkPool, Chunk);
    } else {
        CXPLAT_FREE(Chunk, QUIC_POOL_RECVBUF);
    }
```

src/core/stream_recv.c
```diff
@@ -793,18 +796,19 @@ QuicStreamOnBytesDelivered(
...
-                    "Increasing max RX buffer size to %u (MinRtt=%llu; TimeNow=%llu; LastUpdate=%llu)",
-                    Stream->RecvBuffer.VirtualBufferLength * 2,
-                    Stream->Connection->Paths[0].MinRtt,
-                    TimeNow,
-                    Stream->RecvWindowLastUpdate);
-
-                QuicRecvBufferIncreaseVirtualBufferLength(
-                    &Stream->RecvBuffer,
-                    Stream->RecvBuffer.VirtualBufferLength * 2);
+                if (QuicRecvBufferTryIncreaseVirtualBufferLength(
```
nibanks, line 799: Since you're skipping above, do we need the `Try` part anymore, or the failure case handling?

guhetier: It can probably be replaced with an assertion now - especially if we accept the stream knows about the buffer mode.

src/core/stream_recv.c
guhetier: After giving it some more thought, I think I would rather keep it as it is unless we have a reason to think it has a negative impact.

nibanks: Why? I'd rather reduce the code complexity as much as possible. The assert path is preferrable IMO.

nibanks: Ignore the other comment. We can keep this as is.

src/core/stream_recv.c
nibanks: Thanks. I'd prefer that.

src/core/recv_buffer.c
```diff
@@ -170,10 +205,52 @@ QuicRecvBufferIncreaseVirtualBufferLength(
...
+
+    uint64_t NewBufferLength = RecvBuffer->VirtualBufferLength;
+    for (CXPLAT_LIST_ENTRY* Link = Chunks->Flink;
+         Link != Chunks;
+         Link = Link->Flink) {
+        QUIC_RECV_CHUNK* Chunk = CXPLAT_CONTAINING_RECORD(Link, QUIC_RECV_CHUNK, Link);
+        NewBufferLength += Chunk->AllocLength;
+    }
+
+    if (NewBufferLength > UINT32_MAX) {
```
nibanks, line 231: Is this already ensured at the layer above? If so, we should debug assert and not add branches at multiple layers.

nibanks: And if we can remove this error check, we then simplify the function by no longer needing a return code.

src/core/recv_buffer.c
```diff
@@ -632,6 +724,73 @@ QuicRecvBufferWrite(
     return QUIC_STATUS_SUCCESS;
 }
 
+_IRQL_requires_max_(DISPATCH_LEVEL)
+uint32_t
+QuicRecvBufferReadBufferNeededCount(
+    _In_ const QUIC_RECV_BUFFER* RecvBuffer
+    )
+{
+    if (RecvBuffer->RecvMode == QUIC_RECV_BUF_MODE_SINGLE) {
```
nibanks, line 733: Can we eliminate most of these branches by casting the enum to a number and adding one?

src/core/recv_buffer.c
guhetier: We can't validate the total buffer size before because it depends on the current `VirtualBufferLength`, and adding the buffers can be async.
All that we validate earlier is that the list is not empty, and each individual buffer size is non-zero.

src/core/recv_buffer.c
guhetier: We technically can, but that seems pretty strange and error prone to do, it's only a coincidence that the enum value aligns with the number of buffers. 

I was curious to see if it matters in any way, I did a quick benchmark, and it seems mostly equivalent (I run the benchmark twice with gcc / twice with clang, and got either approach to be slightly faster one on each compiler...)

So I'd rather avoid relying on the value of the enum for no perf gain.

nibanks (approved):

PR #4774, opened 2025-01-29
Add netperf capability to collect CPU traces on Azure

ProjectsByJackHe:
## Description

To improve developer productivity, we enable the option to collect CPU traces in azure for analyzing and identifying bottlenecks.
 
## Testing

Netperf CI

## Documentation

N/A


scripts/quic_callback.ps1
```diff
@@ -3,7 +3,7 @@ param (
 )
 
 if ($PSVersionTable.PSVersion.Major -lt 7) {
-    $isWindows = $true
```
nibanks, line 6: JFYI, PowerShell is case insensitive

nibanks (approved):

PR #4786, opened 2025-01-31
Rust: ConnectionEvent enum

youyuanwu:
## Description
Provide a safe ConnectionEvent enum implementation. Added code to convert from ffi ConectionEvent union.
Removed manual ffi code for ConnectionEvent.
Added StreamRef as an none-owning handle wrapper used in ConnectionEvent.

## Testing
Changed the existing test to use the new ConnectionEvent enum.

## Documentation
NA

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,6 +26,8 @@ extern crate bitfield;
 mod error;
 pub mod ffi;
 pub use error::{Status, StatusCode};
+mod types;
+pub use types::ConnectionEvent;
 
 //
 // The following starts the C interop layer of MsQuic API.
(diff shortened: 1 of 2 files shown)


src/types.rs
```diff
@@ -0,0 +1,178 @@
+use std::ffi::c_void;
```
youyuanwu, line 1: Add copyright.

src/error.rs
```diff
@@ -85,6 +85,19 @@ impl core::fmt::Display for StatusCode {
     }
 }
 
+impl StatusCode {
+    /// If the ffi status is unknown/undefined, use not_supported by default.
+    /// MsQuic returns an undefined status code should not happen ideally,
```
nibanks, line 90: MsQuic absolutely can return other stays codes from the platform. I think we need a better solution for that.

youyuanwu: Removed this. MsQuic returned code will use the Status struct and it can represent arbitrary numbers.
When user wants to check the StatusCode enum, they need to call Status::try_as_status_code, and this enum conversion might return failure if the code is not defined in QUIC_STATUS enum, but some other platform error code.

src/lib.rs
```diff
@@ -1861,37 +1712,55 @@ mod tests {
         event: *mut ffi::QUIC_CONNECTION_EVENT,
     ) -> QUIC_STATUS {
         let connection = unsafe { &*(context as *const Connection) };
-        let event = unsafe { (event as *const ConnectionEvent).as_ref().unwrap() };
-        match event.event_type {
-            crate::CONNECTION_EVENT_CONNECTED => {
+        let ev_ref = unsafe { event.as_ref().unwrap() };
+        let event = ConnectionEvent::from(ev_ref);
+        match event {
```
nibanks, line 1717: This effectively means Rust event handlers have to go through two switch statements now?

youyuanwu: Unfortunately, yes. I hope the optimizer can do the job to eliminate one.

youyuanwu:
> Can you please rebase or merge main?

done

nibanks (approved):

PR #4793, opened 2025-02-04
datapath_winuser: query RioDispatch only if we are using Rio

jianye-chen:
## Description

The `WSAIoctl` call for `RioDispatch` might fail, but it should not block initialization if we have already decided not to use Rio.

## Testing

Existing tests do not cover the issue fixed by this PR.

## Documentation

N/A

### DIFF
--- a/src/platform/datapath_winuser.c
+++ b/src/platform/datapath_winuser.c
@@ -637,26 +637,28 @@ CxPlatDataPathQuerySockoptSupport(
         goto Error;
     }
 
-    Result =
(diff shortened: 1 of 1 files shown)


nibanks:
> ## Description
> The `WSAIoctl` call for `RioDispatch` might fail, but it should not block initialization if we have already decided not to use Rio.

Thanks for making this PR! Can you describe the environment you're running where this query fails? cc @mtfriesen

mtfriesen (approved):

Issue #3516, opened 2023-03-17
Support for QTIP and UDP on QUIC


srsubra:
### Describe the feature you'd like supported

QTIP feature allows QUIC to be sent over TCP/IP for the purposes of framing to leverage SLB fast path support in Azure. But this is a binary flag where the feature works either with UDP or with TCP. It would be better to have a listener support both endpoints to support clients connecting from op-prem and Azure simultaneously.

### Proposed solution

Have an additional flag to support XDP with both QTIP and UDP  in QUIC_EXECUTION_CONFIG_FLAGS like shown below
typedef enum QUIC_EXECUTION_CONFIG_FLAGS {
    QUIC_EXECUTION_CONFIG_FLAG_NONE             = 0x0000,
#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
    QUIC_EXECUTION_CONFIG_FLAG_QTIP             = 0x0001,
   QUIC_EXECUTION_CONFIG_FLAG_ALL            = 0x0002,
#endif
} QUIC_EXECUTION_CONFIG_FLAGS;

ProjectsByJackHe:
UPDATE:

A couple of things:

1. API update: we don't need to add a new EXECUTION_CONFIG_ALL flag, the current design should be interpreted as, if the QTIP flag is set, we *ALLOW* connections to send over QTIP and Listeners to accept QTIP connections. Bottom line, don't change anything with the global flags / state.

2. In order to implement the behavior as specified in this issue, we ADD a new QUIC_SETTINGS parameter that specifies the option to use or don't use QTIP at the Connection/Listener layer. That means we need to update the Connection creation code and the Listener creation code to call into the Datapath layer to apply these state updates. If a Connection is created and we tell it to use QTIP, but the global EXECUTION_CONFIG_QTIP flag is set to OFF, then throw an error or something.

anrossi:
Regarding (2), I think it should be a separate SetParam parameter for listeners and connections, instead of in the QUIC_SETTINGS struct. Putting it in the QUIC_SETTINGS means that this niche feature is visible to many more users of MsQuic, and may encourage more to use it. My understanding is that it's a temporary solution to a specific Azure limitation. I'd rather not immortalize it in the QUIC_SETTINGS struct.

nibanks:
[...]

PR #4803, opened 2025-02-07
Add support for both Qtip and UDP traffic for Server Listeners.

ProjectsByJackHe:
## Description

Currently, we have an execution parameter (QUIC_EXECUTION_CONFIG_FLAG_QTIP) that you set to either process packets via QTIP or normal QUIC/UDP traffic.

For a server listener, that means you can only talk to clients with the same execution profile as you; you either deal over QTIP or QUIC/UDP.

This is inconvenient and sometimes a blocker when you want your listener to accept from clients with varying execution parameters (QTIP through the Azure load balancer, normal QUIC/UDP for on-prem clients)

**Changes**
The main behavior change is that now all Server listeners created will accept both QTIP and QUIC+udp traffic by default when XDP is configured and running correctly.

To support this new design and make testing a bit easier, we also add a knob (QUIC_SETTINGS) to control whether a connection created / listener created is going to send QUIC+udp/QTIP.

This pull-request also updates the behavior of QUIC_EXECUTION_CONFIG_FLAG_QTIP: If you set this flag, all client connections created will default to using QTIP unless otherwise specified. Note that the QUIC_EXECUTION_CONFIG_FLAG_QTIP does not impact server Listeners as they will always accept both QUIC/QTIP traffic assuming XDP is configured and running correctly.

This PR adds an API (connection Get/Set parameter) for client connections to specify whether the connection sends normal QUIC and/or TCP traffic.

Closes: #3516 

## Testing

Added a datatest to alternate between creating normal QUIC connections and QTIP connections, where they all ping the same listener. 

## Documentation

`QTIPEnabled` in `QUIC_SETTINGS` will be documented.


src/inc/msquic.h
```diff
@@ -925,6 +925,11 @@ typedef struct QUIC_SCHANNEL_CREDENTIAL_ATTRIBUTE_W {
 #define QUIC_PARAM_CONN_ORIG_DEST_CID                   0x05000018  // uint8_t[]
 #define QUIC_PARAM_CONN_SEND_DSCP                       0x05000019  // uint8_t
 
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+#define QUIC_PARAM_CONN_SEND_QTIP                       0x0500001A  // uint8_t (BOOLEAN)
+#define QUIC_PARAM_CONN_RECV_QTIP                       0x0500001B  // uint8_t (BOOLEAN)
+#endif
```
nibanks, line 931:
```suggestion
#define QUIC_PARAM_CONN_SEND_DSCP                       0x05000019  // uint8_t
#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
#define QUIC_PARAM_CONN_SEND_QTIP                       0x0500001A  // uint8_t (BOOLEAN)
#define QUIC_PARAM_CONN_RECV_QTIP                       0x0500001B  // uint8_t (BOOLEAN)
#endif
```
As I indicated in the other comment, let's just use one new value `QUIC_PARAM_CONN_QTIP`.

src/core/connection.h
```diff
@@ -208,6 +208,18 @@ typedef union QUIC_CONNECTION_STATE {
...
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+        //
+        // Whether to use QTIP on sends for this connection.
+        //
+        BOOLEAN SendViaQtip : 1;
+
+        //
+        // Whether to allow QTIP on receives for this connection.
+        //
+        BOOLEAN RecvViaQtip : 1;
```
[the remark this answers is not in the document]
ProjectsByJackHe: According to 
image

The RecvViaQtip is for listeners.

src/core/connection.c
```diff
@@ -6754,6 +6754,26 @@ QuicConnParamSet(
         break;
 #endif
 
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
```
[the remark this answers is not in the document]
ProjectsByJackHe: ACK, will fix in the next iteration

src/core/connection.h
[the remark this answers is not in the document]
nibanks: The listener parameter should be `QUIC_PARAM_LISTENER_*`.

src/core/connection.c
```diff
@@ -109,6 +109,7 @@ QuicConnAlloc(
 #if DEBUG
     Connection->RefTypeCount[QUIC_CONN_REF_HANDLE_OWNER] = 1;
 #endif
+    Connection->State.UseQTIP = 0;
```
nibanks, line 112: Unnecessary as we always zero-init the entire connection.

src/core/connection.c
```diff
@@ -6753,6 +6754,14 @@ QuicConnParamSet(
         Status = QUIC_STATUS_SUCCESS;
         break;
 #endif
+    case QUIC_PARAM_CONN_QTIP:
```
nibanks, line 6757: These case statements should be kept ordered by value.

src/core/connection.c
```diff
@@ -6753,6 +6754,14 @@ QuicConnParamSet(
...
+        if (BufferLength != sizeof(BOOLEAN) || Buffer == NULL) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+        Connection->State.UseQTIP = *(BOOLEAN*)Buffer;
```
nibanks, line 6762: You shouldn't let them change after the connection has been started.

src/core/listener.c
```diff
@@ -769,6 +769,17 @@ QuicListenerParamSet(
         return QUIC_STATUS_SUCCESS;
     }
 
+    if (Param == QUIC_PARAM_LISTENER_QTIP) {
```
nibanks, line 772: Probably convert this function back to using a switch statement now that there is more than one.

src/core/listener.h
```diff
@@ -110,6 +110,11 @@ typedef struct QUIC_LISTENER {
     //
     uint8_t CibirId[2 + QUIC_MAX_CIBIR_LENGTH];
 
+    //
+    // Whether or not this listener accepts QTIP traffic (BOOLEAN)
+    //
+    uint8_t UseQTIP : 1;
```
nibanks, line 116: Make this a boolean and put it up with the others for now.

src/inc/quic_datapath.h
```diff
@@ -210,6 +210,7 @@ typedef struct CXPLAT_ROUTE {
     CXPLAT_ROUTE_STATE State;
     CXPLAT_RAW_TCP_STATE TcpState;
 
+    uint8_t UseQTIP : 1;         // TRUE if the route is using QTIP
```
nibanks, line 213: Put this after DatapathType

src/platform/datapath_raw_socket.c
```diff
@@ -1139,4 +1139,4 @@ CxPlatTryAddSocket(
     }
 
     return Status;
-}
+}
```
ProjectsByJackHe, line 1142: will address this.

.github/workflows/test.yml
```diff
@@ -158,18 +162,18 @@ jobs:
...
       shell: pwsh
       timeout-minutes: 120
-      run: scripts/test.ps1 -Config ${{ matrix.vec.config }} -Arch ${{ matrix.vec.arch }} -Tls ${{ matrix.vec.tls }} -OsRunner ${{ matrix.vec.os }} -GHA -LogProfile Full.Light -GenerateXmlResults ${{ matrix.vec.xdp }} ${{ matrix.vec.qtip }}
+      run: scripts/test.ps1 -Config ${{ matrix.vec.config }} -Arch ${{ matrix.vec.arch }} -Tls ${{ matrix.vec.tls }} -OsRunner ${{ matrix.vec.os }} -GHA -LogProfile Full.Light -GenerateXmlResults ${{ matrix.vec.xdp }} ${{ matrix.vec.qtip }} ${{ inputs.filter && '-Filter' }} ${{ inputs.filter || '' }}
     - name: Fix log permissions for Linux XDP
       if: failure() && matrix.vec.plat == 'linux' # (matrix.vec.plat == 'linux' && matrix.vec.xdp == '-UseXdp') doesn't work for some reason
       run: |
         sudo chmod -R 777 artifacts
     - name: Upload on Failure
-      uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08
```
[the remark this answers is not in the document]
ProjectsByJackHe: Yes, the PR is based on an older commit.

guhetier:
`QUIC_EXECUTION_CONFIG_FLAG_QTIP` doesn't seem to be documented.
The new `QUIC_PARAM_CONN_QTIP` also should have some documentation explaining how to use it and why.

src/test/lib/TestConnection.cpp
```diff
@@ -796,6 +796,18 @@ TestConnection::SetResumptionTicket(
...
 
+QUIC_STATUS
+TestConnection::SetQtipPreferences(
+    const uint8_t OnOrOff
+) const {
+    return
+        MsQuic->SetParam(
+            QuicConnection,
+            QUIC_PARAM_CONN_QTIP,
+            sizeof(uint8_t),
```
guhetier, line 807: nit: It is generally better to use `sizeof(variable)` rather than `sizeof(type)`, that way, they can't mismatch.

```suggestion
            sizeof(OnOrOff),
```

src/test/lib/TestConnection.h
```diff
@@ -304,6 +304,8 @@ class TestConnection
 
     QUIC_STATUS SetResumptionTicket(const QUIC_BUFFER* ResumptionTicket) const;
 
+    QUIC_STATUS SetQtipPreferences(const uint8_t OnOrOff) const;
```
guhetier, line 307: nit: `const` on a by-value parameter in a prototype should be removed (it is totally ok to have it in the definition but not the declaration).

ProjectsByJackHe:
> `QUIC_EXECUTION_CONFIG_FLAG_QTIP` doesn't seem to be documented. The new `QUIC_PARAM_CONN_QTIP` also should have some documentation explaining how to use it and why.

Was planning on making a follow up PR to document usage.

src/test/bin/quic_gtest.cpp
```diff
@@ -1751,7 +1751,8 @@ TEST_P(WithSendArgs1, Send) {
             (uint8_t)GetParam().UseSendBuffer,
             (uint8_t)GetParam().UnidirectionalStreams,
             (uint8_t)GetParam().ServerInitiatedStreams,
-            0   // FifoScheduling
+            0,   // FifoScheduling
+            0,   // SendUdpToQtipListener
```
guhetier, line 1755: nit: This one got a comma but not the others.

src/test/bin/quic_gtest.cpp
```diff
@@ -1769,7 +1770,8 @@ TEST_P(WithSendArgs1, Send) {
             GetParam().UseSendBuffer,
             GetParam().UnidirectionalStreams,
             GetParam().ServerInitiatedStreams,
-            false); // FifoScheduling
+            false,  // FifoScheduling
+            true); // SendUdpToQtipListener
```
guhetier, line 1774: This test will have a different behavior in UM and KM, are we OK with that? Or should we create a new test for this that runs only in UM since it doesn't make sense in KM?

src/platform/platform_internal.h
```diff
@@ -1044,7 +1049,9 @@ QUIC_STATUS
 SocketCreateUdp(
     _In_ CXPLAT_DATAPATH* DataPath,
     _In_ const CXPLAT_UDP_CONFIG* Config,
-    _Out_ CXPLAT_SOCKET** NewSocket
+    _Out_ CXPLAT_SOCKET** NewSocket,
```
guhetier, line 1052: It is generally recommended to keep out-parameters last. The two bools should be before `NewSocket`

src/platform/platform_internal.h
```diff
@@ -633,7 +634,9 @@ typedef struct CXPLAT_SOCKET {
     uint8_t Uninitialized : 1;
     uint8_t Freed : 1;
 
-    uint8_t UseTcp : 1;                  // Quic over TCP
+    uint8_t UseTcp : 1;  // This flag represents the expression (THIS SOCKET IS PART OF A CLIENT CONNECTION) && (THE CLIENT CONNECTION SPECIFIED TO USE QTIP)
```
guhetier, line 637: nit: This comment makes the line pretty long, it could go above the variable.
Could also be shorten "Set only for client connections using QTIP" or something similar?

guhetier: Should a similar comment be added on the matching variable for the linux definitions below?

src/core/listener.h
```diff
@@ -115,6 +115,11 @@ typedef struct QUIC_LISTENER {
     //
     uint8_t CibirId[2 + QUIC_MAX_CIBIR_LENGTH];
 
+    //
+    // Whether or not this listener accepts QTIP traffic (BOOLEAN)
+    //
+    BOOLEAN UseQTIP : 1;
```
guhetier, line 121: nit: `AcceptQtip`, given the comment above?

src/core/binding.h
```diff
@@ -279,7 +279,9 @@ _IRQL_requires_max_(PASSIVE_LEVEL)
 QUIC_STATUS
 QuicBindingInitialize(
     _In_ const CXPLAT_UDP_CONFIG* UdpConfig,
-    _Out_ QUIC_BINDING** NewBinding
+    _Out_ QUIC_BINDING** NewBinding,
```
guhetier, line 282: The out parameter should be last.

Also, not related to your change, but the annotation here should be `_Outptr_`.

src/core/library.h
```diff
@@ -555,7 +555,9 @@ _IRQL_requires_max_(PASSIVE_LEVEL)
 QUIC_STATUS
 QuicLibraryGetBinding(
     _In_ const CXPLAT_UDP_CONFIG* UdpConfig,
-    _Out_ QUIC_BINDING** NewBinding
+    _Out_ QUIC_BINDING** NewBinding,
```
guhetier, line 558: Same comment about out params being last + `_Outptr_` sal annotation.

src/core/connection.c
```diff
@@ -7277,6 +7302,28 @@ QuicConnParamGet(
...
+            Status = QUIC_STATUS_BUFFER_TOO_SMALL;
+            break;
+        }
+
+        if (Buffer == NULL) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+
+        *BufferLength = sizeof(BOOLEAN);
```
guhetier, line 7321: nit: Consider sizeof(Connection->State.UseQTIP)

guhetier (commented):
Review in progress :)

src/platform/datapath_xplat.c
```diff
@@ -128,7 +128,9 @@ QUIC_STATUS
 CxPlatSocketCreateUdp(
     _In_ CXPLAT_DATAPATH* Datapath,
     _In_ const CXPLAT_UDP_CONFIG* Config,
-    _Out_ CXPLAT_SOCKET** NewSocket
+    _Out_ CXPLAT_SOCKET** NewSocket,
+    _In_ BOOLEAN UseQTIP,
+    _In_ BOOLEAN OverrideGlobalQTIPSettings
```
guhetier, line 133: A bit below (github won't let me add a comment there):

```
if (Datapath->UseTcp) {
    CxPlatSocketDelete(*NewSocket);
    goto Error;
}
```

Should this take into account the fact that the datapath and the socket might now have a different `UseTcp` value? Should the datapath one still be used?

src/core/connection.c
```diff
@@ -1850,7 +1850,9 @@ QuicConnStart(
     Status =
         QuicLibraryGetBinding(
             &UdpConfig,
-            &Path->Binding);
+            &Path->Binding,
+            Connection->State.UseQTIP,
+            Connection->State.AppDidSetQTIP);
```
guhetier, line 1855: Having to have a second variable to tell if the first one should be considered or not far below is a bit unwieldy.

Could we make the value in the connection always correct, and use it as the single source of truth?
- on connection init, we can set the value based on the global ExecutionConfig
- we already prevent the app from changing the state for server connections
- lower layers simply have to do what the connection tells them to

This might maybe even let us remove the `UseTcp` setting from the datapath? (to double check...)

guhetier:
> > `QUIC_EXECUTION_CONFIG_FLAG_QTIP` doesn't seem to be documented. The new `QUIC_PARAM_CONN_QTIP` also should have some documentation explaining how to use it and why.
> 
> Was planning on making a follow up PR to document usage.

That works, I took the liberty to mention it in the PR description.

src/core/binding.h
ProjectsByJackHe: ACK, next iteration:
- update order of parameters
- update annotation

src/core/connection.c
```diff
@@ -1850,7 +1850,9 @@ QuicConnStart(
     Status =
         QuicLibraryGetBinding(
             &UdpConfig,
-            &Path->Binding);
+            &Path->Binding,
+            Connection->State.UseQTIP,
```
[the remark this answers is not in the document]
ProjectsByJackHe: `UdpConfig` is an ephemeral object here. 

One of the core ideas of this PR was to move state concerning QTIP to the connection layer, and where we do that is in the `Conn->State` field.

src/core/library.h
ProjectsByJackHe: ACK.

src/core/listener.h
ProjectsByJackHe: Will be removing this field in the next iteration. 
This was a change in the earlier iterations of this draft PR.
The decision was made such that listeners always accepted both QTIP/UDP traffic.

src/test/bin/quic_gtest.cpp
ProjectsByJackHe: So the behavior for the test will be the same even if set `SendUdpToQtipListener` to `1` in KM because 
image

src/platform/datapath_xplat.c
ProjectsByJackHe: Awesome catch! 

In this case, the source of truth for clients only is `Socket->UseTcp` and for servers we always delete because servers always initializes everything.

src/core/connection.c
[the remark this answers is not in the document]
guhetier: I got that.
`UdpConfig` is a struct that aggregate all the parameters needed to create a socket (udp or not at this point).
`UseQTIP` can be aggregated with the other parameters. It will still be in the connection, you'll just need a:

`UdpConfig.UseQTIP = Connection->State.UseQTIP`

The idea is to avoid functions with an always increasing number of parameters that are simply forwarded (all the "FALSE, FALSE" you had to add in the PR).

src/test/bin/quic_gtest.cpp
guhetier: Ok. If the result is the same, I'd consider keeping the parameters consistent just to avoid potential weird bugs in the future (let's say we decide to support QTIP in kernel). It's way to easy to miss a difference between the parameters :)

But at this point, I nitpick, either is ok.

src/core/connection.c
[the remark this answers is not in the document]
ProjectsByJackHe: Thanks! That makes sense. The name UdpConfig lead me to me believe this object is specific to non-QTIP logic.

src/core/connection.c
ProjectsByJackHe: I really like this!

I'll do some testing and let you know if I find anything.

src/platform/datapath_xplat.c
ProjectsByJackHe: **Full Context** (you likely already know this, but added for full clarity for other reviewers) 

At the Connection initialization stage while in XDP mode (required for QTIP), we need to instantiate resources like UDP sockets because we need the OS to allocate those sockets for us (we aren't actually going to use them) to get a free port to use for XDP to avoid hijaking traffic from the OS UDP stack.

How it used to work is this global `Datapath->UseTcp` flag dictated whether to create UDP/TCP sockets because we only ever dealt over 1 type of protocol. Now, to support both protocols at the same time via XDP, we require the OS to allocate us both UDP/TCP sockets. The problem we ran into is conflicting port pools for TCP/UDP. We can't use the same port that was given to us when allocating UDP sockets to reserve that port in TCP (for client connections). But, for server listeners, we **can** instantiate both at the same port, and then make decisions to use QTIP/QUIC during runtime depending what type of packets we receive.

src/core/connection.c
ProjectsByJackHe: Hmm... On Connection Init the global ExecutionConfig is NULL.

guhetier: As we were discussing offline, `NULL` is a possibility if the user never set a global parameter, we can use a default.

A potentially bigger problem is synchronization... if going that route, we need to make sure we can't race with global setting change. There might already be a pattern for that, lets look around.

src/platform/platform_internal.h
ProjectsByJackHe: For the linux definition, that field should always be False.

src/test/lib/TestConnection.cpp
ProjectsByJackHe: ACK

src/core/connection.c
ProjectsByJackHe: ACK

src/test/lib/TestConnection.h
ProjectsByJackHe: ACK, fix next iteration.

src/test/bin/quic_gtest.cpp
ProjectsByJackHe: ACK, fix next iteration.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1608,14 +1632,15 @@ CxPlatDpRawTxAlloc(
...
+    UNREFERENCED_PARAMETER(Socket);
     QUIC_ADDRESS_FAMILY Family = QuicAddrGetFamily(&Config->Route->RemoteAddress);
     XDP_QUEUE* Queue = Config->Route->Queue;
     CXPLAT_DBG_ASSERT(Queue != NULL);
     CXPLAT_DBG_ASSERT(&Queue->TxPool != NULL);
     XDP_TX_PACKET* Packet = (XDP_TX_PACKET*)InterlockedPopEntrySList(&Queue->TxPool);
 
     if (Packet) {
-        HEADER_BACKFILL HeaderBackfill = CxPlatDpRawCalculateHeaderBackFill(Family, Socket->UseTcp); // TODO - Cache in Route?
+        HEADER_BACKFILL HeaderBackfill = CxPlatDpRawCalculateHeaderBackFill(Family, Config->Route->UseQTIP); // TODO - Cache in Route?
```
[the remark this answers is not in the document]
ProjectsByJackHe: Ahh, yeah since linux xdp doesn't support QTIP yet, this isn't impacting behavior right now. But I agree, we should keep these updates consistent. Will address this in the next iteration.

src/core/connection.c
ProjectsByJackHe: Actually, updating it to sizeof(Connection->State.UseQTIP) leads to a compilation error (I think it has to do with the way the structure is defined).

Gonna close this thread for now. Feel free to re-open if needed.

src/platform/platform_internal.h
guhetier: The "very long comment" part still needs to be addressed :)

src/core/connection.c
```diff
@@ -6217,6 +6220,8 @@ QuicConnParamSet(
 #ifdef QUIC_OWNING_PROCESS
             UdpConfig.OwningProcess = Connection->Configuration->OwningProcess;
 #endif
+            UdpConfig.UseQTIP = Connection->State.UseQTIP;
+            UdpConfig.OverrideGlobalQTIPSettings = Connection->State.AppDidSetQTIP;
```
guhetier, line 6224: @ProjectsByJackHe I looked around a bit more about how to avoid this "override" variable.

With this PR, we are changing the meaning of `QUIC_EXECUTION_CONFIG_FLAG_QTIP`: it is now a default value for connections that can be overridden (on client connections only), is that correct?

If yes, it really sounds like a setting: it has a default global value, it can be overriden per-connection, it should be set before the connection start...
I think it would make sense to add the new `QUIC_PARAM_CONN_QTIP` in `QUIC_SETTINGS` and to simply remove `QUIC_EXECUTION_CONFIG_FLAG_QTIP`. It is still experimental, so we theoretically can.

Then the existing settings implementation should just work and let you access the setting from the connection: it can become the only source of truth.

@nibanks Is there any reason we can't remove the existing `QUIC_EXECUTION_CONFIG_FLAG_QTIP` I am missing?

nibanks: We can do whatever we want with preview stuff.

src/core/connection.c
```diff
@@ -2243,7 +2244,9 @@ QuicConnGenerateLocalTransportParameters(
     LocalTP->MaxUdpPayloadSize =
         MaxUdpPayloadSizeFromMTU(
             CxPlatSocketGetLocalMtu(
-                Connection->Paths[0].Binding->Socket));
+                Connection->Paths[0].Binding->Socket,
+                Connection->Paths[0].Route.UseQTIP
+            ));
```
nibanks, line 2249: Let's pass the `Route` itself, and make some minor changes to style.
```suggestion
                Connection->Paths[0].Binding->Socket,
                &Connection->Paths[0].Route));
```

src/core/connection.c
```diff
@@ -1843,7 +1843,8 @@ QuicConnStart(
 #ifdef QUIC_OWNING_PROCESS
     UdpConfig.OwningProcess = Configuration->OwningProcess;
 #endif
-
+    UdpConfig.UseQTIP = Connection->State.UseQTIP;
+    UdpConfig.OverrideGlobalQTIPSettings = Connection->State.AppDidSetQTIP;
```
nibanks, line 1847:
```suggestion
    UdpConfig.UseQTIP = Connection->State.UseQTIP;
    UdpConfig.OverrideGlobalQTIPSettings = Connection->State.AppDidSetQTIP;

```

src/core/connection.h
```diff
@@ -1648,7 +1658,7 @@ QuicConnGetMaxMtuForPath(
     //
     uint16_t LocalMtu = Path->LocalMtu;
     if (LocalMtu == 0) {
-        LocalMtu = CxPlatSocketGetLocalMtu(Path->Binding->Socket);
+        LocalMtu = CxPlatSocketGetLocalMtu(Path->Binding->Socket, Path->Route.UseQTIP);
```
nibanks, line 1661:
```suggestion
        LocalMtu = CxPlatSocketGetLocalMtu(Path->Binding->Socket, &Path->Route);
```

src/core/listener.c
```diff
@@ -334,7 +334,8 @@ MsQuicListenerStart(
     Status =
         QuicLibraryGetBinding(
             &UdpConfig,
-            &Listener->Binding);
+            &Listener->Binding
+        );
```
nibanks, line 338:
```suggestion
            &Listener->Binding);
```
Please don't change the style.

src/core/listener.c
```diff
@@ -757,45 +758,49 @@ QuicListenerParamSet(
...
-        }
-        if (BufferLength == 0) {
-            CxPlatZeroMemory(Listener->CibirId, sizeof(Listener->CibirId));
-            return QUIC_STATUS_SUCCESS;
-        }
-        if (BufferLength < 2) { // Must have at least the offset and 1 byte of payload.
-            return QUIC_STATUS_INVALID_PARAMETER;
-        }
+    switch (Param) {
+        case QUIC_PARAM_LISTENER_CIBIR_ID: {
```
nibanks, line 762: Our style has `case` on the same offset as `switch`.

src/core/listener.c
```diff
@@ -757,45 +758,49 @@ QuicListenerParamSet(
         const void* Buffer
     )
 {
-    if (Param == QUIC_PARAM_LISTENER_CIBIR_ID) {
```
nibanks, line 760: Why did you change all this code? AFAICT, you didn't do anything but change from if/else to switch.

src/core/connection.c
nibanks: After looking at the code, I think I agree. Let's make a `QTIPEnabled` setting in the `QUIC_SETTINGS` (inside the PREVIEW block) instead of a parameter. It's cleaner.

src/core/connection.h
```diff
@@ -208,6 +208,16 @@ typedef union QUIC_CONNECTION_STATE {
...
+
+        //
+        // Whether to use QTIP on sends for this connection.
+        //
+        BOOLEAN UseQTIP : 1;
+
+        //
+        // Whether or not the app explicitly set the UseQTIP flag.
+        //
+        BOOLEAN AppDidSetQTIP : 1;
```
nibanks, line 220: Once you convert the parameter to a setting, I don't think this will be needed anymore; possibly not `UseQTIP` either, considering you store the state in the `Route` too.

src/core/path.c
```diff
@@ -35,10 +35,7 @@ QuicPathInitialize(
     Path->EcnValidationState =
         Connection->Settings.EcnEnabled ? ECN_VALIDATION_TESTING : ECN_VALIDATION_FAILED;
 
-    if (MsQuicLib.ExecutionConfig &&
-        MsQuicLib.ExecutionConfig->Flags & QUIC_EXECUTION_CONFIG_FLAG_QTIP) {
-        CxPlatRandom(sizeof(Path->Route.TcpState.SequenceNumber), &Path->Route.TcpState.SequenceNumber);
-    }
+    CxPlatRandom(sizeof(Path->Route.TcpState.SequenceNumber), &Path->Route.TcpState.SequenceNumber);
```
nibanks, line 38: We shouldn't need to do this if we're not using QTIP. Please add back the `if` and fix it up to check if the setting was configured for QTIP.

src/inc/quic_datapath.h
```diff
@@ -578,6 +577,9 @@ typedef struct CXPLAT_UDP_CONFIG {
     uint8_t CibirIdOffsetSrc;           // CIBIR ID offset in source CID
     uint8_t CibirIdOffsetDst;           // CIBIR ID offset in destination CID
     uint8_t CibirId[6];                 // CIBIR ID data
+
+    uint8_t UseQTIP;
+    uint8_t OverrideGlobalQTIPSettings;
```
nibanks, line 582: These can/should be `Flags`. I'm also not sure (yet) why you need two...

src/platform/datapath_raw_xdp_linux.c
```diff
@@ -1064,7 +1062,7 @@ CxPlatDpRawTxAlloc(
 
     Packet = (XDP_TX_PACKET*)xsk_umem__get_data(XskInfo->UmemInfo->Buffer, BaseAddr);
     if (Packet) {
-        HEADER_BACKFILL HeaderBackfill = CxPlatDpRawCalculateHeaderBackFill(Family, Socket->UseTcp); // TODO - Cache in Route?
+        HEADER_BACKFILL HeaderBackfill = CxPlatDpRawCalculateHeaderBackFill(Family, Config->Route->UseQTIP); // TODO - Cache in Route?
```
nibanks, line 1065: Just pass `Route` it has both family and UseQTIP.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1335,48 +1335,72 @@ CxPlatDpRawPlumbRulesOnSocket(
...
 {
+    BOOLEAN UseQTIP = Socket->UseTcp;
     XDP_DATAPATH* Xdp = (XDP_DATAPATH*)Socket->RawDatapath;
     if (Socket->Wildcard) {
-        XDP_RULE Rules[3] = {0};
+        XDP_RULE Rules[5] = {0};
         uint8_t RulesSize = 0;
         if (Socket->CibirIdLength) {
-            Rules[0].Match = Socket->UseTcp ? XDP_MATCH_TCP_QUIC_FLOW_SRC_CID : XDP_MATCH_QUIC_FLOW_SRC_CID;
+            Rules[0].Match = XDP_MATCH_TCP_QUIC_FLOW_SRC_CID;
```
nibanks, line 1344: Put the QUIC rules first.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1398,11 +1422,11 @@ CxPlatDpRawPlumbRulesOnSocket(
         uint8_t* IpAddress;
         size_t IpAddressSize;
         if (Socket->LocalAddress.si_family == QUIC_ADDRESS_FAMILY_INET) {
-            MatchType = Socket->UseTcp ? XDP_MATCH_IPV4_TCP_PORT_SET : XDP_MATCH_IPV4_UDP_PORT_SET;
+            MatchType = UseQTIP ? XDP_MATCH_IPV4_TCP_PORT_SET : XDP_MATCH_IPV4_UDP_PORT_SET;
```
nibanks, line 1425: I don't think you need to create the `UseQTIP` variable above. Just leave this code as it was.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1615,7 +1638,7 @@ CxPlatDpRawTxAlloc(
     XDP_TX_PACKET* Packet = (XDP_TX_PACKET*)InterlockedPopEntrySList(&Queue->TxPool);
 
     if (Packet) {
-        HEADER_BACKFILL HeaderBackfill = CxPlatDpRawCalculateHeaderBackFill(Family, Socket->UseTcp); // TODO - Cache in Route?
+        HEADER_BACKFILL HeaderBackfill = CxPlatDpRawCalculateHeaderBackFill(Family, Config->Route->UseQTIP); // TODO - Cache in Route?
```
nibanks, line 1641: Pass in the Route instead of family and UseQTIP.

src/platform/datapath_xplat.c
```diff
@@ -137,7 +137,8 @@ CxPlatSocketCreateUdp(
         SocketCreateUdp(
             Datapath,
             Config,
-            NewSocket);
+            NewSocket
+        );
```
nibanks, line 141:
```suggestion
            NewSocket);
```
Please don't change style.

src/platform/datapath_xplat.c
```diff
@@ -412,7 +414,28 @@ CxPlatResolveRoute(
...
+        // We always initialize everything. The flag Route->UseQTIP will be set on the receive side.
+        //
+        // For clients, it must be true that Route->UseQTIP == Socket->UseTcp as only 1 set of resources is
+        // allocated.
+        //
+        // For servers, it could be the case that Route->UseQTIP != Socket->UseTcp as servers do not rely
+        // on Socket->UseTcp to initialize resources. For testing purposes though, if we always set
+        // Socket->UseTcp to true in the QTIP scenarios, then both client and servers must have
+        // Route->UseQTIP == Socket->UseTcp.
+        //
```
nibanks, line 434: Instead of having two client and two server comment sections, have one of each.

src/platform/platform_internal.h
```diff
@@ -919,6 +922,8 @@ typedef struct CXPLAT_SOCKET {
 
     uint8_t UseTcp : 1;                  // Quic over TCP
 
+    uint8_t IsServer : 1;
```
nibanks, line 925: `HasFixedRemoteAddress` already is effectively conveys this state. If there is no fixed remote address, then it's a server (unconnected) socket. If there is a fixed remote address, then it's a client socket.

src/test/bin/quic_gtest.cpp
nibanks, line 1774: Instead of changing the behavior of an existing test, let's add a new one, specific to QTIP. I don't want to lose any coverage.

src/test/lib/DataTest.cpp
```diff
@@ -382,6 +396,7 @@ QuicTestConnectAndPing(
     PingStats ClientStats(Length, ConnectionCount, TotalStreamCount, FifoScheduling, UnidirectionalStreams, ServerInitiatedStreams, ClientZeroRtt && !ServerRejectZeroRtt);
 
     MsQuicRegistration Registration(NULL, QUIC_EXECUTION_PROFILE_TYPE_MAX_THROUGHPUT, true);
+
```
nibanks, line 399: nit: please don't go changing all the newline spaces here and below. Leave as is.

src/test/lib/DataTest.cpp
```diff
@@ -431,22 +444,6 @@ QuicTestConnectAndPing(
...
-            QuicAddrFamily,
-            Registration,
-            ServerConfiguration,
-            ClientConfiguration,
-            &ClientStats.ResumptionTicket);
-        if (!ClientStats.ResumptionTicket) {
-            return;
-        }
-    }
-
```
nibanks, line 449: Don't move this code.

src/core/connection.c
ProjectsByJackHe: Updated API to leverage QUIC_SETTINGS.

src/core/connection.c
ProjectsByJackHe: Updated API to use QUIC_SETTINGS

src/core/connection.c
```diff
@@ -1844,6 +1844,12 @@ QuicConnStart(
     UdpConfig.OwningProcess = Configuration->OwningProcess;
 #endif
 
+    if (!Connection->Settings.IsSet.QTIPEnabled) {
+        UdpConfig.UseQTIP = FALSE;
```
nibanks, line 1848: `UdpConfig` is defaulted to `{0}` above, so there is no need to set `FALSE`.

ProjectsByJackHe: Something weird seems to be happening on OpenSSL. Without this guard, weird failures start happening on the CI.

src/core/connection.c
```diff
@@ -6641,6 +6657,25 @@ QuicConnParamSet(
         break;
     }
 
+    case QUIC_PARAM_CONN_QTIP: {
```
nibanks, line 6660: Instead of making a param for this, I think it'd be better to indicate the `UseQTIP` state in the `QUIC_NEW_CONNECTION_INFO` struct that's indicated in the NEW_CONNECTION event to listeners.

src/core/connection.h
nibanks: This should be removed still.

src/core/connection.c
```diff
@@ -1844,6 +1849,13 @@ QuicConnStart(
     UdpConfig.OwningProcess = Configuration->OwningProcess;
 #endif
 
+    UdpConfig.UseQTIP = Connection->Settings.QTIPEnabled;
+    QuicTraceLogConnInfo(
+        IndicateQTIPSettingsBubbledDown,
+        Connection,
+        "Bubbling down UseQTIP setting: %hhu",
+        UdpConfig.UseQTIP);
```
nibanks, line 1857: This log should be unnecessary, IMO, because QuicConnApplyNewSettings would log the value of every settings.

src/core/connection.c
```diff
@@ -1770,6 +1770,11 @@ QuicConnStart(
 
     CXPLAT_TEL_ASSERT(Path->Binding == NULL);
 
+    QuicConnApplyNewSettings(
```
nibanks, line 1773: We want to make sure we don't _also_ call this in QuicConnSetConfiguration for the client path now (but you do for server).

src/inc/msquic.h
```diff
@@ -930,6 +930,10 @@ typedef struct QUIC_SCHANNEL_CREDENTIAL_ATTRIBUTE_W {
 #define QUIC_PARAM_CONN_ORIG_DEST_CID                   0x05000018  // uint8_t[]
 #define QUIC_PARAM_CONN_SEND_DSCP                       0x05000019  // uint8_t
 
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+#define QUIC_PARAM_CONN_QTIP                            0x0500001A  // uint8_t (BOOLEAN)
```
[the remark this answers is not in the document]
nibanks: Let's remove this entirely.

src/inc/quic_datapath.h
```diff
@@ -578,6 +577,8 @@ typedef struct CXPLAT_UDP_CONFIG {
     uint8_t CibirIdOffsetSrc;           // CIBIR ID offset in source CID
     uint8_t CibirIdOffsetDst;           // CIBIR ID offset in destination CID
     uint8_t CibirId[6];                 // CIBIR ID data
+
+    uint8_t UseQTIP;
```
nibanks, line 581: Make this a CXPLAT_SOCKET_FLAG

src/test/bin/quic_gtest.cpp
```diff
@@ -113,7 +113,9 @@ class QuicTestEnvironment : public ::testing::Environment {
 #if defined(QUIC_API_ENABLE_PREVIEW_FEATURES)
             if (UseQTIP) {
                 Config.PollingIdleTimeoutUs = 10000;
-                Config.Flags |= QUIC_EXECUTION_CONFIG_FLAG_QTIP;
+                MsQuicSettings Settings;
+                Settings.SetQtipEnabled(TRUE);
```
nibanks, line 117:
```suggestion
                Settings.SetQtipEnabled(true);
```
C++ code uses `true` instead of `TRUE`.

src/test/lib/DataTest.cpp
```diff
@@ -311,12 +311,34 @@ TestConnection*
...
+        return nullptr;
+    }
+
+
+    if (SendUdpOverQtip && UseQTIP) {
+        // If UseQTIP is true and SendUdpOverQTIP is true, we set it to be the opposite.
+        PerConnSettings.QTIPEnabled = 0;
+        PerConnSettings.IsSet.QTIPEnabled = TRUE;
+        Connection->SetSettings(PerConnSettings);
+        PerConnSettings = Connection->GetSettings();
```
nibanks, line 334: You are trying to use `Connection` before it's been validated to be non-null/valid (below on line 342). Please update theis code.

src/core/connection.c
ProjectsByJackHe: I am removing this in the next few iterations. We are fully migrating off of params to settings. I wanted to get the settings stuff working before deleting code.

src/inc/quic_datapath.h
```diff
@@ -578,6 +577,9 @@ typedef struct CXPLAT_UDP_CONFIG {
...
+    uint8_t IgnoreRawSocketFailure;
```
nibanks, line 582: Please use socket flags instead

src/test/bin/quic_gtest.cpp
ProjectsByJackHe: Fixed with: https://github.com/microsoft/msquic/pull/4803/commits/6b4394a4ede8bb6d33bdb52a51b910134c0e559a

src/core/connection.c
ProjectsByJackHe: Done.

src/inc/quic_datapath.h
ProjectsByJackHe: Done.

src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -185,8 +185,9 @@ QuicMainStart(
 
 #ifndef _KERNEL_MODE
     if (IoMode && IsValue(IoMode, "qtip")) {
-        Config->Flags |= QUIC_EXECUTION_CONFIG_FLAG_QTIP;
-        SetConfig = true;
+        MsQuicSettings Settings;
+        Settings.SetQtipEnabled(TRUE);
+        Settings.SetGlobal();
     }
```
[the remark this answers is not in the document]
ProjectsByJackHe: Done.

src/test/bin/quic_gtest.cpp
ProjectsByJackHe: Done.

src/inc/quic_datapath.h
ProjectsByJackHe: Done.

src/platform/datapath_xplat.c
```diff
@@ -157,11 +160,19 @@ CxPlatSocketCreateUdp(
             QuicTraceLogVerbose(
                 RawSockCreateFail,
                 "[sock] Failed to create raw socket, status:%d", Status);
-            if (Datapath->UseTcp) {
+            BOOLEAN IsWildcardAddr = Config->LocalAddress == NULL || QuicAddrIsWildCard(Config->LocalAddress);
+            if (IsWildcardAddr && TryCount < 1000 && (Config->Flags & CXPLAT_SOCKET_FLAG_QTIP)) {
                 CxPlatSocketDelete(*NewSocket);
+                TryCount += 1;
+                goto Retry;
```
guhetier, line 167: I would avoid this goto.
The semantic you want here is a loop - and if the loop is hard to write, it probably means its body should be moved to a subfunction.

I think the logic if complicated enough it deserves a comment about when/why we are sometimes fine with not creating the raw socket. For instance, it might be correct and intentional, but I am surprised to see we are going to try 1000 times to create a pair of sockets to eventually say "eh, I don't need QTIP anyway, lets succeed", and a comment would help understand why :)

src/platform/platform_internal.h
```diff
@@ -633,7 +638,13 @@ typedef struct CXPLAT_SOCKET {
...
     uint8_t Freed : 1;
 
-    uint8_t UseTcp : 1;                  // Quic over TCP
+    //
+    // This flag determines whether or not we instantiate an auxiliary TCP
+    // socket in XDP mode. For clients, we either only create an auxiliary
+    // TCP socket, or per-proc UDP sockets. For servers, we always create
+    // per-proc UDP sockets, and optionally create an auxiliary TCP socket.
+    //
+    uint8_t UseTcp : 1;
```
guhetier, line 647: nit: Would it make sense to rename this "TcpSocketAvailable" / "TcpSocketInstantiated" / ...? 
Since using QTIP depends on the route for server, is could be confusing to see "UseTcp" and then not use TCP :)

But that's nitpicky, feel free to ignore.

src/platform/datapath_xplat.c
ProjectsByJackHe: Addressed offline + committed fix.

src/platform/platform_internal.h
ProjectsByJackHe: Yeah "useTcp" is a misleading and confusing name. I updated it.

src/perf/lib/PerfClient.cpp
```diff
@@ -236,6 +239,11 @@ PerfClient::Init(
             if (!UsePacing) {
                 Settings.SetPacingEnabled(UsePacing != 0);
             }
+            #ifndef _KERNEL_MODE
```
nibanks, line 242: Need to update the `if` on line 237 too.

ProjectsByJackHe: Updated

src/perf/lib/PerfClient.cpp
```diff
@@ -144,6 +144,9 @@ PerfClient::Init(
     // General configuration options
     //
 
+    #ifndef _KERNEL_MODE
+    TryGetValue(argc, argv, "qtip", &UseQtip);
+    #endif // _KERNEL_MODE
```
nibanks, line 149: Please update help text and docs for secnetperf for this change. Let's also put this after `tcp` and remove the ifdefs here (keep them in the help text).

src/perf/lib/PerfClient.cpp
nibanks, line 242: Just remove the kernel ifdefs

src/core/listener.c
ProjectsByJackHe: Refactoring to use the switch statement. Let me know if you want to reverse this refactor.

src/platform/datapath_raw.c
```diff
@@ -202,11 +198,11 @@ RawSocketDelete(
 _IRQL_requires_max_(DISPATCH_LEVEL)
 uint16_t
 RawSocketGetLocalMtu(
-    _In_ CXPLAT_SOCKET_RAW* Socket
+    _In_ BOOLEAN UseQTIP
```
nibanks, line 201: I'd rather pass the route object in.

src/core/listener.c
```diff
@@ -757,45 +758,49 @@ QuicListenerParamSet(
...
+                }
+                return QUIC_STATUS_SUCCESS;
             }
-            return QUIC_STATUS_SUCCESS;
+            break;
         }
+        default:
+            break;
     }
-
```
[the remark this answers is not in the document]
ProjectsByJackHe: will address in the next iteration.

src/platform/datapath_raw.h
```diff
@@ -62,7 +62,7 @@ typedef struct CXPLAT_DATAPATH_RAW {
     BOOLEAN Uninitialized : 1;
     BOOLEAN Freed : 1;
 #endif
-    BOOLEAN UseTcp;
+    BOOLEAN ReserveAuxTcpSock;
```
nibanks, line 65: Let's add a comment on the same line describing the point of this.

src/core/listener.c
nibanks: Since it's not directly required for this change, let's revert it.

src/core/connection.h
ProjectsByJackHe: Done.

src/inc/msquic.h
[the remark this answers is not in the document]
ProjectsByJackHe: done.

src/inc/quic_datapath.h
ProjectsByJackHe: resolved.

src/core/connection.c
ProjectsByJackHe: Fixed.

src/core/connection.c
ProjectsByJackHe: Fixed.

src/core/connection.h
ProjectsByJackHe: Fixed.

src/platform/platform_internal.h
ProjectsByJackHe: Updated.

src/inc/quic_datapath.h
```diff
@@ -650,7 +652,8 @@ CxPlatSocketUpdateQeo(
 _IRQL_requires_max_(DISPATCH_LEVEL)
 uint16_t
 CxPlatSocketGetLocalMtu(
-    _In_ CXPLAT_SOCKET* Socket
+    _In_ CXPLAT_SOCKET* Socket,
+    _In_ BOOLEAN UseQTIP
```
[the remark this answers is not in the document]
ProjectsByJackHe: Updated.

src/platform/datapath_raw_xdp_win.c
ProjectsByJackHe: Fixed.

src/inc/quic_datapath.h
```diff
@@ -202,14 +202,13 @@ typedef struct CXPLAT_ROUTE {
     uint8_t NextHopLinkLayerAddress[6];
 
     uint16_t DatapathType; // CXPLAT_DATAPATH_TYPE
-
+    uint8_t UseQTIP : 1;   // TRUE if the route is using QTIP
```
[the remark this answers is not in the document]
ProjectsByJackHe: Turned this into a union.

src/core/path.c
ProjectsByJackHe: Fixed.

src/core/listener.c
ProjectsByJackHe: Reverted.

src/core/listener.h
```diff
@@ -115,6 +115,14 @@ typedef struct QUIC_LISTENER {
     //
     uint8_t CibirId[2 + QUIC_MAX_CIBIR_LENGTH];
 
+    //
+    // This flag represents whether or not this listener will try to create an
+    // auxiliary OS TCP socket to reserve the current port when calling Listener.Start().
+    // This is necessary when running msquic in XDP+QTIP mode as we don't want XDP
+    // to steal TCP traffic from other processes.
+    //
+    uint8_t ReserveAuxTcpSock : 1;
```
nibanks, line 124: Instead of caching this at listener create time, eliminate this variable entirely and just use the global settings when you start the listener.

src/inc/quic_datapath.h
```diff
@@ -207,7 +207,12 @@ typedef struct CXPLAT_ROUTE {
     uint8_t LocalLinkLayerAddress[6];
     uint8_t NextHopLinkLayerAddress[6];
 
-    uint16_t DatapathType; // CXPLAT_DATAPATH_TYPE
+    union {
+        struct {
+            uint16_t DatapathType: 15; // CXPLAT_DATAPATH_TYPE
+            uint8_t UseQTIP: 1;        // TRUE if the route is using QTIP
+        };
+    };
```
nibanks, line 215: You shouldn't need all the union and struct stuff.
```suggestion
    union {
        struct {
            uint16_t DatapathType: 15; // CXPLAT_DATAPATH_TYPE
            uint8_t UseQTIP: 1;        // TRUE if the route is using QTIP
        };
    };
```

src/perf/lib/PerfClient.cpp
ProjectsByJackHe: Updated helptext to display the correct/valid io modes.

src/inc/msquic.h
```diff
@@ -697,7 +696,8 @@ typedef struct QUIC_SETTINGS {
             uint64_t OneWayDelayEnabled                     : 1;
             uint64_t NetStatsEventEnabled                   : 1;
             uint64_t StreamMultiReceiveEnabled              : 1;
-            uint64_t RESERVED                               : 21;
+            uint64_t QTIPEnabled                            : 1;
```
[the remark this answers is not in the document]
ProjectsByJackHe: Done.

src/inc/quic_datapath.h
ProjectsByJackHe: Your suggested change is the same as what I already have.

src/inc/msquic.h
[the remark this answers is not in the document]
ProjectsByJackHe: Wait, actually we have to put it last here in Msquic.h because otherwise it breaks down-level compat.

src/test/lib/DataTest.cpp
```diff
@@ -472,13 +492,14 @@ QuicTestConnectAndPing(
         }
 
         UniquePtrArray<TestConnection*> Connections(ConnAlloc);
-
+
```
nibanks, line 495: nit: extra tab?

src/tools/attack/attack.cpp
```diff
@@ -458,7 +458,10 @@ main(
         };
         // flag
         QUIC_EXECUTION_CONFIG_FLAGS Flags = QUIC_EXECUTION_CONFIG_FLAG_XDP;
-        Flags |= AttackType == 0 ? QUIC_EXECUTION_CONFIG_FLAG_QTIP : QUIC_EXECUTION_CONFIG_FLAG_NONE;
+
+        // TODO: Since we moved the QTIP configuration from the datapath to the core layer (MsQuic Settings),
+        //       and this tool seems to be using purely the datapath layer, do we just remove QTIP support from here?
+
```
nibanks, line 464: Let's just remove support.

nibanks (approved):

src/tools/attack/attack.cpp
guhetier: @ProjectsByJackHe I see that the PR was merged before these last comments have been addressed. Are you planning a follow up PR?

ProjectsByJackHe: Yes, I have a follow up PR enqueued.

ProjectsByJackHe: https://github.com/microsoft/msquic/pull/4961

src/test/lib/DataTest.cpp
ProjectsByJackHe: https://github.com/microsoft/msquic/pull/4961

PR #4807, opened 2025-02-12
Add "delay" and "delayType" options to secnetperf tool

sm-msft:
## Description
Add the ability to introduce server-side delays into data streams in the secnetperf tool.

- Default/fixed delay introduces specific microseconds of busy-wait delay
- Variable delay introduces an exponetially distributed random busy-wait delay. 
- Delays are introduced on a separate background thread right before the server sends a response to the client. This thread is pinned to a specific processor (same processor as the one on which the request was received) for predictable behavior.
- Documentation/help text updated
- Minor code reorganization

## Testing
This change updates the test tool. Existing tests are all passing with this change.

## Documentation
Documentation has been updated.


src/perf/lib/PerfServer.cpp
```diff
@@ -181,6 +201,74 @@ PerfServer::ConnectionCallback(
...
+        Now = CxPlatTimeUs64();
+    } while (CxPlatTimeDiff64(Start, Now) <= DelayUs);
+}
+
+double
+PerfServer::CalculateVariableDelay(double lambda)
+{
+    lambda = abs(lambda);
+    std::mt19937 random_generator(CxPlatTimeUs32());
+    std::exponential_distribution<> distribution(lambda);
```
guhetier, line 225: Is there a reason to re-create the distribution on every call?
In the same way that a random seed should usually be provided only once, I expect that recreating the randomness source + the distribution might impact the randomness (and might be a more expensive operation)

src/perf/lib/PerfServer.cpp
```diff
@@ -71,6 +73,24 @@ PerfServer::Init(
         }
     }
 
+    _Benign_race_begin_
```
guhetier, line 76: There is no real race possible, right? This is only the analysis being confused because it can't know initialization is always complete before the stream starts?
`_No_competing_thread_` might be more adapted.

src/perf/lib/PerfServer.cpp
```diff
@@ -181,6 +201,74 @@ PerfServer::ConnectionCallback(
...
+PerfServer::IntroduceVariableDelay(uint32_t DelayUs)
+{
+    if (0 == DelayUs) {
+        return;
+    }
+
+    double lambda = ((double)1) / DelayUs;
+    // Mean value of VariableDelay is expected to be DelayUs
+    double VariableDelay = CalculateVariableDelay(lambda);
+    double MaxFixedDelayUs = 2 * DelayUs;
```
guhetier, line 239: I am surprised to see this depend on `DelayUs` and not be a constant.
For instance, if `DelayUs` is 1us, we should not busy-work for 2us, but if `DelayUs` is 10us, it would be ok to busy-work for 19us?

src/perf/lib/PerfServer.cpp
sm-msft: We are just spinning cycles here. Reseeding the random generator should not make it any less random (and if it does, then it is not a random generator).

src/perf/lib/PerfServer.cpp
sm-msft: Although initialization is run only once, we are not enforcing it in any way. There are competing threads reading the variable and potential competing thread that can write to it. It should not matter in either case, hence the benign race tagging. I am not sure if this will help or hurt the readability here but just keeping it around while I work on developing this.

src/perf/lib/PerfServer.cpp
sm-msft: This depends on DelayUs due to statistical distribution of the exponential distribution generator used here. This simulates a statistically probable large delay.

A fixed or variable delay do not scale down well into low microseconds even if we did not have this rounding issue here. They do not scale well up into multiple milliseconds either. We probably need to impose some limits on the nominal delay parameter.

src/perf/readme.md
```diff
@@ -29,6 +29,8 @@ ecn | `-ecn:<0,1>` | Enables sender-side ECN support.
 exec | `-exec:<lowlat,maxtput,scavenger,realtime>` | The execution profile used for the application.
 pollidle | `-pollidle:<time_us>` | The time, in microseconds, to poll while idle before sleeping (falling back to interrupt-driven IO).
 stats | `-stats:<0,1>` | Prints out statistics at the end of each connection.
+delay | `[-delay:<microseconds>]` | Optional delay in microseconds to be introduced before the server responds to a request.
+delayType | `[-delayType:<fixed|variable>]` | Optional "delay type" can be specified in conjunction with the delay argument. "fixed" delay type introduces the specified delay before the server responds to any request and this is the default behavior. "variable" delay type introduces a statistical variability to the specified delay.
```
sm-msft, line 34: Sample run with cmd ` .\secnetperf.exe -target:127.0.0.1  -down:1KB -ptput:1  -streams:10`
Without server delays:
      _Result: Download 1000 bytes @ 2569 kbps (3.113 ms).
      Result: Download 1000 bytes @ 2457 kbps (3.255 ms).
      Result: Download 1000 bytes @ 2399 kbps (3.334 ms).
      Result: Download 1000 bytes @ 2316 kbps (3.453 ms).
      Result: Download 1000 bytes @ 2251 kbps (3.553 ms).
      Result: Download 1000 bytes @ 2200 kbps (3.636 ms).
      Result: Download 1000 bytes @ 2162 kbps (3.699 ms).
      Result: Download 1000 bytes @ 2130 kbps (3.755 ms).
      Result: Download 1000 bytes @ 2096 kbps (3.816 ms).
      Result: Download 1000 bytes @ 2065 kbps (3.874 ms)._

With 500us fixed server delay:
      _Result: Download 1000 bytes @ 1015 kbps (7.875 ms).
      Result: Download 1000 bytes @ 1000 kbps (7.996 ms).
      Result: Download 1000 bytes @ 992 kbps (8.062 ms).
      Result: Download 1000 bytes @ 985 kbps (8.121 ms).
      Result: Download 1000 bytes @ 977 kbps (8.185 ms).
      Result: Download 1000 bytes @ 970 kbps (8.245 ms).
      Result: Download 1000 bytes @ 962 kbps (8.309 ms).
      Result: Download 1000 bytes @ 956 kbps (8.367 ms).
      Result: Download 1000 bytes @ 948 kbps (8.431 ms).
      Result: Download 1000 bytes @ 942 kbps (8.492 ms)._

With 500us variable server-side delay:
      _Result: Download 1000 bytes @ 368 kbps (21.682 ms).
      Result: Download 1000 bytes @ 366 kbps (21.856 ms).
      Result: Download 1000 bytes @ 364 kbps (21.939 ms).
      Result: Download 1000 bytes @ 363 kbps (22.000 ms).
      Result: Download 1000 bytes @ 362 kbps (22.057 ms).
      Result: Download 1000 bytes @ 361 kbps (22.111 ms).
      Result: Download 1000 bytes @ 360 kbps (22.169 ms).
      Result: Download 1000 bytes @ 360 kbps (22.221 ms).
      Result: Download 1000 bytes @ 359 kbps (22.278 ms).
      Result: Download 1000 bytes @ 358 kbps (22.331 ms)._

sm-msft: The extra delay seems to depend on a few factors including the client cmd parameters and the variable delay. The test was run on an ordinary VM.

src/perf/lib/PerfServer.cpp
```diff
@@ -15,6 +15,8 @@
 #include "PerfServer.cpp.clog.h"
 #endif
 
+#include <random>
```
nibanks, line 18: Unfortunately, since this code is cross-compiled for user mode and kernel mode, we cannot reference the standard library functions/headers in any file in the `lib` directory,

src/perf/lib/PerfServer.cpp
```diff
@@ -71,6 +73,24 @@ PerfServer::Init(
         }
     }
 
+    _Benign_race_begin_
+    if (TryGetValue(argc, argv, "delay", &this->DelayMicroseconds)) {
```
nibanks, line 77: FYI, we have `TryGetVariableUnitValue` in `perfclient.cpp` that allows for the app to specify its own time unit, such as `-delayType:2ms` or `-delayType:100us`. I think it would be useful here.

src/perf/lib/PerfServer.cpp
```diff
@@ -71,6 +73,24 @@ PerfServer::Init(
...
+        // todo: Impose a limit on delay?
```
nibanks, line 78: I would say no. The goal would be to leverage this to mimic a production scenario's workload. If partner X takes 2 seconds to generate responses for requests, we shouldn't limit that just because we think it's too long.

src/perf/lib/PerfServer.cpp
```diff
@@ -181,6 +201,74 @@ PerfServer::ConnectionCallback(
     return QUIC_STATUS_SUCCESS;
 }
 
+void
+PerfServer::IntroduceFixedDelay(uint32_t DelayUs)
+{
+    uint64_t Start = CxPlatTimeUs64();
+    uint64_t Now = Start;
+    WriteOutput("%d\n", DelayUs);
```
nibanks, line 209: Let's remove this. We will want to run tests that might generate millions of requests, each with some fixed delay, and we don't want a console output for them.

src/perf/lib/PerfServer.cpp
```diff
@@ -181,6 +201,74 @@ PerfServer::ConnectionCallback(
...
+    do {
+
+        // Do some stalling work
+        InterlockedIncrement64(&this->WorkCounter);
```
nibanks, line 213: There is no need for this. Simply querying the time (which uses QPC on Windows) is enough work. 😄

src/perf/lib/PerfServer.cpp
```diff
@@ -181,6 +201,74 @@ PerfServer::ConnectionCallback(
...
+
+void
+PerfServer::IntroduceVariableDelay(uint32_t DelayUs)
+{
+    if (0 == DelayUs) {
+        return;
+    }
+
+    double lambda = ((double)1) / DelayUs;
+    // Mean value of VariableDelay is expected to be DelayUs
```
nibanks, line 237: style nit: The comment style in the msquic code base is:
```c++
//
// Mean value of VariableDelay is expected to be DelayUs
//
```

src/perf/lib/PerfServer.cpp
```diff
@@ -205,6 +298,7 @@ PerfServer::StreamCallback(
         break;
     case QUIC_STREAM_EVENT_SEND_COMPLETE:
         Context->OutstandingBytes -= ((QUIC_BUFFER*)Event->SEND_COMPLETE.ClientContext)->Length;
+
```
nibanks, line 301: nit: revert

src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -79,6 +79,10 @@ PrintHelp(
         "  -port:<####>             The UDP port of the server. Ignored if \"bind\" is passed. (def:%u)\n"
         "  -serverid:<####>         The ID of the server (used for load balancing).\n"
         "  -cibir:<hex_bytes>       A CIBIR well-known idenfitier.\n"
+        "  -delay:<microseconds>    Optional delay in microseconds to be introduced before the server responds to a request.\n"
+        "  -delayType:<fixed|variable>    Optional ""delay type"" can be specified in conjunction with the delay argument.\n"
```
nibanks, line 83:
```suggestion
        "  -delayType:<type>       Optional delay type can be specified in conjunction with the 'delay' argument.\n"
```

src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -79,6 +79,10 @@ PrintHelp(
...
+        "                                 ""fixed"" delay type introduces the specified delay before the server responds to any request and this is the default behavior.\n"
+        "                                 ""variable"" delay type introduces a statistical variability to the specified delay.\n"
```
nibanks, line 85:
```suggestion
        "                                 'fixed' - a constant time delay for each request.\n"
        "                                 'variable' - a statistical variability to the specified delay.\n"
```

src/perf/lib/PerfServer.h
```diff
@@ -168,4 +168,20 @@ class PerfServer {
...
+        SYNTHETIC_DELAY_VARIABLE
+    } SYNTHETIC_DELAY_TYPE;
+
+    void IntroduceVariableDelay(uint32_t DelayUs);
+    double CalculateVariableDelay(double lambda);
+    void IntroduceFixedDelay(uint32_t DelayUs);
+    void SimulateDelay(uint32_t DelayUs, SYNTHETIC_DELAY_TYPE DelayType);
+
+
+    volatile int64_t WorkCounter = 0;
```
nibanks, line 183: Unnecessary, since we can just use querying time as the delay's workload.

src/perf/lib/PerfServer.h
```diff
@@ -168,4 +168,20 @@ class PerfServer {
...
+    uint32_t DelayMicroseconds = 0;
+    SYNTHETIC_DELAY_TYPE DelayType = SYNTHETIC_DELAY_FIXED;
```
nibanks, line 185: Please move up to with all the other variables.

src/perf/lib/PerfServer.cpp
guhetier: Depending on how fast you reseed it, it is possible the seed stays the same if you loop fast enough and you get the same value multiple times in a row.

src/perf/lib/PerfServer.cpp
guhetier: Got it, I misunderstood it as a way to avoid spinning for an extensive amount of time. Thanks for the clarification.

src/perf/lib/PerfServer.h
sm-msft: Querying time is an unfenced read, unlike an interlocked operation. The latter creates synchronization between different threads produces interesting pattern of delays.

src/perf/lib/PerfServer.cpp
```diff
@@ -190,6 +278,11 @@ PerfServer::StreamCallback(
     switch (Event->Type) {
     case QUIC_STREAM_EVENT_RECEIVE:
         if (!Context->ResponseSizeSet) {
+            // Introduce delay
+            _Benign_race_begin_
+            SimulateDelay(this->DelayMicroseconds, this->DelayType);
```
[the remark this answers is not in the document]
sm-msft: `In a real-world application, significant delays will not (and cannot) happen on the msquic thread, as this would delay all other network processing (i.e., sending acknowledgements). But small delays (10s, maybe 100s of microseconds) on our thread can be acceptable. ...
Also, the delay should be after all request data has been received to more closely mimic the action of generating the response. So, the QUIC_STREAM_EVENT_PEER_SEND_SHUTDOWN handler will be a better place to handle either running the delay inline or queueing (probably based on a command line arg).`

Thanks for this insight. This explains some of the behaviors I am seeing. I will move the delay to the send shutdown side and see if it fares any differently.

 `So, we must at least support separate async delays (queue to a worker thread, and don't send the response until the delay is complete). `
I plan on subsequently adding support for the async thread model. Your suggestion of not sending a response till delay is complete is an interesting one - I need understand where in the state machine to tap into for this.

src/perf/lib/PerfServer.cpp
sm-msft: This was just for a draft change. Will replace it with a different output subsequently.

src/perf/lib/PerfServer.cpp
sm-msft: This fenced access will synchronize the waiting threads and will create further contention down the line, making it a much richer simulation.

src/perf/lib/PerfServer.cpp
sm-msft: I was seeing unexpected behavior with larger delays - likely due to where the delay is introduced. Got this part about having no upper limit on the delay.

src/perf/lib/PerfServer.cpp
sm-msft: There are different variables at play here and a global RNG is not guaranteed to be the correct solution for a simulation. Statistical results are relevant only with large enough sample sizes and minor inconsistencies should not affect the results at that scale. We will find out once we build and test this out.

src/perf/lib/PerfServer.cpp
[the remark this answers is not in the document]
sm-msft: Discussed this issue separately. Looks like async thread model is the common programming model for Quic applications and it is needed to simulate the delays realistically. Will work towards this.

src/perf/lib/PerfServer.h
sm-msft: I will implement this version in my next iteration. 

I still think the value in simulation is to create contention for a resource with the possibility of lining up threads for simultaneous completion. Perhaps async threads will give us the needed contention by design.

src/perf/lib/PerfServer.cpp
```diff
@@ -71,6 +73,24 @@ PerfServer::Init(
         }
     }
 
+    _Benign_race_begin_
+    if (TryGetValue(argc, argv, "delay", &this->DelayMicroseconds)) {
+        // todo: Impose a limit on delay?
+        const char* DelayTypeString = nullptr;
+        this->DelayType = SYNTHETIC_DELAY_FIXED;
```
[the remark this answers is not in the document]
sm-msft: It is not very productive to argue about style in an established codebase. My take is that code becomes hard to follow when local variables and class members are used within a method without an obvious way to distinguish between them. I also see the use of `This` in this file, but it is not my preferred notation.  I have used `m_` or simply `_` prefix on class members in the past and it makes it easy to follow (and some people hate that notation).

src/perf/lib/PerfServer.cpp
sm-msft: I am not sure how variable delays would even work in Kernel space. I will limit kernel code to using fixed delays.

src/perf/lib/PerfServer.cpp
sm-msft: Removing this. We can add it back later, if needed.

src/perf/lib/PerfServer.cpp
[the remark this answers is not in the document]
sm-msft: Removing use of `this->` in the next iteration.

src/perf/lib/PerfClient.cpp
```diff
@@ -96,6 +96,20 @@ TryGetVariableUnitValue(
...
+_Success_(return != false)
+template
+bool
+TryGetVariableUnitValue<uint32_t>(
+    _In_ int argc,
+    _In_reads_(argc) _Null_terminated_ char* argv[],
+    _In_z_ const char* name,
+    _Out_ uint32_t * pValue,
+    _Out_opt_ bool* isTimed
+);
```
nibanks, line 111: Put this in a header instead?

src/perf/lib/PerfServer.cpp
```diff
@@ -71,6 +84,39 @@ PerfServer::Init(
...
 
+    if (TryGetVariableUnitValue(argc, argv, "delay", &DelayMicroseconds, nullptr)) {
+        const char* DelayTypeString = nullptr;
+        DelayType = SYNTHETIC_DELAY_FIXED;
+
+        if (TryGetValue(argc, argv, "delayType", &DelayTypeString)) {
+#ifndef _KERNEL_MODE
+            if (IsValue(DelayTypeString, "variable")) {
+                //
+                // Initialize variable delay parameters
```
nibanks, line 95: IMO, this comment isn't strictly necessary. It's fairly clear _what_ you're doing. I personally feel comments are useful if there is a _why_ or if the _what_ is really complex.

src/perf/lib/PerfServer.cpp
```diff
@@ -71,6 +84,39 @@ PerfServer::Init(
...
+                //
+                DelayType = SYNTHETIC_DELAY_VARIABLE;
+                Lambda = ((double)1) / DelayMicroseconds;
+                MaxFixedDelayUs = static_cast<uint32_t>(4 * (uint64_t)DelayMicroseconds);
+                if (MaxFixedDelayUs < 1000) MaxFixedDelayUs = 1000;
```
nibanks, line 100: Let's not restrict the max.

src/perf/lib/PerfServer.cpp
```diff
@@ -71,6 +84,39 @@ PerfServer::Init(
...
+
+        ProcCount = (uint16_t)CxPlatProcCount();
+        DelayWorkers = new (std::nothrow) DelayWorker[ProcCount];
+        for (uint16_t i = 0; i < ProcCount; ++i) {
+            if (!DelayWorkers[i].Initialize(this, i)) {
+                WriteOutput("Failed to init delay workers.\n");
+                return QUIC_STATUS_INTERNAL_ERROR;
+            }
+        }
+        DelayPoolUsed = TRUE;
```
nibanks, line 117: Is this variable strictly necessary? Why not just check if `DelayWorkers` is non-null?

src/perf/lib/PerfServer.cpp
```diff
@@ -181,6 +227,93 @@ PerfServer::ConnectionCallback(
...
+    }
+
+    uint64_t Start = CxPlatTimeUs64();
+    uint64_t Now = Start;
+    do {
+        //
+        // Get the current time and check if time has elapsed
+        //
+        Now = CxPlatTimeUs64();
+    } while (CxPlatTimeDiff64(Start, Now) <= DelayUs);
```
nibanks, line 244:
```suggestion
    while (CxPlatTimeDiff64(Start, CxPlatTimeUs64()) <= DelayUs) { /* no op */ }
```

src/perf/lib/PerfServer.cpp
```diff
@@ -217,7 +350,16 @@ PerfServer::StreamCallback(
                 // TODO - Not supported right now
                 MsQuic->StreamShutdown(StreamHandle, QUIC_STREAM_SHUTDOWN_FLAG_ABORT, 0);
             } else {
-                SendResponse(Context, StreamHandle, false);
+                if (DelayPoolUsed) {
+                    //
+                    // Send a delayed response using the background worker thread
+                    //
+                    uint16_t workerNumber = (uint16_t)CxPlatProcCurrentNumber();
```
nibanks, line 357:
```suggestion
                    uint16_t WorkerNumber = (uint16_t)CxPlatProcCurrentNumber();
```

src/perf/lib/PerfServer.cpp
```diff
@@ -376,7 +518,16 @@ PerfServer::TcpReceiveCallback(
 
     } else if (Fin) {
         if (Stream->ResponseSizeSet && Stream->ResponseSize != 0) {
-            Server->SendResponse(Stream, Connection, true);
+            if (Server->DelayPoolUsed) {
+                //
+                // Send a delayed response using the background worker thread
+                //
+                uint16_t workerNumber = (uint16_t)CxPlatProcCurrentNumber();
```
nibanks, line 525:
```suggestion
                uint16_t WorkerNumber = (uint16_t)CxPlatProcCurrentNumber();
```

src/perf/lib/PerfServer.h
```diff
@@ -15,6 +15,72 @@
...
+};
+
+class DelayWorker {
+public:
+    PerfServer* m_Server = nullptr;
+    CXPLAT_EXECUTION_CONTEXT ExecutionContext = { 0 };
+    bool Initialized{ false };
+    CXPLAT_THREAD Thread;
+    CXPLAT_EVENT WakeEvent;
+    CXPLAT_EVENT DoneEvent;
```
nibanks, line 62: Use `CxPlatEvent` instead.

src/perf/lib/PerfServer.h
```diff
@@ -15,6 +15,72 @@
...
+    bool IsTcp;
+    DelayedWorkContext* Next;
+};
+
+class DelayWorker {
+public:
+    PerfServer* m_Server = nullptr;
+    CXPLAT_EXECUTION_CONTEXT ExecutionContext = { 0 };
+    bool Initialized{ false };
+    CXPLAT_THREAD Thread;
```
nibanks, line 60: Maybe use `CxPlatThread`

src/perf/lib/PerfServer.h
```diff
@@ -15,6 +15,72 @@
...
+    CXPLAT_EVENT WakeEvent;
+    CXPLAT_EVENT DoneEvent;
+    CXPLAT_DISPATCH_LOCK Lock;
```
nibanks, line 63: I don't _think_ you need a dispatch lock. MsQuic only ever indicates events up to the app at PASSIVE. So you should be able to use `CxPlatLock`.

src/perf/lib/PerfServer.h
```diff
@@ -43,6 +117,13 @@ class PerfServer {
         );
     QUIC_STATUS Start(_In_ CXPLAT_EVENT* StopEvent);
     QUIC_STATUS Wait(int Timeout);
+    void SimulateDelay();
+    void
+    SendResponse(
+        _In_ StreamContext* Context,
+        _In_ void* Handle,
+        _In_ bool IsTcp
+    );
```
nibanks, line 126:
```suggestion
    void
    SendResponse(
        _In_ StreamContext* Context,
        _In_ void* Handle,
        _In_ bool IsTcp
        );
```

src/perf/lib/PerfServer.h
```diff
@@ -154,6 +216,20 @@ class PerfServer {
     TcpEngine Engine;
     TcpServer Server;
 
+    uint32_t DelayMicroseconds = 0;
+    SYNTHETIC_DELAY_TYPE DelayType = SYNTHETIC_DELAY_FIXED;
+    bool DelayPoolUsed = FALSE;
+    DelayWorker* DelayWorkers = nullptr;
+    uint16_t ProcCount = 0;
```
nibanks, line 223: Let's follow the convention above:
```suggestion
    uint32_t DelayMicroseconds {0};
    SYNTHETIC_DELAY_TYPE DelayType {SYNTHETIC_DELAY_FIXED};
    bool DelayPoolUsed {FALSE};
    DelayWorker* DelayWorkers {nullptr};
    uint16_t ProcCount {0};
```

src/perf/lib/PerfServer.cpp
sm-msft: This is because we use Sleep() beyond MaxFixedDelay and 1ms = 1000us is the smallest sleep value.

nibanks: We aren't always using sleep though, right? We spin too.

src/perf/lib/PerfServer.h
sm-msft: I cant get VSCode or VS2022 to format like this automatically.

src/perf/lib/PerfClient.cpp
sm-msft: This is implemented in perfclient. We should move it to secnetperfmain and include it through a header.

src/perf/lib/PerfServer.h
sm-msft: May be I am missing something. I dont see a CxPlatThread definition in the code...

src/perf/lib/PerfClient.cpp
```diff
@@ -15,87 +15,6 @@
 #include "PerfClient.cpp.clog.h"
 #endif
 
-const char* TimeUnits[] = { "m", "ms", "us", "s" };
```
sm-msft, line 18: This code has been moved to secnetperfmain.cpp

src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -370,3 +378,109 @@ QuicMainGetExtraData(
     CXPLAT_FRE_ASSERT(Client);
     Client->GetExtraData(Data, Length);
 }
+
```
sm-msft, line 381: This is mostly code that has been moved.

src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -370,3 +378,109 @@ QuicMainGetExtraData(
...
+    _Out_ T * pValue,
+    _Out_opt_ bool* isTimed
+)
+{
+    const char* names[] = { name, nullptr };
+    return TryGetVariableUnitValue(argc, argv, names, pValue, isTimed);
+}
+
+/// <summary>
+/// Explicit template instantiation
```
sm-msft, line 464: This is new code

src/perf/lib/PerfServer.cpp
```diff
@@ -181,6 +225,86 @@ PerfServer::ConnectionCallback(
     return QUIC_STATUS_SUCCESS;
 }
 
+void
+PerfServer::IntroduceFixedDelay(uint32_t DelayUs)
+{
+    if (0 == DelayUs) {
+        return;
+    }
```
nibanks, line 233: Theoretically, this isn't really that necessary. It just prevents us from doing two time queries below.

nibanks: Especially since `SimulateDelay` already checks for zero.

src/perf/lib/PerfServer.cpp
```diff
@@ -181,6 +225,86 @@ PerfServer::ConnectionCallback(
...
+}
+
+#else
+
+void
+PerfServer::IntroduceVariableDelay(uint32_t DelayUs)
+{
+    IntroduceFixedDelay(DelayUs);
+}
```
nibanks, line 283: I'd rather just not support variable delay at all on kernel for now, and update the switch in `SimulateDelay`

src/perf/lib/PerfServer.cpp
```diff
@@ -284,6 +412,20 @@ PerfServer::SendResponse(
...
+    _In_ StreamContext* Context,
+    _In_ void* Handle,
+    _In_ bool IsTcp
+    )
+{
+    uint16_t WorkerNumber = (uint16_t)CxPlatProcCurrentNumber();
+    CXPLAT_DBG_ASSERT(WorkerNumber < ProcCount);
+    if (!DelayWorkers[WorkerNumber].QueueWork(Context, Handle, IsTcp)) {
+        WriteOutput("Failed to queue delay worker on processor %d.\n", WorkerNumber);
+    };
```
nibanks, line 426:
```suggestion
    }
```

src/perf/lib/PerfServer.cpp
```diff
@@ -423,3 +569,141 @@ PerfServer::TcpSendCompleteCallback(
...
+
+bool DelayWorker::Initialize(PerfServer* Server, uint16_t PartitionIndex)
+{
+    m_Server = Server;
+    ExecutionContext.Callback = DelayedWork;
+    ExecutionContext.Context = this;
+    InterlockedFetchAndSetBoolean(&ExecutionContext.Ready);
+    ExecutionContext.NextTimeUs = UINT64_MAX;
+
+    uint16_t ThreadFlags = CXPLAT_THREAD_FLAG_SET_IDEAL_PROC;
```
nibanks, line 591: Not really necessary to make a separate variable just for this.

src/perf/lib/PerfServer.cpp
```diff
@@ -423,3 +569,141 @@ PerfServer::TcpSendCompleteCallback(
...
+    //
+    // Not using the high priority flag for delay threads
```
nibanks, line 593: Let's just remove this comment. Most things don't run at high priority, so saying this isn't using that is unnecessary IMO.

src/perf/lib/PerfServer.cpp
```diff
@@ -423,3 +569,141 @@ PerfServer::TcpSendCompleteCallback(
...
+        // delete any pending work items
+        //
+        Lock.Acquire();
+        DelayedWorkContext* CurrentWorkItem = WorkItems;
+        while (nullptr != CurrentWorkItem) {
+            DelayedWorkContext* NextWorkItem;
+            if (WorkItemsTail == &CurrentWorkItem->Next) {
+                NextWorkItem = nullptr;
+            }
+            else {
```
nibanks, line 621:
```suggestion
            } else {
```

src/perf/lib/PerfServer.cpp
```diff
@@ -423,3 +569,141 @@ PerfServer::TcpSendCompleteCallback(
...
+        }
+    }
+    CXPLAT_THREAD_RETURN(0);
+}
+
+BOOLEAN
+DelayWorker::DelayedWork(
+    _Inout_ void* Context,
+    _Inout_ CXPLAT_EXECUTION_STATE* State
+)
```
nibanks, line 656:
```suggestion
    )
```

src/perf/lib/PerfServer.cpp
```diff
@@ -423,3 +569,141 @@ PerfServer::TcpSendCompleteCallback(
...
+DelayWorker::QueueWork(
+    _In_ StreamContext* Context,
+    _In_ void* Handle,
+    _In_ bool IsTcp
+    )
+{
+    DelayedWorkContext* Work = new (std::nothrow) DelayedWorkContext();
+    if (nullptr == Work) {
+        return false;
+    }
```
nibanks, line 697: Since this is test/tool code, we _generally_ assume allocation success. Let's remove this check, which simplifies the whole function (it can't fail then) as well as the callers.

src/perf/lib/PerfServer.h
```diff
@@ -15,6 +15,73 @@
...
+struct DelayedWorkContext {
+    StreamContext* Context;
+    void* Handle;
+    bool IsTcp;
+    DelayedWorkContext* Next;
+};
+
+class DelayWorker {
+public:
+    PerfServer* m_Server = nullptr;
```
nibanks, line 57:
```suggestion
    PerfServer* Server = nullptr;
```

src/perf/lib/PerfServer.h
```diff
@@ -154,6 +223,19 @@ class PerfServer {
...
+    uint32_t DelayMicroseconds {0};
+    SYNTHETIC_DELAY_TYPE DelayType {SYNTHETIC_DELAY_FIXED};
+    DelayWorker* DelayWorkers {nullptr};
+    uint16_t ProcCount {0};
+
+#ifndef _KERNEL_MODE
+    //
+    // Variable delay parameters
+    //
+    double Lambda = 1;
```
nibanks, line 235: Can you replace all ` = ` with ` {value}`? Thanks!

src/perf/lib/SecNetPerf.h
```diff
@@ -224,3 +224,32 @@ QuicPrintStreamStatistics(
...
+_Success_(return != false)
+template <typename T>
+bool
+TryGetVariableUnitValue(
+    _In_ int argc,
+    _In_reads_(argc) _Null_terminated_ char* argv[],
+    _In_z_ const char** names,
+    _Out_ T * pValue,
+    _Out_opt_ bool* isTimed = nullptr
+);
```
nibanks, line 244:
```suggestion
    );
```

src/perf/lib/SecNetPerf.h
```diff
@@ -224,3 +224,32 @@ QuicPrintStreamStatistics(
...
+_Success_(return != false)
+template <typename T>
+bool
+TryGetVariableUnitValue(
+    _In_ int argc,
+    _In_reads_(argc) _Null_terminated_ char* argv[],
+    _In_z_ const char* name,
+    _Out_ T * pValue,
+    _Out_opt_ bool* isTimed = nullptr
+);
```
nibanks, line 255:
```suggestion
    );
```

src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -79,6 +80,10 @@ PrintHelp(
         "  -port:<####>             The UDP port of the server. Ignored if \"bind\" is passed. (def:%u)\n"
         "  -serverid:<####>         The ID of the server (used for load balancing).\n"
         "  -cibir:<hex_bytes>       A CIBIR well-known idenfitier.\n"
+        "  -delay:<####>[unit]      Delay, with an optional unit (def unit is us), to be introduced before the server responds to a request. (def:0)\n"
+        "  -delayType:<fixed|variable>    Optional delay type can be specified in conjunction with the 'delay' argument.\n"
```
nibanks, line 84:
```suggestion
        "  -delay:<####>[unit]      Delay, with an optional unit (def unit is us), to be introduced before the server responds to a request. (def:0)\n"
        "  -delayType:<type>        Optional delay type can be specified in conjunction with the 'delay' argument.\n"
```
We generally use this convention for parameters

src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -370,3 +378,109 @@ QuicMainGetExtraData(
...
+_Success_(return != false)
+template <typename T>
+bool
+TryGetVariableUnitValue(
+    _In_ int argc,
+    _In_reads_(argc) _Null_terminated_ char* argv[],
+    _In_z_ const char** names,
+    _Out_ T * pValue,
+    _Out_opt_ bool* isTimed
+)
```
nibanks, line 398:
```suggestion
    )
```

src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -370,3 +378,109 @@ QuicMainGetExtraData(
...
+_Success_(return != false)
+template <typename T>
+bool
+TryGetVariableUnitValue(
+    _In_ int argc,
+    _In_reads_(argc) _Null_terminated_ char* argv[],
+    _In_z_ const char* name,
+    _Out_ T * pValue,
+    _Out_opt_ bool* isTimed
+)
```
nibanks, line 457:
```suggestion
    )
```

src/perf/lib/SecNetPerfMain.cpp
nibanks: Why do you need these explicit declarations?

src/perf/readme.md
```diff
@@ -29,6 +29,8 @@ ecn | `-ecn:<0,1>` | Enables sender-side ECN support.
 exec | `-exec:<lowlat,maxtput,scavenger,realtime>` | The execution profile used for the application.
 pollidle | `-pollidle:<time_us>` | The time, in microseconds, to poll while idle before sleeping (falling back to interrupt-driven IO).
 stats | `-stats:<0,1>` | Prints out statistics at the end of each connection.
+delay | `[-delay:<value>[units]]` | Delay, with an optional unit (def unit is us), to be introduced before the server responds to a request. (default:0)
```
nibanks, line 32:
```suggestion
delay | `[-delay:<value>[units]]` | Delay, with an optional unit (def unit is us), to be introduced before the server responds to a request.
```

src/perf/lib/PerfServer.cpp
```diff
@@ -423,3 +569,141 @@ PerfServer::TcpSendCompleteCallback(
...
+
+CXPLAT_THREAD_CALLBACK(DelayWorker::WorkerThread, Context)
+{
+    DelayWorker* This = (DelayWorker*)Context;
+    CXPLAT_EXECUTION_STATE DummyState = {
+        0, 0, 0, UINT32_MAX, 0, CxPlatCurThreadID()
+    };
+    while (DelayedWork(This, &DummyState)) {
+        if (!InterlockedFetchAndClearBoolean(&This->ExecutionContext.Ready)) {
+            CxPlatEventWaitForever(This->WakeEvent); // Wait for more work
```
nibanks, line 646:
```suggestion
            This->WakeEvent.WaitForever(); // Wait for more work
```

src/perf/lib/PerfServer.cpp
```diff
@@ -423,3 +569,141 @@ PerfServer::TcpSendCompleteCallback(
...
+void DelayWorker::WakeWorkerThread() {
+    if (!InterlockedFetchAndSetBoolean(&ExecutionContext.Ready)) {
+            CxPlatEventSet(WakeEvent);
+    }
+}
+
+CXPLAT_THREAD_CALLBACK(DelayWorker::WorkerThread, Context)
+{
+    DelayWorker* This = (DelayWorker*)Context;
+    CXPLAT_EXECUTION_STATE DummyState = {
```
nibanks, line 641: You're never using `CXPLAT_EXECUTION_STATE` anywhere. Go ahead and remove all references.

src/perf/lib/PerfServer.cpp
```diff
@@ -423,3 +569,141 @@ PerfServer::TcpSendCompleteCallback(
...
+
+BOOLEAN
+DelayWorker::DelayedWork(
+    _Inout_ void* Context,
+    _Inout_ CXPLAT_EXECUTION_STATE* State
+)
+{
+    DelayWorker* This = (DelayWorker*)Context;
+    if (This->Shuttingdown) {
+        CxPlatEventSet(This->DoneEvent);
```
nibanks, line 660:
```suggestion
        DoneEvent.Set();
```

src/perf/lib/PerfServer.cpp
```diff
@@ -423,3 +569,141 @@ PerfServer::TcpSendCompleteCallback(
...
+        }
+        WorkItem->Next = nullptr;
+    }
+    This->Lock.Release();
+
+    if (nullptr != WorkItem) {
+        This->m_Server->SimulateDelay();
+        This->m_Server->SendResponse(WorkItem->Context, WorkItem->Handle, WorkItem->IsTcp);
+        delete WorkItem;
+        InterlockedFetchAndSetBoolean(&This->ExecutionContext.Ready); // We just did work, let's keep this thread hot.
```
nibanks, line 680: The usage of ExecutionContext and `Ready` is overkill here. I'm pretty sure we can just remove it. In general, you just need the queue and the wake event I think.

src/perf/lib/SecNetPerfMain.cpp
```diff
@@ -16,6 +16,7 @@
 
 const MsQuicApi* MsQuic;
 CXPLAT_WORKER_POOL WorkerPool;
+CXPLAT_WORKER_POOL DelayPool;
```
nibanks, line 19: As far as a I can tell, this isn't used?

src/perf/lib/PerfServer.cpp
```diff
@@ -71,6 +73,48 @@ PerfServer::Init(
         }
     }
 
+    if (TryGetVariableUnitValue(argc, argv, "delay", &DelayMicroseconds, nullptr)) {
```
nibanks, line 76:
```suggestion
    if (TryGetVariableUnitValue(argc, argv, "delay", &DelayMicroseconds, nullptr) && DelayMicroseconds != 0) {
```
Maybe ignore values of zero?

src/perf/lib/SecNetPerfMain.cpp
sm-msft: (I am not an expert on build tools and we are using a vast variety of build tools in this project, each ... so this is a generic explanation of what is happening here) 

The way I look at templates is that it is a shortcut to create copies of a function/class for different types. Copies of functions of only the required type are generated at each step during the build.

Normally, these specific templates are generated/instantiated when secnetperf is compiled into an executable or a dll. They are not generated when just a (secnetperf?) library and object files are created from this code. 

There is some other code that links against functions in the secnetperf library which depends on these specific template objects which wont be found in that library, leading to linker errors. The workaround is to declare these explicit templates. This declaration tells the compiler to instantiate templates with these specific typenames early in the process. 

Some context here: https://learn.microsoft.com/en-us/cpp/cpp/explicit-instantiation?view=msvc-170

src/perf/lib/PerfServer.cpp
sm-msft: We can end up using this elsewhere as well and I want zero delay to mean very close to zero delay as possible. The function can spin for upto 1 microsecond otherwise.

src/perf/lib/PerfServer.cpp
sm-msft: That is apparently valid syntax ...

src/perf/lib/SecNetPerfMain.cpp
sm-msft: This is the threadpool on which the delay threads are scheduled.

src/perf/lib/PerfServer.cpp
sm-msft: We need to put a clang file in place. All of these are default VS 2022/VSCode formatting.

src/perf/lib/PerfServer.cpp
sm-msft: This is in contrast with other threadpools we use in this binary and we have to highlight it somehow..

nibanks: This is not using a thread pool. It is creating a thread directly.

sm-msft: Other threads I guess... I have modified this code and comment as well for the next iteration.

src/perf/lib/SecNetPerfMain.cpp
sm-msft: The parameter here is part of an enumeration. I see other enum parameters in this text shown like so:  <0,1> etc.

Using | is the unix commandline help format to indicate available choice of mutually exclusive options, but I can use the comma if that is used previously here.

sm-msft:
Still seeing some build failures...

nibanks:
> Still seeing some build failures... marking this as a draft

````
perflib.lib(PerfServer.obj) : error LNK2001: unresolved external symbol "void * __cdecl operator new(unsigned __int64)" (??2@YAPEAX_K@Z) [D:\a\msquic\msquic\src\perf\bin\secnetperfdriverpriv.kernel.vcxproj]
```

Means you're calling `new` without `nothrow`.

src/perf/lib/PerfServer.cpp
```diff
@@ -423,3 +559,122 @@ PerfServer::TcpSendCompleteCallback(
...
+                    This->WorkItemsTail = &This->WorkItems;
+                }
+                WorkItem->Next = nullptr;
+            }
+        }
+        This->Lock.Release();
+
+        if (nullptr != WorkItem) {
+            This->Server->SimulateDelay();
+            This->Server->SendResponse(WorkItem->Context, WorkItem->Handle, WorkItem->IsTcp);
```
nibanks, line 654: It's possible that by the time you get here, the stream or parent connection were torn down. This would result in a use-after-free.

sm-msft: As discussed, I did not see this in any of the runs when I ran the client with "-down -streams:xyz" options, even with 5K streams. I saw this only after running the rps-multi scenario a few times, but not with. I have addressed this in my next set of changes, but it touches lot of other code paths. 

I have seen these changes hit a new assert in the msquic path when a delay is used on the server and the client runs rps-multi scenario. (_(Connection->WorkerThreadID == CxPlatCurThreadID())_ in MsQuicStreamSend)

The unchanged tcp server code path seems to be hitting an assert when run with rps-multi scenario and this assert is hit with an increased frequency when a delay is used on the server. (_(Sqe->Overlapped.Internal != 0x103);_ in CxPlatStartDatapathIo)

If the existing tests do not fail, we can finalize these changes.

sm-msft: `I have seen these changes hit a new assert in the msquic path when a delay is used on the server and the client runs rps-multi scenario. ((Connection->WorkerThreadID == CxPlatCurThreadID()) in MsQuicStreamSend)` 
Resolved this with commit 67a77d0

src/perf/lib/Tcp.cpp
```diff
@@ -801,6 +801,10 @@ bool TcpConnection::ProcessTls(const uint8_t* Buffer, uint32_t BufferLength)
 
 bool TcpConnection::SendTlsData(const uint8_t* Buffer, uint16_t BufferLength, uint8_t KeyType)
 {
+    if (!IsActive()) {
```
sm-msft, line 804: I intend to remove this before checkin. This is only meant to be a temp change in this iteration as a possible way to address existing issue of send-after-close scenarios (it causes perf issues if used in all tcp send paths, unfortunately). This normally triggers a debug assert in this tcp code. but otherwise seems harmless and can be left out of the final revision.

sm-msft: No test failures due to this change, but still pulling this out temporary fix.

src/perf/lib/PerfServer.cpp
```diff
@@ -161,12 +204,13 @@ PerfServer::ConnectionCallback(
             if (PrintStats) {
                 QuicPrintConnectionStatistics(MsQuic, ConnectionHandle);
             }
-            MsQuic->ConnectionClose(ConnectionHandle);
+            MsQuic->ConnectionClose((HQUIC)ConnectionHandle);
```
nibanks, line 207: Unnecessary

src/perf/lib/PerfServer.cpp
```diff
@@ -284,6 +404,25 @@ PerfServer::SendResponse(
...
+    )
+{
+    uint16_t WorkerNumber = (uint16_t)CxPlatProcCurrentNumber();
+    CXPLAT_DBG_ASSERT(WorkerNumber < ProcCount);
+    Context->AddRef();
+    if (IsTcp) {
+        //
+        // TcpConnection object is separately reference counted
+        //
+        CXPLAT_DBG_ASSERT(((TcpConnection*)Handle)->TryAddRef());
```
nibanks, line 421: I don't think DBG assert actually executes the inner logic outside of debug builds.

src/perf/lib/PerfServer.cpp
```diff
@@ -415,11 +559,133 @@ PerfServer::TcpSendCompleteCallback(
...
+    DelayedWorkContext* Work = new (std::nothrow) DelayedWorkContext();
+
+    Work->Context = Context;
+    Work->Handle = Handle;
+    Work->IsTcp = IsTcp;
+
+    Lock.Acquire();
+    *WorkItemsTail = Work;
+    WorkItemsTail = &Work->Next;
+    WakeWorkerThread();
```
nibanks, line 689: Wake outside of lock

sm-msft: Same pattern exists in TcpWorker::QueueConnection, but I will not be changing it in this PR.

src/perf/lib/PerfServer.h
```diff
@@ -43,9 +103,23 @@ class PerfServer {
...
+        _In_ StreamContext* Context,
+        _In_ void* Handle,
+        _In_ bool IsTcp
+        );
+    void
+    SendDelayedResponse(
+        _In_ StreamContext* Context,
+        _In_ void* StreamHandle,
+        _In_ bool IsTcp
+        );
```
nibanks, line 118: How that the StreamContext has the handle and IsTcp flag, you don't need to pass these in any more.

src/perf/lib/PerfServer.h
```diff
@@ -15,6 +15,58 @@
...
+    DelayedWorkContext* Next;
+};
+
+class DelayWorker {
+public:
+    PerfServer* Server {nullptr};
+    bool Initialized {false};
+    CxPlatThread Thread;
+    CxPlatEvent WakeEvent;
+    CxPlatEvent DoneEvent;
```
nibanks, line 40:
```suggestion
    CxPlatEvent DoneEvent {true};
```

src/perf/lib/PerfServer.h
```diff
@@ -15,6 +15,58 @@
...
+    CxPlatLock Lock;
+    DelayedWorkContext* WorkItems {nullptr};
+    DelayedWorkContext** WorkItemsTail {&WorkItems};
+    bool Shuttingdown {false};
+
+    DelayWorker() : Thread(true), WakeEvent(false), DoneEvent(true), Lock()
+    {
+    }
```
nibanks, line 48:
```suggestion
    DelayWorker() { }
```
Everything else automatically happens above.

src/perf/lib/PerfServer.cpp
```diff
@@ -415,11 +559,134 @@ PerfServer::TcpSendCompleteCallback(
...
+    _In_ StreamContext* Context,
+    _In_ void* Handle,
+    _In_ bool IsTcp
+    )
+{
+    DelayedWorkContext* Work = new (std::nothrow) DelayedWorkContext();
+
+    Work->Context = Context;
+    Work->Handle = Handle;
+    Work->IsTcp = IsTcp;
```
nibanks, line 684: The handle and IsTcp flag are already saved on the context, so no need to save it on the Work item.

src/perf/lib/PerfServer.cpp
```diff
@@ -415,11 +559,134 @@ PerfServer::TcpSendCompleteCallback(
...
+            CurrentWorkItem = NextWorkItem;
+        }
+        WorkItems = nullptr;
+        Lock.Release();
+    }
+}
+
+void DelayWorker::WakeWorkerThread() {
+    WakeEvent.Set();
+}
```
nibanks, line 618: Small functions like this should go inline in the header.

src/perf/lib/PerfServer.cpp
```diff
@@ -228,8 +349,7 @@ PerfServer::StreamCallback(
         MsQuic->StreamShutdown(StreamHandle, QUIC_STREAM_SHUTDOWN_FLAG_ABORT, 0);
         break;
     case QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE:
-        MsQuic->StreamClose(StreamHandle);
-        StreamContextAllocator.Free(Context);
+        Context->InactivateAndRelease();
```
nibanks, line 352: The 'inactivate' model you're using races with delay thread, but I think it will be ok most of the time.

src/perf/lib/PerfServer.h
```diff
@@ -15,6 +15,58 @@
...
+    SYNTHETIC_DELAY_VARIABLE
+} SYNTHETIC_DELAY_TYPE;
+
+
+struct DelayedWorkContext {
+    StreamContext* Context;
+    void* Handle;
+    bool IsTcp;
+    DelayedWorkContext* Next;
+};
```
nibanks, line 32: You can delete this entire struct now if you simply put a `DelayNext` pointer in the `StreamContext` itself.

src/perf/lib/PerfServer.cpp
nibanks: Instead, how about we just remove the item from the Delay list (if still in, because it could race and just be being removed).

sm-msft: Inactivating a stream will work without races if the send is still in the delay list or even when it is actively holding up the thread. It will race only when it is out of the delay list and is actively sending a response, which is a very minor race and I have not hit that in my testing even with rps-multi scenario.

Getting this to work without any races was the design change I was referring to - we have to sequence the connection close after the send for it to work perfectly all the time.

sm-msft: The race between connection close and send always existed in this code. We will have to figure how to handle that in non-delay cases also.

src/perf/lib/PerfServer.h
sm-msft: I will be updating SendDelayedResponse() and not SendResponse() to keep this change contained.

src/perf/lib/PerfServer.cpp
nibanks: The race didn't exist before because of how msquic guarantees callbacks. It **always** gives stream shutdown callbacks on every stream before it gives the connection shutdown callback (they are all serialized back to the app).

The more I think about this code, it's simpler (and safer) to have this code path lock the worker delay queue, then remove the stream from it if it's still in it, unlock, and then release the ref.

And, on the delay worker, it locks the worker queue, removes the stream from the list, unlocks, delays, sends and then releases the ref.

Finally, I think we must relax these asserts in api.c:
```c
    QUIC_CONN_VERIFY(Connection,
        (Connection->WorkerThreadID == CxPlatCurThreadID()) ||
        !Connection->State.HandleClosed);
```
We shouldn't actually block the app making API calls on a stream even those the connection is closed (we allow this, since the stream implicitly holds a ref on the connection and keeps it alive); but the API will just fail.

----------

So, in summary, let's make the following modifications:

1. Delete the notion of a stream context being active or not (i.e., the `ConnectionClosing` flag).
2. In this function, safely (via the lock) remove the stream context from the queue if it's currently in it.
3. Remove the `QUIC_CONN_VERIFY` in `api.c` that are trying to prevent calls on streams after the connection handle has been closed by the app.

This will ensure we don't do unnecessary work (delay & send) as often as possible when a stream or connection is aborted, keeps proper stream lifetime management, and it prevents msquic from incorrectly complaining about stream usage after connection close.

src/perf/lib/PerfServer.h
nibanks: I strongly prefer we remove this. If we ever need more metadata, we can simply but it on the stream like everything else. In general, I want to apply the principle of keeping code simple now unless there is strong need/plan to add something soon in the future that cannot be supported with the simple design.

src/perf/lib/PerfServer.cpp
nibanks: One other thought: we might also add a flag to the stream `IsShutdown` that the delay polling loop checks and breaks out early.

sm-msft: > The race didn't exist before because of how msquic guarantees callbacks. It always gives stream shutdown callbacks on every stream before it gives the connection shutdown callback (they are all serialized back to the app).

Does the Stream close operation normally synchronize with any new stream send operations? It seems like it would in the existing code because the SteramSend calls are in this single event handler function. I am assuming the connection shutdown is sequenced after (all?) stream close operations succeed.

> In this function, safely (via the lock) remove the stream context from the queue if it's currently in it.

Removing requests from the queue is akin to cancelling server requests. Not all server implementations allow cancelling requests. A closed stream or connection in those cases represents wasted CPU on the server side, which is a valid server simulation, IMO.

Removing the stream context from this queue reduces the window for race condition but does not eliminate it because the worker thread could have already picked up the stream context and is sleeping/attempting to send a response. Other changes are required to achieve this - particularly, the change in api.c.

Couple of follow up questions:
1. Is there a time limit for processing QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE event in the handler function? If there is none, can the handler wait till the stream send completes?

2. In addition to removing the QUIC_CONN_VERIFY statement (which throws exceptions) in api.c, should we check for those conditions and return an error instead? It will expedite further processing on a send that is not going to go through anyway.

sm-msft: I have modified the logic (q2 above) in the next iteration.

src/core/api.c
```diff
@@ -1027,15 +1027,16 @@ MsQuicStreamSend(
 #pragma prefast(suppress: __WARNING_25024, "Pointer cast already validated.")
     Stream = (QUIC_STREAM*)Handle;
 
-    CXPLAT_TEL_ASSERT(!Stream->Flags.HandleClosed);
-    CXPLAT_TEL_ASSERT(!Stream->Flags.Freed);
```
nibanks, line 1031: Don't remove these asserts.

src/core/api.c
```diff
@@ -1027,15 +1027,16 @@ MsQuicStreamSend(
...
-        (Connection->WorkerThreadID == CxPlatCurThreadID()) ||
-        !Connection->State.HandleClosed);
+    if (Stream->Flags.HandleClosed ||
+        Stream->Flags.Freed ||
+        Connection->State.Freed ||
+        (Connection->State.HandleClosed && (Connection->WorkerThreadID != CxPlatCurThreadID())))
+    {
+        Status = QUIC_STATUS_INVALID_STATE;
+        goto Exit;
+    }
```
nibanks, line 1039: These are unnecessary. The `Connection->State.ClosedRemotely` check below catches most of these.

src/core/api.c
```diff
@@ -1027,15 +1027,16 @@ MsQuicStreamSend(
...
 
-    CXPLAT_TEL_ASSERT(!Stream->Flags.HandleClosed);
-    CXPLAT_TEL_ASSERT(!Stream->Flags.Freed);
-
     Connection = Stream->Connection;
 
-    QUIC_CONN_VERIFY(Connection, !Connection->State.Freed);
-    QUIC_CONN_VERIFY(Connection,
-        (Connection->WorkerThreadID == CxPlatCurThreadID()) ||
-        !Connection->State.HandleClosed);
```
nibanks, line 1038: There are 2 other places in this file (I think) that also need to remove these.

src/perf/lib/PerfServer.cpp
```diff
@@ -217,7 +331,13 @@ PerfServer::StreamCallback(
                 // TODO - Not supported right now
                 MsQuic->StreamShutdown(StreamHandle, QUIC_STREAM_SHUTDOWN_FLAG_ABORT, 0);
             } else {
-                SendResponse(Context, StreamHandle, false);
+                CXPLAT_FRE_ASSERT(Context->Handle == (void*)StreamHandle);
+                CXPLAT_FRE_ASSERT(!Context->IsTcp);
```
nibanks, line 335:
```suggestion
                CXPLAT_DBG_ASSERT(Context->Handle == (void*)StreamHandle);
                CXPLAT_DBG_ASSERT(!Context->IsTcp);
```

src/perf/lib/PerfServer.cpp
```diff
@@ -376,7 +511,13 @@ PerfServer::TcpReceiveCallback(
 
     } else if (Fin) {
         if (Stream->ResponseSizeSet && Stream->ResponseSize != 0) {
-            Server->SendResponse(Stream, Connection, true);
+            CXPLAT_FRE_ASSERT(Stream->Handle == (void*)Connection);
+            CXPLAT_FRE_ASSERT(Stream->IsTcp);
```
nibanks, line 515:
```suggestion
            CXPLAT_DBG_ASSERT(Stream->Handle == (void*)Connection);
            CXPLAT_DBG_ASSERT(Stream->IsTcp);
```

src/core/api.c
sm-msft: I see 3 other places, but they dont have additional checks following these asserts to return errors to callers if the connection was already closed. I am unsure what the downstream effects would be. I will publish an iteration with those changes.

src/core/api.c
```diff
@@ -224,11 +224,6 @@ MsQuicConnectionShutdown(
         goto Error;
     }
 
-    QUIC_CONN_VERIFY(Connection, !Connection->State.Freed);
-    QUIC_CONN_VERIFY(Connection,
-        (Connection->WorkerThreadID == CxPlatCurThreadID()) ||
-        !Connection->State.HandleClosed);
-
```
sm-msft, line 231: Unsure if these newly removed verify conditions will be flagged as errors in the code that follows....

sm-msft: Looks like QuicConnTryClose called through the api call below can handle closed connections gracefully.

src/core/api.c
```diff
@@ -1234,11 +1224,6 @@ MsQuicStreamReceiveSetEnabled(
 
     Connection = Stream->Connection;
 
-    QUIC_CONN_VERIFY(Connection, !Connection->State.Freed);
-    QUIC_CONN_VERIFY(Connection,
-        (Connection->WorkerThreadID == CxPlatCurThreadID()) ||
-        !Connection->State.HandleClosed);
-
     Oper = QuicOperationAlloc(Connection->Worker, QUIC_OPER_TYPE_API_CALL);
```
sm-msft, line 1227: QuicStreamRecvSetEnabledState has checks on different Stream flags. It is unclear if any of those checks will fail on a closed stream/closed connection.

src/core/api.c
```diff
@@ -1306,11 +1291,6 @@ MsQuicStreamReceiveComplete(
 
     Connection = Stream->Connection;
 
-    QUIC_CONN_VERIFY(Connection, !Connection->State.Freed);
-    QUIC_CONN_VERIFY(Connection,
-        (Connection->WorkerThreadID == CxPlatCurThreadID()) ||
-        !Connection->State.HandleClosed);
-
```
sm-msft, line 1313: I dont see further checks on connection or stream objects in QuicConnQueueOper() function.

src/core/api.c
nibanks: Sorry, this one should stay, because you shouldn't use a connection after you close (i.e. delete) it. The ones for Stream are all that needed to be deleted, since you can still try to use a stream after the connection is closed (though it generally will gracefully fail).

src/core/api.c
nibanks: This should be fine. The connection stays alive internally as long as the stream maintains a ref on it.

src/core/api.c
sm-msft: Retaining this change since it is in a stream API.

nibanks (approved):

PR #4815, opened 2025-02-14
Rust: StreamEvent enum

youyuanwu:
## Description
Provide safe Stream callback event enum implementation.  Added code to convert from ffi StreamEvent union.
Removed manual ffi code for StreamEvent.

## Testing
Changed the existing test to use the new StreamEvent enum.

## Documentation
NA


src/types.rs
```diff
@@ -177,3 +177,130 @@ impl<'a> From<&'a QUIC_CONNECTION_EVENT> for ConnectionEvent<'a> {
...
+                    peer_accepted: ev.PeerAccepted() != 0,
+                }
+            }
+            crate::ffi::QUIC_STREAM_EVENT_TYPE_QUIC_STREAM_EVENT_RECEIVE => {
+                let ev = unsafe { &mut value.__bindgen_anon_1.RECEIVE };
+                Self::Receive {
+                    absolute_offset: ev.AbsoluteOffset,
+                    total_buffer_length: &mut ev.TotalBufferLength,
+                    buffers: unsafe {
+                        std::slice::from_raw_parts(ev.Buffers, ev.BufferCount as usize)
```
nibanks, line 242: How many dynamic memory allocations does all this require? I assume Rust will dynamically allocate the `Self::Receive` and all the `slice`s here? Or is it able to use stack memory somehow? Bottom line, I see this as introducing a possibly significant bottleneck to a high throughput receive path. In MsQuic we do a LOT of work to completely eliminate dynamic allocations in the datapath, and I'd like to ensure that (somehow) maintains into Rust as well.

youyuanwu: This does not require dynamic alloc. "slice" in rust is like string_view in cpp (on stack). The goal or the requirement I have for the rust wrppers is zero dynamic alloc and zero copy for the buffers.

src/lib.rs
```diff
@@ -1592,7 +1472,7 @@ impl Stream {
         &mut self,
         connection: &Connection,
         flags: StreamOpenFlags,
-        handler: StreamEventHandler,
+        handler: ffi::QUIC_STREAM_CALLBACK_HANDLER,
         context: *const c_void,
     ) -> Result<(), Status> {
         // TODO: remove transmute.
```
guhetier, line 1478: Should this TODO be updated? As well as the clippy annotation?

src/types.rs
```diff
@@ -177,3 +177,130 @@ impl<'a> From<&'a QUIC_CONNECTION_EVENT> for ConnectionEvent<'a> {
...
+/// Stream callback events
+pub enum StreamEvent<'a> {
+    StartComplete {
+        status: crate::Status,
+        id: crate::u62,
+        peer_accepted: bool,
+    },
+    Receive {
+        absolute_offset: u64,
+        total_buffer_length: &'a mut u64,       // out parameter
```
guhetier, line 190: nit: "inout" parameter would be more accurate, since it is also indicating the total number of bytes indicated by the event.

src/types.rs
nibanks: How does it do it statically, when the number of slices is dynamic?

youyuanwu: buffer type is of this:
```rs
buffers: &'a [crate::ffi::QUIC_BUFFER],
```
The memory layout is just
```
struct{
  data: *const crate::ffi::QUIC_BUFFER
  len: usize
}
```
The advantage of using the slice is that u can use iterators etc utilites to traverse it, and makes the lifetime known to the compiler.
I have another branch where QUIC_BUFFER is wrapped and implemented iterators and lifetime etc as well.
The goal is to have wrapper types to expose &[u8] types to users, and the memory the slice points to are owned by msquic. (We will need to figure out for advanced use case where msquic extends the lifetime of the buffer)

src/types.rs
youyuanwu: done.

nibanks (approved):

PR #4832, opened 2025-02-20
Implement Connection Pool API

anrossi:
## Description

Provide an API for pooling connections across multiple RSS cores

## Testing

Manual testing

## Documentation

Documentation will be updated with the new API


src/inc/msquic.h
```diff
@@ -1605,6 +1605,62 @@ QUIC_STATUS
...
+    _In_ HQUIC* Registration,
+    _In_ QUIC_CONNECTION_CALLBACK_HANDLER Handler,
+    _In_opt_ void* Context,
+    _In_opt_ const char* ServerName,
+    _In_opt_ const QUIC_ADDR* ServerAddress,
+    _In_ uint16_t ServerPort,
+    _In_ uint32_t NumberOfConnections,
+    _Out_writes_bytes_(NumberOfConnections * sizeof(HQUIC))
+        HQUIC** ConnectionPool
+    );
```
nibanks, line 1645: I think the extra API table is unnecessary. Let's just add this function directly to the MsQuic table and eliminate all the rest.

src/inc/msquic.h
```diff
@@ -1605,6 +1605,62 @@ QUIC_STATUS
...
+//
+// One of ServerName or ServerAddress *MUST* be supplied, and for the pool to
+// work correctly, the connections *MUST* be started with the same ServerName
+// and ServerPort as supplied in this call.
+//
+typedef
+_IRQL_requires_max_(PASSIVE_LEVEL)
+_Check_return_
+QUIC_STATUS
+(QUIC_API * QUIC_SIMPLE_CONN_POOL_CREATE_FN)(
```
nibanks, line 1635: Let's call this just `QUIC_CONNECTION_POOL_CREATE_FN`, and the usage below `ConnectionPoolCreate`. I'd rather not include simple in the name. We can keep it wrapped in `QUIC_API_ENABLE_PREVIEW_FEATURES` to allow for possible changes in the future if we want to make it more complicated.

src/inc/msquic.h
```diff
@@ -1657,6 +1713,8 @@ typedef struct QUIC_API_TABLE {
     QUIC_STREAM_PROVIDE_RECEIVE_BUFFERS_FN
                                         StreamProvideReceiveBuffers; // Available from v2.5
 #endif
+    QUIC_CONNECTION_POOL_API_OPEN_FN     ConnectionPoolApiOpen; // Available from v2.5
```
nibanks, line 1716: Please put it in `QUIC_API_ENABLE_PREVIEW_FEATURES`, make sure it aligns with the other names and align the comment too.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1882,3 +1882,115 @@ CxPlatIoXdpShutdownEventComplete(
...
+    )
+{
+    QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
+    uint32_t RssConfigSize = 0;
+    XDP_RSS_CONFIGURATION *RawRssConfig = NULL;
+
+    if (RssConfig == NULL) {
+        Status = QUIC_STATUS_INVALID_PARAMETER;
+        goto Error;
+    }
```
nibanks, line 1901: If this is an internal function, we can probably assert this instead of checking it. SAL indicates this is a required output too.

src/platform/unittest/ToeplitzTest.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+#define _CRT_SECURE_NO_WARNINGS 1
+#include "main.h"
+#include "msquic.h"
+#include "msquichelper.h"
+#include "quic_toeplitz.h"
+#include <stdio.h>
+
+#ifdef QUIC_CLOG
+#include "ToeplitzTest.cpp.clog.h"
```
nibanks, line 16: There's a clog file you need to add this too I think, for this to work.

src/platform/unittest/ToeplitzTest.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+    struct QuicTestAddress {
+        QUIC_ADDR Addr;
+        QuicTestAddress(const char* AddrStr, uint16_t Port) {
+            EXPECT_TRUE(QuicAddrFromString(AddrStr, Port, &Addr));
+        }
+    };
+
+    static
+    auto
```
nibanks, line 55: `auto` for a function definition? I've never seen that before.

src/platform/datapath_raw_xdp_win.c
anrossi: Well it's part of the CxPlat interface, so if we make that library separate, it should validate this

src/platform/unittest/ToeplitzTest.cpp
anrossi: I was a little surprised by it too. I couldn't figure out the GTEST type that is returned by the ASSERT macros but apparently the compiler can figure it out.

src/inc/quic_datapath.h
```diff
@@ -858,6 +858,37 @@ CxPlatUpdateRoute(
...
+    uint32_t RssIndirectionTableLength;
+    uint8_t* RssSecretKey;
+    CXPLAT_PROCESSOR_INFO* RssIndirectionTable;
+} CXPLAT_RSS_CONFIG;
+
+_IRQL_requires_max_(PASSIVE_LEVEL)
+QUIC_STATUS
+CxPlatDataPathGetRssConfig(
+    _In_ uint32_t InterfaceIndex,
+    _Outptr_ _At_(*RssConfig, __drv_allocatesMem(Mem))
```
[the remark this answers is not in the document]
anrossi: I did see another instance where a CxPlat function would allocate something and expect the caller to free it with the correct pool tag. I'll write a nice wrapper function to make it explicit, however.

src/inc/msquic.h
anrossi: But with the second level API table, we can extend the connection pool API to more than just an array of connections. I don't know how future-looking we should be with this design

nibanks: Let's keep it as simple as possible. As long as it's wrapped by preview flag we can change later if necessary.

src/inc/msquic.h
```diff
@@ -1605,6 +1605,35 @@ QUIC_STATUS
     _In_opt_ void* ClientSendContext
     );
 
+//
+// Connection Pool API
+//
+
+//
+// Creates a simple connection pool with NumberOfConnections connections
```
nibanks, line 1613:
```suggestion
// Creates a simple pool of NumberOfConnections connections,
```

src/inc/msquic.h
```diff
@@ -1605,6 +1605,35 @@ QUIC_STATUS
...
+// all with the same Context and Handler, and puts them in the
+// caller-supplied array.
```
nibanks, line 1615: Mention that the goal of this pool is to evenly spread these connections across RSS as much as possible.

src/inc/msquic.h
```diff
@@ -1605,6 +1605,35 @@ QUIC_STATUS
...
+//
+typedef
+_IRQL_requires_max_(PASSIVE_LEVEL)
+_Check_return_
+QUIC_STATUS
+(QUIC_API * QUIC_CONN_POOL_CREATE_FN)(
+    _In_ HQUIC Registration,
+    _In_ QUIC_CONNECTION_CALLBACK_HANDLER Handler,
+    _In_opt_ void* Context,
+    _In_opt_ const char* ServerName,
```
nibanks, line 1629: Since `ServerName` can also be an IP address (as a string), let's remove the `ServerAddress` variable.

src/inc/msquic.h
```diff
@@ -1605,6 +1605,35 @@ QUIC_STATUS
...
+    _In_opt_ const QUIC_ADDR* ServerAddress,
+    _In_ uint16_t ServerPort,
+    _In_ uint16_t NumberOfConnections,
+    _Out_writes_bytes_(NumberOfConnections * sizeof(HQUIC))
+        HQUIC** ConnectionPool
```
nibanks, line 1634: Pointer to an array of pointers to handles is a lot of indirection. I'm wondering if we can just take a pointer to an array of handles...

src/inc/msquic_posix.h
```diff
@@ -218,6 +218,10 @@ extern "C" {
 // IP Address Abstraction Helpers
 //
 
+// RFC 6335
+#define QUIC_ADDR_EPHEMERAL_PORT_MIN 49152
+#define QUIC_ADDR_EPHEMERAL_PORT_MAX 65535
```
nibanks, line 223: At least on Windows, I think these ranges are customizable. Are they on Linux as well?

src/inc/quic_datapath.h
```diff
@@ -858,6 +858,43 @@ CxPlatUpdateRoute(
...
+//
+#define CXPLAT_RSS_HASH_TYPE_IPV4        0x001
+#define CXPLAT_RSS_HASH_TYPE_TCP_IPV4    0x002
+#define CXPLAT_RSS_HASH_TYPE_UDP_IPV4    0x004
+#define CXPLAT_RSS_HASH_TYPE_IPV6        0x008
+#define CXPLAT_RSS_HASH_TYPE_TCP_IPV6    0x010
+#define CXPLAT_RSS_HASH_TYPE_UDP_IPV6    0x020
+#define CXPLAT_RSS_HASH_TYPE_IPV6_EX     0x040
+#define CXPLAT_RSS_HASH_TYPE_TCP_IPV6_EX 0x080
+#define CXPLAT_RSS_HASH_TYPE_UDP_IPV6_EX 0x100
```
nibanks, line 872: Why are _types_ set up as _flags_? Can you set all of them simultaneously? I'd assume they are mutually exclusive.

nibanks: And, should these be an enum?

src/platform/datapath_raw_xdp_win.c
nibanks: If we ever actually do that, then there will be plenty to clean up. For now, simpler is better IMO.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1882,3 +1882,124 @@ CxPlatIoXdpShutdownEventComplete(
...
+            Result);
+        Status = Result;
+        goto Error;
+    }
+
+    RssConfigSize = sizeof(CXPLAT_RSS_CONFIG) +
+        RawRssConfig->HashSecretKeySize +
+        RawRssConfig->IndirectionTableSize;
+    CXPLAT_RSS_CONFIG* NewRssConfig =
+        (CXPLAT_RSS_CONFIG*)CXPLAT_ALLOC_NONPAGED(
```
nibanks, line 1958: You used PAGED above and NONPAGED here. I know it doesn't _really_ matter in user mode, but we might adapt this for kernel one day, so we should get it right. I'd say PAGED for both, as this likely wouldn't need to be called at DISPATCH.

src/core/connection_pool.c
```diff
@@ -0,0 +1,350 @@
...
+    for (uint32_t i = 0; i < RssConfig->RssIndirectionTableLength / sizeof(CXPLAT_PROCESSOR_INFO); i++) {
+        uint32_t j;
+        for (j = 0; j < RssProcessorCount; j++) {
+            if (RssProcessors[j].Group == RssConfig->RssIndirectionTable[i].Group &&
+                RssProcessors[j].Index == RssConfig->RssIndirectionTable[i].Index) {
+                break;
+            }
+        }
+        //
+        // This is safe because the RssProcessor array is the same size as the indirection table.
```
nibanks, line 177: If you add a debug assert above that they are the same, SAL should be happy.

src/core/connection_pool.c
```diff
@@ -0,0 +1,350 @@
...
+            QuicTraceLogError(
+                ConnPoolOpenConnection,
+                "[conp] Failed to open connection[%u], 0x%x",
+                i,
+                Status);
+            goto Error;
+        }
+        CreatedConnections++;
+
+        Status = MsQuicSetParam(Connections[i], QUIC_PARAM_CONN_REMOTE_ADDRESS, sizeof(ResolvedRemoteAddress), &ResolvedRemoteAddress);
```
nibanks, line 303: Use `QuicConnSetParam` instead. Also, you can debug assert the result, because it can only fail if you mess up the code here (i.e. invalid parameters). Same for local address.

src/inc/msquic.h
```diff
@@ -1605,6 +1605,35 @@ QUIC_STATUS
...
+// and ServerPort as supplied in this call.
+//
+typedef
+_IRQL_requires_max_(PASSIVE_LEVEL)
+_Check_return_
+QUIC_STATUS
+(QUIC_API * QUIC_CONN_POOL_CREATE_FN)(
+    _In_ HQUIC Registration,
+    _In_ QUIC_CONNECTION_CALLBACK_HANDLER Handler,
+    _In_opt_ void* Context,
```
nibanks, line 1628: They probably need to pass in an array of context pointers, a unique one for each.

src/core/connection_pool.c
```diff
@@ -0,0 +1,350 @@
...
+
+            //
+            // Making it this far means we can create the connection!
+            //
+            CxPlatSocketDelete(TestSocket);
+            break;
+
+        } while (TRUE);
+
+        Status = MsQuicConnectionOpen(Registration, Handler, Context, &Connections[i]);
```
nibanks, line 292: Call `QuicConnAlloc` directly. Another thing: By default, creating a connection uses the "current partition" (i.e., current processor) to affinitize the connection, which isn't really what you want. You want it to be on the RSS partition. So, I would actually recommend modifying `QuicConnAlloc` to expose a new input parameter for `PartitionIndex` so you can set it explicitly.

src/core/connection_pool.c
```diff
@@ -0,0 +1,350 @@
...
+                "[conp] Failed to set local address on connection[%u], 0x%x",
+                i,
+                Status);
+            goto Error;
+        }
+
+        //
+        // The connection was created successfully, add it to the count for this processor.
+        //
+        ConnectionCounts[RssProcIndex]++;
```
nibanks, line 325: You're going to have to call `QuicConnStart` otherwise the port you tried to find that was open might get taken. And, really, you're asking for trouble by using a model of 'try to find an open port, then stop using it, and hope you can get it again'. You should find an initial ephemeral port, and then start creating and starting connections on ports until you've got enough successfully started connections.

anrossi: Indeed, there's a TOCTOU issue there, but the requirements from the customer ask that the connection isn't started yet so they can set their own context (which depends on information that isn't known before the call). 

I'd prefer a model that reserves the ports ahead of creating the connections so they can't be taken, but I don't think there's a cross-platform way to do that

src/inc/msquic_posix.h
```diff
@@ -318,6 +322,20 @@ QuicAddrSetPort(
...
+    )
+{
+    uint32_t Port = QuicAddrGetPort(Addr);
+    Port += 1;
+    if (Port > QUIC_ADDR_EPHEMERAL_PORT_MAX) {
+        Port = QUIC_ADDR_EPHEMERAL_PORT_MIN;
+    }
+    QuicAddrSetPort(Addr, (uint16_t)Port);
+}
+
```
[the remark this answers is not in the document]
anrossi: I was thinking this could also replace some logic in tests (in a separate PR) since incrementing the port number is something that we want to do

src/platform/datapath_raw_xdp_win.c
anrossi: I was using NONPAGED for the out parameter because the platform shouldn't assume the caller won't use this information at DISPATCH. The internal, temporary, allocations can be PAGED because they'll be freed before returning and this call is only supported at PASSIVE (because of all the allocations).

src/inc/quic_datapath.h
anrossi: This is copied from XDP, which copied from [NDIS RSS Hashing Types](https://learn.microsoft.com/en-us/windows-hardware/drivers/network/rss-hashing-types). They don't seem to be mutually exclusive but instead indicate how to handle packets that don't meet all the requirements. e.g. If a packet comes in that is not UDP/TCP, and the NIC doesn't support IPV4 or IPV6 RSS Hashing, then the packet is indicated on core0 (no RSS).

src/inc/msquic_posix.h
anrossi: Yes, it is customizable on both OSes, but these ranges are the recommended defaults via RFC. I think Linux uses a larger default range, however.
Could query for these values instead, would that be preferable?

src/inc/msquic.h
anrossi: What about when the ServerName has to be a specific string for SNI, and it resolves to multiple IP addresses (DNS load-balancing)? If the caller wants to use a specific server by address, this would be the only way to satisfy both, isn't it?

nibanks: Ok. We can keep it.

src/platform/toeplitz.c
```diff
@@ -75,7 +75,7 @@ CxPlatToeplitzHashInitialize(
     //
     // Initialize the Toeplitz->LookupTables.
     //
-    for (uint32_t i = 0; i < CXPLAT_TOEPLITZ_LOOKUP_TABLE_COUNT; i++) {
+    for (uint32_t i = 0; i < Toeplitz->InputSize; i++) {
```
guhetier, line 78: No `* NIBBLES_PER_BYTE` needed here?

anrossi: Thank you! I must have overlooked that during the conversion.

src/core/connection_pool.c
```diff
@@ -0,0 +1,350 @@
...
+#endif
+
+_IRQL_requires_max_(PASSIVE_LEVEL)
+QUIC_STATUS
+QUIC_API
+MsQuicConnectionPoolCreate(
+    _In_ HQUIC Registration,
+    _In_ QUIC_CONNECTION_CALLBACK_HANDLER Handler,
+    _In_opt_ void* Context,
+    _In_opt_ const char* ServerName,
```
guhetier, line 26: The ConnectionStart api takes only a "ServerName" (which can be an IP literal).
Should we stay consistent and have a single mandatory parameter instead of two optional ones? It would leave less room for error.

src/core/connection_pool.c
```diff
@@ -0,0 +1,350 @@
...
+    _In_opt_ const QUIC_ADDR* ServerAddress,
+    _In_ uint16_t ServerPort,
+    _In_ uint16_t NumberOfConnections,
+    _Out_writes_bytes_(NumberOfConnections * sizeof(HQUIC))
+        HQUIC** ConnectionPool
+    )
+{
```
guhetier, line 33: IMO, this is a very big function (>300 lines). It could easily be split into multiple helpers to be easier to understand and less error prone (keep resource management and error handling more scoped, limit the number of imbricated scopes):

At least these would be easy to split:
- Getting the RSS config
- Testing a port

src/core/connection_pool.c
anrossi: I don't want to only assert in case there's a change in behavior in the future. I don't want users of the API to be doing the wrong thing.
But I will add an assert that this shouldn't fail in our test runs so we can catch behavior changes.

src/core/connection_pool.c
anrossi: I agree. I've changed the API to take a configuration struct, and the ServerName is a required parameter.

src/inc/quic_toeplitz.h
```diff
@@ -16,11 +16,23 @@ extern "C" {
 
 //
 // The size (in bytes) of the input.
+//
+typedef uint32_t CXPLAT_TOEPLITZ_INPUT_SIZE;
```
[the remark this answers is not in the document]
anrossi: I was split on whether this should be a new enum or not. It seems like there were conflicting precedents for which uses the *_MAX variable, so I went with this. But an enum is fine too. I just wanted to try to prevent users from using random values, so I defined a type.

src/core/connection.c
```diff
@@ -73,7 +74,8 @@ QuicConnAlloc(
     // partition can be updated accordingly.
     //
     const uint16_t PartitionIndex =
-        IsServer ? Packet->PartitionIndex : QuicLibraryGetCurrentPartition();
+        DesiredPartitionIndex != NULL ? *DesiredPartitionIndex :
+            IsServer ? Packet->PartitionIndex : QuicLibraryGetCurrentPartition();
```
nibanks, line 78: Make the server pass in `Packet->PartitionIndex` so you can get rid of the multiple `?` checks.

src/core/connection.c
```diff
@@ -1737,7 +1739,8 @@ QuicConnStart(
     _In_ QUIC_CONFIGURATION* Configuration,
     _In_ QUIC_ADDRESS_FAMILY Family,
     _In_opt_z_ const char* ServerName,
-    _In_ uint16_t ServerPort // Host byte order
+    _In_ uint16_t ServerPort, // Host byte order
+    _In_ uint32_t StartFlags
```
nibanks, line 1743: We should use a well-defined enum for flags, including a definition for NONE instead of passing zero other places.

src/inc/msquic.h
```diff
@@ -1605,6 +1605,49 @@ QUIC_STATUS
     _In_opt_ void* ClientSendContext
     );
 
+//
+// Connection Pool API
```
nibanks, line 1609: Wrap all this with PREVIEW define.

src/inc/msquic.h
```diff
@@ -1605,6 +1605,49 @@ QUIC_STATUS
...
+DEFINE_ENUM_FLAG_OPERATORS(QUIC_CONNECTION_POOL_FLAGS);
+
+typedef struct QUIC_CONNECTION_POOL_CONFIG {
+    _In_ HQUIC Registration;
+    _In_ HQUIC Configuration;
+    _In_ QUIC_CONNECTION_CALLBACK_HANDLER Handler;
+    _In_reads_opt_(NumberOfConnections)
+        void** Context;
+    _In_ const char* ServerName;
+    _In_opt_ const QUIC_ADDR* ServerAddress;
```
nibanks, line 1625: For things that are optional, just put a comment after it saying so.

src/inc/msquic_posix.h
nibanks: Let's start with what you have. But I see no reason to define this in 3 different platform-specific places, considering they are all the same. Let's move this to quicdef.h or platform.h.

src/platform/datapath_raw_xdp_linux.c
```diff
@@ -1375,3 +1375,25 @@ CxPlatQueueTxIoEventComplete(
...
+    return QUIC_STATUS_NOT_SUPPORTED;
+}
+
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+CxPlatDataPathRssConfigFree(
+    _In_ CXPLAT_RSS_CONFIG* RssConfig
+    )
+{
+    CXPLAT_FRE_ASSERTMSG(FALSE, "CxPlatDataPathRssConfigFree not supported");
```
nibanks, line 1398: I think you'll need an `UNREFERENCED_PARAMETER` here still.

src/core/connection_pool.c
```diff
@@ -0,0 +1,380 @@
...
+    QUIC_CONNECTION** Connections = NULL;
+    CXPLAT_SOCKET* Socket = NULL;
+    CXPLAT_RSS_CONFIG* RssConfig = NULL;
+    CXPLAT_PROCESSOR_INFO* RssProcessors = NULL;
+    uint32_t* ConnectionCounts = NULL;
+    CXPLAT_TOEPLITZ_HASH ToeplitzHash;
+    QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
+    uint32_t CreatedConnections = 0;
+
+    if (Config->NumberOfConnections == 0 || Config->ServerPort == 0 || ConnectionPool == NULL) {
```
nibanks, line 39: Check that `Config` isn't `NULL` either?

src/core/connection_pool.c
nibanks: Let's move the logic to get a local address and interface index into its own function.

src/core/connection_pool.c
```diff
@@ -0,0 +1,380 @@
...
+                continue;
+            }
+
+            //
+            // The connection was created successfully, add it to the count for this processor.
+            //
+            ConnectionCounts[RssProcIndex]++;
+
+            break;
+
```
nibanks, line 343:
```suggestion
            ConnectionCounts[RssProcIndex]++;
            break;
```

src/core/connection_pool.c
anrossi: I thought we avoided single-use helper functions, but I don't feel strongly about it.

src/core/connection_pool.c
```diff
@@ -0,0 +1,380 @@
...
+            ConnPoolRssSecretKeyTooLong,
+            "[conp] RSS secret key too long, 0x%x",
+            Status);
+        goto Error;
+    }
+    CxPlatCopyMemory(&ToeplitzHash.HashKey, RssConfig->RssSecretKey, RssConfig->RssSecretKeyLength);
+    ToeplitzHash.InputSize = CXPLAT_TOEPLITZ_INPUT_SIZE_IP;
+    CxPlatToeplitzHashInitialize(&ToeplitzHash);
+
+    Connections =
```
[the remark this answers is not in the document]
anrossi: To prevent leaking partial state so the app only ever sees a successful output or none.

src/core/connection_pool.c
guhetier: I don't know where this "avoid single-use helper" rule come from ( as in I have no context about what it tries to help with, I'm interested if you know more). In my experience:
- avoiding to add single use helpers *in a "util / common" file is valuable to avoid cluttering. If it is used only in one place, keep it local
- as for everything, there is a balance. It is just as bad having to jump between 50 different 3 lines functions than reading 500 lines long function.
- but breaking down a long function in a succession of sub-tasks, each implemented by a helper function, clarifies what the function does (helper function names are self-documenting), enforce some abstraction between independent tasks (ressource management, error handling, variable being used...)

nibanks: I am a proponent for breaking this down into more easily testable and understandable functions.

src/platform/datapath_raw_xdp_win.c
anrossi: I think I'm going to leave this here. Why would we want less validation? It's on the control path, so this isn't performance sensitive.

src/core/connection_pool.c
anrossi: Well I've split out some of the functions. I'll go over it one more time and see if I can split out anything else that's overly complex.

src/inc/quic_trace.h
```diff
@@ -104,6 +104,7 @@ typedef enum QUIC_TRACE_API_TYPE {
     QUIC_TRACE_API_CONNECTION_COMPLETE_RESUMPTION_TICKET_VALIDATION,
     QUIC_TRACE_API_CONNECTION_COMPLETE_CERTIFICATE_VALIDATION,
     QUIC_TRACE_API_STREAM_PROVIDE_RECEIVE_BUFFERS,
+    QUIC_TRACE_API_CONNECTION_POOL_CREATE,
```
nibanks, line 107: ETW manifest, windbg and wpa plugins will need to be updated accordingly.

src/core/connection_pool.c
```diff
@@ -0,0 +1,508 @@
...
+                Connections[i]->ClientContext = Config->Context[i];
+            }
+
+            //
+            // Set parameters on the connections before starting them.
+            //
+            if (Config->CibirIds) {
+                uint8_t True = TRUE;
+                Status =
+                    QuicConnParamSet(
```
nibanks, line 377: Part of me thinks it's overkill calling QuicConnParamSet for each of these, instead of just setting the parameters directly.

src/core/connection_pool.c
```diff
@@ -0,0 +1,508 @@
...
+
+#ifdef QUIC_CLOG
+#include "connection_pool.c.clog.h"
+#endif
+
+#define MAX_CONNECTION_POOL_RETRY_MULTIPLIER 2
+
+_IRQL_requires_max_(PASSIVE_LEVEL)
+QUIC_STATUS
+QuicConnPoolGetStartingLocalAddress(
```
nibanks, line 23: Please add a test ASAP so all the warnings about uncovered lines go away so I can review more easily 😄

src/test/bin/winkernel/control.cpp
```diff
@@ -528,6 +528,7 @@ size_t QUIC_IOCTL_BUFFER_SIZES[] =
     sizeof(INT32),                           // IOCTL_QUIC_RUN_TEST_ADDR_FUNCTIONS
     0,
     0,
+    sizeof(QUIC_RUN_CONNECTION_POOL_CREATE_PARAMS),
```
nibanks, line 531: Need to update the big `switch` statement below to call the new test function.

src/core/library.c
```diff
@@ -1802,6 +1803,8 @@ MsQuicOpenVersion(
 
     Api->DatagramSend = MsQuicDatagramSend;
 
+    Api->ConnectionPoolCreate = MsQuicConnectionPoolCreate;
```
anrossi, line 1806: This needs to be wrapped in the preview features flag

src/inc/quic_trace.h
anrossi: I updated the manifest, but several of our preview features are not added to the windbg or WPA plugins. 
I want to add them all together in a later PR so this PR doesn't also contain work for e.g. providing receive buffers.

src/core/library.c
anrossi: Well it builds fine when preview features are disabled, so I guess I don't need that

src/core/connection_pool.c
anrossi: `QuicConnParamSet` does some parameter checks, which we need anyway in case the caller passed bad info, before copying the info (particularly for CIBIR).  That doesn't seem like overkill to me.
Sure I could copy the local and remote addresses directly if it seems like such a burden, but I think it keeps the code looking clearer than manually copying.

src/core/connection_pool.c
```diff
@@ -0,0 +1,490 @@
...
+    if (QUIC_FAILED(Status)) {
+        goto Error;
+    }
+
+    CxPlatSocketGetLocalAddress(Socket, LocalAddress);
+
+Error:
+    if (Socket != NULL) {
+        CxPlatSocketDelete(Socket);
+    }
```
nibanks, line 41:
```suggestion
    QUIC_STATUS Status =
        CxPlatSocketCreateUdp(MsQuicLib.Datapath, &UdpConfig, &Socket);
    if (QUIC_SUCCEEDED(Status)) {
        CXPLAT_DBG_ASSERT(Socket != NULL);
        CxPlatSocketGetLocalAddress(Socket, LocalAddress);
        CxPlatSocketDelete(Socket);
    }
```
This change eliminates the 'you don't have code coverage' warning for the error case. Now there is no error case 😄

src/core/connection_pool.c
```diff
@@ -0,0 +1,490 @@
...
+            "[conp] Failed to find local address, 0x%x",
+            Status);
+        goto Error;
+    }
+
+Error:
+
+    if (Addresses != NULL) {
+        CXPLAT_FREE(Addresses, QUIC_POOL_DATAPATH_ADDRESSES);
+    }
```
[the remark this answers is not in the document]
nibanks: Please note, I fixed the `*InterfaceIndex` check above, when comparing it to zero. You were checking the pointer, not the value.

src/core/connection_pool.c
```diff
@@ -0,0 +1,490 @@
...
+{
+    QUIC_CONNECTION** Connections = (QUIC_CONNECTION**)ConnectionPool;
+    CXPLAT_RSS_CONFIG* RssConfig = NULL;
+    uint32_t* RssProcessors = NULL;
+    uint32_t* ConnectionCounts = NULL;
+    CXPLAT_TOEPLITZ_HASH ToeplitzHash;
+    QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
+    uint32_t CreatedConnections = 0;
+
+    if (Config == NULL || Config->NumberOfConnections == 0 || Config->Handler == NULL ||
```
nibanks, line 109: Can we add at least one test case that triggers this code path?

src/core/connection_pool.c
```diff
@@ -0,0 +1,490 @@
...
+        Config->ServerName == NULL || Config->ServerPort == 0 || ConnectionPool == NULL ||
+        (Config->CibirIds != NULL && Config->CibirIdLength == 0)) {
+        Status = QUIC_STATUS_INVALID_PARAMETER;
+        QuicTraceLogError(
+            ConnPoolInvalidParam,
+            "[conp] Invalid parameter, 0x%x",
+            Status);
+        goto Error;
```
nibanks, line 117: Just return instead? Because otherwise you have an unmatched entry/exit API log...

src/platform/unittest/ToeplitzTest.cpp
```diff
@@ -0,0 +1,194 @@
...
+            Data = new uint8_t[Length];
+
+            for (uint16_t i = 0; i < Length; ++i) {
+                Data[i] =
+                    (DecodeHexChar(HexBytes[i * 2]) << 4) |
+                    DecodeHexChar(HexBytes[i * 2 + 1]);
+            }
+        }
+
+        QuicBuffer(const QuicBuffer&) = delete;
```
guhetier, line 41: Need to delete the `operator=` too (rule of 5).

src/platform/unittest/ToeplitzTest.cpp
```diff
@@ -0,0 +1,194 @@
...
+        )
+    {
+        const QuicBuffer KeyBuffer(HashKey);
+
+        CXPLAT_TOEPLITZ_HASH ToeplitzHash{};
+        CxPlatCopyMemory(ToeplitzHash.HashKey, KeyBuffer.Data, KeyBuffer.Length);
+        ToeplitzHash.InputSize = CXPLAT_TOEPLITZ_INPUT_SIZE_IP;
+        CxPlatToeplitzHashInitialize(&ToeplitzHash);
+
+        for (uint32_t i = 0; i < TestCaseCount; i++) {
```
guhetier, line 76: nit:
I think the loop should be kept in the caller (or no loop, and several calls with each test cases).
It adds complexity to understand the test, keeping track of 5 arrays, etc...

The test case could read:

```
ValidateRssToeplitzHash("foo", "bar", "baz", ...);
ValidateRssToeplitzHash("foo2", "bar2", "baz2, ...);
...
```

otherwise, which is both simpler, shorter, and removes room from mis-counting errors.
If the initialization can't be repeated, it could be split in a "prepare" function.

src/inc/msquic.h
```diff
@@ -1605,6 +1605,56 @@ QUIC_STATUS
...
+DEFINE_ENUM_FLAG_OPERATORS(QUIC_CONNECTION_POOL_FLAGS);
+
+typedef struct QUIC_CONNECTION_POOL_CONFIG {
+    _In_ HQUIC Registration;
+    _In_ HQUIC Configuration;
+    _In_ QUIC_CONNECTION_CALLBACK_HANDLER Handler;
+    _Field_size_opt_(NumberOfConnections)
+        void** Context;                         // Optional
+    _In_ const char* ServerName;
+    _In_opt_ const QUIC_ADDR* ServerAddress;    // Optional
```
guhetier, line 1628: I am still not sure I understand why the server address is provided there, instead of relying only on the server name like in the existing connection start API (with an ip address being potentially passed in text form as the server name).

IMO, we should be consistent in the API pattern.

docs/api/QUIC_CONNECTION_POOL_CONFIG.md
```diff
@@ -0,0 +1,90 @@
...
+The UDP port on the server to connect to, in host byte order.
+
+`NumberOfConnections`
+
+The number of connections to create in this connection pool.
+If this number is greater than the number of RSS CPUs configured on the system, then multiple connections will end up on the same CPU, potentially hurting performance.
+
+`CibirIds`
+
+An optional pointer to an array of pointers to `uint8_t`s specifying the CIBIR ID to use for each connection. All CIBIR IDs must be the same length.
```
guhetier, line 74: Just curious, what is a CIBIR? Google isn't being helpful.

src/core/connection_pool.c
```diff
@@ -0,0 +1,521 @@
...
+    QUIC_CONNECTION** Connections = (QUIC_CONNECTION**)ConnectionPool;
+    CXPLAT_RSS_CONFIG* RssConfig = NULL;
+    uint32_t* RssProcessors = NULL;
+    uint32_t* ConnectionCounts = NULL;
+    const char* ServerNameCopy = NULL;
+    CXPLAT_TOEPLITZ_HASH ToeplitzHash;
+    QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
+    uint32_t CreatedConnections = 0;
+
+    if (Config == NULL || ConnectionPool == NULL || Config->Registration == NULL ||
```
guhetier, line 120: nit: That if test is very hard to read.
It could easily be split in several tests and potentially in a "validateConfig" function.

src/core/connection_pool.c
```diff
@@ -0,0 +1,521 @@
...
+    if (RssConfig->RssSecretKeyLength > CXPLAT_TOEPLITZ_KEY_SIZE_MAX) {
+        Status = QUIC_STATUS_INTERNAL_ERROR;
+        QuicTraceLogError(
+            ConnPoolRssSecretKeyTooLong,
+            "[conp] RSS secret key too long (%u bytes), 0x%x",
+            RssConfig->RssSecretKeyLength,
+            Status);
+        goto Error;
+    } else if (RssConfig->RssSecretKeyLength < CXPLAT_TOEPLITZ_KEY_SIZE_MIN) {
+        Status = QUIC_STATUS_INVALID_STATE;
```
guhetier, line 228: Shouldn't this one also be `INTERNAL_ERROR`?
Is there a scenario that can lead here, or is this check to be cautious but essentially an assert?

docs/api/QUIC_CONNECTION_POOL_CONFIG.md
nibanks: https://datatracker.ietf.org/doc/html/draft-banks-quic-cibir

scripts/prepare-machine.ps1
```diff
@@ -258,6 +258,9 @@ function Install-DuoNic {
         $DuoNicScript = (Join-Path $DuoNicPath duonic.ps1)
         if (!(Test-Path $DuoNicScript)) { Write-Error "Missing file: $DuoNicScript" }
         Invoke-Expression "cmd /c `"pushd $DuoNicPath && pwsh duonic.ps1 -Install`""
+        $RssSeedPath = (Join-Path $SetupPath tcprssseed.exe)
+        if (!(Test-Path $RssSeedPath)) { Write-Error "Missing file: $RssSeedPath" }
+        Invoke-Expression "$RssSeedPath set aa55aa55aa55aa55aa55aa55aa55aa55aa55aa55aa55aa55aa55aa55aa55aa55aa55aa55aa55aa55"
```
nibanks, line 263: Could you put a comment to say _why_ we need this?

src/platform/datapath_raw_xdp_win.c
nibanks: I would prefer a DBG assert instead of a check for all internal functions, independent of if it _might_ be exposed later.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1882,3 +1882,133 @@ CxPlatIoXdpShutdownEventComplete(
...
+        Status = QUIC_STATUS_INVALID_PARAMETER;
+        goto Error;
+    }
+
+    *RssConfig = NULL;
+
+    HANDLE InterfaceHandle = NULL;
+    HRESULT Result = XdpInterfaceOpen(InterfaceIndex, &InterfaceHandle);
+    if (FAILED(Result)) {
+        QuicTraceLogError(
```
nibanks, line 1908: In general, we should be using the following format for error traces (not logs) so they are collected in the low volume scenario:
```c
        QuicTraceEvent(
            LibraryErrorStatus,
            "[ lib] ERROR, %u, %s.",
            WsaError,
            "WSAStartup");
   ```

File: src/test/MsQuicTests.h [hunk 50]
```diff
@@ -1366,4 +1375,18 @@ typedef struct {
...
+    bool XdpSupported;
+    bool TestCibirSupport;
+};
+
+#define IOCTL_QUIC_RUN_CONNECTION_POOL_CREATE \
+    QUIC_CTL_CODE(131, METHOD_BUFFERED, FILE_WRITE_DATA)
+    // QUIC_RUN_CONNECTION_POOL_CREATE_PARAMS
+
+#define IOCTL_QUIC_RUN_VALIDATE_CONNECTION_POOL_CREATE \
+QUIC_CTL_CODE(132, METHOD_BUFFERED, FILE_WRITE_DATA)
```
nibanks, line 1390:
```suggestion
#define IOCTL_QUIC_RUN_VALIDATE_CONNECTION_POOL_CREATE \
    QUIC_CTL_CODE(132, METHOD_BUFFERED, FILE_WRITE_DATA)
```

src/test/lib/HandshakeTest.cpp
```diff
@@ -4033,3 +4033,202 @@ QuicTestHandshakeSpecificLossPatterns(
...
+    ListenerMultiAcceptConnection(
+        _In_ TestListener* Listener,
+        _In_ HQUIC ConnectionHandle
+        )
+    {
+        auto* This = (ServerMultiAcceptContext*)Listener->Context;
+        if (This->StartedConnectionCount > This->MaxConnections) {
+            return false;
+        }
+        This->Connections[This->StartedConnectionCount++] =
```
nibanks, line 4055: What happens if/when this is called in parallel on two threads/cpus?

src/core/connection_pool.c
```diff
@@ -0,0 +1,521 @@
...
+
+#define MAX_CONNECTION_POOL_RETRY_MULTIPLIER 2
+
+const char*
+QuicConnPoolAllocServerNameCopy(
+    _In_ const char* ServerName,
+    _In_ size_t ServerNameLength
+    )
+{
+    char* ServerNameCopy = CXPLAT_ALLOC_NONPAGED(ServerNameLength + 1, QUIC_POOL_SERVERNAME);
```
nibanks, line 27: nit: offsets in this function are all off.

src/core/connection_pool.c
```diff
@@ -0,0 +1,521 @@
...
+
+    Copyright (c) Microsoft Corporation.
+    Licensed under the MIT License.
+
+Abstract:
+
+    Definitions for the MsQuic Connection Pool API, which allows clients to
+    create a pool of connections that are spread across RSS cores.
+
+--*/
```
nibanks, line 11: Code coverage is still claiming all the lines in this file are uncovered...

src/core/connection_pool.c
```diff
@@ -0,0 +1,521 @@
...
+QUIC_STATUS
+QuicConnPoolGetInterfaceIndexForLocalAddress(
+    _In_ QUIC_ADDR* LocalAddress,
+    _Out_ uint32_t* InterfaceIndex
+    )
+{
+    CXPLAT_ADAPTER_ADDRESS* Addresses = NULL;
+    uint32_t AddressesCount;
+
+    *InterfaceIndex = 0;
```
nibanks, line 72: Unnecessary, so please remove. This variable is defined as an `_Out_` and `QUIC_STATUS` is annotated with `_Return_type_success_(return >= 0)`.

src/core/connection_pool.c
```diff
@@ -0,0 +1,521 @@
...
+    if (ServerNameLength == QUIC_MAX_SNI_LENGTH + 1) {
+        Status = QUIC_STATUS_INVALID_PARAMETER;
+        QuicTraceLogError(
+            ConnPoolServerNameTooLong,
+            "[conp] ServerName too long (%llu bytes)",
+            ServerNameLength);
+        return Status;
+    }
+
+    QuicTraceEvent(
```
nibanks, line 147: FYI, in general, for other functions in api.c we are logging api enter/exit even for INVALID_PARAMETER cases.

src/core/connection_pool.c
```diff
@@ -0,0 +1,521 @@
...
+
+    uint32_t InterfaceIndex;
+    Status = QuicConnPoolGetInterfaceIndexForLocalAddress(&LocalAddress, &InterfaceIndex);
+    if (QUIC_FAILED(Status)) {
+        goto Error;
+    }
+
+    Status = CxPlatDataPathRssConfigGet(InterfaceIndex, &RssConfig);
+    if (QUIC_FAILED(Status)) {
+        QuicTraceLogError(
```
nibanks, line 200: In general, you don't need to log an error if the function itself already logs an error.

src/core/connection_pool.c
nibanks: I believe this is because code coverage doesn't run with xdp. Probably should change that...

src/core/connection_pool.c
anrossi: We need to initialize it to 0 so that line 88 can detect that no interface was found.

src/core/connection_pool.c
anrossi: Right, but I can't dereference an untrusted pointer without validating it isn't NULL. So either I log entry after the validation, or I log entry without any parameter, which seems to be the requirement for this scenario.

src/test/bin/winkernel/control.cpp
```diff
@@ -1520,6 +1523,20 @@ QuicTestCtlEvtIoDeviceControl(
...
         QuicTestCtlRun(QuicTestStreamAppProvidedBuffersZeroWindow());
         break;
+
+    case IOCTL_QUIC_RUN_CONNECTION_POOL_CREATE:
+        CXPLAT_FRE_ASSERT(Params != nullptr);
+        QuicTestCtlRun(
+            QuicTestConnectionPoolCreate(
+                Params->ConnPoolCreateParams.Family,
+                Params->ConnPoolCreateParams.NumberOfConnections,
+                Params->ConnPoolCreateParams.XdpSupported == TRUE,
```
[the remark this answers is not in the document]
anrossi: We had a pattern (that seems to have been abandoned) where boolean values from user mode were converted for use in kernel mode with `!= FALSE` (which is what I should have done).
These are "nice" booleans, so I should be able to use them directly. Let's try that

src/inc/msquic.h
```diff
@@ -1605,6 +1605,56 @@ QUIC_STATUS
...
+
+typedef enum QUIC_CONNECTION_POOL_FLAGS {
+    QUIC_CONNECTION_POOL_FLAG_NONE =                            0x00000000,
+    QUIC_CONNECTION_POOL_FLAG_CLOSE_CONNECTIONS_ON_FAILURE =    0x00000001,
+} QUIC_CONNECTION_POOL_FLAGS;
+
+DEFINE_ENUM_FLAG_OPERATORS(QUIC_CONNECTION_POOL_FLAGS);
+
+typedef struct QUIC_CONNECTION_POOL_CONFIG {
+    _In_ HQUIC Registration;
```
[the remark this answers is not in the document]
anrossi: Fair point. I'll convert the ones that can be to `_Field_*` and remove the rest.

src/inc/msquic.h
anrossi: So the ServerName doubles as the SNI name, which is used in TLS. You *must* specify the ServerName, because TLS 1.3 requires SNI (at least our implementation does).
In some deployments, you can have DNS-based load balancing, where your DNS name could resolve to multiple IP addresses. A caller may want to control which server they connect to specifically. Our solution to that is to allow the caller to specify the SNI as the ServerName, but then set the RemoteAddress via the SetParam interface before calling Start.
Because this API creates the connection *and* starts it, leaving the caller no opportunity to call the SetParam interface, we must expose that via this API.

src/core/connection_pool.c
anrossi: I'll break it up to make it easier to read, but this is an area where I disagree with making a helper function.

src/inc/msquic.h
guhetier: Thanks for the details.
It would probably be worth adding this to the documentation, to clarify when the `ServerAddress` is expected and what is a valid value (I assume something will fail if the DNS resolution doesn't return the specific address in its list?)

src/core/connection_pool.c
guhetier: I see...
`INVALID_STATE` is generally used when the caller did something at the wrong time or in the wrong order.

Do you know if the user could do something that affect the rss key length? If yes, INVALID_STATE can make sense.
But if it means the OS is in a wrong config (or a config ew don't expect), I think I'd keep INTERNAL_ERROR (or some other more precise error code)

src/core/connection_pool.c
```diff
@@ -0,0 +1,521 @@
...
+                Config->ServerName,
+                ServerNameLength);
+        if (ServerNameCopy == NULL) {
+            Status = QUIC_STATUS_OUT_OF_MEMORY;
+            goto Error;
+        }
+
+        for (; RetryCount < MaxCreationRetries; RetryCount++) {
+
+            uint32_t NewPort = QuicAddrGetPort(&LocalAddress) + 1;
```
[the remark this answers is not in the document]
anrossi: I'll see what I can do. I don't like single-use helper functions because they force the reader to jump around the code just to understand what's going on, but I agree that this function is long and it has different parts which are logically separable.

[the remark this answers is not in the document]
anrossi: Well I looked at this function and thought about what parts can be separated cleanly and did a refactor. Let me know what you think of it Monday.

src/test/MsQuicTests.h
anrossi: Blame Copilot :P

src/test/lib/HandshakeTest.cpp
anrossi: That's when the fun begins.
(good catch, I'll use an interlocked increment)

src/core/connection_pool.c
```diff
@@ -0,0 +1,636 @@
...
+    )
+{
+    QUIC_STATUS Status =
+        QuicConnAlloc(
+            Registration,
+            NULL,
+            NULL,
+            PartitionIndex,
+            Connection);
+    if (QUIC_FAILED(Status)) {
```
vgmahajanshetty, line 245: QuicConnAlloc() can fail for reasons like no memory available. Is it worth retrying on next ports for such failures (at line 590)?

src/core/connection_pool.c
```diff
@@ -0,0 +1,636 @@
...
+                    &ResolvedRemoteAddress,
+                    &LocalAddress,
+                    ServerNameCopy,
+                    Config->ServerPort,
+                    Config->Family,
+                    Config->CibirIdLength,
+                    Config->CibirIds ? Config->CibirIds[i] : NULL,
+                    &Connections[i]);
+                if (QUIC_FAILED(Status)) {
+                    continue;
```
vgmahajanshetty, line 590: If QuicConnAlloc() is successful inside QuicConnPoolTryCreateConnection, then should you free that memory and close connection here?

src/core/connection_pool.c
anrossi: Fair point, out of memory is not an error that will be solved by retrying.

src/core/connection_pool.c
anrossi: I think if the QuicConnAlloc is successful, but an error is returned by a later operation inside TryCreateConnection, then TryCreateConnection should clean up the connection.

src/core/connection_pool.c
anrossi: I'll make a separate PR to do that. In the meantime, when reviewing, you can go into the upper right `...` menu and uncheck "Show Annotations". 
Just remember to uncheck that once you're done reviewing this PR :)

src/core/connection_pool.c
vgmahajanshetty: I think QuicConnPoolTryCreateConnection() is not cleaning up the connection, in cases like QuicConnParamSet() fails then "goto Error" is hit; pls check.

docs/api/QUIC_CONNECTION_POOL_CONFIG.md
```diff
@@ -0,0 +1,93 @@
...
+    _In_ uint8_t CibirIdLength;                 // Optional
+    _In_ QUIC_CONNECTION_POOL_FLAGS Flags;
+} QUIC_CONNECTION_POOL_CONFIG;
+```
+
+# Members
+
+`Registration`
+
+An opened Registration for creating the connection in.
```
nibanks, line 40:
```suggestion
The valid handle to an open registration object.
```
Copied from ConnectionOpen.md

docs/api/QUIC_CONNECTION_POOL_CONFIG.md
```diff
@@ -0,0 +1,93 @@
...
+
+`Configuration`
+
+An opened Configuration that provides the settings used to create each connection in the pool.
```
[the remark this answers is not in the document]
nibanks: I recommend you do pretty much the same for the rest of the parameters, so we can have consistent docs.

docs/api/QUIC_CONNECTION_POOL_CONFIG.md
```diff
@@ -0,0 +1,93 @@
...
+
+The number of bytes in each CIBIR ID. Not allowed to be zero if `CibirIds` is non-NULL.
+
+`Flags`
+
+Flags which affect settings on the connections or how the pool is created.
+| Flag | Effect |
+|------|--------|
+| QUIC_CONNECTION_POOL_FLAG_NONE | Nothing |
+| QUIC_CONNECTION_POOL_FLAG_CLOSE_CONNECTIONS_ON_FAILURE | Tells the API to close all *started* connections in the pool if an error occurrs while creating the pool. **Note:** The application must be able to handle having connections suddenly closed. Without this flag, the application is expected to clean up non-NULL connections when an error is returned. |
```
nibanks, line 89: Thoughts on renaming this more simply to `QUIC_CONNECTION_POOL_FLAG_CLOSE_ON_FAILURE`? It's a _connection_ pool. I don't think a user could assume it would close anything other than _connections_. And also, the docs can clarify things even more.

docs/api/QUIC_CONNECTION_POOL_CONFIG.md
```diff
@@ -0,0 +1,93 @@
...
+
+# See Also
```
nibanks, line 91: I think this and the other md should link to most everything also referenced at the end of ConnectionOpen.md and ConnectionStart.md

src/inc/msquic.h
```diff
@@ -1616,6 +1616,56 @@ QUIC_STATUS
...
+//
+// Creates a simple pool of NumberOfConnections connections, all with the same
+// Handler, and puts them in the caller-supplied array.
+// Connections are spread evenly across RSS CPUs as much as possible.
+// If NumberOfConnections is more than the number of RSS cores, then multiple
+// connections will be put on the same CPU.
+//
+typedef
+_IRQL_requires_max_(PASSIVE_LEVEL)
+_Check_return_
```
nibanks, line 1659: Why did you add this? While I don't disagree with it, it's not consistent with the rest of our API definitions. Let's leave it out for now.

src/inc/quic_platform_winkernel.h
```diff
@@ -232,6 +232,12 @@ CxPlatLogAssert(
 //
 #define CxPlatDebuggerPresent() KD_DEBUGGER_ENABLED
 
+//
+// Enable Supported Tests
+//
+
+#define QUIC_TEST_CONNECTION_POOL 1
```
nibanks, line 239: Why are you enabling this for Windows kernel?

src/core/connection_pool.c
```diff
@@ -0,0 +1,651 @@
...
+
+--*/
+
+#include "precomp.h"
+
+#ifdef QUIC_CLOG
+#include "connection_pool.c.clog.h"
+#endif
+
+#define MAX_CONNECTION_POOL_RETRY_MULTIPLIER 2
```
nibanks, line 19: Maybe we should do more...

src/core/connection_pool.c
```diff
@@ -0,0 +1,651 @@
...
+        (QUIC_CONN_POOL_RSS_PROC_INFO*)CXPLAT_ALLOC_PAGED(
+            RssConfig->RssIndirectionTableCount * sizeof(QUIC_CONN_POOL_RSS_PROC_INFO),
+            QUIC_POOL_TMP_ALLOC);
+    if (RssProcessors == NULL) {
+        Status = QUIC_STATUS_OUT_OF_MEMORY;
+        QuicTraceEvent(
+            AllocFailure,
+            "Allocation of '%s' failed. (%llu bytes)",
+            "RSS Processor List",
+            RssConfig->RssIndirectionTableCount);
```
nibanks, line 61:
```suggestion
            RssConfig->RssIndirectionTableCount * sizeof(QUIC_CONN_POOL_RSS_PROC_INFO));
```

src/core/connection_pool.c
```diff
@@ -0,0 +1,651 @@
...
+static
+_IRQL_requires_max_(PASSIVE_LEVEL)
+QUIC_STATUS
+QuicConnPoolGetStartingLocalAddress(
+    _In_ QUIC_ADDR* RemoteAddress,
+    _Out_ QUIC_ADDR* LocalAddress
+    )
+{
+    CXPLAT_SOCKET* Socket = NULL;
+    CXPLAT_UDP_CONFIG UdpConfig = { .RemoteAddress = RemoteAddress, };
```
nibanks, line 164:
```suggestion
    CXPLAT_UDP_CONFIG UdpConfig = { .RemoteAddress = RemoteAddress, 0 };
```
Let's be a bit more specific here

src/core/connection_pool.c
```diff
@@ -0,0 +1,651 @@
...
+            Registration,
+            NULL,
+            NULL,
+            PartitionIndex,
+            Connection);
+    if (QUIC_FAILED(Status)) {
+        goto Error;
+    }
+    (*Connection)->ClientCallbackHandler = Handler;
+    if (Context != NULL) {
```
nibanks, line 252: No real need to check if NULL here.

src/core/connection_pool.c
```diff
@@ -0,0 +1,651 @@
...
+    QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
+    uint32_t CreatedConnections = 0;
+
+    QuicTraceEvent(
+        ApiEnter,
+        "[ api] Enter %u (%p).",
+        QUIC_TRACE_API_CONNECTION_POOL_CREATE,
+        NULL);
+
+
```
nibanks, line 373:
```suggestion

```

src/core/connection_pool.c
```diff
@@ -0,0 +1,651 @@
...
+    if (Config == NULL || ConnectionPool == NULL) {
+        Status = QUIC_STATUS_INVALID_PARAMETER;
+        QuicTraceEvent(
+            LibraryErrorStatus,
+            "[ lib] ERROR, %u, %s.",
+            Status,
+            "Connection Pool Parameter");
+        return Status;
```
nibanks, line 381: You're not logging API exit by doing a `return` now.

nibanks: Same for below.

src/core/connection_pool.c
```diff
@@ -0,0 +1,651 @@
...
+                    Config->Family,
+                    Config->CibirIdLength,
+                    Config->CibirIds ? Config->CibirIds[i] : NULL,
+                    &Connections[i]);
+
+            //
+            // The connection either owns the ServerNameCopy, or it was freed.
+            //
+            ServerNameCopy = NULL;
+            if (Status == QUIC_STATUS_OUT_OF_MEMORY) {
```
nibanks, line 599: Personally, I wouldn't special case out of memory errors. For one, they may be transient, and two there might be many errors you _could_ special case, but it unnecessarily complicates the logic for such rare edge cases.

docs/api/QUIC_CONNECTION_POOL_CONFIG.md
```diff
@@ -0,0 +1,93 @@
...
+The number of connections to create in this connection pool.
+If this number is greater than the number of RSS CPUs configured on the system, then multiple connections will end up on the same CPU, potentially hurting performance.
+
+`CibirIds`
+
+An optional pointer to an array of pointers to `uint8_t`s specifying the [CIBIR ID](https://datatracker.ietf.org/doc/html/draft-banks-quic-cibir) to use for each connection. All CIBIR IDs must be the same length.  Not allowed to be non-NULL if `CibirIdLength` is zero.
+
+`CibirIdLength`
+
+The number of bytes in each CIBIR ID. Not allowed to be zero if `CibirIds` is non-NULL.
```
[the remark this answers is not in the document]
anrossi: that brings up an interesting question about our API: do we test if the pointer is NULL, or do we look at the length field?

src/core/connection_pool.c
anrossi: Are memory errors not fatal? Should we retry when there's no memory? Most programs respond to OUT_OF_MEMORY by reducing allocations, when they don't just crash outright.

src/inc/msquic.h
anrossi: CoPilot being "helpful"

src/core/connection_pool.c
anrossi: It is cleaning up the connection (by dropping the last reference on the connection), but there's associated state with the connection that is not getting cleaned up. So I'll fix that.

docs/api/QUIC_CONNECTION_POOL_CONFIG.md
anrossi: Sounds reasonable to me.

src/core/connection_pool.c
anrossi: also, each success resets the retry counter, so it's actually ((2 * NumberOfConnections) - 1) * NumberOfConnections worst case number of retries.

src/core/connection_pool.c
anrossi: I prefer the simplicity to the special case as well

docs/api/QUIC_CONNECTION_POOL_CONFIG.md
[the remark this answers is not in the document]
anrossi: I am going to keep this check in the code because it prevents careless bugs in callers of this API. I can remove the documentation about the restriction if you think it's too confusing.

src/inc/quic_platform_winuser.h
```diff
@@ -222,6 +222,12 @@ GetModuleHandleW(
 #define CXPLAT_PASSIVE_CODE() CXPLAT_DBG_ASSERT(CXPLAT_IRQL() == PASSIVE_LEVEL)
 #define CXPLAT_AT_DISPATCH() FALSE
 
+//
+// Enable Supported Tests
+//
+
+#define QUIC_TEST_CONNECTION_POOL 1
```
guhetier, line 229: It seems weird to enable tests based on a macro in a product header. Shouldn't we enable / disable the test based on the usual macros for platform and UM vs KM? (`_WIN32` I believe?)

src/tools/sample/sample.c
```diff
@@ -994,6 +1003,90 @@ RunClient(
...
+    PoolConfig.ServerName = Target;
+    PoolConfig.ServerPort = UdpPort;
+    PoolConfig.NumberOfConnections = (uint16_t)NumberOfConnections;
+
+    printf("Connection Pool Connecting...\n");
+
+    //
+    // Start the connections to the server.
+    //
+    if (QUIC_FAILED(Status = MsQuic->ConnectionPoolCreate(&PoolConfig, Connections))) {
```
guhetier, line 1066: Unless that's something usual in the QUIC codebase, assignment in a if test condition is generally considered error prone.

src/tools/sample/sample.c
```diff
@@ -994,6 +1003,90 @@ RunClient(
...
+    _In_reads_(argc) _Null_terminated_ char* argv[]
+    )
+{
+    //
+    // Load the client configuration based on the "unsecure" command line option.
+    //
+    if (!ClientLoadConfiguration(GetFlag(argc, argv, "unsecure"))) {
+        return;
+    }
+    QUIC_STATUS Status;
```
guhetier, line 1022: nit: I recommend default initializing the variables - for this one, it would be pretty easy to reach the cleanup without having it set to a value.

src/tools/sample/sample.c
```diff
@@ -994,6 +1003,90 @@ RunClient(
...
+    uint32_t NumberOfConnections = 0;
+    HQUIC* Connections = NULL;
+
+    //
+    // Get the target / server name or IP from the command line.
+    //
+    const char* Target;
+    if ((Target = GetValue(argc, argv, "target")) == NULL) {
```
guhetier, line 1030: nit: The assignment can be done on the line above, directly when declaring the variable, rather than in the `if` test. Same below.

src/test/lib/HandshakeTest.cpp
```diff
@@ -4187,10 +4189,10 @@ QuicTestConnectionPoolCreate(
     ServerAddr.SetPort(ServerPort);
 
     {
-        UniquePtrArray<TestConnection*> ServerConnections(new(std::nothrow) TestConnection*[NumberOfConnections]);
+        UniquePtrArray<UniquePtr<TestConnection>> ServerConnections(new(std::nothrow) UniquePtr<TestConnection>[NumberOfConnections]);
```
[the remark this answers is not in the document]
mtfriesen: 😮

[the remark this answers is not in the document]
anrossi: I think it's here because `delete []` is different from `delete` when dealing with an array.

src/tools/sample/sample.c
anrossi: It was used elsewhere in the sample so I followed the convention.

src/tools/sample/sample.c
anrossi: Same pattern seen throughout this file. I want the sample to be consistent, even if it's a pattern that we don't want to use broadly.  We can bulk fix up the sample in a separate PR

src/test/lib/HandshakeTest.cpp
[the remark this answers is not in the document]
guhetier: That's right, but the `UniquePtr<T[]>` specialization also takes care of that. And `UniquePtrArray` doesn't properly define/delete copies and moves, it's a bug waiting for its time to shine ^^

src/inc/quic_platform_winuser.h
anrossi: We introduce a separate macro so we're not dependent upon the platform-specific build macros in general. This is so the tests/feature can be enabled for new platforms as support is added, without having to find all the places where the test was depending on the platform-specific macro.

src/test/lib/HandshakeTest.cpp
```diff
@@ -4237,15 +4239,14 @@ QuicTestConnectionPoolCreate(
...
-                    // memcpy(CibirIdBuffer.get() + (i * CibirIdLength), CibirId, CibirIdLength);
                     CibirIdBuffer[(i * CibirIdLength) + (CibirIdLength - 1)] += (uint8_t)i;
                     CibirIdPtrs[i] = &CibirIdBuffer[i * CibirIdLength];
                 }
                 PoolConfig.CibirIds = CibirIdPtrs.get();
                 PoolConfig.CibirIdLength = CibirIdLength;
             }
 
-            QUIC_STATUS Status = MsQuic->ConnectionPoolCreate(&PoolConfig, Connections.get());
+            QUIC_STATUS Status = MsQuic->ConnectionPoolCreate(&PoolConfig, (HQUIC*)Connections.get());
```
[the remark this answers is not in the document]
anrossi: I don't want to waste memory with two arrays. I could add a C_ASSERT that the size of a `ConnectionScope` is the same as an `HQUIC` to ensure that any size changes in the future are caught and fixed.

src/test/lib/HandshakeTest.cpp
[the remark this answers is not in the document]
anrossi: Oh, I see the specialization now. Interesting that we have both. I updated UniquePtrArray to delete the copy and assignment operators and added the move and move-assignment operators.  We can decide in a separate PR whether to delete UniquePtrArray, or delete the specialization.

src/tools/sample/sample.c
anrossi: Usually, the static analysis catches this if it's an actual issue. But yeah, in the success case, I see this makes it through the whole function without assigning to `Status`. Fixed.

src/test/lib/HandshakeTest.cpp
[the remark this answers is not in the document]
anrossi: The point of a scope guard type is that it's the same size as the type it guards, it just tells the compiler to invoke a cleanup function when it goes out of scope. I added some static asserts on the size of the scope guard, and I removed the cast in favor of more standard language constructs, but the core assumption that the scope guard is the same as the type it guards is preserved.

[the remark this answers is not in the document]
guhetier: I don't think that is true. Ideally a scope guard can be a zero-cost abstraction (and here it is, at least for now).
Pretty often, some extra memory will be needed to enforce invariants: a function pointer, a pointer to a shared context, etc... 
For instance, a std::unique_ptr potentially has a pointer to the cleanup function. A std::shared_ptr comes with a pointer to the shared state.

"Wasting memory" is not a good reason IMO. This is test code, not an optimized hot path where we can't afford a small allocation. I don't think it is worth overriding the type system and potentially having to debug a memory corruption issue in our test code later.

(Just to be pedantic, people would call `ConnectionGuard` and RAII wrapper :). A scope guard is generally understood as an RAII wrapper made to call a generic callback at the end of a scope, like `wil::scope_exit`. But I'll see myself out)

src/inc/quic_platform_winuser.h
guhetier: I get the idea. I just find it strange to basically configure our tests from the product code:
- it takes more searching to know where the test run (the config will be distributed through potentially multiple files if we ever add it to other modes)
- test logic leaks in product logic

Do you think we could achieve something similar with a section in a test header that group those defines for all tests?
Something like:

```
#ifdef _WIN32
#define QUIC_TEST_CONNECTION_POOL 1
#endif

#ifdef LINUX
...
#endif

```

And that way we can just add more to that one place when other tests need the same kind of conditional enablement?

anrossi: Sure, I like that idea. I was looking for a place where put this define and dismayed to discover that we didn't already have a delineated place for it.

anrossi: Hang on, I think I can solve my problem without adding this new flag, simplifying things.

src/test/lib/HandshakeTest.cpp
```diff
@@ -4099,3 +4099,191 @@ QuicTestHandshakeSpecificLossPatterns(
...
+                // and XDP doesn't support the loopback interface, so the
+                // expected error is "NOT_FOUND" since XDP isn't found for that
+                // interface.
+                // In the case the platform doesn't support XDP at all,
+                // NOT_SUPPORTED is expected.
+                //
+                if (Status != QUIC_STATUS_NOT_SUPPORTED &&
+#if defined(_WIN32) && !defined(_KERNEL_MODE)
+                    Status != QUIC_STATUS_FILE_NOT_FOUND &&
+#endif
```
nibanks, line 4279: Maybe just define this on Posix to be the same thing as `QUIC_STATUS_NOT_FOUND` so you don't need this check?

nibanks (approved):

PR #4836, opened 2025-02-22
Receive buffer documentation

guhetier:
## Description

Add Quic dev focused documentation for the receive buffer component.

## Testing

N/A

## Documentation

N/A

### DIFF
--- a/docs/ReceiveBuffer.md
+++ b/docs/ReceiveBuffer.md
@@ -27,58 +27,62 @@ The caller then 'reads' data. The receive buffer reads as much contiguous data a
 It will populate the array of `QUIC_BUFFER` provided by the caller with pointers and length, as a **view of its internal buffers**.
 
 Finally, the caller 'drains' data from the receive buffer. The receive buffer discards the drained data, moving its reading head by the number of drained bytes.
-The receive buffer ensures that the `QUIC_BUFFER`s provided for a 'read' are valid until the data is drained (more on details on when discussing the buffer modes).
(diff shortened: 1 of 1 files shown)


docs/ReceiveBuffer.md
```diff
@@ -0,0 +1,223 @@
...
+                    ReadLength       Capacity                          ReadPendingLength                                 
+```
+
+This receive buffer has 1 retired chunk and 3 active chunks.
+`WrittenRanges` contains 3 data segments:
+- `[0, ReadStart + ReadLength]`, cycling through the end of Chunk 1;
+- The second segment goes over the end of Chunk 1 and the start of Chunk 2;
+- The third segment is at the start of chunk3
+
+**Remark**: The state of the receive buffer presented above would actually never happen in practice - it is so generic it violates each mode constraints.
```
mtfriesen, line 220: This is great documentation - my only nit is that instead of providing one sample of illegal state, it would be better to provide several valid examples covering the same sets of scenarios.

mtfriesen (dismissed):

docs/ReceiveBuffer.md
guhetier: That's fair. I want the "illegal" state, because I find it a great way to frame reasoning about this code (I am planning some simplifications based on it).

I though about adding "per-mode" examples, or a "timeline" to show how operations affect it, but I grew lazy. I'll try to add something so we don't only have the "invalid" case.

nibanks (approved):

docs/ReceiveBuffer.md
mtfriesen: Thanks, this is exactly what I had in mind

PR #4838, opened 2025-02-22
Rust: BufferRef type

youyuanwu:
## Description
Introduce BufferRef type to wrap ffi QUIC_BUFFER type. 
BufferRef has the same abi as QUIC_BUFFER, and it has slice accessor functionality. 
* Apis having arguments using &QUIC_BUFFER and length are converted to use &[BufferRef].
* Callbacks with &[QUIC_BUFFER] fields are converted to use &[BufferRef].

This brings us closer to have a full server client example with relatively safe Apis.

## Testing
Added tests to demonstrate how to use BufferRef to call msquic Apis and how to pass client context to ensure lifetime of buffer memories on heap.

## Documentation
NA

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -303,7 +303,7 @@ impl<'b> From<&'b mut crate::ffi::QUIC_STREAM_EVENT> for StreamEvent<'b> {
                     absolute_offset: ev.AbsoluteOffset,
                     total_buffer_length: &mut ev.TotalBufferLength,
                     buffers: unsafe {
-                        BufferRef::from_ffi_slice_ref(slice_conv(
+                        BufferRef::slice_from_ffi_ref(slice_conv(
                             ev.Buffers,
                             ev.BufferCount as usize,
                         ))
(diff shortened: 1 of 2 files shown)


nibanks (approved):

nibanks (approved):

src/types.rs
```diff
@@ -368,6 +372,57 @@ impl<'b> From<&'b mut crate::ffi::QUIC_STREAM_EVENT> for StreamEvent<'b> {
...
+/// and user needs to ensure that this ref has
+/// the same lifetime as the original buffer
+/// location.
+#[repr(transparent)]
+pub struct BufferRef(pub QUIC_BUFFER);
+
+impl BufferRef {
+    /// Get the bytes of the buffer.
+    pub fn as_bytes(&self) -> &[u8] {
+        unsafe { slice_conv(self.0.Buffer, self.0.Length as usize) }
```
guhetier, line 387: I come after the PR is merged, but isn't this breaking rust safety?
When exiting the unsafe block, nothing guarantees the slice refers to valid memory.

I don't know how we could express it, but as far as I understand, this breaks Rust memory safety guarantees.

youyuanwu: Returning the slice is not unsafe. std Vec does the same thing.
What is unsafe is constructing the BufferRef type in the first place, since the lifetime info of the buffer is not captured by BufferRef struct. We could consider making all the constructing function unsafe.

guhetier: Good point, I should have put that comment on the constructing functions below.

I think we should make the constructing functions unsafe for now - they break safe Rust code guarantees; we should make sure a user of the lib is aware of it.

We can potentially make a smarter wrapper later on that could capture the lifetime one way or another. I am currently thinking the receive completion function could take BufferRefs as a parameter to consume them. It would likely take some extra sate to ensure receives are completed in order though, and that it respects the rules for the different receive modes (multiple receives vs single).

PR #4839, opened 2025-02-22
Rust: Settings type

youyuanwu:
## Description
Add safe wrapper for QUIC_SETTINGS ffi type.
Add setters for each setting entry using macro.
Changed Configuration new() to use new Settings type and removed the manual ffi type.

MISC:
Preview settings is removed/regressed because the auto generated bindings do not have preview apis yet.
Rust code files are accumulated in src dir. We should add a dedicated dir inside src to keep all rust code together. Will do it in a separate PR.

## Testing
Added simple test to set settings using setters, and get the settings using ffi type to verify they match.

## Documentation
NA

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,6 +24,8 @@ pub mod ffi;
 pub use error::{Status, StatusCode};
 mod types;
 pub use types::{BufferRef, ConnectionEvent, ListenerEvent, NewConnectionInfo, StreamEvent};
+mod settings;
+pub use settings::Settings;
 
 //
 // The following starts the C interop layer of MsQuic API.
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1555,19 +1493,10 @@ mod tests {
...
-            Settings::new()
-                .set_peer_bidi_stream_count(100)
-                .set_peer_unidi_stream_count(3)
-                .set_stream_multi_receive_enabled(true),
-            #[cfg(not(feature = "preview-api"))]
-            Settings::new()
-                .set_peer_bidi_stream_count(100)
-                .set_peer_unidi_stream_count(3),
-        );
+        let settings = Settings::new()
```
nibanks, line 1496: What about the use of `#[cfg(not(feature = "preview-api"))]` that was here before?

youyuanwu: We do not have preview apis in generated bindings.
We need to support generating preview apis for bindings first to support the safe wrappers. Let me know if rust support for preview apis are required.

nibanks (dismissed):

nibanks (approved):

PR #4852, opened 2025-02-25
Optimize QuicStreamSetUpdateMaxStreams

### LINKED ISSUE #3538: Improve Perf of QuicStreamSetUpdateMaxStreams

When looking at perf traces, I noticed 2.5% of the connection was spent simply updating the max flow control for stream IDs; essentially allowing queued streams to be started. The reason it's kind of expensive is that we iterate through the entire stream set hash table (including both started and non-started streams).

### Proposed solution

We could have an ordered list of non-started streams that we can walk to improve the perf.

### Additional context

image

guhetier:
## Description

Store new streams that are blocked by Stream Id flow control in a sorted list (instead of the hash table containing all streams) to improve the performances of `QuicStreamSetUpdateMaxStreams`.

It will now be enough to iterate over a subset of this list instead of iterating over the entirety of opened streams when receiving a "MAX_STREAMS" frame increasing the limit.

Closes #3538

## Testing

Already covered by existing tests.

## Documentation

In code documentation added.

### DIFF
--- a/src/core/stream.h
+++ b/src/core/stream.h
@@ -158,6 +158,7 @@ typedef union QUIC_STREAM_FLAGS {
         BOOLEAN Freed                   : 1;    // Freed after last ref count released. Used for Debugging.
 
         BOOLEAN InStreamTable           : 1;    // The stream is currently in the connection's table.
+        BOOLEAN InWaitingList           : 1;    // The stream is currently in the waiting list for stream id FC.
         BOOLEAN DelayIdFcUpdate         : 1;    // Delay stream ID FC updates to StreamClose.
     };
 } QUIC_STREAM_FLAGS;
@@ -236,17 +237,16 @@ typedef struct QUIC_STREAM {
         CXPLAT_HASHTABLE_ENTRY TableEntry;
 
         //
+        // The entry in the connection's list of streams waiting on stream
+        // id flow control.
+        //
+        CXPLAT_LIST_ENTRY WaitingForIdFlowControlLink;
+
+        //
         // The entry in the connection's list of closed streams to clean up.
         //
         CXPLAT_LIST_ENTRY ClosedLink;
     };
-
-    //
-    // TODO guhetier: Check if can be moved in the union or if need to be in hashtable + this at the same time.
-    // The entry in the connection's list of streams waiting on stream
-    // id flow control.
-    //
-    CXPLAT_LIST_ENTRY WaitingForIdFlowControlLink;
 
     //
     // The list entry in the output module's send list.
--- a/src/core/stream.h
+++ b/src/core/stream.h
@@ -237,17 +237,16 @@ typedef struct QUIC_STREAM {
         CXPLAT_HASHTABLE_ENTRY TableEntry;
 
         //
+        // The entry in the connection's list of streams waiting on stream
+        // id flow control.
+        //
+        CXPLAT_LIST_ENTRY WaitingForIdFlowControlLink;
+
+        //
         // The entry in the connection's list of closed streams to clean up.
         //
         CXPLAT_LIST_ENTRY ClosedLink;
     };
-
-    //
-    // TODO guhetier: Check if can be moved in the union or if need to be in hashtable + this at the same time.
-    // The entry in the connection's list of streams waiting on stream
-    // id flow control.
-    //
-    CXPLAT_LIST_ENTRY WaitingForIdFlowControlLink;
 
     //
     // The list entry in the output module's send list.
--- a/src/core/stream.h
+++ b/src/core/stream.h
@@ -237,16 +237,17 @@ typedef struct QUIC_STREAM {
         CXPLAT_HASHTABLE_ENTRY TableEntry;
 
         //
-        // The entry in the connection's list of streams waiting on stream
-        // id flow control.
-        //
-        CXPLAT_LIST_ENTRY WaitingForIdFlowControlLink;
-
-        //
         // The entry in the connection's list of closed streams to clean up.
         //
         CXPLAT_LIST_ENTRY ClosedLink;
     };
+
+    //
+    // TODO guhetier: Check if can be moved in the union or if need to be in hashtable + this at the same time.
+    // The entry in the connection's list of streams waiting on stream
+    // id flow control.
+    //
+    CXPLAT_LIST_ENTRY WaitingForIdFlowControlLink;
 
     //
     // The list entry in the output module's send list.
--- a/src/core/stream.h
+++ b/src/core/stream.h
@@ -242,6 +242,13 @@ typedef struct QUIC_STREAM {
     };
 
     //
+    // TODO guhetier: Check if can be moved in the union or if need to be in hashtable + this at the same time.
+    // The entry in the connection's list of streams waiting on stream
+    // id flow control.
+    //
+    CXPLAT_LIST_ENTRY WaitingForIdFlowControlLink;
+
+    //
     // The list entry in the output module's send list.
     //
     CXPLAT_LIST_ENTRY SendLink;
--- a/src/core/stream.h
+++ b/src/core/stream.h
@@ -230,20 +230,23 @@ typedef struct QUIC_STREAM {
     //
     uint32_t OutstandingSentMetadata;
 
+    //
+    // Linkage in the stream set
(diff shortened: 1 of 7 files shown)


src/core/stream.h
```diff
@@ -158,6 +158,7 @@ typedef union QUIC_STREAM_FLAGS {
         BOOLEAN Freed                   : 1;    // Freed after last ref count released. Used for Debugging.
 
         BOOLEAN InStreamTable           : 1;    // The stream is currently in the connection's table.
+        BOOLEAN InWaitingList           : 1;    // The stream is currently in the waiting list for stream id FC.
```
nibanks, line 161: Please update the copy in `quictypes.h`.

src/core/stream_set.c
```diff
@@ -75,28 +100,29 @@ QuicStreamSetTraceRundown(
...
 
 _IRQL_requires_max_(DISPATCH_LEVEL)
 _Success_(return != FALSE)
 BOOLEAN
-QuicStreamSetInsertStream(
-    _Inout_ QUIC_STREAM_SET* StreamSet,
-    _In_ QUIC_STREAM* Stream
-    )
-{
+QuicStreamSetLazyInitStreamTable(_Inout_ QUIC_STREAM_SET* StreamSet) {
```
nibanks, line 125:
```suggestion
QuicStreamSetLazyInitStreamTable(
    _Inout_ QUIC_STREAM_SET* StreamSet
    )
{
```

In the product code we generally always use this format.

src/core/stream_set.c
```diff
@@ -9,6 +9,18 @@
     keeps track of locally and remotely initiated streams, and synchronizes max
     stream IDs with the peer.
 
+Design:
+
+    The stream set store streams in 3 containers: a hash-table `StreamTable`for
+    open streams (need frequent lookup by ID), a sorted list `WaitingStreams`
+    for streams waiting to be allowed by stream id flow control (they will be
```
nibanks, line 16:
```suggestion
    for streams waiting to be allowed by stream ID flow control (they will be
```

src/core/stream_set.c
```diff
@@ -9,6 +9,18 @@
...
+    inserted in order in `StreamTable` once allowed), and a list `ClosedStreams`
+    for closed streams waiting for deletion.
+    Each stream should be in one and only one container at a time.
```
nibanks, line 19:
```suggestion
    Each stream must be in one and only one container at a time.
```

src/core/stream_set.c
```diff
@@ -23,20 +35,32 @@ QuicStreamSetValidate(
...
-    CXPLAT_HASHTABLE_ENUMERATOR Enumerator;
-    CXPLAT_HASHTABLE_ENTRY* Entry;
-    CxPlatHashtableEnumerateBegin(StreamSet->StreamTable, &Enumerator);
-    while ((Entry = CxPlatHashtableEnumerateNext(StreamSet->StreamTable, &Enumerator)) != NULL) {
-        const QUIC_STREAM* Stream = CXPLAT_CONTAINING_RECORD(Entry, QUIC_STREAM, TableEntry);
+
+    for (CXPLAT_LIST_ENTRY* Link = StreamSet->WaitingStreams.Flink;
+         Link != &StreamSet->WaitingStreams;
+         Link = Link->Flink) {
+        QUIC_STREAM* Stream =
```
nibanks, line 57:
```suggestion
        const QUIC_STREAM* Stream =
```

src/core/stream_set.c
```diff
@@ -303,50 +368,59 @@ QuicStreamSetInitializeTransportParameters(
...
-                FlowBlockedFlagsToRemove |= QUIC_FLOW_BLOCKED_STREAM_FLOW_CONTROL;
-                Stream->SendWindow = (uint32_t)CXPLAT_MIN(Stream->MaxAllowedSendOffset, UINT32_MAX);
+            CXPLAT_DBG_ASSERT(StreamSet->StreamTable != NULL);
+            if (!QuicStreamSetInsertStream(StreamSet, Stream)) {
+                //
+                // The stream hash-table should have been initialized
+                // already when inserting stream in `WaitingStreams`
+                //
+                CXPLAT_FRE_ASSERTMSG(FALSE, "Steam table lazy intialization failed");
             }
```
nibanks, line 396:
```suggestion
            //
            // The stream hash-table should have been initialized
            // already when inserting stream in `WaitingStreams`
            //
            CXPLAT_FRE_ASSERTMSG(
                QuicStreamSetInsertStream(StreamSet, Stream),
                "Steam table lazy intialization failed");
```

src/core/stream_set.c
```diff
@@ -303,50 +368,59 @@ QuicStreamSetInitializeTransportParameters(
...
-            }
+        CXPLAT_DBG_ASSERT(Stream->OutFlowBlockedReasons & QUIC_FLOW_BLOCKED_STREAM_ID_FLOW_CONTROL);
 
-            uint64_t NewMaxAllowedSendOffset =
-                QuicStreamGetInitialMaxDataFromTP(
-                    Stream->ID,
-                    QuicConnIsServer(Connection),
-                    &Connection->PeerTransportParams);
+        uint8_t FlowBlockedFlagsToRemove = 0;
+        if (StreamIndex < Info->MaxTotalStreamCount) {
```
nibanks, line 384: `WaitingStreams` is an ordered list, right? Assuming that, if this condition fails, we should simply `break` out of the loop instead of continuing to walk the list.
```suggestion
        if (StreamIndex >= Info->MaxTotalStreamCount) {
            break;
        }
```

src/core/stream_set.c
```diff
@@ -401,26 +475,46 @@ QuicStreamSetUpdateMaxStreams(
...
+
+            if ((Stream->ID & STREAM_ID_MASK) != Mask) {
+                continue;
+            }
+
+            //
+            // Any stream in the waiting list was blocked by the previous peer id flow control.
+            //
+            CXPLAT_DBG_ASSERT(Index >= Info->MaxTotalStreamCount);
+            if (!QuicStreamRemoveOutFlowBlockedReason(
```
nibanks, line 502: Assert the call directly, as I recommended above.

src/core/stream_set.c
```diff
@@ -401,26 +475,46 @@ QuicStreamSetUpdateMaxStreams(
...
+            if (Index >= MaxStreams) {
+                break;
+            }
+
+            if ((Stream->ID & STREAM_ID_MASK) != Mask) {
+                continue;
+            }
+
+            //
+            // Any stream in the waiting list was blocked by the previous peer id flow control.
```
nibanks, line 499:
```suggestion
            // Any stream in the waiting list was blocked by the previous stream ID flow control.
```

nibanks (commented):
Looks pretty good. Just a few (mostly nit) comments.

src/core/stream_set.c
guhetier: I generally prefer avoiding code with side effect in asserts, I find it error prone. For DBG assert, the call is skipped in release build, and FRE asserts look the same, so it always raises a warning flag when reading this code, "wait, does this actually execute?".
I can do the change if side effect in assert is a common practice in the codebase but would rather not otherwise.

src/core/stream_set.c
guhetier: It is ordered but all types of streams are in the same list so we need to iterate at least up to the maximum "MaxTotalStreamCount" for unidirectional and bidirectional streams.

The values are already available in the function, I'll add that early exit check.

src/core/stream_set.c
nibanks: We do have a lot of code (see posix platform) that follows this pattern.

nibanks (dismissed):
LGTM! Thanks!

src/core/stream_set.c
guhetier: Hum, that change broke the test. Looking closer.

guhetier: Ugh, now I remember why there was no early exit on that loop.
For every stream that is can't be opened with the peer yet, we still need to indicate that the stream is blocked + all that is below in the loop.

For this loop, we do need to iterate through all waiting streams.

nibanks (approved):

PR #4859, opened 2025-02-26
Rust: Support preview apis in ffi generated code

youyuanwu:
## Description
Add preview apis to generated rust code. 

## Testing
NA

## Documentation
NA

### DIFF
--- a/src/ffi/wrapper.hpp
+++ b/src/ffi/wrapper.hpp
@@ -1,5 +1,6 @@
 // header wrapper to feed into bindgen
 
+#define QUIC_API_ENABLE_PREVIEW_FEATURES
 #include "msquic.h"
 
 // undef the macro type and define the enum type.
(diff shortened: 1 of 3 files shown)


src/ffi/wrapper.hpp
```diff
@@ -1,5 +1,6 @@
 // header wrapper to feed into bindgen
 
+#define QUIC_API_ENABLE_PREVIEW_FEATURES
```
nibanks, line 3: Is there a way to make this opt in at the rust layer? We don't want to expose this stuff by default, since it can change.

youyuanwu: Not in the rust FFI layer. Once the code is generated, all apis are there regardless of preview of not.
Or we can generate the code twice with 1 copy containing preview apis and 1 without, but I prefer not to do this since it is a burden to maintain and regenerate.
We should keep using the preview feature flag in the wrapper layer that macro out the preview apis.

PR #4871, opened 2025-02-27
Initialize processor info memory to prevent garbage values

nibanks:
## Description

The `PADDING` field wasn't being set and apparently some code depended on that being explicitly zero.

## Testing

Ci/CD & Locally

## Documentation

N/A

### DIFF
--- a/src/platform/platform_winuser.c
+++ b/src/platform/platform_winuser.c
@@ -153,6 +153,9 @@ CxPlatProcessorInfoInit(
         Status = QUIC_STATUS_OUT_OF_MEMORY;
         goto Error;
     }
+    CxPlatZeroMemory(
+        CxPlatProcessorInfo,
+        ActiveProcessorCount * sizeof(CXPLAT_PROCESSOR_INFO));
 
     CXPLAT_DBG_ASSERT(CxPlatProcessorGroupInfo == NULL);
     CxPlatProcessorGroupInfo =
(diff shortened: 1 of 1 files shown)


src/platform/platform_winuser.c
```diff
@@ -153,6 +153,9 @@ CxPlatProcessorInfoInit(
         Status = QUIC_STATUS_OUT_OF_MEMORY;
         goto Error;
     }
+    CxPlatZeroMemory(
```
mtfriesen, line 156: should this be in `CXPLAT_ALLOC_NONPAGED` ? wouldn't it be best if QUIC could consistently reason about whether a specific allocation is zero'd?

anrossi (dismissed):

src/platform/platform_winuser.c
nibanks: In the future, we can definitely look at putting it directly in there or perhaps creating a `CXPLAT_ALLOC_NONPAGED_ZERO` or something. Right now, I just want a spot fix.

anrossi: I believe HeapAlloc has a flag for zero-initializing the memory, so we wouldn't even need to do anything extra but pass the flag.

mtfriesen: If you change `CxPlatAlloc` for example, you can use `HEAP_ZERO_MEMORY` to do the work as efficiently as possible. `ExAllocatePool2` also has a flag.

nibanks: Yes, it can easily be done. I just want to take some time to figure out the best/right way.

mtfriesen: At the very least, this needs to apply to winkernel, too.

mtfriesen: Hmm, I see, this pattern isn't used in winkernel. NVM.

mtfriesen (dismissed):

guhetier (approved):

PR #4881, opened 2025-03-03
Rust: RegistrationConfig and CredentialConfig types

youyuanwu:
## Description
Provide safe Rust wrapper for RegistrationConfig and CredentialConfig types.
Removed the manual ffi code respectively.
Replaced the crate feature switch schannel with openssl. This makes the rust code by default compiles with schannel on winodws unless openssl feature is enabled. For linux, openssl is the default.

## Testing
Added tests for loading credentials for each credential types.

## Documentation
NA

### DIFF
--- a/src/rs/config.rs
+++ b/src/rs/config.rs
@@ -30,7 +30,7 @@ impl RegistrationConfig {
     }
 
     pub fn set_app_name(mut self, value: String) -> Self {
-        self.app_name = Some(CString::new(value.as_bytes()).unwrap());
+        self.app_name = Some(CString::new(value).unwrap());
         self
     }
 
@@ -78,7 +78,7 @@ impl From<ExecutionProfile> for crate::ffi::QUIC_EXECUTION_PROFILE {
 pub struct CredentialConfig {
     credential_flags: CredentialFlags,
     credential: Credential,
-    principle: Option<CString>, // TODO: support async handler.
+    principal: Option<CString>, // TODO: support async handler.
     allowed_cipher_suites: AllowedCipherSuiteFlags,
     ca_certificate_file: Option<CString>,
 }
@@ -99,8 +99,8 @@ impl CredentialConfig {
         self
     }
 
-    pub fn set_principle(mut self, value: String) -> Self {
-        self.principle = Some(CString::new(value.as_bytes()).unwrap());
+    pub fn set_principal(mut self, value: String) -> Self {
+        self.principal = Some(CString::new(value).unwrap());
         self
     }
 
@@ -112,7 +112,7 @@ impl CredentialConfig {
 
     pub fn set_ca_certificate_file(mut self, value: String) -> Self {
         self.credential_flags |= CredentialFlags::SET_CA_CERTIFICATE_FILE;
-        self.ca_certificate_file = Some(CString::new(value.as_bytes()).unwrap());
+        self.ca_certificate_file = Some(CString::new(value).unwrap());
         self
     }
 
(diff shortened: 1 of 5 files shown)


nibanks (dismissed):

src/rs/config.rs
```diff
@@ -0,0 +1,537 @@
...
+            }
+            Credential::CertificatePkcs12(cert) => {
+                ffi_cfg.Type =
+                    crate::ffi::QUIC_CREDENTIAL_TYPE_QUIC_CREDENTIAL_TYPE_CERTIFICATE_PKCS12;
+                ffi_cfg.__bindgen_anon_1.CertificatePkcs12 =
+                    cert.as_ffi_ref() as *const _ as *mut _;
+            }
+        }
+        ffi_cfg.Principal = self
+            .principle
```
guhetier, line 158: Should this be "principal" instead of "principle" or is it intentional?

src/rs/config.rs
```diff
@@ -0,0 +1,537 @@
...
+}
+
+impl Default for AllowedCipherSuiteFlags {
+    fn default() -> Self {
+        Self::NONE
+    }
+}
+
+// Disable macos because the ffi bindings is using linux
+// for macos and it as error code mismatch.
```
guhetier, line 424:
```suggestion
// for macos and it has error code mismatch.
```

src/rs/config.rs
```diff
@@ -0,0 +1,537 @@
...
+
+    pub fn set_ca_certificate_file(mut self, value: String) -> Self {
+        self.credential_flags |= CredentialFlags::SET_CA_CERTIFICATE_FILE;
+        self.ca_certificate_file = Some(CString::new(value.as_bytes()).unwrap());
+        self
+    }
+
+    /// # Safety
+    /// ffi type returned needs to have the lifetime of self.
+    pub unsafe fn as_ffi(&self) -> QUIC_CREDENTIAL_CONFIG {
```
guhetier, line 121: Just wondering, is there a reason this one is implemented as unsafe instead of following the pattern of a `raw` member like other types?

youyuanwu: Ideally as_ffi() conversion should be implemented for all wrapper types and wrapper types should not contain the ffi types as fields. The reasons is that ffi type field makes the wrapper type unmovable, due to ptr in ffi points to some memory in the wrapper struct, and if user moves the wrapper type to another scope or return it from function, u get ptr pointing to wrong location.
The places where ffi type are fields is due to the inner ffi type memory needs to be valid after QUIC_CREDENTIAL_CONFIG is constructed. If there are internal ffi ptr they point to heap memory so the above-mentioned problem does not occur.

Anyway, I made all ffi functions crate level, so that users cannot call them, and we have the leeway to change them in future.

guhetier: I see, I didn't realize the move issue.

For the types using `ffi_ref`, Isn't that replacing one issue (the Rust type must outlive the ffi type) with another issue (the Rust type must not be moved) then? I don't see anything preventing the user from moving the Rust type and invalidating the ffi object.

Maybe populating the `raw` member only when `as_ffi_ref` is called (that way, the struct can be moved until the ref is requested) would help?
I imagine there is a combination of wrappers and PhantomData that would express that fully.

In any case, it can be done later. Maybe a comment would help not to forget the types must not be moved though.

youyuanwu: If the ffi_ref or as_ffi_ref  returned cannot be moved, the function is marked as unsafe, and user needs to take care of. Some of the ffi_ref is movable, and thus they are safe.
Changing raw member when as_ffi_ref is called requires mutability change on the function signature.

We can standardize more in future PRs.

guhetier: Ah, got it, I missed one level of indirection. Thanks for taking the time to explain.

guhetier (approved):

nibanks (approved):

PR #4882, opened 2025-03-03
Rust: Safe Api for callbacks

youyuanwu:
## Description
Use Rust Fn and FnMut for user callback function or closure, so user no longer need to use unsafe extern C function and managing callback context ptr.
Added macro to define common HandleRef type, and set get context functions to reduce code duplication.

## Testing
Modified the existing test callbacks to use the new APIs.

## Documentation
NA

### DIFF
--- a/src/rs/lib.rs
+++ b/src/rs/lib.rs
@@ -623,11 +623,6 @@ pub struct Stream {
 unsafe impl Sync for Stream {}
 unsafe impl Send for Stream {}
 
-/// Same as Stream but does not own the handle.
-/// Only used in callback wrapping where handle
-/// should not be closed by default.
-pub struct StreamRef(Stream);
-
 impl From<QuicPerformanceCountersParam> for QuicPerformanceCounters {
     fn from(value: QuicPerformanceCountersParam) -> Self {
         Self {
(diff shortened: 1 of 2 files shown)


src/rs/lib.rs
```diff
@@ -1000,6 +1091,7 @@ impl Connection {
             unsafe {
                 Api::ffi_ref().ConnectionClose.unwrap()(self.handle);
             }
+            self.clear_ctx_if_present();
```
youyuanwu, line 1094: This call is causing segfault on linux.
The problem seems to be that GetContext before ConnectionClose call returns the correct context ptr value, but after ConnectionClose call it returns a garbage ptr value. Ctx cleanup code tries to convert the garbage ptr value back to Rust Box and then drop the memory and hit segfault. This behaviour does not manifest on windows.
I ran valgrind and did not find mem corruption before this segfault is hit.
Is GetContext after ConnectionClose well defined? @nibanks

nibanks: Treat `ConnectionClose` like `delete` for the connection. You cannot call anything on it afterwards. The underlying platforms _migh_ manage the memory a bit differently, or have lingering reference counts internally for a bit, but you cannot depend on that.

src/rs/lib.rs
```diff
@@ -950,30 +1007,64 @@ impl Default for Connection {
...
-    /// ffi type and the context ptr makes this function unsafe.
-    pub fn open(
-        &mut self,
-        registration: &Registration,
-        handler: ffi::QUIC_CONNECTION_CALLBACK_HANDLER,
-        context: *const c_void,
-    ) -> Result<(), Status> {
+    pub fn open<F>(&mut self, registration: &Registration, handler: F) -> Result<(), Status>
+    where
+        F: FnMut(ConnectionRef, ConnectionEvent) -> Result<(), Status> + 'static,
```
guhetier, line 1050: Isn't the lifetime of the connection enough here? Callbacks will likely have a context with a limited lifetime.

src/rs/lib.rs
```diff
@@ -1479,11 +1596,7 @@ mod tests {
         );
 
         let mut connection = Connection::new();
-        let res = connection.open(
-            &registration,
-            Some(test_conn_callback),
-            &connection as *const Connection as *const c_void,
-        );
+        let res = connection.open(&registration, test_conn_callback);
```
guhetier, line 1599: It would be interesting to keep a test that provide some context (since any non-trivial application would need to), to make sure it is correctly handled through Rust lifetime validation + unsafe code. A counter for the number of calls could be an easy one.

youyuanwu: I have a full server client test in another branch. Will make it to main soon.

src/rs/lib.rs
```diff
@@ -1423,7 +1561,7 @@ mod tests {
             } => {
                 println!("Stream shutdown complete: {connection_shutdown} {app_close_in_progress} {connection_shutdown_by_app} {connection_closed_remotely} {connection_error_code} {connection_close_status}");
                 // Attach to stream for auto close handle.
-                unsafe { Stream::from_raw(stream) };
+                unsafe { Stream::from_raw(stream.as_raw()) };
```
youyuanwu, line 1564: The only uncertainty I have for this PR is this line.
Ctx is dropped at this line, but the code of execution is inside the callback. This means the current function/closure has already been dropped after this line, and the function has not returned yet. Is this UB?
But naively thinking, after this line if code does not access any closure variables (already dropped), the code is ok. Because the stack is still present, and the remaining function call instruction is just propagating return value to upper stack frame.
But I still worry this is UB (undefined behaviour).

src/rs/lib.rs
youyuanwu: Added the full server client test in this PR.

src/rs/lib.rs
nibanks: So, how do we resolve this uncertainty? Can we test it?

src/rs/lib.rs
guhetier: Right, but isn't `'static` here requiring that the handler has a static lifetime? Why not use the lifetime of `self`?

src/rs/lib.rs
```diff
@@ -804,30 +877,64 @@ impl Default for Connection {
     }
 }
 
+// Remarks on Rust callback in general:
+// Callback function or closure is set to msquic handle context of type `*mut Box<TCallback>>`.
+// We cannot use type `*mut TCallback` because dyn trait cannot be converted to a pointer directly.
```
jrahman, line 882: Issue here being that `&dyn Trait` is implemented as a 'fat pointer' with a pointer to both the underlying impl along with the vtable for that impl's implementation of the trait. So a single 8 byte pointer won't hold the 16 byte 'fat pointer'

src/rs/lib.rs
youyuanwu: Ideally it should be lifetime of self. But Box<ConnectionCallback> cannot be annotated with lifetime. Maybe it is possible somehow or maybe more unsafe trick is needed. It can be a future improvement.

src/rs/lib.rs
youyuanwu: The test executes this path and it seems to work, but I think this is UB, the test runs but there is no guarantee from compiler this will work in the future, or under certain optimizations.
We can document that closing handle inside the callback is not safely supported in rust, and user wishing to do so needs to do proper testing for the UB, or use unsafe to extract the ctx and clean it up outside the callback.
In practice closing the handle in the callback might not be a prevalent use case.

nibanks (dismissed):

nibanks (approved):

PR #4883, opened 2025-03-03
Fix Windows Loader Flags

nibanks:
## Description

Fixes the Windows loader flags to load dependencies from system32 directory.

## Testing

Manually verified locally, and CI/CD for regression checks.

## Documentation

N/A - this was always the expected behavior


gaurav2699 (approved):

AustinWise:
Thanks for fixing this. I built the main branch and confirmed that the load flag is now correctly set.

nibanks:
> Thanks for fixing this. I built the main branch and confirmed that the load flag is now correctly set.

Thanks! It should also be in the two latest release branches.

PR #4894, opened 2025-03-06
Fix Some Synchronization Issues in Receive Fuzzer

nibanks:
## Description

Updates the receive fuzzer's packet queue to have a lock. Also removes usage of the std::list since it's not really needed.

Also cleaned up some of the code while I was at it.

## Testing

Ran locally. CI/CD

## Documentation

N/A

### DIFF
--- a/src/tools/recvfuzz/recvfuzz.cpp
+++ b/src/tools/recvfuzz/recvfuzz.cpp
@@ -726,12 +726,19 @@ void fuzz(CXPLAT_SOCKET* Binding, CXPLAT_ROUTE Route) {
     int64_t HandshakePacketCount = 0;
     int64_t TotalByteCount = 0;
     uint8_t mode;
-    uint64_t StartTimeMs = CxPlatTimeMs64();
+    uint64_t StartTimeMs = CxPlatTimeMs64(), LastPrintTimeMs = StartTimeMs, CurrentTimeMs;
     uint8_t recvBuffer[8192];
     uint32_t bufferoffset = 0;
     bool handshakeComplete = FALSE;
     TlsContext HandshakeClientContext;
-    while (CxPlatTimeDiff64(StartTimeMs, CxPlatTimeMs64()) < RunTimeMs) {
+    while (CxPlatTimeDiff64(StartTimeMs, (CurrentTimeMs = CxPlatTimeMs64())) < RunTimeMs) {
+        if (CxPlatTimeDiff64(LastPrintTimeMs, CurrentTimeMs) > S_TO_MS(60)) {
+            LastPrintTimeMs = CurrentTimeMs;
+            printf("Total Initial Packets sent: %lld\n", (long long)InitialPacketCount);
+            printf("Total Handshake Packets sent: %lld\n", (long long)HandshakePacketCount);
+            printf("Total Bytes sent: %lld\n\n", (long long)TotalByteCount);
+        }
+
         mode = (uint8_t)GetRandom(10);
         if (mode < 1) {
             PacketParams InitialPacketParams = {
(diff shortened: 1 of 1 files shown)


src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -31,9 +31,17 @@ uint64_t MagicCid = 0x989898989898989ull;
 const QUIC_HKDF_LABELS HkdfLabels = { "quic key", "quic iv", "quic hp", "quic ku" };
 uint64_t RunTimeMs = 60000;
 CxPlatEvent RecvPacketEvent(true);
-std::list<QUIC_RX_PACKET*> PacketQueue;
+QUIC_RX_PACKET* PacketQueue;
+QUIC_RX_PACKET** PacketQueueTail = &PacketQueue;
+CxPlatLock PacketQueueLock;
 uint64_t CurrSrcCid = 0;
 
+struct PacketScope {
```
guhetier, line 39: This should use a `std::unique_ptr` instead of re-implementing it.
If keeping it, copy and move constructors should be deleted or implemented.

```
void QuicRxDeletePacket(QUIC_RX_PACKET* p) { CXPLAT_FREE(p, QUIC_POOL_TOOL); }
using QuicRxPacketDeleter = std::integral_constant<decltype(&::QuicRxDeletePacket), &::QuicRxDeletePacket>;
using unique_packet = std::unique_ptr<QUIC_RX_PACKET, QuicRxPacketDeleter>
```

Even better, we could define a generic unique pointer for all CXPLAT allocations (pseudo-code, :

```
template<class T, uint32_t Pool>
void CxPlatFree(T* p) { CXPLAT_FREE(p, Pool); }

template<class T, uint32_t Pool>
using CxPlatDeleter= std::integral_constant<decltype(&::CxPlatFree<T, Pool>), &::CxPlatFree<T, Pool>>;
template<class T, uint32_t Pool>
using unique_cxplat_ptr = std::unique_ptr<T, CxPlatDeleter<T, Pool>>;
```

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -184,13 +192,18 @@ UdpRecvCallback(
                 memcpy((void *)PacketCopy->DestCid, Packet.DestCid, Packet.DestCidLen);
                 PacketCopy->SourceCid = PacketCopy->DestCid + Packet.DestCidLen;
                 memcpy((void *)PacketCopy->SourceCid, Packet.SourceCid, Packet.SourceCidLen);
-                PacketQueue.push_back(PacketCopy);
+                PacketCopy->_.Next = nullptr;
+
+                PacketQueueLock.Acquire();
+                *PacketQueueTail = PacketCopy;
+                PacketQueueTail = (QUIC_RX_PACKET**)&PacketCopy->_.Next;
```
guhetier, line 199: IMO, since this is C++ code, we should just embrace it and simplify our life:
- Put the packet directly in the smart wrapper for memory management so we don't have to care about that part anymore
- Put the smart pointer in a std::dequeue to have a simple push / pop semantic instead of manipulating a C-style linked list by hand
- ... and if we want to go the extra length, it then becomes very simple to wrap that in a struct that take the lock to make the push / pop operations thread safe

nibanks: At least for now, I prefer to use the C type which already has this Next pointer.

src/tools/recvfuzz/recvfuzz.cpp
nibanks: For now, I just want to get this test stable. We can follow up with a more thorough cleaning.

guhetier (approved):

PR #4895, opened 2025-03-06
RecvBuffer simplifications [1/N]

guhetier:
## Description

This is the first of a series of PR to simplify the receive buffer code by unifying the handling of the different buffer modes where possible.

This change:

-  introduces a "RetiredChunk". It is meant to store a chunk when Single/Circular mode duplicates its content but can't free it yet because it is externally referenced.
    - we could add the retired chunk to the chunk list later on if we want to save a pointer, but setting it aside will make it easier to reason about changes to the read, write and drain logics for now

- make sure the "Capacity" is properly updated for single / circular mode when allocating a new buffer

- fixes some comments, add assertions and fix variable name style

## Testing

N/A

## Documentation

No, the documentation doesn't go in that level of details

### DIFF
--- a/src/core/recv_buffer.c
+++ b/src/core/recv_buffer.c
@@ -322,6 +322,7 @@ QuicRecvBufferResize(
                     Span - LengthTillWrap);
             }
             RecvBuffer->ReadStart = 0;
+            CXPLAT_DBG_ASSERT(NewChunk->AllocLength == TargetBufferLength);
             RecvBuffer->Capacity = NewChunk->AllocLength;
 
         } else {
@@ -1186,7 +1187,7 @@ QuicRecvBufferDrain(
     CXPLAT_DBG_ASSERT(DrainLength <= RecvBuffer->ReadPendingLength);
 
     //
-    // Free retired chunks, now that they are no longer referenced.
+    // Free the retired chunk, now that it is no longer referenced.
     //
     if (RecvBuffer->RetiredChunk != NULL) {
         CXPLAT_DBG_ASSERT(
--- a/src/core/recv_buffer.c
+++ b/src/core/recv_buffer.c
@@ -269,7 +269,6 @@ QuicRecvBufferResize(
         TargetBufferLength != 0 &&
         (TargetBufferLength & (TargetBufferLength - 1)) == 0); // Power of 2
     CXPLAT_DBG_ASSERT(!CxPlatListIsEmpty(&RecvBuffer->Chunks)); // Should always have at least one chunk
-
     QUIC_RECV_CHUNK* LastChunk =
         CXPLAT_CONTAINING_RECORD(
             RecvBuffer->Chunks.Blink,
--- a/src/core/recv_buffer.c
+++ b/src/core/recv_buffer.c
@@ -8,10 +8,8 @@ Abstract:
     The receive buffer is a dynamically sized circular buffer for reassembling
     stream data and holding it until it's delivered to the client.
 
-    When the buffer is resized, all bytes in the buffer are copied to a new
-    backing memory of the requested size. The client must keep this in mind and
-    only resize the buffer infrequently, for instance by resizing exponentially,
-    or try to resize when few bytes are buffered.
+    It is implemented as a linked list of buffers to allow for different
+    behaviors (modes) when managing memory.
 
     There are two size variables, AllocBufferLength and VirtualBufferLength.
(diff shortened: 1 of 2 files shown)


src/core/recv_buffer.c
```diff
@@ -268,6 +269,7 @@ QuicRecvBufferResize(
         TargetBufferLength != 0 &&
         (TargetBufferLength & (TargetBufferLength - 1)) == 0); // Power of 2
     CXPLAT_DBG_ASSERT(!CxPlatListIsEmpty(&RecvBuffer->Chunks)); // Should always have at least one chunk
+
```
ProjectsByJackHe, line 272: nit: remove newline

src/core/recv_buffer.c
```diff
@@ -8,10 +8,8 @@
     The receive buffer is a dynamically sized circular buffer for reassembling
     stream data and holding it until it's delivered to the client.
 
-    When the buffer is resized, all bytes in the buffer are copied to a new
```
ProjectsByJackHe, line 11: Has this comment been out-of-sync for a while now?
This change seems to suggest that this copy-and-realloc behavior has been deprecated for some time.
If not, consider keeping some form of it.

src/core/recv_buffer.c
guhetier:
```suggestion
```

ProjectsByJackHe:
So the cargo run https://github.com/microsoft/msquic/actions/runs/14205352895/job/39806083323?pr=4895 failed twice in a row. I know sometimes Cargo can be a bit flaky, so I enqueued a 3rd retry. If it fails again, we'll need to investigate. 

Netperf is failing due to an unrelated reason. Overall LGTM, though I would wait for @anrossi or @nibanks to approve due to the complexity of the Recv Buffer code.

guhetier:
> So the cargo run https://github.com/microsoft/msquic/actions/runs/14205352895/job/39806083323?pr=4895 failed twice in a row. I know sometimes Cargo can be a bit flaky, so I enqueued a 3rd retry. If it fails again, we'll need to investigate.
> 
> Netperf is failing due to an unrelated reason. Overall LGTM, though I would wait for @anrossi or @nibanks to approve due to the complexity of the Recv Buffer code.

Cargo has been especially flaky recently, that is something we need to investigate. I don't think it is related to this PR though.

ProjectsByJackHe (dismissed):
CI passing; LGTM, I would recommend getting another approval, but I approved now to unblock.

src/core/recv_buffer.c
```diff
@@ -23,23 +21,21 @@
...
-    receive a large buffer (given a flight of packets) but not needed to
+    receive a large buffer (given a flight of packets) but not need to
     allocate memory for the entire buffer all at once.
 
     This does expose an attack surface though. In the common case we might be
     able to get by with a smaller buffer, but we need to be careful not to over
     commit. We must always be willing/able to allocate the buffer length
     advertised to the peer.
 
-    Currently, only growing the virtual buffer length is supported.
```
nibanks, line 41: AFAICT, this statement is still true. Why remove it?

src/core/recv_buffer.c
```diff
@@ -321,7 +322,7 @@ QuicRecvBufferResize(
                     Span - LengthTillWrap);
             }
             RecvBuffer->ReadStart = 0;
-            RecvBuffer->Capacity = TargetBufferLength;
+            RecvBuffer->Capacity = NewChunk->AllocLength;
```
nibanks, line 325: I wonder if it's worth a debug assert to say these are the same/interchangeable?

src/core/recv_buffer.c
```diff
@@ -1195,6 +1185,18 @@ QuicRecvBufferDrain(
 {
     CXPLAT_DBG_ASSERT(DrainLength <= RecvBuffer->ReadPendingLength);
 
+    //
+    // Free retired chunks, now that they are no longer referenced.
```
nibanks, line 1189: nit: since we only ever have one, remove the plural from this sentence?

src/core/recv_buffer.c
guhetier: With app-owned mode, it is not strictly true anymore. The virtual buffer length decreases as we consume buffer space since we can't reuse it (and this is fine as long as it doesn't reduce the max byte index acceptable).

nibanks (dismissed):

nibanks (approved):

PR #4897, opened 2025-03-07
issues_4856: Fix Android build errors [draft]

vgmahajanshetty:
## Description
Fix Android build errors.
## Testing

Existing automation should be fine.

## Documentation

No

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -120,10 +120,6 @@ jobs:
         # Android doesn't use system crypto
         - plat: android
           systemcrypto: '-UseSystemOpenSSLCrypto'
-        # TODO: android to support ubuntu-24.04
-        - plat: android
-          os: 'ubuntu-24.04'
-          systemcrypto: '-UseSystemOpenSSLCrypto'
         # No openssl system crypto on ubuntu-22.04
         - plat: linux
           os: 'ubuntu-22.04'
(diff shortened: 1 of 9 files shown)


.github/workflows/build.yml
```diff
@@ -120,9 +120,10 @@ jobs:
         # Android doesn't use system crypto
         - plat: android
           systemcrypto: '-UseSystemOpenSSLCrypto'
-        # TODO: Fix android on ubuntu-24.04
+        # TODO: android to support ubuntu-24.04
         - plat: android
           os: 'ubuntu-24.04'
+          systemcrypto: '-UseSystemOpenSSLCrypto'
```
nibanks, line 126: Instead, just remove this exception completely

src/inc/msquic.h
```diff
@@ -38,6 +38,12 @@ Supported Platforms:
 #error "Unsupported Platform"
 #endif
 
+#if defined(_MSC_VER)
+#define UNUSED_RESULT(x) (void)(x)
+#else
+#define UNUSED_RESULT(x) do { int unused __attribute__((unused)) = (x); } while (0)
+#endif
```
nibanks, line 45: I'd rather not have a wrapper for this. Usually, we explicitly use the `(void)func();` model. Does that not work universally?

vgmahajanshetty: Ok, let me check that.

PR #4900, opened 2025-03-08
Prevent the receive completion operation queued if unnecessary

masa-koz:
## Description
If the receive complete API is called on a user thread before `QUIC_STREAM_EVENT_RECEIVE` finished, MsQuic will call `QuicStreamReceiveComplete` immediately even though a completion operation is queued. And a completion operation will triger `QuicStreamReceiveComplete` even though `RecvCompletionLength` is zero.

So, this pull request introduces a new flag to track if the receive complete API was called inline and updates related logic in the `MsQuic` library. The most important changes include adding the new flag `RecvCompletionInlineCalled`, modifying the condition checks, and updating comments to reflect the new logic.

## Testing

NA

## Documentation

NA


src/core/stream_recv.c
```diff
@@ -953,10 +953,14 @@ QuicStreamRecvFlush(
...
+            // RecvCompletionInlineCalled is one.
             //
-            FlushRecv = (Stream->RecvCompletionLength != 0);
-
+            if (InterlockedCompareExchange16(
+                (short*)&Stream->RecvCompletionInlineCalled, 0, 1) == 1) {
+                FlushRecv = (Stream->RecvCompletionLength != 0);
+            } else {
+                FlushRecv = FALSE;
+            }
```
nibanks, line 963: I don't understand why the old logic wouldn't work. If the app called it inline, then `RecvCompletionLength` is non-zero. Can you please elaborate further to explain why this is necessary?

nibanks: Oh, I see now. I think the better fix is to make the queued receive complete operation that will execute later end up being a no-op, but let me think on this a bit...

nibanks: Or, can we prevent the operation from being queued at all...

nibanks: Perhaps we can do something with the `ReceiveCompleteOperation` pointer.... I need to think on this. But ideally, I think my preference is if the user thread calls receive complete before the callback to the app returns, then I don't want the operation to be queued - msquic will just handle in inline (on its thread) when it returns from the callback. But we have to be careful about the synchronization...

masa-koz: I agree with your preference. I will try to modify this pr according to your suggestion.

masa-koz:
@nibanks I modified this pr according to your suggestion.

src/core/api.c
```diff
@@ -1321,11 +1321,13 @@ MsQuicStreamReceiveComplete(
         Stream,
         BufferLength);
 
+    CxPlatRwLockAcquireShared(&Stream->ReceiveCompleteInlineLock);
```
nibanks, line 1324: I would like to try to do this without a lock. Let me think if there's a way we might be able to do that...

nibanks:
The more I think about this, the more I am wondering what the real problem is with the original behavior:

> If the receive complete API is called on a user thread before QUIC_STREAM_EVENT_RECEIVE finished, MsQuic will call QuicStreamReceiveComplete immediately even though a completion operation is queued. And a completion operation will triger QuicStreamReceiveComplete even though RecvCompletionLength is zero.

What is the downside of calling the queued completion with a zero length?

masa-koz:
Yes. I'm afraid that I forgot the detail, but recv path paused even though the data arrived.

nibanks:
@guhetier can you please add this PR to your to-do list to help drive to completion? We need to get this fixed before the next release.

guhetier:
Hi @masa-koz, are you still interested in working on this PR?
If yes, I'm happy to help you get to a lock-free solution, otherwise, I can take over to fix this issue.

After looking at the problem again, I think we can avoid the lock by storing both `ReceiveCallActive` and `RecvCompletionLength` in the same `uint64_t` and updating them in one atomic operation:

1) The top bit of `RecvCompletionLength` is used for `ReceiveCallActive`
2) In `QuicStreamRecvFlush`, `Stream->Flags.ReceiveCallActive = TRUE;` can become something like:

```
    InterlockedExchangeAdd64((int64_t*)&Stream->RecvCompletionLength, 0x8000000000000000);
```

3) ... and after the notification returns, we can both get the number of bytes to complete and clear the flag with:

```
    BufferLength = InterlockedExchangeAdd64((int64_t*)&Stream->RecvCompletionLength, -0x8000000000000000);
    BufferLength &= ~0x8000000000000000;
    InterlockedExchangeAdd64((int64_t*)&Stream->RecvCompletionLength, BufferLength);
```

4) In `MsQuicStreamReceiveComplete`, we start by always using `InterlockedExchangeAdd64` to add the number of completed bytes to `Stream->RecvCompletionLength`, and we check the return value top bit to know if a receive call is pending, or if we need to queue a work item.

I think this should work. There is still a possibility of a race in "Multiple Receive" mode, but in that mode, a receive completion work item with 0 bytes completed is not a problem, we don't stop indicating receives.

This is similar to what Nick suggested in a comment, but slightly simpler.

masa-koz:
I'm still interested. But I have no time now. Could you mind waiting for a few days?

src/core/stream.h
```diff
@@ -142,7 +142,6 @@ typedef union QUIC_STREAM_FLAGS {
         BOOLEAN ReceiveMultiple         : 1;    // The app supports multiple parallel receive indications.
         BOOLEAN ReceiveFlushQueued      : 1;    // The receive flush operation is queued.
         BOOLEAN ReceiveDataPending      : 1;    // Data (or FIN) is queued and ready for delivery.
-        BOOLEAN ReceiveCallActive       : 1;    // There is an active receive to the app.
```
nibanks, line 145: As the comment at the top of this struct states, please update the copy of this in `quictypes.h`.

src/core/api.c
```diff
@@ -1321,14 +1321,12 @@ MsQuicStreamReceiveComplete(
...
 
-    CxPlatRwLockAcquireShared(&Stream->ReceiveCompleteInlineLock);
-    BOOLEAN ReceiveCompleteInline = Stream->Flags.ReceiveCallActive;
-    InterlockedExchangeAdd64(
-        (int64_t*)&Stream->RecvCompletionLength, (int64_t)BufferLength);
-    CxPlatRwLockReleaseShared(&Stream->ReceiveCompleteInlineLock);
-
-    if (ReceiveCompleteInline) {
-        goto Exit; // No need to queue a completion operation when run inline
+    uint64_t RecvCompletionLength = InterlockedExchangeAdd64(
```
nibanks, line 1324: If the app passes a really big `BufferLength` in, it could mess up this logic. We should probably have some defense against that, maybe limiting `BufferLength <= (UINT64_MAX >> 2)`?

src/core/stream_recv.c
```diff
@@ -971,7 +958,11 @@ QuicStreamRecvFlush(
             Stream->Flags.ReceiveEnabled = TRUE;
 
         } else if (Status == QUIC_STATUS_PENDING) {
-            FlushRecv = ReceiveCompleteCalled;
+            //
+            // The app called the receive complete API
+            // if RecvCompletionLength is non-zero.
+            //
+            FlushRecv = (RecvCompletionLength != 0);
```
nibanks, line 965: Before, we were only invoking this if the `RecvCompletionLength` grew beyond what it was before we ever indicated a receive. This would protect you from the case where the app queued a receive complete right before we started indicating new receives. I think that logic was on the right track, but also flawed. 

If there was any `RecvCompletionLength` before we started indicating receives, that means the app also already queued an operation to process that completion (and it's currently sitting in the queue). If that's the case, then we shouldn't try processing any receive completes inline until that operation has completed.

So, I'd actually modify the logic above where you set the high bit, and grab the return value from that. If the return value is non-zerp (i.e. already something queued to complete) then `FlushRecv` should never be TRUE. Then, if it is zero and the result of removing the high bit is non-zero, then we can flush a receive inline.

Finally, ideally, we should create test cases for all these corner conditions. Some of them could be done by holding locks in the receive callback (in the test code) to ensure another thread can call the completion in parallel. The tricky case will be getting test code to queue receive completion right before the receive indicate happens. We might not be able to guarantee that timing...

src/core/stream_recv.c
```diff
@@ -873,14 +873,10 @@ QuicStreamRecvFlush(
...
-        CxPlatRwLockAcquireExclusive(&Stream->ReceiveCompleteInlineLock);
-        Stream->Flags.ReceiveCallActive = TRUE;
         //
-        // Save the current value of RecvCompletionLength to check if it has
-        // changed during the app callback.
+        // Set the top bit of RecvCompletionLength to indicate that there is an active receive.
         //
-        uint64_t RecvCompletionLength = Stream->RecvCompletionLength;
-        CxPlatRwLockReleaseExclusive(&Stream->ReceiveCompleteInlineLock);
+        InterlockedExchangeAdd64((int64_t*)&Stream->RecvCompletionLength, 0x8000000000000000);
```
nibanks, line 879:
```suggestion
        BOOLEAN RecvCompletionPending =
            InterlockedExchangeAdd64((int64_t*)&Stream->RecvCompletionLength, 0x8000000000000000) != 0;
```

nibanks: Then, down below only go into the `if` that calls `QuicStreamReceiveComplete` if this is `FALSE`.

src/core/stream_recv.c
```diff
@@ -873,14 +873,10 @@ QuicStreamRecvFlush(
     while (FlushRecv) {
```
nibanks, line 873: If we go with my other proposed changes, I think `FlushRecv` is now too ambiguous, and should be renamed to `IndiciateRecvEvents` to indicate that we can continue to indicate up receives to the app, but not necessarily that we can process receive completes yet.

nibanks:
There are also merge conflicts preventing the CI from being run.

src/core/api.c
guhetier: The logic was already vulnerable to an overflow, we only reach that condition earlier now (but still for a value that should be unreachable in a valid scenario).

`BufferLength <= (UINT64_MAX >> 2)` would not be enough to prevent it, because `Stream->RecvCompletionLength` might not be null.

If we care about being safe, I think we can use the second highest bit of `Stream->RecvCompletionLength` as a canary:
- we implement the check you suggest
- after an addition, we check the new value and if the bit is set, we abort

src/core/stream_recv.c
guhetier: I am not sure I follow.

Unless we are in "multiple receive" mode, `RecvCompletionLength` should always be zero when we start a new receive indication (since we won't indicate another receive before the completion). If `RecvCompletionLength` is not zero, the user call the completion api in an invalid way, and the best we could do is to fail or zero the variable (because otherwise, we will complete more than we indicate...)

If we are in "multiple receive" mode, completing a receive from the work queue even with 0 completed byte is not a problem (we never stop indications), so why not continue processing inline as much as possible?

nibanks: I think you're right. I was overthinking things. Though we should add a debug assert to ensure this case.

masa-koz: I've added the debug assert.

src/core/api.c
masa-koz: I've added the check of overflow. Is this your intended?

src/core/api.c
```diff
@@ -1325,20 +1325,32 @@ MsQuicStreamReceiveComplete(
...
         "[strm][%p] Receive complete call [%llu bytes]",
         Stream,
         BufferLength);
 
-    InterlockedExchangeAdd64(
-        (int64_t*)&Stream->RecvCompletionLength, (int64_t)BufferLength);
+    uint64_t RecvCompletionLength = InterlockedExchangeAdd64(
+        (int64_t*)&Stream->RecvCompletionLength,
+        (int64_t)BufferLength);
+    if ((BufferLength & QUIC_STREAM_RECV_COMPLETION_LENGTH_CANARY_BIT) != 0 &&
```
vgmahajanshetty, line 1340: Thought of understanding why is it required to test 59th bit only for overflow detection?

src/core/stream.h
```diff
@@ -469,6 +471,9 @@ typedef struct QUIC_STREAM {
     } BlockedTimings;
 } QUIC_STREAM;
 
+#define QUIC_STREAM_RECEIVE_CALL_ACTIVE_FLAG            0x8000000000000000
+#define QUIC_STREAM_RECV_COMPLETION_LENGTH_CANARY_BIT    0x800000000000000
```
vgmahajanshetty, line 475: Should this be 0x4000000000000000?

nibanks: I agree.

nibanks: Also, they should have the same prefix `QUIC_STREAM_RECV` or `QUIC_STREAM_RECEIVE`. I don't care which, but let's be consistent.

nibanks: Maybe add some comment too please. 😄

src/core/api.c
```diff
@@ -1325,20 +1325,32 @@ MsQuicStreamReceiveComplete(
...
+        (RecvCompletionLength & QUIC_STREAM_RECV_COMPLETION_LENGTH_CANARY_BIT) != 0) {
+        // overflow detected
```
nibanks, line 1342: For one, this doesn't follow the comment style of this problem. For another, I think this comment is too trivial. Either we should remove it or expand upon it.

src/core/api.c
nibanks: Generally, yes. Thanks!

src/core/api.c
```diff
@@ -1325,20 +1325,32 @@ MsQuicStreamReceiveComplete(
...
+        QuicTraceEvent(
+            ConnError,
+            "[conn][%p] ERROR, %s.",
+            Connection,
+            "RecvCompletionLength is overflow");
+        QuicConnSilentlyAbort(Connection);
```
nibanks, line 1348: I think we should just return an error, not abort.

nibanks: I mean don't call the abort function. Just return invalid_state or something.

src/core/stream.h
```diff
@@ -433,7 +432,10 @@ typedef struct QUIC_STREAM {
 
     //
     // The number of received bytes the app has completed but not yet processed
-    // by MsQuic.
+    // by MsQuic. The top bit of RecvCompletionLength is used to indicate that
```
guhetier, line 435: nit: If using the second top-most bit as a canary, we should update the comment with it too.

src/core/stream_recv.c
```diff
@@ -966,7 +974,12 @@ QuicStreamRecvFlush(
 
         QUIC_STATUS Status = QuicStreamIndicateEvent(Stream, &Event);
 
-        Stream->Flags.ReceiveCallActive = FALSE;
+        RecvCompletionLength = Stream->RecvCompletionLength;
```
guhetier, line 977: I am not sure I understand what this does.
It tries to solve the fact that `- Stream->RecvCompletionLength` doesn't work if we set the top bit, right?

And it builds a number such as the addition will do what we need?

- I think a comment is needed, explaining how the operation clear the flags + substracts the bits.
- `NewRecvCompletionLength` could use a better name. It isn't a new completion length, maybe `NumCompletedBytes` since that what is left when removing the top bit?

src/core/api.c
guhetier: @nibanks `MsQuicStreamReceiveComplete` returns `void`, so returning an error code is not an option.
IMO, if `RecvCompletionLength` overflows, it is a clear sign of an application error (no way we could ever complete UINT64_MAX / 2 bytes at once).

- Either we treat it as an irrecoverable error (we can't know how much the app intended to complete, so we can't deliver the next bytes reliably and in order) -> in that case, we either abort the connection or end the process.
- We could treat it as "complete everything pending", but I think it would only be confusing and let a bad state persist longer

I don't think that logging a warning and returning without action is a good solution, in that case, we should simply assume the overflow can't happen and not use a canary bit.

src/core/stream_recv.c
guhetier: Thanks.
Hum, if the issue was only a compiler warning, I am not sure how I feel about this solution.
Technically, `NumCompleteBytes - RecvCompletionLength` is equal to `-QUIC_STREAM_RECV_COMPLETION_LENGTH_RECEIVE_CALL_ACTIVE_FLAG`, this is doing the exact same as before. But it hides the fact we are only trying to remove the bit flag.

Is the warning you get `C4146: unary minus operator applied to unsigned type, result still unsigned`?
If that is the one, we can use the fact that `-QUIC_STREAM_RECV_COMPLETION_LENGTH_RECEIVE_CALL_ACTIVE_FLAG == QUIC_STREAM_RECV_COMPLETION_LENGTH_RECEIVE_CALL_ACTIVE_FLAG`, and just do

```
//
// Clear the active receive flag (by adding it again, as it will clear the byte)
//
RecvCompletionLength = InterlockedExchangeAdd64(
            (int64_t*)&Stream->RecvCompletionLength,
            QUIC_STREAM_RECV_COMPLETION_LENGTH_RECEIVE_CALL_ACTIVE_FLAG);
```

guhetier: Or actually, we should also use `InterlockedAnd64` and `InterlockedOr64` to set and remove the flag, that would be much simpler (not sure why I tunneled vision on the add).

src/core/stream_recv.c
```diff
@@ -966,13 +974,25 @@ QuicStreamRecvFlush(
...
+        uint64_t NumCompletedBytes = RecvCompletionLength & ~QUIC_STREAM_RECV_COMPLETION_LENGTH_RECEIVE_CALL_ACTIVE_FLAG;
+        RecvCompletionLength = InterlockedExchangeAdd64(
+            (int64_t*)&Stream->RecvCompletionLength,
+            NumCompletedBytes - RecvCompletionLength);
+        RecvCompletionLength &= ~QUIC_STREAM_RECV_COMPLETION_LENGTH_RECEIVE_CALL_ACTIVE_FLAG;
 
         if (Status == QUIC_STATUS_CONTINUE) {
             CXPLAT_DBG_ASSERT(!Stream->Flags.SentStopSending);
             InterlockedExchangeAdd64(
```
guhetier, line 992: It seems we could avoid this interlocked operation (and the other one below) since we always add here only to subtract right after.

We could basically have `BufferLength = RecvCompletionLength + Event.RECEIVE.TotalBufferLength` at line 1028, and when we substract from `Stream->RecvCompletionLength`, we can substract only `RecvCompletionLength`.

src/core/stream_recv.c
```diff
@@ -966,13 +974,19 @@ QuicStreamRecvFlush(
...
         QUIC_STATUS Status = QuicStreamIndicateEvent(Stream, &Event);
 
-        Stream->Flags.ReceiveCallActive = FALSE;
+        //
+        // Clear the active receive flag.
+        //
+        RecvCompletionLength = InterlockedAnd64(
+            (int64_t*)&Stream->RecvCompletionLength,
+            ~QUIC_STREAM_RECV_COMPLETION_LENGTH_RECEIVE_CALL_ACTIVE_FLAG);
+        RecvCompletionLength &= ~QUIC_STREAM_RECV_COMPLETION_LENGTH_RECEIVE_CALL_ACTIVE_FLAG;
```
nibanks, line 983:
```suggestion
        RecvCompletionLength =
            InterlockedAnd64(
                (int64_t*)&Stream->RecvCompletionLength,
                ~QUIC_STREAM_RECV_COMPLETION_LENGTH_RECEIVE_CALL_ACTIVE_FLAG) &
            ~QUIC_STREAM_RECV_COMPLETION_LENGTH_RECEIVE_CALL_ACTIVE_FLAG;
```

src/core/stream_recv.c
```diff
@@ -883,6 +883,15 @@ QuicStreamRecvFlush(
     while (FlushRecv) {
         CXPLAT_DBG_ASSERT(!Stream->Flags.SentStopSending);
 
+        //
+        // Set the top bit of RecvCompletionLength to indicate that there is an active receive.
+        //
+        uint64_t RecvCompletionLength = InterlockedOr64(
+            (int64_t*)&Stream->RecvCompletionLength,
+            QUIC_STREAM_RECV_COMPLETION_LENGTH_RECEIVE_CALL_ACTIVE_FLAG);
```
nibanks, line 891:
```suggestion
        uint64_t RecvCompletionLength =
            InterlockedOr64(
                (int64_t*)&Stream->RecvCompletionLength,
                QUIC_STREAM_RECV_COMPLETION_LENGTH_RECEIVE_CALL_ACTIVE_FLAG);
```

src/core/stream_recv.c
```diff
@@ -982,11 +996,10 @@ QuicStreamRecvFlush(
...
         } else if (Status == QUIC_STATUS_PENDING) {
             //
-            // The app called the receive complete API inline if
-            // RecvCompletionLength is non-zero.
+            // The app called the receive complete API
+            // (inline or concurrently) before the callback returned if RecvCompletionLength is non-zero.
             //
-            FlushRecv = (Stream->RecvCompletionLength != 0);
-
+            FlushRecv = (RecvCompletionLength != 0);
```
nibanks, line 1002:
```suggestion
            FlushRecv = (RecvCompletionLength != 0);

```

src/core/api.c
```diff
@@ -1325,20 +1325,30 @@ MsQuicStreamReceiveComplete(
...
         StreamAppReceiveCompleteCall,
         "[strm][%p] Receive complete call [%llu bytes]",
         Stream,
         BufferLength);
 
-    InterlockedExchangeAdd64(
-        (int64_t*)&Stream->RecvCompletionLength, (int64_t)BufferLength);
+    uint64_t RecvCompletionLength = InterlockedExchangeAdd64(
+        (int64_t*)&Stream->RecvCompletionLength,
+        (int64_t)BufferLength);
```
nibanks, line 1339:
```suggestion
    uint64_t RecvCompletionLength =
        InterlockedExchangeAdd64(
            (int64_t*)&Stream->RecvCompletionLength,
            (int64_t)BufferLength);
```

src/core/stream_recv.c
nibanks: Actually, we should use `InterlockedExchange64` since our goal is to clear the whole thing anyways, so we don't need to do another interlocked operation down below to subtract.
```suggestion
        RecvCompletionLength =
            InterlockedExchange64(
                (int64_t*)&Stream->RecvCompletionLength,
                0) &
            ~QUIC_STREAM_RECV_COMPLETION_LENGTH_RECEIVE_CALL_ACTIVE_FLAG;
```
You might have to define this in quic_platform_posix.h to call `__sync_lock_test_and_set`.

Then, you can completely remove the `BufferLength` variable (using `RecvCompletionLength` directly) and the interlocked function below.

src/core/stream_recv.c
```diff
@@ -966,13 +974,19 @@ QuicStreamRecvFlush(
...
+        //
+        // Clear the active receive flag.
+        //
+        RecvCompletionLength = InterlockedAnd64(
+            (int64_t*)&Stream->RecvCompletionLength,
+            ~QUIC_STREAM_RECV_COMPLETION_LENGTH_RECEIVE_CALL_ACTIVE_FLAG);
+        RecvCompletionLength &= ~QUIC_STREAM_RECV_COMPLETION_LENGTH_RECEIVE_CALL_ACTIVE_FLAG;
 
+        uint64_t BufferLength = RecvCompletionLength;
+
```
nibanks, line 986:
```suggestion
```

src/core/stream_recv.c
```diff
@@ -966,13 +974,19 @@ QuicStreamRecvFlush(
...
+        RecvCompletionLength &= ~QUIC_STREAM_RECV_COMPLETION_LENGTH_RECEIVE_CALL_ACTIVE_FLAG;
 
+        uint64_t BufferLength = RecvCompletionLength;
+        
         if (Status == QUIC_STATUS_CONTINUE) {
             CXPLAT_DBG_ASSERT(!Stream->Flags.SentStopSending);
-            InterlockedExchangeAdd64(
-                (int64_t*)&Stream->RecvCompletionLength,
-                (int64_t)Event.RECEIVE.TotalBufferLength);
+            BufferLength += Event.RECEIVE.TotalBufferLength;
```
nibanks, line 989:
```suggestion
            RecvCompletionLength += Event.RECEIVE.TotalBufferLength;
```

src/core/stream_recv.c
```diff
@@ -997,18 +1010,14 @@ QuicStreamRecvFlush(
                 "App failed recv callback",
                 Stream->Connection->Registration->AppName,
                 Status, 0);
-
-            InterlockedExchangeAdd64(
-                (int64_t*)&Stream->RecvCompletionLength,
-                (int64_t)Event.RECEIVE.TotalBufferLength);
+            BufferLength += Event.RECEIVE.TotalBufferLength;
```
nibanks, line 1013:
```suggestion
            RecvCompletionLength += Event.RECEIVE.TotalBufferLength;
```

src/core/stream_recv.c
```diff
@@ -997,18 +1010,14 @@ QuicStreamRecvFlush(
...
             FlushRecv = TRUE;
         }
 
         if (FlushRecv) {
-            uint64_t BufferLength = Stream->RecvCompletionLength;
             InterlockedExchangeAdd64(
```
nibanks, line 1018: Remove this interlocked call.

nibanks (commented):
Here's a slight redesign that makes it cleaner and more efficient (I think).

masa-koz:
@nibanks 
I've changed the code to follow your suggestion.

nibanks (approved):

nibanks:
Could you replace merge/rebase main to get a few recent CI fixes? Then I think we'll be good to go. Thanks so much for working with us on this PR!

src/core/api.c
```diff
@@ -1325,20 +1325,31 @@ MsQuicStreamReceiveComplete(
...
 
-    InterlockedExchangeAdd64(
-        (int64_t*)&Stream->RecvCompletionLength, (int64_t)BufferLength);
+    uint64_t RecvCompletionLength =
+        InterlockedExchangeAdd64(
+            (int64_t*)&Stream->RecvCompletionLength,
+            (int64_t)BufferLength);
+    if ((BufferLength & QUIC_STREAM_RECV_COMPLETION_LENGTH_CANARY_BIT) != 0 &&
+        (RecvCompletionLength & QUIC_STREAM_RECV_COMPLETION_LENGTH_CANARY_BIT) != 0) {
+        QuicTraceEvent(
```
guhetier, line 1343: I am not convinced this is a good way to handle the error...
If we simply `goto Exit` with a log, we silently break our contract with the app: the completed some bytes but we won't restart receive indications (or in multi-receive mode, they are now out of sync).

IMO, this should be a fatal error for the connection.

@nibanks, opinions?

(I thought I commented on this earlier but can't find it anymore...)

nibanks: See this code in StreamSend above:
```c
            //
            // We can't fail the send at this point, because we're already queued
            // the send above. So instead, we're just going to abort the whole
            // connection.
            //
            if (InterlockedCompareExchange16(
                    (short*)&Connection->BackUpOperUsed, 1, 0) != 0) {
                goto Exit; // It's already started the shutdown.
            }
            Oper = &Connection->BackUpOper;
            Oper->FreeAfterProcess = FALSE;
            Oper->Type = QUIC_OPER_TYPE_API_CALL;
            Oper->API_CALL.Context = &Connection->BackupApiContext;
            Oper->API_CALL.Context->Type = QUIC_API_TYPE_CONN_SHUTDOWN;
            Oper->API_CALL.Context->CONN_SHUTDOWN.Flags = QUIC_CONNECTION_SHUTDOWN_FLAG_SILENT;
            Oper->API_CALL.Context->CONN_SHUTDOWN.ErrorCode = (QUIC_VAR_INT)QUIC_STATUS_OUT_OF_MEMORY;
            Oper->API_CALL.Context->CONN_SHUTDOWN.RegistrationShutdown = FALSE;
            Oper->API_CALL.Context->CONN_SHUTDOWN.TransportShutdown = TRUE;
            QuicConnQueueHighestPriorityOper(Connection, Oper);
```

masa-koz: I've added the abort code using Oper.

nibanks (approved):

PR #4921, opened 2025-03-15
Document usage of stream priority set param

### LINKED ISSUE #4826: Documentation for SetParam QUIC_PARAM_STREAM_PRIORITY

Currently the documentation does not elaborate on what the priority parameter actually means, does lower priority number mean the stream is drained first or is it the other way around?

Over [here](https://github.com/microsoft/msquic/blob/8eb966d808aa486fb27bf31c66f6840fbf95f34c/src/inc/msquic.h#L955) 
```c
```
it is mentioned that 0 is low and 0xFFFF is high, which logically seems to imply that 0xFFFF gets drained first

Over [here](https://github.com/microsoft/msquic/blob/4f61d6fa2dcf467937b56805283365ea318a5840/src/core/send.c#L201)
The circular linked list (`Stream->SendLink`) seems to be ordered from highest priority to lowest priority
```c
void
QuicSendUpdateStreamPriority(
    _In_ QUIC_SEND* Send,
    _In_ QUIC_STREAM* Stream
    )
{
    CXPLAT_DBG_ASSERT(Stream->SendLink.Flink != NULL);
    CxPlatListEntryRemove(&Stream->SendLink);

    CXPLAT_LIST_ENTRY* Entry = Send->SendStreams.Blink;
    while (Entry != &Send->SendStreams) {
        //
        // Search back to front for the right place (based on priority) to
        // insert the stream.
        //
        if (Stream->SendPriority <=
            CXPLAT_CONTAINING_RECORD(Entry, QUIC_STREAM, SendLink)->SendPriority) {
            break;
        }
        Entry = Entry->Blink;
    }
    CxPlatListInsertHead(Entry, &Stream->SendLink); // Insert after current Entry
}
```

Over [here](https://github.com/microsoft/msquic/blob/4f61d6fa2dcf467937b56805283365ea318a5840/src/core/send.c#L991)
The first stream which can be sent seems to be sent, which would imply stream with a higher priority number is sent first.
```c
_Success_(return != NULL)
QUIC_STREAM*
QuicSendGetNextStream(
    _In_ QUIC_SEND* Send,
    _Out_ uint32_t* PacketCount
    )
{
    QUIC_CONNECTION* Connection = QuicSendGetConnection(Send);
    CXPLAT_DBG_ASSERT(!QuicConnIsClosed(Connection) || CxPlatListIsEmpty(&Send->SendStreams));

    CXPLAT_LIST_ENTRY* Entry = Send->SendStreams.Flink;
    while (Entry != &Send->SendStreams) {

        //
        // TODO: performance: We currently search through blocked
        // streams repeatedly as we loop.
        //

        QUIC_STREAM* Stream = CXPLAT_CONTAINING_RECORD(Entry, QUIC_STREAM, SendLink);

        //
        // Make sure, given the current state of the connection and the stream,
        // that we can use the stream to frame a packet.
        //
        if (QuicSendCanSendStreamNow(Stream)) {

            if (Connection->State.UseRoundRobinStreamScheduling) {
                //
                // Move the stream after any streams of the same priority. Start
                // with the "next" entry in the list and keep going until the
                // next entry's priority is less. Then move the stream before
                // that entry.
                //
                CXPLAT_LIST_ENTRY* LastEntry = Stream->SendLink.Flink;
                while (LastEntry != &Send->SendStreams) {
                    if (Stream->SendPriority >
                        CXPLAT_CONTAINING_RECORD(LastEntry, QUIC_STREAM, SendLink)->SendPriority) {
                        break;
                    }
                    LastEntry = LastEntry->Flink;
                }
                if (LastEntry->Blink != &Stream->SendLink) {
                    CxPlatListEntryRemove(&Stream->SendLink);
                    CxPlatListInsertTail(LastEntry, &Stream->SendLink);
                }

                *PacketCount = QUIC_STREAM_SEND_BATCH_COUNT;

            } else { // FIFO prioritization scheme
                *PacketCount = UINT32_MAX;
            }

            return Stream;
        }

        Entry = Entry->Flink;
    }

    return NULL;
}
```

### Proposed solution

Add to description of stream priority in [Settings.md](https://github.com/microsoft/msquic/blob/main/docs/Settings.md#stream-parameters).

### Additional context

_No response_

ProjectsByJackHe:
## Description

Not much docs on the usage and behavior of stream priority set parameters for streams.
This PR should fill in the gaps.
Closes #4826 

## Testing

N/A

## Documentation

N/A

### DIFF
--- a/docs/Settings.md
+++ b/docs/Settings.md
@@ -216,7 +216,7 @@ These parameters are access by calling [GetParam](./api/GetParam.md) or [SetPara
 | `QUIC_PARAM_STREAM_ID`<br> 0                      | QUIC_UINT62       | Get-only  | Must be called on a stream after [StreamStart](./api/StreamStart.md) is called.      |
 | `QUIC_PARAM_STREAM_0RTT_LENGTH`<br> 1             | uint64_t          | Get-only  | Length of 0-RTT data received from peer.                                              |
 | `QUIC_PARAM_STREAM_IDEAL_SEND_BUFFER_SIZE`<br> 2  | uint64_t - bytes  | Get-only  | Ideal buffer size to queue to the stream. Assumes only one stream sends steadily.     |
-| `QUIC_PARAM_STREAM_PRIORITY` <br> 3               | uint16_t          | Get/Set   | A value from 0x0 to 0xFFFF that indicates the Stream priority. Data on higher priority stream get sent first. All streams start with priority 0x7FFF by default.  |
+| `QUIC_PARAM_STREAM_PRIORITY` <br> 3               | uint16_t          | Get/Set   | A value from 0x0 to 0xFFFF that indicates the Stream priority. 0xFFFF is highest priority. Data on higher priority stream get sent first. All streams start with priority 0x7FFF by default.  |
 | `QUIC_PARAM_STREAM_STATISTICS` <br> 4             | QUIC_STREAM_STATISTICS | Get-only  | Stream-level statistics. |
 | `QUIC_PARAM_STREAM_RELIABLE_OFFSET` <br> 5        | uint64_t          | Get/Set   | Part of the new Reliable Reset preview feature. Sets/Gets the number of bytes a sender must send before closing SEND path.
 
--- a/docs/Settings.md
+++ b/docs/Settings.md
@@ -216,7 +216,7 @@ These parameters are access by calling [GetParam](./api/GetParam.md) or [SetPara
 | `QUIC_PARAM_STREAM_ID`<br> 0                      | QUIC_UINT62       | Get-only  | Must be called on a stream after [StreamStart](./api/StreamStart.md) is called.      |
 | `QUIC_PARAM_STREAM_0RTT_LENGTH`<br> 1             | uint64_t          | Get-only  | Length of 0-RTT data received from peer.                                              |
 | `QUIC_PARAM_STREAM_IDEAL_SEND_BUFFER_SIZE`<br> 2  | uint64_t - bytes  | Get-only  | Ideal buffer size to queue to the stream. Assumes only one stream sends steadily.     |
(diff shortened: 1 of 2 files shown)


guhetier (approved):

docs/Settings.md
```diff
@@ -216,7 +216,7 @@ These parameters are access by calling [GetParam](./api/GetParam.md) or [SetPara
 | `QUIC_PARAM_STREAM_ID`<br> 0                      | QUIC_UINT62       | Get-only  | Must be called on a stream after [StreamStart](./api/StreamStart.md) is called.      |
 | `QUIC_PARAM_STREAM_0RTT_LENGTH`<br> 1             | uint64_t          | Get-only  | Length of 0-RTT data received from peer.                                              |
 | `QUIC_PARAM_STREAM_IDEAL_SEND_BUFFER_SIZE`<br> 2  | uint64_t - bytes  | Get-only  | Ideal buffer size to queue to the stream. Assumes only one stream sends steadily.     |
-| `QUIC_PARAM_STREAM_PRIORITY` <br> 3               | uint16_t          | Get/Set   | Stream priority. |
+| `QUIC_PARAM_STREAM_PRIORITY` <br> 3               | uint16_t          | Get/Set   | A value from 0x0 to 0xFFFF that indicates the Stream priority. Data on higher priority stream get sent first. All streams start with priority 0x7FFF by default.  |
```
guhetier, line 219: Looking at the bug that this fixes, I wonder if we need to be extra explicit over "higher priority == higher value".

So maybe something like "Data on streams with a higher priority value"? Or just adding another sentence with "a higher value means a higher priority"?

Up to you if you feel it is needed, but the user opening the bug looked confused with whether a high number means a high priority, or a low number means a high priority.

ProjectsByJackHe: I can add a line like "0xFFFF" is highest priority.

nibanks (approved):

PR #4929, opened 2025-03-18
Refactor CxPlat Pool Abstraction

nibanks:
## Description

Updates the CxPlat Pool abstraction to maintain the owning pool pointer so that free always goes back to the allocation source.

## Testing

CI/CD

## Documentation

N/A


src/platform/datapath_winuser.c
```diff
@@ -2780,47 +2770,48 @@ CxPlatSocketContextUninitialize(
...
 RioRecvBufferAllocate(
     _In_ uint32_t Size,
     _In_ uint32_t Tag,
     _Inout_ CXPLAT_POOL* Pool
     )
 {
-    CXPLAT_DATAPATH_PARTITION* DatapathProc =
-        CXPLAT_CONTAINING_RECORD(Pool, CXPLAT_DATAPATH_PARTITION, RioRecvPool);
-    CXPLAT_DATAPATH* Datapath = DatapathProc->Datapath;
+    CXPLAT_POOL_OBJECT* Object = CxPlatLargeAlloc(Size, Tag);
```
guhetier, line 2780: Checking I understand correctly, we need this workaround with the CXPLAT_POOL_OBJECT because we have both a special memory allocator (`VirtualAlloc`) and some "on construction" logic to register the buffer that has been combined with the allocation?

It reminds me of C++ `new` keyword that basically call in succession `operator new` then the constructor. It could make sense to modify the pool to accept an "allocator" and a "constructor", to abstract some of that logic

guhetier (commented):
Nice to see how many paths can be simplified by no longer needing to keep a pointer to the pool around :)

src/inc/quic_platform_winuser.h
```diff
@@ -290,8 +291,19 @@ CxPlatFree(
 
 typedef struct CXPLAT_POOL CXPLAT_POOL;
 
+typedef struct CXPLAT_POOL_OBJECT {
```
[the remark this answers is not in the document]
mtfriesen: On Windows, this should be `__declspec(align(MEMORY_ALLOCATION_ALIGNMENT))`.

src/inc/quic_platform_posix.h
```diff
@@ -478,19 +478,24 @@ typedef struct CXPLAT_POOL {
 
 } CXPLAT_POOL;
 
+typedef struct CXPLAT_POOL_OBJECT {
```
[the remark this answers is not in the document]
mtfriesen: Unsure on POSIX.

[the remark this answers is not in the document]
mtfriesen: Seems like POSIX is similar to Windows, but not quite identical: the Linux malloc man page, at least, says the allocation will be align the largest primitive type that can fit within the allocated buffer size. So, on 64-bit systems, the largest primitive types are usually 128 bits, so 16-byte alignment. IDK what macros are available to get it right on POSIX in general.

src/platform/datapath_winuser.c
nibanks: Yeah, it's kind of like that. But I don't want to split it up any more than it is. These are already very special case scenarios that can afford handling more stuff.

mtfriesen (approved):

src/inc/quic_platform_posix.h
```diff
@@ -541,49 +540,52 @@ CxPlatPoolAlloc(
...
-        return CxPlatAlloc(Pool->Size, Pool->Tag);
-    }
-#endif
     CxPlatLockAcquire(&Pool->Lock);
-    void* Entry = CxPlatListPopEntry(&Pool->ListHead);
-    if (Entry != NULL) {
-        CXPLAT_FRE_ASSERT(Pool->ListDepth > 0);
+    CXPLAT_POOL_HEADER* Header =
+    #if DEBUG
+        CxPlatGetAllocFailDenominator() ? NULL :
```
anrossi, line 546: Is this fixing a bug in the previous behavior? It looks like previously it would allocate new on true, and allocate from the pool on false. Now it's return NULL on true, and allocating from the pool or new on false.

anrossi: ok yeah, it's coming back to me now. The point of this was to skip the pool's lookaside list and always call `CxPlatAlloc` when random memory failures are enabled so that the logic in `CxPlatAlloc` can decide to return NULL.
The problem here is there's now a timing change where it'll always take and then drop the lock before allocating directly, when it would skip the lock before.  Maybe that'll exercise new code paths.

src/inc/quic_platform_winuser.h
```diff
@@ -368,10 +375,7 @@ CxPlatPoolInitialize(
     _Inout_ CXPLAT_POOL* Pool
     )
 {
-#if DEBUG
-    CXPLAT_DBG_ASSERT(Size >= sizeof(CXPLAT_POOL_ENTRY));
-#endif
-    Pool->Size = Size;
+    Pool->Size = Size + sizeof(CXPLAT_POOL_HEADER); // Add space for the object header
```
guhetier, line 378: Do we need to check for overflows? Pool size is usually hardcoded so it is very unlikely, but this could be allocating less than requested and lead to write overflow (for a pool of 4GB buffers... ^^)

src/inc/quic_platform_posix.h
nibanks: This code isn't changing any behavior. It's still skipping the lookaside list when we have either random allocations or custom allocation index dropping enabled. It sets the value to `NULL` here, which forces the code to call alloc below. And I'd rather not change the code once more if it's only for a comment. 😄

src/inc/quic_platform_winuser.h
nibanks: Generally the sizes are small for pool allocations because you're expected to make a LOT of them. So, no, I don't think we need anything here.

anrossi (approved):

PR #4935, opened 2025-03-20
Move Pools to Global Per Proc

nibanks:
## Description

In preparation for creating official partitions, moving all pool allocations into the per-proc structs at the MsQuic layer.

## Testing

Automation

## Documentation

N/A


src/core/operation.c
```diff
@@ -54,11 +54,11 @@ QuicOperationQueueUninitialize(
 _IRQL_requires_max_(DISPATCH_LEVEL)
 QUIC_OPERATION*
 QuicOperationAlloc(
-    _In_ QUIC_WORKER* Worker,
     _In_ QUIC_OPERATION_TYPE Type
     )
 {
-    QUIC_OPERATION* Oper = (QUIC_OPERATION*)CxPlatPoolAlloc(&Worker->OperPool);
+    QUIC_OPERATION* Oper =
+        (QUIC_OPERATION*)CxPlatPoolAlloc(&QuicLibraryGetPerProc()->OperPool);
```
anrossi, line 61: Isn't this going to slow down every allocation because it'll be a function call (to get the current processor), an array indexing operation, and then a pointer dereference, just to hand the pool off to the `CxPlatPoolAlloc` call?

src/core/sent_packet_metadata.h
```diff
@@ -80,10 +80,6 @@ typedef struct QUIC_SENT_FRAME_METADATA {
 
 } QUIC_SENT_FRAME_METADATA;
 
-CXPLAT_STATIC_ASSERT(
```
anrossi, line 83: Why is this moved away from the frame metadata type, which is what the comment is referring to?

src/core/frame.h
```diff
@@ -164,6 +164,10 @@ typedef enum QUIC_FRAME_TYPE {
 
 } QUIC_FRAME_TYPE;
 
+CXPLAT_STATIC_ASSERT(
+    QUIC_FRAME_MAX_SUPPORTED <= (uint64_t)UINT16_MAX,
+    "Sent packet metadata 'Type' field above assumes frames types fit in 16-bits");
```
anrossi, line 169: replace "Sent packet metadata" with `QUIC_SENT_FRAME_METADATA`
remove "above", since the referred-to field is not defined above in this file.

src/core/sent_packet_metadata.h
nibanks: I had to move it because of header include order changes. It would be _really nasty_ to try to keep it here.

src/core/operation.c
nibanks: Not really, and besides, this is very temporary. The very next change (already mostly complete) is to move all these to use an explicit `Partition->OperPool` access,

src/core/frame.h
nibanks: I want to keep both, because they are pointing to two separate requirements that needs to be met.

mtfriesen (approved):

anrossi (approved):

PR #4936, opened 2025-03-20
Rust: Enable server client test on nix

youyuanwu:
## Description
Enable server client test on linux an macos.
Using test certs in the openssl submodule.
Removed code that sets the handle context to null after the handle close. It causes access violation in some cases, and handle after close should not be accessed anymore.

## Testing
NA
## Documentation
NA

### DIFF
--- a/src/rs/lib.rs
+++ b/src/rs/lib.rs
@@ -651,8 +651,16 @@ macro_rules! define_quic_handle_impl {
             }
 
             /// Closes the handle and consumes it.
-            pub fn close(self) {
+            pub fn close(mut self) {
+                // close_inner implementation for listener, connection and stream
+                // is suppose to cleanup the context associated with the handle.
+                // Handle context is dropped after handle ffi close  call completes to ensure that
+                // it outlives the callbacks.
+                // context raw pointer is not set to null after ffi close call because a closed
+                // handle should not be accessed anymore.
                 self.close_inner();
+                // prevent drop to call ffi function again.
+                self.handle = std::ptr::null_mut();
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/rs/lib.rs
```diff
@@ -651,8 +651,16 @@ macro_rules! define_quic_handle_impl {
...
+            pub fn close(mut self) {
+                // close_inner implementation for listener, connection and stream
+                // is suppose to cleanup the context associated with the handle.
+                // Handle context is dropped after handle ffi close  call completes to ensure that
+                // it outlives the callbacks.
+                // context raw pointer is not set to null after ffi close call because a closed
+                // handle should not be accessed anymore.
                 self.close_inner();
+                // prevent drop to call ffi function again.
+                self.handle = std::ptr::null_mut();
```
nibanks, line 663:
```suggestion
                // The close_inner implementation for listener, connection and stream
                // is suppose to cleanup the context associated with the handle.
                // Handle context is dropped after handle ffi close call completes to ensure that
                // it outlives the callbacks.
                // Context raw pointer is not set to null after ffi close call because a closed
                // handle should not be accessed anymore.
                self.close_inner();
                // Prevent drop to call ffi function again.
                self.handle = std::ptr::null_mut();
```

src/rs/server_client_test.rs
```diff
@@ -31,12 +33,24 @@ pub fn get_test_cert_hash() -> String {
...
-    s
+    Credential::CertificateHash(CertificateHash::from_str(&s).unwrap())
+}
+
+/// Use test certs from openssl submodule
+#[cfg(not(target_os = "windows"))]
+pub fn get_test_cred() -> Credential {
+    use crate::CertificateFile;
+    let root_dir = std::env::current_dir().unwrap();
+    let cert_dir = root_dir.join("submodules/openssl/test/certs");
```
nibanks, line 44: I don't care for depending on a specific submodule file.

1. I want to make sure we can support openssl3 in the near future.
2. I am thinking about moving away from submodules and dynamically resolving these dependencies (probably via cmake), and they might be in different directories.

youyuanwu: Changed to use openssl cli to generate test certs in temp dir.

nibanks (approved):

PR #4943, opened 2025-03-24
Refactor State Isolation into Partitions

nibanks:
## Description

This PR updates all the internal state isolation code of to be per-partition, which are per-processor by default, but may be configured to be less than that by the app.

## Testing

CI/CD

## Documentation

N/A - All this is internal


src/core/library.h
```diff
@@ -92,6 +55,11 @@ typedef struct QUIC_LIBRARY {
     //
     BOOLEAN LazyInitComplete : 1;
 
+    //
+    // Tracks whether the library's lazy initialization has completed.
```
nibanks, line 59: Fix comment

src/core/library.h
```diff
@@ -155,7 +123,7 @@ typedef struct QUIC_LIBRARY {
     // Number of processors currently being used.
     //
     _Field_range_(>, 0)
-    uint16_t ProcessorCount;
+    //uint16_t ProcessorCount;
```
nibanks, line 126: Delete

src/core/library.h
```diff
@@ -236,15 +204,15 @@ typedef struct QUIC_LIBRARY {
...
-    _Field_size_(ProcessorCount)
-    QUIC_LIBRARY_PP* PerProc;
+    _Field_size_(PartitionCount)
+    QUIC_PARTITION* Partitions;
 
     //
     // Controls access to the stateless retry keys when rotated.
     //
-    CXPLAT_DISPATCH_LOCK StatelessRetryKeysLock;
+    CXPLAT_DISPATCH_LOCK StatelessRetryKeysLock; // TODO - Move to Partition
```
nibanks, line 215: Go ahead and move this.

src/core/library.c
```diff
@@ -162,55 +158,117 @@ QuicLibraryInitializePartitions(
...
+        QuicPartitionUninitialize(&MsQuicLib.Partitions[j]);
+    }
+
+    CXPLAT_FREE(MsQuicLib.Partitions, QUIC_POOL_PERPROC);
+    MsQuicLib.Partitions = NULL;
+
+    return Status;
+}
+
+QUIC_STATUS // TODO - Move to a partition.c
```
nibanks, line 214: Create partition.c for these.

src/core/worker.c
```diff
@@ -204,6 +204,7 @@ QuicWorkerAssignConnection(
 {
     CXPLAT_DBG_ASSERT(Connection->Worker != Worker);
     Connection->Worker = Worker;
+    Connection->Partition = Worker->Partition;
```
guhetier, line 207: nit: SAL annotations on this function are wrong, it should be `_Inout_` for the connection (and while at it, `const` for the worker).
But not directly tied to the PR.

src/core/connection.h
```diff
@@ -340,6 +340,12 @@ typedef struct QUIC_CONNECTION {
     //
     QUIC_WORKER* Worker;
 
+    //
+    // The partition this connection is currently assigned to. It is changed at
+    // the same time as the worker.
+    //
+    QUIC_PARTITION* Partition;
```
guhetier, line 347: Could this comment clarify that the partition can be set without the worker being set? It gives the impression the two are always set at the same time, while we can have a partition and no worker IIUC.

src/core/operation.h
```diff
@@ -397,6 +401,7 @@ QuicOperationEnqueueFront(
 _IRQL_requires_max_(DISPATCH_LEVEL)
 QUIC_OPERATION*
 QuicOperationDequeue(
+    _In_ QUIC_PARTITION* Partition,
     _In_ QUIC_OPERATION_QUEUE* OperQ
```
guhetier, line 405: nit: All functions above take the queue as the first parameter, but the last two take the partition first.
It'd be nice to be consistent.

src/core/worker.c
nibanks: We can follow up in a separate PR to try to fix these.

src/core/operation.h
guhetier: I actually wonder if given the number of helpers that need the partition, adding the pointer as an OperQ member would make sense.

src/core/partition.h
```diff
@@ -0,0 +1,173 @@
...
+Abstract:
+
+    The partitioned storage for global library state. The partitioning allows
+    multiple threads to operate on the library simultaneously with minimal
+    contention.
+
+    The primary goal of partitioning is to allow multiple threads to allocate
+    and free pool memory simultaneously without contention. It also maintains
+    isolation for other state that may be commonly accessed by multiple threads,
+    such as performance counters and stateless resets and retries.ABC
```
guhetier, line 15: typo?
```suggestion
    such as performance counters and stateless resets and retries.
```

src/core/partition.h
```diff
@@ -0,0 +1,173 @@
...
+
+    A partition is always (soft) affinitized to a specific processor. By default,
+    partitions are one to one with processors. Though, an application may choose
+    to create partitions on a subset of processors. In this case, the partition
+    will be shared by all processors that are not explicitly affinitized to a
+    partition, though generally the library will try to only execute on those
+    specified processors.
```
guhetier, line 22: I am not sure I understand that sentence properly.
Does it mean that one partition can be associated to multiple processors if the application say so? Or that the application can create less partition than the number of processor (but each partition is still associated to one processor)?
If I have 4 CPU and create 2 partitions only as an application, what does it mean for partition sharing between CPU?

mtfriesen: So if an app creates two partitions targeting CPUs A and B, and then submits a request on CPU C, the QUIC library will pick one of the two previously created partitions (neither matching CPU C) and implicitly associate the request to that partition? Or is there a catch-all partition for all remaining processors?

src/core/partition.h
```diff
@@ -0,0 +1,173 @@
...
+    The partitioned storage for global library state. The partitioning allows
+    multiple threads to operate on the library simultaneously with minimal
+    contention.
+
+    The primary goal of partitioning is to allow multiple threads to allocate
+    and free pool memory simultaneously without contention. It also maintains
+    isolation for other state that may be commonly accessed by multiple threads,
+    such as performance counters and stateless resets and retries.ABC
+
+    A partition is always (soft) affinitized to a specific processor. By default,
```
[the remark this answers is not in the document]
nibanks: Right now, a partition is only assigned to one processor.

src/core/partition.h
nibanks: Yes, it will try to pick a close partition and assign the connection to that one. RSS may later move it again.

guhetier:
Wrt to:

> Documentation
> N/A - All this is internal

Once this refactor is completed, it will be a great timing to add some "contributor" documentation about the partitioning and execution context designs.
Not for this PR though, this is more of an early reminder.

nibanks:
> Wrt to:
> 
> > Documentation
> > N/A - All this is internal
> 
> Once this refactor is completed, it will be a great timing to add some "contributor" documentation about the partitioning and execution context designs. Not for this PR though, this is more of an early reminder.

Sounds good for a follow up.

guhetier (approved):

mtfriesen (dismissed):

src/core/partition.h
```diff
@@ -56,9 +61,7 @@ typedef struct QUIC_CACHEALIGN QUIC_PARTITION {
     // Used for generating stateless retries.
     //
     CXPLAT_DISPATCH_LOCK StatelessRetryKeysLock;
-    CXPLAT_KEY* StatelessRetryKeys[2]; // Encryption of retry tokens
-    int64_t StatelessRetryKeysExpiration[2]; // Timestamp when the current stateless retry key expires.
-    BOOLEAN CurrentStatelessRetryKey; // Index into StatelessRetryKeys.
+    QUIC_RETRY_KEY StatelessRetryKeys[2];
```
guhetier, line 64: IMO, there is some non-trivial logic here (keeping only the key for the last two time intervals, access to the right one based on the timestamp). Some documentation would help.

```suggestion
    //
    // The current and previous stateless retry keys.
    // A new key and index are generated every `QUIC_STATELESS_RETRY_KEY_LIFETIME_MS`.
    // They are indexed based on the timestamp and their position on the array depends on the index parity.
    //
    QUIC_RETRY_KEY StatelessRetryKeys[2];
```

guhetier (approved):

src/core/partition.h
nibanks: I added more description to the abstract above.

ProjectsByJackHe:
For someone with less experience with the guts of this processor/partition logic,
I am a little confused with the PR description.

Is it that this PR refactors all the internal state isolation logic to be of per-partition (is partition here a group of CPUs?)?
Or is it that we keep the per-processor logic by default, but add the capability to make it per-partition?

If it's the latter, I don't see where the header changes are for how apps can opt-in/configure this feature. 
Also, (if it's the latter case) the PR description says no documentation is needed cause this refactor is all internal, why is there mention of apps configuring anything? Wouldn't that touch the API?

nibanks:
> For someone with less experience with the guts of this processor/partition logic, I am a little confused with the PR description.
> 
> Is it that this PR refactors all the internal state isolation logic to be of per-partition (is partition here a group of CPUs?)? Or is it that we keep the per-processor logic by default, but add the capability for apps to configure it to be per-partition?
> 
> If it's the latter, I don't see where the header changes are for how apps can opt-in/configure this feature. Also, (if it's the latter case) the PR description says no documentation is needed cause this refactor is all internal, why is there mention of apps configuring anything? Wouldn't that touch the API?

It's the former. This refactors the internal state to be held in partition structs (more detail in the PR in partition.h). This is generally completely hidden from the app layer by default.

guhetier (approved):
LGTM.
A dedicated architecture doc file about partition in general will be a nice follow up, since they are pretty central to MsQuic design.

PR #4950, opened 2025-03-26
Rust: Support linking pre-installed msquic lib

youyuanwu:
## Description
Added new feature to support linking pre-installed msquic lib in system location, and in vcpkg location. 

## Testing
Manual local testing with pre-installed deb pkg, and vcpkg respectively. 

## Documentation
NA

### DIFF
--- a/scripts/build.rs
+++ b/scripts/build.rs
@@ -92,14 +92,13 @@ mod find {
     /// Find and use preinstalled msquic binaries.
     pub(crate) fn find() {
         let out_dir = std::path::PathBuf::from(std::env::var("OUT_DIR").unwrap());
-        // On linux try the known install location of libmsquic pkg.
         let libs = if cfg!(target_os = "linux") {
-            try_system_location()
+            // On linux try the known install location of libmsquic pkg.
+            try_system_location().expect("cannot find msquic in system")
         } else {
-            None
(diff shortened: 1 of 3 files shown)


youyuanwu:
> Could we get some Linux testing that uses the public Linux packages maybe?

Added ubuntu job that installs deb pkg and run all tests.
Windows using vcpkg can be added in future as well, but it might have a long run time.

I removed linux+vcpkg support for now, some paths need to be fixed.

scripts/build.rs
```diff
@@ -65,16 +83,109 @@ fn main() {
...
+        #[cfg(target_os = "windows")]
+        copy_libs(&libs, &out_dir);
+
+        #[cfg(not(target_os = "windows"))]
+        symlink_libs(&libs, &out_dir);
+        println!("cargo:rustc-link-search=native={}", out_dir.display());
+    }
+
+    #[cfg(target_os = "windows")]
+    const LIBS: [&str; 3] = ["bin/msquic.dll", "bin/msquic.pdb", "lib/msquic.lib"];
```
nibanks, line 111: Why do you need the pdb and the lib? Shouldn't you only require the dll?

nibanks (approved):

PR #4952, opened 2025-03-27
Add Cibir tests to Connection Pool tests

anrossi:
## Description

Add CIBIR tests back to the connection pool API tests.

## Testing

Adding new tests

## Documentation

N/A.

### DIFF
--- a/src/test/lib/HandshakeTest.cpp
+++ b/src/test/lib/HandshakeTest.cpp
@@ -4104,28 +4104,36 @@ struct ServerMultiAcceptContext {
     uint32_t StartedConnectionCount;
     uint32_t MaxConnections;
     CxPlatEvent StartedEvent;
-    UniquePtr<TestConnection>* Connections;
+    UniquePtr<MsQuicConnection>* Connections;
 
-    _Function_class_(NEW_CONNECTION_CALLBACK)
+    _Function_class_(MsQuicListenerCallback)
     static
-    bool
+    QUIC_STATUS
     ListenerMultiAcceptConnection(
-        _In_ TestListener* Listener,
-        _In_ HQUIC ConnectionHandle
+        _In_ struct MsQuicListener*,
+        _In_opt_ void* Context,
+        _Inout_ QUIC_LISTENER_EVENT* Event
         )
     {
-        auto* This = (ServerMultiAcceptContext*)Listener->Context;
-        uint32_t NewCount = (uint32_t)InterlockedIncrement((long*)&This->StartedConnectionCount);
-        if (NewCount > This->MaxConnections) {
-            TEST_FAILURE("Too many connections started!");
-            return false;
+        auto* This = (ServerMultiAcceptContext*)Context;
+        if (Event->Type == QUIC_LISTENER_EVENT_NEW_CONNECTION) {
+            uint32_t NewCount = (uint32_t)InterlockedIncrement((long*)&This->StartedConnectionCount);
+            if (NewCount > This->MaxConnections) {
+                TEST_FAILURE("Too many connections started!");
+                return QUIC_STATUS_CONNECTION_REFUSED;
+            }
+            This->Connections[NewCount - 1] =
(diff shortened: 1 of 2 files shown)


src/test/lib/HandshakeTest.cpp
```diff
@@ -4104,32 +4104,44 @@ struct ServerMultiAcceptContext {
...
+        if (Event->Type == QUIC_LISTENER_EVENT_NEW_CONNECTION) {
+            uint32_t NewCount = (uint32_t)InterlockedIncrement((long*)&This->StartedConnectionCount);
+            if (NewCount > This->MaxConnections) {
+                TEST_FAILURE("Too many connections started!");
+                return QUIC_STATUS_CONNECTION_REFUSED;
+            }
+            QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
+            auto Connection = new(std::nothrow) MsQuicConnection(Event->NEW_CONNECTION.Connection, CleanUpAutoDelete, MsQuicConnection::NoOpCallback, nullptr);
+            if (Connection) {
+                if (!This->ServerConfiguration ||
```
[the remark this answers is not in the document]
anrossi: I'll consider it going forward, but it'll be a hard habit to break.
However, I found a way to remove this new code altogether, so this'll go away 😄

ProjectsByJackHe:
I stepped through GDB to try and figure out why https://github.com/microsoft/msquic/actions/runs/14206028558/job/39804328737?pr=4952 is failing, and it seems to be because of a watchdog timeout firing. (Not in this PR) Could we cut down on the flakiness by increasing the timeout or something?

anrossi:
We can increase the timeout, but it's timing out after 60 seconds. Shouldn't a handshake, even with loss, complete well before then?

ProjectsByJackHe:
> We can increase the timeout, but it's timing out after 60 seconds. Shouldn't a handshake, even with loss, complete well before then?

I enqueued a re-run. But given the previous 2 enumerations took 40+ seconds, and we'll have to look at the exact test code, it may be the case that we are doing many many handshakes with loss and it's just taking too long. Unless you're more familiar with that test and it's timing out on a SINGLE handshake (with loss). If that's the case, we'll need to investigate that.

src/test/lib/HandshakeTest.cpp
```diff
@@ -4207,14 +4173,15 @@ QuicTestConnectionPoolCreate(
...
-        TEST_QUIC_SUCCEEDED(Listener.Start(Alpn, &ServerAddr.SockAddr));
+        MsQuicAutoAcceptListener Listener(Registration, ServerConfiguration, MsQuicConnection::NoOpCallback);
+        TEST_QUIC_SUCCEEDED(Listener.GetInitStatus());
+
+        if (TestCibirSupport) {
+            TEST_QUIC_SUCCEEDED(Listener.SetCibirId(CibirId, CibirIdLength));
+        }
+
+        TEST_QUIC_SUCCEEDED(Listener.Start(Alpn, ServerAddr));
+        TEST_QUIC_SUCCEEDED(Listener.GetLocalAddr(ServerAddr));
```
guhetier, line 4184: Just curious, why do we need to get the address? Is it not fully set when we start the listener?

guhetier (approved):

src/test/lib/HandshakeTest.cpp
anrossi: `ServerAddr` doesn't have a port number set before it is passed to `Listener.Start`, this allows the networking stack to select an available port number for the listener. Since we need to tell the clients which port to connect to, we have to ask the listener the full address+port that it is listening on to get the port number.

Not setting the port number is a test reliability pattern; compile-time chosen ports occasionally fail in automation when the port is in use on a test machine. This way, the listener will always get an available port.

guhetier: Makes sense. I saw a Listener, I assumed the port had to be specified :)

PR #4954, opened 2025-03-28
Document Supported Architectures

### LINKED ISSUE #4855: Drop Windows Arm32 Support
We've been building and releasing arm32 binaries for Windows for quite some time, but it seems no one is actually using them. Windows doesn't ship a 32-bit arm OS anymore. .NET removed support over 4 years ago ([link](https://github.com/dotnet/installer/issues/8624)), so the plan is to remove support in MsQuic ASAP.

Just to be clear, this is Windows only. Linux arm32 support will continue on for some time.

nibanks:
## Description

Fixes #4855.

## Testing

N/A

## Documentation

Updated

### DIFF
--- a/docs/Release.md
+++ b/docs/Release.md
@@ -43,9 +43,16 @@ This table describes all officially supported MsQuic releases.
 <br>\* Future **Release Dates** are subject to change.
 <br>\** **End of Support** dates do not include possible [extended support](https://docs.microsoft.com/en-us/windows-server/get-started-19/servicing-channels-19#long-term-servicing-channel-ltsc) extensions.
 
-## Architectures
+## Platforms
 
-Generally, MsQuic is supported for all 64-bit architectures on all platforms, including x64 and arm64. arm32 is also supported, but only on Linux.
+We officially support the following operating systems and architectures.
+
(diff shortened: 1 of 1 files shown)


docs/Release.md
```diff
@@ -43,6 +43,10 @@ This table describes all officially supported MsQuic releases.
 <br>\* Future **Release Dates** are subject to change.
 <br>\** **End of Support** dates do not include possible [extended support](https://docs.microsoft.com/en-us/windows-server/get-started-19/servicing-channels-19#long-term-servicing-channel-ltsc) extensions.
 
+## Architectures
+
+Generally, MsQuic is supported for all 64-bit architectures on all platforms, including x64 and arm64. arm32 is also supported, but only on Linux.
```
guhetier, line 48: I personally think we should have a specific list (and potentially a "while not officially supported, MsQuic should also work on any 64-bit architecture and arm32 on linux).
"Generally" is not that helpful to me for my exotic architecture / platform, or could be seen as a commitment to support something we actually don't intend to, even if we try our best to be compatible with everything.

mtfriesen (approved):

PR #4960, opened 2025-03-31
Add QUIC stream event documentation and other event related doc updates

sm-msft:
## Description

Add/update documentation around events in MsQuic

- Add documentation for QUIC Stream events
- Update/clarify existing documentation on use of events in MSQuic and events for other class types

## Testing

No new tests necessary

## Documentation

This is a documentation change

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -16,8 +16,8 @@ Term | Definition
 *endpoint* | One side of a connection; client or server.
 *peer* | The *other* side of a connection.
 *callback handler* | The function pointer the app registers with an MsQuic object.
-*app context* or<br> *context* | A (possibly null) pointer registered with an MsQuic object. It is passed to callback handlers.
-*event* | An upcall to a callback handler.
+*app context* or<br> *context* | A (possibly null) pointer registered with an MsQuic object. It is passed to App's callback handlers.
+*event* | Beginning or ending of a specific condition or action in MsQuic library that results in the App's callback hander invocation.
 
 # High Level Overview
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -16,8 +16,8 @@ Term | Definition
 *endpoint* | One side of a connection; client or server.
 *peer* | The *other* side of a connection.
 *callback handler* | The function pointer the app registers with an MsQuic object.
-*app context* or<br> *context* | A (possibly null) pointer registered with an MsQuic object. It is passed to App's callback handlers.
-*event* | Beginning or ending of a specific condition or action in MsQuic library that results in the App's callback hander invocation.
+*app context* or<br> *context* | A (possibly null) pointer registered with an MsQuic object. It is passed to the app's callback handlers.
+*event* | Beginning or ending of a specific condition or action in the MsQuic library that results in the app's callback hander invocation.
 
 # High Level Overview
 
--- a/docs/Execution.md
+++ b/docs/Execution.md
@@ -8,8 +8,7 @@ The sections below detail the designs MsQuic uses and the reasons behind these c
 
 MsQuic [Object Model](./API.md#object-model) describes the hierarchy of MsQuic objects.
 
-MsQuic API delivers all state changes and notifications for a specific MsQuic object directly to the corresponding callback handler registered by the application.
-These include connection state changes, new streams being created, stream data being received, and stream sends completing.
+MsQuic API delivers all state changes and notifications for a specific MsQuic object directly to the corresponding callback handler registered by the application. These include connection state changes, new streams being created, stream data being received, and stream sends completing, among others.
 
 Example definition of Listener object callback API:
 
@@ -34,18 +33,18 @@ QUIC_STATUS
     );
 ```
 
-The application must register a callback handler for every MsQuic object it creates. This handler must manage all the events MsQuic may indicate for that object. The handler must also return a status for each event indicating to MsQuic whether not not the event was successfully handled.
+The application must register a callback handler for every MsQuic object it creates. This handler must manage all the events MsQuic may indicate for that object. The handler must also return a status for each event indicating to MsQuic how the event was handled. This returned status is often success/failure, but sometimes indicates MSQuic that further processing is required.
 
 This approach differs significantly from sockets and most networking libraries, where the application must make a call (e.g., `send` or `recv`) to determine if something happened.
 This design choice was made for several reasons:
 
 - The MsQuic API **runs in-process**, eliminating the need for a kernel to user mode boundary switch to notify the application layer. This makes the callback-based design more practical compared to sockets.
 
-- MsQuic, due to the QUIC protocol, has numerous event types. Applications may have hundreds of objects with potential state changes. The callback model allows the application to avoid call management on each object and focus on event management for a given object.
+- The various events defined in MsQuic are derived from the underlying QUIC protocol. Applications may have hundreds of objects with potential state changes. The callback model allows the application to avoid synchronization/call management on each object and focus on event handling for the object.
 
-- Writing correct, scalable code on top of the socket interfaces has proven challenging. Offloading the threading and call managementment to MsQuic enables every application to scalable with minimal effort, making things "just work" out of the box.
+- Writing correct, scalable code in every application built on top of the socket interfaces is a repetetive, challenging task prone to errors. Offloading the threading and synchronization to MsQuic enables every application to be scalable with minimal effort, making things "just work" out of the box.
 
-- Simplified logic flow in MsQuic by eliminating a queue/cached state of yet to be delivered application notifications. This queue/cached state is maintained in the socket model to track yet-to-be-picked-up events/data and the networking stack must wait for call(s) from the application before indicating completion. This represents additional code, complexity and memory usage in socket model that MsQuic does without.
+- Simpler logic flow in MsQuic by eliminating a queue/cached state of yet to be delivered application notifications. This queue/cached state is maintained in the socket model to track yet-to-be-picked-up events/data and the networking stack must wait for call(s) from the application before indicating completion. This represents additional code, complexity and memory usage in socket model that MsQuic does without.
 
 ### Writing Event Handlers
 
@@ -60,7 +59,7 @@ Any substantial work required by the application **must** be performed on thread
 
 This does not imply that the application needs separate threads to perform all of its work.
 Many operations are designed to be most efficient when executed within the callback.
-For example, closing a handle to a connection or stream is ideally done during the "shutdown complete" callback.
+For example, closing a handle to a connection or stream is ideally done during the "shutdown complete" event notification callback.
 
 Some callbacks necessitate the application to call MsQuic API in return, which can be a source of deadlocks.
 MsQuic design ensures that MsQuic API (down) calls made from a callback thread always occur inline (thus avoiding deadlocks) and will take precedence over any calls in progress or queued from a separate thread.
--- a/docs/Execution.md
+++ b/docs/Execution.md
@@ -33,7 +33,7 @@ QUIC_STATUS
     );
 ```
 
-The application must register a callback handler for every MsQuic object it creates. This handler must manage all the events MsQuic may indicate for that object. The handler must also return a status for each event indicating to MsQuic how the event was handled. This returned status is often success/failure, but sometimes indicates MSQuic that further processing is required.
+The application must register a callback handler for every MsQuic object it creates. This handler must manage all the events MsQuic may indicate for that object. The handler must also return a status for each event indicating to MsQuic how the event was handled. This returned status is often success/failure, but sometimes indicates MsQuic that further processing is required.
 
 This approach differs significantly from sockets and most networking libraries, where the application must make a call (e.g., `send` or `recv`) to determine if something happened.
 This design choice was made for several reasons:
@@ -61,8 +61,8 @@ This does not imply that the application needs separate threads to perform all o
 Many operations are designed to be most efficient when executed within the callback.
 For example, closing a handle to a connection or stream is ideally done during the "shutdown complete" event notification callback.
 
-Some callbacks necessitate the application to call MsQuic API in return, which can be a source of deadlocks.
-MsQuic design ensures that MsQuic API (down) calls made from a callback thread always occur inline (thus avoiding deadlocks) and will take precedence over any calls in progress or queued from a separate thread.
+Some callbacks necessitate the application to call MsQuic API in return. Such cyclic call patterns could lead to deadlocks in a generic implementation, but not so in MsQuic.
+Special attention has been paid to ensure that MsQuic API (down) calls made from a callback thread always occur inline (thus avoiding deadlocks) and will take precedence over any calls in progress or queued from a separate thread.
 By default, MsQuic will **never** invoke a recursive callback to the application in these cases. The only exception to this rule is if the application opts in via the `QUIC_STREAM_SHUTDOWN_FLAG_INLINE` flag when calling `StreamShudown` on a callback.
 
 ## Threading
@@ -111,3 +111,12 @@ graph TD
         end
     end
 ```
+
+# See Also
+
+[QUIC_STREAM_CALLBACK](api/QUIC_STREAM_CALLBACK.md)<br>
+[QUIC_STREAM_EVENT](api/QUIC_STREAM_EVENT.md)<br>
+[QUIC_CONNECTION_CALLBACK](api/QUIC_CONNECTION_CALLBACK.md)<br>
+[QUIC_CONNECTION_EVENT](api/QUIC_CONNECTION_EVENT.md)<br>
+[QUIC_LISTENER_CALLBACK](api/QUIC_LISTENER_CALLBACK.md)<br>
+[QUIC_LISTENER_EVENT](api/QUIC_LISTENER_EVENT.md)<br>
--- a/docs/Execution.md
+++ b/docs/Execution.md
@@ -8,7 +8,7 @@ The sections below detail the designs MsQuic uses and the reasons behind these c
 
 MsQuic [Object Model](./API.md#object-model) describes the hierarchy of MsQuic objects.
 
-MsQuic API delivers all state changes and notifications for a specific MsQuic object directly to the corresponding callback handler registered by the application. These include connection state changes, new streams being created, stream data being received, and stream sends completing, among others.
+The MsQuic API delivers all state changes and notifications for a specific MsQuic object directly to the corresponding callback handler registered by the application. These include connection state changes, new streams being created, stream data being received, and stream sends completing, among others.
 
 Example definition of Listener object callback API:
 
--- a/docs/Execution.md
+++ b/docs/Execution.md
@@ -6,8 +6,12 @@ The sections below detail the designs MsQuic uses and the reasons behind these c
 
 ## Event Model
 
-In the MsQuic API, all asynchronous state changes and notifications are indicated directly to the application via a callback.
-This includes connection state changes, new streams being created, stream data being received, and stream sends completing.
+MsQuic [Object Model](./API.md#object-model) describes the hierarchy of MsQuic objects.
+
+MsQuic API delivers all state changes and notifications for a specific MsQuic object directly to the corresponding callback handler registered by the application.
+These include connection state changes, new streams being created, stream data being received, and stream sends completing.
+
+Example definition of Listener object callback API:
 
 ```c
 typedef struct QUIC_LISTENER_EVENT {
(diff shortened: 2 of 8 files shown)


docs/api/QUIC_STREAM_EVENT.md
```diff
@@ -0,0 +1,266 @@
...
+
+`AbsoluteOffset`
+
+Absolute offset of the current data payload from the start of the receive operation.
+
+`TotalBufferLength`
+
+MsQuic indicates the total buffer length of the data in this parameter.
+
+If the application can processes only part of the received buffer synchronously in this call, it can be indicated to MsQuic by setting this parameter to the byte count processed and returning `QUIC_STATUS_CONTINUE` from this call.
```
guhetier, line 120: I think that giving this as a short version can be confusing, because the exact behavior depends on many details (buffering mode, return value...).

Maybe something like:
 
```suggestion
The application must set this parameter to the byte count it processed. If left untouched, it means the application could process all the indicated data. Otherwise, see [Receiving Data On Streams](../Streams.md#Receiving) for details about partially accepting data or asynchronous processing.
```

docs/api/QUIC_STREAM_EVENT.md
```diff
@@ -0,0 +1,266 @@
...
+
+Value | Meaning
+--- | ---
+**QUIC_RECEIVE_FLAG_NONE**<br>0 | No special behavior.
+**QUIC_RECEIVE_FLAG_0_RTT**<br>1 | The data was received in 0-RTT.
+**QUIC_RECEIVE_FLAG_FIN**<br>2 | FIN was included with this data. Used only for streamed data.
+
+## QUIC_STREAM_EVENT_SEND_COMPLETE
+
+Indicates that MsQuic has completed a [StreamSend](StreamSend.md) operation initated by the application.
```
guhetier, line 145: Can we add here that it means the app regain ownership of the `QUIC_BUFFER` and of the referenced memory used for the matching "send" call, and can now free it / re-use it?

We recently got some questions about it.

docs/api/QUIC_STREAM_EVENT.md
```diff
@@ -0,0 +1,266 @@
...
+
+`Canceled`
+Indicates that the StreamSend operation was canceled.
+
+`ClientContext`
+Client context to match this event with the original `StreamSend` operation.
+
+## QUIC_STREAM_EVENT_PEER_SEND_SHUTDOWN
+
+Indicates that the current send operation to the peer on this stream has been shutdown/completed.
```
guhetier, line 159: This seems wrong to me?
In my understanding, an `QUIC_STREAM_EVENT_PEER_SEND_SHUTDOWN` event occurs when the peer gracefully shutdowns its send path. The app gets this event when we got a FIN from the peer AND all the data got delivered to the application.
It directly follows getting a `QUIC_STREAM_EVENT_RECEIVE` event with the `FIN` flag set.

docs/api/QUIC_STREAM_EVENT.md
```diff
@@ -0,0 +1,266 @@
...
+
+## QUIC_STREAM_EVENT_PEER_SEND_ABORTED
+
+Indicates that the peer has aborted `StreamSend` operation.
```
guhetier, line 163: I think we could make this clearer by saying this is when the peer shut down its send path abortively.

```suggestion
Indicates that the peer has abortively closed its send path. The application won't receive any `QUIC_STREAM_EVENT_RECEIVE` event for this stream after this event.
```

docs/api/QUIC_STREAM_EVENT.md
```diff
@@ -0,0 +1,266 @@
...
+
+Ideal send buffer size in bytes for each `StreamSend` operation to avoid idling the connection.
+
+## QUIC_STREAM_EVENT_PEER_ACCEPTED
+
+This event is raised when a peer has provided sufficient flow control to accept a new stream. See [StreamStart](StreamStart.md) for additional information.
+
+## QUIC_STREAM_EVENT_CANCEL_ON_LOSS
+
+This event is raised when a stream is shutdown due to packet loss
```
guhetier, line 247: nit: Would be nice to link to the `QUIC_SEND_FLAG_CANCEL_ON_LOSS` option, since that behavior is not the default.

docs/api/QUIC_STREAM_CALLBACK.md
```diff
@@ -0,0 +1,43 @@
...
+
+The application callback context (optionally) supplied in [StreamOpen](StreamOpen.md), [SetCallbackHandler](SetCallbackHandler.md) or [SetContext](SetContext.md).
+
+`Event`
+
+A pointer to the [QUIC_STREAM_EVENT](QUIC_STREAM_EVENT.md) payload.
+
+# Remarks
+
+This is the signature of the function that handles callbacks from MsQuic for stream events. Apps are expected to keep any execution time in the callback **to a minimum**.
```
guhetier, line 36: > This is the signature of the function that handles callbacks from MsQuic for stream events.

That first sentence should be the main description of the API in my opinion, not just a remark.

docs/Execution.md
```diff
@@ -29,52 +33,50 @@ QUIC_STATUS
     );
 ```
 
-Above is an example of a callback delivered to the listener interface.
-The application must register a per-object callback handler to manage all events MsQuic may indicate for that object, returning a status to show if it was successfully handled or not.
+The application must register a callback handler for every MsQuic object it creates. This handler must manage all the events MsQuic may indicate for that object. The handler must also return a status for each event indicating to MsQuic how the event was handled. This returned status is often success/failure, but sometimes indicates MSQuic that further processing is required.
```
ProjectsByJackHe, line 36: nit: fix MSQuic --> MsQuic for consistency.

docs/Execution.md
```diff
@@ -29,52 +33,50 @@ QUIC_STATUS
...
+Event handlers are **required** for all objects that can receive events, as much of the MsQuic API operates through these callbacks.
 Critical events, such as "shutdown complete" notifications, provide vital information necessary for the application to function correctly.
 Without these events, the application cannot determine when it is safe to clean up objects.
 
-Applications should keep the execution time within callbacks **to a minimum**.
+Applications **must** keep the execution time within callbacks **to a minimum**.
 MsQuic does not use separate threads for protocol execution and upcalls to the application.
 Therefore, any significant delays in the callback **will delay the protocol**.
-Any substantial work required by the application should be performed on its own thread.
+Any substantial work required by the application **must** be performed on threads created by the application.
```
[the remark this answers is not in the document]
sm-msft: If the callback processing takes too long, there could be timeouts and such in the underlying protocol from what I remember.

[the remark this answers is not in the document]
sm-msft: If there are exceptions where long callbacks are permissible, we can add them in later.

docs/Execution.md
```diff
@@ -29,52 +33,50 @@ QUIC_STATUS
...
-In fact, many operations are designed to be most efficient when executed within the callback.
-For example, closing a handle to a connection or stream is ideally done during the "shutdown complete" indication.
+This does not imply that the application needs separate threads to perform all of its work.
+Many operations are designed to be most efficient when executed within the callback.
+For example, closing a handle to a connection or stream is ideally done during the "shutdown complete" event notification callback.
 
-A crucial aspect of this design is that all blocking API (down) calls invoked within a callback always occur inline (to prevent deadlocks) and will take precedence over any calls in progress or queued from a separate thread.
-It's also worth noting that MsQuic will not invoke a recursive callback to the application by default in these cases.
-The one exception to this rule is if the application opts in via the `QUIC_STREAM_SHUTDOWN_FLAG_INLINE` flag when calling `StreamShudown` on a callback.
+Some callbacks necessitate the application to call MsQuic API in return, which can be a source of deadlocks.
```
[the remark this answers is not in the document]
sm-msft: Will try to clarify this in the next iteration. 

I think deadlocks are a general concern with callbacks and we need to state it that it is not a problem with MsQuic.

docs/api/QUIC_CONNECTION_CALLBACK.md
```diff
@@ -21,7 +21,7 @@ QUIC_STATUS
 
 `Connection`
 
-The valid handle to the connection object this event is for.
+The valid handle to the connection object this event handler is for.
```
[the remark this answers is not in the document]
sm-msft: Thats correct.

I guess I confused the event handler registration w/context with event handler handle.

docs/api/QUIC_STREAM_EVENT.md
sm-msft: "must set this parameter" conflicts with "if left untouched", btw.

Our text here should be from the perspective of the application implementation. Will try to reword the text for clarity.
I think the details of receiving data are complex enough that reference to the streams.md is the right way to go.

docs/api/QUIC_STREAM_EVENT.md
sm-msft: Will look further into this...

docs/api/QUIC_STREAM_EVENT.md
sm-msft: I think referring to other events here can get complicated. We should have a sequence diagram for that.

guhetier: +1 for some diagrams.

I do think, though, that a simple version of what it means for the app to get that notification would be helpful (for instance, can I free my receive context, or could a new notification come)?

Maybe we need another file to go over the details, but it could quickly be redundant if we have to list all notifications again to give more details...

docs/api/QUIC_STREAM_CALLBACK.md
sm-msft: This is a boiler plate description in all of the callback method descriptions. I will leave it as-is.

docs/api/QUIC_STREAM_EVENT.md
guhetier: I agree sequence diagrams / state machines, etc... are not in the scope of this PR.
I still think some good (and short) explanation of what the event means for the user is needed here. I don't have a strong opinion on the exact words or if it should refer to other events or not.

docs/Execution.md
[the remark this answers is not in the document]
guhetier: :) I guess depending on how long "long" is, that's correct.
My angle is mostly about not scarring people so much about processing in a callback that they move everything in other threads and pessimize their implementations.
But there is some wording about that below...

docs/api/QUIC_STREAM_EVENT.md
```diff
@@ -117,11 +117,13 @@ Absolute offset of the current data payload from the start of the receive operat
 
 MsQuic indicates the total buffer length of the data in this parameter.
 
-If the application can processes only part of the received buffer synchronously in this call, it can be indicated to MsQuic by setting this parameter to the byte count processed and returning `QUIC_STATUS_CONTINUE` from this call.
-If the application desires to process the received data asynchronously, it should return `QUIC_STATUS_PENDING` from this call.
-
 See [Receiving Data On Streams](../Streams.md#Receiving) for further details on receiving data on a stream.
 
+Most notably, if the application...
+ - processes all the received data synchronously in this call, this parameter must be left unchanged and  `QUIC_STATUS_SUCCESS` must be returned from this call.
```
guhetier, line 123:
```suggestion
 - processes all the received data synchronously in this call, this parameter must be left unchanged and `QUIC_STATUS_SUCCESS` must be returned from this call.
```

extra space :)

docs/api/QUIC_STREAM_EVENT.md
```diff
@@ -117,11 +117,13 @@ Absolute offset of the current data payload from the start of the receive operat
...
+ - can processes only part of the received buffer synchronously in this call, it should be indicated to MsQuic by setting this parameter to the byte count processed and returning `QUIC_STATUS_CONTINUE` from this call.
```
guhetier, line 124:
```suggestion
 - can process only part of the received buffer synchronously in this call, it should be indicated to MsQuic by setting this parameter to the byte count processed and returning `QUIC_STATUS_CONTINUE` from this call.
```

docs/api/QUIC_STREAM_EVENT.md
```diff
@@ -156,7 +163,7 @@ Client context to match this event with the original `StreamSend` operation.
 
 ## QUIC_STREAM_EVENT_PEER_SEND_SHUTDOWN
 
-Indicates that the current send operation to the peer on this stream has been shutdown/completed.
+Indicates that the current send operation from the peer on this stream has been completed/shutdown.
```
guhetier, line 166: I don't think we should use "send operation" here: in MsQuic, we use "send operation" to mean a call to StreamSend.
Maybe we can use "the send direction of the stream" instead?

I would also say "gracefully completed/shutdown" just to be very clear this isn't emmited on an abortive shutdown.

docs/API.md
```diff
@@ -16,8 +16,8 @@ Term | Definition
 *endpoint* | One side of a connection; client or server.
 *peer* | The *other* side of a connection.
 *callback handler* | The function pointer the app registers with an MsQuic object.
-*app context* or<br> *context* | A (possibly null) pointer registered with an MsQuic object. It is passed to callback handlers.
-*event* | An upcall to a callback handler.
+*app context* or<br> *context* | A (possibly null) pointer registered with an MsQuic object. It is passed to App's callback handlers.
```
anrossi, line 19: > app
the word App is not capitalized in the rest of the documentation.

docs/API.md
```diff
@@ -16,8 +16,8 @@ Term | Definition
...
+*event* | Beginning or ending of a specific condition or action in MsQuic library that results in the App's callback hander invocation.
```
anrossi, line 20: > in the MsQuic library
also don't capitalize `app` here.

docs/Execution.md
```diff
@@ -6,15 +6,19 @@ The sections below detail the designs MsQuic uses and the reasons behind these c
 
 ## Event Model
 
-In the MsQuic API, all asynchronous state changes and notifications are indicated directly to the application via a callback.
-This includes connection state changes, new streams being created, stream data being received, and stream sends completing.
+MsQuic [Object Model](./API.md#object-model) describes the hierarchy of MsQuic objects.
+
+MsQuic API delivers all state changes and notifications for a specific MsQuic object directly to the corresponding callback handler registered by the application. These include connection state changes, new streams being created, stream data being received, and stream sends completing, among others.
```
anrossi, line 11: I think this should begin
> The MsQuic API ...
The rest sounds great!

docs/api/QUIC_STREAM_EVENT.md
anrossi: I think it should be more forceful than "...it should be indicated to MsQuic...". If only part of the data is processed, the app MUST change the TotalBufferLength, otherwise they'll lose data that wasn't processed

docs/api/QUIC_STREAM_EVENT.md
```diff
@@ -117,11 +117,13 @@ Absolute offset of the current data payload from the start of the receive operat
...
 
-If the application can processes only part of the received buffer synchronously in this call, it can be indicated to MsQuic by setting this parameter to the byte count processed and returning `QUIC_STATUS_CONTINUE` from this call.
-If the application desires to process the received data asynchronously, it should return `QUIC_STATUS_PENDING` from this call.
-
 See [Receiving Data On Streams](../Streams.md#Receiving) for further details on receiving data on a stream.
 
+Most notably, if the application...
+ - processes all the received data synchronously in this call, this parameter must be left unchanged and  `QUIC_STATUS_SUCCESS` must be returned from this call.
+ - can processes only part of the received buffer synchronously in this call, it should be indicated to MsQuic by setting this parameter to the byte count processed and returning `QUIC_STATUS_CONTINUE` from this call.
+ - desires to process the received data asynchronously, it should return `QUIC_STATUS_PENDING` from this call.
```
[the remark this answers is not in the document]
anrossi: I agree.

docs/api/QUIC_STREAM_EVENT.md
```diff
@@ -0,0 +1,273 @@
...
+
+Absolute offset of the current data payload from the start of the receive operation.
+
+`TotalBufferLength`
+
+MsQuic indicates the total buffer length of the data in this parameter.
+
+See [Receiving Data On Streams](../Streams.md#Receiving) for further details on receiving data on a stream.
+
+Most notably, if the application...
```
anrossi, line 122: More broadly speaking, should we have these key points here, or should we just have a link from here to the comprehensive documentation on all the different receive patterns in Streams.md (above)?
The benefit to a link is having only one place to update.  The benefit to having key points here is that readers don't have to go to a new document.

docs/api/QUIC_STREAM_EVENT.md
anrossi: Ditto.

ProjectsByJackHe (approved):
LGTM, approved to unblock.

anrossi (approved):

docs/Execution.md
```diff
@@ -29,52 +33,50 @@ QUIC_STATUS
     );
 ```
 
-Above is an example of a callback delivered to the listener interface.
-The application must register a per-object callback handler to manage all events MsQuic may indicate for that object, returning a status to show if it was successfully handled or not.
+The application must register a callback handler for every MsQuic object it creates. This handler must manage all the events MsQuic may indicate for that object. The handler must also return a status for each event indicating to MsQuic how the event was handled. This returned status is often success/failure, but sometimes indicates MsQuic that further processing is required.
```
nibanks, line 36: > for every MsQuic obejct it creates.

Some objects are created *by* MsQuic, but still need a callback handler. I'd reword this a bit.

docs/Execution.md
```diff
@@ -29,52 +33,50 @@ QUIC_STATUS
...
+- Writing correct, scalable code in every application built on top of the socket interfaces is a repetetive, challenging task prone to errors. Offloading the threading and synchronization to MsQuic enables every application to be scalable with minimal effort, making things "just work" out of the box.
 
-- It simplifies MsQuic's logic by eliminating the need for a queue or cached state to indicate to the application. In the socket model, the networking stack must wait for a top-down call from the application before indicating completion, increasing code size, complexity, and memory usage.
+- Simpler logic flow in MsQuic by eliminating a queue/cached state of yet to be delivered application notifications. This queue/cached state is maintained in the socket model to track yet-to-be-picked-up events/data and the networking stack must wait for call(s) from the application before indicating completion. This represents additional code, complexity and memory usage in socket model that MsQuic does without.
```
nibanks, line 47: > Simpler logic flow in MsQuic by eliminating

This statement doesn't quite make sense and need fixing.

nibanks, line 47: The second sentence also doesn't make sense to me. IMHO, the text before was clearer.

docs/Streams.md
```diff
@@ -79,6 +79,14 @@ The application can optimize its processing for that case but should be ready to
 
 When the buffer count is 0, it signifies the reception of a QUIC frame with empty data, which also indicates the end of stream data.
 
+## Summary - Common handling of receive data events
+
+If the application...
```
nibanks, line 84: There should be one more case mentioned: returning SUCCESS but reducing TotalBufferLength to indicate no more data can be processed right now, which pauses further receives.

docs/api/QUIC_CONNECTION_EVENT.md
```diff
@@ -267,7 +342,11 @@ The resumption ticket data previously sent to the client via [ConnectionSendResu
 
 ## QUIC_CONNECTION_EVENT_RESUMPTION_TICKET_RECEIVED
 
-This event indicates a TLS resumption ticket has been received from the server.
+This event indicates the client that a TLS resumption ticket has been received from the server.
```
nibanks, line 345:
```suggestion
This event indicates to the client that a TLS resumption ticket has been received from the server.
```

docs/api/QUIC_LISTENER_EVENT.md
```diff
@@ -1,10 +1,20 @@
...
-The payload for QUIC listener events.
+QUIC listener events and the corresponding payload
 
 # Syntax
 
+```C
+typedef enum QUIC_LISTENER_EVENT_TYPE {
+    QUIC_LISTENER_EVENT_NEW_CONNECTION      = 0,
+    QUIC_LISTENER_EVENT_STOP_COMPLETE       = 1,
+    QUIC_LISTENER_EVENT_DOS_MODE_CHANGED    = 2,
```
nibanks, line 12: I don't see any description of this one below.

nibanks (commented):
Generally, this is great. Thanks! Just a few things I think should be fixed in a follow up PR.

PR #4962, opened 2025-04-01
Rust: Change open api signature for handles

youyuanwu:
## Description
Connection, Stream, Listener and Configuration handles all requires create an empty struct using new() and then call open().
open() should return the struct, and new function can be eliminated. This also allows handle variables to be immutable.

## Testing
NA

## Documentation
NA

### DIFF
--- a/src/rs/lib.rs
+++ b/src/rs/lib.rs
@@ -689,6 +689,8 @@ macro_rules! define_quic_handle_ctx_fn {
 
             /// Consume ctx by dropping it.
             /// Set msquic ctx to null.
+            // This is used by Connection and Stream, but not Listener.
+            #[allow(dead_code)]
             fn consume_callback_ctx(&self) {
                 let res = unsafe { self.get_callback_ctx() };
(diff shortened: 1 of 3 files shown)


src/rs/lib.rs
```diff
@@ -874,32 +870,27 @@ extern "C" fn raw_conn_callback(
...
 
 impl Connection {
-    pub fn new() -> Connection {
-        Connection {
-            handle: ptr::null_mut(),
-        }
-    }
-
-    pub fn open<F>(&mut self, registration: &Registration, handler: F) -> Result<(), Status>
+    pub fn open<F>(registration: &Registration, handler: F) -> Result<Self, Status>
```
guhetier, line 873: Since the convention in Rust is generally to use `new` to create an object, do you think that calling the function `new` instead of `open` would make sense?

In C, since we manipulate handles, "open" works well, but in Rust, we create and return the object, so it looks like what a `new` typically does.

guhetier (approved):

PR #4967, opened 2025-04-02
Update documentation to clarify API ordering constraints

### LINKED ISSUE #3635: Explicitly document API ordering constraints

 I encountered a case where MsQuicStreamClose is stuck forever at the following line:

https://github.com/microsoft/msquic/blob/main/src/core/api.c#L783

I hooked debugger into the program and was unable to find another thread stuck in StreamCallback or ConnectionCallback (which, if happens, could explain the stuck StreamClose), so my suspicion is that it's because the connection was already closed, and therefore when Line 779 queued operation to connection, it can never be picked up to be executed, leaving CompletionEvent never to be signaled. 

But since the line 745 does `QUIC_CONN_VERIFY(Connection, !Connection->State.HandleClosed);` verify whether the parent connection is closed or not before queueing operation to it. How can the connection be closed when line 779 runs?

Because my program handles StreamClose on a separate thread, so the StreamClose can happen asynchronously to (e.g. later than) the connection closure. 

In additional, are there other API ordering constraints?

### Proposed solution

Improve documentation of all APIs that has ordering constraints.

### Additional context

_No response_

guhetier:
## Description

Update documentation to clarify some ordering constraints between API.
Focuses on "*Close" calls and the send and receive operations.

Closes #3635.

## Testing

N/A

## Documentation

Doc only

### DIFF
--- a/docs/Streams.md
+++ b/docs/Streams.md
@@ -33,8 +33,15 @@ For peer initiated streams, the app gets a `QUIC_CONNECTION_EVENT_PEER_STREAM_ST
 # Sending
 
 An app can send on any locally initiated stream or a peer initiated bidirectional stream. The app uses the [StreamSend](api/StreamSend.md) API to send data.
-MsQuic holds on to any buffers queued via [StreamSend](api/StreamSend.md) until they have been completed via the `QUIC_STREAM_EVENT_SEND_COMPLETE` event:
-the app must not free, reuse or otherwise access the buffers provided through [StreamSend](api/StreamSend.md) before the matching completion notification.
+
+MsQuic takes ownership of any buffers successfully queued via [StreamSend](api/StreamSend.md). Buffer ownership is
+returned to the application via the `QUIC_STREAM_EVENT_SEND_COMPLETE` event. The application must not free, reuse or
+otherwise access a buffer provided to [StreamSend](api/StreamSend.md) until the matching
+`QUIC_STREAM_EVENT_SEND_COMPLETE` event.
+
+![Note]
+> `QUIC_STREAM_EVENT_SEND_COMPLETE` does not mean the data has been received by the peer, or even the it has been put on
+> the wire - only that MsQuic no longer needs the app send buffer.
 
 ## Send Buffering
 
@@ -59,7 +66,7 @@ To disable internal send buffering and use the second mode, the app must set `Se
 
 The send direction can be shut down in three different ways:
 
-- **Graceful** - The sender can gracefully shut down the send direction by calling [StreamShutdown](api/StreamShutdown.md) with the `QUIC_STREAM_SHUTDOWN_FLAG_GRACEFUL` flag or by including the `QUIC_SEND_FLAG_FIN` flag on the last [StreamSend](api/StreamSend.md) call. In this scenario all data will be delivered to the peer and then the peer is informed the stream has been gracefully shut down.
+- **Graceful** - The sender can gracefully shut down the send direction by calling [StreamShutdown](api/StreamShutdown.md) with the `QUIC_STREAM_SHUTDOWN_FLAG_GRACEFUL` flag or by including the `QUIC_SEND_FLAG_FIN` flag on the last [StreamSend](api/StreamSend.md) call. In this scenario all data will first be delivered to the peer, then the peer is informed the stream has been gracefully shut down.
 
 - **Sender Abort** - The sender can abortively shut down the send direction by calling [StreamShutdown](api/StreamShutdown.md) with the `QUIC_STREAM_SHUTDOWN_FLAG_ABORT_SEND` flag. In this scenario, all outstanding sends are immediately canceled and are not delivered to the peer. The peer is immediately informed of the abort.
 
@@ -186,3 +193,19 @@ MsQuic does not enforce it, and it is legal for an application to provide less b
 
 After the initial receive window is full, flow control will ensure that the peer does not send more data than there is buffer space available.
 However, the application should still provide enough buffer space to keep flow control from impacting performances.
+
+## Receive Shutdown
+
+The receiver can abortively shutdown a stream receive direction by calling [`StreamShutdown`](api/StreamShutdown.md) 
+with the `QUIC_STREAM_SHUTDOWN_FLAG_ABORT_RECEIVE` option.
+
+# Closing a Stream
+
+Once all the directions relevant for a stream have been shutdown, the application receives a
+`QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE` event.
+
+The application must then close the stream using [`StreamClose`](api/StreamClose.md), and can release its notification
+handling context.
+
+Closing a stream before the `QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE` is only allowed if the stream was not started yet.
+Otherwise, it will result in the connection being closed abortively.
--- a/docs/Streams.md
+++ b/docs/Streams.md
@@ -40,10 +40,9 @@ otherwise access a buffer provided to [StreamSend](api/StreamSend.md) until the 
 `QUIC_STREAM_EVENT_SEND_COMPLETE` event.
 
 ![Note]
-> `QUIC_STREAM_EVENT_SEND_COMPLETE` does not mean the data has been received by the peer app, the peer transport, or
-> even the it has been put on the wire - it only means that MsQuic no longer needs the app send buffer and give the
-> owernship back to the application. The app should not assume anything about whether the data has been successfully
-> transmitted based on this notification.
+> `QUIC_STREAM_EVENT_SEND_COMPLETE` does not mean the data has been received by the peer application layer.
+> It only means that MsQuic no longer needs the app send buffer and give the owernship back to the application. The app
+> should assume the data has been successfully transmitted based on this notification.
 
 ## Send Buffering
 
@@ -207,7 +206,7 @@ Once all the directions relevant for a stream have been shutdown, the applicatio
 `QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE` event.
 
 The application must then close the stream using [`StreamClose`](api/StreamClose.md), and can release its notification
-handling context one the call returns.
+handling context once the call returns.
 
 If the app closes a stream before it is shutdown, the stream will be shutdown abortively with an error code of `0`.
 This should be avoided, the app should rather abortively shutdown the stream first with a meaningful error code.
--- a/docs/Streams.md
+++ b/docs/Streams.md
@@ -40,8 +40,10 @@ otherwise access a buffer provided to [StreamSend](api/StreamSend.md) until the 
 `QUIC_STREAM_EVENT_SEND_COMPLETE` event.
 
 ![Note]
-> `QUIC_STREAM_EVENT_SEND_COMPLETE` does not mean the data has been received by the peer, or even the it has been put on
-> the wire - only that MsQuic no longer needs the app send buffer.
+> `QUIC_STREAM_EVENT_SEND_COMPLETE` does not mean the data has been received by the peer app, the peer transport, or
+> even the it has been put on the wire - it only means that MsQuic no longer needs the app send buffer and give the
+> owernship back to the application. The app should not assume anything about whether the data has been successfully
+> transmitted based on this notification.
 
 ## Send Buffering
 
@@ -205,7 +207,9 @@ Once all the directions relevant for a stream have been shutdown, the applicatio
 `QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE` event.
 
 The application must then close the stream using [`StreamClose`](api/StreamClose.md), and can release its notification
-handling context.
+handling context one the call returns.
 
-Closing a stream before the `QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE` is only allowed if the stream was not started yet.
-Otherwise, it will result in the connection being closed abortively.
+If the app closes a stream before it is shutdown, the stream will be shutdown abortively with an error code of `0`.
+This should be avoided, the app should rather abortively shutdown the stream first with a meaningful error code.
+It is however possible for an application to abortively shutdown a stream and immediately close it from the same thread,
+without waiting for the `QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE` event.
--- a/docs/Streams.md
+++ b/docs/Streams.md
@@ -42,7 +42,7 @@ otherwise access a buffer provided to [StreamSend](api/StreamSend.md) until the 
 ![Note]
 > `QUIC_STREAM_EVENT_SEND_COMPLETE` does not mean the data has been received by the peer application layer.
 > It only means that MsQuic no longer needs the app send buffer and give the owernship back to the application. The app
-> should assume the data has been successfully transmitted based on this notification.
+> should *not* assume the data has been successfully transmitted based on this notification.
 
 ## Send Buffering
 
@@ -202,7 +202,7 @@ with the `QUIC_STREAM_SHUTDOWN_FLAG_ABORT_RECEIVE` option.
 
 # Closing a Stream
 
-Once a stream has been shutdown (on both direction for a bi-directional stream), the application receives a
+Once a stream has been shutdown (in both direction for a bi-directional stream), the application receives a
 `QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE` event.
 
 The application must then close the stream using [`StreamClose`](api/StreamClose.md), and can release its context
@@ -210,5 +210,5 @@ pointer safely once the call returns.
 
 If the app closes a stream before it is shutdown, the stream will be shutdown abortively with an error code of `0`.
 This should be avoided; instead the app should abortively shutdown the stream first with a meaningful error code.
-It is however possible for an application to abortively shutdown a stream and immediately close it from the same thread,
+It is possible for an application to abortively shutdown a stream and immediately close it from the same thread,
 without waiting for the `QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE` event.
--- a/docs/Streams.md
+++ b/docs/Streams.md
@@ -202,13 +202,13 @@ with the `QUIC_STREAM_SHUTDOWN_FLAG_ABORT_RECEIVE` option.
 
 # Closing a Stream
 
-Once all the directions relevant for a stream have been shutdown, the application receives a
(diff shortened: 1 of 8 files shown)


docs/api/ConnectionClose.md
```diff
@@ -25,11 +25,15 @@ The valid handle to an open connection object.
 
 `ConnectionClose` cleans up and frees all resources allocated for the connection in `ConnectionOpen`.
 
-A caller should shutdown an active connection via [ConnectionShutdown](ConnectionShutdown.md) before calling `ConnectionClose`; calling `ConnectionClose` without [ConnectionShutdown](ConnectionShutdown.md) will implicitly call [ConnectionShutdown](ConnectionShutdown.md) with the `QUIC_CONNECTION_SHUTDOWN_FLAG_SILENT` flag.
+A caller should shutdown an active connection via [ConnectionShutdown](ConnectionShutdown.md) before calling
+`ConnectionClose`. Calling `ConnectionClose` without [ConnectionShutdown](ConnectionShutdown.md) will abortively and
+silently shutdown the connection and all associated streams through an implicitly call to
```
ProjectsByJackHe, line 30: "through an implicitly" ---> "through an implicit"

docs/api/ConnectionClose.md
```diff
@@ -25,11 +25,15 @@ The valid handle to an open connection object.
 
 `ConnectionClose` cleans up and frees all resources allocated for the connection in `ConnectionOpen`.
 
-A caller should shutdown an active connection via [ConnectionShutdown](ConnectionShutdown.md) before calling `ConnectionClose`; calling `ConnectionClose` without [ConnectionShutdown](ConnectionShutdown.md) will implicitly call [ConnectionShutdown](ConnectionShutdown.md) with the `QUIC_CONNECTION_SHUTDOWN_FLAG_SILENT` flag.
+A caller should shutdown an active connection via [ConnectionShutdown](ConnectionShutdown.md) before calling
+`ConnectionClose`. Calling `ConnectionClose` without [ConnectionShutdown](ConnectionShutdown.md) will abortively and
```
ProjectsByJackHe, line 29: Is it the associated streams that are shutdown abortively and connection shutdown silently? If so, I think we should mention it explicitly.

docs/api/StreamReceiveComplete.md
```diff
@@ -17,14 +17,38 @@ void
...
+
+## Default behavior
+
+The application must ensure that for each `QUIC_STREAM_EVENT_RECEIVE` processed asynchronously (`QUIC_STATUS_PENDING`
+was returned from the callback), `StreamReceiveComplete` is called exactly once.
+
+Duplicate `StreamReceiveComplete` calls after a `QUIC_STREAM_EVENT_RECEIVE` event are ignored
+silently, even when a different `BufferLength` is provided.
+
+If `BufferLength` is smaller than the number of bytes indicated in the matching `QUIC_STREAM_EVENT_RECEIVE`, MsQuic will
```
ProjectsByJackHe, line 35: I personally prefer the extra details, but maybe it goes into too much detail? @nibanks thoughts?

docs/Streams.md
```diff
@@ -33,8 +33,15 @@ For peer initiated streams, the app gets a `QUIC_CONNECTION_EVENT_PEER_STREAM_ST
...
-MsQuic holds on to any buffers queued via [StreamSend](api/StreamSend.md) until they have been completed via the `QUIC_STREAM_EVENT_SEND_COMPLETE` event:
-the app must not free, reuse or otherwise access the buffers provided through [StreamSend](api/StreamSend.md) before the matching completion notification.
+
+MsQuic takes ownership of any buffers successfully queued via [StreamSend](api/StreamSend.md). Buffer ownership is
+returned to the application via the `QUIC_STREAM_EVENT_SEND_COMPLETE` event. The application must not free, reuse or
+otherwise access a buffer provided to [StreamSend](api/StreamSend.md) until the matching
+`QUIC_STREAM_EVENT_SEND_COMPLETE` event.
+
+![Note]
+> `QUIC_STREAM_EVENT_SEND_COMPLETE` does not mean the data has been received by the peer, or even the it has been put on
```
nibanks, line 43: So, I think we need to tease this out. When you are using send buffering, send complete means it's been copied locally. But if you're not, then it does mean the peer _transport_ has received the data - **not** the peer application.

docs/Streams.md
```diff
@@ -186,3 +193,19 @@ MsQuic does not enforce it, and it is legal for an application to provide less b
...
+The receiver can abortively shutdown a stream receive direction by calling [`StreamShutdown`](api/StreamShutdown.md) 
+with the `QUIC_STREAM_SHUTDOWN_FLAG_ABORT_RECEIVE` option.
+
+# Closing a Stream
+
+Once all the directions relevant for a stream have been shutdown, the application receives a
+`QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE` event.
+
+The application must then close the stream using [`StreamClose`](api/StreamClose.md), and can release its notification
+handling context.
```
nibanks, line 208:
```suggestion
handling context once the call returns.
```

docs/Streams.md
```diff
@@ -186,3 +193,19 @@ MsQuic does not enforce it, and it is legal for an application to provide less b
...
+
+Closing a stream before the `QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE` is only allowed if the stream was not started yet.
+Otherwise, it will result in the connection being closed abortively.
```
nibanks, line 211: This is not strictly correct. If you close a stream before you've shut down the stream, the stream will be aborted implicitly (with error code 0). This is not a great behavior because we don't actually know what error 0 will mean - so generally an app really shouldn't do that.

But if the application aborts a stream with an error code, it may immediately close the stream if doesn't care to wait for the completion. This is a completely reasonable behavior.

docs/api/StreamReceiveComplete.md
nibanks: Which extra details do you feel are missing?

docs/api/StreamReceiveComplete.md
```diff
@@ -17,14 +17,38 @@ void
...
+In Multi-receive mode, calls to `StreamReceiveComplete` do not need to match `QUIC_STREAM_EVENT_RECEIVE`: there can be
+either more or less calls to `StreamReceiveComplete` than `QUIC_STREAM_EVENT_RECEIVE` events.
+
+MsQuic will keep on indicating `QUIC_STREAM_EVENT_RECEIVE` irrespectively from calls to `StreamReceiveComplete` and the
+value of `BufferLength`.
+
+The application must keep track of the accumulated `TotalBufferLength` from `QUIC_STREAM_EVENT_RECEIVE` events and ensure that:
+- the sum of all `BufferLength` parameters in `StreamReceiveComplete` calls is always smaller or equal than the number
+    of bytes received on the stream
+- all bytes received are eventually completed in `StreamReceiveComplete` call
```
nibanks, line 51: OR the stream receive is aborted

docs/api/ConnectionClose.md
guhetier: No, the connection is closed abortively and silently (as in, abortively != graceful, and silently = we don't even send an on-the-wire message to the peer about it).

Basically, if you call `ConnectionClose` without calling `ConnectionShutdown` first, we do a `ConnectionShutdown(QUIC_CONNECTION_SHUTDOWN_FLAG_SILENT)` on behalf of the app.

I improved the wording a tiny bit in the next iteration, but if that still feel unclear, I'm open to a suggestion for something better.

docs/Streams.md
guhetier: Hum, I think I wrote this because of this comment in our code, in `QuicStreamClose`:

```
            //
            // TODO - If the stream hasn't been aborted already, then this is a
            // fatal error for the connection. The QUIC transport cannot "just
            // pick an error" to shutdown the stream with. It must abort the
            // entire connection.
            //
```

Is it no longer the case? I will delete the comment if so.

nibanks: Well, we don't actually kill the connection though; just log a warning. My description above still holds I think.

docs/Streams.md
nibanks: Then I would just say it "does not mean the data has been received by the peer application layer."

docs/Streams.md
```diff
@@ -186,3 +194,21 @@ MsQuic does not enforce it, and it is legal for an application to provide less b
...
+# Closing a Stream
+
+Once all the directions relevant for a stream have been shutdown, the application receives a
+`QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE` event.
+
+The application must then close the stream using [`StreamClose`](api/StreamClose.md), and can release its notification
+handling context once the call returns.
+
+If the app closes a stream before it is shutdown, the stream will be shutdown abortively with an error code of `0`.
+This should be avoided, the app should rather abortively shutdown the stream first with a meaningful error code.
```
anrossi, line 212: I think this would read better as
> This should be avoided; instead, the app should abortively shutdown the stream first with a meaningful error code.

docs/Streams.md
```diff
@@ -186,3 +194,21 @@ MsQuic does not enforce it, and it is legal for an application to provide less b
...
+
+The receiver can abortively shutdown a stream receive direction by calling [`StreamShutdown`](api/StreamShutdown.md) 
+with the `QUIC_STREAM_SHUTDOWN_FLAG_ABORT_RECEIVE` option.
+
+# Closing a Stream
+
+Once all the directions relevant for a stream have been shutdown, the application receives a
+`QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE` event.
+
+The application must then close the stream using [`StreamClose`](api/StreamClose.md), and can release its notification
```
anrossi, line 208: Is it called the "notification handling context" elsewhere in the documentation?  It might be clearer to describe the context as the pointer provided in `StreamOpen`, or use whatever description is used elsewhere.

docs/Streams.md
```diff
@@ -186,3 +194,21 @@ MsQuic does not enforce it, and it is legal for an application to provide less b
...
 However, the application should still provide enough buffer space to keep flow control from impacting performances.
+
+## Receive Shutdown
+
+The receiver can abortively shutdown a stream receive direction by calling [`StreamShutdown`](api/StreamShutdown.md) 
+with the `QUIC_STREAM_SHUTDOWN_FLAG_ABORT_RECEIVE` option.
+
+# Closing a Stream
+
+Once all the directions relevant for a stream have been shutdown, the application receives a
```
anrossi, line 205: All the directions relevant? Are there more directions than send and receive? 😄  Joking aside, I get that you're trying to cover unidirectional and bidirectional streams in one statement. Maybe something like 
> Once both directions for a stream have been shutdown (unidirectional streams start with one direction already shutdown) ...

guhetier: Yes, I've been struggling to find a wording that is correct for both uni and bi-directional streams + the send or receive direction.

Maybe "Once a stream has been shutdown (both directions for a bi-directional stream), ..."

docs/Streams.md
```diff
@@ -33,8 +33,16 @@ For peer initiated streams, the app gets a `QUIC_CONNECTION_EVENT_PEER_STREAM_ST
...
+
+MsQuic takes ownership of any buffers successfully queued via [StreamSend](api/StreamSend.md). Buffer ownership is
+returned to the application via the `QUIC_STREAM_EVENT_SEND_COMPLETE` event. The application must not free, reuse or
+otherwise access a buffer provided to [StreamSend](api/StreamSend.md) until the matching
+`QUIC_STREAM_EVENT_SEND_COMPLETE` event.
+
+![Note]
+> `QUIC_STREAM_EVENT_SEND_COMPLETE` does not mean the data has been received by the peer application layer.
+> It only means that MsQuic no longer needs the app send buffer and give the owernship back to the application. The app
+> should assume the data has been successfully transmitted based on this notification.
```
csujedihy, line 45: "data has been successfully **transmitted** based on this notification" -would this statement cause confusions as it could be just buffered by msquic or afd?

docs/Streams.md
```diff
@@ -186,3 +194,21 @@ MsQuic does not enforce it, and it is legal for an application to provide less b
...
 However, the application should still provide enough buffer space to keep flow control from impacting performances.
+
+## Receive Shutdown
+
+The receiver can abortively shutdown a stream receive direction by calling [`StreamShutdown`](api/StreamShutdown.md) 
+with the `QUIC_STREAM_SHUTDOWN_FLAG_ABORT_RECEIVE` option.
+
+# Closing a Stream
+
+Once a stream has been shutdown (on both direction for a bi-directional stream), the application receives a
```
anrossi, line 205: I think it reads better as "in both directions"

anrossi: typo aside, this reads much better than the previous wording!

src/core/stream.c
```diff
@@ -393,12 +393,6 @@ QuicStreamClose(
     if (!Stream->Flags.ShutdownComplete) {
 
         if (Stream->Flags.Started && !Stream->Flags.HandleShutdown) {
-            //
```
anrossi, line 396: Maybe remove the TODO but keep the comment as a signpost for investigating this scenario?

docs/api/StreamReceiveComplete.md
```diff
@@ -17,14 +17,41 @@ void
...
+
+If `BufferLength` is smaller than the number of bytes indicated in the matching `QUIC_STREAM_EVENT_RECEIVE`, MsQuic will
+stop indicating new `QUIC_STREAM_EVENT_RECEIVE` events until a call to [`StreamReceiveSetEnabled`](StreamReceiveSetEnabled.md).
+
+## Multi-receive Mode
+
+If Multi-receive mode has been enabled on the connection, the behavior is different from default behavior detailed above.
+
+In Multi-receive mode, calls to `StreamReceiveComplete` do not need to match `QUIC_STREAM_EVENT_RECEIVE`: there can be
+either more or less calls to `StreamReceiveComplete` than `QUIC_STREAM_EVENT_RECEIVE` events.
```
anrossi, line 45: In this instance, "fewer" is more appropriate than "less" because calls to `StreamReceiveComplete` are countable in the comparative with `QUIC_STREAM_EVENT_RECEIVE` events.

docs/api/StreamClose.md
```diff
@@ -23,12 +23,15 @@ A Stream handle from a previous call to [StreamOpen](StreamOpen.md).
 
 # Remarks
 
-The application should only close a stream after it has been completely shut down or it was never successfully started. Closing a stream before it has been completely shut down produces **undefined behavior** because clean up of the stream **must** be reflected on the wire with an application specific error code. When the app closes a stream without first shutting down, MsQuic has to guess which error code to use (currently uses `0`) when sending the state change to the peer.
+The application should only close a stream after it has been completely shut down or if it was never successfully started. Closing a stream before it has been completely shut down produces **undefined behavior** because clean up of the stream **must** be reflected on the wire with an application specific error code. When the app closes a stream without first shutting down, MsQuic has to guess which error code to use (currently uses `0`) when sending the state change to the peer.
```
anrossi, line 26: I believe that code comment in stream.c is in a block that will abort the entire connection if a stream is closed without being shutdown first. 
This remark here (and possibly also in Streams.md) should mention that this undefined behavior will **terminate the connection**, not just the stream, which seems implied by "[MsQuic] guess[ing] the error code".

docs/Streams.md
nibanks: Yeah, let's go ahead and remove this last sentence.

docs/Streams.md
```diff
@@ -186,3 +194,21 @@ MsQuic does not enforce it, and it is legal for an application to provide less b
...
+`QUIC_STREAM_EVENT_SHUTDOWN_COMPLETE` event.
+
+The application must then close the stream using [`StreamClose`](api/StreamClose.md), and can release its context
+pointer safely once the call returns.
+
+If the app closes a stream before it is shutdown, the stream will be shutdown abortively with an error code of `0`.
+This should be avoided; instead the app should abortively shutdown the stream first with a meaningful error code.
+It is however possible for an application to abortively shutdown a stream and immediately close it from the same thread,
```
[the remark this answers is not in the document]
nibanks: It should work on any thread but a QUIC worker thread that is used to process a different connection.

docs/Streams.md
guhetier: There is a "not" missing. I meant to say "The app should *not* assume the data has been successfully transmitted".

I think there is value in clarifying it, we got some questions about it a month ago or so.

src/core/stream.c
guhetier: The comment doesn't reflect what the code does or what Nick says the behavior is supposed to be (closing the stream without aborting it is supposed to abort the stream implicitly with error code `0`).
IMO, this comment has a negative value, it is misleading + doesn't point to something that actually needs to be done.

docs/api/StreamClose.md
guhetier: :) We are going in a loop!

I started with the documentation stating what the comment says, then Nick mentioned the comment is wrong, so I removed the comment and adapted this paragraph.

Github is not helping, not showing comments on previous iterations.

docs/Streams.md
guhetier:
```suggestion
> should *not* assume the data has been successfully transmitted based on this notification.
```

docs/Streams.md
guhetier:
```suggestion
Once a stream has been shutdown (in both direction for a bi-directional stream), the application receives a
```

docs/Streams.md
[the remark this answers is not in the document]
guhetier:
```suggestion
It is possible for an application to abortively shutdown a stream and immediately close it from the same thread,
```

It can be any thread, as long as the operation will be serialized (inline execution from notification callback in a worker thread or two operations queued from an app thread).

The broader truth is "the app can do whatever as long as the abortive shutdown operation is processed by MsQuic before the close operation", but I think trying to communicate this would be confusing and that it wouldn't be common a pattern.

docs/api/StreamReceiveComplete.md
guhetier:
```suggestion
either more or fewer calls to `StreamReceiveComplete` than `QUIC_STREAM_EVENT_RECEIVE` events.
```

nibanks (approved):

docs/api/StreamClose.md
anrossi: yeah, I wish github would show a tombstone of the previous comment or something. Sorry to run you around in circles.

PR #4969, opened 2025-04-03
Follow up to event documentation updates

sm-msft:
## Description
- Address review feedback from pull request #4960 
- Add missing events/event payload descriptions
- Add a doc for preview features

Fixes [#1795](https://github.com/microsoft/msquic/issues/1795)

### DIFF
--- a/docs/Execution.md
+++ b/docs/Execution.md
@@ -33,7 +33,7 @@ QUIC_STATUS
     );
 ```
 
-The application must register a callback handler for every MsQuic object it creates, as well as for some objects automatically created by MsQuic (TODO: add reference).
+The application must register a callback handler for every MsQuic object it creates, as well as for some objects automatically created by MsQuic (such as those created for [Peer Initiated Streams](Streams.md#Peer_Initiated_Streams) or [Server Connections](./api/QUIC_LISTENER_EVENT.md#New_Connection)).
 
 (TODO: add notes on Preview feature events)
 
--- a/docs/Execution.md
+++ b/docs/Execution.md
@@ -35,9 +35,9 @@ QUIC_STATUS
 
 The application must register a callback handler for every MsQuic object it creates, as well as for some objects automatically created by MsQuic (such as those created for [Peer Initiated Streams](Streams.md#Peer_Initiated_Streams) or [Server Connections](./api/QUIC_LISTENER_EVENT.md#New_Connection)).
 
-(TODO: add notes on Preview feature events)
-
 This handler must manage all the events MsQuic may indicate for that object. The handler must also return a status for each event indicating to MsQuic how the event was handled. This returned status is often success/failure, but sometimes indicates MsQuic that further processing is required.
+
+This handler must return `QUIC_STATUS_SUCCESS` for all events it currently does not handle (default case) to insulate the code from future additions to list of events for a specific object.
 
 This approach differs significantly from sockets and most networking libraries, where the application must make a call (e.g., `send` or `recv`) to determine if something happened.
 This design choice was made for several reasons:
--- a/docs/Execution.md
+++ b/docs/Execution.md
@@ -33,7 +33,11 @@ QUIC_STATUS
     );
 ```
 
-The application must register a callback handler for every MsQuic object it creates. This handler must manage all the events MsQuic may indicate for that object. The handler must also return a status for each event indicating to MsQuic how the event was handled. This returned status is often success/failure, but sometimes indicates MsQuic that further processing is required.
+The application must register a callback handler for every MsQuic object it creates, as well as for some objects automatically created by MsQuic (TODO: add reference).
+
+(TODO: add notes on Preview feature events)
+
+This handler must manage all the events MsQuic may indicate for that object. The handler must also return a status for each event indicating to MsQuic how the event was handled. This returned status is often success/failure, but sometimes indicates MsQuic that further processing is required.
 
 This approach differs significantly from sockets and most networking libraries, where the application must make a call (e.g., `send` or `recv`) to determine if something happened.
 This design choice was made for several reasons:
@@ -44,7 +48,7 @@ This design choice was made for several reasons:
 
 - Writing correct, scalable code in every application built on top of the socket interfaces is a repetetive, challenging task prone to errors. Offloading the threading and synchronization to MsQuic enables every application to be scalable with minimal effort, making things "just work" out of the box.
 
-- Simpler logic flow in MsQuic by eliminating a queue/cached state of yet to be delivered application notifications. This queue/cached state is maintained in the socket model to track yet-to-be-picked-up events/data and the networking stack must wait for call(s) from the application before indicating completion. This represents additional code, complexity and memory usage in socket model that MsQuic does without.
+- Simplifies MsQuic logic by eliminating a queue/cached state to track yet-to-be-delivered application notifications/data. Such state is maintained in the socket model and the networking stack must wait for call(s) from the application before indicating completion. This represents reduced code, complexity and memory usage in the MsQuic library compared to the socket model.
 
 ### Writing Event Handlers
 
--- a/docs/PreviewFeatures.md
+++ b/docs/PreviewFeatures.md
@@ -9,17 +9,17 @@ Preview features
 > Please be aware of these additional risks before using the preview feature API.
 >
 
-New features added to the MsQuic library are exposed though newly added API in a staged manner to enable gradual adaption and roll in.
+New features added to the MsQuic library are exposed through newly added API as a preview, for stabilization and gradual adoption.
 
-These new API are first made available in the export headers only in the **Preview mode**. The underlying functionality is still present in the MsQuic library, but is unreachable through regular API.
+These new API are first made available in the export headers only in the **Preview mode**. The underlying functionality is still present in the MsQuic library but is unreachable through regular API.
 
 When a new feature is deemed ready for wider use, its API is moved from the preview portion to the regular portion of MsQuic library headers.
 
 ## QUIC_API_ENABLE_PREVIEW_FEATURES
 
-A preview feature's API declarations (typically in msquic.hpp/msquic.h) are placed within pre-processor conditional sections that are enabled using the QUIC_API_ENABLE_PREVIEW_FEATURES macro, to isolate it from the rest of the established API.
+A preview feature's API declarations (typically in msquic.h) are placed within pre-processor conditional sections that are enabled using the QUIC_API_ENABLE_PREVIEW_FEATURES macro, to isolate it from the rest of the established API.
 
-Any application wanting to use the preview features must declare this macro (and set to 1) before including the MsQuic headers.
+Any application wanting to use the preview features must declare this macro (and set it to 1) before including the MsQuic headers.
 
 ## Current list of Preview features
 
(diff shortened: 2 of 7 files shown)


docs/PreviewFeatures.md
```diff
@@ -0,0 +1,23 @@
+Preview features
+=========
+
+New features are phased in to the MsQuic library in a staged manner to enable gradual roll in, when each of these features are deemed ready for wider use.
+
+These features are first made available in a library-wide **Preview mode** till they are made part of the mainstream MsQuic library functionality.
```
sm-msft, line 6: I am making this up after reading the code :-) . Any input on this is welcome.

docs/PreviewFeatures.md
```diff
@@ -0,0 +1,23 @@
...
+
```
guhetier, line 7: I would add a big red disclaimer here (maybe using the github tag below) to say that preview features are not a stable API and can be modified or removed at any time without warning.
Just to make sure nobody can say they didn't know!

```
> [!IMPORTANT]  
> Crucial information necessary for users to succeed.
```

docs/PreviewFeatures.md
```diff
@@ -0,0 +1,23 @@
...
+## QUIC_API_ENABLE_PREVIEW_FEATURES
+
+Preview feature source code is placed within c pre-processor conditional sections that are enabled using the QUIC_API_ENABLE_PREVIEW_FEATURES macro, to isolate it from the rest of the established code.
```
guhetier, line 10: To be pedantic, only the API definitions are disabled by the macro. The implementation of a preview feature is always enabled. The goal is more avoiding that an app starts using a preview feature by mistake than isolating the code.

docs/api/QUIC_LISTENER_EVENT.md
```diff
@@ -72,6 +76,22 @@ This flag indicates the server has called [ListenerClose](ListenerClose.md).
 
 This field reserved for future use. Do not use.
 
+## QUIC_LISTENER_EVENT_DOS_MODE_CHANGED
+
+This event indicates an automated change in the DoS mode on the listner object in the MsQuic library. This event is delivered only for listener objects that have opted in for DoS mode change events. Refer to the [Listener Parameters](../Settings.md#Listener_Parameters) documentation for further information.
```
guhetier, line 81:
```suggestion
This event indicates an automated change in the DoS mode on the listener object in the MsQuic library. This event is delivered only for listener objects that have opted in for DoS mode change events. Refer to the [Listener Parameters](../Settings.md#Listener_Parameters) documentation for further information.
```

docs/api/QUIC_LISTENER_EVENT.md
```diff
@@ -72,6 +76,22 @@ This flag indicates the server has called [ListenerClose](ListenerClose.md).
...
+
+### DOS_MODE_CHANGED
+
+Details of the DoS mode change is indicated in the `DOS_MODE_CHANGED` struct/union.
+
+`DosModeEnabled`
+
+If TRUE, DoS mode is currently enabled on the listner object, else DoS mode is currently disabled on the listner object.
```
guhetier, line 89:
```suggestion
If TRUE, DoS mode is currently enabled on the listener object, else DoS mode is currently disabled on the listener object.
```

docs/api/QUIC_CONNECTION_EVENT.md
```diff
@@ -384,10 +410,72 @@ Pointer to a platform/TLS specific certificate chain. Valid only during the call
 
 If `QUIC_CREDENTIAL_FLAG_USE_PORTABLE_CERTIFICATES` was specified in the [QUIC_CREDENTIAL_CONFIG](QUIC_CREDENTIAL_CONFIG.md), this will be a `QUIC_BUFFER` containing the PKCS #7 DER (binary) encoded certificate chain.
 
+## QUIC_CONNECTION_EVENT_RELIABLE_RESET_NEGOTIATED
+
```
guhetier, line 414: nit: I think we should add a "preview feature" tag to the preview feature events.

We have something similar for (at least some) of the preview APIs (https://github.com/microsoft/msquic/blob/main/docs/api/StreamProvideReceiveBuffers.md)

docs/PreviewFeatures.md
```diff
@@ -0,0 +1,23 @@
...
+
+This macro is declared in code and set to be enabled by default. Users can change this macro per their deployment needs and recompile the code to include or exclude the preview features.
```
nibanks, line 12: As Guillaume states above, it's enabled by default in the library, but not for the callers. Let's make sure this is clear in the language here.

docs/PreviewFeatures.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Current list of Preview features
+
```
nibanks, line 15: Instead of empty sections below, let's at least add "TODO" comments.

docs/PreviewFeatures.md
sm-msft: I couldnt make that connection from the code. Will fix the wording.

docs/api/QUIC_CONNECTION_EVENT.md
sm-msft: I will mark these. 

Am I understanding this correctly - it looks like these events will be generated all the time but the application code does not know how to interpret them if it doesnt enable preview mode?

guhetier: There is no check that explicitly prevents the events from being generated.
 For this specific one (and most of the others, I think), they would only be emitted if the app requested a specific feature / setting though, so as long as the setting is under preview, apps that are not onboarding in preview features should not see the event.

sm-msft: Added a TODO for this.

docs/PreviewFeatures.md
```diff
@@ -0,0 +1,43 @@
...
+> [!IMPORTANT]
+>
+> Preview features are recently included features in the MsQuic library.
+>
+> The API for the preview features should be considered unstable / not fully baked in / subject to change.
+>
+> Please be aware of these additional risks before using the preview feature API.
+>
+
+New features added to the MsQuic library are exposed though newly added API in a staged manner to enable gradual adaption and roll in.
```
guhetier, line 12:
```suggestion
New features added to the MsQuic library are exposed through newly added API in a staged manner to enable gradual adoption and roll in.
```

Besides the typos, I am not sure about the "gradual adoption and roll in" part. To me, it sounds like a CFR, which isn't what we do here - it is more about a stabilization period or a preview.

Maybe we can simplify by saying that "New API and features are first made available in the export headers...." etc, basically skipping the first paragraph?

docs/Execution.md
```diff
@@ -33,7 +33,11 @@ QUIC_STATUS
     );
 ```
 
-The application must register a callback handler for every MsQuic object it creates. This handler must manage all the events MsQuic may indicate for that object. The handler must also return a status for each event indicating to MsQuic how the event was handled. This returned status is often success/failure, but sometimes indicates MsQuic that further processing is required.
+The application must register a callback handler for every MsQuic object it creates, as well as for some objects automatically created by MsQuic (TODO: add reference).
+
+(TODO: add notes on Preview feature events)
```
ProjectsByJackHe, line 38: Will there be a follow up PR on this?

ProjectsByJackHe (approved):

docs/PreviewFeatures.md
```diff
@@ -0,0 +1,43 @@
...
+
+These new API are first made available in the export headers only in the **Preview mode**. The underlying functionality is still present in the MsQuic library, but is unreachable through regular API.
+
+When a new feature is deemed ready for wider use, its API is moved from the preview portion to the regular portion of MsQuic library headers.
+
+## QUIC_API_ENABLE_PREVIEW_FEATURES
+
+A preview feature's API declarations (typically in msquic.hpp/msquic.h) are placed within pre-processor conditional sections that are enabled using the QUIC_API_ENABLE_PREVIEW_FEATURES macro, to isolate it from the rest of the established API.
```
guhetier, line 20:
```suggestion
A preview feature's API declarations (typically in msquic.hpp/msquic.h) are placed within pre-processor conditional sections that are enabled using the `QUIC_API_ENABLE_PREVIEW_FEATURES` macro, to isolate it from the rest of the established API.
```

Might need confirmation, but I think the C++ api is basically unstable, so maybe we should only mention `msquic.h`

docs/Execution.md
sm-msft: Edited this based on feedback from prior iteration, but I did not readily spot an example of this requirement. I am identifying the missing reference here.

I can add a reference in this PR if someone can point it out. We can schedule a followup to address all TODOs, but I am not sure when that would be...

docs/PreviewFeatures.md
sm-msft: I need a better spell checker than VS.

guhetier: Let me know if you find a good one, I could use that too...

docs/PreviewFeatures.md
sm-msft: I can reduce the scope here to msquic.h and it can be expanded later if needed.

docs/Execution.md
sm-msft: I will leave the TODO in for now to keep this PR limited in scope. We can come back and fill in the details subsequently.

docs/Execution.md
```diff
@@ -33,7 +33,11 @@ QUIC_STATUS
     );
 ```
 
-The application must register a callback handler for every MsQuic object it creates. This handler must manage all the events MsQuic may indicate for that object. The handler must also return a status for each event indicating to MsQuic how the event was handled. This returned status is often success/failure, but sometimes indicates MsQuic that further processing is required.
+The application must register a callback handler for every MsQuic object it creates, as well as for some objects automatically created by MsQuic (TODO: add reference).
```
nibanks, line 36: For reference, you can mention peer initiated streams, or connections (for server side).

docs/Execution.md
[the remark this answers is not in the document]
sm-msft: We have a couple of events that are defined within the "preview" feature macro. I am assuming that these events will always be generated under the right conditions. That would imply callback handler must ignore these events that it cannot correctly interpret. This is what I was referring to in the TODO.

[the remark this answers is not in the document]
guhetier:
```c
    QUIC_CONNECTION_EVENT_RELIABLE_RESET_NEGOTIATED         = 16,   // Only indicated if QUIC_SETTINGS.ReliableResetEnabled is TRUE.
    QUIC_CONNECTION_EVENT_ONE_WAY_DELAY_NEGOTIATED          = 17,   // Only indicated if QUIC_SETTINGS.OneWayDelayEnabled is TRUE.
    QUIC_CONNECTION_EVENT_NETWORK_STATISTICS                = 18,   // Only indicated if QUIC_SETTINGS.EnableNetStatsEvent is TRUE.
#endif
```

Currently, all these events would only be emitted if a user explicitly enabled a setting, that is preview itself, so we should not emit any "unknown" notification.

It is better for a callback handler to ignore what it doesn't know. I am not sure we can rely on implementations doing it though, so adding new notifications without a setting could be a breaking change.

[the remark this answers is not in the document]
sm-msft: I will add a note about returning QUIC_STATUS_SUCCESS for all events not handled by the event handler, to insulate the code from future updates to the API.

guhetier (approved):

Issue #4326, opened 2024-05-29
Misc/WithCidUpdateArgs.CidUpdate/5 faliure


ami-GS:
### Describe the bug

Packet dropped by decryption failure

[0]0C60.1658::‎2024‎-‎05‎-‎29 00:52:27.501 []***
[0]0C60.1658::‎2024‎-‎05‎-‎29 00:52:27.501 [Microsoft-Quic][conn][0x2AB8631B140] New packet keys created successfully. 
[0]0C60.1658::‎2024‎-‎05‎-‎29 00:52:27.501 [Microsoft-Quic][pack][3298534883335] Decrypting 
[0]0C60.1658::‎2024‎-‎05‎-‎29 00:52:27.501 []***
[0]0C60.1658::‎2024‎-‎05‎-‎29 00:52:27.501 [Microsoft-Quic][ lib] ERROR, 0, EVP_DecryptFinal_ex failed. 
[0]0C60.1658::‎2024‎-‎05‎-‎29 00:52:27.501 [Microsoft-Quic][C][RX][254] SH DestCid: KP:1 SB:0 (Payload 1 bytes) 
[0]0C60.1658::‎2024‎-‎05‎-‎29 00:52:27.501 []***
[0]0C60.1658::‎2024‎-‎05‎-‎29 00:52:27.501 [Microsoft-Quic][conn][0x2AB8631B140] DROP packet Dst=[::1%1]:61860 Src=[::1]:61859 Reason=Decryption failure. 

https://github.com/microsoft/msquic/actions/runs/9278086109/job/25528722387?pr=4298

### Affected OS

- [X] Windows
- [ ] Linux
- [ ] macOS
- [ ] Other (specify below)

### MsQuic version

main

### Steps taken to reproduce bug

Not repro locally

### Expected behavior

succeeds

### Actual outcome

D:\a\msquic\msquic\src\test\lib\HandshakeTest.cpp(2675): error: 1 server packets failed to decrypt!

PR #4973, opened 2025-04-03
Avoid Batching of Packets with Different Keys

vgmahajanshetty:
## Description

Fixes the issue #4326.

Update the HpMask when processing batch of packets in function QuicConnRecvDatagramBatch.
If KeyType of packet changes then use new KeyType of current packet (ReadKeys[Packet->KeyType]->HeaderKey) to compute HpMask which is used to decrypt the packets.

## Testing

Ran existing test cases to make sure it is not breaking anything.

This may not fully solve Misc/WithCidUpdateArgs.CidUpdate/5 faliure, but it solves decryption failure issues caused by change in packet header key.

## Documentation

No.

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -5521,12 +5521,13 @@ QuicConnRecvDatagramBatch(
         CxPlatZeroMemory(HpMask, BatchCount * CXPLAT_HP_SAMPLE_LENGTH);
     }
 
-    uint8_t prevPacket = 0;
+    //uint8_t prevPacket = 0;
     for (uint8_t i = 0; i < BatchCount; ++i) {
         CXPLAT_DBG_ASSERT(Packets[i]->Allocated);
         CXPLAT_ECN_TYPE ECN = CXPLAT_ECN_FROM_TOS(Packets[i]->TypeOfService);
         Packet = Packets[i];
 
+#if 0
         if (Packets[prevPacket]->KeyType != Packet->KeyType) {
             if (Packet->Encrypted &&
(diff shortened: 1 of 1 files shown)


src/core/connection.c
```diff
@@ -5521,10 +5521,32 @@ QuicConnRecvDatagramBatch(
         CxPlatZeroMemory(HpMask, BatchCount * CXPLAT_HP_SAMPLE_LENGTH);
     }
 
+    uint8_t prevPacket = 0;
```
nibanks, line 5524: nit: casing:
```suggestion
    uint8_t PrevPacket = 0;
```
Though, I wonder if we should be changing something more fundamental and these shouldn't be considered part of the same batch to begin with.

vgmahajanshetty: Removed PrevPacket.
I feel grouping same KeyType packets may complicate this further.

nibanks: > I feel grouping same KeyType packets may complicate this further.

Perhaps, but I still think it's likely a better design.

vgmahajanshetty: Please check whether current changes are good. I am avoiding batching of packets if KeyType changes.

src/core/connection.c
```diff
@@ -5711,7 +5712,23 @@ QuicConnRecvDatagrams(
                 BatchCount = 0;
             }
 
+            if ((PrevPackKeyType != QUIC_PACKET_KEY_COUNT) &&
```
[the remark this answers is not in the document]
vgmahajanshetty: Done.

nibanks (approved):

PR #4975, opened 2025-04-04
RecvBuffer simplifications [2/N]

guhetier:
## Description

Part of a series of PR to simplify the receive buffer code by unifying the handling of the different buffer modes where possible.

This change:

- introduces an iterator over the receive buffer chunk list. It provides a view to the contiguous spans of data in chunk list, skipping an initial offset
- refactors the read and write operations using the iterator
- introduce a validation function to enforce the receive buffer invariants consistently

## Testing

CI

## Documentation

N/A

### DIFF
--- a/src/core/recv_buffer.c
+++ b/src/core/recv_buffer.c
@@ -109,7 +109,7 @@ QuicRecvBufferGetChunkIterator(
 //
 _IRQL_requires_max_(DISPATCH_LEVEL)
 _Success_(return == TRUE)
-BOOL
+BOOLEAN
 QuicRecvChunkIteratorNext(
     _Inout_ QUIC_RECV_CHUNK_ITERATOR* Iterator,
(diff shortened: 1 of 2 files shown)


src/core/recv_buffer.c
```diff
@@ -76,6 +191,64 @@ QuicRecvChunkFree(
...
+// Validate the receive buffer invariants.
+// No-op in release builds.
+//
+_IRQL_requires_max_(DISPATCH_LEVEL)
+void
+QuicRecvBufferValidate(
+    _In_ const QUIC_RECV_BUFFER* RecvBuffer
+    )
+{
+#if DEBUG
```
nibanks, line 204: Generally, we replace the function with a no-op define in release mode instead of doing this.

nibanks (approved):

nibanks (approved):

PR #4979, opened 2025-04-08
Rust: Fix rust ci build

youyuanwu:
## Description
Rust ci is broken due to compiler version upgrade.
Fixed the clippy error and pinned the rust version in ci using standard setup rust github action.
CI should not be broken in the future like this time without code changes.

## Testing
NA

## Documentation
NA

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,6 @@
+# Specify which version of rust should be used to compile the project
+# cargo + rustup will use this to consistently build the project
+# with the same version across all checkouts and environments
+[toolchain]
+channel = "1.86.0"
+profile = "default"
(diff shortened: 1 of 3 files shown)


rust-toolchain.toml
```diff
@@ -0,0 +1,6 @@
+# Specify which version of rust should be used to compile the project
+# cargo + rustup will use this to consistently build the project
+# with the same version across all checkouts and environments
+[toolchain]
```
anrossi, line 4: Why is this file needed in addition to the "with: toolchain ..." directive in the cargo.yml file?

youyuanwu: CI yaml only deterimines which version of rust to install, and on a machine different versions of rust can be installed. 
This file determines for this repo what rust version to use to compile.
If on the machine, a newer version of rust is preinstalled in the image, the ci can break again without this file, since by default the latest version installed is picked.

nibanks (approved):

PR #4959, opened 2025-03-31
Add Support for openssl v3.5 as a TLS Backend

nhorman:
## Description

This commit enhances msquic to allow it to build using the canonical openssl sources as a TLS backend.

## Testing

All existing tests should cover this code as far as I can see.

I've run this code base against the quic-interop-runner and it passes all supported test cases there (save for zerortt I believe) which fails with all TLS backends at the moment I think, but we have apis to support early data, so we should be able to get that up and running shortly.

All unit tests pass under windows currently, save for WithHandshakeArgs4.RandomLossResumeRejection/4  and WithHandshakeArgs4.RandomLossResumeRejection/5 which were failing scholastically.  I believe I have those fixed by adjusting some of the asserts when built in debug mode which (If I read the code base right) assume that keys are updated in lock step, which the openssl quic callback api doesn't always do, instead updating keys in independent callbacks.
 
## Documentation

Possibly.  currently this PR renames openssl support (which initially pointed to quictls), to actually be named quictls[3] and takes over the QUIC_TLS=openssl configuration to use the canonical openssl sources.  That may require build doc updates to reflect that, but I've not clearly found any (yet).

PR #4985, opened 2025-04-09
Introduce breaking change to inform users of openssl->quictls naming changes

nhorman:
## Description

As per comment [here](https://github.com/microsoft/msquic/pull/4959#issuecomment-2789336329):
* rename openssl[3] submodule to quictls[3]
* Convert cmakefiles to select submodule based on the new quictls[3] names
* Convert rust crate toml and build scripts to honor QUIC_TLS_LIB
* Update top level CMakeLists.txt to error out if QUIC_TLS is set, to catch users using the old config method

## Testing

They shouldn't, just a mass renaming

## Documentation

I can't find any separate docs that explicitly reference QUIC_TLS, so I don't think so.


.github/workflows/build.yml
```diff
@@ -120,15 +120,15 @@ jobs:
         # Android doesn't use system crypto
         - plat: android
           systemcrypto: '-UseSystemOpenSSLCrypto'
-        # No openssl system crypto on ubuntu-22.04
+        # No quictls system crypto on ubuntu-22.04
```
nibanks, line 123: Undo these comments as we still are using openssl for system crypto library.

.github/workflows/test-down-level.yml
```diff
@@ -25,20 +25,20 @@ jobs:
       matrix:
         vec: [
```
nibanks, line 26: This test will be more complicated, because it downloads previously created artifacts that still use the openssl name. But going forward, they will be named quictls.

nhorman: so, whats the solution here?  We can back out the change I guess until such time as the new artifacts get built.

nibanks: Likely what we will have to do is add a `prevTls` parameter to point it to the old artifact name. We can probably make it optional, and update the logic below to use `prevTls` if set, otherwise `tls`.

nhorman: done, I opted to just introduce an imgtls element to each vector in the matrix to avoid having to do github actions yaml conditionals, which always seem hard to read to me, and made the schannel elements a 1:1 mapping.

nibanks:
`./scripts/prepare-machine.ps1` needs to be updated as well.

nhorman:
not sure whats going on with the nugget test failures:
https://github.com/microsoft/msquic/actions/runs/14364168989/job/40274621851

Error indicates that the Release-windows-windows-2022-x86-quictls artifact can't be found, but this page:
https://github.com/microsoft/msquic/actions/runs/14364168989/artifacts/2913929004
Indicates that the artifact was produced in the prior build step

.github/workflows/code-coverage.yml
```diff
@@ -46,8 +46,8 @@ jobs:
       fail-fast: false
       matrix:
         vec: [
-          { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "quictls", build: "-Test" },
-          { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "quictls", build: "-Test", xdp: "-UseXdp" },
+          { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "quictls", imgtls: "openssl", build: "-Test" },
+          { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "quictls", imgtls: "openssl", build: "-Test", xdp: "-UseXdp" },
```
nibanks: This shouldn't be necessary

scripts/templates/APKBUILD.template
```diff
@@ -16,25 +16,25 @@ subpackages="$pkgname-dev $pkgname-doc"
...
 prepare() {
 	default_prepare
 
 	cd "$builddir/submodules"
-	rm -rf clog googletest openssl openssl3 xdp-for-windows
+	rm -rf clog googletest openssl quictls3 xdp-for-windows
 	mv ../../CLOG-*/ clog/
 	mv ../../googletest-*/ googletest/
-	mv ../../openssl-*/ openssl3/
+	mv ../../openssl-*/ quictls3/
```
anrossi, line 30: Won't the source directory be called "quictls3-*" as well?

scripts/templates/APKBUILD.template
```diff
@@ -16,25 +16,25 @@ subpackages="$pkgname-dev $pkgname-doc"
...
+	quictls3-$_quictls3.tar.gz::https://github.com/quictls/openssl/archive/$_quictls3.tar.gz
 	"
 builddir="$srcdir/msquic-SHA_PLACEHOLDER"
 
 prepare() {
 	default_prepare
 
 	cd "$builddir/submodules"
-	rm -rf clog googletest openssl openssl3 xdp-for-windows
+	rm -rf clog googletest openssl quictls3 xdp-for-windows
```
anrossi, line 27: Why is this keeping around openssl? Shouldn't it be "quictls"?

anrossi:
Left a few comments, but generally, it looks good to me!

anrossi:
> not sure whats going on with the nugget test failures: https://github.com/microsoft/msquic/actions/runs/14364168989/job/40274621851
> 
> Error indicates that the Release-windows-windows-2022-x86-quictls artifact can't be found, but this page: https://github.com/microsoft/msquic/actions/runs/14364168989/artifacts/2913929004 Indicates that the artifact was produced in the prior build step

All the listed artifacts available to download in the log are using the old "*-openssl{3}" names. I wonder if the configuration for the artifact upload has an "openssl" string that was missed?

nhorman:
> > not sure whats going on with the nugget test failures: https://github.com/microsoft/msquic/actions/runs/14364168989/job/40274621851
> > Error indicates that the Release-windows-windows-2022-x86-quictls artifact can't be found, but this page: https://github.com/microsoft/msquic/actions/runs/14364168989/artifacts/2913929004 Indicates that the artifact was produced in the prior build step
> 
> All the listed artifacts available to download in the log are using the old "*-openssl{3}" names. I wonder if the configuration for the artifact upload has an "openssl" string that was missed?

I'm struggling to see where that might be occuring, as, for example, in this build run:
https://github.com/microsoft/msquic/actions/runs/14364168989/job/40274346603

I see this:
```
Artifact Release-windows-windows-2022-x64-quictls has been successfully uploaded! Final size is 91049237 bytes. Artifact ID is 2913929004
```
And in the artifact output list, I can see that artifact:
https://github.com/microsoft/msquic/actions/runs/14364168989/artifacts/2913929004

But when I look at the subsequent job to build the Nugget package:
https://github.com/microsoft/msquic/actions/runs/14364168989/job/40274621851

The download Build artifacts steps attempt to get the corresponding artifact name "Release-windows-windows-2022-x64-quictls", but the only thing they find is artifacts with the openssl suffix, which never appear in the artifact list here:
https://github.com/microsoft/msquic/actions/runs/14364168989

Its like the Download artifact step is attempting to search a cache from a prior run that used the old names

nhorman:
I found (at least part) of the problem with the Nugget build job.  Looking here:
https://github.com/microsoft/msquic/actions/runs/14364168989/job/40274621851

Which is the failed job I'm looking at, we see the run id in the URL is 14364168989.  However, looking at the log output for the Download Build Artifacts step, we see that the the download-artifact action that the ci is using prints this:
```
with:
    name: Release-windows-windows-2022-x64-quictls
    ...
==> Repository: microsoft/msquic
==> Artifact name: Release-windows-windows-2022-x64-quictls
==> Local path: artifacts
==> Workflow name: 47794189
==> Workflow conclusion: success
==> Allow forks: false
==> (found) Run ID: 14357338248    <====== This is not the current workflow run id
==> (found) Run date: 2025-04-09T12:38:29Z
==> (not found) Artifact: Release-windows-windows-2022-x64-quictls
```

So the action is attempting to use a much older run id than the one currently being executed.

There appears to be a bug outstanding in the action repo for this event:
https://github.com/dawidd6/action-download-artifact/issues/240

I'm going to try updating the CI to specify the current run id to see if that works around the issue

scripts/templates/APKBUILD.template
nhorman: this actually needs to stay put, because ../../openssl-* is the extracted directory of the archive fetched from this line:
```
quictls3-$_quictls3.tar.gz::https://github.com/quictls/openssl/archive/$_quictls3.tar.gz
```

In the APKBUILD template.  quictls never really broke from using the openssl name in several places, including the commit level archives that this file downloads (i.e. even though we renamed the variable above, the extracted directory path for quictls3-<commit hash>.tar.gz continues to hold the openssl name in its root.

so thats what we have to target when doing the mv operation.

scripts/templates/APKBUILD.template
nhorman: for the same reason as above, this also needs to stay put :(

nibanks:
FYI, I can't kick off CI because of conflicts.

nhorman:
rebased for you @nibanks

nhorman:
Ok, I think we're just about ready to go on this:
Regarding the failing tests:
1) The BVT and Down level tests - not sure what to do about them, as they aren't build failures, but rather incidental test failures, and I'm not sure I see how changing naming as this PR does might be impacting those tests - thoughts welcome here
2) The CIFuzz test is being addressed in https://github.com/google/oss-fuzz/pull/13222
3) The CLOG test still needs to be fixed, but I'm not sure what the process is for doing so, presumably the changes that PR makes (by running a clog update as the CI test does), just needs to be run locally, and the changes comitted, but doing so locally results in a large number of changes that don't appear relevant, so I'm not sure what the path forward is.  Input appreciated here.
4) Codecov - I presume this is the result of a drop in code coverage, but I'm hesitant to make arbitrary changes to increase that again without fully understanding the impact.

nibanks:
I'm not worried about the BVT. It's a known issue. I triggered yet another rerun of the down-level tests. Those are worrisome because they indicate a change in behavior, somehow.

I saw your oss-fuzz PR. LGTM.

The CLOG failure requires running `.\scripts\update-sidecar.ps1` (on Windows) to fix up the sidecar used for LTTng logging.

I'm not worried about the codecov. It's over sensitive.

nhorman:
@nibanks thank you.  Ack on the BVT tests and code coverage tests

regarding CLOG, I'm a bit concerned about doing that, as running the update-sidecar script locally seems to generate a number of errors about clog not being present, and the submodule doesn't build (unsure as to why).  Several changes are created (among them changes to tls_openssl.clog.h), but the changes are far more widespread than that, so I'm hesitant to accept them without further understanding of whats going on.  Advice appreciated.

nibanks:
I'm not sure what's going on for your clog build, but the changes are required to pass this check. The check is there to ensure our logging sidecar stays in sync with the code base.

nhorman:
@nibanks Think I figured the clog issue out.  Seems I had multiple versions of dotnet installed, and the one that was getting picked up either didn't have an associated sdk, or a version of the sdk that clog didn't like.  Installed 6.0.100 and set my path to point to it, and it seems to have made the appropriate changes.

scripts/make-pgo-pr.ps1
```diff
@@ -41,8 +41,8 @@ git checkout main
 git checkout -b $BranchName
 
 Copy-Item -Path artifacts/PerfDataResults/performance/windows/$($Arch)_$($Config)_schannel/msquic.pgd src/bin/winuser/pgo_$($Arch)/msquic.schannel.pgd -Force
-Copy-Item -Path artifacts/PerfDataResults/performance/windows/$($Arch)_$($Config)_openssl/msquic.pgd src/bin/winuser/pgo_$($Arch)/msquic.openssl.pgd -Force
-Copy-Item -Path artifacts/PerfDataResults/performance/windows/$($Arch)_$($Config)_openssl3/msquic.pgd src/bin/winuser/pgo_$($Arch)/msquic.openssl3.pgd -Force
+Copy-Item -Path artifacts/PerfDataResults/performance/windows/$($Arch)_$($Config)_quictls/msquic.pgd src/bin/winuser/pgo_$($Arch)/msquic.quictls.pgd -Force
+Copy-Item -Path artifacts/PerfDataResults/performance/windows/$($Arch)_$($Config)_quictls3/msquic.pgd src/bin/winuser/pgo_$($Arch)/msquic.quictls3.pgd -Force
```
nibanks, line 45: Did these files get renamed too?

submodules/googletest
nibanks, line 1: Please revert this (accidental?) change.

scripts/make-pgo-pr.ps1
nhorman: no, but I will

pkropachev:
@nibanks , @nhorman , guys, is it still relevant use https://github.com/quictls/openssl repo? The repo was archived. Ot there is plan to use new repo:  https://github.com/quictls/quictls?

nibanks:
> @nibanks , @nhorman , guys, is it still relevant use https://github.com/quictls/openssl repo? The repo was archived. Ot there is plan to use new repo: https://github.com/quictls/quictls?

We plan to immediately make follow up changes to:

1. Remove support for quictls[1], and rename quictls3 to quictls.
2. Migrate from quictls/openssl to quictls/quictls.
3. Add support for openssl.

nhorman:
The current cargo failure is confusing me:
```
running: "cmake" "--build" "D:\\a\\msquic\\msquic\\target\\debug\\build\\msquic-7d379929a28ccd7c\\out\\build" "--target" "install" "--config" "Debug" "--parallel" "4"
```
Which seems to be creating a problem in which the MsQuicEtw.rc file is attempting to be read prior to it getting generated (I think).  Whats increasingly odd is that the --parallel 4 option is being passed, despite the fact that in scripts/build.rs  we have this code:
```
// Disable parallel builds on Windows, as they seems to break manifest builds.
    if cfg!(windows) {
        config.define("CMAKE_BUILD_PARALLEL_LEVEL", "1");
    }
```
so I'm lost as to how we're getting back to a parallel build of 4 threads when we've clearly set it to 1 above

nibanks:
> The current cargo failure is confusing me:

It's not specific to your PR. I've been trying to figure out why it keeps breaking, and have failed to fix it so far...

nibanks:
Would you mind merging/rebasing the latest main to get some CI fixes?

nhorman:
rebase done

nhorman:
FYI, the oss-fuzz build is currently failing in PR 
https://github.com/google/oss-fuzz/pull/13222

Due to a compilation error in which CXPLAT_DBG_ASSERT is defined multiple times.  The issue is addressed in msquic PR:
https://github.com/microsoft/msquic/pull/5016

nibanks:
We either need to (temporarily) disable the CIFuzz workflow until it's fixed or wait on it.

nibanks (approved):

nibanks:
Thanks so much @nhorman for doing this! I have disabled the CIFuzz test while we wait for that external PR to get fixed/merged.

nhorman:
awesome, thanks!  I'll move on to phase 2 of our plan this week.

PR #4986, opened 2025-04-09
Change Formatting of QUIC Version Trace

nibanks:
## Description

Log the version in hex instead of decimal.

## Testing

Automation

## Documentation

N/A

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -579,7 +579,7 @@ QuicConnTraceRundownOper(
     if (Connection->Stats.QuicVersion != 0) {
         QuicTraceEvent(
             ConnVersionSet,
-            "[conn][%p] QUIC Version: %u",
+            "[conn][%p] QUIC Version: 0x%x",
             Connection,
             Connection->Stats.QuicVersion);
     }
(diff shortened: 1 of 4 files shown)


src/core/connection.c
```diff
@@ -3412,19 +3412,17 @@ QuicConnRecvVerNeg(
     // requested version. If it does, we are supposed to ignore it. Cache the
     // first supported version.
     //
-    QuicTraceLogConnVerbose(
-        RecvVerNeg,
-        Connection,
-        "Received Version Negotation:");
+    QuicTraceLogVerbose(
+        PacketRxVersionNegotiation,
+        "[C][TX][-] VN");
```
guhetier, line 3417: IMO, keeping `Version Negotiation` instead of `VN` would make logs easier to understand.

nibanks: The logs on the client receive now match the logs on the server send.

nibanks: But this should be `RX`

guhetier: I'd argue we should update the server side log to make it easier to understand too then, but I won't block the PR on that.

guhetier (approved):

src/core/connection.c
nibanks: We have a standard format for packet logging. Take a look at packet.c and frame.c. This was the one exception to that format.

PR #4987, opened 2025-04-11
Add ConnectionOpenInPartition

nibanks:
## Description

Adds the new API `ConnectionOpenInPartition` that adds a new parameter that allows the application to specify the partition index explicitly so we don't need to infer/guess based on the current processor it's running on.

## Testing

Updated, but the old function goes through the new one, so really it is completely covered already.

## Documentation

Updated/added.

### DIFF
--- a/docs/api/ConnectionOpen.md
+++ b/docs/api/ConnectionOpen.md
@@ -52,7 +52,7 @@ Once the connection has been shutdown, it must be cleaned up with a call to [Con
 
 # See Also
 
-[ConnectionOpenEx1](ConnectionOpenEx1.md)<br>
+[ConnectionOpenInPartition](ConnectionOpenInPartition.md)<br>
 [ConnectionClose](ConnectionClose.md)<br>
 [ConnectionShutdown](ConnectionShutdown.md)<br>
 [ConnectionStart](ConnectionStart.md)<br>
--- a/docs/api/ConnectionOpenInPartition.md
+++ b/docs/api/ConnectionOpenInPartition.md
@@ -47,13 +47,9 @@ The function returns a [QUIC_STATUS](QUIC_STATUS.md). The app may use `QUIC_FAIL
 
 # Remarks
 
-`ConnectionOpenInPartition` is used to create a connection in the client application, with an explicit partition index. In server applications, [ListenerOpen](ListenerOpen.md) and [ListenerStart](ListenerStart.md) must be called to listen for incoming connection attempts, and the server side Connection is created in the `QUIC_LISTENER_EVENT_NEW_CONNECTION` event.
+See [ConnectionOpen](ConnectionOpen.md#Remarks) for more remarks.
 
-'ConnectionOpen' only allocates the resources for the connection, it does not start the connection. To start the connect, the application must call [ConnectionStart](ConnectionStart.md).
-
-Once `ConnectionOpen` completes successfully, the application may create streams, and queue data for sending. This is when 0-RTT streams and data **MUST** be created and queued. See [StreamOpen](StreamOpen.md), and [StreamStart](StreamStart.md).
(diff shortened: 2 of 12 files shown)


docs/api/ConnectionOpen.md
```diff
@@ -52,6 +52,7 @@ Once the connection has been shutdown, it must be cleaned up with a call to [Con
 
 # See Also
 
+[ConnectionOpenEx1](ConnectionOpenEx1.md)<br>
```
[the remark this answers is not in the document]
guhetier: I agree the partern `Ex1` is not something I've seen.

I would suggest `ConnectionOpenWithPartition` though. `Ex` is not helpful to understand the difference.

mtfriesen (dismissed):

docs/api/ConnectionOpenInPartition.md
```diff
@@ -0,0 +1,65 @@
...
+
+# Return Value
+
+The function returns a [QUIC_STATUS](QUIC_STATUS.md). The app may use `QUIC_FAILED` or `QUIC_SUCCEEDED` to determine if the function failed or succeeded.
+
+# Remarks
+
+`ConnectionOpenInPartition` is used to create a connection in the client application, with an explicit partition index. In server applications, [ListenerOpen](ListenerOpen.md) and [ListenerStart](ListenerStart.md) must be called to listen for incoming connection attempts, and the server side Connection is created in the `QUIC_LISTENER_EVENT_NEW_CONNECTION` event.
+
+'ConnectionOpen' only allocates the resources for the connection, it does not start the connection. To start the connect, the application must call [ConnectionStart](ConnectionStart.md).
```
guhetier, line 52: Should `ConnectionOpen` be replaced throughout this document?
Otherwise, we could redirect to the `ConnectionOpen` documentation and only mention the differences here.

nibanks: Updated

src/test/lib/ApiTest.cpp
```diff
@@ -692,6 +692,22 @@ void QuicTestValidateConnection()
                 &Connection.Handle));
     }
 
+    //
```
guhetier, line 695: Should we have a test for the success case just in case? It's partially covered by the existing connection tests, but only when using the current partition index, not for a different one.

nibanks: As I mentioned in the description, I don't feel it's necessary at this time. ConnectionOpen already hits every code path.

anrossi:
Instead of calling `QuicConnAlloc` directly, should the connection pool move to calling this function instead?

nibanks:
> Instead of calling `QuicConnAlloc` directly, should the connection pool move to calling this function instead?

No. It doesn't need to change.

src/inc/msquic.h
```diff
@@ -1714,6 +1725,9 @@ typedef struct QUIC_API_TABLE {
     QUIC_CONNECTION_COMP_RESUMPTION_FN  ConnectionResumptionTicketValidationComplete; // Available from v2.2
     QUIC_CONNECTION_COMP_CERT_FN        ConnectionCertificateValidationComplete;      // Available from v2.2
 
+    QUIC_CONNECTION_OPEN_IN_PARTITION_FN
```
guhetier, line 1728: Should we make this preview, even if only for a short time?

csujedihy, line 1728: This is added in the middle of the struct? What happens if an old client gets the API table?

nibanks: It's only before preview stuff. Nothing below has shipped in a release.

nibanks: As discussed over IM, I've decided that I don't want to.

PR #4993, opened 2025-04-12
Fix a race condition in TCP epoll code

csujedihy:
## Description

It's a race between `PerfClientWorker::WorkerThread` and `TcpWorker::WorkerThread`. In the middle of creating a socket from the perf client worker thread, as long as epoll has been told to monitor the socket, a notification could come up at the same time from tcp worker thread and frees the connection.

```
#1  0x00005b7e577eec9f in CxPlatSocketContextSetEvents
#2  0x00005b7e577ef725 in CxPlatSocketCreateTcpInternal
#3  0x00005b7e577efb8a in SocketCreateTcp
#4  0x00005b7e577ebdb9 in CxPlatSocketCreateTcp
#5  0x00005b7e577db274 in TcpConnection::Start
#6  0x00005b7e577def35 in PerfClientConnection::Initialize
#7  0x00005b7e577df124 in PerfClientWorker::StartNewConnection
#8  0x00005b7e577df140 in PerfClientWorker::WorkerThread
```

The fix is moving the actual connection start to the tcp worker thread.

## Testing

CI

### DIFF
--- a/src/perf/lib/Tcp.cpp
+++ b/src/perf/lib/Tcp.cpp
@@ -404,6 +404,7 @@ TcpConnection::TcpConnection(
         WriteOutput("SecConfig load FAILED\n");
         return;
     }
+    QuicAddrSetFamily(&Route.LocalAddress, QUIC_ADDRESS_FAMILY_UNSPEC);
     Initialized = true;
 }
 
@@ -422,7 +423,9 @@ TcpConnection::Start(
     }
     if (LocalAddress) {
         Family = QuicAddrGetFamily(LocalAddress);
+        Route.LocalAddress = *LocalAddress;
     }
+
     if (RemoteAddress) {
         Route.RemoteAddress = *RemoteAddress;
     } else {
@@ -437,15 +440,7 @@ TcpConnection::Start(
         }
     }
     QuicAddrSetPort(&Route.RemoteAddress, ServerPort);
-    if (QUIC_FAILED(
-        CxPlatSocketCreateTcp(
-            Datapath,
-            LocalAddress,
-            &Route.RemoteAddress,
-            this,
-            &Socket))) {
-        return false;
-    }
+    ConnStartQueued = true;
     Queue();
     return true;
 }
@@ -627,6 +622,19 @@ TcpConnection::TlsReceiveTicketCallback(
 
 void TcpConnection::Process()
 {
+    if (ConnStartQueued) {
+        ConnStartQueued = false;
+        if (QUIC_FAILED(
+            CxPlatSocketCreateTcp(
+                Datapath,
+                QuicAddrGetFamily(&Route.LocalAddress) == QUIC_ADDRESS_FAMILY_UNSPEC ?
+                    nullptr : &Route.LocalAddress,
+                &Route.RemoteAddress,
+                this,
+                &Socket))) {
+            Shutdown = true;
+        }
+    }
     if (IndicateAccept) {
         IndicateAccept = false;
         TcpServer* Server = (TcpServer*)Context;
--- a/src/perf/lib/Tcp.h
+++ b/src/perf/lib/Tcp.h
@@ -183,6 +183,7 @@ class TcpConnection {
     bool Closed{false};
     bool QueuedOnWorker{false};
     bool StartTls{false};
+    bool ConnStartQueued{false};
     bool IndicateAccept{false};
     bool IndicateConnect{false};
     bool IndicateSendComplete{false};


guhetier (approved):

guhetier (dismissed):

src/platform/platform_posix.c
```diff
@@ -98,6 +101,17 @@ CxPlatSystemLoad(
     void
     )
 {
+    //
+    // There are a few cases where we might call CxPlatSystemLoad more than
+    // once. Let's guard against that. The caller is responsible for ensuring
+    // it's not called concurrently.
```
nibanks, line 107: Where? No other platforms protect against this, and I personally think it's a bug in the caller and not something we should protect against here (best case assert instead).

src/platform/datapath_epoll.c
```diff
@@ -18,6 +18,7 @@
 #include <linux/filter.h>
 #include <linux/in6.h>
 #include <netinet/udp.h>
+#include <stdatomic.h>
```
nibanks, line 21: nit: this is already included from platform_internal.h so it's unnecessary here too.

src/platform/datapath_epoll.c
```diff
@@ -1175,7 +1176,7 @@ CxPlatSocketContextUninitialize(
     SocketContext->Uninitialized = TRUE;
 #endif
 
-    if (!SocketContext->IoStarted) {
+    if (!atomic_load_explicit(&SocketContext->IoStarted, memory_order_acquire)) {
```
nibanks, line 1179: Are you sure we need these? If so, what do you think about making a platform abstraction in quic_platform_posix.h. I suspect this wouldn't be the only place we could/should do this across the code base.

src/platform/platform_posix.c
csujedihy: In secnetperf, it's called twice. Once before we allocate msquic object. Once inside the object allocation.

nibanks: I assume you're using static linking of msquc?

src/platform/datapath_epoll.c
csujedihy: Theoretically yes to make read and write before visible in other processor before seeing setting iostarted to true.

Maybe I should add an abstraction and use those __sync_fetch_xxx instead. In the future, if stdatomics is fully supported on msvc, we remove all the abstraction and use the built-in ones that come with c11.

src/platform/platform_posix.c
csujedihy: Oh, haha, you are right. But, it wouldn't to make this change to make things easier for static build, right? The binaries built not seeking the so file in the same folder was the reason I use static build.

csujedihy:
Thinking about the race more, I think the fix only makes the race hard to happen.

The race can happen because creating the socket isn't on the TcpWorker thread. It's actually a race between `PerfClientWorker::WorkerThread` and `TcpWorker::WorkerThread`. If we create the socket on TcpWorker::WorkerThread, we would have serialized everything and freeing a socket could only happen after we are completely done with creating it.

```
#1  0x00005b7e577eec9f in CxPlatSocketContextSetEvents
#2  0x00005b7e577ef725 in CxPlatSocketCreateTcpInternal
#3  0x00005b7e577efb8a in SocketCreateTcp
#4  0x00005b7e577ebdb9 in CxPlatSocketCreateTcp
#5  0x00005b7e577db274 in TcpConnection::Start
#6  0x00005b7e577def35 in PerfClientConnection::Initialize
#7  0x00005b7e577df124 in PerfClientWorker::StartNewConnection
#8  0x00005b7e577df140 in PerfClientWorker::WorkerThread
```

csujedihy:
> Thinking about the race more, I think the fix only makes the race hard to happen.
> 
> The race can happen because creating the socket isn't on the TcpWorker thread. It's actually a race between `PerfClientWorker::WorkerThread` and `TcpWorker::WorkerThread`. If we create the socket on TcpWorker::WorkerThread, we would have serialized everything and freeing a socket could only happen after we are completely done with creating it.
> 
> ```
> #1  0x00005b7e577eec9f in CxPlatSocketContextSetEvents
> #2  0x00005b7e577ef725 in CxPlatSocketCreateTcpInternal
> #3  0x00005b7e577efb8a in SocketCreateTcp
> #4  0x00005b7e577ebdb9 in CxPlatSocketCreateTcp
> #5  0x00005b7e577db274 in TcpConnection::Start
> #6  0x00005b7e577def35 in PerfClientConnection::Initialize
> #7  0x00005b7e577df124 in PerfClientWorker::StartNewConnection
> #8  0x00005b7e577df140 in PerfClientWorker::WorkerThread
> ```

This PR will be repurposed to fix this particular race.

submodules/googletest
nibanks, line 1: Please revert this change.

nibanks:
With this PR, the Azure epoll netperf tests for TCP HPS all ran successfully:
```
==============================
  RUN 1:
  Result: 2826 HPS
  ==============================
  RUN 2:
  Result: 1187 HPS
  ==============================
  RUN 3:
  Result: 711 HPS
```
FYI, compared to QUIC results:
```
  ==============================
  RUN 1:
  Result: 2924 HPS
  ==============================
  RUN 2:
  Result: 2871 HPS
  ==============================
  RUN 3:
  Result: 2920 HPS
```

csujedihy:
> With this PR, the Azure epoll netperf tests for TCP HPS all ran successfully:
> 
> ```
>   ==============================
>   RUN 1:
>   Result: 2826 HPS
>   ==============================
>   RUN 2:
>   Result: 1187 HPS
>   ==============================
>   RUN 3:
>   Result: 711 HPS
> ```
> 
> FYI, compared to QUIC results:
> 
> ```
>   ==============================
>   RUN 1:
>   Result: 2924 HPS
>   ==============================
>   RUN 2:
>   Result: 2871 HPS
>   ==============================
>   RUN 3:
>   Result: 2920 HPS
> ```

good to know. I will revert that submodule change.

nibanks (approved):

PR #4998, opened 2025-04-14
Fix non-Preview Builds

nibanks:
## Description

Internal builds are built without preview features enabled externally (i.e. in tests). Those were broken recently. This fixes them.

## Testing

CI/CD

## Documentation

N/A

### DIFF
--- a/.github/workflows/build-reuse-win.yml
+++ b/.github/workflows/build-reuse-win.yml
@@ -56,6 +56,10 @@ on:
         required: false
         default: ''
         type: string
+      official:
+        required: false
+        default: ''
+        type: string
       sanitize:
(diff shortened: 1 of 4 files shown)


.github/workflows/build-reuse-win.yml
```diff
@@ -56,6 +56,10 @@ on:
         required: false
         default: ''
         type: string
+      official:
```
mtfriesen, line 59: I still find "official" to be a confusing term, especially in GitHub, which should never build production/shipping/signed bits.

mtfriesen (dismissed):

mtfriesen (approved):

PR #4999, opened 2025-04-14
Try An Environmental Variable to Disable Rust Parallel Builds on Windows

nibanks:
## Description

Another attempt at fixing the parallel build failure with Rust.

## Testing

CI/CD

## Documentation

N/A

### DIFF
--- a/scripts/build.rs
+++ b/scripts/build.rs
@@ -43,7 +43,7 @@ fn cmake_build() {
 
     // Disable parallel builds on Windows, as they seems to break manifest builds.
     if cfg!(windows) {
-        config.define("CMAKE_BUILD_PARALLEL_LEVEL", "1");
+        env::set_var("CMAKE_BUILD_PARALLEL_LEVEL", "1");
     }
 
     // By default enable schannel on windows, unless openssl feature is selected.


scripts/build.rs
```diff
@@ -43,7 +43,7 @@ fn cmake_build() {
 
     // Disable parallel builds on Windows, as they seems to break manifest builds.
     if cfg!(windows) {
-        config.define("CMAKE_BUILD_PARALLEL_LEVEL", "1");
+        env::set_var("CMAKE_BUILD_PARALLEL_LEVEL", "1");
```
guhetier, line 46: Shouldn't this be set in `cmake::env`?
It is possible (not sure) that the cmake environment variable default to the system environment variables, but conceptually, I think they are different.

- https://docs.rs/cmake/latest/cmake/struct.Config.html#method.env
- https://cmake.org/cmake/help/latest/envvar/CMAKE_BUILD_PARALLEL_LEVEL.html
- https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#cmake-language-environment-variables

guhetier:
```suggestion
        config.env("CMAKE_BUILD_PARALLEL_LEVEL", "1");
```

nibanks: I can try it. The build succeeded with what I have. Let's see if does with this change...

nibanks: This didn't work.

guhetier (approved):

PR #5002, opened 2025-04-15
Print out when test cases are skipped

anrossi:
## Description

Print out when a test case is skipped so it's possible to tell the difference from tests that ran and passed.

## Testing

N/A

## Documentation

N/A

### DIFF
--- a/scripts/run-gtest.ps1
+++ b/scripts/run-gtest.ps1
@@ -663,7 +663,7 @@ function Wait-TestCase($TestCase) {
                 if ($StdOutTxt) { Write-Host $StdOutTxt }
                 if ($StdErrorTxt) { Write-Host $StdErrorTxt }
             } elseif($AnyTestSkipped) {
-                Log "SKIPPED $($TestCase.Name) (in $($Delta.TotalSeconds) sec)"
+                Log "$($TestCase.Name) skipped"
             } else {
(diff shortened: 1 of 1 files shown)


scripts/run-gtest.ps1
```diff
@@ -660,6 +662,8 @@ function Wait-TestCase($TestCase) {
                 LogErr "$($TestCase.Name) failed (in $($Delta.TotalSeconds) sec):"
                 if ($StdOutTxt) { Write-Host $StdOutTxt }
                 if ($StdErrorTxt) { Write-Host $StdErrorTxt }
+            } elseif($AnyTestSkipped) {
+                Log "SKIPPED $($TestCase.Name) (in $($Delta.TotalSeconds) sec)"
```
nibanks, line 666:
```suggestion
                Log "$($TestCase.Name) skipped"
```
Let's follow the same format and keep test name first. I also don't see a reason to log the time.

anrossi: I rearranged it to make the skipped tests stand out in the thousands of tests that we run. The timestamp is of low utility, so I can remove it

nibanks: Bottom line, I prefer the consistency.

nibanks (approved):

PR #5001, opened 2025-04-14
Fix secnetperf calling some cxplat functions twice

csujedihy:
## Description

In secnetperf, we call CxPlatSystemLoad and CxPlatInitialize before calling into msquic open. When statically linking to msquic, CxPlatSystemLoad and CxPlatInitialize access the set of global variables in memory. This leads to double free.

Add a safety guard to prevent it from happening.

## Testing

CI

PR #5004, opened 2025-04-15
Fix a double free when secnetperf statically linking to msquic

csujedihy:
## Description

We hit a double free because we call CxPlatInitialize and CxPlatSystemLoad twice in secnetperf, which work on the same set of memory when statically linked to msquic.

The fix is instantiating msquic for static builds in the main function of secnetperf and let msquic handle ref counting.

## Testing

CI

### DIFF
--- a/src/tools/spin/spinquic.cpp
+++ b/src/tools/spin/spinquic.cpp
@@ -202,9 +202,7 @@ public:
 };
 
 static QUIC_API_TABLE MsQuic;
-//
-// This lock is used to protect concurrent MsQuic library open and close.
-//
+// This locks MsQuicOpen2 in RunThread when statically linked with libmsquic
 CXPLAT_LOCK RunThreadLock;
 
 const uint32_t MaxBufferSizes[] = { 0, 1, 2, 32, 50, 256, 500, 1000, 1024, 1400, 5000, 10000, 64000, 10000000 };
@@ -245,10 +243,7 @@ struct SpinQuicGlobals {
 #ifndef FUZZING
             DumpMsQuicPerfCounters(MsQuic);
 #endif
-
-            CxPlatLockAcquire(&RunThreadLock);
             MsQuicClose(MsQuic);
-            CxPlatLockRelease(&RunThreadLock);
         }
         delete [] SendBuffer;
     }
@@ -1369,9 +1364,13 @@ CXPLAT_THREAD_CALLBACK(RunThread, Context)
     do {
         Gbs Gb;
 
+#ifdef QUIC_BUILD_STATIC
         CxPlatLockAcquire(&RunThreadLock);
         QUIC_STATUS Status = MsQuicOpen2(&Gb.MsQuic);
         CxPlatLockRelease(&RunThreadLock);
+#else
+        QUIC_STATUS Status = MsQuicOpen2(&Gb.MsQuic);
+#endif
         if (QUIC_FAILED(Status)) {
             break;
         }
--- a/src/perf/bin/appmain.cpp
+++ b/src/perf/bin/appmain.cpp
@@ -290,9 +290,6 @@ main(
     QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
     QUIC_CREDENTIAL_CONFIG* SelfSignedCredConfig = nullptr;
     uint8_t CipherSuite = 0;
-    const char* DriverName = nullptr;
-    bool PrivateTestLibrary = false;
-    const char* FileName = nullptr;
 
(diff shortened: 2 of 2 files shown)


csujedihy:
I hope this can address #5001.

src/tools/spin/spinquic.cpp
```diff
@@ -243,7 +243,14 @@ struct SpinQuicGlobals {
 #ifndef FUZZING
             DumpMsQuicPerfCounters(MsQuic);
 #endif
+
+#ifdef QUIC_BUILD_STATIC
```
guhetier, line 247: Not an expert on static / dynamic linking, but a quick search says that for linux, dynamic linking also share global variables.
Should we simply always use the lock?

A comment to explain this would be useful, this isn't an obvious pattern IMO.

csujedihy: You beat me to it. I was reasoning about this part too as it didn't make sense to me.

csujedihy: @nibanks do you have any context? It seems wrong but I feel like changes like this were added for a reason.

src/tools/spin/spinquic.cpp
```diff
@@ -202,7 +202,9 @@ class SpinQuicWatchdog {
 };
 
 static QUIC_API_TABLE MsQuic;
-// This locks MsQuicOpen2 in RunThread when statically linked with libmsquic
+//
+// This lock is used to protect concurrent MsQuic library open and close.
+//
 CXPLAT_LOCK RunThreadLock;
```
ProjectsByJackHe, line 208: Was the intention of `RunThreadLock` always to protect concurrent accesses to MsQuic library open/close?
If so, maybe a better name would be "MsQuicLibLock" or something?

csujedihy: Yeah, that's a better name. Let me see if this PR fixes the problem first.

csujedihy:
Hmm, we are still hitting the assert. How the heck is that possible?

nibanks:
> Hmm, we are still hitting the assert. How the heck is that possible?

I'm equally perplexed. I don't see how your PR yesterday would be causing this to hit so often either...

csujedihy:
> > Hmm, we are still hitting the assert. How the heck is that possible?
> 
> I'm equally perplexed. I don't see how your PR yesterday would be causing this to hit so often either...

I don't know if this is relevant to the assert being hit but the use of ref count is problematic in CxPlatInitialize. If we somehow bail in this function, we don't decrement the ref count. Subsequent calls to CxPlatInitialize will be skipped.

csujedihy:
It's also weird that we set CxPlatform.Heap in both CxPlatInitialize and CxPlatSystemLoad.

csujedihy:
I am trying a different approach to fix the original static build problem by just instantiating msquic instead of calling those plat and system init function directly. Let me know what you think.

csujedihy:
> I am trying a different approach to fix the original static build problem by just instantiating msquic instead of calling those plat and system init function directly. Let me know what you think.

Oh crap, this won't work if msquic is not statically linked.

src/tools/spin/spinquic.cpp
nibanks: I think the intention was to protect it only if doing static link (because dynamic link MUST support parallel callers in the same process). I'm fine with changing the name though.

src/tools/spin/spinquic.cpp
```diff
@@ -1364,13 +1369,9 @@ CXPLAT_THREAD_CALLBACK(RunThread, Context)
     do {
         Gbs Gb;
 
-#ifdef QUIC_BUILD_STATIC
```
nibanks, line 1367: We don't want to use a lock with dynamic linking, because it must support parallel callers. We can't crash if two things call into the dll/so in the same process.

src/tools/spin/spinquic.cpp
nibanks: As I mentioned in the other comments, this was done on purpose. Dynamic linking MUST support parallel callers without synchronization from the callers.

src/tools/spin/spinquic.cpp
csujedihy: How things are different for static linking in this case? If we support parallel callers to open and close msquic for dynamic linking, static linked msquic should work too, right?

src/tools/spin/spinquic.cpp
guhetier: Can you explain why? I think we are all confused here about how it would be different between static and dynamic linking.

nibanks (approved):

PR #5011, opened 2025-04-16
disable stub log event check

mtfriesen:
## Description

Linux perf shows QuicConnLogOutFlowStats is a hot spot on Linux, which is because a slow logging path is unconditionally enabled with the "stub" logger. Instead, treat the logs as always disabled when optimizing hot paths.

## Testing

CI.

## Documentation

No.


src/inc/quic_trace.h
```diff
@@ -180,7 +180,12 @@ clog_stdout(struct clog_param * head, const char * format, ...)
 #endif
 
 #if defined(QUIC_EVENTS_STDOUT) || defined(QUIC_LOGS_STUB)
+
```
nibanks, line 183: Let's just do FALSE for both

PR #5015, opened 2025-04-17
Refactor TCP Security Config for Reuse on Client Connections

nibanks:
## Description

By reusing the security config for TCP, this should significiantly improve performance on the client side.

## Testing

CI/CD

## Documentation

N/A

### DIFF
--- a/src/perf/lib/Tcp.cpp
+++ b/src/perf/lib/Tcp.cpp
@@ -49,44 +49,55 @@ const uint8_t FixedIv[CXPLAT_MAX_IV_LENGTH] = { 0 };
 
 const QUIC_HKDF_LABELS TcpHkdfLabels = { "tcp key", "tcp iv", "tcp hp", "tcp ku" };
 
-struct LoadSecConfigHelper {
-    LoadSecConfigHelper() : SecConfig(nullptr) { CxPlatEventInitialize(&CallbackEvent, TRUE, FALSE); }
-    ~LoadSecConfigHelper() { CxPlatEventUninitialize(CallbackEvent); }
-    CXPLAT_SEC_CONFIG* Load(const QUIC_CREDENTIAL_CONFIG* CredConfig) {
-        if (QUIC_FAILED(
-            CxPlatTlsSecConfigCreate(
-                CredConfig,
-                CXPLAT_TLS_CREDENTIAL_FLAG_NONE,
-                &TcpEngine::TlsCallbacks,
-                this,
-                SecConfigCallback))) {
-            return nullptr;
-        }
-        CxPlatEventWaitForever(CallbackEvent);
-        return SecConfig;
+TcpSecConfig::TcpSecConfig() noexcept
+{
+    CxPlatEventInitialize(&CallbackEvent, TRUE, FALSE);
+}
+
+TcpSecConfig::TcpSecConfig(_In_ const QUIC_CREDENTIAL_CONFIG* CredConfig) noexcept
+{
+    CxPlatEventInitialize(&CallbackEvent, TRUE, FALSE);
+    if (!Load(CredConfig)) {
+        CXPLAT_DBG_ASSERT(false);
     }
-private:
-    static
-    _IRQL_requires_max_(PASSIVE_LEVEL)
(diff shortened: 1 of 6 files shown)


src/perf/lib/Tcp.h
```diff
@@ -135,23 +135,33 @@ class TcpWorker {
...
+        _In_ const QUIC_CREDENTIAL_CONFIG* /* CredConfig */,
         _In_opt_ void* Context,
         _In_ QUIC_STATUS Status,
         _In_opt_ CXPLAT_SEC_CONFIG* SecurityConfig
         );
+public:
+    CXPLAT_SEC_CONFIG* SecConfig{nullptr};
+    TcpSecConfig() noexcept;
+    TcpSecConfig(const QUIC_CREDENTIAL_CONFIG* CredConfig) noexcept;
+    ~TcpSecConfig() noexcept;
```
guhetier, line 155: Please delete or implement the copy and move constructors if you add a destructor (rule of 5)

src/perf/lib/Tcp.cpp
```diff
@@ -49,44 +49,55 @@ const uint8_t FixedIv[CXPLAT_MAX_IV_LENGTH] = { 0 };
...
-                this,
-                SecConfigCallback))) {
-            return nullptr;
-        }
-        CxPlatEventWaitForever(CallbackEvent);
-        return SecConfig;
-    }
-private:
-    static
-    _IRQL_requires_max_(PASSIVE_LEVEL)
```
guhetier, line 70: Aren't the SAL annotations supposed to stay in sync between the declaration and definition?

src/perf/lib/Tcp.h
nibanks: We don't do that for anything else, so I prefer to not do it here, especially since it's never going to be used for this one.

guhetier: Well, not doing it on anything else is a problem too, but I can't really ask you to fix them all at once...
If the copy/move are not going to be used, deleting them enforces it. It can avoid hard to debug double free / memory corruptions letter for the price of adding 4 lines right now.

```
TcpSecConfig(const TcpSecConfig&) = delete;
TcpSecConfig(TcpSecConfig&&) = delete;
TcpSecConfig& operator=(const TcpSecConfig&) = delete;
TcpSecConfig& operator=(TcpSecConfig&&) = delete;
```

nibanks: How about this: I will do a follow up PR to delete all these for the TCP C++ classes immediately after this is merged?

src/perf/lib/Tcp.cpp
```diff
@@ -762,7 +751,7 @@ bool TcpConnection::ProcessTls(const uint8_t* Buffer, uint32_t BufferLength)
             &BufferLength,
             &TlsState);
     if (Results & CXPLAT_TLS_RESULT_ERROR) {
-        WriteOutput("CxPlatTlsProcessData FAILED\n");
+        //WriteOutput("CxPlatTlsProcessData FAILED, Alert=0x%hx\n", TlsState.AlertCode);
```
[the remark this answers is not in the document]
nibanks: I'd rather keep for now, and might look into more later.

src/perf/lib/Tcp.cpp
```diff
@@ -384,9 +381,9 @@ TcpServer::AcceptCallback(
 
 TcpConnection::TcpConnection(
     TcpEngine* Engine,
-    const QUIC_CREDENTIAL_CONFIG* CredConfig,
+    TcpConfiguration* Config,
```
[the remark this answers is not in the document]
nibanks: Sure. Will fix.

src/platform/datapath_epoll.c
```diff
@@ -1740,7 +1740,7 @@ CxPlatSocketHandleErrors(
             SocketContext->Binding,
             errno,
             "getsockopt(SO_ERROR) failed");
-    } else {
+    } else if (ErrNum != 0) {
```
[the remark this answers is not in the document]
nibanks: I think it's possible to get an ErrNum 0, but not positive.

csujedihy (approved):

guhetier (approved):

PR #5023, opened 2025-04-18
Sum Throughput Rates Across Streams/Connections

nibanks:
## Description

Updates the output for throughput prints to sum the rates across all the streams/connections that were run. Makes it much easier to understand total throughput with multi-connection tests.

## Testing

Locally. CI/CD

## Documentation

N/A

### DIFF
--- a/src/perf/lib/PerfClient.cpp
+++ b/src/perf/lib/PerfClient.cpp
@@ -150,6 +150,7 @@ PerfClient::Init(
     TryGetValue(argc, argv, "pacing", &UsePacing);
     TryGetValue(argc, argv, "sendbuf", &UseSendBuffering);
     TryGetValue(argc, argv, "ptput", &PrintThroughput);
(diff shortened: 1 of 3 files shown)


mtfriesen (approved):

src/perf/lib/PerfClient.cpp
```diff
@@ -381,6 +382,16 @@ PerfClient::Wait(
             unsigned long long RPS = CompletedStreams * 1000 * 1000 / RunTime;
             WriteOutput("Result: %llu RPS\n", RPS);
         }
+    } else if (PrintThroughput) {
+        unsigned long long UploadRate = GetUploadRate();
```
ProjectsByJackHe, line 386: nit: Does using uint64_t work here?

nibanks: Nope. Linux build breaks.

ProjectsByJackHe (approved):

PR #5030, opened 2025-04-19
Expose TLS Group through MsQuic API

nibanks:
## Description

The TLS group is useful to understand exactly what level of security is being negotiated during the TLS handshake. For instance, this allows perf tests to ensure they are apples to apples.

On a side note, OpenSSL negotiates weaker/simpler/less-CPU-intensive groups by default which is why their HPS numbers were so high. A follow up PR needs to fix that.

## Testing

Local testing

## Documentation

Added some minimal docs.


src/inc/msquic.h
```diff
@@ -430,6 +430,23 @@ typedef enum QUIC_KEY_EXCHANGE_ALGORITHM {
     QUIC_KEY_EXCHANGE_ALGORITHM_NONE  = 0,
 } QUIC_KEY_EXCHANGE_ALGORITHM;
 
+//
+// See the following IANA registry for the TLS groups:
+//   https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8
+//
+typedef enum QUIC_TLS_GROUP {
+    QUIC_TLS_GROUP_NONE                 = 0,
```
nibanks, line 438: Wonder if I should call this `_UNKNOWN` instead...

src/platform/tls_schannel.c
```diff
@@ -3058,7 +3058,7 @@ CxPlatTlsParamGet(
             break;
 
         case QUIC_PARAM_TLS_HANDSHAKE_INFO: {
-            if (*BufferLength < sizeof(QUIC_HANDSHAKE_INFO)) {
+            if (*BufferLength < CXPLAT_STRUCT_SIZE_THRU_FIELD(QUIC_HANDSHAKE_INFO, CipherSuite)) {
```
anrossi, line 3061: Won't this need to be updated every time the `QUIC_HANDSHAKE_INFO` is changed?
Why is sizeof insufficient? To prevent failure of older applications using a newer MsQuic.dll passing the old struct size?

anrossi: Ah yeah, that's exactly it. and no it doesn't need to be updated. I see it now.

anrossi (approved):

PR #5031, opened 2025-04-20
Remove Support for quictls v1.1

nibanks:
## Description

Since quictls/openssl v1.1 is no longer supported, remove support from msquic.

## Testing

CI/CD

## Documentation

Updated


nibanks:
@nhorman perhaps you could help me with this weird build issue we're seeing with Rust:

https://github.com/microsoft/msquic/actions/runs/14577495817/job/40886758174?pr=5031#step:16:29

```
Caused by:
  Source directory was modified by build.rs during cargo publish. Build scripts should not modify anything outside of OUT_DIR.
  Added: /Users/runner/work/msquic/msquic/target/package/msquic-2.5.0-beta/submodules/quictls/doc/html
  	/Users/runner/work/msquic/msquic/target/package/msquic-2.5.0-beta/submodules/quictls/doc/html/man1
  	/Users/runner/work/msquic/msquic/target/package/msquic-2.5.0-beta/submodules/quictls/doc/html/man3
  	/Users/runner/work/msquic/msquic/target/package/msquic-2.5.0-beta/submodules/quictls/doc/html/man5
  	/Users/runner/work/msquic/msquic/target/package/msquic-2.5.0-beta/submodules/quictls/doc/html/man7
  	/Users/runner/work/msquic/msquic/target/package/msquic-2.5.0-beta/submodules/quictls/doc/man
  	/Users/runner/work/msquic/msquic/target/package/msquic-2.5.0-beta/submodules/quictls/doc/man/man1
  	/Users/runner/work/msquic/msquic/target/package/msquic-2.5.0-beta/submodules/quictls/doc/man/man3
  	/Users/runner/work/msquic/msquic/target/package/msquic-2.5.0-beta/submodules/quictls/doc/man/man5
  	/Users/runner/work/msquic/msquic/target/package/msquic-2.5.0-beta/submodules/quictls/doc/man/man7
 ```

For some reason when we build Rust against quictls v3 it seems to be changing files under the `doc` subdirectory. Any ideas why? Rust doesn't like that it's now "dirty".

**Comment by nhorman** [comment]
Just an initial guess, but my first thought would be the fact that openssl generates the files that you are hitting errors on using perldoc, and so perhaps those files are getting left over from the build?

Its odd, as I would have thought anything in this area would have been hit in quictls1.1 (unless that tree did something internally to not build docs).  

If this hypothesis is vaguely accurate, I would think that the fast solution would be to configure openssl in submodules/CMakeLists.txt such that OPENSSL_CONFIG_FLAGS included the no-docs option.  That would prevent any doc building, and given that (I think) you don't actually package the openssl docs, speed up the build somewhat considerably).

**Comment by nibanks** [comment]
> Just an initial guess, but my first thought would be the fact that openssl generates the files that you are hitting errors on using perldoc, and so perhaps those files are getting left over from the build?
> 
> Its odd, as I would have thought anything in this area would have been hit in quictls1.1 (unless that tree did something internally to not build docs).
> 
> If this hypothesis is vaguely accurate, I would think that the fast solution would be to configure openssl in submodules/CMakeLists.txt such that OPENSSL_CONFIG_FLAGS included the no-docs option. That would prevent any doc building, and given that (I think) you don't actually package the openssl docs, speed up the build somewhat considerably).

Ok. Thanks! I'll try the `no-docs` option. I had to add the `docs` folder to get quictls v3 to build. It wasn't there (nor seemed to be required) for v1.1....

**Comment by nhorman** [comment]
that would support the theory that either openssl1.1 separated out the doc build somehow, or quictls modified the build process.

**Comment by nhorman** [comment]
grrr, quictls3 doesn't appear to support no-docs.  Let me look into that

**Comment by nhorman** [comment]
damn, looks like the no-docs option was added in openssl-3.2.0, so the build for quictls is always going to produce docs.

Next best option (I think) would be to, after the build is complete for quictls, cd into docs, and run git clean.  Ugly, but it should work.

File: .azure/OneBranch.Publish.yml [hunk 1]
```diff
@@ -22,40 +22,7 @@ parameters:
...
-        - microsoft-ubuntu-bionic-prod-apt   # 18.04 4.15
-        - microsoft-ubuntu-focal-prod-apt    # 20.04 5.4
-        - microsoft-ubuntu-groovy-prod-apt   # 20.10 5.8
-        - microsoft-ubuntu-hirsute-prod-apt  # 21.04 5.11
-        - microsoft-debian-stretch-prod-apt  #  9 4.9
-        - microsoft-debian-buster-prod-apt   # 10 4.19
-        - microsoft-debian-bullseye-prod-apt # 11 5.10
-    - name: CBL
-      destinations:
-        - cbl-mariner-1.0-prod-Microsoft-x86_64-rpms-yum
```
anrossi, line 55: no more mariner support?

.azure/templates/download-artifacts.yml
anrossi, line 1: shouldn't all these deletes be their own separate PR since it's not really related to removal of quictls 1.1.1?  Suppose we need to revive AZP in the future; how can we make this easier to find?

.github/workflows/build.yml
```diff
@@ -227,33 +224,28 @@ jobs:
...
+      uses: actions/download-artifact@95815c38cf2ff2164869cbab79da8d1f422bc89e
       with:
-        run_id: ${{github.run_id}}
         name: Release-${{ matrix.vec.plat }}-windows-2022-x64-${{ matrix.vec.tls }}
         path: artifacts
-        if_no_artifact_found: ignore
     - name: Download Build Artifacts
-      uses: dawidd6/action-download-artifact@07ab29fd4a977ae4d2b275087cf67563dfdf0295
+      uses: actions/download-artifact@95815c38cf2ff2164869cbab79da8d1f422bc89e
+      if: matrix.vec.tls != 'quictls' # TODO: Remove once quictls is supported
```
anrossi, line 238: Are these TODOs for this PR or a future one?

scripts/get-buildconfig.ps1
```diff
@@ -90,28 +90,6 @@ if ("" -eq $Tls) {
         $Tls = "schannel"
     } else {
         $Tls = "quictls"
-        try {
```
anrossi, line 93: Why is this being removed? Don't we need this logic in the future for determining whether to build OpenSSL or quictls?

anrossi (commented):
I see some changes to the packaging automation; does the CI run this as well so we know if this breaks things?

.azure/OneBranch.Publish.yml
nibanks: We need to figure out how to build for a more recent version. The old one used v1.1 openssl.

.azure/templates/download-artifacts.yml
nibanks: I had to fix this stuff up to get this to work here.

.github/workflows/build.yml
nibanks: A future PR. I cleaned up how these things worked. If you look above, you will see build for arm64 is also a TODO. I hope that by moving to the latest version of quictls/openssl we can remove these.

scripts/get-buildconfig.ps1
nibanks: The was trying to see if we're building for openssl3. Now, we **always** build for openssl3.

nibanks:
> I see some changes to the packaging automation; does the CI run this as well so we know if this breaks things?

If it's in GitHub, then it runs it all; otherwise we will have to merge it and see what breaks.

anrossi:
Are these ubuntu test failures unrelated? I know the babeltrace crash is unrelated, but the test failure is the same on two of the test runs. I started a rerun on one of them

nibanks:
> Are these ubuntu test failures unrelated? I know the babeltrace crash is unrelated, but the test failure is the same on two of the test runs. I started a rerun on one of them

Yes, they are unrelated.

anrossi (approved):

PR #5049, opened 2025-04-25
Always use host silo for ETW registration

csujedihy:
## Description

We don't support per-silo ETW tracing. So, always use the host silo for everything in DriverEntry and DriverUnload,

## Testing

Locally tested in a VM. CI.

### DIFF
--- a/src/bin/winkernel/driver.c
+++ b/src/bin/winkernel/driver.c
@@ -95,6 +95,8 @@ Return Value:
     NTSTATUS Status;
     WDF_DRIVER_CONFIG Config;
     WDFDRIVER Driver;
+    QUIC_SILO PrevSilo = QuicSiloAttach(QuicSiloGetHostServerSilo());
+    CXPLAT_DBG_ASSERT(!QuicSiloIsServerSilo());
 
     //
     // We explicitly load the MsQuic library upfront (instead of letting it
@@ -143,6 +145,7 @@ Error:
         MsQuicLibraryUnload();
     }
 
+    QuicSiloDetatch(PrevSilo);
     return Status;
 }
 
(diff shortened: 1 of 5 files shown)


src/inc/quic_platform_winkernel.h
```diff
@@ -955,6 +955,8 @@ CxPlatRandom(
 #define QUIC_SILO PESILO
 #define QUIC_SILO_INVALID ((PESILO)(void*)(LONG_PTR)-1)
 
+#define QuicSiloGetHostServer() (PsGetHostSilo())
```
Robo210, line 958: While Windows refers to these silos as "Server Containers" or "Server Silos", the use of the word "Server" at the end of the names here is a bit confusing because it seems to imply a "client / server" type situation.

`QuicSiloGetHostSilo`, `QuicSiloIsInSilo`, `QuicSiloGetCurrentSilo` might make more sense.

csujedihy: I agree. I followed the existing bad naming pattern, `QuicSiloGetCurrentServer`. I will update all of them.

csujedihy:
I will move the attach and detach to driver entry. It makes more sense there.

src/bin/winkernel/driver.c
```diff
@@ -143,6 +145,7 @@ Return Value:
         MsQuicLibraryUnload();
     }
 
+    QuicSiloDetatch(PrevSilo);
```
anrossi, line 148: Does `DriverUnload` get called when `DriverEntry` returns error?

csujedihy: Nope. Is that an issue?

Robo210: If you register an ETW provider and then unload the image (dll/sys) that contains the enable callback without unregistering the provider, there will be a crash when the callback is next invoked.

Robo210: (I don't know if that's happening here or not, but just FYI)

csujedihy: We have proper error handling in DriverEntry to ensure everything gets cleaned up when bailing out.

nibanks (approved):

PR #5061, opened 2025-04-30
Add io_uring support

mtfriesen:
## Description

## Testing

## Documentation


src/inc/quic_platform_posix.h
```diff
@@ -517,21 +548,21 @@ CxPlatPoolInitialize(
...
     _Inout_ CXPLAT_POOL* Pool
     )
 {
-    CXPLAT_POOL_HEADER* Entry;
-    while ((Entry = (CXPLAT_POOL_HEADER*)CxPlatListPopEntry(&Pool->ListHead)) != NULL) {
-        CXPLAT_DBG_ASSERT(Entry->SpecialFlag == CXPLAT_POOL_FREE_FLAG);
-        CxPlatFree(Entry, Pool->Tag);
-    }
-    CxPlatLockUninitialize(&Pool->Lock);
+    CxPlatPoolInitializeEx(IsPaged, Size, Tag, NULL, NULL, Pool);
```
nibanks, line 565: Instead of NULLs and always doing if checks in every alloc/free, should we have defaults that call what we want?

src/platform/datapath_epoll.c
```diff
@@ -189,155 +167,10 @@ typedef struct CXPLAT_RECV_MSG_CONTROL_BUFFER {
 
 } CXPLAT_RECV_MSG_CONTROL_BUFFER;
 
-#ifdef DEBUG
-#define CXPLAT_DBG_ASSERT_CMSG(CMsg, type) \
-    if (CMsg->cmsg_len < CMSG_LEN(sizeof(type))) { \
-        printf("%u: cmsg[%u:%u] len (%u) < exp_len (%u)\n", \
```
nibanks, line 195: I never noticed we had a printf assert. We should replace this with a real assert.

src/platform/platform_internal.h
```diff
@@ -917,6 +917,13 @@ typedef struct CXPLAT_SOCKET {
 
 } CXPLAT_SOCKET;
 
+typedef struct CXPLAT_REGISTERED_BUFFER_POOL {
+    void *Ring;
+    uint8_t *Buffers;
```
nibanks, line 922: nit, style puts the `*` with the type
```suggestion
    void* Ring;
    uint8_t* Buffers;
```

src/platform/platform_worker.c
```diff
@@ -556,16 +556,18 @@ CxPlatProcessEvents(
...
 #if DEBUG // Debug statistics
         Worker->CqeCount += CqeCount;
 #endif
         State->NoWorkCount = 0;
-        for (uint32_t i = 0; i < CqeCount; ++i) {
-            CXPLAT_SQE* Sqe = CxPlatCqeGetSqe(&Cqes[i]);
-            Sqe->Completion(&Cqes[i]);
+        while (CqeCount > 0) {
+            CXPLAT_SQE* Sqe = CxPlatCqeGetSqe(CurrentCqe);
+            Sqe->Completion(&CurrentCqe, &CurrentCqeCount);
```
nibanks, line 570: Every SQE has its own completion function. Why are you changing that?

src/inc/quic_platform_posix.h
```diff
@@ -968,73 +1046,148 @@ void
...
     )
 {
-    io_uring_cq_advance(queue, count);
+    return (CXPLAT_SQE*)(uintptr_t)(*cqe)->user_data;
+}
+
+inline
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+CxPlatSqeClassicCompletion(
```
nibanks, line 1178: Why add this interface to just do what the caller was doing anyways?

src/platform/platform_internal.h
mtfriesen: Yeah, I will have to do one sweep for these after, because they keep creeping in.

src/inc/quic_platform_posix.h
mtfriesen: I might back out this change if I don't get as far as zerocopy TX, which is the only use case for this model.

src/platform/platform_worker.c
mtfriesen: The new completion function supports batching completions, which is better than calling Foo() 100 times, and having Foo() do boilerplate (lock, submit, etc.) each time.

src/inc/quic_platform_posix.h
mtfriesen: See the other comment (The new completion function supports batching completions) and the comment in this function (it's a hack to avoid updating all code).

src/platform/platform_worker.c
nibanks: Batching of what exactly though? If you have a list of 10 completion functions to call, what is being batched? The callbacks are all outside of a lock, right?

mtfriesen: The `CurrentCqe` and `CurrentCqeCount` are both `_Inout_` parameters, so the routine can consume [1, CurrentCqeCount] CQEs.

Here's the only routine taking advantage:
https://github.com/microsoft/msquic/blob/0752c6da0e1835622ba7e1a46d7d652de111194f/src/platform/datapath_iouring.c#L2106-L2116

src/platform/datapath_iouring.c
```diff
@@ -0,0 +1,2128 @@
...
+    SendData->MsgHdr.msg_control = SendData->ControlBuffer;
+    SendData->MsgHdr.msg_controllen = SendData->ControlBufferLength;
+    if (SendData->ControlBufferLength == 0) {
+        CxPlatSendDataPopulateAncillaryData(SendData, &SendData->MsgHdr);
+    } else {
+        SendData->MsgHdr.msg_controllen = SendData->ControlBufferLength;
+    }
+
+    io_uring_prep_sendmsg(Sqe, SendData->SocketContext->SocketFd, &SendData->MsgHdr, 0);
+    io_uring_sqe_set_data(Sqe, (void *)&SendData->Sqe);
```
guhetier, line 1922: Shouldn't there be a io_uring_submit call around here?
I am not finding where the initial submission occurs for the send side.

mtfriesen: Yes and no, and in the yes case, it is indeed missing :)

In the no case, the submit happens in CxPlatSocketContextIoEventComplete.

src/inc/quic_platform_posix.h
mtfriesen: I reverted this for now, but I'm actually not sure whether a default handler (through an indirect function pointer) is higher performance than a NULL check and inline fallback. The answer is probably "it depends".

src/platform/datapath_epoll.c
mtfriesen: Agreed, done.

src/platform/platform_internal.h
mtfriesen: Swept. There were a lot...

CMakeLists.txt
```diff
@@ -455,6 +456,15 @@ if (NOT MSVC AND NOT APPLE AND NOT ANDROID)
     endif()
 endif()
 
+if (QUIC_LINUX_IOURING_ENABLED)
+    find_library(LIBURING NAMES liburing-ffi.so)
+    if (LIBURING)
+        message(STATUS "Found liburing: ${LIBURING}")
+    else()
+        message(STATUS "liburing not found. If build fails, install liburing")
```
anrossi, line 464: should this be a warning instead of just a status?

mtfriesen: Yeah, probably. I suspect I copied this from the XDP example several months ago - I can fix both if you tell me how 😄

src/inc/msquic_posix.h
```diff
@@ -543,6 +579,8 @@ typedef struct QUIC_SQE {
     QUIC_EVENT_COMPLETION_HANDLER Completion;
 } QUIC_SQE;
 
+#endif
```
anrossi: Can we get a comment on this tying it to the `#if` it belongs to?

CMakeLists.txt
mtfriesen: Looks like I copied libnuma directly above, not XDP

src/inc/quic_platform_posix.h
```diff
@@ -988,73 +1025,171 @@ void
...
+(CXPLAT_EVENT_BATCH_COMPLETION)(
+    _Inout_ CXPLAT_CQE** Cqes,
+    _Inout_ uint32_t* Count
+    );
+typedef CXPLAT_EVENT_BATCH_COMPLETION *CXPLAT_EVENT_BATCH_COMPLETION_HANDLER;
 typedef struct CXPLAT_SQE {
-    CXPLAT_EVENT_COMPLETION_HANDLER Completion;
+    CXPLAT_EVENT_COMPLETION_HANDLER ClassicCompletion;
+    CXPLAT_EVENT_BATCH_COMPLETION* Completion;
+    void* Context; // Review: create an encapsulating structure for this?
```
anrossi: Why is a void* insufficient and an encapsulating structure needed?

src/inc/quic_platform_posix.h
```diff
@@ -1066,35 +1201,48 @@ CxPlatSqeInitialize(
...
+    sqe->Signature = CXPLAT_SQE_SIGNATURE_INITIALIZED;
+#endif
     return TRUE;
 }
 
 QUIC_INLINE
-void
-CxPlatSqeCleanup(
+BOOLEAN
+CxPlatSqeInitialize2(
```
anrossi: Initialize**2**?!

src/inc/quic_platform_posix.h
```diff
@@ -1066,35 +1201,48 @@ CxPlatSqeInitialize(
...
     _In_ CXPLAT_EVENTQ* queue,
-    _In_ CXPLAT_SQE* sqe
+    _In_ CXPLAT_EVENT_BATCH_COMPLETION_HANDLER completion,
```
anrossi, line 1215: nit: consistent captialization on parameters

src/inc/quic_platform_posix.h
mtfriesen: Oh, I really should have fixed this - any context is typically stored in an encapsulating structure, e.g.:

```
CXPLAT_URING_SQE {
    CXPLAT_SQE Sqe;
    struct MY_URING_CONTEXT MyContext;
};
```

I'll update

src/inc/quic_platform_posix.h
mtfriesen: Yeah, this deserves a better name - I'll call it a batched SQE

src/inc/quic_platform_posix.h
mtfriesen: Luckily addressing the other comment ("what is a `void *` doing here?) will resolve this one.

src/platform/datapath_iouring.c
```diff
@@ -0,0 +1,2264 @@
...
+typedef enum DATAPATH_CONTEXT_TYPE {
+    DatapathContextRecv,
+    DatapathContextSend,
+} DATAPATH_CONTEXT_TYPE;
+
+//
+// Contains all the info for a single RX IO operation. Multiple RX packets may
+// come from a single IO operation.
+//
+typedef struct __attribute__((aligned(16))) DATAPATH_RX_IO_BLOCK {
```
anrossi: didn't you define a handy constant, `CXPLAT_MEMORY_ALIGNMENT` in another header?

src/platform/datapath_iouring.c
```diff
@@ -0,0 +1,2264 @@
...
+    //DATAPATH_RX_PACKET Packets[0];
+
+    //
+    // Buffer that actually stores the UDP payload.
+    //
+    //uint8_t Buffer[]; // CXPLAT_SMALL_IO_BUFFER_SIZE or CXPLAT_LARGE_IO_BUFFER_SIZE
+
+} DATAPATH_RX_IO_BLOCK;
+
+typedef struct __attribute__((aligned(16))) DATAPATH_RX_PACKET {
```
anrossi: Ditto

src/platform/datapath_iouring.c
```diff
@@ -0,0 +1,2264 @@
...
+            DatapathErrorStatus,
+            "[data][%p] ERROR, %u, %s.",
+            DatapathPartition,
+            Status,
+            "io_uring_register_buf_ring failed");
+            goto Exit;
+    }
+
+    //
+    // Review: we may also want to io_uring_register_buffers for
```
anrossi, line 309: we may, but I am ill-equipped to answer one way or another.

anrossi: why would we want to do that? and how is it different from what is being done now?

src/platform/datapath_iouring.c
```diff
@@ -0,0 +1,2264 @@
...
+    SocketContext->DatapathPartition = &Datapath->Partitions[PartitionIndex];
+    CxPlatRefIncrement(&SocketContext->DatapathPartition->RefCount);
+
+    Status = CxPlatSocketContextSqeInitialize(SocketContext);
+    if (QUIC_FAILED(Status) || SocketType == CXPLAT_SOCKET_TCP_SERVER) {
+        goto Exit;
+    }
+
+    //
+    // Create datagram socket. (Review: these steps could be performed using the io_uring).
```
anrossi, line 613: is there a performance benefit to using the io_uring for these steps? File a task to track that work if it's important, otherwise we can wait for a perf benchmark to prove it's needed

src/platform/datapath_iouring.c
mtfriesen: This egg came before the chicken...

src/platform/datapath_iouring.c
mtfriesen: It just pre-registers more stuff with the kernel so it doesn't have to take and fully validate and handle raw fds/pointers for each IO. Theoretically, that would improve perf, but also increases complexity.

src/platform/datapath_iouring.c
mtfriesen: I think the premise of this module is that all IO should occur on the IO uring, eventually. Each IO may have minor improvements (or minor regressions) but it promises consistently asynchronous rather than randomly blocking IO.

src/platform/datapath_iouring.c
```diff
@@ -0,0 +1,2264 @@
...
+    IoPayload = (uint8_t*)IoBlock + DatapathPartition->Datapath->RecvBlockBufferOffset;
+    RecvMsgOut = io_uring_recvmsg_validate(IoPayload, Cqe->res, (struct msghdr*)&CxPlatRecvMsgHdr);
+    CXPLAT_FRE_ASSERT(RecvMsgOut != NULL); // Review: can this legally fail?
+
+    CXPLAT_DBG_ASSERT((uintptr_t)IoBlock % CXPLAT_MEMORY_ALIGNMENT == 0);
+
+    IoBlock->Route.State = RouteResolved;
+
+    //
+    // Review: these can be batched by propagating the CQE array here.
```
anrossi, line 1671: We should track this with a backlog task; batching usually provides tangible performance benefits and we should make use of that when possible.

mtfriesen: IMO let's not clutter up our backlog with these items yet - these are just ideas, hence "review"

CMakeLists.txt
mtfriesen: I'm going to wontfix this for now because I need to stop churning, and it seems to follow the precedent. We can sweep them all later, if needed.

anrossi: replace "STATUS" with "WARNING" and I think that'll do it

mtfriesen: Opened #5330

anrossi (approved):

PR #5062, opened 2025-04-30
RecvBuffer simplifications [3/N]

guhetier:
## Description

Refactor the "drain" operation in the receive buffer.

The "drain" operation has a lot of mode-specific behavior and ended-up being implemented in a way that required a lot of special casing.
In particular, the "full" and "partial" drain logic was convoluted.

This PR simplify the "drain" operation by using a more generic approach and limit the number of special cases. It still uses a "full" and "partial" drain helpers, but in a more systematic way:

- the "full" drain deals with all chunks that are fully consumed (all bytes have been written, read and drained)
- the "partial" (or "firstchunk") drain updates the chunk now at the front of the list

## Testing

C/I

## Documentation

No changes, the existing documentation cover this.

### DIFF
--- a/src/core/recv_buffer.c
+++ b/src/core/recv_buffer.c
@@ -44,10 +44,10 @@ Abstract:
 #endif
 
 typedef struct QUIC_RECV_CHUNK_ITERATOR {
-    QUIC_RECV_CHUNK* CurrentChunk;
+    QUIC_RECV_CHUNK* NextChunk;
     CXPLAT_LIST_ENTRY* IteratorEnd;
-    uint32_t StartOffset; // Offset of the first byte to read in the current chunk.
-    uint32_t EndOffset;   // Offset of the last byte to read in the current chunk (inclusive!).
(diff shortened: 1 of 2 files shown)


nibanks:
```
/__w/msquic/msquic/src/core/recv_buffer.c: In function 'QuicRecvBufferDrainFullChunks':
/__w/msquic/msquic/src/core/recv_buffer.c:960:9: error: 'return' with no value, in function returning non-void [-Werror=return-type]
  960 |         return;
      |         ^~~~~~
/__w/msquic/msquic/src/core/recv_buffer.c:934:1: note: declared here
  934 | QuicRecvBufferDrainFullChunks(
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/__w/msquic/msquic/src/core/recv_buffer.c: At top level:
/__w/msquic/msquic/src/core/recv_buffer.c:995:1: error: unknown type name 'VOID'
  995 | VOID
      | ^~~~
```

nibanks (approved):

PR #5063, opened 2025-05-01
Randomize Base Address in Kernel Projects

nibanks:
## Description

See [here](https://learn.microsoft.com/en-us/cpp/build/reference/base-base-address?view=msvc-170) for more details on what and why.

We will want to backport this to v2.3 and v2.4.

## Testing

CI/CD

## Documentation

N/A


csujedihy (approved):

Harvester57:
For reference, here is the correct MSFT documentation for the linker option `RandomizedBaseAddress`: https://learn.microsoft.com/en-us/cpp/build/reference/dynamicbase-use-address-space-layout-randomization?view=msvc-170

PR #5074, opened 2025-05-05
Remove duplicate inline.c definitions for clang/Linux

mtfriesen:
## Description

It seems header-only libraries/functions are idiomatically declared `static inline` on Linux. This removes the need for a dummy .c file to produce an object file containing definitions.

## Testing

Builds locally on Linux.

## Documentation

No.


src/inc/msquic.h
```diff
@@ -225,7 +225,7 @@ typedef enum QUIC_STREAM_SHUTDOWN_FLAGS {
     QUIC_STREAM_SHUTDOWN_FLAG_ABORT_RECEIVE = 0x0004,   // Abruptly closes the receive path.
     QUIC_STREAM_SHUTDOWN_FLAG_ABORT         = 0x0006,   // Abruptly closes both send and receive paths.
     QUIC_STREAM_SHUTDOWN_FLAG_IMMEDIATE     = 0x0008,   // Immediately sends completion events to app.
-    QUIC_STREAM_SHUTDOWN_FLAG_INLINE        = 0x0010,   // Process the shutdown immediately inline. Only for calls on callbacks.
+    QUIC_STREAM_SHUTDOWN_FLAG_INLINE        = 0x0010,   // Process the shutdown immediately QUIC_INLINE. Only for calls on callbacks.
```
guhetier, line 228: nit: Looks like a spurious replace here

src/inc/msquic_posix.h
```diff
@@ -38,6 +38,8 @@
 #include <errno.h>
 #include "quic_sal_stub.h"
 
+#define QUIC_INLINE static inline
```
guhetier, line 41: Could it backfire to have `QUIC_INLINE` always be static under the hood?
It there situations where we need only inline in a `forceinline` meaning, and `QUIC_INLINE` would be confusing?

(since it is more meant as a "inline in header")?

mtfriesen: To partially answer your question, there is a separate `FORCEINLINE` macro defined that I also had to fix.

guhetier: In that case, maybe `QUIC_HEADER_INLINE`?
Or actually, is there a problem with `static inline` for all platforms?

In any case, the PR is an improvement over the current state.

mtfriesen: I don't see a substantial win with QUIC_HEADER_INLINE, and now the code is straightforward to refactor in the future if a significantly better name occurs to us.

src/test/lib/KArray.h
```diff
@@ -44,7 +44,7 @@ namespace wistd     // ("Windows Implementation" std)
 //
 
 template <class _Tp>
-inline constexpr
```
nibanks, line 47: These changes are unnecessary for templates because they're C++, right?

src/inc/quic_platform.h
```diff
@@ -68,7 +68,7 @@ typedef struct CXPLAT_SLIST_ENTRY {
 #if (_MSC_VER >= 1200)
 #define FORCEINLINE __forceinline
 #else
-#define FORCEINLINE __inline
+#define FORCEINLINE QUIC_INLINE
```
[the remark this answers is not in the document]
mtfriesen: MSDN [says](https://learn.microsoft.com/en-us/cpp/cpp/inline-functions-cpp?view=msvc-170):

> The __inline keyword is equivalent to inline.

src/test/lib/KArray.h
mtfriesen: Hmm, good question. I didn't think too hard about this, and I'm not a C++ expert. I can say the code compiles and tests seem to pass, but I can't say this change is necessary. I'm happy to revert here, since there was not affected inline.c file.

nibanks (approved):

PR #5077, opened 2025-05-05
Upload rust binding diff as an artifact

guhetier:
## Description

Upload rust binding diff as an artifact + print it in the logs: the comment on the PR doesn't work for forks based PRs.

## Testing

See on this PR CI run.

## Documentation

N/A


ProjectsByJackHe (approved):
LGTM once you remove the dummy code to break Cargo.

csujedihy (approved):

guhetier:
Should be ready to go.
An additional improvement could be to try to edit the diff comment instead of having multiple ones, but I am not sure it is worth the additional complexity, so lets see how it is in practice first.

PR #5080, opened 2025-05-06
Add TX checksum offload for Windows XDP

mtfriesen:
## Description

This PR adds support for TX checksum offload for Windows XDP. The offload may not be supported for at least the following reasons:

1. The version of XDP does not support the offload yet
2. The interface the socket is bound to does not support the offload
3. The offload is defined as experimental, so the version of XDP may no longer support the offload

Note the version of the XDP runtime in MsQuic is _not_ being updated to a prerelease build with checksum support, so this code will not be practically enabled yet in CI/netperf/etc.

## Testing

We rely on the XDP interop tests for correctness, though this is not robust if the VMs running the interop test do not advertise checksum support. I've also manually run interop tests between XDP client and the full stack IOCP server in netperf, which we should eventually onboard both for performance and conformance coverage.

## Documentation

Mostly self-explanatory implementation details.


ProjectsByJackHe:
For clarity, the TX offloading capabilities were already there (set IP->headerCheckSum = 0)

Would a better title for this PR be: "Refactor XDP TX checksum offloads"?

src/platform/datapath_raw_socket.c
```diff
@@ -763,26 +769,36 @@ CxPlatFramingWriteHeaders(
         IPv4->HeaderChecksum = 0;
         CxPlatCopyMemory(IPv4->Source, &Route->LocalAddress.Ipv4.sin_addr, sizeof(Route->LocalAddress.Ipv4.sin_addr));
         CxPlatCopyMemory(IPv4->Destination, &Route->RemoteAddress.Ipv4.sin_addr, sizeof(Route->RemoteAddress.Ipv4.sin_addr));
-        IPv4->HeaderChecksum = SkipNetworkLayerXsum ? 0 : ~CxPlatFramingChecksum((uint8_t*)IPv4, sizeof(IPV4_HEADER), 0);
+        if (SkipNetworkLayerXsum) {
+            IPv4->HeaderChecksum = 0;
```
nibanks, line 773: This is set like 5 lines up already.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -660,6 +670,27 @@ CxPlatDpRawInterfaceInitialize(
             goto Error;
         }
 
+        //
+        // TODO: check whether the interface currently supports the offload
+        // before enabling it (XSK_SOCKOPT_TX_OFFLOAD_CURRENT_CONFIG_CHECKSUM).
```
nibanks, line 675: If it's not supported would it simply fail? If so, why check first?

mtfriesen: Yeah, even I forgot how this works. The app can set `XSK_SOCKOPT_TX_OFFLOAD_CHECKSUM` to enable the checksum offload ring descriptors, which is independent of whether the underlying NIC supports checksum offload and/or has it currently enabled.

The reason for that is the NDIS model (rightly, probably) allows the NIC hardware capabilities and configuration to change on the fly. XDP follows this model by allowing the app to opt in to the checksum ring extensions, even if the NIC doesn't happen to have the offload supported _right now_.

For MsQuic, we'd probably make a simplifying assumption that the NIC capabilities don't change, and enable the ring extension only if the HW currently has the offload enabled, which is the TODO here.

mtfriesen:
> For clarity, the TX offloading capabilities were already there (set IP->headerCheckSum = 0) in `datapath_raw_socket` prior to this PR, correct?
> 
> Would a better title for this PR be: "Set correct socket options for TX checksum offloads"?

Until now, the common `datapath_raw_socket` conditionally computed the checksums based on the datapath capabilities, and the Windows XDP datapath never declared that capability. So this PR adds TX checksum offloads for Windows XDP.

src/platform/datapath_raw_socket.c
```diff
@@ -516,19 +516,24 @@ CxPlatFramingTransportChecksum(
     _In_ uint32_t AddrLength,
     _In_ uint16_t NextHeader,
     _In_reads_(IPPayloadLength) uint8_t* IPPayload,
-    _In_ uint32_t IPPayloadLength
+    _In_ uint32_t IPPayloadLength,
+    _In_ BOOLEAN PseudoHeaderOnly
```
ProjectsByJackHe, line 520: What's this for?

src/platform/datapath_raw_xdp_win.c
```diff
@@ -71,6 +72,14 @@ typedef struct XDP_QUEUE {
     CXPLAT_SQE TxIoSqe;
     XSK_RING TxRing;
     XSK_RING TxCompletionRing;
+    struct {
```
ProjectsByJackHe, line 75: Just curious why not flatten this out instead of making it a nested struct? Is this a memory layout optimization?

mtfriesen: Yeah, this is just copy/pasta. I kept it similar because (a) my own personal preference on the matter doesn't outweigh simply following the existing pattern in this project and (b) it seemed reasonable to me, in terms of being somewhat more compact in memory and more friendly for auto-complete, code searching, etc.

src/platform/datapath_raw_socket.c
csujedihy: HW xsum offload requires a pseudoHeader to be passed down to the NIC

mtfriesen: `PseudoHeaderOnly` computes the framing checksum's IP pseudoheader checksum without computing the final checksum over the transport layer header and data.

mtfriesen: > HW xsum offload requires a pseudoHeader to be passed down to the NIC

Sorry, I missed this answer when writing mine. BTW, from what I could tell when skimming DPDK, the pseudoheader might or might not need to be filled in, depending on the PMD. We don't have any tests yet, but I think we have a latent bug in our own DPDK code, because we need to conditionally zero/initialize with a pseudoheader checksum for DPDK.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1734,6 +1911,22 @@ CxPlatXdpTx(
         Queue->Error = TRUE;
     }
 
+    if (XskRingOffloadChanged(&Queue->TxRing) &&
```
csujedihy, line 1914: What happens if offload change races with us submitting packets without xsum computed to xdp? Does xdp drop or compute on behalf of the caller?

mtfriesen: XDP never compares the current hardware state, it just sets the OOBs and lets the packets rip.

csujedihy:
It looks like there are 2 sources of truth now for offload enablement.

1. `Interface->OffloadStatus.Transmit.TransportLayerXsum`

2. `CxPlatDpRawIsL3TxXsumOffloadedOnQueue`

1 is only used in QTIP code path. Is this expected?

src/platform/datapath_raw_xdp_win.c
csujedihy: Ok assuming that wouldn't cause the NIC to freak out

csujedihy:
very nit: "Describe the purpose of and changes within this Pull Request." we probably don't need the placeholder in the PR describption.

src/platform/datapath_raw_xdp_win.c
mtfriesen: As always in the Windows ecosystem, there are no guarantees, and we do know some NICs reuse/abuse OOBs they don't indicate support for. To a limited extent, the point of XDP is to avoid the runtime and maintenance costs of supporting rare/non-compliant drivers, so we're assuming NICs used in XDP scenarios will not freak until proven otherwise.

mtfriesen:
> It looks like there are 2 sources of truth now for offload enablement.
> 
> 1. `Interface->OffloadStatus.Transmit.TransportLayerXsum`
> 2. `CxPlatDpRawIsL3TxXsumOffloadedOnQueue`
> 
> 1 is only used in QTIP code path. Is this expected?

Thanks, I meant to update QTIP too, but got forgot in the chaos. Great catch.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -624,39 +672,79 @@ CxPlatDpRawInterfaceInitialize(
...
+        if (QUIC_FAILED(Status)) {
+            CXPLAT_DBG_ASSERT(!TxChecksumOffload);
+            QuicTraceEvent(
+                LibraryErrorStatus,
+                "[ lib] ERROR, %u, %s.",
+                Status,
+                "GetTxOffloadConfig(Queue, &TxChecksumOffload)");
+        }
+
+        if (TxChecksumOffload || QUIC_SUCCEEDED(Status) /* TODO: Just do the offload, even if the NIC says it can't. */) {
```
nibanks, line 702: Shouldn't this only offload if it's supported?

src/platform/datapath_raw_xdp_win.c
```diff
@@ -624,39 +672,79 @@ CxPlatDpRawInterfaceInitialize(
...
+        //
+        // Before enabling the TX offload descriptors, check whether the offload
+        // is currently enabled on the interface. If it isn't, assume the
+        // interface (or XDP driver) does not and will not support the offload.
+        //
+        uint32_t TxChecksumOffload = FALSE;
+        Status = GetTxOffloadConfig(Queue, &TxChecksumOffload);
+        if (QUIC_FAILED(Status)) {
+            CXPLAT_DBG_ASSERT(!TxChecksumOffload);
+            QuicTraceEvent(
```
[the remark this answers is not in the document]
nibanks: OR, should we always simply try to set and don't even query?

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1735,6 +1918,22 @@ CxPlatXdpTx(
         Queue->Error = TRUE;
     }
 
+    if (XskRingOffloadChanged(&Queue->TxRing) &&
+        Queue->OffloadStatus.Transmit.ChecksumOffloadExtensions) {
+        uint32_t TxChecksumOffload;
+        QUIC_STATUS Status = GetTxOffloadConfig(Queue, &TxChecksumOffload);
+        if (QUIC_SUCCEEDED(Status)) {
+            Queue->OffloadStatus.Transmit.ChecksumOffload = !!TxChecksumOffload;
```
nibanks, line 1926: If this gets enabled, should we be setting `XSK_SOCKOPT_TX_OFFLOAD_CHECKSUM` now?

src/platform/datapath_raw_xdp_win.c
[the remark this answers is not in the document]
mtfriesen: We need to know whether the NIC will implement the offload before we skip computing the offload in software - that is essential for correctness.

Re: the logging, I'll update this based on whether MsQuic logs only within leaf functions or within each function in the stack. There's a fundamental difference between XDP not supporting the API (or other random failures, e.g. low resources) and the NIC simply not having the offload enabled/supported, though the end behavior is the same.

src/platform/datapath_raw_xdp_win.c
mtfriesen: No, it's too late.

mtfriesen: https://github.com/microsoft/xdp-for-windows/blob/48117ff9c7322816c3393a2793c617c21f2837f3/published/external/afxdp_experimental.h#L80-L95

mtfriesen:
Converting back to draft status while experimentation continues.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -371,6 +382,38 @@ CxPlatDpRawInterfaceUninitialize(
...
+    uint32_t OptionLength = sizeof(ChecksumConfig);
+
+    Status =
+        XskGetSockopt(
+            Queue->TxXsk, XSK_SOCKOPT_TX_OFFLOAD_CURRENT_CONFIG_CHECKSUM, &ChecksumConfig,
+            &OptionLength);
+    if (QUIC_FAILED(Status) ||
+        OptionLength < XDP_SIZEOF_CHECKSUM_CONFIGURATION_REVISION_1 ||
+        ChecksumConfig.Header.Revision != XDP_CHECKSUM_CONFIGURATION_REVISION_1 ||
+        ChecksumConfig.Header.Size < XDP_SIZEOF_CHECKSUM_CONFIGURATION_REVISION_1) {
```
nibanks, line 402: So, if XDP starts returning a version 2, it breaks our support for the offload? If so, that doesn't seem like a great design.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -371,6 +382,38 @@ CxPlatDpRawInterfaceUninitialize(
...
+{
+    QUIC_STATUS Status;
+    XDP_CHECKSUM_CONFIGURATION ChecksumConfig;
+    uint32_t OptionLength = sizeof(ChecksumConfig);
+
+    Status =
+        XskGetSockopt(
+            Queue->TxXsk, XSK_SOCKOPT_TX_OFFLOAD_CURRENT_CONFIG_CHECKSUM, &ChecksumConfig,
+            &OptionLength);
+    if (QUIC_FAILED(Status) ||
```
nibanks, line 399: We should log at the lowest point (not at the caller). Additionally, I think we should have separate logs to distinguish between failure was returned vs unsupported version.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -371,6 +382,38 @@ CxPlatDpRawInterfaceUninitialize(
...
+        ChecksumConfig.Header.Revision != XDP_CHECKSUM_CONFIGURATION_REVISION_1 ||
+        ChecksumConfig.Header.Size < XDP_SIZEOF_CHECKSUM_CONFIGURATION_REVISION_1) {
+        *TxChecksumOffload = FALSE;
+        if (QUIC_SUCCEEDED(Status)) {
+            Status = QUIC_STATUS_NOT_SUPPORTED;
+        }
+        goto Exit;
+    }
+
+    *TxChecksumOffload = ChecksumConfig.Enabled;
```
nibanks, line 410: Should we log (informational?) a success result too?

src/platform/datapath_raw_xdp_win.c
```diff
@@ -624,39 +667,80 @@ CxPlatDpRawInterfaceInitialize(
...
+                Queue->OffloadStatus.Transmit.ChecksumOffloadExtensions = TRUE;
+            } else {
+                QuicTraceEvent(
+                    LibraryErrorStatus,
+                    "[ lib] ERROR, %u, %s.",
+                    Status,
+                    "XskSetSockopt(XSK_SOCKOPT_TX_OFFLOAD_CHECKSUM)");
+            }
+        } else {
+            QuicTraceEvent(
```
nibanks, line 715: This is a duplicate log in the QUIC_FAILED case above.

src/platform/datapath_raw_xdp_win.c
mtfriesen: Version 2 is by definition a breaking change, and version 2 has not yet been defined. The only legal action the application can take here is to ignore the payload.

src/platform/datapath_raw_xdp_win.c
mtfriesen: Yeah. I'm not sure how to (cheaply) add an informational log, though.

nibanks:
```c
QuicTraceLogInfo(
    XdpInterfaceGetOffload,
    "[data][%p] Shut down (complete)",
    Socket);
```
Something like the above, you can use printf syntax.

src/platform/datapath_raw_xdp_win.c
mtfriesen: I don't have a strong opinion, other than that XDP already consistently uses this pattern everywhere, and I don't see much justification to deviate here.

src/platform/datapath_raw_xdp_win.c
mtfriesen: I'm not sure - `GetTxOffloadConfig` can return QUIC_SUCCEEDED and `TxChecksumOffload` can be false.

csujedihy:
There's a code defect (probably introduced by myself) that's not part of your PR. If we don't rely on these flags for xdp anymore, maybe delete these lines?

In datapath_raw_xdp_win.c:
```
    Interface->OffloadStatus.Transmit.NetworkLayerXsum = Xdp->SkipXsum;
    Interface->OffloadStatus.Transmit.NetworkLayerXsum = Xdp->SkipXsum;
```

The second line should be setting the transport layer xsum.

csujedihy (approved):

PR #5088, opened 2025-05-09
always collect Get-NetView in secnetperf

mtfriesen:
## Description

The secnetperf script collects Get-NetView only if a session exists, which isn't very useful if a session doesn't exist.

Update the script to always collect local data.

## Testing

CI.

## Documentation

No.

### DIFF
--- a/scripts/secnetperf.ps1
+++ b/scripts/secnetperf.ps1
@@ -196,9 +196,11 @@ if (!($Session -eq "NOT_SUPPORTED")) {
     Copy-Item -ToSession $Session ./artifacts -Destination "$RemoteDir/artifacts" -Recurse
     Copy-Item -ToSession $Session ./scripts -Destination "$RemoteDir/scripts" -Recurse
     Copy-Item -ToSession $Session ./src/manifest/MsQuic.wprp -Destination "$RemoteDir/scripts"
+}
 
-    # Create the logs directories on both machines.
-    New-Item -ItemType Directory -Path ./artifacts/logs | Out-Null
+# Create the logs directories on both machines.
+New-Item -ItemType Directory -Path ./artifacts/logs | Out-Null
+if ($Session -ne "NOT_SUPPORTED") {
     Invoke-Command -Session $Session -ScriptBlock {
         New-Item -ItemType Directory -Path $Using:RemoteDir/artifacts/logs | Out-Null
     }
--- a/scripts/secnetperf.ps1
+++ b/scripts/secnetperf.ps1
@@ -205,7 +205,7 @@ if (!($Session -eq "NOT_SUPPORTED")) {
 }
 
 # Collect some info about machine state.
-if (!$NoLogs -and $isWindows -and !($Session -eq "NOT_SUPPORTED")) {
+if (!$NoLogs -and $isWindows) {
     $Arguments = "-SkipNetsh"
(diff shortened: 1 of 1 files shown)


scripts/secnetperf.ps1
```diff
@@ -224,18 +224,20 @@ if (!$NoLogs -and $isWindows -and !($Session -eq "NOT_SUPPORTED")) {
...
-            Invoke-Expression "Get-NetView -OutputDirectory $Using:RemoteDir/artifacts/logs $Using:Arguments"
-            Remove-Item $Using:RemoteDir/artifacts/logs/msdbg.$env:COMPUTERNAME -recurse
-            $filePath = (Get-ChildItem -Path $Using:RemoteDir/artifacts/logs/ -Recurse -Filter msdbg.$env:COMPUTERNAME*.zip)[0].FullName
-            Rename-Item $filePath "get-netview.peer.zip"
-        }
-        Copy-Item -FromSession $Session -Path "$RemoteDir/artifacts/logs/get-netview.peer.zip" -Destination ./artifacts/logs/
-        Write-Host "Generated get-netview.peer.zip"
-    } catch { Write-Host $_ }
-    Write-Host "::endgroup::"
+    if ($Session -ne "NOT_SUPPORTED") {
```
ProjectsByJackHe, line 227: To be clear, this will only impact Azure 1ES runs as lab runners will always have a valid remote powershell session.
So on Azure 1ES runs, this will always crash when calling Copy-Item or Invoke-Command on a "NOT_SUPPORTED" string. The try catch should handle it I suppose but we will need to make sure the netperf CI isn't broken before merging this PR.

mtfriesen: I'm not sure I follow - why would this crash in Azure? BTW, I did kick off a run here: https://github.com/microsoft/netperf/actions/runs/14936326713/job/41964637564

It seems to be working as intended there, and collecting only the local info:

```
Collecting information on local machine state
Fetching watermark_regression.json
Setup complete! Running all tests
```

mtfriesen: It wasn't actually entirely working as intended (a prerequisite was hidden in a different `$Session` check) but should be fixed below.

ProjectsByJackHe: > I'm not sure I follow - why would this crash in Azure? 

Sorry, should clarify my language. Because it's in a try-catch it wouldn't crash, but I am saying `Invoke-Command -Session` will always fail.

Azure 1ES does not use remote powershell to synchronize test orchestration due to security (AFAIU remote powershelling between Azure VMs in a cluster is not allowed because if 1 VM gets hacked, the entire cluster then gets hacked and start mining bitcoin).

If you see around the script, everywhere that has `$Session == "NOT_SUPPORTED"` it uses custom netperf callers to "send" a command to the peer and the peer polls whatever command and runs it (in `quic_callback.ps1`).

ProjectsByJackHe: So my comment is emphasizing that we need to ensure the 1ES Azure netperf runs aren't broken with this PR prior to merging.

mtfriesen: Can you take a look at this run? It seems fine to me:

https://github.com/microsoft/netperf/actions/runs/14936737831/job/41966123529

ProjectsByJackHe: yeah, that looks good.

ProjectsByJackHe (approved):

PR #5089, opened 2025-05-09
Add XDP Interop Test

nibanks:
## Description

Enabled interop testing with XDP.

## Testing

CI/CD

## Documentation

N/A

### DIFF
--- a/src/tools/interop/interop.cpp
+++ b/src/tools/interop/interop.cpp
@@ -739,6 +739,8 @@ RunInteropTest(
     Settings.IsSet.SendBufferingEnabled = TRUE;
     Settings.IdleTimeoutMs = WaitTimeoutMs;
     Settings.IsSet.IdleTimeoutMs = TRUE;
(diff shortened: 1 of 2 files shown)


mtfriesen (approved):

src/tools/interop/interop.cpp
```diff
@@ -739,6 +739,8 @@ RunInteropTest(
     Settings.IsSet.SendBufferingEnabled = TRUE;
     Settings.IdleTimeoutMs = WaitTimeoutMs;
     Settings.IsSet.IdleTimeoutMs = TRUE;
+    Settings.XdpEnabled = TRUE;
```
ProjectsByJackHe, line 742: wouldn't you need something like `if (Feature == XDP_enabled)` here?

nibanks: No. Features here are protocol features, so XDP isn't at that level.

PR #5091, opened 2025-05-10
Remove DataIndication Queuing from  Kernel Datapath

nibanks:
## Description

For several years now, the Windows kernel datapath has been always copying the recevied data in to local buffers because we found that any significant queuing can cause problems lower in the stack (i.e., the NIC); even if we queued only 1 indication. So, let's just remove all the code that supported the queuing mode to eliminate unnecessary branches in the kernel datapath code.

## Testing

CI/CD

## Documentation

N/A

### DIFF
--- a/src/platform/datapath_winkernel.c
+++ b/src/platform/datapath_winkernel.c
@@ -2001,7 +2001,6 @@ CxPlatDataPathSocketReceive(
     }
 
     CXPLAT_DBG_ASSERT(Context != NULL);
-
     CXPLAT_SOCKET* Binding = (CXPLAT_SOCKET*)Context;
 
     const uint32_t CurProcNumber = CxPlatProcCurrentNumber();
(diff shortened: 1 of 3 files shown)


src/platform/datapath_winkernel.c
```diff
@@ -2373,12 +2319,10 @@ CxPlatDataPathSocketReceive(
...
-    if (ReleaseChain != NULL) {
-        //
-        // Release any dropped or copied datagrams.
-        //
-        Binding->DgrmSocket->Dispatch->WskRelease(Binding->Socket, ReleaseChain);
-    }
+    //
+    // Release any dropped or copied datagrams.
+    //
+    Binding->DgrmSocket->Dispatch->WskRelease(Binding->Socket, DataIndicationHead);
```
nibanks, line 2325: Should/can we just return STATUS_SUCCESS at the end to indicate inline processing and then not call `WskRelease` at all?

mtfriesen: > A WSK application calls the WskRelease function to release a linked list of [WSK_DATAGRAM_INDICATION](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wsk/ns-wsk-_wsk_datagram_indication) structures that it previously retained by returning STATUS_PENDING from either its [WskReceiveEvent](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wsk/nc-wsk-pfn_wsk_receive_event) or its [WskReceiveFromEvent](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wsk/nc-wsk-pfn_wsk_receive_from_event) event callback function.

From the docs, return success and do not call release seems simplest.

nibanks: Updated.

PR #5097, opened 2025-05-12
Only One Send or Recevie Flush per Connection Drain

nibanks:
## Description

Updates the connection operation draining logic to only flush send and receive once we call, instead of over and over. The goal here is to improve fairness between connections when they are CPU limited (i.e. both sending or receiving a lot).

## Testing

CI/CD

## Documentation

N/A

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -7686,12 +7686,6 @@ QuicConnDrainOperations(
     _Inout_ BOOLEAN* StillHasPriorityWork
     )
 {
-    QUIC_OPERATION* Oper;
-    const uint32_t MaxOperationCount =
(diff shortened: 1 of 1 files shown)


src/core/connection.c
```diff
@@ -7823,6 +7825,15 @@ QuicConnDrainOperations(
         QuicPerfCounterIncrement(Connection->Partition, QUIC_PERF_COUNTER_CONN_OPER_COMPLETED);
     }
 
+    if (MoreFlushRecvOper) {
```
vgmahajanshetty, line 7828: Whether this can have negative impact on this connection throughput? Because I think this operation will be dequeued and processed on next schedule for this connection in worker loop, which might introduce some delay.

PR #5107, opened 2025-05-19
Resend Close Frames

### LINKED ISSUE #4933: Resend CLOSE Frame in Closing State
### Describe the bug

Per spec ([link](https://www.rfc-editor.org/rfc/rfc9000.html#section-10.2.1)):

> An endpoint in the closing state sends a packet containing a CONNECTION_CLOSE frame in response to any incoming packet that it attributes to the connection.
> 
> An endpoint SHOULD limit the rate at which it generates packets in the closing state. For instance, an endpoint could wait for a progressively increasing number of received packets or amount of time before responding to received packets.

We don't seem to ever be sending a CLOSE frame beyond the first time we send it. So, in the case of packet loss the peer doesn't ever get the indication of connection closure. We should fix this.

### Affected OS

- [x] Windows
- [x] Linux
- [x] macOS
- [ ] Other (specify below)

### Additional OS information

_No response_

### MsQuic version

main

Close a connection and then lose that first packet.

### Expected behavior

The peer should usually get the close on a restrasmit.

### Actual outcome

It never does

### Additional details

_No response_

nibanks:
## Description

Resend the connection close if more data is received after shutdown. Fixes #4933.

## Testing

CI/CD

## Documentation

N/A


src/core/connection.c
```diff
@@ -4411,6 +4411,18 @@ QuicConnRecvFrames(
...
 
+    //
+    // RFC 9000 Section 10.2.1: 
+    // In closing state, respond to any packet with CONNECTION_CLOSE, rate-limited
+    // to once every 5ms.
+    //
+    if (Closed && !Connection->State.ShutdownComplete) {
+        if (RecvTime - Connection->LastCloseResponseTimeUs >= QUIC_CLOSING_RESPONSE_MIN_INTERVAL_US) {
+            QuicSendSetSendFlag(&Connection->Send, QUIC_CONN_SEND_FLAG_CONNECTION_CLOSE);
+            Connection->LastCloseResponseTimeUs = RecvTime;
```
guhetier, line 4422: Shouldn't `LastCloseResponseTimeUs` when closing the connection?
This probably works as long as it has been zeroed out, but we might send a second CONNECTION_CLOSE right after the first instead of 5us later.

guhetier:
Are the submodule update intentional?

guhetier (approved):

src/core/api.h
```diff
@@ -5,6 +5,8 @@
 
 --*/
 
+#define QUIC_CONNECTION_SHUTDOWN_FLAG_STATUS ((QUIC_CONNECTION_SHUTDOWN_FLAGS)0x8000)
```
mtfriesen, line 13: Seems like the rest of these are defined elsewhere:

```
typedef enum QUIC_CONNECTION_SHUTDOWN_FLAGS {
    QUIC_CONNECTION_SHUTDOWN_FLAG_NONE      = 0x0000,
    QUIC_CONNECTION_SHUTDOWN_FLAG_SILENT    = 0x0001,   // Don't send the close frame over the network.
} QUIC_CONNECTION_SHUTDOWN_FLAGS;

DEFINE_ENUM_FLAG_OPERATORS(QUIC_CONNECTION_SHUTDOWN_FLAGS);
```

1. Will this break the C++ operator overrides?
2. Can this be defined in the enum directly, even if it is set to the same 0x8000 value?

mtfriesen (dismissed):

src/core/api.h
mtfriesen: Nit: this also needs a bit of documentation, even if it is an internal flag.

src/core/connection.c
```diff
@@ -703,6 +706,13 @@ QuicConnQueueOper(
         CXPLAT_DBG_ASSERT(QuicConnIsServer(Connection));
         CXPLAT_DBG_ASSERT(Connection->SourceCids.Next != NULL || CxPlatIsRandomMemoryFailureEnabled());
     }
+    if (Oper->Type == QUIC_OPER_TYPE_API_CALL) {
```
guhetier, line 709: Would it be better to have this assert when processing the operation rather than when queuing it?
Long term, if we apply this pattern, this function could become very messy, with validation for many different operations, instead of being completely operation agnostic.

src/core/api.h
nibanks: I do this to specifically hide it from the app. We don't want to expose this complexity to the app, but we do use it internally.

src/core/connection.c
nibanks: I did add another assert in the processing path (in the connection close function).

src/core/api.h
mtfriesen: Makes sense. A very brief bit of internal documentation would still be good, mentioning it is reserved for internal use and that it implies the type of the error code.

src/core/connection.c
```diff
@@ -4411,6 +4422,19 @@ QuicConnRecvFrames(
     uint16_t PayloadLength = Packet->PayloadLength;
     uint64_t RecvTime = CxPlatTimeUs64();
 
+    //
+    // In closing state, respond to any packet with a new close frame (rate-limited).
+    //
+    if (Closed && !Connection->State.ShutdownComplete) {
+        if (RecvTime - Connection->LastCloseResponseTimeUs >= QUIC_CLOSING_RESPONSE_MIN_INTERVAL) {
```
sm-msft, line 4429: Is there an obligation on the server to retain the connection object in the closed state for certain period of time before it is shutdown? 

How do we normally test the state machines on server and client?

nibanks: I believe the spec defines the closing state to last 3 x PTO (Probe time out, which is a multiple of RTT).

src/core/api.h
guhetier: I understand the motivation, but I think a better solution would be to define an internal enum type and convert at the API layer.
It is a bit more work now - but it will be less confusing on the long term and let us leverage tooling (completion, compilers, etc... can't guess this value is a valid part of the enum)

src/core/connection.c
guhetier: In this case, is that one still useful?

src/core/api.h
mtfriesen: Yes, I agree there are even clearer solutions. Personally, though, I dislike converting types through each layer of the stack - some codebases pass the same flag from the API layer deep down into the stack, and reading such simple code feels like a miracle compared to the equivalent Windows code, which may use 3 or 4 or 5 different flags at various layers (sometimes with wildly different names) to express the same API.

mtfriesen: (I also don't want to overthink this - the consistent naming here and the cast to the enum type seem like they should interact reasonably well with most tools.)

guhetier: > feels like a miracle compared to the equivalent Windows code, which may use 3 or 4 or 5 different flags at various layers

I guess it all depends on the code you got to read :)
I remember some parameters plumbed with zero conversions from user-mode to a miniport driver, making any change a nightmare PR.

I don't see it as a blocking issue though, I am fine with whatever Nick decides to go for as long as it is documented.

src/core/connection.c
nibanks: In my experience, the more asserts the better. Different code paths hit differently.

src/core/api.h
nibanks: I'll just add a comment for now, and leave the rest as is.

src/core/api.h
```diff
@@ -5,6 +5,13 @@
 
 --*/
 
+//
+// Private flag of the QUIC_CONNECTION_SHUTDOWN_FLAGS enum used for indicating
+// the type of error code being passed in. Not exposed to apps because they
```
guhetier, line 10: nti:
```suggestion
// a QUIC_STATUS is being passed (and not a QUIC_ERROR). 
// Not exposed to apps because they
```

guhetier (approved):
Slight nitpick to make the comment for precise and readily useful, good to go otherwise.

src/core/api.h
nibanks: The "not a QUIC_ERROR" isn't actually correct either 😄 I think I'd rather just leave it as is for now and not write a massive comment explaining the various types of errors in MsQuic

PR #5109, opened 2025-05-20
added no-asm to the config flag for openssl build

gaurav2699:
## Description

MSQuic builds fail for amd64 using native compiler. It found that the issue is the assembly implementation of openssl doesnt support control flow guards. Adding no-asm to the config flags fixes the issue. 
## Testing

## Documentation


nibanks:
As mentioned over IM, we can't take this even if it fixes some internal build issues. No assembly would hurt perf.

PR #5111, opened 2025-05-20
Add Support for GetParam for Stat Sizes

### LINKED ISSUE #5110: Add Support for a new Global Parameter to Query the Sizes of QUIC_STATISTICS_V2

A new MsQuic API global GetParam should be added that returns the well-known sizes of each version of the QUIC_STATISTICS_V2 struct.

### Proposed solution

Add the new parameter. Support GetParam only. Return hard coded sizes for the various API version sizes.

### Additional context

_No response_

nibanks:
## Description

Add a new global parameter that returns the supported sizes of the QUIC_STATISTICS_V2 struct. Fixes #5110.

## Testing

Updated. CI/CD

## Documentation

Updated


mtfriesen (dismissed):

docs/api/GetParam.md
```diff
@@ -71,6 +71,36 @@ Sample of double-call:
     }
 ```
 
+# Special Parameters
+
+## QUIC_PARAM_GLOBAL_STATISTICS_V2_SIZES
+
+Returns an array of well-known sizes (in bytes) for each version of the `QUIC_STATISTICS_V2` struct. This allows applications to determine the correct buffer size for statistics queries, even as new versions are added in future MsQuic releases.
```
guhetier, line 78: I think it would be useful to explain when this is needed: this isn't useful for an app that just want access to the stats, but only for a client that wants to forward the stats to a final consumer that will be aware of the new size.

For most app, getting the size directly from the header constant should be enough and they should not overthink it.

src/core/library.c
```diff
@@ -1498,6 +1498,33 @@ QuicLibraryGetGlobalParam(
...
+        if (MaxSizes == 0) {
+            *BufferLength = NumSizes * sizeof(uint32_t); // Indicate the max size.
+            Status = QUIC_STATUS_BUFFER_TOO_SMALL;
+            break;
+        }
+        if (Buffer == NULL) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+        uint32_t ToCopy = MaxSizes < NumSizes ? MaxSizes : NumSizes;
```
guhetier, line 1519: Would be better to use `min` for clarity.

src/test/lib/ApiTest.cpp
```diff
@@ -2787,6 +2787,52 @@ void QuicTestGlobalParam()
...
+            QUIC_STATUS_BUFFER_TOO_SMALL,
+            MsQuic->GetParam(
+                nullptr,
+                QUIC_PARAM_GLOBAL_STATISTICS_V2_SIZES,
+                &Length,
+                nullptr));
+        TEST_TRUE(Length >= sizeof(uint32_t));
+
+        //
+        // Second call: retrieve the sizes
```
guhetier, line 2809: IMO, the main risk of bug in this API is when we the size provided is smaller than the max (easy to overflow).
It'd be good to have a test case for a partial query then the full query.

src/core/library.c
nibanks: It's really the same thing, and I don't feel it matters all that much. I'd prefer to leave as is and not make another change.

guhetier: IMO, it is worth the effort now - code gets written once and read many times, and `min` is straighforward when a ternary forces the reader to think to now if this is a min / max / something else.

One click option :)
```suggestion
        uint32_t ToCopy = CXPLAT_MIN(MaxSizes, NumSizes);
```

src/test/lib/ApiTest.cpp
```diff
@@ -2831,6 +2844,30 @@ void QuicTestGlobalParam()
...
+        //
+        // Partial retrieve
+        //
+        Length = sizeof(uint32_t);
+        TEST_QUIC_SUCCEEDED(
+            MsQuic->GetParam(
+                nullptr,
+                QUIC_PARAM_GLOBAL_STATISTICS_V2_SIZES,
+                &Length,
+                Sizes));
```
guhetier, line 2857: nit: It'd be even better to pass a pointer to a single uint32 here, to catch an overflow in the copy (and not only the value of `Length` being too big).

guhetier (dismissed):
Some nitpicks but good to go otherwise.

guhetier (approved):

PR #5112, opened 2025-05-21
Allow Configuration of Stateless Retry Key

### LINKED ISSUE #5005: External DDoS Support for Retry Token Key Configuration

MsQuic supports Retry Token mechanism to validate client's address. To generate/validate Retry Token, a key is used as well to add entropy. Ask is to support a configuration path for these keys i.e. keys should be accepted as configuration from some external service. This external service will rotate and send new keys every X seconds.

### Proposed solution

This is needed so that Ddos solution and MsQuic solution can share keys. Idea is for ddos and msquic to work in conjunction and use same keys and same encryption APIs so that either can generate a token and both can validate it correctly.

### Additional context

_No response_

anrossi:
## Description

Add a new Global SetParam to set the Stateless Retry key, algorithm, and key rotation interval.
Fixes #5005.

## Testing

New tests added covering the invalid parameter cases.

## Documentation

Settings.md was updated to describe the new parameter.


src/test/lib/ApiTest.cpp
```diff
@@ -2773,6 +2773,78 @@ void QuicTestGlobalParam()
     }
 #endif
 
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
```
nibanks, line 2776: The stuff you added to msquic.h isn't preview (which is fine) so this doesn't need to be either.

src/core/library.c
```diff
@@ -2556,3 +2570,27 @@ MsQuicExecutionPoll(
 }
 
 #endif
+
+// Allows external services to set the retry key configuration for stateless retry tokens at runtime.
```
nibanks, line 2574: nit: Most of the comments in this PR don't follow our style.

src/core/library.c
```diff
@@ -2556,3 +2570,27 @@ MsQuicExecutionPoll(
...
+QuicLibrarySetRetryKeyConfig(
+    _In_ const QUIC_RETRY_KEY_CONFIG* Config
+    )
+{
+    if (Config == NULL || Config->BaseRetrySecret == NULL) {
+        return QUIC_STATUS_INVALID_PARAMETER;
+    }
+    // Validate AEAD algorithm and secret length.
+    size_t SecretLen = CxPlatKeyLength(Config->AeadAlgorithm);
+    if (SecretLen == 0) {
```
nibanks, line 2587: At least DBG assert `SecretLen` is <= `sizeof(MsQuicLib.BaseRetrySecret)`

src/core/library.c
```diff
@@ -1205,6 +1205,20 @@ QuicLibrarySetGlobalParam(
...
+    case QUIC_PARAM_GLOBAL_STATELESS_RETRY_CONFIG: {
+        if (Buffer == NULL || BufferLength != sizeof(QUIC_STATELESS_RETRY_CONFIG)) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+        const QUIC_STATELESS_RETRY_CONFIG* Config = (const QUIC_STATELESS_RETRY_CONFIG*)Buffer;
+        if (Config->Algorithm > CXPLAT_AEAD_CHACHA20_POLY1305 || Config->Secret == NULL || Config->RotationMs == 0) {
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
```
nibanks, line 1217: Move all this validation into `MsQuicLibrarySetRetryKeyConfig` instead.

src/core/library.c
```diff
@@ -2556,3 +2570,27 @@ MsQuicExecutionPoll(
...
+
+// Allows external services to set the retry key configuration for stateless retry tokens at runtime.
+// The BaseRetrySecret length must match the AEAD algorithm requirements.
+_IRQL_requires_max_(PASSIVE_LEVEL)
+QUIC_STATUS
+QuicLibrarySetRetryKeyConfig(
+    _In_ const QUIC_RETRY_KEY_CONFIG* Config
+    )
+{
+    if (Config == NULL || Config->BaseRetrySecret == NULL) {
```
nibanks, line 2582: SAL already requires `Config` to be non-NULL. Just DBG assert that instead of checking.

src/core/library.h
```diff
@@ -203,11 +203,26 @@ typedef struct QUIC_LIBRARY {
     _Field_size_(PartitionCount)
     QUIC_PARTITION* Partitions;
 
+    //
+    // Lock protecting the stateless retry configuration.
+    //
+    CXPLAT_DISPATCH_RW_LOCK StatelessRetryLock;
```
nibanks, line 209: I don't see you initializing/uninitializing this.

src/core/library.h
```diff
@@ -203,11 +203,26 @@ typedef struct QUIC_LIBRARY {
...
 
+    //
+    // The AEAD algorithm to use for the retry key.
+    //
+    CXPLAT_AEAD_TYPE RetryAeadAlgorithm;
+
+    //
+    // The number of milliseconds between key rotations.
+    //
+    uint32_t RetryKeyRotationMs;
```
nibanks, line 224: Are you defaulting this? Probably should be set to `QUIC_STATELESS_RETRY_KEY_LIFETIME_MS`.

src/inc/msquic.h
```diff
@@ -946,6 +946,16 @@ void
 #endif
 #define QUIC_PARAM_GLOBAL_TLS_PROVIDER                  0x0100000A  // QUIC_TLS_PROVIDER
 #define QUIC_PARAM_GLOBAL_STATELESS_RESET_KEY           0x0100000B  // uint8_t[] - Array size is QUIC_STATELESS_RESET_KEY_LENGTH
+#define QUIC_PARAM_GLOBAL_STATELESS_RETRY_CONFIG        0x0100000C  // QUIC_STATELESS_RETRY_CONFIG
```
nibanks, line 949: We also need to support registry based configuration. Do you plan to add that to this PR or a follow up?

src/inc/msquic.h
```diff
@@ -946,6 +946,16 @@ void
...
+
+//
+// Structure for QUIC_PARAM_GLOBAL_STATELESS_RETRY_CONFIG
+//
+typedef struct QUIC_STATELESS_RETRY_CONFIG {
+    CXPLAT_AEAD_TYPE Algorithm;      // AEAD algorithm for the key
+    uint32_t RotationMs;             // Key rotation interval in milliseconds
+    const uint8_t* Secret;           // Pointer to the base secret
+} QUIC_STATELESS_RETRY_CONFIG;
```
nibanks, line 958: I think most all other general types are declared at the top of the file.

src/test/lib/ApiTest.cpp
```diff
@@ -2773,6 +2773,78 @@ void QuicTestGlobalParam()
...
+                nullptr,
+                QUIC_PARAM_GLOBAL_STATELESS_RETRY_CONFIG,
+                sizeof(QUIC_STATELESS_RETRY_CONFIG),
+                nullptr));
+
+        // Wrong size
+        TEST_QUIC_STATUS(
+            QUIC_STATUS_INVALID_PARAMETER,
+            MsQuic->SetParam(
+                nullptr,
```
nibanks, line 2801: Need to not pass NULL for this to hit the size check.

src/inc/msquic.h
anrossi: I'll add it in this PR.

src/core/library.c
anrossi: I was torn on having it here, closest to the user, or in `QuicLibrarySetRetryKeyConfig`

src/core/library.c
anrossi: Hmm, CxPlatKeyLength will FRE assert before it returns 0. Maybe it's not necessary to check the result

src/inc/msquic.h
```diff
@@ -946,6 +946,16 @@ void
...
 #define QUIC_PARAM_GLOBAL_STATELESS_RESET_KEY           0x0100000B  // uint8_t[] - Array size is QUIC_STATELESS_RESET_KEY_LENGTH
+#define QUIC_PARAM_GLOBAL_STATELESS_RETRY_CONFIG        0x0100000C  // QUIC_STATELESS_RETRY_CONFIG
+
+//
+// Structure for QUIC_PARAM_GLOBAL_STATELESS_RETRY_CONFIG
+//
+typedef struct QUIC_STATELESS_RETRY_CONFIG {
+    CXPLAT_AEAD_TYPE Algorithm;      // AEAD algorithm for the key
+    uint32_t RotationMs;             // Key rotation interval in milliseconds
+    const uint8_t* Secret;           // Pointer to the base secret
```
anrossi, line 957: @nibanks
Is it fine for this to be a pointer (to caller memory), or should the struct have a fixed-sized array for the secret? Or a zero-length array at the end with the secret at the end?

I picked this way so there's fewer copies of the secret in memory (and fewer places the app has to zero out to protect the secret)

nibanks: Instead, I think we should have a variable length buffer at the end of the struct, so it's all contained. That allows the input length to be validated as well.

src/core/library.c
nibanks: All in `QuicLibrarySetRetryKeyConfig` keeps it in one place.

src/inc/msquic.h
anrossi: Zero-length arrays are non-standard, and we don't enable the compiler option to use them, so I made an array of size 1, and the caller is expected to allocate enough additional memory beyond the end of the struct to hold the entire secret.

nibanks: Well, we still use them 😄

src/core/library.c
```diff
@@ -334,6 +335,47 @@ QuicPerfCounterSnapShot(
...
+                Storage,
+                QUIC_SETTING_RETRY_KEY_ALGORITHM,
+                (uint8_t*)&RetryConfig.Algorithm,
+                &AlgLength)) &&
+        QUIC_SUCCEEDED(
+            CxPlatStorageReadValue(
+                Storage,
+                QUIC_SETTING_RETRY_KEY_SECRET,
+                Secret,
+                &RetryConfig.SecretLength))) {
```
nibanks, line 372: nit: the style in this project is more like this:
```suggestion
    if (QUIC_SUCCEEDED(
        CxPlatStorageReadValue(
            Storage,
            QUIC_SETTING_RETRY_KEY_ROTATION_MS,
            (uint8_t*)&RetryConfig.RotationMs,
            &RotationLength)) &&
        QUIC_SUCCEEDED(
        CxPlatStorageReadValue(
            Storage,
            QUIC_SETTING_RETRY_KEY_ALGORITHM,
            (uint8_t*)&RetryConfig.Algorithm,
            &AlgLength)) &&
        QUIC_SUCCEEDED(
        CxPlatStorageReadValue(
            Storage,
            QUIC_SETTING_RETRY_KEY_SECRET,
            Secret,
            &RetryConfig.SecretLength))) {
```

src/core/library.c
```diff
@@ -377,6 +419,8 @@ MsQuicLibraryReadSettings(
     QuicSettingsSetDefault(&MsQuicLib.Settings);
     if (MsQuicLib.Storage != NULL) {
         QuicSettingsLoad(&MsQuicLib.Settings, MsQuicLib.Storage);
+
+        QuicLibraryLoadRetryConfig(MsQuicLib.Storage);
```
nibanks, line 423:
```suggestion
        QuicLibraryLoadRetryConfig(MsQuicLib.Storage);
```

src/core/library.c
```diff
@@ -2556,3 +2619,48 @@ MsQuicExecutionPoll(
...
+    }
+    if (Config->RotationMs == 0) {
+        QuicTraceLogError(
+            LibrarySetRetryKeyRotationInvalid,
+            "[ lib] Invalid retry key rotation ms: %u.",
+            Config->RotationMs);
+        return QUIC_STATUS_INVALID_PARAMETER;
+    }
+
+    if (Config->Secret == NULL) {
```
nibanks, line 2644: This can't be NULL anymore, right?

src/core/library.c
```diff
@@ -2556,3 +2619,48 @@ MsQuicExecutionPoll(
...
+    }
+    if (Config->SecretLength < CxPlatKeyLength((CXPLAT_AEAD_TYPE)Config->Algorithm)) {
+        QuicTraceLogError(
+            LibrarySetRetryKeySecretLengthInvalid,
+            "[ lib] Invalid retry key secret length: %u. Expected %u.",
+            Config->SecretLength,
+            CxPlatKeyLength((CXPLAT_AEAD_TYPE)Config->Algorithm));
+        return QUIC_STATUS_INVALID_PARAMETER;
+    }
+    size_t SecretLen = CXPLAT_MIN(sizeof(MsQuicLib.BaseRetrySecret), Config->SecretLength);
```
nibanks, line 2658: Shouldn't this just be `CxPlatKeyLength((CXPLAT_AEAD_TYPE)Config->Algorithm)`?

src/core/library.c
```diff
@@ -2556,3 +2619,48 @@ MsQuicExecutionPoll(
...
+        return QUIC_STATUS_INVALID_PARAMETER;
+    }
+
+    if (Config->Secret == NULL) {
+        QuicTraceLogError(
+            LibrarySetRetryKeySecretNull,
+            "[ lib] Invalid retry key secret: NULL.");
+        return QUIC_STATUS_INVALID_PARAMETER;
+    }
+    if (Config->SecretLength < CxPlatKeyLength((CXPLAT_AEAD_TYPE)Config->Algorithm)) {
```
nibanks, line 2650: Shouldn't this just be `!=`?

src/core/partition.c
```diff
@@ -77,7 +77,7 @@ QuicPartitionUninitialize(
 _IRQL_requires_max_(DISPATCH_LEVEL)
 _Ret_maybenull_
 CXPLAT_KEY*
-QuicPartitioGetStatelessRetryKey(
+QuicPartitionGetStatelessRetryKey(
```
nibanks, line 80: Oops. 😄

src/core/partition.c
```diff
@@ -96,15 +96,15 @@ QuicPartitioGetStatelessRetryKey(
     CxPlatCopyMemory(
         RawKey,
         MsQuicLib.BaseRetrySecret,
-        sizeof(MsQuicLib.BaseRetrySecret));
+        CxPlatKeyLength(MsQuicLib.RetryAeadAlgorithm));
```
nibanks, line 99: Instead of recalculating length every time (I know it's just a switch statement), we should optimize for perf and save the length.

src/inc/msquic.h
```diff
@@ -875,6 +875,20 @@ typedef struct QUIC_STREAM_STATISTICS {
     uint64_t StreamBlockedByAppUs;
 } QUIC_STREAM_STATISTICS;
 
+typedef enum QUIC_AEAD_ALGORITHM_TYPE {
+    QUIC_AEAD_ALGORITHM_AES_128_GCM = 0,
+    QUIC_AEAD_ALGORITHM_AES_256_GCM = 1,
+    QUIC_AEAD_ALGORITHM_CHACHA20_POLY1305 = 2, // Not supported on Schannel
```
nibanks, line 881: Should we _fail_ then to set this type on Schannel?

anrossi: actually, in this case, we might be able to support this on Schannel; the only reason we don't support ChachaPoly with Schannel is because bcrypt doesn't allow us to set the counter value to a specific value, which is needed for the header protection.
But for the retry ticket, I don't think we're using a specific counter value, so it can start at 0 by default.

src/core/library.c
anrossi: I was thinking of the future change where we use a different key derivation algorithm. But I'll just update all this for that PR when it happens and use `!=` here.

src/test/lib/ApiTest.cpp
```diff
@@ -2773,6 +2773,93 @@ void QuicTestGlobalParam()
...
+                QUIC_PARAM_GLOBAL_STATELESS_RETRY_CONFIG,
+                sizeof(QUIC_STATELESS_RETRY_CONFIG),
+                Config.get()));
+        Config->RotationMs = 60000;
+
+        TEST_QUIC_SUCCEEDED(
```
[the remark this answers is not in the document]
anrossi: Our storage tests are in bad shape. First of all, the storage unittests are written for TAEF and aren't compiled at all.

src/core/library.c
```diff
@@ -413,6 +458,9 @@ MsQuicLibraryInitialize(
     MsQuicLib.ToeplitzHash.InputSize = CXPLAT_TOEPLITZ_INPUT_SIZE_QUIC;
     CxPlatToeplitzHashInitialize(&MsQuicLib.ToeplitzHash);
 
+    CxPlatDispatchRwLockInitialize(&MsQuicLib.StatelessRetryLock);
+    RetryLockInitialized = TRUE;
```
nibanks, line 462: There is no failure path between this and `PlatformInitialized = TRUE;`. Let's instead just use `PlatformInitialized` and set it here.

src/inc/msquic.h
```diff
@@ -875,6 +875,20 @@ typedef struct QUIC_STREAM_STATISTICS {
...
+    QUIC_AEAD_ALGORITHM_AES_256_GCM = 1,
+    QUIC_AEAD_ALGORITHM_CHACHA20_POLY1305 = 2,
+} QUIC_AEAD_ALGORITHM_TYPE;
+
+typedef struct QUIC_STATELESS_RETRY_CONFIG {
+    QUIC_AEAD_ALGORITHM_TYPE Algorithm; // AEAD algorithm for the key.
+    uint32_t RotationMs;                // Key rotation interval in milliseconds.
+    uint32_t SecretLength;              // Length of the secret.
+    _Field_size_bytes_(SecretLength)
+        uint8_t Secret[1];              // Secret to generate the key.
```
nibanks, line 889: You should use `[0]` because otherwise, it's breaking all your size validation checks in the core code. You are doing sizeof(struct) + SecretLength, which is now off-by-one.

src/core/library.c
```diff
@@ -196,6 +196,7 @@ QuicLibraryInitializePartitions(
     uint8_t ResetHashKey[20];
     CxPlatRandom(sizeof(ResetHashKey), ResetHashKey);
     CxPlatRandom(sizeof(MsQuicLib.BaseRetrySecret), MsQuicLib.BaseRetrySecret);
+    MsQuicLib.RetryKeyRotationMs = QUIC_STATELESS_RETRY_KEY_LIFETIME_MS;
```
nibanks, line 199: Should we be calling `QuicLibrarySetRetryKeyConfig` with some defaults?

src/inc/msquic.h
anrossi: That's the problem; zero-length array isn't a standard feature, and while we have it used internally (e.g. `QUIC_CID`) putting that in the official API header will now require every (C++) app including the header to enable a compiler option or disable a warning. 
From Bing:
> C++ does not officially support zero-length arrays as part of the standard. Declaring an array with a size of zero (e.g., int arr[0];) is not valid in standard C++ and will result in a compilation error.
However, some compilers (like GCC) may allow zero-length arrays as a non-standard extension, primarily for compatibility with C or specific use cases like struct padding or dynamic memory management. This behavior is not portable and should be avoided in standard-compliant C++ code.

I think this either needs to be a pointer to a buffer with a length, or a fixed-length buffer at the end of the struct.  I prefer the pointer solution for forward-compatibility.

nibanks: We disable other nonstandard extensions at the top of our file. For this, while I hear your feedback, I think we should use `[0]`.

src/core/library.h
```diff
@@ -203,11 +203,31 @@ typedef struct QUIC_LIBRARY {
     _Field_size_(PartitionCount)
     QUIC_PARTITION* Partitions;
 
+    //
+    // Lock protecting the stateless retry configuration.
+    //
+    CXPLAT_DISPATCH_RW_LOCK StatelessRetryLock;
```
guhetier, line 209: nit: Should we consider grouping all the members related to stateless retry in a struct to keep the library context cleaner?

src/inc/msquic.h
```diff
@@ -875,6 +875,23 @@ typedef struct QUIC_STREAM_STATISTICS {
...
+} QUIC_AEAD_ALGORITHM_TYPE;
+
+#pragma warning(push)
+#pragma warning(disable:4200)  // nonstandard extension used: zero-sized array in struct/union
+typedef struct QUIC_STATELESS_RETRY_CONFIG {
+    QUIC_AEAD_ALGORITHM_TYPE Algorithm; // AEAD algorithm for the key.
+    uint32_t RotationMs;                // Key rotation interval in milliseconds.
+    uint32_t SecretLength;              // Length of the secret.
+    _Field_size_bytes_(SecretLength)
+        uint8_t Secret[0];              // Secret to generate the key.
```
guhetier, line 891: I think this is the first zero-sized array we introduce in the API, which is significant (zero-sized array are error prone).
Other structure seems to be using a pointer to memory the user can put wherever they see fit (`QUIC_SCHANNEL_CREDENTIAL_ATTRIBUTE_W`).

We should stay consistent.

anrossi: I prefer your suggestion as well.  This being the first zero-length array in the header has caused some annoyance getting the compiler errors ignored.

guhetier: After the fact, I found one case: `uint16_t ProcessorList[1];`.

I still don't think we should introduce more of it (I'd rather consider fixing that one since it is still preview).
With a pointer, the user still can allocate everything in one memory block if they want to, and it is less error prone.

anrossi: The version settings use pointers and a count.  Given the push for standardization, I think that we may find more hurdles to zero-length arrays in the header (if not also the internal uses of it) in the future.

src/core/library.c
```diff
@@ -195,10 +195,22 @@ QuicLibraryInitializePartitions(
...
+    uint8_t RetrySecret[CXPLAT_AEAD_AES_256_GCM_SIZE];
+    CxPlatRandom(sizeof(RetrySecret), RetrySecret);
+
+    QUIC_STATELESS_RETRY_CONFIG RetryConfig;
+    RetryConfig.SecretLength = sizeof(RetrySecret);
+    RetryConfig.Secret = RetrySecret;
+    RetryConfig.RotationMs = QUIC_STATELESS_RETRY_KEY_LIFETIME_MS;
+    RetryConfig.Algorithm = QUIC_AEAD_ALGORITHM_AES_256_GCM;
+
+    QUIC_STATUS Status = QuicLibrarySetRetryKeyConfig(&RetryConfig);
```
nibanks, line 208: This can only fail if you pass invalid defaults (i.e. you fundamentally wrote the code above wrong), right? I'd actually FRE_ASSERT the result and don't add another error path.

src/test/lib/ApiTest.cpp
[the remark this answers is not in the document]
anrossi: Finally added tests.

src/inc/msquic.h
```diff
@@ -886,6 +886,23 @@ typedef struct QUIC_STREAM_STATISTICS {
...
 } QUIC_STREAM_STATISTICS;
 
+typedef enum QUIC_AEAD_ALGORITHM_TYPE {
+    QUIC_AEAD_ALGORITHM_AES_128_GCM = 0,
+    QUIC_AEAD_ALGORITHM_AES_256_GCM = 1,
+    QUIC_AEAD_ALGORITHM_CHACHA20_POLY1305 = 2,
+} QUIC_AEAD_ALGORITHM_TYPE;
+
+#pragma warning(push)
+#pragma warning(disable:4200)  // nonstandard extension used: zero-sized array in struct/union
```
nibanks: No longer needed

anrossi: Thanks for the reminder!

docs/Settings.md
```diff
@@ -120,6 +120,7 @@ These parameters are accessed by calling [GetParam](./api/GetParam.md) or [SetPa
 | `QUIC_PARAM_GLOBAL_STATELESS_RESET_KEY`<br> 11    | uint8_t[]               | Set-Only  | Globally change the stateless reset key for all subsequent connections.                               |
 | `QUIC_PARAM_GLOBAL_STATISTICS_V2_SIZES`<br> 12    | uint32_t[]               | Get-only  | Array of well-known sizes for each version of the QUIC_STATISTICS_V2 struct. The output array length is variable; pass a buffer of uint32_t and check BufferLength for the number of sizes returned. See GetParam documentation for usage details. |
 | `QUIC_PARAM_GLOBAL_VERSION_NEGOTIATION_ENABLED`<br> (preview) | uint8_t (BOOLEAN) | Both | Globally enable the version negotiation extension for all client and server connections. |
+| `QUIC_PARAM_GLOBAL_STATELESS_RETRY_CONFIG`<br> 12    | QUIC_STATELESS_RETRY_CONFIG | Set-Only | Configure the stateless retry token key, key algorithm, and key rotation interval. The secret length *must* match the AEAD algorithm key length. |
```
guhetier: Should the `QUIC_STATELESS_RETRY_CONFIG` type be documented too?

I see we haven't done it for other types in this list (at least `QUIC_GLOBAL_EXECUTION_CONFIG`), but I think we should.

src/core/library.c
```diff
@@ -195,7 +195,18 @@ QuicLibraryInitializePartitions(
...
+    uint8_t RetrySecret[CXPLAT_AEAD_AES_256_GCM_SIZE];
+    CxPlatRandom(sizeof(RetrySecret), RetrySecret);
+
+    QUIC_STATELESS_RETRY_CONFIG RetryConfig;
+    RetryConfig.SecretLength = sizeof(RetrySecret);
+    RetryConfig.Secret = RetrySecret;
+    RetryConfig.RotationMs = QUIC_STATELESS_RETRY_KEY_LIFETIME_MS;
+    RetryConfig.Algorithm = QUIC_AEAD_ALGORITHM_AES_256_GCM;
+
+    CXPLAT_FRE_ASSERT(QUIC_SUCCEEDED(QuicLibrarySetRetryKeyConfig(&RetryConfig)));
```
guhetier: question: Is code with side effects in a fre assert a pattern we already use?

I don't see it being used in a very fast scan, and IMO, we should not: code in an assert is generally treated as being there for validation only (even if that specific assert will always execute its content).
Unless this is largely used already, I think using a `Status` variable would be clearer.

src/core/library.c
```diff
@@ -334,6 +345,68 @@ QuicPerfCounterSnapShot(
         sizeof(PerfCounterSamples));
 }
 
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
```
guhetier, line 349: question: Should we allow an "incomplete" registry config, or should we validate that all 3 keys are there before using any of the values? For instance, is it ok to provide a secret only, without specifying the algorithm in the registry config?

src/core/partition.c
```diff
@@ -77,7 +77,7 @@ QuicPartitionUninitialize(
 _IRQL_requires_max_(DISPATCH_LEVEL)
 _Ret_maybenull_
 CXPLAT_KEY*
```
guhetier, line 87: Consider adding a comment about the locking requirement for this function

src/core/library.c
```diff
@@ -1525,6 +1614,38 @@ QuicLibraryGetGlobalParam(
         break;
     }
 
+    case QUIC_PARAM_GLOBAL_STATELESS_RETRY_CONFIG: {
+#ifdef DEBUG
```
guhetier, line 1631: Please add a comment explaining why this is only in debug builds.

Also, does it really need to?
MsQuic callers can already access the values if they really want to, it is their memory.

src/core/library.h
```diff
@@ -203,11 +203,31 @@ typedef struct QUIC_LIBRARY {
...
+
     //
     // The base secret used to generate keys for the stateless retry token.
     //
     uint8_t BaseRetrySecret[CXPLAT_AEAD_AES_256_GCM_SIZE];
```
guhetier: Now that there can be multiple types of key, we should define a `CXPLAT_AEAD_MAX_SIZE` that is guaranteed to be big enough for all supported key types.

src/core/library.c
```diff
@@ -2583,3 +2704,56 @@ MsQuicExecutionPoll(
...
+        return QUIC_STATUS_INVALID_PARAMETER;
+    }
+    if (Config->RotationMs == 0) {
+        QuicTraceLogError(
+            LibrarySetRetryKeyRotationInvalid,
+            "[ lib] Invalid retry key rotation ms: %u.",
+            Config->RotationMs);
+        return QUIC_STATUS_INVALID_PARAMETER;
+    }
+    if (Config->SecretLength != CxPlatKeyLength((CXPLAT_AEAD_TYPE)Config->Algorithm)) {
```
guhetier: nit: For clarity, instead of calling `CxPlatKeyLength` 3 times, lets store the key length in a variable.

src/core/library.c
```diff
@@ -2583,3 +2704,56 @@ MsQuicExecutionPoll(
...
+            CxPlatKeyLength((CXPLAT_AEAD_TYPE)Config->Algorithm));
+        return QUIC_STATUS_INVALID_PARAMETER;
+    }
+    uint16_t SecretLen = CxPlatKeyLength((CXPLAT_AEAD_TYPE)Config->Algorithm);
+    CXPLAT_DBG_ASSERT(SecretLen <= sizeof(MsQuicLib.BaseRetrySecret));
+
+    CxPlatDispatchRwLockAcquireExclusive(&MsQuicLib.StatelessRetryLock, PrevIrql);
+    if (Config->Secret != MsQuicLib.BaseRetrySecret) {
+        CxPlatCopyMemory(MsQuicLib.BaseRetrySecret, Config->Secret, SecretLen);
+        MsQuicLib.RetrySecretLength = SecretLen;
```
guhetier: I think you should either update the length even if the buffer is the same, or not update the algorithm if the key is the same.
Here, if I provide the function the same (big enough buffer), but with an algo that requires a smaller length (and I set the length right), we end up in an invalid config.

src/inc/msquic.h
```diff
@@ -886,6 +886,20 @@ typedef struct QUIC_STREAM_STATISTICS {
     uint64_t StreamBlockedByAppUs;
 } QUIC_STREAM_STATISTICS;
 
+typedef enum QUIC_AEAD_ALGORITHM_TYPE {
```
guhetier, line 889: Do we need to mark all the changes in the header as preview?

src/platform/storage_winuser.c
```diff
@@ -348,5 +348,116 @@ CxPlatStorageClear(
     _In_ CXPLAT_STORAGE* Storage
     )
 {
-    return HRESULT_FROM_WIN32(RegDeleteTreeA(Storage->RegKey, NULL));
```
guhetier: Was this incorrect? What is the problem with it?

src/platform/storage_winuser.c
```diff
@@ -348,5 +348,116 @@ CxPlatStorageClear(
...
+
+        while (Error == ERROR_MORE_DATA && AllocatedLength < 32767) {
+            //
+            // Reallocate buffer only if current buffer is too small
+            //
+            CXPLAT_DBG_ASSERT(ValueName != NULL);
+
+            CXPLAT_FREE(ValueName, QUIC_POOL_PLATFORM_TMP_ALLOC);
+
+            AllocatedLength *= 2; // The API doesn't tell us the needed length; just double it.
```
guhetier: This is very unfortunate :(

src/platform/storage_winuser.c
```diff
@@ -348,5 +348,116 @@ CxPlatStorageClear(
     _In_ CXPLAT_STORAGE* Storage
     )
 {
-    return HRESULT_FROM_WIN32(RegDeleteTreeA(Storage->RegKey, NULL));
+    QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
+    DWORD Error = NO_ERROR;
+    DWORD AllocatedLength = 30;
```
guhetier: We have some MsQuic setting keys that are more than 30 characters long. We should make this 255 so that at least for valid MsQuic values, we never need to loop.

src/test/bin/quic_gtest.cpp
```diff
@@ -2501,7 +2501,23 @@ TEST(Basic, TestVersionStorage) {
         QuicTestVersionStorage();
     }
 }
-#endif
+#endif // QUIC_API_ENABLE_PREVIEW_FEATURES
+
+#if  defined(DEBUG) // This test needs a GetParam API that is only available in debug builds.
```
guhetier:
```suggestion
#ifdef DEBUG // This test needs a GetParam API that is only available in debug builds.
```

src/test/lib/ApiTest.cpp
```diff
@@ -2773,6 +2773,103 @@ void QuicTestGlobalParam()
...
 #endif
 
+    //
+    // QUIC_PARAM_GLOBAL_STATELESS_RETRY_CONFIG
+    //
+    {
+        TestScopeLogger LogScope0("QUIC_PARAM_GLOBAL_STATELESS_RETRY_CONFIG");
+        const uint32_t SecretLength = 32;
+        const auto BufferLength = sizeof(QUIC_STATELESS_RETRY_CONFIG) + SecretLength;
+        UniquePtr<uint8_t[]> Buffer (new (std::nothrow) uint8_t[BufferLength]);
```
guhetier: nit: I would suggest to use a on-stack QUIC_STATELESS_RETRY_CONFIG and only allocate the buffer for the secret

- Cleaner from an alignment point of view
- Test the more complex config (no assumption over the shape of the buffer we could still have in the code)

... actually, you can even have the buffer as an on-stack array, it is only 32 bytes :)

src/test/lib/ApiTest.cpp
```diff
@@ -2773,6 +2773,103 @@ void QuicTestGlobalParam()
...
+        TEST_QUIC_STATUS(
+            QUIC_STATUS_INVALID_PARAMETER,
+            MsQuic->SetParam(
+                nullptr,
+                QUIC_PARAM_GLOBAL_STATELESS_RETRY_CONFIG,
+                sizeof(QUIC_STATELESS_RETRY_CONFIG) - 1,
+                Config));
+
+        // Invalid algorithm
+        Config->Algorithm = (QUIC_AEAD_ALGORITHM_TYPE)1000;
```
guhetier: Because you added this test case... did you try a negative number?
Enum signedness is compiler dependent and I have seen bugs where a negative value would allow to bypass parameter validation (because, as in this PR, we generally only check the upper bound)

src/test/lib/ApiTest.cpp
```diff
@@ -6530,3 +6627,225 @@ QuicTestValidateConnectionPoolCreate()
...
+    uint32_t TestRotationMs = 54321;
+    QUIC_AEAD_ALGORITHM_TYPE TestAlg = QUIC_AEAD_ALGORITHM_AES_128_GCM;
+    MsQuicRegistration Registration("TestRetryConfigSetting");
+    {
+        QuicStorageSettingScopeGuard GlobalSettings = QuicStorageSettingScopeGuard::Create();
+
+        //
+        // Test that the defaults were correctly picked up.
+        //
+        TEST_QUIC_SUCCEEDED(
```
guhetier, line 6670: nit: It'd be nice to test the case where no buffer / a buffer too short is provided, to check we get the length as expected

src/test/lib/ApiTest.cpp
```diff
@@ -6530,3 +6627,225 @@ QuicTestValidateConnectionPoolCreate()
...
+            CxPlatStorageWriteValue(
+                GlobalSettings,
+                QUIC_SETTING_RETRY_KEY_ROTATION_MS,
+                CXPLAT_STORAGE_TYPE_UINT32,
+                sizeof(TestRotationMs),
+                (uint8_t*)&TestRotationMs));
+
+        CxPlatSleep(100);
```
guhetier: This makes me affraid of test flakyness.
In similar situations, I have used a "TryUntil" wrapper, that read in a loop with a short sleep time (100ms), re-try until a condition is acheived, or that a long timeout (2s) elapsed.

It is pretty readable in C++ (make the condition a lambda) and you get the best of:
- a fast test execution
- no flakiness

src/core/library.c
csujedihy: We have a few instances where we add ref in an assert but it does look a little weird.

src/core/library.c
anrossi: Suppose the admin wants to use the default algorithm and rotation interval, but use a chosen key. Or they want to use a random key, but a shorter timeout period.  Nick suggested we allow this, as opposed to requiring setting all three before any take effect.

src/core/library.c
anrossi: I'll use a status variable.

src/core/library.c
anrossi: When it comes to setting encryption secrets, they should be write-only. This is a test hook, which is why it is only provided in debug builds.
Yes, the memory is in the application's address space, but it doesn't need to be read by the application, and it's best not to leave it available for accidental (or malicious) use.

src/inc/msquic.h
anrossi: these ones unfortunately don't get the preview treatment; we gotta get it right this time.

src/platform/storage_winuser.c
anrossi: When using the scope guards on the global registry, this would delete all the per-app settings keys as well, which is undesired and inconsistent with how kernel mode works (kernel mode only deletes the values, not the entire tree of keys and values).

src/platform/storage_winuser.c
anrossi: Agreed. This API is completely backwards: updates the name length on success, and but not on insufficient memory. (but the optional length values are set on failure)

src/platform/storage_winuser.c
anrossi: According to the documentation, registry names can be up to 32,767 bytes long 😮

src/core/library.c
anrossi: Hmm, I need to think carefully about how to do this then.
The point of `Config->Secret != MsQuicLib.BaseRetrySecret` was to provide a "no-change" check so when the registry change notification changes the non-Secret members, Secret is set to a valid "default" which doesn't get copied.
See `QuicLibraryLoadRetryConfig` in library.c

docs/Settings.md
```diff
@@ -120,6 +120,7 @@ These parameters are accessed by calling [GetParam](./api/GetParam.md) or [SetPa
 | `QUIC_PARAM_GLOBAL_STATELESS_RESET_KEY`<br> 11    | uint8_t[]               | Set-Only  | Globally change the stateless reset key for all subsequent connections.                               |
 | `QUIC_PARAM_GLOBAL_STATISTICS_V2_SIZES`<br> 12    | uint32_t[]               | Get-only  | Array of well-known sizes for each version of the QUIC_STATISTICS_V2 struct. The output array length is variable; pass a buffer of uint32_t and check BufferLength for the number of sizes returned. See GetParam documentation for usage details. |
 | `QUIC_PARAM_GLOBAL_VERSION_NEGOTIATION_ENABLED`<br> (preview) | uint8_t (BOOLEAN) | Both | Globally enable the version negotiation extension for all client and server connections. |
+| `QUIC_PARAM_GLOBAL_STATELESS_RETRY_CONFIG`<br> 12    | [QUIC_STATELESS_RETRY_CONFIG](./api/QUIC_STATELESS_RETRY_CONFIG.md) | Set-Only | Configure the stateless retry token key, key algorithm, and key rotation interval. The secret length *must* match the AEAD algorithm key length. |
```
nibanks: The value is the same as `QUIC_PARAM_GLOBAL_STATISTICS_V2_SIZES`

src/inc/quic_crypt.h
```diff
@@ -66,11 +66,19 @@ typedef enum CXPLAT_AEAD_TYPE {
...
+    "CXPLAT AEAD algorithm enum values must match the QUIC API enum values.");
+
 typedef enum CXPLAT_AEAD_TYPE_SIZE {
 
     CXPLAT_AEAD_AES_128_GCM_SIZE       = 16,
     CXPLAT_AEAD_AES_256_GCM_SIZE       = 32,
-    CXPLAT_AEAD_CHACHA20_POLY1305_SIZE = 32
+    CXPLAT_AEAD_CHACHA20_POLY1305_SIZE = 32,
+
+    CXPLAT_AEAD_MAX_SIZE               = 32
```
nibanks: Put a commend indicating this should be a `max` of all values above?

src/test/lib/ApiTest.cpp
nibanks: I agree that you might want retry here. Also 100 ms might not be enough based on prior experience.

src/core/partition.c
```diff
@@ -74,10 +74,16 @@ QuicPartitionUninitialize(
     CxPlatHashFree(Partition->ResetTokenHash);
 }
 
+//
+// MUST be called while holding the per-partition StatelessRetryKeysLock to
```
nibanks, line 78: You can annotate this requirement with SAL

src/platform/storage_winuser.c
guhetier: I'm not sure I follow, isn't it desirable to wipe both values and keys when clearing the storage? Why leave something behind? Isn't the kernel mode the one that needs to be fixed?

src/core/library.c
guhetier: I agree this bit is tricky. I have a hard time convincing myself that every combination of input parameters will result in a consistent state, if you start going in adding then removing values in the registry (like, set a short secret + length + algo. Then delete the secret, set a long length and compatible algo. I think our current code would accept that, but we would use a secret that is not properly random.)

I am starting to think that if settings are changed and don't contain a key, we should always re-generate one. Re-using the same secret with different length / algos leaves room for problems.

src/platform/storage_winuser.c
anrossi: The way the storage is architected, global settings are under the `HKLM\CCS\Services\MsQuic\Parameters` key, while application-specific settings are under the `HKLM\CCS\Services\MsQuic\Parameters\Apps\AppName` key.
Subkeys aren't used or parsed below theses specific keys.
So when user mode `CxPlatStorageClear` is called on the global storage path, it also deletes **all** per-app settings, including those unrelated to the test.  In kernel mode, only the values under the given key are deleted, which achieves the desired goal.

src/test/lib/ApiTest.cpp
anrossi: Is the idea that the lambda would just poll the setting until the desired output is observed?

docs/Settings.md
anrossi: Whoops. I thought I fixed that when I merged main

src/platform/storage_winuser.c
guhetier: Ok, but then this function becomes more a "clear storage partially"...
I am also not sure I get why it is a problem to clear per-app parameters. This function is meant for test code, why would we want to preserve per-app settings?

src/test/lib/ApiTest.cpp
guhetier: Yes, something along the lines of:

```
template <class P>
bool TryUntil(uint32 retryMs, uint32 timeoutMs, P condition) {
    uint32 nb_try = timeoutMs / retryMs + 1;
    for (uint32_t i = 0; i < nb_try; ++i) {
        bool success = condition();
        if (success) { return true; }
        sleep(retryMs);
    }
    return false;
}
```

And you use it like:

```
TEST_IS_TRUE(TryUntil(100, 2000, [&]{ return QUIC_SUCCEEDED(CxPlatStorageWriteValue(...)); });
```

Maybe with a richer return type than a boolean to help with understanding why the test fail.

src/platform/storage_winuser.c
anrossi: Let me reframe the question: should clearing the global storage also clear ALL per-app storage at the same time? Or should clearing the global storage only clear global settings, and clearing a given per-app storage only clears that app's settings?
I've been favoring the latter, and it sounds like you prefer the former.

guhetier: The way I see it:
- the "storage" module has no concept of global / per app. So it makes sense that "clear" clears everything, that what you expect of a generic clear function
- we use this function in test code only. and we should leave a clean state after a test to make sure we don't impact another test. So clearing everything makes sense here.

Is there a situation you think we need to preserve per-app settings in tests?

anrossi: `CxPlatStorageClear` takes a `CXPLAT_STORAGE` handle and so I was taking that to mean that it should only clear the settings in the handle passed. If it took no parameter, then I would interpret it to clear all storage regardless of global or per-app.

guhetier: Oh, you have a point, I was thinking of the `CXPLAT_STORAGE` as the registry key without realizing that a CXPLAT_STORAGE could be a subkey of the base key for another CXPLAT_STORAGE.

The current version makes sense then, thanks for the explanation :)

src/core/library.c
anrossi: Maybe it would just be simpler to require all three set via registry before accepting any of them? This is how the `SetParam` interface works (must set all three) and it avoid these combinatorial problems.

src/test/lib/ApiTest.cpp
anrossi: I've implemented this because I agree with the utility, but not all of these test cases work in this manner. Some of them test that nothing has changed, so there's nothing to poll for, but we still need to wait until some time has passed for the registry notification to fire and be processed.  I'm not sure there's something I can poll for to determine if the registry has been read and processed in that instance.

src/core/library.c
anrossi: Here's what I'll do: I'll allow the rotation interval to be set independently, but the secret and the algorithm must be set together.

src/core/quicdef.h
```diff
@@ -709,4 +709,4 @@ CXPLAT_STATIC_ASSERT(
 
 #define QUIC_SETTING_RETRY_KEY_ALGORITHM            "RetryKeyAlgorithm"
 #define QUIC_SETTING_RETRY_KEY_SECRET               "RetrySecret"
-#define QUIC_SETTING_RETRY_KEY_ROTATION_MS          "RetryKeyRotationMs"
+#define QUIC_SETTING_RETRY_KEY_ROTATION_MS          "StatelessRetry.KeyRotationMs"
```
guhetier: Is this intentional? The dot seems unusual + not done for the other key names.

src/core/library.c
```diff
@@ -396,6 +397,11 @@ QuicLibraryLoadRetryConfig(
             QUIC_SETTING_RETRY_KEY_SECRET,
             Secret,
             &SecretLength))) {
+        SecretChanged = TRUE;
+    }
+
+    if (SecretChanged && AlgorithmChanged) {
```
guhetier, line 404:
```suggestion
    //
    // If only the secret of the algorithm is specified in the registry, we ignore it: both must be present.
    //
    if (SecretChanged && AlgorithmChanged) {
```

guhetier (dismissed):

docs/api/QUIC_STATELESS_RETRY_CONFIG.md
```diff
@@ -0,0 +1,44 @@
...
+} QUIC_STATELESS_RETRY_CONFIG;
+```
+
+# Members
+
+`Algorithm`
+
+The AEAD algorithm used for protecting the retry token. Must be one of the following constants:
+
+Constant |  Key Length
```
sm-msft: Should this be "key size" to match the description of "SecretLength" below?

docs/api/QUIC_STATELESS_RETRY_CONFIG.md
```diff
@@ -0,0 +1,44 @@
...
+**QUIC_AEAD_ALGORITHM_AES_128_GCM**<br> 0 | 16
+**QUIC_AEAD_ALGORITHM_AES_256_GCM**<br>1<br> *The default* | 32
+**QUIC_AEAD_ALGORITHM_CHACHA20_POLY1305**<br>2 | 32
+
+`RotationMs`
+
+The interval to rotate the retry key. 30,000ms is the default.
+
+`SecretLength`
+
```
sm-msft, line 34: SecretLengthBytes to disambiguate between the bits used in algorithm name and the bytes required to be passed here?

docs/api/QUIC_STATELESS_RETRY_CONFIG.md
```diff
@@ -0,0 +1,44 @@
...
+Constant |  Key Length
+---------|------------
+**QUIC_AEAD_ALGORITHM_AES_128_GCM**<br> 0 | 16
+**QUIC_AEAD_ALGORITHM_AES_256_GCM**<br>1<br> *The default* | 32
+**QUIC_AEAD_ALGORITHM_CHACHA20_POLY1305**<br>2 | 32
+
+`RotationMs`
+
+The interval to rotate the retry key. 30,000ms is the default.
```
sm-msft: I am just curious about the default value here - 30 second key rotation seems too frequent. Does this stem from requirements or something we do in practice?

src/core/partition.c
```diff
@@ -144,15 +157,19 @@ QuicPartitionGetStatelessRetryKeyForTimestamp(
     )
 {
     const int64_t Now = CxPlatTimeEpochMs64();
```
sm-msft, line 159: Unrelated to your PR, but this code is susceptible to issues arising from system clock rollback. This should use CxPlatTimeUs64() instead of CxPlatTimeEpochMs64.

This applies to all functions that access QuicPartitionGetCurrentStatelessRetryKey() function.

src/core/library.c
```diff
@@ -2583,3 +2718,57 @@ MsQuicExecutionPoll(
...
+    uint16_t AlgSecretLen = CxPlatKeyLength((CXPLAT_AEAD_TYPE)Config->Algorithm);
+    if (Config->SecretLength != AlgSecretLen) {
+        QuicTraceLogError(
+            LibrarySetRetryKeySecretLengthInvalid,
+            "[ lib] Invalid retry key secret length: %u. Expected %u.",
+            Config->SecretLength,
+            AlgSecretLen);
+        return QUIC_STATUS_INVALID_PARAMETER;
+    }
+    CXPLAT_DBG_ASSERT(AlgSecretLen <= sizeof(MsQuicLib.StatelessRetry.BaseSecret));
```
sm-msft, line 2761: We could make this FRE assert due to potential implication of the condition not being met.

src/core/library.c
```diff
@@ -2583,3 +2718,57 @@ MsQuicExecutionPoll(
...
+
+    CxPlatDispatchRwLockAcquireExclusive(&MsQuicLib.StatelessRetry.Lock, PrevIrql);
+    if (Config->Secret != MsQuicLib.StatelessRetry.BaseSecret) {
```
sm-msft, line 2764: Can we return an error under this condition?

src/core/partition.c
```diff
@@ -144,15 +157,19 @@ QuicPartitionGetStatelessRetryKeyForTimestamp(
...
-    const int64_t CurrentKeyIndex = Now / QUIC_STATELESS_RETRY_KEY_LIFETIME_MS;
-    const int64_t KeyIndex = Timestamp / QUIC_STATELESS_RETRY_KEY_LIFETIME_MS;
+    CxPlatDispatchRwLockAcquireShared(&MsQuicLib.StatelessRetry.Lock, PrevIrql);
+    const int64_t CurrentKeyIndex = Now / MsQuicLib.StatelessRetry.KeyRotationMs;
```
sm-msft, line 161: How would this indexing be impacted if the settings change a small key rotation interval to a large key rotation interval?

src/core/unittest/main.h
```diff
@@ -7,6 +7,7 @@
 
 #pragma once
 
+#define CXPLAT_STORAGE_ENABLE_WRITE_SUPPORT
```
sm-msft, line 10: 👍

src/core/library.c
```diff
@@ -334,6 +345,78 @@ QuicPerfCounterSnapShot(
         sizeof(PerfCounterSamples));
 }
 
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+QuicLibraryLoadRetryConfig(
```
sm-msft, line 350: Msg'd Anthony separately about this function.

guhetier: For the sake of the review, it'd be awesome to have a quick summary of what the discussion was about and what the conclusion was :)

src/core/partition.c
nibanks: This time is required to be synchronized across all load balanced machines to allow for distributed encryption and decryption.

src/platform/storage_winuser.c
```diff
@@ -348,5 +348,116 @@ CxPlatStorageClear(
...
+                Storage->RegKey,
+                0, // Always use index 0 since we delete as we go
+                ValueName,
+                &NameLength,
+                NULL,   // Reserved
+                NULL,   // Type
+                NULL,   // Data
+                NULL);  // DataLength
+
+        while (Error == ERROR_MORE_DATA && AllocatedLength < 32767) {
```
guhetier: nit: While looking at #5182, it made me realize we should use a named constant for 32767, especially since it is used multiple times in this function. Don't we have `SHORT_MAX` somewhere?

src/core/library.c
sm-msft: Discussed in today's meeting. The follow-up was non-trivial for Anthony. The conclusion was that we should be ok with this usage pattern.

src/core/library.c
```diff
@@ -2583,3 +2718,57 @@ MsQuicExecutionPoll(
...
+    }
+    if (Config->Algorithm > QUIC_AEAD_ALGORITHM_CHACHA20_POLY1305 ||
+        Config->Algorithm < QUIC_AEAD_ALGORITHM_AES_128_GCM) {
+        QuicTraceLogError(
+            LibrarySetRetryKeyAlgorithmInvalid,
+            "[ lib] Invalid retry key algorithm: %d.",
+            Config->Algorithm);
+        return QUIC_STATUS_INVALID_PARAMETER;
+    }
+    if (Config->RotationMs == 0) {
```
sm-msft, line 2745: Should we limit the maximum rotation duration as well?

src/core/partition.c
sm-msft: Read the code some more. Change in key rotation interval will cause the indexing will change and it will impact whether or not we can retrieve a key. When key rotation interval is changed, we also expect a new key to be supplied. 

Even if a key is retrieved after this, it is unlikely that it would be able to decrypt the cyphertext.

No further comments from me on this - please feel free to resolve.

src/core/partition.c
```diff
@@ -95,16 +103,16 @@ QuicPartitioGetStatelessRetryKey(
     uint8_t RawKey[CXPLAT_AEAD_AES_256_GCM_SIZE];
     CxPlatCopyMemory(
         RawKey,
-        MsQuicLib.BaseRetrySecret,
-        sizeof(MsQuicLib.BaseRetrySecret));
+        MsQuicLib.StatelessRetry.BaseSecret,
+        MsQuicLib.StatelessRetry.SecretLength);
     for (size_t i = 0; i < sizeof(KeyIndex); ++i) {
```
sm-msft, line 108: Assuming relatively constant key rotation setting, it seems like we potentially use a given BaseSecret in this path for as long as it is left untouched.

In the decode (QuicPartitionGetStatelessRetryKeyForTimestamp()) path, we seem to be attempting to use the key for 2x key rotation time duration.

Please ack if this is accurate.

src/core/partition.c
sm-msft: Given that we are permitting key usage for decryption only in a 2x key rotation window and the expected key rotation, we seem to have a pretty fail-safe behavior even if the clock were to somehow jump suddenly.

Resolving this comment thread.

docs/api/QUIC_STATELESS_RETRY_CONFIG.md
sm-msft: Discussed the rotation interval offline and 30s is fine.

Only upshot here is that all the servers behind a loadbalancer must have their clocks synchronized to within this window to support this feature. There could be some failing corner cases if one of the servers has a clock in the future, which could be an additional caveat as well.

I think we need to call out the limitations in the documentation something to this effect:

- All servers deployed in a cluster and sharing the secret keys must have their clocks synchronized within RotationMs of UTC.
- A server whose clock is ahead of UTC may produce encrypted data that other servers in that deployment are unable to decrypt.

src/core/library.c
```diff
@@ -334,6 +345,78 @@ QuicPerfCounterSnapShot(
         sizeof(PerfCounterSamples));
 }
 
+_IRQL_requires_max_(PASSIVE_LEVEL)
+void
+QuicLibraryLoadRetryConfig(
+    _In_ CXPLAT_STORAGE* Storage
+    )
+{
```
sm-msft, line 353: If this is the regular registry callback, we can expect calls for each registry change - adding/updating a value and deleting a value. If we are expecting N values to be written to by an external agent, we would get N callbacks effectively etc. 

We need to ensure that our code below is idempotent, and we need some test cases to cover that.

We are also reading the values in a particular order - we have to ensure the agent updates them in the same order, so that we can read in consistent set of values rather than a split/mismatching set of values. 

Perhaps requiring local tickcount64 or some other counter to be saved in the registry to help ensure consistency?

sm-msft: There are various ways to handle this. A potential workaround would be to encode the 3 setting values into a string REG_SZ value - perhaps separated by semicolons - and read in a single value read. This should limit the notification to change in that value (addition/removal) and would also keep any value read in self-consistent.

docs/api/QUIC_STATELESS_RETRY_CONFIG.md
anrossi: I've updated the documentation to specify this is the length in bytes.

docs/api/QUIC_STATELESS_RETRY_CONFIG.md
anrossi: Updated to use length in both places

docs/api/QUIC_STATELESS_RETRY_CONFIG.md
anrossi: I've updated the documentation to call this out.

src/core/library.c
anrossi: It could, but a DBG assert is sufficient in this case because the length value comes from our code, not the caller, so this assert is to catch developers who did not update the `CXPLAT_AEAD_SIZE_MAX` definition.

src/core/library.c
anrossi: This is not an error condition; this is a sentinel value to allow the storage change callback, `QuicLibraryLoadRetryConfig` to update the `RotationMs` without updating the Secret/Algorithm.

src/platform/storage_winuser.c
anrossi: I was not particularly creative in searching for the constant to use here. It's `MAXSHORT` instead of `MAX_SHORT` or `SHORT_MAX` which I tried and then moved on.
Since this change will come in via a different PR; I'll just revert the change in this file.

src/core/library.c
anrossi: The RFC doesn't make any specific claims other than to say
> Servers SHOULD ensure that tokens sent in Retry packets are only accepted for a short time, as they are returned immediately by clients.

I've documented that the `RotationMs` should be set for a short period, but I'd rather not pick and enforce an arbitrary upper limit.

src/core/library.c
anrossi: I already have tests for updating the registry configuration out of order of the read order and ensure that the settings are only applied when a valid configuration is present.

guhetier: I think that Sarath may have a point here.

The rotation time value can probably be independent. But if I currently have a key KeyA for AlgoA, and I want to go to KeyB with AlgoB, and AlgoA and AlgoB have the same key length, for a bit, KeyA and AlgoB or KeyB and AlgoA could be used together.

Is it a problem if it happens?

src/core/partition.c
anrossi: I don't fully understand what you're saying in the first sentence, but the decode path will use the current or previous key (2x RotationMs) to handle a retry token, yes that is correct.

src/core/library.c
anrossi: Changing any of the values--RotationMs, Algorithm, Secret--will invalidate all previously issued tokens. In your examples, there's a window of a few milliseconds where tokens could get issued with an algorithm or key that is transitory, and those tokens won't validate. There's low-no risk using the same base key material to generate a key for another algorithm.
If there's concern (say, there's been a break in AlgoA and they're moving to AlgoB), deployers can delete the reg existing values so when they put the new values, it only picks up the AlgoB + KeyB pair, and there's no chance for a split config.
I added more documentation about this.

mtfriesen: The registry supports transactions for consistency, and we could leverage that here too, which has some benefits vs cooking up our own string formats and writing parsing code.

guhetier: Awesome, I'm happy there is no need for a custom string format.
I also didn't know the registry supported transaction, I learned something :)

guhetier (dismissed):
LGTM

docs/api/QUIC_STATELESS_RETRY_CONFIG.md
```diff
@@ -0,0 +1,52 @@
...
+A non-NULL pointer to a buffer containing `SecretLength` bytes of randomness. Used to generate the keys protecting the retry token.
+
+# Remarks
+
+`RotationMs` should be kept to a short interval, less than a minute, as retry tokens are returned immediately by clients.
+Changing `RotationMs`, `Algorithm`, or `Secret` will invalidate all retry tokens issued prior to the change.
+All servers deployed in a cluster and sharing the secret must have their clocks synchronized within `RotationMs` of UTC.
+A server whose clock is ahead of UTC may produce a retry token that other servers in that deployment are unable to validate.
+
+If configuring via the registry
```
sm-msft: Incomplete sentence?

anrossi: ah, yeah. The registry talk is in Settings.md
I've removed this fragment.

src/core/library.c
sm-msft: The testing here would be insufficient to test all possible race conditions between registry modifications and change notification handling. I think requiring ordering of operations should be **good enough** with the understanding that transient encryption state can exist.

Regarding registry transactions (or even ordering of changes) - we are placing this requirement on whoever is implementing the code that rotates the keys - it will be ignored even if documented, unless we write a tool for it.

Custom formatted string values in registry are widely used, transactional and it is orthogonal to using registry transactions. 

Please use "MUST perform.." in the documentation - if you expect deleting to happen before addition.

sm-msft: Please add a comment in the code to explain the expectation as well (or a reference to the documentation). It will be hard to understand the intent at a later time.

src/core/partition.c
sm-msft: Discussed offline. 

Summary: "If the keys in the registry do not get rotated, the current base secret will be used indefinitely and it will be rotated using the fixed "KeyIndex" pattern." 

The key lifetime ix 2x RotationMs.

This was existing behavior before the current change - just fyi to the readers.

src/core/library.c
anrossi: I've added documentation to this effect and a comment.

src/inc/quic_crypt.h
```diff
@@ -66,11 +66,18 @@ typedef enum CXPLAT_AEAD_TYPE {
...
+    (uint32_t)CXPLAT_AEAD_AES_128_GCM == (uint32_t)QUIC_AEAD_ALGORITHM_AES_128_GCM &&
+    (uint32_t)CXPLAT_AEAD_AES_256_GCM == (uint32_t)QUIC_AEAD_ALGORITHM_AES_256_GCM,
+    "CXPLAT AEAD algorithm enum values must match the QUIC API enum values.");
+
 typedef enum CXPLAT_AEAD_TYPE_SIZE {
 
     CXPLAT_AEAD_AES_128_GCM_SIZE       = 16,
     CXPLAT_AEAD_AES_256_GCM_SIZE       = 32,
-    CXPLAT_AEAD_CHACHA20_POLY1305_SIZE = 32
+    CXPLAT_AEAD_CHACHA20_POLY1305_SIZE = 32,
```
sm-msft, line 78: Nit: should we remove this reference to chacha poly?

anrossi: No, because `CXPLAT_AEAD_CHACHA20_POLY1305` is still defined above.  The `CXPLAT_AEAD_TYPE` is used for more than the retry key algorithm.

sm-msft:
Looks good to me.

sm-msft:
(Pls ignore the re-review. Hit the wrong button)

PR #5114, opened 2025-05-22
Add versioning support for Connection resumption ticket

sm-msft:
**Update**: Given the various unrelated changes that have accumulated in this PR, I will split this into two PRs and abandon this one.

The crux of the changes are in this PR: https://github.com/microsoft/msquic/pull/5131
Will post a link to the second PR when it is ready.

Addresses the issue #5115
 
## Description

This change adds the ability to version the connection resumption ticket on the server and the ability to filter out desirable versions of the tickets on the client.

- Min and max resumption ticket versions are hardcoded to 1 and 2 by default
- Add several missing tests for existing settings and add storage platform API to support these tests
- Check if the received resumption ticket is in a valid range on the client
- V2 resumption ticket includes a 64 byte fixed size buffer for usage in the near future
- NULL ptr checks in CxPlatStorageReadValue
- Other minor test updates

## Testing

Unit tests have been updated and expanded to test this change.

## Documentation

TBD

### DIFF
--- a/docs/BUILD.md
+++ b/docs/BUILD.md
@@ -156,7 +156,7 @@ For more info, take a look at the [build.ps1](../scripts/build.ps1) script.
 By default the build output will go in the `build` folder and the final build binaries in the `artifacts` folder. Under that it will create per-platform folders with subfolders for architecture/tls combinations. This allows for building different platforms and configurations at the same time.
 
 ## Updating Clog Sidecar
-Some code changes such as adding/updating new Quic traces require updating the Clog sidecar for successful Linux builds. This is done by running the following command:
+Some code changes such as adding/updating new MsQuic traces require updating the CLOG sidecar for successful Linux builds. This is done by running the following command:
 ```PowerShell
 ./scripts/update-sidecar.ps1
 ```
--- a/docs/BUILD.md
+++ b/docs/BUILD.md
@@ -155,6 +155,13 @@ For more info, take a look at the [build.ps1](../scripts/build.ps1) script.
 
 By default the build output will go in the `build` folder and the final build binaries in the `artifacts` folder. Under that it will create per-platform folders with subfolders for architecture/tls combinations. This allows for building different platforms and configurations at the same time.
 
+## Updating Clog Sidecar
+Some code changes such as adding/updating new Quic traces require updating the Clog sidecar for successful Linux builds. This is done by running the following command:
+```PowerShell
+./scripts/update-sidecar.ps1
+```
+This makes any necessary updates to the clog sidecar manifest and the generated files under `src/generated/` folder. The modified files must be committed along with the rest of the code changes for addressing the Linux build failures.
+
 # Building with CMake
 
 The following section details how to build MsQuic purely with CMake commands.
--- a/src/core/crypto.c
+++ b/src/core/crypto.c
@@ -2144,34 +2144,22 @@ QuicCryptoUpdateKeyPhase(
     PacketSpace->CurrentKeyPhaseBytesSent = 0;
 }
 
-uint8_t
-QuicGetOutgoingResumptionTicketVersion(
-    _In_opt_ QUIC_CONNECTION* Connection
+BOOLEAN
+IsQuicIncomingResumptionTicketSupported(
+    _In_ QUIC_VAR_INT TicketVersion
     )
 {
-
-    CXPLAT_DBG_ASSERT(QuicVarIntSize(CXPLAT_TLS_RESUMPTION_TICKET_VERSION) == sizeof(Connection->Settings.ResumptionTicketMaxVersion));
-    CXPLAT_STATIC_ASSERT(sizeof(uint8_t) == sizeof(Connection->Settings.ResumptionTicketMaxVersion), "Resumption ticket version setting field must be uint8_t");
-
-    if (Connection == NULL) {
-        return (uint8_t)CXPLAT_TLS_RESUMPTION_TICKET_VERSION;
-    }
-
-    return Connection->Settings.ResumptionTicketMaxVersion;
-}
-
-BOOLEAN
-IsQuicIncomingResumptionTicketSupported(_In_ QUIC_CONNECTION* Connection, QUIC_VAR_INT TicketVersion)
-{
-    if (TicketVersion >= Connection->Settings.ResumptionTicketMinVersion &&
-        TicketVersion <= Connection->Settings.ResumptionTicketMaxVersion) {
+    if (TicketVersion >= CXPLAT_TLS_RESUMPTION_TICKET_VERSION &&
+        TicketVersion <= CXPLAT_TLS_RESUMPTION_TICKET_MAX_VERSION) {
         return TRUE;
     }
 
     return FALSE;
 }
 
+//
 // Server calls this function to generate the resumption ticket for a specific client
+//
 QUIC_STATUS
 QuicCryptoEncodeServerTicket(
     _In_opt_ QUIC_CONNECTION* Connection,
@@ -2235,6 +2223,7 @@ QuicCryptoEncodeServerTicket(
         QuicVarIntSize(AppDataLength) +
         AlpnLength +
         EncodedTPLength +
+        RESUMPTION_TICKET_V2_EXTENSION_LENGTH +
         AppDataLength);
 
     TicketBuffer = CXPLAT_ALLOC_NONPAGED(TotalTicketLength, QUIC_POOL_SERVER_CRYPTO_TICKET);
@@ -2261,7 +2250,7 @@ QuicCryptoEncodeServerTicket(
     //
 
     _Analysis_assume_(sizeof(*TicketBuffer) >= 8);
-    uint8_t* TicketCursor = QuicVarIntEncode(QuicGetOutgoingResumptionTicketVersion(Connection), TicketBuffer);
+    uint8_t* TicketCursor = QuicVarIntEncode(CXPLAT_TLS_RESUMPTION_TICKET_MAX_VERSION, TicketBuffer);
     CxPlatCopyMemory(TicketCursor, &QuicVersion, sizeof(QuicVersion));
     TicketCursor += sizeof(QuicVersion);
     TicketCursor = QuicVarIntEncode(AlpnLength, TicketCursor);
@@ -2271,6 +2260,12 @@ QuicCryptoEncodeServerTicket(
     TicketCursor += AlpnLength;
     CxPlatCopyMemory(TicketCursor, EncodedHSTP + CxPlatTlsTPHeaderSize, EncodedTPLength);
     TicketCursor += EncodedTPLength;
+
+    //
+    // TODO: add processing of V2 extension here
+    //
+    TicketCursor += RESUMPTION_TICKET_V2_EXTENSION_LENGTH; // Skip the fixed length V2 extension
+
     if (AppDataLength > 0) {
         CxPlatCopyMemory(TicketCursor, AppResumptionData, AppDataLength);
         TicketCursor += AppDataLength;
@@ -2291,7 +2286,9 @@ Error:
     return Status;
 }
 
+//
 // Server uses this function to decode the resumption ticket presented by the client
+//
 QUIC_STATUS
 QuicCryptoDecodeServerTicket(
     _In_ QUIC_CONNECTION* Connection,
@@ -2322,17 +2319,13 @@ QuicCryptoDecodeServerTicket(
         goto Error;
     }
 
-    if (!IsQuicIncomingResumptionTicketSupported(Connection, TicketVersion)) {
+    if (!IsQuicIncomingResumptionTicketSupported(TicketVersion)) {
         QuicTraceEvent(
             ConnError,
             "[conn][%p] ERROR, %s.",
             Connection,
             "Resumption Ticket version unsupported");
         goto Error;
-    }
-
-    if (TicketVersion == CXPLAT_TLS_RESUMPTION_TICKET_VERSION_V2) {
-        // Handle V2 ticket specific extensions
     }
 
     if (TicketLength < Offset + sizeof(uint32_t)) {
@@ -2425,6 +2418,25 @@ QuicCryptoDecodeServerTicket(
         goto Error;
     }
     Offset += (uint16_t)TPLength;
+
+    if (TicketVersion == CXPLAT_TLS_RESUMPTION_TICKET_VERSION_V2) {
+        //
+        // V2 resumption ticket has a fixed length extension.
+        //
+        if (TicketLength < Offset + RESUMPTION_TICKET_V2_EXTENSION_LENGTH) {
+            QuicTraceEvent(
+                ConnError,
+                "[conn][%p] ERROR, %s.",
+                Connection,
+                "Resumption Ticket too small for V2 extensions");
+            goto Error;
+        }
+
+        //
+        // TODO: add processing of V2 extension here
+        //
+        Offset += RESUMPTION_TICKET_V2_EXTENSION_LENGTH;
+    }
 
     if (TicketLength == Offset + AppTicketLength) {
         Status = QUIC_STATUS_SUCCESS;
--- a/src/core/crypto.c
+++ b/src/core/crypto.c
@@ -2145,11 +2145,13 @@ QuicCryptoUpdateKeyPhase(
 }
 
 uint8_t
-QuicGetOutgoingResumptionTicketVersion(_In_opt_ QUIC_CONNECTION* Connection)
+QuicGetOutgoingResumptionTicketVersion(
+    _In_opt_ QUIC_CONNECTION* Connection
+    )
 {
 
-    CXPLAT_FRE_ASSERT(QuicVarIntSize(CXPLAT_TLS_RESUMPTION_TICKET_VERSION) == sizeof(Connection->Settings.ResumptionTicketMaxVersion));
-    CXPLAT_FRE_ASSERT(sizeof(uint8_t) == sizeof(Connection->Settings.ResumptionTicketMaxVersion));
+    CXPLAT_DBG_ASSERT(QuicVarIntSize(CXPLAT_TLS_RESUMPTION_TICKET_VERSION) == sizeof(Connection->Settings.ResumptionTicketMaxVersion));
+    CXPLAT_STATIC_ASSERT(sizeof(uint8_t) == sizeof(Connection->Settings.ResumptionTicketMaxVersion), "Resumption ticket version setting field must be uint8_t");
 
     if (Connection == NULL) {
         return (uint8_t)CXPLAT_TLS_RESUMPTION_TICKET_VERSION;
--- a/src/core/crypto.c
+++ b/src/core/crypto.c
@@ -2144,6 +2144,31 @@ QuicCryptoUpdateKeyPhase(
     PacketSpace->CurrentKeyPhaseBytesSent = 0;
 }
 
+uint8_t
+QuicGetOutgoingResumptionTicketVersion(_In_opt_ QUIC_CONNECTION* Connection)
+{
+
+    CXPLAT_FRE_ASSERT(QuicVarIntSize(CXPLAT_TLS_RESUMPTION_TICKET_VERSION) == sizeof(Connection->Settings.ResumptionTicketMaxVersion));
+    CXPLAT_FRE_ASSERT(sizeof(uint8_t) == sizeof(Connection->Settings.ResumptionTicketMaxVersion));
+
+    if (Connection == NULL) {
+        return (uint8_t)CXPLAT_TLS_RESUMPTION_TICKET_VERSION;
+    }
+
+    return Connection->Settings.ResumptionTicketMaxVersion;
+}
+
+BOOLEAN
+IsQuicIncomingResumptionTicketSupported(_In_ QUIC_CONNECTION* Connection, QUIC_VAR_INT TicketVersion)
+{
+    if (TicketVersion >= Connection->Settings.ResumptionTicketMinVersion &&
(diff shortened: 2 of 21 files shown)


anrossi:
Why does the application need to set resumption ticket versions? The resumption ticket is opaque to the application, so the version of a ticket is for MsQuic's own benefit for parsing.

anrossi:
For careful resume, the application should control whether it wants to use careful resume, not which version of the ticket to send. If careful resume is enabled on a server, send a version 2 ticket. If a client receives a version 2 ticket, but doesn't have careful resume enabled, ignore the version 2 field, and just use the version 1 fields. 
The intent is much clearer than controlling the version of the ticket being sent/received to determine careful resume.

sm-msft:
> Why does the application need to set resumption ticket versions? The resumption ticket is opaque to the application, so the version of a ticket is for MsQuic's own benefit for parsing.

Part of the resumption ticket is visible to the client application and this versioning will help the client filter out tickets it cannot parse.

src/inc/msquic.h
```diff
@@ -838,6 +840,10 @@ typedef struct QUIC_SETTINGS {
     uint32_t StreamRecvWindowBidiRemoteDefault;
     uint32_t StreamRecvWindowUnidiDefault;
 
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+    uint8_t ResumptionTicketMinVersion;
+    uint8_t ResumptionTicketMaxVersion;
+#endif
```
nibanks, line 846: Why does the app need to control this?

src/core/crypto.c
```diff
@@ -2144,6 +2144,31 @@ QuicCryptoUpdateKeyPhase(
     PacketSpace->CurrentKeyPhaseBytesSent = 0;
 }
 
+uint8_t
+QuicGetOutgoingResumptionTicketVersion(_In_opt_ QUIC_CONNECTION* Connection)
+{
+
+    CXPLAT_FRE_ASSERT(QuicVarIntSize(CXPLAT_TLS_RESUMPTION_TICKET_VERSION) == sizeof(Connection->Settings.ResumptionTicketMaxVersion));
+    CXPLAT_FRE_ASSERT(sizeof(uint8_t) == sizeof(Connection->Settings.ResumptionTicketMaxVersion));
```
nibanks, line 2152: Use a static assert instead.

src/core/crypto.c
```diff
@@ -2144,6 +2144,31 @@ QuicCryptoUpdateKeyPhase(
     PacketSpace->CurrentKeyPhaseBytesSent = 0;
 }
 
+uint8_t
+QuicGetOutgoingResumptionTicketVersion(_In_opt_ QUIC_CONNECTION* Connection)
+{
+
+    CXPLAT_FRE_ASSERT(QuicVarIntSize(CXPLAT_TLS_RESUMPTION_TICKET_VERSION) == sizeof(Connection->Settings.ResumptionTicketMaxVersion));
```
nibanks, line 2151: Use dbg assert instead.

src/core/crypto.c
```diff
@@ -2144,6 +2144,31 @@ QuicCryptoUpdateKeyPhase(
     PacketSpace->CurrentKeyPhaseBytesSent = 0;
 }
 
+uint8_t
+QuicGetOutgoingResumptionTicketVersion(_In_opt_ QUIC_CONNECTION* Connection)
```
nibanks, line 2148: nit: please match out style

src/core/crypto.c
```diff
@@ -2233,7 +2258,7 @@ QuicCryptoEncodeServerTicket(
     //
 
     _Analysis_assume_(sizeof(*TicketBuffer) >= 8);
-    uint8_t* TicketCursor = QuicVarIntEncode(CXPLAT_TLS_RESUMPTION_TICKET_VERSION, TicketBuffer);
+    uint8_t* TicketCursor = QuicVarIntEncode(QuicGetOutgoingResumptionTicketVersion(Connection), TicketBuffer);
```
nibanks, line 2261: Why does this need to be different for each connection? Why can't it be a constant for the library?

src/core/unittest/SettingsTest.cpp
```diff
@@ -335,14 +339,218 @@ TEST(SettingsTest, StreamRecvWindowDefaultGetsOverridenByIndividualLimits)
...
+    ASSERT_EQ(s.SendBufferingEnabled, 1);
+    ASSERT_EQ(s.PacingEnabled, 1);
+    ASSERT_EQ(s.ResumptionTicketMinVersion, 40);
+
+    // But an unset field should be set to default
+    ASSERT_EQ(s.MigrationEnabled, QUIC_DEFAULT_MIGRATION_ENABLED);
+}
+
+
+#include <winreg.h>
```
nibanks, line 430: All this code is Widows specific.

csujedihy:
> > Why does the application need to set resumption ticket versions? The resumption ticket is opaque to the application, so the version of a ticket is for MsQuic's own benefit for parsing.
> 
> Part of the resumption ticket is visible to the client application and this versioning will help the client filter out tickets it cannot parse.

Looking at the code, it doesn't seem like this is the case. The ticket, where ticket version, TP and whatnot are encoded in QuicCryptoEncodeServerTicket and then passed to schannel/quictls to encrypt (QuicCryptoEncodeServerTicket->QuicCryptoProcessAppData->CxPlatTlsProcessData). I don't see any code that skips the first few bytes.

nibanks:
> > > Why does the application need to set resumption ticket versions? The resumption ticket is opaque to the application, so the version of a ticket is for MsQuic's own benefit for parsing.
> > 
> > 
> > Part of the resumption ticket is visible to the client application and this versioning will help the client filter out tickets it cannot parse.
> 
> Looking at the code, it doesn't seem like this is the case. The ticket, where ticket version, TP and whatnot are encoded in QuicCryptoEncodeServerTicket and then passed to schannel/quictls to encrypt (QuicCryptoEncodeServerTicket->QuicCryptoProcessAppData->CxPlatTlsProcessData). I don't see any code that skips the first few bytes.

The format/version of the ticket is not exposed to the app. They can have their own blob as a part of it, but we only give them back their part. They should have their own version, which is completely independent of MsQuic's.

anrossi:
> > Why does the application need to set resumption ticket versions? The resumption ticket is opaque to the application, so the version of a ticket is for MsQuic's own benefit for parsing.
> 
> Part of the resumption ticket is visible to the client application and this versioning will help the client filter out tickets it cannot parse.

The client shouldn't need to think about things like this. The client application, when it gets the resumption token from MsQuic, is supposed to treat it like an opaque string of bytes and save the ticket without modification for use later.  

If a client which supports careful resume receives a V1 resumption ticket from the server, that means it can't do careful resume, but it can still do resumption.
If a client which doesn't support careful resume (or has it turned off) receives a V2 resumption ticket from the server, it can't do careful resume, but it must still be able to do resumption.
We can't break resumption for careful resume.

csujedihy:
> > > > Why does the application need to set resumption ticket versions? The resumption ticket is opaque to the application, so the version of a ticket is for MsQuic's own benefit for parsing.
> 
> > > 
> 
> > > 
> 
> > > Part of the resumption ticket is visible to the client application and this versioning will help the client filter out tickets it cannot parse.
> 
> > 
> 
> > Looking at the code, it doesn't seem like this is the case. The ticket, where ticket version, TP and whatnot are encoded in QuicCryptoEncodeServerTicket and then passed to schannel/quictls to encrypt (QuicCryptoEncodeServerTicket->QuicCryptoProcessAppData->CxPlatTlsProcessData). I don't see any code that skips the first few bytes.
> 
> 
> 
> The format/version of the ticket is not exposed to the app. They can have their own blob as a part of it, but we only give them back their part. They should have their own version, which is completely independent of MsQuic's.

So, the metadata like ticket version, quic version and TP etc can actually be decrypted and be decoded if the client is also msquic? My bad. I thought it couldn't be even decrypted by the client.

sm-msft:
> Part of the resumption ticket is visible to the client application and this versioning will help the client filter out tickets it cannot parse.

Discussed this with Anthony - this is not the case here.

src/inc/msquic.h
sm-msft: There was some confusion on my end about the visibility of the ticket contents. Will address this in my next iteration.

src/core/crypto.c
sm-msft: The resumption ticket version settings can be dynamically configured through the registry. I believe this will be essential at least while we develop the feature and perhaps even going forward if we were to account for the new SChannel functionality being available in Srv2025 only.

This settings for a connection appear to be sticky for some time by means of reading them off of a copy on the connection object itself and this follows the same pattern.

docs/BUILD.md
```diff
@@ -155,6 +155,13 @@ For more info, take a look at the [build.ps1](../scripts/build.ps1) script.
 
 By default the build output will go in the `build` folder and the final build binaries in the `artifacts` folder. Under that it will create per-platform folders with subfolders for architecture/tls combinations. This allows for building different platforms and configurations at the same time.
 
+## Updating Clog Sidecar
+Some code changes such as adding/updating new Quic traces require updating the Clog sidecar for successful Linux builds. This is done by running the following command:
```
nibanks, line 159:
```suggestion
Some code changes such as adding/updating new MsQuic traces require updating the CLOG sidecar for successful Linux builds. This is done by running the following command:
```

src/core/crypto.c
nibanks: I still don't see a reason why we shouldn't always write just the latest version. It's simpler.

src/core/crypto.c
```diff
@@ -2144,6 +2144,34 @@ QuicCryptoUpdateKeyPhase(
...
+    if (Connection == NULL) {
+        return (uint8_t)CXPLAT_TLS_RESUMPTION_TICKET_VERSION;
+    }
+
+    return Connection->Settings.ResumptionTicketMaxVersion;
+}
+
+BOOLEAN
+IsQuicIncomingResumptionTicketSupported(_In_ QUIC_CONNECTION* Connection, QUIC_VAR_INT TicketVersion)
+{
```
nibanks, line 2165:
```suggestion
BOOLEAN
QuicIsIncomingResumptionTicketSupported(
    _In_ QUIC_CONNECTION* Connection,
    _In_ QUIC_VAR_INT TicketVersion
    )
{
```

src/core/crypto.c
```diff
@@ -2144,6 +2144,34 @@ QuicCryptoUpdateKeyPhase(
...
+    if (TicketVersion >= Connection->Settings.ResumptionTicketMinVersion &&
+        TicketVersion <= Connection->Settings.ResumptionTicketMaxVersion) {
+        return TRUE;
+    }
+
+    return FALSE;
+}
+
+// Server calls this function to generate the resumption ticket for a specific client
```
nibanks, line 2174:
```suggestion
//
// Server calls this function to generate the resumption ticket for a specific client
//
```

src/core/crypto.c
```diff
@@ -2301,6 +2331,10 @@ QuicCryptoDecodeServerTicket(
         goto Error;
     }
 
+    if (TicketVersion == CXPLAT_TLS_RESUMPTION_TICKET_VERSION_V2) {
+        // Handle V2 ticket specific extensions
+    }
```
nibanks, line 2336: Don't add empty if statements. Either remove it or implement it please.

src/core/quicdef.h
```diff
@@ -444,11 +444,22 @@ CXPLAT_STATIC_ASSERT(
...
 #define QUIC_DEFAULT_SERVER_RESUMPTION_LEVEL    QUIC_SERVER_NO_RESUME
 
+
 //
-// Version of the wire-format for resumption tickets.
-// This needs to be incremented for each change in order or count of fields.
+// Valid Resumption Ticket Versions - these must be contiguous
+//
+#define CXPLAT_TLS_RESUMPTION_TICKET_VERSION_V1   1
+#define CXPLAT_TLS_RESUMPTION_TICKET_VERSION_V2   2
```
nibanks, line 452: Do we have something to change in the ticket that requires a v2? Otherwise, let's **not** just add it for test reasons.

src/inc/msquic.h
```diff
@@ -837,7 +837,6 @@ typedef struct QUIC_SETTINGS {
     uint32_t StreamRecvWindowBidiLocalDefault;
     uint32_t StreamRecvWindowBidiRemoteDefault;
     uint32_t StreamRecvWindowUnidiDefault;
-
```
nibanks, line 840: nit: revert changes in here.

src/core/settings.h
```diff
@@ -65,7 +65,9 @@ typedef struct QUIC_SETTINGS_INTERNAL {
             uint64_t XdpEnabled                             : 1;
             uint64_t QTIPEnabled                            : 1;
             uint64_t RioEnabled                             : 1;
-            uint64_t RESERVED                               : 13;
+            uint64_t ResumptionTicketMinVersion             : 1;
+            uint64_t ResumptionTicketMaxVersion             : 1;
+            uint64_t RESERVED                               : 11;
```
nibanks, line 70: We shouldn't make this version configurable. I'd say remove all settings changes for this.

src/core/unittest/SpinFrame.cpp
```diff
@@ -75,7 +75,10 @@ TEST(SpinFrame, SpinFrame1000000)
         }
 
         do {
-            TEST_QUIC_SUCCEEDED(CxPlatRandom(sizeof(FrameType), &FrameType));
+            uint16_t randValue;
+            TEST_QUIC_SUCCEEDED(CxPlatRandom(sizeof(randValue), &randValue));
+            // map the randvalue to Frame type range
+            FrameType = (uint16_t)(randValue % (QUIC_FRAME_MAX_SUPPORTED + 1));
```
nibanks, line 81: We specifically want to test unsupported ranges. I don't think we need this change.

sm-msft: This test runs an indeterminately large amount of time without this fix. If we want to incorporate part of unsupported range, we can double the FrameType range here and get a test with an upper limit on runtime.

src/core/quicdef.h
sm-msft: I will be adding an extension field for Careful Resume that would require this versioning.

src/core/crypto.c
sm-msft: This is a placeholder in the draft PR. It will be populated in the upcoming iterations.

src/core/crypto.c
sm-msft: Missed this one...

src/core/unittest/SpinFrame.cpp
sm-msft: Also, there are unsupported ranges within the (0, QUIC_FRAME_MAX_SUPPORTED] range. Doubling this to (0, 2*QUIC_FRAME_MAX_SUPPORTED] should include a wider unsupported range.

src/core/crypto.c
sm-msft: Discussed this offline - we will now go with the fixed/latest version number when creating a ticket. The configurability was to address the scenario of mixed version deployments, which is not a potential concern now.

src/core/settings.c
```diff
@@ -2000,7 +2001,7 @@ QuicSettingsGetSettings(
         QUIC_SETTINGS* Settings
     )
 {
-    uint32_t MinimumSettingsSize = (uint32_t)CXPLAT_STRUCT_SIZE_THRU_FIELD(QUIC_SETTINGS, MtuDiscoveryMissingProbeCount);
+    uint32_t MinimumSettingsSize = (uint32_t)CXPLAT_STRUCT_SIZE_THRU_FIELD(QUIC_SETTINGS, StreamRecvWindowUnidiDefault);
```
nibanks, line 2004: Revert?

src/core/unittest/SettingsTest.cpp
```diff
@@ -289,7 +289,7 @@ TEST(SettingsTest, StreamRecvWindowDefaultDoesNotOverrideIndividualLimitsWhenSet
     ASSERT_EQ(Destination.StreamRecvWindowUnidiDefault, Original);
 }
 
-TEST(SettingsTest, StreamRecvWindowDefaultGetsOverridenByIndividualLimits)
+TEST(SettingsTest, CertainDefaultsGetOverridenByIndividualLimits)
```
nibanks, line 292: Why the rename?

src/core/unittest/SettingsTest.cpp
```diff
@@ -335,14 +335,283 @@ TEST(SettingsTest, StreamRecvWindowDefaultGetsOverridenByIndividualLimits)
...
+        QUIC_STATUS_SUCCESS,
+        CxPlatStorageOpen(
+            nullptr,
+            nullptr,
+            nullptr,
+            &Storage));
+
+    ResetMsQuicTestSettings(Storage);
+
+    if (SkipStorageTests) {
```
nibanks, line 459: Instead of making this a global, can we just have `ResetMsQuicTestSettings` return a `bool` to indicate if storage tests are supported?

src/core/unittest/SettingsTest.cpp
```diff
@@ -394,7 +663,7 @@ TEST(SettingsTest, SettingsSizesSet)
     QUIC_SETTINGS_INTERNAL InternalSettings;
     CxPlatZeroMemory(&InternalSettings, sizeof(InternalSettings));
 
-    uint32_t MinimumSettingsSize = (uint32_t)SETTINGS_SIZE_THRU_FIELD(QUIC_SETTINGS, MtuDiscoveryMissingProbeCount);
+    uint32_t MinimumSettingsSize = (uint32_t)SETTINGS_SIZE_THRU_FIELD(QUIC_SETTINGS, StreamRecvWindowUnidiDefault);
```
nibanks, line 666: Revert

src/test/lib/ApiTest.cpp
```diff
@@ -2969,7 +2969,7 @@ void QuicTestConfigurationParam()
                     QUIC_PARAM_CONFIGURATION_SETTINGS,
                     &Length,
                     nullptr));
-            TEST_EQUAL(Length, SETTINGS_SIZE_THRU_FIELD(QUIC_SETTINGS, MtuDiscoveryMissingProbeCount));
+            TEST_EQUAL(Length, SETTINGS_SIZE_THRU_FIELD(QUIC_SETTINGS, StreamRecvWindowUnidiDefault));
```
nibanks, line 2972: Revert?

src/core/settings.c
```diff
@@ -1819,7 +1820,7 @@ QuicSettingsSettingsToInternal(
     _Out_ QUIC_SETTINGS_INTERNAL* InternalSettings
     )
 {
-    if (!CXPLAT_STRUCT_HAS_FIELD(QUIC_SETTINGS, SettingsSize, MtuDiscoveryMissingProbeCount)) {
+    if (!CXPLAT_STRUCT_HAS_FIELD(QUIC_SETTINGS, SettingsSize, StreamRecvWindowUnidiDefault)) {
```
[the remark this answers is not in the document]
sm-msft: The existing code seems to be measuring the size of the struct incorrectly.  I have updated this to point to the last member of QUIC_SETTINGS (https://github.com/microsoft/msquic/blob/2e4a9f04f7407392b9444aa06fffb612d210d717/src/inc/msquic.h#L839)

src/core/unittest/SettingsTest.cpp
```diff
@@ -8,7 +8,7 @@
     Unit test for the settings logic.
 
 --*/
-
+#define QUIC_UNIT_TESTS
```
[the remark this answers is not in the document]
sm-msft: The settings tests need these API.

src/core/unittest/SettingsTest.cpp
sm-msft: We set this global when the first action in ResetMsQuicTestSettings() fails. Subsequent steps/actions could fail also, but it would not necessarily indicate that the storage was not available, but some other test issue. Returning a single bool from this function would not necessarily capture the complexity here.

src/core/unittest/SettingsTest.cpp
```diff
@@ -335,14 +335,283 @@ TEST(SettingsTest, StreamRecvWindowDefaultGetsOverridenByIndividualLimits)
...
+            "TEST",
+            &TempStore));
+
+    // Set up storage values for a few settings
+    ASSERT_EQ(
+        QUIC_STATUS_SUCCESS,
+        CxPlatStorageSaveUIntValue(
+            TempStore,
+            QUIC_SETTING_SEND_BUFFERING_DEFAULT,
+            1));
```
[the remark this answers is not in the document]
sm-msft: Some of these are courtesy copilot :). Will take a closer look.

src/core/settings.c
sm-msft: Downlevel tests are failing... got to revert this.

sm-msft:
Moved the resumption ticket related code into a separate PR: https://github.com/microsoft/msquic/pull/5131
Please continue the review there.

src/core/unittest/SettingsTest.cpp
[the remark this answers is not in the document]
sm-msft: Tests updated.

src/core/settings.c
sm-msft: Would it be correct to say that MtuDiscoveryMissingProbeCount is part of the earliest released QUIC_SETTINGS struct and our API contract requires us to accept these smaller sized structs? 

Is there reference documentation around this?

sm-msft:
This PR is superseded by the following PRs:
https://github.com/microsoft/msquic/pull/5131
https://github.com/microsoft/msquic/pull/5133

Closing this PR to declutter.

PR #5119, opened 2025-05-23
Added GetParam for Network Statistics

Johan511:
## Description

Currently Network Statistics can not be queried, it is only indicated by `QUIC_CONNECTION_EVENT`.

Should I user be forced to subscribe to all the network statistics update events? No
Also, we do not signal the event on every update

This PR proposes a method where the application can query for Network Statistics when it wants to using GetParam

## Testing

Added tests similar to `QuicTest_QUIC_PARAM_CONN_STATISTICS*`

## Documentation

Added documentation for the newly introduced `QUIC_PARAM_CONN_NETWORK_STATISTICS`


src/inc/msquic.h
```diff
@@ -1359,14 +1387,7 @@ typedef struct QUIC_CONNECTION_EVENT {
...
         } ONE_WAY_DELAY_NEGOTIATED;
-        struct {
-           uint32_t BytesInFlight;              // Bytes that were sent on the wire, but not yet acked
-           uint64_t PostedBytes;                // Total bytes queued, but not yet acked. These may contain sent bytes that may have potentially lost too.
-           uint64_t IdealBytes;                 // Ideal number of bytes required to be available to  avoid limiting throughput
-           uint64_t SmoothedRTT;                // Smoothed RTT value
-           uint32_t CongestionWindow;           // Congestion Window
-           uint64_t Bandwidth;                  // Estimated bandwidth
-        } NETWORK_STATISTICS;
+        NETWORK_STATISTICS_INTERNAL NETWORK_STATISTICS;
```
nibanks, line 1390: I'm not a big fan of doing it this way. Since this is the preview features define, we can still change it. So maybe we just have a `NETWORK_STATISTICS` as the payload of this local struct. I don't like all the extra stuff you had to put in the `NETWORK_STATISTICS_INTERNAL`. It's too messy.

Johan511: Made the necessary change

Johan511:
Requesting review on few things in particular
- All members of `NETWORK_STATISTICS` are now `uint64_t`, previously `BytesInFlight` and `CongestionWindow` were uint32_t. Why make the change? Congestion window has no bound of `UINT32_MAX` and would like to see all references to it changed to `uint64_t` in the future. The size of the struct still remains the same on most platforms (because padding), we weren't really saving space in the first place. 

- [This](https://github.com/Johan511/msquic//blob/2ffe127ce2ddd9c603ccaf8d9e786df666896571/src/inc/msquic.h#L1377) weirdness in naming scheme where type and the member are named the same. Please suggest an alternative name for the type if this is an issue.

- Please have a thorough look at `QuicConnGetNetworkStatistics(...)`, if in the future there are `NETWORK_STATISTICS` members which are supported by **only** one of Bbr or Cubic, we have no way to signal the user in the API (same issue exists also for `QUIC_CONNECTION_EVENT_NETWORK_STATISTICS`).

src/inc/msquic.h
```diff
@@ -1012,6 +1023,10 @@ typedef struct QUIC_SCHANNEL_CREDENTIAL_ATTRIBUTE_W {
 #define QUIC_PARAM_CONN_ORIG_DEST_CID                   0x05000018  // uint8_t[]
 #define QUIC_PARAM_CONN_SEND_DSCP                       0x05000019  // uint8_t
 
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+#define QUIC_PARAM_CONN_NETWORK_STATISTICS              0x05000020  // struct NETWORK_STATISTICS
+#endif
```
nibanks:
```suggestion
#define QUIC_PARAM_CONN_SEND_DSCP                       0x05000019  // uint8_t
#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
#define QUIC_PARAM_CONN_NETWORK_STATISTICS              0x05000020  // struct NETWORK_STATISTICS
#endif
```

src/inc/msquic.h
```diff
@@ -641,6 +641,17 @@ typedef struct QUIC_STATISTICS_V2 {
...
 } QUIC_STATISTICS_V2;
 
+typedef struct NETWORK_STATISTICS
+{
+    uint32_t BytesInFlight;                      // Bytes that were sent on the wire, but not yet acked
+    uint64_t PostedBytes;                        // Total bytes queued, but not yet acked. These may contain sent bytes that may have potentially lost too.
+    uint64_t IdealBytes;                         // Ideal number of bytes required to be available to  avoid limiting throughput
+    uint64_t SmoothedRTT;                        // Smoothed RTT value
+    uint32_t CongestionWindow;                   // Congestion Window
+    uint64_t Bandwidth;                          // Estimated bandwidth
```
nibanks:
```suggestion
    uint32_t BytesInFlight;              // Bytes that were sent on the wire, but not yet acked
    uint64_t PostedBytes;                // Total bytes queued, but not yet acked. These may contain sent bytes that may have potentially lost too.
    uint64_t IdealBytes;                 // Ideal number of bytes required to be available to  avoid limiting throughput
    uint64_t SmoothedRTT;                // Smoothed RTT value
    uint32_t CongestionWindow;           // Congestion Window
    uint64_t Bandwidth;                  // Estimated bandwidth
```

src/core/connection.c
```diff
@@ -6902,6 +6902,40 @@ QuicConnGetV2Statistics(
...
+
+    if (*StatsLength == 0) {
+        *StatsLength = sizeof(NETWORK_STATISTICS);
+        return QUIC_STATUS_BUFFER_TOO_SMALL;
+    }
+
+    if (*StatsLength < MinimumStatsSize) {
+        *StatsLength = MinimumStatsSize;
+        return QUIC_STATUS_BUFFER_TOO_SMALL;
+    }
```
nibanks:
```suggestion
    if (*StatsLength < sizeof(NETWORK_STATISTICS)) {
        *StatsLength = sizeof(NETWORK_STATISTICS);
        return QUIC_STATUS_BUFFER_TOO_SMALL;
    }
```

src/core/connection.c
```diff
@@ -6902,6 +6902,40 @@ QuicConnGetV2Statistics(
...
+
+    if (Stats == NULL) {
+        return QUIC_STATUS_INVALID_PARAMETER;
+    }
+
+    memset(Stats, 0, MinimumStatsSize);
+
+    Connection->CongestionControl.QuicCongestionControlGetNetworkStatistics(
+                        Connection, &Connection->CongestionControl, Stats);
```
nibanks:
```suggestion
    Connection->CongestionControl.QuicCongestionControlGetNetworkStatistics(
        Connection, &Connection->CongestionControl, Stats);
```

src/core/connection.c
```diff
@@ -6902,6 +6902,40 @@ QuicConnGetV2Statistics(
...
+    if (*StatsLength < MinimumStatsSize) {
+        *StatsLength = MinimumStatsSize;
+        return QUIC_STATUS_BUFFER_TOO_SMALL;
+    }
+
+    if (Stats == NULL) {
+        return QUIC_STATUS_INVALID_PARAMETER;
+    }
+
+    memset(Stats, 0, MinimumStatsSize);
```
nibanks: Please use CxPlatZeroMemory

nibanks (commented):
Looks pretty good. Just a few minor comments.

Johan511:
Made the requested changes

src/core/connection.c
Johan511: I had copied that snippet of code from [here](https://github.com/Johan511/msquic//blob/e104677251f7bc0b72fb23eb0fbf39f581fe0972/src/core/connection.c#L6814-L6824)

src/rs/lib.rs
```diff
@@ -327,6 +327,7 @@ pub const PARAM_CONN_VERSION_SETTINGS: u32 = 0x05000014;
 pub const PARAM_CONN_INITIAL_DCID_PREFIX: u32 = 0x05000015;
 pub const PARAM_CONN_STATISTICS_V2: u32 = 0x05000016;
 pub const PARAM_CONN_STATISTICS_V2_PLAT: u32 = 0x05000017;
+pub const PARAM_CONN_NETWORK_STATISTICS: u32 = 0x05000020;
```
youyuanwu, line 330: No need to add here. All these param numbers should be removed here (at some point). We migrated to use ffi mod generated numbers.

src/rs/ffi/win_bindings.rs
```diff
@@ -5807,50 +5852,6 @@ const _: () = {
 };
 #[repr(C)]
 #[derive(Debug, Copy, Clone)]
-pub struct QUIC_CONNECTION_EVENT__bindgen_ty_1__bindgen_ty_19 {
```
youyuanwu, line 5810: In the Rust wrappers u need to change the conversion function to access the newly named struct instead of the old anonymous struct.

youyuanwu:
Please run cargo build with "--features overwrite" to update rust bindings, the current code is not updated.
After that the conversion code should use as_ref() in `let ev = unsafe { value.__bindgen_anon_1.CONNECTED.as_ref() };` to access the additional layer introduced in the anonymous wrapper.

Johan511:
Hi @youyuanwu 
Thank you for the help
Pardon my ignorance, I haven't worked with the rust ecosystem

I have changed all the 

`let ev = unsafe { value.__bindgen_anon_1.*QUIC_CONNECTION_EVENT__bindgen_ty_1 variants*`
to
`let ev = unsafe { value.__bindgen_anon_1.*QUIC_CONNECTION_EVENT__bindgen_ty_1 variants*.as_ref()`

I still get one same error (rest are resolved)
```
error[E0412]: cannot find type `NETWORK_STATISTICS` in this scope
    --> src/rs/ffi/linux_bindings.rs:5335:49
     |
5335 |     pub NETWORK_STATISTICS: __BindgenUnionField<NETWORK_STATISTICS>,
     |                                                 ^^^^^^^^^^^^^^^^^^ not found in this scope
     |
help: you might be missing a type parameter
     |
5305 | pub struct QUIC_CONNECTION_EVENT__bindgen_ty_1<NETWORK_STATISTICS> {
     |                                               ++++++++++++++++++++

```

Do I need to explicitly register/declare `NETWORK_STATISTICS` somewhere?

Also, for `win_bindings.rs`, do I need to run cargo on a VM or is there someway I can run cargo with target_os as windows?

youyuanwu:
The bindings does not seem to be generated and checked in correctly.
The easy way is to comment out the conversion code that has the problem and let the github action run and compile successfully. It will post the bindings patch to the PR and you can apply it.

Usually I generate the bindings on linux and windows both locally.

Johan511:
@youyuanwu 
I had a look at `build.rs` and saw `allowlist_item` not taking the pattern for the struct I wanted to expose (`NETWORK_STATISTICS`)

I have tried the following diff in `build.rs` and as you mentioned `cargo build --features overwrite` works fine
```
-        .allowlist_item("QUIC.*|BOOLEAN|BYTE|HQUIC|HRESULT")
+        .allowlist_item("QUIC.*|BOOLEAN|BYTE|HQUIC|HRESULT|NETWORK_STATISTICS")
```

With this change it seems I dont need `as_ref()` either
Do you believe this is an appropriate change?

youyuanwu:
This looks good.
But in hindsight the struct should probably be named "QUIC_NETWORK_STATISTICS", to avoid name collisions.
@nibanks

Johan511:
> But in hindsight the struct should probably be named "QUIC_NETWORK_STATISTICS", to avoid name collisions.

I much prefer this, making the change

youyuanwu (approved):

src/core/connection.c
```diff
@@ -32,6 +32,7 @@
 --*/
 
 #include "precomp.h"
+#include "quic_platform.h"
```
nibanks: This shouldn't be necessary, right? I thought precomp already had this.

Johan511: My bad
My clangd setup aggressively adds includes
Fixed it

PR #5123, opened 2025-05-26
Patch prefix 'L' on unevaluated string literal having no effect causing an Error

snw7:
…ng error

## Description

## Testing

## Documentation

### DIFF
--- a/src/inc/quic_platform_posix.h
+++ b/src/inc/quic_platform_posix.h
@@ -292,7 +292,7 @@ CxPlatLogAssert(
     _In_z_ const char* Expr
     );
 
-#define CXPLAT_STATIC_ASSERT(X,Y) static_assert(X, Y);
+#define CXPLAT_STATIC_ASSERT(X,Y) static_assert(X, #Y);
 #define CXPLAT_ANALYSIS_ASSERT(X)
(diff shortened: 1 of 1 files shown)


guhetier:
Hi! Sorry for the delay.
Could you please clarify what was the issue with "L" prefixes? I don't see it used with `CXPLAT_STATIC_ASSERT` at any point. Is the problem that certain compilation options have `static_assert` expect a "L" prefixed string?

snw7:
> Hi! Sorry for the delay. Could you please clarify what was the issue with "L" prefixes? I don't see it used with `CXPLAT_STATIC_ASSERT` at any point. Is the problem that certain compilation options have `static_assert` expect a "L" prefixed string?

Hi Guillaume! Thanks for following up! The L prefix is required because, under certain compiler configurations (for example, when UNICODE is defined), the string literal used in static_assert is expected to be a wide string. Without the L prefix, the build fails on the toolchains due to a type mismatch in the diagnostic message literal.
Adding the prefix ensures the assertion compiles consistently across all supported build configurations and avoids failures on some platforms. I initially encountered the issues on my debian server distro :-)

guhetier:
Thanks for confirming, and for the contribution!

PR #5130, opened 2025-05-29
Fix DSCP detection on Windows. Add DSCP support to SecNetPerf

anrossi:
## Description

DSCP detection on Windows was incorrect and setting an invalid value, which would always error, causing MsQuic to determine DSCP to not be supported.  This change fixes that bug, and also adds support to SecNetPerf to set DSCP on outgoing packets. It also adds support for receiving DSCP to the Windows datapath layers, which enables the DSCP tests to function.

## Testing

Updated existing DSCP tests to validate DSCP is getting sent and received as expected.

## Documentation

SecNetPerf help text was updated to list the new option.


sm-msft:
Suggest updating the readme file with help and sample test run data: https://github.com/microsoft/msquic/blob/main/src/perf/readme.md

nibanks:
> TODO: Tests should be updated/verified to be checking for DSCP

Do you plan to fix this up soon? We need to get this into the release/2.5 branch.

anrossi:
> > TODO: Tests should be updated/verified to be checking for DSCP
> 
> Do you plan to fix this up soon? We need to get this into the release/2.5 branch.

Ah the problem is that, on receive, windows doesn't fetch the IP_TOS field, just ECN, so this test is failing because now it's actually testing DSCP, and we weren't fetching the DSCP values.
I'd added fetch support for DSCP, which should work downlevel since it has been around for a while.

src/platform/datapath_winuser.c
```diff
@@ -159,6 +159,7 @@ typedef struct DATAPATH_RX_IO_BLOCK {
         WSA_CMSG_SPACE(sizeof(IN6_PKTINFO)) +   // IP_PKTINFO
         WSA_CMSG_SPACE(sizeof(DWORD)) +         // UDP_COALESCED_INFO
         WSA_CMSG_SPACE(sizeof(INT)) +           // IP_ECN
+        WSA_CMSG_SPACE(sizeof(INT)) +           // IP_TOS
```
nibanks: Why do you keep the space for ECN?

anrossi: experiment to try to see if the recv is failing is because there's not enough space. It's not the reason.

src/platform/unittest/DataPathTest.cpp
```diff
@@ -801,7 +801,9 @@ TEST_P(DataPathTest, UdpData)
     UdpRecvContext RecvContext;
     CxPlatDataPath Datapath(&UdpRecvCallbacks);
     RecvContext.TtlSupported = Datapath.IsSupported(CXPLAT_DATAPATH_FEATURE_TTL);
-    RecvContext.DscpSupported = Datapath.IsSupported(CXPLAT_DATAPATH_FEATURE_SEND_DSCP);
+    RecvContext.DscpSupported =
+        Datapath.IsSupported(CXPLAT_DATAPATH_FEATURE_SEND_DSCP) &&
```
sm-msft: Could we turn this into a macro or a function in this test file to avoid repetition? It is easy to miss one of those conditions when writing a new test otherwise.

src/platform/datapath_winuser.c
```diff
@@ -3472,9 +3518,12 @@ CxPlatDataPathUdpRecvComplete(
                     LocalAddr->Ipv4.sin_port = SocketProc->Parent->LocalAddress.Ipv6.sin6_port;
                     LocalAddr->Ipv6.sin6_scope_id = PktInfo->ipi_ifindex;
                     FoundLocalAddr = TRUE;
+                } else if (CMsg->cmsg_type == IP_TOS) {
+                    TypeOfService = *(PINT)WSA_CMSG_DATA(CMsg);
+                    CXPLAT_DBG_ASSERT(TypeOfService < UINT8_MAX);
                 } else if (CMsg->cmsg_type == IP_ECN) {
-                    ECN = *(PINT)WSA_CMSG_DATA(CMsg);
-                    CXPLAT_DBG_ASSERT(ECN < UINT8_MAX);
+                    TypeOfService = *(PINT)WSA_CMSG_DATA(CMsg);
```
sm-msft, line 3525: Are we guaranteed to receive only one of these msg headers? It seems like we are requesting at least two of the three here in case DSCP is enabled...

anrossi: Since we don't request ECN at the same time as requesting TOS, yes, we're only going to receive one or the other, not both.

sm-msft: Ah.. the options have different names in IPv6 and IPv4 (TClass vs TOS)... We are good here.

src/platform/unittest/DataPathTest.cpp
anrossi: what if I change it to `Datapath.IsSupported(CXPLAT_DATAPATH_FEATURE_SEND_DSCP | CXPLAT_DATAPATH_FEATURE_RECV_DSCP)`?

src/platform/datapath_winuser.c
```diff
@@ -158,7 +158,7 @@ typedef struct DATAPATH_RX_IO_BLOCK {
         RIO_CMSG_BASE_SIZE +
         WSA_CMSG_SPACE(sizeof(IN6_PKTINFO)) +   // IP_PKTINFO
         WSA_CMSG_SPACE(sizeof(DWORD)) +         // UDP_COALESCED_INFO
-        WSA_CMSG_SPACE(sizeof(INT)) +           // IP_ECN
+        WSA_CMSG_SPACE(sizeof(INT)) +           // IP_ECN/IP_TOS
```
csujedihy: why do they share the same space?

anrossi: On platforms that don't support receiving TOS, we request ECN instead.  It saves 24 bytes for each receive instead of taking extra space which will never be used.

csujedihy: So, we never query both? We might need a comment for this. It's not very obvious.

anrossi: ok, I'll add a comment.

csujedihy (approved):

Issue #5115, opened 2025-05-23
Add support for bumping up connection resumption ticket version


sm-msft:
### Describe the feature you'd like supported

Connection resumption tickets currently use a fixed default version number. We need the ability to version these tickets in order to make future format changes/updates to them.

### Proposed solution

This version number is inserted by the MsQuic server when it generates the resumption ticket and validated by the client when it receives the said ticket - there is no handshake/negotiation around this version number.

Implementing this feature includes modifying the server and client to code to enable supporting one or more versions of these tickets.

This (sub)feature will help us extend the resumption ticket to incorporate data necessary for features such as "Careful Resume".

nibanks:
The tickets already have a version number they use internally. It's not clear to me what else on top of that is required.

PR #5131, opened 2025-05-29
Careful resume structs + encode/decode code

sm-msft:
Addresses these sub-feature requests 
- https://github.com/microsoft/msquic/issues/5115
- https://github.com/microsoft/msquic/issues/5173 

## Description

- Update the version on resumption tickets and added room to include additional data for Careful Resume in the future.
-- Uplevel server can accept v1 tickets but always issue v2 tickets
-- Updated ticket tests
- Define Careful Resume struct 
- Added encoding/decoding functionality for Careful Resume Structs
- Integrated Careful Resume encoding/decoding into existing custom app resumption ticket code path
- Added unit tests for the new code 

## Testing

Ran updated unit tests + running pipeline tests

## Documentation

N/A


src/core/crypto.c
```diff
@@ -2243,6 +2260,12 @@ QuicCryptoEncodeServerTicket(
     TicketCursor += AlpnLength;
     CxPlatCopyMemory(TicketCursor, EncodedHSTP + CxPlatTlsTPHeaderSize, EncodedTPLength);
     TicketCursor += EncodedTPLength;
+
+    //
+    // TODO: add processing of V2 extension here
+    //
+    TicketCursor += RESUMPTION_TICKET_V2_EXTENSION_LENGTH; // Skip the fixed length V2 extension
```
csujedihy, line 2267: I think we very likely need a variable length region for careful resume. The draft says `saved_remote_endpoint` should be part of the cc params. `saved_remote_endpoints` would have to be an IP address, which can be v4 or v6, which have different lengths.

sm-msft: There are additional complications with this field because quic allows a client's ip address to change (to some extent..). We may have to store multiple ip addresses here.

The good news is that this field is read only by the server and we have some flexibility in updating it at a later time with minimal versioning impact.

Any idea on how big a resumption ticket can currently get?

csujedihy: 1. The feature you are referring to is client migration where a quic connection can migrate to use a different network path (which typically has a different source IP address). We can't allow careful resume to happen when the network path has changed because the saved CC params do not make sense to the new path, e.g. switching from ethernet to satellite, the server should not resume with the cwnd saved from the ethernet path.
2. The server will not be able to know beforehand which IP address the client is going to migrate to at the moment the server is saving the CC params and issuing the session ticket. So, regardless, it's not feasible to store a set of client IP addresses.

So, the IP address actually has to be one AFACT.

For schannel, the app data size including whatever we package in msquic layer is 2KB.

src/core/crypto.c
```diff
@@ -2392,6 +2419,25 @@ QuicCryptoDecodeServerTicket(
     }
     Offset += (uint16_t)TPLength;
 
+    if (TicketVersion == CXPLAT_TLS_RESUMPTION_TICKET_VERSION_V2) {
+        //
+        // V2 resumption ticket has a fixed length extension.
```
anrossi: Is the Careful Resume standard close enough to finalization that you know you won't need more space? Or is this a placeholder for future parsing logic?

nibanks: I don't think we should take this change until we actually know what should go into the next version of the ticket.

sm-msft: We reached an agreement last week on what goes into the Careful Resume portion of the resumption ticket. However, that is somewhat orthogonal to this incremental change. Since Resumption tickets are consumed only on the server, changes to it do not impact the protocol beyond the size of the ticket.

What is being tested in this change is that we handle bumping up the version of this ticket on the server and that a larger ticket does not break clients.

sm-msft: The actual careful resume portion of the changes should take much less space than this and are upcoming.

src/core/crypto.c
```diff
@@ -2166,6 +2166,321 @@ QuicCryptoUpdateKeyPhase(
...
+            ConnError,
+            "[conn][%p] ERROR, %s.",
+            Connection,
+            "Unexpected address size");
+        CXPLAT_DBG_ASSERT(FALSE);
+        return;
+    }
+
+    uint8_t* TicketCursor = QuicVarIntEncode(AddrLen, Buffer);
+    TicketCursor = QuicVarIntEncode(CarefulResumeState->SmoothedRtt, TicketCursor);
```
csujedihy, line 2384: Why do we encode each individual member of `CarefulResumeState` into the ticket instead of just appending the whole careful resume struct into it with an encoded length like how we did it for TP and AppData?

The fields being encoded here do not always qualify as QuicVarInt. E.g. there are no restrictions on CWnd for being less than 62-bit max.

csujedihy: IMO, by doing that, you don't have to encode address separately. You can just use that `QUIC_ADDR` and use `QuicAddrCompareIp` to compare IP addresses.

src/core/unittest/TicketTest.cpp
```diff
@@ -578,8 +582,443 @@ TEST(ResumptionTicketTest, ServerEncDecNoAppData)
...
+                &DecodedCarefulResumeState,
+                &DecodedAppData,
+                &DecodedAppDataLength));
+
+        ASSERT_EQ((uint16_t)DecodedAppDataLength, 0);
+        ASSERT_EQ(DecodedAppData, nullptr);
+        CompareTransportParameters(&ServerTP, &DecodedServerTP);
+
+        // Validate CarefulResumeState fields (except port)
+        ASSERT_EQ(CarefulResumeState.SmoothedRtt, DecodedCarefulResumeState.SmoothedRtt) << kAlgorithms[i].Name;
```
csujedihy, line 677: I have not seen this syntax, ` << kAlgorithms[i].Name;`. What does it do?

sm-msft: This prints the variable (algorithm name in this case) along with the asserted condition in case of a failure.

src/core/crypto.c
sm-msft: 1. Regarding encoding ip addresses - your comment for an earlier iteration against a fixed length buffer was that ip address are of variable length. This is the exact opposite of that comment :-). We now have a variable length implementation here. I think I will leave this as-is for now. If we see better reasons, we can modify the packing.

2. Regarding serializing the struct as a buffer vs packing individual members of the struct - there are minor space gains to be had from packing individual members and I decided to go with that. This also makes minor versioning possible, if we nitpick about that.

3. Careful resume is a server-side thing and transport parameters seems to be a collection of various possible structs in various circumstances - the pack/unpack code there is getting very unwieldy. Not sure adding further to that makes it more readable.

4. Regarding size of CWND - I thought we agreed that CWND will be Uint32 and that is what I implemented here with. Also, can the CWND be larger than 2^32 at any point in time?

src/core/unittest/TicketTest.cpp
csujedihy: Ok, JFYI. I am not sure if we compile unittest in kernel mode. If we do, this won't be compiled.

src/core/crypto.c
csujedihy: 1. Sorry, the comment made in previous iteration was when I didn't know what the careful resume state struct would look like and thus didn't think too much about its impact. 
2. I agree you might squeeze a few bytes but the code becomes a lot more complicated, especially the IP address. I don't think serializing the struct stops minor versioning.
3. All I was suggesting was just encode the length of the careful resume state struct and dump the memory of the struct in the ticket. When retrieving, you can simply cast the buffer back to the struct.
4. CWnd is a bad example. Its 32bit. But the other fields are uint64, right? I don't think they will be bigger than 62-bit max in practice, but that's a limit we don't previously have elsewhere, isn't it?

src/core/unittest/TicketTest.cpp
```diff
@@ -578,8 +582,443 @@ TEST(ResumptionTicketTest, ServerEncDecNoAppData)
...
+        QUIC_TP_FLAG_INITIAL_MAX_STRM_DATA_BIDI_REMOTE |
+        QUIC_TP_FLAG_INITIAL_MAX_STRM_DATA_UNI |
+        QUIC_TP_FLAG_INITIAL_MAX_STRMS_BIDI |
+        QUIC_TP_FLAG_INITIAL_MAX_STRMS_UNI;
+    ServerTP.ActiveConnectionIdLimit = QUIC_TP_ACTIVE_CONNECTION_ID_LIMIT_MIN;
+
+    // Set IPv4 address 192.0.2.1 (do not set port)
+    CxPlatZeroMemory(&CarefulResumeState.RemoteEndpoint, sizeof(CarefulResumeState.RemoteEndpoint));
+    CarefulResumeState.RemoteEndpoint.Ipv4.sin_family = QUIC_ADDRESS_FAMILY_INET;
+    CarefulResumeState.RemoteEndpoint.Ipv4.sin_addr.s_addr = htonl(0xC0000201); // 192.0.2.1
```
csujedihy: Can we use `QuicAddrFromString` to set IP?

csujedihy: Same for v6 test case

src/core/unittest/TicketTest.cpp
```diff
@@ -578,8 +582,443 @@ TEST(ResumptionTicketTest, ServerEncDecNoAppData)
...
+        CompareTransportParameters(&ServerTP, &DecodedServerTP);
+
+        // Validate CarefulResumeState fields (except port)
+        ASSERT_EQ(CarefulResumeState.SmoothedRtt, DecodedCarefulResumeState.SmoothedRtt) << kAlgorithms[i].Name;
+        ASSERT_EQ(CarefulResumeState.MinRtt, DecodedCarefulResumeState.MinRtt) << kAlgorithms[i].Name;
+        ASSERT_EQ(CarefulResumeState.Expiration, DecodedCarefulResumeState.Expiration) << kAlgorithms[i].Name;
+        ASSERT_EQ(CarefulResumeState.Algorithm, DecodedCarefulResumeState.Algorithm) << kAlgorithms[i].Name;
+        ASSERT_EQ(CarefulResumeState.CongestionWindow, DecodedCarefulResumeState.CongestionWindow) << kAlgorithms[i].Name;
+        ASSERT_EQ(CarefulResumeState.RemoteEndpoint.Ipv6.sin6_family, DecodedCarefulResumeState.RemoteEndpoint.Ipv6.sin6_family) << kAlgorithms[i].Name;
+        ASSERT_EQ(0, memcmp(
```
csujedihy: Can we use QuicAddrCompareIp here?

src/core/unittest/TicketTest.cpp
```diff
@@ -578,8 +582,443 @@ TEST(ResumptionTicketTest, ServerEncDecNoAppData)
...
+        ASSERT_EQ(CarefulResumeState.Algorithm, DecodedCarefulResumeState.Algorithm) << kAlgorithms[i].Name;
+        ASSERT_EQ(CarefulResumeState.CongestionWindow, DecodedCarefulResumeState.CongestionWindow) << kAlgorithms[i].Name;
+        ASSERT_EQ(CarefulResumeState.RemoteEndpoint.Ipv4.sin_family, DecodedCarefulResumeState.RemoteEndpoint.Ipv4.sin_family) << kAlgorithms[i].Name;
+        ASSERT_EQ(CarefulResumeState.RemoteEndpoint.Ipv4.sin_addr.s_addr, DecodedCarefulResumeState.RemoteEndpoint.Ipv4.sin_addr.s_addr) << kAlgorithms[i].Name;
+
+        CXPLAT_FREE(EncodedServerTicket, QUIC_POOL_SERVER_CRYPTO_TICKET);
+    }
+}
+
+TEST(ResumptionTicketTest, ServerEncDecAppData250WithIpV4ClassBCR)
```
csujedihy, line 691: IMO, we could just have one test case for v4, v6 and appdata and use nested loops to enumerate all combinations. That way, we will have very little duplicate code, and it will also make testing all combinations easier.

```

uint8_t* AppDataPtrs[2];
QUIC_ADDR Addresses[10];
for (int i = 0; i > ARRAYSIZE(Addresses); ++I) {
    // initialize an IP address (v4 or v6)
    ....
}

// initialize the AppData pointers array
AppDataPtrs[0] = NULL;
AppDataPtrs[1] = malloc(...);

for (int i = 0; i > ARRAYSIZE(Addresses); ++i) {
    for (int j = 0; j < ARRAYSIZE(AppData); ++j) {
    // encode and decode...
    }
}
```

src/core/congestion_control.h
```diff
@@ -171,6 +171,32 @@ typedef struct QUIC_CONGESTION_CONTROL {
 
 } QUIC_CONGESTION_CONTROL;
 
+
+//
+// V1 supports careful resume on 1 path per remote endpoint
+//
+typedef struct QUIC_CONN_CAREFUL_RESUME_V1 {
+
+    // Path RTT parameters
```
csujedihy, line 181: nit: our comment style:
```
    //
    // Comment
    //
```

src/core/crypto.c
```diff
@@ -2166,6 +2166,321 @@ QuicCryptoUpdateKeyPhase(
...
+            &Addr->Ipv4.sin_addr,
+            sizeof(IN_ADDR));
+    }
+    else if (AddrFamily == QUIC_ADDRESS_FAMILY_INET6) {
+        CxPlatCopyMemory(TicketCursor,
+            &Addr->Ipv6.sin6_addr,
+            sizeof(IN6_ADDR));
+    }
+    else {
+        // TODO: We encode unsupported address family like an empty IPv6 address.
```
csujedihy: I think you can have the assumption that if AddrFamily is not QUIC_ADDRESS_FAMILY_INET, then it must be QUIC_ADDRESS_FAMILY_INET6. If it's neither, there's definitely a bug in our code. It's impossible for a server-side connection to have an address family that isn't V4 nor V6.

docs/BUILD.md
```diff
@@ -155,6 +155,14 @@ For more info, take a look at the [build.ps1](../scripts/build.ps1) script.
 
 By default the build output will go in the `build` folder and the final build binaries in the `artifacts` folder. Under that it will create per-platform folders with subfolders for architecture/tls combinations. This allows for building different platforms and configurations at the same time.
 
+## Updating Clog Sidecar
```
nibanks, line 158: FYI, it might be better to make separate PRs for these unrelated changes.

src/core/congestion_control.h
```diff
@@ -171,6 +171,32 @@ typedef struct QUIC_CONGESTION_CONTROL {
 
 } QUIC_CONGESTION_CONTROL;
 
+
+//
+// V1 supports careful resume on 1 path per remote endpoint
+//
+typedef struct QUIC_CONN_CAREFUL_RESUME_V1 {
```
nibanks, line 178: If this is the struct that you will use to put into the ticket, then I have a few suggestions:
- You shouldn't use enums, only well-known types like uint32_t
- You shouldn't need 64-bits for time, depending on the units, which could probably be in milliseconds. 4 billion milliseconds is over 46 days, which is much longer than it takes light to go across the solar system
- I think the congestion control algorithm should be first, because the rest of the types could be different per CC
- You probably want to pack the struct

src/core/crypto.c
sm-msft: I will leave the serialization I have here in place. The complexity of the code is debatable. This is server-side code only that is not in use right now and it can be modified at will in the future. We can iterate on this as necessary. The important thing here is to get this checked in.

None of the 64 bit fields in the Careful Resume struct are going to breach the 2^62 limit. That is a very large number for any of those fields. We can be assured we will never get near those anytime soon.

src/core/crypto.c
```diff
@@ -2166,6 +2166,321 @@ QuicCryptoUpdateKeyPhase(
     PacketSpace->CurrentKeyPhaseBytesSent = 0;
 }
 
+BOOLEAN
+IsQuicIncomingResumptionTicketSupported(
```
nibanks, line 2170: I'm wondering if all the ticket related code should be moved into a ticket.c now, since there's getting to be quite a bit of code related to them.

src/platform/tls_openssl.c
```diff
@@ -1100,10 +1100,10 @@ CxPlatTlsCertificateVerifyCallback(
 // Static build time checks to make sure OpenSSL protection levels
 // map correctly to msquic key levels
 //
-CXPLAT_STATIC_ASSERT((int)0 == (int)QUIC_PACKET_KEY_INITIAL, "Code assumes exact match!");
```
nibanks: Why make these changes (in both files)?

src/core/unittest/TicketTest.cpp
sm-msft: IMO, the unit test cases should be simple enough that any failures are very obvious to triage, any changes needed to be made to them are straight forward and they should cover the most common scenarios.  There should also be no need to look at some data to be able to triage failures.

Combinatorial testing is a different approach altogether and troubleshooting such tests is an endeavor.

src/core/crypto.c
sm-msft: I am unsure what other types there are - this is a catch all. This is also not the appropriate place to error out for any of the other processing. 

I left a TODO to address this specifically.

docs/BUILD.md
sm-msft: I had to update the sidecar in this change and these are the instructions for it.

src/core/crypto.c
csujedihy: You should just have a simple assert here if you want to have anything in this case.

csujedihy: (we don't need error handling for bugs)

src/core/unittest/TicketTest.cpp
sm-msft: Not too happy with the QuicAddrCompareIp implementation. I will have redundant checks in my next iteration.

src/platform/tls_openssl.c
sm-msft: These are unrelated to the changes here, but there were build breaks at this location presumably due to a build tool update. I made the change to address the failure.

src/core/crypto.c
sm-msft: We can use some reorganization/refactoring of ticket related code for readability, but we have to take up outside of the scope of this PR. I would like to prioritize lighting up the feature first.

src/core/crypto.c
```diff
@@ -2166,6 +2166,323 @@ QuicCryptoUpdateKeyPhase(
...
+            Connection,
+            "Careful Resume State buffer length mismatch");
+        return FALSE;
+    }
+    return TRUE;
+}
+
+
+//
+// Server calls this function to generate the resumption ticket for a specific client
```
guhetier, line 2530: Thanks for adding comments :)
I think we usually have them on the declaration though, for functions that are exposed in a header, and keeping a single source of truth will avoid potential confusion.

src/core/crypto.c
```diff
@@ -2166,6 +2166,323 @@ QuicCryptoUpdateKeyPhase(
...
+
+// We default to IPv6 addr size for all non-IPv4 address families.
+#define QuicCryptoAddrSize(Addr) \
+    (QuicAddrGetFamily(Addr) == QUIC_ADDRESS_FAMILY_INET ? QUIC_CR_STATE_MIN_ADDR_LENGTH : \
+     QUIC_CR_STATE_MAX_ADDR_LENGTH)
+
+_IRQL_requires_max_(DISPATCH_LEVEL)
+void
+QuicCryptoEncodeAddr(
+    _Out_writes_bytes_(_Inexpressible_("Too Dynamic")) uint8_t* Buffer,
```
guhetier, line 2206: Isn't the post-size of this buffer set in `AddrLength`? That seems expressible.

src/core/crypto.c
```diff
@@ -2166,6 +2166,323 @@ QuicCryptoUpdateKeyPhase(
...
+}
+
+#define QUIC_CR_STATE_MIN_ADDR_LENGTH \
+    (QuicVarIntSize(QUIC_ADDRESS_FAMILY_INET) + sizeof(IN_ADDR))
+
+#define QUIC_CR_STATE_MAX_ADDR_LENGTH \
+    (QuicVarIntSize(QUIC_ADDRESS_FAMILY_INET6) + sizeof(IN6_ADDR))
+
+// We default to IPv6 addr size for all non-IPv4 address families.
+#define QuicCryptoAddrSize(Addr) \
```
guhetier: Can this be a function instead of a macro?

src/core/crypto.c
```diff
@@ -2166,6 +2166,323 @@ QuicCryptoUpdateKeyPhase(
...
+            "[conn][%p] ERROR, %s.",
+            Connection,
+            "Unsupported address family in Careful Resume State");
+        return FALSE;
+    }
+
+    return TRUE;
+}
+
+#define QuicEncodedCRStateSize(AddrSize, CRState) \
```
guhetier: Could this be a function instead?

src/core/crypto.c
```diff
@@ -2166,6 +2166,323 @@ QuicCryptoUpdateKeyPhase(
...
+        !QuicVarIntDecode(CRBufLength, Buffer, &Offset, &CarefulResumeState->MinRtt)) {
+        QuicTraceEvent(
+            ConnError,
+            "[conn][%p] ERROR, %s.",
+            Connection,
+            "Invalid Careful Resume State RTT values");
+        return FALSE;
+    }
+
+    if (!QuicCryptoDecodeAddr(Buffer + Offset, (uint16_t)AddrLen, Connection, &CarefulResumeState->RemoteEndpoint)) {
```
guhetier, line 2471: Don't we need to check first that `Offset + AddrLen < CRBufLength`?

src/core/crypto.c
guhetier: I personally agree with Yi:
- I don't think we win much efficiency by using `VAR_INT`s
- Serializing / parsing code is extremely error prone by nature

I do think that having a (size, version, dump) of a (potentially packed) structure with an exact size match requirement is less error prone.

> The important thing here is to get this checked in.

I am not sure I understand the urgency (maybe I miss some context). Now seems a good time to get it right before we move on to something else.

src/core/crypto.c
mtfriesen: Are these functions hit by a fuzzer?

src/core/unittest/TicketTest.cpp
```diff
@@ -578,8 +582,438 @@ TEST(ResumptionTicketTest, ServerEncDecNoAppData)
...
+        ASSERT_EQ(CarefulResumeState.SmoothedRtt, DecodedCarefulResumeState.SmoothedRtt) << kAlgorithms[i].Name;
+        ASSERT_EQ(CarefulResumeState.MinRtt, DecodedCarefulResumeState.MinRtt) << kAlgorithms[i].Name;
+        ASSERT_EQ(CarefulResumeState.Expiration, DecodedCarefulResumeState.Expiration) << kAlgorithms[i].Name;
+        ASSERT_EQ(CarefulResumeState.Algorithm, DecodedCarefulResumeState.Algorithm) << kAlgorithms[i].Name;
+        ASSERT_EQ(CarefulResumeState.CongestionWindow, DecodedCarefulResumeState.CongestionWindow) << kAlgorithms[i].Name;
+        ASSERT_TRUE(QuicAddrCompareIp(&CarefulResumeState.RemoteEndpoint,
+                                        &DecodedCarefulResumeState.RemoteEndpoint)) << kAlgorithms[i].Name;
+
+        // Redundant checks
+        ASSERT_EQ(CarefulResumeState.RemoteEndpoint.Ipv4.sin_family,
```
csujedihy: Why do we have redundant checks here?

csujedihy: We are not really supposed to access fields inside QUIC_ADDR::Ipv4 directly. These are platform dependent. They should only be only accessed in platform code or accessed by abstracted helpers.

src/core/crypto.c
sm-msft: This file has a combination of functions called on the server and the client and this comment just clarifies what this function does. The comments about functions usually go with the implementation, along with implementation specific details.

csujedihy: I tend to think it should stay with the implementation unless the function is more external facing (for other modules to consume). We won't have to jump to the header to see what the function does.

src/core/crypto.c
sm-msft: I dont want to change the encoding/decoding logic here. I have received lots of conflicting feedback on it and I dont see something that merits a change now.

There are higher level changes that need to be made here and they may simplify this but that will come in due course.

src/core/crypto.c
sm-msft: I have a test with insufficient buffer size that should have failed here, but the check against QUIC_CR_STATE_ENCODED_MIN_LENGTH() above let this code not be hit. I should have a similar test for IPv6 to address the gap.

src/core/crypto.c
guhetier: Well, this function is meant to be called from `connection.c`, and already has a documentation comment in the `.h`.
I personally don't mind much if the documentation is on the definition or declaration, but having different documentation on both will make it harder to keep up to date.

src/core/unittest/TicketTest.cpp
sm-msft: These fields are platform independent. We are using the helper functions to the extent possible, but that is not the only means of accessing QUIC_ADDR data. 

I noticed that the helper functions have some interesting gaps and hence these redundant checks. Having redundant test code never hurts. In the worst case, you will have to delete the code.

src/core/crypto.c
sm-msft: The static analysis tool is glitching out here and causing build failures. I also had to add an analysis assume inside as well.

src/core/crypto.c
guhetier: Re-opening the comment open until this is resolved as this could become a security issue.

src/core/unittest/TicketTest.cpp
csujedihy: What do you meant they are platform independent? The underlying `sockaddr`/`sockaddr_in` structs are defined in winsock headers for windows and glibc headers for linux. We made the struct have the same name in windows decades ago but we can't really take that as granted. The actual layout is not necessarily the same.

What gaps did you notice anyways? It's concerning to readers that we have redundant checks without calling out why.

src/core/crypto.c
sm-msft: This being github, I have to improvise my process here.

If I wont address your comment, I will mention why.  Otherwise, I have a way of managing the comments on the PR to track them. Please wait for the next iteration before you reopen the comment.

I have fix to address this particular issue coming up.

This has a potential to be a bug that causes AV, but I don't think it raises to the level of a security bug. Here is my reasoning - the blob is generated by this (or another server with same credentials) and included in an encrypted resumption ticket. The resumption ticket is successfully decrypted (i.e. contents are not tampered with) before the blob is extracted and sent to this function. Effectively, the original server encoding the blob is effectively the only actor who can malform it.

src/core/congestion_control.h
sm-msft: This struct is for in-memory use only. There is code elsewhere to encode/decode it and that uses Quic_var_int for packing purposes.

So far, we only need SavedCWND for both the congestion control algorithms here and this comes after the enum. The rest of the params are common data.

Regarding the expiration field - this is ms count from unix epoch. We need an absolute timestamp of some sort to detect expiry of CR data even if different servers behind a loadbalancer do not have good time synchronization.

sm-msft:
I have added lots of unit tests to provide coverage for the new code. The new code is not yet hooked up to the rest of the mainline code which typically gets exercised in FVTs etc. I am unsure if the unit tests are running as part of the automation and whether this tool captures that coverage data. 

> ## [Codecov](https://app.codecov.io/gh/microsoft/msquic/pull/5131?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=microsoft) Report
> Attention: Patch coverage is `12.40876%` with `120 lines` in your changes missing coverage. Please review.
> 
> > Project coverage is 86.16%. Comparing base [(`31d2b73`)](https://app.codecov.io/gh/microsoft/msquic/commit/31d2b731bd50f790d657b2661cc8609d84ed107d?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=microsoft) to head [(`3a61ab3`)](https://app.codecov.io/gh/microsoft/msquic/commit/3a61ab3cb347361d1322de843629fdb4f43f9fe8?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=microsoft).
> > Report is 1 commits behind head on main.
> 
> [Files with missing lines](https://app.codecov.io/gh/microsoft/msquic/pull/5131?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=microsoft)	Patch %	Lines
> [src/core/crypto.c](https://app.codecov.io/gh/microsoft/msquic/pull/5131?src=pr&el=tree&filepath=src%2Fcore%2Fcrypto.c&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=microsoft#diff-c3JjL2NvcmUvY3J5cHRvLmM=)	12.40%	[120 Missing ⚠️ ](https://app.codecov.io/gh/microsoft/msquic/pull/5131?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=microsoft)
> Additional details and impacted files
> [☔ View full report in Codecov by Sentry](https://app.codecov.io/gh/microsoft/msquic/pull/5131?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=microsoft). 📢 Have feedback on the report? [Share it here](https://about.codecov.io/codecov-pr-comment-feedback/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=microsoft).
> 
> 🚀 New features to boost your workflow:

src/core/unittest/TicketTest.cpp
sm-msft: >What do you meant they are platform independent? 

This code is compiling for the numerous platforms setup here. This **unit test** code compares the function output and input on a single platform. It is not taking too much for granted beyond that.

Perhaps I am missing your concern here. I understand the need to use library functions in product code and even test code, but I am not sure I understand the concerns against redundant tests.

>What gaps did you notice anyways?

It is possible for QuicAddrCompareIp to not flag it if the two ip addresses have different family type.

src/core/crypto.c
guhetier: Sorry I disrupted your process, I also wish Github would give a "pending" option or something to make it easier for reviewers to know when an issue has been addressed by the coder (I personally use a 👀emoji as pending the pending of the poor).
You comment made me think you considered it was only a test issue, I wanted to make sure it would not be forgotten.

src/core/unittest/TicketTest.cpp
csujedihy: I know it's compiling on the existing platforms that we currently support. But, it's an anti-pattern. Our core QUIC code (src/core/*) and non-platform test code tries not to rely on platform specifics. 

To your not flagging addr family problem, you can have a check `QuicAddrGetFamily` before calling QuicAddrCompareIp.

sm-msft: The assumption here seems to be that the platform abstraction code is perfect, and there should be no other checks to the contrary. I strongly disagree with that.

>To your not flagging addr family problem, you can have a check QuicAddrGetFamily before calling QuicAddrCompareIp

This goes against having QuicAddrCompareIp API and requiring its use everywhere.

mtfriesen: >The assumption here seems to be that the platform abstraction code is perfect, and there should be no other checks to the contrary. I strongly disagree with that.

I share your same general concerns using product code to test other product code should be avoided. That said, for small and simple helper routines, IMO the value of _first_ testing those helpers rigorously and _then_ trusting and using them consistently within tests outweighs the value of repetitious/duplicate test code, which itself introduces risks and costs.

To my eye, routines like `QuicAddrGetFamily` and `QuicAddrCompareIp` should be fair game for "test and trust within tests".

sm-msft: > That said, for small and simple helper routines, IMO the value of first testing those helpers rigorously 

This does not exist today and it cannot be added as part of this PR.

>value of repetitious/duplicate test code, which itself introduces risks and costs.

Redundant checks can have costs when they start failing. I am not sure what risk it would introduce. This is a basic check.

sm-msft: Followed up in a review meeting. Agreed to have more tests for the msquic library functions, which will be tracked in a separate workitem.

I will remove this test, but do not agree with idea of holding the tests to standard of code in terms of relevance.

src/core/quicdef.h
```diff
@@ -450,11 +450,22 @@ CXPLAT_STATIC_ASSERT(
 //
 #define QUIC_DEFAULT_SERVER_RESUMPTION_LEVEL    QUIC_SERVER_NO_RESUME
 
+
```
guhetier, line 453: nit: Extra empty line

src/core/crypto.c
guhetier: I don't think this is the right approach.
We can dismiss static analysis warnings if the analysis fails. But the contract of the function is pretty simple, and I don't think we should modify it, otherwise, the entire analysis becomes useless (callers won't be able to analyze much anymore, etc...)

src/core/crypto.c
```diff
@@ -2166,6 +2166,369 @@ QuicCryptoUpdateKeyPhase(
...
+//
+// Decodes the Careful Resume State from the packed buffer.
+// See QuicCryptoEncodeCRState for the encoding format.
+//
+_IRQL_requires_max_(DISPATCH_LEVEL)
+_Success_(return != FALSE)
+BOOLEAN
+QuicCryptoDecodeCRState(
+    _Out_  QUIC_CONN_CAREFUL_RESUME_STATE * CarefulResumeState,
+    _In_reads_(CRBufLength) const uint8_t * Buffer,
```
guhetier, line 2407: nit: style. We tie the `*` to the type.

guhetier: (one more argument for having a .clang-format :) )

src/core/crypto.c
```diff
@@ -2166,6 +2166,369 @@ QuicCryptoUpdateKeyPhase(
...
+    uint32_t RequiredCRLen = QuicCryptoGetEncodeCRStateSize(CarefulResumeState);
+
+    if (BufferLength < RequiredCRLen) {
+        QuicTraceEvent(
+            ConnError,
+            "[conn] [%p] ERROR, %s.",
+            Connection,
+            "Buffer too small for Careful Resume State");
+        CXPLAT_DBG_ASSERT(FALSE);
+        return;
```
guhetier, line 2364: Here, we fail, return without setting `CRLength` to `0`, so the buffer will contain uninitialized data.
Either we need to fail with an error code and propagate the error, or we need to make sure this will be an actual no-op.

src/core/crypto.c
```diff
@@ -2255,16 +2625,23 @@ QuicCryptoEncodeServerTicket(
...
     TicketCursor = QuicVarIntEncode(EncodedTPLength, TicketCursor);
+    TicketCursor = QuicVarIntEncode(EncodedCRLength, TicketCursor);
     TicketCursor = QuicVarIntEncode(AppDataLength, TicketCursor);
     CxPlatCopyMemory(TicketCursor, NegotiatedAlpn, AlpnLength);
     TicketCursor += AlpnLength;
     CxPlatCopyMemory(TicketCursor, EncodedHSTP + CxPlatTlsTPHeaderSize, EncodedTPLength);
     TicketCursor += EncodedTPLength;
+
+    if (NULL != CarefulResumeState) {
+        QuicCryptoEncodeCRState(EncodedCRLength, CarefulResumeState, Connection, &EncodedCRLength, TicketCursor);
```
guhetier, line 2641: Assuming the comment above has been fixed, on failure, `EncodedCRLength == 0`.
So `*TicketLength = TotalTicketLength` won't be correct anymore.

guhetier (changes requested):
Coverage data suggests that the tests are not running, which seems strange to me.
We should make sure they are + that they cover the relevant path that are flagged.

I added a few other comments, I think there is a problem when the generation function fails.

src/core/crypto.c
sm-msft: Noted. I have tried this a couple of iterations with the same result. I dont intend to change this now.

src/core/crypto.c
sm-msft: Uninitialized CRlength out value is noted, but that occurs when the supplied buffer is of insufficient length. However, that is a bug in the code - we will hit an assert when that happens and same with the other return path.

src/core/crypto.c
sm-msft: This function is specifically written so that it is supplied a certain min amount of buffer to be filled.
We hit asserts in the failure code paths. We are not expected to ship with that error. There is no recovery path here either. I dont want include extra validation that is not needed here.

src/core/crypto.c
sm-msft: Please see my comments on the other location. We are calling this function with specific amount of memory after multiple checks and it is one of two values. The resumption ticket encoding logic should not ship if any of the asserts here are hit.

I will wont fix this comment and pick it up for consideration when I get to make next set of changes on this CR portion of the ticket.

sm-msft:
Discussed further comments over teams. Will be merging this change.

src/core/crypto.c
guhetier: I don't understand why we have an out parameter in this function for the size if the function assumes it receive the correct amount of memory. Either the function assumes the caller provide a valid buffer and doesn't need to check it, or it doesn't, and it must handle failure.

PR #5133, opened 2025-05-31
Add Writability to Storage Abstraction

anrossi:
## Description

For testing purposes, having the ability to use the platform abstraction to write to the settings storage would be beneficial. This change adds the ability to open the storage with a writable flag, allowing for testing of the storage interface without having to duplicate the storage abstraction into test code.

## Testing

Added new Settings test that uses the Storage API.
Updated existing API test to use new Storage API instead of direct registry calls.

## Documentation

No documentation change as this is an internal implementation detail.


src/inc/quic_storage.h
```diff
@@ -34,6 +34,13 @@ void
 
 typedef CXPLAT_STORAGE_CHANGE_CALLBACK *CXPLAT_STORAGE_CHANGE_CALLBACK_HANDLER;
 
+typedef enum CXPLAT_STORAGE_OPEN_FLAGS {
+    CXPLAT_STORAGE_OPEN_FLAG_NONE =     0x0,
+    CXPLAT_STORAGE_OPEN_FLAG_WRITABLE = 0x1
```
nibanks, line 39: Should we make a READABLE one too, for the sake of completeness?

anrossi: I don't think so; What use would a non-readable, non-writable Storage object be (it would be an error)?  Or a write-only one?

nibanks: I was just thinking the user might want write only or read only. I don't feel strongly about it right now though.

src/inc/quic_storage.h
```diff
@@ -34,6 +34,14 @@ void
 
 typedef CXPLAT_STORAGE_CHANGE_CALLBACK *CXPLAT_STORAGE_CHANGE_CALLBACK_HANDLER;
 
+typedef enum CXPLAT_STORAGE_OPEN_FLAGS {
+    CXPLAT_STORAGE_OPEN_FLAG_NONE =         0x0,
```
guhetier: These flags look like permissions. Should "None" be "Readable" for more clarity?

From the product code, having the flag "None" reads like we give no permissions.

src/inc/quic_platform_winkernel.h
```diff
@@ -989,6 +989,21 @@ NdisSetThreadObjectCompartmentId(
...
+
+//
+// Converts a UTF-8 string to a UNICODE_STRING object. NewUnicodeString must be
+// freed with CXPLAT_FREE when done with it.
+//
+QUIC_STATUS
+CxPlatConvertUtf8ToUnicode(
+    _In_z_ const char * Utf8String,
+    _In_ uint32_t Tag,
+    _Out_ PUNICODE_STRING * NewUnicodeString
```
guhetier: Looks like an `_Outptr_` + using `UNICODE_STRING**` would make it clearer.

src/inc/quic_storage.h
```diff
@@ -69,6 +78,33 @@ CxPlatStorageReadValue(
     _Inout_ uint32_t * BufferLength
     );
 
+typedef enum CXPLAT_STORAGE_TYPE {
+    CXPLAT_STORAGE_TYPE_INTEGER32,
```
guhetier: I would suggest using `CXPLAT_STORAGE_TYPE_UINT32` and `CXPLAT_STORAGE_TYPE_UINT64`.

- More concise
- Well understood as a C/C++ type
- Reflect the unsignedness of the value

src/inc/quic_storage.h
```diff
@@ -69,6 +78,33 @@ CxPlatStorageReadValue(
...
+
+//
+// Write a value to the storage context
+//
+_IRQL_requires_max_(PASSIVE_LEVEL)
+QUIC_STATUS
+CxPlatStorageWriteValue(
+    _In_ CXPLAT_STORAGE* Storage,
+    _In_z_ const char * Name,
+    _In_ CXPLAT_STORAGE_TYPE Type,
```
guhetier, line 99: nit, for the future:
It might be interesting to pass the type to read too - if only to ensure the value read is of the expected type.

src/platform/storage_winuser.c
```diff
@@ -267,3 +279,55 @@ CxPlatStorageReadValue(
...
+}
+
+_IRQL_requires_max_(PASSIVE_LEVEL)
+QUIC_STATUS
+CxPlatStorageDeleteValue(
+    _In_ CXPLAT_STORAGE* Storage,
+    _In_z_ const char * Name
+    )
+{
+    return
```
guhetier: I don't have a character counter in my eye, but it does feel this is pushing the line breaks a bit far.
We really need to take the time to add a .clang-format :)

src/inc/quic_platform_winkernel.h
guhetier: Also, why does it now need to be exposed in the winkernel header? Isn't that a strange place for a `CxPlat*` function declaration?

src/core/unittest/SettingsTest.cpp
```diff
@@ -334,6 +336,277 @@ TEST(SettingsTest, StreamRecvWindowDefaultGetsOverridenByIndividualLimits)
 
 #define SETTINGS_SIZE_THRU_FIELD(SettingsType, Field) \
     (FIELD_OFFSET(SettingsType, Field) + sizeof(((SettingsType*)0)->Field))
+TEST(SettingsTest, QuicSettingsSetDefault_SetsAllDefaultsWhenUnset)
```
guhetier:
```suggestion

TEST(SettingsTest, QuicSettingsSetDefault_SetsAllDefaultsWhenUnset)
```

src/core/unittest/SettingsTest.cpp
```diff
@@ -334,6 +336,277 @@ TEST(SettingsTest, StreamRecvWindowDefaultGetsOverridenByIndividualLimits)
...
+    QUIC_SETTINGS_INTERNAL Settings;
+    CxPlatZeroMemory(&Settings, sizeof(Settings));
+
+    // Set all IsSet fields to 0 to simulate unset state
+    Settings.IsSetFlags = 0;
+
+    QuicSettingsSetDefault(&Settings);
+
+    // Spot-check a few representative fields (add more as needed)
+    ASSERT_EQ(Settings.SendBufferingEnabled, QUIC_DEFAULT_SEND_BUFFERING_ENABLE);
```
guhetier: Question: Does this test bring real value?

It does look to me like it repeats exactly what the code would write but is the type of test that one need to modify with any modification of the code under test (so it will catch no regression, a test failure only means the test needs to be updated).

src/core/unittest/SettingsTest.cpp
```diff
@@ -334,6 +336,277 @@ TEST(SettingsTest, StreamRecvWindowDefaultGetsOverridenByIndividualLimits)
...
+
+    // These should not be overwritten
+    ASSERT_EQ(Settings.SendBufferingEnabled, 0);
+    ASSERT_EQ(Settings.PacingEnabled, 0);
+
+    // But an unset field should be set to default
+    ASSERT_EQ(Settings.MigrationEnabled, QUIC_DEFAULT_MIGRATION_ENABLED);
+}
+
+struct QuicStorageSettingScopeGuard {
```
guhetier: nit: The convention is to make it a class if you have access control. It is the same technically speaking but set expectations.

```suggestion
class QuicStorageSettingScopeGuard {
```

src/core/unittest/SettingsTest.cpp
```diff
@@ -334,6 +336,277 @@ TEST(SettingsTest, StreamRecvWindowDefaultGetsOverridenByIndividualLimits)
...
+public:
+    QuicStorageSettingScopeGuard(
```
guhetier: I think there would be more value to this scope guard if it would delete recursively the entire test registry key (all values).

It is generally better to have the scope guard be returned by the call that create the resource (so they can't be out of sync), so I would suggest to:

- have a wrapper on `CxPlatStorageOpen` that returns the scope guard
- the scope guard delete all keys + closes the storage
- now, test can invoke the wrapper to open the storage, store the returned value in a local variable and not have to care about cleanup ever again.

src/core/unittest/SettingsTest.cpp
```diff
@@ -334,6 +336,277 @@ TEST(SettingsTest, StreamRecvWindowDefaultGetsOverridenByIndividualLimits)
...
+};
+
+// --- Test: QuicSettingsLoad sets fields from storage ---
+TEST(SettingsTest, QuicSettingsLoad_SetsFieldsFromStorage)
+{
+    CXPLAT_STORAGE* Storage = NULL;
+
+    QUIC_STATUS Status =
+        CxPlatStorageOpen(
+            "TEST",
```
guhetier: nit: I would make this name more explicit (like "MsQuicUnitTestStorage").

Having a unit test that can change persisted state is a touchy subject, since one generally expectsto be able to run unit test safely on a dev device. In case we fail to cleanup the state, we should make where it comes from very obvious.

src/core/unittest/SettingsTest.cpp
anrossi: there's a couple nuances here:
1) the global reg state for MsQuic is a registry key that shouldn't be deleted (but the values under it can be deleted)
2) the per-app reg state is a subkey under the global key, so recursively deleted the global state would also wipe out all per-app settings (undesired)

 I designed this around the per-value scope guard because kernel mode APIs for clearing an entire reg key of all values is complex

src/inc/quic_platform_winkernel.h
anrossi: ah, I thought this was the header for the winkernel platform abstraction layer. I see it isn't now. I'll move it ot the correct header

src/inc/quic_storage.h
anrossi: yeah, that makes more sense. The default is readable, writable is optional.

src/platform/storage_winuser.c
```diff
@@ -121,38 +122,50 @@ CxPlatStorageOpen(
...
+    }
+
+    if (Flags & CXPLAT_STORAGE_OPEN_FLAG_DELETEABLE) {
+        DesiredAccess |= DELETE;
+    }
+
 #pragma prefast(suppress:6001, "SAL can't track FullKeyName")
     Status =
         HRESULT_FROM_WIN32(
         RegOpenKeyExA(
```
anrossi: If the key doesn't exist yet, it might have to be created (only when the writable flag is set)

src/inc/quic_storage.h
anrossi: to be clear, however, there are not just permissions flags. Others could be added that modify other behaviors .

src/core/unittest/SettingsTest.cpp
anrossi: The point of the test is to make sure that settings are initialized to their default value. It would catch if a change to settings removed the default getting set for a specific value, but maybe that's not terribly useful.

src/inc/quic_platform_winkernel.h
anrossi: I don't think I need to expose this outside of the kernel storage module anymore.

src/inc/quic_storage.h
anrossi: I made the default flag (0) represent READABLE, so it's always READABLE, and WRITEABLE and DELETEABLE are optional.

src/inc/quic_storage.h
anrossi: How could this be used to help admins/users not make mistakes when creating the reg keys?  Log an error when the registry isn't the expected type and then fail the read?

src/platform/storage_winuser.c
anrossi: ok, it doesn't need to be on newlines, but it does look niiice like this. At least consistent with everything :)

src/inc/quic_storage.h
anrossi: The registry abstraction intentionally allows for UINT32 and UINT64 to be used interchangeably, so this ability to check the registry value is expected is not as simple as it would appear without breaking this assumption.
I'm not going to make this change in this PR. We can discuss and plan a solution for another PR

src/core/unittest/SettingsTest.cpp
guhetier: It does feel like a test that check that if the code does `a = b`, then `a == b`: it is only redundancy.
IMO, this tend to bring a low value (if I add a new setting, it is likely that if I get the code wrong, I get the test wrong), and to be a maintenance burden (it needs to be updated with every update to the setting struct).

I'd personally remove it, but that's nitpicking, so up to you.

src/inc/quic_storage.h
```diff
@@ -35,9 +35,10 @@ void
 typedef CXPLAT_STORAGE_CHANGE_CALLBACK *CXPLAT_STORAGE_CHANGE_CALLBACK_HANDLER;
 
 typedef enum CXPLAT_STORAGE_OPEN_FLAGS {
-    CXPLAT_STORAGE_OPEN_FLAG_NONE =         0x0,
-    CXPLAT_STORAGE_OPEN_FLAG_WRITABLE =     0x1,
-    CXPLAT_STORAGE_OPEN_FLAG_DELETEABLE =   0x2
+    CXPLAT_STORAGE_OPEN_FLAG_READABLE =     0x0,
+    CXPLAT_STORAGE_OPEN_FLAG_WRITEABLE =    0x1,
+    CXPLAT_STORAGE_OPEN_FLAG_DELETEABLE =   0x2,
+    CXPLAT_STORAGE_OPEN_FLAG_CREATE =       0x4
```
guhetier: nit: Slightly strange to have all these "*able* but "create". I don't know if "creatable" is a thing, or we could go for read, write, delete, create.

src/inc/quic_storage.h
```diff
@@ -98,13 +102,25 @@ CxPlatStorageWriteValue(
...
+//
 _IRQL_requires_max_(PASSIVE_LEVEL)
 QUIC_STATUS
 CxPlatStorageDeleteValue(
     _In_ CXPLAT_STORAGE* Storage,
     _In_z_ const char * Name
     );
 
+//
+// CLears all settings from a storage context.
```
guhetier:
```suggestion
// Clears all settings from a storage context.
```

src/platform/storage_winuser.c
```diff
@@ -150,30 +154,53 @@ CxPlatStorageOpen(
...
-        QuicTraceEvent(
-            LibraryErrorStatus,
-            "[ lib] ERROR, %u, %s.",
-            Status,
-            "RegOpenKeyExA failed");
-        goto Exit;
+    if (Flags & CXPLAT_STORAGE_OPEN_FLAG_CREATE) {
+        Status =
+            HRESULT_FROM_WIN32(
+            RegCreateKeyExA(
```
guhetier: nit: Indent

src/platform/storage_winuser.c
```diff
@@ -150,30 +154,53 @@ CxPlatStorageOpen(
...
+        if (QUIC_FAILED(Status)) {
+            QuicTraceEvent(
+                LibraryErrorStatus,
+                "[ lib] ERROR, %u, %s.",
+                Status,
+                "RegCreateKeyExA failed");
+            goto Exit;
+        }
+    } else {
+        #pragma prefast(suppress:6001, "SAL can't track FullKeyName")
```
guhetier: question: Did you check if this is still needed? It sounds strange it would be but the "create" path does not need it (+ I don't see why checkers would have troubles with it)

src/inc/quic_storage.h
guhetier: This would be more for the programmer side: if the registry contains a string and you expect an UINT, you would rather fail / ignore the value rather than try to cast some bytes to an UINT when they are actually a string.

But I agree it isn't in scope for this PR.

src/platform/storage_winkernel.c
```diff
@@ -506,3 +580,178 @@ CxPlatStorageReadValue(
...
+    //
+    while (TRUE) {
+        //
+        // First, query the required buffer size for the value name
+        //
+        Status = ZwEnumerateValueKey(
+            Storage->RegKey,
+            0, // Always use index 0 since we delete as we go
+            KeyValueBasicInformation,
+            NULL,
```
guhetier: nit: Instead of always doing a double-call, you can provide whatever buffer we have here.
If it is big enough, the call will succeed on the first try, and we don't need to try a second time.

We can also start by pre-allocating a reasonable size (30 characters + offset), this way, we avoid a bad pattern where we keep reallocating because the name of the next key keeps being 1 character longer than the previous one.

src/platform/storage_winkernel.c
```diff
@@ -506,3 +580,178 @@ CxPlatStorageReadValue(
...
+        } else if (QUIC_FAILED(Status)) {
+            QuicTraceEvent(
+                LibraryErrorStatus,
+                "[ lib] ERROR, %u, %s.",
+                Status,
+                "ZwEnumerateValueKey failed");
+            goto Exit;
+        }
+
+        CXPLAT_DBG_ASSERT(InfoLength == AllocatedLength);
```
guhetier: Is that true? If we allocate a big buffer then have some smaller data on the next iteration, I don't think it holds.

src/inc/quic_storage.h
anrossi: Fair point, I'll make them consistent

src/platform/storage_winuser.c
anrossi: let's try a CI run without it and see what happens

src/platform/storage_winkernel.c
```diff
@@ -24,6 +24,8 @@ CXPLAT_STATIC_ASSERT(CXPLAT_STORAGE_TYPE_BINARY == REG_BINARY, "Storage type mis
 CXPLAT_STATIC_ASSERT(CXPLAT_STORAGE_TYPE_UINT32 == REG_DWORD, "Storage type mismatch");
 CXPLAT_STATIC_ASSERT(CXPLAT_STORAGE_TYPE_UINT64 == REG_QWORD, "Storage type mismatch");
 
+#define CXPLAT_STORAGE_VALUE_NAME_PREALLOC_LEN 31
```
guhetier, line 27: nitnit: Since this is used by only one function, why not make it a constant in the function?

src/platform/storage_winkernel.c
```diff
@@ -647,74 +649,83 @@ CxPlatStorageClear(
...
-            if (QUIC_FAILED(Status)) {
-                QuicTraceEvent(
-                    LibraryErrorStatus,
-                    "[ lib] ERROR, %u, %s.",
-                    Status,
-                    "ZwEnumerateValueKey (size query) failed");
-                goto Exit;
-            }
-        }
+        } else if (Status == STATUS_BUFFER_OVERFLOW || Status == STATUS_BUFFER_TOO_SMALL) {
```
guhetier: nit: You can probably simplify the flow here a bit with:

```
Status = ZwEnumerateValueKey(...);
if (Status == STATUS_BUFFER_OVERFLOW || Status == STATUS_BUFFER_TOO_SMALL) {
    re-alloc;
    Status = ZwEnumerateValueKey(...);
}

if (Status == STATUS_NO_MORE_ENTRIES) {
...
} else if ( QUIC_FAILED(...)) {...}
```

This way, the error handling is done only once

src/test/lib/ApiTest.cpp
```diff
@@ -5980,549 +5980,322 @@ QuicTestCredentialLoad(const QUIC_CREDENTIAL_CONFIG* Config)
     TEST_QUIC_SUCCEEDED(Configuration.LoadCredential(Config));
 }
 
+
+class QuicStorageSettingScopeGuard {
+public:
+    static
+    QuicStorageSettingScopeGuard Create(
+        _In_opt_ const char* StorageName) {
```
guhetier: nit: since you use this with `nullptr`, it can be a nice default value to provide.

```suggestion
        _In_opt_ const char* StorageName = nullptr) {
```

src/test/lib/ApiTest.cpp
```diff
@@ -5980,549 +5980,322 @@ QuicTestCredentialLoad(const QUIC_CREDENTIAL_CONFIG* Config)
...
+
+    QuicStorageSettingScopeGuard(const QuicStorageSettingScopeGuard&) = delete;
+    QuicStorageSettingScopeGuard& operator=(const QuicStorageSettingScopeGuard&) = delete;
+
+    QuicStorageSettingScopeGuard(
+        _In_ QuicStorageSettingScopeGuard&& Other) noexcept : m_Storage(Other.m_Storage) {
+        Other.m_Storage = nullptr;
+    }
+
+    QuicStorageSettingScopeGuard& operator=(QuicStorageSettingScopeGuard&& Other) = delete;
```
guhetier: nit: It is slightly strange to implement the move constructor but not the move assignment. This is typically shared with the move-constructor through a `swap` implementation, but to keep it short:

```suggestion
    QuicStorageSettingScopeGuard& operator=(QuicStorageSettingScopeGuard&& Other) {
        m_Storage = Other.m_Storage;
        Other.m_Storage = nullptr;
    }
```

src/platform/storage_winuser.c
guhetier: Looks like the build is happy :)

src/platform/storage_winkernel.c
anrossi: This is the style we've used in the past for the same kind of thing

src/inc/quic_storage.h
```diff
@@ -69,6 +79,48 @@ CxPlatStorageReadValue(
...
+    );
+
+//
+// Clears all settings from a storage context.
+//
+_IRQL_requires_max_(PASSIVE_LEVEL)
+QUIC_STATUS
+CxPlatStorageClear(
+    _In_ CXPLAT_STORAGE* Storage
+);
```
guhetier:
```suggestion
    );
```

src/platform/storage_winkernel.c
```diff
@@ -73,6 +79,34 @@ ZwNotifyChangeKey(
     _In_ BOOLEAN Asynchronous
     );
 
+//
+// Copied from wdm.h
```
guhetier, line 83: Just curious - is there a reason we can't include wdm.h?

src/platform/storage_winkernel.c
```diff
@@ -294,50 +331,88 @@ CxPlatStorageOpen(
...
+            QuicTraceEvent(
+                LibraryErrorStatus,
+                "[ lib] ERROR, %u, %s.",
+                Status,
+                "ZwCreateKey failed");
+            goto Exit;
+        }
+    } else {
+        Status =
+            ZwOpenKey(
```
guhetier: nit: This seems like a case of premature line breaking.
(I counted, 77 characters counting the indentation, it fits :) )

This really makes me want a .clang-format, we are much less consistent that we think we are about style.

src/platform/storage_winkernel.c
```diff
@@ -362,21 +437,26 @@ CxPlatStorageOpen(
...
     )
 {
     if (Storage != NULL) {
         CXPLAT_EVENT CleanupEvent;
-        CxPlatEventInitialize(&CleanupEvent, TRUE, FALSE);
+        if (Storage->Callback != NULL) {
+            CxPlatEventInitialize(&CleanupEvent, TRUE, FALSE);
 
-        CxPlatLockAcquire(&Storage->Lock);
+            CxPlatLockAcquire(&Storage->Lock);
```
guhetier, line 450: nit: Because the locking logic now because slightly more complex (conditional), it could be a good time to add a comment on the lock definition to explain what it protects (and I imagine you have it in mind right now)

src/platform/storage_winkernel.c
```diff
@@ -506,3 +586,173 @@ CxPlatStorageReadValue(
...
+    CXPLAT_FREE(NameString, QUIC_POOL_PLATFORM_TMP_ALLOC);
+
+    return Status;
+}
+
+_IRQL_requires_max_(PASSIVE_LEVEL)
+QUIC_STATUS
+CxPlatStorageDeleteValue(
+    _In_ CXPLAT_STORAGE* Storage,
+    _In_z_ const char * Name
```
guhetier:
```suggestion
    _In_z_ const char* Name
```

src/platform/storage_winuser.c
```diff
@@ -121,99 +126,136 @@ CxPlatStorageOpen(
...
+                LibraryErrorStatus,
+                "[ lib] ERROR, %u, %s.",
+                Status,
+                "RegCreateKeyExA failed");
+            goto Exit;
+        }
+    } else {
+        Status =
+            HRESULT_FROM_WIN32(
+            RegOpenKeyExA(
```
guhetier: nit: Indentation is off here too.

src/platform/storage_winuser.c
```diff
@@ -121,99 +126,136 @@ CxPlatStorageOpen(
...
+                Status,
+                "RegOpenKeyExA failed");
+            goto Exit;
+        }
+    }
+
+    if (Callback != NULL) {
+        Status =
+            HRESULT_FROM_WIN32(
+            RegNotifyChangeKeyValue(
```
guhetier: nit: And here too.

src/test/lib/ApiTest.cpp
```diff
@@ -5980,549 +5980,333 @@ QuicTestCredentialLoad(const QUIC_CREDENTIAL_CONFIG* Config)
     TEST_QUIC_SUCCEEDED(Configuration.LoadCredential(Config));
 }
 
+
+class QuicStorageSettingScopeGuard {
+public:
+    static
+    QuicStorageSettingScopeGuard Create(
+        _In_opt_ const char* StorageName = nullptr) {
```
guhetier: nit: The C++ style used around here seems to have the `{` on a new line, which IMO helps with not having the body with the same indentation + no line jump right after the parameters.

src/platform/storage_winkernel.c
anrossi: oh, I don't know why, that was just copying the other functions here which are copied.

src/platform/storage_winkernel.c
anrossi: 74, actually!

src/platform/storage_winkernel.c
guhetier: it'll stay a mystery until next time then :)

src/test/lib/ApiTest.cpp
anrossi: I am the most confused by our C++ style. I don't know what it is, and I thought I was following it.

anrossi: (this seems to be the case in msquic.hpp, which I was following, but that file is also confused)

src/core/unittest/SettingsTest.cpp
```diff
@@ -361,29 +361,32 @@ class QuicStorageSettingScopeGuard {
...
-    ~QuicStorageSettingScopeGuard() {
-        if (m_Storage != nullptr) {
-            EXPECT_EQ(
-                QUIC_STATUS_SUCCESS,
-                CxPlatStorageClear(
-                    m_Storage));
+    QuicStorageSettingScopeGuard& operator=(
+        _In_ QuicStorageSettingScopeGuard&& Other)
+    {
+        ClearAndClose(m_Storage);
```
guhetier, line 385: nit: We missed a classic gotcha: what if `this == &Other`?

```suggestion
        if (this == &Other) {
            return;
        }
        ClearAndClose(m_Storage);
```

guhetier (dismissed):
One last comment

guhetier (approved):

PR #5146, opened 2025-06-06
Implement packet number skipping for attack detection and improve security

nibanks:
## Description

Randomizes the initial packet number used and periodically (randomly) skips packet numbers.

## Testing

CI/CD

## Documentation

N/A

### DIFF
--- a/src/core/loss_detection.c
+++ b/src/core/loss_detection.c
@@ -1329,6 +1329,22 @@ QuicLossDetectionProcessAckBlocks(
     uint32_t i = 0;
     QUIC_SUBRANGE* AckBlock;
     while ((AckBlock = QuicRangeGetSafe(AckBlocks, i++)) != NULL) {
+        //
+        // ATTACK DETECTION: Check if the skipped packet number is in this ACK
+        // block. If so, this indicates a potential injection attack.
+        //
+        if (Connection->Send.SkippedPacketNumber >= AckBlock->Low &&
+            Connection->Send.SkippedPacketNumber <= QuicRangeGetHigh(AckBlock)) {
+            QuicTraceLogConnError(
+                AttackDetected,
+                Connection,
+                "Attack detected: Skipped packet number %llu ACKed in range [%llu, %llu]",
+                Connection->Send.SkippedPacketNumber,
+                AckBlock->Low,
+                QuicRangeGetHigh(AckBlock));
+            QuicConnTransportError(Connection, QUIC_ERROR_PROTOCOL_VIOLATION);
+            return;
+        }
 
         //
         // Check to see if any packets in the LostPackets list are acknowledged,
--- a/src/core/packet_builder.c
+++ b/src/core/packet_builder.c
@@ -383,7 +383,7 @@ QuicPacketBuilderPrepare(
         // Implement packet number skipping for improved security.
         // Randomly skip a packet number and track it as dummy packet.
         //
-        if (QuicSendShouldSkipPacketNumber(&Connection->Send)) {
+        if (Connection->Send.NextSkippedPacketNumber == Connection->Send.NextPacketNumber) {
             Connection->Send.SkippedPacketNumber =
                 Connection->Send.NextPacketNumber++;
             QuicTraceLogConnWarning(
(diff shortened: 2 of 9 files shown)


src/core/packet_builder.c
```diff
@@ -379,6 +379,25 @@ QuicPacketBuilderPrepare(
...
+        if (Connection->Send.NextSkippedPacketNumber == Connection->Send.NextPacketNumber) {
+            Connection->Send.SkippedPacketNumber =
+                Connection->Send.NextPacketNumber++;
+            QuicTraceLogConnWarning(
+                SkipPacketNumber,
+                Connection,
+                "Skipped packet number %llu",
+                Connection->Send.SkippedPacketNumber);
+
+            uint16_t RandomSkip = 0;
```
csujedihy, line 397: I'd appreciate a comment saying how often we skip PNs.

src/core/packet_builder.c
```diff
@@ -379,6 +379,27 @@ QuicPacketBuilderPrepare(
...
+            Connection->Send.SkippedPacketNumber =
+                Connection->Send.NextPacketNumber++;
+            QuicTraceLogConnWarning(
+                SkipPacketNumber,
+                Connection,
+                "Skipped packet number %llu",
+                Connection->Send.SkippedPacketNumber);
+
+            //
+            // Randomly skip a packet number (from 0 to 65535).
```
csujedihy, line 395: "The server skips a packet number with a uniform probability of at most one every 131,072 packets"

Just pasting  how quiche fixed the issue here just in case you missed it in the paper. I don't have an opinion on this.

nibanks: I decided to go a slightly different direction than quiche. 😄

src/core/packet_builder.c
```diff
@@ -379,6 +379,27 @@ QuicPacketBuilderPrepare(
             Builder->Metadata->PacketId,
             Builder->BatchId);
 
+        //
+        // Occassionally skip a packet number for improved security.
```
guhetier, line 383: nit: I think it would be nice to give a bit more context.
Right now, from reading the code without a knowing what this is for, it would be hard to find how this improve security / what is the problem. Maybe having "opportunistic ack protection" somewhere would help finding relevant info?

src/core/packet_builder.c
guhetier: What are the chances that we send `NextSkippedPacketNumber` before we receive the hypothetical ack for `SkippedPacketNumber`? If we send skipped ack to fast and keep track only of the last one, we could defeat our own protection.

How fast are we generally consuming packets numbers? I don't have a good estimate in mind.

nibanks: It depends on the speed and RTT. But at the end of the day, this is a best effort feature.

src/core/loss_detection.c
```diff
@@ -1329,6 +1329,22 @@ QuicLossDetectionProcessAckBlocks(
...
+        if (Connection->Send.SkippedPacketNumber >= AckBlock->Low &&
+            Connection->Send.SkippedPacketNumber <= QuicRangeGetHigh(AckBlock)) {
+            QuicTraceLogConnError(
+                AttackDetected,
+                Connection,
+                "Attack detected: Skipped packet number %llu ACKed in range [%llu, %llu]",
+                Connection->Send.SkippedPacketNumber,
+                AckBlock->Low,
+                QuicRangeGetHigh(AckBlock));
+            QuicConnTransportError(Connection, QUIC_ERROR_PROTOCOL_VIOLATION);
```
sm-msft, line 1345: Is this reported back to the receiver?

src/core/loss_detection.c
```diff
@@ -1329,6 +1329,22 @@ QuicLossDetectionProcessAckBlocks(
     uint32_t i = 0;
     QUIC_SUBRANGE* AckBlock;
     while ((AckBlock = QuicRangeGetSafe(AckBlocks, i++)) != NULL) {
+        //
+        // ATTACK DETECTION: Check if the skipped packet number is in this ACK
+        // block. If so, this indicates a potential injection attack.
+        //
+        if (Connection->Send.SkippedPacketNumber >= AckBlock->Low &&
+            Connection->Send.SkippedPacketNumber <= QuicRangeGetHigh(AckBlock)) {
```
sm-msft, line 1337: Are these [low , high] ackblock values contiguous?

nibanks: Yes. It's a range of packet numbers being acknowledged. You can see further examples below using it.

src/core/loss_detection.c
nibanks: Yes.

src/core/packet_builder.c
nibanks: There's a balance here. On one hand, I agree. On the other, the more info you provide, the more info you provide to would-be attackers too. So, for now at least, I'm going to leave as is.

PR #5147, opened 2025-06-07
Add Kernel Static Assert Correct Version is Targetted

nibanks:
## Description

This adds a static assert to the kernel header to ensure we're targetting the right downlevel release (Fe). Otherwise, it can break back-compat.

## Testing

CI/CD

## Documentation

N/A


guhetier (approved):
LGTM

PR #5132, opened 2025-05-30
Additional settings tests and misc changes

sm-msft:
This PR includes the settings related test changes from this original draft PR: https://github.com/microsoft/msquic/pull/5114

## Description

- New Windows user-mode storage API for writing MsQuic test settings
- Additional settings tests, including tests with stored settings (windows user mode only)
- Minor build documentation change

## Testing

Existing tests + new tests

## Documentation

N/A

PR #5165, opened 2025-06-13
CP: Add Writability to Storage Abstraction (#5133)

anrossi:
* Add Writability to Storage Abstraction

* Add tests from #5132

### DIFF
--- a/src/core/unittest/SettingsTest.cpp
+++ b/src/core/unittest/SettingsTest.cpp
@@ -9,6 +9,8 @@ Abstract:
 
 --*/
 
+#define QUIC_UNIT_TESTS
+
 #include "main.h"
 #ifdef QUIC_CLOG
 #include "SettingsTest.cpp.clog.h"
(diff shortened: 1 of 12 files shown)


guhetier:
Just to understand, why do we need to cherry-pick this to 2.5? (Could be nice to add in the PR description for posterity)

anrossi:
@guhetier This PR provides the foundation for tests that #5112 needs. #5112 will also be cherry-picked as it's the last feature needed for 2.5 release. I'll add this in the description in the future.

src/inc/quic_storage.h
```diff
@@ -69,6 +79,48 @@ CxPlatStorageReadValue(
...
+    // Non-Windows Registry types begin at 16 or above.
+    //
+} CXPLAT_STORAGE_TYPE;
+
+//
+// Writes a value to the storage context
+//
+_IRQL_requires_max_(PASSIVE_LEVEL)
+QUIC_STATUS
+CxPlatStorageWriteValue(
```
sm-msft, line 96: I dont recall if we discussed this already. 

These are quite risky functions from production perspective. We should surround these declarations with a #define and allow the functions to be used in the test code for now. If we need it in the msquic lib, we can discuss each usage specifically.

sm-msft: SettingsTest.cpp uses QUIC_UNIT_TESTS macro for this.

src/core/unittest/SettingsTest.cpp
```diff
@@ -335,6 +337,247 @@ TEST(SettingsTest, StreamRecvWindowDefaultGetsOverridenByIndividualLimits)
...
+    QuicStorageSettingScopeGuard(
+        _In_opt_ const char* StorageName)
+    {
+        EXPECT_EQ(
+            QUIC_STATUS_SUCCESS,
+            CxPlatStorageOpen(
+                StorageName,
+                nullptr,
+                nullptr,
+                CXPLAT_STORAGE_OPEN_FLAG_DELETE | CXPLAT_STORAGE_OPEN_FLAG_WRITE | CXPLAT_STORAGE_OPEN_FLAG_CREATE,
```
sm-msft, line 410: Nit: Perhaps this line is too long?
A scope-local param "flags" could be set to these flags and used instead for better readability.

src/platform/storage_winkernel.c
```diff
@@ -73,6 +79,34 @@ ZwNotifyChangeKey(
     _In_ BOOLEAN Asynchronous
     );
 
+//
+// Copied from wdm.h
```
sm-msft, line 83: We have a couple of references to wdm.h here. Just curious if it is a public file.

guhetier:
Your "retry" PR made some extra changes to the "clear" function.
- Will that other PR be cherry-picked too?
- If not, do we need to bring that change in?

guhetier:
The change looks good to me, but can you check the test failure is unrelated?
It seems likely since storage should only affect Windows, but let's do our due diligences.

anrossi:
> Your "retry" PR made some extra changes to the "clear" function.
> 
> * Will that other PR be cherry-picked too?
> * If not, do we need to bring that change in?

yes, that PR will also be cherry-picked once it is completed.

src/test/lib/ApiTest.cpp
```diff
@@ -5980,549 +5980,343 @@ QuicTestCredentialLoad(const QUIC_CREDENTIAL_CONFIG* Config)
     TEST_QUIC_SUCCEEDED(Configuration.LoadCredential(Config));
 }
 
+
+class QuicStorageSettingScopeGuard {
```
sm-msft, line 5984: Perhaps a comment to mention this is duplicated in the settings test as well...

src/test/lib/ApiTest.cpp
```diff
@@ -5980,549 +5980,343 @@ QuicTestCredentialLoad(const QUIC_CREDENTIAL_CONFIG* Config)
...
+private:
+    QuicStorageSettingScopeGuard(
+        _In_opt_ const char* StorageName)
+    {
+        TEST_QUIC_SUCCEEDED(
+            CxPlatStorageOpen(
+                StorageName,
+                nullptr,
+                nullptr,
+                CXPLAT_STORAGE_OPEN_FLAG_DELETE | CXPLAT_STORAGE_OPEN_FLAG_WRITE | CXPLAT_STORAGE_OPEN_FLAG_CREATE,
```
sm-msft, line 6033: Please see my comment in the settings unit test code.

src/platform/storage_winkernel.c
```diff
@@ -294,50 +334,84 @@ CxPlatStorageOpen(
 
     CxPlatZeroMemory(Storage, sizeof(CXPLAT_STORAGE));
     CxPlatLockInitialize(&Storage->Lock);
-    Storage->Callback = Callback;
-    Storage->CallbackContext = CallbackContext;
+    if (Callback != NULL) {
```
sm-msft, line 337: Should we also validate or assert that CallbackContext is not NULL?

src/platform/storage_winuser.c
```diff
@@ -121,99 +126,136 @@ CxPlatStorageOpen(
     }
 
     CxPlatZeroMemory(Storage, sizeof(CXPLAT_STORAGE));
-    Storage->Callback = Callback;
-    Storage->CallbackContext = CallbackContext;
+    if (Callback != NULL) {
```
sm-msft, line 129: Should we also check/assert if CallbackContext is not NULL as well?

PR #5182, opened 2025-06-24
Removed RegDeleteTreeA usage in storage_winuser

gaurav2699:
## Description

Gamecore builds are failing because it is not able to find the definition of RegDeleteTreeA in the undock pipelines.

## Testing

## Documentation


nibanks:
It seems like there is a missing include instead.

mtfriesen (approved):

mtfriesen:
> It seems like there is a missing include instead.

I was wondering how this could be broken, but the documentation for `RegDeleteTreeA` says to include `windows.h` which we surely haven't missed?

anrossi:
@nibanks I think gamecore specifically doesn't define the registry API, so we must define it here ourselves to build.

src/platform/storage_winuser.c
```diff
@@ -57,6 +57,14 @@ RegNotifyChangeKeyValue(
     _In_opt_ HANDLE hEvent,
     _In_ BOOL fAsynchronous
     );
+
+WINADVAPI
```
guhetier: nit: Given the slowly increasing number of "Restricted build" declarations, should we consider grouping them in a header made specifically for this purpose?
It would keep them out of the way for the 99% of the time we are not concerned with them :)

nibanks:
I missed that this was gamecore specific. Generally, we just have to remove usage for things not on gamecore.

mtfriesen:
> I missed that this was gamecore specific. Generally, we just have to remove usage for things not on gamecore.

Yeah, now that I read the documentation for these partitions more carefully, if we declare functions ourselves, it's possible for Gamecore to successfully compile and maybe even link to APIs that will not be present on every Gamecore system, leading to runtime failures down the line.

guhetier:
Oh, that would be bad and annoying.

Also, I just remembered that #5112 actually remove the use of `RegDeleteTreeA` and replace it with an iteration over the relevant keys. This PR might not be needed if we merge #5112.

gaurav2699:
Added the storage_winuser part from #5112 as discussed with Anthony to unblock gamecore builds as the PR still requires changes before it gets merged.

guhetier:
You need to update the CLog manifest (scripts/update-sidecar.ps1)

src/platform/storage_winuser.c
```diff
@@ -348,5 +348,116 @@ CxPlatStorageClear(
...
+        while (Error == ERROR_MORE_DATA && AllocatedLength < 32767) {
+            //
+            // Reallocate buffer only if current buffer is too small
+            //
+            CXPLAT_DBG_ASSERT(ValueName != NULL);
+
+            CXPLAT_FREE(ValueName, QUIC_POOL_PLATFORM_TMP_ALLOC);
+
+            AllocatedLength *= 2; // The API doesn't tell us the needed length; just double it.
+            if (AllocatedLength > 32767) {
```
mtfriesen:
```suggestion
            if (AllocatedLength > MAXSHORT) {
```

src/platform/storage_winuser.c
```diff
@@ -348,5 +348,116 @@ CxPlatStorageClear(
...
+        QuicTraceEvent(
+            AllocFailure,
+            "Allocation of '%s' failed. (%llu bytes)",
+            "RegEnumValueA ValueName",
+            AllocatedLength);
+        goto Exit;
+    }
+
+    //
+    // Iterate through all values and delete them
```
mtfriesen, line 404: I expect the old RegDeleteTreeA would recurse, too. Is that true, and would we ever rely on the recursion?

guhetier: Because I had a similar comment on #5112: it does recurse, and that was a problem, because it would delete both the global settings and all app specific settings given the structure of the data in the registry.

The function is meant to delete only the global or the app data (depending on the storage used), not to wipe everything.

src/platform/storage_winuser.c
```diff
@@ -348,5 +348,116 @@ CxPlatStorageClear(
...
+
+        while (Error == ERROR_MORE_DATA && AllocatedLength < 32767) {
+            //
+            // Reallocate buffer only if current buffer is too small
+            //
+            CXPLAT_DBG_ASSERT(ValueName != NULL);
+
+            CXPLAT_FREE(ValueName, QUIC_POOL_PLATFORM_TMP_ALLOC);
+
+            AllocatedLength *= 2; // The API doesn't tell us the needed length; just double it.
```
mtfriesen: Apparently the `RegQueryInfoKeyA` function returns `lpcbMaxValueNameLen` which is the answer.

src/platform/storage_winuser.c
mtfriesen: Thanks for the context. Let's put that design decision (and now API contract) into the comments for this function.

(Bonus points if you have a link to your original comment handy)

guhetier: Here  you go: https://github.com/microsoft/msquic/pull/5112#discussion_r2150964425

src/platform/storage_winuser.c
anrossi: Ah, so at the top call `RegQueryInfoKeyA` to get the initial value to use for `AllocatedLength`?

mtfriesen: Yup, that's the theory.

mtfriesen (approved):
LGTM as-is, though the `RegQueryInfoKeyA` approach may be tidier.

src/platform/storage_winuser.c
guhetier: +1 on using it then, it seems to me like a more natural pattern!

src/platform/storage_winuser.c
```diff
@@ -348,5 +348,122 @@ CxPlatStorageClear(
...
+        }
+    }
+
+Exit:
+    if (ValueName != NULL) {
+        CXPLAT_FREE(ValueName, QUIC_POOL_PLATFORM_TMP_ALLOC);
+    }
+
+    return Status;
+}
```
anrossi: nit: Don't forget to add the newline at the end of the document.

anrossi:
Agree with Michael's comment, but this looks good too.

anrossi (approved):

src/platform/storage_winuser.c
gaurav2699: I have modified it according to the suggested method, please see if it looks fine. Thanks.

src/platform/storage_winuser.c
```diff
@@ -348,5 +348,107 @@ CxPlatStorageClear(
...
+        NULL,                   // SubKeys
+        NULL,                   // MaxSubKeyLen
+        NULL,                   // MaxClassLen
+        NULL,                   // Values
+        &AllocatedLength,       // MaxValueNameLen
+        NULL,                   // MaxValueLen
+        NULL,                   // SecurityDescriptor
+        NULL);                  // LastWriteTime
+
+    if (Error != NO_ERROR) {
```
nibanks:
```suggestion
        NULL);                  // LastWriteTime
    if (Error != NO_ERROR) {
```

nibanks (approved):

guhetier (approved):

PR #5216, opened 2025-07-02
CP: Allow Configuration of Stateless Retry Key (#5112)

anrossi:
Cherry pick from #5112

### DIFF
--- a/docs/Settings.md
+++ b/docs/Settings.md
@@ -90,6 +90,27 @@ All restrictions and effects on the versions mentioned in [QUIC_VERSION_SETTINGS
 
 Particularly, on server, these must be set **GLOBALLY** if you want them to take effect for servers.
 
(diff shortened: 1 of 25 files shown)


sm-msft:
I think we need to discuss the changes that go into release 2.5 at this time.

src/platform/tls_quictls.c
nibanks, line 1: FYI, thanks for this. I think this is causing msh3 to fail to build on the latest atm.

nibanks (approved):

guhetier:
It seems strange to me to have an API change that is going to ship in v2.5.**1**, which is a patch version.

On the one hand, some changes leading to it have already been cherry-picked. On the other hand, we could also very easily create a v2.6.

Opinions?

anrossi:
I don't want to break the rules around semantic versioning unless there's a really good reason, but I need to make sure the partners who requested this feature can go with 2.6 or if they need it in 2.5. I think they need it in 2.5.

PR #5219, opened 2025-07-02
Fix size check for QUIC_HANDSHAKE_INFO

### LINKED ISSUE #5217: [API] OpenSSL 3.5 QUIC_HANDSHAKE_INFO
`QUIC_HANDSHAKE_INFO` used to **NOT** contain `internal QUIC_TLS_GROUP TlsGroup;`. 

The new OpenSSL implementation ignores this and using this version of lib with older API definition fails with `QUIC_STATUS_BUFFER_TOO_SMALL`.
https://github.com/microsoft/msquic/blob/2623c07df62b4bd171f469fb29c2714b6735b676/src/platform/tls_openssl.c#L3433-L3437

Note that both quictls and schannel are backward compatible and check for older size, e.g.:
https://github.com/microsoft/msquic/blob/2623c07df62b4bd171f469fb29c2714b6735b676/src/platform/tls_quictls.c#L2290-L2295

cc @nhorman

nhorman:
Size checks on QUIC_HANDSHAKE_INFO need to take into account backward compatibility, as some prior versions were a shorter length.  Quictls and schannel account for this but openssl does not.

Fix that up

Fixes #5217

### DIFF
--- a/src/platform/tls_openssl.c
+++ b/src/platform/tls_openssl.c
@@ -3431,7 +3431,7 @@ CxPlatTlsParamGet(
     switch (Param) {
 
         case QUIC_PARAM_TLS_HANDSHAKE_INFO: {
-            if (*BufferLength < sizeof(QUIC_HANDSHAKE_INFO)) {
+            if (*BufferLength < CXPLAT_STRUCT_SIZE_THRU_FIELD(QUIC_HANDSHAKE_INFO, CipherSuite)) {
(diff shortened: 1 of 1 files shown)


nhorman:
@guhetier It has, or at least its part of the CI.  There may be missing test cases I'm unaware of however

guhetier:
> @guhetier It has, or at least its part of the CI. There may be missing test cases I'm unaware of however

Good to know, thanks! We might have to follow up to add a test case for querying the TLS params with various buffer sizes in that case.

PR #5220, opened 2025-07-02
More silo enlightenment

csujedihy:
## Description

Attach to the right silo when uninitializing a connection. Otherwise, the security context could be freed in a silo that's not where the context was allocated.

## Testing

CI and local testing

## Documentation

N/A

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -375,6 +375,11 @@ QuicConnFree(
     QuicDatagramUninitialize(&Connection->Datagram);
     if (Connection->Configuration != NULL) {
 #ifdef QUIC_SILO
+        //
+        // Take a ref on the silo before releasing the configuration
+        // to prevent the silo from being destroyed while we are still
+        // holding onto the thread to clean up other stuff for this connection.
+        //
         Silo = Connection->Configuration->Silo;
         QuicSiloAddRef(Silo);
 #endif
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -309,6 +309,7 @@ QuicConnFree(
     )
 {
     QUIC_PARTITION* Partition = Connection->Partition;
+    QuicConfigurationAttachSilo(Connection->Configuration);
     CXPLAT_FRE_ASSERT(!Connection->State.Freed);
     CXPLAT_TEL_ASSERT(Connection->RefCount == 0);
     if (Connection->State.ExternalOwner) {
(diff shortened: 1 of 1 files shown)


guhetier:
Can you please add a few more details about what this does / why it is needed (or link to a bug with those details if one exists)?
I don't know much about silos, so I don't know why we need to enter it in this function.

src/core/connection.c
```diff
@@ -408,6 +409,7 @@ QuicConnFree(
     InterlockedDecrement(&MsQuicLib.ConnectionCount);
 #endif
     QuicPerfCounterDecrement(Partition, QUIC_PERF_COUNTER_CONN_ACTIVE);
+    QuicConfigurationDetachSilo();
```
anrossi, line 426: The call to `QuicConfigurationRelease` on line 373 calls `QuicConfigurationUninitialize` if the ref count drops to zero, which calls `QuicSiloRelease`. Will this cause a crash if we detach from a silo that has been released? Or will the silo release crash because the silo is currently in use?

csujedihy: Good catch. Technically, detaching from a released silo is fine. Keeping the thread in a released silo might be problematic. I will move the detach up.

src/core/connection.c
```diff
@@ -317,7 +317,9 @@ QuicConnFree(
     CXPLAT_TEL_ASSERT(Connection->SourceCids.Next == NULL);
     CXPLAT_TEL_ASSERT(CxPlatListIsEmpty(&Connection->Streams.ClosedStreams));
     QuicRangeUninitialize(&Connection->DecodedAckRanges);
+    QuicConfigurationAttachSilo(Connection->Configuration);
```
nibanks: If this guaranteed to be non-null?

csujedihy: `QuicConfigurationAttachSilo` =>

```
#define QuicConfigurationAttachSilo(Configuration) \
    QUIC_SILO PrevSilo = (Configuration == NULL || Configuration->Silo == NULL) ? \
        QUIC_SILO_INVALID : QuicSiloAttach(Configuration->Silo)
```

csujedihy: We also deref configuration in this function. So, it actually has to be non-null.

anrossi (approved):

src/core/connection.c
csujedihy: Now, the entire function is run in the silo where the config was allocated, which is identical to the behavior before #3938

nibanks (approved):

PR #5227, opened 2025-07-07
Fix MDL mapping size when allocating send buffers in Windows kernel mode

guhetier:
## Description

#4929 added a header to all pool allocation to simplify their management.
However, the computation of the actual data buffer size was not adapted when initializing the MDL in Windows Kernel mode and overestimated the buffer size by the size of the header.

This PR corrects the size computation.

## Testing

CI + running HTTP stress tests that discovered the issue.

## Documentation

N/A


csujedihy (approved):

mtfriesen:
@microsoft-github-policy-service rerun

PR #5228, opened 2025-07-08
Fixing bug in CxPlatDataPathResolveAddress due to different definitio…

Johan511:
`CxPlatDataPathResolveAddress` in [`src/platform/datapath_unix.c`](https://github.com/microsoft/msquic/blob/main/src/platform/datapath_unix.c#L102-L163) forwards the return value of `getaddrinfo` as `QUIC_STATUS` to signal the success/failure of the function

This is incorrect, the return value of `getaddrinfo` indicates success if 0 else it is to be considered as failure
[Reference: `getaddrinfo` man page](https://man7.org/linux/man-pages/man3/getaddrinfo.3.html#RETURN_VALUE)

In case of QUIC_STATUS, a positive value indicates failure and a negative value indicates success.

This patch fixes this.

PS: All the error return values are negative for `getaddrinfo`, hence the existing implementation would never indicate an error.


src/platform/datapath_unix.c
```diff
@@ -155,9 +154,10 @@ CxPlatDataPathResolveAddress(
         "[%p] Couldn't resolve hostname '%s' to an IP address",
         Datapath,
         HostName);
-    Status = (QUIC_STATUS)Result;
 
 Exit:
-
-    return Status;
+    if(Result != 0)
+        return QUIC_STATUS_INVALID_PARAMETER;
```
ProjectsByJackHe: So, according to the manpage, there are multiple error types (https://man7.org/linux/man-pages/man3/getaddrinfo.3.html#RETURN_VALUE), why is it ok to encapsulate all of them with QUIC_STATUS_INVALID_PARAMETER? Should we have some logic here to better map unix errors with QUIC errors?

csujedihy:
The description is slightly off- "a negative value indicates success." => "a non-positive value indicates success."

src/platform/datapath_unix.c
```diff
@@ -155,9 +154,10 @@ CxPlatDataPathResolveAddress(
         "[%p] Couldn't resolve hostname '%s' to an IP address",
         Datapath,
         HostName);
-    Status = (QUIC_STATUS)Result;
 
 Exit:
-
-    return Status;
+    if(Result != 0)
```
csujedihy: Please ensure the coding style is consistent throughout this function.

```
if (Result != 0) {
else {
    return QUIC_STATUS_SUCCESS
}
```

src/platform/datapath_unix.c
guhetier: Interestingly, the windows specific version of this maps uses `HRESULT_FROM_WIN32(WSAHOST_NOT_FOUND)`, which doesn't have a `QUIC_STATUS` explicit equivalent.

Maybe we could define `QUIC_STATUS_HOST_NOT_FOUND` and use it in both cases.

nibanks: I think a platform specific host not found error would be good.

Johan511: > Maybe we could define QUIC_STATUS_HOST_NOT_FOUND and use it in both cases.

All MsQuic errors we use right now are a mapping from certain posix errors (or windows errors)
Can you suggest which error should it be mapped to (for POSIX)?

Also to answer @ProjectsByJackHe 
I do not see much value is translating each of the return values of `getaddrinfo` into the respective `QUIC_STATUS_*`. Please let me know if you think otherwise.

I can get behind a `QUIC_STATUS_ADDR_RESOLVE_ERROR` but not an equivalent for each error possible by `getaddrinfo`

Also there exists a `QUIC_STATUS_ADDRESS_NOT_AVAILABLE` which we never seem to use and is defined only in `msquic_posix.h` and not in it's windows counterpart (probably the reason we do not use it). Any opinions on it?

Johan511:
If this issue is being discussed in depth
Just want to tag another issue I have with which led me to finding this bug

Over [here](https://github.com/Johan511/msquic//blob/a8d66820db2c8d9a2a4c1f4c699665ac362d134e/src/inc/msquic_posix.h#L195) we define `QUIC_LOCALHOST_FOR_AF(Af)` as `localhost` or `ip6-localhost` on posix machines

I use a posix machine and my `/etc/hosts` does not have an entry for `ip6-localhost` which led to a failure in `getaddrinfo` (which was not found in `getaddrinfo` which led to this PR)

In case of windows we always use `localhost`, any opinions on changing the posix variant to always be `localhost` too?

src/platform/datapath_unix.c
guhetier: `QUIC_STATUS_ADDRESS_NOT_AVAILABLE ` was added in https://github.com/microsoft/msquic/pull/3038. I am not sure of the context, but I would be cautious before modifying it.

Maybe it could be re-used if it makes sense (I am not familiar enough with `EADDRNOTAVAIL` to know if it would make sense to use it here, and mapped to `HRESULT_FROM_WIN32(WSAHOST_NOT_FOUND)` for Windows.
Or we can go with a new name if another Unix error code makes more sense here.

Johan511: Looked into man pages for `EADDRNOTAVAIL`, it seems to mean the socket addr is not valid
Nothing to do with IP address or IP path

guhetier: Would `EHOSTUNREACH` make sense?

Johan511: > EHOSTUNREACH
              No valid routing table entry matches the destination
              address.  This error can be caused by an ICMP message from
              a remote router or for the local routing table.

I'm not well versed with windows so will discuss from a POSIX perspective 

This will lead to all the error codes being represented by HOST UNREACHABLE 

What if getaddrfamily fails because the application mentioned it is ipv4 but the string is provides is unmistakeably ipv6?

Either we go overkill and add all these error codes in QUIC

Or return a simple EINVAL

guhetier: My angle is more from the point of view of the caller of `ConnectionStart`. It would be nice if we could fail with an error code that indicate "something went wrong with the address resolution" so the app developer has an idea on what to start with when debugging.

But that is honestly a minor concern, and EINVAL already be better than the current state (with the equivalent for the windows side)

Johan511: Yes, you are right.

I have added a macro `NETDB_ERR_CODE_TO_QUIC_STATUS`
Can you please review it?

Johan511: Alternatively we can have 

```
#define QUIC_STATUS_NETDB_GENERIC_ERROR     NETDB_ERR_CODE_TO_QUIC_STATUS(0)

#ifdef EAI_*
#define QUIC_STATUS_NETDB_ERR_*      NETDBNETDB_ERR_CODE_TO_QUIC_STATUS(EAI_*)
#else
#define QUIC_STATUS_NETDB_ERR_* QUIC_STATUS_GENERIC_NETDB_ERR
#endif
```

This would make translating each of these codes to their windows counter parts much easier and vice versa
Also removes the issue of getting the platform / libc version support for an error code wrong

src/inc/msquic_posix.h
```diff
@@ -133,6 +134,38 @@ QUIC_INLINE ENUMTYPE &operator ^= (ENUMTYPE &a, ENUMTYPE b) throw() { return (EN
 #define QUIC_STATUS_ALPN_IN_USE             ((QUIC_STATUS)EPROTOTYPE)       // 91   (41 on macOS)
 #define QUIC_STATUS_ADDRESS_NOT_AVAILABLE   ((QUIC_STATUS)EADDRNOTAVAIL)    // 99   (47 on macOS)
 
+#define NETDB_ERR_CODE_TO_QUIC_STATUS(Alert)        (Alert + QUIC_NET_DB_ERROR_BASE)
+
+# define QUIC_STATUS_BADFLAGS               NETDB_ERR_CODE_TO_QUIC_STATUS(EAI_BADFLAGS) // 200001023
```
guhetier, line 139: I am not sure I like this solution.

Largely (ignoring `QUIC_STATUS_ADDRESS_NOT_AVAILABLE` which seems a strange case), `QUIC_STATUS` are defined for both Windows and Unix with similar names (and different values based on what makes sense for the OS). This allows the application code to be platform agnostic.

If we add `QUIC_STATUS` in the Unix header only, the application code stops being compatible cross-plateform if it uses one of them.

Unless we can find an ERRNO that would make sense as a generic "address resolution failed" (in which case, we can define a single new `QUIC_STATUS_ADDRESS_RESOLUTION_FAILED` mapped to `HRESULT_FROM_WIN32(WSAHOST_NOT_FOUND)` on Windows and that ERRNO on Unix), I would rather simply use one of the existing `QUIC_STATUS`, like `QUIC_STATUS_UNREACHABLE` or `QUIC_STATUS_INVALID_PARAMETER`.

Issue #5193, opened 2025-06-28
Use standardized KDF with retry


anrossi:
Stateless retry keys are generated using a non-standard KDF. 
Use a standardized KDF, SP800-108 Hashed Counter, to generate the key.

This work will involve exposing a new crypto API through the CXPLAT interface, for both bcrypt and libcrypto (openSSL), which perform this functionality.

PR #5232, opened 2025-07-10
Use standard KDF for stateless retry

anrossi:
## Description

Addresses #5193

## Testing

New unit tests added to validate implementation is correctly using the underlying crypto libraries correctly.

## Documentation

No documentation change is needed.

### DIFF
--- a/src/core/partition.c
+++ b/src/core/partition.c
@@ -98,19 +98,24 @@ QuicPartitionGetStatelessRetryKey(
     }
 
     //
-    // Generate a new key from the base retry key combined with the index.
+    // Generate a new key from the base retry secret using SP800-108 CTR-HMAC.
     //
-    uint8_t RawKey[CXPLAT_AEAD_AES_256_GCM_SIZE];
-    CxPlatCopyMemory(
-        RawKey,
-        MsQuicLib.StatelessRetry.BaseSecret,
-        MsQuicLib.StatelessRetry.SecretLength);
-    for (size_t i = 0; i < sizeof(KeyIndex); ++i) {
-        RawKey[i] ^= ((uint8_t*)&KeyIndex)[i];
+    uint8_t RawKey[CXPLAT_AEAD_MAX_SIZE];
+    QUIC_STATUS Status =
+        CxPlatKbKdfDerive(
+            MsQuicLib.StatelessRetry.BaseSecret,
+            MsQuicLib.StatelessRetry.SecretLength,
+            "QUIC Stateless Retry Key",
+            (uint8_t*)&KeyIndex,
+            sizeof(KeyIndex),
+            MsQuicLib.StatelessRetry.SecretLength,
+            RawKey);
+    if (QUIC_FAILED(Status)) {
+        return NULL;
     }
 
     CXPLAT_KEY* NewKey;
-    QUIC_STATUS Status =
+    Status =
         CxPlatKeyCreate(
             MsQuicLib.StatelessRetry.AeadAlgorithm,
             RawKey,
@@ -121,12 +126,14 @@ QuicPartitionGetStatelessRetryKey(
             "[ lib] ERROR, %u, %s.",
             Status,
             "Create stateless retry key");
+        CxPlatSecureZeroMemory(RawKey, sizeof(RawKey));
         return NULL;
     }
 
     CxPlatKeyFree(Partition->StatelessRetryKeys[KeyIndex & 1].Key);
     Partition->StatelessRetryKeys[KeyIndex & 1].Key = NewKey;
     Partition->StatelessRetryKeys[KeyIndex & 1].Index = KeyIndex;
+    CxPlatSecureZeroMemory(RawKey, sizeof(RawKey));
 
     return NewKey;
 }
--- a/src/platform/crypt_bcrypt.c
+++ b/src/platform/crypt_bcrypt.c
@@ -807,7 +807,7 @@ CxPlatKbKdfDerive(
     ULONG LabelLength = 0;
     ULONG DerivedKeyLength;
 
-    LabelLength = strnlen_s(Label, 255);
+    LabelLength = (ULONG)strnlen_s(Label, 255);
 
     // Create symmetric key from the secret using HMAC-SHA256
     Status =
--- a/src/platform/crypt_bcrypt.c
+++ b/src/platform/crypt_bcrypt.c
@@ -46,12 +46,14 @@ BCRYPT_ALG_HANDLE CXPLAT_HMAC_SHA384_ALG_HANDLE;
 BCRYPT_ALG_HANDLE CXPLAT_HMAC_SHA512_ALG_HANDLE;
 BCRYPT_ALG_HANDLE CXPLAT_AES_ECB_ALG_HANDLE;
 BCRYPT_ALG_HANDLE CXPLAT_AES_GCM_ALG_HANDLE;
+BCRYPT_ALG_HANDLE CXPLAT_SP800108_CTR_HMAC_ALG_HANDLE;
 #else
 BCRYPT_ALG_HANDLE CXPLAT_HMAC_SHA256_ALG_HANDLE = BCRYPT_HMAC_SHA256_ALG_HANDLE;
 BCRYPT_ALG_HANDLE CXPLAT_HMAC_SHA384_ALG_HANDLE = BCRYPT_HMAC_SHA384_ALG_HANDLE;
 BCRYPT_ALG_HANDLE CXPLAT_HMAC_SHA512_ALG_HANDLE = BCRYPT_HMAC_SHA512_ALG_HANDLE;
 BCRYPT_ALG_HANDLE CXPLAT_AES_ECB_ALG_HANDLE = BCRYPT_AES_ECB_ALG_HANDLE;
 BCRYPT_ALG_HANDLE CXPLAT_AES_GCM_ALG_HANDLE = BCRYPT_AES_GCM_ALG_HANDLE;
+BCRYPT_ALG_HANDLE CXPLAT_SP800108_CTR_HMAC_ALG_HANDLE = BCRYPT_SP800108_CTR_HMAC_ALG_HANDLE;
 #endif
 BCRYPT_ALG_HANDLE CXPLAT_CHACHA20_POLY1305_ALG_HANDLE = NULL;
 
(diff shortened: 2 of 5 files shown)


src/platform/crypt_bcrypt.c
```diff
@@ -204,6 +206,21 @@ CxPlatCryptInitialize(
         }
     }
 
+    Status =
+        BCryptOpenAlgorithmProvider(
+            CXPLAT_SP800108_CTR_HMAC_ALG_HANDLE,
+            BCRYPT_SP800108_CTR_HMAC_ALGORITHM,
+            NULL,
```
anrossi: make sure to us MS_PRIMITIVE_PROVIDER here.

src/platform/crypt_bcrypt.c
```diff
@@ -204,6 +206,21 @@ CxPlatCryptInitialize(
         }
     }
 
+    Status =
+        BCryptOpenAlgorithmProvider(
+            CXPLAT_SP800108_CTR_HMAC_ALG_HANDLE,
```
anrossi: This needs to be an address-of

src/platform/crypt_bcrypt.c
```diff
@@ -764,3 +787,87 @@ CxPlatHashCompute(
...
+
+    LabelLength = (ULONG)strnlen_s(Label, 255);
+
+    NTSTATUS Status =
+        BCryptGenerateSymmetricKey(
+            CXPLAT_SP800108_CTR_HMAC_ALG_HANDLE,
+            &KeyHandle,
+            NULL, // Let BCrypt manage the memory
+            0,
+            (PUCHAR)Secret,
```
guhetier: nit: Please avoid P... pointer alias, it interacts poorly with type specifiers like `const`

src/platform/crypt_bcrypt.c
```diff
@@ -764,3 +787,87 @@ CxPlatHashCompute(
...
+    _In_ uint32_t ContextLength,
+    _In_ uint32_t OutputLength,
+    _Out_writes_(OutputLength) uint8_t* Output
+    )
+{
+    BCRYPT_KEY_HANDLE KeyHandle = NULL;
+    BCryptBuffer ParameterList[3];
+    BCryptBufferDesc Parameters;
+    ULONG LabelLength = 0;
+    ULONG DerivedKeyLength;
```
guhetier: nit: I think we should default initialize these parameters and/or declare them closer to use.
It would make it less likely to use uninitialized memory.

src/platform/unittest/CryptTest.cpp
```diff
@@ -488,6 +518,49 @@ TEST_F(CryptTest, HpMaskAes128)
     CxPlatHpKeyFree(HpKey);
 }
 
+TEST_F(CryptTest, KbKdfDerive)
+{
+    QuicBuffer Key256("3edc6b5b8f7aadbd713732b482b8f979286e1ea3b8f8f99c30c884cfe3349b83");
+    uint64_t Context = 1752112221;
```
guhetier: nit: Label and Context stays the same throughout all tests.
Since you use a helper function, I think they should be abstracted by it, so it is easier to see what varies between test cases.

src/platform/unittest/CryptTest.cpp
```diff
@@ -348,6 +348,36 @@ struct CryptTest : public ::testing::TestWithParam<int32_t>
...
+                Label,
+                (uint8_t*)Context,
+                ContextLength,
+                OutputLength,
+                Output));
+        
+        if (memcmp(ExpectedOutput.Data, Output, ExpectedOutput.Length) != 0) {
+            LogTestBuffer("Expected Output:     ", ExpectedOutput.Data, ExpectedOutput.Length);
+            LogTestBuffer("Calculated Output:   ", Output, ExpectedOutput.Length);
+            FAIL();
```
guhetier: Please consider returning a boolean or error code and failing in the test body.
That way, the failure line makes it very easy to know which specific test case failed, rather than having to guess which of the 4 it is.

src/platform/crypt_openssl.c
```diff
@@ -711,3 +711,62 @@ CxPlatHashCompute(
...
+    _In_ uint32_t SecretLength,
+    _In_z_ const char* Label,
+    _In_reads_opt_(ContextLength) const uint8_t* Context,
+    _In_ uint32_t ContextLength,
+    _In_ uint32_t OutputLength,
+    _Out_writes_(OutputLength) uint8_t* Output
+    )
+{
+    QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
+    EVP_KDF *kdf;
```
guhetier: nit: * are usually associated with the type, not the variable in msquic

csujedihy (dismissed):

src/platform/crypt_openssl.c
csujedihy: Also nit: K should be capitalized, right?

src/platform/crypt_bcrypt.c
csujedihy: the parameter in the function definition is already PUCHAR. Would it make any difference if we change it to UCHAR*?

src/platform/crypt_bcrypt.c
```diff
@@ -764,3 +787,87 @@ CxPlatHashCompute(
...
+
+    LabelLength = (ULONG)strnlen_s(Label, 255);
+
+    NTSTATUS Status =
+        BCryptGenerateSymmetricKey(
+            CXPLAT_SP800108_CTR_HMAC_ALG_HANDLE,
```
guhetier, line 806: Question: Is this another name for KB KDF?

anrossi: This is the full name of the specific algorithm that is being used. KbKdf stands for "Key-based Key Derivation Function" and is a family of KDFs.  I was tempted by brevity to use KbKdf than to have Sp800108CtrHmacKdf everywhere in our code.  But I don't feel strongly about the name and am open to changing it to be more specific.

src/platform/crypt_bcrypt.c
guhetier: It doesn't make a difference on a specific use, it's more about building the habit/convention in general. I have seen too many `const PUCHAR` not doing what people expect.

src/platform/crypt_bcrypt.c
guhetier: Got it - I am fine with the name not referencing the algorithm, it's an implementation detail.
Is OpenSSL using the same algo by default? (I assumed KBKDF was a specific one because of the OpenSSL provider name) Or do we simply don't care much about the specific algo in this case?

src/platform/crypt_bcrypt.c
anrossi: Fair point. I thought I needed to match the type of the BCrypt API specifically, but UCHAR* is accepted (which makes sense in the C type system). I think that's C++ that needs to match the type of the function parameter.

src/platform/crypt_bcrypt.c
anrossi: Sure, I'll do that

src/platform/crypt_openssl.c
anrossi: Oops, followed the OSSL sample too closely

src/platform/crypt_bcrypt.c
anrossi: Yes, this usage of EVP_KDF-KB is the same SP800-108 Counter HMAC as BCrypt. The way to switch to a different KB KDF algorithm, is to pass different parameters to EVP_KDF-KB.

anrossi: FYI https://docs.openssl.org/3.0/man7/EVP_KDF-KB/

guhetier: Thanks for the details :)

src/platform/crypt_bcrypt.c
guhetier: `PUCHAR` is just a typedef for `UCHAR*`, it is only syntactic sugar (but the same type).
The problem is that `const PUCHAR` is not `const UCHAR*` (or `UCHAR const*`), but `UCHAR* const`: a constant pointer to a mutable `UCHAR`, instead of a mutable pointer to a constant `UCHAR`.

src/platform/unittest/CryptTest.cpp
anrossi: I wanted to be able to vary those too, but since I'm not, they can be abstracted away too.

src/platform/unittest/CryptTest.cpp
anrossi: Good point. I've done so.

src/platform/crypt_bcrypt.c
anrossi: That is quite the doozy, but I was aware of that.
I thought C++ treats the `PUCHAR` as a distinct type separate from `UCHAR*`?

guhetier: Sorry for the lecture then.
But no, C++ and C are the same on that, a typedef / using is just a different name for the same type.

anrossi: Oh it's no problem; it's a good reminder.

src/platform/unittest/CryptTest.cpp
```diff
@@ -488,6 +529,44 @@ TEST_F(CryptTest, HpMaskAes128)
     CxPlatHpKeyFree(HpKey);
 }
 
+TEST_F(CryptTest, KbKdfDerive)
+{
+    QuicBuffer Key256("3edc6b5b8f7aadbd713732b482b8f979286e1ea3b8f8f99c30c884cfe3349b83");
+
+    QuicBuffer ExpectedOutput256_Key256("B7BFF374C8928335AA41589D41084B64211876771C459C23B06BA4A2EA89B5AE");
```
sm-msft, line 536: Just curious - Are these values output from a prior run and hardcoded into the tests for automation?

src/core/partition.c
```diff
@@ -98,19 +98,24 @@ QuicPartitionGetStatelessRetryKey(
...
+    // Generate a new key from the base retry secret using SP800-108 CTR-HMAC.
     //
-    uint8_t RawKey[CXPLAT_AEAD_AES_256_GCM_SIZE];
-    CxPlatCopyMemory(
-        RawKey,
-        MsQuicLib.StatelessRetry.BaseSecret,
-        MsQuicLib.StatelessRetry.SecretLength);
-    for (size_t i = 0; i < sizeof(KeyIndex); ++i) {
-        RawKey[i] ^= ((uint8_t*)&KeyIndex)[i];
+    uint8_t RawKey[CXPLAT_AEAD_MAX_SIZE];
```
sm-msft, line 103: Do we have unit test coverage for this function? 
I understand that may be non-trivial due to the necessary setup before this function can be tested.

src/platform/unittest/CryptTest.cpp
anrossi: Yeah. I used the output from OpenSSL and hardcoded it to make sure both BCrypt and OpenSSL are outputting the same values.

src/core/partition.c
anrossi: There isn't a unit test for `QuicPartitionGetStatelessRetryKey` because it is not easily unit testable. There is a unit test for `CxPlatKbKdfDerive`.
There is an end-to-end test for stateless retry; several in fact.

src/core/partition.c
```diff
@@ -98,19 +98,24 @@ QuicPartitionGetStatelessRetryKey(
...
+    QUIC_STATUS Status =
+        CxPlatKbKdfDerive(
```
nibanks, line 105: Our algorithm for calculating stateless retry must be documented.

anrossi: Documented in #5258

PR #5240, opened 2025-07-10
CP: More silo enlightenment (#5220) - v2.4

csujedihy:
## Description

#5220 Attach to the right silo when uninitializing a connection. Otherwise, the security context could be freed in a silo that's not where the context was allocated.

## Testing

CI and local repro

## Documentation

N/A

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -312,6 +312,11 @@ QuicConnFree(
     _In_ __drv_freesMem(Mem) QUIC_CONNECTION* Connection
     )
 {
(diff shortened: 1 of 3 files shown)


csujedihy:
> nit: Should we cherry-pick the build fix PRs separately instead of bringing them with other changes?

Probably unnecessary. We won't need that fix anymore as we have gotten rid of v2.3. Also, the build fix can't be simply cherry-picked from 2.5. The code is different and I had to fix it manually.

PR #5256, opened 2025-07-16
Add `rust-toolchain` to Dependabot configuration

JamieMagee:
## Description

This change enables updates for `rust-toolchain.toml` with Dependabot.

## Testing

N/A

## Documentation

N/A

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,6 @@
 
 version: 2
+enable-beta-ecosystems: true
 updates:
 
   - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,5 +1,6 @@
 
 version: 2
+enable-beta-ecosystems: true
```
anrossi, line 3: Is this necessary?  According to the documentation here it doesn't do anything: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#enable-beta-ecosystems-

anrossi: Spoke offline and understand the usage now.

PR #5257, opened 2025-07-17
Build + Build automation + test automation changes to build binaries targeting latest WinSDK

sm-msft:
## Description
- Add a new build switch to target the binaries at the latest WinSDK version to use the latest platform features.
Binaries built on Server2025-specific WinSDK with this option will not run on Server2022 because all features may not be present downlevel. 

- Modified build and test yml files to build these binaries and test them on Server2025.

## Testing

CI

## Documentation

N/A


ProjectsByJackHe:
I know this is unrelated to your PR, but the Stress test failing is somewhat concerning. Please create an issue (or link to an existing one) to track it.

sm-msft:
There is a simpler way of achieving the same result - We can use binaries generated using the newest WinSDK on downlevel machines as well. I have a different PR with that change and am abandoning this PR.

Issue #5247, opened 2025-07-11
Kernel BVT is flaky


ProjectsByJackHe:
### Describe the bug

Quite a few CI runs contain flaky Kernel runs. Frequently the runner times out or loses connection with the GH backend (bugcheck): https://github.com/microsoft/msquic/actions/runs/16229331024/job/45829284212?pr=5237

Due to the unavailability of memory dumps, it's been challenging trying to diagnose the problem.

### Affected OS

- [x] Windows
- [ ] Linux
- [ ] macOS
- [ ] Other (specify below)

### MsQuic version

main

### Steps taken to reproduce bug

Run CI

### Expected behavior

Kernel mode BVT should pass on PRs not changing product code

### Actual outcome

Runner bugchecks or times out.

csujedihy:
Hanging kernel test: https://github.com/microsoft/msquic/actions/runs/16281186886/job/45971341570?pr=5253#logs

It seems to be relatively consistent, right after `[07/15/2025 00:55:55] Basic/WithFamilyArgs.Ecn/1 succeeded (in 0.5177443 sec)`.

anrossi:
https://github.com/microsoft/msquic/actions/runs/16339050671/job/46157435548?pr=5258

anrossi:
Another one here: https://github.com/microsoft/msquic/actions/runs/16390514964/job/46315995193?pr=5267

ProjectsByJackHe:
An idea that I want to pursue for this issue:

Allocate some of our idle Netperf machines to be used by MsQuic Kernel BVT CI as self-hosted runners and add automation that will upload any MEMORY.DMP dumps generated.

guhetier:
I though you got a local repro, was it a different issue?
Have you asked Michael about how XDP is configured to upload (sometimes at least) kernel dump to Azure Watson?

This might be easier than configuring another runner pool (unless that's what XDP is doing).

ProjectsByJackHe:
Well this was more so that we get an automation that would allow us to debug kernel crashes, without having to locally repro each time. 

My local repro was not consistent. I had to re-run many times with a debugger, and the runs that failed all hit the watchdog timeout issue.

guhetier:
I fully agree for the long term, we should have dump collection one way or another!
(I was just asking for that one)

Actually, I'm going to create a task for setting up the dump collection properly.

csujedihy:
[...]

PR #5261, opened 2025-07-17
Upgrade dependabot to support dockerfile updates

csujedihy:
## Description

Sync linux cross compile docker base image version via dependabot

## Testing

CI

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -32,3 +32,15 @@ updates:
     labels:
       - "dependencies"
       - "Language: Rust"
+
+  - package-ecosystem: "docker"
+    directory: "/.docker/ubuntu-22.04"
+    schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -32,3 +32,15 @@ updates:
     labels:
       - "dependencies"
       - "Language: Rust"
+
+  - package-ecosystem: "docker"
+    directory: "/.docker/ubuntu-22.04"
```
guhetier: nit: the doc mention "directories" to define a list instead of duplicating the config.

csujedihy: Hmm, AI was fooling me..

guhetier (approved):

guhetier:
CI failure tracked by #5247

Issue #5024, opened 2025-04-18
[CI-TEST] Handshake/WithHandshakeArgs10.HandshakeSpecificLossPatterns/2


nibanks:
### Describe the bug

This test has been flaky on Linux recently:

https://github.com/microsoft/msquic/actions/runs/14539110187/job/40793694940?pr=5021

```
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
Initializing for User Mode tests
[----------] 1 test from Handshake/WithHandshakeArgs10
[ RUN      ] Handshake/WithHandshakeArgs10.HandshakeSpecificLossPatterns/2

bash: line 1: 37327 Aborted                 (core dumped) LSAN_OPTIONS=report_objects=1 ASAN_OPTIONS=disable_coredump=0:abort_on_error=1 UBSAN_OPTIONS=halt_on_error=1:print_stacktrace=1 /home/runner/work/msquic/msquic/artifacts/bin/linux/x64_Debug_openssl/msquictest --gtest_catch_exceptions=0 --gtest_filter=Handshake/WithHandshakeArgs10.HandshakeSpecificLossPatterns/2 --gtest_output=xml:/home/runner/work/msquic/msquic/artifacts/logs/msquictest/04.18.2025.17.59.30/Handshake_WithHandshakeArgs10.HandshakeSpecificLossPatterns_2_1/results.xml --timeout 60000 --osRunner=ubuntu-22.04 -PfxPath:/home/runner/work/msquic/msquic/artifacts/bin/linux/x64_Debug_openssl/selfsignedservercert.pfx
```

https://github.com/microsoft/msquic/actions/runs/14539110187/artifacts/2971207093

### Affected OS

- [ ] Windows
- [x] Linux
- [ ] macOS
- [ ] Other (specify below)

### MsQuic version

main

### Steps taken to reproduce bug

Run CI/CD

### Expected behavior

All tests pass

### Actual outcome

Sometimes this test fails

anrossi:
Moved from 2506 to 2507

guhetier:
Adding some link to logs:

Failure on Linux with logs: https://github.com/microsoft/msquic/actions/runs/15797479560/artifacts/3375763515
Failure on Windows: https://github.com/microsoft/msquic/actions/runs/15797479560/artifacts/3375873810

guhetier:
Similar failure here? https://github.com/microsoft/msquic/actions/runs/16209226758/job/45814518534?pr=5241

[...]

PR #5276, opened 2025-07-24
Use O0 instead of Og in Linux debug builds

guhetier:
## Description

Optimization level -Og doesn't allow proper debugging, as most functions and variables still end up being optimized.
This PR uses -O0 instead for Linux debug builds.

Related to #5024

## Testing

CI.

## Documentation

N/A

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -646,7 +646,7 @@ else() #!WIN32
         set(MARCH -march=native)
     endif()
 
-    set(CMAKE_C_FLAGS_DEBUG "-Og -fno-omit-frame-pointer")
+    set(CMAKE_C_FLAGS_DEBUG "-O0 -fno-omit-frame-pointer")
     set(CMAKE_C_FLAGS_MINSIZEREL "-Os -DNDEBUG")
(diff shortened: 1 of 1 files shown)


csujedihy (approved):

CMakeLists.txt
```diff
@@ -646,7 +646,7 @@ else() #!WIN32
         set(MARCH -march=native)
     endif()
 
-    set(CMAKE_C_FLAGS_DEBUG "-Og -fno-omit-frame-pointer")
+    set(CMAKE_C_FLAGS_DEBUG "-O0 -fno-omit-frame-pointer")
```
csujedihy, line 649: JFYI. You could also selectively disable some optimizations, e.g. `-fno-inline-functions`.

guhetier: Yes, but I am not sure it is worth hunting for them. Unless -O0 becomes a performance problem even in debug mode, the optimization with -Og were making debugging completely impossible for me.

(I may not have the same definition of "greatly interfere with debugging" than GCC people).

mtfriesen: > (I may not have the same definition of "greatly interfere with debugging" than GCC people).

:wink:

mtfriesen (approved):

PR #5288, opened 2025-07-26
Move clang-incompatible flag under separate check

anrossi:
## Description

When building with clang, -fno-var-tracking-assignments is not recognized and causes an error. This change moves the flag behind a compiler check to ensure it is applied to compilers that support it.

## Testing

CI will validate the change.

## Documentation

N/A

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -676,7 +676,11 @@ else() #!WIN32
 
     if(QUIC_ENABLE_SANITIZERS)
         message(STATUS "Configuring sanitizers")
-        list(APPEND QUIC_COMMON_FLAGS -fsanitize=address,leak,undefined,alignment -fsanitize-address-use-after-scope -Og -fno-omit-frame-pointer -fno-optimize-sibling-calls -fno-var-tracking-assignments)
+        list(APPEND QUIC_COMMON_FLAGS -fsanitize=address,leak,undefined,alignment -fsanitize-address-use-after-scope -Og -fno-omit-frame-pointer -fno-optimize-sibling-calls)
+        check_c_compiler_flag(-fno-var-tracking-assignments HAS_NO_VAR_TRACKING)
+        if (HAS_NO_VAR_TRACKING)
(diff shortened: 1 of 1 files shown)


guhetier:
Question: Which versions of Clang don't support this flag? Are there recent / relevant enough we want to support them (vs simply asking to upgrade the LLVM version used)?

guhetier:
There are multiple CI failures here that are not the 2 we are currently investigating.

Please have a look and if you are 100% confident they are not related to this change, create a new issue for each of them so we follow up.

anrossi:
> Question: Which versions of Clang don't support this flag? Are there recent / relevant enough we want to support them (vs simply asking to upgrade the LLVM version used)?

Clang 18.1.8 is the version I'm seeing the problem with, but it looks like no version of clang supports this. Looking at the runner definition, MsQuic doesn't build with clang with address sanitizer, so this case never gets hit in the MsQuic CI.

CMakeLists.txt
```diff
@@ -676,7 +676,11 @@ else() #!WIN32
 
     if(QUIC_ENABLE_SANITIZERS)
         message(STATUS "Configuring sanitizers")
-        list(APPEND QUIC_COMMON_FLAGS -fsanitize=address,leak,undefined,alignment -fsanitize-address-use-after-scope -Og -fno-omit-frame-pointer -fno-optimize-sibling-calls -fno-var-tracking-assignments)
+        list(APPEND QUIC_COMMON_FLAGS -fsanitize=address,leak,undefined,alignment -fsanitize-address-use-after-scope -Og -fno-omit-frame-pointer -fno-optimize-sibling-calls)
+        check_c_compiler_flag(-fno-var-tracking-assignments HAS_NO_VAR_TRACKING)
+        if (HAS_NO_VAR_TRACKING)
```
guhetier, line 682: nit: I'd add a comment explaining when this flag is not supported (here, for clang)

anrossi: Added a comment

guhetier (approved):

PR #5289, opened 2025-07-28
add ifdef and ifndefs for UM and KM

ProjectsByJackHe:
## Description

## Testing

## Documentation

### DIFF
--- a/src/platform/datapath_xplat.c
+++ b/src/platform/datapath_xplat.c
@@ -116,7 +116,7 @@ CxPlatDataPathIsPaddingPreferred(
 
 _IRQL_requires_max_(PASSIVE_LEVEL)
 QUIC_STATUS
-CxPlatSocketCreateUdp(
+CxPlatUMSocketCreateUdp(
     _In_ CXPLAT_DATAPATH* Datapath,
     _In_ const CXPLAT_UDP_CONFIG* Config,
     _Out_ CXPLAT_SOCKET** NewSocket
@@ -179,6 +179,69 @@ Error:
 
 _IRQL_requires_max_(PASSIVE_LEVEL)
 QUIC_STATUS
+CxPlatKMSocketCreateUdp(
+    _In_ CXPLAT_DATAPATH* Datapath,
+    _In_ const CXPLAT_UDP_CONFIG* Config,
+    _Out_ CXPLAT_SOCKET** NewSocket
+)
+{
+    QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
+
+    Status =
+        SocketCreateUdp(
+            Datapath,
+            Config,
+            NewSocket);
+    if (QUIC_FAILED(Status)) {
+        QuicTraceLogVerbose(
+            SockCreateFail,
+            "[sock] Failed to create socket, status:%d", Status);
+        goto Error;
+    }
+
+    (*NewSocket)->RawSocketAvailable = 0;
+    if (Datapath->RawDataPath) {
+        Status =
+            RawSocketCreateUdp(
+                Datapath->RawDataPath,
+                Config,
+                CxPlatSocketToRaw(*NewSocket));
+        (*NewSocket)->RawSocketAvailable = QUIC_SUCCEEDED(Status);
+        if (QUIC_FAILED(Status)) {
+            QuicTraceLogVerbose(
+                RawSockCreateFail,
+                "[sock] Failed to create raw socket, status:%d", Status);
+            if (Config->Flags & CXPLAT_SOCKET_FLAG_QTIP) {
+                CxPlatSocketDelete(*NewSocket);
+                goto Error;
+            }
+            Status = QUIC_STATUS_SUCCESS;
+        }
+    }
+
+Error:
+    return Status;
+}
+
+_IRQL_requires_max_(PASSIVE_LEVEL)
+QUIC_STATUS
+CxPlatSocketCreateUdp(
+    _In_ CXPLAT_DATAPATH* Datapath,
+    _In_ const CXPLAT_UDP_CONFIG* Config,
+    _Out_ CXPLAT_SOCKET** NewSocket
+    )
+{
+    #ifdef _KERNEL_MODE
+    return CxPlatKMSocketCreateUdp(Datapath, Config, NewSocket);
+    #endif
+
+    #ifndef _KERNEL_MODE
+    return CxPlatUMSocketCreateUdp(Datapath, Config, NewSocket);
+    #endif
+}
+
+_IRQL_requires_max_(PASSIVE_LEVEL)
+QUIC_STATUS
 CxPlatSocketCreateTcp(
     _In_ CXPLAT_DATAPATH* Datapath,
     _In_opt_ const QUIC_ADDR* LocalAddress,
@@ -231,10 +294,19 @@ CxPlatSocketGetLocalMtu(
     )
 {
     CXPLAT_DBG_ASSERT(Socket != NULL);
+    #ifndef _KERNEL_MODE
     if (Route->UseQTIP || (Socket->RawSocketAvailable &&
         !IS_LOOPBACK(Socket->RemoteAddress))) {
         return RawSocketGetLocalMtu(Route);
     }
+    #endif
+
+    #ifdef _KERNEL_MODE
+    if ((Socket->RawSocketAvailable &&
+        !IS_LOOPBACK(Socket->RemoteAddress))) {
+        return RawSocketGetLocalMtu(Route);
+    }
+    #endif
     return Socket->Mtu;
 }
 
@@ -421,6 +493,7 @@ CxPlatResolveRoute(
     _In_ CXPLAT_ROUTE_RESOLUTION_CALLBACK_HANDLER Callback
     )
 {
+    #ifndef _KERNEL_MODE
     if (Socket->HasFixedRemoteAddress) {
         //
         // For clients,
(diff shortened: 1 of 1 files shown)


StuffByJackHe:
TODO: try using ifdef/ifndef WIN_32

csujedihy:
Disable QuicTestConnectionPriority too and revert your previous changes since they are not relevant.

.github/workflows/test.yml
```diff
@@ -252,7 +252,7 @@ jobs:
     - name: Test
       shell: pwsh
       timeout-minutes: 120
-      run: scripts/test.ps1 -Config ${{ matrix.vec.config }} -Arch ${{ matrix.vec.arch }} -Tls ${{ matrix.vec.tls }} -OsRunner ${{ matrix.vec.os }} -GHA -LogProfile ${{ !inputs.log_level && 'Full.Light' || inputs.log_level }} -GenerateXmlResults -Kernel -Filter -*ValidateConfiguration:*ValidAlpnLengths:*ResumeRejection*:*ClientCertificate*:*LoadBalanced*
+      run: for ($i = 1; $i -le 15; $i++) { scripts/test.ps1 -Config ${{ matrix.vec.config }} -Arch ${{ matrix.vec.arch }} -Tls ${{ matrix.vec.tls }} -OsRunner ${{ matrix.vec.os }} -GHA -LogProfile ${{ !inputs.log_level && 'Full.Light' || inputs.log_level }} -GenerateXmlResults -Kernel -Filter Handshake/WithHandshakeArgs10.HandshakeSpecificLossPatterns/* }
```
ProjectsByJackHe: I will revert these changes once we get an approval for a fix.

src/core/connection.c
```diff
@@ -2478,12 +2478,11 @@ QuicConnSetConfiguration(
...
-        QuicConnApplyNewSettings(
-            Connection,
-            FALSE,
-            &Configuration->Settings);
-    }
+    QuicConnApplyNewSettings(
+        Connection,
+        FALSE,
+        &Configuration->Settings);
+
```
sm-msft: An active client connection may not consume new config settings - need some investigation/insight into what gets updated with this call vs what doesnt.

sm-msft:
Could you please add the technical change details to the PR description?

ProjectsByJackHe:
> Could you please add the technical change details to the PR description?

done

src/platform/datapath_xplat.c
```diff
@@ -436,6 +436,18 @@ CxPlatResolveRoute(
         Route->UseQTIP = Socket->ReserveAuxTcpSock;
     }
 
+    #ifdef _KERNEL_MODE
```
guhetier: That's a great step forward if it addresses the flakiness, but I don't think it addresses the root of the problem.
`Route` is an `_Inout_` parameter, it should be properly initialized by the caller.

These checks could be a debug assert (validating that QTIP is only enabled in configuration we support), but we should fix the caller initialization.

ProjectsByJackHe: resolved offline.

src/core/connection.c
ProjectsByJackHe: resolved offline.

src/core/path.c
```diff
@@ -26,6 +26,7 @@ QuicPathInitialize(
     )
 {
     CxPlatZeroMemory(Path, sizeof(QUIC_PATH));
+    CxPlatInitializeRoute(&Path->Route);
```
csujedihy: Redundant as it's zero'd out above.

ProjectsByJackHe: Yes, but in the future when we add new fields to Route with non-zero default values, we can just update CxPlatInitializeRoute.

src/platform/datapath_xplat.c
```diff
@@ -444,3 +448,17 @@ CxPlatResolveRoute(
...
+//
+// Initializes the CXPLAT_ROUTE structure to default values.
+//
+void
+CxPlatInitializeRoute(
+    _Out_writes_bytes_(sizeof(CXPLAT_ROUTE)) CXPLAT_ROUTE* Route
+    )
+{
+    //
+    // For platform specific logic, use the corresponding #ifdef.
```
csujedihy: It doesn't seem like there's a value having a wrapper. The platform code basically does the initialization and this initializer is basically a memory eraser. I'd just remove the wrapper.

src/core/path.c
csujedihy: The thing is you now have an initializer that only zeros out memory and the actual initializing code is all over the place. IMO, it's making the code even more confusing.

src/core/connection.c
guhetier: Can you please sum up the resolution here? I'm interested in understanding too.

src/inc/quic_datapath.h
```diff
@@ -877,6 +877,14 @@ CxPlatResolveRoute(
     _In_ CXPLAT_ROUTE_RESOLUTION_CALLBACK_HANDLER Callback
     );
 
+//
+// Initializes the CXPLAT_ROUTE structure to default values.
+//
+void
+CxPlatInitializeRoute(
+    _Out_writes_bytes_(sizeof(CXPLAT_ROUTE)) CXPLAT_ROUTE* Route
```
guhetier: Is this different from `_Out_`?

src/platform/datapath_xplat.c
```diff
@@ -436,6 +436,10 @@ CxPlatResolveRoute(
         Route->UseQTIP = Socket->ReserveAuxTcpSock;
     }
 
+    #if defined(_KERNEL_MODE) || defined(CX_PLATFORM_LINUX) || defined(CX_PLATFORM_DARWIN)
```
csujedihy, line 439: is qtip really not supported by linux xdp code path?

src/platform/datapath_epoll.c
```diff
@@ -1986,6 +1986,7 @@ CxPlatSocketReceiveCoalesced(
             goto Exit;
         }
 
+        CxPlatInitializeRoute(&IoBlock->Route);
```
guhetier: The `IoBlock` contains more than the route (here, a `refCount`). I am not sure how it is used, but it seems like the full `IoBlock` should be initialized, not only the route.

src/platform/datapath_xplat.c
guhetier: I agree with this. A wrapper could be relevant to initialize the `IoBlock` the route is a part of (although the init logic is pretty platform specific), but seems strange for the route.

I also don't understand the comment. If there is platform specific logic, it should probably move out of the xplat header.

src/platform/datapath_xplat.c
ProjectsByJackHe: not currently.

src/core/connection.c
ProjectsByJackHe: There was a misunderstanding that `QuicConnSetConfiguration` was a publicly documented function, instead of an internal worker API. 

We decided that this change was a no-op, and will revert it.

src/inc/quic_datapath.h
ProjectsByJackHe: got rid of wrapper.

src/platform/datapath_epoll.c
ProjectsByJackHe: got rid of wrapper.

src/platform/datapath_xplat.c
ProjectsByJackHe: got rid of wrapper.

src/core/path.c
ProjectsByJackHe: got rid of wrapper.

src/platform/datapath_winkernel.c
```diff
@@ -2003,7 +2003,9 @@ CxPlatSocketAllocRxIoBlock(
 
     DATAPATH_RX_IO_BLOCK* IoBlock = CxPlatPoolAlloc(Pool);
 
+
```
guhetier: nit: Extra empty line

guhetier:
LGTM modulo nit and the test.yml cleanup.

Still curious / concerned about whether other parts of the IOBlocks might not be initialized and cause issues in the future.

EDIT: "AppData/WithSendArgs1.SendQtip/135" failing might be concerning though

ProjectsByJackHe:
> LGTM modulo nit and the test.yml cleanup.
> 
> Still curious / concerned about whether other parts of the IOBlocks might not be initialized and cause issues in the future.
> 
> EDIT: "AppData/WithSendArgs1.SendQtip/135" failing might be concerning though

This has nothing to do with this PR. As per the logs: 
image

For some reason the advertised interface on the GH runner did not support TX checksum offload. JFYI, the Github backend will allocate a random runner to be used for the CI from a pool of machines. Some will have hardware NICs that advertises this support, others, evidently, won't.

guhetier:
> > LGTM modulo nit and the test.yml cleanup.
> > Still curious / concerned about whether other parts of the IOBlocks might not be initialized and cause issues in the future.
> > EDIT: "AppData/WithSendArgs1.SendQtip/135" failing might be concerning though
> 
> This has nothing to do with this PR. As per the logs: image
> 
> For some reason the advertised interface on the GH runner did not support TX checksum offload. JFYI, the Github backend will allocate a random runner to be used for the CI from a pool of machines. Some will have hardware NICs that advertises this support, others, evidently, won't.

Good to know. If it happens more than once in a blue moon, we should consider adapting the test though.

PR #5291, opened 2025-07-28
ALLOC has random bit flips?

ProjectsByJackHe:
## Description

## Testing

## Documentation

### DIFF
--- a/src/platform/datapath_winkernel.c
+++ b/src/platform/datapath_winkernel.c
@@ -1293,7 +1293,8 @@ SocketCreateUdp(
         Status = QUIC_STATUS_OUT_OF_MEMORY;
         goto Error;
     }
-
+    
+    Binding->ReserveAuxTcpSock = FALSE;
     //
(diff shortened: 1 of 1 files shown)


src/platform/datapath_winkernel.c
```diff
@@ -1293,7 +1293,8 @@ SocketCreateUdp(
         Status = QUIC_STATUS_OUT_OF_MEMORY;
         goto Error;
     }
-
+    
+    Binding->ReserveAuxTcpSock = FALSE;
```
nibanks, line 1297: Binding is zero'ed a few lines below.

ProjectsByJackHe: It does, but we set *NewBinding before we zero it out so I am wondering perhaps if there is a race condition.

nibanks: If you're worried about that then I'd move that line down. But, I don't think you need to worry about it because no other thread will reference that variable until you pose a receive, which happens later, after everything is initialized.

PR #5294, opened 2025-07-29
Set the Target Architecture on OpenSSL/quicTLS arm64 builds

anrossi:
## Description

In non-onebranch scenarios, the target architecture needs to be set in OpenSSL/quicTLS configure step for clang builds cross-compiling for arm64.

## Testing

CI

## Documentation

N/A


guhetier (approved):
LGTM, please let the CI test run to make sure there is no unexpected regression.

PR #5310, opened 2025-07-31
disable all optimizations in linux debug builds

mtfriesen:
## Description

The `-Og` flag reduces optimizations, but debugging is still slower and less productive than necessary due to the remaining compiler optimizations. Turn the optimizations all the way back to `-O0` and add a separate `-g` flag.

## Testing

Seems to improve debugging efficiency locally.

## Documentation

No.


mtfriesen:
Evidently a dup of #5276

PR #5313, opened 2025-08-02
Allow the app to react when app-provided buffer is insufficient

guhetier:
## Description

## Testing

## Documentation

### DIFF
--- a/src/core/stream.c
+++ b/src/core/stream.c
@@ -1012,3 +1012,34 @@ QuicStreamProvideRecvBuffers(
     }
     return Status;
 }
+
+_IRQL_requires_max_(DISPATCH_LEVEL)
+BOOLEAN
+QuicStreamNotifyRecvBufferTooSmall(
+    _In_ QUIC_STREAM* Stream,
+    _In_ uint64_t BufferLengthNeeded,
+    _Out_ QUIC_VAR_INT* ShutdownErrorCode
+    )
+{
+    CXPLAT_DBG_ASSERT(Stream->RecvBuffer.RecvMode == QUIC_RECV_BUF_MODE_APP_OWNED);
+
+    *ShutdownErrorCode = 0;
+
+    QUIC_STREAM_EVENT Event = {0};
+    // TODO guhetier: Consider renaming to "INSUFICIENT_RECEIVE_BUFFER"
+    Event.Type = QUIC_STREAM_EVENT_RECEIVE_BUFFER_NEEDED;
+    Event.RECEIVE_BUFFER_NEEDED.BufferLengthNeeded = BufferLengthNeeded;
+
+    QuicTraceLogStreamVerbose(
+        StreamNotifyRecvBufferTooSmall,
+        Stream,
+        "Indicating QUIC_STREAM_EVENT_RECEIVE_BUFFER_NEEDED [BufferLengthNeeded=%llu]",
+        Event.RECEIVE_BUFFER_NEEDED.BufferLengthNeeded);
+
+    if (QUIC_FAILED(QuicStreamIndicateEvent(Stream, &Event))) {
+        return FALSE;
+    };
+
+    *ShutdownErrorCode = Event.RECEIVE_BUFFER_NEEDED.ErrorCode;
+    return Event.RECEIVE_BUFFER_NEEDED.ShutdownStream;
+}
--- a/src/core/stream.c
+++ b/src/core/stream.c
@@ -1018,32 +1018,23 @@ QuicStreamProvideRecvBuffers(
 }
 
 _IRQL_requires_max_(DISPATCH_LEVEL)
-BOOLEAN
-QuicStreamNotifyRecvBufferTooSmall(
+void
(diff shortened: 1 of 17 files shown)


docs/Streams.md
```diff
@@ -170,6 +170,17 @@ The application regains full ownership of a buffer after it get a receive notifi
 If the application accepts all the buffer's bytes **inline** from the receive notification, by returning `QUIC_STATUS_SUCCESS` and setting `TotalBufferLength` appropriately,
 it can free or reuse the buffer while in the notification handler.
 
+If more data is received on the stream than buffer space was provided by the application,
+MsQuic will emit a `QUIC_STREAM_EVENT_INSUFFICIENT_RECEIVE_BUFFER` notification.
```
nibanks:
```suggestion
If more data is received on the stream than buffer space was provided by the application, MsQuic will emit a `QUIC_STREAM_EVENT_INSUFFICIENT_RECEIVE_BUFFER` notification.
```
nit: for docs, in general we put each sentence on a single line. No manual line wraps.

src/inc/msquic.h
```diff
@@ -1543,6 +1543,9 @@ typedef enum QUIC_STREAM_EVENT_TYPE {
     QUIC_STREAM_EVENT_IDEAL_SEND_BUFFER_SIZE    = 8,
     QUIC_STREAM_EVENT_PEER_ACCEPTED             = 9,
     QUIC_STREAM_EVENT_CANCEL_ON_LOSS            = 10,
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+    QUIC_STREAM_EVENT_INSUFFICIENT_RECEIVE_BUFFER = 11,
```
nibanks: I'm not a big fan of the name. I'm trying to think of something better. Perhaps `QUIC_STREAM_EVENT_RECEIVE_BUFFER_NEEDED`...

src/core/stream.h
```diff
@@ -1056,3 +1056,16 @@ QuicStreamProvideRecvBuffers(
...
+// incoming data.
+// The app can increase the buffer size or shutdown the stream inline
+// as a reaction. Only for app-owned buffering mode.
+//
+_IRQL_requires_max_(DISPATCH_LEVEL)
+void
+QuicStreamNotifyInsufficientRecvBuffer(
+    _In_ QUIC_STREAM* Stream,
+    _In_ uint64_t BufferLengthNeeded
+    );
```
nibanks: nit: missing trailing newline

src/core/stream.c
```diff
@@ -999,16 +999,42 @@ QuicStreamProvideRecvBuffers(
...
+    Event.INSUFFICIENT_RECEIVE_BUFFER.BufferLengthNeeded = BufferLengthNeeded;
+
+    QuicTraceLogStreamVerbose(
+        StreamNotifyInsufficientRecvBuffer,
+        Stream,
+        "Indicating QUIC_STREAM_EVENT_INSUFFICIENT_RECEIVE_BUFFER [BufferLengthNeeded=%llu]",
+        Event.INSUFFICIENT_RECEIVE_BUFFER.BufferLengthNeeded);
+
+    (void)QuicStreamIndicateEvent(Stream, &Event);
+}
```
nibanks: nit missing trailing newline

src/inc/msquic.h
guhetier: I actually started with `QUIC_STREAM_EVENT_RECEIVE_BUFFER_NEEDED`, and changed it later. It felt it could encourage the app developer to see it as the expected way to provide more receive buffer space.
I think it should rather be seen as a warning that they didn't anticipate correctly and are now trying to recover.

I am open to going back to `QUIC_STREAM_EVENT_RECEIVE_BUFFER_NEEDED` or any other better option.

nibanks: I expect if we're really going to support this model of allowing an app to dynamically supply receive buffers, some will want to go all in with this, possibly always waiting for this event before they post a receive buffer. So, considering this a warning-only case is an assumption I don't think we should have.

guhetier: I am sure some will, but I don't want to encourage it, and have the app providing small chunks in the callback all the time (or worst, allocating memory in each callback).
The expected flow is for the app to provide buffer space in advance, and use this callback only to react to the unexpected. Maybe I am too concerned about the app pessimizing itself and the warning in the doc is enough.

docs/Streams.md
guhetier: Fixed. I dislike long lines in md documents (hard to diff and review), but I guess I'll just do small sentences :)

src/inc/msquic.h
guhetier: I reverted to `QUIC_STREAM_EVENT_RECEIVE_BUFFER_NEEDED` - but I am open to other good idea for a name.

ProjectsByJackHe:
Overall LGTM, can you include more details in the description about how the tests have been adapted? @guhetier

ProjectsByJackHe (approved):

guhetier:
@ProjectsByJackHe I updated the description

PR #5327, opened 2025-08-06
make posix header more compatible with c++

anrossi:
## Description

C++ projects with strict validation don't like the C-style casts used for the QUIC_STATUS definitions.

## Testing

Existing tests should cover this

## Documentation

N/A


mtfriesen (dismissed):

src/inc/msquic_posix.h
```diff
@@ -79,9 +79,30 @@ QUIC_INLINE ENUMTYPE &operator ^= (ENUMTYPE &a, ENUMTYPE b) throw() { return (EN
 
 #define QUIC_API
 #define QUIC_MAIN_EXPORT
-#define QUIC_STATUS                     unsigned int
+#ifdef __cplusplus
+typedef int QUIC_STATUS;
```
guhetier: `using` should be preferred

```suggestion
using QUIC_STATUS = int;
```

src/inc/msquic_posix.h
```diff
@@ -79,9 +79,30 @@ QUIC_INLINE ENUMTYPE &operator ^= (ENUMTYPE &a, ENUMTYPE b) throw() { return (EN
...
+static inline
```
guhetier: Consider `constexpr` instead.
Do you have a specific reason for `static` here? It is very rare it makes sense to have a `static inline`, especially in a header

guhetier:
Should similar changes be made to other headers declaring error codes for consistency?

guhetier (approved):

guhetier:
Please open bugs for CI failures you do not resolve as part of the PR.

I suggest we go for the following format so they are easy to find in the future:
"[CI-TEST] <testname>"

We can then:
- reference the issue from the bug to "link" them together (and do that on potential CI failure)
- add the exact testpass name in the bug description

src/inc/msquic_posix.h
sm-msft: This can't be a constexpr function since its output depends on the input. "static" is likely to allow it to access C++ static members or for use in static methods.

anrossi: I think constexpr does allow for functions that depend on the input, because if the compiler can deduce that the input is constant, then it optimizes away the function call entirely.

src/inc/msquic_posix.h
sm-msft: Also, want to make sure the change from "unsigned int" -> "int" was intentional.

src/inc/msquic_posix.h
sm-msft: The nuevo macro it seems :)

edit: 
Reading this page left me with more questions than answers. (Until C++23?)

<html><body>

Its return type (if exists) is a literal type.  Each of its parameter types is a literal type. | (until C++23)

</body>
</html>

[Literal Type](https://en.cppreference.com/w/cpp/language/constant_expression.html#Literal_type)

guhetier: `constexpr` on a function will let the compiler evaluate it at compile time as long as the parameters are known at compile time. In the recent C++ standard, where much more things are constexpr compatible, you can do a whole lot through it.

The page is pretty hard to read because the concept evolved at lot over the last decade, generally speaking, it keeps growing in term of what can be constexpr.

mtfriesen:
> Please open bugs for CI failures you do not resolve as part of the PR.
> 
> I suggest we go for the following format so they are easy to find in the future: "[CI-TEST] "
> 
> We can then:
> 
> * reference the issue from the bug to "link" them together (and do that on potential CI failure)
> * add the exact testpass name in the bug description

The eBPF team has automated something along these lines, FWIW. More could probably be done to get useful bug titles from individual test cases. Example issue: https://github.com/microsoft/ebpf-for-windows/issues/4470

src/inc/msquic_posix.h
mtfriesen: Ok, so TL;DR `constexpr` doesn't mean the parameter must be const and the function evaluated at compile time, it just means the function is compatible with compile time evaluation if the parameters all happen to be known?

src/inc/msquic_posix.h
mtfriesen: Yeah, that was by design, though I didn't understand why it was unsigned in the first place.

mtfriesen:
> Should similar changes be made to other headers declaring error codes for consistency?

I'm fine with doing this lazily, both in the compsci and work-prioritization senses.

src/inc/msquic_posix.h
guhetier: Yes, or to be nitpicky, that the function *must* be compatible with compile time evaluation (so the compiler will tell you if you try to do something not compatible)

sm-msft: This is a commentary on the standard - The intuitive readability has taken a hit with the expansion of "constexpr" definition to include functions that are also const and don't rely on dynamically allocated memory but don't necessarily return a constant value. Also, the revisions to the definition are not helping in this, unfortunately.

src/inc/msquic_posix.h
```diff
@@ -79,7 +79,7 @@ QUIC_INLINE ENUMTYPE &operator ^= (ENUMTYPE &a, ENUMTYPE b) throw() { return (EN
 
 #define QUIC_API
 #define QUIC_MAIN_EXPORT
-#define QUIC_STATUS                     unsigned int
+#define QUIC_STATUS                     int
```
guhetier: Thinking more about the `unsigned int` -> `int` change, why do we need it?

I am starting to have concerns about compile time breaking changes.

anrossi: I think it's because `QUIC_STATUS_PENDING` is -2 and `QUIC_STATUS_CONTINUE` is -1.  On windows, HRESULT is defined as `long` which is a signed type as well, so this would bring them both in line, but you're right, it could break existing projects

guhetier: IMHO, with a cast, that is not justification enough.
In the absolute, `QUIC_STATUS` being signed here from the start could have made sense. But it has not been since the start and changing might have consequences.

My question is more about "what is this change trying to fix" / can't the fix be on the other side. It may take some extra ugly casts, but we are weighting against a potentially breaking api change.

anrossi: That's a fair point. I've reverted the type changes and simply have the casting

guhetier (approved):

PR #5328, opened 2025-08-06
Remove experimental RIO data path

mtfriesen:
## Description

The RIO data path added maintenance costs without any use cases. Performance tests showed limited upside, as well, and the scenarios where RIO should be better than IOCP are scenarios where XDP, DPDK, RDMA, etc. will be even better.

## Testing

I couldn't find any tests onboarded.

## Documentation

Removed.


docs/api/QUIC_SETTINGS.md
```diff
@@ -57,7 +57,7 @@ typedef struct QUIC_SETTINGS {
             uint64_t StreamMultiReceiveEnabled              : 1;
             uint64_t XdpEnabled                             : 1;
             uint64_t QTIPEnabled                            : 1;
-            uint64_t RioEnabled                             : 1;
+            uint64_t ReservedRioEnabled                     : 1;
```
nibanks, line 60: This is in the preview section so you should just remove it.

PR #5333, opened 2025-08-08
Improve test and add documentation to the connection pool

### LINKED ISSUE #5300: Connection pools are sending spurious connection notifications to the app
### Describe the bug

By design, connection pools create multiple connections and close them immediately if they are created on an RSS Core for which a connection was already created.

However, these transient connections are still sending notifications to the application on creation / connection / stop. This is problematic because it is very complex for an application to know which callbacks should be ignored and which callbacks correspond to an actual connection.

The connection pool should prevent notifications from the transient connection from reaching the app. 

A potential fix would be to provide a temporary callback handler when creating the connection, and to switch to the app callback only once the connection is confirmed to be on a new RSS core.
The switch should likely be done from inside of a callback to limit concurrency issues.

### Affected OS

- [ ] Windows
- [ ] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

_No response_

### MsQuic version

main

Create a connection pool

### Expected behavior

No spurious notifications

### Actual outcome

Spurious notifications are received

### Additional details

_No response_

guhetier:
## Description

Since connection pool can start and discard connections as part of its attempt to place them a different RSS cores, there were concerns the discarded connection would send spurious notifications.
Review of the code and designed confirm this isn't the case, and test coverage was improved.

- Add some comments to the connection_pool explaining the overall design
- Validate that no spurious notifications are received in tests

Closes #5300

## Testing

CI

## Documentation

Improved

### DIFF
--- a/src/test/lib/HandshakeTest.cpp
+++ b/src/test/lib/HandshakeTest.cpp
@@ -4170,14 +4170,18 @@ QuicTestConnectionPoolCreate(
         QuicAddrSetToDuoNic(&ServerAddr.SockAddr);
     }
 
-    UniquePtrArray<ConnectionScope> Connections(new(std::nothrow) ConnectionScope[NumberOfConnections]);
-    TEST_NOT_EQUAL(nullptr, Connections);
-
+    //
+    // Make sure to create the connection contexts before the connections,
+    // to ensure they are not freed before the connection is closed.
+    //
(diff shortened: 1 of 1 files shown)


guhetier:
Netperf failing because of the iouring PR.

PR #5347, opened 2025-08-11
CP: Allow the app to react when app-provided buffer is insufficient

guhetier:
## Description

Cherry-pick of #5313

When using app-provided buffers, the flow control window was constrained to the amount of buffer provided by the application.
Assuming the application mis-calculates the amount of buffer needed, this could cause a stream to be soft-locked: the receive window would be zero and the sender wait forever.

This changes this behavior:
- the flow control window is no longer bound by the amount of buffer provided
- if more data is received than there is buffer space, a `QUIC_STREAM_EVENT_INSUFFICIENT_RECEIVE_BUFFER` is sent to the app which can:
    - provide more buffer space inline
    - close abortively the receive side of the stream inline
  Otherwise, the connection is aborted.

## Testing

Test have been improved and adapted for the new behavior.

Instead of validating that when the sender queue more data than the receiver has buffer for, the receive window limits the amount of data sent (previous behavior), the updated tests validate that the receive buffer is no longer a limiting factor and that when more data is received than there is buffer available for:
- the notification is emmited
- the app can provide more buffer space
- the app can shutdown the stream inline

## Documentation

Documentation was updated

### DIFF
--- a/docs/PreviewFeatures.md
+++ b/docs/PreviewFeatures.md
@@ -1,4 +1,4 @@
-Preview features
+GPreview features
 =========
 > [!IMPORTANT]
 >
@@ -35,8 +35,7 @@ TODO
 
 TODO
 
-### StreamProvideReceiveBuffers
+### App-provided receive buffers
 
-[StreamProvideReceiveBuffers](StreamProvideReceiveBuffers.md)
-
-TODO
+- [StreamProvideReceiveBuffers](api/StreamProvideReceiveBuffers.md)
+- [QUIC_API_ENABLE_PREVIEW_FEATURES](api/QUIC_STREAM_EVENT.md#quic_stream_event_receive_buffer_needed)
--- a/docs/PreviewFeatures.md
+++ b/docs/PreviewFeatures.md
@@ -1,4 +1,4 @@
-GPreview features
+Preview features
 =========
 > [!IMPORTANT]
 >
--- a/docs/Streams.md
+++ b/docs/Streams.md
@@ -170,9 +170,16 @@ The application regains full ownership of a buffer after it get a receive notifi
 If the application accepts all the buffer's bytes **inline** from the receive notification, by returning `QUIC_STATUS_SUCCESS` and setting `TotalBufferLength` appropriately,
 it can free or reuse the buffer while in the notification handler.
 
+If more data is received on the stream than buffer space was provided by the application, MsQuic will emit a `QUIC_STREAM_EVENT_RECEIVE_BUFFER_NEEDED` notification.
+When receiving this notification, the app can:
+- provide a sufficient amount of buffer space **inline** from the callback using [`StreamProvideReceiveBuffers`](./api/StreamProvideReceiveBuffers.md)
+- shutdown the stream receive direction of the stream **inline** by calling [`StreamShutdown`](api/StreamShutdown.md) with the `QUIC_STREAM_SHUTDOWN_FLAG_INLINE` flag
+
+When the callback returns, if the stream has not been shutdown and a sufficient amount of memory is not available, the connection is closed abortively.
+Providing memory in reaction to `QUIC_STREAM_EVENT_RECEIVE_BUFFER_NEEDED` can impact performances negatively.
+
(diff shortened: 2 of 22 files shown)


csujedihy:
> this could cause a stream to be soft-locked: the receive window would be zero and the sender wait forever.

Why would the sender wait forever? Would the sender be unblocked when the app provides buffers to msquic on the receiver side?

It seems like with app-owned buffer, we will never be doing flow control? (I have not caught up all these receive changes recently, probably never will :) )

csujedihy (approved):

guhetier:
> > this could cause a stream to be soft-locked: the receive window would be zero and the sender wait forever.
> 
> Why would the sender wait forever? Would the sender be unblocked when the app provides buffers to msquic on the receiver side?
> 
> It seems like with app-owned buffer, we will never be doing flow control? (I have not caught up all these receive changes recently, probably neve will :) )

In the first version of app-provided receive buffer (before this PR), the receive window for the stream was limited to the amount of buffer provided by the app. The problem is if the app has a bug and miscount how much buffer it needs to provide. You could end up with a 0-size received window. The sender is waiting because the receive window size is zero, the receiver is waiting because it think more data should arrive, and the stream is stalled.

After this PR, when running out of buffer space, the app receive a notification. So if the receiver is buggy and mis-counted how much buffer it needs to provide, it now has a clear signal (to either give more buffer, or simply close the stream because something went wrong).

csujedihy:
> > > this could cause a stream to be soft-locked: the receive window would be zero and the sender wait forever.
> > 
> > 
> > Why would the sender wait forever? Would the sender be unblocked when the app provides buffers to msquic on the receiver side?
> > It seems like with app-owned buffer, we will never be doing flow control? (I have not caught up all these receive changes recently, probably neve will :) )
> 
> In the first version of app-provided receive buffer (before this PR), the receive window for the stream was limited to the amount of buffer provided by the app. The problem is if the app has a bug and miscount how much buffer it needs to provide. You could end up with a 0-size received window. The sender is waiting because the receive window size is zero, the receiver is waiting because it think more data should arrive, and the stream is stalled.
> 
> After this PR, when running out of buffer space, the app receive a notification. So if the receiver is buggy and mis-counted how much buffer it needs to provide, it now has a clear signal (to either give more buffer, or simply close the stream because something went wrong).

Ok, the notification is useful. Though, it seems less ideal if we remove the capability for the receiver to put back pressure on the sender if the receiver simply cannot keep up. In that case, we'd either kill the connection or provide a buffer, but if the system cannot keep up, providing a buffer upon getting such notification is going to exacerbate the load/queuing/memory usage?

guhetier:
> > > > this could cause a stream to be soft-locked: the receive window would be zero and the sender wait forever.
> > > 
> > > 
> > > Why would the sender wait forever? Would the sender be unblocked when the app provides buffers to msquic on the receiver side?
> > > It seems like with app-owned buffer, we will never be doing flow control? (I have not caught up all these receive changes recently, probably neve will :) )
> > 
> > 
> > In the first version of app-provided receive buffer (before this PR), the receive window for the stream was limited to the amount of buffer provided by the app. The problem is if the app has a bug and miscount how much buffer it needs to provide. You could end up with a 0-size received window. The sender is waiting because the receive window size is zero, the receiver is waiting because it think more data should arrive, and the stream is stalled.
> > After this PR, when running out of buffer space, the app receive a notification. So if the receiver is buggy and mis-counted how much buffer it needs to provide, it now has a clear signal (to either give more buffer, or simply close the stream because something went wrong).
> 
> Ok, the notification is useful. Though, it seems less ideal if we remove the capability for the receiver to put back pressure on the sender if the receiver simply cannot keep up. In that case, we'd either kill the connection or provide a buffer, but if the system cannot keep up, providing a buffer upon getting such notification is going to exacerbate the load/queuing/memory usage?

The "usual" receive window growth heuristic still apply, some backpressure is applied at that layer and by the speed at which the application manage to provide the memory needed.
But overall, the feature is not meant to let an application provide memory buffers for an unknown stream of data - rather to let the app provide buffers for a specific payload of roughly known size to save a copy between MsQuic buffers and its final destination.

The feature is still in preview, so happy to talk about how to improve the design if you have ideas (the backport PR might not be the best spot :) )

Issue #5344, opened 2025-08-11
Netperf pipeline broken due to #5335


ProjectsByJackHe:
### Describe the bug

It seems like PR (https://github.com/microsoft/msquic/pull/5335) causes Netperf pipeline to crash:

<img width="1463" height="1330" alt="Image" src="https://github.com/user-attachments/assets/16b81797-26dd-4522-bfe9-6cc702e5bd50" />

We also would like to figure out steps to take to prevent merging PRs that break Netperf CI.

### Affected OS

- [x] Windows
- [x] Linux
- [ ] macOS
- [ ] Other (specify below)

### MsQuic version

main

### Steps taken to reproduce bug

Trigger Netperf CI from the latest main

### Expected behavior

Netperf passes

### Actual outcome

Lab runs fail due to this bug

mtfriesen:
Welp, my bad, I didn't expect my changes to affect Windows and assumed the lab failures were just flakiness.

(Separately, but unhelpfully, it would be ideal if the same code that runs clean on Azure machines implied it would run clean on lab machines.)

ProjectsByJackHe:
Fixed with #5353

PR #5348, opened 2025-08-12
Fix Netperf CI

ProjectsByJackHe:
## Description

Fixes: #5344

Addresses the CI failures due to recent changes in the secnetperf scripts.

## Testing

Netperf CI: https://github.com/microsoft/netperf/actions/runs/16894832432/job/47862515766

## Documentation

N/A

### DIFF
--- a/scripts/secnetperf.ps1
+++ b/scripts/secnetperf.ps1
@@ -277,10 +277,7 @@ $allScenarios = @("upload", "download", "hps", "rps", "rps-multi", "latency")
 $hasFailures = $false
 
 try {
-    # TODO: Running the prepare-machine script on the linux lab machine is running into quite a few snags currently.
-    # PROBLEM: We don't copy over the entire repo via remote powershell (should we?) so "git submodule init..." commands fail.
-    #          Windows lab does not have this problem because prepare-machine does not run any of that on Windows.
-    if (!(!$IsWindows -and $environment -eq "lab")) {
+    if ($IsWindows -or $environment -eq "azure") {
         Prepare-MachineForTest $Session $RemoteDir
     }
 
--- a/scripts/secnetperf.ps1
+++ b/scripts/secnetperf.ps1
@@ -280,7 +280,7 @@ try {
     # TODO: Running the prepare-machine script on the linux lab machine is running into quite a few snags currently.
     # PROBLEM: We don't copy over the entire repo via remote powershell (should we?) so "git submodule init..." commands fail.
     #          Windows lab does not have this problem because prepare-machine does not run any of that on Windows.
-    if (!(!$IsWindows -and -$environment -eq "lab")) {
+    if (!(!$IsWindows -and $environment -eq "lab")) {
         Prepare-MachineForTest $Session $RemoteDir
     }
 
--- a/scripts/secnetperf.ps1
+++ b/scripts/secnetperf.ps1
@@ -277,12 +277,22 @@ $allScenarios = @("upload", "download", "hps", "rps", "rps-multi", "latency")
 $hasFailures = $false
 
 try {
-Prepare-MachineForTest $Session $RemoteDir
+    # TODO: Running the prepare-machine script on the linux lab machine is running into quite a few snags currently.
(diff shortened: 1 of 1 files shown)


scripts/secnetperf.ps1
```diff
@@ -277,14 +277,24 @@ $allScenarios = @("upload", "download", "hps", "rps", "rps-multi", "latency")
 $hasFailures = $false
 
 try {
-Prepare-MachineForTest $Session $RemoteDir
+    # TODO: Running the prepare-machine script on the linux lab machine is running into quite a few snags currently.
+    # PROBLEM: We don't copy over the entire repo via remote powershell (should we?) so "git submodule init..." commands fail.
+    #          Windows lab does not have this problem because prepare-machine does not run any of that on Windows.
+    if (!(!$IsWindows -and $environment -eq "lab")) {
```
guhetier: Please use the De Morgan's law to simplify this.
`if ($IsWindows -or $environment -neq "lab")` is much simpler to read.

scripts/secnetperf.ps1
```diff
@@ -277,14 +277,24 @@ $allScenarios = @("upload", "download", "hps", "rps", "rps-multi", "latency")
 $hasFailures = $false
 
 try {
-Prepare-MachineForTest $Session $RemoteDir
+    # TODO: Running the prepare-machine script on the linux lab machine is running into quite a few snags currently.
```
guhetier: Please address the TODO or file a bug to track it.

Lab don't have internet access? I doubt the test runs needs the submodules (only the artifacts and the script folder, IIRC), do you know why `git submodule init` is even running as part of the "for test" setup?

ProjectsByJackHe: Created the issue with: https://github.com/microsoft/msquic/issues/5349

The labs do have internet, but as the TODO says, we don't copy over the entire git history. As for submodule dependency, this seems to be a Linux Logging related dependency. I am sure there is a way to remove it. I really don't want to have to copy over the .git history from client to remote as part of the automation.

scripts/secnetperf.ps1
ProjectsByJackHe: Will do.

scripts/secnetperf.ps1
```diff
@@ -277,14 +277,21 @@ $allScenarios = @("upload", "download", "hps", "rps", "rps-multi", "latency")
 $hasFailures = $false
 
 try {
-Prepare-MachineForTest $Session $RemoteDir
+    if ($IsWindows -or $environment -eq "azure") {
```
mtfriesen, line 280: I think the point is that MsQuic should not have this logic.

scripts/secnetperf.ps1
```diff
@@ -277,14 +277,21 @@ $allScenarios = @("upload", "download", "hps", "rps", "rps-multi", "latency")
...
+        Prepare-MachineForTest $Session $RemoteDir
+    }
 
 if ($isWindows -and !($environment -eq "azure")) {
     $HasTestSigning = $false
     try { $HasTestSigning = ("$(bcdedit)" | Select-String -Pattern "testsigning\s+Yes").Matches.Success } catch { }
     if (!$HasTestSigning) { Write-Host "Test Signing Not Enabled!" }
 }
 
+if (!($Session -eq "NOT_SUPPORTED")) {
```
mtfriesen, line 290: MsQuic should not have this logic.

mtfriesen (changes requested):
MsQuic should not have logic like `$environment -eq "azure"` nor `$Session -eq "NOT_SUPPORTED"` sprinkled throughout.

scripts/secnetperf.ps1
guhetier: Now that I see the detail of the issue, I agree with Michael's comments - if `prepare-machinefortest` requires the git repo to run, we should clone the git repo. Or fix the script so that it doesn't need the git repo. But the script is meant to prepare the machine before a test run, so we should run it before a test run.

ProjectsByJackHe:
Closing in lieu of #5353

PR #5351, opened 2025-08-13
Split Sanitizer Configuration Options

anrossi:
## Description

Split the configuration of sanitizers to allow turning them on individually.  Also enable configuration of sanitizers on OpenSSL separate from the rest of MsQuic.

## Testing

The CI should cover that this doesn't regress the build.

## Documentation

N/A


CMakeLists.txt
```diff
@@ -86,7 +86,11 @@ option(QUIC_BUILD_TEST "Builds the test code" OFF)
 option(QUIC_BUILD_PERF "Builds the perf code" OFF)
 option(QUIC_BUILD_SHARED "Builds msquic as a dynamic library" ON)
 option(QUIC_ENABLE_LOGGING "Enables logging" OFF)
-option(QUIC_ENABLE_SANITIZERS "Enables sanitizers" OFF)
+option(QUIC_ENABLE_SANITIZERS "Enables all sanitizers" OFF)
+option(QUIC_ENABLE_ASAN "Enables only Address Sanitizer" OFF)
+option(QUIC_ENABLE_LSAN "Enables only Leak Sanitizer" OFF)
+option(QUIC_ENABLE_UBSAN "Enables only Undefined Behavior Sanitizer" OFF)
+option(QUIC_SANITIZE_OPENSSL "Pass sanitizer settings to OpenSSL build" ON)
```
guhetier: Based on our earlier discussion, I think this setting is overly specific.

- what about QUIC_TLS? Or other submodules?
- aren't the option above already allowing you to configure sanitizer settings in a way that works?

IMO, `QUIC_SANITIZE_SUBMODULES` would be better.

guhetier: It is also confusing that setting `QUIC_SANITIZE_OPENSSL` will actually do nothing unless `QUIC_ENABLE_SANITIZERS` is set.

CMakeLists.txt
```diff
@@ -115,6 +119,16 @@ if(NOT "${QUIC_LOGGING_TYPE}" MATCHES "^(etw|lttng|stdout|)$")
     message(FATAL_ERROR "QUIC_LOGGING_TYPE must be \"etw\", \"lttng\" or \"stdout\", not \"${QUIC_LOGGING_TYPE}\"")
 endif()
 
+if (QUIC_ENABLE_SANITIZERS OR QUIC_ENABLE_ASAN OR QUIC_ENABLE_LSAN OR QUIC_ENABLE_UBSAN)
```
guhetier: I would simplify the rest of the code by setting QUIC_ENABLE_ASAN, QUIC_ENABLE_LSAN and QUIC_ENABLE_UBSAN to TRUE here if QUIC_ENABLE_SANITIZERS is set, and then not have to check the different combinations.

Although there might be "alignment" missing? To check

guhetier: Looking more, I think I'd add an extra variable for "all other sanitizers".
Then, after you set all your variables here to the right value based on the inputs, the logic can be a simple "if (X) enable Y", without `else` or `OR`.

CMakeLists.txt
anrossi: I've renamed it to QUIC_SKIP_SANITIZE_SUBMODULES

CMakeLists.txt
anrossi: Sounds good to me. Done

CMakeLists.txt
```diff
@@ -115,6 +120,16 @@ if(NOT "${QUIC_LOGGING_TYPE}" MATCHES "^(etw|lttng|stdout|)$")
     message(FATAL_ERROR "QUIC_LOGGING_TYPE must be \"etw\", \"lttng\" or \"stdout\", not \"${QUIC_LOGGING_TYPE}\"")
 endif()
 
+if (QUIC_ENABLE_SANITIZERS)
+    set(QUIC_SANITIZER_ACTIVE ON)
```
mtfriesen, line 124: Is there a way to set these only if the caller hasn't explicitly set them to off? It might be useful for a caller to be able to set `QUIC_ENABLE_SANITIZERS = ON` and then override one noisy sanitizer with `QUIC_ENABLE_UBSAN = OFF` for example.

CMakeLists.txt
```diff
@@ -115,6 +120,16 @@ if(NOT "${QUIC_LOGGING_TYPE}" MATCHES "^(etw|lttng|stdout|)$")
...
+    set(QUIC_ENABLE_ASAN ON)
+    set(QUIC_ENABLE_LSAN ON)
+    set(QUIC_ENABLE_UBSAN ON)
+    set(QUIC_ENABLE_EXTRA_SANITIZERS ON)
+else()
+    set(QUIC_SANITIZER_ACTIVE OFF)
```
mtfriesen, line 138: How is this intended to work, exactly? Is the caller supposed to be able to turn on, say, just `QUIC_ENABLE_ASAN`, which automatically turns on `QUIC_SANITIZER_ACTIVE`?

CMakeLists.txt
anrossi: I think the way to achieve that is to have separate variables, one as the configuration parameter, and then one used internally. That would be a pattern that MsQuic hasn't used yet (configuration parameters as the variable used), but could  be used here to solve the problem. The downside is added complexity

CMakeLists.txt
anrossi: Ah yes, Good catch!

CMakeLists.txt
guhetier: For the sake of simplicity, I don't think we need that capability.
Or we very quickly reach the point where we could just forward a string with the exact sanitizer options :)

Maybe renaming `QUIC_ENABLE_SANITIZERS` to `QUIC_ENABLE_ALL_SANITIZERS` makes it clearer that you can't disable only some checks?

anrossi: I was trying to avoid renaming QUIC_ENABLE_SANITIZERS since it's external and third parties may have taken a dependency on it.

anrossi: Ok, let's go with simplicity. I'll rename `QUIC_ENABLE_SANITIZERS` to `QUIC_ENABLE_ALL_SANITIZERS` and I won't allow turning off specific ones. Users will enable all of them, or enable only the specific ones they want.

mtfriesen: > I was trying to avoid renaming QUIC_ENABLE_SANITIZERS since it's external and third parties may have taken a dependency on it.

Agreed we should not create wanton disruption for consumers, but increasing code clarity in the long run (and future consumers) is justified.

mtfriesen (dismissed):

guhetier:
Please file a bug / reference the existing bug for failing test passes. It looks like we have a new kernel test frequent failure.

guhetier (approved):

Issue #5334, opened 2025-08-08
Netperf session differences leak across codebase


mtfriesen:
### Describe the bug

Netperf has different remote-control code paths depending on how machines were created, and MsQuic code has to have duplicate (redundant) code to control the remote peer in both netperf modes.

We need to have a single unified layer we build our business logic on top of.

### Affected OS

- [x] Windows
- [x] Linux
- [x] macOS
- [ ] Other (specify below)

### MsQuic version

main

### Steps taken to reproduce bug

Added a prepare-machine.ps1 step to each netperf deployment.

### Expected behavior

Preparing a machine should be as simple as `ExecuteOnPeer { prepare-machine.ps1 }`, modulo a little bit of boilerplate.

### Actual outcome

Multiple files had to be edited, with `if`, `else`, and `return`, and non-obvious requirements.

ProjectsByJackHe:
#5353 Presents a solution to rectify this problem, but that PR only includes changes for a subset of the Netperf script.

We can have follow up PRs to refactor the rest of the code to remove these session differences. We can keep this issue open until that's done.

PR #5353, opened 2025-08-14
Simplify Netperf API

ProjectsByJackHe:
## Description

Attempt to address issues #5348 #5334 

The idea here is that for projects like MsQuic to leverage both lab-based and Azure-vm based perf testing, we have a standardized interface such that no internal implementation details leak.

Here is the proposed design:

For projects that only need to use the lab, they can still leverage remote powershell.
Every project that wants to leverage both Azure + Lab in their workflows needs to run:

- InitNetperfLib "project_callback.ps1"
- CopyRepoToPeer

Each time the client needs to orchestrate some remote commands to run on the peer, you always use:
- NetperfSendCommand -Command X
- NetperfWaitServerFinishExecution

And you add details to "project_callback.ps1" where:

if (command == X) {
   ...do XYZ...
}

### NOTE: 

Needs dependent Netperf PR: https://github.com/microsoft/netperf/pull/710 to merge first.

This pull request is a step in the direction to address the debt, but does not address all debt. Future PRs will be made to replace all instances of the secnetperf scripts to remove any leakage of implementation details.

## Testing

Netperf CI: https://github.com/microsoft/netperf/actions/runs/16973623322

## Documentation

Once more feedback comes, we can add this to the offical docs on netperf.

### DIFF
--- a/scripts/quic_callback.ps1
+++ b/scripts/quic_callback.ps1
@@ -124,6 +124,7 @@ if ($Command.Contains("/home/secnetperf/_work/quic/artifacts/bin/linux/x64_Relea
     .\scripts\log.ps1 -Stop -OutputPath "artifacts/logs/$artifactName/server" -RawLogOnly
 } elseif ($Command.Contains("Config_DumpCollection_Windows")) {
     $DumpDir = "C:/_work/quic/artifacts/crashdumps"
+    $WerDumpRegPath = "HKLM:\Software\Microsoft\Windows\Windows Error Reporting\LocalDumps\secnetperf.exe"
     New-Item -Path $DumpDir -ItemType Directory -ErrorAction Ignore | Out-Null
     New-Item -Path $WerDumpRegPath -Force -ErrorAction Ignore | Out-Null
     Set-ItemProperty -Path $WerDumpRegPath -Name DumpFolder -Value $DumpDir | Out-Null
--- a/scripts/quic_callback.ps1
+++ b/scripts/quic_callback.ps1
@@ -1,5 +1,9 @@
-param (
-    [string]$Command
+param(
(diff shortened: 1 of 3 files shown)


mtfriesen (dismissed):

scripts/quic_callback.ps1
```diff
@@ -81,8 +85,8 @@ if ($Command.Contains("/home/secnetperf/_work/quic/artifacts/bin/linux/x64_Relea
     Write-Host "Executing command: $(pwd)/artifacts/bin/windows/x64_Release_schannel/secnetperf -exec:$mode -io:$io -stats:$stats"
     ./artifacts/bin/windows/x64_Release_schannel/secnetperf -exec:$mode -io:$io -stats:$stats
 } elseif ($Command.Contains("Prepare_MachineForTest")) {
-    Write-Host "Executing command: Prepare_MachineForTest"
-    .\scripts\prepare-machine.ps1 -ForTest -InstallSigningCertificates
+    Write-Host "Executing command: $WorkingDir\scripts\prepare-machine.ps1 -ForTest -InstallSigningCertificates"
+    iex "$WorkingDir\scripts\prepare-machine.ps1 -ForTest -InstallSigningCertificates"
```
guhetier: nit: iex -> InvokeExpression.

In scripts, it is often better to use the long name for clarity.

guhetier:
I looks like we have a new kernel BVT instability. Can you file a bug?

ProjectsByJackHe:
> I looks like we have a new kernel BVT instability. Can you file a bug?

https://github.com/microsoft/msquic/issues/5355

scripts/quic_callback.ps1
ProjectsByJackHe: Done.

PR #5357, opened 2025-08-14
CP: Modified cmake to disable xdp for arm architecture 

guhetier:
## Description

The undock pipeline uses the same cmake flags for all the archs of a target container which means it will throw an error when QUIC_LINUX_XDP_ENABLED is on for arm architectures as its not supported. This PR adds a check to disable the flag for arm archs.

## Testing

CI

## Documentation

N/A

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -293,6 +293,12 @@ else()
             set(QUIC_LOGGING_TYPE "lttng")
             message(STATUS "Choosing lttng as default logging type for platform")
         endif()
+        
+        if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(arm|aarch64)$")
+            set(QUIC_LINUX_XDP_ENABLED OFF CACHE BOOL "XDP not supported on ARM architectures" FORCE)
+        else()
+            option(QUIC_LINUX_XDP_ENABLED "Enables XDP support" OFF)
+        endif()
 
     elseif(CX_PLATFORM STREQUAL "darwin")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -293,6 +293,12 @@ else()
             set(QUIC_LOGGING_TYPE "lttng")
             message(STATUS "Choosing lttng as default logging type for platform")
         endif()
+        
+        if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(arm|aarch64)$")
+            set(QUIC_LINUX_XDP_ENABLED OFF CACHE BOOL "XDP not supported on ARM architectures" FORCE)
+        else()
+            option(QUIC_LINUX_XDP_ENABLED "Enables XDP support" OFF)
```
anrossi, line 300: I know this is a cherry-pick, but this works?  It looks like it just turns off Linux XDP always

guhetier: As discussed offline: it seems like when setting the option, the user provided value is used. I am not sure why the `option` needs to be duplicated from the top of the file, but I'll keep what works in main for here.

guhetier: @gaurav2699 If you can chime in about this it would be helpful, so we make sure the state in main is clean.

gaurav2699: @anrossi As Guillaume mentioned, the user provided value will be used and it wont stay on. @guhetier I agree the else part is useless.

PR #5367, opened 2025-08-19
Windows: Allow customizing CompletionKey for IOCP usage

### LINKED ISSUE #5366: [Custom Execution] Allow distinguishing msquic VS non-msquic IO completions
Hello,

MSQuic recently added bring-your-own threading, which is great for complex applications that want complete control over all of this.

My discussion here is about Windows, I haven't really brainstormed about kpoll/epoll equivalents, but since _custom execution_ is inherently platform-specific, maybe this suggestion could apply to Windows only?

One thing that I think is missing is the ability to mix IO of any shape, with that of MSQuic.
The associated doc says this:

> In a real application, these completion events may come both from MsQuic and the application itself, therefore, this means **the application must use the same base format for its own submission entries**.
This is necessary to be able to share the same event queue object.

For my specific program, it will not be possible for me to change the layout of my own IO's to match that of QUIC_SQE. I believe that this is overly restrictive and that it would be possible to relax this at next to no cost or complexity.

Again from the docs, looking at the sample:
```c++
if (GetQueuedCompletionStatusEx(IOCP, Overlapped, ARRAYSIZE(Overlapped), &OverlappedCount, WaitTime, FALSE)) {
        for (ULONG i = 0; i < OverlappedCount; ++i) {
            QUIC_SQE* Sqe = CONTAINING_RECORD(Overlapped[i].lpOverlapped, QUIC_SQE, Overlapped);
            Sqe->Completion(&Overlapped[i]);
        }
    }
```

Right off the bat, thanks to the fact that the API shape chosen by MSQuic doesn't leverage the `completionKey` (In GQCSE this would be `OVERLAPPED_ENTRY::lpCompletionKey`)  at all, we can see it is totally possible to distinguish msquic-based IO from other arbitrary IO by the completionKey. MSQuic currently hardcodes that to 0 [here](https://github.com/microsoft/msquic/blob/28649f704c72d1e1e286cfeeca24afca9a390c78/src/inc/quic_platform_winuser.h#L751). So even if we changed nothing at all, it would be possible for applications to tell IOs apart by ensuring their own IOs are non-0.

However, basing it on zero/non-zero is a bit brittle and limited.

I would like to propose that applications could tell MSQuic what value to use as completionKey, as a configuration value. 
MSQuic would pass this in `CxPlatEventQAssociateHandle` to `CreateIoCompletionPort`.
It would also pass it to `CxPlatEventQEnqueue(Ex)`

I would be happy to contribute the PR if this idea is approved.

Thanks!

cc @nibanks since you committed the original work for custom execution
ref #4616, #1923

fbrosseau:
## Description

Applications can mix their own arbitrary IOs with those of MsQuic on the same "event queues" (Windows IOCP, Linux epoll, etc).

The simplest way to accomplish this is to ensure that their non-MsQuic follow the same layout and convention as that of `CXPLAT_SQE`, which has some platform-specific state (`OVERLAPPED` in Windows, for example), and a callback called `CXPLAT_SQE::Completion` that should be invoked by the application.

However, it is not always possible for complex applications to follow MsQuic's convention for their own arbitrary IOs.

On Windows specifically, it is possible to let applications distinguish the origin of IO completion packets with the use of the `CompletionKey` that gets configured by `CreateIoCompletionPort` and passed back by `GetQueuedCompletionStatus/Ex`.

This PR changes the definition of `CXPLAT_EVENTQ` for Windows so that applications may tell MsQuic what value to use as `CompletionKey`. The value they provide is application-defined and is not interpreted by MsQuic.

## Testing

I adapted the Windows Execution sample.

## Documentation

Yes updated.

Fixes #5366

### DIFF
--- a/docs/Execution.md
+++ b/docs/Execution.md
@@ -131,7 +131,10 @@ To create an execution context, the app much first create an event queue object,
 On Windows, the following types are defined:
 
 ```c++
-typedef HANDLE QUIC_EVENTQ;
+typedef struct CXPLAT_EVENTQ {
+    HANDLE IoCompletionPort;
+    ULONG_PTR CompletionKey;
+} CXPLAT_EVENTQ;
 
 typedef OVERLAPPED_ENTRY CXPLAT_CQE;
 
@@ -149,22 +152,26 @@ typedef struct CXPLAT_SQE {
 } CXPLAT_SQE;
 ```
 
-You will also notice the definiton for `QUIC_SQE` (SQE stands for submission queue entry), which defines the format that all completion events must take so they may be generically processed from the event queue (more on this below).
+You will also notice the definiton for `CXPLAT_SQE` (SQE stands for submission queue entry), which defines the format that all completion events must take so they may be generically processed from the event queue (more on this below).
 
 Once the app has the event queue, it may create the execution context with the `ExecutionCreate` function:
 
 ```c++
 HANDLE IOCP = CreateIoCompletionPort(INVALID_HANDLE_VALUE, nullptr, 0, 1);
-QUIC_EXECUTION_CONFIG ExecConfig = { 0, &IOCP };
+ULONG_PTR completionKey = 0; // Optional
+CXPLAT_EVENTQ queueConfig = { IOCP, completionKey };
+QUIC_EXECUTION_CONFIG ExecConfig = { 0, &queueConfig };
 
 QUIC_EXECUTION* ExecContext = nullptr;
 QUIC_STATUS Status = MsQuic->ExecutionCreate(QUIC_GLOBAL_EXECUTION_CONFIG_FLAG_NONE, 0, 1, &ExecConfig, &ExecContext);
 ```
 
-The above code createa a new IOCP (for Windows), sets up an execution config, indicating an ideal processor of 0 and the pointer to the IOCP, and then calls MsQuic to create 1 execution context.
+The above code creates a new IOCP (for Windows), sets up an execution config, indicating an ideal processor of 0 and the pointer to the IOCP, and then calls MsQuic to create 1 execution context.
 An application may expand this code to create multiple execution contexts, depending on their needs.
 
 To drive this execution context, the app will need to to periodically call `ExecutionPoll` and use the platform specific function to drain completion events from the event queue.
+The return value of `ExecutionPoll` tells the application how long it should wait (in milliseconds) before calling `ExecutionPoll` again, even if no new IO has completed. This is used by MsQuic to implement its internal timeouts. 
+If the application operates a simple loop around `GetQueuedCompletionStatus(Ex)` as in the following example, then it may conveniently pass that directly as its timeout value.
 
 ```c
 bool AllDone = false;
@@ -175,7 +182,7 @@ while (!AllDone) {
     OVERLAPPED_ENTRY Overlapped[8];
     if (GetQueuedCompletionStatusEx(IOCP, Overlapped, ARRAYSIZE(Overlapped), &OverlappedCount, WaitTime, FALSE)) {
         for (ULONG i = 0; i < OverlappedCount; ++i) {
-            QUIC_SQE* Sqe = CONTAINING_RECORD(Overlapped[i].lpOverlapped, QUIC_SQE, Overlapped);
+            CXPLAT_SQE* Sqe = CONTAINING_RECORD(Overlapped[i].lpOverlapped, CXPLAT_SQE, Overlapped);
             Sqe->Completion(&Overlapped[i]);
         }
     }
@@ -185,8 +192,39 @@ while (!AllDone) {
 Above, you can see a simple loop that properly drives a single execution context on Windows.
 `OVERLAPPED_ENTRY` objects received from `GetQueuedCompletionStatusEx` are used to get the submission queue entry and then call its completion handler.
 
-In a real application, these completion events may come both from MsQuic and the application itself, therefore, this means **the application must use the same base format for its own submission entries**.
-This is necessary to be able to share the same event queue object.
+Applications may also perform their own IOs alongside those of MsQuic, pointing to the same IOCP, so that the same threads may efficiently service both MsQuic and non-MsQuic IOs together.
+
+The sample above makes the assumption that _all_ dequeued OVERLAPPED_ENTRY values are in fact pointing to `CXPLAT_SQE` objects (which embed the traditional Win32 `OVERLAPPED` structure).
+To simplify mixing MsQuic with non-MsQuic IO, the application may decide to use the same base format as `CXPLAT_SQE` for its own IOs and streamline the event loop as in the above example. This approach will work for all platforms and not only Windows user-mode.
+
+On Windows, if the application has its own structure embedding the `OVERLAPPED` for its own non-MsQuic IOs and cannot align with the same convention as `CXPLAT_SQE`, it may configure MsQuic to use a specific `CompletionKey`
+when posting packets to the IOCP. By inspecting this `CompletionKey`, the application may distinguish MsQuic-based IO (which means it is a `CXPLAT_SQE`), from non-MsQuic IO that may be in any other arbitrary shape.
+
+```c++
+HANDLE IOCP = CreateIoCompletionPort(INVALID_HANDLE_VALUE, nullptr, 0, 1);
+ULONG_PTR completionKeyForMsQuic = 0x11223344;
+CXPLAT_EVENTQ queueConfig = { IOCP, completionKeyForMsQuic };
+QUIC_EXECUTION_CONFIG ExecConfig = { 0, &queueConfig };
+
+// [...]
+
+while (true) {
+    uint32_t WaitTime = MsQuic->ExecutionPoll(ExecContext);
+
+    ULONG OverlappedCount = 0;
+    OVERLAPPED_ENTRY Overlapped[8];
+    if (GetQueuedCompletionStatusEx(IOCP, Overlapped, ARRAYSIZE(Overlapped), &OverlappedCount, WaitTime, FALSE)) {
+        for (ULONG i = 0; i < OverlappedCount; ++i) {
+            if (Overlapped[i].lpCompletionKey == completionKeyForMsQuic) {
+                CXPLAT_SQE* Sqe = CONTAINING_RECORD(Overlapped[i].lpOverlapped, CXPLAT_SQE, Overlapped);
+                Sqe->Completion(&Overlapped[i]);
+            } else {
+                // This IO did not originate from MSQuic and is app-specific.
+            }
+        }
+    }
+}
+```
 
 # See Also
 
--- a/src/inc/msquic_winuser.h
+++ b/src/inc/msquic_winuser.h
@@ -382,7 +382,7 @@ QuicAddrToString(
 // Event Queue Abstraction
 //
 
-typedef HANDLE QUIC_EVENTQ;
+typedef struct CXPLAT_EVENTQ QUIC_EVENTQ;
 
 typedef OVERLAPPED_ENTRY QUIC_CQE;
 
--- a/src/inc/quic_platform_winuser.h
+++ b/src/inc/quic_platform_winuser.h
@@ -713,7 +713,11 @@ CxPlatEventWaitWithTimeout(
 // Event Queue Interfaces
 //
 
-typedef HANDLE CXPLAT_EVENTQ;
+typedef struct CXPLAT_EVENTQ {
+    HANDLE IoCompletionPort;
+    ULONG_PTR CompletionKey;
+} CXPLAT_EVENTQ;
+
 typedef OVERLAPPED_ENTRY CXPLAT_CQE;
 typedef
 _IRQL_requires_max_(PASSIVE_LEVEL)
@@ -736,7 +740,9 @@ CxPlatEventQInitialize(
     _Out_ CXPLAT_EVENTQ* queue
     )
 {
-    return (*queue = CreateIoCompletionPort(INVALID_HANDLE_VALUE, NULL, 0, 1)) != NULL;
+    queue->CompletionKey = 0;
+
+    return (queue->IoCompletionPort = CreateIoCompletionPort(INVALID_HANDLE_VALUE, NULL, 0, 1)) != NULL;
 }
 
 QUIC_INLINE
@@ -745,7 +751,7 @@ CxPlatEventQCleanup(
     _In_ CXPLAT_EVENTQ* queue
     )
 {
-    CloseHandle(*queue);
+    CloseHandle(queue->IoCompletionPort);
 }
 
 QUIC_INLINE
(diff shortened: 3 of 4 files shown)


src/inc/quic_platform_winuser.h
```diff
@@ -755,38 +761,33 @@ CxPlatEventQAssociateHandle(
...
     )
 {
-    return *queue == CreateIoCompletionPort(fileHandle, *queue, 0, 0);
+    return queue->IoCompletionPort == CreateIoCompletionPort(fileHandle, queue->IoCompletionPort, queue->CompletionKey, 0);
 }
 
 QUIC_INLINE
 BOOLEAN
-CxPlatEventQEnqueue(
+CxPlatEventQEnqueueEx( // Windows specific extension
```
fbrosseau, line 769: I simply inverted the Ex and regular function to reduce dup a little bit. Now regular calls ex with num_bytes=0.

mtfriesen:
@anrossi could you help review this? It might overlap with our own upcoming execution context work.

I'd like to try to find a cross-platform way to solve this without MsQuic imposing its view of the IO world onto the application, and instead allow MsQuic to delegate the lowest level of IO to the application. On Windows, that would mean the application submits and completes the OVERLAPPED structure.

src/inc/msquic_winuser.h
```diff
@@ -382,7 +382,7 @@ QuicAddrToString(
 // Event Queue Abstraction
 //
 
-typedef HANDLE QUIC_EVENTQ;
+typedef struct CXPLAT_EVENTQ QUIC_EVENTQ;
```
nibanks: For now, msquic*.h files can't require dependency on the quic_platform*.h files. So please duplicate the definition (with the different name) instead of typedef'ing the CXPLAT one.

fbrosseau: Ha, I copied this exact line from one of the other abstraction files.

I started with exactly what you're suggesting, but then there's generic code that expects to just assign it by value

fbrosseau: hold on I can't find it anymore. Maybe I'm confused and I made that up

fbrosseau: [here](https://github.com/microsoft/msquic/blob/28649f704c72d1e1e286cfeeca24afca9a390c78/src/inc/msquic_posix.h#L552)

conflicts with [here](https://github.com/microsoft/msquic/blob/28649f704c72d1e1e286cfeeca24afca9a390c78/src/platform/platform_worker.c#L393)

edit: sorry I remembered the wrong build break. I just updated the second link to the correct location.

docs/Execution.md
```diff
@@ -185,8 +197,39 @@ while (!AllDone) {
...
-This is necessary to be able to share the same event queue object.
+Applications may also perform their own IOs alongside those of MsQuic, pointing to the same IOCP, so that the same threads may efficiently service both MsQuic and non-MsQuic IOs together.
+
+The sample above makes the assumption that _all_ dequeued OVERLAPPED_ENTRY values are in fact pointing to `CXPLAT_SQE` objects (which embed the traditional Win32 `OVERLAPPED` structure).
+To simplify mixing MsQuic with non-MsQuic IO, the application may decide to use the same base format as `CXPLAT_SQE` for its own IOs and streamline the event loop as in the above example. This approach will work for all platforms and not only Windows user-mode.
+
+On Windows, if the application has its own structure embedding the `OVERLAPPED` for its own non-MsQuic IOs and cannot align with the same convention as `CXPLAT_SQE`, it may configure MsQuic to use a specific `CompletionKey`
+when posting packets to the IOCP. By inspecting this `CompletionKey`, the application may distinguish MsQuic-based IO (which means it is a `CXPLAT_SQE`), from non-MsQuic IO that may be in any other arbitrary shape.
```
nibanks: > it may configure MsQuic to use a specific `CompletionKey`

But that's not what this PR is actually doing. No where in here are you providing a mechanism to tell the MsQuic library to use a specific key when it creates its own SQEs internally. At best, this API allows the app to configure its own non-zero keys that could be use to distinguished from MsQuic's zero key IOs.

fbrosseau: Sorry, I am not following?

> when it creates its own SQEs internally

This short sentence multiple meanings, but there are two cases:

1. Actual IO, such as MSQuic issuing WSARecv or anything else. It is the kernel that will post a packet to the IOCP. In this case, the kernel will pick the completion key that the library associated to the socket at binding time (`CxPlatEventQAssociateHandle` aka `CreateIoCompletionPort`)

2. Internal packet posting that's not actually IO (`CxPlatEventQEnqueueEx` aka `PostQueuedCompletionStatus`). This system call takes in the CompletionKey that should be delivered to the eventual `GQCS`.

I have updated both to take the value that the application can now provide as part of the `CXPLAT_EVENTQ`, which does mean that 100% of packets being posted to the IOCP, that came from MsQuic in any form, will have the requested CompletionKey.

If the application does NOT provide custom `CXPLAT_EVENTQ`, then it remains as before - calls `CxPlatEventQInitialize` and the key will be 0.

src/inc/msquic_winuser.h
fbrosseau: ahhhhh I understood CXPLAT_ <-> QUIC_ in reverse ok. QUIC are the public types. Let me fix that.

fbrosseau: Ok I fixed my new ones, also updated the doc accordingly. I did not touch posix.h however.

fbrosseau:
> @anrossi could you help review this? It might overlap with our own upcoming execution context work.
> 
> I'd like to try to find a cross-platform way to solve this without MsQuic imposing its view of the IO world onto the application, and instead allow MsQuic to delegate the lowest level of IO to the application. On Windows, that would mean the application submits and completes the OVERLAPPED structure.

Thank you. I don't know the details of these future expansion plans, but I believe my proposal is the most natural way to let applications mix diverse uses to existing Windows IOCP loops. This is nothing new - Winsock RIO does exactly this in IOCP mode [here](https://learn.microsoft.com/en-us/windows/win32/api/mswsock/ns-mswsock-rio_notification_completion) (it takes in the IOCP handle, the completion key, and in their case a fixed reusable OVERLAPPED, but that last part doesn't apply here).

At a high level, the idea is simply that the program must be in control of all completion keys. This allows it to dispatch each completion to the appropriate subsystem.

mtfriesen:
> Thank you. I don't know the details of these future expansion plans, but I believe my proposal is the most natural way to let applications mix diverse uses to existing Windows IOCP loops.
> At a high level, the idea is simply that the program must be in control of all completion keys. This allows it to dispatch each completion to the appropriate subsystem.

I agree this is a viable solution for some Windows apps: those that do not rely on a specific OVERLAPPED layout and have not already taken some incompatible dependency on the completion key. If an existing app already does something like `struct MY_OVERLAPPED { OVERLAPPED o; VOID (*MyCompletionRoutine)(VOID* Context); };` and hypothetically uses the completion key to store the `Context`, then this will not work. We want to allow applications to provide arbitrary inputs to the lowest level IO, rather than impose any opinions/restrictions from MsQuic. More importantly, perhaps, is the case of all the cross-platform IO completion patterns: we want to allow cross-platform applications to be built above MsQuic with minimal code differences.

My high level idea (not thought through, since this is future work) is to allow apps to provide IO wrapper callback routines along the lines of `OVERLAPPED* AppGetSqe(CXPLAT_SQE* MsQuicSqe)` on Windows and `VOID* AppGetSqe(CXPLAT_SQE* MsQuicSqe)` on Linux to wrangle the CXPLAT_SQE structure into the standard IO types for the platforms. The apps could then poll the completion ports and unwrap their opaque IO types and obtain the original CXPLAT_SQE provided by MsQuic.

mtfriesen:
More concisely: this is a general gap in the MsQuic execution context API, and we want to try to solve it for all platforms MsQuic supports. Solving today for a subset of Windows apps will make a more general solution more difficult to implement.

fbrosseau:
I understand the desire to keep platforms as uniform as can be, but this is a feature that only Windows invented (the permanent hard-linking of a file to an IOCP, along with a key), and it would be unfortunate to not allow it as an extensibility point in MSQuic. The way I see it, allowing consumers to pick their CompletionKey is orthogonal to implementing a true cross-platform solution. Applications could choose one way or the other. Basically, all I am suggesting is that instead of leaving the concept of CompletionKey completely unused by MsQuic (in favor of generic solutions since all other platforms only have a single context pointer instead of 2), let the application (optionally) drive it, if it does care about completion keys.

> I agree this is a viable solution for some Windows apps: those that do not rely on a specific OVERLAPPED layout and have not already taken some incompatible dependency on the completion key. If an existing app already does something like struct MY_OVERLAPPED { OVERLAPPED o; VOID (*MyCompletionRoutine)(VOID* Context); }; and hypothetically uses the completion key to store the Context, then this will not work. 

I am not sure I am following the problem in this description. Say that yes, the application already has dynamic, per-file context (typically, some heap-allocated state, and the context is its pointer). All this application has to do to integrate MSQuic, is to assign a magic CompletionKey that will never ever be a valid context address to MsQuic - ULONGPTR_MAX or anything of the sort. If it's not the magic key, then it's app-private IO with a dynamic key. Or simply streamline the whole thing and have a dynamic context pointer for MSQuic too.

mtfriesen:
> I understand the desire to keep platforms as uniform as can be, but this is a feature that only Windows invented (the permanent hard-linking of a file to an IOCP, along with a key), and it would be unfortunate to not allow it as an extensibility point in MSQuic. The way I see it, allowing consumers to pick their CompletionKey is orthogonal to implementing a true cross-platform solution. Applications could choose one way or the other. Basically, all I am suggesting is that instead of leaving the concept of CompletionKey completely unused by MsQuic (in favor of generic solutions since all other platforms only have a single context pointer instead of 2), let the application (optionally) drive it, if it does care about completion keys.

I agree this is, at the lowest level, orthogonal, but it adds more complexity that MsQuic will need to support until the next breaking API change, which we hope never to do. I want to be clear that I'm not strictly opposed to this change, just that I think we need to balance taking this as-is versus what it will cost us in the long run once we provide a general solution.

> I am not sure I am following the problem in this description. Say that yes, the application already has dynamic, per-file context (typically, some heap-allocated state, and the context is its pointer). All this application has to do to integrate MSQuic, is to assign a magic CompletionKey that will never ever be a valid context address to MsQuic - ULONGPTR_MAX or anything of the sort. If it's not the magic key, then it's app-private IO with a dynamic key.

We need to avoid assuming the IO is associated with a file handle: IOCPs can contain arbitrary IO completions, such as those submitted via [PostQueuedCompletionStatus](https://learn.microsoft.com/en-us/windows/win32/fileio/postqueuedcompletionstatus). In the most general case, it is easy to imagine an application that will not support this completion key concept, as it can be used for other purposes, and other libraries imposing similar (but different) restrictions on IO may have conflicting designs.

fbrosseau:
Applications that use PQCS and that care about completion key must be disciplined to pass the right key to PQCS, or they will not function. That is true even outside of MSQuic.

MSQuic does use PQCS also, of course. In this PR, I am handling that too. MSQuic's PQCS usages carry the right key, like IOs. This PR makes every completion packet that originates in any way from MSQuic carry the key, instead of 0.

src/inc/quic_platform_winuser.h
```diff
@@ -736,7 +740,9 @@ CxPlatEventQInitialize(
     _Out_ CXPLAT_EVENTQ* queue
     )
 {
-    return (*queue = CreateIoCompletionPort(INVALID_HANDLE_VALUE, NULL, 0, 1)) != NULL;
+    queue->CompletionKey = 0;
+
+    return (queue->IoCompletionPort = CreateIoCompletionPort(INVALID_HANDLE_VALUE, NULL, 0, 1)) != NULL;
```
guhetier, line 745: nit: As this line becomes more complex, please split the assignment and the comparison/return.

src/inc/quic_platform_winuser.h
```diff
@@ -755,38 +761,33 @@ CxPlatEventQAssociateHandle(
...
 QUIC_INLINE
 BOOLEAN
-CxPlatEventQEnqueueEx( // Windows specific extension
+CxPlatEventQEnqueue(
     _In_ CXPLAT_EVENTQ* queue,
-    _In_ CXPLAT_SQE* sqe,
-    _In_ uint32_t num_bytes
-    )
+    _In_ CXPLAT_SQE* sqe
+)
```
guhetier, line 788: nit: `)` should be aligned with the parameters

guhetier (commented):
The change looks technically ok (modulo a few nits).

I agree with the comments above that a more holistic solution would be preferable - which would potentially mean a redesign of the feature.

IMO, as long as the feature is in preview, we could go with this change for now and simply revert it when we implement something different.

mtfriesen:
> IMO, as long as the feature is in preview, we could go with this change for now and simply revert it when we implement something different.

As a platform team, _citation needed_. I'm assuming this PR is coming to us because this PR author needs to use this feature in important software, with its own ship cycle, etc.? Are we free to break their app at will?

guhetier:
> > IMO, as long as the feature is in preview, we could go with this change for now and simply revert it when we implement something different.
> 
> As a platform team, _citation needed_. I'm assuming this PR is coming to us because this PR author needs to use this feature in important software, with its own ship cycle, etc.? Are we free to break their app at will?

As long as the feature is in "Preview", yes. That is the entire point of the "Preview". If anybody uses a feature marked as preview, they do agree to being broken, at any time, in any way. If they can't afford it, they can work with us to move the feature out of preview, once it is ready.
The fact that this is preview is what allows us to do any of the improvement mentioned here at all - this PR is a breaking api change by itself.

fbrosseau:
Although I believe that the flexibility of controlling CompletionKey is the most standard way to "lend" IOCPs in Windows, with longstanding existing precedents (RIO, win32 threadpool, etc), I had missed your explanation above with the plan of having the application provide an OVERLAPPED of its liking (and unwrap it on the other end). That plan obviously makes total sense too, and as you point out, will be cross-platform. And this plan would also fit well in the IO loops I know of.

If you believe that my proposal would no longer be justified once the OVERLAPPED-wrapping API ships, then yes I think it's better to not go ahead with my proposal and just wait for OVERLAPPED-wrapping.

mtfriesen:
> > As a platform team, _citation needed_. I'm assuming this PR is coming to us because this PR author needs to use this feature in important software, with its own ship cycle, etc.? Are we free to break their app at will?
> 
> As long as the feature is in "Preview", yes. That is the entire point of the "Preview". If anybody uses a feature marked as preview, they do agree to being broken, at any time, in any way. If they can't afford it, they can work with us to move the feature out of preview, once it is ready. The fact that this is preview is what allows us to do any of the improvement mentioned here at all - this PR is a breaking api change by itself.

Theoretically, sure. My point was hinting at the long experience of our platform teams finding that APIs cannot change, in practice, once a critical mass of consumers exist. Again, I fully agree with your interpretation of the rules, but the world runs on de facto.

mtfriesen:
> Although I believe that the flexibility of controlling CompletionKey is the most standard way to "lend" IOCPs in Windows, with longstanding existing precedents (RIO, win32 threadpool, etc), I had missed your explanation above with the plan of having the application provide an OVERLAPPED of its liking (and unwrap it on the other end). That plan obviously makes total sense too, and as you point out, will be cross-platform. And this plan would also fit well in the IO loops I know of.
> 
> If you believe that my proposal would no longer be justified once the OVERLAPPED-wrapping API ships, then yes I think it's better to not go ahead with my proposal and just wait for OVERLAPPED-wrapping.

Do you have a timeframe you need to be unblocked? I expect we will get to the OVERLAPPED model in the next few months, and that it should be a relatively small change, but it's not committed to a specific timeline yet.

fbrosseau:
Oh no no timeframe for this, it was just exploration of an idea to integrate MsQuic better. So definitely if my suggestion derails your existing plans, I am totally fine with cancelling this and waiting. And I do appreciate the very very quick response times from all the members of this project too!

Ok for now let me cancel this PR. Your suggested feature will work when you get to it.

PR #5368, opened 2025-08-20
Fix incorrect cmake option name in build.ps1 and cmakepreset

penglei0:
## Description

- QUIC_LINUX_LOG_ENCODER is an incorrect CMake options, should be modified to `QUIC_LOGGING_TYPE`.

## Testing

No new tests should be required.

## Documentation

No docs need be updated.

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -440,9 +440,6 @@ function CMake-Generate {
     $Arguments += " -DQUIC_TLS_LIB=" + $Tls
     $Arguments += " -DQUIC_OUTPUT_DIR=""$ArtifactsDir"""
 
-    if ($IsLinux) {
-        $LoggingType = "lttng"
-    }
     if (!$DisableLogs) {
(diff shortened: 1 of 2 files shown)


penglei0:

scripts/build.ps1
```diff
@@ -441,7 +441,7 @@ function CMake-Generate {
     $Arguments += " -DQUIC_OUTPUT_DIR=""$ArtifactsDir"""
 
     if ($IsLinux) {
-        $Arguments += " -DQUIC_LINUX_LOG_ENCODER=lttng"
+        $LoggingType = "lttng"
```
guhetier: This would override any value set to `$LoggingType` as a script parameter.
I think that line should simply be removed. `$LoggingType == ""` is fine if the value is already in the CMake cache or if logging should not be enabled.

guhetier (approved):
Thanks for the PR!

Issue #5307, opened 2025-07-31
Support Careful Resume - Part 3


sm-msft:
This task starts where task https://github.com/microsoft/msquic/issues/5214 left off.

Tentatively planned work items:
 Schannel Support for Application Managed Resumption tickets + testing

sm-msft:
Refer to the prototype + full code change for this feature in PR https://github.com/microsoft/msquic/pull/5311

Issue #5162, opened 2025-06-12
Add Schannel Active Session Management


anrossi:
### Describe the feature you'd like supported

Add support for setting custom data in resumption tickets on Schannel.  

### Proposed solution

A new CREDENTIAL_FLAG would need to be added to opt in to this functionality. This flag changes the flags passed to Schannel during initialization and controls whether the `ConnectionSendResumptionTicket` calls down into Schannel or not. 

This enables the Careful Resume feature for Schannel.

PR #5311, opened 2025-08-01
Prototype - Custom resumption tickets over Schannel

sm-msft:
## Description
Goal: Introduce custom resumption tickets and resumption ticket handling to schannel code in MsQuic

-Code to push/retrieve server app state in a resumption ticket over SChannel

- Code to push/retrieve client resumption ticket over SChannel
- Few corresponding test changes

## Testing

1 existing unit test runs successfully. Remaining scenarios need further work to succeed.

## Documentation

N/A

PR #5371, opened 2025-08-20
Server app managed custom resumption tickets/Client app managed resumption tickets over SChannel

sm-msft:
## Description
SChannel in Windows 2025 and newer releases allows 
- Server app to generate/manage resumption tickets that can include custom Server application state
- Client app to manage the resumption ticket blobs it receives from the server

These features help implement custom 1-RTT resumption tickets on SChannel server applications and custom resumption ticket handling on SChannel clients.

This PR adds support for these features and integrates it into existing MsQuic code that utilizes similar features supported in OpenSSL and quictls libraries. Bulk of these changes are in tls_schannel.c wrapper.

Where available (and when not explicitly disabled), MsQuic has been using the automatic SChannel 1-RTT resumption handling thus far and it would remain the default mechanism in the library. 

This change introduces a new QUIC_CREDENTIAL_FLAG for SChannel use
 - for a server app to opt into exclusively managing custom application state tickets/resumption ticket issuance 
 - for a client app to opt into exclusively managing resumption ticket reception and use.

An additional QUIC_CREDENTIAL_FLAG allows an app to turn disable the use of 1-RTT/resumption in SChannel.
These flags are available only under the preview features macro.

Methods added to tls_schannel.c and other wrappers help enable the new functionality on Server2025 and newer releases and also to target 1-RTT tests on the appropriate platforms.

This change is linked to these issues: https://github.com/microsoft/msquic/issues/5307 and #5162 

## Testing

Existing 1-RTT unit tests, BVTs have been enabled for SChannel on Server2025+ and new unit tests are added for additional coverage in SChannel/OpenSSL/quictls.

Existing automatic resumption tests cover the default scenario and older Windows OS scenarios.

## Documentation

Documentation has been added to the newly introduced credential flags.


sm-msft:
This draft PR was used to develop this change: https://github.com/microsoft/msquic/pull/5311

src/platform/tls_schannel.c
```diff
@@ -2717,50 +3105,141 @@ CxPlatTlsWriteDataToSchannel(
...
+                TlsContext->RxSessionTicketAllocLength = OutSecBufferDesc.pBuffers[i].cbBuffer;
+                OutSecBufferDesc.pBuffers[i].pvBuffer = (void*)TlsContext->RxSessionTicket;
+                Result |= CXPLAT_TLS_RESULT_CONTINUE;
+                SessionTicketAllocated = TRUE;
             }
-            if (TlsContext->PeerTransportParams != NULL) {
+
+            if (TLSExtensionAllocated && AppSessionStateAllocated && SessionTicketAllocated) {
                 break;
             }
```
csujedihy, line 3235: Is this check still needed when you have the check below
```
        if (!(Result & CXPLAT_TLS_RESULT_ERROR) &&
            (TLSExtensionAllocated || AppSessionStateAllocated || SessionTicketAllocated)) {
            break;
        }
```

src/platform/tls_schannel.c
```diff
@@ -3336,3 +3918,34 @@ QuicTlsPopulateOffloadKeys(
...
+}
+
+BOOLEAN
+CxPlatSupportsInline1RTTAppStateValidation()
+{
+    //
+    // Inline 1-RTT app state validation is not supported on Windows
+    //
+    return FALSE;
+}
```
csujedihy, line 3942: What does "Inline 1-RTT app state validation" mean?

src/test/bin/quic_gtest.cpp
```diff
@@ -777,6 +785,57 @@ TEST_P(WithHandshakeArgs1, Resume) {
     }
 }
 
+extern "C"
+BOOLEAN
+CxPlatSupportsTicketManagement();
```
guhetier, line 790: Test should not use internal functions.
To skip tests, please use the pattern used below (`QUIC_DISABLE_0RTT_TESTS`) so it is controled as a test configuration.

src/test/bin/quic_gtest.cpp
```diff
@@ -777,6 +785,57 @@ TEST_P(WithHandshakeArgs1, Resume) {
...
+            (uint8_t)GetParam().GreaseQuicBitExtension,
+            QUIC_TEST_ASYNC_CONFIG_DISABLED,
+            (uint8_t)GetParam().MultiPacketClientInitial,
+            QUIC_TEST_RESUMPTION_ENABLED,
+            0,   // RandomLossPercentage
+            true // True Resumption
+        };
+        ASSERT_TRUE(DriverClient.Run(IOCTL_QUIC_RUN_CONNECT, Params));
+    }
+    else {
```
guhetier, line 823:
```suggestion
    } else {
```

src/test/lib/HandshakeTest.cpp
```diff
@@ -36,13 +36,17 @@ void QuicTestUninitialize()
...
+
 void
 QuicTestPrimeResumption(
     _In_ QUIC_ADDRESS_FAMILY QuicAddrFamily,
     _In_ MsQuicRegistration& Registration,
     _In_ MsQuicConfiguration& ServerConfiguration,
     _In_ MsQuicConfiguration& ClientConfiguration,
-    _Out_ QUIC_BUFFER** ResumptionTicket
+    _Out_ QUIC_BUFFER** ResumptionTicket,
+    _In_ bool TrueResume
```
guhetier, line 49: I think a better name is needed. `TrueResume` is very vague. I think it means "send a non-empty resume ticket"? A name capturing this would be better.

guhetier (commented):
Review in progress

src/platform/tls_schannel.c
```diff
@@ -612,6 +615,56 @@ typedef struct CXPLAT_TLS {
...
+
+    //
+    // Length of allocated application state buffer on the server
+    //
+    uint32_t RxAppSessionStateAllocLength;
+
+    //
+    // Received application session state on the server, typically notified to the server app.
+    //
+    PSEC_APP_SESSION_STATE RxAppSessionState;
```
csujedihy, line 645: Can we have "client" and "server" in the name for these variables? It's a little difficult to follow when looking at how they are used.

Maybe we can match the bit flags defined above?

src/inc/msquic.h
```diff
@@ -147,6 +147,14 @@ typedef enum QUIC_CREDENTIAL_FLAGS {
     QUIC_CREDENTIAL_FLAG_INPROC_PEER_CERTIFICATE                = 0x00080000, // Schannel only
     QUIC_CREDENTIAL_FLAG_SET_CA_CERTIFICATE_FILE                = 0x00100000, // OpenSSL only currently
     QUIC_CREDENTIAL_FLAG_DISABLE_AIA                            = 0x00200000, // Schannel only currently
+
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+
+    QUIC_CREDENTIAL_FLAG_DISABLE_RESUMPTION                     = 0x00400000, // TODO: Combine this with CXPLAT_TLS_CREDENTIAL_FLAG_DISABLE_RESUMPTION
```
guhetier, line 153: TODO to be solved before merging.

src/test/bin/quic_gtest.cpp
```diff
@@ -121,7 +122,14 @@ class QuicTestEnvironment : public ::testing::Environment {
             }
 #endif
             memcpy(&ServerSelfSignedCredConfig, SelfSignedCertParams, sizeof(QUIC_CREDENTIAL_CONFIG));
+            memcpy(&ServerSelfSignedCredConfigSChannelResumption, SelfSignedCertParams, sizeof(QUIC_CREDENTIAL_CONFIG));
             memcpy(&ServerSelfSignedCredConfigClientAuth, SelfSignedCertParams, sizeof(QUIC_CREDENTIAL_CONFIG));
+
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
```
guhetier, line 128: Question: Do we actually need this ifdef?
As far as I know:
- test code is always compiled with `QUIC_API_ENABLE_PREVIEW_FEATURES`
- preview features implementation is always compiled in MsQuic, `QUIC_API_ENABLE_PREVIEW_FEATURES` only control the visibility of definitions in the header

These uses of `QUIC_API_ENABLE_PREVIEW_FEATURES` in test code might be a remain from the past and not have any effect.

src/test/bin/quic_gtest.cpp
```diff
@@ -777,6 +785,57 @@ TEST_P(WithHandshakeArgs1, Resume) {
...
+    return CxPlatSupportsTicketManagement();
+}
+
+TEST_P(WithHandshakeArgs1, TrueResume) {
+    if (!CanRunExplicitResumptionTests()) {
+        GTEST_SKIP_("Explicit resumption tests not supported on this platform");
+    }
+
+#ifndef QUIC_API_ENABLE_PREVIEW_FEATURES
+    GTEST_SKIP_("Resumption ticket management is not supported in this build.");
```
guhetier, line 804: That skip reason is misleading, the build always support preview features.

src/test/bin/quic_gtest.cpp
```diff
@@ -777,6 +785,57 @@ TEST_P(WithHandshakeArgs1, Resume) {
...
+BOOLEAN
+CxPlatSupportsTicketManagement();
+
+BOOLEAN
+CanRunExplicitResumptionTests()
+{
+    return CxPlatSupportsTicketManagement();
+}
+
+TEST_P(WithHandshakeArgs1, TrueResume) {
```
guhetier, line 798: Please adapt the test name to describe what is tested here. If needed, feel free to change the `Resume` test name above to make it clear what the difference is.

src/test/lib/HandshakeTest.cpp
```diff
@@ -51,6 +55,7 @@ QuicTestPrimeResumption(
     struct PrimeResumption {
         CxPlatEvent ShutdownEvent;
         MsQuicConnection* Connection {nullptr};
+        bool TrueResume;
```
guhetier, line 58: This is a C++ struct, please intialize this

src/test/lib/HandshakeTest.cpp
```diff
@@ -59,13 +64,28 @@ QuicTestPrimeResumption(
...
+                    QUIC_SEND_RESUMPTION_FLAG_FINAL,
+                    TicketBufferLength,
+                    TicketBufferPtr);
             }
             return QUIC_STATUS_SUCCESS;
         }
     };
 
     PrimeResumption Context;
+    Context.TrueResume = TrueResume;
```
guhetier, line 88: To simplify this and limit the number of test specific branches in the code, I suggest:

- don't set a `TrueResume` flag, instead, store the ticket in the context
- when in the `TrueResume` test, populate the ticket when initializing the context, otherwise, keep the ticket empty (zero length)
- in the validation, always check that the ticket receive is the one set

This way, the validation part of the test is independent from the parameters and it gives more flexibility to test different tickets in the future

docs/api/QUIC_CREDENTIAL_CONFIG.md
```diff
@@ -163,7 +163,23 @@ Enable CA certificate file provided in the `CaCertificateFile` member.
...
 `QUIC_CREDENTIAL_FLAG_DISABLE_AIA`
 
-The following flag can be set to explicitly disable AIA retrievals. Only valid on Windows.
+Explicitly disable AIA retrievals. Only valid on Windows.
+
+`QUIC_CREDENTIAL_FLAG_DISABLE_RESUMPTION`
+
+**Preview feature**: This flag is in [preview](../PreviewFeatures.md). It should be considered unstable and can be subject to breaking changes.
+
+Disables TLS session resumption. Only valid on Windows with SChannel.
```
guhetier, line 172:
```suggestion
Disables TLS session resumption. Only supported on Windows with SChannel.
```

docs/api/QUIC_CREDENTIAL_CONFIG.md
```diff
@@ -163,7 +163,23 @@ Enable CA certificate file provided in the `CaCertificateFile` member.
...
+
+`QUIC_CREDENTIAL_FLAG_ALLOW_RESUMPTION_TICKET_MANAGEMENT`
+
+**Preview feature**: This flag is in [preview](../PreviewFeatures.md). It should be considered unstable and can be subject to breaking changes.
+
+Enables exclusive application management of TLS session resumption tickets and disables any automatic resumption ticket management in the underlying TLS library. Only valid on Windows Server 2025 or newer OS, with SChannel TLS.
```
guhetier, line 178:
```suggestion
Enables exclusive application management of TLS session resumption tickets and disables any automatic resumption ticket management in the underlying TLS library. Only supported on Windows Server 2025 or newer OS, with SChannel TLS.
```

guhetier: I also think that some expension of "exclusive application management" would be helpful, it is pretty vague.
Maybe something along the line of "when set, MsQuic will use the TLS session resumption tickets provided by the app and forward any TLS resumption ticket received directly to the app"?

src/inc/msquic.h
```diff
@@ -147,6 +147,14 @@ typedef enum QUIC_CREDENTIAL_FLAGS {
...
+    QUIC_CREDENTIAL_FLAG_ALLOW_RESUMPTION_TICKET_MANAGEMENT     = 0x00800000, // Schannel only currently
```
guhetier, line 154: I think this name can be improved, the idea is that it allows the app to manage resumption tickets, not that they would not be managed otherwise.

Maybe `QUIC_CREDENTIAL_FLAG_APP_MANAGED_RESUMPTION_TICKETS`?

src/platform/unittest/TlsTest.cpp
```diff
@@ -32,6 +33,26 @@ const QUIC_HKDF_LABELS HkdfLabels = { "quic key", "quic iv", "quic hp", "quic ku
 
 bool IsWindows2019() { return OsRunner && strcmp(OsRunner, "windows-2019") == 0; }
 bool IsWindows2022() { return OsRunner && strcmp(OsRunner, "windows-2022") == 0; }
+uint8_t* AppSessionState = nullptr;
```
guhetier, line 36: Please avoid global mutable state in unit tests. It makes it possible that different tests interfere with each others.

src/platform/tls_schannel.c
```diff
@@ -550,6 +550,9 @@ typedef struct CXPLAT_TLS {
     BOOLEAN PeerTransportParamsReceived : 1;
     BOOLEAN HandshakeKeyRead : 1;
     BOOLEAN ApplicationKeyRead : 1;
+    BOOLEAN ServerAppSessionStateRx : 1;       // If TRUE, AppSessionState was received on the server
```
guhetier, line 553: `ServerAppSessionStateReceive` is more readable, understandable, and matches the precedent from `PeerTransportParamsReceived`.
Same for the other two flags.

src/platform/tls_schannel.c
```diff
@@ -612,6 +615,56 @@ typedef struct CXPLAT_TLS {
...
+    //
+    // Length of the allocated session ticket struct
+    //
+    uint32_t TxSessionTicketAllocLength;
+
+    //
+    // Resumption/Reconnection Session ticket on the client used when establishing the connection with a server
+    // Use this if ClientSessionStateTicketForTx is TRUE and this can be used only once.
+    //
+    PSEC_SESSION_TICKET TxSessionTicket;
```
guhetier, line 666: nit: Please avoid `P` pointer types, same above. `SEC_SESSION_TICKET*` is better.

src/platform/tls_schannel.c
```diff
@@ -612,6 +615,56 @@ typedef struct CXPLAT_TLS {
     //
     QUIC_TLS_SECRETS* TlsSecrets;
 
+    //
+    // Session state and resumption have traditionally been managed by the TLS layer without
```
guhetier, line 619: Please don't refer to the past in comments, what is "traditionnally" the case now will be different in the future.
It is better to state what is the case in the present.

src/platform/tls_schannel.c
```diff
@@ -959,6 +1012,26 @@ CxPlatTlsGetProvider(
...
+    if (!(flags & QUIC_CREDENTIAL_FLAG_DISABLE_RESUMPTION) &&
+        (flags & QUIC_CREDENTIAL_FLAG_ALLOW_RESUMPTION_TICKET_MANAGEMENT)) {
+        return TRUE;
+    }
+
+    return FALSE;
+}
+
+BOOLEAN
+AreResumptionTicketsManaged(_In_ const QUIC_CREDENTIAL_FLAGS flags)
```
guhetier, line 1030: Please clarify managed by who. Managed is already a cursed words in names because it is so generic :)
Same comment for all the names about managing resumption tickets.

`AreResumptionTicketsManagedByApp`.

src/platform/tls_schannel.c
```diff
@@ -1124,8 +1197,22 @@ CxPlatTlsSecConfigCreate(
...
+        }
+    }
+
+    if (ResumptionTicketManagementAllowed(AchContext->SecConfig->Flags)) {
+        if (CxPlatSupportsTicketManagement()) {
+            QuicTraceLogVerbose(
+                SchannelTlsInitVerbose,
+                "[ tls] Explicit resumption ticket management enabled");
+        } else {
+            QuicTraceLogWarning(
```
guhetier, line 1210: Does this mean that if the app tries to set the flag to manage ticket but it is not supported by the schannel version, we silently ignore the flag?
This could break the expectation of the app, that is waiting for a callback. I would consider failing here and let the app retry without the flag so they are in control.

src/platform/tls_schannel.c
```diff
@@ -1599,17 +1692,128 @@ CxPlatTlsInitialize(
...
+            memcpy(TlsContext->TxSessionTicket->SessionTicket,
+                Config->ResumptionTicketBuffer,
+                Config->ResumptionTicketLength);
+            TlsContext->TxSessionTicket->SessionTicketSize =
+                (uint16_t)Config->ResumptionTicketLength;
+            TlsContext->ClientSessionStateTicketForTx = TRUE;
+        }
+
+        //
+        // TODO: Config is marked as an _in_ parameter, but we are required to free this buffer in this call
```
guhetier, line 1795: TODO to be resolved.

src/platform/tls_schannel.c
```diff
@@ -2717,50 +3105,141 @@ CxPlatTlsWriteDataToSchannel(
...
+                }
+                TlsContext->RxAppSessionState = OutSecBufferDesc.pBuffers[i].pvBuffer = NULL;
+                TlsContext->RxAppSessionStateAllocLength = 0;
 
-                    QuicTraceLogConnInfo(
-                        SchannelTransParamsBufferTooSmall,
-                        TlsContext->Connection,
-                        "Peer TP too large for available buffer (%u vs. %u)",
+                TlsContext->RxAppSessionState =
+                    (PSEC_APP_SESSION_STATE)CXPLAT_ALLOC_NONPAGED(
```
guhetier, line 3177: Cast should not be needed on the `void*` pointer returned by the allocation.
Same for other allocations.

src/platform/tls_schannel.c
```diff
@@ -2858,34 +3357,117 @@ CxPlatTlsProcessData(
...
+        else if (!TlsContext->IsServer &&
+            TlsContext->ClientSessionStateTicketRx) {
+            //
+            // Notify client application of the RX TLS Session State/Resumption Ticket
+            // This is typically opaque data for the client.
+            //
+            if (TlsContext->SecConfig->Callbacks.ReceiveTicket(
+                TlsContext->Connection,
+                TlsContext->RxSessionTicket->SessionTicketSize,
+                (void*)TlsContext->RxSessionTicket->SessionTicket)) {
```
guhetier, line 3442: Please consider avoiding side effects, especially one as important as an app notification, from inside an `if` condition. Using a temporary variable to store the result is more readable.

Same above.

guhetier: (A bonus is that it can let you aggregate the two logs as "Resumption session ticket delivered to the app. Ticket size: %u bytes. Accepted: %d)

src/platform/tls_schannel.c
```diff
@@ -1599,17 +1692,128 @@ CxPlatTlsInitialize(
...
+                    QUIC_POOL_TLS_EXTRAS);
+
+            if (TlsContext->RxAppSessionState == NULL) {
+                QuicTraceEvent(
+                    AllocFailure,
+                    "Allocation of '%s' failed. (%llu bytes)",
+                    "Initial App Session State",
+                    TlsContext->RxAppSessionStateAllocLength);
+                Status = QUIC_STATUS_OUT_OF_MEMORY;
+                goto Error;
```
guhetier, line 1714: I can't comment on it directly because Github is annoying, but above, there is a "TLS context Created" log.
There used to be no failure possible after that log, this is introducing new ones. I think the log should be moved lower so that it doesn't show up if we actually are going to fail.

src/platform/tls_schannel.c
```diff
@@ -1599,17 +1692,128 @@ CxPlatTlsInitialize(
...
+                    TlsError,
+                    "[ tls][%p] ERROR, %s.",
+                    Config->Connection,
+                    "Resumption ticket is too big");
+                Status = QUIC_STATUS_INVALID_PARAMETER;
+                goto Error;
+            }
+
+            //
+            // Allocate the resumption ticket buffer with an extra byte
```
guhetier, line 1762: Do we need the extra byte? If yes, please explain the "why" in the comment.
If no, please use an "offset of" macro to allocate the exact size and clarify the intent.

src/platform/tls_schannel.c
```diff
@@ -1599,17 +1692,128 @@ CxPlatTlsInitialize(
...
+                    TlsContext->TxSessionTicketAllocLength);
+                Status = QUIC_STATUS_OUT_OF_MEMORY;
+                goto Error;
+            }
+
+            RtlZeroMemory(
+                TlsContext->TxSessionTicket,
+                TlsContext->TxSessionTicketAllocLength);
+
+            memcpy(TlsContext->TxSessionTicket->SessionTicket,
```
guhetier, line 1786: style, the first parameter should be on its own line.

src/platform/tls_schannel.c
```diff
@@ -1812,49 +2029,80 @@ CxPlatTlsWriteDataToSchannel(
 
     uint8_t AlertBufferRaw[2];
 
-    if (*InBufferLength == 0) {
+    //
+    // Ticket data refers only to the server application session state
+    //
+    CXPLAT_DBG_ASSERT(!IsTicketData ||
+                      (IsTicketData &&
```
guhetier, line 2036: redundant check on `IsTicketData`, if it was false, the assert would already be valid with the first line.

src/platform/tls_schannel.c
```diff
@@ -1812,49 +2029,80 @@ CxPlatTlsWriteDataToSchannel(
...
+                       TlsContext->IsServer &&
+                       ManagedResumptionTickets));
+
+    if (!IsTicketData) {
 
         //
-        // If the input length is zero, then we are initializing the client
-        // side, and have a few special differences in this code path.
+        // Call to encode ticket data has different ordering of inputs and different inputs as well.
```
guhetier, line 2043: I don't understand this comment. Different from what? It would be more helpfult to explain the ordering needs in for the different scenarios

src/platform/tls_schannel.c
```diff
@@ -1879,12 +2127,56 @@ CxPlatTlsWriteDataToSchannel(
...
+
+    CXPLAT_STATIC_ASSERT(
+        ISC_REQ_EXPLICIT_SESSION == ASC_REQ_EXPLICIT_SESSION,
+        "To simplify the code, we use the same value for both ISC and ASC");
+
     TlsContext->Workspace.InSecFlags.Flags = ISC_REQ_MESSAGES;
+
+    if (ManagedResumptionTickets) {
+        //
+        // Explicit session management is in force for this context.
```
guhetier, line 2139: "in force": is this a typo? Did you mean "enforced"?

src/platform/tls_schannel.c
```diff
@@ -2817,37 +3296,57 @@ CxPlatTlsWriteDataToSchannel(
     }
 #endif
 
+    if (AppSessionState != NULL) {
```
guhetier, line 3299: :O that function is so long, I had to scroll forever to find this cleanup.
We should break it down somehow (not for this PR)

src/platform/tls_schannel.c
```diff
@@ -2817,37 +3296,57 @@ CxPlatTlsWriteDataToSchannel(
...
 
-    if (!TlsContext->IsServer && State->BufferOffset1Rtt > 0 &&
-        State->HandshakeComplete) {
+    if (State->HandshakeComplete &&
+        State->BufferOffset1Rtt > 0 &&
+        !TlsContext->IsServer &&
+        !AreResumptionTicketsManaged(TlsContext->SecConfig->Flags)){
         //
-        // Schannel currently sends the NST after receiving client finished.
+        // Older client behavior: Schannel sends the NST after receiving client finished.
```
guhetier, line 3349: What is "older" in this context?

src/platform/tls_schannel.c
```diff
@@ -2858,34 +3357,117 @@ CxPlatTlsProcessData(
...
+                    SchannelClientIgnoringTicket,
+                    TlsContext->Connection,
+                    "Client app rejected session ticket. Proceeding with the connection. Ticket size: %u bytes",
+                    TlsContext->RxSessionTicket->SessionTicketSize);
+            }
+
+            //
+            // Client session state ticket is delivered only once.
+            //
+            TlsContext->RxSessionTicket->SessionTicketSize = 0;
```
guhetier, line 3466: Shouldn't we free the memory too in this case?

guhetier (commented):
Review in progress

src/platform/tls_schannel.c
csujedihy: I thought I was the only confused by this name. Whenever I see "managed", I always assume its managed by something else like "managed code". I believe Schannel folks refer to this feature as "explict/active session management" and we should have "active" or "explicit" in the name. "App managed" works too.

src/platform/tls_schannel.c
csujedihy: This isn't client behavior. Sending NST after receiving client finished is a server behavior. Its still the case, I believe, because NSTs can only be sent from the server after handshake but now the server with active management can defer sending the ticket. This piece of code fakes a ticket received event for schannel.

PR #5376, opened 2025-08-22
Add strong partitioning to listeners, connections, and streams

### LINKED ISSUE #5372: Add partition affinity options for all callbacks

All objects generating callbacks (e.g. listener, connection, stream) should have options to set the partition affinity of the callbacks. For listeners, connections should inherit the affinity of the listener.

### Proposed solution

Add `SetParam` options for each object. Streams already inherit the affinity of the connection. 

### Additional context

_No response_

mtfriesen:
## Description

Resolves #5372 

Adds a new (preview) listener parameter `QUIC_PARAM_LISTENER_PARTITION_INDEX` to affinitize all event indications onto that partition and affinitizes all related object (connections, and by transitivity, streams) to that same partition. There is already a preview feature for client connections, `ConnectionOpenInPartition`. This is currently only supported end-to-end on linux with epoll, though much of the churn is cross-platform.

This feature is primarily useful when operating in external execution context modes: it allows fine-grained control over where MsQuic processes all activity for a listener.

I found quite a few issues with the existing (preview) external execution context mode, which I tried to fix independently in #5384, but that fix was incomplete. I eventually had to add an asynchronous `RegistrationClose2` API, else it was impossible to close and unload the registration, execution context, and MsQuic.

## Testing

Added basic functional testing for the existing (preview) execution context feature, as well as targeted functional tests for partition affinity of listeners and connections. Streams are already implicitly aligned to connections and were not tested.

## Documentation

All of these features are in preview; documentation was updated and improved.

### DIFF
--- a/src/test/lib/ApiTest.cpp
+++ b/src/test/lib/ApiTest.cpp
@@ -6864,15 +6864,11 @@ struct TestPartitionCallbackContext {
     bool CloseInStop;
 };
 
-// TODO: remove
-void WrongThread() {}
-
 void
 TestPartitionVerifyCallback(TestPartitionCallbackContext* Context)
 {
     if (TestCurThreadID() != Context->ExpectedThreadId) {
         Context->ActualThreadId = TestCurThreadID();
-        WrongThread();
     }
 }
 


src/core/binding.h
```diff
@@ -210,6 +210,11 @@ typedef struct QUIC_BINDING {
     //
     BOOLEAN Connected : 1;
 
+    //
+    // Indicates that the binding is partitioned.
```
guhetier: The use of "partitioned" to mean "tied to a partition" sound confusing to me, because something partitioned sounds like it is split in multiple pieces, with is opposite.

I would advocate for "Afinitized" or "TiedToPartition" instead

mtfriesen: I'm open to wordsmithing here, though I think it _is_ partitioned: the binding is logically split into multiple pieces (multiple partitions) and this particular binding only represents one of the partitions?

It's possible the bindings for the rest of the partitions don't exist, but that's orthogonal.

guhetier: If I understand your point of view, you conceptually see the binding struct as representing the part of the binding tied to the CPU, with potentially other parts of the same binding that could exist and be tied to a different CPU?

And I see the binding struct as the full binding, restricted/afinitized to a single CPU - since we prevent inserting multiple bindings for the same 4 tuple but different partition Id.

I don't know if there is canonical point of view here. Because only one binding struct exists (vs one per CPU core...), it feels to me more complicated to consider their potential future existence.

mtfriesen: In the future, it would be a goal to allow partitions to coexist at a binding level; however, our current partner requesting this feature doesn't need it, and it's just one more thing we'd have to test and provide official support for.

Conceptually, an endgame here is for an app to open one listener per partition, all binding to UDP:443, set some kind of CID pattern, and have inbound and outbound connections reliably and deterministically execute within those partitions from the API layer all the way down through the kernel to the NIC HW.

That is a long ways off, obviously, so currently the concept of a partitioned binding is nebulous, other than to say: it applies only to one partition, and the rest of the checks are there simply to prevent compat problems today and tomorrow.

src/test/lib/ApiTest.cpp
```diff
@@ -6636,6 +6659,448 @@ QuicTestValidateConnectionPoolCreate()
...
+        auto &Ec = Ecs[i];
+        Ec.QuicEc = Execution[i];
+        CXPLAT_THREAD_CONFIG Config{};
+        Config.Name = "QuicTestPartitionWorker";
+        Config.Callback = QuicTestPartitionWorker;
+        Config.Context = &Ec;
+        TEST_QUIC_SUCCEEDED(Ec.Thread.Create(&Config));
+        TEST_TRUE(Ec.Ready.WaitTimeout(TestWaitTimeout));
+    }
+
```
mtfriesen, line 7112: extra blank line

CMakeLists.txt
```diff
@@ -714,7 +714,7 @@ else() #!WIN32
     if (QUIC_SANITIZER_ACTIVE)
         message(STATUS "Configuring sanitizers: ASAN:${QUIC_ENABLE_ASAN} LSAN:${QUIC_ENABLE_LSAN} UBSAN:${QUIC_ENABLE_UBSAN} EXTRA:${QUIC_ENABLE_EXTRA_SANITIZERS}")
         # Append common flags for all sanitizer options
-        list(APPEND QUIC_COMMON_FLAGS -Og -fno-omit-frame-pointer -fno-optimize-sibling-calls)
+        list(APPEND QUIC_COMMON_FLAGS -O0 -fno-omit-frame-pointer -fno-optimize-sibling-calls)
```
mtfriesen, line 717: dup of #5409

src/platform/platform_posix.c
```diff
@@ -317,7 +317,7 @@ CxPlatRefIncrement(
     _Inout_ CXPLAT_REF_COUNT* RefCount
     )
 {
-    if (__atomic_add_fetch(RefCount, 1, __ATOMIC_SEQ_CST)) {
+    if (__atomic_add_fetch(RefCount, 1, __ATOMIC_SEQ_CST) > 1) {
```
mtfriesen, line 320: dup of #5407

docs/api/RegistrationClose2.md
```diff
@@ -0,0 +1,46 @@
...
+
+# Parameters
+
+`Registration`
+
+A registration handle from a previous call to [RegistrationOpen](RegistrationOpen.md).
+
+`Handler`
+
+A registration close completion handler. It will be invoked once the registration is closed.
```
anrossi: The documentation should explain the expected format of the callback too.

docs/api/RegistrationClose2.md
```diff
@@ -0,0 +1,46 @@
+RegistrationClose2 function
+======
+
+> **Preview**
+> This routine is in preview and is subject to breaking changes.
+
+Closes an existing registration asynchronously.
```
anrossi, line 7: Instead of calling it RegistrationClose2, why not RegistrationCloseAsync, to indicate that it's specifically for async usage?

src/core/binding.h
anrossi: Can we capture this in the comment to help explain what the `partitioned` flag means?
> it applies only to one partition, and the rest of the checks are there simply to prevent compat problems today and tomorrow.

src/core/connection.c
```diff
@@ -1870,7 +1870,7 @@ QuicConnStart(
     UdpConfig.LocalAddress = Connection->State.LocalAddressSet ? &Path->Route.LocalAddress : NULL;
     UdpConfig.RemoteAddress = &Path->Route.RemoteAddress;
     UdpConfig.Flags = CXPLAT_SOCKET_FLAG_NONE;
-    UdpConfig.InterfaceIndex = Connection->State.LocalInterfaceSet ? (uint32_t)Path->Route.LocalAddress.Ipv6.sin6_scope_id : 0, // NOLINT(google-readability-casting)
+    UdpConfig.InterfaceIndex = Connection->State.LocalInterfaceSet ? (uint32_t)Path->Route.LocalAddress.Ipv6.sin6_scope_id : 0; // NOLINT(google-readability-casting)
```
anrossi, line 1873: That's quite the subtle bug!  Does the NOLINT statement have anything to do with the issue here?

src/core/library.c
```diff
@@ -685,31 +708,83 @@ MsQuicLibraryUninitialize(
...
+    }
+
     QuicSettingsCleanup(&MsQuicLib.Settings);
 
     CXPLAT_FREE(MsQuicLib.DefaultCompatibilityList, QUIC_POOL_DEFAULT_COMPAT_VER_LIST);
     MsQuicLib.DefaultCompatibilityList = NULL;
 
     CxPlatDispatchRwLockUninitialize(&MsQuicLib.StatelessRetry.Lock);
 
+    if (MsQuicLib.RegistrationCloseCleanupWorker) {
```
anrossi: In what case would this be null if the library was initialized successfully?

src/core/library.c
```diff
@@ -685,31 +708,83 @@ MsQuicLibraryUninitialize(
...
 #endif
     CxPlatUninitialize();
 }
 
+CXPLAT_THREAD_CALLBACK(RegistrationCleanupWorker, Context)
+{
+    UNREFERENCED_PARAMETER(Context);
+
+    while (!MsQuicLib.RegistrationCloseCleanupShutdown) {
+        CxPlatEventWaitForever(MsQuicLib.RegistrationCloseCleanupEvent);
```
anrossi, line 764: I think this needs to check `RegistrationCloseCleanupShutdown` before attempting to take the lock because an APC could cause a spurious wake of this thread and in that case it should go back to waiting on the event.

src/core/library.c
```diff
@@ -685,31 +708,83 @@ MsQuicLibraryUninitialize(
...
+
+        CxPlatLockAcquire(&MsQuicLib.RegistrationCloseCleanupLock);
+        while (!CxPlatListIsEmpty(&MsQuicLib.RegistrationCloseCleanupList)) {
+            CXPLAT_LIST_ENTRY* Entry =
+                CxPlatListRemoveHead(&MsQuicLib.RegistrationCloseCleanupList);
+            QUIC_REGISTRATION* Registration =
+                CXPLAT_CONTAINING_RECORD(Entry, QUIC_REGISTRATION, CloseCleanupEntry);
+            CxPlatLockRelease(&MsQuicLib.RegistrationCloseCleanupLock);
+
+            CxPlatThreadWait(&Registration->CloseThread);
```
anrossi, line 774: Is this missing a call to the async close callback?

src/core/listener.c
```diff
@@ -439,6 +457,45 @@ QuicListenerIndicateDispatchEvent(
...
+        Listener);
+
+    Listener->StopCompleteThreadID = CxPlatCurThreadID();
+    (void)QuicListenerIndicateEvent(Listener, &Event);
+    Listener->StopCompleteThreadID = 0;
+
+    QuicListenerDetachSilo();
+
+    //
+    // If !Listener->NeedsCleanup, then another thread is waiting on this event
```
anrossi: This comment is referencing behavior that is no longer true/not near this comment anymore. Please update/move or remove.

src/core/listener.c
```diff
@@ -905,6 +1007,26 @@ QuicListenerParamGet(
...
+        if (Buffer == NULL) {
+            return QUIC_STATUS_INVALID_PARAMETER;
+        }
+
+        if (!Listener->Partitioned) {
+            return QUIC_STATUS_INVALID_STATE;
+        }
+
+        *BufferLength = sizeof(Listener->PartitionIndex);
+        *(uint16_t*)Buffer = Listener->PartitionIndex;
```
anrossi, line 1043: There's no guarantee this pointer is aligned so this can crash on ARM platforms. Use `memcpy` instead.

src/core/listener.c
```diff
@@ -918,18 +1040,58 @@ _IRQL_requires_max_(DISPATCH_LEVEL)
...
+            QuicListenerDetachSilo();
+        } else {
+            //
+            // Best effort mode synchronization: the non-partitioned case is
+            // also racy.
+            //
+            Listener->DosModeEnabled = DosModeEnabled;
+            if (!InterlockedFetchAndSetBoolean(&Listener->NeedsDosModeModeEvent)) {
+                QuicListenerReference(Listener);
+                QuicWorkerQueueListener(Listener->Worker, Listener);
```
anrossi, line 1083: So this is queueing the listener object directly onto the worker? Is this to avoid an allocation and the possibility of OOM failure?

src/core/listener.h
```diff
@@ -32,15 +32,30 @@ typedef struct QUIC_LISTENER {
...
 
     //
     // Indicates the listener opted in for DoS Mode event.
     //
     BOOLEAN DosModeEventsEnabled;
 
+    //
+    // Indicates a DoS Mode event indication is required.
+    //
+    BOOLEAN NeedsDosModeModeEvent;
```
anrossi, line 48: Why are these full BOOLEANs and not part of the bitfield?

src/core/listener.h
```diff
@@ -70,10 +102,15 @@ typedef struct QUIC_LISTENER {
...
     //
-    // Active reference count on the listener.
+    // Active reference count on the listener preventing a stop completion.
     //
     CXPLAT_REF_COUNT RefCount;
 
+    //
+    // Internal reference count on the listener preventing cleanup.
+    //
+    CXPLAT_REF_COUNT InternalRefCount;
```
anrossi: Two separate ref counts seem like a design that invites bugs difficult bugs if they get out of sync. Is there a way to combine them, like how connection.h has `RefTypeCount` (for debugging purposes)?

src/core/registration.c
```diff
@@ -24,6 +24,89 @@
...
+        }
+        CxPlatLockRelease(&MsQuicLib.Lock);
+    }
+}
+
+CXPLAT_THREAD_CALLBACK(RegistrationCloseWorker, Context)
+{
+    QUIC_REGISTRATION* Registration = Context;
+
+    CxPlatEventWaitForever(Registration->CloseEvent);
```
anrossi, line 82: This can spuriously wake so you'll want this to be in a loop checking to make sure it's time to clean up and going back to waiting on the event when it's not.

src/core/library.c
anrossi: oh I see how it works. The RegistrationClose thread is waiting on an event to actually close (and invoke the async callback), and this is waiting for that thread to exit.

src/core/registration.c
mtfriesen: Woah, really? Does "spuriously wake" mean the wait forever returns without CloseEvent being set?

src/core/listener.h
mtfriesen: Yeah, I still want to think more about this, and in particular, naming. The listener logically has two different types of references of two different lifetimes: references that keep the listener started (i.e., not completing a stop) and references that keep the listener valid in memory, including until ListenerClose is set.

IMO, part of the complexity in the existing code is due to trying to avoid a reference count on the second case by introducing separate boolean fields for each potential reference holder during teardown.

A `RefTypeCount` would be nice, but right now I'm not sure we have enough references to make it worthwhile.

src/core/listener.h
mtfriesen: They are unsynchronized - there is a comment where they get set, but I should add a comment here, too.

src/core/listener.c
mtfriesen: It's for simplicity (avoid creating an entire `OperQ` for one operation) and also to take advantage of the fact the booleans toggling can be elided into simply posting the most recent change as a notification.

src/core/listener.c
mtfriesen: I'm not sure I follow - MsQuic can demand callers provide aligned pointers? There are plenty of other precedents with MsQuic where the API expects valid alignment.

src/core/library.c
mtfriesen: A non-alertable wait (QUIC should not be using alertable waits?) can be interrupted by kernel APCs but the wait is not abandoned/control does not return to the caller.

src/core/connection.c
mtfriesen: Haha I have no clue how this snuck in. The NOLINT probably doesn't help, and ironically the casting was the lesser of the readability issues...

docs/api/RegistrationClose2.md
mtfriesen: I'm open to renaming this, though I used MsQuicOpen2 as the precedent.

I had been thinking about this over the weekend, though, and while it would complicate this PR even more, it might be most consistent to add a callback handler to the registration itself and continue calling RegistrationClose. We might even be able to avoid adding a new entry to the API table by using the existing `SetCallbackHandler` infrastructure.

That being said, I had thought even more that perhaps we should either add a `RegistrationShutdown` or `ExecutionContextShutdown` function that serves the purpose of cleaning up lazily-created objects asynchronously while being consistent with `XxxClose` function in MsQuic always being a terminal, blocking call.

anrossi: Oh, the RegistrationClose callback could be an optional parameter in the QUIC_REGISTRATION_CONFIG object! (append to the end of the struct).  There already is a RegistrationShutdown call, which could do what you are describing (synchronously without a callback passed in RegistrationOpen, or asynchronously with a callback in RegistrationOpen)

src/core/registration.c
anrossi: I was thinking of the APC thing you addressed in your other comment.

src/core/listener.c
anrossi: I see we don't have a consistent stance on this, so I'll drop it for this PR. We should either require callers to use aligned pointers (and document it) or expect callers to provide unaligned pointers and handle it appropriately.

src/tools/spin/spinquic.cpp
```diff
@@ -860,7 +860,7 @@ const uint32_t ParamCounts[] = {
     QUIC_PARAM_GLOBAL_LIBRARY_GIT_HASH + 1,
     0,
     QUIC_PARAM_CONFIGURATION_SCHANNEL_CREDENTIAL_ATTRIBUTE_W + 1,
-    QUIC_PARAM_LISTENER_CIBIR_ID + 1,
```
anrossi, line 863: I assume this is temporary; we should still fuzz the CIBIR ID in addition to this new parameter.

mtfriesen: Oh, I maybe misunderstood what this line does. I thought it was supposed to be the highest known parameter for the given type (listener).

src/core/listener.h
mtfriesen: I've clarified this, I hope, by renaming the refcounts and their helper functions.

docs/api/RegistrationClose2.md
mtfriesen: RegistrationClose does substantially different work from RegistrationShutdown, and while I'm not sure the work can't be migrated from Close to Shutdown, I also don't want to change that behavior for a preview feature. For now, I think the RegistrationClose2 unblocks our AFP partner without risking breaking changes to the rest of the ecosystem; we should be able to carefully refactor that preview API in the future.

src/core/library.c
mtfriesen: Good question, this was just a habit from having a common cleanup routine for both success and failure cases.

src/test/lib/ApiTest.cpp
```diff
@@ -6816,35 +6856,49 @@ TestCurThreadID()
...
 
+// TODO: remove
+void WrongThread() {}
+
+void
+TestPartitionVerifyCallback(TestPartitionCallbackContext* Context)
+{
+    if (TestCurThreadID() != Context->ExpectedThreadId) {
+        Context->ActualThreadId = TestCurThreadID();
+        WrongThread();
```
anrossi: What's the purpose of this? I see a TODO to remove it.

src/tools/spin/spinquic.cpp
anrossi: Ah, I see how this is being used. Ok, this change is fine as-is.

src/test/lib/ApiTest.cpp
mtfriesen: Good catch, thanks. It just made setting breakpoints easier.

src/core/api.c
```diff
@@ -173,39 +204,51 @@
...
+
+        if (MsQuicLib.CustomExecutions) {
+            //
+            // For custom executions, ConnectionClose completes asynchronously.
+            //
+            Connection->CloseApiContext.Completed = NULL;
+            WaitForCompletion = FALSE;
+        } else {
+            CxPlatEventInitialize(&CompletionEvent, TRUE, FALSE);
+            Connection->CloseApiContext.Completed = &CompletionEvent;
```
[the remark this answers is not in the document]
anrossi: I wonder how we suppress this warning in other places, because this certainly isn't the only place where we have a wait on a local object.  Maybe it's because of the `WaitForCompletion` block being further down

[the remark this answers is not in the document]
mtfriesen: Yeah, I dismissed it - I think it's fine to let them be raised each time and evaluated, since it is indeed risky.

anrossi (approved):

PR #5383, opened 2025-08-25
Allow CLOG to run on newer frameworks in test

mtfriesen:
## Description

CLOG by default requires exactly version 6.0 of the .NET framework, which is obsolete and not secure. Allow CLOG to run with newer version of the framework in the logging script, which is used by test infra.

## Testing

Tested locally.

## Documentation

No.

### DIFF
--- a/scripts/log.ps1
+++ b/scripts/log.ps1
@@ -376,9 +376,19 @@ function Log-Decode {
 #                     Main Execution                         #
 ##############################################################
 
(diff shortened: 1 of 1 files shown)


guhetier:
Upgrading CLOG DotNet version would also be good (IMO, might as well have both)
https://github.com/microsoft/CLOG/pull/117

PR #5392, opened 2025-08-27
Use Clang for sanitize builds on linux

### LINKED ISSUE #5400: Build and test sanitizers with clang in CI
Add pipeline sanitizer runs that build MsQuic with clang and get them to run clean.

anrossi:
## Description

Use clang for the sanitize builds on linux because clang's tooling for sanitizers is high quality. Fixes #5400 

## Testing

Existing testing should catch any issues.

## Documentation

N/A

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -189,7 +189,7 @@ jobs:
       uses: actions/download-artifact@634f93cb2916e3fdff6788551b99b062d0335ce0
       if: matrix.vec.plat == 'linux'
       with:
-        name: ${{ matrix.vec.config }}-${{ matrix.vec.plat }}-${{ matrix.vec.os }}-${{ matrix.vec.arch }}-${{ matrix.vec.tls }}${{ matrix.vec.systemcrypto }}${{ matrix.vec.sanitize}}${{ matrix.vec.sanitize }}${{ matrix.vec.xdp }}${{ matrix.vec.iouring }}${{ matrix.vec.build }}
+        name: ${{ matrix.vec.config }}-${{ matrix.vec.plat }}-${{ matrix.vec.os }}-${{ matrix.vec.arch }}-${{ matrix.vec.tls }}${{ matrix.vec.systemcrypto }}${{ matrix.vec.clang}}${{ matrix.vec.sanitize }}${{ matrix.vec.xdp }}${{ matrix.vec.iouring }}${{ matrix.vec.build }}
         path: artifacts
     - name: Fix permissions for Unix
       if: matrix.vec.plat == 'linux' || matrix.vec.plat == 'macos'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -99,17 +99,23 @@ jobs:
       matrix:
         vec: [
           { config: "Debug", plat: "linux", os: "ubuntu-22.04", arch: "x64", tls: "quictls", sanitize: "-Sanitize", build: "-Test", clang: "-Clang" },
+          { config: "Debug", plat: "linux", os: "ubuntu-22.04", arch: "x64", tls: "quictls", sanitize: "-Sanitize", build: "-Test" },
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -125,6 +125,7 @@ jobs:
       iouring: ${{ matrix.vec.iouring }}
       ref: ${{ inputs.ref || '' }}
       repo: ${{ github.repository }}
+      clang: ${{ matrix.vec.clang }}
```
mtfriesen, line 134: at risk of further bloating our test matrix, but why not use both clang and whatever we were using before (gcc?) for maximum coverage?

anrossi: I can do that; the problem I don't know is if the artifacts store can tell the difference between the clang-build sanitize binaries and the gcc-build sanitize binaries. I decided to try doing a clang-only run here to see if it works and then to bloat the matrix with both.

mtfriesen: yeah, you often need to add yet another disambiguator to the artifact path.

another reason, other than maximizing coverage, is that we ideally want msquic to be usable by other projects using common tools, which still includes gcc afaik.

src/platform/toeplitz.c
```diff
@@ -55,6 +55,7 @@
 // Initializes the state required for a Toeplitz hash computation. We
 // maintain per-nibble lookup tables, and we initialize them here.
 //
+QUIC_NO_SANITIZE("unsigned-integer-overflow")
```
guhetier: That is scary to read. Is this because of an over-zealous check that cause a false positive? Or is it an actual runtime situation?
If we disable the check, I think the scope should be as small as possible + a comment should be mandatory to clearly explain what can overflow and why it is fine in that case.

src/platform/datapath_linux.c
```diff
@@ -202,13 +202,13 @@ CxPlatSocketConfigureRss(
     int Result = 0;
 
     struct sock_filter BpfCode[] = {
-        {BPF_LD | BPF_W | BPF_ABS, 0, 0, SKF_AD_OFF | SKF_AD_CPU}, // Load CPU number
+        {(uint32_t)(BPF_LD | BPF_W | BPF_ABS, 0, 0, SKF_AD_OFF | SKF_AD_CPU)}, // Load CPU number
```
guhetier: Sooo, I know that `,` is an operator and that this boils down to a single value you can cast, but I doubt that's actually doing what you want it to do.

anrossi: The error I get is
>  runtime error: implicit conversion from type 'int' of value -4060 (32-bit, signed) to type '__u32' (aka 'unsigned int') changed the value to 4294963236 (32-bit, unsigned)

I suppose I could just suppress it, but I'd like to understand better what is going on; I thought this was just ORing a number of flags together, but now that I notice the comma operation, I'm lost as to what's going on.

src/platform/toeplitz.c
anrossi: this is part of the toeplitz hash, which expects the overflow. I could instead make it a comment above the specific lines where the overflow occurs with explanation there instead of ignoring overflow for the entire function.

guhetier: I think that would be worth it, both for safety and as a precedent.

src/platform/datapath_linux.c
guhetier: I think that what you wrote is basically equivalent to `{(uint32_t)(SKF_AD_OFF | SKF_AD_CPU), 0, 0, 0}`.
The issue is likely that a set of flags is seen as signed and set the top bit, raising alarms about the cast.

`{(uint32_t)(BPF_LD | BPF_W | BPF_ABS), 0, 0, (uint32_t)(SKF_AD_OFF | SKF_AD_CPU)}` is probably what you are looking for.

src/platform/toeplitz.c
anrossi: I found a way to fix this without just suppressing the warning.

Issue #5356, opened 2025-08-14
[CI-TEST] Misc.StreamAppProvidedBuffers crash with Iouring


anrossi:
```[08/14/2025 19:30:59] Process had nonzero exit code: 134
Warning: [08/14/2025 19:30:59] Core file(s) generated

warning: Can't open file anon_inode:[io_uring] which was expanded to anon_inode:[io_uring] during file-backed mapping note processing

warning: Can't open file /dev/shm/shm-ust-consumer-6082 (deleted) during file-backed mapping note processing

warning: Can't open file /dev/shm/shm-event-notifier-error-accounting-6066 (deleted) during file-backed mapping note processing

warning: Can't open file /dev/shm/shm-event-notifier-error-accounting-5157 (deleted) during file-backed mapping note processing

warning: could not find '.gnu_debugaltlink' file for /lib/x86_64-linux-gnu/liburing-ffi.so.2

warning: could not find '.gnu_debugaltlink' file for /lib/x86_64-linux-gnu/liblttng-ust-tracepoint.so.1

warning: could not find '.gnu_debugaltlink' file for /lib/x86_64-linux-gnu/liblttng-ust.so.1

warning: 44	./nptl/pthread_kill.c: No such file or directory
Undefined command: "".  Try "help".
Undefined command: "".  Try "help".
==================================================================================
 msquictest.18473.1755199858.core
==================================================================================
Error: [08/14/2025 19:31:00] Misc.StreamAppProvidedBuffers failed (in 2.4548807 sec):
Note: Google Test filter = Misc.StreamAppProvidedBuffers
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
Initializing for User Mode tests
[----------] 1 test from Misc
[ RUN      ] Misc.StreamAppProvidedBuffers

[...]

PR #5395, opened 2025-08-27
QTIP: Configure XDP rules properly

ProjectsByJackHe:
## Description

As part of the change to have listeners support both QTIP and QUIC/UDP, we configure the XDP rules to listen to both 
UDP and TCP packets bound to the listener port. 

But the problem is, we always set these rules, even in cases where XDP is enabled but not QTIP. What that means is we don't create an auxiliary TCP socket, so other applications bound to the same TCP port as the msquic listener will have their traffic stolen. 

## Testing

CI

## Documentation

N/A

### DIFF
--- a/src/platform/datapath_raw_xdp_win.c
+++ b/src/platform/datapath_raw_xdp_win.c
@@ -1452,6 +1452,7 @@ CxPlatDpRawPlumbRulesOnSocket(
 
             memcpy(Rules[0].Pattern.QuicFlow.CidData, Socket->CibirId, Socket->CibirIdLength);
             memcpy(Rules[1].Pattern.QuicFlow.CidData, Socket->CibirId, Socket->CibirIdLength);
+            RulesSize = 2;
 
             if (Socket->ReserveAuxTcpSock) {
                 Rules[2].Match = XDP_MATCH_TCP_QUIC_FLOW_SRC_CID;
@@ -1479,8 +1480,6 @@ CxPlatDpRawPlumbRulesOnSocket(
                 Rules[4].Redirect.TargetType = XDP_REDIRECT_TARGET_TYPE_XSK;
                 Rules[4].Redirect.Target = NULL;
                 RulesSize = 5;
-            } else {
-                RulesSize = 2;
             }
             CXPLAT_DBG_ASSERT(RulesSize <= RTL_NUMBER_OF(Rules));
         } else {
@@ -1489,6 +1488,7 @@ CxPlatDpRawPlumbRulesOnSocket(
             Rules[0].Action = XDP_PROGRAM_ACTION_REDIRECT;
             Rules[0].Redirect.TargetType = XDP_REDIRECT_TARGET_TYPE_XSK;
             Rules[0].Redirect.Target = NULL;
+            RulesSize = 1;
 
             if (Socket->ReserveAuxTcpSock) {
                 Rules[1].Match = XDP_MATCH_TCP_DST;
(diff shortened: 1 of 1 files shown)


src/platform/datapath_raw_xdp_win.c
```diff
@@ -1450,52 +1450,58 @@ CxPlatDpRawPlumbRulesOnSocket(
...
-            memcpy(Rules[3].Pattern.QuicFlow.CidData, Socket->CibirId, Socket->CibirIdLength);
-
-            Rules[4].Match = XDP_MATCH_TCP_CONTROL_DST;
-            Rules[4].Pattern.Port = Socket->LocalAddress.Ipv4.sin_port;
-            Rules[4].Action = XDP_PROGRAM_ACTION_REDIRECT;
-            Rules[4].Redirect.TargetType = XDP_REDIRECT_TARGET_TYPE_XSK;
-            Rules[4].Redirect.Target = NULL;
-            RulesSize = 5;
 
+            if (Socket->ReserveAuxTcpSock) {
```
guhetier: Can you move the `memcpy` with their respective rule blocks, and merge the two `if` blocks together? It'll be more readable

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1450,52 +1450,58 @@ CxPlatDpRawPlumbRulesOnSocket(
...
+                memcpy(Rules[2].Pattern.QuicFlow.CidData, Socket->CibirId, Socket->CibirIdLength);
+                memcpy(Rules[3].Pattern.QuicFlow.CidData, Socket->CibirId, Socket->CibirIdLength);
+                Rules[4].Match = XDP_MATCH_TCP_CONTROL_DST;
+                Rules[4].Pattern.Port = Socket->LocalAddress.Ipv4.sin_port;
+                Rules[4].Action = XDP_PROGRAM_ACTION_REDIRECT;
+                Rules[4].Redirect.TargetType = XDP_REDIRECT_TARGET_TYPE_XSK;
+                Rules[4].Redirect.Target = NULL;
+                RulesSize = 5;
+            } else {
+                RulesSize = 2;
```
guhetier: nit: Maybe a personal taste, but I would put the size to 2 right below the rules 0 and 1 are set, and overwrite it to 5 in the if block.

It reduces the risk the count ends up out of sync in the future

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1450,52 +1450,55 @@ CxPlatDpRawPlumbRulesOnSocket(
...
+
+                memcpy(Rules[2].Pattern.QuicFlow.CidData, Socket->CibirId, Socket->CibirIdLength);
+                memcpy(Rules[3].Pattern.QuicFlow.CidData, Socket->CibirId, Socket->CibirIdLength);
+
+                Rules[4].Match = XDP_MATCH_TCP_CONTROL_DST;
+                Rules[4].Pattern.Port = Socket->LocalAddress.Ipv4.sin_port;
+                Rules[4].Action = XDP_PROGRAM_ACTION_REDIRECT;
+                Rules[4].Redirect.TargetType = XDP_REDIRECT_TARGET_TYPE_XSK;
+                Rules[4].Redirect.Target = NULL;
+                RulesSize = 5;
```
mtfriesen: IMO it would make this code more readable/maintainable to do something like:

```
RulesSize = 0;
...
// Add another rule.
Rules[RuleSize].Match = ...
Rules[RuleSize].Pattern = ...
RuleSize++;
...
```

guhetier:
CI failure #5356

src/platform/datapath_raw_xdp_win.c
ProjectsByJackHe: Sure, we can do that too.

PR #5401, opened 2025-08-28
Always use the latest base image for build dockers

csujedihy:
## Description

Always use the latest base image for build dockers

## Testing

CI

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -35,8 +35,7 @@ updates:
 
   - package-ecosystem: "docker"
     directories:
-      - "/.docker/ubuntu-22.04"
-      - "/.docker/ubuntu-24.04"
+      - "/.docker/*"
     schedule:
       interval: "weekly"
       day: "saturday"


mtfriesen:
Is it better to use dependabot to automatically update these? I see code warnings for the unpinned dependency.

csujedihy:
> Is it better to use dependabot to automatically update these? I see code warnings for the unpinned dependency.

I was puzzled by dependabot not updating the dockerfile so I made this PR. But, I just found a way to look at the dependabot logs and figured out why. There's a stale path in the config file.

mtfriesen (approved):

.github/dependabot.yml
```diff
@@ -35,8 +35,8 @@ updates:
 
   - package-ecosystem: "docker"
     directories:
-      - "/.docker/ubuntu-20.04"
       - "/.docker/ubuntu-22.04"
+      - "/.docker/ubuntu-24.04"
```
mtfriesen: > The directories key supports globbing and the wildcard character *. These features are not supported by the directory key.

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#directories-or-directory--

Can we prevent this in the future by just doing `- "/.docker/*`?

csujedihy: good find

PR #5415, opened 2025-09-02
Removed unnecessary if-else in TimeDiff functions

penglei0:
## Description

- 1. in any cases, `T2 + (0xFFFFFFFF - T1) + 1` is equal to `T2 - T1`;
- 2. subtraction (`T2 - T1`) can handle wraparound automatically in the ring 2^32;
- 3. removed if-else in TimeDiff function improves the performance;

## Testing

No new testing required.

## Documentation

No docs need to be updated.

### DIFF
--- a/src/inc/quic_platform_posix.h
+++ b/src/inc/quic_platform_posix.h
@@ -745,11 +745,8 @@ CxPlatTimeDiff32(
     _In_ uint32_t T2      // Second time measured
     )
 {
-    if (T2 > T1) {
-        return T2 - T1;
-    } else { // Wrap around case.
-        return T2 + (0xFFFFFFFF - T1) + 1;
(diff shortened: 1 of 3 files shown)


mtfriesen:
I wonder - do we really even need this 32-bit function? Skimming the few locations it is used, using 64-bit time would be simpler.

penglei0:
> I wonder - do we really even need this 32-bit function? Skimming the few locations it is used, using 64-bit time would be simpler.

The places where uint32_t is used for time MS are initialized with `CxPlatTimeMs32`; and `CxPlatTimeMs32` is just a wrapper of `CxPlatTimeMs64`; so, it's possible to remove CxPlatTimeDiff32 with changes only in a few lines.

PR #5419, opened 2025-09-04
Fix do the 1way delay caculation only when feature is enabled

penglei0:
## Description

`QuicConnUpdateRtt` tries to calculate `Path`'s OneWayDelay even when this feature is in disabled state; it has the following minor impacts:
- Misleading OneWayDelay value in trace log (see log details below)
- Unnecessary calculation on an invalid Packet's  `SendTimestamp` (SendTimestamp is fixed to 0 when OneWayDelay is disabled)

Expected behavior:
- Skip OneWayDelay calculation when it's disabled;
- in trace log，1Way should be shown as 0  when it's disabled.

Before fix (with OneWayDelay disabled):
```log
  685,51: [0][71.7b][10:01:07.062688][conn][0x7daca4065da0] Updated Rtt=143.229 ms, Var=71.614 1Way=71.614 ms
  1866,51: [0][71.7b][10:01:07.204465][conn][0x7daca4065da0] Updated Rtt=142.849 ms, Var=54.469 1Way=70.97 ms
  2006,51: [0][71.7b][10:01:07.204575][conn][0x7daca4065da0] Updated Rtt=142.525 ms, Var=41.499 1Way=70.91 ms
  2528,51: [0][71.7b][10:01:07.205713][conn][0x7daca4065da0] Updated Rtt=142.223 ms, Var=31.727 1Way=70.57 ms
  2723,51: [2][71.7b][10:01:07.216949][conn][0x7daca4065da0] Updated Rtt=141.980 ms, Var=24.279 1Way=69.921 ms
  3749,51: [6][71.7b][10:01:07.344662][conn][0x7daca4065da0] Updated Rtt=141.746 ms, Var=18.676 1Way=70.56 ms
  4234,51: [6][71.7b][10:01:07.345034][conn][0x7daca4065da0] Updated Rtt=141.572 ms, Var=14.353 1Way=70.39 ms
  4325,51: [6][71.7b][10:01:07.345904][conn][0x7daca4065da0] Updated Rtt=141.391 ms, Var=11.125 1Way=69.887 ms
  4741,51: [6][71.7b][10:01:07.347169][conn][0x7daca4065da0] Updated Rtt=141.258 ms, Var=8.609 1Way=69.619 ms
  5460,51: [6][71.7b][10:01:07.357118][conn][0x7daca4065da0] Updated Rtt=141.117 ms, Var=6.737 1Way=68.117 ms
  6117,51: [6][71.7b][10:01:07.484959][conn][0x7daca4065da0] Updated Rtt=141.1 ms, Var=5.283 1Way=50.832 ms
  6627,51: [6][71.7b][10:01:07.485265][conn][0x7daca4065da0] Updated Rtt=140.906 ms, Var=4.151 1Way=35.675 ms
  7291,51: [6][71.7b][10:01:07.487110][conn][0x7daca4065da0] Updated Rtt=140.808 ms, Var=3.309 1Way=22.166 ms
  7631,51: [6][71.7b][10:01:07.487251][conn][0x7daca4065da0] Updated Rtt=140.724 ms, Var=2.649 1Way=10.330 ms
  8091,51: [6][71.7b][10:01:07.487549][conn][0x7daca4065da0] Updated Rtt=140.675 ms, Var=2.83 1Way=3917010173.912 ms
  8123,51: [6][71.7b][10:01:07.487568][conn][0x7daca4065da0] Updated Rtt=140.627 ms, Var=1.657 1Way=206158421.87 ms
  8158,51: [6][71.7b][10:01:07.487576][conn][0x7daca4065da0] Updated Rtt=140.577 ms, Var=1.342 1Way=3401614081.356 ms
  8199,51: [6][71.7b][10:01:07.487584][conn][0x7daca4065da0] Updated Rtt=140.524 ms, Var=1.112 1Way=2439541400.81 ms
  8229,51: [6][71.7b][10:01:07.487591][conn][0x7daca4065da0] Updated Rtt=140.474 ms, Var=0.933 1Way=3745211451.961 ms
  ```

## Testing

Any stream/Datagram sending/receiving tests can repeat this;

## Documentation

No docs need to be updated.

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -826,7 +826,7 @@ QuicConnUpdateRtt(
         Path->SmoothedRtt = (7 * Path->SmoothedRtt + LatestRtt) / 8;
     }
 
-    if (OurSendTimestamp != UINT64_MAX) {
+    if (Connection->Settings.OneWayDelayEnabled && OurSendTimestamp != UINT64_MAX) {
         if (Connection->Stats.Timing.PhaseShift == 0 || NewMinRtt) {
             Connection->Stats.Timing.PhaseShift =
                 (int64_t)PeerSendTimestamp - (int64_t)OurSendTimestamp - (int64_t)LatestRtt / 2;

### CODE REVIEW

**Comment by ProjectsByJackHe** [comment]
Hey thanks for this contribution. Overall LGTM. Let's wait for the CI to complete and it should be good.

File: src/core/connection.c [hunk 1]
```diff
@@ -826,7 +826,7 @@ QuicConnUpdateRtt(
         Path->SmoothedRtt = (7 * Path->SmoothedRtt + LatestRtt) / 8;
     }
 
-    if (OurSendTimestamp != UINT64_MAX) {
+    if (Connection->Settings.OneWayDelayEnabled && OurSendTimestamp != UINT64_MAX) {
```
nibanks: Wasn't `OurSendTimestamp ` meant to stay as `UINT64_MAX` when the feature wasn't in use?

penglei0: `OurSendTimestamp` is calculated by `LargestAckedPacket->SentTime - Connection->Stats.Timing.Start` which apparently is not UINT64_MAX.

guhetier: Is there a point keeping the `OurSendTimestamp != UINT64_MAX` check in this case?

I wonder if it was meant to be `PeerSendTimestamp != UINT64_MAX`, as the peer timestamp is initialized to `UINT64_MAX` for each packet, and set to the peer value only when a `QUIC_FRAME_TIMESTAMP` is received. 

Otherwise, it seems to me that if a packet is received without the timestamp frame, we would use `UINT64_MAX` as the peer timestamp and get the wrong one way delay.

penglei0: Yes, I think we should remove the `OurSendTimestamp != UINT64_MAX` in this case.

As for the initial value of `PeerSendTimestamp`, it's set to UINT64_MAX in `QuicBindingReceive` at the beginning, but later it has been reset to 0 in `QuicBindingPreprocessPacket`.

So, I think we can do the checks for the enabled status of valid 1w-delay with the following code:

```cpp
  if (PeerSendTimestamp != UINT64_MAX && PeerSendTimestamp != 0) {
    // ...
  }
```

What do you think of this?
@guhetier

guhetier: Hum, I am worried about the `!= 0` test, `0` could be a valid value, especially for the initial peer timestamp (chances are their epoch is the time of the first send - I am not too familiar with this feature though).

The zeroing in `QuicBindingPreprocessPacket` seems suspicious to me.
`QuicBindingReceive` initialize most of the packet members, only to have them erased and re-initialized right after?

My first thought is that the zeroing should be removed and the proper initialization done in `QuicBindingReceive`, but I haven't looked at the details and whether there could be consequences.

penglei0: > I am worried about the != 0 test, 0 could be a valid value
```cpp
    if (Builder->Connection->State.TimestampSendNegotiated &&
        Builder->EncryptLevel == QUIC_ENCRYPT_LEVEL_1_RTT) {
        QUIC_TIMESTAMP_EX Frame = { Timestamp - Builder->Connection->Stats.Timing.Start };
```
true, 0 is a valid value for `PeerSendTimestamp`.


> My first thought is that the zeroing should be removed and the proper initialization done in QuicBindingReceive..

https://github.com/microsoft/msquic/blob/4a6febc80e6b4464563c13080fe7e1b95621fbca/src/core/binding.c#L1645-L1656

Perhaps, the above redundant code should be removed but keep the minimal fields that test datapath hooks rely on (at least AvailBuffer and AvailBufferLength ), since **QuicBindingPreprocessPacket intentionally zeroes every field from PacketNumber onward**.

https://github.com/microsoft/msquic/blob/4a6febc80e6b4464563c13080fe7e1b95621fbca/src/core/binding.c#L1139-L1141

In summary, another possible fix could be like this:

```patch
diff --git a/src/core/binding.c b/src/core/binding.c
index 443a6520..9006bbfe 100644
--- a/src/core/binding.c
+++ b/src/core/binding.c
@@ -1137,6 +1137,7 @@ QuicBindingPreprocessPacket(
         sizeof(QUIC_RX_PACKET) - offsetof(QUIC_RX_PACKET, PacketNumber));
     Packet->AvailBuffer = Packet->Buffer;
     Packet->AvailBufferLength = Packet->BufferLength;
+    Packet->SendTimestamp = UINT64_MAX;

     *ReleaseDatagram = TRUE;

@@ -1638,18 +1639,9 @@ QuicBindingReceive(
         QUIC_RX_PACKET* Packet = (QUIC_RX_PACKET*)Datagram;
         Packet->PacketId =
             PartitionShifted | InterlockedIncrement64((int64_t*)&Partition->ReceivePacketId);
-        Packet->PacketNumber = 0;
-        Packet->SendTimestamp = UINT64_MAX;
+        // Preserve buffer pointers/length so test hooks see valid values.
         Packet->AvailBuffer = Datagram->Buffer;
-        Packet->DestCid = NULL;
-        Packet->SourceCid = NULL;
         Packet->AvailBufferLength = Datagram->BufferLength;
-        Packet->HeaderLength = 0;
-        Packet->PayloadLength = 0;
-        Packet->DestCidLen = 0;
-        Packet->SourceCidLen = 0;
-        Packet->KeyType = QUIC_PACKET_KEY_INITIAL;
-        Packet->Flags = 0;

         CXPLAT_DBG_ASSERT(Packet->PacketId != 0);
         QuicTraceEvent(
diff --git a/src/core/connection.c b/src/core/connection.c
index 8648dfe4..896e0035 100644
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -826,7 +826,7 @@ QuicConnUpdateRtt(
         Path->SmoothedRtt = (7 * Path->SmoothedRtt + LatestRtt) / 8;
     }

-    if (OurSendTimestamp != UINT64_MAX) {
+    if (PeerSendTimestamp != UINT64_MAX) {
         if (Connection->Stats.Timing.PhaseShift == 0 || NewMinRtt) {
             Connection->Stats.Timing.PhaseShift =
```

guhetier: Doesn't the test hook only take the `CXPLAT_RECV_DATA` and not the `QUIC_RX_PACKET`? I am not sure I see where the field you keep initialized in the diff are needed.

Otherwise, this approach seems good to me.
If you don't mind adding a comment to clarify `SendTimestamp` must be initialized to `UINT64_MAX` when it isn't received?

PR #5428, opened 2025-09-06
Refactor confusing datapath code.

ProjectsByJackHe:
## Description

While debugging other issues, it has become apparent that some implicit assumptions were causing a lot of confusion. 

Let's clear all this up here, and refactor some code while we're at it.

## Testing

CI

## Documentation

N/A

### DIFF
--- a/src/platform/datapath_xplat.c
+++ b/src/platform/datapath_xplat.c
@@ -134,7 +134,6 @@ CxPlatSocketCreateUdp(
     // sockets the same as client sockets, in that they bind to some random free UDP port.
     // However, what's free in UDP may not be free in TCP. So we loop until we find a free port.
     //
-
     for (uint32_t TryCount = 0; TryCount < 1000; TryCount++) {
         Status =
             SocketCreateUdp(


src/platform/datapath_raw_win.c
```diff
@@ -123,34 +123,45 @@ RawSocketCreateUdp(
         memcpy(Socket->CibirId, Config->CibirId, Config->CibirIdLength);
     }
 
+    //
```
guhetier: Would this be more discoverable close to the `CXPLAT_UDP_CONFIG` declaration rather than in one specific datapath?

src/platform/datapath_raw_win.c
```diff
@@ -123,34 +123,45 @@ RawSocketCreateUdp(
...
+
     if (Config->RemoteAddress) {
+        //
+        // This CxPlatSocket is part of a client connection.
+        //
         CXPLAT_FRE_ASSERT(!QuicAddrIsWildCard(Config->RemoteAddress));  // No wildcard remote addresses allowed.
         if (Socket->ReserveAuxTcpSock) {
+            //
+            // TODO: Not sure what's special about QTIP. We always set Socket->RemoteAddress in SocketCreateUdp()
```
guhetier: Please don't add TODOs. Create a task if there is something to be done.

src/platform/datapath_raw_win.c
```diff
@@ -123,34 +123,45 @@ RawSocketCreateUdp(
...
+            //
             Socket->RemoteAddress = *Config->RemoteAddress;
+            if (Socket->LocalAddress != NULL) {
+                CXPLAT_FRE_ASSERT(!QuicAddrIsWildCard(Config->LocalAddress));
+                Socket->LocalAddress = *Config->LocalAddress;
+            } else {
+                QuicAddrSetFamily(&Socket->LocalAddress, QUIC_ADDRESS_FAMILY_INET6);
+            }
         }
         Socket->Connected = TRUE;
```
guhetier, line 139: note: `Socket->Connected == TRUE` means this is a "client" socket.
In this context, "connected" means bound to a specific remote address, and while not "connected" means it listen for incoming traffic from any peers.

src/platform/datapath_raw_win.c
```diff
@@ -123,34 +123,45 @@ RawSocketCreateUdp(
...
         if (Socket->ReserveAuxTcpSock) {
+            //
+            // TODO: Not sure what's special about QTIP. We always set Socket->RemoteAddress in SocketCreateUdp()
+            //       earlier and why don't we always set Socket->LocalAddress if LocalAddress is specified
+            //       regardless of if we're using QTIP or not? Also if LocalAddress was not specified why do we
+            //       need to set Socket->LocalAddress to IPv6 if we're using QTIP?
+            //
             Socket->RemoteAddress = *Config->RemoteAddress;
+            if (Socket->LocalAddress != NULL) {
+                CXPLAT_FRE_ASSERT(!QuicAddrIsWildCard(Config->LocalAddress));
```
guhetier: Why does this deserve a FRE assert?

src/platform/datapath_raw_win.c
```diff
@@ -123,34 +123,45 @@ RawSocketCreateUdp(
...
-        }
-        if (QuicAddrIsWildCard(Config->LocalAddress)) {
-            if (!Socket->Connected) {
-                Socket->Wildcard = TRUE;
-            }
-        } else if (!Socket->Connected) {
-            // Assumes only connected sockets fully specify local address
+    } else {
+        //
+        // This CxPlatSocket is part of a server listener.
```
guhetier, line 142: I would extend this comment with "It is assumed listeners always listen on the wildcard address".

Do we even need that assumption? I am curious abot what would go wrong if we simply accepted an interface address here (which would not be exercised by the core, but is there a reason for the datapath to reject it?)

ProjectsByJackHe: So with XDP/QTIP, the current logic is that if Socket->WildCard == TRUE, then we will start snooping for BOTH UDP and TCP packets using XDP. Obviously, there would be a serious problem if a client connection was doing that, or a server listener was NOT doing that.

src/platform/datapath_raw_win.c
ProjectsByJackHe: Fixed in the latest iteration and modified my comments regarding assumptions made by MsQuic core code.

src/platform/datapath_raw_win.c
ProjectsByJackHe: I guess this isn't really a task, rather a Q&A. It'd be great to figure out an answer to this question (maybe in this comment thread! If not, we can take it offline)

src/platform/datapath_raw_win.c
ProjectsByJackHe: I think here is fine. There are 2 places (listener / connection) that passes down the udp/datapath config, having the comment here would explain the asserts.

src/platform/datapath_raw_win.c
```diff
@@ -123,34 +123,45 @@ RawSocketCreateUdp(
         memcpy(Socket->CibirId, Config->CibirId, Config->CibirIdLength);
     }
 
+    //
+    // Key assumptions by MsQuic core code:
```
mtfriesen: If these assumptions apply to MsQuic core, this comment should be somewhere common to all data paths, not just winuser.

src/platform/datapath_raw_win.c
```diff
@@ -123,34 +123,45 @@ RawSocketCreateUdp(
...
+            //       earlier and why don't we always set Socket->LocalAddress if LocalAddress is specified
+            //       regardless of if we're using QTIP or not? Also if LocalAddress was not specified why do we
+            //       need to set Socket->LocalAddress to IPv6 if we're using QTIP?
+            //
             Socket->RemoteAddress = *Config->RemoteAddress;
+            if (Config->LocalAddress != NULL) {
+                CXPLAT_FRE_ASSERT(!QuicAddrIsWildCard(Config->LocalAddress));
+                Socket->LocalAddress = *Config->LocalAddress;
+            } else {
+                QuicAddrSetFamily(&Socket->LocalAddress, QUIC_ADDRESS_FAMILY_INET6);
```
mtfriesen, line 136: I don't think we should hardcore IPv6 here unless we understand why it's necessary. Otherwise, the datapath is even more confusing, IMO, since it is now doing even more but with no clearer reasons.

mtfriesen: Oh, this is moved from a block below, so it is not more confusing, just equally confusing 😄

src/platform/datapath_raw_win.c
ProjectsByJackHe: Yes, I agree. I should also change this comment to say: "Key assumptions **made** by MsQuic core code"

src/platform/datapath_raw_win.c
ProjectsByJackHe: Yeah, there must be a reason for having this. Too many features crammed into spaghetti often leads to side effects like this.

src/platform/datapath_raw_win.c
ProjectsByJackHe: Feel free to reopen this thread if you have more questions.

src/platform/datapath_raw_win.c
ProjectsByJackHe: Moved my comment to datapath_xplat.

src/platform/datapath_xplat.c
```diff
@@ -134,7 +134,15 @@ CxPlatSocketCreateUdp(
     // sockets the same as client sockets, in that they bind to some random free UDP port.
     // However, what's free in UDP may not be free in TCP. So we loop until we find a free port.
     //
+
     for (uint32_t TryCount = 0; TryCount < 1000; TryCount++) {
+        //
+        // When examining datapath code that parses CXPLAT_UDP_CONFIG, keep in mind the assumptions made by MsQuic core code about the config:
+        //      - A non-NULL remote address specified by the config means this CxPlat socket MUST be part of a client connection.
```
guhetier: It is nitpicky, but for clarity, I would turn these sentences around:

- "A server listener must specify have a NULL remote address"
- "A client connection can specify anything for the local address:  NULL, wildcard, any <ip/port tupple>"
    - Bonus point if you add a sentence to say what we do with NULL / wildcard

... which makes me realize that the first and third point are actually redundant.

Can a client connection have a null / wildcard address? I would add that info too.
- "A client connection must have a fully specified remote addresss (ip + port)"

src/platform/datapath_xplat.c
```diff
@@ -134,7 +134,15 @@ CxPlatSocketCreateUdp(
     // sockets the same as client sockets, in that they bind to some random free UDP port.
     // However, what's free in UDP may not be free in TCP. So we loop until we find a free port.
     //
+
     for (uint32_t TryCount = 0; TryCount < 1000; TryCount++) {
+        //
```
guhetier: nit: I would consider adding this on the definition of `CXPLAT_UDP_CONFIG` instead - that is personally the place I would think about looking at if I was trying to understand how to deal with the config.

src/platform/datapath_xplat.c
guhetier: Another tiny nitpick: a socket is not really "part" of a connection. Maybe "used by" or "bound" is better.... the issue being that there is no 1-1 mapping between sockets and connections.

src/platform/datapath_raw_win.c
```diff
@@ -124,35 +124,39 @@ RawSocketCreateUdp(
...
-            QuicAddrSetFamily(&Socket->LocalAddress, QUIC_ADDRESS_FAMILY_INET6);
-        }
-        if (!Socket->Connected) {
-            Socket->Wildcard = TRUE;
-        }
+        Socket->Wildcard = TRUE;
     }
 
+    //
+    // Note here that the socket COULD have local address be a wildcard AND Socket->Wildcard == FALSE.
```
guhetier, line 156: I would consider fixing that name, that is extremely confusing! Or maybe removing it? If it is a XOR with `Connected`, we might not need both.

src/platform/datapath_xplat.c
ProjectsByJackHe: addressed offline.

src/platform/datapath_raw_win.c
ProjectsByJackHe: addressed offline.

src/platform/datapath_xplat.c
ProjectsByJackHe: addressed offline.

src/platform/datapath_xplat.c
```diff
@@ -134,6 +134,7 @@ CxPlatSocketCreateUdp(
     // sockets the same as client sockets, in that they bind to some random free UDP port.
     // However, what's free in UDP may not be free in TCP. So we loop until we find a free port.
     //
+
```
guhetier:
```suggestion
```

Spurious new line

src/platform/datapath_raw_win.c
guhetier: Offline discussion: Not for this PR, but let's make sure we follow up on this and rename / delete `Wildcard`, as this is very confusing.

ProjectsByJackHe:
> An extra line jump sneaked in

Fixed, will need a re-approval.

PR #5443, opened 2025-09-12
CI: validate Rust crate size under 10MB in cargo workflow

### LINKED ISSUE #5243: Add a Rust crate size validation in the CI
> We should consider adding a validation during the CI that the compressed crate size stays below 10MB. 

 _Originally posted by @guhetier in [#5175](https://github.com/microsoft/msquic/issues/5175#issuecomment-3059730165)_

crates.io only accepts uploads smaller than 10MB, and we are dangerously close to the limit. We need to be aware of our crate being too big before we snap a release so we can prune it first, or we can't publish it.

cnaples79:
## Summary
Add a CI check to ensure the Rust crate archive stays under the crates.io 10MB limit.

## Motivation
As noted in #5243, crates.io rejects packages over 10MB. We’re close to the limit, and without CI validation we might only discover this at release time.

## Changes
- Workflow: `.github/workflows/cargo.yml`
  - After `cargo publish --dry-run`, run `cargo package` to produce the `.crate` tarball under `target/package`.
  - On `ubuntu-latest`, compute the `.crate` size and fail the job if it exceeds 10MB.

## Notes
- Limited to `ubuntu-latest` to avoid shell differences across OS runners; behavior is OS-independent because crates package source files.
- Keeps the check in the existing cargo workflow so it runs on PRs and pushes.

Fixes #5243

Update: Per maintainer feedback, temporarily exempt OpenSSL TLS builds from the crate size check. The validation still runs on ubuntu-latest for non-OpenSSL TLS configurations (e.g., quictls). Once package contents are reduced to consistently stay <10MiB, we can re-enable the check for OpenSSL as well.

### DIFF
--- a/.github/workflows/cargo.yml
+++ b/.github/workflows/cargo.yml
@@ -103,7 +103,9 @@ jobs:
     - name: Cargo Publish (dry run)
       run: cargo publish --dry-run --allow-dirty --no-verify
     - name: Validate crate size (< 10MB)
-      if: matrix.os == 'ubuntu-latest'
+      # Temporary: skip size check for OpenSSL TLS builds which currently exceed 10MiB.
+      # We still validate for other TLS configs (e.g., quictls) on ubuntu-latest.
+      if: matrix.os == 'ubuntu-latest' && matrix.vec.tls != 'openssl'
       shell: bash
       run: |
         set -euo pipefail


cnaples79:

cnaples79:
@mtfriesen Are there any changes you would like me to make to this PR to help fix #5243 ?

mtfriesen:
> @mtfriesen Are there any changes you would like me to make to this PR to help fix #5243 ?

Hmm, I'm not sure whether we should try to get all the packages under 10MiB (this seems like an ideal but perhaps slow outcome) or set this CI to fail only for the baseline package configurations currently under 10MiB. @guhetier, thoughts?

From what I can see in the last CI run, only ubuntu-latest with openssl TLS was over the limit; quictls builds were just under. @cnaples79, if you want to exempt openssl from this check, that might be the best short term option.

cnaples79:
@mtfriesen Sure, I can update that check for the PR.

cnaples79:
I've updated the PR with the CI exclusion and added a note!

mtfriesen:
> I've updated the PR with the CI exclusion and added a note!

Thanks! LGTM now with the Cargo CI passing, and we can chip away at the openssl case later.

mtfriesen (approved):

mtfriesen:
I just realized the CI won't pass until #5453 fixing MacOS builds is merged into this PR's branch.

cnaples79:
@mtfriesen I updated the PR to include the changes from 5453, it should pass CI for macos now.

mtfriesen (approved):

guhetier:
> > @mtfriesen Are there any changes you would like me to make to this PR to help fix #5243 ?
> 
> Hmm, I'm not sure whether we should try to get all the packages under 10MiB (this seems like an ideal but perhaps slow outcome) or set this CI to fail only for the baseline package configurations currently under 10MiB. @guhetier, thoughts?
> 
> From what I can see in the last CI run, only ubuntu-latest with openssl TLS was over the limit; quictls builds were just under. @cnaples79, if you want to exempt openssl from this check, that might be the best short term option.

The crate includes many useless files, mainly in the OpenSSL / QuicTLS submodules. Last time we hit the issue, I excluded a lot of test content. The next big folder was documentation but excluding it carelessly was breaking the build - there are probably parameters we could tune to exclude all files that are not strictly required.

Issue #5450, opened 2025-09-15
QUIC_MAX_RANGE_ALLOC_SIZE Limit Causes Performance Degradation Under Extreme Packet Loss


stevefan1999-personal:
### Describe the bug

Performance degradation and potential DoS vulnerability in QUIC range tracking under extreme packet loss scenarios. The `QUIC_MAX_RANGE_ALLOC_SIZE` limit of 1MB (~67,786 subranges) is exhausted rapidly when handling pathological packet loss patterns, causing range operations to fail and connections to become unresponsive.

### Affected OS

- [x] Windows
- [x] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

Linux userspace applications (tested on various Linux distributions)

### MsQuic version

main

### Steps taken to reproduce bug

1. Set up a QUIC connection with high packet loss simulation
2. Send packets with large sequence number gaps (e.g., packets 1, 1000, 2, 999, 3, 998...)
3. Monitor range set usage in `QUIC_MAX_RANGE_ALLOC_SIZE` constrained structures
4. Observe connection performance degradation when range limit is reached
5. Connection becomes unresponsive or fails range operations

### Expected behavior

QUIC connections should handle extreme packet loss gracefully without hitting artificial memory limits. Range tracking should either:
- Dynamically expand memory allocation, or
- Implement range compaction to coalesce adjacent ranges and free up space

### Actual outcome

Under consistent packet loss with large gaps:
- Range set hits 67,786 subrange limit quickly
- Further range operations fail
- Connection performance degrades significantly
- Potential for DoS attacks via crafted packet sequences

### Additional details

The 1MB limit appears designed for Windows kernel memory constraints but is too restrictive for userspace Linux applications. 

**Root cause analysis:**
The issue occurs because each packet gap requires a separate subrange entry. Pathological sequences like alternating high/low packet numbers cause O(n) insertion costs due to array shifting.

**Proposed solutions:**
1. **Quick workaround:** Increase `QUIC_MAX_RANGE_ALLOC_SIZE` to 4GB or remove the limit entirely
2. **Proper fix:** Implement `QuicRangeCompact()` function to coalesce adjacent ranges before expanding allocation

**Code reference:**
https://github.com/microsoft/msquic/blob/1a1ef818035222c3bbc274e55ec71a89ff084ce1/src/core/quicdef.h#L244

**Pathological input example:**
```c
// This sequence creates worst-case O(n) behavior:
[...]

PR #5451, opened 2025-09-15
Range compaction

stevefan1999-personal:
## Description

This Pull Request adds range compaction and shrinking functionality to optimize memory usage in the QUIC range management system. The key changes include:

- **New Functions**:
  - `QuicRangeCompact`: Merges overlapping or adjacent subranges to reduce fragmentation and improve search performance.
  - `QuicRangeShrink`: Dynamically reduces the allocation size of the subranges array when usage is significantly below allocated capacity.
  - `QuicRangeCalculateShrinkLength`: Helper function to determine optimal shrink thresholds based on configurable parameters.

- **Integration**: Modified existing range operations (`QuicRangeAddRange`, `QuicRangeRemoveRange`, `QuicRangeSetMin`, `QuicRangeRemoveSubranges`) to automatically compact ranges after modifications that could create adjacent subranges.

These optimizations are particularly beneficial for long-running, highly fragmented QUIC connections where range operations accumulate over time, helping to maintain efficient memory usage and performance.

Remediates the problem in #5450, but still does not prevent worst case mentioned.

## Testing

Existing unit tests for range operations (in `src/test/`) should continue to pass and provide coverage for basic functionality. Additional tests are recommended for:
- Compaction logic with various overlap scenarios
- Shrink behavior at different usage thresholds
- Edge cases like single subrange compaction and allocation failure handling

## Documentation

No user-facing documentation impact, as these are internal implementation optimizations. If any public APIs are indirectly affected, ensure API documentation remains accurate.

### DIFF
--- a/src/core/range.c
+++ b/src/core/range.c
@@ -15,6 +15,16 @@ Abstract:
 #include "range.c.clog.h"
 #endif
 
+//
+// Forward declarations for internal helper functions
+//
+_IRQL_requires_max_(DISPATCH_LEVEL)
+uint32_t QuicRangeCalculateShrinkLength(
+    _In_ QUIC_RANGE* Range, 
+    _In_ uint32_t MinAllocMultiplier, 
+    _In_ uint32_t UsedThresholdDenominator
+    );
+
 _IRQL_requires_max_(DISPATCH_LEVEL)
 void
 QuicRangeInitialize(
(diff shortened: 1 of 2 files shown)


csujedihy:
> Remediates the problem in https://github.com/microsoft/msquic/issues/5450, but still does not prevent worst case mentioned.

Just to be clear, "Pathological sequences like alternating high/low packet numbers cause O(n) insertion costs due to array shifting." is the worst case you were referring to, right?

Looks like we need a better data structure to handle ranges in the future, like a RB tree, which we did in Windows TCP.

stevefan1999-personal:
> > Remediates the problem in https://github.com/microsoft/msquic/issues/5450, but still does not prevent worst case mentioned.
> 
> Just to be clear, "Pathological sequences like alternating high/low packet numbers cause O(n) insertion costs due to array shifting." is the worst case you were referring to, right?
> 
> Looks like we need a better data structure to handle ranges in the future, like a RB tree, which we did in Windows TCP.

Correct. Incidentally I'm writing a first boot memory allocator in kernel context, and I need to solve the memblock problem and face a similar problem, then I used a [Scapegoat Tree](https://en.m.wikipedia.org/wiki/Scapegoat_tree) to act as a BTreeMap and hacked out an easier to understand memblock :)

This is eeriely similar to the range data structure here, though

PR #5454, opened 2025-09-16
Added /profile flag only in case of pipeline build

### LINKED ISSUE #5360: Make MsQuic.dll compatible with .NET Core APIScan
.NET Core API Scan requires binary to be Vulcan ready.

For MsQuic, what seems to be missing is compiling with the `/profile` flag on MSVC
(Although we need to confirm this fully addresses the issue)

- Ensure compiling with the `/profile` flag won't have any unexpected effect on the binary (only on the PDB)
- Needed on v2.5 and v2.4
- The `/profile` flag prevent incremental compilation: it should be enabled only for pipeline runs, not on local compilation

Internal link: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2510825

Santhosha-bk:
## Description

Using /profile disables the incremental linking, to avoid it in local build, using /profile flag in case of pipleine build only.  fixes #5360 
## Testing
I ran MaxDep locally and verified.

## Documentation

### DIFF
--- a/src/bin/CMakeLists.txt
+++ b/src/bin/CMakeLists.txt
@@ -239,13 +239,11 @@ endif()
 set_property(TARGET msquic PROPERTY FOLDER "${QUIC_FOLDER_PREFIX}libraries")
 
 if(WIN32)
+    set(MSQUIC_LINK_FLAGS "/DEF:\"${CMAKE_CURRENT_BINARY_DIR}/msquic.def\"")
     if(QUIC_CI)
-        SET_TARGET_PROPERTIES(msquic
-            PROPERTIES LINK_FLAGS "/DEF:\"${CMAKE_CURRENT_BINARY_DIR}/msquic.def\" /PROFILE")
-    else()
-        SET_TARGET_PROPERTIES(msquic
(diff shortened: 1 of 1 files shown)


src/bin/CMakeLists.txt
```diff
@@ -239,8 +239,13 @@ endif()
...
 
 if(WIN32)
-    SET_TARGET_PROPERTIES(msquic
-        PROPERTIES LINK_FLAGS "/DEF:\"${CMAKE_CURRENT_BINARY_DIR}/msquic.def\" /PROFILE")
+    if(QUIC_CI)
+        SET_TARGET_PROPERTIES(msquic
+            PROPERTIES LINK_FLAGS "/DEF:\"${CMAKE_CURRENT_BINARY_DIR}/msquic.def\" /PROFILE")
+    else()
+        SET_TARGET_PROPERTIES(msquic
+            PROPERTIES LINK_FLAGS "/DEF:\"${CMAKE_CURRENT_BINARY_DIR}/msquic.def\"")
```
mtfriesen: is it possible to always set the common properties, and then have one `if (QUIC_CI)` that only appends `/PROFILE`?

Santhosha-bk: Updated the PR with suggested changes, Thanks.

PR #5459, opened 2025-09-17
ensure msquic doesn't unload while registration cleanup is pending

mtfriesen:
## Description

This PR fixes a race condition missed in PR #5376: during library uninitialization, the internal registration cleanup worker thread was not guaranteed to have flushed all registrations that had been closed to the app, causing leaks. These were caught by sanitizer in another environment; it's not clear why sanitizer has not caught the bug within the MsQuic CI, which should be executing similar code.

## Testing

We have code coverage of the scenario and sanitizer; we simply just don't hit this internal race condition.

## Documentation

No.

### DIFF
--- a/src/core/registration.c
+++ b/src/core/registration.c
@@ -84,6 +84,10 @@ CXPLAT_THREAD_CALLBACK(RegistrationCloseWorker, Context)
     if (Registration->CloseCompleteHandler != NULL) {
         BOOLEAN WakeCleanupWorker;
 
+        CXPLAT_FRE_ASSERTMSG(
+            CxPlatRundownAcquire(&MsQuicLib.RegistrationCloseCleanupRundown),
+            "The library is being unloaded while a registration is not fully closed!");
+
         QuicRegistrationClose(Registration);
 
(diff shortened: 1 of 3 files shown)


src/core/registration.c
```diff
@@ -84,6 +84,10 @@ CXPLAT_THREAD_CALLBACK(RegistrationCloseWorker, Context)
     if (Registration->CloseCompleteHandler != NULL) {
         BOOLEAN WakeCleanupWorker;
 
+        CXPLAT_FRE_ASSERTMSG(
+            CxPlatRundownAcquire(&MsQuicLib.RegistrationCloseCleanupRundown),
```
anrossi, line 88: Shouldn't the rundown acquire happen when the registration is created, so it gets released when the registration is closed? That would prevent the race this assert is supposed to catch

mtfriesen: The fundamental race condition between illegal use of API open/close and registration open/close will always exist, so it boils down to a question of where the race should be and what kind of undefined behavior is best. IMO this is no worse than undefined behavior during registration create.

PR #5464, opened 2025-09-24
Disable DSCP send && recv

ProjectsByJackHe:
## Testing

## Documentation

### DIFF
--- a/src/platform/datapath_winkernel.c
+++ b/src/platform/datapath_winkernel.c
@@ -592,60 +592,63 @@ CxPlatDataPathQuerySockoptSupport(
 
     } while (FALSE);
 
-    do {
-        DWORD TypeOfService = CXPLAT_DSCP_LE << 2;
+    // TODO: Uncomment this when Windows supports DSCP on the UDP fast indication path.
+    // do {
+    //     DWORD TypeOfService = CXPLAT_DSCP_LE << 2;
 
-        IoReuseIrp(Irp, STATUS_SUCCESS);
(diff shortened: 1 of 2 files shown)


csujedihy:
"If you know you know.", what exactly are you trying to do?

mtfriesen:
I think we should just add a knob rather than compile out the code.

ProjectsByJackHe:
> I think we should just add a knob rather than compile out the code.

Sure, this was just a quick iteration so HTTP people can validate it actually addresses the problem.

guhetier:
Converting to a draft since it sounds like there is no intent of merging that change in its current state.

src/platform/datapath_winkernel.c
```diff
@@ -592,60 +592,63 @@ CxPlatDataPathQuerySockoptSupport(
...
 
     //
     // Some USO/URO bug blocks TTL/Recv DSCP feature support on Windows Server 2022.
     //
+
     if (CxPlatform.dwBuildNumber != 20348) {
         Datapath->Features |= CXPLAT_DATAPATH_FEATURE_TTL;
-        Datapath->Features |= CXPLAT_DATAPATH_FEATURE_RECV_DSCP;
+        // TODO: Uncomment this when Windows Optimization for DSCP fast UDP indication path is added.
+        // Datapath->Features |= CXPLAT_DATAPATH_FEATURE_RECV_DSCP;
```
anrossi, line 651: This change should just disable the receive DSCP support on windows. The send support shouldn't be problematic.

PR #5467, opened 2025-09-27
Do not always enable DSCP on Recv

ProjectsByJackHe:
## Description

DSCP on Recv, when enabled, seriously regresses performance. 

At the same time, we cannot just delete the code. 

So let's come to a compromise and only enable it via a global SetParam. 

Most users of MsQuic SHOULD NOT set this on. But we set it on in our CI so our test code can exercise DSCP.

## Testing

CI

## Documentation

No

### DIFF
--- a/src/core/library.c
+++ b/src/core/library.c
@@ -1342,13 +1342,6 @@ QuicLibrarySetGlobalParam(
 #endif
 
     case QUIC_PARAM_GLOBAL_DATAPATH_DSCP_RECV_ENABLED: {
-        if (MsQuicLib.LazyInitComplete) {
-            //
-            // Not allowed to change the DSCP config after we've already started running the library.
-            //
-            Status = QUIC_STATUS_INVALID_STATE;
-            break;
-        }
 
         if (BufferLength != sizeof(BOOLEAN)) {
             Status = QUIC_STATUS_INVALID_PARAMETER;
@@ -1360,7 +1353,20 @@ QuicLibrarySetGlobalParam(
             break;
         }
 
+        if (MsQuicLib.LazyInitComplete) {
+            //
+            // Not allowed to change the DSCP config after we've already started running the library.
+            //
+            Status = QUIC_STATUS_INVALID_STATE;
+            break;
+        }
+
         MsQuicLib.EnableDscpOnRecv = *(BOOLEAN*)Buffer;
+
+        QuicTraceLogInfo(
+            LibraryDscpRecvEnabledSet,
+            "[ lib] Setting Dscp on recv = %d", MsQuicLib.EnableDscpOnRecv);
+
         Status = QUIC_STATUS_SUCCESS;
         break;
     }
--- a/src/core/library.h
+++ b/src/core/library.h
@@ -67,7 +67,7 @@ typedef struct QUIC_LIBRARY {
 
     //
     // Whether the datapath will be initialized with support for DSCP on receive.
-    // As of Windows 26100, requesting DSCP on the receive path causes packet to fall out of
+    // As of Windows 26100, requesting DSCP on the receive path causes packets to fall out of
     // the Windows fast path causing a large performance regression.
     //
     BOOLEAN EnableDscpOnRecv : 1;
--- a/src/inc/msquicp.h
+++ b/src/inc/msquicp.h
@@ -121,7 +121,7 @@ typedef struct QUIC_PRIVATE_TRANSPORT_PARAMETER {
 
 //
 // Sets whether the datapath will be initialized with support for DSCP on receive.
-// As of Windows 26100, requesting DSCP on the receive path causes packet to fall out of
+// As of Windows 26100, requesting DSCP on the receive path causes packets to fall out of
 // the Windows fast path causing a large performance regression.
 //
 #define QUIC_PARAM_GLOBAL_DATAPATH_DSCP_RECV_ENABLED    0x81000007 // BOOLEAN
--- a/src/inc/quic_datapath.h
+++ b/src/inc/quic_datapath.h
@@ -472,7 +472,7 @@ typedef CXPLAT_DATAPATH_SEND_COMPLETE *CXPLAT_DATAPATH_SEND_COMPLETE_HANDLER;
 typedef struct CXPLAT_DATAPATH_INIT_CONFIG {
     //
     // Whether the datapath will be initialized with support for DSCP on receive.
-    // As of Windows 26100, requesting DSCP on the receive path causes packet to fall out of
+    // As of Windows 26100, requesting DSCP on the receive path causes packets to fall out of
     // the Windows fast path causing a large performance regression.
     //
     BOOLEAN EnableDscpOnRecv;
--- a/src/test/bin/quic_gtest.cpp
+++ b/src/test/bin/quic_gtest.cpp
@@ -122,7 +122,6 @@ public:
 #endif
             //
             // Enable DSCP on the receive path. This is needed to test DSCP Send path.
-            // This will regress performance on Windows.
             //
             BOOLEAN Option = TRUE;
             ASSERT_TRUE(QUIC_SUCCEEDED(MsQuic->SetParam(
--- a/src/test/bin/winkernel/control.cpp
+++ b/src/test/bin/winkernel/control.cpp
@@ -708,11 +708,14 @@ QuicTestCtlEvtIoDeviceControl(
 #endif
         {
             //
-            //
-            // Our functional CI tests needs to enable recv dscp datapath features. This will regress windows performance.
+            // We don't want to hinge the result of 'Status = ' on this setparam call because
+            // this SetParam will only succeed the first time, before the datapath initializes.
+            // User mode tests already ensure at most 1 setparam call. But in Kernel mode, this IOCTL
+            // can be invoked many times.
+            // If the datapath is already initialized, this setparam call should fail silently.
             //
             BOOLEAN EnableDscpRecvOption = TRUE;
-            Status = MsQuic->SetParam(
+            MsQuic->SetParam(
                     nullptr,
                     QUIC_PARAM_GLOBAL_DATAPATH_DSCP_RECV_ENABLED,
                     sizeof(BOOLEAN),
--- a/src/generated/linux/library.c.clog.h
+++ b/src/generated/linux/library.c.clog.h
@@ -208,6 +208,23 @@ tracepoint(CLOG_LIBRARY_C, LibraryExecutionConfigSet );\
 
 
 /*----------------------------------------------------------
+// Decoder Ring for LibraryDscpRecvEnabledSet
+// [ lib] Setting Dscp on recv = %d
+// QuicTraceLogInfo(
+            LibraryDscpRecvEnabledSet,
+            "[ lib] Setting Dscp on recv = %d", MsQuicLib.EnableDscpOnRecv);
+// arg2 = arg2 = MsQuicLib.EnableDscpOnRecv = arg2
+----------------------------------------------------------*/
+#ifndef _clog_3_ARGS_TRACE_LibraryDscpRecvEnabledSet
+#define _clog_3_ARGS_TRACE_LibraryDscpRecvEnabledSet(uniqueId, encoded_arg_string, arg2)\
+tracepoint(CLOG_LIBRARY_C, LibraryDscpRecvEnabledSet , arg2);\
+
+#endif
+
+
+
+
+/*----------------------------------------------------------
 // Decoder Ring for LibraryInUse
 // [ lib] Now in use.
 // QuicTraceLogInfo(
--- a/src/generated/linux/library.c.clog.h.lttng.h
+++ b/src/generated/linux/library.c.clog.h.lttng.h
@@ -177,6 +177,24 @@ TRACEPOINT_EVENT(CLOG_LIBRARY_C, LibraryExecutionConfigSet,
 
 
 /*----------------------------------------------------------
+// Decoder Ring for LibraryDscpRecvEnabledSet
+// [ lib] Setting Dscp on recv = %d
+// QuicTraceLogInfo(
+            LibraryDscpRecvEnabledSet,
+            "[ lib] Setting Dscp on recv = %d", MsQuicLib.EnableDscpOnRecv);
+// arg2 = arg2 = MsQuicLib.EnableDscpOnRecv = arg2
+----------------------------------------------------------*/
+TRACEPOINT_EVENT(CLOG_LIBRARY_C, LibraryDscpRecvEnabledSet,
+    TP_ARGS(
+        int, arg2), 
+    TP_FIELDS(
+        ctf_integer(int, arg2, arg2)
+    )
+)
+
+
+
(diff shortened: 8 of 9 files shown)


src/inc/msquicp.h
```diff
@@ -117,6 +117,7 @@ typedef struct QUIC_PRIVATE_TRANSPORT_PARAMETER {
 #define QUIC_PARAM_GLOBAL_IN_USE                        0x81000004  // BOOLEAN
 #define QUIC_PARAM_GLOBAL_DATAPATH_FEATURES             0x81000005  // uint32_t
 #define QUIC_PARAM_GLOBAL_PLATFORM_WORKER_POOL          0x81000006  // CXPLAT_WORKER_POOL*
+#define QUIC_PARAM_GLOBAL_DATAPATH_DSCP_RECV_ENABLED    0x81000007  // BOOLEAN
```
guhetier: The comment says the option takes a `BOOLEAN`, but the implementation does not.
I think taking a boolean is the right move so both enabling and disabling is possible.

src/platform/platform_internal.h
```diff
@@ -608,6 +618,11 @@ typedef struct CX_PLATFORM {
 
     void* Reserved; // Nothing right now.
```
guhetier, line 609: Consider removing this, IIUC, it was there only to avoid an empty struct on release builds.

src/test/bin/quic_gtest.cpp
```diff
@@ -71,6 +71,7 @@ class QuicTestEnvironment : public ::testing::Environment {
             printf("Initializing for Kernel Mode tests\n");
             const char* DriverName;
             const char* DependentDriverNames;
+            CxPlatSetDscpEnabled();
```
guhetier: Please add a comment about why this is needed.

src/inc/msquicp.h
guhetier: Also, please add a comment about what this does, and why it exists + is private (vs being a default).
This is important context to avoid the same regression in the future.

src/platform/platform_internal.h
```diff
@@ -148,6 +148,11 @@ typedef struct CX_PLATFORM {
     //
     DWORD dwBuildNumber;
 
+    //
+    // Enabled whether the datapath supports RECV_DSCP.
```
guhetier: This comment seems to mix two sentences?

src/platform/datapath_winuser.c
```diff
@@ -580,7 +580,9 @@ CxPlatDataPathQuerySockoptSupport(
     //
     if (CxPlatform.dwBuildNumber != 20348) {
         Datapath->Features |= CXPLAT_DATAPATH_FEATURE_TTL;
-        Datapath->Features |= CXPLAT_DATAPATH_FEATURE_RECV_DSCP;
+        if (CxPlatGetDscpEnabled()) {
```
guhetier: Please add a comment here or when the option is actually set on the socket (same in other datapath implementations) to warn about the performance implication of setting this flag.

src/platform/platform_internal.h
ProjectsByJackHe: I'd prefer to keep it in, for consistency.

guhetier: Consistency with what? I don't see having a `Reserved` pointer being a pattern, and it doesn't make much sense outside of a public, frozen API.

src/platform/platform_internal.h
csujedihy, line 139: I don't have a strong opinion but it seems like global params are usually part of `MsQuicLib`. So, just curious why is it now in `CX_PLATFORM`?

src/platform/platform_posix.c
```diff
@@ -605,6 +605,20 @@ CxPlatGetAllocFailDenominator(
 }
 #endif
 
+void
+CxPlatSetDscpEnabled(
```
csujedihy: nit: I'd make "Recv" explicit in the function name

src/platform/platform_internal.h
ProjectsByJackHe: This was a quick decision made during an offline meeting with @anrossi. 

I had wanted to put this field in QUIC_LIBRARY  as well.

ProjectsByJackHe: @anrossi  what was your reasoning again? I am inclined to put this as a private field in `MsQuicLib`.

ProjectsByJackHe:
> Two high level comments:
> 
> * should we consider a "common" sub-struct to the CX_PLATFORM to avoid duplicating this code for all platforms?
> * I am not a fan of storing this setting in the `CX_PLATFORM` in the first place. It isn't a property of the platform. It would be better as a parameter provided to `DatapathInitialize`, and stored with most other global settings in the library object.

yes I agree, storing this in the library object would be the way to go. I had an offline meeting with @anrossi and he wanted it in CX_PLATFORM but I forgot the reasoning. @anrossi what was the reason again?

src/core/library.h
```diff
@@ -65,6 +65,12 @@ typedef struct QUIC_LIBRARY {
     //
     BOOLEAN CustomPartitions : 1;
 
+    //
+    // This datapath will be initialized with support for DSCP on receive.
+    // Enabling this option seriously MsQuic regresses performance.
```
anrossi: On *Windows*. Linux doesn't have a performance penalty for this

src/inc/msquicp.h
```diff
@@ -118,6 +118,15 @@ typedef struct QUIC_PRIVATE_TRANSPORT_PARAMETER {
 #define QUIC_PARAM_GLOBAL_DATAPATH_FEATURES             0x81000005  // uint32_t
 #define QUIC_PARAM_GLOBAL_PLATFORM_WORKER_POOL          0x81000006  // CXPLAT_WORKER_POOL*
 
+
+//
+// Set this global parameter to TURN ON DSCP on the receive path.
+// IMPORTANT: this will regress MsQuic performance, which is why this field
```
anrossi: Again, only on Windows does this reduce performance.

src/inc/quic_datapath.h
```diff
@@ -468,6 +468,15 @@ void
 
 typedef CXPLAT_DATAPATH_SEND_COMPLETE *CXPLAT_DATAPATH_SEND_COMPLETE_HANDLER;
 
+
+typedef struct CXPLAT_DATAPATH_INIT_CONFIG {
+    //
+    // Whether the datapath will be initialized with support for DSCP on receive.
+    // Enablement will seriously regress perf.
```
anrossi: Probably only need to mention the performance cost of this in one place; the `QUIC_PARAM_GLOBAL_DATAPATH_DSCP_RECV_ENABLED` setting is how apps would interact with this option, so that's the best place for the comment about performance impact.

src/inc/quic_datapath.h
```diff
@@ -478,7 +487,8 @@ CxPlatDataPathInitialize(
     _In_opt_ const CXPLAT_UDP_DATAPATH_CALLBACKS* UdpCallbacks,
     _In_opt_ const CXPLAT_TCP_DATAPATH_CALLBACKS* TcpCallbacks,
     _In_ CXPLAT_WORKER_POOL* WorkerPool,
-    _Out_ CXPLAT_DATAPATH** NewDatapath
+    _Out_ CXPLAT_DATAPATH** NewDatapath,
+    _In_ CXPLAT_DATAPATH_INIT_CONFIG* InitConfig
```
anrossi: Nit: `_In_` parameters should come before the `_Out_` parameters

src/test/bin/winkernel/control.cpp
```diff
@@ -706,7 +706,14 @@ QuicTestCtlEvtIoDeviceControl(
 #if defined(QUIC_API_ENABLE_PREVIEW_FEATURES)
         // TODO - XDP stuff, if/when supported
 #endif
-
+        //
+        // Our func test CI needs to enable recv dscp datapath features. This will regress performance.
```
anrossi: nit: we can use the full word functional

src/platform/unittest/DataPathTest.cpp
```diff
@@ -60,13 +60,15 @@ struct QuicAddr
     void Resolve(QUIC_ADDRESS_FAMILY af, const char* hostname) {
         CXPLAT_WORKER_POOL* WorkerPool = CxPlatWorkerPoolCreate(nullptr);
         CXPLAT_DATAPATH* Datapath = nullptr;
+        CXPLAT_DATAPATH_INIT_CONFIG InitConfig = {0};
```
anrossi, line 63: For the datapath tests to actually test DSCP, I think this needs to have the `DSCPRecvEnabled` field set to `TRUE`

src/platform/platform_internal.h
anrossi: It's specifically related to the platform datapath, so it made sense to have the setting on the CX_PLATFORM instead of in the library.
However, the current design (as of commit 7) with the flag being passed to the datapath during creation from the global `MsQuicLib` flag is acceptable.

src/inc/msquicp.h
anrossi: Should also mention that this parameter only has an effect on Windows too; the other platforms always have DSCP receive support enabled

src/platform/unittest/DataPathTest.cpp
ProjectsByJackHe: I can set that manually, but it looks like for this particular Resolve() test, it just initializes the datapath, runs CxPlatDatapathResolveAddress, then unintializes the datapath. I wasn't sure if it ran any of the DSCP stuff here.

anrossi: I just commented on the first occurrence in the file; looking more closely, it needs to be set on the second occurrence, since that's used for all the tests which actually check the DSCP value is sent/received.

src/inc/quic_datapath.h
guhetier: I personally think having some details at the top and bottom is useful. A bit redundant, but it makes sure it won't be ignored by people tweaking the relevant places (setting the option / implementing it in the datapath)

> // Enablement will seriously regress perf.

Please be more specific. I have bad memory, in 2 years, this comment will be useless to me ("regress perf? Why? I don't know. I guess I'll have to ignore the comment and see").
"As of Windows 26100, requesting DSCP on the receive path causes packet to fall out of the Windows fast path causing a large performance regression" is much more useful.

src/platform/platform_internal.h
ProjectsByJackHe: Removed from CX_PLATFORM.

src/platform/datapath_winuser.c
ProjectsByJackHe: Added plenty of warnings in comments.

src/platform/unittest/DataPathTest.cpp
```diff
@@ -462,12 +465,14 @@ struct CxPlatDataPath {
     {
         WorkerPool =
             CxPlatWorkerPoolCreate(Config ? Config : &DefaultExecutionConfig);
+        CXPLAT_DATAPATH_INIT_CONFIG InitConfig = {0};
```
anrossi, line 468: This one needs the DscpRecvEnabled flag also set to true.

ProjectsByJackHe: Done.

src/inc/quic_datapath.h
ProjectsByJackHe: Yeah, I like that comment. Let's opt for verbosity.

src/core/library.c
```diff
@@ -1337,6 +1341,30 @@ QuicLibrarySetGlobalParam(
     }
 #endif
 
+    case QUIC_PARAM_GLOBAL_DATAPATH_DSCP_RECV_ENABLED: {
+        if (MsQuicLib.LazyInitComplete) {
```
guhetier: nit: I would have the parameter validation before the state validation.

- parameters are the first thing to check / the call doesn't even make sense if the parameters are wrong (while we could hypothetically accept a call even after the library is inited). Checking them first is less error prone.
- most code here check parameters first, then state (even if it isn't fully consistent)

src/core/library.c
```diff
@@ -1337,6 +1341,30 @@ QuicLibrarySetGlobalParam(
     }
 #endif
 
+    case QUIC_PARAM_GLOBAL_DATAPATH_DSCP_RECV_ENABLED: {
```
guhetier, line 1344: Should we also implement the GET for completeness sake?

src/core/library.h
```diff
@@ -65,6 +65,13 @@ typedef struct QUIC_LIBRARY {
     //
     BOOLEAN CustomPartitions : 1;
 
+    //
+    // Whether the datapath will be initialized with support for DSCP on receive.
+    // As of Windows 26100, requesting DSCP on the receive path causes packet to fall out of
```
guhetier:
```suggestion
    // As of Windows 26100, requesting DSCP on the receive path causes packets to fall out of
```

src/test/bin/winkernel/control.cpp
```diff
@@ -706,7 +706,18 @@ QuicTestCtlEvtIoDeviceControl(
 #if defined(QUIC_API_ENABLE_PREVIEW_FEATURES)
         // TODO - XDP stuff, if/when supported
 #endif
-
+        {
+            //
```
guhetier, line 710: Extra comment line

src/test/bin/quic_gtest.cpp
```diff
@@ -120,6 +120,16 @@ class QuicTestEnvironment : public ::testing::Environment {
                 ASSERT_TRUE(QUIC_SUCCEEDED(Settings.SetGlobal()));
             }
 #endif
+            //
+            // Enable DSCP on the receive path. This is needed to test DSCP Send path.
+            // This will regress performance on Windows.
```
guhetier: nit: I'd either add more details or delete this line.

Since we are already explicit about the performance in all the other places that matter, here, I'd tend to delete this second line.

src/core/library.c
```diff
@@ -896,12 +896,16 @@ QuicLibraryLazyInitialize(
     }
 #endif
 
+    CXPLAT_DATAPATH_INIT_CONFIG InitConfig = {0};
+    InitConfig.EnableDscpOnRecv = MsQuicLib.EnableDscpOnRecv;
```
guhetier, line 900: Is there any place we log the options we set?
It could be helpful to know if recv dscp has been enabled when investigating.

src/core/library.c
mtfriesen: I vote against completeness if this is merely a test param - it means we have to either add another level of test code for test code, or leave the code completely untested.

src/inc/quic_datapath.h
mtfriesen: or the other way around (e.g. memcpy) which in Windows is now preferred, i believe. either way is OK as long as they're not interleaved.

src/core/library.c
ProjectsByJackHe: I agree.

src/core/library.c
ProjectsByJackHe: yeah I added a log when we first set this option.

PR #5480, opened 2025-10-03
Leverage Netperf Lab to Collect CI MEMORY dumps

### LINKED ISSUE #5273: Allow the collection of kernel dumps when the test pipeline fails
Issues like #5247 are hard to debug without access to a kernel dump.

We should have a way to collect kernel dump on failure, either leveraging Azure Watson or another mechanism

ProjectsByJackHe:
Netperf, in addition to measuring performance data on MsQuic, should also be able to run MsQuic functional CI (a subset) in Kernel mode and collect crash dumps for analysis.

Fixes #5273 
We can leverage this pattern for the XDP-for-windows project too.

Validation: 
https://github.com/microsoft/netperf/actions/runs/18208239591,

which ran on MsQuic Ref: https://github.com/microsoft/msquic/compare/main...jackhe/bugcheck-on-purpose

image

image

Here is another run which doesn't run on an MsQuic version that does not bugcheck:

https://github.com/microsoft/netperf/actions/runs/18206803000

### DIFF
--- a/.github/workflows/netperf-kernel-bvt.yml
+++ b/.github/workflows/netperf-kernel-bvt.yml
@@ -1,0 +1,32 @@
+name: Netperf Kernel BVT
+
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
+        required: true
+        description: 'Git Ref'
+        type: string
+      filter:
+        required: false
+        description: 'Custom Test Filter'
+        default: '-*ValidateConfiguration:*ValidAlpnLengths:*ResumeRejection*:*ClientCertificate*:*LoadBalanced*'
+        type: string
+
+permissions: read-all
+
+jobs:
+  run:
+    name: Run Perf
+    runs-on: windows-latest
+    steps:
+    - name: Run NetPerf Workflow
+      timeout-minutes: 120
+      shell: pwsh
+      run: |
+        $url = "https://raw.githubusercontent.com/microsoft/netperf/main/run-workflow.ps1"
+        if ('${{ secrets.NET_PERF_TRIGGER }}' -eq '') {
+            Write-Host "Not able to run because no secrets are available!"
+            return
+        }
+        iex "& { $(irm $url) } ${{ secrets.NET_PERF_TRIGGER }} msquic-kernel-bvt 'N/A' ${{ inputs.ref }} 'N/A'"
--- a/.github/workflows/netperf-kernel-bvt.yml
+++ b/.github/workflows/netperf-kernel-bvt.yml
@@ -17,10 +17,10 @@ permissions: read-all
 
 jobs:
   run:
-    name: Run Perf
+    name: Run Kernel BVT on Netperf
     runs-on: windows-latest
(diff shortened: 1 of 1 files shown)


guhetier:
Can you please link this to the issue it is addressing?

guhetier:
> TODO: We can leverage this pattern for the XDP-for-windows project too.

Please do not let a TODO in the PR description, if there is a follow up needed, we can create a task for it so we don't lose track of it.

guhetier:
IIUC, this run the kernel BVT on the lab machines shared with netperf.
- How / when is the run triggered?
    - Can we add some documentation about it, to persist the instruction to start a run?
    - If this is a manual trigger in an attempt to reproduce a previously seen issue:
        - it could be good if there was a possibility to schedule multiple runs to help with getting a repro
        - it could be good to be able to provide a filter (so that multiple runs don't take forever :D)
- Are the BVT the only tests running in KM for which we had a dump collection problem?

ProjectsByJackHe:
> IIUC, this run the kernel BVT on the lab machines shared with netperf.
> 
> * How / when is the run triggered?
>   
>   * Can we add some documentation about it, to persist the instruction to start a run?
>   * If this is a manual trigger in an attempt to reproduce a previously seen issue:
>     
>     * it could be good if there was a possibility to schedule multiple runs to help with getting a repro
>     * it could be good to be able to provide a filter (so that multiple runs don't take forever :D)
> * Are the BVT the only tests running in KM for which we had a dump collection problem?

1) This run is triggered via a manual dispatch / scheduled dispatch run (once a day). 
2) 
- Sure
- Sure

3) Primarily. We can onboard the stress ones too. Just add another workflow yaml to netperf.

.github/workflows/netperf-kernel-bvt.yml
```diff
@@ -0,0 +1,32 @@
+name: Netperf Kernel BVT
+
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
+        required: true
+        description: 'Git Ref'
+        type: string
+      filter:
```
guhetier, line 10: Is this unused?

guhetier:
> 3. Primarily. We can onboard the stress ones too. Just add another workflow yaml to netperf.

I would consider it if it isn't a lot of extra work, so we are fully done with it rather than having to come back to it when stress fails.

ProjectsByJackHe:
> > 3. Primarily. We can onboard the stress ones too. Just add another workflow yaml to netperf.
> 
> I would consider it if it isn't a lot of extra work, so we are fully done with it rather than having to come back to it when stress fails.

Stress CI isn't running in Kernel mode MsQuic today. Failures in stress tests do not generate a memory dump. Let's be careful about scope creep.

.github/workflows/netperf-kernel-bvt.yml
ProjectsByJackHe: updated

guhetier (approved):

PR #5453, opened 2025-09-15
fix: build on macOS ARM64

fernandoareias:
## Description

Fix #5446 

## Testing

CI

## Documentation

N/A

PR #5492, opened 2025-10-06
CP: Do not send CONNECTION_CLOSE frames in draining state.

guhetier:
## Description

Cherry-pick #5477 

Some spurious test failures show a pattern when the two peers entered a loop of sending CONNECTION_CLOSE frames to each other. This regression was introduced by #5107.

According to RFC 9000:
- When a peer sends a CONNECTION_CLOSE frame, it enters the "closing" state and should answer any incoming packet on the connection with a new CONNECTION_CLOSE frame (subject to rate limiting), to ensure the peer get the CONNECTION_CLOSE.
- When a peer receives a CONNECTION_CLOSE frame, it enters the "draining" state and should not send any more packets.

(The RFC is not explicit about what happens when a peer in the "closing" state receives a CONNECTION_CLOSE, but since it means both peers are closing the connection, it can stop sending additional CONNECTION_CLOSE).

However, #5107 was re-sending CONNECTION_CLOSE frame both in "closing" and "draining" state. This PR fixes it.

## Testing

CI. The issue is timing dependent and can't be reliably reproduced.

## Documentation

N/A

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -1526,6 +1526,7 @@ QuicConnTryClose(
         if (!SilentClose) {
             //
             // Enter 'draining period' to flush out any leftover packets.
+            // During that time, no packet should be sent.
             //
             QuicConnTimerSet(
                 Connection,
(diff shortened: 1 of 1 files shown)


guhetier:
MIght need to CP #5453 too.

csujedihy (approved):

csujedihy (approved):

PR #5493, opened 2025-10-06
CP: Do not always enable DSCP on Recv

ProjectsByJackHe:
## Description

CP of: https://github.com/microsoft/msquic/pull/5467

Description copied for convenience: 

DSCP on Recv, when enabled, seriously regresses performance.

At the same time, we cannot just delete the code.

So let's come to a compromise and only enable it via a global SetParam.

Most users of MsQuic SHOULD NOT set this on. But we set it on in our CI so our test code can exercise DSCP.

## Testing

CI

## Documentation

N/A


src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -541,6 +541,14 @@ void WriteCryptoFrame(
         0, ClientContext->State.BufferLength, ClientContext->State.Buffer
     };
 
+    //
+    // TODO: The code in the recvfuzzer assumes that all data produced in
```
guhetier: Was this TODO in the original PR? It seems unrelated to the fix.

ProjectsByJackHe: No, but I wanted to keep this comment rather than leave it out when resolving the merge conflict.

More docs = better anyway.

guhetier: That's a bit more complicated. More good documentation is better.
But if you pick a TODO / comment outside of the context it was written in, it might not be true or relevant for this branch.

IMO, we should pick this TODO if and when we backport the PR that introduced it.

ProjectsByJackHe: sure

Issue #5481, opened 2025-10-03
Defer queue livelock problem under high packet loss and high bandwidth scenario


stevefan1999-personal:
> Thanks for the PR.
> 
> A concern with prioritizing connection during the handshake is whether this could have a negative impact on established connections, in particular in a scenario with a very high rate of new connections, and potentially, allow for a DOS.
> 
> How much loss/reordering are you seeing and how much load (WorkerQueueDelay) are we talking about?

In our lab test environment with 20% packet loss, 20ms latency, and 1Gbps symmetric uplink/downlink, we've developed a custom benchmark tool to stress-test QUIC connections under adverse conditions. This setup simulates real-world scenarios like satellite links or congested Wi-Fi networks, where packet loss and reordering are common. Our testing revealed a complex failure mode during handshakes that combines packet deferral, work queue scheduling, and interrupt handling issues, leading to connection collapses that appear as "mysterious handshake timeouts."

Also, here's our findings for this issue:

#### Root Cause Breakdown

1. **Handshake Confirmation Packet Loss in Coalesced Datagrams**:  
   The client sends a handshake confirmation packet (e.g., containing the `HANDSHAKE_DONE` frame) to the server, often coalesced with other packets into a single UDP datagram for efficiency (as per QUIC's packet coalescing in `src/core/packet_builder.c`). However, in lossy networks, the datagram may arrive corrupted (e.g., due to bit errors or interference), causing the entire frame to be dropped at the UDP layer. Wireshark traces show the datagram arriving intact at the network level, but MsQuic's integrity checks (in `QuicConnRecvDecryptAndAuthenticate` in `src/core/connection.c`) fail, discarding it. This loss prevents the server from advancing `Connection->State.HandshakeConfirmed` to `TRUE`, stalling key upgrades.

2. **Deferred Packet Queue Overflow Due to Unconfirmed Handshake**:  
[...]

PR #5499, opened 2025-10-08
Infer max pending datagram to configured inital window packet

stevefan1999-personal:
This PR increases the type of `DeferredPacketsCount` from `uint8_t` to `uint16_t` to support a larger number of deferred packets. It also adjusts the maximum pending datagrams calculation to be based on the `InitialWindowPackets` setting plus a small buffer. This prevents the connection from dropping packets unnecessarily when a large initial window is configured, as the previous hardcoded limit of `QUIC_MAX_PENDING_DATAGRAMS` could be easily exceeded.

The change in [`src/core/connection.c`](src/core/connection.c:3716) dynamically calculates the limit, making the packet deferral mechanism more robust and adaptive to different connection settings. The type change in [`src/core/packet_space.h`](src/core/packet_space.h:61) ensures the counter can accommodate this larger, dynamic limit.

Mitigates #5481

### DIFF
--- a/src/core/connection.c
+++ b/src/core/connection.c
@@ -3716,7 +3716,7 @@ QuicConnGetKeyOrDeferDatagram(
         } else {
             QUIC_ENCRYPT_LEVEL EncryptLevel = QuicKeyTypeToEncryptLevel(Packet->KeyType);
             QUIC_PACKET_SPACE* Packets = Connection->Packets[EncryptLevel];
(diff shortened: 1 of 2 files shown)


src/core/connection.c
csujedihy, line 6016: You will need to change `uint8_t` to `uint16_t` to get the builds to pass.

PR #5503, opened 2025-10-08
separate io_uring flags into locked bitfield

mtfriesen:
## Description

One of the recent stress failures can be explained by a race between `CxPlatSocketContextUninitialize` setting `Uninitialized` and the io_uring data path reading/writing to `Shutdown` or `MultiRecvStarted`.

Split out the io_uring fields used on the data path under lock into a separate bitfield struct.

## Testing

Builds locally.

## Documentation

No.

### DIFF
--- a/src/platform/platform_internal.h
+++ b/src/platform/platform_internal.h
@@ -805,18 +805,20 @@ typedef struct QUIC_CACHEALIGN CXPLAT_SOCKET_CONTEXT {
     BOOLEAN IoStarted : 1;
 
 #ifdef CXPLAT_USE_IO_URING
-    //
-    // Indicates if the socket has started shutting down.
-    //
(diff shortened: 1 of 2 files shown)


src/platform/platform_internal.h
```diff
@@ -805,18 +805,20 @@ typedef struct QUIC_CACHEALIGN CXPLAT_SOCKET_CONTEXT {
     BOOLEAN IoStarted : 1;
 
 #ifdef CXPLAT_USE_IO_URING
-    //
-    // Indicates if the socket has started shutting down.
-    //
-    BOOLEAN Shutdown : 1;
+    struct {
```
guhetier, line 808: nit: Might be worth mentioning which lock in a comment (the partion event queue lock) - since it isn't a lock local to the struct.

PR #5506, opened 2025-10-09
Support TSAN builds

mtfriesen:
## Description

Add support for building with thread sanitizer. This is mutually exclusive with address sanitizer and (GCC-only) leak sanitizer.

## Testing

Builds Linux locally. Windows may need more work, based on the special cases required for regular sanitizer.

Added a config to the stress CI build as well,. I haven't onboarded any tests on the build, since they will time out and/or fail.

## Documentation

N/A.


CMakeLists.txt
guhetier, line 124: That name feels misleading now. Not sure what to do with it though, except maybe a comment?
Or "QUIC_ENABLE_ALMOST_ALL_SANITIZERS"? :P

mtfriesen: Yeah, this is not ideal, but we can resolve it if and when clang/gcc resolve the incompatibilities between address/leak and thread sanitizer. I see no fundamental reason they can't all be run together.

PR #5515, opened 2025-10-10
Auto-BVT bug issuer bugfix

ProjectsByJackHe:
## Description

During validation, an edge case was missed where if GH download artifacts encounters no logs, 
it will break some assumptions the script code makes about the directory structure of the BVT artifacts.

## Testing

https://github.com/microsoft/msquic/actions/runs/18392689186

## Documentation

No

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -415,13 +415,18 @@ jobs:
           const issuesToFile = [...dedup.values()];
 
           if (issuesToFile.length > 10) {
-            console.log('Too Many CI Failures...');
+            console.log('Too many CI failures: ' + issuesToFile.length);
+            let body = run_url;
+            for (const it of issuesToFile) {
+              body += it.title + ", "
+            }
             await github.rest.issues.create({
               owner, repo,
-              title: "Too Many CI Failures...",
-              body: run_url,
+              title: 'Too many CI failures: ' + issuesToFile.length,
+              body: body,
               labels: ['bvt', 'ci-failure'] // tweak/remove labels as you wish
             });
+            process.exit(0);
           }
 
           // --- Fetch all open issues once (pagination safe) ---
@@ -433,12 +438,7 @@ jobs:
           const openByTitle = new Map(openIssues.map(i => [i.title, i]));
           // --- Upsert: comment on existing (open) or create a new issue ---
 
-          let issuesFiled = 0;
           for (const it of issuesToFile) {
-            if (issuesFiled >= 10) {
-              break;
-            }
-            issuesFiled++;
             const existing = openByTitle.get(it.title);
             if (existing) {
               await github.rest.issues.createComment({


.github/workflows/test.yml
```diff
@@ -382,16 +380,26 @@ jobs:
               }
           }
 
+          find_issues('', 'msquictest');
```
mtfriesen, line 383: This seems like a potentially tricky point of maintenance - is there anything we can do to avoid enumerating binaries here?

ProjectsByJackHe: These are not binaries. They are folder names.

guhetier: I think the same question applies to folder names. Can we avoid listing them somehow? Maybe inspect all subfolders?

.github/workflows/test.yml
```diff
@@ -400,34 +408,37 @@ jobs:
...
-
           const dedup = new Map();
           for (const it of potential_issues) {
             if (!dedup.has(it.title)) dedup.set(it.title, it);
           }
           const issuesToFile = [...dedup.values()];
 
-          if (issuesToFile.length > 5) {
+          if (issuesToFile.length > 10) {
+            console.log('Too Many CI Failures...');
```
guhetier: Weird capitalization + it would be nice to log the number.

```suggestion
            console.log("Too many CI failures ${$issuesToFile.length}");
```

(or whatever the exact syntax for printing the variable is)

.github/workflows/test.yml
```diff
@@ -400,34 +408,37 @@ jobs:
...
             await github.rest.issues.create({
               owner, repo,
               title: "Too Many CI Failures...",
```
guhetier: I should have caught that in the previous PR, but:
- The title should contain [CI-TEST]
- This is a public, visible issue (not just a CI log), lets make it a bit more formal: "[CI-TEST] <numIssues> failures in CI run" (or "many failures" if having the number is problematic for aggregation)

.github/workflows/test.yml
```diff
@@ -400,34 +408,37 @@ jobs:
...
           );
-
           // Build a quick lookup by title
           const openByTitle = new Map(openIssues.map(i => [i.title, i]));
-
           // --- Upsert: comment on existing (open) or create a new issue ---
+
+          let issuesFiled = 0;
           for (const it of issuesToFile) {
+            if (issuesFiled >= 10) {
```
guhetier: If we are already filing an issue for too many issues in the pipeline, I don't think we should file the arbitrary first 10. We already have the "many issue" bug, we can work from that.

When we have many issues, it is likely something broke (vs flakyness) anyway.

.github/workflows/test.yml
guhetier: Any chance we can dump the full list of failures in the body of this issue?

.github/workflows/test.yml
ProjectsByJackHe: No matter how we slice this, the scripting code will need to make some assumptions about what the folder structure + names looks like to parse test failures from the BVT artifacts. The other solution is we don't use BVT artifacts as our source of truth of test failures, and we parse the entire STDOUT from all the jobs and try to extract information there with RegEx or something. IMO, that approach is equally if not more complex.

.github/workflows/test.yml
ProjectsByJackHe: See latest iteration.

.github/workflows/test.yml
ProjectsByJackHe: See latest iteration.

PR #5519, opened 2025-10-12
Rust build script fixes

nishanthkarthik:
## Description

1. Enable logging to stdout using Cargo feature flags
2. Fix build failure when `static` feature is enabled but libnuma is missing

See commit messages for more explanation

## Testing

This PR does not (and should not) change any defaults. Fixes are opt-in through feature flags.

## Documentation

Not sure


scripts/build.rs
```diff
@@ -43,6 +43,14 @@ fn cmake_build() {
         .define("QUIC_ENABLE_LOGGING", logging_enabled)
         .define("QUIC_OUTPUT_DIR", quic_output_dir.to_str().unwrap());
 
+    if cfg!(unix) && cfg!(feature = "logging") {
+        config.define("QUIC_LOGGING_TYPE", "stdout");
```
guhetier, line 47: Logging using `stdout` is a big performance hit, it should not be a default.
`lttng` would be better, with `stdout` taking some explicit opt-in.

PR #5521, opened 2025-10-13
Add io_uring sqe synchronization diagnostics and fix submission bug

### LINKED ISSUE #5350: [CI-TEST] Stress Test failed with iouring
It crashed, no log to report.

Failure here:
https://github.com/microsoft/msquic/actions/runs/16916730458/job/47932965211?pr=5327

### LINKED ISSUE #5490: [CI-TEST] Stress (Debug, linux, ubuntu-24.04, x64, quictls, -Sanitize, -Test, -UseIoUring)
Stress test failure in https://github.com/microsoft/msquic/pull/5483

mtfriesen:
## Description

Add more debug assertions to detect certain unsynchronized patterns. While adding these and reviewing io_uring documentation, I realized one code path calls `io_uring_submit` without any external synchronization and fixed that too, after verifying the new diagnostics catch the error locally.

The symptoms should lead exactly to the io_uring flakiness we were seeing, so optimistically:

Resolves #5350
Resolves #5490

## Testing

Builds and spins locally, and detects the bug fixed.

## Documentation

No.

### DIFF
--- a/src/inc/quic_platform_posix.h
+++ b/src/inc/quic_platform_posix.h
@@ -1020,7 +1020,8 @@ typedef struct CXPLAT_EVENTQ {
     //
     CXPLAT_LOCK Lock;
 #if DEBUG
-    uint32_t ContentionCount;
+    uint32_t CqContentionCount;
+    uint32_t SqContentionCount;
 #endif
     BOOLEAN NeedsSubmit;
 } CXPLAT_EVENTQ;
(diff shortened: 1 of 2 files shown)


src/inc/quic_platform_posix.h
```diff
@@ -1121,22 +1162,28 @@ CxPlatEventQDequeue(
...
     if (Queue->NeedsSubmit) {
         //
-        // Review: can be batched with waits below.
+        // Review: can be batched with waits below if fully partitioned.
         //
-        io_uring_submit(&Queue->Ring);
+        CxPlatLockAcquire(&Queue->Lock);
+        CxPlatEventQSubmit(Queue);
         Queue->NeedsSubmit = FALSE;
+        CxPlatLockRelease(&Queue->Lock);
```
ProjectsByJackHe, line 1175: So, this is the key code change that fixes the flakiness?

The rest of the changes looks to be ASSERT modifications and wrapping io_uring_submit calls.

mtfriesen: Yep, that's the actual fix. io_uring works a bit different than some of the other ring APIs, where in others the call into the kernel can be made thread-safe, in io_uring the submission modifies the submission ring and needs mutual exclusion with other submission code paths.

src/inc/quic_platform_posix.h
```diff
@@ -1087,6 +1088,46 @@ CxPlatEventQCleanup(
...
+QUIC_INLINE
+struct io_uring_sqe*
+CxPlatEventGetSqe(
+    _In_ CXPLAT_EVENTQ* Queue
+    )
+{
+    struct io_uring_sqe* io_sqe;
+#if DEBUG
+    CxPlatLockAcquire(&Queue->Lock);
+    CXPLAT_DBG_ASSERT(Queue->SqContentionCount++ == 0);
```
guhetier, line 1100:
```suggestion
    CXPLAT_DBG_ASSERT(Queue->SqContentionCount == 0);
    Queue->SqContentionCount++;
```

For the sake of clarity, I'd split this. The 1 line saved is, in my opinion, not worth having to think about the postfix / prefix behaviors for `++` and `--`.

src/inc/quic_platform_posix.h
```diff
@@ -1087,6 +1088,46 @@ CxPlatEventQCleanup(
...
+    CxPlatLockAcquire(&Queue->Lock);
+    CXPLAT_DBG_ASSERT(--Queue->SqContentionCount == 0);
+    CxPlatLockRelease(&Queue->Lock);
+#endif
+    return io_sqe;
+}
+
+QUIC_INLINE
+void
+CxPlatEventQSubmit(
```
guhetier, line 1114: The issue was that this call needs to happen under lock, right?
Could we add some SAL annotation / comment?
Since the lock is re-entrant, should we simply take the lock inside of the function?

src/inc/quic_platform_posix.h
mtfriesen: 6 of one, half dozen the other. Sticking with concision.

src/inc/quic_platform_posix.h
mtfriesen: No lock is required, and eventually the goal is to remove all locks from the io_uring implementation and rely on internal/external end-to-end synchronization via explicit partitioning.

guhetier: I'm not sure I understand. On another comment, you mention that the fix is to call this under lock because these call needs synchronization for io_uring. If you have a pointer to an explanation, I'd take it.

mtfriesen: io_uring is not thread-safe for many operations; callers of io_uring APIs must provide some external synchronization.

This PR fixes a thread-unsafe operation by acquiring a lock. This function does not require holding a lock.

mtfriesen:
@microsoft-github-policy-service rerun

PR #5522, opened 2025-10-15
Prevent sending CONN_CLOSE when closing the connection silently

guhetier:
## Description

This fixes an intermittent test failure due to connection being able to send a CONN_CLOSE frame after being silently closed.
This was a regression introduce by #5107.

If a connection was closed locally with the "silent" flag, it would still send a CONN_CLOSE frame when receiving a frame from the peer.
This fix prevents this behavior in a generic way by adding a check when adding send flags.

This minimal fix needs to be backported to v2.5.
A follow up PR with a refactoring of the connection close state will follow.

## Testing

Already covered.

## Documentation

N/A

### DIFF
--- a/src/core/connection.h
+++ b/src/core/connection.h
@@ -35,7 +35,7 @@ typedef union QUIC_CONNECTION_STATE {
         BOOLEAN Connected       : 1;    // Handshake completed.
         BOOLEAN ClosedLocally   : 1;    // Locally closed.
         BOOLEAN ClosedRemotely  : 1;    // Remotely closed.
-        BOOLEAN ClosedSilently  : 1;    // Closed with the QUIC_CLOSE_SILENT flag - go directly to draining state.
+        BOOLEAN ClosedSilently  : 1;    // Closed with the QUIC_CLOSE_SILENT flag.
         BOOLEAN AppClosed       : 1;    // Application (not transport) closed connection.
         BOOLEAN ShutdownComplete : 1;   // Shutdown callback delivered for handle.
         BOOLEAN HandleClosed    : 1;    // Handle closed by application layer.


src/core/connection.c
mtfriesen, line 1481: Does it make sense to set `ClosedSilently` if the connection is already closed? Is it required?

src/core/connection.h
```diff
@@ -35,6 +35,7 @@ typedef union QUIC_CONNECTION_STATE {
         BOOLEAN Connected       : 1;    // Handshake completed.
         BOOLEAN ClosedLocally   : 1;    // Locally closed.
         BOOLEAN ClosedRemotely  : 1;    // Remotely closed.
+        BOOLEAN ClosedSilently  : 1;    // Closed with the QUIC_CLOSE_SILENT flag - go directly to draining state.
```
mtfriesen: The "go directly to draining state" seems like a possibly misplaced comment - should it go into `QuicConnTryClose` instead?

src/manifest/clog.sidecar
```diff
@@ -2707,18 +2707,6 @@
       ],
       "macroName": "QuicTraceLogWarning"
     },
-    "DatapathQueryRioDispatchFailed": {
-      "ModuleProperites": {},
-      "TraceString": "[data] Query for SIO_GET_MULTIPLE_EXTENSION_FUNCTION_POINTER failed, 0x%x",
```
mtfriesen, line 2712: What happened here? Why is this changing?

mtfriesen: I remember deleting RIO a while ago, which was nice, but I don't know why clog waited until now to remove the traces.

mtfriesen: RIO code was removed in #5328. CLOG seemed satisfied at that time, which was ~two months ago.

guhetier: I suspect that CLOG doesn't remove deleted logs from the sidecar if you don't delete this file + regenerate it from zero (which I had to do, because it doesn't like modification to a log otherwise).

src/core/connection.c
guhetier: Yes. We could be in "closing" state, and be sending CONN_CLOSE frames as the answer to any incoming packet, which we should stop doing when we get a second "silent" close.

I am not sure when potentially get a second close, but the code below makes it clear it can happen.

src/core/connection.h
guhetier: Good point, that doesn't make sense anymore with the current state of the PR.


```suggestion
        BOOLEAN ClosedSilently  : 1;    // Closed with the QUIC_CLOSE_SILENT flag.
```

src/manifest/clog.sidecar
mtfriesen: Can you file a bug to track that in clog? Generated code should reliably produce the same output, and I can't imagine why previous output should feed back into the next output.

mtfriesen: Actually, I clicked the built-in button on this thread to open https://github.com/microsoft/CLOG/issues/118.

src/core/connection.c
mtfriesen: It would be nice to provide a specific high-level example scenario where that behavior is expected or required, but I can take on face value that a connection may "close" gracefully in some context, and then decide it wants to more forcefully ("silently") close.

guhetier: I'll add that in the follow up refactor PR I am working on.
My best guess is that it could happen if we have a protocol error causing a silent close right after an application close.

mtfriesen (approved):

PR #5523, opened 2025-10-16
Change the format specifier

gaurav2699:
## Description
The msquic pipeline in undock using UCRT compiler is failing due to format specifier mismatch. BOOLEAN is typedef'd as unsigned char, but the format string uses %d which expects a signed integer. When an unsigned char is passed to a function that expects a signed type, the compiler generates a C6340 warning.

## Testing
The new changes are tested on the undock pipeline and its passing.

## Documentation
No


guhetier (approved):
Do we need to backport this to v2.5?

gaurav2699:
> Do we need to backport this to v2.5?

Yes

PR #5547, opened 2025-10-29
Fix maximum stream count value in Streams.md

guhetier:
## Description

Corrected the maximum stream count limit from 2 ^ 16 to 65,535 in the documentation.

## Testing

N/A

## Documentation

N/A


guhetier:
@kitlith FYI

PR #5548, opened 2025-10-29
Update Dependabot to ignore updating Ubuntu 22.04 minor versions

### LINKED ISSUE #5536: Internal mirror raises duplicate dependabot PRs
### Describe the bug

Our internal mirror repo has a separate ES that automatically raises dependabot PRs that duplicate or conflict with the public dependabot racing it.

Figure out how to suppress the internal dependabot when the internal mirror is within a certain grace period.

Also, make sure the two dependabots share a common configuration and common goal state at any given time, since some internal PRs do not match the public dependabot.

### Affected OS

- [ ] Windows
- [ ] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

_No response_

### MsQuic version

main

Create an internal ADO mirror? 

### Expected behavior

The internal mirror's dependabot or other automation should catch out-of-date dependencies beyond a reasonable grace period for the public repo to merge an equivalent dependabot PR and sync with the mirror.

### Actual outcome

Dependabot fights.

### Additional details

_No response_

anrossi:
## Description

Update Dependabot to skip updating major and minor versions on Ubuntu 22.04. This should cause Dependabot to only update the patch version.
Also update Ubuntu 22.04 to the latest patch.

Hopefully fixes #5536.

## Testing

Manually ran dependabot CLI with this ignore statement to test that it correctly only bumps the patch/commit version.

## Documentation

N/A

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -35,14 +35,6 @@ updates:
   - package-ecosystem: "docker"
     directories:
       - "/.docker/ubuntu-22.04"
-    schedule:
-      interval: "monthly"
-    ignore:
-      - dependency-name: "ubuntu"
-        update-types: ["version-update:semver-major", "version-update:semver-minor"]
-
-  - package-ecosystem: "docker"
-    directories:
       - "/.docker/ubuntu-24.04"
     schedule:
       interval: "monthly"


.github/dependabot.yml
```diff
@@ -40,6 +40,7 @@ updates:
     ignore:
       - dependency-name: "ubuntu"
         versions: ["24.x", ">= 23.0.0", "23.x"]
```
csujedihy: Versions are stale now?

csujedihy:
Dependabot has a cli which you can use to test the changes.

.github/dependabot.yml
anrossi: What do you mean by stale? Like they're replaced by the update-types info?

csujedihy: I meant redundant now because the update-types limit the updates to patch version only.

.github/dependabot.yml
guhetier: A similiar change would be needed here?

.github/dependabot.yml
```diff
@@ -39,7 +39,7 @@ updates:
       interval: "monthly"
     ignore:
       - dependency-name: "ubuntu"
-        versions: ["24.x", ">= 23.0.0", "23.x"]
+        update-types: ["version-update:semver-major", "version-update:semver-minor"]
 
   - package-ecosystem: "docker"
```
csujedihy: You don't need to 2 entries for docker now. We had 2 because they had different ignore conditions.

.docker/ubuntu-22.04/Dockerfile
```diff
@@ -55,7 +55,7 @@ RUN git config --global safe.directory '*'
 FROM base AS xdp-build
 
 RUN apt-get update -y && apt-get install -y \
-    && echo "deb [arch=amd64] http://cz.archive.ubuntu.com/ubuntu noble main" > /etc/apt/sources.list.d/xdp.list \
```
mtfriesen, line 58: how did `cz` ever get here? wow...

guhetier: Contribution from .Net?

guhetier (approved):

PR #5549, opened 2025-10-30
Add test cases for unit testing QuicBufferResetRead function.

saikat107:
## Description
Add test cases for unit testing `QuicRecvBufferResetRead` function in `src\core\recv_buffer.c`. 
Currently none of the unit tests in `src\core\unittest\RecvBufferTest.cpp` test this functionality. 

## Testing
Two new tests are added as part of this PR `RecvBufferResetReadTest.ResetClearsPendingAndExternalReference_SingleMode`

## Documentation
N/A


src/core/unittest/RecvBufferTest.cpp
```diff
@@ -2157,3 +2157,117 @@ INSTANTIATE_TEST_SUITE_P(
...
+        }
+
+        // 2. Corrupt the list: mark it as non-empty but make `Flink` a bogus
+        //    (non-dereferenceable) pointer. The value `0x1` is in a guard
+        //    page on all supported Windows architectures and is guaranteed
+        //    to AV when accessed.
+        RecvBuf.RecvBuf.Chunks.Flink = reinterpret_cast<CXPLAT_LIST_ENTRY*>(0x1);
+
+        // 3. Invoke the API under test. Attempting to touch the first chunk
+        //    through the corrupted pointer should crash.
```
guhetier: I don't think this is a behavior we want to test for: dereferencing a corrupted pointer is undefined behavior, we would generally observe a crash - but not always, so this test could fail intermittently.

And more importantly, it isn't part of the contract the receive buffer makes with its callers: if the caller violates the pre-condition for calling in the receive buffer (here "the receive buffer is in a valid state"), calling the function is undefined behavior.

saikat107: Fair enough! 
Then I am removing this test. What about the other test?

src/core/unittest/RecvBufferTest.cpp
```diff
@@ -2157,3 +2157,79 @@ INSTANTIATE_TEST_SUITE_P(
...
+This test exercises the normal control-flow path of
+QuicRecvBufferResetRead where:
+1. The receive buffer is in SINGLE-chunk mode.
+2. The internal chunk list is non-empty.
+3. A read operation has been performed so that
+    – The first chunk's `ExternalReference` flag is TRUE, and
+    – `ReadPendingLength` is non-zero.
+
+After invoking QuicRecvBufferResetRead we expect:
+    • The first chunk's ExternalReference flag is cleared (FALSE).
```
guhetier: nit: Those ` • ` should be replaced with `-` to keep the character set simple

src/core/unittest/RecvBufferTest.cpp
```diff
@@ -2157,3 +2157,79 @@ INSTANTIATE_TEST_SUITE_P(
     WithMode,
     ::testing::Values(QUIC_RECV_BUF_MODE_SINGLE, QUIC_RECV_BUF_MODE_CIRCULAR, QUIC_RECV_BUF_MODE_MULTIPLE, QUIC_RECV_BUF_MODE_APP_OWNED),
     testing::PrintToStringParamName());
+
+/*
```
guhetier: We avoid this comment style, please favor `//`

This comment is also very verbose, a more concise description would be better (stating the intent of the path that will be covered rather than the details.

Basically `Happy path validation` the bit that matters here.

src/core/unittest/RecvBufferTest.cpp
```diff
@@ -2157,3 +2157,79 @@ INSTANTIATE_TEST_SUITE_P(
...
+    ASSERT_TRUE(NewDataReady);
+
+    // 3. Read once to create a pending read and mark the first chunk as having
+    //    an external reference.
+    uint64_t ReadOffset = 0;
+    QUIC_BUFFER Buffers[3] = {};
+    uint32_t BufferCount = ARRAYSIZE(Buffers);
+    RecvBuf.Read(&ReadOffset, &BufferCount, Buffers);
+
+    // Sanity – we should have read exactly the 16 bytes we wrote.
```
guhetier: More of an opinionated comment - I think we don't need these sanity checks.

`Read` has already been validated by the unit test targeting this function, so in this test targeting `QuicRecvBufferResetRead`, we can assume it is correct, and make this test more concise and easier to understand.

Same comment for some of the following checks before the Reset call.
(Asserting success on the function calls is still good for error handling)

src/core/unittest/RecvBufferTest.cpp
saikat107: Removed the whole comment block. I think the test name is self-explanatory.

src/core/unittest/RecvBufferTest.cpp
saikat107: See above.

src/core/unittest/RecvBufferTest.cpp
saikat107: Good point. Updated to only keep the necessary asserts.

PR #5569, opened 2025-11-04
Update the macos runner version

ProjectsByJackHe:
## Description

See https://github.com/actions/runner-images/issues/13046

MacOs-13 is being deprecated soon. Let's switch to use macos-15.

## Testing

CI

## Documentation

No


guhetier:
Please provide an accurate PR description (even if short for a simple change). PR descriptions are useful in the future, the link won't persist and is not helpful without context.

guhetier (approved):

Issue #5570, opened 2025-11-05
Add ref count tags to all MsQuic objects


anrossi:
MsQuic has ref count tagging for Connections and Streams which helps when debugging ref count mismatches.  This same tagging should be expanded for all MsQuic object types to help debugging more issues.

PR #5573, opened 2025-11-07
Add detailed Registration reference tracking

anrossi:
## Description

Add detailed reference tracking to `QUIC_REGISTRATION` objects to help with debugging hangs at `RegistrationClose`.  Part of #5570 .

## Testing

The CI should cover all relevant usages.

## Documentation

N/A

### DIFF
--- a/src/core/registration.c
+++ b/src/core/registration.c
@@ -186,7 +186,7 @@ MsQuicRegistrationOpen(
     CxPlatRundownInitialize(&Registration->Rundown);
 #if DEBUG
     CxPlatRefInitializeMultiple(Registration->RefTypeBiasedCount, QUIC_REG_REF_COUNT);
-    CxPlatRefInitializeEx(&Registration->RefTypeBiasedCount[QUIC_REG_REF_HANDLE_OWNER], 2);
+    CxPlatRefIncrement(&Registration->RefTypeBiasedCount[QUIC_REG_REF_HANDLE_OWNER]);
 #endif
     CxPlatEventInitialize(&Registration->CloseEvent, TRUE, FALSE);
     Registration->AppNameLength = (uint8_t)(AppNameLength + 1);


src/core/registration.h
```diff
@@ -148,6 +166,55 @@ typedef struct QUIC_REGISTRATION {
 #define QUIC_REG_VERIFY(Registration, Expr)
 #endif
 
+//
+// Adds a reference to the Registration.
+//
+_IRQL_requires_max_(DISPATCH_LEVEL)
+QUIC_INLINE
+BOOLEAN
+QuicRegistrationAddRef(
```
guhetier: Given that this is a rundown ref count, not an ownership ref count, I would reflect it on the helper name to avoid confusion.
The concepts are close but treated differently (if they should is another discussion), I think we should keep them clear.

anrossi: Fair point. The name should reflect the actual operation underneath.

src/core/registration.h
```diff
@@ -21,6 +21,17 @@ typedef enum QUIC_CONNECTION_ACCEPT_RESULT {
     QUIC_CONNECTION_REJECT_APP
 } QUIC_CONNECTION_ACCEPT_RESULT;
 
+typedef enum QUIC_REGISTRATION_REF {
+
+    QUIC_REGI_REF_HANDLE_OWNER,         // Application or Core.
```
mtfriesen: The `QUIC_REGI_` prefix seems novel - I'd recommend the full `QUIC_REGISTRATION_`, or otherwise truncate to 3 characters (`QUIC_REG_`), which is a Windows/NT style norm.

src/core/registration.h
```diff
@@ -21,6 +21,17 @@ typedef enum QUIC_CONNECTION_ACCEPT_RESULT {
...
+    QUIC_REGI_REF_CONFIGURATION,        // Per Configuration.
+    QUIC_REGI_REF_CONNECTION,           // Per Connection.
+    QUIC_REGI_REF_LISTENER,             // Per Listener.
```
mtfriesen: I don't think `LISTENER, // Listener` style comments add any value to offset their noise.

src/core/registration.h
```diff
@@ -100,6 +111,13 @@ typedef struct QUIC_REGISTRATION {
     //
     CXPLAT_RUNDOWN_REF Rundown;
 
+#if DEBUG
+    //
+    // Detailed ref counts. The actual reference count is in the Rundown.
+    //
+    short RefTypeCount[QUIC_REGI_REF_COUNT];
```
mtfriesen: Why a `short`? That seems like a recipe for underflow/overflow. The type of each refcount tracker should be at least as wide as the underlying refcount.

src/core/registration.h
```diff
@@ -148,6 +166,55 @@ typedef struct QUIC_REGISTRATION {
...
+QUIC_INLINE
+void
+QuicRegistrationRundownRelease(
+    _In_ QUIC_REGISTRATION* Registration,
+    _In_ QUIC_REGISTRATION_REF Ref
+    )
+{
+
+#if DEBUG
+    CXPLAT_TEL_ASSERT(Registration->RefTypeCount[Ref] > 0);
```
mtfriesen: Please don't read a refcount field without at least a volatile accessor. The assert below of the result of the interlocked result is strictly a superset of this check.

src/core/registration.h
```diff
@@ -148,6 +166,55 @@ typedef struct QUIC_REGISTRATION {
...
+    uint16_t result = (uint16_t)InterlockedDecrement16((volatile short*)&Registration->RefTypeCount[Ref]);
+    CXPLAT_TEL_ASSERT(result != 0xFFFF);
```
mtfriesen: We have a best practice within Windows of avoiding `0xFFFF` literal typedefs in favor of `MAXUSHORT`. There have been non-zero bugs we've had to fix where the literal was `0xFFF` or `0xFFFFF` and easily avoidable.

src/core/registration.c
```diff
@@ -239,6 +246,9 @@ MsQuicRegistrationOpen(
         CXPLAT_DBG_ASSERT(!Registration->CloseThread);
         CxPlatEventUninitialize(Registration->CloseEvent);
         CxPlatRundownUninitialize(&Registration->Rundown);
+#if DEBUG
```
mtfriesen, line 258: This should probably be above `CxPlatRundownUninitialize` if we're reversing the order when unwinding, the usual pattern.

src/core/registration.h
mtfriesen: In comments (not necessarily function names) I suggest the phrase "rundown reference" when describing what gets added/removed.

src/core/registration.h
```diff
@@ -148,6 +166,55 @@ typedef struct QUIC_REGISTRATION {
...
+//
+_IRQL_requires_max_(DISPATCH_LEVEL)
+QUIC_INLINE
+void
+QuicRegistrationRelease(
+    _In_ QUIC_REGISTRATION* Registration,
+    _In_ QUIC_REGISTRATION_REF Ref
+    )
+{
+
```
mtfriesen: Nit: blank line can be removed

src/core/registration.h
mtfriesen: Oh, wow, what a mess, I see this is copied from existing patterns:

```
    //
    // Number of references to the handle.
    //
    long RefCount;

#if DEBUG
    //
    // Detailed ref counts
    //
    short RefTypeCount[QUIC_CONN_REF_COUNT];
#endif
```

The `short` is most egregious, but `long` is also wrong. All of these should be of type `CXPLAT_REF_COUNT`.

src/core/registration.h
guhetier: +1, a single "indicate the resource type the registration rundown must wait for" would be more helpful.

src/core/registration.h
anrossi: I'll go fix the location where this was copied from and remove this

src/core/registration.h
anrossi: The `CxPlatRefIncrement` routine requires the ref count to start at 1, which isn't the case for these detailed counters. So, I'm using the `CXPLAT_REF_COUNT` type, so that the size matches up with the rundown, but using the `Interlocked*` functions to manipulate the values.
I don't think initializing the detailed counters to 1 with `CxPlatRefInitialize` is the right move since that'll require future investigations to remember that the fields are one-greater than the real count of references.

guhetier: If we are not manipulating the value with `CxPlatRef*` helpers, I don't think it should be of the `CXPLAT_REF_COUNT` type.  That would be confusing.

src/core/registration.h
```diff
@@ -148,6 +166,53 @@ typedef struct QUIC_REGISTRATION {
...
+QUIC_INLINE
+void
+QuicRegistrationRundownRelease(
+    _In_ QUIC_REGISTRATION* Registration,
+    _In_ QUIC_REGISTRATION_REF Ref
+    )
+{
+#if DEBUG
+    QuicDecrementLongPtrRelease(&Registration->RefTypeCount[Ref]);
+    CXPLAT_DBG_ASSERT(QuicReadLongPtrNoFence(&Registration->RefTypeCount[Ref]) >= 0);
```
mtfriesen: We should assert the result of the `QuicDecrementLongPtrRelease`, not peek at a volatile value which may have bounced back up to.

src/core/registration.h
```diff
@@ -148,6 +166,53 @@ typedef struct QUIC_REGISTRATION {
...
+    _In_ QUIC_REGISTRATION_REF Ref
+    )
+{
+    BOOLEAN Result = CxPlatRundownAcquire(&Registration->Rundown);
+#if DEBUG
+    if (Result) {
+        //
+        // Only increment the detailed ref count if the Rundown acquire succeeded.
+        //
+        QuicIncrementLongPtrNoFence(&Registration->RefTypeCount[Ref]);
```
mtfriesen: We should assert we didn't bounce off `-1`.

mtfriesen: (The assert should work for all negative bounced-from values)

src/core/registration.h
mtfriesen: I tend to agree solving this with the proper refcount helpers (with an implicit `+1` bias) would be most consistent and also allow us to benefit from their existing and robust overflow/underflow/bounce checks. That's what I would do.

That said, I tend to lean towards a `REF_COUNT` type for a reference count being more expressive and intrinsically guaranteeing the required properties of a reference count, which is what these are. Any other type leads towards the developer choosing their own idea of what a correct reference type is, which empirically gets chosen dangerously wrong.

src/core/registration.c
```diff
@@ -171,6 +184,10 @@ MsQuicRegistrationOpen(
     CxPlatListInitializeHead(&Registration->Connections);
     CxPlatListInitializeHead(&Registration->Listeners);
     CxPlatRundownInitialize(&Registration->Rundown);
+#if DEBUG
+    CxPlatRefInitializeMultiple(Registration->RefTypeBiasedCount, QUIC_REG_REF_COUNT);
+    CxPlatRefInitializeEx(&Registration->RefTypeBiasedCount[QUIC_REG_REF_HANDLE_OWNER], 2);
```
mtfriesen: IMO the ref should just be bumped here, not re-initialized to `2` from `1`.

src/inc/quic_platform_winkernel.h
```diff
@@ -401,6 +401,18 @@ CxPlatRefInitialize(
     *RefCount = 1;
 }
 
+QUIC_INLINE
+void
+CxPlatRefInitializeMultiple(
```
mtfriesen, line 406: I'm kind of surprised this function needs to be implemented in multiple places, but it seems to be somewhat the norm within MsQuic. Ideally it would have just one definition, even if it needs to be declared in each of these platform-specific files.

mtfriesen (dismissed):

src/core/registration.c
anrossi: I wanted to avoid the cost of an interlocked operation when the object hasn't even been shared yet. I expect the compiler can decide how best to optimize the two initializations

src/inc/quic_platform_winkernel.h
anrossi: I was thinking the same thing, but I think the reason CxPlat copies the ref count APIs in each platform file is because the underlying CXPLAT_REF_COUNT type is a platform-specific type (`LONG_PTR` on Windows and `int64_t` on POSIX) and isn't defined in the top-level quic_platform.h.
There's definitely some deduplication/reordering that could be done in the CxPlat layer.

src/core/registration.c
mtfriesen: I don't feel strongly this should be changed, but I do feel strongly there's no performance justification in this debug-only code; such premature micro-optimizations are what left us with `short` types instead of properly sized types. If anything, debug code should tend towards being robust, maintainable, and obvious, rather than bleeding-edge fast.

PR #5580, opened 2025-11-12
Fix actions permissions

ProjectsByJackHe:
## Description

To allow for the automatic creation of GH issues, we had to elevate test.yml to write issues.
But unknowingly, this action restricted certain permissions in test.yml. 
Let's be explicit and enumerate all.

## Testing

CI

## Documentation

N/A

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,20 +46,7 @@ concurrency:
   cancel-in-progress: true
 
 permissions:
-  actions: read,
-  attestations: read
-  checks: read
   contents: read
-  deployments: read
-  discussions: read
-  models: read
-  packages: read
-  pages: read
-  pull-requests: read
-  repository-projects: read
-  statuses: read
-  security-events: read
-  id-token: read
   issues: write
 
 jobs:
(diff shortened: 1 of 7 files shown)


.github/workflows/test.yml
```diff
@@ -46,7 +46,20 @@ concurrency:
   cancel-in-progress: true
 
 permissions:
+  actions: read,
```
guhetier: Did you try giving the `build-reuse-*.yml` the `contents: read` permission only instead of adding everything here?
That should be what they all mostly need, and that's only 4 files (it doesn't look like we have a dependency tree getting more).

mtfriesen: I think there's also a `read-all` type option that would probably be fine, and then `issues: write` could extend that permission?

guhetier: Looking at the doc, you can't combine `read-all` + `issues: write`, either you use a `*-all` or you enumerate :(

mtfriesen: Brutal

.github/workflows/test.yml
```diff
@@ -46,7 +46,20 @@ concurrency:
...
   contents: read
+  deployments: read
+  discussions: read
+  models: read
+  packages: read
+  pages: read
+  pull-requests: read
+  repository-projects: read
+  statuses: read
+  security-events: read
```
mtfriesen: We'd want to be careful with `security-events` in particular, since those could be sensitive.

Agreed with Guillaume's comments above: let's grant the least privileges required, even if it's a little tedious to get there.

PR #5581, opened 2025-11-12
Update Stream detailed Ref Counts

anrossi:
## Description

Update the detailed ref counts for streams to use `CXPLAT_REF_COUNT` and the associated functions. Part of #5570.

## Testing

CI and local test run

## Documentation

N/A

### DIFF
--- a/src/core/stream.c
+++ b/src/core/stream.c
@@ -74,7 +74,7 @@ QuicStreamInitialize(
         Connection->Settings.StreamMultiReceiveEnabled &&
         !Stream->Flags.UseAppOwnedRecvBuffers;
     Stream->RecvMaxLength = UINT64_MAX;
-    Stream->RefCount = 1;
+    CxPlatRefInitialize(&Stream->RefCount);
     Stream->SendRequestsTail = &Stream->SendRequests;
     Stream->SendPriority = QUIC_STREAM_PRIORITY_DEFAULT;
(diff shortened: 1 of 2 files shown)


src/core/stream.c
```diff
@@ -89,7 +89,8 @@ QuicStreamInitialize(
     Stream->ReceiveCompleteOperation->API_CALL.Context->Type = QUIC_API_TYPE_STRM_RECV_COMPLETE;
     Stream->ReceiveCompleteOperation->API_CALL.Context->STRM_RECV_COMPLETE.Stream = Stream;
 #if DEBUG
-    Stream->RefTypeCount[QUIC_STREAM_REF_APP] = 1;
+    CxPlatRefInitializeMultiple(Stream->RefTypeBiasCount, QUIC_STREAM_REF_COUNT);
```
mtfriesen: `RefTypeBiasedCount` in registration. I prefer both consistency and the `Biased` word, as a `BiasCount` would be a count of the bias.

src/core/stream.c
```diff
@@ -173,6 +174,7 @@ QuicStreamInitialize(
 
     if (Stream) {
 #if DEBUG
+        CxPlatRefDecrement(&Stream->RefTypeBiasCount[QUIC_STREAM_REF_APP]);
```
mtfriesen: missing assert the refcount didn't go to 1

mtfriesen: or below 1, or whatever 😄

PR #5584, opened 2025-11-13
Update Connection reference counting to use ref counting APIs

anrossi:
## Description

Update `QUIC_CONNECTION` RefCount to use `CXPLAT_REF_COUNT` and also update the detailed ref counts to use the `CXPLAT_REF_COUNT` type.  Related to #5570 

## Testing

CI

## Documentation

N/A

### DIFF
--- a/src/inc/msquic_winkernel.h
+++ b/src/inc/msquic_winkernel.h
@@ -25,6 +25,7 @@ Environment:
 #include <ntstatus.h>
 #include <basetsd.h>
 #include <netioddk.h>
+#include <ntintsafe.h>
 
 #define QUIC_INLINE inline
 
@@ -38,7 +39,6 @@ typedef UINT16 uint16_t;
 typedef UINT32 uint32_t;
 typedef UINT64 uint64_t;
 
-#define INT32_MAX   0x7fffffffi32
 #define UINT8_MAX   0xffui8
 #define UINT16_MAX  0xffffui16
 #define UINT32_MAX  0xffffffffui32


src/core/connection.h
```diff
@@ -1106,21 +1108,24 @@ QuicConnRelease(
...
         }
 #endif
         if (Ref == QUIC_CONN_REF_LOOKUP_RESULT) {
+            //
+            // The ref count has hit zero and we need to reset to 1 to avoid
+            // hitting an assert when the next reference is taken.
+            //
+            CxPlatRefInitialize(&Connection->RefCount);
             //
             // Lookup results cannot be the last ref, as they can result in the
```
mtfriesen, line 1121: This is confusing - the comment below seems to directly contradict this new comment: "Lookup results cannot be the last ref". How could the refcount be zero here?

mtfriesen: Oh, maybe I'm misinterpreting the comment. Either way, this changes the code to increase the refcount by 1, and I don't see where a corresponding change is to decrease the refcount by 1 later?

anrossi: oh, I see it now. Hmm, this will be tricky to solve...

src/plugins/dbg/quictypes.h
```diff
@@ -97,6 +97,12 @@ typedef union QUIC_CONNECTION_STATE {
         BOOLEAN HandleClosed    : 1;    // Handle closed by application layer.
         BOOLEAN Freed           : 1;    // Freed. Used for Debugging.
         BOOLEAN Partitioned     : 1;    // The connection cannot move across partitions.
+        BOOLEAN CloseAsync      : 1;    // The connection will close without waiting for callbacks.
```
guhetier, line 100: Good catch... I wish we had a way to enforce coherence :/

src/core/connection.h
```diff
@@ -43,6 +43,11 @@ typedef union QUIC_CONNECTION_STATE {
         BOOLEAN Partitioned     : 1;    // The connection cannot move across partitions.
         BOOLEAN CloseAsync      : 1;    // The connection will close without waiting for callbacks.
 
+        //
+        // The connection's ref count has hit zero, but needs to be cleaned up on a worker.
+        //
+        BOOLEAN CleanupStarted  : 1;
```
guhetier: nit: `Cleanup` feels slightly unclear when we have many steps to end a connection (close / shutdown / free). Maybe `FreePending` would not allow confusion?

src/core/connection.h
```diff
@@ -1088,7 +1093,15 @@ QuicConnAddRef(
     UNREFERENCED_PARAMETER(Ref);
 #endif
 
-    CxPlatRefIncrement(&Connection->RefCount);
+    if (!Connection->State.CleanupStarted) {
```
guhetier: I think we should have this check in the caller instead.
There is only a single place in the code where we should be trying to add a ref to a connection that reached a refcount of zero.
Any other place would be a pretty bad bug I think we should catch even in release builds.

src/core/connection.h
mtfriesen: I prefer `CleanupStarted`: its ambiguity is a feature, not a bug. `FreePending` is something that is either always true for any allocated thing, in a general interpretation, or else is falsely precise, in that the code setting this expects more work to be done than calling `free`.

src/core/connection.h
mtfriesen: Agreed, this is a special case warranting special code in a special place.

src/plugins/dbg/quictypes.h
mtfriesen: (Or prevent duplication)

guhetier: Side project for when I have too much time - we could probably have a special comment in code to let a script know it needs to copy the type and generate the dbg header from that.

src/core/connection.h
anrossi: Moved to the callsite

src/core/connection.h
anrossi: I am going to stick with `CleanupStarted` this time.

src/core/worker.c
```diff
@@ -271,7 +271,19 @@ QuicWorkerQueueConnection(
...
+            // If the Connection is in the CleanupStarted state, its ref count
+            // is zero and it is queued here for freeing.
+            // Only increment the detailed ref count in debug builds.
+            // Just don't increment the main ref count in this state.
+            //
+            CxPlatRefIncrement(&Connection->RefTypeBiasedCount[QUIC_CONN_REF_WORKER]);
+#endif
+        } else {
+            QuicConnAddRef(Connection, QUIC_CONN_REF_WORKER);
+        }
```
guhetier: nit:
To make this easier to understand, I recommend starting with the golden path (here, what happen in non-debug scenarios). It also avoid an empty `if` body that some checks don't like.


```suggestion
        if (!Connection->State.CleanupStarted) {
            //
            // Only take a ref if the Connection is not in the CleanupStarted state.
            // It is only queued for freeing otherwise with a ref count of zero.
            //
            QuicConnAddRef(Connection, QUIC_CONN_REF_WORKER);
        }
#if DEBUG
        else {
            //
            // Only increment the detailed ref count.
            // 
            CxPlatRefIncrement(&Connection->RefTypeBiasedCount[QUIC_CONN_REF_WORKER]);
        }
#endif
```

And after writing that, I also think we should not increase the debug ref count because since we don't have a ref count. It would be confusing, the sum of the details would no longer match the refcount.

anrossi: The debug add ref is there because without it, then the final decrement on the connection will make the debug ref go below 1.

guhetier: That makes it sound like it was decremented somewhere it should not have been.
`\sum_0^n(RefTypeBiasedCount[i] - 1) == RefCount` should be an invariant at any time.

anrossi: ok, I'll rework things so that's not yet another thing to work around

anrossi: I rearranged the `QuicConnRelease` function to not need this special increment. There's probably room for further optimization (for conciseness or readability). I just don't have a handy pattern memorized for this.

src/core/connection.h
```diff
@@ -1112,13 +1112,28 @@ QuicConnRelease(
 {
     QuicConnValidate(Connection);
 
+    BOOLEAN LastRef = FALSE;
+    QUIC_CONNECTION_STATE State = {0};
+    State.Flags = CxPlatReadNoFence64((volatile const int64_t*)&Connection->State.Flags);
```
mtfriesen: I'm not sure we should start adding volatile accessors to some subset of flags / code paths within the connection. If we need special synchronization for one or two flags, let's split them out of this bitfield; otherwise, let's make sure the synchronization requirements are coherent and consistently applied across all the code.

anrossi: I talked this over with Guillaume and looked at the other flags; this is not the first flag which is used concurrently. I'll leave this flag unsynchronized in this PR but will make a task to fix access to the flags which need synchronization, so we're consistent

anrossi: ok, this flag approach was exposing a race condition between two threads in quicspin.  At this point, I think it's better to not touch it if it isn't broken, and only fix the per-source references to use ref count API.

src/core/connection.h
```diff
@@ -1106,20 +1109,13 @@ QuicConnRelease(
     QuicConnValidate(Connection);
 
 #if DEBUG
-    CXPLAT_TEL_ASSERT(Connection->RefTypeCount[Ref] > 0);
-    uint16_t result = (uint16_t)InterlockedDecrement16((volatile short*)&Connection->RefTypeCount[Ref]);
-    CXPLAT_TEL_ASSERT(result != UINT16_MAX);
+    CXPLAT_TEL_ASSERT(!CxPlatRefDecrement(&Connection->RefTypeBiasedCount[Ref]));
```
guhetier, line 1112: Should this simply be a debug assert, instead of a TEL assert in a debug block?

anrossi: Ah it probably should have been. But TEL asserts are just debug asserts

guhetier: Let's follow up on that when we have an opportunity, not a big deal otherwise.

Issue #5346, opened 2025-08-11
[CI-TEST] SendQtip failed


guhetier:
AppData/WithSendArgs1.SendQtip/272 failed

```
[ RUN      ] AppData/WithSendArgs1.SendQtip/272
C:\a\msquic\msquic\src\test\lib\TestConnection.cpp(864): error: Unexpected transport Close Error, expected=0x0, actual=0x80410003
C:\a\msquic\msquic\src\test\lib\DataTest.cpp(146): error: Not all data sent.
C:\a\msquic\msquic\src\test\lib\DataTest.cpp(149): error: Not all data received.
C:\a\msquic\msquic\src\test\lib\DataTest.cpp(146): error: Not all data sent.
C:\a\msquic\msquic\src\test\lib\DataTest.cpp(149): error: Not all data received.
C:\a\msquic\msquic\src\test\lib\DataTest.cpp(231): error: Connection->GetTransportClosed() not false
C:\a\msquic\msquic\src\test\lib\DataTest.cpp(599): error: Wait for server to complete timed out after 3000 ms.
[  FAILED  ] AppData/WithSendArgs1.SendQtip/272, where GetParam() = v6/0/4/2/NoSendBuffer/Bidi/Client (3090 ms)
```

Link to failure:
https://github.com/microsoft/msquic/actions/runs/16885870731/job/47833919388

ProjectsByJackHe:
This is related to https://github.com/microsoft/xdp-for-windows/issues/876

guhetier:
```
1455 2025-08-11T16:47:43.132 0E90.033C e [Microsoft-Quic] [sock] Failed to create raw socket, status:-2147014848
1457 2025-08-11T16:47:43.132 0E90.033C e [Microsoft-Quic] [bind][0x00000272BF2AA9F0] ERROR, Binding ephemeral port reuse encountered.
1461 2025-08-11T16:47:43.133 0E90.033C e [Microsoft-Quic] [conn][0x00000272BF2047F0] Indicating QUIC_CONNECTION_EVENT_SHUTDOWN_INITIATED_BY_TRANSPORT [0x80410003]
1462 2025-08-11T16:47:43.133 0E90.033C e [Microsoft-Quic] [test] FAILURE - C:\a\msquic\msquic\src\test\lib\TestConnection.cpp:864 - Unexpected transport Close Error, expected=0x0, actua
```

```
[...]

PR #5586, opened 2025-11-13
Increase log level for QTIP BVT

guhetier:
## Description

Increase the log level for QTIP BVT.
To investigate #5346, OS logs are needed, and the issue is infrequent enough it didn't reproduce on manual pipeline re-run.

This change is meant to be temporary for the duration of the investigation.

## Testing

CI.

## Documentation

N/A

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -147,19 +147,19 @@ jobs:
           { config: "Debug", plat: "linux", os: "ubuntu-22.04", arch: "x64", tls: "openssl", sanitize: "-SanitizeAddress", build: "-Test"  },
           { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "schannel", sanitize: "-SanitizeAddress", build: "-Test" },
           { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "schannel", xdp: "-UseXdp", sanitize: "-SanitizeAddress", build: "-Test" },
-          { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "schannel", xdp: "-UseXdp", qtip: "-UseQtip", sanitize: "-SanitizeAddress", build: "-Test" },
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -147,19 +147,19 @@ jobs:
...
           { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "quictls", build: "-Test" },
           { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "quictls", xdp: "-UseXdp", build: "-Test" },
-          { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "quictls", xdp: "-UseXdp", qtip: "-UseQtip", build: "-Test" },
+          { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "quictls", xdp: "-UseXdp", qtip: "-UseQtip", build: "-Test", log: "Full.Verbose" },
           { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "openssl", build: "-Test" },
           { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "openssl", xdp: "-UseXdp", build: "-Test" },
-          { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "openssl", xdp: "-UseXdp", qtip: "-UseQtip", build: "-Test" },
+          { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "openssl", xdp: "-UseXdp", qtip: "-UseQtip", build: "-Test", log: "Full.Verbose" },
```
mtfriesen, line 156: Should these just always be Full.Verbose? Will we have to update each matrix entry each time that test starts failing?

guhetier: I considered it - but most of the time, test issues we debug only need the MsQuic logs and keeping the verbosity lower makes it easier to navigate the logs without having multi-gigabyte files.

If needing OS logs becomes a pattern, we can make it a default.

mtfriesen: Yeah, makes sense. It's always a trade-off.

PR #5597, opened 2025-11-18
Fix double deref in connection pool error path

### LINKED ISSUE #5550: [CI - FAILURE] Handshake/WithHandshakeArgs12.ConnectionPoolCreate failed hitting an assertion
https://github.com/microsoft/msquic/actions/runs/18915645862/job/53999007778

```
[10/29/2025 17:14:17] Handshake/WithHandshakeArgs12.ConnectionPoolCreate/8 succeeded (in 0.2291524 sec)
[10/29/2025 17:14:17] Process had nonzero exit code: -1073740768
Error: [10/29/2025 17:14:17] Handshake/WithHandshakeArgs12.ConnectionPoolCreate/9 failed (in 0.2247218 sec):
Note: Google Test filter = Handshake/WithHandshakeArgs12.ConnectionPoolCreate/9
[----------] 1 test from Handshake/WithHandshakeArgs12
[ RUN      ] Handshake/WithHandshakeArgs12.ConnectionPoolCreate/9

[10/29/2025 17:14:19] Process had nonzero exit code: -1073740768
Error: [10/29/2025 17:14:19] Handshake/WithHandshakeArgs12.ConnectionPoolCreate/9 failed (in 0.2932292 sec):
Note: Google Test filter = Handshake/WithHandshakeArgs12.ConnectionPoolCreate/9
[----------] 1 test from Handshake/WithHandshakeArgs12
[ RUN      ] Handshake/WithHandshakeArgs12.ConnectionPoolCreate/9
```

```
[2]0810.1BDC::2025/10/29-17:14:17.540017000 [Microsoft-Quic][ lib] ASSERT, D:\a\msquic\msquic\src\core\connection.h:1064 - !Connection->State.Freed.
```

Logs have been collected but no dump.

guhetier:
## Description

When creating a connection pool, if a `QuicConnStart` fails:
- the connection was marked as `ExternalOwner` to prevent it from sending notification to the app
    - but this also mean that the closing logic will take care of releasing the owner refcount, since the application is not the owner yet
- the connection was closed using `MsQuicConnectionClose`, which release the refcount of the application

This caused a double release, triggering an assertion.

We should not call APIs from internal call (it makes logging confusing and breaks some assumptions), so queue the connection close manually instead.

Fixes #5550.

## Testing

C/I.
Need to consider if there is a simply way to deterministically test the connection pool failure paths.

## Documentation

N/A

### DIFF
--- a/src/core/connection_pool.c
+++ b/src/core/connection_pool.c
@@ -714,8 +714,8 @@ CleanUpConnections:
         //
         // Close every connection that was created.
         // The application will receive the shutdown notification.
-        // Wait for the task to be complete so that when this function returns to the app,
-        // all connections are closed (since the shutdown notification is visible).
+        // Wait for the task to complete so that when this function returns to the app,
+        // all connections are already closed (since the shutdown notification is visible).
         //
         for (uint32_t i = 0; i < CreatedConnections; i++) {
             QuicConnPoolQueueConnectionClose(Connections[i], TRUE);


src/core/connection_pool.c
```diff
@@ -236,6 +236,39 @@ QuicConnPoolGetInterfaceIndexForLocalAddress(
...
+    _In_ const BOOLEAN WaitForCompletion
+    )
+{
+    Connection->CloseOper.Type = QUIC_OPER_TYPE_API_CALL;
+    Connection->CloseOper.FreeAfterProcess = FALSE;
+    Connection->CloseOper.API_CALL.Context = &Connection->CloseApiContext;
+    Connection->CloseApiContext.Type = QUIC_API_TYPE_CONN_CLOSE;
+    Connection->CloseApiContext.Status = NULL;
+
+    CXPLAT_EVENT CompletionEvent = {0};
```
mtfriesen: Style: declare local variables at top of function.

src/core/connection_pool.c
```diff
@@ -673,13 +707,19 @@ MsQuicConnectionPoolCreate(
...
+    if (QUIC_FAILED(Status) &&
+        (Config->Flags & QUIC_CONNECTION_POOL_FLAG_CLOSE_ON_FAILURE) != 0) {
+
+        //
+        // Close every connection that was created.
+        // The application will receive the shutdown notification.
+        //
+        for (uint32_t i = 0; i < CreatedConnections; i++) {
+            QuicConnPoolQueueConnectionClose(Connections[i], TRUE);
+            QuicConnRelease(Connections[i], QUIC_CONN_REF_HANDLE_OWNER);
```
mtfriesen, line 722: This thread waits just so it can release a reference? I also don't entirely understand the reason for this, since `MsQuicConnectionClose` itself has a queue-an-operation-and-wait mode.

guhetier: I am not using `MsQuicConnectionClose` to not call an API from internal code. That's what essentially cause the initial issue + logging gets confusing when you have "[api]" logs that are actually not the app calling an API.

The wait is to ensure that on failure, all the connections are closed before we return from the `MsQuicConnectionPoolCreate`. Since connection start / connection shutdown are sent to the app for these connections, it is a visible behavior.

I personally think that we should remove the `QUIC_CONNECTION_POOL_FLAG_CLOSE_ON_FAILURE` flag before getting this out of preview because it makes it complicated to know which connection need to be closed by the app after receiving a shutdown notification.

src/core/connection_pool.c
```diff
@@ -236,6 +236,39 @@
...
+    Connection->CloseOper.Type = QUIC_OPER_TYPE_API_CALL;
+    Connection->CloseOper.FreeAfterProcess = FALSE;
+    Connection->CloseOper.API_CALL.Context = &Connection->CloseApiContext;
+    Connection->CloseApiContext.Type = QUIC_API_TYPE_CONN_CLOSE;
+    Connection->CloseApiContext.Status = NULL;
+
+    CXPLAT_EVENT CompletionEvent = {0};
+    if (WaitForCompletion) {
+        CxPlatEventInitialize(&CompletionEvent, TRUE, FALSE);
+        Connection->CloseApiContext.Completed = &CompletionEvent;
```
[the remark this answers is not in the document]
guhetier: This is safe because we wait for the task to be completed before exit the scope.

src/core/connection_pool.c
mtfriesen: Thanks, agreed with all of the above - let's add some comments. (Avoiding calling `MsQuicXxx` we can leave as implicitly a bad thing, though I was unaware of the rule.)

anrossi (approved):

PR #5598, opened 2025-11-18
Add global Stream tracker in debug builds

anrossi:
## Description

Help debug lost reference issues in MsQuic by keeping a global list of streams.

## Testing

CI

## Documentation

N/A

### DIFF
--- a/src/core/library.c
+++ b/src/core/library.c
@@ -61,6 +61,9 @@ MsQuicLibraryLoad(
         CxPlatSystemLoad();
         CxPlatLockInitialize(&MsQuicLib.Lock);
         CxPlatDispatchLockInitialize(&MsQuicLib.DatapathLock);
+#if DEBUG
+        CxPlatDispatchLockInitialize(&QuicStreamTrackerLock);
+#endif
         CxPlatListInitializeHead(&MsQuicLib.Registrations);
         CxPlatListInitializeHead(&MsQuicLib.Bindings);
         QuicTraceRundownCallback = QuicTraceRundown;
@@ -87,6 +90,9 @@ MsQuicLibraryUnload(
         QUIC_LIB_VERIFY(MsQuicLib.OpenRefCount == 0);
         QUIC_LIB_VERIFY(!MsQuicLib.InUse);
         MsQuicLib.Loaded = FALSE;
+#if DEBUG
+        CxPlatDispatchLockUninitialize(&QuicStreamTrackerLock);
+#endif
         CxPlatDispatchLockUninitialize(&MsQuicLib.DatapathLock);
         CxPlatLockUninitialize(&MsQuicLib.Lock);
         CxPlatSystemUnload();
--- a/src/core/stream.c
+++ b/src/core/stream.c
@@ -15,6 +15,14 @@ Abstract:
 #include "stream.c.clog.h"
 #endif
 
+#if DEBUG
+//
+// Global stream object tracker for debugging.
+//
+CXPLAT_LIST_ENTRY QuicStreamTrackerList = { &QuicStreamTrackerList, &QuicStreamTrackerList };
+CXPLAT_DISPATCH_LOCK QuicStreamTrackerLock;
+#endif
+
 _IRQL_requires_max_(DISPATCH_LEVEL)
 QUIC_STATUS
 QuicStreamInitialize(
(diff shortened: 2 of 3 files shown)


src/core/library.c
```diff
@@ -61,6 +61,9 @@ MsQuicLibraryLoad(
         CxPlatSystemLoad();
         CxPlatLockInitialize(&MsQuicLib.Lock);
         CxPlatDispatchLockInitialize(&MsQuicLib.DatapathLock);
+#if DEBUG
+        CxPlatDispatchLockInitialize(&QuicStreamTrackerLock);
```
mtfriesen, line 65: Nit: I'd suggest simplifying this to one lock for all the trackers.

src/core/stream.c
```diff
@@ -15,6 +15,14 @@
 #include "stream.c.clog.h"
 #endif
 
+#if DEBUG
+//
+// Global stream object tracker for debugging.
+//
+CXPLAT_LIST_ENTRY QuicStreamTrackerList = { &QuicStreamTrackerList, &QuicStreamTrackerList };
```
mtfriesen, line 22: I've never seen this pattern before - is there a precedent for it in MsQuic? If not, let's use the init routine.

src/core/stream.c
```diff
@@ -175,6 +187,10 @@ QuicStreamInitialize(
     if (Stream) {
 #if DEBUG
         CXPLAT_DBG_ASSERT(!CxPlatRefDecrement(&Stream->RefTypeBiasedCount[QUIC_STREAM_REF_APP]));
+        CxPlatDispatchLockAcquire(&QuicStreamTrackerLock);
+        CxPlatListEntryRemove(&Stream->TrackerLink);
```
mtfriesen, line 191: I think we may want to introduce some abstractions for these trackers, such that if perf becomes a practical bottleneck on debug builds, we can:

1. Easily add some more fields beyond the list entry
2. Easily replace the single global list with something more sophisticated, such as a per-CPU list

guhetier:
I am not convinced this helps given that `AllStreamsLink` already exists + the debugger extension. The PR is adding / removing to the list in the exact same spot as for `AllStreamsLink`. It seems redundant to me.

mtfriesen:
> I am not convinced this helps given that `AllStreamsLink` already exists + the debugger extension. The PR is adding / removing to the list in the exact same spot as for `AllStreamsLink`. It seems redundant to me.

The debugger extension is fine, but isn't cross platform. We need a way to provide simple debugging steps for all platforms, so having simple, manually-inspectable lists seems the path of least cost and most reliability.

In terms of the stream tracker specifically being redundant, I agree it's perhaps a lower value object if the owning connections can be reliably found. I'd think of this PR more as a litmus test for the general model than a standalone PR that will dramatically change the debugging experience by itself.

guhetier:
> > I am not convinced this helps given that `AllStreamsLink` already exists + the debugger extension. The PR is adding / removing to the list in the exact same spot as for `AllStreamsLink`. It seems redundant to me.
> 
> The debugger extension is fine, but isn't cross platform. We need a way to provide simple debugging steps for all platforms, so having simple, manually-inspectable lists seems the path of least cost and most reliability.
> 
> In terms of the stream tracker specifically being redundant, I agree it's perhaps a lower value object if the owning connections can be reliably found. I'd think of this PR more as a litmus test for the general model than a standalone PR that will dramatically change the debugging experience by itself.

Good point for the debugger extension. It still feels redundant to me, or at least, I have a hard time thinking of a scenario where this global list brings much value over the per-connection list.

If there are scenarios, it would be good to extend our trouble shooting guide (or have another doc file focused on debugger) with a list of those debug helpers and when they are relevant. It will be easier to do as we go.

mtfriesen:
> Good point for the debugger extension. It still feels redundant to me, or at least, I have a hard time thinking of a scenario where this global list brings much value over the per-connection list.
> 
> If there are scenarios, it would be good to extend our trouble shooting guide (or have another doc file focused on debugger) with a list of those debug helpers and when they are relevant. It will be easier to do as we go.

I stumbled onto a concrete example of why we need the simple, robust object tracker: https://github.com/microsoft/msquic/blob/9117f56038d4470e91708599d72a8bce08304a6b/src/core/connection.c#L335-L344

We can see that following the chain of references here by transitivity will fail, more or less by the existing design, and practically necessarily so, if expected object lifetime relationships break down. Since the lifetimes breaking down is exactly one interesting use case, splitting this into separate lists seems best.

PR #5603, opened 2025-11-19
Change Dependabot update schedule to weekly

anrossi:
## Description

Update Dependabot to update docker images weekly

## Testing

N/A

## Documentation

N/A

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -38,6 +38,7 @@ updates:
       - "/.docker/ubuntu-24.04"
     schedule:
       interval: "weekly"
+      day: "saturday"
     ignore:
       - dependency-name: "ubuntu"
         update-types: ["version-update:semver-major", "version-update:semver-minor"]


.github/dependabot.yml
```diff
@@ -37,7 +37,7 @@ updates:
       - "/.docker/ubuntu-22.04"
       - "/.docker/ubuntu-24.04"
     schedule:
-      interval: "monthly"
+      interval: "weekly"
```
guhetier, line 40: I think that for weekly, we can specify the day of the week and it is set to `      day: "saturday"` for other config.

We should do the same, it is convenient to go through all the dependabot PRs at once on Monday.

mtfriesen: shall we do the github actions config above at the same time? it would be nice to have a uniform policy

anrossi: Let's leave the github actions at the monthly cadence. They were generating a lot of noise and low-value work at the weekly cadence.

PR #5604, opened 2025-11-20
Fix the Key update test

### LINKED ISSUE #3463: QuicTestKeyUpdate test sometimes fail
### Describe the bug

Sometimes QuicTestKeyUpdate test fail
https://dev.azure.com/ms/msquic/_build/results?buildId=431415&view=logs&j=3667e279-20cb-5b66-4f7d-50a166799211&t=69945233-3564-536d-0950-ae97d4372f33&l=1480

maybe windows only?

### Affected OS

- [X] Windows
- [ ] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

_No response_

### MsQuic version

main

.\artifacts\bin\windows\x64_Debug_schannel\msquictest.exe --gtest_filter="Misc/WithKeyUpdateArgs1.KeyUpdate/*"

### Expected behavior

Success

### Actual outcome

Sometimes Failed (feels 0.0001% level on local machine, more frequently on CI)

### Additional details

_No response_

guhetier:
## Description

The "KeyUpdate" test is sometime failing spuriously.

This was due to a sequence of frames where, after a client initiated key update, it would not need to send a frame in response to the server acknowledging the key update. Then, the server would still be waiting for the key update confirmation when the test tries to initiate a key update from the server side.
This behavior follows the RFC, there is no need to acknowledge a key change a packet needs to be sent for another reason.

The test was relying on the assumption that no frames would be sent without a test action, which is not true in general (only most of the time). It was also mixing two largely independent test scenarios in one test, making it hard to understand.

This PR:
- split the test in two based on the two scenarios validated (force updates / trigger updates by sending packets)
- refactor the test, adding helpers, removing non-needed scopes
- fix the issue by ensuring both the client and the server send a frame to acknowledge a key update.

Fixes #3463.

## Testing

C/I

## Documentation

N/A.

### DIFF
--- a/src/test/MsQuicTests.h
+++ b/src/test/MsQuicTests.h
@@ -9,6 +9,12 @@ Abstract:
 
 --*/
 
+//
+// For official releases, disable tests for preview features:
+// the official release test binary for a version is used for down-level compatibility testing
+// against newer versions of MsQuic, and since preview features are subject to change, those tests
+// would break.
+//
 #ifndef QUIC_OFFICIAL_RELEASE
 #define QUIC_API_ENABLE_PREVIEW_FEATURES
 #endif
@@ -94,6 +100,7 @@ void QuicTestValidateStreamEvents(uint32_t Test);
 // Basic Functionality Tests
 //
 
+void QuicTestRegistrationOpenClose();
 void QuicTestCreateListener();
 void QuicTestStartListener();
 void QuicTestStartListenerMultiAlpns();
@@ -1401,8 +1408,11 @@ struct QUIC_RUN_CONNECTION_POOL_CREATE_PARAMS {
 #define IOCTL_QUIC_RUN_VALIDATE_PARTITION \
     QUIC_CTL_CODE(136, METHOD_BUFFERED, FILE_WRITE_DATA)
 
-#define IOCTL_QUIC_RUN_FORCE_KEY_UPDATE \
+#define IOCTL_QUIC_RUN_REGISTRATION_OPEN_CLOSE \
     QUIC_CTL_CODE(137, METHOD_BUFFERED, FILE_WRITE_DATA)
+    
+#define IOCTL_QUIC_RUN_FORCE_KEY_UPDATE \
+    QUIC_CTL_CODE(138, METHOD_BUFFERED, FILE_WRITE_DATA)
     // int - Family
 
-#define QUIC_MAX_IOCTL_FUNC_CODE 137
+#define QUIC_MAX_IOCTL_FUNC_CODE 138
--- a/src/test/bin/winkernel/control.cpp
+++ b/src/test/bin/winkernel/control.cpp
@@ -535,6 +535,7 @@ size_t QUIC_IOCTL_BUFFER_SIZES[] =
     0,
     0,
     0,
+    0,
     sizeof(INT32),
 };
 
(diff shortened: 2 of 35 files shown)


mtfriesen (approved):

src/test/bin/winkernel/control.cpp
```diff
@@ -426,7 +426,7 @@ size_t QUIC_IOCTL_BUFFER_SIZES[] =
     sizeof(uint32_t),
     sizeof(uint32_t),
     sizeof(INT32),
-    sizeof(QUIC_RUN_KEY_UPDATE_PARAMS),
+    sizeof(INT32),
```
anrossi, line 429: Why is the `Family` parameter a 4-byte int?

src/test/MsQuicTests.h
```diff
@@ -1415,4 +1401,8 @@ struct QUIC_RUN_CONNECTION_POOL_CREATE_PARAMS {
 #define IOCTL_QUIC_RUN_VALIDATE_PARTITION \
     QUIC_CTL_CODE(136, METHOD_BUFFERED, FILE_WRITE_DATA)
 
-#define QUIC_MAX_IOCTL_FUNC_CODE 136
+#define IOCTL_QUIC_RUN_FORCE_KEY_UPDATE \
+    QUIC_CTL_CODE(137, METHOD_BUFFERED, FILE_WRITE_DATA)
+    // int - Family
+
+#define QUIC_MAX_IOCTL_FUNC_CODE 137
```
anrossi: Did your editor delete the trailing new line that git so loves to have at the end of every file?

src/test/lib/HandshakeTest.cpp
```diff
@@ -2385,259 +2407,333 @@ QuicTestKeyUpdateRandomLoss(
...
-                    TEST_FAILURE("%llu client packets failed to decrypt!", Stats.RecvDecryptionFailures);
-                    return;
-                }
+        TEST_QUIC_SUCCEEDED(SendFrame(*Server, Client));
 
-                if (Stats.KeyUpdateCount < 1) {
-                    TEST_FAILURE("%u Key updates occured. Expected at least 1", Stats.KeyUpdateCount);
-                    return;
-                }
+        CxPlatSleep(50);
```
anrossi, line 2474: How sensitive is this to VM timing?

src/test/bin/winkernel/control.cpp
guhetier: That's how it is defined in `QUIC_IOCTL_PARAMS`, I am just applying the pattern here.

I wish we could avoid that list, and may refactor that dispatch code if I find a clean solution.

src/test/MsQuicTests.h
guhetier: I can add one.

Git doesn't care about whether there is a newline at the end or not, it simply tells you so that you know if you care. Some people care because some Unix tools display files poorly when they don't end with a newline.

src/test/lib/HandshakeTest.cpp
guhetier: This 50 ms sleep was already there (line 2455 of the old code).

If anything, I would just remove it. `SendFrame` should have made sure everything is synchronized and sleep in tests are just slowing things down without reliability. But I tried to keep the change minimal.

anrossi (approved):

src/test/lib/HandshakeTest.cpp
anrossi: ok, no worries then.

Issue #5369, opened 2025-08-20
MSQuic will stall in OOM conditions on Windows


fbrosseau:
### Describe the bug

All calls to `CxPlatEventQEnqueue/Ex` must check for return value, otherwise MSQuic will stall.

At first glance, I can find these problematic places:
```
Q:\src\msquic\src\platform\datapath_raw_xdp_win.c(1951):CxPlatEventQEnqueue(Partition->EventQ, &Partition->ShutdownSqe);
Q:\src\msquic\src\platform\platform_worker.c(228):CxPlatEventQEnqueue(&Worker->EventQ, &Worker->ShutdownSqe);
Q:\src\msquic\src\platform\platform_worker.c(504):CxPlatEventQEnqueue(&Worker->EventQ, &Worker->UpdatePollSqe);
Q:\src\msquic\src\platform\platform_worker.c(515):CxPlatEventQEnqueue(&Worker->EventQ, &Worker->WakeSqe);
```

I am going to write this bug report all in terms of Windows, but I would assume that it applies to some of the other implementations as well.

`CxPlatEventQEnqueue`, aka `PostQueuedCompletionStatus`, can fail for two reasons:
1- The IOCP was unexpectedly closed - this would be a program error.
2- OOM. The reason for this is that `NtSetIoCompletion` underneath PQCS must allocate an IO completion packet to add to the target IOCP *.

It would not be easy to manually repro this, but in complex programs that like to live on the edge of 100% memory (like a database), this will happen 100%.

* If we want to get technical, it is possible to leverage "reserve objects" through `NtAllocateReserveObject` and eliminate the possibility of OOM entirely by pre-creating the reusable IO packet. Not only does this remove the problem, but it is also an optimization. However, MSQuic would have to get OS folks to document it to MSDN it in order to be able to use it.

### Affected OS

- [x] Windows
- [ ] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

All Windows versions

### MsQuic version

main

### Steps taken to reproduce bug

Be at memory exhaustion levels
CxPlatEventQEnqueue

### Expected behavior

N/A

### Actual outcome

Error ignored

PR #5617, opened 2025-11-21
Added failure path in case of Event Enqueue fails

Santhosha-bk:
## Description
CxPlatEventQEnqueue, aka PostQueuedCompletionStatus, can fail because of OOM. The reason for this is that NtSetIoCompletion underneath PQCS must allocate an IO completion packet to add to the target IOCP *.

Same applies in case of epoll or kqueue io_uring.

Added a Wake Completion Packet (WCP) mechanism as backup incase of PostQueuedCompletionStatus fails.
#5369 

## Testing

## Documentation

### DIFF
--- a/src/inc/quic_platform_winuser.h
+++ b/src/inc/quic_platform_winuser.h
@@ -777,10 +777,14 @@ typedef struct CXPLAT_SQE {
 #if DEBUG
     BOOLEAN IsQueued; // Debug flag to catch double queueing.
 #endif
-    HANDLE WcpEvent;      // Manual-reset event for wake packets
-    HANDLE WaitCompletionPacket;    // Wait completion packet bound to Event
 } CXPLAT_SQE;
 
+// Extended SQE with Wait Completion Packet support for manual events
+typedef struct CXPLAT_SQE_WCP {
+    CXPLAT_SQE BaseSqe;   // Base SQE must be first for correct casting.
+    HANDLE WcpEvent;      // Manual-reset event for wake packets
+    HANDLE WaitCompletionPacket;    // Wait completion packet bound to Event
+} CXPLAT_SQE_WCP;
 //
 // Wait Completion Packet functions from ntdll.dll.
 //
@@ -807,9 +811,19 @@ typedef NTSTATUS (NTAPI *FuncNtCancelWaitCompletionPacket)(
     );
 
 // Global function pointers (initialized in CxPlatInitialize)
-extern FuncNtCreateWaitCompletionPacket CxPlatNtCreateWaitCompletionPacket;
-extern FuncNtAssociateWaitCompletionPacket CxPlatNtAssociateWaitCompletionPacket;
-extern FuncNtCancelWaitCompletionPacket CxPlatNtCancelWaitCompletionPacket;
+extern FuncNtCreateWaitCompletionPacket NtCreateWaitCompletionPacket;
+extern FuncNtAssociateWaitCompletionPacket NtAssociateWaitCompletionPacket;
+extern FuncNtCancelWaitCompletionPacket NtCancelWaitCompletionPacket;
+
+// Helper to check if WCP APIs are available
+QUIC_INLINE
+BOOLEAN
+CxPlatWcpAvailable(void)
+{
+    return NtCreateWaitCompletionPacket != NULL &&
+           NtAssociateWaitCompletionPacket != NULL &&
+           NtCancelWaitCompletionPacket != NULL;
+}
 
 QUIC_INLINE
 BOOLEAN
@@ -851,15 +865,7 @@ CxPlatEventQEnqueue(
     sqe->IsQueued;
 #endif
     CxPlatZeroMemory(&sqe->Overlapped, sizeof(sqe->Overlapped));
-    if (!PostQueuedCompletionStatus(*queue, 0, 0, &sqe->Overlapped)){
-        if (sqe->WcpEvent) {
-            CxPlatEventSet(sqe->WcpEvent);
-            return TRUE;
-        } else {
-            return FALSE;
-        }
-    }
-    return TRUE;
+    return PostQueuedCompletionStatus(*queue, 0, 0, &sqe->Overlapped) != 0;
 }
 
 QUIC_INLINE
@@ -876,6 +882,22 @@ CxPlatEventQEnqueueEx( // Windows specific extension
 #endif
     CxPlatZeroMemory(&sqe->Overlapped, sizeof(sqe->Overlapped));
     return PostQueuedCompletionStatus(*queue, num_bytes, 0, &sqe->Overlapped) != 0;
+}
+
+// Enqueue WCP-based SQE to the event queue.
+// Falls back to standard PQCS if WCP is unavailable.
+QUIC_INLINE
+void
+CxPlatEventQEnqueueWcp(
+    _In_ CXPLAT_EVENTQ* queue,
+    _In_ CXPLAT_SQE_WCP* sqe
+    )
+{
+    if (sqe->WcpEvent) {
+        CxPlatEventSet(sqe->WcpEvent);
+    } else {
+        CxPlatEventQEnqueue(queue, &sqe->BaseSqe);
+    }
 }
 
 QUIC_INLINE
@@ -914,6 +936,7 @@ CxPlatEventQReturn(
     UNREFERENCED_PARAMETER(count);
 }
 
+// Only WakeSqe and UpdatePollSqe will be using this.
 QUIC_INLINE
 BOOLEAN
 CxPlatSqeInitialize(
@@ -925,6 +948,28 @@ CxPlatSqeInitialize(
     UNREFERENCED_PARAMETER(queue);
     CxPlatZeroMemory(sqe, sizeof(*sqe));
     sqe->Completion = completion;
+    return TRUE;
+}
+
+// Currently only ShutdownSqe will be using this, but it can be used by any SQE that wants to take advantage of WCP for manual events.
+QUIC_INLINE
+BOOLEAN
+CxPlatSqeInitializeWcp(
+    _In_ CXPLAT_EVENTQ* queue,
+    _In_ CXPLAT_EVENT_COMPLETION completion,
+    _Out_ CXPLAT_SQE_WCP* sqe
+    )
+{
+    UNREFERENCED_PARAMETER(queue);
+    CxPlatZeroMemory(sqe, sizeof(*sqe));
+    sqe->BaseSqe.Completion = completion;
+
+    // Only attempt WCP initialization if APIs are available
+    if (!CxPlatWcpAvailable()) {
+        // WCP not available - this is acceptable on older Windows versions
+        // SQE will fall back to standard PQCS (with OOM risk)
+        return TRUE;
+    }
 
     if (sqe->WcpEvent == NULL) {
         CxPlatEventInitialize(&sqe->WcpEvent, TRUE, FALSE);
@@ -933,7 +978,7 @@ CxPlatSqeInitialize(
         return FALSE;
     }
 
-    NTSTATUS status = CxPlatNtCreateWaitCompletionPacket(&sqe->WaitCompletionPacket,
+    NTSTATUS status = NtCreateWaitCompletionPacket(&sqe->WaitCompletionPacket,
                                                   GENERIC_ALL,
                                                   NULL);
     if (!NT_SUCCESS(status)) {
@@ -941,29 +986,34 @@ CxPlatSqeInitialize(
         return FALSE;
     }
 
-    status = CxPlatNtAssociateWaitCompletionPacket(
+    status = NtAssociateWaitCompletionPacket(
         sqe->WaitCompletionPacket,
         *queue,
         sqe->WcpEvent,
         NULL,
-        sqe,
+        &sqe->BaseSqe,
         0,          // IoStatus STATUS_SUCCESS
         0,
         NULL);
 
     if (!NT_SUCCESS(status)) {
-        NTSTATUS CancelStatus = CxPlatNtCancelWaitCompletionPacket(
+        NTSTATUS CancelStatus = NtCancelWaitCompletionPacket(
                 sqe->WaitCompletionPacket,   // WaitCompletionPacketHandle
                 TRUE);              // RemoveSignaledPacket
         CXPLAT_DBG_ASSERT(NT_SUCCESS(CancelStatus));
-        CloseHandle(sqe->WaitCompletionPacket);
-        CloseHandle(sqe->WcpEvent);
+        // Close the Handle Only if the cancel succeeded.
+        // In failure case leak it to avoid a memory corruption.
+        if (NT_SUCCESS(CancelStatus)) { 
+            CloseHandle(sqe->WaitCompletionPacket);
+            CloseHandle(sqe->WcpEvent);
+        }
         return FALSE;
     }
 
     return TRUE;
 }
 
+// Most of the SQEs that are created through CxPlatStartDatapathIo will be using this.
 QUIC_INLINE
 void
 CxPlatSqeInitializeEx(
@@ -985,14 +1035,38 @@ CxPlatSqeCleanup(
     _In_ CXPLAT_SQE* sqe
     )
 {
+    UNREFERENCED_PARAMETER(queue);
+    UNREFERENCED_PARAMETER(sqe);
+    // No-op for base SQE
+}
+
+QUIC_INLINE
+void
+CxPlatSqeCleanupWcp(
+    _In_ CXPLAT_EVENTQ* queue,
+    _In_ CXPLAT_SQE_WCP* sqe
+    )
+{
     if (sqe->WcpEvent)
     {
-        NTSTATUS CancelStatus = CxPlatNtCancelWaitCompletionPacket(
-            sqe->WaitCompletionPacket,   // WaitCompletionPacketHandle
-            TRUE);              // RemoveSignaledPacket
-        CXPLAT_DBG_ASSERT(NT_SUCCESS(CancelStatus));
-        CloseHandle(sqe->WaitCompletionPacket);
-        CloseHandle(sqe->WcpEvent);
+        if (sqe->WaitCompletionPacket) {
+            if (CxPlatWcpAvailable()) {
+                NTSTATUS CancelStatus = NtCancelWaitCompletionPacket(
+                    sqe->WaitCompletionPacket,   // WaitCompletionPacketHandle
+                    TRUE);              // RemoveSignaledPacket
+                CXPLAT_DBG_ASSERT(NT_SUCCESS(CancelStatus));
+                // Close the Handle Only if the cancel succeeded.
+                // In failure case leak it to avoid a memory corruption.
+                if (NT_SUCCESS(CancelStatus)) { 
+                    CloseHandle(sqe->WaitCompletionPacket);
+                    CloseHandle(sqe->WcpEvent);
+                }
+            }
+        } else {
+            CloseHandle(sqe->WcpEvent);
+        }
+        sqe->WcpEvent = NULL;
+        sqe->WaitCompletionPacket = NULL;
     }
     UNREFERENCED_PARAMETER(queue);
     UNREFERENCED_PARAMETER(sqe);
--- a/src/platform/platform_winuser.c
+++ b/src/platform/platform_winuser.c
@@ -37,9 +37,9 @@ QUIC_TRACE_RUNDOWN_CALLBACK* QuicTraceRundownCallback;
 typedef LONG (WINAPI *FuncRtlGetVersion)(RTL_OSVERSIONINFOW *);
 
 // Global function pointers (initialized in CxPlatInitialize)
-FuncNtCreateWaitCompletionPacket CxPlatNtCreateWaitCompletionPacket = NULL;
-FuncNtAssociateWaitCompletionPacket CxPlatNtAssociateWaitCompletionPacket = NULL;
-FuncNtCancelWaitCompletionPacket CxPlatNtCancelWaitCompletionPacket = NULL;
+FuncNtCreateWaitCompletionPacket NtCreateWaitCompletionPacket = NULL;
+FuncNtAssociateWaitCompletionPacket NtAssociateWaitCompletionPacket = NULL;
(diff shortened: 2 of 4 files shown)


src/platform/datapath_epoll.c
```diff
@@ -1021,10 +1021,20 @@ CxPlatSocketContextUninitialize(
...
+            int Errno = errno;
+            QuicTraceEvent(
+                DatapathErrorStatus,
+                "[data][%p] ERROR, %u, %s.",
+                SocketContext->Binding,
+                Errno,
+                "CxPlatEventQEnqueue failed (Shutdown)");
+
+            // Queue can’t run the completion, so do it inline to finish teardown.
+            CxPlatSocketContextUninitializeEventComplete(&SocketContext->ShutdownSqe.Cqe);
```
mtfriesen: I'm not sure we can simply run inline - in at least some of these files, I expect the orderly completion is relied on for serialization / mutual exclusion. Instead, we need to provide a backup mechanism that ensures orderly completion runs even if the underlying OS queue fails.

Santhosha-bk: @mtfriesen  Rather than inline, we can try to re-add the event in queue with a delay in case of failure as a backup mechanism, we can add a 1ms delay and re-add the event like below. Please share your thoughts on this approach.
```
while (!CxPlatEventQEnqueue(&Worker->EventQ, &Worker->ShutdownSqe)) {
            if (++RetryCount >= MaxRetries) {
                QuicTraceLogWarning(
                    WorkerShutdownEnqueueRetry,
                    "[wrkr][%p] Shutdown enqueue failed, retry %u/%u",
                    Worker,
                    RetryCount,
                    MaxRetries);
                break;
            }
            // Exponential backoff: 1ms, 2ms, 4ms, 8ms, ...
            CxPlatSleep(1U << (RetryCount - 1));
 }
```

guhetier: I don't think that re-trying with delay is an acceptable option, and it only pushes back the problem: what happens if it keeps failing, do we hang forever? Or are we back to a fre assert?

It might also be worth checking whether the issue reported in the bug can happen for Unix.
Typically, memory management won't let an allocation failure happen (a process will either get memory or be killed).

Santhosha-bk: I was thinking we could have loop of 5-10 re-tries, but this is not an acceptable solution.

MsQuic stalls only in CxPlatWorkerPoolDestroyWorker because the thread waits for the worker thread to rejoin after a failed CxPlatEventQEnqueue. In all other cases where CxPlatEventQEnqueue is used, this behavior does not occur.

Santhosha-bk: Using WCP backup mechanism for windows

src/inc/quic_platform_winuser.h
```diff
@@ -740,8 +740,44 @@ typedef struct CXPLAT_SQE {
...
+    HANDLE WcpEvent;      // Manual-reset event for wake packets
+    HANDLE WaitCompletionPacket;    // Wait completion packet bound to Event
 } CXPLAT_SQE;
 
+#define IO_WAIT_COMPLETION_PACKET_MODIFY_STATE 0x0001
+
+NTSYSCALLAPI
+NTSTATUS
+NTAPI
+NtCreateWaitCompletionPacket (
```
guhetier: These functions must be loaded with `LoadLibrary` and `GetProcAddress`. The prototypes should be grouped with other similar functions.

src/inc/quic_platform_winuser.h
```diff
@@ -782,7 +818,16 @@ CxPlatEventQEnqueue(
     sqe->IsQueued;
 #endif
     CxPlatZeroMemory(&sqe->Overlapped, sizeof(sqe->Overlapped));
-    return PostQueuedCompletionStatus(*queue, 0, 0, &sqe->Overlapped) != 0;
+    if (!PostQueuedCompletionStatus(*queue, 0, 0, &sqe->Overlapped)){
+        if (sqe->WcpEvent){
+            SetEvent(sqe->WcpEvent);
```
guhetier: nit: Please follow the codebase style for spaces and brace positions.

```c
if (...) {

} else {

}
```

src/inc/quic_platform_winuser.h
```diff
@@ -740,8 +740,44 @@ typedef struct CXPLAT_SQE {
 #if DEBUG
     BOOLEAN IsQueued; // Debug flag to catch double queueing.
 #endif
+    HANDLE WcpEvent;      // Manual-reset event for wake packets
```
guhetier: nit: Lets make this name more explicit. `Wcp` is not a well known acronym.
For instance, something along the line of `WaitCompletinoPacketEvent` would be more helpful.

src/inc/quic_platform_winuser.h
```diff
@@ -740,8 +740,44 @@ typedef struct CXPLAT_SQE {
...
+    HANDLE WaitCompletionPacket;    // Wait completion packet bound to Event
```
guhetier: There is no `Event`, this comment needs to be updated.

It would also be more helpful to have a comment about the "why" than the "what", explaining that these two handles control a wait completion packets allowing to safely wait until the IOCP is fully drained.

src/inc/quic_platform_winuser.h
guhetier:
```suggestion
    if (!GetQueuedCompletionStatusEx(*queue, events, count, &out_count, wait_time, FALSE)) {
        return 0;
    }
```

src/inc/quic_platform_winuser.h
```diff
@@ -846,6 +891,39 @@ CxPlatSqeInitialize(
     UNREFERENCED_PARAMETER(queue);
     CxPlatZeroMemory(sqe, sizeof(*sqe));
     sqe->Completion = completion;
+
+    if (sqe->WcpEvent == NULL) {
+        sqe->WcpEvent = CreateEvent(NULL, TRUE, FALSE, NULL);
```
guhetier: Consider using `CxPlatEventInitialize` instead.

src/inc/quic_platform_winuser.h
guhetier: I don't think this follows the suggestion from the bug.
The idea was not to use WaitCompletionPackets as a backup - associating a wait completion packet and an event to each SQE is not acceptable.

The idea is to replace *some specific uses* of `CxPlatEventQEnqueue`, where the only point of queuing the SQE is to trigger some queue activity and flush it, by a wait completion packet.

For instance, in `CxPlatWorkerPoolDestroyWorker`, instead of calling `CxPlatEventQEnqueue`, we would set a `ShutdownEvent`. This would cause the packet to be added in the IOCP and eventually `GetQueuedCompletionStatusEx` would retrieve the packet and run the callback function.

Some of the code related to the shutdown sqe is in the cross-platform code, so you may have to look at how to abstract it - but in a first time, a windows only solution would be enough. Don't hesitate to reach out if you have questions.

src/inc/quic_platform_winuser.h
```diff
@@ -782,7 +818,16 @@ CxPlatEventQEnqueue(
     sqe->IsQueued;
 #endif
     CxPlatZeroMemory(&sqe->Overlapped, sizeof(sqe->Overlapped));
-    return PostQueuedCompletionStatus(*queue, 0, 0, &sqe->Overlapped) != 0;
+    if (!PostQueuedCompletionStatus(*queue, 0, 0, &sqe->Overlapped)){
+        if (sqe->WcpEvent){
```
guhetier: I don't think using a wait completion packet should ever be a backup: it should be an OR. If we created the SQE with the intent of using a wait completion we should always use it.

src/inc/quic_platform_winuser.h
```diff
@@ -846,6 +891,39 @@ CxPlatSqeInitialize(
     UNREFERENCED_PARAMETER(queue);
     CxPlatZeroMemory(sqe, sizeof(*sqe));
     sqe->Completion = completion;
+
+    if (sqe->WcpEvent == NULL) {
+        sqe->WcpEvent = CreateEvent(NULL, TRUE, FALSE, NULL);
+        CXPLAT_DBG_ASSERT(sqe->WcpEvent != NULL);
+    }
+
+    NTSTATUS status = NtCreateWaitCompletionPacket(&sqe->WaitCompletionPacket,
```
guhetier, line 981: Most SQE are created as part of `CxPlatStartDatapathIo`, and the expectation is that the networking stack will queue the completion (unless they complete inline).

I don't think we should create an event and all the related completion packet setup when we expect the sqe to be completed by the kernel and not explicitly by our code (at least for now...)

src/inc/quic_platform_winuser.h
```diff
@@ -846,6 +891,39 @@ CxPlatSqeInitialize(
...
+    if (!NT_SUCCESS(status)) {
+        CloseHandle(sqe->WcpEvent);
+        return FALSE;
+    }
+
+    status = NtAssociateWaitCompletionPacket(
+        sqe->WaitCompletionPacket,
+        *queue,
+        sqe->WcpEvent,
+        sqe,
```
guhetier: Are you sure the key context is what is going to map to the overlapped when we receive the event? Copilot suggest it should be set as the apc context instead.

src/inc/quic_platform_winuser.h
```diff
@@ -846,6 +891,39 @@ CxPlatSqeInitialize(
...
+        NULL,
+        0,          // IoStatus STATUS_SUCCESS
+        0,
+        NULL);
+
+    if (!NT_SUCCESS(status)) {
+        NtCancelWaitCompletionPacket(
```
guhetier: We should debug assert this doesn't return pending

src/inc/quic_platform_winuser.h
```diff
@@ -870,6 +948,14 @@ CxPlatSqeCleanup(
     _In_ CXPLAT_SQE* sqe
     )
 {
+    if (sqe->WcpEvent)
+    {
+        NtCancelWaitCompletionPacket(
```
guhetier: What if this return pending?

src/inc/quic_platform_winuser.h
```diff
@@ -880,6 +966,9 @@ CxPlatCqeGetSqe(
     _In_ const CXPLAT_CQE* cqe
     )
 {
+    if (cqe->lpOverlapped == NULL) {
+        return (CXPLAT_SQE*)cqe->lpCompletionKey;
```
guhetier: See the comment above - I think that if you set the sqe as the APC context, it will be mapped to `lpOverlapped` here.

src/inc/quic_platform_winuser.h
guhetier: Maybe we should consider having an "extension" to the SQE type so that we pay the price for these extra members only if we are going to use them.

Maybe something like:

```
struct CXPLAT_SQE_WCP {
    CXPLAT_SQE sqe;
    HANDLE Event;
    HANDLE WaitCompletionPacket;
}
```

Not fully sure if it is worth it or not.

mtfriesen: Agreed, I don't think every single SQE needs this

src/inc/quic_platform_winuser.h
```diff
@@ -740,8 +740,44 @@ typedef struct CXPLAT_SQE {
 #if DEBUG
     BOOLEAN IsQueued; // Debug flag to catch double queueing.
 #endif
+    HANDLE WcpEvent;      // Manual-reset event for wake packets
+    HANDLE WaitCompletionPacket;    // Wait completion packet bound to Event
 } CXPLAT_SQE;
 
+#define IO_WAIT_COMPLETION_PACKET_MODIFY_STATE 0x0001
```
mtfriesen: I don't see any public documentation for this flag - we can't use it in OSS.

src/inc/quic_platform_winuser.h
```diff
@@ -846,6 +891,39 @@ CxPlatSqeInitialize(
     UNREFERENCED_PARAMETER(queue);
     CxPlatZeroMemory(sqe, sizeof(*sqe));
     sqe->Completion = completion;
+
+    if (sqe->WcpEvent == NULL) {
+        sqe->WcpEvent = CreateEvent(NULL, TRUE, FALSE, NULL);
+        CXPLAT_DBG_ASSERT(sqe->WcpEvent != NULL);
```
mtfriesen: This isn't necessarily true, AFAIK.

mtfriesen: (Assertions should assert only statements that are necessarily true.)

src/inc/quic_platform_winuser.h
mtfriesen: I'm not sure I understand this comment, but I think I have the same general idea.

The way I'd look at this problem is we need a 100% reliable queue of SQE+CQEs and a reliable mechanism to trigger the main event completion loop to check the reliable queue. It should be sufficient to have one wake event associated with the underlying event queue.

The backup queue also needs to provide ordering with the main queue, because MsQuic requires `CxPlatEventQSubmit(A)`, `CxPlatEventQSubmit(B)` always produce `CQE(A)`, `CQE(B)` in order.

Finally, once we implement the reliable, ordered backup queue, should we prefer it for _all_ non-IO completions? In order words, if we post a million internal messages, is there a reason to round-trip them all through the kernel queue, or should we just round-trip one wake event and use the internal queue for all messaging?

guhetier: What I was trying to say above is that we don't want to call `PostQueuedCompletionStatus`, and to fall back to `WcpEvent` if it fails: we need to know which of the two control the event (based on how it was setup) and call the right one and only that one.

Having our own event queue + a single trigger event might be a good way to simplify it (but will take a bigger refactor, with the risk of re-inventing the wheel?).

I think we can keep this one PR scope on the smaller side by focusing on how to queue the event in the IOCP: we can start by using wait completion events for specific events only. When that works, we can have follow up with our own queue to avoid useless queuing in KM (might be worth perf measurements).

mtfriesen: The problem, though, is that unless we have a single backup wake event and explicit backup queue, then _every_ SQE needs to preallocate the backup event to ensure it is "queueable" upon PQCS failure, which is unknowable a priori.

That is, unless we change the design such that some SQEs are created as "reliable" and most SQEs are initialized as unreliable. I'd rather not introduce the notion of unreliable SQEs, but it may be sufficient, at least for now.

src/inc/quic_platform_winuser.h
Santhosha-bk: Corrected it

src/inc/quic_platform_winuser.h
Santhosha-bk: Done

src/inc/quic_platform_winuser.h
Santhosha-bk: Updated to use CxPlatEventInitialize

src/inc/quic_platform_winuser.h
Santhosha-bk: Done

src/inc/quic_platform_winuser.h
Santhosha-bk: Added the debug assert

src/inc/quic_platform_winuser.h
Santhosha-bk: removed this flag

src/inc/quic_platform_winuser.h
Santhosha-bk: Passing sqe in the APC context, it will be mapped to lpOverlapped

src/inc/quic_platform_winuser.h
Santhosha-bk: There are two different paths to initialize/create SQEs in Windows IOCP.
1. CxPlatSqeInitialize ---> This function gets called to create the SQEs for fake packets, and this is the path where the failure in PostQueuedCompletionStatus because of OOM can happen.
2. CxPlatSqeInitializeEx ---> This function gets called for the case you have mentioned above, these will be real packets, and the expectation is networking stack will queue the completion.

For second case we will not allocate the WcpEvent, only in case of fake packets WcpEvent will be allocated and used as backup mechanism, for the real data packets SQEs from CxPlatStartDatapathIo WcpEvent will be NULL

src/inc/quic_platform_winuser.h
Santhosha-bk: Done

src/inc/quic_platform_winuser.h
```diff
@@ -782,7 +814,15 @@ CxPlatEventQEnqueue(
     sqe->IsQueued;
 #endif
     CxPlatZeroMemory(&sqe->Overlapped, sizeof(sqe->Overlapped));
-    return PostQueuedCompletionStatus(*queue, 0, 0, &sqe->Overlapped) != 0;
+    if (!PostQueuedCompletionStatus(*queue, 0, 0, &sqe->Overlapped)){
```
guhetier: I think using the `WcpEvent` should not be a fallback, but the main and only way for SQE using that mechanism.

Basically:
```
if (sqe->WcpEvent) {
    CxPlatEventSet(sqe->WcpEvent);
    return TRUE;
}
return PostQueuedCompletionStatus(*queue, 0, 0, &sqe->Overlapped) != 0;
```

src/platform/platform_winuser.c
```diff
@@ -36,6 +36,11 @@ QUIC_TRACE_RUNDOWN_CALLBACK* QuicTraceRundownCallback;
 //
 typedef LONG (WINAPI *FuncRtlGetVersion)(RTL_OSVERSIONINFOW *);
 
+// Global function pointers (initialized in CxPlatInitialize)
+FuncNtCreateWaitCompletionPacket CxPlatNtCreateWaitCompletionPacket = NULL;
```
guhetier: Please keep the "original" name for these function, (`NtCreateWaitCompletionPacket`). They are not `CxPlat` functions.

src/inc/quic_platform_winuser.h
```diff
@@ -846,6 +888,42 @@ CxPlatSqeInitialize(
...
     sqe->Completion = completion;
+
+    if (sqe->WcpEvent == NULL) {
+        CxPlatEventInitialize(&sqe->WcpEvent, TRUE, FALSE);
+    }
+    if (sqe->WcpEvent == NULL) {
+        return FALSE;
+    }
+
+    NTSTATUS status = CxPlatNtCreateWaitCompletionPacket(&sqe->WaitCompletionPacket,
```
guhetier: `CxPlatNtCreateWaitCompletionPacket` could be null if loading it failed. We need to check it is non-null before calling it. A small wrapper that check for non-null and return an error code if it is null is also an option.

guhetier: Same for all calls to dynamically loaded functions.

src/inc/quic_platform_winuser.h
guhetier, line 942: You explained offline that we initialize the `WaitCompletionPacket` on this function only (and not `CxPlatSqeInitializeEx`) because that is the one used to initialize the SQE for events we queue manually.

Some comments should be added to explain it. The function should maybe also be renamed to make the distinction clear. This goes in pair with the likely need of two distinct types of SQE struct.

src/inc/quic_platform_winuser.h
```diff
@@ -846,6 +888,42 @@ CxPlatSqeInitialize(
...
+                                                  GENERIC_ALL,
+                                                  NULL);
+    if (!NT_SUCCESS(status)) {
+        CloseHandle(sqe->WcpEvent);
+        return FALSE;
```
guhetier, line 986: This function being able to fail seems to be a new thing, across all implementations. We need to double check failures are handled properly (both handled, and we do the right thing on failure).

src/inc/quic_platform_winuser.h
```diff
@@ -846,6 +888,42 @@ CxPlatSqeInitialize(
...
+        sqe,
+        0,          // IoStatus STATUS_SUCCESS
+        0,
+        NULL);
+
+    if (!NT_SUCCESS(status)) {
+        NTSTATUS CancelStatus = CxPlatNtCancelWaitCompletionPacket(
+                sqe->WaitCompletionPacket,   // WaitCompletionPacketHandle
+                TRUE);              // RemoveSignaledPacket
+        CXPLAT_DBG_ASSERT(NT_SUCCESS(CancelStatus));
```
guhetier, line 1003: Is it possible the cancelation fails? What happens if it fails? Is it still ok to close the handle?
Or should we leak it to avoid a memory corruption? Or abort the process if this is truly unrecoverable?

src/inc/quic_platform_winuser.h
```diff
@@ -870,6 +948,15 @@ CxPlatSqeCleanup(
     _In_ CXPLAT_SQE* sqe
     )
 {
+    if (sqe->WcpEvent)
+    {
+        NTSTATUS CancelStatus = CxPlatNtCancelWaitCompletionPacket(
+            sqe->WaitCompletionPacket,   // WaitCompletionPacketHandle
+            TRUE);              // RemoveSignaledPacket
+        CXPLAT_DBG_ASSERT(NT_SUCCESS(CancelStatus));
```
guhetier: Same comment about cleanup as above.

src/inc/quic_platform_winuser.h
```diff
@@ -782,7 +814,15 @@ CxPlatEventQEnqueue(
     sqe->IsQueued;
 #endif
     CxPlatZeroMemory(&sqe->Overlapped, sizeof(sqe->Overlapped));
-    return PostQueuedCompletionStatus(*queue, 0, 0, &sqe->Overlapped) != 0;
+    if (!PostQueuedCompletionStatus(*queue, 0, 0, &sqe->Overlapped)){
+        if (sqe->WcpEvent) {
+            CxPlatEventSet(sqe->WcpEvent);
```
guhetier: Can the event be re-used to queue the same SQE multiple times?
Or is it only valid one time, and then another completion packet must be created?

`ShutdownSqe` would only be used once, but `WakeSqe` or `UpdatePollSqe` needs to be reusable.

guhetier: https://learn.microsoft.com/en-us/windows/win32/devnotes/ntassociatewaitcompletionpacket

This seems to say the association is valid only once. That wouldn't make it a great mechanism for events we need to queue multiple times (because we would have to deal with the association failure instead of the queuing failure, which will be a similar problem).

Issue #5520, opened 2025-10-13
Unintended handle context leak in Rust


nishanthkarthik:
### Describe the bug

The Rust bindings make a distinction between owned and ref types, for instance, `Connection` and `ConnectionRef`. On the server side, a listener callback receives a `ConnectionRef` for the `NewConnection` event. However, the callback handler attached to a `ConnectionRef` is not automatically cleaned up internally.

### Affected OS

- [ ] Windows
- [ ] Linux
- [ ] macOS
- [ ] Other (specify below)

### MsQuic version

main

### Steps taken to reproduce bug

1. Modify the example to allocate a large `Box<[u8; N]>` inside the connection handler closure
2. Watch the memory usage grow for each new connection

### Expected behavior

Setting `connection_ref.set_callback_handler(handler)` should automatically destroy the handler, if present. For every `ConnectionRef`, I assume someone owns a `Connection` internally. This `conn: Connection` should ideally invoke the cleanup.

### Actual outcome

The handler leaks. The example below provides a way to handle this, but I am not sure if this is the idiomatic way to clean up resources.

https://github.com/microsoft/msquic/blob/f7657ad5358c9b3282435de90547053fbd4eb959/src/rs/server_client_test.rs#L162

guhetier:
Thanks for reporting this.
Our Rust wrappers are a work in progress we haven't had the opportunity to focus on much recently - but a PR is very welcome if you want to give it a try.

PR #5618, opened 2025-11-21
Fix Rust connection callback context cleanup and add regression test

LindonAliu:
## Description
Closes: #5520 
- Clean up connection callback contexts on ShutdownComplete so handlers set via ConnectionRef are freed automatically (no leak).
- Allow connection callbacks to return success when the context was already cleared (e.g., post-cleanup), avoiding panics.
- Add a regression test that asserts the server-side ConnectionRef handler is dropped after shutdown.
Testing
## Testing

- cargo test
## Documentation

### DIFF
--- a/src/rs/lib.rs
+++ b/src/rs/lib.rs
@@ -705,9 +705,9 @@ macro_rules! define_quic_handle_ctx_fn {
             /// Gets the callback context without clearing it from the handle.
             ///
             /// # Safety
-            /// Only call while closing the handle, so MsQuic will not read the
-            /// context again after this function returns. Calling this in other
-            /// situations can cause double frees or use-after-free.
+            /// Only call while closing the handle. The return value must be kept
+            /// alive until the handle is fully closed and no more callbacks can
+            /// be emitted. Dropping it earlier can cause use-after-free.
             unsafe fn peek_callback_ctx(&self) -> Option<Box<Box<$callback_type>>> {
                 let ctx = self.get_context();
                 if ctx.is_null() {
--- a/src/rs/server_client_test.rs
+++ b/src/rs/server_client_test.rs
@@ -397,13 +397,9 @@ fn connection_ref_callback_cleanup() {
         .expect("Client did not complete shutdown");
 
     // Drop the server connection to trigger cleanup of the callback context.
+    // close_inner drops the context synchronously after ConnectionClose returns.
     drop(server_conn);
 
-    let mut retries = 50;
-    while drop_counter.load(Ordering::SeqCst) == 0 && retries > 0 {
-        std::thread::sleep(Duration::from_millis(10));
-        retries -= 1;
-    }
     assert_eq!(
         drop_counter.load(Ordering::SeqCst),
         1,


LindonAliu:

src/rs/lib.rs
```diff
@@ -714,6 +714,18 @@ macro_rules! define_quic_handle_ctx_fn {
                     None
                 }
             }
+
+            /// # Safety
+            /// Removes the callback context from the handle and returns it.
```
guhetier: nit: The comment describing the behavior of the function come above the safety section.

```suggestion
            /// Removes the callback context from the handle and returns it.
            /// # Safety
            /// <Safety invariants the caller should respect>
```

Does this still need to be unsafe?

src/rs/lib.rs
```diff
@@ -697,15 +697,15 @@ macro_rules! define_quic_handle_ctx_fn {
...
-                    unsafe { self.set_context(std::ptr::null_mut()) };
+                if let Some(ctx) = unsafe { self.take_callback_ctx() } {
+                    std::mem::drop(ctx);
                 }
             }
 
             /// # Safety
             /// Caller is responsible for clearing the context if needed.
             /// This does not clear the ctx.
+            #[allow(dead_code)]
```
guhetier: If this is now dead code, it can be deleted.

src/rs/lib.rs
```diff
@@ -861,17 +873,29 @@ extern "C" fn raw_conn_callback(
...
+        }
+        // Context already cleaned (e.g. after ShutdownComplete). Nothing to do.
+        None => StatusCode::QUIC_STATUS_SUCCESS.into(),
     };
-    let event = ConnectionEvent::from(unsafe { event.as_ref().unwrap() });
-    match f(conn, event) {
-        Ok(_) => StatusCode::QUIC_STATUS_SUCCESS.into(),
-        Err(e) => e.0,
+
+    if cleanup_ctx {
```
guhetier: I think this might not be addressing the root cause of the issue: the API contract of MsQuic is that when the listener accept a connection (`ListenerEvent::NewConnection`), the application takes ownership of the connection and must eventually close it (`Api::ffi_ref().ConnectionClose`).
The callback context should already be dropped when the connection is closed.

But `ListenerEvent::NewConnection` currently provide a ConnectionRef, not an owning Connection.
There is a todo in "types.rs" the event definition about this needing to be fixed.

src/rs/lib.rs
LindonAliu: `take_callback_ctx` stays `unsafe` because it reads a raw context pointer from the FFI handle, assumes the handle is still valid, and assumes the stored pointer was produced via our boxing convention. If those invariants are violated (foreign/invalid pointer, handle already closed), we could double‑free or touch invalid memory. I’ll leave it unsafe and spell that out in the safety comment.

src/rs/lib.rs
LindonAliu: Thanks a lot for the review and for pointing out the proper fix path. I agree the ownership should be adjusted on ListenerEvent::NewConnection; I’ll look into that direction next.

src/rs/lib.rs
```diff
@@ -861,17 +866,21 @@ extern "C" fn raw_conn_callback(
...
+        Some(f) => {
+            let event = ConnectionEvent::from(event_ref);
+            let conn = unsafe { ConnectionRef::from_raw(connection) };
+            match f(conn, event) {
+                Ok(_) => StatusCode::QUIC_STATUS_SUCCESS.into(),
+                Err(e) => e.0,
+            }
+        }
+        // Context already cleaned (e.g. after ShutdownComplete). Nothing to do.
+        None => StatusCode::QUIC_STATUS_SUCCESS.into(),
```
guhetier, line 896: I don't think this scenario is possible assuming the logic is sound. ShutdownComplete is the last event delivered to an application by MsQuic (see QUIC_CONNECTION_EVENT.md), and the rust wrapper own the context so it should never be freed before ShutdownComplete.

We can `expect` it to be present, I think.

src/rs/server_client_test.rs
```diff
@@ -286,3 +293,123 @@ fn test_server_client() {
...
+                        ConnectionEvent::ShutdownComplete { .. } => {
+                            let _ = server_handle_tx.send(unsafe { conn.as_raw() });
+                        }
+                        _ => {}
+                    };
+                    Ok(())
+                });
+                connection.set_configuration(&server_config)?;
+                // Keep connection alive until ShutdownComplete closes it.
+                let _ = unsafe { connection.into_raw() };
```
guhetier: It would be great if the test could take ownership of the connection safely. It would serve as an example of what apps are expected to be doing and could simplify the cleanup.

src/rs/lib.rs
LindonAliu: Thanks for the review! I found a case where the context can be `None`:
When `Connection::drop()` is called from outside the callback (e.g., after transferring ownership via a channel), `close_inner()` does:
1. `take_callback_ctx()` → context becomes `null`
2. `ConnectionClose()` → MsQuic triggers `ShutdownComplete` **synchronously** before returning
3. The callback is invoked with a `null` context

This is demonstrated in the test where we `drop(server_conn)` from the main thread. Should we:
1. Keep handling `None` to support both patterns (close from callback vs close from outside)?
2. Or change `close_inner()` to drop the context **after** the handle is fully closed?

I lean towards option 2, dropping the context after `ConnectionClose()` returns would ensure the callback always has a valid context, and `expect()` would be safe. This also aligns better with your expectation that `ShutdownComplete` always has a context. What do you think?

guhetier: Thanks for explaining.

It seems to me like an unintentional change of behavior introduced by the switch to `take_callback_ctx`: before, `get_callback_ctx` would get the context but not take it, so it would still be present during the call to `ConnectionClose` and usable in the callbacks (`ConnectionClose` is blocking until all callbacks it could triggers are completed).

Now, if `ConnectionClose` is called without a `ConnectionShutdown` call first, the application would not get the `Shutdown` notification anymore.
I think we need to adapt `close_inner` so the context stays set in the connection until `ConnectionClose` returns.

src/rs/lib.rs
```diff
@@ -1102,7 +1108,7 @@ impl Listener {
     fn close_inner(&self) {
         if !self.handle.is_null() {
             // consume the context and drop it after handle close.
-            let ctx = unsafe { self.get_callback_ctx() };
+            let ctx = unsafe { self.take_callback_ctx() };
```
guhetier: I think we have the same problem as above here.
`ListenerClose` could potentially trigger notifications (I don't think it currently does, but that might happen in the future). The context needs to stay set until `ListenerClose` returns.

LindonAliu: Thanks for the follow-up! I updated the close paths to keep the callback context alive until the Close call returns, so callbacks triggered synchronously during Close still see a valid context:

- Added peek_callback_ctx (read without clearing).
- Connection/Listener/Stream now use peek_callback_ctx in close_inner(), call Close, then drop the context after Close returns.

This aligns with the previous behavior you highlighted (context still present during Close) and prevents None when Close fires final notifications without a prior Shutdown.

src/rs/lib.rs
```diff
@@ -697,21 +697,41 @@ macro_rules! define_quic_handle_ctx_fn {
...
                 }
             }
 
+            /// Gets the callback context without clearing it from the handle.
+            ///
             /// # Safety
-            /// Caller is responsible for clearing the context if needed.
-            /// This does not clear the ctx.
-            unsafe fn get_callback_ctx(&self) -> Option<Box<Box<$callback_type>>> {
+            /// Only call while closing the handle, so MsQuic will not read the
```
guhetier: nit: To clarify the comment, we could write:

`Only call while closing the handle. The return value must be kept alive until the handle is fully closed and not more callback can be emitted"

In the current comment "MsQuic will not read the context again after this function return" is not fully true, since the context will be used for the final notifications until the handle `Close` call returns.

src/rs/server_client_test.rs
```diff
@@ -286,3 +293,121 @@ fn test_server_client() {
...
+        .recv_timeout(Duration::from_secs(5))
+        .expect("Server did not receive connection");
+    client_done_rx
+        .recv_timeout(Duration::from_secs(5))
+        .expect("Client did not complete shutdown");
+
+    // Drop the server connection to trigger cleanup of the callback context.
+    drop(server_conn);
+
+    let mut retries = 50;
```
guhetier: Do we still need retries here? `drop_inner` should drop the context inline, I think.

LindonAliu:
Hi @guhetier! Thanks a lot for the review and approval. Since I don’t have merge permissions, could you please merge this PR when you get a chance? Thank you!

guhetier:
I'll merge as soon as the CI completes (a few unrelated failures).
Thanks @LindonAliu for the contribution!

PR #5620, opened 2025-11-23
Build for Android Arm/Arm64 in CI

kitlith:
Arm and Arm64 android builds feel much more applicable than the x86_64 android builds that CI already produces. Since it was a simple change to make to the CI workflow, I figured I'd send the proposal to add them to the CI flow along with the PR to actually add it. It's up to you whether the extra CI work makes sense, though.

The Build workflow that I modified here completed successfully when I [ran it myself](<https://github.com/kitlith/msquic/actions/runs/19561650233>), though I was unable to complete the rest of the CI run due to not having access to `macos-latest-xlarge` instances.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -99,7 +99,7 @@ jobs:
       fail-fast: false
       matrix:
         config: ['Debug', 'Release']
-        plat: [linux]
+        plat: [android, linux]
         os: ['ubuntu-22.04', 'ubuntu-24.04']
         arch: [arm, arm64]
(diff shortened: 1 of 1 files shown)


guhetier:
Thanks for the PR! Let's see if the CI succeeds :)

kitlith:
The failing workflows look unrelated (spurious?) to me, but I'm not super familiar with this repo's further CI flows.

guhetier:
First time I am seeing this failure, but I don't see how it would be related either. I started a re-run.

guhetier (approved):

PR #5638, opened 2025-12-04
Track XDP main

ProjectsByJackHe:
## Description

Let's track XDP main instead of a release branch as currently MsQuic has dependencies on stuff outside of the release branches.

## Testing

CI

## Documentation

N/A

### DIFF
--- a/docs/Release.md
+++ b/docs/Release.md
@@ -62,14 +62,17 @@ MsQuic may work on other platforms, including macOS, iOS, Android, x86, etc. but
 1. Add (via PR) notes above for the new release.
 1. Fork `main` branch to `release/X.Y` where `X` is the major version and `Y` is the minor version.
 1. Update (via PR) the minor version for the `main` branch:
-   - Run `./scripts/update-version.ps1 -Part Minor` to generate the relavent changes.
+   - Run `./scripts/update-version.ps1 -Part Minor` to generate the relevant changes.
    - Also add the new version to the bug_report.yaml issue template file.
+1. Update `.gitmodules` for the xdp-for-windows submodule to track a specific XDP release branch instead of `main`. If there
+is no release branch in XDP that has the dependencies needed for this new MsQuic release,
+you should create a new XDP release first.
 
 ## Servicing a Release Branch
 
-1. Changes first go into the `main` branch, and then are cherry-picked into the relavent `release/X.Y` branches.
+1. Changes first go into the `main` branch, and then are cherry-picked into the relevant `release/X.Y` branches.
 1. Update (via PR) the patch version for the release branches:
-   - Run `./scripts/update-version.ps1 -Part Patch` to generate the relavent changes.
+   - Run `./scripts/update-version.ps1 -Part Patch` to generate the relevant changes.
 
 ## Publishing a Release Branch
 
@@ -77,7 +80,7 @@ MsQuic may work on other platforms, including macOS, iOS, Android, x86, etc. but
    - Make sure to pick the correct `release/X.Y` branch
    - The tag should be the full version number: `vX.Y.Z`
    - The release title should be `MsQuic vX.Y.Z`
-   - Put relavent information in the notes of the release (see previous releases for examples)
+   - Put relevant information in the notes of the release (see previous releases for examples)
 1. Wait for [msquic-Official-Tests](https://mscodehub.visualstudio.com/msquic/_build?definitionId=1824&_a=summary) pipeline to run for the newly created tag.
 1. Download the `distribution` packages from the artifacts and upload them to the GitHub release:
    - msquic_gamecore_console_x64_Release_schannel.zip


nibanks:
I only caution that at msquic release branches only reference xdp release branches.

mtfriesen:
> I only caution that at msquic release branches only reference xdp release branches.

This is a good point - @ProjectsByJackHe can you update the release docs to make changing the submodule to the latest XDP release branch part of the release steps? Though if we do that, we in turn could make releasing MsQuic into a refactoring exercise if the most recent XDP release branch is missing compile-time dependencies. Hmm.

guhetier:
> > I only caution that at msquic release branches only reference xdp release branches.
> 
> This is a good point - @ProjectsByJackHe can you update the release docs to make changing the submodule to the latest XDP release branch part of the release steps? Though if we do that, we in turn could make releasing MsQuic into a refactoring exercise if the most recent XDP release branch is missing compile-time dependencies. Hmm.

I think having to change the submodule when creating a release branch is somewhat error prone / easy to forget.

If there are difference in the main and xdp release branch, it might mean we introduce a bug in the release branch that was not happening in main.

If we need to update the submodule by hand from time to time, I'd rather have to do it in main, rather than at release time.

ProjectsByJackHe:
> > I only caution that at msquic release branches only reference xdp release branches.
> 
> This is a good point - @ProjectsByJackHe can you update the release docs to make changing the submodule to the latest XDP release branch part of the release steps? Though if we do that, we in turn could make releasing MsQuic into a refactoring exercise if the most recent XDP release branch is missing compile-time dependencies. Hmm.

I updated the release docs. I think to avoid the refactoring hassle in MsQuic whenever you create a new release, if there is no release branch that integrates nicely with this new MsQuic release, devs should make a new XDP release first. 

That way, active development in main for MsQuic tracks the latest features in XDP, and releases for MsQuic track releases in XDP, which I think should be the way to go.

guhetier:
> That way, active development in main for MsQuic tracks the latest features in XDP, and releases for MsQuic track releases in XDP, which I think should be the way to go.

I am not sure I agree with that. MsQuic devs and XDP devs happen to be largely the same group, but it should not be the responsibility of MsQuic release process to figure out when to create an XDP release or if a specific build is stable.

docs/Release.md
```diff
@@ -62,22 +62,25 @@ MsQuic may work on other platforms, including macOS, iOS, Android, x86, etc. but
 1. Add (via PR) notes above for the new release.
 1. Fork `main` branch to `release/X.Y` where `X` is the major version and `Y` is the minor version.
 1. Update (via PR) the minor version for the `main` branch:
-   - Run `./scripts/update-version.ps1 -Part Minor` to generate the relavent changes.
+   - Run `./scripts/update-version.ps1 -Part Minor` to generate the relevant changes.
    - Also add the new version to the bug_report.yaml issue template file.
+1. Update `.gitmodules` for the xdp-for-windows submodule to track a specific XDP release branch instead of `main`. If there
+is no release branch in XDP that has the dependencies needed for this new MsQuic release,
```
guhetier, line 68: "Does this release branch of XDP has the dependencies needed for this MsQuic release" is not an easy question to answer if MsQuic has been tracking XDP main for development or testing.

 It sounds to me like a "fuzzy" step in a process that should be precise and easy to follow otherwise.

mtfriesen:
Agreed with all points above: XDP and MsQuic releases _must_ be decoupled, but MsQuic must also avoid taking fragile dependencies on unreleased XDP code within its release branches.

I've been thinking about this a bit more, and maybe MsQuic needs to take build dependencies two different ways:

1. The official/released SDK nuget provided by XDP matching the official XDP runtime package. This will be the only mode supported by official/release builds, and needs to also be part of the `main` test matrix. This allows MsQuic + XDP to be built and tested only against officially supported artifacts.
2. MsQuic can also track prerelease/main code, either continuing to use the submodule on some non-release commit, or by taking a parallel SDK nuget + runtime nuget dependency, with both tracking prerelease code. Since most XDP features require runtime nuget changes, it implies XDP must create a prerelease for any cross-project engineering, so there's no additional cost to publishing the prerelease SDK nuget. Alternatively, MsQuic could use the XDP submodule to consume both the compile time and build its runtime test dependencies.

nibanks:
Many of these problems are why we had decided to always have official MsQuic branches referenced official releases of XDP. I still feel pretty strongly that should stay the case. It's fine if there is a way for private branches to take temporary dependencies on unreleased XDP (or other dependencies), but merging that into main causes problems...

ProjectsByJackHe:
> Many of these problems are why we had decided to always have official MsQuic branches referenced official releases of XDP. I still feel pretty strongly that should stay the case. It's fine if there is a way for private branches to take temporary dependencies on unreleased XDP (or other dependencies), but merging that into main causes problems...

If we want to go back to this model, we need to wait until XDP releases v1.2, update submodules to point to v1.2, and make sure future devs don't check in MsQuic code that depends on some unreleased feature in XDP main into MsQuic main. I think this is the easiest way forward for now.

mtfriesen:
To be clear, the problem isn't to determine what is easiest for us now, but what is the best (balancing safe, reliable, efficient, agile) way to maintain the dependency between these two projects.

We've discovered tracking the latest XDP release branch keeps MsQuic a little too far behind, but we're now in a bad state where MsQuic is _ahead_ of release but still supposedly tracking the release branch.

I generally agree with both Nick and Guillaume that we need to track this dependency in a more robust and predictable way, including, but not limited to, we must depend only on released features in release branches.

PR #5643, opened 2025-12-08
Add logging for spinquic tests.

ProjectsByJackHe:
## Description

The Spin powershell script had support for gathering MsQuic logs but the CI is never using it.

Let's add an option to configure logging in spin tests. For workflow dispatch, the user can control the log profile. For pull request and merge events, Basic.Light is used.

## Testing

CI

## Documentation

N/A

### DIFF
--- a/.github/workflows/stress.yml
+++ b/.github/workflows/stress.yml
@@ -145,12 +145,12 @@ jobs:
       if: github.event_name == 'pull_request'
       timeout-minutes: 15
       shell: pwsh
-      run: scripts/spin.ps1 -AZP -Config ${{ matrix.vec.config }} -Arch ${{ matrix.vec.arch }} -Tls ${{ matrix.vec.tls }} -Timeout ${{ env.pr-timeout }} -RepeatCount ${{ env.pr-repeat }} -AllocFail ${{ env.pr-allocfail }} ${{ matrix.vec.xdp }} -LogProfile ${{ inputs.logprofile || 'Basic.Light' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/stress.yml
```diff
@@ -2,6 +2,20 @@ name: Stress
...
   workflow_dispatch:
+    inputs:
+      logprofile:
+        description: 'Log profile to use'
+        required: false
+        default: "None"
+        type: choice
+        options:
+          - "None"
+          - SpinQuic.Light
```
guhetier, line 13: Wouldn't this one make more sense as a default?
Or if it isn't enough (look like it might contain very little), maybe we should consider adding what makes sense to it - no point in having a SpinQuic profile that is not enough to diagnose issues.

ProjectsByJackHe:
> Seems good to me in concept.
> 
> Most of the time though, spin test matter only when they trigger a crash. Shouldn't we focus on dump collection?

If the script does what it's supposed to do, having log-profile set to non-NULL should automatically coalesce all additional logs created into the artifacts folder. Today, the artifacts folder does not contain any ETL.

.github/workflows/stress.yml
ProjectsByJackHe: Yeah, we'll see what these look like in the next few iterations.

ProjectsByJackHe:
> Seems good to me in concept.
> 
> Most of the time though, spin test matter only when they trigger a crash. Shouldn't we focus on dump collection?

Also, if what you mean by "dump collection" is a crash dump, the spinquic tool does not produce any crash dumps in user mode.

.github/workflows/stress.yml
guhetier: Sounds good. If it doesn't appear useful, I would not hesitate to delete the profile from the wprpr file and replace it by less specific one.

PR #5645, opened 2025-12-09
Add verbose logs to some test passes to catch schannel timeout issue

guhetier:
## Description

Handshake/WithHandshakeArgsustomClientCertificateValidation has been failing with a call to schannel taking over 3sec.

This temporarily enables schannel logs to more test passes in the hope to get enough logs to investigate next repro.

## Testing

CI

## Documentation

N/A

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -145,13 +145,13 @@ jobs:
           { config: "Debug", plat: "linux", os: "ubuntu-24.04", arch: "x64", tls: "quictls", systemcrypto: "-UseSystemOpenSSLCrypto", build: "-Test", iouring: "-UseIoUring"  },
           { config: "Release", plat: "linux", os: "ubuntu-24.04", arch: "x64", tls: "quictls", systemcrypto: "-UseSystemOpenSSLCrypto", build: "-Test", iouring: "-UseIoUring"  },
           { config: "Debug", plat: "linux", os: "ubuntu-22.04", arch: "x64", tls: "openssl", sanitize: "-SanitizeAddress", build: "-Test"  },
-          { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "schannel", sanitize: "-SanitizeAddress", build: "-Test" },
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -145,13 +145,13 @@ jobs:
...
           { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "schannel", xdp: "-UseXdp", sanitize: "-SanitizeAddress", build: "-Test" },
           { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "schannel", xdp: "-UseXdp", qtip: "-UseQtip", sanitize: "-SanitizeAddress", build: "-Test", log: "Full.Verbose" },
-          { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "quictls", build: "-Test" },
+          { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "quictls", build: "-Test", log: "Full.Verbose" },
           { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "quictls", xdp: "-UseXdp", build: "-Test" },
           { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "quictls", xdp: "-UseXdp", qtip: "-UseQtip", build: "-Test", log: "Full.Verbose" },
-          { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "openssl", build: "-Test" },
+          { config: "Debug", plat: "windows", os: "windows-2022", arch: "x64", tls: "openssl", build: "-Test", log: "Full.Verbose" },
```
mtfriesen, line 154: We seem to be churning logs on and off a lot in CI.

One solution I've seen is to create a logging session per test case and only save/upload sessions for test cases that fail, at least by default. This avoids the massive log file problem while automatically producing artifacts for flaky or one-off failures.

Shall we do something similar in MsQuic?

guhetier: That may eventually be a solution.
For now, I don't think it deserves the investment yet (these are to investigate long standing corner case issues, not common ones).

Issue #5491, opened 2025-10-06
[CI - FAILURE] Stress (Debug, windows, windows-2022, x64, schannel, -Sanitize, -Test)


guhetier:
Stress test failure in https://github.com/microsoft/msquic/pull/5487

~Likely related to #5490~

guhetier:
- No dump collected (there should be in theory)
- No much logs

First step might be to collect diag data.

ProjectsByJackHe:
https://github.com/microsoft/msquic/pull/5643 Should add logging for stress tests.

guhetier:
#5585 might be relevant for collecting dumps.

ProjectsByJackHe:
Looks like we observed a failure while running the CI for #5643, which luckily included the logging: https://github.com/microsoft/msquic/actions/runs/20075142156/job/57593883201

<img width="1571" height="157" alt="Image" src="https://github.com/user-attachments/assets/90d093a9-6323-4fb4-bea2-95b19c368322" />

Looks like a classic watchdog timeout.

ProjectsByJackHe:
This issue should be addressed by https://github.com/microsoft/msquic/issues/5491

PR #5647, opened 2025-12-09
increase spinquic watchdog timeout

ProjectsByJackHe:
## Description

As discussed in issue #5491 , from logs, the watchdog assert is firing.
For now, let's increase it by 100%.

## Testing

CI

## Documentation

N/A

### DIFF
--- a/src/tools/spin/spinquic.cpp
+++ b/src/tools/spin/spinquic.cpp
@@ -167,7 +167,7 @@ public:
 // The amount of extra time (in milliseconds) to give the watchdog before
 // actually firing.
 //
-#define WATCHDOG_WIGGLE_ROOM 20000
+#define WATCHDOG_WIGGLE_ROOM 30000
 
 class SpinQuicWatchdog {
     CXPLAT_THREAD WatchdogThread;


guhetier:
I am not that familiar with the spin test, but isn't this only going to cause the spintest to run for a longer time?
Looking at the sources very fast, the time you change control the time spent spinning, and there is a WATCHDOG_WIGGLE_ROOM that gives a bit of extra time for the watchdog.

ProjectsByJackHe:
> I am not that familiar with the spin test, but isn't this only going to cause the spintest to run for a longer time? Looking at the sources very fast, the time you change control the time spent spinning, and there is a WATCHDOG_WIGGLE_ROOM that gives a bit of extra time for the watchdog.

Yes! good catch

guhetier:
Did you investigate, based on the traces, what was pending when the timeout fired? 2 / 3 seconds is already quite a lot. It is possible something was delayed on a slow VM, but it is possible too that a softlock / deadlock was happening in MsQuic.

ProjectsByJackHe:
> Did you investigate, based on the traces, what was pending when the timeout fired? 2 / 3 seconds is already quite a lot. It is possible something was delayed on a slow VM, but it is possible too that a softlock / deadlock was happening in MsQuic.

Based on the ETL trace from the link I added in the issue, I couldn't find any deadlocks happening. Although, there are comments in SpinQuic itself that notes certain code paths will lead to deadlocks, but those are all disabled.

guhetier:
Ok. This might help, but I suspect going from 2sec to 3sec won't be a definitive fix.
We should make sure dumps are collected so that next time, we can check the state of pending threads.

guhetier (approved):

guhetier:
@ProjectsByJackHe What is the status of this PR? Do you still intend to merge it? Or should it be abandoned?

ProjectsByJackHe:
> @ProjectsByJackHe What is the status of this PR? Do you still intend to merge it? Or should it be abandoned?

Excuse the latency, the actual fix was merged in https://github.com/microsoft/msquic/pull/5673

Closing this PR now.

PR #5648, opened 2025-12-09
[Abandoned] [DeepTest] Add Unit Test cases for Cubic Congestion Control 

saikat107:
**This PR is abandoned**
# New PR #5653 

## Description

This PR adds test suite contains 17 tests covering initialization, congestion window management, data tracking, congestion events, and utility functions. While achieving 63.42% line coverage, significant functionality remains untested, particularly advanced scenarios like HyStart++, congestion avoidance calculations, and spurious congestion recovery.

---

### Current Test Cases (17 tests)

1. **CubicTest.InitializeComprehensive** - Validates initialization with various settings
2. **CubicTest.InitializeBoundaries** - Tests boundary conditions during initialization
3. **CubicTest.MultipleSequentialInitializations** - Verifies multiple init cycles
4. **CubicTest.CanSendScenarios** - Tests send permission logic with various states
5. **CubicTest.SetExemption** - Validates exemption packet handling
6. **CubicTest.GetSendAllowanceScenarios** - Tests send allowance with/without pacing
7. **CubicTest.GetSendAllowanceWithActivePacing** - Validates pacing calculations
8. **CubicTest.GetterFunctions** - Tests getter APIs for exposing internal state
9. **CubicTest.ResetScenarios** - Validates full and partial reset behavior
10. **CubicTest.OnDataSent_IncrementsBytesInFlight** - Tests flight tracking on send
11. **CubicTest.OnDataInvalidated_DecrementsBytesInFlight** - Tests invalidation handling
12. **CubicTest.OnDataAcknowledged_BasicAck** - Validates ACK processing in slow start
13. **CubicTest.OnDataLost_WindowReduction** - Tests loss-based congestion events
14. **CubicTest.OnEcn_CongestionSignal** - Validates ECN-based congestion detection
15. **CubicTest.GetNetworkStatistics_RetrieveStats** - Tests statistics retrieval
16. **CubicTest.MiscFunctions_APICompleteness** - Covers stub/simple functions
17. **CubicTest.HyStart_StateTransitions** - Tests HyStart++ state machine basics

### Tested paths
✅ **Comprehensive initialization testing** - All init paths covered  
✅ **Good basic operation coverage** - Core send/ack/loss cycle tested  
✅ **Function pointer testing** - APIs accessed through function table  
✅ **Memory safety** - Mock connection structure verified with static asserts  
✅ **Boundary condition testing** - Edge values for window sizes tested  

### To be tested in future PR
❌ **No congestion avoidance testing** - Core CUBIC algorithm untested  
❌ **No HyStart++ testing** - Advanced optimization completely skipped  
❌ **Limited recovery testing** - Only basic loss handling tested  
❌ **No persistent congestion** - Severe congestion path untested  
❌ **No spurious recovery** - False congestion detection untested  
❌ **Single-phase testing** - Tests don't run through multiple phases (slow start → congestion avoidance → recovery)

## Documentation
N/A

### DIFF
--- a/src/core/unittest/BbrTest.cpp
+++ b/src/core/unittest/BbrTest.cpp
@@ -1005,3 +1005,160 @@ TEST(BbrTest, MaxAckHeightFilterResetOnEmpty)
     QUIC_STATUS Status = QuicSlidingWindowExtremumGet(&Cc.Bbr.MaxAckHeightFilter, &Entry);
     ASSERT_EQ(QUIC_STATUS_NOT_FOUND, Status);
 }
+
+//
+// Tests for uncovered branches in BbrCongestionControlGetCongestionWindow
+//
+
+TEST(BbrTest, GetCongestionWindowInProbeRttState)
+{
+    QUIC_CONGESTION_CONTROL Cc;
+    CxPlatZeroMemory(&Cc, sizeof(Cc));
+
+    QUIC_SETTINGS_INTERNAL Settings;
+    CxPlatZeroMemory(&Settings, sizeof(Settings));
+    Settings.CongestionControlAlgorithm = QUIC_CONGESTION_CONTROL_ALGORITHM_BBR;
+    Settings.InitialWindowPackets = 10;
+
+    QuicCongestionControlInitialize(&Cc, &Settings);
+
+    // Save the initial congestion window
+    uint32_t InitialCwnd = Cc.Bbr.CongestionWindow;
+
+    // Transition to PROBE_RTT state (BBR_STATE_PROBE_RTT = 3)
+    Cc.Bbr.BbrState = 3; // BBR_STATE_PROBE_RTT
+
+    // In PROBE_RTT, should return minimum congestion window
+    uint32_t CwndInProbeRtt = Cc.QuicCongestionControlGetCongestionWindow(&Cc);
+    
+    // Minimum should be kMinCwndInMss * DatagramPayloadLength (kMinCwndInMss = 4)
+    ASSERT_LT(CwndInProbeRtt, InitialCwnd);
+    ASSERT_NE(0u, CwndInProbeRtt);
+}
+
+TEST(BbrTest, GetCongestionWindowInRecoveryState)
+{
+    QUIC_CONGESTION_CONTROL Cc;
+    CxPlatZeroMemory(&Cc, sizeof(Cc));
+
+    QUIC_SETTINGS_INTERNAL Settings;
+    CxPlatZeroMemory(&Settings, sizeof(Settings));
+    Settings.CongestionControlAlgorithm = QUIC_CONGESTION_CONTROL_ALGORITHM_BBR;
+    Settings.InitialWindowPackets = 10;
+
+    QuicCongestionControlInitialize(&Cc, &Settings);
+
+    // Enter recovery state (RECOVERY_STATE_CONSERVATIVE = 1)
+    Cc.Bbr.RecoveryState = 1;
+    
+    // Set recovery window smaller than congestion window
+    Cc.Bbr.RecoveryWindow = Cc.Bbr.CongestionWindow / 2;
+
+    // Should return the minimum of CongestionWindow and RecoveryWindow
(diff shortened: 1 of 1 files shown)


src/core/unittest/CubicTest.cpp
```diff
@@ -0,0 +1,564 @@
...
+#include "main.h"
+#ifdef QUIC_CLOG
+#include "CubicTest.cpp.clog.h"
+#endif
+
+//
+// Mock structures for testing
+//
+
+typedef struct MOCK_CONNECTION {
```
guhetier: Ah, it's annoying the connection is used indirectly in the congestion controller.
I think we should either use the real connection object (initializing only the relevant bits) or have static assert to ensure the memory layout matches a `QUIC_CONNECTION`.

I don't think that is the case currently, that will likely cause some runtime OOB accesses.

src/core/unittest/CubicTest.cpp
```diff
@@ -0,0 +1,564 @@
...
+
+Abstract:
+
+    Unit test for CUBIC congestion control initialization logic.
+
+--*/
+
+#include "main.h"
+#ifdef QUIC_CLOG
+#include "CubicTest.cpp.clog.h"
```
guhetier, line 14: This needs to be added to claog.inputs I think, so the log generation picks it up, the same way as "RecvBufferTest.cpp.clog.h"

src/core/unittest/CubicTest.cpp
```diff
@@ -0,0 +1,564 @@
+/*++
+
+    Copyright (c) Microsoft Corporation.
+    Licensed under the MIT License.
+
+Abstract:
+
+    Unit test for CUBIC congestion control initialization logic.
```
guhetier: nit: Hopefully we eventually get more than just the initialization logic test here :)


```suggestion
    Unit tests for CUBIC congestion control
```

src/core/unittest/CubicTest.cpp
saikat107: Makes sense. Created objects of `QUIC_CONNECTION`. Please let me know if this is what you meant. 

PS: did not understand the static assert bit though!

src/core/unittest/CubicTest.cpp
saikat107: Added in the `src/generated`. But the test file, in current version, does not reference any clog functions directly (not sure if it does indirectly). We could remove this, and add back in future if a clog functionality is accessed.

saikat107: Added to the clog.inputs

saikat107:
Not sure why the test in the CI is failing. In my local run, the test case pass. 

BTW, here is (AI-agent generated) documentation of the added tests (I did check, but may have missed something)
[COVERAGE_REPORT.md](https://github.com/user-attachments/files/24090070/CUBIC_INITIAL_COVERAGE_REPORT.md)

guhetier:
I started a re-run for the test, it seems unrelated to this change. I'll look into it.
I'll try to review the PR shortly.

src/core/unittest/CubicTest.cpp
guhetier: When you call a function like `QuicCongestionControlGetConnection`, it will use `QuicCongestionControlGetConnection` to get the parent `QUIC_CONNECTION` object from the pointer to the `QUIC_CONGESTION_CONTROL`.

This is a (sadly) common pattern in C, that assume the congestion control struct is always a member of a `QUIC_CONNECTION`. But here, the congestion control struct is a member of the `MOCK_CONNECTION`. If the `MOCK_CONNECTION` binary layout is not matching the one of a `QUIC_CONNECTION`, the test will access memory at invalid offset.

Static assert can help validate the binary layout is the same.
But the simpler / safer is probably to not use a `MOCK_CONNECTION` struct and use a `QUIC_CONNECTION` (which I don't see done yet in this iteration?)

saikat107:
@guhetier, abandoning this PR, as we pushed commits for bbr testing (which is not yet complete and require more work) to this branch. 
Opened a new PR #5653

PR #5650, opened 2025-12-12
Implement a generic test ioctl for kernel mode tests

guhetier:
## Description

Adding system tests to MsQuic is convoluted because they must run in both user-mode and kernel mode.

Before this PR, to add a system test, 5 code locations must be edited:
- define a new IOCTL
    - IOCTL use sequential numbers: the new test will be at the end of the list, even if it would better to group it with related tests
- handle the IOCTL from the test driver
- define a type for the IOCTL parameters
- add the parameter size type to the size array
- populate the IOCTL parameter struct from the google test parameters
- invoke the IOCTL

With this PR, only two file locations must be edited, with trivial changes:
- register the test in kernel mode (1 line)
- invoke the test from user mode, forwarding the gtest parameters directly (no additional parameter type, 1 line)

This PR introduces some constraint over the test function: they must accept a single parameter by constant reference. This isn't a limitation: test don't need non-constant parameters; multiple parameters can be packaged in a struct (and generally already are when provided by gtest); and this will encourage multiple small test with few parameters rather than complex test functions trying to handle multiple scenarios.

Exisiting test will be converted to this schema in follow up PRs.

## Testing

CI

## Documentation

Code comments

### DIFF
--- a/src/test/bin/quic_gtest.cpp
+++ b/src/test/bin/quic_gtest.cpp
@@ -263,7 +263,6 @@ bool InvokeKernelTest(const std::string& Name, FunType, const ParamType& Params)
 TEST(ParameterValidation, ValidateGlobalParam) {
     TestLogger Logger("QuicTestValidateGlobalParam");
     if (TestingKernelMode) {
-        ASSERT_TRUE(InvokeKernelTest(FUNC(QuicTestGlobalParam)));
         ASSERT_TRUE(DriverClient.Run(IOCTL_QUIC_RUN_VALIDATE_GLOBAL_PARAM));
     } else {
         QuicTestGlobalParam();
--- a/src/test/bin/winkernel/control.cpp
+++ b/src/test/bin/winkernel/control.cpp
@@ -650,7 +650,7 @@ ExecuteTestRequest(
 
     // Fail if no function matched
     char Buffer[256];
-    (void)_vsnprintf_s(Buffer, sizeof(Buffer), _TRUNCATE, "Unknown function name in test RPC call: %s", Request->FunctionName);
+    (void)_vsnprintf_s(Buffer, sizeof(Buffer), _TRUNCATE, "Unknown function name in IOCTL test request: %s", Request->FunctionName);
 
     QuicTraceEvent(LibraryError, "[ lib] ERROR, %s.", Buffer);
 


src/test/bin/winkernel/control.cpp
```diff
@@ -536,7 +538,7 @@ size_t QUIC_IOCTL_BUFFER_SIZES[] =
     0,
     0,
     0,
-    sizeof(INT32),
```
mtfriesen, line 539: is this intentional?

src/test/bin/winkernel/control.cpp
```diff
@@ -585,6 +587,76 @@ typedef union {
...
+        } \
+    } while (false)
+
+QUIC_STATUS
+ExecuteTestRequest(
+    _In_ QUIC_RUN_TEST_REQUEST* RpcRequest
+    )
+{
+    // Ensure null termination
+    RpcRequest->FunctionName[255] = '\0';
```
mtfriesen: 255 hard-coded?

src/test/bin/winkernel/control.cpp
```diff
@@ -627,7 +699,46 @@ QuicTestCtlEvtIoDeviceControl(
...
+                "WdfRequestRetrieveInputBuffer failed for run test request");
+            goto Error;
+        }
+
+        if (Length < sizeof(QUIC_RUN_TEST_REQUEST) + TestRequest->ParameterSize) {
+            Status = STATUS_INVALID_PARAMETER;
+            QuicTraceEvent(
+                LibraryError,
+                "[ lib] ERROR, %s.",
+                "RPC request buffer too small for parameters");
```
guhetier:
```suggestion
                "IOCTL buffer too small for the test parameters");
```

src/test/bin/winkernel/control.cpp
guhetier: No, I added a size then removed it and lost the comma in the process.

I'll fix it - but that entire array should disappear when tests are migrated!

src/test/bin/quic_gtest.cpp
```diff
@@ -227,10 +229,35 @@ struct TestLoggerT {
...
+
+bool InvokeTestInKernel(const std::string& Name) {
+    QUIC_RUN_TEST_REQUEST Request{};
+    Name.copy(Request.FunctionName, sizeof(Request.FunctionName));
+
+    return DriverClient.Run(IOCTL_QUIC_RUN_TEST, (void*)&Request, (uint32_t)sizeof(Request));
+}
+
+template <class ParamType>
+bool InvokeTestInKernel(const std::string& Name, const ParamType& Params) {
```
mtfriesen: IMO the slightly more succinct `InvokeKernelTest` would be just as clear.

src/test/bin/winkernel/control.cpp
```diff
@@ -585,6 +587,76 @@ typedef union {
     X; \
     Status = Client->TestFailure ? STATUS_FAIL_FAST_EXCEPTION : STATUS_SUCCESS;
 
+// Base template providing a readable error for unsupported scenarios
+template<class... Args>
```
anrossi, line 591: Ohhh, so that's how this is done. I just learned something!

src/test/bin/winkernel/control.cpp
```diff
@@ -585,6 +587,76 @@ typedef union {
...
+
+    const Arg& arg = *reinterpret_cast<const Arg*>(argBuffer);
+    func(arg);
+    return QUIC_STATUS_SUCCESS;
+}
+
+// Specialization for functions with no arguments
+template<>
+QUIC_STATUS InvokeTestFunction(void(*func)(), const uint8_t*, uint32_t argBufferSize) {
+    if (0 != argBufferSize) {
```
anrossi, line 615: shouldn't this check that the arg buffer is also null?

src/test/bin/winkernel/control.cpp
```diff
@@ -585,6 +587,76 @@ typedef union {
...
+
+    // Register any test functions here
+    RegisterTestFunction(QuicTestAckSendDelay);
+    RegisterTestFunction(QuicTestValidateApi);
+    RegisterTestFunction(QuicTestStreamAppProvidedBuffers);
+    RegisterTestFunction(QuicTestStreamAppProvidedBuffersOutOfSpace);
+
+    // Fail if no function matched
+    char Buffer[256];
+    (void)_vsnprintf_s(Buffer, sizeof(Buffer), _TRUNCATE, "Unknown function name in test RPC call: %s", Request->FunctionName);
```
anrossi: IOCTL not RPC right?

src/test/bin/winkernel/control.cpp
guhetier: There is no arg buffer pointer. The arguments are in the same buffer as the request, just past the request struct.
I can always get that address, the question is only whether there is data I am allowed to access there or not.

src/test/bin/winkernel/control.cpp
guhetier: Good catch, I missed one after I decided not to call this a UM to KM RPC :)

src/test/bin/winkernel/control.cpp
```diff
@@ -585,6 +587,76 @@ typedef union {
...
+            "Parameter provided for a test function expecting none");
+        return QUIC_STATUS_INVALID_PARAMETER;
+    }
+
+    func();
+    return QUIC_STATUS_SUCCESS;
+}
+
+#define RegisterTestFunction(Function) \
+    do { \
```
anrossi, line 628: Why does the macro need the do...while loop around the if-block? Isn't the if-block providing enough of a scope?

src/test/bin/winkernel/control.cpp
```diff
@@ -585,6 +587,76 @@ typedef union {
...
+
+QUIC_STATUS
+ExecuteTestRequest(
+    _In_ QUIC_RUN_TEST_REQUEST* Request
+    )
+{
+    // Ensure null termination
+    Request->FunctionName[sizeof(Request->FunctionName) - 1] = '\0';
+
+    // Register any test functions here
```
anrossi, line 645: aww, so we still have to register the test function with the driver? :(
At least it's better than having to get the IOCTL code correct AND update the switch statement list.

src/test/bin/winkernel/control.cpp
```diff
@@ -627,7 +699,46 @@ QuicTestCtlEvtIoDeviceControl(
...
+                LibraryError,
+                "[ lib] ERROR, %s.",
+                "IOCTL buffer too small for test parameters");
+            goto Error;
+        }
+
+        // Invoke the test function
+        Client->TestFailure = false;
+        Status = ExecuteTestRequest(TestRequest);
+        if (Status == QUIC_STATUS_SUCCESS && Client->TestFailure) {
```
anrossi, line 734: Why also checking the status? `QuicTestCtlRun` doesn't currently consider the return status.

src/test/bin/quic_gtest.cpp
```diff
@@ -10,6 +10,8 @@
 #include "quic_gtest.cpp.clog.h"
 #endif
 
+#include <array>
```
anrossi, line 13: *gasp* STL! 
If you need a dynamic array, we have a kernel-friendly `std::vector` replacement called KArray.

src/test/bin/quic_gtest.cpp
```diff
@@ -227,27 +229,41 @@ struct TestLoggerT {
...
 }
 
+#define FUNC(TestFunction) \
+    #TestFunction, TestFunction
+
 TEST(ParameterValidation, ValidateGlobalParam) {
     TestLogger Logger("QuicTestValidateGlobalParam");
     if (TestingKernelMode) {
+        ASSERT_TRUE(InvokeKernelTest(FUNC(QuicTestGlobalParam)));
         ASSERT_TRUE(DriverClient.Run(IOCTL_QUIC_RUN_VALIDATE_GLOBAL_PARAM));
```
anrossi, line 266: Did you mean to delete this? Seems like this is running the test twice, once using the old way, and once with the new way.

src/test/bin/winkernel/control.cpp
guhetier: Yes, there is no solution around it in kernel mode. Typically, you could use self-registering object (which is what gtest does to register test functions for instance), but that requires a runtime with support for complex initialization of static variables, which kernel mode doesn't allow.

src/test/bin/winkernel/control.cpp
guhetier: The `do ... while` is never for the scope (you could just use `{ .. }` for that), but is a common trick to require a `;` at the end of the macro invocation so it looks like a normal statement.

src/test/bin/quic_gtest.h
```diff
@@ -74,6 +74,19 @@ class WithFamilyArgs : public testing::Test,
     public testing::WithParamInterface<FamilyArgs> {
 };
 
+struct WithFamilyArgs2 :
```
anrossi, line 77: I'm confused why this exists separate from `WithFamilyArgs`. Just a different way of doing the same thing?

src/test/bin/winkernel/control.cpp
guhetier: Technically, it would probably be fine because `Client->TestFailure == TRUE` implies `Status == QUIC_STATUS_SUCCESS`, but in general, if I already have a status failure, I don't want to override it, with a new one for something happening later - I want to report the first one.

src/test/bin/quic_gtest.cpp
guhetier: A static array is all I need here, and this code is user-mode only, so STL is welcome :)

src/test/bin/winkernel/control.cpp
anrossi: I guess I'm confused by the parameter list having 3 parameters. You're saying the middle parameter just isn't present in this case, so there's no point in verifying it is null?

src/test/bin/winkernel/control.cpp
anrossi: I always forget that's the reason. Maybe I'll remember now.

src/test/bin/quic_gtest.cpp
guhetier: Good catch, I forgot to get rid of the old way.

src/test/bin/quic_gtest.cpp
anrossi: oh right! This is UM

src/test/bin/quic_gtest.h
guhetier: I added this one just to demonstrate how to migrate a test - but since I don't migrate all the tests using `WithFamilyArgs`, I had to make a new one temporarily. It will go away (see draft PRs) as soon as I can migrate all the tests.

I need it because I needed `FamilyArgs2` temporarily, to have a type that doesn't depend on STL and is visible from KM.
`WithFamilyArgs2` can host the `Generate` function and stay UM only.

src/test/bin/winkernel/control.cpp
guhetier: No, look at the call site.

The argument buffer is always `(uint8_t*)(Request + 1)`, it is never null, because if there are arguments, that's where they are. But we are only allowed to read `argBufferSize` bytes from that address, so when there are no arguments, we should be allowed to read 0 bytes.

src/test/bin/quic_gtest.cpp
guhetier: Oh, actually that's probably why the CI is failing. It is not registered in the kernel side.

PR #5653, opened 2025-12-12
[DeepTest] Add Unit Test cases for Cubic Congestion Control

saikat107:
## Description

_Some unrelated change to cubic testing got pushed to the previous PR #5648. Starting this new PR to maintain the PRs well._

This PR adds test suite contains 17 tests covering initialization, congestion window management, data tracking, congestion events, and utility functions. While achieving 63.42% line coverage, significant functionality remains untested, particularly advanced scenarios like HyStart++, congestion avoidance calculations, and spurious congestion recovery.

---

### Current Test Cases (17 tests)

1. **CubicTest.InitializeComprehensive** - Validates initialization with various settings
2. **CubicTest.InitializeBoundaries** - Tests boundary conditions during initialization
3. **CubicTest.MultipleSequentialInitializations** - Verifies multiple init cycles
4. **CubicTest.CanSendScenarios** - Tests send permission logic with various states
5. **CubicTest.SetExemption** - Validates exemption packet handling
6. **CubicTest.GetSendAllowanceScenarios** - Tests send allowance with/without pacing
7. **CubicTest.GetSendAllowanceWithActivePacing** - Validates pacing calculations
8. **CubicTest.GetterFunctions** - Tests getter APIs for exposing internal state
9. **CubicTest.ResetScenarios** - Validates full and partial reset behavior
10. **CubicTest.OnDataSent_IncrementsBytesInFlight** - Tests flight tracking on send
11. **CubicTest.OnDataInvalidated_DecrementsBytesInFlight** - Tests invalidation handling
12. **CubicTest.OnDataAcknowledged_BasicAck** - Validates ACK processing in slow start
13. **CubicTest.OnDataLost_WindowReduction** - Tests loss-based congestion events
14. **CubicTest.OnEcn_CongestionSignal** - Validates ECN-based congestion detection
15. **CubicTest.GetNetworkStatistics_RetrieveStats** - Tests statistics retrieval
16. **CubicTest.MiscFunctions_APICompleteness** - Covers stub/simple functions
17. **CubicTest.HyStart_StateTransitions** - Tests HyStart++ state machine basics

### Tested paths
✅ **Comprehensive initialization testing** - All init paths covered  
✅ **Good basic operation coverage** - Core send/ack/loss cycle tested  
✅ **Function pointer testing** - APIs accessed through function table  
✅ **Memory safety** - Mock connection structure verified with static asserts  
✅ **Boundary condition testing** - Edge values for window sizes tested  

### To be tested in future PR
❌ **No congestion avoidance testing** - Core CUBIC algorithm untested  
❌ **No HyStart++ testing** - Advanced optimization completely skipped  
❌ **Limited recovery testing** - Only basic loss handling tested  
❌ **No persistent congestion** - Severe congestion path untested  
❌ **No spurious recovery** - False congestion detection untested  
❌ **Single-phase testing** - Tests don't run through multiple phases (slow start → congestion avoidance → recovery)

## Documentation
N/A

### DIFF
--- a/src/core/unittest/CubicTest.cpp
+++ b/src/core/unittest/CubicTest.cpp
@@ -331,7 +331,7 @@ TEST(CubicTest, GetSendAllowanceWithActivePacing)
     ASSERT_GT(Allowance, 0u); // Should allow some sending
     ASSERT_LT(Allowance, AvailableWindow); // But less than full window due to pacing
 
-    // Exact value is caldulated considering the current implementation is right and this test is meant to
+    // Exact value is calculated considering the current implementation is right and this test is meant to
     // prevent future regressions
     uint32_t ExpectedPacedAllowance = 4928; // Pre-calculated expected value
     ASSERT_EQ(Allowance, ExpectedPacedAllowance);
@@ -745,7 +745,7 @@ TEST(CubicTest, HyStart_StateTransitions)
 
     QUIC_ACK_EVENT AckEvent;
     CxPlatZeroMemory(&AckEvent, sizeof(AckEvent));
-    AckEvent.TimeNow = 1000000; //CxPlatTimeUs64();
+    AckEvent.TimeNow = 1000000;
     AckEvent.LargestAck = 5;
     AckEvent.LargestSentPacketNumber = 10;
     AckEvent.NumRetransmittableBytes = 5000;


src/core/unittest/CubicTest.cpp
```diff
@@ -0,0 +1,808 @@
...
+//
+// Test 1: Comprehensive initialization verification
+// Scenario: Verifies CubicCongestionControlInitialize correctly sets up all CUBIC state
+// including settings, function pointers, state flags, HyStart fields, and zero-initialized fields.
+// This consolidates basic initialization, function pointer, state flags, HyStart, and zero-field checks.
+//
+TEST(CubicTest, InitializeComprehensive)
+{
+    QUIC_CONNECTION Connection;
+    QUIC_SETTINGS_INTERNAL Settings;
```
guhetier: nit: Since this is C++, a zero-init can simply by `QUIC_SETTINGS_INTERNAL Settings{};`

src/core/unittest/CubicTest.cpp
```diff
@@ -0,0 +1,808 @@
...
+#ifdef QUIC_CLOG
+#include "CubicTest.cpp.clog.h"
+#endif
+
+//
+// Helper to create a minimal valid connection for testing CUBIC initialization.
+// Uses a real QUIC_CONNECTION structure to ensure proper memory layout when
+// QuicCongestionControlGetConnection() does CXPLAT_CONTAINING_RECORD pointer arithmetic.
+//
+static void InitializeMockConnection(QUIC_CONNECTION* Connection, uint16_t Mtu)
```
guhetier: nit: We annotate every pointer with SAL (`_Inout_` here).

Since this is C++, you can also make it a `QUIC_CONNECTION&` to make the semantic clear.

src/core/unittest/CubicTest.cpp
```diff
@@ -0,0 +1,808 @@
...
+
+    // Verify HyStart fields
+    ASSERT_EQ(Cubic->HyStartState, HYSTART_NOT_STARTED);
+    ASSERT_EQ(Cubic->HyStartRoundEnd, 0u);
+    ASSERT_EQ(Cubic->HyStartAckCount, 0u);
+    ASSERT_EQ(Cubic->MinRttInLastRound, UINT64_MAX);
+    ASSERT_EQ(Cubic->MinRttInCurrentRound, UINT64_MAX);
+    ASSERT_EQ(Cubic->CWndSlowStartGrowthDivisor, 1u);
+
+    // Verify zero-initialized fields
```
guhetier: Looking at the `CubicCongestionControlInitialize` code quickly... the function actually doesn't initialize those fields at all - it assumes that they have been zeroed out at allocation time.

Whether it is a good thing or not is open to discussion, but it means currently, all the assertions below are validating `InitializeMockConnection`, not product code.

src/core/unittest/CubicTest.cpp
guhetier: (multiple occurrences in the PR)

src/core/unittest/CubicTest.cpp
```diff
@@ -0,0 +1,808 @@
...
+    AckEvent.HasLoss = FALSE;
+    AckEvent.IsLargestAckedPacketAppLimited = FALSE;
+    AckEvent.AdjustedAckTime = AckEvent.TimeNow;
+    AckEvent.AckedPackets = NULL;
+
+    Connection.CongestionControl.QuicCongestionControlOnDataAcknowledged(
+        &Connection.CongestionControl,
+        &AckEvent);
+
+    // HyStart may transition states based on RTT measurements
```
guhetier, line 766: Can't we control the RTT here?
The only call to an OS time function in cubic.c seems to be `CxPlatTimeUs64();` which I don't think interfers with this test, so we should be able to see if test the transition happen when expected.

src/core/unittest/CubicTest.cpp
```diff
@@ -0,0 +1,808 @@
...
+    ASSERT_EQ(Exemptions, 3u);
+
+    // Test GetBytesInFlightMax
+    uint32_t MaxBytes = Connection.CongestionControl.QuicCongestionControlGetBytesInFlightMax(&Connection.CongestionControl);
+    ASSERT_EQ(MaxBytes, Cubic->BytesInFlightMax);
+    ASSERT_EQ(MaxBytes, Cubic->CongestionWindow / 2);
+
+    // Test GetCongestionWindow
+    uint32_t CongestionWindow = Connection.CongestionControl.QuicCongestionControlGetCongestionWindow(&Connection.CongestionControl);
+    ASSERT_EQ(CongestionWindow, Cubic->CongestionWindow);
```
guhetier, line 373: nit: Some of these getter tests feel very low value - the assertion is basically the function implementation.

src/core/unittest/CubicTest.cpp
```diff
@@ -0,0 +1,808 @@
...
+        &Connection.CongestionControl, TimeSinceLastSend, TRUE);
+
+    // Pacing should limit the allowance to less than the full available window
+    ASSERT_GT(Allowance, 0u); // Should allow some sending
+    ASSERT_LT(Allowance, AvailableWindow); // But less than full window due to pacing
+
+    // Verify it's approximately the expected pacing calculation
+    uint32_t ExpectedPacedAllowance = (uint32_t)(((uint64_t)Cubic->CongestionWindow * TimeSinceLastSend) / Connection.Paths[0].SmoothedRtt);
+
+    // Allow some margin due to integer arithmetic and min/max clamping
```
guhetier: Shouldn't this code be deterministic? A +- 50% interval seems huge.

guhetier (approved):
As a general feedback, the generated tests seem to be very tied to the internal implementation details rather than testing using the module interface only.

Since the module is C, there is no member encapsulation, but conceptually those member are private. Using them directly in tests means that if the module is refactored, the tests have to be re-written instead of helping to validate the behavior stay consistent.

Typically, we would try to write test by setting up and validating the state through calls to the interface (as far as it is realistically possible of course), mimicking a sequence of operations that could happen in the product.

src/core/unittest/CubicTest.cpp
saikat107: Resolved

src/core/unittest/CubicTest.cpp
saikat107: Resolved

src/core/unittest/CubicTest.cpp
saikat107: Good point. Added a constant `TimeNow` for `AckEvent`, we do not need to rely on the `CxPlatTimeUs64()`

src/core/unittest/CubicTest.cpp
saikat107: That function is not initializing these fields either. These assertions are good in a way that it validates the assumption.

src/core/unittest/CubicTest.cpp
saikat107: I don't have a clear understanding on how to calculate these lower bond and the upper bound.

saikat107:
Thanks for the valuable feedback @guhetier. Addressed some of the comments in a recent change. We will add the high-level feedback you made about the behavioral test into DeepTest's working procedure.

src/core/unittest/CubicTest.cpp
guhetier: I don't know if we could find existing test vectors.
But otherwise, an easy way (since we expect a deterministic result) is to assume for now that the code is correct and used the returned value in the assert.
It will check for regression in the future.

src/core/unittest/CubicTest.cpp
guhetier: What I meant is that these asserts are valid only because `InitializeMockConnection` zero initialize the connection. And they would fail if we were to remove the `CxPlatZeroMemory` at line 51.

Whether they succeed or fail depends only on test code, not on product code: are they providing value?

saikat107: Understood. Fair enough. Removing those assertions -- only keeping the assertions to the fields explicitly assigned in the cubic.c

src/core/unittest/CubicTest.cpp
```diff
@@ -356,12 +331,10 @@ TEST(CubicTest, GetSendAllowanceWithActivePacing)
     ASSERT_GT(Allowance, 0u); // Should allow some sending
     ASSERT_LT(Allowance, AvailableWindow); // But less than full window due to pacing
 
-    // Verify it's approximately the expected pacing calculation
-    uint32_t ExpectedPacedAllowance = (uint32_t)(((uint64_t)Cubic->CongestionWindow * TimeSinceLastSend) / Connection.Paths[0].SmoothedRtt);
-
-    // Allow some margin due to integer arithmetic and min/max clamping
-    ASSERT_GE(Allowance, ExpectedPacedAllowance / 2);
-    ASSERT_LE(Allowance, ExpectedPacedAllowance * 2);
+    // Exact value is caldulated considering the current implementation is right and this test is meant to
```
guhetier: typo, calculated.

src/core/unittest/CubicTest.cpp
```diff
@@ -782,7 +745,7 @@ TEST(CubicTest, HyStart_StateTransitions)
 
     QUIC_ACK_EVENT AckEvent;
     CxPlatZeroMemory(&AckEvent, sizeof(AckEvent));
-    AckEvent.TimeNow = CxPlatTimeUs64();
+    AckEvent.TimeNow = 1000000; //CxPlatTimeUs64();
```
guhetier: nit: The comment should be deleted.

guhetier (dismissed):
A few nits, but looks good otherwise.

PR #5655, opened 2025-12-13
Add ref count tracking to CXPLAT_WORKER_POOL

anrossi:
## Description

Track the reference count sources on the CXPLAT_WORKER_POOL object in debug builds. Related to #5570.

## Testing

Existing tests and CI should cover this.

## Documentation

No.

### DIFF
--- a/src/platform/platform_worker.c
+++ b/src/platform/platform_worker.c
@@ -347,6 +347,8 @@ CxPlatWorkerPoolCreate(
 #if DEBUG
     CxPlatRefInitializeMultiple(WorkerPool->RefTypeBiasedCount, CXPLAT_WORKER_POOL_REF_COUNT);
     CxPlatRefIncrement(&WorkerPool->RefTypeBiasedCount[RefType]);
+#else
+    UNREFERENCED_PARAMETER(RefType);
 #endif
 
     return WorkerPool;


src/core/library.c
```diff
@@ -2772,7 +2772,7 @@ MsQuicExecutionCreate(
         //
         // Clean up any previous worker pool and create a new one.
         //
-        CxPlatWorkerPoolDelete(MsQuicLib.WorkerPool);
+        CxPlatWorkerPoolDelete(MsQuicLib.WorkerPool, CXPLAT_WORKER_POOL_REF_EXTERNAL);
```
guhetier, line 2775: This is guaranteed to be external because the worker pool would have been allocated by the lib only during the lazy init, is that right?

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -1107,7 +1107,7 @@ void SetupAndFuzz() {
 
     CxPlatSocketDelete(Binding);
     CxPlatDataPathUninitialize(Datapath);
-    CxPlatWorkerPoolDelete(WorkerPool);
+    CxPlatWorkerPoolDelete(WorkerPool, CXPLAT_WORKER_POOL_REF_LIBRARY);
```
guhetier: These refs in tools make me think that we should:
- Add a ref type parameter to `CxPlatWorkerPoolCreate`
- Create a ref type for tools

Just to be consistent

src/platform/platform_worker.c
```diff
@@ -417,12 +436,24 @@ CxPlatWorkerPoolCreateExternal(
...
+#if DEBUG
+        CXPLAT_DBG_ASSERT(!CxPlatRefDecrement(&WorkerPool->RefTypeBiasedCount[RefType]));
+#else
+        UNREFERENCED_PARAMETER(RefType);
+#endif
         CxPlatRundownReleaseAndWait(&WorkerPool->Rundown);
 
+#if DEBUG
+        for (uint32_t i = 0; i < CXPLAT_WORKER_POOL_REF_COUNT; i++) {
+            CXPLAT_TEL_ASSERT(WorkerPool->RefTypeBiasedCount[i] == 1);
```
guhetier: This should be a debug assert

src/core/library.c
anrossi: Correct. If the application triggers the lazy init (by calling `RegistrationOpen()`) and then tries to call `ExecutionCreate()`, this will fail. This path should only happen if the application calls `ExecutionCreate()` twice, where we can assume that the previous WorkerPool was from the previous `ExecutionCreate()` call.

PR #5656, opened 2025-12-13
Fix static build with CMAKE_DEBUG_POSTFIX

Maiyu27:
### Description
This patch fixes an issue where building a static version of msquic with `CMAKE_DEBUG_POSTFIX` set fails during the linking step. Previously, the build would attempt to link a library file without the postfix, causing a fatal LNK1181 error. This change ensures the correct filename with the postfix is used for static builds.

### Steps to reproduce
1. Configure a Debug build with `QUIC_BUILD_SHARED=OFF` and `CMAKE_DEBUG_POSTFIX=d`.
2. Attempt to build msquic.
3. Observe the linking failure without this fix.

### Expected behavior
The build should succeed using the correct library filename with the debug postfix.

### Actual behavior
The build fails during linking because the library file without the postfix is referenced.

### Notes
- Verified locally.
- Minor change in CMakeLists.txt to respect `CMAKE_<CONFIG>_POSTFIX` for static builds.


src/bin/CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ if(BUILD_SHARED_LIBS)
     endif()
 else()
     add_library(msquic_static STATIC static/empty.c)
+    set_target_properties(msquic_static PROPERTIES OUTPUT_NAME "${QUIC_LIBRARY_NAME}${CMAKE_DEBUG_POSTFIX}")
```
guhetier, line 23: The bug / PR description mentions the issue when `QUIC_BUILD_SHARED=ON`, but this is in the `QUIC_BUILD_SHARED=OFF` path. A similar change might be needed above to add the postfix

j-schultz: Sorry, that was an obvious typo in the issue. `QUIC_BUILD_SHARED=ON` (the default) obviously implies building a shared version of the library.

Issue #5585, opened 2025-11-13
Test Crashes in Windows Pipeline Runner Don't Collect Dumps


anrossi:
### Describe the bug

If the test crashes during a pipeline run on Windows OSes, no dump file is collected.
This used to work in the past and is harming our ability to debug recent test crashes.

### Affected OS

- [x] Windows
- [ ] Linux
- [ ] macOS
- [ ] Other (specify below)

### MsQuic version

main

### Steps taken to reproduce bug

1. CI Pipeline run on Windows Crashes

### Expected behavior

2. Crash Dump is available in the test results.

### Actual outcome

2. No crash dump is available in the test results. (repros on SpinQuic, BVT, and down level testruns)

PR #5665, opened 2025-12-15
Use ProcDump instead of WER to collect user mode crash dumps

ProjectsByJackHe:
## Description

Crash dump collection in spin tests has been broken for the longest time.

The scripts always tried to use WER, but WER is notoriously brittle in CI environments. Even after setting the HKEY regkeys properly, due to group policy or other security enforcement policies, WER in CIs has been egregiously broken.

Let us use ProcDump instead.

Fixes: #5585 

## Testing

I forked the topic branch of this PR and intentionally added a NULL pointer de-ref.

Then, I kicked off a manual dispatch and verified the dump is generated properly: https://github.com/microsoft/msquic/actions/runs/20246218870

## Documentation

N/A

### DIFF
--- a/scripts/run-executable.ps1
+++ b/scripts/run-executable.ps1
@@ -416,16 +416,27 @@ function Wait-Executable($Exe) {
         $Exe.Process.WaitForExit()
         $exitCode = $Exe.Process.ExitCode
 
-        # When using ProcDump, the process is ProcDump itself, not the target.
-        # The ProcDump exit codes does not map cleanly to the status of the target process, so we will just log it.
-        if (!$UseProcDump -or !$IsWindows) {
+        # Extract target process exit code when using ProcDump
+        if ($UseProcDump -and $IsWindows) {
+            # ProcDump writes the target process exit code in its output
+            # Look for pattern like "[13:17:55]Process Exit: PID 39412, Exit Code 0x00000000"
+            if ($stdout -match "Process Exit:.*Exit Code 0x([0-9a-fA-F]+)") {
+                $targetExitCodeHex = $Matches[1]
+                $targetExitCode = [Convert]::ToInt32($targetExitCodeHex, 16)
+                Log "Target process exit code: $targetExitCode (0x$targetExitCodeHex) (ProcDump wrapper exit code: $exitCode)"
+                if ($targetExitCode -ne 0) {
+                    LogErr "Target process had nonzero exit code: $targetExitCode"
+                    $KeepOutput = $true
+                }
+            } else {
+                LogWrn "Could not parse target exit code from ProcDump output. ProcDump exit code: $exitCode"
+            }
+        } else {
             # Normal process (not wrapped by ProcDump)
             if ($exitCode -ne 0) {
                 LogErr "Process had nonzero exit code: $exitCode"
                 $KeepOutput = $true
             }
-        } else {
-            Log "ProcDump exit code: $exitCode"
         }
 
         # List files in log directory


scripts/run-executable.ps1
```diff
@@ -124,6 +127,40 @@ function Test-Administrator
...
+    $pd = Get-ProcDumpPath
+    if ($pd) { return $pd }
+
+    if (!$IsWindows) { throw "ProcDump is Windows-only." }
+
+    $toolDir = Join-Path $RootDir "artifacts" "tools" "procdump"
+    New-Item -ItemType Directory -Force -Path $toolDir | Out-Null
+
+    $zipPath = Join-Path $toolDir "procdump.zip"
+    $url = "https://download.sysinternals.com/files/Procdump.zip"  # official Sysinternals download
```
guhetier: I have concerns than trying to download an exe from the internet get pretty flaky.
Are we already doing this for other tools? Are there more stable ways to install it? (winget?)

scripts/run-executable.ps1
```diff
@@ -124,6 +127,40 @@ function Test-Administrator
...
+    $cmd = Get-Command "procdump.exe" -ErrorAction SilentlyContinue
+    if ($cmd) { return $cmd.Source }
+
+    $candidate = Join-Path $RootDir "artifacts" "tools" "procdump" "procdump.exe"
+    if (Test-Path $candidate) { return $candidate }
+
+    return $null
+}
+
+function Ensure-ProcDump {
```
guhetier: This returns "procdump.exe" path. Why do we need the `Get-ProcDumpPath` function, that might end up not being in sync? Can't the caller simply store the path in a variable?

anrossi: This should be moved to be part of the prepare-machine.ps1 script.

scripts/run-executable.ps1
```diff
@@ -382,6 +447,23 @@ function Wait-Executable($Exe) {
             LogErr "Process had nonzero exit code: $($Exe.Process.ExitCode)"
             $KeepOutput = $true
         }
+
+        # When using ProcDump, give it a moment to finish writing dumps
```
guhetier: Any reason we don't need this for WER?

scripts/run-executable.ps1
```diff
@@ -382,6 +447,23 @@ function Wait-Executable($Exe) {
...
+        if ($UseProcDump -and $IsWindows) {
+            Write-Host "Waiting for dump files to be written..."
+            Start-Sleep -Seconds 2
```
guhetier: Is 2 seconds enough? Can't we wait more, and only if the exit code was non-zero?

And if we use procdump to launch the executable, do we actually need to wait at all? I expect procdump would not return until it is done writting.

scripts/run-executable.ps1
ProjectsByJackHe: moved

ProjectsByJackHe: I moved this logic to prepare-machine and simplified the code a bit more

scripts/run-executable.ps1
```diff
@@ -227,11 +264,39 @@ function Start-Executable {
...
+                    $pinfo.Arguments = "-accepteula -ma -e -x `"$LogDir`" `"$Path`" $Arguments"
+                    Write-Host "ProcDump command: $($pinfo.FileName) $($pinfo.Arguments)"
+                } catch {
+                    Write-Warning "Failed to setup ProcDump: $_"
+                    Write-Warning "Falling back to direct execution without crash dumps"
+                    $pinfo.FileName = $Path
+                    $pinfo.Arguments = $Arguments
+                }
+            } else {
+                # Direct execution with WER
```
guhetier: If WER is broken and not expected to be fixed, should we simply simplify the script by removing support and rely on procdump only?

scripts/run-executable.ps1
ProjectsByJackHe: fair points, I looked up some online examples that included a timeout. And I do not know why WER did not have a timeout. But from local testing this is indeed unnecessary so I will remove it.

scripts/run-executable.ps1
ProjectsByJackHe: dup of the other thread on the same line

scripts/run-executable.ps1
ProjectsByJackHe: Locally WER works sort of. I guess the question is, since the online download of procdump has been moved to prepare-machine, do we want users to have to download procdump in order to debug locally? It's not a huge deal.

scripts/run-executable.ps1
ProjectsByJackHe: Not as flaky as our CI ;) 

Procdump is pretty famous and a lot of other companies / teams rely on it for CI process dump collection.

If indeed we are seeing download failures often, then we can move to more of an in-house solution.

scripts/run-executable.ps1
```diff
@@ -124,6 +127,17 @@ function Test-Administrator
     (New-Object Security.Principal.WindowsPrincipal $user).IsInRole([Security.Principal.WindowsBuiltinRole]::Administrator)
 }
 
+function Get-ProcDumpPath {
+    $cmd = Get-Command "procdump.exe" -ErrorAction SilentlyContinue
+    if ($cmd) { return $cmd.Source }
```
guhetier: Are you sure this returns a path? When running `Get-Command`, the souce show "Appx", "Dism", etc...

If we expect procdump in a known path, we should just use that. Otherwise, we can add it to the PATH when we install it and not have to duplicate the install path.

scripts/run-executable.ps1
guhetier: If you want to locally debug a test, you are expected to run `Prepare-Machine -ForTest` to setup everything, or otherwise, to deal with your dump collection.

I think keeping the script simple is worth it.

ProjectsByJackHe: sounds good

scripts/run-executable.ps1
ProjectsByJackHe: Let's use the known path and assume all users will run prepare-machine first.

scripts/run-executable.ps1
```diff
@@ -124,6 +127,15 @@ function Test-Administrator
     (New-Object Security.Principal.WindowsPrincipal $user).IsInRole([Security.Principal.WindowsBuiltinRole]::Administrator)
 }
 
+function Get-ProcDumpPath {
+
+    $toolDir = Join-Path $RootDir "artifacts" "tools" "procdump"
+    $candidate = Join-Path $toolDir "procdump.exe"
```
guhetier:
```suggestion
    $candidate = Join-Path $RootDir "artifacts" "tools" "procdump" "procdump.exe"
```

scripts/run-executable.ps1
```diff
@@ -378,10 +415,30 @@ function Wait-Executable($Exe) {
...
+        # When using ProcDump, the process is ProcDump itself, not the target.
+        # The ProcDump exit codes does not map cleanly to the status of the target process, so we will just log it.
+        if (!$UseProcDump -or !$IsWindows) {
+            # Normal process (not wrapped by ProcDump)
+            if ($exitCode -ne 0) {
+                LogErr "Process had nonzero exit code: $exitCode"
+                $KeepOutput = $true
+            }
+        } else {
+            Log "ProcDump exit code: $exitCode"
```
guhetier: If the process return an error code but doesn't crash, does this mean we don't set `KeepOutput` to true, and won't be keeping logs?

If we can't know the target return code when called with procdump, maybe we should use it as a monitor instead (not `-x` + run the target separately)

ProjectsByJackHe: If the process returns an error code, ProcDump will generate a dump. There is additional logic in the script that detects any dmps generated and that would fail the run and upload the logs.

ProjectsByJackHe: I am not sure if doing 2 concurrent runs in the same script is the way to go.

I think it's better to abstract that logic to the YAML and do 2 sequential runs: 1 with procdump 1 without if we really need the target exit code.

guhetier: > If the process returns an error code, ProcDump will generate a dump

Are you sure about that? If the program returns (not crash), what is there to dump?

> I am not sure if doing 2 concurrent runs in the same script is the way to go.

That's not what I am suggesting. `ProcDump` without the `-x` option runs in a "monitor" mode. It will wait for a process of the expected name to start, attach to it, and create dumps if it fails with what procdump is configured to monitor.

ProjectsByJackHe: > > If the process returns an error code, ProcDump will generate a dump
> 
> Are you sure about that? If the program returns (not crash), what is there to dump?
> 
> > I am not sure if doing 2 concurrent runs in the same script is the way to go.
> 
> That's not what I am suggesting. `ProcDump` without the `-x` option runs in a "monitor" mode. It will wait for a process of the expected name to start, attach to it, and create dumps if it fails with what procdump is configured to monitor.

Launching procdump as a sidecar was how I tried implementing it originally, but it was running into too many race conditions in the CI, making it unreliable. If the main issue is we need to extract what the inner process's exit code is, we can just parse it from stdout. Procdump prints that there.

And good point about the non zero exit code. In those instances, the only log persisted was `stdout.txt`, which is the least bit helpful. That's how the stress tests was working before. I thought it was ok to discard those logs, but the entire job passing would indeed be misleading. I modified the script now to parse the procdump reported exit code of the inner process, and fail the pipeline if it's non-zero.

guhetier:
Looks good, thanks for finding a good solution to that dump collection problem!

PR #5668, opened 2025-12-16
Fix spurious failure in Handshake/WithHandshakeArgs1.ResumeRejection

guhetier:
## Description

`QuicTestConnect` can fail spuriously if the test thread wins a race against a shutdown notification.

- Add a wait for the peer shutdown notification before asserting on whether it was received
- Remove useless scopes

## Testing

CI

## Documentation

N/A

### DIFF
--- a/src/test/lib/HandshakeTest.cpp
+++ b/src/test/lib/HandshakeTest.cpp
@@ -227,222 +227,219 @@ QuicTestConnect(
     PrivateTransportHelper TpHelper(MultiPacketClientInitial, !!ResumptionTicket);
     RandomLossHelper LossHelper(RandomLossPercentage);
 
+    if (SessionResumption == QUIC_TEST_RESUMPTION_REJECTED) {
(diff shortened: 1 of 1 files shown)


src/test/lib/HandshakeTest.cpp
```diff
@@ -227,223 +227,220 @@ QuicTestConnect(
...
-            } else {
-                Server->Shutdown(QUIC_CONNECTION_SHUTDOWN_FLAG_SILENT, 0);
-            }
+            TEST_FALSE(Client.GetPeerClosed());
+            TEST_FALSE(Client.GetTransportClosed());
         }
     }
+
+    if (RandomLossPercentage == 0) {
+        Server->WaitForPeerClose();
```
guhetier, line 438: Github shows this diff horribly.
This is the only line added, everything else is removing some indentation level and braces (the "hide whitespace" option makes it better)

mtfriesen (approved):

PR #5673, opened 2025-12-16
Fix invalid handle access for RegistrationCloseCleanup Events

ProjectsByJackHe:
## Description

When MsQuic is first initialized, there may not be enough memory to allocate the datapath layer.

This happens so rarely that there is little to no coverage for error handling when this happens.

But when it happens, you skip directly to the exit condition, and that tries free-ing memory that was never initialized.

So fix up the logic so cleanup only happens if initialized.

Fixes: https://github.com/microsoft/msquic/issues/5491

## Testing

Stress tests.

## Documentation

No

### DIFF
--- a/src/core/library.c
+++ b/src/core/library.c
@@ -607,11 +607,6 @@ Error:
             CxPlatThreadDelete(&MsQuicLib.RegistrationCloseCleanupWorker);
             MsQuicLib.RegistrationCloseCleanupWorker = 0;
         }
-        if (PlatformInitialized) {
-            CxPlatRundownUninitialize(&MsQuicLib.RegistrationCloseCleanupRundown);
-            CxPlatEventUninitialize(MsQuicLib.RegistrationCloseCleanupEvent);
-            CxPlatLockUninitialize(&MsQuicLib.RegistrationCloseCleanupLock);
-        }
         if (MsQuicLib.Storage != NULL) {
             CxPlatStorageClose(MsQuicLib.Storage);
             MsQuicLib.Storage = NULL;
@@ -621,6 +616,9 @@ Error:
             MsQuicLib.DefaultCompatibilityList = NULL;
         }
         if (PlatformInitialized) {
+            CxPlatRundownUninitialize(&MsQuicLib.RegistrationCloseCleanupRundown);
+            CxPlatEventUninitialize(MsQuicLib.RegistrationCloseCleanupEvent);
+            CxPlatLockUninitialize(&MsQuicLib.RegistrationCloseCleanupLock);
             CxPlatDispatchRwLockUninitialize(&MsQuicLib.StatelessRetry.Lock);
             CxPlatUninitialize();
         }


src/core/library.c
```diff
@@ -528,6 +529,8 @@ MsQuicLibraryInitialize(
     CxPlatListInitializeHead(&MsQuicLib.RegistrationCloseCleanupList);
     CxPlatRundownInitialize(&MsQuicLib.RegistrationCloseCleanupRundown);
 
+    RegistrationCloseCleanupInitialized = TRUE;
```
guhetier: Instead of adding a variable and runtime logic, can we simply move the initialization of these up before to any failure path?

ProjectsByJackHe: I think that would break the pattern of error handling for the rest of this entire function just to preserve the old error handling code. I would argue it's cleaner to follow the rest of the code in the error handling: if resource is allocated, deallocate it.

ProjectsByJackHe: And also, as a preference, I think CxPlatInitialize() should always run first. The first step before initializing locks and events is to first ensure that there is enough memory on the heap / can we even initialize the datapath :D

guhetier: Yes, `CxPlatInitialize()` needs to stay first (if only for logging and other setup), but you can simply move this before `PlatformInitialized = TRUE;` and use that variable to control the cleanup. `CxPlatStorageOpen` can come after these handles.

It will also help to keep the cleanup in reverse order from the init, which is part of the problem.
No point multiplying the cleanup variables :)

ProjectsByJackHe: That makes more sense, I thought you meant to do the lock/event init before we allocate the datapath memory :)

src/core/library.c
```diff
@@ -606,9 +606,11 @@ MsQuicLibraryInitialize(
             CxPlatThreadDelete(&MsQuicLib.RegistrationCloseCleanupWorker);
             MsQuicLib.RegistrationCloseCleanupWorker = 0;
         }
-        CxPlatRundownUninitialize(&MsQuicLib.RegistrationCloseCleanupRundown);
-        CxPlatEventUninitialize(MsQuicLib.RegistrationCloseCleanupEvent);
-        CxPlatLockUninitialize(&MsQuicLib.RegistrationCloseCleanupLock);
+        if (PlatformInitialized) {
+            CxPlatRundownUninitialize(&MsQuicLib.RegistrationCloseCleanupRundown);
```
guhetier: Please move these 3 with the other cleanups controled by `PlatformInitialized`.

Generally speaking, cleanups should be done in the reverse order of inits: the lifetime of your variables will match a virtual "scope", and it will avoid unexpected dependency issues.

src/core/library.c
```diff
@@ -502,9 +502,17 @@ MsQuicLibraryInitialize(
     CxPlatToeplitzHashInitialize(&MsQuicLib.ToeplitzHash);
 
     CxPlatDispatchRwLockInitialize(&MsQuicLib.StatelessRetry.Lock);
-    PlatformInitialized = TRUE;
 
     CxPlatZeroMemory(&MsQuicLib.Settings, sizeof(MsQuicLib.Settings));
+    MsQuicLibraryReadSettings(NULL); // NULL means don't update registrations.
```
guhetier: This one line depends on the storage to be initialized already, it needs to stay after `CxPlatStorageOpen`.
(it would have been nice if it took the storage as a parameter to reflect that dependency)

ProjectsByJackHe: Done

src/core/library.c
ProjectsByJackHe: Done

PR #5674, opened 2025-12-17
Migrate more test to the new kernel invocation pattern

guhetier:
## Description

Migrate all tests taking no parameters, using Copilot.
The array of sizes and the IOCTL definitions are not cleaned up yet as remaining tests still depend on the proper indexing.

## Testing

CI

## Documentation

N/A

### DIFF
--- a/src/test/bin/quic_gtest.cpp
+++ b/src/test/bin/quic_gtest.cpp
@@ -264,7 +264,6 @@ TEST(ParameterValidation, ValidateGlobalParam) {
     TestLogger Logger("QuicTestValidateGlobalParam");
     if (TestingKernelMode) {
         ASSERT_TRUE(InvokeKernelTest(FUNC(QuicTestGlobalParam)));
-        ASSERT_TRUE(DriverClient.Run(IOCTL_QUIC_RUN_VALIDATE_GLOBAL_PARAM));
     } else {
         QuicTestGlobalParam();
     }
@@ -273,7 +272,7 @@ TEST(ParameterValidation, ValidateGlobalParam) {
 TEST(ParameterValidation, ValidateCommonParam) {
     TestLogger Logger("QuicTestValidateCommonParam");
     if (TestingKernelMode) {
-        ASSERT_TRUE(DriverClient.Run(IOCTL_QUIC_RUN_VALIDATE_COMMON_PARAM));
+        ASSERT_TRUE(InvokeKernelTest(FUNC(QuicTestCommonParam)));
     } else {
         QuicTestCommonParam();
     }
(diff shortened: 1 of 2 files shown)


src/test/bin/winkernel/control.cpp
```diff
@@ -645,6 +645,82 @@ ExecuteTestRequest(
...
+    RegisterTestFunction(QuicTestConfigurationParam);
+    RegisterTestFunction(QuicTestListenerParam);
+    RegisterTestFunction(QuicTestConnectionParam);
+    RegisterTestFunction(QuicTestTlsParam);
+    RegisterTestFunction(QuicTestStreamParam);
+    RegisterTestFunction(QuicTestGetPerfCounters);
+    RegisterTestFunction(QuicTestValidateConfiguration);
+    RegisterTestFunction(QuicTestValidateListener);
+    RegisterTestFunction(QuicTestValidateConnection);
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
```
guhetier, line 661: Those are probably not needed, even if we run the downlevel tests against a more recent MsQuic that broke the preview feature, it isn't a problem the driver is able to invoke the test.
Same for all defines that copilot added in that list, I think.

src/test/bin/quic_gtest.cpp
```diff
@@ -266,7 +266,7 @@ TEST(ParameterValidation, ValidateApi) {
 TEST(ParameterValidation, ValidateRegistration) {
     TestLogger Logger("QuicTestValidateRegistration");
     if (TestingKernelMode) {
-        ASSERT_TRUE(DriverClient.Run(IOCTL_QUIC_RUN_VALIDATE_REGISTRATION));
+        ASSERT_TRUE(InvokeTestInKernel("QuicTestValidateRegistration"));
```
mtfriesen: Can we make these look like plain function names/references rather than strings?

mtfriesen: That is, make this look like: `InvokeTestInKernel(QuicTestValidateRegistration);`

mtfriesen: Dang, I meant to comment on the original PR.

guhetier: I was considering it.
It'll take an extra layer of macro, but it might let me add more compile time validation on the parameter type. I'll give it a stab.

mtfriesen: Yeah, IDK if it's a lesser evil. I don't have strong feelings either way.

guhetier: The best I can manage is:

`ASSERT_TRUE(InvokeKernelTest(FUNC(QuicTestValidateRegistration)));` / `ASSERT_TRUE(InvokeKernelTest(FUNC(QuicTestValidateRegistration), Param));`
or
`ASSERT_TRUE(InvokeKernelTest(QuicTestValidateRegistration));` / `ASSERT_TRUE(InvokeKernelTestWithParam(QuicTestValidateRegistration, Param));`

the problem being there is no good way to handle optional parameters and overloads with macros.

I am leaning towards the first option (less macro), but I don't like either much.

mtfriesen: Yeah, let's skip the macros.

guhetier: Ah, I saw this too late. Well, I can undo it and you get to see it in the diff in the meantime.
The one good point about macros is that it allows to check at compile time the function name exists and is compatible with the argument type.

mtfriesen:
@guhetier is this PR a duplicate?

guhetier:
> @guhetier is this PR a duplicate?

Just a draft as an example, I'll close it. Not meant to be merged.
I wish Github made it easier to distinguish drafts from opened PRs.

PR #5678, opened 2025-12-17
Fix race between causing spurious failure in the ConnectTcp datapath test

### LINKED ISSUE #5679: [CI - FAILURE] ConnectTcp fails
Test failure observed on multiple PR CI runs

guhetier:
## Description

`CxPlatDataPathSocketProcessAcceptCompletion` can race with `CxPlatSocketDelete` if the test code delete the socket right after it receive the "accept" callback.

`CxPlatSocketDelete` doesn't wait on the rundown ref when `IoStarted` is not true (I don't know why, and I wonder if we couldn't simply remove that special case, but that is out of scope of this PR).

This would allow a socket to be already in the "uninitialzed" state when then end of `CxPlatDataPathSocketProcessAcceptCompletion` tries to post the first receive SQE on it.

Fixes #5679

## Testing

CI. The failure was observed regularly on recent PR CI runs but is not consistent, so we may have to keep an eye for a time.

## Documentation

N/A

### DIFF
--- a/src/platform/datapath_winuser.c
+++ b/src/platform/datapath_winuser.c
@@ -2661,6 +2661,7 @@ CxPlatDataPathSocketProcessAcceptCompletion(
         //
         AcceptSocketProc->IoStarted = TRUE;
         if (!CxPlatRundownAcquire(&AcceptSocketProc->RundownRef)) {
+            AcceptSocketProc = NULL;
             goto Error;
         }
 


src/platform/datapath_winuser.c
```diff
@@ -2655,7 +2655,11 @@ CxPlatDataPathSocketProcessAcceptCompletion(
             0,
             "AcceptEx Completed!");
 
-
+        //
+        // Mark IO Started before taking a rundown reference,
+        // otherwise the cleanup might not wait for the rundown.
+        //
+        AcceptSocketProc->IoStarted = TRUE;
         if (!CxPlatRundownAcquire(&AcceptSocketProc->RundownRef)) {
```
anrossi, line 2663: if it fails to acquire the rundown, it means the socket is being shutdown, and the IO should be canceled, and `IoStarted` should be set to false.

guhetier: I am not sure I understand.
`IoStarted` is never set back to FALSE in other path. It only enables a potentially faster cleanup by closing the socket directly. I am not sure why we even have that extra complexity, as far as I can see, it doesn't seem worth it.

Although your comment makes me notice that if we fail to take the ref, we still deref it in the `Error` cleanup, which seems bad?

guhetier: Discussed offline - there is no problem setting IoStarted to TRUE here.

Fixed the spurious deref.

mtfriesen (approved):

guhetier:
Interesting, the CI is catching a leak in the epoll datapath which is not modified in this PR.
It's likely an unrelated issue but I'll have a look while at it.

PR #5680, opened 2025-12-18
Migrate a test with multiple parameters

guhetier:
## Description

Demo how it would be for migrating a test with multiple params.

## Testing

CI

## Documentation

N/A

### DIFF
--- a/src/test/MsQuicTests.h
+++ b/src/test/MsQuicTests.h
@@ -36,6 +36,15 @@ extern QUIC_CREDENTIAL_CONFIG ClientCertCredConfig;
 #endif
 extern char CurrentWorkingDirectory[MAX_PATH + 1];
 
+struct FamilyArgs2 {
+    int Family;
+};
+
+struct KeyUpdateArgs2 {
+    int Family;
+    uint8_t RandomLossPercentage;
+};
+
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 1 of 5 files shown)


src/test/MsQuicTests.h
```diff
@@ -550,13 +558,9 @@ QuicTestReceiveResumeNoData(
...
     );
 
-struct FamilyArgs2 {
-    int Family;
-};
-
 void
 QuicTestAckSendDelay(
-    const FamilyArgs2& Params
+    const FamilyArgs2& Family
```
guhetier, line 563: Ignore this, just a failed merge when I extracted just the relevant commit.

src/test/bin/quic_gtest.h
```diff
@@ -577,26 +577,22 @@ class WithSend0RttArgs2 : public testing::Test,
...
     return o <<
         (args.Family == 4 ? "v4" : "v6") << "/" <<
         args.RandomLossPercentage;
 }
 
-class WithKeyUpdateArgs2 : public testing::Test,
+struct WithKeyUpdateArgs2 : public testing::Test,
     public testing::WithParamInterface<KeyUpdateArgs2> {
+
+    static ::std::vector<KeyUpdateArgs2> Generate() {
```
guhetier, line 589: Note that this is moved to `WithKeyUpdateArgs2` because:
- it uses the STL, so it can't go in MsQuicTest.h
- it makes more sense to have the generator in this class anyway: different tests might take the same parameter type, but want to generate a different set of values.

PR #5682, opened 2025-12-18
New packaging pipeline + Improve current PR based one

liveans:
## Description

Add new local container-based package validation script + on-demand container-based package validation CI + improve current PR-based validation.

To test new pipeline you can fork: [liveans/msquic-package-validation](https://github.com/liveans/msquic-package-validation) and run it

## Testing

CI + Local Runs

## Documentation

No

### DIFF
--- a/scripts/validate-msquic-docker.ps1
+++ b/scripts/validate-msquic-docker.ps1
@@ -205,7 +205,8 @@ function Initialize-PackageFolderStructure {
         if (-not (Test-Path $distroPath)) {
             New-Item -ItemType Directory -Path $distroPath -Force | Out-Null
             Write-Host "  Created: $distro/" -ForegroundColor Green
-        } else {
+        }
+        else {
             Write-Host "  Exists:  $distro/" -ForegroundColor Yellow
         }
     }
@@ -232,9 +233,9 @@ function Get-PackageVersion {
     # RPM: libmsquic-2.4.8-1.x86_64.rpm -> 2.4.8
     if ($Filename -match 'libmsquic[_-](\d+)\.(\d+)\.(\d+)') {
         return @{
-            Major = [int]$Matches[1]
-            Minor = [int]$Matches[2]
-            Patch = [int]$Matches[3]
+            Major    = [int]$Matches[1]
+            Minor    = [int]$Matches[2]
+            Patch    = [int]$Matches[3]
             Original = $Filename
         }
     }
@@ -261,9 +262,9 @@ function Get-LatestPackage {
         return $null
     }
     
-    $sorted = $versionedList | Sort-Object -Property @{Expression={$_.Major}; Descending=$true}, 
-                                                      @{Expression={$_.Minor}; Descending=$true}, 
-                                                      @{Expression={$_.Patch}; Descending=$true}
+    $sorted = $versionedList | Sort-Object -Property @{Expression = { $_.Major }; Descending = $true }, 
+    @{Expression = { $_.Minor }; Descending = $true }, 
+    @{Expression = { $_.Patch }; Descending = $true }
     
     # Handle single item case
     if ($sorted -is [hashtable]) {
@@ -305,7 +306,7 @@ function Get-PackageByVersion {
 }
 
 # Function to download packages from packages.microsoft.com
-function Download-LatestPackages {
+function Download-Packages {
     param(
         [string]$BasePath,
         [bool]$UseTestingRepo = $false,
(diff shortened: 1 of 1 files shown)


scripts/validate-msquic-docker.ps1
```diff
@@ -0,0 +1,1868 @@
...
+            '# Install findutils if not available (needed for minimal images like openSUSE Tumbleweed)',
+            'if ! command -v find > /dev/null 2>&1; then',
+            '    if command -v zypper > /dev/null 2>&1; then',
+            '        echo "Installing findutils..."',
+            '        zypper --non-interactive --no-gpg-checks install findutils > /dev/null 2>&1 || true',
+            '    fi',
+            'fi',
+            "# Look in distro-specific folder first, then fall back to root",
+            "RPM_FILE=`$(find /packages/${distroFolder} -name '*${rpmArch}.rpm' -type f 2>&1 | grep -v 'No such file' | head -1)",
+            'if [ -z "$RPM_FILE" ]; then RPM_FILE=$(find /packages -maxdepth 1 -name "*' + $rpmArch + '.rpm" -type f | head -1); fi',
```
guhetier, line 1282: For the package validation, is there a specific reason you download the package then install it from the local folder instead of installing it directly from the repository?

scripts/validate-msquic-docker.ps1
```diff
@@ -0,0 +1,1868 @@
...
+        '    print(f"ERROR: Failed to load libmsquic: {e}")',
+        '    sys.exit(1)',
+        'PYEOF',
+        '    if [ $? -ne 0 ]; then',
+        '        echo "ERROR: Library validation failed"',
+        '        exit 1',
+        '    fi',
+        'else',
+        '    echo "WARNING: python3 not available, skipping library API validation"',
+        '    echo "Library installation was verified via ldconfig"',
```
guhetier: ldconfig will validate that the lib is installed, but it could still fail to load, right?
Should we fail in this case? A python3 availability issue could otherwise disable this validation with the pipeline still succeeding

.github/workflows/package-linux.yml
guhetier, line 37: @gaurav2699 This pipeline ends up building and validating linux packages, but the scripts used in ADO for the official packages are different.

How much do you think we could unify the two?

Validating packages in Github is nice to shift the validation left, but not that useful if the package we validate here and the one published end up being different.

.github/workflows/validate-linux-packages.yml
```diff
@@ -0,0 +1,54 @@
...
+concurrency:
+  group: validate-linux-packages-${{ github.sha }}
+  cancel-in-progress: true
+
+permissions: read-all
+
+jobs:
+  validate:
+    name: Validate
+    uses: ./.github/workflows/validate-linux-packages-reuse.yml
```
guhetier, line 48: Question: Why do you use a reusable workflow here, when this is the only place using it?

scripts/validate-msquic-docker.ps1
```diff
@@ -0,0 +1,1868 @@
...
+
+        $downloadedFromDistro = $false
+
+        foreach ($urlInfo in $urlsToTry) {
+            if ($downloadedFromDistro) { break }
+
+            $baseUrl = $urlInfo.url
+            $urlLabel = $urlInfo.label
+
+        try {
```
guhetier: nit: The try should be indented?

scripts/validate-msquic-docker.ps1
```diff
@@ -0,0 +1,1868 @@
...
+                    $amd64Files = $allDebFiles | Where-Object { $_ -match '_amd64\.deb$' -and $_ -match '~rc' }
+                    $arm64Files = $allDebFiles | Where-Object { $_ -match '_arm64\.deb$' -and $_ -match '~rc' }
+                    $armhfFiles = $allDebFiles | Where-Object { $_ -match '_armhf\.deb$' -and $_ -match '~rc' }
+                } else {
+                    $amd64Files = $allDebFiles | Where-Object { $_ -match '_amd64\.deb$' -and $_ -notmatch '~rc' }
+                    $arm64Files = $allDebFiles | Where-Object { $_ -match '_arm64\.deb$' -and $_ -notmatch '~rc' }
+                    $armhfFiles = $allDebFiles | Where-Object { $_ -match '_armhf\.deb$' -and $_ -notmatch '~rc' }
+                }
+
+                # Select package by version (or latest if no version specified)
```
guhetier: It would be interesting to have a warning/failure if the latest version is not the same for all packages.
This would indicate that either the publishing partially failed or is still in progress - in both cases, we need to re-run the validation later.

scripts/validate-msquic-docker.ps1
```diff
@@ -0,0 +1,1868 @@
...
+            }
+        }
+    }
+
+    # Version not found
+    return $null
+}
+
+# Function to download packages from packages.microsoft.com
+function Download-LatestPackages {
```
guhetier: nit: `Download-Packages` since it takes a version?

scripts/validate-msquic-docker.ps1
liveans: Maybe we can but it's not necessary, because it will be validated on .NET again, and if we can't load it, we will fail anyway.

.github/workflows/validate-linux-packages.yml
liveans: I was thinking to separate it initially, I ended up with one pipeline.

scripts/validate-msquic-docker.ps1
liveans: Not really, I wanted to be able to provide custom packages if needed, so that's why I implemented this way

guhetier: Got it. I actually like it (it might let us also use this script to test unpublished packages from the pipeline).

The only thing I can think of is that it means we don't validate that the package manager is able to fetch the package properly, but we can probably trust that the repository infrastructure deal with that.

.github/workflows/validate-linux-packages.yml
guhetier: ok - it might be worth merging the two just to avoid future dev wondering about it.

liveans: I think i'll do it with another PR next month.

scripts/validate-msquic-docker.ps1
liveans: Done

scripts/validate-msquic-docker.ps1
liveans: Done

guhetier (approved):

gaurav2699:
I think we skipped RHEL 10 for the validation.

PR #5684, opened 2025-12-18
Docs: CA cert files are only supported in OpenSSL.

Myriachan:
## Description

Added a note saying that `QUIC_CREDENTIAL_FLAG_SET_CA_CERTIFICATE_FILE` is only supported in OpenSSL.  It's supported in QuicTLS too, but that isn't mentioned for the other options.

## Testing

## Documentation

This is a documentation change.

### DIFF
--- a/docs/api/QUIC_CREDENTIAL_CONFIG.md
+++ b/docs/api/QUIC_CREDENTIAL_CONFIG.md
@@ -160,6 +160,7 @@ Obtain the peer certificate using a faster in-process API call. Only available o
 `QUIC_CREDENTIAL_FLAG_SET_CA_CERTIFICATE_FILE`
 
 Enable CA certificate file provided in the `CaCertificateFile` member.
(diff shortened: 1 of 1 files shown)


Myriachan:
I don't know what I should do regarding the contribution license.  I work for an Xbox studio, so I don't know how things would be different because of that.  I can discuss that more privately.

I'm working on adding support for `QUIC_CREDENTIAL_FLAG_SET_CA_CERTIFICATE_FILE` under SChannel, as well as adding a new `QUIC_CREDENTIAL_FLAG_SET_CA_CERTIFICATE_BLOB` that takes the same data in memory as opposed to a disk file.

guhetier:
Thanks for the PR.
For the license, the CLA bot seems happy so I expect you already agreed to the terms for contributing to a Microsoft repo, but please reach out internally if you have any concerns.

guhetier (approved):

PR #5685, opened 2025-12-18
[DeepTest] CUBIC Congestion Control Test Suite Enhancement

saikat107:
## Summary

Rewrote `src/core/unittest/CubicTest.cpp` from 17 standalone tests (628 lines) to 50 fixture-based tests + 16 parametric `CubeRootTest` cases (1754 lines).

## What changed

- **Test fixture** with shared helpers (`InitializeDefaultWithRtt`, `EnterCongestionAvoidance`, `EnterHyStartActive`, `MakeAckEvent`/`MakeLossEvent`) and named RTT constants for HyStart thresholds.
- **16 parametric `CubeRootTest` cases** covering boundaries, perfect/non-perfect cubes, and large inputs.
- **HyStart++ state transition table** in-file, mapping every transition to its test.

## Test areas

| Area | Tests | What's covered |
|------|-------|----------------|
| Core CUBIC lifecycle | 14 | Initialization, reset, data sent/invalidated/acked, loss, ECN, recovery entry/exit, fast convergence, spurious congestion, zero-byte ACK, statistics, app-limited API |
| Congestion avoidance & AIMD | 7 | CUBIC formula vs AIMD selection, AIMD accumulator (half-rate, full-rate, overflow), sequential linear convergence, Reno-friendly region, CUBIC overflow to BytesInFlightMax |
| Pacing | 3 | Slow start estimation (2×CW), congestion avoidance estimation (1.25×CW), LastSendAllowance carryover |
| HyStart++ state machine | 16 | All 6 state transitions (NOT_STARTED→ACTIVE, NOT_STARTED→DONE, ACTIVE→DONE×2, ACTIVE→NOT_STARTED, DONE absorbing), delay increase/decrease detection, Eta calculation, conservative round counting, disabled mode, state persistence across recovery/spurious events, growth divisor invariant |
| Edge cases & boundaries | 8 | Minimum window floor, 2×BytesInFlightMax clamp, bounded growth clamp (1.5×CW), time gap overflow, persistent congestion window reset, double-loss guard, recovery stay-on-old-ACK, ECN spurious rollback asymmetry |
| Send allowance & control | 2 | CanSend scenarios (7 states), GetSendAllowance scenarios (5 states) |

## Coupling classification

| Category | Tests |
|----------|-------|
| Spec-conformance (RFC 8312, RFC 9002, HyStart++) | 23 |
| Implementation-specific, API-level assertions | 7 |
| Implementation-specific, internal struct assertions | 20 |

## Testing

```
[==========] 66 tests from 2 test suites ran.
[  PASSED  ] 66 tests.
```

### DIFF
--- a/src/core/unittest/CubicTest.cpp
+++ b/src/core/unittest/CubicTest.cpp
@@ -246,6 +246,910 @@ protected:
     }
 };
 
+//====================================================================
+//
+//  Specification-conformance tests
+//
+//  These tests validate behavior defined by RFC 8312 (CUBIC),
+//  RFC 9002 (QUIC Loss Detection), and RFC 9406 (HyStart++).
+//  They would remain valid across any conforming implementation.
+//
+//  References:
+//    https://www.rfc-editor.org/rfc/rfc8312
+//    https://www.rfc-editor.org/rfc/rfc9002
+//    https://www.rfc-editor.org/rfc/rfc9406
+//
+//====================================================================
+
+//
+// Test: OnDataAcknowledged - Basic ACK Processing in Slow Start
+// Scenario: Tests CubicCongestionControlOnDataAcknowledged for basic slow-start
+// window growth. Sends data via OnDataSent (properly tracking BytesInFlightMax),
+// then acknowledges a portion. The window should grow by BytesAcked since we are
+// in slow start (CongestionWindow < SlowStartThreshold) with HyStart disabled
+// (CWndSlowStartGrowthDivisor = 1).
+//
+TEST_F(CubicTest, OnDataAcknowledged_BasicAck)
+{
+    InitializeDefaultWithRtt(/*WindowPackets = */ 10, /*HyStart = */ false);
+
+    uint32_t InitialWindow = Cubic->CongestionWindow;
+
+    // Send data via OnDataSent to properly track BytesInFlightMax.
+    // Must send enough that BytesInFlightMax > (InitialWindow + BytesAcked) / 2
+    // to avoid the 2*BytesInFlightMax clamping guard.
+    uint32_t BytesSent = 10000;
+    CC->QuicCongestionControlOnDataSent(CC, BytesSent);
+
+    uint32_t BytesAcked = 5000;
+    QUIC_ACK_EVENT AckEvent = MakeAckEvent(1050000, 5, 10, BytesAcked);
+
+    CC->QuicCongestionControlOnDataAcknowledged(CC, &AckEvent);
+
+    // Slow start growth: window increases by BytesAcked (divisor=1, HyStart disabled)
+    // NewWindow = 12320 + 5000 = 17320, clamped by 2*BytesInFlightMax = 2*10000 = 20000 (no clamp)
+    ASSERT_EQ(Cubic->CongestionWindow, InitialWindow + BytesAcked);
+    ASSERT_EQ(Cubic->BytesInFlight, BytesSent - BytesAcked);
+}
+
+//
+// Test: OnDataLost - Packet Loss Handling and Window Reduction
+// Scenario: Tests CUBIC's response to packet loss. When packets are declared lost,
+// the congestion window should be reduced by beta (0.7) and the connection enters
+// recovery. Verifies window reduction, threshold update, and recovery state flags.
+//
+TEST_F(CubicTest, OnDataLost_WindowReduction)
+{
+    InitializeDefaultWithRtt(/*WindowPackets = */ 20, /*HyStart = */ true);
+    uint32_t InitialWindow = Cubic->CongestionWindow;
+
+    // Send data via OnDataSent to properly track BytesInFlightMax
+    CC->QuicCongestionControlOnDataSent(CC, 10000);
+
+    QUIC_LOSS_EVENT LossEvent = MakeLossEvent(3600, 10, 15);
+
+    CC->QuicCongestionControlOnDataLost(CC, &LossEvent);
+
+    // Verify window was reduced (CUBIC multiplicative decrease)
+    // New window = InitialWindow * 0.7 = InitialWindow * 7 / 10
+    // InitialWindow = (1280 - 48) * 20 = 24640 (IPv6 formula: MTU - 40 - 8)
+    // Expected = 24640 * 7 / 10 = 17248
+    uint32_t ExpectedWindow = InitialWindow * 7 / 10;
+    ASSERT_EQ(Cubic->CongestionWindow, ExpectedWindow);
+    ASSERT_EQ(Cubic->SlowStartThreshold, ExpectedWindow);
+
+    // Verify recovery state transitions
+    ASSERT_TRUE(Cubic->IsInRecovery);
+    ASSERT_TRUE(Cubic->HasHadCongestionEvent);
+}
+
+//
+// Test: OnEcn - ECN Marking Handling
+// Scenario: Tests Explicit Congestion Notification (ECN) handling. When ECN-marked packets
+// are received, CUBIC should treat it as a congestion signal and reduce the window.
+// Unlike loss, ECN does NOT save previous state (`if (!Ecn)` guard),
+// so spurious congestion rollback cannot undo an ECN event.
+//
+TEST_F(CubicTest, OnEcn_CongestionSignal)
+{
+    InitializeDefaultWithRtt(/*WindowPackets = */ 20, /*HyStart = */ true);
+    uint32_t InitialWindow = Cubic->CongestionWindow;
+
+    // Send data via OnDataSent to properly track BytesInFlightMax
+    CC->QuicCongestionControlOnDataSent(CC, 10000);
+
+    QUIC_ECN_EVENT EcnEvent{};
+    EcnEvent.LargestPacketNumberAcked = 10;
+    EcnEvent.LargestSentPacketNumber = 15;
+
+    CC->QuicCongestionControlOnEcn(CC, &EcnEvent);
+
+    // Verify window was reduced due to ECN congestion signal (same as loss: 0.7x)
+    // InitialWindow = (1280 - 48) * 20 = 24640 (IPv6 formula: MTU - 40 - 8)
+    // Expected = 24640 * 7 / 10 = 17248
+    uint32_t ExpectedWindow = InitialWindow * 7 / 10;
+    ASSERT_EQ(Cubic->CongestionWindow, ExpectedWindow);
+
+    // Verify recovery state transitions
+    ASSERT_TRUE(Cubic->IsInRecovery);
+    ASSERT_TRUE(Cubic->HasHadCongestionEvent);
+}
+
+//
+// Test: Fast Convergence - Window Reduction Path
+// Scenario: Tests CUBIC's fast convergence algorithm. When a new congestion event occurs
+// before reaching the previous WindowMax, CUBIC applies an additional reduction factor
+// to converge faster with other flows. This tests the WindowLastMax > WindowMax path.
+//
+TEST_F(CubicTest, FastConvergence_AdditionalReduction)
+{
+    InitializeDefaultWithRtt(/*WindowPackets = */ 30, /*HyStart = */ true);
+
+    // Simulate first congestion event to establish WindowMax
+    // Send data to fill the window
+    uint32_t InitialWindow = Cubic->CongestionWindow;
+    CC->QuicCongestionControlOnDataSent(CC, InitialWindow);
+
+    // Trigger first loss event
+    QUIC_LOSS_EVENT FirstLoss = MakeLossEvent(3000, 5, 10);
+
+    CC->QuicCongestionControlOnDataLost(CC, &FirstLoss);
+
+    // ACK while still in recovery (LargestAck == RecoverySentPacketNumber, so
+    // recovery doesn't exit). Window stays at post-loss value.
+    QUIC_ACK_EVENT AckEvent = MakeAckEvent(1000000, 10, 15, 5000);
+
+    CC->QuicCongestionControlOnDataAcknowledged(CC, &AckEvent);
+
+    // Send more data
+    CC->QuicCongestionControlOnDataSent(CC, 3000);
+
+    // Trigger second loss event (before reaching previous WindowMax)
+    // Must use LargestPacketNumberLost > RecoverySentPacketNumber (which is 10 from first loss)
+    QUIC_LOSS_EVENT LossEvent = MakeLossEvent(3000, 15, 20);
+
+    CC->QuicCongestionControlOnDataLost(CC, &LossEvent);
+
+    // Fast convergence fires on the second loss because WindowLastMax (36960, set
+    // during first loss) > WindowMax (25872, the current CW at second loss entry).
+    // The code sets WindowLastMax = WindowMax = 25872, then applies:
+    //   WindowMax = WindowMax * (10 + BETA) / 20 = 25872 * 17 / 20 = 21991
+    uint32_t WindowAfterFirstLoss = InitialWindow * 7 / 10;  // 25872
+    uint32_t ExpectedWindowMax = WindowAfterFirstLoss * 17 / 20;  // 21991
+    ASSERT_EQ(Cubic->WindowMax, ExpectedWindowMax);
+}
+
+//
+// Test: Recovery Exit Path
+// Scenario: Tests exiting from recovery state when an ACK is received for a packet
+// sent after recovery started (LargestAck > RecoverySentPacketNumber). Uses a
+// non-persistent loss to enter recovery cleanly.
+//
+TEST_F(CubicTest, Recovery_ExitOnNewAck)
+{
+    InitializeDefaultWithRtt(/*WindowPackets = */ 20,  /*HyStart = */ true);
+
+    // Enter recovery via non-persistent loss
+    CC->QuicCongestionControlOnDataSent(CC, 5000);
+    Connection.Send.NextPacketNumber = 10;
+
+    QUIC_LOSS_EVENT LossEvent = MakeLossEvent(1200, 8, 10);
+
+    CC->QuicCongestionControlOnDataLost(CC, &LossEvent);
+
+    // Now in recovery state
+    ASSERT_TRUE(Cubic->IsInRecovery);
+    ASSERT_FALSE(Cubic->IsInPersistentCongestion);
+
+    // Send new packet after recovery started
+    Connection.Send.NextPacketNumber = 15;
+
+    QUIC_ACK_EVENT AckEvent = MakeAckEvent(1000000, 15, 20, 1200);
+
+    CC->QuicCongestionControlOnDataAcknowledged(CC, &AckEvent);
+
+    // Should exit recovery
+    ASSERT_FALSE(Cubic->IsInRecovery);
+    ASSERT_FALSE(Cubic->IsInPersistentCongestion);
+}
+
+//
+// Test: ACK During Recovery Stays in Recovery
+// Scenario: When an ACK is received with LargestAck <= RecoverySentPacketNumber,
+// the connection remains in recovery. Only an ACK for a packet sent after
+// recovery started (LargestAck > RecoverySentPacketNumber) exits recovery.
+//
+TEST_F(CubicTest, Recovery_StayOnOldAck)
+{
+    InitializeDefaultWithRtt(/*WindowPackets = */ 20, /*HyStart = */ false);
+
+    CC->QuicCongestionControlOnDataSent(CC, 5000);
+    Connection.Send.NextPacketNumber = 10;
+
+    // Enter recovery via loss. RecoverySentPacketNumber = 10.
+    QUIC_LOSS_EVENT LossEvent = MakeLossEvent(1200, 5, 10);
+    CC->QuicCongestionControlOnDataLost(CC, &LossEvent);
+
+    ASSERT_TRUE(Cubic->IsInRecovery);
+    uint32_t WindowInRecovery = Cubic->CongestionWindow;
+
+    // ACK with LargestAck == RecoverySentPacketNumber (10): should stay in recovery
+    QUIC_ACK_EVENT StayAck = MakeAckEvent(1050000, 10, 15, 1200);
+    CC->QuicCongestionControlOnDataAcknowledged(CC, &StayAck);
+
+    ASSERT_TRUE(Cubic->IsInRecovery);
+    // Window should not grow during recovery (goes to Exit label)
+    ASSERT_EQ(Cubic->CongestionWindow, WindowInRecovery);
+
+    // ACK with LargestAck < RecoverySentPacketNumber (8): also stays
+    QUIC_ACK_EVENT OlderAck = MakeAckEvent(1060000, 8, 15, 600);
+    CC->QuicCongestionControlOnDataAcknowledged(CC, &OlderAck);
+
+    ASSERT_TRUE(Cubic->IsInRecovery);
+    ASSERT_EQ(Cubic->CongestionWindow, WindowInRecovery);
+
+    // ACK with LargestAck > RecoverySentPacketNumber: exits recovery
+    Connection.Send.NextPacketNumber = 20;
+    QUIC_ACK_EVENT ExitAck = MakeAckEvent(1100000, 11, 25, 600);
+    CC->QuicCongestionControlOnDataAcknowledged(CC, &ExitAck);
+
+    ASSERT_FALSE(Cubic->IsInRecovery);
+}
+
+//
+// Test: Double Loss During Recovery - No Additional Window Reduction
+// Scenario: When a second loss occurs during recovery with LargestPacketNumberLost
+// <= RecoverySentPacketNumber, the guard at cubic.c prevents a second window
+// reduction. Only BytesInFlight is decremented; the congestion window stays at
+// its post-first-loss value.
+//
+TEST_F(CubicTest, Recovery_DoubleLossNoDoubleReduction)
+{
+    InitializeDefaultWithRtt(/*WindowPackets = */ 20, /*HyStart = */ false);
+
+    CC->QuicCongestionControlOnDataSent(CC, 10000);
+    Connection.Send.NextPacketNumber = 10;
+
+    // First loss: enters recovery, reduces window
+    QUIC_LOSS_EVENT FirstLoss = MakeLossEvent(1200, 5, 10);
+    CC->QuicCongestionControlOnDataLost(CC, &FirstLoss);
+
+    ASSERT_TRUE(Cubic->IsInRecovery);
+    uint32_t WindowAfterFirstLoss = Cubic->CongestionWindow;
+    uint32_t BytesInFlightAfterFirstLoss = Cubic->BytesInFlight;
+
+    // Second loss during recovery: LargestPacketNumberLost (8) <= RecoverySentPacketNumber (10)
+    // Guard prevents window reduction; only BytesInFlight decreases.
+    uint32_t SecondLostBytes = 600;
+    QUIC_LOSS_EVENT SecondLoss = MakeLossEvent(SecondLostBytes, 8, 10);
+    CC->QuicCongestionControlOnDataLost(CC, &SecondLoss);
+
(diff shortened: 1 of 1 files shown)


src/core/congestion_control.h
```diff
@@ -411,4 +411,4 @@ QuicCongestionControlSetAppLimited(
     )
 {
     Cc->QuicCongestionControlSetAppLimited(Cc);
-}
+}
```
guhetier: nit: we try to keep the end of file carriage return

src/core/unittest/CubicTest.cpp
```diff
@@ -394,10 +419,19 @@ TEST(CubicTest, ResetScenarios)
...
     // Scenario 1: Partial reset (FullReset=FALSE) - preserves BytesInFlight
-    Cubic->BytesInFlight = 5000;
-    Cubic->SlowStartThreshold = 10000;
-    Cubic->IsInRecovery = TRUE;
-    Cubic->HasHadCongestionEvent = TRUE;
+    // First, send some data and trigger a congestion event to set internal flags
+    Connection.CongestionControl.QuicCongestionControlOnDataSent(&Connection.CongestionControl, 5000);
+
+    // Trigger congestion event via loss
+    QUIC_LOSS_EVENT LossEvent;
```
guhetier: Default initialization in C++ is better done with `QUIC_LOSS_EVENT LossEvent{};`.
In C, `QUIC_LOSS_EVENT LossEvent = {0};` is also generally enough.

src/core/unittest/CubicTest.cpp
```diff
@@ -505,9 +550,10 @@ TEST(CubicTest, OnDataAcknowledged_BasicAck)
     Settings.SendIdleTimeoutMs = 1000;
 
     InitializeMockConnection(Connection, 1280);
+    Connection.Settings.HyStartEnabled = TRUE;  // Must set on Connection for runtime checks
```
guhetier: I am not sure I understand why we need HyStart.

src/core/unittest/CubicTest.cpp
```diff
@@ -505,9 +550,10 @@ TEST(CubicTest, OnDataAcknowledged_BasicAck)
...
     Connection.Paths[0].GotFirstRttSample = TRUE;
     Connection.Paths[0].SmoothedRtt = 50000; // 50ms in microseconds
-
+    Connection.Settings.NetStatsEventEnabled = TRUE; // Enable logging
```
guhetier: Enabling this is probably good for coverage, but it seems there should either be a callback set to validate the data in the event, or a comment explaining this intentionally tests the case where no callback is present.

src/core/unittest/CubicTest.cpp
```diff
@@ -759,13 +817,2265 @@ TEST(CubicTest, HyStart_StateTransitions)
...
+        &Connection.CongestionControl,
+        &PersistentLoss);
+
+    // After persistent congestion:
+    // - Window is reset to 2 * MTU (approximately 2560, but may vary)
+    // - Threshold is set to old_window * 0.7
+    uint32_t WindowAfterPC = Cubic->CongestionWindow;
+    uint32_t ThresholdAfterPC = Cubic->SlowStartThreshold;
+
+    ASSERT_LT(WindowAfterPC, 3000u); // Should be around 2*MTU
```
guhetier: This should be an equality. The test can control the MTU exactly.
A comparison brings little validation, if we were to make a mistake, it would be a coin-flip whether it does in the right direction or not to cause a test failure.

src/core/unittest/CubicTest.cpp
```diff
@@ -759,13 +817,2265 @@ TEST(CubicTest, HyStart_StateTransitions)
...
+    AckEvent.IsLargestAckedPacketAppLimited = FALSE;
+    AckEvent.AdjustedAckTime = AckEvent.TimeNow;
+    AckEvent.AckedPackets = NULL;
+
+    Connection.CongestionControl.QuicCongestionControlOnDataAcknowledged(
+        &Connection.CongestionControl,
+        &AckEvent);
+
+    // Window should be valid (overflow was handled)
+    ASSERT_GT(Cubic->CongestionWindow, 0u);
```
guhetier: What is this validating? `CongestionWindow` is a uint32_t, so all this ensures is that the value is neither `0` nor `UINT32_MAX`.

The test seems to assume `CongestionWindow` is a signed value, like an int64_t maybe.

src/core/unittest/CubicTest.cpp
```diff
@@ -759,13 +817,2265 @@ TEST(CubicTest, HyStart_StateTransitions)
...
+        if (round < QUIC_CONSERVATIVE_SLOW_START_DEFAULT_ROUNDS - 1) {
+            // Still in HYSTART_ACTIVE for first n-1 rounds
+            ASSERT_EQ(Cubic->HyStartState, HYSTART_ACTIVE);
+        } else {
+            // for the rest of the rounds, state should transition
+            // to HYSTART_DONE and stay there
+            ASSERT_EQ(Cubic->HyStartState, HYSTART_DONE);
+            ASSERT_EQ(Cubic->ConservativeSlowStartRounds, 0u);
+            // Verify SlowStartThreshold is set to current congestion window
+            ASSERT_LT(Cubic->SlowStartThreshold, UINT32_MAX);
```
guhetier: `SlowStartThreshold` is a uint32, this is not validating much.

guhetier:
A general comment is that while the tests do provide coverage (which is good), the validation itself seems pretty weak in term of catching logic issues or regressions.
And in terms of finding undefined behavior, memory issues, etc... this probably doesn't beat fuzzing.

src/core/unittest/CubicTest.cpp
saikat107: Resolved

src/core/unittest/CubicTest.cpp
saikat107: Resolved with `ASSERT_EQ`

src/core/unittest/CubicTest.cpp
saikat107: Not sure how to set the callbacks. Agent was not able to resolve that either. Added the comment -- please suggest if (and how) we should add callbacks.

src/core/unittest/CubicTest.cpp
saikat107: You are right -- this is not necessary for this test case.

saikat107:
Thanks for the review @guhetier -- these do not only make the current test cases better, but also the DeepTest agent better.

saikat107:
@guhetier this PR is ready for review.

src/core/unittest/CubicTest.cpp
```diff
@@ -864,7 +895,7 @@ TEST(CubicTest, Recovery_ExitOnNewAck)
     Connection.CongestionControl.QuicCongestionControlOnDataSent(&Connection.CongestionControl, 5000);
     Connection.Send.NextPacketNumber = 10; // Set packet number before loss
 
-    QUIC_LOSS_EVENT LossEvent;
+    QUIC_LOSS_EVENT LossEvent{};
     CxPlatZeroMemory(&LossEvent, sizeof(LossEvent));
```
guhetier: nit: The previous comment about zero-initializing using `{}` was addressed but the `CxPlatZeroMemory` right below is still there. Same for multiple other occurrences in the file.

guhetier: On top of this, there are generally multiple fields that are always set to a default value below.
According to Copilot review, 500 lines (~18%) of the file is redundant initialization (I didn't check the numbers :)), it seems like an area to improve.

src/core/unittest/CubicTest.cpp
```diff
@@ -371,7 +405,9 @@ TEST(CubicTest, GetterFunctions)
     // Test GetCongestionWindow
     uint32_t CongestionWindow = Connection.CongestionControl.QuicCongestionControlGetCongestionWindow(&Connection.CongestionControl);
     ASSERT_EQ(CongestionWindow, Cubic->CongestionWindow);
```
guhetier: nit: Re-reading through the code, this assertion is now IMO needlessly intrusive.

What matters is that `QuicCongestionControlGetCongestionWindow` returns the expected value.
How it does it internally doesn't matter, it could be a trivial getter or make some computation.

This is a general pattern, I don't know if that's something DeepTest can get smarter about.
The tests should ideally not depend on the internal implementation details more than needed: this allows us to change the code without changing the test and get confirmation that the behavior didn't regress. If we need to modify both at once, we lose some validation.

src/core/unittest/CubicTest.cpp
```diff
@@ -537,8 +598,10 @@ TEST(CubicTest, OnDataAcknowledged_BasicAck)
     Connection.CongestionControl.QuicCongestionControlOnDataAcknowledged(
         &Connection.CongestionControl,
         &AckEvent);
-    // Verify window may have grown (depends on slow start vs congestion avoidance)
-    ASSERT_GE(Cubic->CongestionWindow, InitialWindow);
+
+    // Window unchanged: LargestAck (5) >= HyStartRoundEnd (0) triggers round boundary
+    // crossing which resets HyStartAckCount to 0. With insufficient samples, no growth.
+    ASSERT_EQ(Cubic->CongestionWindow, InitialWindow);
```
guhetier: **[AI-generated]** This test passes for the wrong reason. The comment on line 602 claims *"HyStart round boundary crossing resets HyStartAckCount, preventing window growth"*, but HyStart is **disabled** in this test (`InitializeMockConnection` sets `HyStartEnabled = FALSE` and this test never overrides it). The HyStart code path at `cubic.c:476` is skipped entirely.

What actually happens: the window **does** grow from 12320 to 17320 during slow start (`+= BytesAcked`), but is then **clamped back** to `2 * BytesInFlightMax = 2 * 6160 = 12320` at `cubic.c:683`. The clamping fires because `BytesInFlight` is set directly (line 578: `Cubic->BytesInFlight = 5000`) instead of via `OnDataSent`, so `BytesInFlightMax` is never updated from its init value (`CongestionWindow / 2 = 6160`).

This test accidentally validates the `BytesInFlightMax` clamping guard, not HyStart behavior. If someone modifies the clamping logic, this test will fail for a reason unrelated to its stated purpose.

**Suggestion:** Either use `OnDataSent` to set BytesInFlight (which updates BytesInFlightMax), fix the comment to describe the real behavior, or restructure the test to actually test what it claims.

src/core/unittest/CubicTest.cpp
```diff
@@ -759,13 +840,2412 @@ TEST(CubicTest, HyStart_StateTransitions)
...
+    AckEvent.AdjustedAckTime = AckEvent.TimeNow;
+    AckEvent.AckedPackets = NULL;
+
+    Connection.CongestionControl.QuicCongestionControlOnDataAcknowledged(
+        &Connection.CongestionControl,
+        &AckEvent);
+
+    // Accumulator is 0: LargestAck (15) > RecoverySentPacketNumber (10) exits recovery,
+    // but code jumps to Exit label after clearing IsInRecovery, so AIMD logic not reached.
+    ASSERT_EQ(Cubic->AimdAccumulator, 0u);
```
guhetier: **[AI-generated]** Tests 24 and 25 (`AIMD_AccumulatorBelowWindowPrior` and `AIMD_AccumulatorAboveWindowPrior`) both claim to test AIMD accumulator logic (0.5 vs 1.0 MSS/RTT slope), but neither test actually exercises that code.

In both tests, `LargestAck=15 > RecoverySentPacketNumber=10`, so recovery exits at `cubic.c:454`, which then `goto Exit` at line 468 **before** reaching the AIMD accumulator logic at `cubic.c:649`. Both tests only verify `AimdAccumulator == 0` after recovery exit — they are functionally identical (differing only in BytesAcked: 600 vs 1200, which doesn't matter since the AIMD code is never reached).

The AIMD accumulator paths (`AimdAccumulator += BytesAcked / 2` when `AimdWindow < WindowPrior`, vs `AimdAccumulator += BytesAcked` otherwise) have **no test coverage**.

**Suggestion:** To actually test AIMD, exit recovery first (via a prior ACK with `LargestAck > RecoverySentPacketNumber`), then send a subsequent ACK with non-zero `BytesAcked` that will reach the congestion avoidance code at `cubic.c:563+`.

src/core/unittest/CubicTest.cpp
```diff
@@ -759,13 +840,2412 @@ TEST(CubicTest, HyStart_StateTransitions)
...
 
-    // HyStart may transition states based on RTT measurements
-    // Just verify state is valid and divisor is set appropriately
-    ASSERT_TRUE(Cubic->HyStartState >= HYSTART_NOT_STARTED &&
-                Cubic->HyStartState <= HYSTART_DONE);
-    ASSERT_GE(Cubic->CWndSlowStartGrowthDivisor, 1u);
+    // LargestAck (10) is NOT > RecoverySentPacketNumber (10), so we stay in recovery
+    // Window doesn't grow during recovery - just BytesInFlight decreases
+    ASSERT_TRUE(Cubic->IsInRecovery);
+    ASSERT_EQ(Cubic->CongestionWindow, WindowBefore);
```
guhetier: **[AI-generated]** This test claims to test *"AIMD vs CUBIC window selection logic"* but it stays in recovery the entire time. `LargestAck=10` is NOT `> RecoverySentPacketNumber=10` (the comparison is strict `>`), so recovery is never exited and the code jumps to `Exit` at `cubic.c:468`.

The AIMD vs CUBIC selection logic at `cubic.c:659-670` (`if (Cubic->AimdWindow > CubicWindow)`) is never reached. The only thing this test validates is that the window doesn't change during recovery — which is already covered by tests 26 and 30.

**Suggestion:** Exit recovery first, then ACK data to exercise the actual AIMD vs CUBIC window selection in congestion avoidance. Or rename this test to accurately describe what it tests (e.g., `Recovery_WindowUnchangedDuringRecovery`).

src/core/unittest/CubicTest.cpp
```diff
@@ -759,13 +840,2412 @@ TEST(CubicTest, HyStart_StateTransitions)
...
+    AckEvent.AckedPackets = NULL;
+
+    Connection.CongestionControl.QuicCongestionControlOnDataAcknowledged(
+        &Connection.CongestionControl,
+        &AckEvent);
+
+    // Still in recovery because LargestAck (10) is NOT > RecoverySentPacketNumber (10)
+    // Window stays the same during recovery
+    ASSERT_TRUE(Cubic->IsInRecovery);
+    ASSERT_EQ(Cubic->CongestionWindow, WindowAfterLoss);
```
guhetier: **[AI-generated]** This test claims to test *"overflow handling in CUBIC window calculation"* but it stays in recovery (`LargestAck=10` NOT `> RecoverySentPacketNumber=10`). The CUBIC window formula at `cubic.c:620` and the overflow check at `cubic.c:625` (`if (CubicWindow < 0)`) are never reached.

The test only verifies that the window doesn't change during recovery — identical to what tests 23 and 30 already verify.

**Suggestion:** Exit recovery before the ACK that should trigger the CUBIC window overflow, so the congestion avoidance code path is actually exercised.

src/core/unittest/CubicTest.cpp
```diff
@@ -759,13 +840,2412 @@ TEST(CubicTest, HyStart_StateTransitions)
...
+    AckEvent2.AckedPackets = NULL;
+
+    Connection.CongestionControl.QuicCongestionControlOnDataAcknowledged(
+        &Connection.CongestionControl,
+        &AckEvent2);
+
+    // Still in recovery because LargestAck (10) is NOT > RecoverySentPacketNumber (10)
+    // Window stays the same during recovery
+    ASSERT_TRUE(Cubic->IsInRecovery);
+    ASSERT_EQ(Cubic->CongestionWindow, WindowAfterFirst);
```
guhetier: **[AI-generated]** This test claims to test *"behavior with large time gaps between ACKs"* (idle period handling), but both ACKs stay in recovery (`LargestAck` is never `> RecoverySentPacketNumber=10`). The time gap adjustment logic at `cubic.c:580-588` is only reached **outside** recovery (the `IsInRecovery` check at `cubic.c:453` causes a `goto Exit` before the time gap code).

The 5-second gap between ACK1 and ACK2 has no effect — the test just verifies the window doesn't change during recovery, same as tests 23 and 26.

**Suggestion:** Exit recovery, establish `TimeOfLastAckValid=TRUE` with an initial ACK, then send a second ACK after a long idle gap to actually exercise the `TimeSinceLastAck > SendIdleTimeoutMs` path at `cubic.c:582`.

guhetier (commented):
AI-generated review (GitHub Copilot CLI). Focused on test correctness: 6 high-severity issues where tests don't exercise the code paths they claim to test.

src/core/unittest/CubicTest.cpp
```diff
@@ -759,13 +840,2412 @@ TEST(CubicTest, HyStart_StateTransitions)
...
+    LossEvent.NumRetransmittableBytes = 2400;
+    LossEvent.PersistentCongestion = FALSE;
+    LossEvent.LargestPacketNumberLost = 5;
+    LossEvent.LargestSentPacketNumber = 10;
+
+    Connection.CongestionControl.QuicCongestionControlOnDataLost(&Connection.CongestionControl, &LossEvent);
+
+    // Invariant 2: DONE → divisor = 1
+    ASSERT_EQ(Cubic->HyStartState, HYSTART_DONE);
+    ASSERT_EQ(Cubic->CWndSlowStartGrowthDivisor, 1u);
```
guhetier: **[AI-generated]** This test claims to verify the growth divisor invariant across all three HyStart states (*"NOT_STARTED → divisor=1, ACTIVE → divisor=4, DONE → divisor=1"*), but it only transitions NOT_STARTED → DONE via loss. The ACTIVE state (divisor=4) is **never entered or verified** in this test.

Test 42 (`HyStart_DelayIncreaseDetection_TriggerActiveTransition`) already verifies the ACTIVE state divisor at line 2669.

**Suggestion:** Either add the ACTIVE state transition to this test (transition NOT_STARTED → ACTIVE → DONE and check divisor at each step), or update the test description to reflect that it only verifies NOT_STARTED and DONE states.

src/core/unittest/CubicTest.cpp
saikat107: Done in the new version.

src/core/unittest/CubicTest.cpp
saikat107: In the current version, expected values were calculated based on formulae, and used APIs to manistates. Hope this is better.

src/core/unittest/CubicTest.cpp
```diff
@@ -23,749 +25,2394 @@ static void InitializeMockConnection(
...
-    // Set to max
-    Connection.CongestionControl.QuicCongestionControlSetExemption(&Connection.CongestionControl, 255);
-    ASSERT_EQ(Cubic->Exemptions, 255u);
-}
+// //
+// // Test: SetExemption
+// // Scenario: Tests SetExemption to verify it correctly sets the number of packets that
+// // can bypass congestion control. Used for probe packets and other special cases.
+// //
+// TEST_F(CubicTest, SetExemption)
```
guhetier: Commented test?

src/core/unittest/CubicTest.cpp
```diff
@@ -23,749 +25,2394 @@ static void InitializeMockConnection(
...
+    CC->QuicCongestionControlOnDataAcknowledged(CC, &ExitAck);
+    ASSERT_FALSE(Cubic->IsInRecovery);
+
+    // Force extreme conditions to trigger CUBIC overflow.
+    // Set a huge WindowMax so KCubic is very large, making DeltaT deeply negative
+    // when TimeInCongAvoid is small. The cubic term DeltaT^3 overflows negative.
+    Cubic->WindowMax = UINT32_MAX;
+    // Recalculate KCubic for the extreme WindowMax (mirrors the loss handler logic)
+    Cubic->KCubic =
+        CubeRoot(
```
guhetier: It seems weird to me to pull a private function from the cubic compilation unit just to set a test value.
Pulling a function in to test it would have been a thing, but here, it is used as a helper to set test data, and combined with a "magic formula".

At this point, why not set a value directly in KCubic, that will trigger the scenario we care about and explain why this value? (like it is done for WindowsMax)?

src/core/unittest/CubicTest.cpp
saikat107: Mistake

saikat107:
Check this out, @guhetier

## Overview

| Category | Count | % of 50 |
|----------|-------|---------|
| Specification-conformance | 23 | 46% |
| Implementation-specific, loosely coupled | 7 | 14% |
| Implementation-specific, tightly coupled | 20 | 40% |

**Specification-conformance** tests validate behavior defined by RFC 8312 (CUBIC), RFC 9002 (QUIC Loss Detection), or the HyStart++ draft. These tests would remain valid across any conforming implementation.

**Implementation-specific, loosely coupled** tests exercise MsQuic-specific features (pacing API, exemptions, statistics) but assert only through public API outputs. They would survive an internal refactor.

**Implementation-specific, tightly coupled** tests directly read or write internal `Cubic->` struct fields. If the implementation is refactored — even while preserving identical congestion control behavior — these tests would break.

---

## Specification-Conformance Tests (23)

These test behavior mandated or described by protocol specifications.

| Test | Spec Basis |
|------|-----------|
| `OnDataLost_WindowReduction` | RFC 8312 §4.6: β_cubic = 0.7 multiplicative decrease |
| `OnEcn_CongestionSignal` | RFC 9002 §7.1: ECN treated as congestion signal |
| `FastConvergence_AdditionalReduction` | RFC 8312 §4.7: fast convergence algorithm |
| `Recovery_ExitOnNewAck` | RFC 9002 §7.3.2: exit recovery on ACK of new data |
| `OnDataAcknowledged_BasicAck` | RFC 8312 §4.1: slow start window growth |
| `CongestionAvoidance_AIMDvsCubicSelection` | RFC 8312 §4.2–4.4: W_cubic vs W_est selection |
| `AIMD_SequentialLinearConvergence` | RFC 8312 §4.3: TCP-friendly estimation |
| `CongestionAvoidance_RenoFriendlyRegion` | RFC 8312 §4.3: Reno-friendly region |
| `CongestionAvoidance_BoundedGrowthClamp` | RFC 8312 §4.2: bounded window growth |
| `MinimumWindowFloor_AfterLoss` | RFC 9002 §7.2: minimum congestion window of 2×MTU |
| `SlowStart_BytesInFlightMaxClamp` | RFC 9002 §7.8: pacing/limiting based on bytes in flight |
| `Recovery_DoubleLossNoDoubleReduction` | RFC 9002 §7.3.1: single reduction per recovery period |
| `Recovery_StayOnOldAck` | RFC 9002 §7.3.2: stay in recovery until new ACK |
| `SpuriousCongestion_StateRollback` | RFC 8312 §4.8: response to spurious congestion |
| `HyStart_NotStartedToDone_ViaLoss` | HyStart++ draft: loss triggers transition to DONE |
| `HyStart_NotStartedToDone_ViaECN` | HyStart++ draft: ECN triggers transition to DONE |
| `HyStart_ActiveToDone_ViaLoss` | HyStart++ draft: loss from ACTIVE triggers DONE |
| `HyStart_ActiveToDone_ViaECN` | HyStart++ draft: ECN from ACTIVE triggers DONE |
| `HyStart_AnyToDone_ViaPersistentCongestion` | HyStart++ draft: persistent congestion triggers DONE |
| `HyStart_TerminalState_DoneIsAbsorbing` | HyStart++ draft: DONE is a terminal state |
| `HyStart_DelayIncreaseDetection_TriggerActiveTransition` | HyStart++ draft: delay increase → ACTIVE |
| `HyStart_RttDecreaseDetection_ReturnToNotStarted` | HyStart++ draft: RTT decrease → NOT_STARTED |
| `HyStart_ConservativeSlowStartRounds_TransitionToDone` | HyStart++ draft: conservative rounds → DONE |

---

## Implementation-Specific Tests — Loosely Coupled (7)

These test MsQuic-specific features but assert only through public API outputs (`CanSend`, `GetSendAllowance`, `GetNetworkStatistics`, etc.). They would survive an internal refactor that preserves the same API contract.

| Test | Asserts On |
|------|-----------|
| `CanSendScenarios` | `CanSend()` return value |
| `GetSendAllowanceScenarios` | `GetSendAllowance()` return value |
| `CongestionWindowCalculation` | `CongestionWindow` (primary output) |
| `GetNetworkStatistics_RetrieveStats` | `GetNetworkStatistics()` output struct |
| `CongestionControlOnSpuriousCongestionEvent` | `OnSpuriousCongestionEvent()` return value |
| `ZeroBytesAcked_EarlyExit` | `CongestionWindow` unchanged after zero-byte ACK |
| `AppLimited_APICoverage` | `IsAppLimited()` / `SetAppLimited()` return values |

---

## Implementation-Specific Tests — Tightly Coupled (20)

These tests directly read or write internal `QUIC_CONGESTION_CONTROL_CUBIC` struct fields (e.g., `AimdAccumulator`, `TimeOfCongAvoidStart`, `HyStartState`, `PrevCongestionWindow`). If the implementation is refactored while preserving identical congestion control behavior, these tests would fail.

| Test | Internal Fields Touched |
|------|------------------------|
| `ResetScenarios` | `MinRttInLastRound` (asserts sentinel value) |
| `OnDataSent_IncrementsBytesInFlight` | `BytesInFlight`, `BytesInFlightMax`, `NumRetransmittableExemptions` |
| `OnDataInvalidated_DecrementsBytesInFlight` | `BytesInFlight` |
| `Pacing_SlowStartWindowEstimation` | Writes `SlowStartThreshold`; derives expected from exact `EstWnd = 2×CW` formula |
| `Pacing_CongestionAvoidanceEstimation` | Derives expected from exact `EstWnd = 1.25×CW` formula |
| `Pacing_LastSendAllowanceCarryover` | Tests `LastSendAllowance` accumulation mechanism |
| `AIMD_AccumulatorBelowWindowPrior` | Asserts `AimdAccumulator` exact value (half-rate path) |
| `AIMD_AccumulatorAboveWindowPrior` | Asserts `AimdAccumulator` exact value (full-rate path) |
| `AIMD_AccumulatorTriggersWindowGrowth` | Asserts `AimdAccumulator` overflow triggers +DPL |
| `CubicWindow_OverflowToBytesInFlightMax` | Writes `WindowMax`, `KCubic`, `TimeOfLastAckValid` |
| `TimeGap_IdlePeriodHandling` | Asserts `TimeOfCongAvoidStart` adjustment |
| `CongestionAvoidance_TimeGapOverflowProtection` | Writes and asserts `TimeOfCongAvoidStart` |
| `ECN_SpuriousRollbackRestoresStaleState` | Asserts `PrevCongestionWindow`, `PrevAimdWindow`, `PrevSlowStartThreshold` |
| `SlowStart_WindowOverflowAfterPersistentCongestion` | Asserts `TimeOfCongAvoidStart`, `AimdWindow` |
| `HyStart_InitialStateVerification` | `HyStartState`, `CWndSlowStartGrowthDivisor`, `HyStartAckCount`, etc. |
| `HyStart_Disabled_NoTransitions` | `HyStartState` |
| `HyStart_StateInvariant_GrowthDivisor` | `HyStartState`, `CWndSlowStartGrowthDivisor` |
| `HyStart_RecoveryExit_StatePersistence` | `HyStartState` |
| `HyStart_SpuriousCongestion_StateNotRolledBack` | `HyStartState` |
| `HyStart_DelayIncreaseDetection_EtaCalculationAndCondition` | `HyStartAckCount`, `MinRttInCurrentRound`, `HyStartState` |

---

## CubeRootTest (separate test suite)

16 parametric test cases (`CubeRootCases/CubeRootTest.ReturnsExpectedValue/0..15`) test the `CubeRoot` utility function with boundary values, perfect cubes, non-perfect cubes, and large inputs. These are **specification-conformance** tests — they validate a pure mathematical function against known correct values.
---

Shall we remove the test of the last class of tests?

guhetier:
@saikat107 Nice categorization. I think all categories can stay in, ideally we can separate them with a banner comment:

```
//
//  Implementation specific test - tightly coupled
//
```

so future devs know what to expect

guhetier:
@saikat107 Can you please rebase the PR one last time? We fixed the cargo build, but this PR will need to pickup the change to complete the CI.
All looks good otherwise.

PR #5687, opened 2025-12-19
Add Datapath Cleanup Logs

anrossi:
## Description

Add logging to datapath clean up to better diagnose issues when clean up gets stuck.

## Testing

N/A

## Documentation

N/A

### DIFF
--- a/src/platform/datapath_epoll.c
+++ b/src/platform/datapath_epoll.c
@@ -309,8 +309,8 @@ CxPlatDataPathRelease(
 #endif
         CxPlatWorkerPoolRelease(Datapath->WorkerPool, CXPLAT_WORKER_POOL_REF_EPOLL);
 
-        QuicTraceLogInfo(
-            CxPlatDataPathRelease,
+        QuicTraceLogVerbose(
+            EpollDataPathRelease,
             "[data][%p] Datapath Freed",
             Datapath);
         CXPLAT_FREE(Datapath, QUIC_POOL_DATAPATH);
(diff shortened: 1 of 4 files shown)


src/platform/datapath_epoll.c
```diff
@@ -308,6 +308,11 @@ CxPlatDataPathRelease(
         Datapath->Freed = TRUE;
 #endif
         CxPlatWorkerPoolRelease(Datapath->WorkerPool, CXPLAT_WORKER_POOL_REF_EPOLL);
+
+        QuicTraceLogInfo(
```
anrossi: Should this be verbose instead of info?  It's verbose level on windows

src/platform/datapath_epoll.c
```diff
@@ -323,6 +328,10 @@ CxPlatProcessorContextRelease(
         CXPLAT_DBG_ASSERT(!DatapathPartition->Uninitialized);
         DatapathPartition->Uninitialized = TRUE;
 #endif
+        QuicTraceLogInfo(
+            CxPlatProcessorContextRelease,
```
anrossi: Remove the CxPlat part of the event name and use the datapath name instead

src/platform/datapath_epoll.c
ProjectsByJackHe: For consistency, yes.

guhetier: Verbose. It's a noisy, generally not relevant operation, it doesn't deserve the info level: if you start looking at the alloc / free logs, you are already deep in debugging.

Issue #5417, opened 2025-09-03
Failed to build msquic v2.5.1-beta on Windows from the Rust crate


masa-koz:
### Describe the bug

I tried to use the latest rust crate (msquic v2.5.1-beta), but failed to build it on Windows. The build error is following.

` C:\Users\MasahiroKozuka\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\msquic-2.5.1-beta\src\platform\datapath_raw_xdp_win.c(16,1): error C1083: Cannot open include file: 'xdp/wincommon.h': No such file or directory [C:\Users\MasahiroKozuka\source\repos\h3-masque\target\debug\build\msquic-25a18f5e61db4d50\out\build\src\platform\msquic_platform.vcxproj]`

I checked the directory of msquic v2.5.1-beta, and I found that "/submodules/xdp-for-windows/published/external" was missing.

### Affected OS

- [x] Windows
- [ ] Linux
- [ ] macOS
- [ ] Other (specify below)

### MsQuic version

v2.5

### Steps taken to reproduce bug

> cargo new msquic-test
> cd msquic-test
> cargo add msquic
> cargo build

### Expected behavior

Succeed in building.

### Actual outcome

Failed to build.

shinexus:
c:\Users\JimTsui\Documents\GitHub\LearnToProgram\msquic-2.5.5>cmake --build --preset release
适用于 .NET Framework MSBuild 版本 17.14.23+b0019275e

  MsQuicEtw_Resource.vcxproj -> C:\Users\JimTsui\Documents\GitHub\LearnToProgram\msquic-2.5.5\build\win-release-schanne
  l\MsQuicEtw_Resource.dir\Release\MsQuicEtw_Resource.lib
  core.vcxproj -> C:\Users\JimTsui\Documents\GitHub\LearnToProgram\msquic-2.5.5\build\win-release-schannel\obj\Release\
  core.lib
  core_fuzz.vcxproj -> C:\Users\JimTsui\Documents\GitHub\LearnToProgram\msquic-2.5.5\build\win-release-schannel\obj\Rel
  ease\core_fuzz.lib
  datapath_raw_xdp_win.c
C:\Users\JimTsui\Documents\GitHub\LearnToProgram\msquic-2.5.5\src\platform\datapath_raw_xdp_win.c(16,1): error C1083: 无
[...]

PR #5688, opened 2025-12-19
Add compiler toggle for xdp on windows

mriise:
XDP for windows [requires windows server](https://github.com/microsoft/xdp-for-windows/blob/main/docs/usage.md#prerequisites), add a toggle so msquic can be built without xdp.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,6 +102,7 @@ option(QUIC_EXTERNAL_TOOLCHAIN "Enable if system libs and include paths are conf
 option(QUIC_PGO "Enables profile guided optimizations" OFF)
 option(QUIC_LINUX_IOURING_ENABLED "Enables io_uring support" OFF)
 option(QUIC_LINUX_XDP_ENABLED "Enables XDP support" OFF)
+option(QUIC_WINDOWS_XDP_ENABLED "Enables Windows XDP support" OFF)
 option(QUIC_SOURCE_LINK "Enables source linking on MSVC" ON)
 option(QUIC_EMBED_GIT_HASH "Embed git commit hash in the binary" ON)
 option(QUIC_PDBALTPATH "Enable PDBALTPATH setting on MSVC" ON)
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -91,6 +91,8 @@ find = []
 # This feature enables openssl on windows, and has no effect on linux.
 openssl = []
 quictls = []
+# Enable Windows XDP datapath support when building from source.
+windows-xdp = []
 static = ["src"]
 preview-api = []
 # Overwrite generated binding by reruning the bindgen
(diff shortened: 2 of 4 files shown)


mriise:
@microsoft-github-policy-service agree company="Jinxxy Technologies LLC"

mriise:
@microsoft-github-policy-service agree company="Jinxxy Technologies LLC"

guhetier:
Can you please clarify the issue you are trying to solve?

MsQuic should compile without issue on Windows Client, XDP is a header only dependency.
Whether XDP is used or not is controlled by a runtime setting.

ProjectsByJackHe:
Expanding on what [guhetier](https://github.com/guhetier) said, there are 2 components to XDP-for-windows: the runtime kernel driver, and the header-only user space library. MsQuic does not build the runtime kernel driver, it just links against the header only user space library.

The documentation link you referenced pertains to the kernel driver, which MsQuic does not compile.

mriise:
Apologies for incorrect documentation, should have done a bit more work explaining.  

`datapath_raw_xdp_win.c` brings in a header [here](https://github.com/microsoft/msquic/blob/9639f53d7358e4990000dd70bfd61d771d242bef/src/platform/datapath_raw_xdp_win.c#L16) `#include <xdp/wincommon.h>` to this [file](https://github.com/microsoft/xdp-for-windows/blob/release/1.1/published/external/xdp/wincommon.h), which is inside the the xdp submodule.

This causes a build failure if the submodule is not pulled in, and when `xdp-for-windows/published/external` is otherwise [not included](https://github.com/microsoft/msquic/pull/5688/files#diff-dcb90aa17f202ae1dc738b8ef233f6d0dcba3bc4aaed0b3c55563ba8ba9acae2R137) in the target.

This is the same build failure in #5417 as well. If this is supposed to reference a header in the main repo then it is missing, otherwise I would really prefer to have a compile flag to avoid needing to pull in a submodule I don't have the intention of using.

mtfriesen:
@mriise, MsQuic depends on several submodules being present on build machines - is there a reason why the XDP submodule can't be initialized in your builds? I understand the code may be effectively unused in your scenario, but right now we're trying to simplify and reduce our build config matrix because of exploding overhead of validating and testing all the combinations.

We'd like to avoid adding the no-XDP config to our CI test matrix, but we also don't want to frustrate you or other developers who find the hypothetical no-XDP config broken.

mriise:
I am coming from trying to use this in rust. The published crate does not include xdp in the published version (see https://docs.rs/crate/msquic/2.5.1-beta/source/submodules/). Publisher of the crate likely didn't have submoudles (at least not win-xdp) initialized during publish. From what I can tell CI is correct, but `prepare-machine.ps1` wont init win-xdp on linux. That may be cause of the mistake as im assuming the publish was done manually.

This just an approach to fixing builds for an improperly published rust crate. I am fine with closing this assuming a usable crate is published and/or release procedure for the rust crate is updated.

mtfriesen:
That makes sense. @guhetier, could you take a look at the rust build/packaging?

guhetier:
The rust package issue is a known one we didn't get time to address. I'll look at it.
See #5417

PR #5690, opened 2025-12-19
Persists the virtual buffer size when switching to app owned buffers

### LINKED ISSUE #5672: StreamProvideReceiveBuffers doesn't invoke QUIC_STREAM_EVENT_RECEIVE_BUFFER_NEEDED  on server end for a request
### Describe the bug

As a part of my feature development, I was trying to use StreamProvideReceiveBuffers on the server end to provide buffers for requests with an initial buffer size of 4K. I observed the following:
1. When the request size <= 4K, I get a QUIC_STREAM_EVENT_RECEIVE on the server and the RPCs complete without issues.
2. When the request size > 4K, I am getting a QUIC_STREAM_EVENT_SEND_SHUTDOWN_COMPLETE event instead of QUIC_STREAM_EVENT_RECEIVE_BUFFER_NEEDED. 

Client streams with app provided buffers throw these events properly and work without issues. So this seems to be a bug only for server side streams

### Affected OS

- [x] Windows
- [ ] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

_No response_

### MsQuic version

v2.5

1. Initiate a request from client with a request payload size of X bytes
2. Provide buffers for a new stream on the server end as a part of QUIC_CONNECTION_EVENT_PEER_STREAM_STARTED through StreamProvideReceiveBuffers  and provide a buffer of size < X bytes

### Expected behavior

QUIC_STREAM_EVENT_RECEIVE_BUFFER_NEEDED

### Actual outcome

QUIC_STREAM_EVENT_SEND_SHUTDOWN_COMPLETE 

### Additional details

_No response_

guhetier:
## Description

When an endpoint receives a new stream indication, it can provide some receive buffers inline to convert the stream to app-owned buffer mode.
When doing so, the receive buffer struct of the stream is reinitialized, and the virtual buffer length (which correspond to the stream receive window) was set to zero.

This is incorrect since we introduced the "more buffer needed" notification and the receive window is no longer tied to the amount of buffer provided.

This would result in a receive failure with data received outside of the virtual size.

When converting the receive buffer to app-buffer mode, the virtual size is now preserved.

Fixes #5672.

## Testing

Test didn't catch this issue because the amount of buffer provided were too big and a single datagram would not allow to go over the virtual size.

Adding more test case:
- test all sender / receiver configurations
- test with various buffers numbers and sizes

While this is a large churn of test code to implement the various scenarios, the tested logic is still the same:
- provide some amount of receive buffer
- send data
- provide more receive buffers after some amount of data is received, or upon a buffer needed notification
- validate all the data is received correctly

## Documentation

Clarify documentation about the app provided receive buffer ownership.

### DIFF
--- a/src/test/bin/quic_gtest.cpp
+++ b/src/test/bin/quic_gtest.cpp
@@ -2379,7 +2379,9 @@ TEST(Misc, StreamMultiReceive) {
 struct WithAppProvidedBuffersConfigArgs: public testing::TestWithParam<AppProvidedBuffersConfig> {
     static ::std::vector<AppProvidedBuffersConfig> Generate() {
         return {
-            { 8, 0x500, 8, 0x500}
+            { 8, 0x500, 8, 0x500}, // Base scenario
+            { 1, 100, 1, 100}, // Small buffers
+            { 150, 0x50, 150, 0x50}, // Many buffers
         };
     }
 };


src/test/lib/DataTest.cpp
```diff
@@ -5043,19 +5181,112 @@ QuicTestStreamAppProvidedBuffersOutOfSpace_ServerSend(
...
+    TEST_EQUAL(0, memcmp(Buffers.SendDataBuffer.get(), Buffers.ReceiveDataBuffer.get(), Buffers.SendDataSize));
+}
+
 void
 QuicTestStreamAppProvidedBuffersOutOfSpace(
     )
 {
-    QuicTestStreamAppProvidedBuffersOutOfSpace_ClientSend();
-    QuicTestStreamAppProvidedBuffersOutOfSpace_ServerSend();
+    QuicTestStreamAppProvidedBuffersOutOfSpace_ClientSend_AbortStream();
```
ProjectsByJackHe: Nit / opinion, but would it be better to just define 1 helper, and use parameter to configure the use case? There is quite a bit of duplication in all these functions.

guhetier: I will change that bit once the test refactor PR is merged, but actually, going the other direction to make the test fully independent.
There is some amount of duplication, especially in the boilerplate to setup a connection, but that is preferrable in test code over parametrization that makes the test logic and validation conditional. That makes a single test failure very hard to investigate, because you need to keep all the possible code paths in your head to understand what is the scenario (look at "ConnectAndPing"...)

It is generally better to go with a toolbox approach, with simple helpers you can compose (like the struct helper that takes care of buffer allocations) but a single scenario per tests. We could definitely use a "Setup two connected MsQuic connection for me" helper, that's on my list ^^
(also,
- parameters that don't change the test scenario, like IPv4 or IPv6, a perfectly fine, this is really about keeping what you try to test clear
- this applies to test code, in product code, you generally want to parametrize more to handle all possible scenarios
)

docs/Streams.md
```diff
@@ -167,6 +167,7 @@ There is no guarantee the `QUIC_BUFFER`s indicated in a receive notification wil
 
 The application is responsible for tracking the amount of data received and when a buffer it provided has been fully used.
 The application regains full ownership of a buffer after it get a receive notification for all bytes in the buffer and accept them by calling [StreamReceiveComplete](api/StreamReceiveComplete.md).
+MsQuic will not re-use a buffer for subsequent receives once it has been fully written, unless the app submit the same memory again.
```
mtfriesen:
```suggestion
MsQuic will not re-use a buffer for subsequent receives once it has been fully written, unless the app submits the same buffer again.
```

src/core/stream.c
```diff
@@ -977,6 +977,15 @@ QuicStreamSwitchToAppOwnedBuffers(
     _In_ QUIC_STREAM* Stream
     )
 {
+    //
+    // Preserve the initial receive window size.
+    // It should not have evolved yet since the stream creation.
```
mtfriesen:
```suggestion
    // Preserve the initial receive window size: it should not have changed
    // since the stream's creation.
```

src/core/stream.c
```diff
@@ -977,6 +977,15 @@ QuicStreamSwitchToAppOwnedBuffers(
...
+    //
+    const uint32_t InitialControlFlow = Stream->RecvBuffer.VirtualBufferLength;
+    CXPLAT_DBG_ASSERT(
+        InitialControlFlow == Stream->Connection->Settings.StreamRecvWindowBidiRemoteDefault ||
+        InitialControlFlow == Stream->Connection->Settings.StreamRecvWindowBidiLocalDefault);
```
mtfriesen: What about `StreamRecvWindowUnidiDefault`?

guhetier: Good catch, `StreamRecvWindowUnidiDefault` should have been here instead of `StreamRecvWindowBidiLocalDefault`.
This code path is only possible for a stream opened by the peer (the API is only callable in a "peer stream start" event callback), so `StreamRecvWindowBidiLocalDefault` is not relevant here.


```suggestion
    CXPLAT_DBG_ASSERT(
        InitialControlFlow == Stream->Connection->Settings.StreamRecvWindowBidiRemoteDefault ||
        InitialControlFlow == Stream->Connection->Settings.StreamRecvWindowUnidiDefault);
```

src/core/stream.c
```diff
@@ -978,13 +978,13 @@ QuicStreamSwitchToAppOwnedBuffers(
...
-    // Preserve the initial receive window size.
-    // It should not have evolved yet since the stream creation.
+    // Preserve the initial receive window size: it should not have changed
+    // since the stream's creation.
     //
     const uint32_t InitialControlFlow = Stream->RecvBuffer.VirtualBufferLength;
     CXPLAT_DBG_ASSERT(
         InitialControlFlow == Stream->Connection->Settings.StreamRecvWindowBidiRemoteDefault ||
-        InitialControlFlow == Stream->Connection->Settings.StreamRecvWindowBidiLocalDefault);
+        InitialControlFlow == Stream->Connection->Settings.StreamRecvWindowUnidiDefault);
```
mtfriesen, line 987: Presumably this unidirectional stream remains untested?

guhetier: It is tested, but the default value happens to be the same.

PR #5693, opened 2025-12-23
Add support for in-memory certificate root stores.

Myriachan:
## Description

Add support for in-memory certificate root stores.
This adds new credential flag `QUIC_CREDENTIAL_FLAG_SET_CA_CERTIFICATE_BLOB` and new `QUIC_CREDENTIAL_CONFIG` fields `CaCertificateBlob` and `CaCertificateBlobLength`.

Only the OpenSSL and quictls backends are supported for now.

## Testing

This needs new tests.  As a result, this is a draft pull request only.

## Documentation

This pull request updates the documentation according to the changes.


Myriachan:
I submitted this pull request as a draft to get comments on it and talk about next steps.

Known remaining things as I write this comment:

 * Testing.  The current codebase has no test for `QUIC_CREDENTIAL_FLAG_SET_CA_CERTIFICATE_FILE`, so a test for that is a prerequisite to a test for the new `QUIC_CREDENTIAL_FLAG_SET_CA_CERTIFICATE_BLOB`.  I don't know how to create or run the tests in this codebase.
 * Rust.  As this is an API change, the Rust bindings need to be updated.  I don't know how to do this as I don't know Rust.
 * C#.  I know enough C# to get by, but I don't know what needs to change for its bindings.
 * SChannel.  I have another change I'm working on to add support for `QUIC_CREDENTIAL_FLAG_SET_CA_CERTIFICATE_FILE` and `_BLOB`.  It's considerably more complicated, though, as it involves searching a text file for `----BEGIN CERTIFICATE-----` and passing those to SChannel's `CertAddCertificateContextToStore`.
   * This might make more sense as a separate pull request to do after this one.
 * NT kernel mode.  I have absolutely no idea what the TLS interface is in kernel mode.

guhetier:
Hi @Myriachan, thanks for opening this PR. Sorry for the delay answering, I'll try to review it soon.

src/inc/msquic.h
```diff
@@ -416,6 +417,8 @@ typedef struct QUIC_CREDENTIAL_CONFIG {
     QUIC_CREDENTIAL_LOAD_COMPLETE_HANDLER AsyncHandler; // Optional
     QUIC_ALLOWED_CIPHER_SUITE_FLAGS AllowedCipherSuites;// Optional
     const char* CaCertificateFile;                      // Optional
+    const uint8_t* CaCertificateBlob;                   // Optional
```
guhetier, line 420: These two new members are exclusive with `CaCertificateFile`, I think it would make sense to group them in a union or to re-use `CaCertificateFile` as suggeseted by #5715 (since `CaCertificateBlob` is meant to hold a string IIUC)

src/platform/tls_openssl.c
```diff
@@ -2135,20 +2135,125 @@ CxPlatTlsSecConfigCreate(
         }
     }
 
-    if (CredConfigFlags & QUIC_CREDENTIAL_FLAG_SET_CA_CERTIFICATE_FILE &&
+    if ((CredConfigFlags & QUIC_CREDENTIAL_FLAG_SET_CA_CERTIFICATE_FILE) &&
         CredConfig->CaCertificateFile) {
+        X509_STORE* CertStore = SSL_CTX_get_cert_store(SecurityConfig->SSLCtx);
```
guhetier, line 2140: What is the motivation behind that change? `SSL_CTX_load_verify_locations` seems to me the typical way to handle this change.

src/platform/tls_openssl.c
```diff
@@ -2135,20 +2135,125 @@ CxPlatTlsSecConfigCreate(
...
+    if (CredConfigFlags & QUIC_CREDENTIAL_FLAG_SET_CA_CERTIFICATE_BLOB &&
+        CredConfig->CaCertificateBlob) {
+        X509_STORE* CertStore = NULL;
+        BIO* CertBio = NULL;
+        STACK_OF(X509_INFO)* CertStack = NULL;
+
+        Status = QUIC_STATUS_TLS_ERROR;
+
+        // BIO_new_mem_buf() takes 'int' length
+        if (CredConfig->CaCertificateBlobLength > (unsigned)INT_MAX) {
```
guhetier, line 2175: We try to use sized integer types, and this is the type of `CaCertificateBlobLength`

```suggestion
        if (CredConfig->CaCertificateBlobLength > (uint32_t)INT_MAX) {
```

src/platform/tls_openssl.c
```diff
@@ -2135,20 +2135,125 @@ CxPlatTlsSecConfigCreate(
...
+                        "X509_STORE_add_crl failed");
+                    Status = QUIC_STATUS_TLS_ERROR;
+                    goto BlobExit;
+                }
+            }
+        }
+
+        Status = QUIC_STATUS_SUCCESS;
+
+    BlobExit:
```
guhetier, line 2248: Could you please extract the code above in an helper function, if it needs it own cleanup section? It will be less error prone.

guhetier (commented):
> I submitted this pull request as a draft to get comments on it and talk about next steps.
> 
> Known remaining things as I write this comment:
> 
> * Testing.  The current codebase has no test for `QUIC_CREDENTIAL_FLAG_SET_CA_CERTIFICATE_FILE`, so a test for that is a prerequisite to a test for the new `QUIC_CREDENTIAL_FLAG_SET_CA_CERTIFICATE_BLOB`.  I don't know how to create or run the tests in this codebase.

Test for this would likely fit in the "HandshakeTest.cpp" file. A simplified version of `QuicTestConnect` that only focuses on establishing a connection with the certificate provided through different mean would be a good starting point. 
"ApiTest.cpp" is typically the home for test validating the invalid API invocation cases.

Test supported only for a specific TLS provider are skipped using defines like `QUIC_TEST_SCHANNEL_FLAGS`.

> * Rust.  As this is an API change, the Rust bindings need to be updated.  I don't know how to do this as I don't know Rust.

Rust bindings are generated automatically - if you are not familiar with Rust, the diff you need to apply will be reported by the CI.
The Rust wrappers need to be written by hand, but I don't think it is a problem if they don't support this feature for now, they are a work in progress.

> * C#.  I know enough C# to get by, but I don't know what needs to change for its bindings.

Running `generate-dotnet.ps1` will automatically generate the C# bindings, no manual code is needed.

> * SChannel.  I have another change I'm working on to add support for `QUIC_CREDENTIAL_FLAG_SET_CA_CERTIFICATE_FILE` and `_BLOB`.  It's considerably more complicated, though, as it involves searching a text file for `----BEGIN CERTIFICATE-----` and passing those to SChannel's `CertAddCertificateContextToStore`.
>   
>   * This might make more sense as a separate pull request to do after this one.

Sounds good, I also think a different PR will make more sense. We already have some features that are supported only for some crypto provider.

> * NT kernel mode.  I have absolutely no idea what the TLS interface is in kernel mode.

In my opinion, this doesn't need to be supported for the NT kernel.

I added a few comments. I am far from an expert in OpenSSL, but the approach seems good to me.

src/platform/tls_openssl.c
Myriachan: I read somewhere that `SSL_CTX_load_verify_locations` is deprecated.  But maybe it isn't anymore?

src/platform/tls_openssl.c
Myriachan: My logic was that I only cared that the right side is an unsigned type for comparison, and `unsigned` is the unsigned version of `int`.  In other situations, it would be more robust against type changes.  This is an artifact of my coding style showing through, and not something from this particular situation.

However, here, it doesn't matter.  If `int` were to someday become 64-bit, a `uint32_t` cast here would do an unsigned compare against `0xFFFFFFFF` rather than `0x7FFFFFFFFFFFFFFF`.  It probably would cause a truncation-of-constant warning.  But `CaCertificateBlobLength` as a `uint32_t` can't be larger than `0xFFFFFFFF` anyway, so functionality-wise, it doesn't matter.

I'm just showing my paranoid style that led to why it's `(unsigned)` here rather than `(uint32_t)`.  But it's fine either way, so I don't mind committing it.

src/platform/tls_openssl.c
Myriachan: Yeah that sounds better; will do.

src/inc/msquic.h
Myriachan: Yes, that could work.  The certificate store format is traditionally a text file, though as far as I know nothing in OpenSSL stops you from passing a binary blob that happens to have a `----- BEGIN CERTIFICATE -----` text block inside it.  OpenSSL ignores anything outside these blocks.  That's why mentally I came up with `uint8_t`, but it ultimately doesn't matter.

I'll look into the naming/union more here.

src/platform/tls_openssl.c
guhetier: I see your reasoning, it makes sense. I was thinking the other way, casting to the type of `CaCertificateBlobLength` as the maximum value that the parameter can takes.

Either works here for me.

src/platform/tls_openssl.c
guhetier: Good catch, I could find this commit:
https://github.com/openssl/openssl/commit/6dcb100f89d0ef081771d533fed342412ac7a13f

PR #5707, opened 2026-01-08
Fix hang in ExecutionDelete

anrossi:
## Description

In external execution mode, when cleaning up the last socket (binding) and registration and calling `ExecutionDelete`, it's possible the worker pool will attempt to be deleted by `ExecutionDelete` before the asynchronous shut down callbacks have had time to run. This will result in a hang because the ref count on the worker pool is non-zero, since there's still cleanup work to do. However, the caller, by calling `ExecutionDelete` has indicated that they have stopped the execution engine, so that remaining cleanup work won't get a chance to run.

## Testing

Tested the fix with the partner scenario

## Documentation

No documentation change.

### DIFF
--- a/src/platform/platform_worker.c
+++ b/src/platform/platform_worker.c
@@ -18,8 +18,9 @@ Abstract:
 typedef struct QUIC_CACHEALIGN CXPLAT_WORKER {
 
     //
-    // Thread used to drive the worker. Only set when the worker is created and
-    // managed internally (default case).
+    // Thread used to drive the worker, when the worker is created and
+    // managed internally (default case). In the external execution case,
+    // this thread is the cleanup thread.
     //
     CXPLAT_THREAD Thread;
 
@@ -68,6 +69,12 @@ typedef struct QUIC_CACHEALIGN CXPLAT_WORKER {
     //
     CXPLAT_SLIST_ENTRY* ExecutionContexts;
 
+    //
+    // The event for the cleanup worker to wait on to start cleanup.
+    // NULL in the internally-managed execution case.
+    //
+    CXPLAT_EVENT* CleanupEvent;
+
 #if DEBUG // Debug statistics
     uint64_t LoopCount;
     uint64_t EcPollCount;
@@ -92,7 +99,6 @@ typedef struct QUIC_CACHEALIGN CXPLAT_WORKER {
     BOOLEAN StoppingThread : 1;
     BOOLEAN StoppedThread : 1;
     BOOLEAN DestroyedThread : 1;
-    BOOLEAN DrainingEvents : 1;
 #if DEBUG // Debug flags - Must not be in the bitfield.
     BOOLEAN ThreadStarted;
     BOOLEAN ThreadFinished;
@@ -108,6 +114,7 @@ typedef struct QUIC_CACHEALIGN CXPLAT_WORKER {
 typedef struct CXPLAT_WORKER_POOL {
 
     CXPLAT_RUNDOWN_REF Rundown;
+    CXPLAT_EVENT CleanupEvent; // Only used with external workers.
     uint32_t WorkerCount;
     BOOLEAN External;
 
@@ -163,11 +170,6 @@ UpdatePollCompletion(
 }
 
 void
-CxPlatWorkerPoolDrainEvents(
-    _In_ CXPLAT_WORKER_POOL* WorkerPool
-    );
-
-void
 CxPlatProcessEvents(
     _In_ CXPLAT_WORKER* Worker
     );
@@ -175,9 +177,10 @@ CxPlatProcessEvents(
 BOOLEAN
 CxPlatWorkerPoolInitWorker(
     _Inout_ CXPLAT_WORKER* Worker,
+    _In_ CXPLAT_WORKER_POOL* WorkerPool,
     _In_ uint16_t IdealProcessor,
     _In_opt_ CXPLAT_EVENTQ* EventQ, // Only for external workers
-    _In_opt_ CXPLAT_THREAD_CONFIG* ThreadConfig // Only for internal workers
+    _In_ CXPLAT_THREAD_CONFIG* ThreadConfig
     )
 {
     CxPlatLockInitialize(&Worker->ECLock);
@@ -186,6 +189,7 @@ CxPlatWorkerPoolInitWorker(
     Worker->IdealProcessor = IdealProcessor;
     Worker->State.WaitTime = UINT32_MAX;
     Worker->State.ThreadID = UINT32_MAX;
+    Worker->CleanupEvent = &WorkerPool->CleanupEvent;
 
     if (EventQ != NULL) {
         Worker->EventQ = *EventQ;
@@ -351,7 +355,7 @@ CxPlatWorkerPoolCreate(
 
         CXPLAT_WORKER* Worker = &WorkerPool->Workers[i];
         if (!CxPlatWorkerPoolInitWorker(
-                Worker, IdealProcessor, NULL, &ThreadConfig)) {
+                Worker, WorkerPool, IdealProcessor, NULL, &ThreadConfig)) {
             goto Error;
         }
     }
@@ -379,6 +383,15 @@ Error:
     CXPLAT_FREE(WorkerPool, QUIC_POOL_PLATFORM_WORKER);
 
     return NULL;
+}
+
+CXPLAT_THREAD_CALLBACK(CxPlatExecutionCleanupThread, Context)
+{
+    CXPLAT_WORKER* Worker = (CXPLAT_WORKER*)Context;
+
+    CxPlatEventWaitForever(*Worker->CleanupEvent);
+
+    return CxPlatWorkerThread(Context);
 }
 
 _Success_(return != NULL)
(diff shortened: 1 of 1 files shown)


src/platform/platform_worker.c
```diff
@@ -666,6 +687,42 @@ CxPlatWorkerPoolWorkerPoll(
     return Worker->State.WaitTime;
 }
 
+void
+CxPlatWorkerPoolWorkerDrainEvents(
```
anrossi: This runs the workers all on the same thread, which is not the original thread the worker was operating on; this is a deviation from the assumptions the workers have.
I think the only way to guarantee the workers run on their original threads is to have the caller keep running the workers after the last Registration is cleaned up, and provide some signal that all work has ceased, before calling `ExecutionDelete`

src/platform/platform_worker.c
```diff
@@ -449,6 +459,17 @@ CxPlatWorkerPoolDelete(
 #else
         UNREFERENCED_PARAMETER(RefType);
 #endif
+
+        if (RefType == CXPLAT_WORKER_POOL_REF_EXTERNAL) {
```
mtfriesen: I don't think we should lean on the `RefType` for functionality. See the `UNREFERENCED_PARAMETER(RefType)` in the non-`DEBUG` case directly above.

src/platform/platform_worker.c
```diff
@@ -666,6 +687,42 @@ CxPlatWorkerPoolWorkerPoll(
...
+        //
+        // Set the wait time to zero here to process as soon as possible.
+        // Otherwise, CxPlatProcessEvents may wait this many milliseconds.
+        //
+        Worker->State.WaitTime = 0;
+
+        //
+        // Assume there is no work to do, and this will update to zero if work was done.
+        //
+        Worker->State.NoWorkCount = 1;
```
mtfriesen: Can we deduplicate/refactor this function? I don't think it will be maintainable to keep this as a separate function used only for external execution cleanup.

src/platform/platform_worker.c
mtfriesen: I don't think there's any requirement or expectation that execution (or "work") that could have or would have been done by the external execution contexts be affinitized to any given thread.

src/platform/platform_worker.c
anrossi: I'll look into storing a boolean on the WorkerPool

src/platform/platform_worker.c
```diff
@@ -666,6 +687,42 @@ CxPlatWorkerPoolWorkerPoll(
...
+        if (Worker->State.WaitTime && InterlockedFetchAndClearBoolean(&Worker->Running)) {
+            Worker->State.TimeNow = CxPlatTimeUs64();
+            CxPlatRunExecutionContexts(Worker); // Run once more to handle race conditions
+        }
+
+        //
+        // Set the wait time to zero here to process as soon as possible.
+        // Otherwise, CxPlatProcessEvents may wait this many milliseconds.
+        //
+        Worker->State.WaitTime = 0;
```
mtfriesen: 1. We do want to wait, _somewhere_, else we've implemented a busy loop, wasting CPU cycles and watts (at best) or preempting some other code that would unblock the events we're waiting for (worse).
2. The _somewhere_ should span all the workers being run down, since there may be dependencies across the workers, e.g., an event on worker[0] won't fire until worker[1] executes for a bit.

src/platform/platform_worker.c
```diff
@@ -666,6 +687,42 @@ CxPlatWorkerPoolWorkerPoll(
...
+
+        //
+        // Assume there is no work to do, and this will update to zero if work was done.
+        //
+        Worker->State.NoWorkCount = 1;
+        CxPlatProcessEvents(Worker);
+
+#if DEBUG
+        CXPLAT_DBG_ASSERTMSG(++Iterations < 10, "Is the library still active?");
```
mtfriesen: We can't assert a limit on the number of iterations, though we may want to assert a maximum wall-clock time spent attempting to stop. That said, those watchdogs would probably best be implemented at a higher level.

src/platform/platform_worker.c
```diff
@@ -666,6 +687,42 @@ CxPlatWorkerPoolWorkerPoll(
...
+#endif
+    } while (Worker->State.NoWorkCount == 0);
```
mtfriesen: The worker not doing any work in one iteration (especially in a busy loop, where it doesn't get a chance to wait for events) doesn't imply it's done doing work forever.

I'm actually not sure what we'd consider "done" to be? The notion hasn't been formalized in this context AFAIK.

src/platform/platform_worker.c
anrossi: The key problem to solve with waiting is that if a `CXPLAT_EXECUTION_CONTEXT` has no work to do, it'll tell the worker to wait indefinitely (`UINT32_MAX` milliseconds). But there's also a case where the next-time-to-run timestamp is more than `UINT32_MAX` milliseconds in the future, so the value `UINT32_MAX - 1` is used to indicate that. (See `CxPlatRunExecutionContexts` line 661).
This `WaitTime` member is used as the timeout parameter to `CxPlatEventQDequeue` which will pass it along to a platform function (e.g. `epoll_wait`) which will block for that time, or until an event comes in. (see `CxPlatProcessEvents` line 797).

Since `ExecutionDelete` is supposed to be the second-to-last API called (before `MsQuicClose`), we can't expect an event to come in, so it'll block indefinitely.

I'm thinking that we could iterate over all workers calling `CxPlatRunExecutionContexts` to get the largest wait that is less-than `UINT32_MAX - 1`, wait that length, then iterate over all the workers calling `CxPlatProcessEvents`.  This process repeats until all workers set the `WaitTime` to `UINT32_MAX` in `CxPlatRunExecutionContexts`. Then there's no new work scheduled.
A FRE assert should be added crash if the `WaitTime` is set to `UINT32_MAX - 1`. But maybe that should be any value over 10,000.

anrossi: Actually, I think part of that is wrong. DrainEvents shouldn't run all the contexts and then process events. `ProcessEvents` should be run after each call to `RunContexts`.
However, if a context sets `WaitTime` to `UINT32_MAX`, `DrainEvents` should set `WaitTime` to 0 before calling process events to see if there are any events waiting without blocking. This way, it will pick up any pending events, while also not blocking indefinitely.

src/platform/platform_worker.c
anrossi: I think "done" in this context is when the worker has no more work left to do.  The way to measure that is the Worker's `WaitTime` being `UINT32_MAX`, and the `NoWorkCount` being non-zero.  This would mean the worker is waiting for an external event to tell it to do work, and `NoWorkCount` being non-zero means we waited for work and timed out with no work to do.

src/platform/platform_worker.c
anrossi, line 258: I just noticed this TODO.  Maybe if WorkerPool is refactored to use a plain reference count instead of a rundown, somehow this work could be put into this case?
This is an idle thought and much more thinking needs to be done to see if that even makes sense.

src/platform/platform_worker.c
mtfriesen: >Since ExecutionDelete is supposed to be the second-to-last API called (before MsQuicClose), we can't expect an event to come in

I'm not sure why we can't expect an event - the essence of the bug we're trying to fix is that non-zero work is still running after this "second-to-last API called" and events are just a type of work.

src/platform/platform_worker.c
```diff
@@ -160,12 +169,18 @@ UpdatePollCompletion(
     CxPlatUpdateExecutionContexts(Worker);
 }
 
+void
+CxPlatProcessEvents(
```
anrossi: delete this.

guhetier:
I am not sure I understand the issue fully: isn't `RegistrationShutdown` supposed to wait for all the callbacks?
Having a thread hanging there for the full lifetime of each worker seems pretty expensive and potentially very error prone.
It could also be problematic for systems that want/need full control over the execution, if we now start creating threads.

What is the scenario when we can't require the app to drive the execution until fully completed?

src/platform/platform_worker.c
```diff
@@ -368,6 +380,15 @@ CxPlatWorkerPoolCreate(
     return NULL;
 }
 
+CXPLAT_THREAD_CALLBACK(CxPlatExecutionCleanupThread, Context)
```
guhetier, line 383: This function needs comments, it isn't easily understandable without context.
Outside of this PR, it would be very hard to understand why we would ever need to start the worker thread on cleanup (and it already takes some looking around to know that `CxPlatWorkerThread` starts the worker thread, and doesn't, for instance, returns the worker thread handle from the context).

guhetier: The comment could echo what the comment lower ("In the case of external execution...") says, as in "When we take ownership of the execution context after the app call ExecutionDelete, this drive the execution ... yadayada... to execute the cleanup"

guhetier:
> I am not sure I understand the issue fully: isn't `RegistrationShutdown` supposed to wait for all the callbacks? Having a thread hanging there for the full lifetime of each worker seems pretty expensive and potentially very error prone. It could also be problematic for systems that want/need full control over the execution, if we now start creating threads.
> 
> What is the scenario when we can't require the app to drive the execution until fully completed?

After offline discussion:
- there seem to be agreement that this solution is a temporary hack and a better solution would be an asynchronous "MsQuicClose" call (or another mechanism allowing MsQuic to indicate to the app it can stop running the execution)
- there are risks with the approach:
    - if there are work items on the message queue that are not msquic ones, taking ownership of the queue would be a problem
    - if the app drive the execution because it wants full control over the threads, we break the contract

If this goes in as a temporary fix, I strongly think we need to address it properly before it goes out of preview, and even before we create a release 2.6 branch.

mtfriesen (changes requested):
Please describe the cost/risk/benefit of the possible fix strategies and reach consensus or convert to a draft PR.

PR #5714, opened 2026-01-14
Migrate a batch of test to the new IOCTL pattern

guhetier:
## Description

A generic IOCTL has been created to simplify running test in kernel mode.
This PR migrate tests taking no parameters and some with a single parameter to the new pattern.
Follow up PR will migrate more tests until the old pattern can be cleaned up fully.

## Testing

CI

## Documentation

N/A


mtfriesen:
LGTM but the kernel CI disagrees

guhetier:
> LGTM but the kernel CI disagrees

Looks like I missed one call, should be fixed now.

PR #5727, opened 2026-01-19
Added 1rtt fuzzing

### LINKED ISSUE #5721: Implement 1rtt fuzzer
Recvfuzz is a fuzzing tool in msquic that fuzzes packets in receive path. Currently we fuzz the intial packet and handshake packet only. We need to support 1rtt packet fuzzing too to increase code coverage.

gaurav2699:
## Description
Implemented 1rtt fuzzing in recvfuzz tool

Fixes #5721 .

## Purpose
Fuzzing is done to find vulnerabilites

## Testing
Locally

## Documentation
No


src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -627,6 +677,17 @@ void WriteClientPacket(
                 Buffer + *HeaderLength,
                 ClientContext,
                 PacketParams);
+        } else if (PacketParams->FrameTypes[i] == QUIC_FRAME_HANDSHAKE_DONE) {
+            WriteHandshakeDoneFrame(
+                &PayloadLength,
+                BufferSize,
+                Buffer + *HeaderLength);
+        } else if (PacketParams->FrameTypes[i] >= 0x08 && PacketParams->FrameTypes[i] <= 0x0f) {
```
guhetier: Please use constants here, `QUIC_FRAME_STREAM` to `QUIC_FRAME_STREAM_7`

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -856,44 +1079,29 @@ void FuzzInitial(
...
-    PacketParams.Mode = 1;
-    GetRandomBytes(sizeof(uint64_t), &PacketParams.SourceCid);
-    memcpy(&CurrSrcCid, PacketParams.SourceCid, sizeof(uint64_t)); // Save for receive path
-
-    TlsContext ClientContext;
-    ClientContext.CreateContext(PacketParams.SourceCid);
-    CXPLAT_FRE_ASSERT(ClientContext.ProcessData() & CXPLAT_TLS_RESULT_DATA);
-
     //
-    // Keep sending the packet until we receive a response or the time runs out.
```
guhetier, line 882: The old comment contains aditionnal information we lose, i'd keep it (adding your new line in addition)

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -903,33 +1111,32 @@ void FuzzHandshake(
 
         if (!Packet->DestCidLen ||
             memcmp(Packet->DestCid, &CurrSrcCid, sizeof(uint64_t)) != 0) {
-            continue; // Packet doesn't match our current connection
```
guhetier, line 906: Same here, please don't delete comment that explain the code flow.

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -953,37 +1160,114 @@ void FuzzHandshake(
...
+        PacketParams.PacketType = QUIC_HANDSHAKE_V1;
+        PacketParams.NumFrames = 1;
+        PacketParams.FrameTypes[0] = QUIC_FRAME_CRYPTO;
+        PacketParams.NumPackets = 1;
+        BuildAndSendPackets(Binding, Route, &PacketParams, &ClientContext, false);
+
+        if (ClientContext.State.WriteKeys[QUIC_PACKET_KEY_1_RTT] != nullptr) {
             //
-            // Send the rest of the handshake packets.
+            // Send HANDSHAKE_DONE packet in 1-RTT
```
guhetier: Consider randomizing this? (low chance we don't send it)

guhetier: And low chance we send it multiple time maybe?

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -856,44 +1079,29 @@ void FuzzInitial(
...
-    ClientContext.CreateContext(PacketParams.SourceCid);
-    CXPLAT_FRE_ASSERT(ClientContext.ProcessData() & CXPLAT_TLS_RESULT_DATA);
-
     //
-    // Keep sending the packet until we receive a response or the time runs out.
+    // Send initial packet to start handshake
     //
     do {
-        BuildAndSendPackets(Binding, Route, &PacketParams, &ClientContext, false); // Don't fuzz this one
+        BuildAndSendPackets(Binding, Route, PacketParams, ClientContext, false);
```
guhetier: Consider renaming this to `BuildAndSendLongHeaderPackets`?

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -768,6 +829,168 @@ void BuildAndSendPackets(
...
+                BufferSize,
+                Buffer + *HeaderLength);
+        } else if (PacketParams->FrameTypes[i] >= 0x08 && PacketParams->FrameTypes[i] <= 0x0f) {
+            // STREAM frame (types 0x08-0x0f)
+            WriteStreamFrame(
+                &PayloadLength,
+                BufferSize,
+                Buffer + *HeaderLength);
+        }
+        // 1-RTT packets typically contain STREAM frames, PING, etc.
```
guhetier: It would be interesting to add more frame types, and potentially allow unknown frame / invalid frames with some probablity.

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -995,11 +1279,13 @@ void FuzzReceivePath(CXPLAT_SOCKET* Binding, CXPLAT_ROUTE* Route) {
...
-
-        if (GetRandom<uint8_t>(16) == 0) {
+        uint8_t FuzzMode = GetRandom<uint8_t>(32);
+        if (FuzzMode == 0) {
             FuzzInitial(Binding, Route);
-        } else {
+        } else if (FuzzMode < 16) {
             FuzzHandshake(Binding, Route, StartTimeMs);
+        } else {
+            Fuzz1Rtt(Binding, Route, StartTimeMs);
```
guhetier, line 1635: IIUC, `FuzzHandshake` does the handshake + send a number of randomized long header packets. And `Fuzz1Rtt` does the handshake + send a number of randomized short header packets.

Is this correct?
Could we / should we have the capacity of sending a mix of short and long header packets?

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -768,6 +829,168 @@ void BuildAndSendPackets(
...
+                PacketParams->LargestAcknowledge,
+                &PayloadLength,
+                BufferSize,
+                Buffer + *HeaderLength);
+        } else if (PacketParams->FrameTypes[i] == QUIC_FRAME_HANDSHAKE_DONE) {
+            WriteHandshakeDoneFrame(
+                &PayloadLength,
+                BufferSize,
+                Buffer + *HeaderLength);
+        } else if (PacketParams->FrameTypes[i] >= 0x08 && PacketParams->FrameTypes[i] <= 0x0f) {
```
guhetier: Same here, please use constants.
Can this code be factored with the similar bit of `WriteClientPacket`? It seems similiar, so a `WriteFrame` function might be used to contain that loop after the header has been written.

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -953,37 +1160,114 @@ void FuzzHandshake(
...
-            break;
+            PacketParams.FrameTypes[0] = QUIC_FRAME_HANDSHAKE_DONE;
+            PacketParams.NumPackets = 1;
+            BuildAndSendShortHeaderPackets(Binding, Route, &PacketParams, &ClientContext, false);
+
+            //
+            // Send fuzzed 1-RTT packets with STREAM frames
+            //
+            PacketParams.NumFrames = 1;
+            PacketParams.FrameTypes[0] = QUIC_FRAME_STREAM; // STREAM frame base type
```
guhetier: Do we always only send a single type of frame per packet?

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -627,6 +677,17 @@ void WriteClientPacket(
...
                 ClientContext,
                 PacketParams);
+        } else if (PacketParams->FrameTypes[i] == QUIC_FRAME_HANDSHAKE_DONE) {
+            WriteHandshakeDoneFrame(
+                &PayloadLength,
+                BufferSize,
+                Buffer + *HeaderLength);
+        } else if (PacketParams->FrameTypes[i] >= 0x08 && PacketParams->FrameTypes[i] <= 0x0f) {
+            // STREAM frame (types 0x08-0x0f)
+            WriteStreamFrame(
```
guhetier: This always send a base stream frame, that doesn't match what the condition promise

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -519,6 +519,56 @@ void WriteAckFrame(
...
+{
+    uint64_t StreamId = GetRandom<uint8_t>(4); // Use stream IDs 0-3 (bidirectional client-initiated)
+    uint64_t StreamOffset = GetRandom<uint8_t>(2) == 0 ? 0 : GetRandom<uint16_t>(1000);
+    uint16_t DataLength = GetRandom<uint8_t>(200) + 10; // 10-210 bytes of data
+    uint8_t Fin = GetRandom<uint8_t>(4) == 0 ? 1 : 0; // 25% chance of FIN
+
+    // STREAM frame type with LEN and OFF bits set (0x0A)
+    uint8_t FrameType = 0x08; // Base STREAM frame
+    if (StreamOffset > 0) FrameType |= 0x04; // OFF bit
+    FrameType |= 0x02; // LEN bit (always include length)
```
guhetier: Why always include the length? We should randomize this too

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -519,6 +519,56 @@ void WriteAckFrame(
...
+    Buffer[(*Offset)++] = QUIC_FRAME_HANDSHAKE_DONE;
+}
+
+void WriteStreamFrame(
+    _Inout_ uint16_t* Offset,
+    _In_ uint16_t BufferLength,
+    _Out_writes_to_(BufferLength, *Offset) uint8_t* Buffer
+    )
+{
+    uint64_t StreamId = GetRandom<uint8_t>(4); // Use stream IDs 0-3 (bidirectional client-initiated)
```
guhetier: Could we throw some unidirectional in the pool?

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -519,6 +519,56 @@ void WriteAckFrame(
...
+        *Offset += QuicVarIntSize(StreamOffset);
+    }
+
+    // Encode Length
+    QuicVarIntEncode(DataLength, Buffer + *Offset);
+    *Offset += QuicVarIntSize(DataLength);
+
+    // Write random stream data
+    CXPLAT_FRE_ASSERT(*Offset + DataLength <= BufferLength);
+    GetRandomBytes(DataLength, Buffer + *Offset);
```
guhetier: Should we sometime provide an more or less bytes than we promise in the stream frame?

src/tools/recvfuzz/recvfuzz.cpp
gaurav2699: Fuzz handshake does the handshake and sends a number of randomized handshake packets.  Essentially the packet that is supposed to have the handshake crypto data.

src/tools/recvfuzz/recvfuzz.cpp
gaurav2699: Currently yeah, my thinking is to fist make a working model and add on top of that.

guhetier: Sounds good. Please don't hesitate to file a task with what you think we need to add, so we don't forget and progressively extend support.

src/tools/recvfuzz/recvfuzz.cpp
gaurav2699: > Could we / should we have the capacity of sending a mix of short and long header packets?

I think I will add that in 1rtt fuzz then, a small chance we are sending a long header packet

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -953,37 +1377,158 @@ void FuzzHandshake(
...
+    _In_ uint64_t StartTimeMs
+    )
+{
+    PacketParams PacketParams = {
+        sizeof(uint64_t),
+        sizeof(uint64_t),
+        0,
+        1
+    };
+    PacketParams.PacketType = QUIC_INITIAL_V1;
```
guhetier: nit: I would avoid mixing an init list and direct member initialization, it makes it hard to know what is initialized.

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -953,37 +1377,158 @@ void FuzzHandshake(
...
+    PacketParams.FrameTypes[0] = QUIC_FRAME_CRYPTO;
+    PacketParams.Mode = 2;
+    GetRandomBytes(sizeof(uint64_t), &PacketParams.SourceCid);
+    memcpy(&CurrSrcCid, PacketParams.SourceCid, sizeof(uint64_t)); // Save for receive path
+
+    TlsContext ClientContext;
+    ClientContext.CreateContext(PacketParams.SourceCid);
+    CXPLAT_FRE_ASSERT(ClientContext.ProcessData() & CXPLAT_TLS_RESULT_DATA);
+
+    if (CompleteHandshake(Binding, Route, StartTimeMs, &PacketParams, &ClientContext)) {
```
guhetier: nit: Consider returning early instead of nesting if blocks.
It will make the flow easier to understand: if we didn't get to complete the handshake, we do
nothing. If we don't get 1 RTT write keys, we do nothing. Then, if we have all we need, we get to
the actual body

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -853,36 +1293,22 @@ void FuzzInitial(
...
-    TlsContext ClientContext;
-    ClientContext.CreateContext(PacketParams.SourceCid);
-    CXPLAT_FRE_ASSERT(ClientContext.ProcessData() & CXPLAT_TLS_RESULT_DATA);
-
     //
     // Keep sending the packet until we receive a response or the time runs out.
     //
     do {
-        BuildAndSendPackets(Binding, Route, &PacketParams, &ClientContext, false); // Don't fuzz this one
+        BuildAndSendLongHeaderPackets(Binding, Route, PacketParams, ClientContext, false);
```
guhetier, line 1387: Please keep / improve the comment here.
Even better if you clarify which packet this is (I assume this is meant to be the INITIAL?)

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -953,37 +1377,158 @@ void FuzzHandshake(
...
+
+    if (CompleteHandshake(Binding, Route, StartTimeMs, &PacketParams, &ClientContext)) {
+        //
+        // Complete handshake by sending final handshake packets
+        //
+        PacketParams.PacketType = QUIC_HANDSHAKE_V1;
+        PacketParams.NumFrames = 1;
+        PacketParams.FrameTypes[0] = QUIC_FRAME_CRYPTO;
+        PacketParams.NumPackets = 1;
+        BuildAndSendLongHeaderPackets(Binding, Route, &PacketParams, &ClientContext, false);
```
guhetier: I am slightly confused by this, why isn't `CompleteHandshake` taking care of it?

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -519,6 +519,215 @@ void WriteAckFrame(
...
+    uint64_t StreamType = GetRandom<uint8_t>(2) == 0 ? 0x00 : 0x02; // Bidi or Uni
+    uint64_t StreamIndex = GetRandom<uint8_t>(4); // 0-3
+    uint64_t StreamId = (StreamIndex * 4) | StreamType; // Generate stream ID
+    uint64_t StreamOffset = GetRandom<uint8_t>(2) == 0 ? 0 : GetRandom<uint16_t>(1000);
+    uint16_t DataLength = GetRandom<uint8_t>(200) + 10; // 10-210 bytes of data
+
+    // Use the provided frame type (which encodes FIN, LEN, and OFF bits)
+    bool HasOffset = (FrameType & 0x04) != 0;
+    bool HasLength = (FrameType & 0x02) != 0;
+    bool HasFin = (FrameType & 0x01) != 0;
```
guhetier: Dead variable.

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -519,6 +519,215 @@ void WriteAckFrame(
             Buffer));
 }
 
+void WriteHandshakeDoneFrame(
+    _Inout_ uint16_t* Offset,
+    _In_ uint16_t BufferLength,
+    _Out_writes_to_(BufferLength, *Offset) uint8_t* Buffer
+    )
+{
+    CXPLAT_FRE_ASSERT(*Offset < BufferLength);
```
guhetier: nit: The AI review pointed out that this assert won't prevent writing out of bounds.
While the assert is true and makes sense, a bit more validation could help avoid memory corruption issues (at least asserting the final offset is still smaller than `BufferLength`, if you assume the caller knowns what it does.

Same for all Write* functions below.

src/tools/recvfuzz/recvfuzz.cpp
gaurav2699: Mainly because this step is the packet we fuzz in FuzzHandshake, but we need to do this step in 1rtt to do the remaining operations, I tried to keep the fuzzing stuff outside the completeHandshake function but I agree it would make some sense to keep this step there

src/tools/recvfuzz/recvfuzz.cpp
guhetier: nit: can be removed.
```suggestion
```

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -768,6 +1079,161 @@ void BuildAndSendPackets(
...
+    _In_ bool FuzzPacket = true
+    )
+{
+    const uint16_t DatagramLength = 1200; // Standard datagram size
+    CXPLAT_SEND_CONFIG SendConfig = { Route, DatagramLength, CXPLAT_ECN_NON_ECT, 0, CXPLAT_DSCP_CS0 };
+    CXPLAT_SEND_DATA* SendData = CxPlatSendDataAlloc(Binding, &SendConfig);
+    CXPLAT_FRE_ASSERT(SendData != nullptr);
+
+    uint8_t numPacketsSent = 0;
+    while (!CxPlatSendDataIsFull(SendData) && numPacketsSent <= PacketParams->NumPackets) {
```
guhetier: Shouldn't this be `<`? It doesn't matter much but this would send one more packet than planned.

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -768,6 +1079,161 @@ void BuildAndSendPackets(
...
+        QUIC_BUFFER* SendBuffer =
+            CxPlatSendDataAllocBuffer(SendData, DatagramLength);
+        CXPLAT_FRE_ASSERT(SendBuffer != nullptr);
+        CxPlatZeroMemory(SendBuffer->Buffer, DatagramLength);
+
+        uint16_t PacketLength, HeaderLength;
```
guhetier: In general, please avoid declaring multiple variables in one statement and systematically default initialize variables.

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -555,12 +789,90 @@ void WriteCryptoFrame(
...
+    for (int i = 0; i < PacketParams->NumFrames; i++) {
+        *PayloadLength += GetRandom<uint8_t>(64); // Random padding
+
+        if (PacketParams->FrameTypes[i] == QUIC_FRAME_ACK) {
+            WriteAckFrame(
+                PacketParams->LargestAcknowledge,
+                PayloadLength,
+                BufferSize,
+                Buffer);
+
```
guhetier:
```suggestion
```

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -517,6 +518,238 @@ void WriteAckFrame(
...
+        ActualDataLength = DataLength > 5 ? DataLength - GetRandom<uint8_t>(5) - 1 : 0;
+    } else if (LengthFuzz == 1) {
+        // 5% chance: write more bytes than promised (if space allows)
+        uint16_t Extra = GetRandom<uint8_t>(10) + 1;
+        if (*Offset + DataLength + Extra <= BufferLength) {
+            ActualDataLength = DataLength + Extra;
+        }
+    }
+    
+    CXPLAT_FRE_ASSERT(*Offset + ActualDataLength <= BufferLength);
```
guhetier: I'm not sure I understand how this assert holds. Is the caller supposed to have enough memory allocated for all maximum random values?

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -517,6 +518,238 @@ void WriteAckFrame(
...
+
+void WriteHandshakeDoneFrame(
+    _Inout_ uint16_t* Offset,
+    _In_ uint16_t BufferLength,
+    _Out_writes_to_(BufferLength, *Offset) uint8_t* Buffer
+    )
+{
+    CXPLAT_FRE_ASSERT(*Offset + 1 <= BufferLength);
+    Buffer[(*Offset)++] = QUIC_FRAME_HANDSHAKE_DONE;
+    CXPLAT_FRE_ASSERT(*Offset <= BufferLength);
```
guhetier: Why this assert? This is demonstrably true given the assert above. Same comment for other functions.

src/tools/recvfuzz/recvfuzz.cpp
gaurav2699: You are right. We can clamp to available buffer space which would be better for fuzzing scenario.

anrossi:
@copilot+gpt-5.3-codex Could you help review this? Check the field ordering of the generated frames is consistent with RFC 9000 and RFC 9001, if you have RFC 9000 and 9001 in your training data.

anrossi:
@copilot+gpt-5.3-codex Could you help review this? Check the field ordering of the generated frames is consistent with RFC 9000 and RFC 9001, if you have RFC 9000 and 9001 in your training data.

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -519,6 +518,233 @@ void WriteAckFrame(
...
+    _In_ uint16_t BufferLength,
+    _Out_writes_to_(BufferLength, *Offset) uint8_t* Buffer
+    )
+{
+    CXPLAT_FRE_ASSERT(*Offset + 1 <= BufferLength);
+    Buffer[(*Offset)++] = QUIC_FRAME_RESET_STREAM;
+    
+    // Stream ID
+    uint64_t StreamId = GetRandom<uint8_t>(4) * 4; // Client-initiated bidi
+    CXPLAT_FRE_ASSERT(*Offset + QuicVarIntSize(StreamId) <= BufferLength);
```
guhetier: Same issue as earlier for this ASSERT and many of the following asserts.

You can't have an assert in the fuzzer code unless you can prove it is true (and it is there only to catch a future dev mistake). I don't see any code that ensure `BufferLength` is large enough.

The function must fail (silently or not) and leave the buffer in a state we are ok feeding to MsQuic (which means pretty much anything as long as it is intentional since we are fuzzing).

gaurav2699: I am now returning early when the buffer size is not enough, is that fine

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -504,19 +502,246 @@ void WriteAckFrame(
     _Out_writes_to_(BufferLength, *Offset) uint8_t* Buffer
     )
 {
+    if (*Offset >= BufferLength) return;
```
guhetier: nit: Please follow the repository style

```
if (...) {
    return;
}
```

src/tools/recvfuzz/recvfuzz.cpp
guhetier: That's a reasonable approach as long as the buffer is large enough in general so that we don't always skip most of the data.

src/tools/recvfuzz/recvfuzz.cpp
```diff
@@ -953,39 +1403,170 @@ void FuzzHandshake(
...
+        PacketParams.NumFrames = 1;
+        PacketParams.FrameTypes[0] = QUIC_FRAME_HANDSHAKE_DONE;
+        PacketParams.NumPackets = HandshakeDoneCount;
+        BuildAndSendShortHeaderPackets(Binding, Route, &PacketParams, &ClientContext, false);
+    }
+
+    //
+    // Send fuzzed 1-RTT packets with randomized frame types
+    //
+    PacketParams.NumFrames = 1;
```
guhetier, line 1585: We should inject packets with multiple frames.
It doesn't need to be done in this PR, but if we keep it for later, please create a task for it.

guhetier (commented):
Looks good to me once the style is fixed.

src/tools/recvfuzz/recvfuzz.cpp
gaurav2699: https://github.com/microsoft/msquic/issues/5907

PR #5739, opened 2026-01-23
Treat SEC_E_INTERNAL_ERROR as SEC_E_NO_CREDENTIAL when querying certi…

guhetier:
## Description

`SECPKG_ATTR_REMOTE_CERTIFICATES` can result in a `SEC_E_INTERNAL_ERROR` when the certificate is not found.
Schannel intentionally preserve this behavior to keep their error behavior unchanged.

When this happens, MsQuic would close the connection immediately, even when TLS was configured to require client certificate validation but defer the validation to the app (which could let the app accept a connection even if no certificate is provided).

This PR treats `SEC_E_INTERNAL_ERROR` the same way as `SECPKG_ATTR_REMOTE_CERTIFICATES`.

## Testing

CI. Partner validation pending.

## Documentation

N/A

### DIFF
--- a/src/platform/tls_schannel.c
+++ b/src/platform/tls_schannel.c
@@ -2311,10 +2311,14 @@ CxPlatTlsWriteDataToSchannel(
             if ((SecStatus == SEC_E_NO_CREDENTIALS || SecStatus == SEC_E_INTERNAL_ERROR) &&
                 (TlsContext->SecConfig->Flags & QUIC_CREDENTIAL_FLAG_DEFER_CERTIFICATE_VALIDATION)) {
                 //
-                // Ignore this case.
+                // Certificate validation is being deferred to the application:
+                // indicate no certificate is present but let the application decide if this is a
+                // failure.
+                // SEC_E_INTERNAL_ERROR can be returned on SECPKG_ATTR_REMOTE_CERT_CONTEXT if no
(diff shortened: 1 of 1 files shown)


src/platform/tls_schannel.c
```diff
@@ -2308,7 +2308,7 @@ CxPlatTlsWriteDataToSchannel(
                         QUIC_CERT_BLOB_CONTEXT : QUIC_CERT_BLOB_NONE;
 #endif
             }
-            if (SecStatus == SEC_E_NO_CREDENTIALS &&
+            if ((SecStatus == SEC_E_NO_CREDENTIALS || SecStatus == SEC_E_INTERNAL_ERROR) &&
                 (TlsContext->SecConfig->Flags & QUIC_CREDENTIAL_FLAG_DEFER_CERTIFICATE_VALIDATION)) {
                 //
                 // Ignore this case.
```
mtfriesen: Could you extend this comment with an explanation similar to the PR description - we're ignoring this case to allow the app to decide what to do? (Maybe it's obvious, in which case, the current tautological comment should just be removed.)

guhetier: Improved the comment and on a second thought, normalized the error code.
The function needs to be split/refactored (the control flow is really hard to follow), in the meantime, clarifying the intent won't hurt.

PR #5747, opened 2026-01-26
Fix false hash table clashes in raw datapath creation 

### LINKED ISSUE #5718: [CI - FAILURE]:  / AppData_WithSendArgs1.SendQtip_214 (msquictest.exe)
https://github.com/microsoft/msquic/actions/runs/21005317942

ProjectsByJackHe:
## Description

Since the XDP datapath is initialized on a "best-effort" basis, it will fall back to normal OS sockets whenever there is an issue with creating the XDP datapath. 

One check that it does is whether or not we previously have already created the same datapath on the same IP/port for local/remote. The assumption is, we already hooked into XDP and it's already snooping on the same IP/port, no need to do it again. It will fall back to using normal OS sockets. 

This is misleading because when using QTIP, it is possible that a client machine has connections bound to the same ephemeral port, except the difference is it's UDP/TCP. Previously, this was not possible as MsQuic is using either ALL QTIP, or None at all. 

Fixes #5718

image

NOTE:
On any machine, either ALL listeners have QTIP enabled or all of them have them disabled. 
This is controlled via the GLOBAL MsQuic settings. 

It's only client connections that can vary this on a per-connection basis with local settings. 

## Testing

Logs from CI. 

## Documentation

QTIP.md


src/platform/datapath_raw_socket.c
```diff
@@ -1138,15 +1138,32 @@ CxPlatTryAddSocket(
...
+    // and determine if this new datapath is redundant or not.
+    //
     Entry = CxPlatHashtableLookup(&Pool->Sockets, Socket->LocalAddress.Ipv4.sin_port, &Context);
     while (Entry != NULL) {
         CXPLAT_SOCKET_RAW* Temp = CXPLAT_CONTAINING_RECORD(Entry, CXPLAT_SOCKET_RAW, Entry);
         if (CxPlatSocketCompare(Temp, &Socket->LocalAddress, &Socket->RemoteAddress)) {
+            if (Temp->ReserveAuxTcpSock != Socket->ReserveAuxTcpSock) {
+                //
+                // We previously created a datapath on the same local IP and port, but in fact
+                // not the same QTIP settings. This is not considered a "clash."
```
mtfriesen: I'm new to this QTIP code - could you share a specific example (maybe snippets from the logs) to illustrate concrete values for what all these sockets are bound/connected to?

src/platform/datapath_raw_socket.c
```diff
@@ -1138,15 +1138,32 @@ CxPlatTryAddSocket(
...
     }
 
+    //
+    // Look up all previously established datapaths bound to a specific port,
+    // and determine if this new datapath is redundant or not.
+    //
     Entry = CxPlatHashtableLookup(&Pool->Sockets, Socket->LocalAddress.Ipv4.sin_port, &Context);
     while (Entry != NULL) {
         CXPLAT_SOCKET_RAW* Temp = CXPLAT_CONTAINING_RECORD(Entry, CXPLAT_SOCKET_RAW, Entry);
         if (CxPlatSocketCompare(Temp, &Socket->LocalAddress, &Socket->RemoteAddress)) {
```
mtfriesen: For other code depending on `CxPlatSocketCompare`, will behavior still be deterministic if we start allowing duplicates within the hash table?

src/platform/datapath_raw_socket.c
```diff
@@ -1138,15 +1138,32 @@ CxPlatTryAddSocket(
...
+            if (Temp->ReserveAuxTcpSock != Socket->ReserveAuxTcpSock) {
+                //
+                // We previously created a datapath on the same local IP and port, but in fact
+                // not the same QTIP settings. This is not considered a "clash."
+                //
+                continue;
+            }
+
+            //
+            // Redundant datapath. We already created one with the same IP port for local/remote AND QTIP settings.
```
mtfriesen: This could be an ignorant question, but what are "QTIP settings"? I have an understanding of how TCP/UDP local and remote address/port settings are defined and used, but I'm not sure what we're comparing when we say "QTIP settings" are the same or not the same.

src/platform/datapath_raw_socket.c
guhetier:
```suggestion
                // not the same QTIP settings. This is not considered a collision.
```

or

```suggestion
                // not the same QTIP settings. This is not incompatible.
```

src/platform/datapath_raw_socket.c
```diff
@@ -1138,15 +1138,32 @@ CxPlatTryAddSocket(
...
     Entry = CxPlatHashtableLookup(&Pool->Sockets, Socket->LocalAddress.Ipv4.sin_port, &Context);
     while (Entry != NULL) {
         CXPLAT_SOCKET_RAW* Temp = CXPLAT_CONTAINING_RECORD(Entry, CXPLAT_SOCKET_RAW, Entry);
         if (CxPlatSocketCompare(Temp, &Socket->LocalAddress, &Socket->RemoteAddress)) {
+            if (Temp->ReserveAuxTcpSock != Socket->ReserveAuxTcpSock) {
+                //
+                // We previously created a datapath on the same local IP and port, but in fact
+                // not the same QTIP settings. This is not considered a "clash."
+                //
+                continue;
```
guhetier: Can you explain why this is not a collision?
I don't get why having two sockets with the same port, local and remote would be ok if one is using QTIP and not the other.

mtfriesen: +1 - the existing model is implicitly defined only by the code itself, yet reasonably intuitive to networking generalists.

Even after looking at this for a while, I couldn't infer what the "QTIP settings" are.

src/platform/datapath_raw_socket.c
ProjectsByJackHe: Sure. QTIP settings revers to whether or not a connection using QTIP, which translates to whether or not the underlying CXPLAT socket backing the connection is using QTIP. If QTIP is used (i.e. ReserveAuxTcpSock = TRUE) the underlying datapath backing that socket has XDP snooping for TCP packets. 
Otherwise, XDP will be snooping for UDP packets. 

That's why if we observe a "clash" with same local IP/port, but different QTIP settings, they are snooping for different packets.

src/platform/datapath_raw_socket.c
ProjectsByJackHe: Sure. 

When QTIP is enabled, we pick a random TCP port reserved to us by the OS for the CXPLAT socket. 
Otherwise, we pick a random UDP port reserved to us by the OS. 

Now, it is possible, in our tests where we create multiple QTIP and non-QTIP connections, that eventually the OS will reserve for us a TCP port that has the same value as the random UDP port that previously got assigned. 

This is because, client connections only reserve 1 OS socket (TCP or UDP depending on QTIP settings). 
Listeners, on the other hand, reserve 2 (both UDP and TCP). 
So in the case of listeners, it will be impossible for the OS to assign a random UDP or TCP port that clashes with  ports assigned to a socket backing a listener.

src/platform/datapath_raw_socket.c
ProjectsByJackHe: resolved offline.

src/platform/datapath_raw_socket.c
ProjectsByJackHe: resolved offline

src/core/binding.c
```diff
@@ -312,6 +312,18 @@ QuicBindingGetRemoteAddress(
 #endif
 }
 
+//
+// Queries the QTIP settings of the binding.
+//
+_IRQL_requires_max_(DISPATCH_LEVEL)
+BOOLEAN
+QuicBindingGetQtipSettings(
```
mtfriesen: Broadly speaking, across this PR, the phrase "QTIP settings" and this function returning a BOOLEAN strike me as both intuitively confusing (ambiguous) and a grammatical contradiction. 

If you're checking if QTIP is "enabled", just say "enabled".

src/core/library.c
```diff
@@ -2224,23 +2225,23 @@ QuicLibraryLookupBinding(
...
             }
 
         } else {
             //
             // For server (unconnected/listening) bindings we always use wildcard
-            // addresses, so we simply need to match on local port.
+            // addresses, so we simply need to match on local port and QTIP settings.
             //
-            if (QuicAddrGetPort(&BindingLocalAddr) == QuicAddrGetPort(LocalAddress)) {
+            if (QuicAddrGetPort(&BindingLocalAddr) == QuicAddrGetPort(LocalAddress) && QuicBindingGetQtipSettings(Binding) == UseQtip) {
```
mtfriesen: Nit: 136 characters exceeds our line style. I'm not sure what the magic number is for MsQuic, but this stands out.

src/core/library.c
```diff
@@ -2270,6 +2271,7 @@ QuicLibraryGetBinding(
     const BOOLEAN ShareBinding = !!(UdpConfig->Flags & CXPLAT_SOCKET_FLAG_SHARE);
     const BOOLEAN ServerOwned = !!(UdpConfig->Flags & CXPLAT_SOCKET_SERVER_OWNED);
     const BOOLEAN Partitioned = !!(UdpConfig->Flags & CXPLAT_SOCKET_FLAG_PARTITIONED);
+    const BOOLEAN UseQtip = !!(UdpConfig->Flags & CXPLAT_SOCKET_FLAG_QTIP);
```
mtfriesen: Perhaps nitpicky and pedantic, but perhaps important to the confusion here: does this flag imply QTIP will be used (that is, traffic will go over QTIP) or that QTIP _may_ be used? If the latter, I'd suggest the same word as above: "EnableQtip".

src/core/binding.c
guhetier: I second this, I was writing a similar comment on the datapath code. "enabled" or "supported" would be clearer depending on what it means.

src/core/binding.c
```diff
@@ -312,6 +312,18 @@ QuicBindingGetRemoteAddress(
...
+    _In_ QUIC_BINDING* Binding
```
guhetier: Please try to enforce some const-correctness. If the function is not meant to modify the binding, take it as a pointer to const.

src/platform/datapath_xplat.c
```diff
@@ -225,6 +225,16 @@ CxPlatSocketDelete(
...
 }
 
+_IRQL_requires_max_(DISPATCH_LEVEL)
+BOOLEAN
+CxPlatSocketGetQtipSettings(
+    _In_ CXPLAT_SOCKET* Socket
+    )
+{
+    CXPLAT_DBG_ASSERT(Socket != NULL);
+    return Socket->ReserveAuxTcpSock;
```
guhetier, line 235: nit: We should rename this.
`ReserveAuxTcpSock` is hard to understand, and as far as I can tell, is equivalent to using QTip, which is a documented feature.

src/core/library.c
```diff
@@ -2224,23 +2225,23 @@ QuicLibraryLookupBinding(
...
+                if (QuicAddrCompare(RemoteAddress, &BindingRemoteAddr) && QuicBindingGetQtipSettings(Binding) == UseQtip) {
                     return Binding;
                 }
             }
 
         } else {
             //
             // For server (unconnected/listening) bindings we always use wildcard
-            // addresses, so we simply need to match on local port.
+            // addresses, so we simply need to match on local port and QTIP settings.
```
guhetier: For a server binding, when QTIP is enabled, we listen on both QTIP and UDP traffic. And if QTIP is disabled, we listen only on UDP traffic.

Is that correct? If yes, it sounds to me that a UDP only listener and a UDP+QTIP listener on the same port / local address should conflict, they both try to listen to the same UDP traffic.

src/core/library.c
```diff
@@ -2202,7 +2202,8 @@ QuicLibraryLookupBinding(
     _In_ QUIC_COMPARTMENT_ID CompartmentId,
 #endif
     _In_ const QUIC_ADDR* LocalAddress,
-    _In_opt_ const QUIC_ADDR* RemoteAddress
+    _In_opt_ const QUIC_ADDR* RemoteAddress,
+    _In_ const BOOLEAN UseQtip
```
guhetier: I am not a fan of the multiplication of parameters on this function (especially because I already can see an "UseRdma" bool being added soon.
Maybe it is time to wrap them in a struct. That would also allow to document them better (IMO, the binding lookup and collision logic could use more clarity in what it intends to achieve)

src/core/library.c
guhetier: I read your updated PR description, which clarifies that either all listeners have QTIP enabled or none.
I think this should be clarified in a comment here, as it matters.
And I think we should then *not* check the QTIP state for listeners:
- we expect it to always match per the above
- if we one day allow different QTIP enablement state for different listeners, the code would be correct

src/platform/datapath_xplat.c
mtfriesen: Agreed, this seems an over-abstraction. That said, I'm waiting to converge on the proposed model before renaming fields to reflect it.

src/core/library.c
ProjectsByJackHe: That's fair, I added this check to be extra careful as without it then it would be incorrect if we one day broke the model of how listeners use QTIP. It's this implicit assumption that I was worried about. 

We can certainly call it out in a comment instead of the check though, but we will have to remember to update this again if one day we make it so listeners can have different QTIP settings.

src/core/library.c
guhetier: Please add some details to the logic in this function, especially related to Qtip.
The fact we consider server-side bindings to match independently from the Qtip enablement state, but check it matches for client-side binding is not something a reader can understand by himself.

ProjectsByJackHe: Yes, added.

src/core/library.c
ProjectsByJackHe: Sounds good, wrapped it up in a struct.

src/core/library.c
ProjectsByJackHe: yeah, I see what you're saying. This binding lookup function can be used for both client connections and server listeners, where the use case is pretty different. Client connections must use QTIP if this flag is there, Server listeners MAY use QTIP in addition to normal XDP. I can do the rename.

src/platform/datapath_xplat.c
ProjectsByJackHe: Sure. We can decide on a best name after the overall approach for solving this problem has been decided.

ProjectsByJackHe:
> This will allow bindings with the same 4-tupple to coexist as long as they use a different transport. What does it mean for the upper layers? Is there any risk of this allowing to confuse the stack by sending a mix a UDP and QTIP packets with the same 4-tuple (in everything related to QUIC_PATH, path validation or client migration for instance)?
> 
> At a minimum, we need to update the documentation to clarify the app can't rely on the 4-tuple only to distinguish connections when QTIP is in use, even with port sharing disabled.

Updated QTIP.md

docs/QTIP.md
```diff
@@ -24,7 +24,28 @@ This setting can be overridden per client connection, allowing you to create som
 > [!IMPORTANT]
 > Crucial information necessary for users to succeed.
 
-Using QTIP will initialize a TCP socket and attempt to bind to your listener's local address. This is to reserve a TCP port for your listener to ensure
-XDP does not steal any TCP traffic from your other processes later. That also means you need to ensure no other processes are listening on the same TCP port as your listener's local address prior
-to starting your listener, otherwise the OS will throw a socket access denied / address in use error,
-and your listener will fail to initialize.
+**Client connections with different QTIP enablements CAN exist on the same local port.**
+
+MsQuic over vanilla XDP relies on the OS to assign us ephemeral UDP ports to reserve them and configure XDP to snoop UDP traffic on that port.
```
mtfriesen: We should avoid idioms like "vanilla" if words like "plain" carry the same meaning, both as a matter of style and because not all readers/writers have English as their native language.

docs/QTIP.md
```diff
@@ -24,7 +24,28 @@ This setting can be overridden per client connection, allowing you to create som
...
+
+MsQuic over QTIP XDP relies on the OS to assign us ephemeral TCP ports to reserve them and configure XDP to snoop TCP traffic on that port.
```
mtfriesen: How much of this is intrinsic to "XDP" vs a general port pooling problem?

docs/QTIP.md
mtfriesen: We should avoid "us" in public-facing documentation and make clear who each actor is.

docs/QTIP.md
```diff
@@ -24,7 +24,28 @@ This setting can be overridden per client connection, allowing you to create som
...
+
+
+**Listeners with different QTIP enablements shall NOT be able to exist on the same local port.**
+
+A QTIP-enabled listener will reserve both UDP/TCP ports equal to the local port of the listener
+and configure XDP to snoop UDP/TCP packets on that local port. A non-QTIP listener will just reserve a UDP port
+and have XDP snoop UDP packets on that port. To avoid conflicting traffic, we disallow 2 listeners with the same
+local port but different QTIP enablements to exist.
+
+In general, listeners should either be ALL QTIP enabled, or none of them are. Since QTIP enablement is controlled via the global setting, You should NOT alter the global settings for QTIP once you start a listener to avoid other listeners
```
mtfriesen: > You should NOT alter the global settings

This should be enforced at the API level. Similar to the comment about "us" above, "you" should also be avoided and actors should be specified.

docs/QTIP.md
```diff
@@ -24,7 +24,28 @@ This setting can be overridden per client connection, allowing you to create som
...
+But since we can create many client connections with different QTIP enablements, sometimes the OS assigns us
+the same TCP and UDP port. This isn't necessarily a problem but may be confusing when debugging logs and
+you observe 2 client connections on the same local port. It is important to differentiate the transport.
+
+
+**Listeners with different QTIP enablements shall NOT be able to exist on the same local port.**
+
+A QTIP-enabled listener will reserve both UDP/TCP ports equal to the local port of the listener
+and configure XDP to snoop UDP/TCP packets on that local port. A non-QTIP listener will just reserve a UDP port
+and have XDP snoop UDP packets on that port. To avoid conflicting traffic, we disallow 2 listeners with the same
```
mtfriesen: We should distinguish between snooping and stealing: snooping passively observes without modification; stealing absorbs and prevents others from receiving.

I believe the problem here is stealing, not snooping.

docs/QTIP.md
```diff
@@ -24,7 +24,28 @@ This setting can be overridden per client connection, allowing you to create som
 > [!IMPORTANT]
 > Crucial information necessary for users to succeed.
 
-Using QTIP will initialize a TCP socket and attempt to bind to your listener's local address. This is to reserve a TCP port for your listener to ensure
```
mtfriesen, line 27: Is this removed comment no longer useful? The notion of conflicting ports with other processes on the system seems potentially still relevant.

docs/QTIP.md
ProjectsByJackHe: In this case, it's like 99% XDP. Normal user mode apps that just uses the OS networking stack never need to worry about port collisions between TCP/UDP sockets as the OS handles proper packet de-muxing.

docs/QTIP.md
ProjectsByJackHe: Had an offline discussion about this, and the conclusion is actually we may still want to support multiple listeners, different QTIP enablement scenario in the future. I will update the docs.

docs/QTIP.md
ProjectsByJackHe: This seemed implicit based on the rest of the doc, but I can add this back.

src/core/binding.c
ProjectsByJackHe: Updated the naming.

docs/QTIP.md
ProjectsByJackHe: Updated

docs/QTIP.md
```diff
@@ -24,7 +24,25 @@ This setting can be overridden per client connection, allowing you to create som
...
+
+**Client connections with different QTIP enablements CAN exist on the same local port.**
+
+MsQuic connections over UDP XDP creates an OS UDP socket only and relies on the OS to assign the app an ephemeral UDP port to reserve them and configure XDP to snoop UDP traffic on that port.
+
+MsQuic connections over QTIP XDP creates an OS TCP socket only and relies on the OS to assign the app an ephemeral TCP port to reserve them and configure XDP to snoop TCP traffic on that port.
+
+But since apps can create many client connections with different QTIP enablements, sometimes the OS assigns
+the same TCP and UDP port. This isn't necessarily a problem but may be confusing when debugging logs. Apps should
+expect in rare cases some client connections will have the same local port when using different QTIP enablements.
```
guhetier: More than "This isn't necessarily a problem..." (which is true but not that helpful, I'd replace the rest of the sentence by a recommendation:
"If using client connections with and without QTIP enabled simultaneously, the application should not assume 4-tupples uniquely identify a connection and also track the QTIP state."

ProjectsByJackHe: Yup, that sounds good. I'll push an update.

docs/QTIP.md
```diff
@@ -24,7 +24,24 @@ This setting can be overridden per client connection, allowing you to create som
 > [!IMPORTANT]
 > Crucial information necessary for users to succeed.
 
-Using QTIP will initialize a TCP socket and attempt to bind to your listener's local address. This is to reserve a TCP port for your listener to ensure
-XDP does not steal any TCP traffic from your other processes later. That also means you need to ensure no other processes are listening on the same TCP port as your listener's local address prior
+Listeners with QTIP enabled will initialize a TCP and UDP socket and attempt to bind to your listener's local address. This is to reserve a TCP/UDP port for your listener to ensure
+XDP does not steal any traffic from your other processes later. That also means you need to ensure no other processes are listening on the same port as your listener's local address prior
```
mtfriesen:
```suggestion
XDP does not steal any traffic from other sockets later. That also means you need to ensure no other processes are listening on the same port as your listener's local address prior
```

docs/QTIP.md
```diff
@@ -24,7 +24,24 @@ This setting can be overridden per client connection, allowing you to create som
...
 to starting your listener, otherwise the OS will throw a socket access denied / address in use error,
 and your listener will fail to initialize.
+
+**Client connections with different QTIP enablements CAN exist on the same local port.**
+
+MsQuic connections over UDP XDP creates an OS UDP socket only and relies on the OS to assign the app an ephemeral UDP port to reserve them and configure XDP to snoop UDP traffic on that port.
```
mtfriesen: This is a run-on sentence - it needs some punctuation.

mtfriesen: Same for the TCP case below.

docs/QTIP.md
```diff
@@ -24,7 +24,24 @@ This setting can be overridden per client connection, allowing you to create som
...
+
+MsQuic connections over QTIP XDP creates an OS TCP socket only and relies on the OS to assign the app an ephemeral TCP port to reserve them and configure XDP to snoop TCP traffic on that port.
+
+But since apps can create many client connections with different QTIP enablements, sometimes the OS assigns
```
mtfriesen:
```suggestion
Since apps can create many client connections with different QTIP enablements, sometimes the OS assigns
```

docs/QTIP.md
```diff
@@ -24,7 +24,24 @@ This setting can be overridden per client connection, allowing you to create som
...
+the same TCP and UDP port. If using client connections with and without QTIP enabled simultaneously, the application should not assume 4-tuples uniquely identify a connection and also track the QTIP state.
```
mtfriesen:
```suggestion
the same TCP and UDP port number. If using client connections with and without QTIP enabled simultaneously, the application should not assume 4-tuples uniquely identify a connection and also track the QTIP state.
```

mtfriesen: 4-tuples or 5-tuples?

ProjectsByJackHe: apps should not assume 4-tuples are enough if QTIP is enabled.

src/core/library.c
```diff
@@ -2224,21 +2224,27 @@ QuicLibraryLookupBinding(
         if (Binding->Connected) {
             //
             // For client/connected bindings we need to match on both local and
-            // remote addresses/ports.
+            // remote addresses/ports, along with QTIP settings. This is because if QTIP is enabled, the OS
+            // may choose to assign us the same ephemeral TCP and UDP local port. We allow client connections to use
```
mtfriesen: Can we rewrite this to make the intuition clearer: we allow two client sockets to "share" a 4-tuple as long as the full 5-tuple (that is, including UDP vs TCP) is unique?

src/core/library.c
```diff
@@ -2224,21 +2224,27 @@ QuicLibraryLookupBinding(
...
+            // the same local port IF they have different QTIP settings. So any binding a client connection shares
```
mtfriesen:
```suggestion
            // the same local port IF they have different QTIP settings. Any binding a client connection shares
```

src/core/library.c
ProjectsByJackHe: Yes. The core intuition maybe could be phrased as: "we need to match on the 5-tuple, to include transport type for client socket bindings, because client connections cannot share the binding if the transport does not match."

docs/QTIP.md
ProjectsByJackHe: Added punctuation.

src/core/binding.c
```diff
@@ -312,6 +312,18 @@ QuicBindingGetRemoteAddress(
...
+//
+// Queries the QTIP settings of the binding.
+//
+_IRQL_requires_max_(DISPATCH_LEVEL)
+BOOLEAN
+QuicBindingGetQtipEnabled(
+    _In_ const QUIC_BINDING* Binding
+    )
+{
+    return CxPlatSocketGetQtipSettings(Binding->Socket);
```
guhetier: `CxPlatSocketGetQtipSettings` should also be renamed to `CxPlatSocketGetQtipEnabled`

src/core/binding.c
```diff
@@ -312,6 +312,18 @@ QuicBindingGetRemoteAddress(
 #endif
 }
 
+//
+// Queries the QTIP settings of the binding.
```
guhetier: Some "QTIP settings" still remaning.
We should also not duplicate the comment between declaration and definition, or they'll end up out of sync

src/core/library.c
```diff
@@ -2224,21 +2224,25 @@ QuicLibraryLookupBinding(
...
+                    QuicBindingGetQtipEnabled(Binding) == EnableQtip) {
                     return Binding;
                 }
             }
 
         } else {
             //
             // For server (unconnected/listening) bindings we always use wildcard
-            // addresses, so we simply need to match on local port.
+            // addresses, so we simply need to match on the local port. We need not consider the
```
guhetier, line 2243: The code structure makes those comment confusing:
- above, we say we use the QTIP enable state in the match because client connection can't share the binding
- here, we say we *don't* use the QTIP enable state in the match because listeners can't share the binding

The difference being that code in the parent will do extra validation and either create a binding or fail.
I am not sure how to rephrase them better though...

src/core/library.c
```diff
@@ -2384,26 +2384,30 @@ QuicLibraryGetBinding(
         // tuple, so we need to do collision detection based on the whole
         // 4-tuple.
         //
+        CXPLAT_UDP_CONFIG Config = {0};
+        #ifdef QUIC_COMPARTMENT_ID
```
guhetier: nit: preprocessor code should not be indented

src/core/library.c
```diff
@@ -2300,19 +2305,14 @@ QuicLibraryGetBinding(
...
-            UdpConfig->LocalAddress,
-            UdpConfig->RemoteAddress);
+        QuicLibraryLookupBinding(UdpConfig);
     if (Binding != NULL) {
         if (!ShareBinding || Binding->Exclusive ||
             (ServerOwned != Binding->ServerOwned) ||
             (Partitioned != Binding->Partitioned) ||
-            (Partitioned && UdpConfig->PartitionIndex != Binding->PartitionIndex)) {
+            (Partitioned && UdpConfig->PartitionIndex != Binding->PartitionIndex) ||
+            (!Binding->Connected && QuicBindingGetQtipEnabled(Binding) != EnableQtip)) {
```
guhetier, line 2315: The condition is starting to be pretty big, we could extract it to its own function `CanUseBinding` (which would allow to comment on each decision)

src/platform/datapath_raw.h
```diff
@@ -334,21 +334,29 @@ BOOLEAN
...
+        // always give a match if the listener Socket has enabled QTIP (can accept both UDP/QTIP traffic).
+        // Otherwise, we want to NOT match if UseQtip is TRUE but the listener socket does not enable QTIP.
+        //
+        return Socket->ReserveAuxTcpSock || !UseQtip;
     }
 
     //
-    // Make sure the local IP matches and the full remote address matches.
+    // For a client socket,
+    // Make sure the local IP matches and the full remote address matches along with QTIP settings.
```
guhetier: nit:
```suggestion
    // make sure the local IP matches and the full remote address matches along with QTIP settings.
```

src/platform/datapath_raw.h
```diff
@@ -334,21 +334,29 @@ BOOLEAN
...
-    _In_ const QUIC_ADDR* RemoteAddress
+    _In_ const QUIC_ADDR* RemoteAddress,
+    _In_ const BOOLEAN UseQtip
     )
 {
     CXPLAT_DBG_ASSERT(QuicAddrGetPort(&Socket->LocalAddress) == QuicAddrGetPort(LocalAddress));
     if (Socket->Wildcard) {
-        return TRUE; // The local port match is all that is needed.
+        //
+        // For a listener socket,
```
mtfriesen: Let's fix up the line lengths here and avoid breaking the comment after 4 words.

src/platform/datapath_raw.c
```diff
@@ -238,7 +239,7 @@ CxPlatDpRawRxEthernet(
                     if (i == PacketCount - 1 ||
                         Packets[i+1]->Reserved != SocketType ||
```
guhetier, line 240: When a QTIP listener receive UDP traffic, won't this check cause us to break every packet? Or is it taking into account we can accept both UDP and TCP on the listener?
I am not sure how the `PacketChain` get its `UseQTIP` tag but the assert makes it look like it would not be smart.

ProjectsByJackHe: The function `CxPlatSocketCompare` will always match ALL packets if the socket has QTIP enabled and is a listener socket.

src/core/library.c
ProjectsByJackHe: We can just call it out that the caller enforces this.

src/platform/datapath_raw.c
guhetier: We still break here if `Packets[i+1]->Reserved != SocketType`, independently of what `CxPlatSocketCompare` returns.

ProjectsByJackHe: The only case we reject a packet for a listener socket is if the packet has UseQTIP = true but the listener did not enable QTIP.

This scenario is effectively impossible because that would imply some other QTIP listener socket was created on the same port as our listener socket.

ProjectsByJackHe: > We still break here if `Packets[i+1]->Reserved != SocketType`, independently of what `CxPlatSocketCompare` returns.

For this, the code appears to be doing a nested loop for a singular traversal of a 1D list of packets. 

SocketType is derived PER PACKET.

Suppose the list looks like [UDP], [UDP], [TCP], [UDP], [TCP], [TCP] and our socket is a QTIP listener socket.

An outer for loop will start an inner loop that tries to chain together as many packets of the same type as possible. 

So [UDP] [UDP] will be processed, because the 'SocketType' variable looks at the first packet. BREAK.

And next, SocketType will be TCP... etc. etc.

Now, if the Socket is a client QTIP socket, it will only process the [TCP] packets because CxPlatSocketCompare will return false. Vice versa.

src/core/library.c
ProjectsByJackHe: That would be more complex IMO. The parameter list would explode and IDK if creating a struct just for this is justified. @mtfriesen thoughts? We can perhaps save this for a bigger refactor PR.

docs/QTIP.md
```diff
@@ -24,7 +24,24 @@ This setting can be overridden per client connection, allowing you to create som
...
+Since apps can create many client connections with different QTIP enablements, sometimes the OS assigns
+the same TCP and UDP port number. If using client connections with and without QTIP enabled simultaneously, the application should not assume 4-tuples uniquely identify a connection and also track the QTIP state.
+
+
+**Listeners with different QTIP enablements shall NOT be able to exist on the same local port.**
+
+A QTIP-enabled listener will reserve both UDP/TCP ports equal to the local port of the listener
+and configure XDP to intercept UDP/TCP packets on that local port. A non-QTIP listener will just reserve a UDP port
+and have XDP intercept UDP packets on that port. To avoid conflicting traffic, we disallow 2 listeners with the same
+local port but different QTIP enablements to exist.
```
mtfriesen, line 47: I'm actually confused: if a non-QTIP listener reserves UDP port N, wouldn't any other type of listener (QTIP or non-QTIP) also attempt to reserve UDP port N, and thus also be prohibited? Is there any reason why the special case of QTIP needs to be described?

src/core/library.c
mtfriesen: I don't have a strong opinion; my weak opinion is to echo the prevailing preference on our team not to refactor code used only once into subroutines: it tends to obfuscate the code. I don't see how this particular code would be significantly easier to comprehend if it were relocated somewhere else. 

Ultimately it's a subjective choice in each case.

src/platform/datapath_raw.h
```diff
@@ -334,21 +334,29 @@ BOOLEAN
...
-    // Make sure the local IP matches and the full remote address matches.
+    // For a client socket, make sure the local IP matches and the full
+    // remote address matches along with QTIP settings.
     //
     CXPLAT_DBG_ASSERT(Socket->Connected);
     return
         QuicAddrCompareIp(&Socket->LocalAddress, LocalAddress) &&
-        QuicAddrCompare(&Socket->RemoteAddress, RemoteAddress);
+        QuicAddrCompare(&Socket->RemoteAddress, RemoteAddress) &&
+        Socket->ReserveAuxTcpSock == UseQtip;
```
mtfriesen, line 359: There may have already been a comment about this, but I'd recommend renaming `ReserveAuxTcpSock` to include `Qtip` in its name, and ideally still keep `Tcp` in the name.

Something like `ReserveTcpQtipSock` might be clearest? Obviously it breaks the abstraction that the TCP socket could be used for arbitrary TCP traffic, but since that feature is not currently used and we have no roadmap for it, let's make this code as intuitive as possible by scoping down the abstraction.

src/platform/datapath_raw.c
mtfriesen: >This scenario is effectively impossible because that would imply some other QTIP socket was created on the same port as our listener socket.

Should we DBG_ASSERT this?

src/core/library.c
guhetier: I'll let Jack decides on what he wants, this was mostly a nit.

I don't know where this "not to refactor code used only once into subroutines: it tends to obfuscate the code" emerged from. I have guesses based on our legacy codebases. IMO, this is a code smell, it both progressively leads to unmaintainable many-hundred line of code long functions with complex flow control, and is a lost opportunity to self-document code. Here, the test is essentially a "is the binding that I found compatible with my requirement" test. That can be indicated in a function name, or left to the reader to figure out every time, whether they care or not.

src/platform/datapath_raw.h
guhetier: IMO, `UseQtip` or `EnableQtip` (or something along those lines) would be fine.
We use `ReserveAuxTcpSock` for things that have nothing to do with reserving a TCP socket, where the only thing that matters is that Qtip is used. Reserving a TCP socket is also only a consequence of using QTIP.

src/platform/datapath_raw.c
guhetier: Oh, got it, in `CXPLAT_DBG_ASSERT(!Socket->HasFixedRemoteAddress || Socket->ReserveAuxTcpSock == PacketChain->Route->UseQTIP);`, `!Socket->HasFixedRemoteAddress` exclude the server side scenario.

We leak some assumption from the core layer in the datapath, but that's not the goal of this PR to fix it.

src/core/library.c
mtfriesen: Our team roster has changed...substantially...since the last time we had a general discussion about this, so we can re-open the debate. Ultimately there is no universal truth, with strong arguments on both sides, so it is most efficient (IMO) to simply keep on keeping on.

You can read into the above my own opinion: the subroutine-for-one debate is at best marginally more productive than the origin story of the "endianness" debate 😉

mtfriesen: As a bonus, just like endianness, changing the team preference on subroutines immediately obsoletes the style of all existing code, adding to our technical debt...

docs/QTIP.md
guhetier: An app could be trying strange things. I can't really think of a scenario where it would make sense to try it, but that has not prevented interesting API calls to be written in the past :)

src/platform/datapath_raw.c
ProjectsByJackHe: Yup, that's the intention with that assert.

src/core/library.c
ProjectsByJackHe: Let's table this refactor in a separate PR

src/platform/datapath_raw.h
ProjectsByJackHe: Let's table this refactor in a separate PR. My opinion is ReserveAuxTcpSock is the best name (I renamed it to this from 'UseTcp' previously. But we can have this discussion in a follow up.

docs/QTIP.md
ProjectsByJackHe: this part is called out because it was possible for 2 listeners on the same local port to exist as long as they had different ALPNs, and MsQuic WILL share their bindings. 

The code changes in `library.c` in this PR specifically prohibits 2 listeners from sharing a binding IFF they have different QTIP enablements. (listeners sharing a binding was a common case where they had the same IP (listener A: 443/HTTP3, listener B: 443/some_other_app_protocol)

mtfriesen: Oh yeah, that's a critical point. I don't know whether ALPNs belong in the description here, but maybe it wouldn't hurt to explain why callers within QUIC would want to share ports. I'll leave that up to you.

PR #5762, opened 2026-02-02
Add an RDMA dev doc for WIP

ProjectsByJackHe:
## Description

Let's add a WIP document for integrating the RDMA datapath. 

This is so that one day when everything is working, we have a wealth of information to formally 
document the official usage of MsQuic over RDMA.

## Testing

No. 

## Documentation

RDMA.md

### DIFF
--- a/docs/experimental/RDMA.md
+++ b/docs/experimental/RDMA.md
@@ -1,7 +1,9 @@
 # RDMA Datapath Proposal
 
-Important: This article is the proposed API design for how applications consume MsQuic over RDMA.
-This datapath is still very much a work in progress. Application developers should not expect any support until
+Important: This article is a work-in-progress document for implementing
+the RDMA datapath.
+
+Application developers should not expect any support until
 otherwise indicated.
 
 
@@ -34,22 +36,23 @@ over datapath initialization and worker thread allocation.
 
 Additionally, MsQuic requires the RNIC adapter interface IPv4/IPv6 address for datapath initialization.
 
-Thus, we will add `RdmaEnabledOnInterfaceIp: QUIC_ADDR` into `QUIC_REGISTRATION_CONFIG`,
-which is provided to the OpenRegistration call that initializes the datapath layer and will need this information. `RdmaEnabledOnInterfaceIp` will be
-set to NULL by default. If it's any valid adapter address, then MsQuic will
-initialize the RDMA datapath.
-
+Proposed API design:
 ```C
+typedef struct RDMA_DATAPATH_CONFIG {
+    BOOLEAN RdmaEnabled;
+    QUIC_ADDR RdmaInterfaceAdapterIp;
+}
+
 typedef struct QUIC_REGISTRATION_CONFIG {
     const char* AppName;
     QUIC_EXECUTION_PROFILE ExecutionProfile;
-    QUIC_ADDR RdmaEnabledOnInterfaceIp;
+    RDMA_DATAPATH_CONFIG RdmaConfig;
 } QUIC_REGISTRATION_CONFIG;
 ```
 
 
 If applications would like to use XDP, normal OS sockets, and RDMA on the same machine, they
-should create multiple registrations.
+should create multiple registrations. 
 
 ## RDMA facts
 


docs/experimental/RDMA.md
```diff
@@ -0,0 +1,54 @@
+# RDMA Datapath Proposal
+
+Important: This article is the proposed API design for how applications consume MsQuic over RDMA.
```
guhetier: nit: Having API design work document checked in repos can be nice, but we should get them in their own folder, not the general "docs" folder IMO.

docs/RDMA.md
```diff
@@ -0,0 +1,54 @@
...
+This is primarily due to the RDMA datapath impacting the core logic of QUIC itself (much of MsQuic loss detection logic can be disabled), along with major updates to core MsQuic structures and the entire datapath initialization stack.
+With how fluid `QUIC_SETTINGS` can update based on many different contexts,
+the integration logic for RDMA will be very complex.
+Having multiple connections or listeners with varying RDMA settings is a non-goal.
+
+
+Thus the `RdmaEnabled` option should live at the registration level; to give maximal control
+over datapath initialization and worker thread allocation.
+
+A new flag `QUIC_EXECUTION_PROFILE_RDMA` can be added to `QUIC_EXECUTION_PROFILE`.
```
guhetier: I think we should keep execution options (largely about scheduling options) and datapath options separated. Conceptually, nothing prevents "max throughput + RDMA" or "low latency + RDMA".

docs/experimental/RDMA.md
```diff
@@ -0,0 +1,54 @@
...
+
+Thus the `RdmaEnabled` option should live at the registration level; to give maximal control
+over datapath initialization and worker thread allocation.
+
+A new flag `QUIC_EXECUTION_PROFILE_RDMA` can be added to `QUIC_EXECUTION_PROFILE`.
+When enabled for a registration, completely disallow XDP/QTIP/OS sockets usage under this registration.
+The core logic for MsQuic will also update accordingly.
+
+If applications would like to use XDP, normal OS sockets, and RDMA on the same machine, they
+should create multiple registrations.
```
guhetier: I like that way of doing it - we should double check with partners it will work.
It would also give us the option to make "XDP only" registrations / "XDP disabled" registrations if we wanted to, or we can easily allow RDMA + XDP if we realize it is better to have them under the same registration.

docs/RDMA.md
```diff
@@ -0,0 +1,54 @@
...
+When enabled for a registration, completely disallow XDP/QTIP/OS sockets usage under this registration.
+The core logic for MsQuic will also update accordingly.
+
+If applications would like to use XDP, normal OS sockets, and RDMA on the same machine, they
+should create multiple registrations.
+
+Having XDP/QTIP available, normal OS sockets available, and RDMA available all under the same
```
guhetier: nit: I don't think test matrix should be a consideration in API design, but I agree with the idea: RDMA going to be different enough that preventing interaction with non-RDMA is a good thing.

docs/experimental/RDMA.md
```diff
@@ -0,0 +1,54 @@
...
+The model for using MsQuic over the RDMA datapath can **not**
+easily inherit from the general model where MsQuic lazily initializes RDMA on a best effort basis and the datapath implicitly reads from `QUIC_SETTINGS`.
+
+This is primarily due to the RDMA datapath impacting the core logic of QUIC itself (much of MsQuic loss detection logic can be disabled), along with major updates to core MsQuic structures and the entire datapath initialization stack.
+With how fluid `QUIC_SETTINGS` can update based on many different contexts,
+the integration logic for RDMA will be very complex.
+Having multiple connections or listeners with varying RDMA settings is a non-goal.
+
+
+Thus the `RdmaEnabled` option should live at the registration level; to give maximal control
```
guhetier, line 34: When enabling RDMA, we need to provide an adapter address (IPv4 or IPv6) for the datapath initialization. The API must let the app provide it.

docs/experimental/RDMA.md
mtfriesen: Agreed but I'd still keep the WIP documents within `docs` - just another directory deeper.

docs/RDMA.md
```diff
@@ -0,0 +1,54 @@
...
+registration where different listeners and different connections can have different datapath enablements and depending on the datapath enablement would change the QUIC logic would explode the test matrix and state machine.
+
+## RDMA facts
+
+RDMA over RoCEv2, which is what MsQuic will use,
+ intercepts packets and redirects them at the hardware level.
+The RNIC is programmed by the provider driver (MANA or Mellanox) to redirect all packets matching a specific well known port X (usually 4791) and DMA them into the RDMA datapath then de-muxed using the QPN (queue pair number) into specific queue pairs (QP). A QP is a socket-like endpoint for applications using RDMA.
```
mtfriesen: Let's be neutral to IHVs here - MsQuic should interoperate with any and all NICs providing the required RDMA features.

docs/experimental/RDMA.md
ProjectsByJackHe: noted

docs/experimental/RDMA.md
guhetier: I see this was moved to an "experimental" folder.
To me, that name sounds like where we would put documentation for experimental / preview features.

This document is not documentation, but a design proposal (it doesn't describe what is, but what may be in the future). "proposal" / "design_proposal" would be a better name IMO.

docs/experimental/RDMA.md
```diff
@@ -0,0 +1,51 @@
...
+over datapath initialization and worker thread allocation.
+
+Additionally, MsQuic requires the RNIC adapter interface IPv4/IPv6 address for datapath initialization.
+Thus, the RDMA datapath will require `RdmaEnabled: BOOLEAN` and `RdmaInterfaceIp: QUIC_ADDR` be added to the `QUIC_REGISTRATION_CONFIG` structure.
+`QUIC_REGISTRATION_CONFIG` is provided to the OpenRegistration call, which initialize the datapath layer and will need this information.
+
+If applications would like to use XDP, normal OS sockets, and RDMA on the same machine, they
+should create multiple registrations.
+
+## RDMA facts
```
guhetier, line 57: I don't think this section belong in this doc.
What it contains is largely correct, but not relevant to the API design.

docs/experimental/RDMA.md
```diff
@@ -0,0 +1,51 @@
...
+This is primarily due to the RDMA datapath impacting the core logic of QUIC itself (much of MsQuic loss detection logic can be disabled), along with major updates to core MsQuic structures and the entire datapath initialization stack.
+With how fluid `QUIC_SETTINGS` can update based on many different contexts,
+the integration logic for RDMA will be very complex.
+Having multiple connections or listeners with varying RDMA settings is a non-goal.
+
+
+Thus the `RdmaEnabled` option should live at the registration level; to give maximal control
+over datapath initialization and worker thread allocation.
+
+Additionally, MsQuic requires the RNIC adapter interface IPv4/IPv6 address for datapath initialization.
```
guhetier, line 37: If you want real feedback over the API surface, I'd suggest adding a code block in this doc showing the change you have in mind (only the API surface, no need for any implementation).

It'll let us comment on details like wrapping things in a union / exact names, ...

docs/experimental/RDMA.md
```diff
@@ -0,0 +1,51 @@
...
+During the datapath initialization step, we **always** attempt to initialize the XDP datapath on
+a best-effort basis.
+
+
+## RDMA Datapath Consumption
+
+The model for using MsQuic over the RDMA datapath can **not**
+easily inherit from the general model where MsQuic lazily initializes RDMA on a best effort basis and the datapath implicitly reads from `QUIC_SETTINGS`.
+
+This is primarily due to the RDMA datapath impacting the core logic of QUIC itself (much of MsQuic loss detection logic can be disabled), along with major updates to core MsQuic structures and the entire datapath initialization stack.
```
guhetier: That's not really the case, we could handle that at the connection level.
The constrain is knowing if we need to initialize the RDMA datapath, and providing the IP addr of the local adapter.
We need that config when creating the registration.

docs/experimental/RDMA.md
ProjectsByJackHe: Can we give some wiggle room for MDs in the /experimental folder?

Yes, this information might not be relevant currently, but I still think it's useful to know.

Once we actually standardize this, and move this doc out of /experimental then we can make it more formalized.

docs/experimental/RDMA.md
ProjectsByJackHe: Updated to call that out

PR #5775, opened 2026-02-05
Add automation to reverse mirror PRs

mtfriesen:
## Description

Instead of ad-hoc process, or documenting process to follow, simply automate reverse mirroring internal PRs.

## Testing

Seems to work locally. YMMV - contributions welcome.

Sample: https://github.com/microsoft/msquic/pull/5774

## Documentation

N/A.

### DIFF
--- a/scripts/Create-ReverseMirrorPR.ps1
+++ b/scripts/Create-ReverseMirrorPR.ps1
@@ -87,7 +87,6 @@ foreach ($tool in $RequiredTools) {
 # Ensure we're logged into Azure DevOps
 Write-Host "Checking Azure DevOps authentication..." -ForegroundColor Yellow
 try {
-    az devops configure --defaults organization="https://dev.azure.com/$AdoOrg" project=$AdoProject | Out-Null
     $null = az account show 2>$null
     if ($LASTEXITCODE -ne 0) {
         Write-Host "Logging into Azure DevOps..." -ForegroundColor Yellow


scripts/Create-ReverseMirrorPR.ps1
```diff
@@ -0,0 +1,226 @@
+<#
```
ProjectsByJackHe, line 1: Thanks a lot for this! Super useful.

I guess the unofficial process is *someone* polls ADO and manually runs this script when they see a PR filed.

Could we document this somewhere? Not in this PR necessarily but it would be our team policy going forward.

mtfriesen: Yeah, we should document that somewhere, though I'm not sure where is best.

scripts/Create-ReverseMirrorPR.ps1
```diff
@@ -0,0 +1,226 @@
...
+    [string]$AdoRemote = "undock",
+
+    [Parameter(Mandatory = $false)]
+    [string]$GitHubRemote = "origin"
+)
+
+Set-StrictMode -Version 'Latest'
+$PSDefaultParameterValues['*:ErrorAction'] = 'Stop'
+
+# Parse PR ID from URL if needed
```
mtfriesen, line 70: Future work: some of the other script parameters could be deduced from the URL.

(It's still a nice bit of sugar to support non-URL IDs)

scripts/Create-ReverseMirrorPR.ps1
```diff
@@ -0,0 +1,226 @@
...
+    if (-not (Get-Command $tool -ErrorAction SilentlyContinue)) {
+        Write-Error "Required tool '$tool' not found. Please install it first."
+        exit 1
+    }
+}
+
+# Ensure we're logged into Azure DevOps
+Write-Host "Checking Azure DevOps authentication..." -ForegroundColor Yellow
+try {
+    az devops configure --defaults organization="https://dev.azure.com/$AdoOrg" project=$AdoProject | Out-Null
```
guhetier: The script changing the defaults of the user is not great. Is this needed? Commands below seem to be providing the org.

mtfriesen: Good catch, I doubt it's needed - probably more AI hallucination. The AI was quite insistent the project was important, yet didn't even exist in most APIs.

mtfriesen: Seems to work fine without it.

scripts/Create-ReverseMirrorPR.ps1
guhetier: We could have a sub folder in our docs for "dev" doc (for MsQuic internal design doc) and/or "maintainer" doc (for all these kind of info we should keep in mind).

ProjectsByJackHe (approved):

scripts/Create-ReverseMirrorPR.ps1
mtfriesen: "Maintainer" is a nice term.

PR #5776, opened 2026-02-05
Ignore all Ubuntu Docker dependabot updates

ProjectsByJackHe:
## Description

ADO policy should dictate the source of truth for the SHA version, not Github.
Let's disable it completely and use the reverse mirror script in #5775 

## Testing

No 

## Documentation

No

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -48,7 +48,6 @@ updates:
       day: "saturday"
     ignore:
       - dependency-name: "ubuntu"
-        update-types: ["version-update:semver-major", "version-update:semver-minor"]
 
   - package-ecosystem: "docker"
     directories:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
mtfriesen, line 43: I suspect the 22 and 24 ubuntu configs can be re-merged now that we're simply ignoring both.

ProjectsByJackHe: True. But idk maybe if in the future we find a holistic solution that doesn't involve the reverse mirror, we may still want to keep it separate.

guhetier, line 43: Why not delete that entire section instead of declaring it to ignore everything?

ProjectsByJackHe: Similar to my response above, I want to leave the skeleton in case we find a more holistic solution in the future.

PR #5783, opened 2026-02-06
Do not use bitflags in an attempt to solve flakyness on ValidateAlpns test

### LINKED ISSUE #5752: [CI - FAILURE]:  / Alpn.ValidAlpnLengths (msquictest.exe)
https://github.com/microsoft/msquic/actions/runs/21447602051

guhetier:
## Description

The only think I can think of triggering test instability is something strange happening with bitflags of `bool` type.
Move to bools, there is no reason to keep bitflags here

By moving everything to bools and solving runtime errors, I found some members were not properly initialized. This + bitfields might have caused undefined behavior. I moved initializations to the declaration point, where it is easy to spot missing ones.

Fixes #5752

## Testing

C/I

## Documentation

N/A

### DIFF
--- a/src/test/lib/TestConnection.cpp
+++ b/src/test/lib/TestConnection.cpp
@@ -19,6 +19,8 @@ TestConnection::TestConnection(
     _In_opt_ NEW_STREAM_CALLBACK_HANDLER NewStreamCallbackHandler
     ) :
     QuicConnection(Handle),
+    IsServer(true),
+    IsStarted(true),
     NewStreamCallback(NewStreamCallbackHandler)
 {
     CxPlatEventInitialize(&EventConnectionComplete, TRUE, FALSE);


ProjectsByJackHe (dismissed):

src/test/lib/TestConnection.cpp
```diff
@@ -19,18 +19,7 @@ TestConnection::TestConnection(
     _In_opt_ NEW_STREAM_CALLBACK_HANDLER NewStreamCallbackHandler
     ) :
     QuicConnection(Handle),
-    IsServer(true), IsStarted(true), IsConnected(false), Resumed(false),
```
anrossi, line 22: You can't just delete this block wholesale. While most members here are default initialized, this constructor initializes `IsServer` and `IsStarted` to `true` intentionally (it's because it's for wrapping server-side connections)

guhetier: Good catch, I scanned though the list and missed these two.
I'll double check those lists in case there is another one I missed.

PR #5784, opened 2026-02-10
platform/posix: reject IPv6 link-local addresses without scope id in `QuicAddrIsValid`

### LINKED ISSUE #3813: Incorrect handling of link local addresses and IPv6 scope
### Describe the bug

I created listener to listen on LLA `fe80::7260:3b3d:bd2b:c845%6` and I open connection to it from same host.
On Windows local & remote addresses are correct and have scope as expected. 
```
Client [fe80::7260:3b3d:bd2b:c845%6]:65289 [fe80::7260:3b3d:bd2b:c845%6]:65288
Server [fe80::7260:3b3d:bd2b:c845%6]:65288 [fe80::7260:3b3d:bd2b:c845%6]:65289
```

Same test on Linux:
```
Client [fe80::9c3a:b64d:6249:1de8]:54041 [fe80::9c3a:b64d:6249:1de8%2]:48276
```

Client is missing scopeID.

Interestingly, in some cases scope is set for addresses other than LLA.
Connecting to `::1` on Linux and Windows
```
Server [::1%1]:41741 [::1]:50859
Server [::1%1]:65292 [::1]:65293
```

loopback address has scope set even if it should not. 

In all cases so far the connection succeeded but it makes it difficult for companions and I'm. to sure (yet) about multi-machine setup where the scope is essential for selecting correct interface. 

### Affected OS

- [X] Windows
- [X] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

_No response_

### MsQuic version

main

make simple listener and connect to it using LLA. 
observe addresses on established connection on both client & server.  

### Expected behavior

LLA should always have scope set, global addresses should not have scope set. 
Windows and Linux would have same outcome for same scenario. 
Local and remote address on client and server should be exactly opposite. 
But there should not be differences in scope. 

### Actual outcome

scope is missing in some cases on Linux. It can be  incorrectly set for global addresses. 

### Additional details

_No response_

unstabler:
## Description

> **NOTE: LLM-ASSISTED PR**
> This PR was drafted with the assistance of an LLM.

This PR addresses an issue in the POSIX implementation of `QuicAddrIsValid()`, where IPv6 link-local addresses were considered valid even without a specified scope/interface index (`sin6_scope_id`).

- Updated `QuicAddrIsValid` on POSIX platforms to strictly validate that IPv6 link-local addresses (fe80::/10) have a non-zero scope ID. (partially resolves #3813)

## Testing

- **New Test Case**: Added a specific test case for POSIX targets to verify that `QuicAddrIsValid()` returns `FALSE` when provided with an IPv6 link-local address compliant with the format but missing a scope ID.

## Documentation

### DIFF
--- a/src/test/lib/ApiTest.cpp
+++ b/src/test/lib/ApiTest.cpp
@@ -3758,6 +3758,28 @@ void QuicTest_QUIC_PARAM_CONN_LOCAL_ADDRESS(MsQuicRegistration& Registration, Ms
                     &Dummy));
         }
 
+#ifndef _WIN32
+        //
+        // IPv6 link-local addresses require a non-zero scope id on POSIX.
+        //
+        {
+            TestScopeLogger LogScope2("IPv6 link-local without scope id");
+            MsQuicConnection Connection(Registration);
+            TEST_QUIC_SUCCEEDED(Connection.GetInitStatus());
+
+            QUIC_ADDR LinkLocalAddr = {};
+            TEST_TRUE(QuicAddrFromString("fe80::1", 0, &LinkLocalAddr));
+            LinkLocalAddr.Ipv6.sin6_scope_id = 0;
+
+            TEST_QUIC_STATUS(
+                QUIC_STATUS_INVALID_PARAMETER,
+                Connection.SetParam(
+                    QUIC_PARAM_CONN_LOCAL_ADDRESS,
+                    sizeof(LinkLocalAddr),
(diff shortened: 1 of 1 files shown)


guhetier:
Thanks for opening this PR.
I am not sure a partial fix is enough: if we make the validation stricter without fixing the underlying issue (why isn't the scope ID set), it may regress existing working scenario.

PR #5785, opened 2026-02-10
core/connection: fix EXC_BAD_ACCESS in send flush after local address migration

unstabler:
## Description

> **NOTE: LLM-ASSISTED PR**
> This PR was drafted with the assistance of an LLM.

This PR fixes a crash during local address rebinding when multipath is active.

### Background

We observed intermittent `EXC_BAD_ACCESS` on the client worker thread in a real environment:

- Topology: macOS host (server, MsQuic) <-> iPad (client, MsQuic)
- Client and server were on the same network, with iPad also connected over USB for Xcode debugging
- In this setup, connecting via `macoshost.local` can create a USB network path and effectively enable multipath conditions
- IPv6 link-local addresses changed frequently (roughly every 2-3 minutes in the observed network)
- The client called `SetParam(QUIC_PARAM_CONN_LOCAL_ADDRESS, newAddress)` whenever local interface/address changed (for migration)

Observed crash stack (symbol-only):
- `pthread_mutex_lock`
- `CxPlatPoolAlloc`
- `CxPlatSendDataAlloc`
- `QuicPacketBuilderPrepare`
- `QuicPacketBuilderPrepareForControlFrames`
- `QuicSendFlush`
- `QuicConnDrainOperations`
- `QuicWorkerProcessConnection`
- `QuicWorkerLoop`

### Root cause

During `QUIC_PARAM_CONN_LOCAL_ADDRESS` handling, binding replacement/reset behavior was not robust for active multipath state transitions.

A stale per-path route queue pointer could survive rebinding in certain path-ordering/swap cases.
That stale queue was later consumed from send flush paths, which could lead to invalid memory access.

### Fix

In local-address rebind handling (`QuicConnParamSet`), after acquiring the new binding:

- Normalize **all** paths to the new binding
- Reset each path route state to unresolved
- Clear each path route queue pointer (`Route.Queue = NULL`)

This guarantees no stale queue/binding state remains after rebinding, even if active path ordering changed earlier under multipath.

### Impact

- Prevents worker-thread send-path crashes after local address rebinding in multipath scenarios
- Makes local rebind behavior deterministic across path state transitions
- Improves robustness even when network paths are volatile (e.g., link-local churn + USB/debug path changes)

## Testing

- **New regression test**: `ClientLocalAddrRebindWithMultiplePaths`
- Scenario validated by test:
  - Establish connection
  - Trigger multiple paths / peer-address change
  - Rebind client local address via `SetLocalAddr`
  - Keep traffic flowing and verify stability
- **Existing related test** also re-run: `LocalPathChanges`

Commands run:
- `cmake --build cmake-build-debug --target msquictest -j8`
- `./bin/Debug/msquictest --gtest_filter='*ClientLocalAddrRebindWithMultiplePaths*'`
- `./bin/Debug/msquictest --gtest_filter='*LocalPathChanges*'`

Result:
- New regression tests passed
- Existing related tests passed

## Documentation

No public API surface changed.
No documentation update is required for this fix.

### DIFF
--- a/src/test/bin/quic_gtest.cpp
+++ b/src/test/bin/quic_gtest.cpp
@@ -703,6 +703,15 @@ TEST_P(WithFamilyArgs, LocalPathChanges) {
     }
 }
 
+TEST_P(WithFamilyArgs, ClientLocalAddrRebindWithMultiplePaths) {
+    TestLoggerT<ParamType> Logger("QuicTestClientLocalAddrRebindWithMultiplePaths", GetParam());
+    if (TestingKernelMode) {
+        GTEST_SKIP_("No kernel test control opcode wired for this scenario.");
+    } else {
+        QuicTestClientLocalAddrRebindWithMultiplePaths(GetParam().Family);
+    }
+}
+
 TEST(Mtu, Settings) {
     TestLogger Logger("QuicTestMtuSettings");
     if (TestingKernelMode) {
--- a/src/test/MsQuicTests.h
+++ b/src/test/MsQuicTests.h
@@ -153,6 +153,11 @@ QuicTestLocalPathChanges(
     const FamilyArgs& Params
     );
 
+void
+QuicTestClientLocalAddrRebindWithMultiplePaths(
+    _In_ int Family
+    );
+
 //
 // Handshake Tests
 //
--- a/src/test/lib/PathTest.cpp
+++ b/src/test/lib/PathTest.cpp
@@ -42,6 +42,11 @@ struct PathTestContext {
     }
 };
 
+struct ClientPathTestContext {
+    CxPlatEvent PeerStreamsChanged;
+    CxPlatEvent PeerAddrChanged;
+};
+
 static
 QUIC_STATUS
 QUIC_API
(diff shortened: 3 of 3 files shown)


src/test/bin/quic_gtest.cpp
```diff
@@ -821,6 +703,15 @@ TEST_P(WithFamilyArgs, LocalPathChanges) {
     }
 }
 
+TEST_P(WithFamilyArgs, ClientLocalAddrRebindWithMultiplePaths) {
+    TestLoggerT<ParamType> Logger("QuicTestClientLocalAddrRebindWithMultiplePaths", GetParam());
+    if (TestingKernelMode) {
+        GTEST_SKIP_("No kernel test control opcode wired for this scenario.");
```
guhetier: Unless there is a reason this test can't run in kernel mode, please enable it.
We just refactored the kernel mode invocation logic, it should take little boilerplate now.

Please also have `QuicTestClientLocalAddrRebindWithMultiplePaths` take the test param directly as a const ref to follow the new pattern.

src/core/connection.c
```diff
@@ -6326,8 +6326,18 @@ QuicConnParamSet(
...
-            Connection->Paths[0].Route.Queue = NULL;
+
+            //
+            // Keep all existing paths consistent with the newly selected
+            // binding. Non-active paths can still be used for path challenge
+            // traffic; leaving them on the old binding/queue can result in
+            // stale datapath pointers after old binding teardown.
+            //
+            for (uint8_t i = 0; i < Connection->PathsCount; ++i) {
+                Connection->Paths[i].Binding = Connection->Paths[0].Binding;
```
guhetier: I don't think we can simply do that.
A PATH represents a 4-tuple. A binding is, generally speaking, tied to at least the local 2-tuple. This would have the new binding no longer matching the path local address.

For `Path[0]`, we update the address lower in the function. For the other paths, I expect we need to invalidate them completely. I'll have to dig more in the PATH logic to be sure, I am not familiar enough with it to say from the top of my head.

PR #5796, opened 2026-02-17
Update setup perl action

guhetier:
## Description

The build CI started to fail with errors like:

```
Run shogo82148/actions-setup-perl@fa3bdcb0144a50651e317f30949bb4491e74a6e4
install perl
  verifying download: D:\a\_temp\d7077245-f8e5-4d0c-bca6-93c64dc8f837
  Warning: failed to verify attestation: TypeError: Cannot read properties of null (reading 'mediaType'), continue to next attestation
  Warning: failed to verify attestation: TypeError: Cannot read properties of null (reading 'mediaType'), continue to next attestation
Error: Error: failed to verify D:\a\_temp\d7077245-f8e5-4d0c-bca6-93c64dc8f837
```

https://github.com/microsoft/msquic/actions/runs/22020983438/job/63904220521?pr=5794

This has been fixed in https://github.com/shogo82148/actions-setup-perl/issues/2481.

Update the action to fix the CI.

## Testing

CI

## Documentation

N/A

### DIFF
--- a/.github/workflows/build-reuse-win.yml
+++ b/.github/workflows/build-reuse-win.yml
@@ -83,7 +83,7 @@ jobs:
         repository: ${{ inputs.repo}}
         ref: ${{ inputs.ref }}
     - name: Install Perl
-      uses: shogo82148/actions-setup-perl@fa3bdcb0144a50651e317f30949bb4491e74a6e4
+      uses: shogo82148/actions-setup-perl@b6cfc127f1b0a63ac98797800f641494d80f42f8
       with:
         perl-version: '5.34'
     - name: Install NASM
--- a/.github/workflows/cargo.yml
+++ b/.github/workflows/cargo.yml
@@ -40,7 +40,7 @@ jobs:
       shell: pwsh
     - name: Install Perl
       if: runner.os == 'Windows'
-      uses: shogo82148/actions-setup-perl@fa3bdcb0144a50651e317f30949bb4491e74a6e4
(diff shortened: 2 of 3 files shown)


mtfriesen:
LGTM but we need to figure out how to not depend on `shogo82148`.

guhetier:
> LGTM but we need to figure out how to not depend on `shogo82148`.

It seems to be the standard way to install perl in CI. I don't think we have a need / reason to not depend on it in our public CI.

mtfriesen:
> > LGTM but we need to figure out how to not depend on `shogo82148`.
> 
> It seems to be the standard way to install perl in CI. I don't think we have a need / reason to not depend on it in our public CI.

Hmm, maybe the more pertinent but ignorant question is "why do we even use perl?" if the rest of the industry seems not to bother supporting it.

In general, we want to take dependencies with both clear support and security policies - if the action is not a first party (GitHub or Microsoft) action, then it becomes a more nuanced decision.

guhetier:
> ...
> 
> Hmm, maybe the more pertinent but ignorant question is "why do we even use perl?" if the rest of the industry seems not to bother supporting it.
> 
> In general, we want to take dependencies with both clear support and security policies - if the action is not a first party (GitHub or Microsoft) action, then it becomes a more nuanced decision.

Perl is needed to build OpenSSL.

mtfriesen:
> > Hmm, maybe the more pertinent but ignorant question is "why do we even use perl?" if the rest of the industry seems not to bother supporting it.
> 
> Perl is needed to build OpenSSL.

Now it all makes sense! 😆

PR #5798, opened 2026-02-18
Add support for multi-process port sharing with CIBIR.

### LINKED ISSUE #5795: [Partner-SQL] Multiple Quic listeners needed on same port on Windows
### Describe the bug

Requirements are documented in this draft: https://www.ietf.org/archive/id/draft-banks-quic-cibir-00.html

Multiple processes need to reuse the same port on Windows VMs. Cannot expose different ports or IPs for each process running on the VM, hence different processes need to share/reuse the same listener port. Using CIBIRID the incoming connections should get demultiplexed as each process can register a different CIBIR ID but use the same port. 

This is a blocker for production use case.

Currently if we spawn two different processes sharing same port and different CIBIR would get this WIN Error: 
**[3]1FA4.1E04::2026/02/15-22:47:26.551337800 [Microsoft-Quic][data][0x1BF50EF47E0] ERROR, 10048, SIO_ACQUIRE_PORT_RESERVATION.
[3]1FA4.1E04::2026/02/15-22:47:26.551338600 [Unknown]Unknown( 32): GUID=bbccf6c1-6cd1-48c4-80ff-839482e37671 (No Format Information found).**
[3]1FA4.1E04::2026/02/15-22:47:26.551338600 [Microsoft-Quic][data][0x1BF50EF47E0] Destroyed
[3]1FA4.1E04::2026/02/15-22:47:26.551381300 [Microsoft-Quic][data][000001BF50EF4860] Socket context shutdown

### Affected OS

- [x] Windows
- [ ] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

_No response_

### MsQuic version

v2.5

1. Create two processes listening on the same Port. 
2. Both are using xdp configuration 
3. Second process listener would fail with the above-mentioned error

### Expected behavior

Both processes should reuse port and use cibir id to route the incoming connections

### Actual outcome

Getting the error mentioned above.

### Additional details

_No response_

ProjectsByJackHe:
## Description

Fixes #5795 

The XDP datapath can be configured to intercept packets based on QUIC Connection ID instead of local port.
This behavior existed in MsQuic but was not heavily exercised until recently.
One issue was that MsQuic always attempted to reserve UDP / TCP sockets for each application server process.
But for multiple server processes that may want to share a single port, we would run into port collision errors.
This PR adds support for CIBIR across multiple processes on the same port and document the behavior

## Potential options to allow for multi-process port sharing:

- **Option 1:** MsQuic delegates port protection to applications and provides best practice recommendations. 
 > **Analysis:** Deferring the responsibility of port protection and isolation to the application has the upside of enabling the most potential scenarios but could also be a footgun. 
- **Option 2:** MsQuic makes sure *some* persistent reservation exists at a port. 
 > **Analysis:** Note that LookUpPersistentReservation does not require admin privileges, but CreatePersistentReservation does require admin. This is useful in that if any reservation exists on a port, we can reasonably trust that an admin knew what they were doing when they created it. safety, and ensure the consumers of CIBIR must know what they are doing.
- **Option 3:** MsQuic creates per-proc sockets with SIO_CPU_AFFINITY, but does not reserve the port. 
> **Analysis:** If another unrelated app creates a socket with SIO_CPU_AFFINITY, then they can bind to the CIBIR shared port.  But for all other apps, trying to bind a socket to a CIBIR port will result in a collision.  

## Option chosen: 1

MsQuic's stance is that the application takes responsibility for book-keeping and protecting sharing shared local ports when using XDP + CIBIR.

- Multiple MsQuic processes in Cibir+XDP mode can share a local port for **server sockets only.**
- Applications should also not assume the shared port is safe from other non-Msquic processes binding to it.
>MsQuic will NOT make an OS port reservation for server sockets when CIBIR+XDP is enabled. Clients on the other hand, MsQuic will always make OS port reservations.
- Applications using server sockets + CIBIR/XDP must specify a well-known local port.

## What changed

- Server sockets with XDP+CIBIR both enabled/available will skip OS port reservation and OS socket creation to rely on XDP.
> any failures plumbing xdp rules will bubble up as a socket creation error to the app. Can't fall back to OS sockets.
- Client sockets with XDP+CIBIR both enabled/available will still do OS port reservation and socket creation but rely on XDP.
> any failures plumbing xdp rules will silently fall back to using OS sockets. CIBIR transport negotiation can still work without XDP.
- Server sockets with CIBIR enabled but XDP not available/enabled will do OS port reservation and fall back to OS sockets
- Client sockets with CIBIR enabled but XDP not available/enabled will do OS port reservation and fall back to OS sockets

## Port protection options

- Windows has the https://learn.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-createpersistentudpportreservation API, to allow sysadmins to pre-allocate a block of ports and disallow other applications from binding to it. 
- A well known CIBIR registry key can be used to detail shared ports, and sysadmins can coordinate their system such that other apps will not bind to those ports.
- ALE policies; applications can configure WFP to block certain ports from being binded to by other apps.

## Testing

A new DataPathTest was added.

## Documentation

Settings.md

### DIFF
--- a/src/platform/datapath_xplat.c
+++ b/src/platform/datapath_xplat.c
@@ -198,7 +198,7 @@ CxPlatSocketCreateUdp(
             Status = QUIC_STATUS_INVALID_STATE;
             goto Error;
         } else if (CibirRequested) {
-             QuicTraceLogWarning(
+            QuicTraceLogWarning(
                 WarnNoXdpForCibirSockets,
                 "[sock] Warning: app requested CIBIR but XDP not enabled/available/initialized. "
                 "Falling back to normal OS sockets to allow for CIBIR transport parameter negotiation.");
--- a/src/test/lib/HandshakeTest.cpp
+++ b/src/test/lib/HandshakeTest.cpp
@@ -3920,6 +3920,7 @@ QuicTestCibirExtension(
         // CIBIR + XDP requires an explicit local port. Reserve an ephemeral port
         // up front so the listener always binds to a known port.
         //
+        TEST_NOT_EQUAL(0, PortReservation.Port);
         ServerLocalAddr.SetPort(PortReservation.Port);
     }
 #endif
@@ -4490,6 +4491,7 @@ QuicTestConnectionPoolCreate(
         // If Cibir+XDP mode is active, we can't pass in a 0 local port
         // hoping the stack will assign us an ephemeral port.
         //
+        TEST_NOT_EQUAL(0, PortReservation.Port);
         ServerAddr.SetPort(PortReservation.Port);
     }
 #endif


src/platform/datapath_raw_socket.c
```diff
@@ -890,7 +890,7 @@ CxPlatTryAddSocket(
     // binding an auxiliary (dual stack) socket.
     //
 
-    if (Socket->ReserveAuxTcpSock) {
+    if (Socket->ReserveAuxTcpSock && !Socket->SkipCreatingOsSockets) {
```
mtfriesen: Let's follow up on our [commitment ](https://github.com/microsoft/msquic/pull/5747#discussion_r2761572184) to refactor `ReserveAuxTcpSock` to clearly mention QTIP - either as a prerequisite for this PR, or within this PR.

src/platform/datapath_raw_socket.c
```diff
@@ -929,6 +929,14 @@ CxPlatTryAddSocket(
         }
 
         if (Socket->CibirIdLength) {
+            //
+            // Setting SO_REUSEADDR does NOT robustly allow
```
mtfriesen: Can you explain in this comment how/why the option is not robust on Windows and how it is robust on Linux?

Does it make sense for this socket option to be set on both (all?) platforms if it has different semantics?

src/platform/datapath_raw_socket.c
```diff
@@ -929,6 +929,14 @@ CxPlatTryAddSocket(
...
+            // multiple processes to share the same port on
+            // Windows (WinSock).
+            // This code was added primarily for Linux XDP,
```
mtfriesen: Is this implying this code block is known to be buggy on Windows? Should we have different code on Windows, now that we've thought through the full implications?

src/platform/datapath_winuser.c
```diff
@@ -1724,6 +1724,27 @@ SocketCreateUdp(
                         Socket,
                         WsaError,
                         "SIO_ACQUIRE_PORT_RESERVATION");
+
+                    if (Config->CibirIdLength > 0 && IsServerSocket) {
+                        //
+                        // CIBIR is configured. A port collision here is actually
```
mtfriesen: A port collision error can be expected, but something like `WSAEINVAL` would remain unexpected, right?

src/platform/datapath_winuser.c
```diff
@@ -1831,6 +1852,25 @@ SocketCreateUdp(
 
 Skip:
 
+    if (Socket->SkipCreatingOsSockets) {
+        //
+        // Clean up all partially-initialized per-proc sockets since
+        // we're skipping OS socket creation (XDP-only via CIBIR).
+        //
+        for (uint16_t i = 0; i < SocketCount; i++) {
```
mtfriesen: This seems fairly duplicative of part of `CxPlatSocketContextUninitialize` - can it be refactored into a common subroutine?

src/platform/platform_internal.h
```diff
@@ -212,6 +212,8 @@ typedef struct CXPLAT_SOCKET {
     uint8_t HasFixedRemoteAddress : 1;
     uint8_t RawSocketAvailable : 1;
 
+    uint8_t SkipCreatingOsSockets : 1;
```
mtfriesen, line 216: nit: newline spacing of these bitfields is inconsistent.

src/platform/platform_internal.h
mtfriesen: Please address (remove) this `// always false?` comment while refactoring `ReserveAuxTcpSock `.

docs/Settings.md
```diff
@@ -169,7 +169,7 @@ These parameters are accessed by calling [GetParam](./api/GetParam.md) or [SetPa
 |-------------------------------------------|---------------------------|-----------|-----------------------------------------------------------|
 | `QUIC_PARAM_LISTENER_LOCAL_ADDRESS`<br> 0 | QUIC_ADDR                 | Get-only  | Get the full address tuple the server is listening on.    |
 | `QUIC_PARAM_LISTENER_STATS`<br> 1         | QUIC_LISTENER_STATISTICS  | Get-only  | Get statistics specific to this Listener instance.        |
-| `QUIC_PARAM_LISTENER_CIBIR_ID`<br> 2      | uint8_t[]                 | Both      | The CIBIR well-known idenfitier.                          |
+| `QUIC_PARAM_LISTENER_CIBIR_ID`<br> 2      | uint8_t[]                 | Both      | Sets a CIBIR (CID-Based Identification and Routing) well-known identifier. CIBIR does 2 things when set: 1. XDP will now steer packets to the correct process/listener by matching the CIBIR prefix within the packet QUIC Connection ID. 2. In the case of a port collision when reserving OS UDP/TCP sockets, MsQuic will continue with initializing the datapath. If XDP is not available/enabled, then no traffic will flow for the listener that experiences a collision. |
```
mtfriesen: > If XDP is not available/enabled, then no traffic will flow for the listener that experiences a collision.

Is the collision bubbled up to the app? If not, what is the benefit of allowing a listener to be created that will silently not receive the intended traffic? If it is bubbled up as an error, are we saying we'll create the listener, but it just won't receive traffic, or we'll return an error and no listener?

guhetier: Do we have a dedicated page / paragraph for CIBIR? This description becomes long for a table, and a link to some extended doc might be better.

src/platform/platform_internal.h
```diff
@@ -900,6 +904,8 @@ typedef struct CXPLAT_SOCKET {
 
     uint8_t RawSocketAvailable : 1;
 
+    uint8_t SkipCreatingOsSockets : 1;
```
guhetier, line 908: Consider `ReserveOsSockets` as a variable instead, negative variable names tend to be more confusing.
(In C, it needs to be balanced with zero-init)

src/platform/datapath_winuser.c
```diff
@@ -1724,6 +1724,27 @@ SocketCreateUdp(
...
+                        // the same UDP and/or TCP port.
+                        //
+                        CxPlatRefInitializeEx(&Socket->RefCount, 1);
+                        Socket->SkipCreatingOsSockets = TRUE;
+                        BOOLEAN XdpAvailable = Datapath->RawDataPath != NULL;
+                        BOOLEAN XdpEnabled = Config->Flags & CXPLAT_SOCKET_FLAG_XDP;
+                        if (!XdpAvailable || !XdpEnabled) {
+                            QuicTraceLogWarning(
+                                DatapathCibirSkipNoXdp,
+                                "[data][%p] CIBIR configured, skipping OS socket reservation but XDP not %s",
```
guhetier: I'm not sure I understand this comment. If XDP is not available and we skip setting OS sockets, what do we have? Shouldn't we just fail?

src/platform/datapath_winuser.c
```diff
@@ -1724,6 +1724,27 @@ SocketCreateUdp(
                         Socket,
                         WsaError,
                         "SIO_ACQUIRE_PORT_RESERVATION");
+
+                    if (Config->CibirIdLength > 0 && IsServerSocket) {
+                        //
+                        // CIBIR is configured. A port collision here is actually
+                        // expected if multiple server processes were configured to share
```
guhetier: I am not sure what is the best place for it, but I think a comment should document that when a CIBIR is configured, XDP will redirect only the traffic matching this CIBIR to the process, which largely solves the traffic stealing situation that happens without CIBIR.

src/platform/datapath_winuser.c
guhetier, line 1745: Question: Should we even make a port reservation is CIBIR is configured? Or should we consider that XDP redirects the traffic matching the CIBIR, and everything else can flow to whoever is bound?

Side-question: What happens if two processes configure rules to XDP for the same CIBIR? Is it an undefined behavior? Would XDP reject the second rule?

src/platform/datapath_winuser.c
mtfriesen: https://github.com/microsoft/msquic/pull/5798#discussion_r2823264294

src/platform/datapath_winuser.c
mtfriesen: XDP doesn't officially arbitrate between CIBIR rules AFAIK, but I _think_ XDP's behavior is still well-defined, in the sense that the first matching rule would be applied. (The overall scenario is undefined in my mind, because I don't know how MsQuic assembles the list of XDP rules.)

mtfriesen: XDP doesn't officially arbitrate between ports, in the same way: if Rule1 says "redirect UDP port 1234" and so does Rule2, then Rule1 is the only one to execute, since redirect is a terminating rule.

ProjectsByJackHe: > Question: Should we even make a port reservation is CIBIR is configured? Or should we consider that XDP redirects the traffic matching the CIBIR, and everything else can flow to whoever is bound?
> 
> Side-question: What happens if two processes configure rules to XDP for the same CIBIR? Is it an undefined behavior? Would XDP reject the second rule?

We can say let's skip reserving any ports if CIBIR is configured, but that would lead to some undefined behavior.
IMO, we still need at least one of the CIBIR processes to reserve a port on the machine. 

The XDP CIBIR rule still considers local ports.
The exact matching logic is for all received packets: 
Match packets with local port == xyz AND QUIC_HEADER.hasCIDPrefix(CIBIR).

XDP parses the QUIC_HEADER with very little logic for actually validating it conforms to QUIC. So long as there is a sufficient number of bytes. @mtfriesen can correct me. So, in theory, if say we had an unrelated process created on the same local port by the OS, an arbitrary packet COULD contain just the right bytes to satisfy the QUIC_HEADER.hasCIDPrefix(CIBIR) but would otherwise be a bogus packet... 

Additionally, just having a port reservation would align better with tooling/validation.

src/platform/platform_internal.h
ProjectsByJackHe: Noted, will fix all nits once core logic addressed.

src/platform/platform_internal.h
ProjectsByJackHe: Noted, will fix all nits once core logic addressed.

src/platform/platform_internal.h
ProjectsByJackHe: Noted, will fix all nits once core logic addressed.

src/platform/datapath_raw_socket.c
ProjectsByJackHe: This code block frankly was never truly exercised. Not in test code. Not in prod.

From looking at the Blame history, these were added as part of the Linux XDP passion project.

The behavior expected by using CIBIR, as specified in Nick Bank's paper, would not have been possible using the SO_REUSEADDR option on windows, but only on linux. 

With the current changes in this PR, if CIBIR was enabled, the entire outer block of this portion of code will not execute.

src/platform/datapath_raw_socket.c
ProjectsByJackHe: Sure thing for 1)

For 2), no it does not make sense, but we never ran into problems because it was never exercised.

ProjectsByJackHe:
> In general, I am concerned that we keep adding incremental exceptions to the port reservation logic to solve the next issue, but without having a clear design goal. This code will be hard to maintain and confusing for apps as there is no simple rule about what can be done with ports.
> 
> I think we need to take the time soon to come up with a clear story about when can port be shared and when they can't + document it + check we implement it.

I agree. I added a CIBIR.md.

docs/Settings.md
ProjectsByJackHe: added a CIBIR.md

ProjectsByJackHe: Really good points. This PR was opened early so we can unblock SQL. We can flesh out the details more.

I chose the design of logging a warning when XDP is not available/not enabled for ease of debugging / unit testing. But we can find a better solution. 

I am leaning towards just straight up failing the binding creation if you have a listener trying to use CIBIR but xdp is not available/enabled.

src/platform/datapath_winuser.c
ProjectsByJackHe: Noted

src/platform/datapath_raw_socket.c
ProjectsByJackHe: Noted

src/platform/datapath_winuser.c
ProjectsByJackHe: Yes. We can add an explicit check for the port collision error code.

src/platform/datapath_winuser.c
ProjectsByJackHe: CIBIR.md is best IMO

src/platform/datapath_raw_socket.c
ProjectsByJackHe: Done

src/platform/datapath_raw_socket.c
ProjectsByJackHe: I don't want to fully get rid of this until we officially deprecate/delete Linux XDP support.

src/platform/datapath_winuser.c
ProjectsByJackHe: Done

src/platform/platform_internal.h
ProjectsByJackHe: I prefer to keep the name like this if that's OK. I don't want to have to 1-init it.

src/platform/platform_internal.h
ProjectsByJackHe: fixed

ProjectsByJackHe:
> In general, I am concerned that we keep adding incremental exceptions to the port reservation logic to solve the next issue, but without having a clear design goal. This code will be hard to maintain and confusing for apps as there is no simple rule about what can be done with ports.
> 
> I think we need to take the time soon to come up with a clear story about when can port be shared and when they can't + document it + check we implement it.

please see the updated XDP.md and CIBIR.md

src/platform/datapath_raw_socket.c
csujedihy: This piece of code was introduced in https://github.com/microsoft/msquic/commit/5e5b7aea048694ee36d75a1cea4ed235173f73e4, which is way before we added linux xdp datapath.

I'd be interested to know why its not robust on Windows.

ProjectsByJackHe: SO_REUSEADDR option on Windows is dangerous because it risks port hijaking.

But more importantly, SO_REUSEADDR does NOT allow 2 processes to share the same port. That would be the SO_REUSEPORT option, which is only on linux.

From this piece of code in the context of CIBIR, I can only infer that the intention was to share ports across processes. But clearly the SO_REUSEADDR does not accomplish this.

csujedihy: The code never intended to share a port across multiple processes (in a deterministic and load-balanced way that SO_REUSEPORT does). It just allows the bind to succeed across multiple processes.

ProjectsByJackHe: The core problem and the key motivation of this PR was precisely that bind() did not succeed across multiple processes because we got a ADDR_INUSE error. I think this particular usage of SO_REUSEADDR covered the use case of QTIP (it only set this option on the aux socket). But overall, I think the better approach (and the one in this PR) is to just not reserve those sockets if one is already reserved AND cibir is enabled.

csujedihy: Its interesting that bind() failed with `ADDR_INUSE` with this socket option set. Anyways, I think the alternative would work too.

ProjectsByJackHe: To be clear, bind() failed with the UDP sockets when the 2nd process tried to reserve on the same port (with CIBIR enabled). The UDP code didn't have SO_REUSEADDR option added. 

Anyways, it's confusing why that option was not there but set here in the AuxSocket, which only dealt with QTIP...

csujedihy: The code originally set SO_REUSEADDR for UDP when this was all first introduced. See https://github.com/microsoft/msquic/commit/5e5b7aea048694ee36d75a1cea4ed235173f73e4. This is pre-QTIP and pre-linux-XDP.

nibanks:
Maybe we should actually go through the rest of the standardization process for https://datatracker.ietf.org/doc/draft-banks-quic-cibir/ since you're still using it....

guhetier:
> Maybe we should actually go through the rest of the standardization process for https://datatracker.ietf.org/doc/draft-banks-quic-cibir/ since you're still using it....

Do you have more context about it? I am interested in understanding whether we should work on standardizing CIBIR or if another solution was adopted by other deployments that we should align with.

src/platform/datapath_raw_socket.c
ProjectsByJackHe: I don't see it setting SO_REUSEADDR for udp in https://github.com/microsoft/msquic/blob/5e5b7aea048694ee36d75a1cea4ed235173f73e4/src/platform/datapath_winuser.c, or anywhere else in the frozen snapshot of 5e5b7aea048694ee36d75a1cea4ed235173f73e4 except for in AuxSocket.

nibanks:
> > Maybe we should actually go through the rest of the standardization process for https://datatracker.ietf.org/doc/draft-banks-quic-cibir/ since you're still using it....
> 
> Do you have more context about it? I am interested in understanding whether we should work on standardizing CIBIR or if another solution was adopted by other deployments that we should align with.

I know of no other solution/proposal that achieves what CIBIR does. We can chat more on Discord about the process.

docs/Settings.md
ProjectsByJackHe: The behavior should be to return an INVALID_STATE error if CIBIR is used by XDP is not available/enabled.

ProjectsByJackHe: Updated

src/platform/datapath_raw_socket.c
csujedihy: It's exactly in `AuxSocket`, isn't it?

ProjectsByJackHe: The way XDP enablement is configured is at compile time way back then. As in, the datapath predates the "best-effort" initialization of XDP. So I got confused by AuxSocket because I thought the per-proc UDP socket in the WinUser datapath was the way XDP reserved UDP ports. I always associated AuxSocket with TCP sockets. Anyways, that clears things up. Whomever unified the datapaths to make XDP initialization best-effort forgot to set SOCKET_REUSEADDR in the normal WinUser datapath...

ProjectsByJackHe: UPDATE: Got rid of this entire code block. It was a remnant of a refactor to try and unify the raw/winuser datapaths that frankly does not translate well to the new model of how msquic creates WSA sockets today.

src/platform/datapath_winuser.c
ProjectsByJackHe: Actually, this inline cleanup is the cleanest IMO. The CxPlatSocketContextUninitialize also decrements the parent CXPLAT socket refcount too for deletion. We intentionally set the parent refcount to 1 once we know we are meant to skip creating OS sockets. We would need to refactor that method to ignore the parent refcount, which is a bit hairier.

src/platform/platform_internal.h
ProjectsByJackHe: Fixed

docs/images/XDP-arch.png
mtfriesen, line 1: Similar to the comments in XDP, let's try to make future diagrams more maintainable by using Mermaid, ASCII art, etc.

If we need to tweak the PNG for future maintenance, that's quite a bit more painful, and more difficult for text-based AIs to interact with.

docs/CIBIR.md
```diff
@@ -0,0 +1,28 @@
...
+
+See [XDP](./XDP.md) first to understand the context.
+
+When CIBIR is used, rather than programming XDP to filter packets on port numbers,
+we now filter and de-mux packets based on QUIC connection ID.
+
+CIBIR (CID-Based Identification and Routing) is just a prefix substring that XDP
+will use to match and filter all packets with a QUIC CID that contains the prefix substring equal to CIBIR.
+
+What using CIBIR also enables is allowing 2 separate server processes to share a single
```
mtfriesen:
```suggestion
What using CIBIR also enables is allowing 2 or more separate server processes to share a single
```

src/core/connection.c
```diff
@@ -6708,12 +6708,18 @@ QuicConnParamSet(
         Connection->CibirId[0] = (uint8_t)BufferLength - 1;
         memcpy(Connection->CibirId + 1, Buffer, BufferLength);
 
+        uint64_t CibirIdValue = 0;
+        for (uint8_t i = 0; i < Connection->CibirId[0]; ++i) {
+            CibirIdValue = (CibirIdValue << 8) | Connection->CibirId[2 + i];
```
mtfriesen: This seems pretty fragile - can we not just print the ID as an array of bytes? If not, should we create (format) a string in a buffer of characters?

src/platform/datapath_winuser.c
```diff
@@ -1724,6 +1726,40 @@ SocketCreateUdp(
...
+                        //
+                        BOOLEAN XdpAvailable = Datapath->RawDataPath != NULL;
+                        BOOLEAN XdpEnabled = Config->Flags & CXPLAT_SOCKET_FLAG_XDP;
+                        if (!XdpAvailable || !XdpEnabled) {
+                            QuicTraceLogWarning(
+                                DatapathCibirWarning,
+                                "[data][%p] CIBIR detected,  %s",
+                                Socket,
+                                !XdpAvailable ?
+                                "but XDP not available. NO TRAFFIC WILL FLOW ON THIS LISTENER." :
```
mtfriesen: I don't think we need ALL CAPs - the severity of the message should be provided by the log level, where warning seems OK but error might be even better.

src/platform/datapath_winuser.c
```diff
@@ -1724,6 +1726,40 @@ SocketCreateUdp(
...
+                            goto Error;
+                        }
+
+                        QuicTraceLogWarning(
+                            DatapathCibirWarning,
+                            "[data][%p] CIBIR detected,  %s",
+                            Socket,
+                            "ignoring port collision by assuming some \
+                            other MsQuic CIBIR process has reserved the OS port. \
+                            Let's continue with initialization and skip port reservation.");
```
mtfriesen: I'd drop the entire line "Let's continue with initialization and skip port reservation" which can be a comment, and not a log message.

src/platform/datapath_winuser.c
```diff
@@ -1724,6 +1726,40 @@ SocketCreateUdp(
...
+                            Status = QUIC_STATUS_INVALID_STATE;
+                            goto Error;
+                        }
+
+                        QuicTraceLogWarning(
+                            DatapathCibirWarning,
+                            "[data][%p] CIBIR detected,  %s",
+                            Socket,
+                            "ignoring port collision by assuming some \
+                            other MsQuic CIBIR process has reserved the OS port. \
```
mtfriesen: It's too bad we don't have a way to test if MsQuic with CIBIR is the other endpoint - unless we do?

Could we create a very simple (cooperative, not adversarial) protocol to send a UDP packet to the port and check if the response seems MsQuic-CIBIR-like? I suppose this would introduce a path where the other process has exited, and then we have to retry binding...potentially forever.

src/core/listener.c
```diff
@@ -884,12 +884,18 @@ QuicListenerParamSet(
         Listener->CibirId[0] = (uint8_t)BufferLength - 1;
         memcpy(Listener->CibirId + 1, Buffer, BufferLength);
 
+        uint64_t CibirIdValue = 0;
```
guhetier: nit: Since this is reused, I'd consider extracting it in a small inline helper function in a CIBIR related header. 
It will let you call it directly from the log `QuicTraceLogVerbose(..., GetCibirIdValue(CibirId));`

docs/XDP.md
```diff
@@ -0,0 +1,44 @@
...
+
+For instance: "drop all packets with a UDP header and destination port
+42."
+
+## Port reservation logic
+
+The type of logic MsQuic programs into XDP looks like:
+ "redirect all packets with a destination port X to an AF_XDP socket."
+
+This runs into the issue of **packet stealing.** If say there was an unrelated process
```
guhetier:
```suggestion
This runs into the issue of **packet stealing.** If there was an unrelated process
```

src/platform/datapath_winuser.c
guhetier: nit: imo, the "NO TRAFFIC WILL FLOW ON THIS LISTENER." seems a bit much, especially if we fail.

On the other hand, since CIBIR requires XDP, we should probably fail to create a connection if we can't honor the CIBIR configuration, rather to fallback to the OS stack and ignore the config.

src/platform/datapath_winuser.c
guhetier: Why do we even create those sockets in the first place and have a need to clean them up here?

The function being > 1000 lines long makes resource management very easy to get wrong...

src/platform/datapath_winuser.c
ProjectsByJackHe: I think your suggestion is: when a port collision occurs and CIBIR is enabled, instead of assuming the
  other process that previously reserved the port is also a CIBIR process, we should first do a quick test to
  verify that?
 
If that is indeed the suggestion I would like to offer a little pushback because that seems a bit overkill.
I think it's reasonable to expect apps that enable CIBIR to know what they are doing. 

Moving this discussion offline...

src/core/connection.c
ProjectsByJackHe: QuicTraceLogConnInfo doesn't support byte array types. 

Formatting to a hex string in a local buffer adds `sprintf` overhead.

To deal with the fragility, I have abstracted this logic into its own helper.

src/platform/datapath_winuser.c
ProjectsByJackHe: ACK

src/platform/datapath_winuser.c
ProjectsByJackHe: ACK

src/platform/datapath_winuser.c
ProjectsByJackHe: ACK

src/core/listener.c
ProjectsByJackHe: ACK

docs/images/XDP-arch.png
ProjectsByJackHe: ACK

src/platform/datapath_winuser.c
mtfriesen: Reactivating: my request is more generally to consider whether we can apply any heuristic beyond "it's probably CIBIR" before potentially stealing an OS port.

A concrete suggestion is to follow the existing port reservation model here, but add support for persistent reservations, which can span processes, rather than ephemeral reservations, which are constrained to one process.

src/inc/quic_datapath.h
```diff
@@ -624,6 +625,23 @@ typedef struct CXPLAT_UDP_CONFIG {
...
+// Converts a CIBIR ID byte array to a uint64 for trace logging.
+//
+inline
+uint64_t
+QuicCibirIdToUint64(
+    _In_reads_(Length) const uint8_t* Id,
+    _In_ uint8_t Length
+    )
+{
+    uint64_t Value = 0;
```
guhetier, line 638: please assert that Length <= 8

src/platform/datapath_winuser.c
```diff
@@ -688,6 +688,8 @@ DataPathInitialize(
     }
 #endif
 
+    Datapath->Features |= CXPLAT_DATAPATH_FEATURE_CIBIR;
```
guhetier: Cibir is supported only if XDP is supported, right?

src/platform/datapath_winuser.c
```diff
@@ -1724,6 +1726,46 @@ SocketCreateUdp(
...
+                        // the same UDP and/or TCP port.
+                        //
+                        BOOLEAN XdpAvailable = Datapath->RawDataPath != NULL;
+                        BOOLEAN XdpEnabled = Config->Flags & CXPLAT_SOCKET_FLAG_XDP;
+                        if (!XdpAvailable || !XdpEnabled) {
+                            //
+                            // In the case of a port collision while cibir is enabled,
+                            // MsQuic assumes another MsQuic cibir process has done the
+                            // job of reserving the OS ports via socket creation/bind.
+                            // At least that process can fall back to using the OS stack if
```
guhetier: That comment concerns me: if CIBIR is enabled, we need XDP to honor it. Falling back to OS sockets seems problematic to me, we don't honor an option the app set without reporting it.

src/platform/datapath_winuser.c
guhetier, line 1755: nit: Consider not logging an error when we are actually fine because CIBIR is in use. It will avoid confusion when looking at logs, "fake errors" are misleading.

guhetier:
In my opinion, this entire code path is becoming a maintenance problem (independently from this specific PR). Reviewing changes in a function that long and with multiple level of loops, tests and jump is not reasonable. We need to refactor it, with intentional design.

We cannot expect a user to understand all the special cases and subtly different behavior that have emerged from this code.

src/platform/datapath_winuser.c
guhetier: ... except if you consider that CIBIR can also be used for dispatching between different host, with a load balancer dispatching instead of XDP. But that should not be a transparent fallback.

src/platform/datapath_winuser.c
mtfriesen: Agreed, this falls into my general concern that we don't have clear and robust error detection and error handling here. Our behavior should be clearly defined, with the ability to fill out a matrix of possible actions by MsQuic and non-MsQuic (and CIBIR and non-CIBIR) processes on the system, and the matrix should have reasonable observabvle behavior in each entry.

mtfriesen:
> In my opinion, this entire code path is becoming a maintenance problem (independently from this specific PR). Reviewing changes in a function that long and with multiple level of loops, tests and jump is not reasonable. We need to refactor it, with intentional design.
> 
> We cannot expect a user to understand all the special cases and subtly different behavior that have emerged from this code.

This comment has been made multiple times, and while I agree there is probably room for improvement, as we've discussed, port pools rich with features tend to become intrinsically complicated when properly implemented.

guhetier:
Complex and complicated are different. Portpool are complex, but what we expose to our users or the way we implement it doesn't have to be complicated and confusing. A good implementation can abstract the complexity and present a clear behavior, this is not what we are doing now.

src/platform/datapath_winuser.c
guhetier: @mtfriesen Maybe you can give some more details about how you see things?
Do you mean enforcing a permanent port reservation to exist? Or trying to use a potentially existing one and ignore the failure if it doesn't exist? Something else?
I am not very familiar with permanent port reservations (from the doc, I am somewhat confused about what the token is meant to help with).

mtfriesen: To be super clear, while I provided one concrete suggestion, my request is to demonstrate clear reasoning for whatever we choose to do:

1. Describe the problem (potential port stealing) and reason about its likelihood and impact. Random, hard-to-reproduce failures on 2% of prod machines would be frustrating, as would tests or dev environments randomly failing.
2. Come up with some possible solutions, including existing APIs that might help, or anything else we can implement with reasonable upfront and maintenance cost.
3. Evaluate whether the costs/risks of any of the solutions in (2) would be enough to offset eliminating the expected reliability issues in (1).

It's not obvious to me, after quite a few years in networking at Microsoft, that customers/deployments can be trusted to "know what they're doing" and some rather popular concepts (containers) exist precisely because it's hard to know what you're doing at scale.

src/platform/datapath_winuser.c
ProjectsByJackHe: The raw datapath initialization logic happens after this call. I am not sure I want to refactor that mess in this PR right now.

guhetier: I'm not asking for a refactor now.
But it is strange to me to say that only the winuser datapath supports CIBIR, when it is actually either supported by all datapath, or only when XDP is used.

There might be a better way to control when the test run (this seems to be the only use?)

ProjectsByJackHe:
> Waiting for the engineering analysis of port sharing and port protection options, i.e., a description of how CIBIR ports can and can't be shared across processes, and how that interacts with other CIBIR, other MsQuic, and other unrelated processes; and how we decided on that behavior.

updated the description

src/platform/datapath_winuser.c
ProjectsByJackHe: Discussed and resolved offline

src/platform/datapath_winuser.c
ProjectsByJackHe: discussed offline and resolved.

src/platform/datapath_winuser.c
mtfriesen: I'm not sure this has been resolved offline, and we should summarize any offline discussion in the PR.

ProjectsByJackHe: The discussion was moved to an internal engineering document. The ultimate decision seems to be converging on the approach currently laid out in this PR.

mtfriesen: I sign off on the internal doc; let's summarize the options table (by updating the options table in this PR) and take its proposed action by adding the necessary documentation to recommend safe usage of the CIBIR port sharing feature.

ProjectsByJackHe: Updated CIBIR.md and description.

src/test/lib/TestHelpers.h
```diff
@@ -96,6 +96,78 @@ QuitTestIsFeatureSupported(uint32_t Feature) {
...
+    QuicTestPortReservation& operator=(const QuicTestPortReservation&) = delete;
+
+    //
+    // Releases the port reservation by closing the held socket. Call this
+    // immediately before the listener binds to the same port.
+    //
+    void Release()
+    {
+#ifndef _KERNEL_MODE
+        if (Sock != INVALID_SOCKET) { closesocket(Sock); Sock = INVALID_SOCKET; }
```
guhetier: nit: style, please avoid multiple statements on one line + on the same line as the condition.

src/test/lib/TestHelpers.h
```diff
@@ -96,6 +96,78 @@ QuitTestIsFeatureSupported(uint32_t Feature) {
...
+//
+// Reserves an ephemeral UDP port by binding a socket to port 0. The socket
+// stays open to hold the reservation until Release() is called (or the object
+// is destroyed). This minimizes the TOCTOU window between discovering a free
+// port and having the listener bind to it.
+//
+// In kernel mode, raw socket APIs are unavailable, so this is a no-op
+// (Port stays 0 and the listener falls back to OS-assigned ephemeral ports).
+//
+#ifdef _WIN32
```
guhetier: Rather than having multiple level of #ifdef, pick a single boundary:
- either `QuicTestPortReservation` exists only for Windows UM and you add a #ifdef at the call site
- or `QuicTestPortReservation` always exists and you have a dummy implementation for Windows KM and Linux.

Please don't mix the two. Also, consider having a single #ifdef for the entire struct definition, rather than 3 like currently done. The dummy struct is trivial to implement, it might be less than the 6 lines of defines.

src/test/lib/HandshakeTest.cpp
```diff
@@ -3913,10 +3913,34 @@ QuicTestCibirExtension(
...
     QUIC_ADDRESS_FAMILY QuicAddrFamily = (Family == 4) ? QUIC_ADDRESS_FAMILY_INET : QUIC_ADDRESS_FAMILY_INET6;
     QuicAddr ServerLocalAddr(QuicAddrFamily);
+#ifdef _WIN32
+    QuicTestPortReservation PortReservation(QuicAddrFamily);
+    if (UseDuoNic) {
+        //
+        // CIBIR + XDP requires an explicit local port. Reserve an ephemeral port
+        // up front so the listener always binds to a known port.
+        //
+        if (PortReservation.Port != 0) {
```
guhetier: Why are we not failing / handling the error here when the port is `0`? The comment above just said 0 is not ok for CIBIR + XDP, but the test would proceed with 0.

src/test/lib/HandshakeTest.cpp
```diff
@@ -3913,10 +3913,34 @@ QuicTestCibirExtension(
...
+            ServerLocalAddr.SetPort(PortReservation.Port);
+        }
+    } else {
+        //
+        // If no XDP, CIBIR just falls back to normal OS sockets.
+        // We can just rely upon the Listener created sockets. Passing in
+        // a local port of 0 is fine here.
+        //
+        PortReservation.Release();
```
guhetier: Why do we create a port reservation for nothing? Create it only when `UseDuoNic` is true.

I assume you are trying to work around scopes to use RAII. Use `std::move`, or allow a neutral "unreserved" state to your port reservation.

guhetier: Or just always use the reservation and set a specific port all the time, that also work.

src/platform/datapath_xplat.c
```diff
@@ -160,17 +163,45 @@ CxPlatSocketCreateUdp(
...
+                    Status = QUIC_STATUS_SUCCESS;
                 } else {
                     CxPlatSocketDelete(*NewSocket);
                 }
                 goto Error;
             }
+        } else if (RequiresQtip) {
+            QuicTraceLogWarning(
+                ErrNoXdpForRaw,
+                "[sock] Error: app requested QTIP but XDP not enabled/available/initialized.");
```
guhetier, line 196: `QuicTraceLogWarning` but `[sock] Error`, let's be consistent about what this is.

src/platform/datapath_xplat.c
```diff
@@ -160,17 +163,45 @@ CxPlatSocketCreateUdp(
                     RawSockCreateFail,
                     "[sock] Failed to create raw socket, status:%d", Status);
                 BOOLEAN IsWildcardAddr = Config->LocalAddress == NULL || QuicAddrIsWildCard(Config->LocalAddress);
-                if (IsWildcardAddr && (Config->Flags & CXPLAT_SOCKET_FLAG_QTIP)) {
+                if (IsWildcardAddr && RequiresQtip) {
+                    //
+                    // This retry loop is purely for QTIP listener sockets that try to reserve both a UDP/TCP port,
+                    // which may run into a WSA_ADDR_INUSE for TCP if the UDP ephemeral port collides with something
```
guhetier: Unless `WSA_ADDR_INUSE` is the only possible error, lets be generic. On top of it, this is cross-platform code.
The same port number not being available for both TCP and UDP is a cross platform concept.

docs/CIBIR.md
```diff
@@ -0,0 +1,26 @@
+# CIBIR
+
+## What is it
+
+See [XDP](./XDP.md) first to understand the context.
```
mtfriesen: Same as our internal doc, please lead with a link to the CIBIR draft RFC.

docs/CIBIR.md
```diff
@@ -0,0 +1,26 @@
...
+## CIBIR port sharing logic
+- Applications must provide a well-known local port for server sockets when using CIBIR and XDP.
+- **IMPORTANT:** MsQuic will **NOT** reserve an OS port for server sockets when both CIBIR and XDP is enabled and available.
+    > Client sockets can never share ports, so MsQuic will reserve an OS port in that scenario.
+- The responsbility of book-keeping shared ports and ensuring robust protection for those shared ports is delegated to the application.
+
+
+## Port protection recommendation for shared ports
+
+MsQuic strongly recommends applications leverage the Windows [persistent port reservations API](https://learn.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-createpersistentudpportreservation) to secure shared CIBIR ports prior to serving multi-process CIBIR traffic on a shared port.
```
mtfriesen, line 24: I think we should also mention ALE filters as a viable option - let's just verify with our internal `sockbind` tool that creating an ALE block filter on the CIBIR port prevents regular apps from binding to it.

docs/CIBIR.md
```diff
@@ -0,0 +1,26 @@
...
+- One time setup by a system admin to create the persistent reservation.
+    > A good option for book-keeping persistent port reservations is via registry keys.
+- Persistent port reservations survive reboots, allowing for robust portection in the event of crashes.
```
mtfriesen: "Portection" haha. Let's use "protection" though.

docs/CIBIR.md
```diff
@@ -0,0 +1,26 @@
...
+- Having a persistent reservation makes sure critical ports are taken out of the ephemeral port pool, so an unsuspecting application process won't get accidently assigned an ephemeral port that collides with a CIBIR port.
```
mtfriesen: It's not just taken out of the ephemeral pool, it also prevents specific binds, right?

docs/XDP.md
```diff
@@ -0,0 +1,130 @@
+# MsQuic over XDP
+
+To avoid confusion, "XDP" refers to [XDP-for-windows](https://github.com/microsoft/xdp-for-windows). While Linux XDP has been experimented
+upon in the past and shown some promise for running MsQuic, it is NOT a stable datapath actively being maintained today.
```
mtfriesen: In fact, it's gone!

docs/XDP.md
```diff
@@ -0,0 +1,130 @@
...
+
+There are *exceptions* to this port reservation.
+
+- Sometimes, MsQuic may create a TCP OS socket instead, or both TCP and UDP (see [QTIP](./QTIP.md)).
+- Sometimes, MsQuic may NOT create any OS sockets at all (see [CIBIR](./CIBIR.md)).
+
+
+## MsQuic over XDP general architecture:
+
+```mermaid
```
mtfriesen, line 39: Nice, this mermaid looks really good, huge improvement over checked in image binaries.

src/platform/datapath_xplat.c
mtfriesen: Agreed, this has been a common problem across this PR. Let's let the log levels indicate the log level.

mtfriesen (commented):
This is looking close, aside from the refinements requested by various comments.

docs/CIBIR.md
guhetier: +1, port exclusions prevent ephemeral port assignment, port reservations prevent binding.

src/test/lib/HandshakeTest.cpp
ProjectsByJackHe: the non-zero check is useless because it's under an ifdef. will remove next iteration.

docs/CIBIR.md
ProjectsByJackHe: Yup, sockbind verified that creating an ALE block filter also works. will add docs.

docs/CIBIR.md
ProjectsByJackHe: will specify that it also forbids binds modulo a token.

src/platform/datapath_xplat.c
ProjectsByJackHe: All err logs now do not get the TraceLogWarning level.

src/test/lib/TestHelpers.h
ProjectsByJackHe: added ifdefs at the callsites to enforce this.

src/test/lib/HandshakeTest.cpp
ProjectsByJackHe: untangled the logic a bit more... I went with just using RAII to release the reservation.

docs/CIBIR.md
```diff
@@ -0,0 +1,35 @@
...
+When CIBIR is used, rather than programming [XDP](./XDP.md) to filter and demux packets based on on address and port number,
+XDP with CIBIR will instead filter and de-mux packets based on address, port number, and QUIC connection ID.
+
+What CIBIR allows for is 2 or more separate server processes to share a single
+port on the same machine, as long as their CIBIR ID is different.
+
+## CIBIR port sharing logic
+- Applications must provide a well-known local port for server sockets when using CIBIR and XDP.
+- **IMPORTANT:** MsQuic will **NOT** reserve an OS port for server sockets when both CIBIR and XDP is enabled and available.
+    > Client sockets can never share ports, so MsQuic will reserve an OS port in that scenario.
```
guhetier: nit: It is strange to use `>`, which is markdown for a quote, here.
A `-` or `*` would be more appropriate as a sub-point in a list.
https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#quoting-text

Also, is this correct even taking `QUIC_PARAM_CONN_SHARE_UDP_BINDING` into account?

docs/CIBIR.md
```diff
@@ -0,0 +1,35 @@
...
+- The responsbility of book-keeping shared ports and ensuring robust protection for those shared ports is delegated to the application.
+
+
+## Port protection recommendations for shared ports
+
+### Option 1: Persistent port reservations (Recommended)
+
+MsQuic strongly recommends applications leverage the Windows [persistent port reservations API](https://learn.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-createpersistentudpportreservation) to secure shared CIBIR ports prior to serving multi-process CIBIR traffic on a shared port.
+- One time setup by a system admin to create the persistent reservation.
+    > A good option for book-keeping persistent port reservations is via registry keys.
```
guhetier: nit: Same as above.

I guess Github markdown uses block quotes for alerts, but then I'd add the `[!NOTE]` tag to follow the semantics of md.
https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

src/test/lib/TestHelpers.h
```diff
@@ -96,6 +96,70 @@ QuitTestIsFeatureSupported(uint32_t Feature) {
...
+            int AddrLen = AddrSize;
+            if (getsockname(Sock, (struct sockaddr*)&Addr, &AddrLen) == 0) {
+                Port = QuicAddrGetPort(&Addr);
+            }
+        }
+
+        if (Port == 0) {
+            Release();
+        }
+        UNREFERENCED_PARAMETER(Family);
```
guhetier: Family is referenced.

src/test/lib/TestHelpers.h
```diff
@@ -96,6 +96,70 @@ QuitTestIsFeatureSupported(uint32_t Feature) {
...
+        _In_ QUIC_ADDRESS_FAMILY Family
+        )
+    {
+        int af = (Family == QUIC_ADDRESS_FAMILY_INET) ? AF_INET : AF_INET6;
+
+        Sock = socket(af, SOCK_DGRAM, IPPROTO_UDP);
+        if (Sock == INVALID_SOCKET) { return; }
+
+        QUIC_ADDR Addr;
+        CxPlatZeroMemory(&Addr, sizeof(Addr));
```
guhetier: this is C++:

```suggestion
        QUIC_ADDR Addr{};
```

src/test/lib/TestHelpers.h
```diff
@@ -96,6 +96,70 @@ QuitTestIsFeatureSupported(uint32_t Feature) {
...
+    QuicTestPortReservation() = default;
+
+    QuicTestPortReservation(
+        _In_ QUIC_ADDRESS_FAMILY Family
+        )
+    {
+        int af = (Family == QUIC_ADDRESS_FAMILY_INET) ? AF_INET : AF_INET6;
+
+        Sock = socket(af, SOCK_DGRAM, IPPROTO_UDP);
+        if (Sock == INVALID_SOCKET) { return; }
```
guhetier: nit: style, return should be on the next line

src/test/lib/TestHelpers.h
```diff
@@ -96,6 +96,70 @@ QuitTestIsFeatureSupported(uint32_t Feature) {
...
+    QuicTestPortReservation& operator=(const QuicTestPortReservation&) = delete;
+
+    //
+    // Releases the port reservation by closing the held socket. Call this
+    // immediately before the listener binds to the same port.
+    //
+    void Release()
+    {
+        if (Sock != INVALID_SOCKET) {
+            closesocket(Sock); Sock = INVALID_SOCKET;
```
guhetier: style, one statement per line

src/platform/datapath_xplat.c
```diff
@@ -160,17 +163,45 @@ CxPlatSocketCreateUdp(
...
+                ErrNoXdpForQtip,
+                "[sock] Error: app requested QTIP but XDP not enabled/available/initialized.");
+            CxPlatSocketDelete(*NewSocket);
+            Status = QUIC_STATUS_INVALID_STATE;
+            goto Error;
+        } else if (CibirRequested) {
+             QuicTraceLogWarning(
+                WarnNoXdpForCibirSockets,
+                "[sock] Warning: app requested CIBIR but XDP not enabled/available/initialized. \
+                Falling back to normal OS sockets to allow for CIBIR transport parameter negotiation.");
```
guhetier: The `\` string continuation will include ~16 spaces from the indentation in the middle of the log.

src/platform/datapath_winuser.c
```diff
@@ -1269,19 +1269,52 @@ SocketCreateUdp(
...
+                    "[data][%p] ERROR, %u, %s.",
+                    Socket,
+                    (uint32_t)QUIC_STATUS_INVALID_PARAMETER,
+                    "CIBIR server socket with XDP requires an explicit local port");
+                Status = QUIC_STATUS_INVALID_PARAMETER;
+                goto Error;
+            }
+            QuicTraceLogWarning(
+                DatapathCibirWarning,
+                "[data][%p] CIBIR detected,  %s",
```
guhetier: nit: double space before %s

src/platform/unittest/DataPathTest.cpp
```diff
@@ -848,6 +850,47 @@ TEST_P(DataPathTest, UdpData)
...
+        Server1.CreateUdp(Datapath, &unspecAddress.SockAddr, nullptr, &RecvContext);
+    }
+    VERIFY_QUIC_SUCCESS(Server1.GetInitStatus());
+    ASSERT_NE(nullptr, Server1.Socket);
+
+    //
+    // Try creating a CIBIR-aware socket on the same port.
+    //
+    CxPlatSocket Server2;
+    Server2.CibirIdLength = 8;
```
guhetier: CXPLAT_UDP_CONFIG.CibirId[6] is only 6 bytes, 8 is an invalid value (I'm surprise it doesn't trigger an assert, we should add one).

docs/CIBIR.md
```diff
@@ -0,0 +1,35 @@
...
+XDP with CIBIR will instead filter and de-mux packets based on address, port number, and QUIC connection ID.
+
+What CIBIR allows for is 2 or more separate server processes to share a single
+port on the same machine, as long as their CIBIR ID is different.
+
+## CIBIR port sharing logic
+- Applications must provide a well-known local port for server sockets when using CIBIR and XDP.
+- **IMPORTANT:** MsQuic will **NOT** reserve an OS port for server sockets when both CIBIR and XDP is enabled and available.
+    > Client sockets can never share ports, so MsQuic will reserve an OS port in that scenario.
+- The responsbility of book-keeping shared ports and ensuring robust protection for those shared ports is delegated to the application.
```
guhetier:
```suggestion
- The responsibility of book-keeping shared ports and ensuring robust protection for those shared ports is delegated to the application.
```

docs/CIBIR.md
ProjectsByJackHe: QUIC_PARAM_CONN_SHARE_UDP_BINDING  is for multiple msquic client connections sharing the same client CXPLAT socket, not multiple client CXPLAT sockets sharing the same port.

src/manifest/clog.sidecar
```diff
@@ -4146,6 +4222,20 @@
...
       "macroName": "QuicTraceLogVerbose"
     },
+    "ErrNoXdpForQtip": {
+      "ModuleProperites": {},
+      "TraceString": "[sock] Error: app requested QTIP but XDP not enabled/available/initialized.",
+      "UniqueId": "ErrNoXdpForQtip",
+      "splitArgs": [],
+      "macroName": "QuicTraceLogError"
+    },
+    "ErrNoXdpForRaw": {
```
guhetier: This event seems to be orphaned, it isn't used in the PR. You probably need to delete the sidecar and re-generate it to clear those spurious events.
I'm surprised the CI doesn't complain about it.

guhetier:
The PR description says:

> any failures plumbing xdp rules will bubble up as a socket creation error to the app. Can't fall back to OS sockets.

But errors when setting up XDP rules seems to be silently eaten:

```
   RawSocketCreateUdp                     → returns QUIC_STATUS
     └─ CxPlatTryAddSocket(...)           → returns QUIC_STATUS (checked ✅)
     └─ CxPlatDpRawPlumbRulesOnSocket(..) → void (NOT checked ❌)
          └─ CxPlatDpRawInterfaceAddRules → void, silently fails on alloc failure or rule overflow
               └─ CxPlatDpRawInterfaceUpdateRules → void, has TODO: "Figure out how to better handle failure"
                    └─ XdpCreateProgram → returns QUIC_STATUS (failure logged, then `continue`)
```

When XDP is not backed up with OS sockets, we would silently ignore all traffic. Probably applies to QTIP too.

ProjectsByJackHe:
> The PR description says:
> 
> > any failures plumbing xdp rules will bubble up as a socket creation error to the app. Can't fall back to OS sockets.
> 
> But errors when setting up XDP rules seems to be silently eaten:
> 
> ```
>    RawSocketCreateUdp                     → returns QUIC_STATUS
>      └─ CxPlatTryAddSocket(...)           → returns QUIC_STATUS (checked ✅)
>      └─ CxPlatDpRawPlumbRulesOnSocket(..) → void (NOT checked ❌)
>           └─ CxPlatDpRawInterfaceAddRules → void, silently fails on alloc failure or rule overflow
>                └─ CxPlatDpRawInterfaceUpdateRules → void, has TODO: "Figure out how to better handle failure"
>                     └─ XdpCreateProgram → returns QUIC_STATUS (failure logged, then `continue`)
> ```
> 
> When XDP is not backed up with OS sockets, we would silently ignore all traffic. Probably applies to QTIP too.

Yeah, looking through the code, xdp failures in rule plumbing get silently ignored. This is a latent bug impacting QTIP (now CIBIR as well). Let's create a separate issue for this and address it in a follow up PR. This PR is starting to balloon into a very hard to review size.

src/platform/unittest/DataPathTest.cpp
ProjectsByJackHe: added assert

ProjectsByJackHe:
> The PR description says:
> 
> > any failures plumbing xdp rules will bubble up as a socket creation error to the app. Can't fall back to OS sockets.
> 
> But errors when setting up XDP rules seems to be silently eaten:
> 
> ```
>    RawSocketCreateUdp                     → returns QUIC_STATUS
>      └─ CxPlatTryAddSocket(...)           → returns QUIC_STATUS (checked ✅)
>      └─ CxPlatDpRawPlumbRulesOnSocket(..) → void (NOT checked ❌)
>           └─ CxPlatDpRawInterfaceAddRules → void, silently fails on alloc failure or rule overflow
>                └─ CxPlatDpRawInterfaceUpdateRules → void, has TODO: "Figure out how to better handle failure"
>                     └─ XdpCreateProgram → returns QUIC_STATUS (failure logged, then `continue`)
> ```
> 
> When XDP is not backed up with OS sockets, we would silently ignore all traffic. Probably applies to QTIP too.

https://github.com/microsoft/msquic/issues/5935

src/platform/datapath_xplat.c
```diff
@@ -160,17 +163,45 @@ CxPlatSocketCreateUdp(
...
+                ErrNoXdpForQtip,
+                "[sock] Error: app requested QTIP but XDP not enabled/available/initialized.");
+            CxPlatSocketDelete(*NewSocket);
+            Status = QUIC_STATUS_INVALID_STATE;
+            goto Error;
+        } else if (CibirRequested) {
+             QuicTraceLogWarning(
+                WarnNoXdpForCibirSockets,
+                "[sock] Warning: app requested CIBIR but XDP not enabled/available/initialized. "
+                "Falling back to normal OS sockets to allow for CIBIR transport parameter negotiation.");
```
guhetier: nit: Missing indentation.
```suggestion
             QuicTraceLogWarning(
                 WarnNoXdpForCibirSockets,
                 "[sock] Warning: app requested CIBIR but XDP not enabled/available/initialized. "
                 "Falling back to normal OS sockets to allow for CIBIR transport parameter negotiation.");
```

src/test/lib/HandshakeTest.cpp
```diff
@@ -3913,10 +3913,21 @@ QuicTestCibirExtension(
 
     QUIC_ADDRESS_FAMILY QuicAddrFamily = (Family == 4) ? QUIC_ADDRESS_FAMILY_INET : QUIC_ADDRESS_FAMILY_INET6;
     QuicAddr ServerLocalAddr(QuicAddrFamily);
+#if defined(_WIN32) && !defined(_KERNEL_MODE)
+    QuicTestPortReservation PortReservation(QuicAddrFamily);
```
guhetier, line 3917: nit: Consider asserting `ASSERT_NE(0, PortReservation.Port)` here (and same below) so we get a better error if the test fails because we failed to get a port.

Otherwise, we will fail later with "invalid parameter", which will take longer to triage as a spurious failure.

src/platform/datapath_xplat.c
ProjectsByJackHe: I fixed the indent alignment by moving the signature 'QuicTrace...' back a bit.

src/test/lib/HandshakeTest.cpp
ProjectsByJackHe: added a test_not_equal case here

PR #5799, opened 2026-02-18
Added support for custom OpenSSL selection

### LINKED ISSUE #5614: Allow to use custom OpenSSL version

OpenSSL is a complex dependency, and in many projects it is actually a dependency of multiple packages. As a critical component, it is important to keep OpenSSL up-to-date. Hence, we want each one of our dependencies to link against the same OpenSSL version that we supply, to avoid the danger of one package shipping its own potentially outdated and vulnerable OpenSSL version, and having to wait for each individual package to update its dependencies. This also helps when statically linking dependencies, because you cannot easily statically link two different versions of the same library.

### Proposed solution

I propose that msquic adds a possibility for OpenSSL (and potentially also for other vendored components) for the user to choose how to find OpenSSL:
- Use submodule
- Supply `OPENSSL_LIB_DIR` and `OPENSSL_INCLUDE_DIR` variables
- Use `find_package`

### Additional context

See https://github.com/microsoft/msquic/discussions/5166

toxicteddy00077:
Closes #5614 

Added support for choosing custom OpenSSL version with three prospective methods:
1. Use submodule
2. Use OPENSSL_LIB_DIR and OPENSSL_INCLUDE_DIR variables
3. Use find_package

I mostly built on the suggestion given by the issue author

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,6 +111,9 @@ option(QUIC_CI "CI Specific build" OFF)
 option(QUIC_SKIP_CI_CHECKS "Disable CI specific build checks" OFF)
 option(QUIC_TELEMETRY_ASSERTS "Enable telemetry asserts in release builds" OFF)
 option(QUIC_USE_SYSTEM_LIBCRYPTO "Use system libcrypto if quictls TLS" OFF)
+option(QUIC_USE_EXTERNAL_OPENSSL "Use external OpenSSL instead of building from submodules" OFF)
+set(QUIC_OPENSSL_INCLUDE_DIR "" CACHE PATH "Path to OpenSSL include directory")
+set(QUIC_OPENSSL_LIB_DIR "" CACHE PATH "Path to OpenSSL library directory")
 option(QUIC_HIGH_RES_TIMERS "Configure the system to use high resolution timers" OFF)
 option(QUIC_OFFICIAL_RELEASE "Configured the build for an official release" OFF)
 set(QUIC_FOLDER_PREFIX "" CACHE STRING "Optional prefix for source group folders when using an IDE generator")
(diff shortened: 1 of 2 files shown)


toxicteddy00077:

CMakeLists.txt
```diff
@@ -764,14 +764,36 @@ endif()
 if(QUIC_TLS_LIB STREQUAL "quictls" OR QUIC_TLS_LIB STREQUAL "openssl")
     add_library(OpenSSL INTERFACE)
 
-    include(FetchContent)
+    option(QUIC_USE_EXTERNAL_OPENSSL "Use external OpenSSL instead of building from submodules" OFF)
```
guhetier: Please declare these parameters with the others at the top of the file, so they are easy to find.

CMakeLists.txt
```diff
@@ -764,14 +764,36 @@ endif()
...
+            target_link_libraries(OpenSSLQuic INTERFACE ${LIB_SSL} ${LIB_CRYPTO})
+        else()
+            if(QUIC_TLS_LIB STREQUAL "openssl")
+                find_package(OpenSSL 3.5.0 REQUIRED)
+            else()
+                find_package(OpenSSL REQUIRED)
+            endif()
+            target_link_libraries(OpenSSLQuic INTERFACE OpenSSL::SSL OpenSSL::Crypto)
+        endif()
+        add_library(OpenSSLQuic::OpenSSLQuic ALIAS OpenSSLQuic)
```
guhetier: Consider using a different name for the alias, and aliasing `OpenSSLQuic::OpenSSLQuic` to that name too.
I find it somewhat confusing as it is (unless this is a common Cmake pattern, I am not fluent in Cmake)

CMakeLists.txt
```diff
@@ -764,14 +764,36 @@ endif()
...
     add_library(OpenSSL INTERFACE)
 
-    include(FetchContent)
+    option(QUIC_USE_EXTERNAL_OPENSSL "Use external OpenSSL instead of building from submodules" OFF)
+    set(QUIC_OPENSSL_INCLUDE_DIR "" CACHE PATH "Path to OpenSSL include directory")
+    set(QUIC_OPENSSL_LIB_DIR "" CACHE PATH "Path to OpenSSL library directory")
+
+    if(QUIC_USE_EXTERNAL_OPENSSL OR QUIC_OPENSSL_INCLUDE_DIR OR QUIC_OPENSSL_LIB_DIR)
+        add_library(OpenSSLQuic INTERFACE)
+        if(QUIC_OPENSSL_INCLUDE_DIR AND QUIC_OPENSSL_LIB_DIR)
```
guhetier, line 778: If only one of the two is set, we will go through the `else` path.
We should have an error instead.

CMakeLists.txt
```diff
@@ -764,14 +764,36 @@ endif()
...
+            target_include_directories(OpenSSLQuic INTERFACE ${QUIC_OPENSSL_INCLUDE_DIR})
+            find_library(LIB_CRYPTO NAMES crypto libcrypto PATHS ${QUIC_OPENSSL_LIB_DIR} NO_DEFAULT_PATH REQUIRED)
+            find_library(LIB_SSL NAMES ssl libssl PATHS ${QUIC_OPENSSL_LIB_DIR} NO_DEFAULT_PATH REQUIRED)
+            target_link_libraries(OpenSSLQuic INTERFACE ${LIB_SSL} ${LIB_CRYPTO})
+        else()
+            if(QUIC_TLS_LIB STREQUAL "openssl")
+                find_package(OpenSSL 3.5.0 REQUIRED)
+            else()
+                find_package(OpenSSL REQUIRED)
```
guhetier: When the TLS provider is quictls, we need to link QuicTLS version of libssl statically.
It is a fork of OpenSSL with support for QUIC, OpenSSL prior to 3.5 doesn't expose the required APIs, so dynamically linking to the system libssl won't work.

CMakeLists.txt
toxicteddy00077: Can't say im sure about the naming convention either but for the sake of brevity ive kept it as `OpenSSL::MsQuic`

CMakeLists.txt
toxicteddy00077: Alright ive added a condition for this

CMakeLists.txt
toxicteddy00077: done

CMakeLists.txt
toxicteddy00077: done

masa-koz:
@toxicteddy00077 Hi, I found a missing entry for Win32 in submodules/CMakeLists.txt.

```
--- a/submodules/CMakeLists.txt
+++ b/submodules/CMakeLists.txt
@@ -165,6 +165,7 @@ if (WIN32)
     )

     add_library(OpenSSLQuic::OpenSSLQuic ALIAS OpenSSLQuic)
+    add_library(OpenSSL::MsQuic ALIAS OpenSSLQuic)
```

toxicteddy00077:
Thanks @masa-koz good catch, @guhetier please check now

CMakeLists.txt
guhetier: Why do we need two aliases now?

What I meant previously is that after all that code, we will have one of multiple OpenSSL library added. We should ensure they are all aliased to a unique name, so that the rest of the CMake code can refer to them easily.
To avoid confusion, the unique name should not favor one particular way of getting OpenSSL.

`OpenSSL::MsQuic` also seems reversed, shouldn't it be `MsQuic::OpenSSL` to make clarify this is the OpenSSL library used in MsQuic, not the MsQuic library used in OpenSSL?

toxicteddy00077: You're right, I wasnt sure about completely removing the old name, but its neater to keep just one. I've made the refactoring

CMakeLists.txt
```diff
@@ -759,18 +767,40 @@ endif()
...
+    if(QUIC_TLS_LIB STREQUAL "quictls" AND (QUIC_USE_EXTERNAL_OPENSSL OR QUIC_OPENSSL_INCLUDE_DIR OR QUIC_OPENSSL_LIB_DIR))
+        message(FATAL_ERROR "Manual OpenSSL selection is not supported with the 'quictls'. 'quictls' must be built from submodules to ensure static linkage with required QUIC APIs. Use 'openssl' provider for external 3.5.0+ OpenSSL.")
+    endif()
+
+    # Only 'openssl' provider supports external selection for now
+    if(QUIC_TLS_LIB STREQUAL "openssl" AND (QUIC_USE_EXTERNAL_OPENSSL OR QUIC_OPENSSL_INCLUDE_DIR OR QUIC_OPENSSL_LIB_DIR))
+        add_library(OpenSSLQuic INTERFACE)
+        if(QUIC_OPENSSL_INCLUDE_DIR AND QUIC_OPENSSL_LIB_DIR)
+            target_include_directories(OpenSSLQuic INTERFACE ${QUIC_OPENSSL_INCLUDE_DIR})
+            find_library(LIB_CRYPTO NAMES crypto libcrypto PATHS ${QUIC_OPENSSL_LIB_DIR} NO_DEFAULT_PATH REQUIRED)
```
guhetier: `REQUIRED` doesn't work with `find_library` until CMake 3.18 and we still allow 3.16.
The result must be manually validated.

See above:
```
find_library(ATOMIC NAMES atomic libatomic.so.1)
    if (ATOMIC)
        message(STATUS "Found libatomic: ${ATOMIC}")
    else()
        message(STATUS "libatomic not found. If build fails, install libatomic")
    endif()
```

toxicteddy00077: Changes made @guhetier

PR #5807, opened 2026-02-21
fix(forwarder): null-check allocations and fix resource leaks in quicforward

### LINKED ISSUE #5806: Null pointer dereference on allocation failure in quicforward
### Describe the bug

In src/tools/forwarder/quicforward.cpp, three allocation failure paths 
are unhandled:

1. In ConnectionCallback (PEER_STREAM_STARTED), MsQuicStream objects are 
allocated with new(std::nothrow) but never null-checked before being 
dereferenced. On allocation failure, PeerStream->Context crashes with a 
null pointer dereference. Additionally, if LocalStream allocation fails, 
the raw stream handle from Event->PEER_STREAM_STARTED.Stream is never 
closed, leaking it.

2. In ListenerCallback (NEW_CONNECTION), MsQuicConnection objects are 
allocated with new(std::nothrow) but never null-checked before being 
dereferenced. On allocation failure, BackEndConn->Context crashes with a 
null pointer dereference.

3. In StreamCallback (RECEIVE), if StreamSend returns an unexpected error 
status, CXPLAT_FRE_ASSERT fires without first freeing SendContext, causing 
a resource leak before process termination.

The correct null-check pattern already exists in MsQuicAutoAcceptListener 
in msquic.hpp and is not consistently applied in quicforward.cpp.

### Affected OS

- [x] Windows
- [x] Linux
- [x] macOS
- [ ] Other (specify below)

### Additional OS information

_No response_

### MsQuic version

main

1. Run the quicforward tool to establish an active forwarding session.
2. Trigger an OOM condition during stream or connection setup to cause 
   new(std::nothrow) to return null in ConnectionCallback or 
   ListenerCallback.
3. For the third bug: cause StreamSend to return a failure status other than 
   QUIC_STATUS_ABORTED or QUIC_STATUS_INVALID_STATE during an active 
   forwarding session.

Note: these paths require OOM conditions or fault injection to trigger 
reliably and cannot be easily reproduced without a dedicated test harness.

### Expected behavior

Allocation failures are handled gracefully. On stream allocation failure,  the inbound stream handle is closed cleanly and the connection continues. On connection allocation failure, QUIC_STATUS_OUT_OF_MEMORY is returned and MsQuic refuses the connection cleanly. On unexpected send failure, SendContext is freed before process termination.

### Actual outcome

Null pointer dereference and crash when new(std::nothrow) returns null in ConnectionCallback or ListenerCallback. Resource leak before process termination when StreamSend returns an unexpected error in StreamCallback.

### Additional details

Root cause traced across three files:

- quicforward.cpp: allocation results used without null checks
- msquic.hpp: MsQuicStream and MsQuicConnection constructors have no 
  internal null guard on the outer pointer; destructor calls StreamClose/
  ConnectionClose, so delete on a successfully constructed wrapper will 
  close the handle — manual close only needed when new returns null
- listener.c: confirmed that returning QUIC_STATUS_OUT_OF_MEMORY from 
  ListenerCallback is safe — MsQuic refuses the connection and retains 
  handle ownership on failure returns
- spinquic.cpp: confirms that PEER_STREAM_STARTED requires explicit 
  StreamClose on rejection, not a bare break

I created a fix and will be submit it as a PR referencing this issue.

MarkedMuichiro:
## Description
Fixes #5806.

In quicforward.cpp, MsQuicStream and MsQuicConnection objects are allocated with new(std::nothrow) in ConnectionCallback and ListenerCallback, but the results are never null-checked before being dereferenced. Under OOM conditions this causes a null pointer dereference and crash. The correct guard pattern already exists in MsQuicAutoAcceptListener in msquic.hpp and is applied here consistently.

Ownership notes:
- ListenerCallback returns QUIC_STATUS_OUT_OF_MEMORY on allocation failure.
  Confirmed via listener.c that failure returns are permitted — MsQuic
  refuses the connection and retains handle ownership, so no manual cleanup
  of the raw connection handle is needed.
- ConnectionCallback explicitly closes the raw stream handle on allocation
  failure, matching the pattern in spinquic.cpp. Sequential allocation is
  intentional — MsQuicConnection's constructor calls SetCallbackHandler
  immediately, so a consolidated null-check would risk closing a handle
  not yet owned by the caller.

## Testing
No existing tests cover these OOM paths. Built successfully on Linux (GitHub Codespaces) and validated normal forwarding behavior is unaffected. Reliably triggering these paths requires fault injection that does not currently exist for this code. Happy to add coverage if maintainers have a preferred approach.

## Documentation
None.

### DIFF
--- a/src/tools/forwarder/forwarder.cpp
+++ b/src/tools/forwarder/forwarder.cpp
@@ -207,7 +207,9 @@ QUIC_STATUS ListenerCallback(
     if (Event->Type == QUIC_LISTENER_EVENT_NEW_CONNECTION) {
         auto BackEndConn = new(std::nothrow) MsQuicConnection(*Registration, CleanUpAutoDelete, ConnectionCallback);
         if (!BackEndConn || !BackEndConn->IsValid()) {
-            if (BackEndConn) BackEndConn->Close();
+            if (BackEndConn) {
+                BackEndConn->Close();
+            }
             return QUIC_STATUS_OUT_OF_MEMORY;
         }
         auto FrontEndConn = new(std::nothrow) MsQuicConnection(Event->NEW_CONNECTION.Connection, CleanUpAutoDelete, ConnectionCallback, BackEndConn);


src/tools/forwarder/forwarder.cpp
```diff
@@ -119,6 +119,10 @@ QUIC_STATUS StreamCallback(
             ForwardedSend::Delete(SendContext);
             return QUIC_STATUS_SUCCESS;
         }
+        // Any other failure is an unexpected invariant violation — free before fatal assert.
+        if (QUIC_FAILED(Status)) {
```
guhetier: We assert success on the next line, which implies this new code path is unreachable.
We also don't need to cleanup resources manually since we are about to terminate the process.

MarkedMuichiro: Good point, I'll remove the QUIC_FAILED block.

src/tools/forwarder/forwarder.cpp
```diff
@@ -171,7 +171,16 @@ QUIC_STATUS ConnectionCallback(
     case QUIC_CONNECTION_EVENT_PEER_STREAM_STARTED: {
         //printf("c[%p] Peer stream started\n", Connection);
         auto PeerStream = new(std::nothrow) MsQuicStream(*PeerConn, Event->PEER_STREAM_STARTED.Flags, CleanUpAutoDelete, StreamCallback);
+        if (!PeerStream) {
+            MsQuic->StreamClose(Event->PEER_STREAM_STARTED.Stream);
+            return QUIC_STATUS_SUCCESS;
+        }
         auto LocalStream = new(std::nothrow) MsQuicStream(Event->PEER_STREAM_STARTED.Stream, CleanUpAutoDelete, StreamCallback, PeerStream);
+        if (!LocalStream) {
+            delete PeerStream;
```
guhetier: Actually, on a second read, I think this will be a problem.
`PeerStream` was created with the `CleanUpAutoDelete` flag, so it will be freed when the stream is closed as part of the destructor, and then again by delete.

We should call `LocalStream.Close()` here and let it take care of the cleanup.
There is a similar problem for all the `delete` on object created with the `CleanUpAutoDelete` flag.

src/tools/forwarder/forwarder.cpp
```diff
@@ -171,7 +171,16 @@ QUIC_STATUS ConnectionCallback(
     case QUIC_CONNECTION_EVENT_PEER_STREAM_STARTED: {
         //printf("c[%p] Peer stream started\n", Connection);
         auto PeerStream = new(std::nothrow) MsQuicStream(*PeerConn, Event->PEER_STREAM_STARTED.Flags, CleanUpAutoDelete, StreamCallback);
+        if (!PeerStream) {
+            MsQuic->StreamClose(Event->PEER_STREAM_STARTED.Stream);
+            return QUIC_STATUS_SUCCESS;
+        }
         auto LocalStream = new(std::nothrow) MsQuicStream(Event->PEER_STREAM_STARTED.Stream, CleanUpAutoDelete, StreamCallback, PeerStream);
+        if (!LocalStream) {
```
guhetier, line 182: On top of checking the allocation, if we want to handle errors properly, we should check `IsValid`, otherwise the internal handle could be `NULL` when the initialization fails.

src/tools/forwarder/forwarder.cpp
MarkedMuichiro: Good catches — I replaced `delete` with `Close()` for all `CleanUpAutoDelete` objects and add `IsValid()` checks after allocation. Thank you.

src/tools/forwarder/forwarder.cpp
```diff
@@ -171,7 +171,17 @@ QUIC_STATUS ConnectionCallback(
     case QUIC_CONNECTION_EVENT_PEER_STREAM_STARTED: {
         //printf("c[%p] Peer stream started\n", Connection);
         auto PeerStream = new(std::nothrow) MsQuicStream(*PeerConn, Event->PEER_STREAM_STARTED.Flags, CleanUpAutoDelete, StreamCallback);
+        if (!PeerStream || !PeerStream->IsValid()) {
+            if (PeerStream) PeerStream->Close();
```
guhetier:
```suggestion
            if (PeerStream) {
                PeerStream->Close();
            }
```

Please always use braces on if tests, for consistency and as a less error prone style

MarkedMuichiro:
> Looks good to me, except some nit style.

Done. Thank you!

src/tools/forwarder/forwarder.cpp
```diff
@@ -194,7 +206,15 @@ QUIC_STATUS ListenerCallback(
 {
     if (Event->Type == QUIC_LISTENER_EVENT_NEW_CONNECTION) {
         auto BackEndConn = new(std::nothrow) MsQuicConnection(*Registration, CleanUpAutoDelete, ConnectionCallback);
+        if (!BackEndConn || !BackEndConn->IsValid()) {
+            if (BackEndConn) BackEndConn->Close();
```
guhetier: I should have mentioned, there are a few occurences.

MarkedMuichiro: That was my bad for not double checking. It should be good now. Thank you.

Issue #5479, opened 2025-10-02
Scorecard validation is failing since March 2024


guhetier:
https://github.com/microsoft/msquic/security/code-scanning/tools/Scorecard/status/configurations/actions-FZTWS5DIOVRC653POJVWM3DPO5ZS643DN5ZGKY3BOJSHGLLBNZQWY6LTNFZS46LNNQ

> 2025/10/01 21:48:49 scorecard had an error: repo unreachable: GET https://api.github.com/repos/microsoft/msquic: 401 Bad credentials []

This should be fixed or fully disabled.

PR #5808, opened 2026-02-23
Disable the scorecard analysis run temporarily

guhetier:
## Description

The Scorecard analysis workflow systematically fails because of an error 401. It generates noise and makes it harder to check if something else went wrong in the CI run: the main branch always shows a failed status (https://github.com/microsoft/msquic/commits/main/).

This PR disable the run.
Fixing the permission is already tracked by #5479

## Testing

CI

## Documentation

N/A

### DIFF
--- a/.github/workflows/scorecards-analysis.yml
+++ b/.github/workflows/scorecards-analysis.yml
@@ -1,10 +1,10 @@
 name: Scorecards supply-chain security
 
-on: workflow_dispatch
-#  push:
-#    branches: [ main ]
-#  pull_request:
-#    branches: [ main ]
+on:
+  push:
+    branches: [ main ]
+  pull_request:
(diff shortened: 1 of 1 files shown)


guhetier:
@copilot Remove the scorecard analysis completely, cleaning up after yourself.

guhetier:
While trying to get a better idea whether this is a problem to remove fully, I saw the doc seems to say the token is needed only for a corner case:

https://github.com/ossf/scorecard-action/blob/main/docs/authentication/fine-grained-auth-token.md
> For repositories that want to detect their classic Branch Protection rules

Maybe just deleting it will fix it. Otherwise, we can delete the workflow: scanning happens anyway, the action helps the maintainer of ossf to spend less resources.

PR #5810, opened 2026-02-24
Multiple Quic listeners on same port

DennySun2020:
## Description

This change is to fix this issue https://github.com/microsoft/msquic/issues/5806

## Testing

A test case is added to cover this change

## Documentation

No


guhetier:
The issue referenced in the PR seems to be the wrong one.
Work is already ongoing for this issue in #5798.

@ProjectsByJackHe FYI

PR #5811, opened 2026-02-24
Add the `/utf-8` flag to the compiler options

masa-koz:
## Description

Some files include non-ASCII (UTF-8) characters. So, we have to make sure that source files are interpreted as UTF-8 by adding the `/utf-8` flag to the compiler options for Windows builds.

## Testing

CI

## Documentation

No

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -234,7 +234,7 @@ include(GNUInstallDirs)
 
 if (WIN32)
     set(QUIC_WARNING_FLAGS /WX /W4 /sdl /wd4206 CACHE INTERNAL "")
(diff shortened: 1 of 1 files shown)


guhetier:
Thanks for the PR.
All the utf-8 characters I find are in comments, where it should not be a problem. Was the character set an issue for you?
The option is probably good to have anyway.

masa-koz:
@guhetier Yes, in my environment the default code page is CP932 and the utf-8 characters will be interpreted as Shift-JIS without "/utf-8" option and it will produce the C4819 warning. With "/utf-8" option, I see no warning.

PR #5816, opened 2026-02-25
Enable LTTng logging with sanitizers and fix TlsTest leak

leikong:
Remove the guard that blocked LTTng logging when sanitizers are active. The incompatibility claim (from 2020) does not seem to be an issue anymore, all tests pass with LTTng+ASAN enabled, verified with both gcc and clang.

Also fix a memory leak in TlsTest::TearDownTestSuite where CaSelfSignedCertParams and CaClientCertParams were allocated via CxPlatGetSelfSignedCert but never freed.


leikong:
> Thanks for the PR, looks good to me

Thanks for the review.
Just as FYI, I don't have permission to merge, so would you please merge it for me when you guys are ready to take this?

PR #5846, opened 2026-03-04
Update DotNet SDK version from 10.0.101 to 10.0.103

guhetier:
## Description

Component governance requires a DotNet SDK upgrade due to a security patch in 10.0.103.
https://dotnet.microsoft.com/en-us/download/dotnet/10.0

## Testing

CI

## Documentation

N/A

### DIFF
--- a/.github/workflows/check-clog.yml
+++ b/.github/workflows/check-clog.yml
@@ -22,7 +22,9 @@ jobs:
     - name: Setup .NET
       uses: actions/setup-dotnet@baa11fbfe1d6520db94683bd5c7a3818018e4309
       with:
-        dotnet-version: 8.0.x
+        dotnet-version: |
+          10.0.x
+          8.0.x
     - name: Checkout repository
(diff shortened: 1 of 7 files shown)


.github/workflows/check-clog.yml
```diff
@@ -22,7 +22,9 @@ jobs:
     - name: Setup .NET
       uses: actions/setup-dotnet@baa11fbfe1d6520db94683bd5c7a3818018e4309
       with:
-        dotnet-version: 8.0.x
+        dotnet-version: |
+          10.0.x
```
mtfriesen, line 26: Why do we keep these older versions around? Can we just put `10.0.x` everywhere?

guhetier: When I tried at first to just upgrade, it broke - I think because some pipeline need one version to build / package the binary, and must run it on 10.0.103.

We can probably clean some (maybe also upgrading the .Net version used for some of the tools). For now, I was trying to get it to build and I don't have good ways of validating it outside of running the pipeline.

I'll likely try to get copilot to iterate on another branch to clean some up.

Issue #4870, opened 2025-02-27
MsQuic needs to use zero-initialized pool by default


mtfriesen:
### Describe the bug

Windows requires zero-initialized pool allocations by default, with uninitialized allocations explicitly opt-in on a case-by-case basis. MsQuic should conform with these security requirements.

### Affected OS

- [x] Windows
- [x] Linux
- [x] macOS
- [ ] Other (specify below)

### MsQuic version

main

### Steps taken to reproduce bug

Use `CxPlatProcessorInfo` variable to interact with OS.

### Expected behavior

The OS accepts the `CxPlatProcessorInfo` as valid parameters.

### Actual outcome

The OS rejects `CxPlatProcessorInfo` because the `PADDING` field, which maps to the OS `Reserved` field, is not zero.

ProjectsByJackHe:
Seems like a patch was already made in the past: https://github.com/microsoft/msquic/pull/4871 as a sort of "band-aid".

ProjectsByJackHe:
Fixed in #5855

PR #5855, opened 2026-03-10
Use zero-init pools

ProjectsByJackHe:
## Description

Following up on https://github.com/microsoft/msquic/pull/4871 to introduce a broader fix to address https://github.com/microsoft/msquic/issues/4870.

Makes pool allocs zero-init by default in user mode.

## Testing

Added an alloc test to test zero-inits.

## Documentation

N/A

### DIFF
--- a/src/platform/datapath_winkernel.c
+++ b/src/platform/datapath_winkernel.c
@@ -2606,6 +2606,10 @@ CxPlatSendDataAllocDataBuffer(
     _In_ CXPLAT_POOL* BufferPool
     )
 {
+    //
+    // Use the *PoolAllocUninitialized method here to preserve the MDL
+    // pre-initializations (the default *PoolAlloc method zero-inits).
+    //
     CXPLAT_DATAPATH_SEND_BUFFER* SendBuffer = CxPlatPoolAllocUninitialized(BufferPool);
     if (SendBuffer == NULL) {
         return NULL;


src/inc/quic_platform_posix.h
```diff
@@ -367,6 +367,13 @@ CxPlatAlloc(
     _In_ uint32_t Tag
     );
 
+_Ret_maybenull_
+void*
+CxPlatAllocUninitialized(
```
guhetier, line 372: nit: Consider `CxPlatAllocNonZeroed` instead, there is a small distinction between initializing something and zeroing it (typically if zero is not a valid value for the type).

mtfriesen:
Should we also do the same for kernel mode?

src/inc/quic_platform_winuser.h
```diff
@@ -448,7 +456,9 @@ CxPlatPoolAlloc(
     Header->SpecialFlag = CXPLAT_POOL_ALLOC_FLAG;
 #endif
     Header->Owner = Pool;
-    return (void*)(Header + 1);
+    void* Result = (void*)(Header + 1);
+    RtlZeroMemory(Result, Pool->Size - sizeof(CXPLAT_POOL_HEADER));
```
guhetier, line 460: Question: Any idea if we should be concerned about the perf impact of this?

Especially for the datapath and data packet allocation, should we already move to the non-zeroed helpers?

src/platform/unittest/AllocTest.cpp
```diff
@@ -0,0 +1,141 @@
...
+//
+#define TEST_ALLOC_SIZE 256
+
+static bool
+IsZeroMemory(
+    _In_reads_(Size) const uint8_t* Buffer,
+    _In_ size_t Size
+    )
+{
+    for (size_t i = 0; i < Size; i++) {
```
guhetier: nit: You have access to C++ and the STL here.
`std::all_of(Buffer, Buffer + Size, [](auto i) { return i == 0 });` + make `Buffer` a `std::span` if the C++ standard used allows it.

guhetier: The same applies to the entire file - when coding in C++ user-mode, let's try to make it idiomatic.

src/platform/unittest/AllocTest.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+--*/
+
+#include "main.h"
+
+#ifdef QUIC_CLOG
+#include "AllocTest.cpp.clog.h"
+#endif
+
+#ifndef _KERNEL_MODE
```
guhetier: Unit tests are not compiled in kernel mode anyway, we don't need this.

src/inc/quic_platform_posix.h
mtfriesen: I think we're just using the terms aligned with `ExAllocatePoolUninitialized` / `ExAllocatePoolZero` here. IMO it's better to align to existing well-defined precedent than to try setting a slightly better precedent.

In other words, the principle of least surprise is that `CxPlatAllocUninitialized` calls existing function `ExAllocatePoolUninitialized`.

src/inc/quic_platform_winuser.h
mtfriesen: I'm curious how much a regression the netperf results show with the overhead of zeroing each packet - it shouldn't be a trivial cost.

src/inc/quic_platform_posix.h
guhetier: If we have precedents, fine by me. I was thinking of the pool flags (HEAP_ZERO_MEMORY)

src/platform/unittest/AllocTest.cpp
ProjectsByJackHe: sure, refactored to use std::all_of

src/inc/quic_platform_winuser.h
ProjectsByJackHe: From looking at the results in https://github.com/microsoft/netperf/actions/runs/22923733783, and focusing purely on QUIC + iocp, and comparing it to the results on the [dashboard ](https://microsoft.github.io/netperf/dist/)from the last run (on Azure VMs): 

Slight regression, not too bad. Barely a rounding error.


UPLOAD:
(this PR)
image

(dashboard)
image

DOWNLOAD:
(this PR)
image

(dashboard)
image

Latency looks to be about the same...
RPS is slightly lower (136K on the dashboard vs. ~130K average on the PR run)

ProjectsByJackHe:
> Should we also do the same for kernel mode?

I think adding it for Kernel makes sense. Added it.

src/inc/quic_platform_winuser.h
guhetier: > Slight regression, not too bad. Barely a rounding error.

It is nice to see this is not a drastic change. It would be good to have better data though (without error bars / variance, CPU usage, etc... this is not very significant data).

I still think that we should consider not zeroing the payload buffers directly in this PR. That is the one place in the codebase where I would advocate for not initializing a buffer before use!

mtfriesen: Conceptually I agree, but if the data show uninitialized memory is a trivial perf gain for a nontrivial risk of memory disclosure, I'm happy to both simplify our code and make it more secure - let's just zero!

src/platform/datapath_winkernel.c
```diff
@@ -2606,7 +2606,7 @@ CxPlatSendDataAllocDataBuffer(
     _In_ CXPLAT_POOL* BufferPool
     )
 {
-    CXPLAT_DATAPATH_SEND_BUFFER* SendBuffer = CxPlatPoolAlloc(BufferPool);
+    CXPLAT_DATAPATH_SEND_BUFFER* SendBuffer = CxPlatPoolAllocUninitialized(BufferPool);
```
mtfriesen, line 2613: Why did we decide here to go with the Uninitialized variant?

ProjectsByJackHe: in the winkernel datapath, it has it's own custom allocator that pre-initializes MDLs. Zeroing them out overwrites those inits.

mtfriesen: Ah, that explains why the CI failed. Please add a code comment here since that's a perfectly reasonable justification for "uninitialized", though arguably this falls into the claim of being very carefully initialized, just not zero'd.

ProjectsByJackHe: Added.

PR #5882, opened 2026-03-20
Fix use-after-free in QuicRangeRemoveRange middle-split path

### LINKED ISSUE #5824: Potential use-after-free in QuicRangeRemoveRange middle-split path.
At line 435 in range.c, `*NewSub = *Sub` may read from freed memory when:
1. SubRanges array is at capacity (`UsedLength == AllocLength`)
2. `AllocLength > QUIC_RANGE_INITIAL_SUB_COUNT` (so old array is heap-allocated)
3. Middle-split requires `MakeSpace -> Grow`, which frees old `SubRanges`
4. `Sub` (captured before `MakeSpace` at line 415) becomes dangling

**This bug is latent:** freed memory typically still contains old data on
standard heaps, masking the UB. It would manifest under ASan, Page Heap,
or memory pressure. A proper fix: capture Sub's values (Low/Count) before
calling MakeSpace, then assign them to NewSub.

**Reproducing Scenario:** Fill 16 subranges (AllocLength=16, heap-allocated), then split one
  via RemoveRange to trigger growth from 16 to 32. Verify data integrity.
**How:** Add [100..199] and 15 single values. Remove [140..159] from [100..199].
**Assertions:** Split produces [100..139] and [160..199], ValidCount == 17,
  removed values are actually gone.

### Affected OS

- [x] Windows
- [ ] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

_No response_

### MsQuic version

main

1. Add the following test in `src\core\unittest\RangeTest.cpp`
```cpp
TEST(DeepTest_Range, RemoveRangeMiddleSplitWithGrow)
{
    SmartRange range;

    // Add a range subrange that will be split
    range.Add(100, 100);  // [100..199], will be at index 0 (sorted)

    // Add 15 more single-value subranges to fill to 16 total
    // These have higher values so [100..199] stays at index 0
    for (uint32_t i = 1; i <= 15; i++) {
        range.Add(1000 + i * 100);
    }
    ASSERT_EQ(range.ValidCount(), 16u);
    ASSERT_EQ(range.Min(), 100ull);

    // Verify [100..199] is present
    uint64_t GetCount = 0;
    BOOLEAN IsLastRange = FALSE;
    ASSERT_EQ(QuicRangeGetRange(&range.range, 100, &GetCount, &IsLastRange), TRUE);
    ASSERT_EQ(GetCount, 100ull);

    // Now remove [140..159] from the middle of [100..199].
    // This requires splitting into [100..139] and [160..199], needing 17 slots.
    // AllocLength is 16, so MakeSpace calls Grow (16 -> 32).
    // Grow frees the old 16-element array -> Sub becomes dangling.
    // *NewSub = *Sub reads from freed memory (use-after-free).
    BOOLEAN result = QuicRangeRemoveRange(&range.range, 140, 20);
    ASSERT_EQ(result, TRUE);

    // After the split, we should have 17 subranges:
    // [100..139], [160..199], and the 15 single-value subranges
    ASSERT_EQ(range.ValidCount(), 17u);

    // Verify the split produced correct subranges
    // Due to the use-after-free bug, these may fail with corrupted values
    auto sub0 = QuicRangeGet(&range.range, 0);
    ASSERT_EQ(sub0->Low, 100ull);
    ASSERT_EQ(sub0->Count, 40ull);   // [100..139]

    auto sub1 = QuicRangeGet(&range.range, 1);
    ASSERT_EQ(sub1->Low, 160ull);
    ASSERT_EQ(sub1->Count, 40ull);   // [160..199]

    // Verify overall range integrity
    ASSERT_EQ(range.Min(), 100ull);
    ASSERT_EQ(range.Max(), 2500ull);

    // Verify the removed values are actually gone
    uint64_t Dummy;
    BOOLEAN DummyBool;
    ASSERT_EQ(QuicRangeGetRange(&range.range, 140, &Dummy, &DummyBool), FALSE);
    ASSERT_EQ(QuicRangeGetRange(&range.range, 150, &Dummy, &DummyBool), FALSE);
    ASSERT_EQ(QuicRangeGetRange(&range.range, 159, &Dummy, &DummyBool), FALSE);
}
```
2. Build in `Debug` mode, and test with `-CodeCoverage`. The test fails with assertion error.

**[Note]** This failure can only be reproduced when tested with `-CodeCoverage`.

### Expected behavior

None of the assertions in the test should fail, even with `-CodeCoverage`.

### Actual outcome

The assertions following the block 
```cpp
    // Verify the split produced correct subranges
    // Due to the use-after-free bug, these may fail with corrupted values
    ...
``` 
in the test fails.

### Additional details

_No response_

guhetier:
## Description

Fix use-after-free in `QuicRangeRemoveRange` (`src/core/range.c`). In the middle-split path, the `Sub` pointer captured before `QuicRangeMakeSpace` becomes dangling when `MakeSpace` triggers `QuicRangeGrow`, which frees the old heap-allocated `SubRanges` array. The subsequent `*NewSub = *Sub` reads from freed memory.

**Trigger conditions** (all required):
1. `UsedLength == AllocLength` (array at capacity)
2. `AllocLength > QUIC_RANGE_INITIAL_SUB_COUNT` (8)  old array is heap-allocated
3. Middle-overlap case in `RemoveRange`

**Fix:** The function is unused: delete it, and adapt the tests to reflect the code flow used in the product.

Fixes #5824

## Testing

Removed or adapted tests relying on the function.

## Documentation

No documentation impact.


ProjectsByJackHe (dismissed):

src/core/unittest/RangeTest.cpp
```diff
@@ -787,3 +787,33 @@ TEST(RangeTest, SearchRangeThree)
...
+
+    //
+    // Fill to AllocLength=16 (heap-allocated) so a middle-split
+    // triggers Grow, which frees the old SubRanges array.
+    //
+    range.Add(100, 100);  // [100..199] at index 0, will be split
+    for (uint32_t i = 1; i <= 15; i++) {
+        range.Add(1000 + i * 100);
+    }
+    ASSERT_EQ(range.ValidCount(), 16u);
```
mtfriesen: I find AI is a little too happy to copy/paste or otherwise create fragile dependencies when more robust abstractions are readily available. Couldn't this function at least rely on `QUIC_RANGE_INITIAL_SUB_COUNT` before making the assumption / implementation detail dependency that triggering a heap alloc will double the length? IMO since this unit test is effectively a white box test, peeking directly at the AllocLength field after the first `QUIC_RANGE_INITIAL_SUB_COUNT + 1` entries are added would be a little better, so we can tweak/tune the allocation growth/reduction strategy in the future without silently breaking the test.

mtfriesen (dismissed):

src/core/unittest/RangeTest.cpp
guhetier: Good point, I'll improve that.

I agree with the AI tendency, maybe that something we can try to tune with better instructions...

mtfriesen (dismissed):

mtfriesen (approved):

mtfriesen:
Note: description should be updated with the actual fix.

PR #5887, opened 2026-03-22
[DeepTest] Add unit tests for BBR congestion control

saikat107:
## Description

Adds comprehensive unit tests for the BBR congestion control implementation (`src/core/bbr.c`), achieving **100% line coverage** (484/484 lines) across 75 tests.

### Test Architecture

Tests use a `BbrTest_DeepTest` GoogleTest fixture that constructs a minimal mock `QUIC_CONNECTION` and directly initializes the BBR congestion control vtable — no network, no TLS, no threading. This enables deterministic, sub-millisecond testing of the full BBR state machine.

The test suite is organized into three tiers:

| Section | Count | Description |
|---------|-------|-------------|
| **Spec-conformance** | 24 | Validates behavior defined by the BBR draft (state machine transitions, recovery, MinRtt expiration, BtlbwFound detection). Would survive any conforming reimplementation. |
| **Loosely-coupled** | 16 | Exercises MsQuic-specific features (pacing API, exemptions, blocked state, statistics) through the CC vtable only. Survives internal refactors. |
| **Tightly-coupled** | 35 | Directly asserts internal `QUIC_CONGESTION_CONTROL_BBR` struct fields (CW formulas, pacing gain arithmetic, bandwidth filter, ack aggregation, SendQuantum). Validates implementation correctness at the cost of coupling. |

### Key Behaviors Covered

- **State machine**: STARTUP → DRAIN → PROBE_BW → PROBE_RTT (all transitions and exit conditions)
- **Recovery**: CONSERVATIVE → GROWTH transitions, persistent congestion, window adjustment, exit
- **Bandwidth estimation**: Delivery rate with/without LastAckedPacketInfo, zero-elapsed fallbacks, app-limited filtering, multi-packet ACKs
- **Pacing**: STARTUP formula (kHighGain), non-STARTUP formula (cycle gains), MinRtt sentinel, CW>>2 cap, pacing-disabled fallback
- **MinRtt**: Update, expiration, no-update-when-larger, PROBE_RTT timer start/exit
- **Congestion window**: STARTUP growth, PROBE_RTT reduction, recovery min(CW, RW), ack-height filter
- **Edge cases**: ExitingQuiescence suppression, zero-length packets, implicit ACKs, data invalidation, exemption underflow guard

### Fixture Helpers

- `InitializeWithDefaults(windowPkts, mtu, pacing, netStats)` — one-line init
- `PumpBandwidthSample(...)` — injects a bandwidth sample into the filter
- `DriveToBtlbwFound()` — drives STARTUP to completion (BtlbwFound=TRUE, state=PROBE_BW)
- `DriveToProbeRtt()` — drives to PROBE_RTT via MinRtt expiration
- `EnterRecovery(...)` — enters CONSERVATIVE recovery via loss event
- `MakeBbrPacket(...)` / `MakeBbrAckEvent(...)` — construct test metadata

## Testing

This PR **is** the tests. Verified:

```
[  PASSED  ] 75 tests.
```

Coverage (via OpenCppCoverage):
```
lines-covered="484" lines-valid="484" → 100% line coverage on src/core/bbr.c
```

## Documentation

No documentation impact. Tests are self-documenting with block comments describing each scenario.


src/core/unittest/BbrTest.cpp
```diff
@@ -0,0 +1,2260 @@
...
+//    https://datatracker.ietf.org/doc/html/draft-cardwell-iccrg-bbr-congestion-control
+//
+//====================================================================
+
+//
+// Test: OnDataLost - Enter CONSERVATIVE Recovery
+// Scenario: Sends 5000 bytes via OnDataSent, then triggers a loss event of 1200 bytes
+// with LargestPacketNumberLost=5 and LargestSentPacketNumber=10. BBR should enter
+// CONSERVATIVE recovery on the first loss event. Asserts that RecoveryState ==
+// CONSERVATIVE, EndOfRecoveryValid is TRUE, and BytesInFlight decrements from
```
guhetier: nit: The comment about the assert is rather redundant, it describes exactly what the code does.
I find the sentence before describing what the test does and expect at a high level useful, but the last sentence doesn't add value. 

Same for the other tests.

src/core/unittest/BbrTest.cpp
```diff
@@ -0,0 +1,2260 @@
...
+    void InitializeWithDefaults(
+        uint32_t WindowPackets = 10,
+        uint16_t Mtu = 1280,
+        bool PacingEnabled = false,
+        bool NetStatsEnabled = false)
+    {
+        Settings.InitialWindowPackets = WindowPackets;
+        InitBbrMockConnection(Connection, Mtu);
+        Connection.Settings.PacingEnabled = PacingEnabled ? TRUE : FALSE;
+        Connection.Settings.NetStatsEventEnabled = NetStatsEnabled ? TRUE : FALSE;
```
guhetier, line 171: When `NetStatsEventEnabled` is true, it would be interesting to set a dummy callback and to potentially use it to validate the notification is emitted when expected.

src/core/unittest/BbrTest.cpp
```diff
@@ -0,0 +1,2260 @@
...
+
+    uint32_t Allowance = CC->QuicCongestionControlGetSendAllowance(CC, 50000, TRUE);
+    uint32_t CW = CC->QuicCongestionControlGetCongestionWindow(CC);
+    ASSERT_EQ(Allowance, CW - 1000u);
+}
+
+//
+// Test: GetSendAllowance - MinRtt at UINT64_MAX With Pacing
+// Scenario: Initializes with PacingEnabled=TRUE. MinRtt starts at UINT64_MAX (no RTT
+// sample yet). Due to BUG-1 (MinRtt compared to UINT32_MAX not UINT64_MAX), the pacing
```
guhetier: What is `BUG-1`? It sounds like an issue we should potentially fix, but not mention in test comments (and the test should be adapted if a fix is needed of course)

src/core/unittest/BbrTest.cpp
```diff
@@ -0,0 +1,2260 @@
...
+    //
+    // Helper: pump a bandwidth sample into the BBR filter via OnDataAcknowledged.
+    // Creates a packet metadata with HasLastAckedPacketInfo so that
+    // BbrBandwidthFilterOnPacketAcked computes a delivery rate.
+    // Returns the TimeNow used so callers can build on it.
+    //
+    //
+    // Pump a bandwidth sample into the BBR filter via OnDataAcknowledged.
+    // Constructs packet metadata so that BbrBandwidthFilterOnPacketAcked
+    // computes DeliveryRate = BW_UNIT * SendRate_BytesPerSec.
```
guhetier: This comment says the same thing twice.

src/core/unittest/BbrTest.cpp
```diff
@@ -0,0 +1,2260 @@
...
+}
+
+//
+// Test: OnDataAcknowledged - STARTUP to DRAIN Transition
+// Scenario: Calls DriveToBtlbwFound() to detect bottleneck bandwidth in STARTUP. Once
+// BtlbwFound is set, BBR transitions from STARTUP to DRAIN. Since BytesInFlight is
+// near zero after the helper, the DRAIN to PROBE_BW transition fires immediately.
+// Asserts BtlbwFound == TRUE and BbrState == PROBE_BW.
+//
+TEST_F(BbrTest, OnDataAcknowledged_StartupToDrain)
```
guhetier: This test is a duplicate of the next one.
Assuming the comment is correct and the transition to PROBE_BW is due to a low number of bytes in flight, we should set it up to pretend that more bytes are in flight.

src/core/unittest/BbrTest.cpp
saikat107: Resolved.

src/core/unittest/BbrTest.cpp
saikat107: Fixed

src/core/unittest/BbrTest.cpp
saikat107: Added the callback.

src/core/unittest/BbrTest.cpp
saikat107: Fixed

src/core/unittest/BbrTest.cpp
```diff
@@ -0,0 +1,2260 @@
...
+//    https://datatracker.ietf.org/doc/html/draft-cardwell-iccrg-bbr-congestion-control
+//
+//====================================================================
+
+//
+// Test: OnDataLost - Enter CONSERVATIVE Recovery
+// Scenario: Sends 5000 bytes via OnDataSent, then triggers a loss event of 1200 bytes
+// with LargestPacketNumberLost=5 and LargestSentPacketNumber=10. BBR should enter
+// CONSERVATIVE recovery on the first loss event.
+TEST_F(BbrTest, OnDataLost_EnterRecovery)
```
guhetier: I think it would be good to label test generated by DeepTest as such.

```suggestion
TEST_F(BbrTest_DeepTest, OnDataLost_EnterRecovery)
```

I think it will help with future maintenance: it is useful to know what is a generated test that exercise and validate the current behavior of the code, vs what is a human generated test (which generally validate the intended behavior of the code).

Given that those tests are pretty tightly coupled to the implementation, knowing in which category a test is will help understanding what should be done if a change cause a test failure.

src/core/unittest/BbrTest.cpp
```diff
@@ -0,0 +1,2260 @@
...
+enum RECOVERY_STATE {
+    RECOVERY_STATE_NOT_RECOVERY = 0,
+    RECOVERY_STATE_CONSERVATIVE = 1,
+    RECOVERY_STATE_GROWTH       = 2
+};
+
+//
+// Tracks the most recent NETWORK_STATISTICS event delivered by the callback.
+//
+static bool NetStatsCallbackInvoked = false;
```
guhetier: nit: Could be added to the `BbrTest` fixture, with providing the fixture as the callback context.

Avoiding globals is generally desirable for tests to avoid any possible cross-test contamination

src/core/unittest/BbrTest.cpp
```diff
@@ -0,0 +1,2260 @@
...
+    Pkt.Next = NULL;
+    return PacketBuf;
+}
+
+//
+// GoogleTest fixture for BBR congestion control tests.
+//
+class BbrTest : public ::testing::Test {
+protected:
+    static constexpr uint16_t kIPv6UdpOverhead = 48;
```
guhetier: Seems unused?
Why a `k` prefix?

src/core/unittest/BbrTest.cpp
```diff
@@ -0,0 +1,2260 @@
...
+{
+    InitializeWithDefaults();
+
+    CC->QuicCongestionControlOnDataSent(CC, 10000);
+    Connection.Send.NextPacketNumber = 10;
+
+    QUIC_LOSS_EVENT Loss = MakeBbrLossEvent(1200, 5, 10, FALSE);
+    CC->QuicCongestionControlOnDataLost(CC, &Loss);
+
+    // RecoveryWindow = max(BIF=8800, MinCW=4928) - 1200 = 7600
```
guhetier: This differs from the formula in the function level comment, let's keep only the correct one.

src/core/unittest/BbrTest.cpp
```diff
@@ -0,0 +1,2260 @@
...
+
+    // Filter maximum should NOT be replaced by the lower app-limited rate
+    ASSERT_EQ(BbrCongestionControlGetBandwidth(CC), BwBefore);
+}
+
+//
+// Test: BandwidthFilter - Multiple Packets in Single ACK
+// Scenario: Creates two chained packets (Packet1.Next = &Packet2) with different
+// timing data and acknowledges them together. Packet1 has SendRate=400000/AckRate=480000
+// (DeliveryRate=min=400000) and Packet2 has a higher rate. BbrBandwidthFilterOnPacketAckedprocesses both packets
```
guhetier:
```suggestion
// (DeliveryRate=min=400000) and Packet2 has a higher rate. BbrBandwidthFilterOnPacketAcked processes both packets
```

Missing space

src/core/unittest/BbrTest.cpp
```diff
@@ -0,0 +1,2260 @@
...
+    ASSERT_TRUE(Bbr->ExitingQuiescence);
+}
+
+//
+// Test: OnDataSent - Updates BytesInFlightMax High-Water Mark
+// Scenario: Records the initial BytesInFlightMax, then sends InitialMax + 1000 bytes.
+// OnDataSent should update BytesInFlightMax when BytesInFlight exceeds the previous
+// maximum.
+//
+TEST_F(BbrTest, OnDataSent_UpdateBytesInFlightMax)
```
guhetier: Seems very redundant (iddentical?) to GetBytesInFlightMax.

Other potential case of redundancy:
- BtlbwFoundDetection + DrainToProbeBw + ProbeBw_HighGainNoAdvance all collapse to "DriveToBtlbwFound(); check different field of same end state."
  - CanSend_Blocked (L1376) ≡ UpdateBlockedState_BecameBlocked (L1658).
  - GetBytesInFlightMax (L1752) ≡ OnDataSent_UpdateBytesInFlightMax (L1904).
  - OnDataAcknowledged_WithNetStats (L2032) vs _ImplicitWithNetStats (L2010) — perfect parameterized-test candidates.
  - InRecovery_NotInRecovery, IsAppLimited_NotLimited, GetBandwidth_Empty are all subsumed by Initialize_DefaultState.

src/core/unittest/BbrTest.cpp
```diff
@@ -0,0 +1,2260 @@
...
+//  struct fields. If the BBR implementation is refactored -- even
+//  while preserving identical congestion control behavior -- these
+//  tests would break.
+//
+//====================================================================
+
+//
+// Test: Initialize - Verifies Correct Default State
+// Scenario: Initializes BBR with default settings (10 window packets, MTU=1280) and
+// inspects all internal fields. Verifies Name=="BBR", BbrState==STARTUP,
```
guhetier: The comment is very redundant with the test (which is already mostly redundant with the code).
We should at least remove the comment, maybe the test altogether.

guhetier (dismissed):

guhetier (approved):

PR #5889, opened 2026-03-23
fixed .net test issue for opensuse 16 and sles 16

### LINKED ISSUE #5571: Validate linux packages on more distros
Currently, packages are only validated on Ubuntu and Azure Linux, this should be expanded to include other distros to find issues before packages are released.

gaurav2699:
## Description
Linux Validation was failing on OpenSUSE 16 and SLES 16 due to QuicHello not found even though it is there. We use ls+grep instead of find negation, which is unreliable on BusyBox (OpenSUSE/SLES).
Also there was a misleading description for package version that said leave empty for latest version which doesnt work as curl us used to download the package which requires exact package version. 
fixes #5571 
## Documentation


.github/workflows/validate-linux-packages-reuse.yml
```diff
@@ -6,9 +6,9 @@ on:
   workflow_call:
     inputs:
       package_version:
-        description: 'Package version to validate (empty for latest)'
+        description: 'Package version to validate'
         required: false
-        default: ''
+        default: '2.5.6'
```
guhetier: Can we avoid to fix this?
It will get out of date all the time. If "latest" is not something we can determine easily, lets have this parameter be a mandatory one.

.github/workflows/validate-linux-packages-reuse.yml
```diff
@@ -609,7 +609,8 @@ jobs:
         run: |
           docker exec ${{ steps.container.outputs.name }} sh -c '
             # Find and run the self-contained QuicHello executable
-            EXE=$(find /dotnet -maxdepth 1 -name "QuicHello*" -type f ! -name "*.dll" ! -name "*.pdb" ! -name "*.json" 2>/dev/null | head -1)
+            # Use ls+grep instead of find negation, which is unreliable on BusyBox (OpenSUSE/SLES)
```
guhetier: I am not sure I understand this comment.
I assume it meant the old version was unreliable? But since the old version is not there anymore, that comment is just confusing.

.github/workflows/validate-linux-packages-reuse.yml
```diff
@@ -6,9 +6,9 @@ on:
   workflow_call:
     inputs:
       package_version:
```
guhetier: Another limitation we need to address is support for release candidates
(since we would want to run this on release candidates before we publish the release).

We need some code to go from a `X.Y.Z-rcN` version to the version used by each distro.
It can also be done in a follow up PR.

.github/workflows/validate-linux-packages.yml
guhetier, line 1: As discussed in the original PR, we should merge these two files in a single one - we are not reusing the other workflow and this one is only 50 lines of boilerplate to call the other one without any added processing.
(can also be done in follow up PR)

.github/workflows/validate-linux-packages-reuse.yml
gaurav2699: I believe if you put 2.5.6-rc1 as the version number it will test the rc package

guhetier: I think it doesn't, because the version number for various linux distro doesn't follow this formatting (which is what nuget uses), for the package file.

`FILENAME="libmsquic_${VERSION}_${{ matrix.pkg_arch }}.deb"`

`FILENAME="libmsquic-${VERSION}-1.${{ matrix.pkg_arch }}.rpm"`

I expect those filenames will be invalid.

guhetier: From https://packages.microsoft.com/sles/16/testing/Packages/l/:

`libmsquic-2.4.17~rc2-1.aarch64.rpm`
`libmsquic-2.4.17~rc2-1.x86_64.rpm`

~, not -. There are other similar variations for other package managers.

gaurav2699: Yeah , fixed now

guhetier (approved):

PR #5892, opened 2026-03-24
Added new APIs to Spinquic 

### LINKED ISSUE #5590: Update SpinQuic to call new APIs
SpinQuic should be updated to cover the new/experimental APIs:
- `ConnectionPoolCreate`
- `RegistrationClose2`
- App-provided receive buffers APIs: `StreamProvideReceiveBuffers`

gaurav2699:
## Description
SpinQuic updated to cover the new/experimental APIs:

ConnectionPoolCreate
RegistrationClose2
App-provided receive buffers APIs: StreamProvideReceiveBuffers_Describe the purpose of and changes within this Pull Request._
Fixes #5590 

## Testing

## Documentation


src/tools/spin/spinquic.cpp
```diff
@@ -975,7 +984,11 @@ void Spin(Gbs& Gb, LockableVector<HQUIC>& Connections, std::vector<HQUIC>* Liste
             BAIL_ON_NULL_CONNECTION(Connection);
             HQUIC Stream;
             auto ctx = new SpinQuicStream(*SpinQuicConnection::Get(Connection));
-            QUIC_STATUS Status = MsQuic.StreamOpen(Connection, (QUIC_STREAM_OPEN_FLAGS)GetRandom(8), SpinQuicHandleStreamEvent, ctx, &Stream);
+            QUIC_STREAM_OPEN_FLAGS OpenFlags = (QUIC_STREAM_OPEN_FLAGS)GetRandom(8);
+            if (GetRandom(5) == 0) {
+                OpenFlags |= QUIC_STREAM_OPEN_FLAG_APP_OWNED_BUFFERS;
```
anrossi: Why is this needed? Doesn't the `GetRandom(8)` call include this flag?

src/tools/spin/spinquic.cpp
```diff
@@ -1166,6 +1179,66 @@ void Spin(Gbs& Gb, LockableVector<HQUIC>& Connections, std::vector<HQUIC>* Liste
...
+                    for (uint16_t i = 0; i < PoolSize; i++) {
+                        auto ctx = new(std::nothrow) SpinQuicConnection(PoolConnections[i], ThreadID);
+                        if (ctx != nullptr) {
+                            std::lock_guard<std::mutex> Lock(Connections);
+                            Connections.push_back(PoolConnections[i]);
+                        } else {
+                            MsQuic.ConnectionClose(PoolConnections[i]);
+                        }
+                    }
+                }
```
anrossi, line 1291: depending on the `PoolConfig.Flags` value, you may need to close the connection handles on failure.

src/tools/spin/spinquic.cpp
```diff
@@ -237,7 +237,14 @@ struct SpinQuicGlobals {
             free(Alpns);
         }
         if (Registration) {
-            MsQuic->RegistrationClose(Registration);
+            if (rand() % 2 == 0) {
+                MsQuic->RegistrationClose2(
+                    Registration,
+                    [](void*) -> void { }, // No-op callback
```
anrossi: If the spin test ever runs msquic in external execution mode, then this callback will need to be waited on.

src/tools/spin/spinquic.cpp
gaurav2699: GetRandom(8) returns a random value in the range [0, 7] we need 8 for the APP_OWNED_BUFFER

.github/workflows/validate-linux-packages-reuse.yml
```diff
@@ -609,7 +609,8 @@ jobs:
         run: |
           docker exec ${{ steps.container.outputs.name }} sh -c '
             # Find and run the self-contained QuicHello executable
-            EXE=$(find /dotnet -maxdepth 1 -name "QuicHello*" -type f ! -name "*.dll" ! -name "*.pdb" ! -name "*.json" 2>/dev/null | head -1)
+            # Use ls+grep instead of find negation, which is unreliable on BusyBox (OpenSUSE/SLES)
+            EXE=$(ls /dotnet/QuicHello* 2>/dev/null | grep -v -e "\.dll$" -e "\.pdb$" -e "\.json$" | head -1)
```
guhetier: Should not be in this PR?

src/tools/spin/spinquic.cpp
guhetier: Event when not running in external execution mode I think.
This callback is non-blocking, and I think `MsQuicClose` must not be called until the registration closure is completed.

src/tools/spin/spinquic.cpp
guhetier: Why not make it `GetRandom(2 << 4)` which would cover all combinations?

src/tools/spin/spinquic.cpp
```diff
@@ -975,7 +984,11 @@ void Spin(Gbs& Gb, LockableVector<HQUIC>& Connections, std::vector<HQUIC>* Liste
             BAIL_ON_NULL_CONNECTION(Connection);
             HQUIC Stream;
             auto ctx = new SpinQuicStream(*SpinQuicConnection::Get(Connection));
-            QUIC_STATUS Status = MsQuic.StreamOpen(Connection, (QUIC_STREAM_OPEN_FLAGS)GetRandom(8), SpinQuicHandleStreamEvent, ctx, &Stream);
+            QUIC_STREAM_OPEN_FLAGS OpenFlags = (QUIC_STREAM_OPEN_FLAGS)GetRandom(8);
+            if (GetRandom(5) == 0) {
+                OpenFlags |= QUIC_STREAM_OPEN_FLAG_APP_OWNED_BUFFERS;
+            }
+            QUIC_STATUS Status = MsQuic.StreamOpen(Connection, OpenFlags, SpinQuicHandleStreamEvent, ctx, &Stream);
```
guhetier: When `QUIC_STREAM_OPEN_FLAG_APP_OWNED_BUFFERS` is set, you should consider randomly providing some amount of buffer before starting the stream.
`SpinQuicAPICallStreamProvideReceiveBuffers` is meant to be more of a fallback when you couldn't provide enough space upfront.

src/tools/spin/spinquic.cpp
```diff
@@ -1166,6 +1179,66 @@ void Spin(Gbs& Gb, LockableVector<HQUIC>& Connections, std::vector<HQUIC>* Liste
...
+                if (Stream == nullptr) continue;
+                uint32_t BufCount = GetRandom(3) + 1; // 1-3 buffers
+                QUIC_BUFFER* Buffers = new(std::nothrow) QUIC_BUFFER[BufCount];
+                if (Buffers == nullptr) continue;
+                for (uint32_t i = 0; i < BufCount; i++) {
+                    Buffers[i].Length = MaxBufferSizes[GetRandom(BufferCount)];
+                    Buffers[i].Buffer = Gb.SendBuffer; // Reuse send buffer as receive buffer
+                }
+                MsQuic.StreamProvideReceiveBuffers(Stream, BufCount, Buffers);
+                delete[] Buffers;
```
guhetier: This is user-mode C++. Please use modern C++ construct line `std::vector` and smart pointers.
You should almost never have to write a `new` or `delete`.

src/tools/spin/spinquic.cpp
```diff
@@ -1166,6 +1179,66 @@ void Spin(Gbs& Gb, LockableVector<HQUIC>& Connections, std::vector<HQUIC>* Liste
...
+            break;
+        }
+        case SpinQuicAPICallStreamProvideReceiveBuffers: {
+            auto Connection = Connections.TryGetRandom();
+            BAIL_ON_NULL_CONNECTION(Connection);
+            auto ctx = SpinQuicConnection::Get(Connection);
+            {
+                std::lock_guard<std::mutex> Lock(ctx->Lock);
+                auto Stream = ctx->TryGetStream();
+                if (Stream == nullptr) continue;
```
guhetier: style: Always use brackets for a `if`, with the body on a separate line from the condition.

src/tools/spin/spinquic.cpp
```diff
@@ -1166,6 +1179,66 @@ void Spin(Gbs& Gb, LockableVector<HQUIC>& Connections, std::vector<HQUIC>* Liste
...
+                uint32_t BufCount = GetRandom(3) + 1; // 1-3 buffers
+                QUIC_BUFFER* Buffers = new(std::nothrow) QUIC_BUFFER[BufCount];
+                if (Buffers == nullptr) continue;
+                for (uint32_t i = 0; i < BufCount; i++) {
+                    Buffers[i].Length = MaxBufferSizes[GetRandom(BufferCount)];
+                    Buffers[i].Buffer = Gb.SendBuffer; // Reuse send buffer as receive buffer
```
guhetier: Let's allocate a receive buffer instead, or at least rename this, if it isn't only a send buffer.

guhetier (commented):
It would be nice to also add a call to `StreamProvideReceiveBuffers` when a new stream is notified to a connection.

src/tools/spin/spinquic.cpp
```diff
@@ -237,7 +239,20 @@ struct SpinQuicGlobals {
             free(Alpns);
         }
         if (Registration) {
-            MsQuic->RegistrationClose(Registration);
+            if (rand() % 2 == 0) {
+                CXPLAT_EVENT CloseComplete;
```
guhetier: nit: Consider using a `CxPlatEvent` since this is C++.

src/tools/spin/spinquic.cpp
```diff
@@ -218,11 +218,13 @@ struct SpinQuicGlobals {
     uint32_t AlpnCount {0};
     const size_t SendBufferSize { MaxBufferSizes[BufferCount - 1] + UINT8_MAX };
     uint8_t* SendBuffer;
```
guhetier: nit: Consider replacing this two with `std::vector<uint8_t>` since this is C++

src/tools/spin/spinquic.cpp
```diff
@@ -237,7 +239,20 @@ struct SpinQuicGlobals {
             free(Alpns);
         }
         if (Registration) {
-            MsQuic->RegistrationClose(Registration);
+            if (rand() % 2 == 0) {
```
guhetier: Shouldn't this use `GetRandom`?

src/tools/spin/spinquic.cpp
```diff
@@ -491,8 +509,21 @@ QUIC_STATUS QUIC_API SpinQuicHandleConnectionEvent(HQUIC Connection, void* , QUI
         if (GetRandom(2) == 0) {
             Event->PEER_STREAM_STARTED.Flags |= QUIC_STREAM_OPEN_FLAG_DELAY_ID_FC_UPDATES;
         }
+        if (GetRandom(5) == 0) {
+            Event->PEER_STREAM_STARTED.Flags |= QUIC_STREAM_OPEN_FLAG_APP_OWNED_BUFFERS;
+        }
         auto StreamCtx = new SpinQuicStream(*ctx, Event->PEER_STREAM_STARTED.Stream);
         MsQuic.SetCallbackHandler(Event->PEER_STREAM_STARTED.Stream, (void *)SpinQuicHandleStreamEvent, StreamCtx);
+        if (Event->PEER_STREAM_STARTED.Flags & QUIC_STREAM_OPEN_FLAG_APP_OWNED_BUFFERS) {
+            static uint8_t RecvBuffer[64000];
```
guhetier: This needs to stay alive until MsQuic is done writing to it, which is this case, basically means until the stream is closed.
Maybe `static` saves you, but you should consider using the global receive buffer instead (and keep track of how much of it has been used already, a `std::span` might be your friend.

src/tools/spin/spinquic.cpp
```diff
@@ -975,9 +1006,19 @@ void Spin(Gbs& Gb, LockableVector<HQUIC>& Connections, std::vector<HQUIC>* Liste
             BAIL_ON_NULL_CONNECTION(Connection);
             HQUIC Stream;
             auto ctx = new SpinQuicStream(*SpinQuicConnection::Get(Connection));
-            QUIC_STATUS Status = MsQuic.StreamOpen(Connection, (QUIC_STREAM_OPEN_FLAGS)GetRandom(8), SpinQuicHandleStreamEvent, ctx, &Stream);
+            QUIC_STREAM_OPEN_FLAGS OpenFlags = (QUIC_STREAM_OPEN_FLAGS)GetRandom(16);
+            QUIC_STATUS Status = MsQuic.StreamOpen(Connection, OpenFlags, SpinQuicHandleStreamEvent, ctx, &Stream);
             if (QUIC_SUCCEEDED(Status)) {
                 ctx->Handle = Stream;
+                if ((OpenFlags & QUIC_STREAM_OPEN_FLAG_APP_OWNED_BUFFERS) && GetRandom(2) == 0) {
+                    uint32_t BufCount = GetRandom(3) + 1; // 1-3 buffers
```
guhetier: nit: Please go to at least 4 or 5.

3 is a magic number used in the codebase to go from stack allocated to heap allocated in path that this will hopefully exercise, we will get more coverage with at least 4 buffers.

src/tools/spin/spinquic.cpp
guhetier: Actually, the global receive buffer is probably a bad idea too (it'll stay alive but be used concurrently to receive data by multiple streams).
Having the receive buffer in the stream context is probably best.

src/tools/spin/spinquic.cpp
gaurav2699: CxPlatEvent throws an error. I think it requires the CX_PLATFORM_TYPE which comes from the platform headers which we dont include

src/tools/spin/spinquic.cpp
gaurav2699: That actually doesnt work. Not fully sure why but GetRandom has a ThreadID param which throws the error as ThreadID: Undeclared identifier, this is in the destructor where GetRandom with ThreadID context doesn't make sense anyway.

guhetier: `#define GetRandom(UpperBound) GetRandom(UpperBound, ThreadID)`

Wow. That is very bad code.
Can you please delete this macro, do a research / replace to provide the thread ID explicitly in current calls, and use it?

src/tools/spin/spinquic.cpp
guhetier: I think you just need to include msquic.hpp after msquichelper.h

src/tools/spin/spinquic.cpp
```diff
@@ -1011,11 +1018,18 @@ void Spin(Gbs& Gb, LockableVector<HQUIC>& Connections, std::vector<HQUIC>* Liste
...
-                    uint32_t BufCount = GetRandom(3) + 1; // 1-3 buffers
+                    uint32_t BufCount = GetRandom(5) + 1; // 1-5 buffers
                     std::vector<QUIC_BUFFER> Buffers(BufCount);
+                    size_t TotalSize = 0;
                     for (uint32_t i = 0; i < BufCount; i++) {
                         Buffers[i].Length = MaxBufferSizes[GetRandom(BufferCount)];
-                        Buffers[i].Buffer = Gb.RecvBuffer;
+                        TotalSize += Buffers[i].Length;
+                    }
+                    ctx->EnsureRecvBuffer(TotalSize);
```
guhetier: nit: `EnsureRecvBuffer` is not very meaningful as a name. Maybe `AllocatedRecvBuffer` or `ReserveRecvBuffer`?

Another approach can also be a `std::vector<QUIC_BUFFER> GetReceiveBuffer(std::span<uint32_t> sizes) { ... }` that both receive the buffer backing up the data and initialize the `QUIC_BUFFER`.

src/tools/spin/spinquic.cpp
gaurav2699: That doesnt work either cause it causes name collisions:
`C:\Users\gaurasingh\msquic\src\tools\spin\spinquic.cpp(203,23): error C2040: 'MsQuic': 'QUIC_API_TABLE' differs in leve
ls of indirection from 'const MsQuicApi *'`
msquic.hpp declares extern const MsQuicApi* MsQuic; while spinquic.cpp declares static QUIC_API_TABLE MsQuic. Including msquic.hpp causes a name collision.

src/tools/spin/spinquic.cpp
gaurav2699: Done

src/tools/spin/spinquic.cpp
guhetier: Those are things you can fix if decide to go that way.
This is otherwise a nitpicking comment, so up to you how you resolve it.

src/tools/spin/spinquic.cpp
```diff
@@ -297,11 +289,27 @@ struct SpinQuicStream {
...
-        return (SpinQuicStream*)MsQuic.GetContext(Stream);
+        return (SpinQuicStream*)MsQuicTable.GetContext(Stream);
+    }
+    std::vector<QUIC_BUFFER> AllocateReceiveBuffers(const std::vector<uint32_t>& Sizes) {
+        size_t TotalSize = 0;
+        for (auto Size : Sizes) {
+            TotalSize += Size;
+        }
+        size_t Offset = RecvBuffer.size();
+        RecvBuffer.resize(Offset + TotalSize);
```
guhetier: I think this will be a problem: if you `resize` a vector, it potentially creates a new allocation for the entire vector, so any pointer to the pre-existing data will become dangling.

Either we need to guarantee only one set of buffers is used at once, or your need a vector of vectors (or unique_ptr).

guhetier (commented):
Looks largely good, but the failures in the CI are the sign of either a bug in MsQuic the new test is finding, or that the spintest is doing something wrong.

Please address it before we merge this PR (in a different PR if the issue is a MsQuic bug)

gaurav2699:
> Looks largely good, but the failures in the CI are the sign of either a bug in MsQuic the new test is finding, or that the spintest is doing something wrong.
> 
> Please address it before we merge this PR (in a different PR if the issue is a MsQuic bug)

Pulling the latest changes seem to have removed the issue. Maybe due to the latest xdp removal effort.

guhetier (dismissed):

src/tools/spin/spinquic.cpp
```diff
@@ -1234,8 +1237,15 @@ void Spin(Gbs& Gb, LockableVector<HQUIC>& Connections, std::vector<HQUIC>* Liste
...
+            //
+            // Pool create adds up to N connections per call (vs. 1 for
+            // ConnectionOpen), so fire rarely to keep the per-iteration
+            // connection count bounded. Otherwise teardown blows the
+            // watchdog under Debug + alloc_fail.
+            //
+            if (GetRandom(10, ThreadID) != 0) break;
             if (!IsServer) {
-                uint16_t PoolSize = (uint16_t)(GetRandom(4, ThreadID) + 1); // 1-4 connections
+                uint16_t PoolSize = (uint16_t)(GetRandom(2, ThreadID) + 1); // 1-2 connections
```
guhetier, line 1248: Both skipping this most of the time + limiting the number of connections to 2 max seems pretty extreme.

src/tools/spin/spinquic.cpp
```diff
@@ -1702,11 +1724,11 @@ void start() {
         SpinSettings.RunTimeMs = SpinSettings.RunTimeMs / SpinSettings.RepeatCount;
         for (uint32_t i = 0; i < SpinSettings.RepeatCount; i++) {
 
-            CXPLAT_THREAD_CONFIG Config = {
-                0, 0, "spin_run", RunThread, nullptr
-            };
             CXPLAT_THREAD Threads[4];
             uint32_t Count = FuzzData ? (uint32_t)FuzzingData::NumSpinThread / 2 : (uint32_t)(rand() % (ARRAYSIZE(Threads) - 1) + 1);
+            CXPLAT_THREAD_CONFIG Config = {
+                0, 0, "spin_run", RunThread, (void*)(uintptr_t)Count
```
guhetier, line 1730: Please don't use the context pointer to store a numerical value.
You wait for the thread right below, just pass a pointer to the `Count`.

src/tools/spin/spinquic.cpp
```diff
@@ -1641,7 +1657,13 @@ void start() {
...
 
     {
-        SpinQuicWatchdog Watchdog((uint32_t)SpinSettings.RunTimeMs + SpinSettings.RepeatCount*WATCHDOG_WIGGLE_ROOM);
+        //
+        // Matches the per-iteration watchdog in RunThread scaled by the
+        // maximum possible sibling-thread count (see ARRAYSIZE(Threads) - 1
+        // below) so the outer watchdog is always looser than the sum of the
+        // inner ones.
+        //
+        SpinQuicWatchdog Watchdog((uint32_t)SpinSettings.RunTimeMs + SpinSettings.RepeatCount * 3 * WATCHDOG_WIGGLE_ROOM);
```
guhetier, line 1666: Why 3?

src/tools/spin/spinquic.cpp
```diff
@@ -1490,8 +1500,14 @@ void PrintHelpText(void)
 
 CXPLAT_THREAD_CALLBACK(RunThread, Context)
 {
-    UNREFERENCED_PARAMETER(Context);
-    SpinQuicWatchdog Watchdog((uint32_t)SpinSettings.RunTimeMs + WATCHDOG_WIGGLE_ROOM);
+    //
+    // Context carries the number of sibling RunThreads for this iteration.
+    // Cleanup serializes on the debug-heap critical section, so wiggle scales
+    // with concurrency: more sibling threads means proportionally longer
+    // teardown under NTGLOBALFLAG=0x70.
```
guhetier, line 1507: `under NTGLOBALFLAG=0x70` what does this mean?

src/tools/spin/spinquic.cpp
```diff
@@ -1234,8 +1237,15 @@ void Spin(Gbs& Gb, LockableVector<HQUIC>& Connections, std::vector<HQUIC>* Liste
             break;
         }
         case SpinQuicAPICallConnectionPoolCreate: {
+            //
+            // Pool create adds up to N connections per call (vs. 1 for
+            // ConnectionOpen), so fire rarely to keep the per-iteration
+            // connection count bounded. Otherwise teardown blows the
+            // watchdog under Debug + alloc_fail.
```
guhetier, line 1244: Why? That doesn't seem obvious, and we should probably fix the teardown if it can take that long?

src/tools/spin/spinquic.cpp
gaurav2699: I was just experimenting the root cause, have reverted these changes. This definitely reduced the pool create activity by 17x

src/tools/spin/spinquic.cpp
gaurav2699: It's ARRAYSIZE(Threads) - 1 from the line below (rand() % (ARRAYSIZE(Threads) - 1) + 1, which picks 1..3), i.e. the
  maximum possible concurrent-thread count so the outer watchdog is never tighter than the sum of the per-iter ones

src/tools/spin/spinquic.cpp
gaurav2699: It actually refers to the windows debug heap, unnecessary tho.

src/tools/spin/spinquic.cpp
gaurav2699: The dumps show three spin_client threads simultaneously inside RtlDebugFreeHeap, specifically one of them blocked on RtlEnterCriticalSectionContended waiting for the heap CS, while the other two walk free lists. Every free() on the Windows debug heap takes a per-heap critical  section, so N concurrent closing connections serialize on it

src/tools/spin/spinquic.cpp
gaurav2699: You are right, Have4 reverted this.

src/tools/spin/spinquic.cpp
```diff
@@ -993,14 +987,31 @@ void Spin(Gbs& Gb, LockableVector<HQUIC>& Connections, std::vector<HQUIC>* Liste
...
 
-        switch (GetRandom(SpinQuicAPICallCount, ThreadID)) {
+        uint32_t APICall = GetRandom(SpinQuicAPICallCount, ThreadID);
+
+        //
+        // Above SOFT_CONNECTION_TARGET, replace the random pick with a
+        // ConnectionClose with probability (Excess / SOFT_CONNECTION_TARGET).
+        // Pulls the count back toward the target without forbidding bursts.
+        //
+        {
```
guhetier, line 993: nit: no need for an extra context.

src/tools/spin/spinquic.cpp
```diff
@@ -1157,13 +1236,113 @@ void Spin(Gbs& Gb, LockableVector<HQUIC>& Connections, std::vector<HQUIC>* Liste
...
+                // to SpinQuicConnection*, AddStream into stack memory, and
+                // orphan the stream for the rest of the run. Wrappers created
+                // with the single-arg ctor have Connection == nullptr; their
+                // dtor calls ConnectionClose(nullptr) which is a documented
+                // no-op, so unwinding them on failure is safe.
+                //
+                std::vector<SpinQuicConnection*> Wrappers(PoolSize, nullptr);
+                std::vector<void*> Contexts(PoolSize, nullptr);
+                for (uint16_t i = 0; i < PoolSize; i++) {
+                    Wrappers[i] = new(std::nothrow) SpinQuicConnection(ThreadID);
```
guhetier: If the allocation fails, a connection may end up with a nullptr context, which would be dereferenced when handling notifications like SHUTDOWN_COMPLETE.

The simpler is probably to abort creating the connection pool if any allocation failed.

guhetier (dismissed):

src/tools/spin/spinquic.cpp
```diff
@@ -1249,27 +1244,24 @@ void Spin(Gbs& Gb, LockableVector<HQUIC>& Connections, std::vector<HQUIC>* Liste
...
-                // landed before wrapper creation would cast that stack address
-                // to SpinQuicConnection*, AddStream into stack memory, and
-                // orphan the stream for the rest of the run. Wrappers created
-                // with the single-arg ctor have Connection == nullptr; their
-                // dtor calls ConnectionClose(nullptr) which is a documented
-                // no-op, so unwinding them on failure is safe.
-                //
                 std::vector<SpinQuicConnection*> Wrappers(PoolSize, nullptr);
                 std::vector<void*> Contexts(PoolSize, nullptr);
+                bool AllocFailed = false;
```
guhetier, line 1250: Consider a `std::vector<std::unique_ptr<SpinQuicConnection>> Wrappers` instaead.
Then `Contexts[i] = Wrappers[i].release()`, and you can simply `break` on alloc failure and let the smart pointer deal with any cleanup.
It is rare you should call `delete` in user-mode C++.

src/tools/spin/spinquic.cpp
guhetier: I meant the `{ ... }` context, not the comment :)

The comment wording was a bit convoluted, but a quick "if we have too many connections open, close one with an increasingly high probability" is helpful to understand the intent.

guhetier (dismissed):

src/tools/spin/spinquic.cpp
```diff
@@ -1157,13 +1232,85 @@ void Spin(Gbs& Gb, LockableVector<HQUIC>& Connections, std::vector<HQUIC>* Liste
...
+            if (!IsServer) {
+                uint16_t PoolSize = (uint16_t)(GetRandom(4, ThreadID) + 1); // 1-4 connections
+                std::vector<HQUIC> PoolConnections(PoolSize, nullptr);
+                std::vector<std::unique_ptr<SpinQuicConnection>> Wrappers(PoolSize);
+                std::vector<void*> Contexts(PoolSize, nullptr);
+                bool AllocFailed = false;
+                for (uint16_t i = 0; i < PoolSize; i++) {
+                    Wrappers[i].reset(new(std::nothrow) SpinQuicConnection(ThreadID));
+                    if (!Wrappers[i]) {
+                        AllocFailed = true;
```
guhetier, line 1254: nit: This kind of patterns is the sign this code should be factored in helper functions.
But lets keep it for another PR.

src/tools/spin/spinquic.cpp
```diff
@@ -1157,13 +1232,85 @@ void Spin(Gbs& Gb, LockableVector<HQUIC>& Connections, std::vector<HQUIC>* Liste
...
+                PoolConfig.CibirIds = nullptr;
+                PoolConfig.CibirIdLength = 0;
+                PoolConfig.Flags = (QUIC_CONNECTION_POOL_FLAGS)GetRandom(2, ThreadID);
+                QUIC_STATUS Status = MsQuicTable.ConnectionPoolCreate(&PoolConfig, PoolConnections.data());
+                if (QUIC_SUCCEEDED(Status)) {
+                    for (uint16_t i = 0; i < PoolSize; i++) {
+                        SpinQuicConnection* ctx = Wrappers[i].release();
+                        ctx->Set(PoolConnections[i]);
+                        std::lock_guard<std::mutex> Lock(Connections);
+                        Connections.push_back(PoolConnections[i]);
```
guhetier, line 1281: nit:

```suggestion
                        Wrappers[i]->Set(PoolConnections[i]);
                        std::lock_guard<std::mutex> Lock(Connections);
                        Connections.push_back(PoolConnections[i]);
                        Wrappers[i].release();
```

Release the smart pointer only once the ownership has been successfully transferred to the vector.
If `push_back` throws, you would otherwise leak the `SpinConnection`.
Not a real since exceptions are not handled anyway.

guhetier (approved):

guhetier:
Feel free to merge as is, the comments are non blocking remarks.

PR #5897, opened 2026-03-27
[CP] Fix SAL annotation placement on function templates (#5895)

guhetier:
Cherry-pick of #5895

## Description
SAL annotations (`_Success_`, `_Function_class_`) placed before the `template` keyword cause MSVC compilation errors (C2059) when TVS / static analysis is enabled. The `template` head must always come first; annotations go between it and the return type.

Fixed 5 occurrences:
- `src/perf/lib/SecNetPerfMain.cpp`: 4 instances of `_Success_` before `template`  
- `src/test/lib/ApiTest.cpp`: 1 instance of `_Function_class_` before `template`  

## Testing
Existing tests cover this change. No new tests needed — this is a compilation fix only (no behavioral change).

## Documentation
No documentation impact.

### DIFF
--- a/.github/workflows/build-reuse-win.yml
+++ b/.github/workflows/build-reuse-win.yml
@@ -82,7 +82,7 @@ jobs:
         repository: ${{ inputs.repo}}
         ref: ${{ inputs.ref }}
     - name: Install Perl
-      uses: shogo82148/actions-setup-perl@22423f01bde48fb88785c007e3166fbbbd8e892a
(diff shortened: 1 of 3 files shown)


mtfriesen:
Should the perl action/CP be a separate chore?

guhetier:
> Should the perl action/CP be a separate chore?

It can if you feel strongly, I wanted to check fast it was enough to solve the failure.

mtfriesen:
> > Should the perl action/CP be a separate chore?
> 
> It can if you feel strongly, I wanted to check fast it was enough to solve the failure.

I think ideally they're separate, but I don't feel strongly at all.

PR #5900, opened 2026-03-27
fix(recv_buffer): skip already-written bytes on overlapping writes

### LINKED ISSUE #5819: Overlapping CRYPTO frames overwrite previously received data
#### Tested Build
v2.5.5

#### Observation
The msquic server accepts overlapping CRYPTO frames containing different data at an already received offset and updates previously stored data instead of rejecting or ignoring the overlap.

#### Expected Behavior

RFC 9000, Section 2.2 states:

> An endpoint could receive data for a stream at the same stream offset multiple times. Data that has already been received can be discarded. The data at a given offset MUST NOT change if it is sent multiple times; an endpoint MAY treat receipt of different data at the same offset within a stream as a connection error of type PROTOCOL_VIOLATION.

MarkedMuichiro:
Fixes #5819

## Problem
QuicRecvBufferCopyIntoChunks previously overwrote already-received data
when a new write partially overlapped with existing WrittenRanges.
QuicRangeAddRange correctly returns WrittenRangesUpdated=TRUE for partial
overlaps, causing CopyIntoChunks to write the full incoming buffer
including bytes at already-received offsets.

## Fix
Walk existing WrittenRanges before calling QuicRangeAddRange to identify
gap regions within the incoming write window. Call
QuicRecvBufferCopyIntoChunks once per gap, copying only new bytes.
ReadLength update moved out of QuicRecvBufferCopyIntoChunks into
QuicRecvBufferWrite after the range is committed.

This fixes the issue for both CRYPTO and STREAM frame processing.

## Testing
Added four parameterized unit tests across all four buffer modes:
- OverlapWritePreservesExistingData: partial overlap at end
- OverlapWriteAtFront: fill gap and overlap
- OverlapWriteExactDuplicate: pure duplicate write
- OverlapWriteMultipleGaps: swiss-cheese pattern covered by large write

@guhetier

### DIFF
--- a/src/core/unittest/RecvBufferTest.cpp
+++ b/src/core/unittest/RecvBufferTest.cpp
@@ -2453,3 +2453,108 @@ TEST_P(WithMode, OverlapWriteMultipleGaps)
     ASSERT_EQ(30u, TotalRead);
     RecvBuf.Drain(30);
 }
+TEST_P(WithMode, OverlapWriteThreeRangesTwoOverlaps)
+{
+    //
+    // Write [0, 15), then [10, 25) overlapping [10, 15), then [20, 30) overlapping [20, 25).
+    // Sentinel values (0xff) cover both overlap regions to verify existing data is preserved.
+    //
+    RecvBuffer RecvBuf;
+    ASSERT_EQ(QUIC_STATUS_SUCCESS, RecvBuf.Initialize(GetParam()));
+
+    uint64_t InOutWriteLength = LARGE_TEST_BUFFER_LENGTH;
+    BOOLEAN NewDataReady = FALSE;
+
+    // Write [0, 15) with pattern: byte[i] = (uint8_t)i
+    ASSERT_EQ(QUIC_STATUS_SUCCESS, RecvBuf.Write(0, 15, &InOutWriteLength, &NewDataReady));
+    ASSERT_TRUE(NewDataReady);
+
+    // Write [10, 25): 0xff sentinel in overlap [10, 15), valid data in [15, 25)
+    uint8_t SecondWrite[15];
+    for (uint16_t i = 0; i < 5;  ++i) SecondWrite[i] = 0xff;
+    for (uint16_t i = 5; i < 15; ++i) SecondWrite[i] = (uint8_t)(10 + i);
+    InOutWriteLength = LARGE_TEST_BUFFER_LENGTH;
+    NewDataReady = FALSE;
+    ASSERT_EQ(QUIC_STATUS_SUCCESS, RecvBuf.WriteCustom(10, 15, SecondWrite, &InOutWriteLength, &NewDataReady));
+    ASSERT_TRUE(NewDataReady);
+
+    // Write [20, 30): 0xff sentinel in overlap [20, 25), valid data in [25, 30)
+    uint8_t ThirdWrite[10];
+    for (uint16_t i = 0; i < 5;  ++i) ThirdWrite[i] = 0xff;
+    for (uint16_t i = 5; i < 10; ++i) ThirdWrite[i] = (uint8_t)(20 + i);
+    InOutWriteLength = LARGE_TEST_BUFFER_LENGTH;
+    NewDataReady = FALSE;
+    ASSERT_EQ(QUIC_STATUS_SUCCESS, RecvBuf.WriteCustom(20, 10, ThirdWrite, &InOutWriteLength, &NewDataReady));
+    ASSERT_TRUE(NewDataReady);
+    ASSERT_EQ(30ull, RecvBuf.GetTotalLength());
+
+    // Expected: [0,15) from first write, [15,25) from SecondWrite[5..14], [25,30) from ThirdWrite[5..9]
+    uint8_t Expected[30];
+    for (uint32_t i = 0;  i < 15; ++i) Expected[i] = (uint8_t)i;
+    for (uint32_t i = 15; i < 25; ++i) Expected[i] = (uint8_t)(10 + (i - 10));
+    for (uint32_t i = 25; i < 30; ++i) Expected[i] = (uint8_t)(20 + (i - 20));
+
+    uint64_t ReadOffset;
+    QUIC_BUFFER ReadBuffers[3];
+    uint32_t BufferCount = ARRAYSIZE(ReadBuffers);
+    QuicRecvBufferRead(&RecvBuf.RecvBuf, &ReadOffset, &BufferCount, ReadBuffers);
+    ASSERT_EQ(0ull, ReadOffset);
+    uint32_t TotalRead = 0;
+    for (uint32_t b = 0; b < BufferCount; ++b) {
+        ASSERT_EQ(0, memcmp(ReadBuffers[b].Buffer, Expected + TotalRead, ReadBuffers[b].Length));
+        TotalRead += ReadBuffers[b].Length;
+    }
+    ASSERT_EQ(30u, TotalRead);
+    RecvBuf.Drain(30);
+}
+
+TEST_P(WithMode, OverlapWriteThreeContiguousNonOverlapping)
+{
+    //
+    // Write three non-overlapping contiguous ranges: [0,10), [10,20), [20,30).
+    // Ranges touch but do not overlap. All data should land correctly without corruption.
+    //
+    RecvBuffer RecvBuf;
+    ASSERT_EQ(QUIC_STATUS_SUCCESS, RecvBuf.Initialize(GetParam()));
+
+    uint64_t InOutWriteLength = LARGE_TEST_BUFFER_LENGTH;
+    BOOLEAN NewDataReady = FALSE;
+
+    // Write [0, 10) with pattern: byte[i] = (uint8_t)i
+    ASSERT_EQ(QUIC_STATUS_SUCCESS, RecvBuf.Write(0, 10, &InOutWriteLength, &NewDataReady));
+    ASSERT_TRUE(NewDataReady);
+
+    // Write [10, 20) — contiguous, no overlap
+    uint8_t SecondWrite[10];
+    for (uint16_t i = 0; i < 10; ++i) SecondWrite[i] = (uint8_t)(10 + i);
+    InOutWriteLength = LARGE_TEST_BUFFER_LENGTH;
+    NewDataReady = FALSE;
+    ASSERT_EQ(QUIC_STATUS_SUCCESS, RecvBuf.WriteCustom(10, 10, SecondWrite, &InOutWriteLength, &NewDataReady));
+    ASSERT_TRUE(NewDataReady);
+
+    // Write [20, 30) — contiguous, no overlap
+    uint8_t ThirdWrite[10];
+    for (uint16_t i = 0; i < 10; ++i) ThirdWrite[i] = (uint8_t)(20 + i);
+    InOutWriteLength = LARGE_TEST_BUFFER_LENGTH;
+    NewDataReady = FALSE;
+    ASSERT_EQ(QUIC_STATUS_SUCCESS, RecvBuf.WriteCustom(20, 10, ThirdWrite, &InOutWriteLength, &NewDataReady));
+    ASSERT_TRUE(NewDataReady);
+    ASSERT_EQ(30ull, RecvBuf.GetTotalLength());
+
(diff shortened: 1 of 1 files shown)


src/core/recv_buffer.c
```diff
@@ -712,7 +689,60 @@ QuicRecvBufferWrite(
     }
 
     //
-    // Set the write offset/length as a valid written range.
+    // Copy only the new (non-overlapping) bytes into the chunks.
```
guhetier: Please factor this into a function `QuicRecvBufferWriteNewBytes` (or similar).

src/core/recv_buffer.c
```diff
@@ -742,9 +769,15 @@ QuicRecvBufferWrite(
     *NewDataReady = UpdatedRange->Low == 0;
 
     //
-    // Write the data into the chunks now that everything has been validated.
+    // Update the amount of data readable in the first chunk.
+    // (Moved here from QuicRecvBufferCopyIntoChunks since the range is now committed.)
```
guhetier: Please keep comment referring only to the present state of the code, not the past.

`Update the amount of data readable in the first chunk, committing the write.`

src/core/recv_buffer.c
```diff
@@ -661,8 +651,6 @@ QuicRecvBufferWrite(
     //
     // Check that the write is not going to cause us to consume more than the
     // quota allocated by the connection flow control.
-    // If it is in bound, update the output to indicate how much of the quota is
```
guhetier, line 664: Please do not delete existing comments unless they are no longer correct.

MarkedMuichiro: Restored — apologies for the removal.

src/core/recv_buffer.c
MarkedMuichiro: Got it; done.

src/core/recv_buffer.c
MarkedMuichiro: Done. Thank you.

src/core/recv_buffer.c
```diff
@@ -742,9 +809,14 @@ QuicRecvBufferWrite(
     *NewDataReady = UpdatedRange->Low == 0;
 
     //
-    // Write the data into the chunks now that everything has been validated.
+    // Update the amount of data readable in the first chunk, committing the write.
```
guhetier: nit: What "commit" the write is updating the ranges, here we only keep cached data in sync

```suggestion
    // Update the amount of data readable in the first chunk.
```

src/core/unittest/RecvBufferTest.cpp
```diff
@@ -2205,3 +2205,148 @@ TEST(RecvBufferResetReadTest, ResetClearsPendingAndExternalReference_SingleMode)
...
+    BOOLEAN NewDataReady = FALSE;
+
+    // Write [0, 20)
+    ASSERT_EQ(QUIC_STATUS_SUCCESS, RecvBuf.Write(0, 20, &InOutWriteLength, &NewDataReady));
+    ASSERT_TRUE(NewDataReady);
+
+    // Write [10, 30) — overlaps with [10, 20) already written
+    InOutWriteLength = LARGE_TEST_BUFFER_LENGTH;
+    NewDataReady = FALSE;
+    ASSERT_EQ(QUIC_STATUS_SUCCESS, RecvBuf.Write(10, 20, &InOutWriteLength, &NewDataReady));
```
guhetier: Currently, `Write` always write data based on the byte stream offset.
So even if the new code is incorrect and it override the old data, this test would not detect it: the new write and the old one are always identical.

We will need to improve the test helper to allow writing different data. Since `Read` also validate the data read follow what is expected, we will have to find a solution that provide the expected data to both read and write.

src/core/unittest/RecvBufferTest.cpp
```diff
@@ -2205,3 +2205,148 @@ TEST(RecvBufferResetReadTest, ResetClearsPendingAndExternalReference_SingleMode)
...
+    // Read and verify data integrity — buffer[i] should equal (uint8_t)i
+    uint64_t ReadOffset;
+    QUIC_BUFFER ReadBuffers[3];
+    uint32_t BufferCount = ARRAYSIZE(ReadBuffers);
+    RecvBuf.Read(&ReadOffset, &BufferCount, ReadBuffers);
+    ASSERT_EQ(0ull, ReadOffset);
+    ASSERT_GE(BufferCount, 1u);
+
+    // Validate first 20 bytes were not corrupted by the overlapping write
+    RecvBuffer::ValidateBuffer(ReadBuffers[0].Buffer, ReadBuffers[0].Length, ReadOffset);
```
guhetier: Not needed, already done as part of `Read`.

src/core/unittest/RecvBufferTest.cpp
guhetier: A possibility is to overload `Write` to allow providing the data to write.
We can then craft our write buffer to have "invalid" values (like `0xff`) for the bytes that should not be written

MarkedMuichiro:
> The product code update looks good, thanks for the contribution! The tests need some changes to actually validate the feature.

Thanks for the detailed feedback! I'm glad to help. Here's the approach I took in addressing your comments:

For the test issue: added a `WriteCustom` helper that accepts an explicit buffer, and a `ValidateBufferCustom` that validates against a provided expected array. The overlap tests now write `0xff` sentinel values in the already-written regions of the second write. If the fix were broken and those bytes got overwritten, the sentinels would appear in the buffer and `ValidateBufferCustom` would catch the corruption immediately.

Also removed the redundant `ValidateBuffer` call you flagged — it was already handled by `Read`.

src/core/unittest/RecvBufferTest.cpp
```diff
@@ -2205,3 +2246,222 @@ TEST(RecvBufferResetReadTest, ResetClearsPendingAndExternalReference_SingleMode)
...
+    }
+
+    uint64_t ReadOffset;
+    QUIC_BUFFER ReadBuffers[3];
+    uint32_t BufferCount = ARRAYSIZE(ReadBuffers);
+    QuicRecvBufferRead(&RecvBuf.RecvBuf, &ReadOffset, &BufferCount, ReadBuffers);
+    ASSERT_EQ(0ull, ReadOffset);
+    uint32_t TotalRead = 0;
+    for (uint32_t b = 0; b < BufferCount; ++b) {
+        RecvBuffer::ValidateBufferCustom(ReadBuffers[b].Buffer, ReadBuffers[b].Length, Expected + TotalRead);
```
guhetier: nit: We could simply call `memcmp` or `std::equal_n`.

src/core/unittest/RecvBufferTest.cpp
```diff
@@ -2205,3 +2246,222 @@ TEST(RecvBufferResetReadTest, ResetClearsPendingAndExternalReference_SingleMode)
...
     EXPECT_EQ(16ull, RecvBuf.RecvBuf.ReadPendingLength);
 }
+
+// Tests for overlapping write behavior - verifies that overlapping writes
+// do not corrupt previously received data.
+//
+// Each test uses WriteCustom to write sentinel values (0xff) in the overlap
+// region of the second write. If the fix is working, the original bytes
+// survive. If broken, the 0xff values would overwrite them and
+// ValidateBufferCustom would catch the corruption.
```
guhetier: Please do not mention "the fix" in comments: that is meaningless to a future reader.
The first sentence is enough to explain what the test does.

Several other occurrences below.

src/core/unittest/RecvBufferTest.cpp
MarkedMuichiro: Good call out, switched to `memcmp`. Thanks!

src/core/unittest/RecvBufferTest.cpp
MarkedMuichiro: Removed all references to 'the fix' — comments now only describe what the test does.

src/core/unittest/RecvBufferTest.cpp
```diff
@@ -249,6 +280,16 @@ struct RecvBuffer {
...
+    // Validates the buffer matches an explicitly provided expected array.
+    static void ValidateBufferCustom(
+        _In_reads_(BufferLength) const uint8_t* Buffer,
+        _In_ uint32_t BufferLength,
+        _In_reads_(BufferLength) const uint8_t* Expected
+        ) {
+        for (uint32_t i = 0; i < BufferLength; ++i) {
+            ASSERT_EQ(Expected[i], Buffer[i]);
+        }
+    }
```
guhetier: This is now dead code.

```suggestion
```

MarkedMuichiro:
> Looks good to me one the dead code is removed.

Done. Thank you.

guhetier:
Thanks for the contribution! I'll merge this once the CI succeeds.

MarkedMuichiro:
@guhetier Any idea why some test cases are failing?

guhetier:
There is a regression (or at least change of behavior) in CI runners. We just checked in a workaround for the "WinPrerelease" ones. I am investigating the linux one.

If you rebase, more should succeed, I don't think this is related to your PR in any way but we try to limit churn until we solve it.

src/core/unittest/RecvBufferTest.cpp
```diff
@@ -2205,3 +2236,220 @@ TEST(RecvBufferResetReadTest, ResetClearsPendingAndExternalReference_SingleMode)
...
+// region of the second write, so that any unintended overwrite of already-
+// received bytes would be detectable when validating the read output.
+
+TEST_P(WithMode, OverlapWritePreservesExistingData)
+{
+    //
+    // Write [0, 20) with pattern A (offset-based).
+    // Then write [10, 30) with 0xff in the overlap region [10, 20).
+    // Verify bytes [0, 20) still contain pattern A, and [20, 30) contain
+    // the new data from the second write.
```
sm-msft, line 2253: Could you please include tests for the following scenarios:
- 3 ranges that have 2 overlaps between them 
- 3 non-overlapping contiguous ranges
- 4 ranges - 3 with 2 range overlaps and one range that is non-contiguous with the others

MarkedMuichiro:
> There is a regression (or at least change of behavior) in CI runners. We just checked in a workaround for the "WinPrerelease" ones. I am investigating the linux one.
> 
> If you rebase, more should succeed, I don't think this is related to your PR in any way but we try to limit churn until we solve it.

Thanks for the update and for investigating the CI issue. Rebased onto main.

src/core/unittest/RecvBufferTest.cpp
MarkedMuichiro: Added `OverlapWriteThreeRangesTwoOverlaps` and `OverlapWriteThreeContiguousNonOverlapping` per your request. Thanks for the suggestion!

src/core/recv_buffer.c
```diff
@@ -611,15 +611,76 @@ QuicRecvBufferCopyIntoChunks(
...
+        //
+        if (Sub->Low >= WriteEnd) {
+            break;
+        }
+
+        //
+        // If this subrange ends before our current copy position, skip it.
+        //
+        if (Sub->Low + Sub->Count <= CopyOffset) {
+            continue;
```
sm-msft, line 653: We should trace this as a datapath event since there is no other tracking of this overlap or an indication of this occurring. A new type of DataPath event traced here seems appropriate, but I will leave it to @guhetier to comment on specifics.

guhetier: Receiving duplicated stream data is not uncommon or problematic, if a packet was considered lost and is eventually received.

Our logs on RX would already trace the stream frame. A trace here could be noisy and would be redundant.

MarkedMuichiro: Understood — no trace added.

guhetier:
One more rebase or fix seems to be needed, after the changes to remove Linux XDP that was blocking the pipeline. Sorry about this back and forth, last week was a bad one for our CI.

MarkedMuichiro:
> One more rebase or fix seems to be needed, after the changes to remove Linux XDP that was blocking the pipeline. Sorry about this back and forth, last week was a bad one for our CI.

No worries, rebased.

MarkedMuichiro:
Thanks for the approval, @guhetier! Anything else needed on my end? I see two tests have come back as neutral.

guhetier:
All good, thanks for the contribution!

PR #5940, opened 2026-04-12
Fix a link failure when QUIC_USE_EXTERNAL_OPENSSL=on and QUIC_BUILD_SHARED=off

masa-koz:
## Description
This pull request updates the OpenSSL integration logic in `CMakeLists.txt` to improve how pkg-config and transitive dependencies (like zlib and zstd) are handled, especially when building static libraries. The main goal is to ensure reliable detection and linking of OpenSSL and its required dependencies, reducing build errors related to missing libraries.

OpenSSL pkg-config handling improvements:
* Added logic to search for the OpenSSL `pkgconfig` directory under the specified `QUIC_OPENSSL_ROOT_DIR` and set the `PKG_CONFIG_PATH` environment variable accordingly, with a warning if not found.

Transitive dependency linking enhancements:
* Updated the way transitive dependencies (such as zlib and zstd) are handled by explicitly searching for and linking their static libraries if required by OpenSSL, addressing issues with incomplete propagation by `FindOpenSSL.cmake`.
* Improved messaging to clarify when dependencies like zlib or zstd are found and linked, or skipped if not found.

Link interface adjustments:
* Changed the linking approach so only `libcrypto` and `libssl` are added directly, and transitive dependencies are handled separately, to avoid duplicate or missing static linkages.

## Testing

No

## Documentation

No

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -787,17 +787,56 @@ if(QUIC_TLS_LIB STREQUAL "quictls" OR QUIC_TLS_LIB STREQUAL "openssl")
         else()
             if(QUIC_OPENSSL_ROOT_DIR)
                 set(OPENSSL_ROOT_DIR "${QUIC_OPENSSL_ROOT_DIR}")
+                find_path(OPENSSL_PKGCONFIG_DIR
+                    NAMES openssl.pc libcrypto.pc libssl.pc
+                    PATHS
+                        "${QUIC_OPENSSL_ROOT_DIR}/lib/pkgconfig"
+                        "${QUIC_OPENSSL_ROOT_DIR}/lib64/pkgconfig"
+                        "${QUIC_OPENSSL_ROOT_DIR}/share/pkgconfig"
+                    NO_DEFAULT_PATH
+                )
+                if(OPENSSL_PKGCONFIG_DIR)
+                    message(STATUS "Found OpenSSL pkgconfig dir: ${OPENSSL_PKGCONFIG_DIR}")
+                    set(ENV{PKG_CONFIG_PATH} "${OPENSSL_PKGCONFIG_DIR}")
+                else()
+                    message(WARNING "No pkgconfig directory found under ${QUIC_OPENSSL_ROOT_DIR}, pkg-config may use system OpenSSL metadata")
+                endif()
             endif()
             if(NOT BUILD_SHARED_LIBS)
                 set(OPENSSL_USE_STATIC_LIBS TRUE)
             endif()
             find_package(OpenSSL 3.5.0 REQUIRED)
+
             # OpenSSL::SSL and OpenSSL::Crypto imported targets propagate transitive dependencies
-            # (for example zstd). BUILD_SHARED_LIBS is OFF and those transitive static libs can be linked again via
-            # src/platform/CMakeLists.txt, causing duplicate/static link failures.
-            # Keep link ownership consistent to avoid duplicate transitive linkage.
+            # (for example zstd) when BUILD_SHARED_LIBS is OFF. However, the paths of those transitive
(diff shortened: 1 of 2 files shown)


guhetier:
Thanks for the fix.

I am not sure what is the best way to do it (my CMake-foo is not great), but we start to have a lot of logic related to how to build/link and external OpenSSL, spread among multiple files. I think it would be better if we could factor it in a CMakeLists that is pulled in when we are not using the submodule.

It doesn't need to be in this PR, but since you are the main contributor for this OpenSSL build configuration, I'm interested in your thoughts.

guhetier:
@masa-koz Looking at the details, I would rather not add these dependencies in MsQuic.
Building your OpenSSL with `no-zlib no-comp` should fix the build too (like we do for the submodule), and QUIC / TLS 1.3 forbids compression, so MsQuic doesn't need those libs.

masa-koz:
I'm sorry for my late repsonse.
I found that Debian had already adopted OpenSSL 3.5.5 for stable and specified enable-zstd enable-zlib for building OpenSSL.
https://salsa.debian.org/debian/openssl/-/blob/debian/trixie/debian/rules?ref_type=heads#L33

I suppose that the user of QUIC_USE_EXTERNAL_OPENSSL would like to use libcrypto and libssl on the system if any. So, we should link zlib and zstd at least when we find they are needed.

masa-koz:
@guhetier I've modified this PR to link zlib and zstd if needed. And I've moved adding zlib and zstd to target_link_libraries  from platform/CMakeLists.txt to top CMakeLists.txt.

guhetier:
> I'm sorry for my late repsonse. I found that Debian had already adopted OpenSSL 3.5.5 for stable and specified enable-zstd enable-zlib for building OpenSSL. https://salsa.debian.org/debian/openssl/-/blob/debian/trixie/debian/rules?ref_type=heads#L33
> 
> I suppose that the user of QUIC_USE_EXTERNAL_OPENSSL would like to use libcrypto and libssl on the system if any. So, we should link zlib and zstd at least when we find they are needed.

Thanks for the details. I think I disagree with the conclusion.
The system OpenSSL builds with zstd and zlib is not a problem as long as we dynamically link to it (which is our goal for OpenSSL 3.5, we are planning our migration to it by default as I speak).

If a user wants to statically link to OpenSSL 3.5, we are happy to expose a few knobs to facilitate it, but adding dependencies that are not needed by MsQuic goes too far. This level of detail can be configured by whoever needs it when they re-build MsQuic, if they really need to statically link against an OpenSSL with zstd/zlib enabled.

If I am missing something or you have a specific use case in mind, I'm happy to discuss it (either here or in private if needed).

masa-koz:
Thanks for your explanation. I agree with your explanation of reject if you mention my first PR, which links zlib and zstd everytime. However, my curtrent PR will check whether libcrypto.a needs zlib and zstd and link them if they are needed (https://github.com/microsoft/msquic/pull/5940/changes#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR819-R839). Do you think that it also goes too far?

guhetier:
Yes, I think so, at least without more context about what you try to achieve specifically.
I assume that someone in the business of statically linking MsQuic against an external OpenSSL, is also building OpenSSL and have control over its build config.

It isn't a goal for MsQuic to support statically linking to any configuration of OpenSSL. We are happy to facilitate custom configuration, but this looks like something that should be address in the OpenSSL build configuration or some additional build config you would provide for your project.

Our long-term goal is actually to mostly/only support dynamically linking to OpenSSL 3.5 once it is supported broadly enough, we can trust it is present for all our supported platforms.

masa-koz:
Thanks for your detail explanation. I've understood it. I'll close this PR.

PR #5948, opened 2026-04-14
fix(xdp): propagate XDP rule plumbing failures to caller

### LINKED ISSUE #5935: XDP rule configuration failures are suppressed in MsQuic
### Describe the bug

MsQuic has datapaths that explicitly requires XDP to be available and enabled.
However, failures to initialize XDP get silently ignored, and does not bubble up to the application.

```
   RawSocketCreateUdp                     → returns QUIC_STATUS
     └─ CxPlatTryAddSocket(...)           → returns QUIC_STATUS (checked ✅)
     └─ CxPlatDpRawPlumbRulesOnSocket(..) → void (NOT checked ❌)
          └─ CxPlatDpRawInterfaceAddRules → void, silently fails on alloc failure or rule overflow
               └─ CxPlatDpRawInterfaceUpdateRules → void, has TODO: "Figure out how to better handle failure"
                    └─ XdpCreateProgram → returns QUIC_STATUS (failure logged, then `continue`)
```

### Affected OS

- [x] Windows
- [ ] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

_No response_

### MsQuic version

main

XDP fails to plumb rules (out of memory errors)

### Expected behavior

Failure must bubble up to the app.

### Actual outcome

Failures get silently ignored.

### Additional details

_No response_

MarkedMuichiro:
Fixes #5935

## Problem

`CxPlatDpRawPlumbRulesOnSocket`, `CxPlatDpRawInterfaceAddRules`, and `CxPlatDpRawInterfaceUpdateRules` all returned `void`, silently dropping failures from `XdpCreateProgram`, allocation failures, and rule overflow. When XDP is required without OS socket fallback (e.g. CIBIR + XDP server sockets as introduced in #5798), a silent failure means the socket accepts no traffic with no indication to the application.

## Fix

Changed all three functions to return `QUIC_STATUS` and propagate errors up through the call chain to the caller.

**`CxPlatDpRawInterfaceUpdateRules`:** Tracks the first `XdpCreateProgram` failure across all queues and returns it after all queues are attempted.

**`CxPlatDpRawInterfaceAddRules`:** Returns `QUIC_STATUS_BUFFER_TOO_SMALL` on rule overflow, `QUIC_STATUS_OUT_OF_MEMORY` on alloc failure, and propagates the return value from `CxPlatDpRawInterfaceUpdateRules`. Fixed a pre-existing PortSet memory leak — the buffer is now freed when `AddRules` fails before copying the rule into `Interface->Rules`.

**`CxPlatDpRawPlumbRulesOnSocket`:** Propagates failures from `CxPlatDpRawInterfaceAddRules`. On failure, both Wildcard and non-Wildcard branches now break at the first failure and perform best-effort rollback of already-configured interfaces before returning. The Wildcard branch calls `CxPlatDpRawInterfaceRemoveRules` on interfaces that were already configured. The non-Wildcard branch clears the port bit on interfaces that were already configured.

**`datapath_raw.h`:** Updated declaration from `void` to `QUIC_STATUS`.

**`datapath_raw.c` (`RawSocketDelete`):** Deletion is best-effort — logs a warning on failure rather than silently discarding the return value.

**`datapath_raw_win.c`:** Creation path now checks the return value of `CxPlatDpRawPlumbRulesOnSocket` and calls `CxPlatRemoveSocket` to roll back on failure.

Note: `CxPlatDpRawInterfaceRemoveRules` is intentionally left as `void` — rule removal failures are out of scope for this fix.

## Testing

All 30 `DataPathTest` cases pass on Linux (`msquicplatformtest --gtest_filter="DataPathTest*"`).

The XDP rule plumbing cleanup paths in `CxPlatDpRawPlumbRulesOnSocket` are Windows-only — `datapath_raw_xdp_win.c` is excluded from the Linux build entirely. The existing `QUIC_TEST_DATAPATH_HOOKS` infrastructure operates at the packet layer in `binding.c` and cannot reach `CxPlatDpRawInterfaceAddRules` in the XDP platform layer. Wiring failure injection there would require adding a new callback to the hook struct — happy to do this if the team considers it worth the effort.

### DIFF
--- a/src/platform/datapath_raw.c
+++ b/src/platform/datapath_raw.c
@@ -174,7 +174,19 @@ RawSocketDelete(
     _In_ CXPLAT_SOCKET_RAW* Socket
     )
 {
-    CxPlatDpRawPlumbRulesOnSocket(Socket, FALSE);
+    //
+    // Rule removal on socket deletion is best-effort. A failure here means
+    // some XDP rules may linger until the interface is reinitialized, but
+    // there is nothing useful we can do at this point.
+    //
+    QUIC_STATUS PlumbStatus = CxPlatDpRawPlumbRulesOnSocket(Socket, FALSE);
+    if (QUIC_FAILED(PlumbStatus)) {
+        QuicTraceEvent(
+            LibraryErrorStatus,
+            "[ lib] ERROR, %u, %s.",
+            PlumbStatus,
+            "CxPlatDpRawPlumbRulesOnSocket (delete)");
+    }
     CxPlatRemoveSocket(&Socket->RawDatapath->SocketPool, Socket);
     CxPlatRundownReleaseAndWait(&Socket->RawRundown);
     if (Socket->PausedTcpSend) {
--- a/src/platform/datapath_raw_xdp_win.c
+++ b/src/platform/datapath_raw_xdp_win.c
@@ -938,11 +938,7 @@ CxPlatDpRawInterfaceAddRules(
         // they were not transferred to Interface->Rules.
         //
         for (uint8_t i = 0; i < Count; i++) {
-            if ((Rules[i].Match == XDP_MATCH_IPV4_UDP_PORT_SET ||
-                 Rules[i].Match == XDP_MATCH_IPV4_TCP_PORT_SET ||
-                 Rules[i].Match == XDP_MATCH_IPV6_UDP_PORT_SET ||
-                 Rules[i].Match == XDP_MATCH_IPV6_TCP_PORT_SET) &&
-                Rules[i].Pattern.IpPortSet.PortSet.PortSet != NULL) {
+            if (Rules[i].Pattern.IpPortSet.PortSet.PortSet) {
                 CxPlatFree(
                     (uint8_t*)Rules[i].Pattern.IpPortSet.PortSet.PortSet,
                     PORT_SET_TAG);
@@ -967,11 +963,7 @@ CxPlatDpRawInterfaceAddRules(
         // they were not transferred to Interface->Rules.
         //
         for (uint8_t i = 0; i < Count; i++) {
-            if ((Rules[i].Match == XDP_MATCH_IPV4_UDP_PORT_SET ||
-                 Rules[i].Match == XDP_MATCH_IPV4_TCP_PORT_SET ||
-                 Rules[i].Match == XDP_MATCH_IPV6_UDP_PORT_SET ||
-                 Rules[i].Match == XDP_MATCH_IPV6_TCP_PORT_SET) &&
-                Rules[i].Pattern.IpPortSet.PortSet.PortSet != NULL) {
+            if (Rules[i].Pattern.IpPortSet.PortSet.PortSet) {
                 CxPlatFree(
                     (uint8_t*)Rules[i].Pattern.IpPortSet.PortSet.PortSet,
                     PORT_SET_TAG);
--- a/src/platform/datapath_raw_xdp_win.c
+++ b/src/platform/datapath_raw_xdp_win.c
@@ -853,7 +853,7 @@ Error:
 
 _IRQL_requires_max_(PASSIVE_LEVEL)
 _Requires_lock_held_(Interface->RuleLock)
-void
+QUIC_STATUS
 CxPlatDpRawInterfaceUpdateRules(
     _In_ XDP_INTERFACE* Interface
     )
@@ -863,6 +863,8 @@ CxPlatDpRawInterfaceUpdateRules(
         .Direction = XDP_HOOK_RX,
         .SubLayer = XDP_HOOK_INSPECT,
     };
+
+    QUIC_STATUS ReturnStatus = QUIC_STATUS_SUCCESS;
 
     for (uint32_t i = 0; i < Interface->QueueCount; i++) {
 
@@ -882,16 +884,14 @@ CxPlatDpRawInterfaceUpdateRules(
                 Interface->RuleCount,
                 &NewRxProgram);
         if (QUIC_FAILED(Status)) {
-            //
-            // TODO - Figure out how to better handle failure and revert changes.
-            // This will likely require working with XDP to get an improved API;
-            // possibly to update all queues at once.
-            //
             QuicTraceEvent(
                 LibraryErrorStatus,
                 "[ lib] ERROR, %u, %s.",
                 Status,
                 "XdpCreateProgram");
+            if (QUIC_SUCCEEDED(ReturnStatus)) {
+                ReturnStatus = Status;
+            }
             continue;
         }
 
@@ -901,10 +901,18 @@ CxPlatDpRawInterfaceUpdateRules(
 
         Queue->RxProgram = NewRxProgram;
     }
+
+    return ReturnStatus;
 }
 
+//
+// Takes ownership of any PortSet buffers in the input rules. On any failure
+// path (pre-copy or post-copy via CxPlatDpRawInterfaceUpdateRules), the
+// PortSet buffer is freed by either this function or the interface — the
+// caller must not free it.
+//
 _IRQL_requires_max_(PASSIVE_LEVEL)
-void
+QUIC_STATUS
 CxPlatDpRawInterfaceAddRules(
     _In_ XDP_INTERFACE* Interface,
     _In_reads_(Count) const XDP_RULE* Rules,
@@ -913,6 +921,8 @@ CxPlatDpRawInterfaceAddRules(
 {
 #pragma warning(push)
 #pragma warning(disable:6386) // Buffer overrun while writing to 'NewRules' - FALSE POSITIVE
+
+    QUIC_STATUS Status;
 
     CxPlatLockAcquire(&Interface->RuleLock);
     // TODO - Don't always allocate a new array?
@@ -923,7 +933,22 @@ CxPlatDpRawInterfaceAddRules(
             "[ lib] ERROR, %s.",
             "No more room for rules");
         CxPlatLockRelease(&Interface->RuleLock);
-        return;
+        //
+        // We own any PortSet buffers in the input rules; free them since
+        // they were not transferred to Interface->Rules.
+        //
+        for (uint8_t i = 0; i < Count; i++) {
+            if ((Rules[i].Match == XDP_MATCH_IPV4_UDP_PORT_SET ||
+                 Rules[i].Match == XDP_MATCH_IPV4_TCP_PORT_SET ||
+                 Rules[i].Match == XDP_MATCH_IPV6_UDP_PORT_SET ||
+                 Rules[i].Match == XDP_MATCH_IPV6_TCP_PORT_SET) &&
+                Rules[i].Pattern.IpPortSet.PortSet.PortSet != NULL) {
+                CxPlatFree(
+                    (uint8_t*)Rules[i].Pattern.IpPortSet.PortSet.PortSet,
+                    PORT_SET_TAG);
+            }
+        }
+        return QUIC_STATUS_BUFFER_TOO_SMALL;
     }
 
     const size_t OldSize = sizeof(XDP_RULE) * (size_t)Interface->RuleCount;
@@ -937,7 +962,22 @@ CxPlatDpRawInterfaceAddRules(
             "XDP_RULE",
             NewSize);
         CxPlatLockRelease(&Interface->RuleLock);
-        return;
+        //
+        // We own any PortSet buffers in the input rules; free them since
+        // they were not transferred to Interface->Rules.
+        //
+        for (uint8_t i = 0; i < Count; i++) {
+            if ((Rules[i].Match == XDP_MATCH_IPV4_UDP_PORT_SET ||
+                 Rules[i].Match == XDP_MATCH_IPV4_TCP_PORT_SET ||
+                 Rules[i].Match == XDP_MATCH_IPV6_UDP_PORT_SET ||
+                 Rules[i].Match == XDP_MATCH_IPV6_TCP_PORT_SET) &&
+                Rules[i].Pattern.IpPortSet.PortSet.PortSet != NULL) {
+                CxPlatFree(
+                    (uint8_t*)Rules[i].Pattern.IpPortSet.PortSet.PortSet,
+                    PORT_SET_TAG);
+            }
+        }
+        return QUIC_STATUS_OUT_OF_MEMORY;
     }
 
     if (Interface->RuleCount > 0) {
@@ -952,11 +992,13 @@ CxPlatDpRawInterfaceAddRules(
     }
     Interface->Rules = NewRules;
 
-    CxPlatDpRawInterfaceUpdateRules(Interface);
+    Status = CxPlatDpRawInterfaceUpdateRules(Interface);
 
     CxPlatLockRelease(&Interface->RuleLock);
 
 #pragma warning(pop)
+
+    return Status;
 }
 
 _IRQL_requires_max_(PASSIVE_LEVEL)
@@ -1423,12 +1465,13 @@ CxPlatDpRawClearPortBit(
 }
 
 _IRQL_requires_max_(PASSIVE_LEVEL)
-void
+QUIC_STATUS
 CxPlatDpRawPlumbRulesOnSocket(
     _In_ CXPLAT_SOCKET_RAW* Socket,
     _In_ BOOLEAN IsCreated
     )
 {
+    QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
     XDP_DATAPATH* Xdp = (XDP_DATAPATH*)Socket->RawDatapath;
     if (Socket->Wildcard) {
         XDP_RULE Rules[5] = {0};
@@ -1504,7 +1547,17 @@ CxPlatDpRawPlumbRulesOnSocket(
         for (Entry = Xdp->Interfaces.Flink; Entry != &Xdp->Interfaces; Entry = Entry->Flink) {
             XDP_INTERFACE* Interface = CONTAINING_RECORD(Entry, XDP_INTERFACE, Link);
             if (IsCreated) {
-                CxPlatDpRawInterfaceAddRules(Interface, Rules, RulesSize);
+                QUIC_STATUS AddStatus = CxPlatDpRawInterfaceAddRules(Interface, Rules, RulesSize);
+                if (QUIC_FAILED(AddStatus)) {
+                    //
+                    // Stop on first failure and propagate. The caller is
+                    // responsible for invoking this function again with
+                    // IsCreated=FALSE to roll back any rules already
+                    // installed on previous interfaces.
+                    //
+                    Status = AddStatus;
+                    break;
+                }
             } else {
                 CxPlatDpRawInterfaceRemoveRules(Interface, Rules, RulesSize);
             }
@@ -1566,14 +1619,29 @@ CxPlatDpRawPlumbRulesOnSocket(
                             "Allocation of '%s' failed. (%llu bytes)",
                             "PortSet",
                             XDP_PORT_SET_BUFFER_SIZE);
-                        return;
+                        Status = QUIC_STATUS_OUT_OF_MEMORY;
+                        break;
                     }
                     CxPlatDpRawSetPortBit(
                         (uint8_t*)NewRule.Pattern.IpPortSet.PortSet.PortSet,
                         Socket->LocalAddress.Ipv4.sin_port);
                     memcpy(
                         &NewRule.Pattern.IpPortSet.Address, IpAddress, IpAddressSize);
-                    CxPlatDpRawInterfaceAddRules(Interface, &NewRule, 1);
+                    QUIC_STATUS AddStatus = CxPlatDpRawInterfaceAddRules(Interface, &NewRule, 1);
+                    if (QUIC_FAILED(AddStatus)) {
+                        Status = AddStatus;
+                        //
+                        // CxPlatDpRawInterfaceAddRules takes ownership of the
+                        // PortSet buffer on every path — including failure —
+                        // so we don't free it here.
+                        //
+                        // Stop on first failure and propagate. The caller is
+                        // responsible for invoking this function again with
+                        // IsCreated=FALSE to roll back any port bits already
+                        // set on previous interfaces.
+                        //
+                        break;
+                    }
                 }
             } else {
                 //
@@ -1588,6 +1656,8 @@ CxPlatDpRawPlumbRulesOnSocket(
             }
         }
     }
+
+    return Status;
 }
 
 _IRQL_requires_max_(DISPATCH_LEVEL)
--- a/src/platform/unittest/DataPathTest.cpp
+++ b/src/platform/unittest/DataPathTest.cpp
@@ -1357,4 +1357,55 @@ TEST_P(DataPathTest, TcpDataServer)
     ASSERT_TRUE(CxPlatEventWaitWithTimeout(ClientContext.ReceiveEvent, 500));
 }
 
+//
+// CxPlatSetAllocFailDenominator is only available in DEBUG builds, so
+// this test must be DEBUG-only.
+//
+#ifdef DEBUG
+TEST_F(DataPathTest, XdpRuleAddOomCleanup)
+{
+    //
+    // Verify the XDP rule plumbing failure path: when CxPlatDpRawInterfaceAddRules
+    // fails (e.g. due to allocation failure), CxPlatDpRawPlumbRulesOnSocket must
+    // propagate the error and RawSocketCreateUdp must invoke
+    // CxPlatDpRawPlumbRulesOnSocket(FALSE) for best-effort cleanup of any
+    // partially-installed state, then return failure to the caller without
+    // leaking resources or crashing. Only meaningful when running with the
+    // XDP/DuoNic datapath.
+    //
+    if (!UseDuoNic) {
+        GTEST_SKIP();
+    }
+
+    QUIC_GLOBAL_EXECUTION_CONFIG Config = { QUIC_GLOBAL_EXECUTION_CONFIG_FLAG_NONE, 0, 1, {0} };
+    CxPlatDataPath Datapath(&EmptyUdpCallbacks, nullptr, 0, &Config);
+    VERIFY_QUIC_SUCCESS(Datapath.GetInitStatus());
+    ASSERT_TRUE(Datapath.IsSupported(CXPLAT_DATAPATH_FEATURE_RAW, CXPLAT_SOCKET_FLAG_XDP));
+
+    //
+    // Create a client (connected) QTIP socket — providing a RemoteAddress
+    // makes RawSocketCreateUdp take the connected-socket path which reaches
+    // CxPlatDpRawPlumbRulesOnSocket. A non-wildcard local + no remote would
+    // be rejected with QUIC_STATUS_INVALID_STATE before rule plumbing runs.
+    // QTIP also disables the wildcard retry loop and the OS-socket fallback
+    // in CxPlatSocketCreateUdp so the rule-plumbing failure propagates
+    // deterministically.
+    //
+    QuicAddr RemoteAddr = GetNewLocalIPv4();
+
+    //
+    // Fail every 2nd allocation so the socket struct (alloc #1) succeeds and
+    // the XDP rule array allocation inside CxPlatDpRawInterfaceAddRules
+    // (alloc #2) is the one that fails, exercising both the failure-propagation
+    // path and the RawSocketCreateUdp-driven cleanup path.
+    //
+    CxPlatSetAllocFailDenominator(-2);
+    {
+        CxPlatSocket Socket(Datapath, nullptr, &RemoteAddr.SockAddr, nullptr, CXPLAT_SOCKET_FLAG_XDP | CXPLAT_SOCKET_FLAG_QTIP);
+        ASSERT_TRUE(QUIC_FAILED(Socket.GetInitStatus()));
+    }
+    CxPlatSetAllocFailDenominator(0);
+}
+#endif // DEBUG
+
 INSTANTIATE_TEST_SUITE_P(DataPathTest, DataPathTest, ::testing::Values(4, 6), testing::PrintToStringParamName());
--- a/src/generated/linux/datapath_raw.c.clog.h
+++ b/src/generated/linux/datapath_raw.c.clog.h
@@ -42,6 +42,26 @@ tracepoint(CLOG_DATAPATH_RAW_C, AllocFailure , arg2, arg3);\
 
 
 /*----------------------------------------------------------
+// Decoder Ring for LibraryErrorStatus
+// [ lib] ERROR, %u, %s.
+// QuicTraceEvent(
+            LibraryErrorStatus,
+            "[ lib] ERROR, %u, %s.",
+            PlumbStatus,
+            "CxPlatDpRawPlumbRulesOnSocket (delete)");
+// arg2 = arg2 = PlumbStatus = arg2
+// arg3 = arg3 = "CxPlatDpRawPlumbRulesOnSocket (delete)" = arg3
+----------------------------------------------------------*/
+#ifndef _clog_4_ARGS_TRACE_LibraryErrorStatus
+#define _clog_4_ARGS_TRACE_LibraryErrorStatus(uniqueId, encoded_arg_string, arg2, arg3)\
+tracepoint(CLOG_DATAPATH_RAW_C, LibraryErrorStatus , arg2, arg3);\
+
+#endif
+
+
+
+
+/*----------------------------------------------------------
 // Decoder Ring for DatapathRecv
 // [data][%p] Recv %u bytes (segment=%hu) Src=%!ADDR! Dst=%!ADDR!
 // QuicTraceEvent(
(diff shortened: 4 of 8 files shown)


src/platform/datapath_raw_linux.c
```diff
@@ -164,7 +164,11 @@ RawSocketCreateUdp(
         goto Error;
     }
 
-    CxPlatDpRawPlumbRulesOnSocket(NewSocket, TRUE);
+    Status = CxPlatDpRawPlumbRulesOnSocket(NewSocket, TRUE);
```
ProjectsByJackHe: We got rid of Linux XDP support. Not sure why the removal PR (copilot AI) didn't include these files https://github.com/microsoft/msquic/pull/5921

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1504,7 +1516,10 @@ CxPlatDpRawPlumbRulesOnSocket(
         for (Entry = Xdp->Interfaces.Flink; Entry != &Xdp->Interfaces; Entry = Entry->Flink) {
             XDP_INTERFACE* Interface = CONTAINING_RECORD(Entry, XDP_INTERFACE, Link);
             if (IsCreated) {
-                CxPlatDpRawInterfaceAddRules(Interface, Rules, RulesSize);
+                QUIC_STATUS AddStatus = CxPlatDpRawInterfaceAddRules(Interface, Rules, RulesSize);
+                if (QUIC_FAILED(AddStatus) && QUIC_SUCCEEDED(Status)) {
```
ProjectsByJackHe, line 1521: Why do we have this conditional instead of just returning the failure status here? Same question elsewhere in this function.

src/platform/datapath_raw_linux.c
MarkedMuichiro: Good catch — reverted `datapath_raw_linux.c`. Thanks!

src/platform/datapath_raw_xdp_win.c
MarkedMuichiro: Added a comment explaining the reasoning — we continue through all interfaces because we don't know which one hosts the socket's address until the optimization noted in the TODO is implemented. Added the same comment in the PortSet alloc failure path.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1566,14 +1586,25 @@ CxPlatDpRawPlumbRulesOnSocket(
...
+                        continue;
                     }
                     CxPlatDpRawSetPortBit(
                         (uint8_t*)NewRule.Pattern.IpPortSet.PortSet.PortSet,
                         Socket->LocalAddress.Ipv4.sin_port);
                     memcpy(
                         &NewRule.Pattern.IpPortSet.Address, IpAddress, IpAddressSize);
-                    CxPlatDpRawInterfaceAddRules(Interface, &NewRule, 1);
+                    QUIC_STATUS AddStatus = CxPlatDpRawInterfaceAddRules(Interface, &NewRule, 1);
+                    if (QUIC_FAILED(AddStatus) && QUIC_SUCCEEDED(Status)) {
```
guhetier: If we assume this can fail, should we now cleanup `NewRule.Pattern.IpPortSet.PortSet.PortSet`?

guhetier: The leak is actually not new, but now that we have an error code, we can fix it.

MarkedMuichiro:
Simplified the conditional — a success on a later interface no longer overwrites an earlier failure. Also fixed a potential use-after-free in the non-Wildcard branch: the PortSet free is now guarded by `QUIC_STATUS_OUT_OF_MEMORY` and `QUIC_STATUS_BUFFER_TOO_SMALL` only, since those are the two pre-copy failure paths where the caller still owns the buffer. On `XdpCreateProgram` failure the rule is already in `Interface->Rules` and the interface owns the buffer.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1504,7 +1516,15 @@ CxPlatDpRawPlumbRulesOnSocket(
         for (Entry = Xdp->Interfaces.Flink; Entry != &Xdp->Interfaces; Entry = Entry->Flink) {
             XDP_INTERFACE* Interface = CONTAINING_RECORD(Entry, XDP_INTERFACE, Link);
             if (IsCreated) {
-                CxPlatDpRawInterfaceAddRules(Interface, Rules, RulesSize);
+                //
+                // Continue iterating even on failure — for Wildcard sockets, rules must
```
guhetier: That doesn't seem right. On failure, we are going to fail creating the socket (RawSocketCreateUdp), so the rules we plumb won't matter.

Shouldn't a best-effort cleanup of the rules that got set be done instead, to avoid leaving unused rules if possible?

src/platform/datapath_raw.c
```diff
@@ -174,7 +174,7 @@ RawSocketDelete(
     _In_ CXPLAT_SOCKET_RAW* Socket
     )
 {
-    CxPlatDpRawPlumbRulesOnSocket(Socket, FALSE);
+    (void)CxPlatDpRawPlumbRulesOnSocket(Socket, FALSE);
```
guhetier: nit: Since we ignore the return value intentionally, lets comment about the removal being a best-effort and log a failure as a warning.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -952,11 +961,13 @@ CxPlatDpRawInterfaceAddRules(
     }
     Interface->Rules = NewRules;
 
-    CxPlatDpRawInterfaceUpdateRules(Interface);
+    Status = CxPlatDpRawInterfaceUpdateRules(Interface);
```
guhetier, line 964: I think that here too, if we fail to plumb the rules, we need to improve the cleanup now that we handle failures.
We probably can free the old rules later, so we can put them back if the new ones were not plumbed.

If new rules are only partially plumbed, it might get hard to keep things in sync, we may have to give it some thought (is it a problem if we have an XDP rule set but not in the array and we try to set it again?)

ProjectsByJackHe:
My biggest concern (when I created this issue) is partial instantiation...... and the non-determinism of it all. This isn't an issue with your PR, I am just braindumping the current problem we have and hopefully go for a more complete + deterministic solution. 

Today, MsQuic creates 1 program per interface per queue, and each program has its own rules engine. So, if a system has N interfaces, and each interface has K queues, MsQuic will create N * K XDP programs, and tries to replicate the same ruleset to those N * K programs.

Now if we fail to plumb rules on some programs, the flawed logic is we continue on failure and ignore it. It's good that your PR at least adds an error status, but that doesn't solve the partial instantiation, and the necessary cleanup. So, continuing on failure ideally should be avoided. We have to have a coherent cleanup strategy on failure.

Ideally, we have something atomic. Either all programs get their rules plumbed, or none do and the application can try it again later.

src/platform/datapath_raw_xdp_win.c
ProjectsByJackHe: +1. I added another comment in the PR. Continuing on error does not address the non-determinism. It's good that now at least an app will know of some failure, but would still be left in an undefined state.

MarkedMuichiro:
I can take care of the two concrete items — adding the best-effort comment and warning log in `RawSocketDelete`, and cleaning up already-installed rules on failure in the Wildcard branch before returning. For the atomic all-or-nothing guarantee across N×K programs, that sounds like it goes beyond what this PR can reasonably solve and might need a separate issue and possibly XDP API changes. Should I scope this PR to those cleanup improvements and track the larger problem separately, or would you all prefer a different approach?

ProjectsByJackHe:
> I can take care of the two concrete items — adding the best-effort comment and warning log in `RawSocketDelete`, and cleaning up already-installed rules on failure in the Wildcard branch before returning. For the atomic all-or-nothing guarantee across N×K programs, that sounds like it goes beyond what this PR can reasonably solve and might need a separate issue and possibly XDP API changes. Should I scope this PR to those cleanup improvements and track the larger problem separately, or would you all prefer a different approach?

That's fair, sorry for the scope creep. But the gist of it is, we should try to clean up at least on a best effort basis partially plumbed interfaces/queues. Maybe "**guarantee**" is a strong word :)

MarkedMuichiro:
> > I can take care of the two concrete items — adding the best-effort comment and warning log in `RawSocketDelete`, and cleaning up already-installed rules on failure in the Wildcard branch before returning. For the atomic all-or-nothing guarantee across N×K programs, that sounds like it goes beyond what this PR can reasonably solve and might need a separate issue and possibly XDP API changes. Should I scope this PR to those cleanup improvements and track the larger problem separately, or would you all prefer a different approach?
> 
> That's fair, sorry for the scope creep. But the gist of it is, we should try to clean up at least on a best effort basis partially plumbed interfaces/queues. Maybe "**guarantee**" is a strong word :)

Done — pushed the best-effort cleanup. On failure in the Wildcard branch the loop now breaks at the first failure and rolls back already-configured interfaces. Also added the warning log in `RawSocketDelete`. Best-effort is plenty :)

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1566,14 +1597,35 @@ CxPlatDpRawPlumbRulesOnSocket(
                             "Allocation of '%s' failed. (%llu bytes)",
                             "PortSet",
                             XDP_PORT_SET_BUFFER_SIZE);
-                        return;
+                        //
```
ProjectsByJackHe: We need to also add the cleanup for the partial rules that have already been plumbed here as well. Same thing as what you did above.

MarkedMuichiro: Ah yes, good catch; working on that now. Thanks!

ProjectsByJackHe:
There are still a couple of CLOG failures - you might need to clear auto-generated sidecar files then run update-sidecar.ps1 again.

src/platform/datapath_raw_xdp_win.c
ProjectsByJackHe: The logic of iterating on all interfaces to exhaustively apply the ruleset to get traffic to flow to a socket should fail if applying rules on any specific interface fails. Allowing it leads to undefined behavior. I guess since now you have updated the code to cleanup partially initialized sockets, and stop iterating on failure, that's a step in the right direction.

ProjectsByJackHe:
There is currently no coverage for the cleanup routines and your error handling code. 

If even as a sanity check, can you verify that there is no crash or undefined behavior when your failure path runs?

You can hard-code a conditional to go into that path through the usage of MsQuic Test Hooks. It would be ideal if this can be added to our automated CI.

But at a bare minimum, can you verify manually the cleanup code works as expected + update PR description with how you tested this?

MarkedMuichiro:
> There is currently no coverage for the cleanup routines and your error handling code.
> 
> If even as a sanity check, can you verify that there is no crash or undefined behavior when your failure path runs?
> 
> You can hard-code a conditional to go into that path through the usage of MsQuic Test Hooks. It would be ideal if this can be added to our automated CI.
> 
> But at a bare minimum, can you verify manually the cleanup code works as expected + update PR description with how you tested this?

I looked into using `QUIC_TEST_DATAPATH_HOOKS` but the existing callbacks (`Create`, `Send`, `Receive`, `GetLocalAddress`, `GetRemoteAddress`) operate at the packet layer in `binding.c` and can't reach `CxPlatDpRawInterfaceAddRules` in the XDP platform layer — so there's no existing hook point to inject a failure there. Wiring one in would require adding a new callback to the hook struct. Happy to do that if the team considers it worth it, just want guidance on the right approach before touching the hook infrastructure. In the meantime I ran all 30 DataPath tests on Linux and they pass cleanly, and updated the PR description with testing notes.

ProjectsByJackHe:
> > There is currently no coverage for the cleanup routines and your error handling code.
> > If even as a sanity check, can you verify that there is no crash or undefined behavior when your failure path runs?
> > You can hard-code a conditional to go into that path through the usage of MsQuic Test Hooks. It would be ideal if this can be added to our automated CI.
> > But at a bare minimum, can you verify manually the cleanup code works as expected + update PR description with how you tested this?
> 
> I looked into using `QUIC_TEST_DATAPATH_HOOKS` but the existing callbacks (`Create`, `Send`, `Receive`, `GetLocalAddress`, `GetRemoteAddress`) operate at the packet layer in `binding.c` and can't reach `CxPlatDpRawInterfaceAddRules` in the XDP platform layer — so there's no existing hook point to inject a failure there. Wiring one in would require adding a new callback to the hook struct. Happy to do that if the team considers it worth it, just want guidance on the right approach before touching the hook infrastructure. In the meantime I ran all 30 DataPath tests on Linux and they pass cleanly, and updated the PR description with testing notes.

I see. If it's do-able for you to add those hook callbacks to inject those failures in the Datapath, that would be great. If it's too much, then we can save it for another PR. But at least having a manual pass through the codepath to add some coverage is needed. BTW, you said "ran 30 DataPath tests on Linux" --- I am not sure how this relates to the current changes that are only in the WinUser xdp path. Did you mean windows?

MarkedMuichiro:
> > > There is currently no coverage for the cleanup routines and your error handling code.
> > > If even as a sanity check, can you verify that there is no crash or undefined behavior when your failure path runs?
> > > You can hard-code a conditional to go into that path through the usage of MsQuic Test Hooks. It would be ideal if this can be added to our automated CI.
> > > But at a bare minimum, can you verify manually the cleanup code works as expected + update PR description with how you tested this?
> > 
> > 
> > I looked into using `QUIC_TEST_DATAPATH_HOOKS` but the existing callbacks (`Create`, `Send`, `Receive`, `GetLocalAddress`, `GetRemoteAddress`) operate at the packet layer in `binding.c` and can't reach `CxPlatDpRawInterfaceAddRules` in the XDP platform layer — so there's no existing hook point to inject a failure there. Wiring one in would require adding a new callback to the hook struct. Happy to do that if the team considers it worth it, just want guidance on the right approach before touching the hook infrastructure. In the meantime I ran all 30 DataPath tests on Linux and they pass cleanly, and updated the PR description with testing notes.
> 
> I see. If it's do-able for you to add those hook callbacks to inject those failures in the Datapath, that would be great. If it's too much, then we can save it for another PR. But at least having a manual pass through the codepath to add some coverage is needed. BTW, you said "ran 30 DataPath tests on Linux" --- I am not sure how this relates to the current changes that are only in the WinUser xdp path. Did you mean windows?


Yes, I did mean Windows, sorry about that. I dug a bit deeper and found that the layering, `datapath_raw_xdp_win.c` has no include path to `core/library.h` so accessing `MsQuicLib.TestDatapathHooks` from the platform layer would invert the dependency graph — that's a real architectural violation. Here are two viable options:

**Option A:** Add a static function pointer `QUIC_STATUS (*CxPlatDpRawInterfaceAddRulesHook)(void) = NULL` in `datapath_raw_xdp_win.c` itself, consulted at the top of `CxPlatDpRawInterfaceAddRules` under `#if QUIC_TEST_DATAPATH_HOOKS_ENABLED`. Stays entirely within the platform layer, zero breakage elsewhere, covers all failure cases.

**Option B:** Use the existing `CxPlatSetAllocFailDenominator` mechanism to trigger the `QUIC_STATUS_OUT_OF_MEMORY` branch inside `CxPlatDpRawInterfaceAddRules` with zero new code — only covers the OOM path but may be sufficient.

Which approach would you prefer? I'm happy to consider other approaches as well.

ProjectsByJackHe:
> > > > There is currently no coverage for the cleanup routines and your error handling code.
> > > > If even as a sanity check, can you verify that there is no crash or undefined behavior when your failure path runs?
> > > > You can hard-code a conditional to go into that path through the usage of MsQuic Test Hooks. It would be ideal if this can be added to our automated CI.
> > > > But at a bare minimum, can you verify manually the cleanup code works as expected + update PR description with how you tested this?
> > > 
> > > 
> > > I looked into using `QUIC_TEST_DATAPATH_HOOKS` but the existing callbacks (`Create`, `Send`, `Receive`, `GetLocalAddress`, `GetRemoteAddress`) operate at the packet layer in `binding.c` and can't reach `CxPlatDpRawInterfaceAddRules` in the XDP platform layer — so there's no existing hook point to inject a failure there. Wiring one in would require adding a new callback to the hook struct. Happy to do that if the team considers it worth it, just want guidance on the right approach before touching the hook infrastructure. In the meantime I ran all 30 DataPath tests on Linux and they pass cleanly, and updated the PR description with testing notes.
> > 
> > 
> > I see. If it's do-able for you to add those hook callbacks to inject those failures in the Datapath, that would be great. If it's too much, then we can save it for another PR. But at least having a manual pass through the codepath to add some coverage is needed. BTW, you said "ran 30 DataPath tests on Linux" --- I am not sure how this relates to the current changes that are only in the WinUser xdp path. Did you mean windows?
> 
> Yes, I did mean Windows, sorry about that. I dug a bit deeper and found that the layering, `datapath_raw_xdp_win.c` has no include path to `core/library.h` so accessing `MsQuicLib.TestDatapathHooks` from the platform layer would invert the dependency graph — that's a real architectural violation. Here are two viable options:
> 
> **Option A:** Add a static function pointer `QUIC_STATUS (*CxPlatDpRawInterfaceAddRulesHook)(void) = NULL` in `datapath_raw_xdp_win.c` itself, consulted at the top of `CxPlatDpRawInterfaceAddRules` under `#if QUIC_TEST_DATAPATH_HOOKS_ENABLED`. Stays entirely within the platform layer, zero breakage elsewhere, covers all failure cases.
> 
> **Option B:** Use the existing `CxPlatSetAllocFailDenominator` mechanism to trigger the `QUIC_STATUS_OUT_OF_MEMORY` branch inside `CxPlatDpRawInterfaceAddRules` with zero new code — only covers the OOM path but may be sufficient.
> 
> Which approach would you prefer? I'm happy to consider other approaches as well.

Yeah you can just trigger the OOM branch with CxPlatSetAllocFailDenominator. The specific failure doesn't matter, I just want to make sure that the cleanup code runs and no egregious errors.

MarkedMuichiro:
> > > > > There is currently no coverage for the cleanup routines and your error handling code.
> > > > > If even as a sanity check, can you verify that there is no crash or undefined behavior when your failure path runs?
> > > > > You can hard-code a conditional to go into that path through the usage of MsQuic Test Hooks. It would be ideal if this can be added to our automated CI.
> > > > > But at a bare minimum, can you verify manually the cleanup code works as expected + update PR description with how you tested this?
> > > > 
> > > > 
> > > > I looked into using `QUIC_TEST_DATAPATH_HOOKS` but the existing callbacks (`Create`, `Send`, `Receive`, `GetLocalAddress`, `GetRemoteAddress`) operate at the packet layer in `binding.c` and can't reach `CxPlatDpRawInterfaceAddRules` in the XDP platform layer — so there's no existing hook point to inject a failure there. Wiring one in would require adding a new callback to the hook struct. Happy to do that if the team considers it worth it, just want guidance on the right approach before touching the hook infrastructure. In the meantime I ran all 30 DataPath tests on Linux and they pass cleanly, and updated the PR description with testing notes.
> > > 
> > > 
> > > I see. If it's do-able for you to add those hook callbacks to inject those failures in the Datapath, that would be great. If it's too much, then we can save it for another PR. But at least having a manual pass through the codepath to add some coverage is needed. BTW, you said "ran 30 DataPath tests on Linux" --- I am not sure how this relates to the current changes that are only in the WinUser xdp path. Did you mean windows?
> > 
> > 
> > Yes, I did mean Windows, sorry about that. I dug a bit deeper and found that the layering, `datapath_raw_xdp_win.c` has no include path to `core/library.h` so accessing `MsQuicLib.TestDatapathHooks` from the platform layer would invert the dependency graph — that's a real architectural violation. Here are two viable options:
> > **Option A:** Add a static function pointer `QUIC_STATUS (*CxPlatDpRawInterfaceAddRulesHook)(void) = NULL` in `datapath_raw_xdp_win.c` itself, consulted at the top of `CxPlatDpRawInterfaceAddRules` under `#if QUIC_TEST_DATAPATH_HOOKS_ENABLED`. Stays entirely within the platform layer, zero breakage elsewhere, covers all failure cases.
> > **Option B:** Use the existing `CxPlatSetAllocFailDenominator` mechanism to trigger the `QUIC_STATUS_OUT_OF_MEMORY` branch inside `CxPlatDpRawInterfaceAddRules` with zero new code — only covers the OOM path but may be sufficient.
> > Which approach would you prefer? I'm happy to consider other approaches as well.
> 
> Yeah you can just trigger the OOM branch with CxPlatSetAllocFailDenominator. The specific failure doesn't matter, I just want to make sure that the cleanup code runs and no egregious errors.

Added `XdpRuleAddOomCleanup` test to `DataPathTest.cpp` that uses `CxPlatSetAllocFailDenominator(-2)` to force the XDP rule array allocation inside `CxPlatDpRawInterfaceAddRules` to fail — the socket struct allocation succeeds first, then the rule allocation fails, triggering the cleanup path in `CxPlatDpRawPlumbRulesOnSocket`. The test is DEBUG-only and runtime-skips when DuoNic is unavailable, consistent with other XDP tests. Verified all 30 existing DataPath tests still pass on Linux (`msquicplatformtest --gtest_filter=\"DataPathTest*\"`). The new test requires a DuoNic-capable Windows runner to actually execute — let me know if CI has one available, otherwise this test will runtime-skip and the cleanup verification remains theoretical.

src/platform/datapath_raw_xdp_win.c
MarkedMuichiro: Good point — here's what I'm thinking for this: defer freeing the old rules array until after `CxPlatDpRawInterfaceUpdateRules` succeeds. If it fails, free the new array, restore `Interface->Rules` and `Interface->RuleCount` to the old values, and do a best-effort re-apply of the old rules to resync queues that got the new rules installed before failure.

That brings software state back to a known-good baseline on failure. I can't fully fix the hardware divergence for queues where the re-apply itself fails — that's the limitation the existing TODO in `UpdateRules` acknowledges — but this at least keeps `Interface->Rules` consistent and avoids the double-count scenario where a later `AddRules` for the same port would duplicate a rule.

Scope is ~10 lines in `CxPlatDpRawInterfaceAddRules`, no signature changes. Does that match what you had in mind?

MarkedMuichiro:
I'm hitting a CLOG regeneration blocker — running `update-sidecar.ps1` in Codespaces/Linux fails on an unrelated `WarnNoXdpForCibirSockets` collision in `datapath_xplat.c`, and using `CLOG_DEVELOPMENT_MODE=1` to bypass it produces sidecars in a different order than CI expects (which is why Validate keeps failing).

The diff CI wants is just a reordering of the `LibraryErrorStatus` block in `datapath_raw.c.clog.h` and `datapath_raw.c.clog.h.lttng.h` — the content is identical, only the position of the block differs. Any suggestion on how to get CI-canonical sidecars from a Linux environment? Or can someone regenerate them on Windows?

ProjectsByJackHe:
A bunch of BVT CI failures @MarkedMuichiro 

For the CLOG errors, have you tried just cloning the repo to your local environment instead of using Github codespaces? 
Our scripts generally are the most stable in a windows environment.

guhetier:
For the CLOG failures, if running the script causes issues, the CI failure should report the diff and you can fix it by hand.
It might be some whitespace / return character problem.
https://github.com/microsoft/msquic/actions/runs/24915489496/job/73243800270?pr=5948

MarkedMuichiro:
Pushed the trailing-space fix for the CLOG diff. Workflows are waiting on maintainer approval to run — could either of you approve when you get a chance @guhetier @ProjectsByJackHe ?

MarkedMuichiro:
`XdpRuleAddOomCleanup` is failing on the Windows BVTs because `Socket.GetInitStatus()` returned success — the `CxPlatSetAllocFailDenominator(-2)` value doesn't reliably fail the rule array allocation in this CI environment. I think `-2` was based on assumed allocation ordering that may differ on Windows XDP. I have two options: (1) reduce the test to just verify no crash regardless of init status, or (2) try a different denominator value. Since the goal of the test was just to ensure the cleanup path doesn't crash, would option 1 be acceptable? It would change `ASSERT_TRUE(QUIC_FAILED(...))` to just running the code and asserting no crash.

guhetier:
> `XdpRuleAddOomCleanup` is failing on the Windows BVTs because `Socket.GetInitStatus()` returned success — the `CxPlatSetAllocFailDenominator(-2)` value doesn't reliably fail the rule array allocation in this CI environment. I think `-2` was based on assumed allocation ordering that may differ on Windows XDP. I have two options: (1) reduce the test to just verify no crash regardless of init status, or (2) try a different denominator value. Since the goal of the test was just to ensure the cleanup path doesn't crash, would option 1 be acceptable? It would change `ASSERT_TRUE(QUIC_FAILED(...))` to just running the code and asserting no crash.

A test that depends on the order / number of memory allocations will be brittle, I don't think we should merge it.
Adding a unit test validating the XDP path can be initialized successfully is probably useful if we don't already have one, but does little to help validating this specific change.

A manual validation, modifying the datapath code to pretend an allocation fails, and checking the error path execute as expected would be welcome.

@ProjectsByJackHe since it can be difficult to setup XDP, maybe you can assist with running this validation?

src/platform/unittest/DataPathTest.cpp
```diff
@@ -1357,4 +1357,37 @@ TEST_P(DataPathTest, TcpDataServer)
     ASSERT_TRUE(CxPlatEventWaitWithTimeout(ClientContext.ReceiveEvent, 500));
 }
 
+#ifdef DEBUG
```
guhetier, line 1364: If we keep this test, please add a comment to point at why it must be enabled for DEBUG only
(`CxPlatSetAllocFailDenominator`)

src/platform/datapath_raw_xdp_win.c
guhetier: nit: Lets remove this comment now that we do handle errors.

src/platform/datapath_raw_xdp_win.c
guhetier: Actually, looking back at it, I think it was a bad idea. See my other comments.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1423,12 +1434,13 @@ CxPlatDpRawClearPortBit(
 }
 
 _IRQL_requires_max_(PASSIVE_LEVEL)
-void
+QUIC_STATUS
 CxPlatDpRawPlumbRulesOnSocket(
```
guhetier, line 1461: Looking back at this function, I think we could simplify this best-effort cleanup by pushing it up. Since `CxPlatDpRawPlumbRulesOnSocket` already handles adding and removing rules, we can leverage it.

- `CxPlatDpRawPlumbRulesOnSocket(socket, TRUE)` set the rules as a best effort. It stops at the first error and return a failure.
- `CxPlatDpRawPlumbRulesOnSocket(socket, TRUE)` remove the rules as a best effort. It continues on error and returns a failure.

If `CxPlatDpRawPlumbRulesOnSocket(socket, TRUE)` fails, we can call `CxPlatDpRawPlumbRulesOnSocket(socket, FALSE)` directly to do a best effort cleanup + log it. That way, we don't have to duplicate most of the code. We are going to try to remove some rules that have not been set yet, but that should not be a problem, XDP is simply going to return an error code.

This should let us remove most of the more complicated cleanup code from this function, and only focus on returning an error and not leaking resources.

guhetier: @ProjectsByJackHe let me know if you see an issue with that approach.

ProjectsByJackHe: That approach sounds good to me, let's leverage existing functions to clean up rules.

ProjectsByJackHe:
Hey @MarkedMuichiro, I created https://github.com/MarkedMuichiro/msquic/pull/1 to fix up the BVT errors and confirmed from a local debugger that the added cleanup path has code coverage from my fixed up DataPathTest. Once we get that merged, and clean up some nits from @guhetier, this PR should be good to go!

MarkedMuichiro:
Pushed. Three changes:

- Reverted `CxPlatDpRawInterfaceUpdateRules` back to `void` — per-queue XdpCreateProgram failures are now best-effort as before.
- Removed the obsolete TODO comment in the `XdpCreateProgram` failure branch.
- Added a comment above `#ifdef DEBUG` in `DataPathTest.cpp` explaining why the test must be DEBUG-only (`CxPlatSetAllocFailDenominator` is debug-only).

Also includes Jack's PR #1 changes for the simplified cleanup approach — `RawSocketCreateUdp` now invokes `PlumbRulesOnSocket(socket, FALSE)` for best-effort cleanup on install failure, and the test was updated to use a connected QTIP socket so the failure propagates deterministically.

ProjectsByJackHe (dismissed):
It seems a bit too much got reverted after my helper PR got merged into this topic branch.

src/platform/datapath_raw_xdp_win.c
guhetier: Don't we need to propagate the error code from here?

guhetier, line 895: Shouldn't this report the error (and potentially stop the iteration)?

If the error is not reported, we will never get to the best-effort rollback

guhetier:
> Reverted CxPlatDpRawInterfaceUpdateRules back to void — per-queue XdpCreateProgram failures are now best-effort as before.

I'm confused by that part, why? Am I missing a problem with still returning an error from CxPlatDpRawInterfaceUpdateRules and keeping the rollback as it is done now?

ProjectsByJackHe:
> * Reverted `CxPlatDpRawInterfaceUpdateRules` back to `void` — per-queue XdpCreateProgram failures are now best-effort as before.

@MarkedMuichiro we still need to propagate this up. You can just set the HEAD of this PR to equal the HEAD of https://github.com/microsoft/msquic/pull/5968.

To clear up the objective, our goal is to notify the app whenever *any* problems occur when setting up XDP for data paths that requires it (i.e. does not have a fallback option (QTIP, CIBIR))

We can have the cleanup be best-effort and non-deterministic. Surprisingly, `PlumbRules(FALSE)` is a reasonably robust cleanup routine, even for partially initialized programs (if XdpCreateProgram fails).

MarkedMuichiro:
Updated branch to align with #5968 — `CxPlatDpRawInterfaceUpdateRules` and `CxPlatDpRawInterfaceAddRules` propagate the error code, and the simplified cleanup pattern remains: on install failure, `RawSocketCreateUdp` invokes `PlumbRulesOnSocket(socket, FALSE)` for best-effort rollback before returning the error to the caller.

Also addressed the two remaining nits:
- Removed the obsolete TODO comment in the `XdpCreateProgram` failure branch
- Added a comment above `#ifdef DEBUG` in `DataPathTest.cpp` explaining the test is DEBUG-only because of `CxPlatSetAllocFailDenominator`

Sorry for the earlier confusion @guhetier — I misread your previous comment about `UpdateRules` returning `void`. Thanks for the helper PR @ProjectsByJackHe.

src/platform/datapath_raw_xdp_win.c
```diff
@@ -1566,14 +1583,37 @@ CxPlatDpRawPlumbRulesOnSocket(
...
                     CxPlatDpRawSetPortBit(
                         (uint8_t*)NewRule.Pattern.IpPortSet.PortSet.PortSet,
                         Socket->LocalAddress.Ipv4.sin_port);
                     memcpy(
                         &NewRule.Pattern.IpPortSet.Address, IpAddress, IpAddressSize);
-                    CxPlatDpRawInterfaceAddRules(Interface, &NewRule, 1);
+                    QUIC_STATUS AddStatus = CxPlatDpRawInterfaceAddRules(Interface, &NewRule, 1);
+                    if (QUIC_FAILED(AddStatus)) {
+                        Status = AddStatus;
+                        if (AddStatus == QUIC_STATUS_OUT_OF_MEMORY ||
```
guhetier: Let's not have conditional memory management.
If `CxPlatDpRawInterfaceAddRules` is meant to take ownership of the `PortSet`, it should always take ownership of it and clean it up on failure.

guhetier: This is also currently incorrect, `XdpCreateProgram` could return `QUIC_STATUS_OUT_OF_MEMORY` and lead us in this path while we should not release the buffer.

MarkedMuichiro: Good catch, I will fix the logic and refactor the code. Thank you.

ProjectsByJackHe: Suggestion:

Put the cleanup for freeing the `PortSet` buffers in the `AddRules` function in code paths where it exits early, before it has a chance to set the pointer to `Interface->Rules.`

ProjectsByJackHe: I can open another helper PR for this if you'd like.

MarkedMuichiro: Thanks @ProjectsByJackHe — I just pushed that exact refactor. `CxPlatDpRawInterfaceAddRules` now takes ownership of any input PortSet buffers and frees them on its pre-copy failure paths (`BUFFER_TOO_SMALL` and `OUT_OF_MEMORY`). After the rule is copied into `Interface->Rules`, ownership transfers and `CxPlatDpRawInterfaceRemoveRules` handles eventual cleanup. The conditional free at the caller is gone — that was the use-after-free path when `XdpCreateProgram` returned `OUT_OF_MEMORY`. Appreciate the offer to open a helper PR but I think we're good now.

ProjectsByJackHe (approved):

guhetier (approved):

PR #5949, opened 2026-04-14
Add a CI failure investigation skill to the repository

guhetier:
## Description

Add a skill generated using prompt kit and tuned over a few bug investigations to more easily investigate CI failures.

The skill:
- guide the agent through an investigation based on a link to a CI run failure or auto-filed bug
- pushes the agent to always investigate based on logs and hard data
- contains a list of well-known issues we won't / can't fix
     - this should be used only for issues outside of our control (like overprovisioned runners), not what we can fix, like poor test synchronization.

It still requires a decent understanding of the test and code to know when to push the agent to look further

## Testing

Tested on 3 CI failures, it correctly identified the root causes with human assistance.

## Documentation

N/A


mtfriesen (dismissed):

guhetier:
Added a few improvements based on running the skill on the CI failure for this PR, trying to get the agent to investigate the issue more in depth without having to push it. The updated version found the right root cause (known, see #5899) without additional probing questions.

anrossi (approved):

PR #5953, opened 2026-04-17
fix(retry): ensure Retry Source CID differs from client DCID

### LINKED ISSUE #5302: QUIC: Retry Source CID May Match Client DCID
## Description

According to RFC 9000 [17.2.5.1. ](https://www.rfc-editor.org/rfc/rfc9000.html#section-17.2.5.1)[Sending a Retry Packet](https://www.rfc-editor.org/rfc/rfc9000.html#name-sending-a-retry-packet)

The server includes a connection ID of its choice in the Source Connection ID field. This value MUST NOT be equal to the Destination Connection ID field of the packet sent by the client. A client MUST discard a Retry packet that contains a Source Connection ID field that is identical to the Destination Connection ID field of its Initial packet.

In the [QUIC_OPER_TYPE_RETRY code path](https://github.com/microsoft/msquic/blob/2623c07df62b4bd171f469fb29c2714b6735b676/src/core/binding.c#L978), the implementation generates a new Source Connection ID (CID) using random bytes. However, it does not explicitly check that the generated Source CID differs from the client’s Destination CID, as required by the RFC above.

https://github.com/microsoft/msquic/blob/2623c07df62b4bd171f469fb29c2714b6735b676/src/core/binding.c#L976-L978

This correctly generates a random Source CID for the Retry packet. However, there is no check to ensure it does not accidentally match the client’s DCID. This check is missing, and while collisions are rare, they are possible — leading to non-compliant Retry packets:
if (memcmp(NewDestCid, RecvPacket->DestCid, RecvPacket->DestCidLen) == 0)

MarkedMuichiro:
Fixes #5302

## Problem

When generating a Retry packet, `CxPlatRandom` produces a random Source Connection ID but does not check that it differs from the client's Destination CID. RFC 9000 §17.2.5.1 requires that the Source Connection ID in a Retry packet MUST NOT equal the Destination Connection ID sent by the client.

While a collision is astronomically unlikely, the check is required for RFC compliance.

## Fix

Wrapped the `CxPlatRandom` call in a do-while loop that regenerates if the new CID matches the client's DCID:

```c
do {
    CxPlatRandom(MsQuicLib.CidTotalLength, NewDestCid);
} while (RecvPacket->DestCidLen == MsQuicLib.CidTotalLength &&
         memcmp(NewDestCid, RecvPacket->DestCid, MsQuicLib.CidTotalLength) == 0);
```

The length guard is load-bearing — the client DCID length is client-controlled and can legitimately differ from `CidTotalLength`, so skipping it would risk comparing against a shorter buffer.

This also fixes a minor issue in the original: `CxPlatRandom` was called with `sizeof(NewDestCid)` which randomizes the full `QUIC_CID_MAX_LENGTH` buffer, but only `MsQuicLib.CidTotalLength` bytes are ever used downstream. The fix randomizes only the bytes that matter.

### DIFF
--- a/src/core/binding.c
+++ b/src/core/binding.c
@@ -995,9 +995,18 @@ QuicBindingProcessStatelessOperation(
             goto Exit;
         }
 
+        const uint8_t CidTotalLength = MsQuicLib.CidTotalLength;
         uint8_t NewDestCid[QUIC_CID_MAX_LENGTH];
-        CXPLAT_DBG_ASSERT(sizeof(NewDestCid) >= MsQuicLib.CidTotalLength);
-        CxPlatRandom(sizeof(NewDestCid), NewDestCid);
+        CXPLAT_DBG_ASSERT(sizeof(NewDestCid) >= CidTotalLength);
+        //
+        // RFC 9000 sec. 17.2.5.1: "This value MUST NOT be equal to the
+        // Destination Connection ID field of the packet sent by the client."
+        // Regenerate until the new Source CID differs from the client's DCID.
+        //
+        do {
+            CxPlatRandom(CidTotalLength, NewDestCid);
+        } while (RecvPacket->DestCidLen == CidTotalLength &&
+                 memcmp(NewDestCid, RecvPacket->DestCid, CidTotalLength) == 0);
 
         QUIC_TOKEN_CONTENTS Token = { 0 };
         Token.Authenticated.Timestamp = (uint64_t)CxPlatTimeEpochMs64();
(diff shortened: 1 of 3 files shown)


src/core/binding.c
```diff
@@ -997,7 +997,10 @@ QuicBindingProcessStatelessOperation(
 
         uint8_t NewDestCid[QUIC_CID_MAX_LENGTH];
         CXPLAT_DBG_ASSERT(sizeof(NewDestCid) >= MsQuicLib.CidTotalLength);
-        CxPlatRandom(sizeof(NewDestCid), NewDestCid);
+        do {
```
guhetier, line 1005: Please leave a comment pointing at the RFC section that state CIDs must be different.

MarkedMuichiro: Done.

src/core/binding.c
```diff
@@ -997,7 +997,15 @@ QuicBindingProcessStatelessOperation(
...
         uint8_t NewDestCid[QUIC_CID_MAX_LENGTH];
         CXPLAT_DBG_ASSERT(sizeof(NewDestCid) >= MsQuicLib.CidTotalLength);
-        CxPlatRandom(sizeof(NewDestCid), NewDestCid);
+        //
+        // RFC 9000 sec. 17.2.5.1: "This value MUST NOT be equal to the
+        // Destination Connection ID field of the packet sent by the client."
+        // Regenerate until the new Source CID differs from the client's DCID.
+        //
+        do {
+            CxPlatRandom(MsQuicLib.CidTotalLength, NewDestCid);
```
guhetier: We should make a local copy of `MsQuicLib.CidTotalLength` before the ASSERT so it stays consistent in case it is concurrently changed from `QuicLibApplyLoadBalancingSetting`.

This is unlikely and the same issue exists in the codebase already, but let's be better.

guhetier (commented):
Thanks for the fix. 
Looks good, a minor suggestion to avoid some potential concurrency issues.

src/core/binding.c
MarkedMuichiro: Thanks for the suggestion! Added a local snapshot of `MsQuicLib.CidTotalLength` at the top of the Retry block and updated all downstream references to use it — assert, random generation, loop condition, memcmp, IV construction, encode call, and trace log.

guhetier (dismissed):
Looks good to me. Thanks for the fix!

guhetier:
You'll need to run the update-sidecar.ps1 script to fix the loging.

MarkedMuichiro:
> Looks good to me. Thanks for the fix!

You're welcome! Glad to help!

MarkedMuichiro:
> You'll need to run the update-sidecar.ps1 script to fix the loging.

Done — ran `update-sidecar.ps1` and pushed the regenerated sidecars.

guhetier (approved):

PR #5962, opened 2026-04-23
Fix null deref in QuicCryptoCustomCertValidationComplete after shutdown

guhetier:
## Description

Fix a null pointer dereference crash in `QuicCryptoCustomCertValidationComplete` when certificate validation completes after the connection has already been shut down.

**Root cause**: When a remote peer closes a QUIC connection while async certificate validation is pending on the server, `QuicConnOnShutdownComplete` frees all SourceCids. When the app later calls `ConnectionCertificateValidationComplete(TRUE)`, `QuicCryptoProcessTlsCompletion` dereferences `Connection->SourceCids.Next` (now NULL), causing an access violation. This was observed as a kernel bugcheck (0x7E) in production.

**Fix**: Add a `ShutdownComplete` guard in `QuicCryptoCustomCertValidationComplete` to bail out early if the connection has already completed shutdown, preventing the code from proceeding into `QuicCryptoProcessTlsCompletion` with freed state.

## Testing

Two new integration tests added:
- `Handshake.CustomServerCertValidationAfterShutdown`  Client-side async cert validation with client shutdown before completion.
- `Handshake.CustomClientCertValidationAfterShutdown`  Server-side async client cert validation with client disconnect before server completes. Mirrors the exact production crash scenario.

Both tests use event-based synchronization (`WaitForPeerCertReceived`) to deterministically ensure cert validation is pending before shutdown.

## Documentation

No documentation impact.

### DIFF
--- a/src/core/crypto.c
+++ b/src/core/crypto.c
@@ -1740,15 +1740,26 @@ QuicCryptoCustomCertValidationComplete(
     _In_ QUIC_TLS_ALERT_CODES TlsAlert
     )
 {
+    QUIC_CONNECTION* Connection = QuicCryptoGetConnection(Crypto);
+
     if (!Crypto->CertValidationPending) {
         return;
     }
 
     Crypto->CertValidationPending = FALSE;
+
+    //
+    // The connection might have been shutdown during the cert validation.
+    // Nothing to do in that case.
+    //
+    if (Connection->State.ShutdownComplete) {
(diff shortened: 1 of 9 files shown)


guhetier:
Note: There is a similar issue that will be addressed by #5963.

mtfriesen (approved):

PR #5963, opened 2026-04-24
Fix null deref in QuicCryptoCustomTicketValidationComplete after shutdown

guhetier:
## Description

This PR contains two related fixes for the deferred ticket validation path:

### 1. Null deref in `QuicCryptoCustomTicketValidationComplete` after shutdown

Fix a null pointer dereference when ticket validation completes after the
connection has shut down. This is the same class of bug fixed for cert
validation in #5962, but for the ticket (resumption) validation path.

After `ShutdownComplete`, `QuicConnOnShutdownComplete` frees all SourceCids via
`QuicLookupRemoveLocalCids`, leaving `Connection->SourceCids.Next == NULL`. If
the app then completes the deferred ticket validation (accept or reject), the
code calls `QuicCryptoProcessDataComplete` or `QuicCryptoInitializeTls`, both
of which dereference the freed SourceCids.

The fix adds a `ShutdownComplete` guard that clears `TicketValidationPending`
and `PendingValidationBufferLength` and returns early, mirroring the cert
validation fix.

### 2. Infinite send loop on `CONNECTION_CLOSE` when `WriteKey` is 0-RTT

While adding the new test, an unrelated pre-existing bug in the send loop
surfaced. When a client uses a pre-shared key (PSK) for resumption,
OpenSSL/QuicTLS derives 0-RTT keys preemptively while building the
ClientHello, advancing `Connection->Crypto.TlsState.WriteKey` to
`QUIC_PACKET_KEY_0_RTT`, even though no 0-RTT data is being sent. If the
client app calls `ConnectionShutdown` before the ServerHello arrives,
`QuicSendFlush` enters an infinite loop:

- `QuicPacketBuilderGetPacketTypeAndKeyForControlFrames` correctly refuses to
  put a `CONNECTION_CLOSE` frame in a 0-RTT packet and substitutes `Initial`
  as the packet key (RFC 9000 12.5).
- `QuicSendWriteFrames` then fails to clear the close send-flag, because
  `Builder->Key->Type` (Initial) `!= WriteKey` (0-RTT).
- The next iteration picks Initial again (0-RTT < HANDSHAKE so no downgrade),
  and the loop spins until the debug deadlock-detection counter aborts the
  process.

The fix adds a disjunct to the clear-flag check: when `WriteKey == 0-RTT`,
the close just got written at Initial and Initial is the only level it can
ever go out at (there is no higher level to upgrade to without first
receiving the ServerHello, which would advance `ReadKey` and trigger a new
send). The flag is cleared immediately in that case.

## Testing

New parameterized test `Handshake/WithAcceptTicket.CustomTicketValidationAfterShutdown`
(Accept/Reject) covers both fixes. The test primes resumption, defers ticket
validation with `QUIC_STATUS_PENDING`, shuts down the client while validation
is pending, then completes validation on the server. Without fix 1) the
server crashes on null deref; without fix 2) the client hangs/asserts in the
send loop.

## Documentation

No documentation impact.


src/test/lib/HandshakeTest.cpp
```diff
@@ -1423,6 +1423,118 @@ QuicTestCustomClientCertValidationAfterShutdown()
     TEST_QUIC_SUCCEEDED(Server->SetCustomValidationResult(TRUE));
 }
 
+void
+QuicTestCustomTicketValidationAfterShutdown(
+    bool AcceptTicket
+    )
+{
+#ifdef QUIC_DISABLE_0RTT_TESTS
```
guhetier: Need to remove this.

mtfriesen (approved):

anrossi (approved):

PR #5969, opened 2026-04-30
Add Ubuntu 26.04 package support

wfurt:
Adds support for Ubuntu 26.04 that was released to public earlier this month.

### DIFF
--- a/.github/workflows/package-linux.yml
+++ b/.github/workflows/package-linux.yml
@@ -33,6 +33,9 @@ jobs:
           { config: "Release", os: "ubuntu-24.04", arch: "arm", tls: "quictls", time64: "-Time64Distro" },
           { config: "Release", os: "ubuntu-24.04", arch: "arm64", tls: "quictls", time64: "-Time64Distro" },
           { config: "Release", os: "ubuntu-24.04", arch: "x64", tls: "quictls", time64: "-Time64Distro" },
+          { config: "Release", os: "ubuntu-26.04", arch: "arm", tls: "quictls", time64: "-Time64Distro" },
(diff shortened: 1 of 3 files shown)


scripts/validate-msquic-docker.ps1
```diff
@@ -68,6 +68,9 @@
 .EXAMPLE
     .\validate-msquic-docker.ps1 -Distro Ubuntu_24_04
 
+.EXAMPLE
+    .\validate-msquic-docker.ps1 -Distro Ubuntu_26_04
+
```
guhetier: Let's not have an example for every distro :)


```suggestion
```

gaurav2699 (approved):

.github/workflows/package-linux.yml
```diff
@@ -33,6 +33,9 @@ jobs:
           { config: "Release", os: "ubuntu-24.04", arch: "arm", tls: "quictls", time64: "-Time64Distro" },
           { config: "Release", os: "ubuntu-24.04", arch: "arm64", tls: "quictls", time64: "-Time64Distro" },
           { config: "Release", os: "ubuntu-24.04", arch: "x64", tls: "quictls", time64: "-Time64Distro" },
+          { config: "Release", os: "ubuntu-26.04", arch: "arm", tls: "quictls", time64: "-Time64Distro" },
```
guhetier, line 36: I didn't realize at first, but these are used to define the runner the packaging runs on, and ubuntu-26.04 is not available yet for runners.

guhetier: @Copilot Revert the changes in this file.

guhetier:
Superseeded by #5985

Issue #5982, opened 2026-05-05
[IMPLEMENTATION PLAN] Add support for XDP maps.


ProjectsByJackHe:
# XDP Map-Based Isolation for Multi-Process Worker Security

## Summary

This tracks the MsQuic changes needed to support XDP map-based packet isolation for multi-process deployments. Instead of each worker process configuring its own XDP rules (which allows a compromised worker to intercept traffic for other workers), a trusted orchestrator configures rules and provides pre-created XDP maps to workers. Workers can only register AF_XDP sockets into their assigned maps; they cannot read, write, or interfere with other workers' rules.

## Design

```
┌─────────────────┐         ┌──────────────────┐
│  Orchestrator    │         │   XDP Driver     │
│  (trusted)       │────────▶│   (kernel)       │
│                  │ creates │                  │
│  • Creates maps  │ rules + │  • Matches pkts  │
│  • Creates rules │  maps   │  • Redirects via │
│  • DuplicateHandle         │    map → AF_XDP  │
└────────┬─────────┘         └──────────────────┘
         │ map handle
         ▼
┌─────────────────┐
│  Worker Process  │
│  (less trusted)  │
│                  │
│  • Receives map  │
│  • Creates XSK   │
│    sockets       │
│  • Inserts into  │
│    map           │
│  • Cannot create │
│    rules         │
└──────────────────┘
```

**Resource model:** Each worker receives M maps (one per NIC interface), each map has Q entries (one per RSS queue). All dimensions are static and known at startup.

## PRs (topologically sorted)

| # | PR | Issue | Status |
|---|-----|-------|--------|
| 1 | API: Add `QUIC_XDP_MAP_CONFIG` | #5971 | Merged: #5983 |
| 2 | Datapath: Skip rule creation in map mode, insert sockets into map. End to end map tests. | #5972 | Merged #6003 |
| 3 | Documentation | #5973 | in-progress |

**Dependency graph:**
```
PR 1 (API) ──┬──▶ PR 2 (Datapath + Tests) 
             └──▶ PR 3 (Docs) ──────┘
```

## External dependencies

- **[xdp-for-windows#1009](https://github.com/microsoft/xdp-for-windows/issues/1009)** — New `XSKMAP` redirect action type + map create/delete API. Blocks the actual `XdpMapInsert`/`XdpMapRemove` calls in PR 2 and integration tests in PR 4.
[...]

PR #5983, opened 2026-05-05
[PLAN PR] [1/3] Add Global API to provide map configurations to the MsQuicLib prior to opening a registration.

### LINKED ISSUE #5971: Add an API to provide an XDP maps to MsQuic
An XDP map must be configured in MsQuic before datapath initialization.
The API should be experimental.

Multiple maps could be provided, for multiple interfaces.

Potential API surface:

```
typedef struct XDP_DATAPATH_CONFIG { 
     QUIC_XDP_MAP XdpSocketMap; // QUIC_XDP_MAP is defined per-platform, HANDLE on Windows UM. 
} 

typedef struct QUIC_REGISTRATION_CONFIG { 
     const char* AppName; 
     QUIC_EXECUTION_PROFILE ExecutionProfile; 
     XDP_DATAPATH_CONFIG XdpConfig; 
} QUIC_REGISTRATION_CONFIG; 
```

ProjectsByJackHe:
## Description
 PR [1/3] as part of the implementation plan #5982 

 Adds a SetParam/GetParam API for passing pre-created XDP map handles into MsQuic prior to opening a registration.

Fixes #5971.
 
 ## Testing
 
 Added 7 new test cases in `ApiTest.cpp` covering: null buffer, bad alignment, zero QueueCount, null handle, valid set/get roundtrip, clear, and buffer-too-small.
 
 ```powershell
 ./scripts/test.ps1 -Tls schannel -Config Debug -Arch x64 -Filter "*RegistrationParam*"
```

## Documentation

None in this PR. A docs/XdpMaps.md guide will follow with the datapath integration PR.

### CODE REVIEW

File: src/core/registration.h [hunk 1]
```diff
@@ -161,6 +161,14 @@ typedef struct QUIC_REGISTRATION {
     // Name of the application layer.
     //
     uint8_t AppNameLength;
+
+    //
+    // XDP map configurations (deep-copied from SetParam).
+    // One per interface, provided by external orchestrator.
+    //
+    QUIC_XDP_MAP_CONFIG* XdpMapConfigs;
```
mtfriesen:
```suggestion
    const QUIC_XDP_MAP_CONFIG* XdpMapConfigs;
```

src/test/lib/ApiTest.cpp
```diff
@@ -3071,15 +3071,15 @@ void QuicTestRegistrationParam()
...
+    // Unknown registration param should fail
     //
     {
         uint32_t Dummy = 0;
         TEST_QUIC_STATUS(
             QUIC_STATUS_INVALID_PARAMETER,
             MsQuic->SetParam(
                 Registration.Handle,
-                QUIC_PARAM_PREFIX_REGISTRATION,
+                QUIC_PARAM_PREFIX_REGISTRATION | 0xFF,
```
ProjectsByJackHe: This change is necessary because this test has a bug which assumed no registration parameters could exist, and used it to mark an invalid parameter. Now, we have added a registration parameter following the rest of the pattern in msquic.h. Instead of changing that value, let's fix the bug in the test.

src/inc/msquic.h
```diff
@@ -353,6 +353,23 @@ uint32_t
...
+// XDP Map configuration for multi-process isolation.
+// Passed via SetParam (QUIC_PARAM_REGISTRATION_XDP_MAP_CONFIG) after
+// RegistrationOpen but before creating any connections/listeners.
+//
+typedef void* QUIC_XDP_MAP_HANDLE;
+
+typedef struct QUIC_XDP_MAP_CONFIG {
+    uint32_t InterfaceIndex;        // Network interface this map applies to.
+    uint32_t QueueCount;            // Number of RSS queues for this interface.
+    QUIC_XDP_MAP_HANDLE MapHandle;  // XDP map handle (from orchestrator).
```
mtfriesen:
```suggestion
    QUIC_XDP_MAP_HANDLE MapHandle;  // XDP map handle.
```

src/inc/msquic.h
```diff
@@ -353,6 +353,23 @@ uint32_t
 
 #endif // QUIC_API_ENABLE_PREVIEW_FEATURES
 
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+
+//
+// XDP Map configuration for multi-process isolation.
+// Passed via SetParam (QUIC_PARAM_REGISTRATION_XDP_MAP_CONFIG) after
```
guhetier: I think this should not use one more `SetParam`, and instead be a part of the `QUIC_REGISTRATION_CONFIG`.
That's the direction we are going toward with RDMA, and it generally makes sense to have this kind of parameters that should be set before any connection is created set on init.

src/inc/msquic.h
```diff
@@ -353,6 +353,23 @@ uint32_t
 
 #endif // QUIC_API_ENABLE_PREVIEW_FEATURES
 
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+
+//
+// XDP Map configuration for multi-process isolation.
```
mtfriesen:
```suggestion
// XDP Maps configured for external XDP programs.
```

guhetier (changes requested):

src/core/registration.c
guhetier, line 519: Needs cleanup.

src/core/registration.c
```diff
@@ -513,6 +517,62 @@ QuicRegistrationParamSet(
         const void* Buffer
     )
 {
+    if (Param == QUIC_PARAM_REGISTRATION_XDP_MAP_CONFIG) {
+        if (BufferLength == 0) {
+            //
+            // Clear any existing XDP map config.
```
guhetier: I don't think we want / can support this at runtime.

src/core/registration.h
```diff
@@ -161,6 +161,14 @@ typedef struct QUIC_REGISTRATION {
     // Name of the application layer.
     //
     uint8_t AppNameLength;
+
+    //
+    // XDP map configurations (deep-copied from SetParam).
+    // One per interface, provided by external orchestrator.
```
mtfriesen: > One per interface

I don't see this validated - maybe just drop the comment?

mtfriesen: > provided by external orchestrator

Let's just say it's optional?

src/core/registration.c
```diff
@@ -513,6 +517,62 @@ QuicRegistrationParamSet(
         const void* Buffer
     )
 {
+    if (Param == QUIC_PARAM_REGISTRATION_XDP_MAP_CONFIG) {
```
guhetier: Consider using a subfunction. That's 50 lines for a single config, it'll age much better if you split dispatching and parameter handling.

src/core/registration.c
```diff
@@ -513,6 +517,62 @@ QuicRegistrationParamSet(
...
+        }
+
+        const uint32_t Count = BufferLength / sizeof(QUIC_XDP_MAP_CONFIG);
+        const QUIC_XDP_MAP_CONFIG* Configs = (const QUIC_XDP_MAP_CONFIG*)Buffer;
+
+        //
+        // Validate each config entry.
+        //
+        for (uint32_t i = 0; i < Count; i++) {
+#pragma warning(suppress: 6385) // BufferLength >= Count * sizeof check above guarantees bounds
```
guhetier: Do we really need this? Static analysis should understand it.

src/core/registration.c
```diff
@@ -513,6 +517,62 @@ QuicRegistrationParamSet(
...
+        }
+
+        //
+        // Deep-copy the new config.
+        //
+        Registration->XdpMapConfigs =
+            CXPLAT_ALLOC_NONPAGED(BufferLength, QUIC_POOL_REGISTRATION);
+        if (Registration->XdpMapConfigs == NULL) {
+            Registration->XdpMapConfigCount = 0;
+            return QUIC_STATUS_OUT_OF_MEMORY;
```
mtfriesen: I'm not sure if this error-handling is ideal, because we may have freed the previous config without successfully setting the new one.

If we can trivially make this API an atomic transaction, that seems best.

src/inc/msquic.h
```diff
@@ -353,6 +353,23 @@ uint32_t
...
+// Passed via SetParam (QUIC_PARAM_REGISTRATION_XDP_MAP_CONFIG) after
+// RegistrationOpen but before creating any connections/listeners.
+//
+typedef void* QUIC_XDP_MAP_HANDLE;
```
guhetier: Let's make this a handle on Windows. There are already some types in the API that are defined differently for Windows and Linux, we can follow the pattern

src/inc/msquic.h
mtfriesen: Yeah, this aligns with my question (not really answerable in this PR) of how this setter synchronizes with the datapath.

On the other hand, if the registration must be provided once, upfront, then how would the settings respond to classic networking events such as NICs being added or deleted?

guhetier: Hum, good point for NIC being added / deleted, if that is something we want/need to support.

I think we still should have a registration-init level parameter that says "use XDP with maps", even if the map ends up being provided differently. Eventually it will align better with other datapath selection option (RDMA, XDP without fallback, QMUX...)

src/test/lib/ApiTest.cpp
```diff
@@ -3091,12 +3091,152 @@ void QuicTestRegistrationParam()
...
                 &Length,
                 &Buffer));
         TEST_EQUAL(Length, 65535);
         TEST_EQUAL(Buffer, 65535);
     }
+
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
```
guhetier: In general, let's try not to introduce more `QUIC_API_ENABLE_PREVIEW_FEATURES` in tests.

The only reason they are here in the first place is pretty weak (it is because if the preview feature is modified in the future, and we run this as a down-level test against the future MsQuic, the test would fail).
But a better solution for this problem is to keep test focused on a feature, and skip tests that fail because of a breaking change on a preview feature.

src/inc/msquic.h
mtfriesen: Agree declaring intent/dependencies upfront is useful.

src/core/registration.c
ProjectsByJackHe: the compiler does path-based analysis. I can only cleanup `Param`

src/inc/msquic.h
ProjectsByJackHe: Surfacing this offline discussion:

There is a fundamental ABI mismatch problem here. 
 
```
The problem:

 // Non-preview caller sees:
 struct QUIC_REGISTRATION_CONFIG {  // 12 bytes (x64: 8 ptr + 4 enum)
     const char* AppName;
     QUIC_EXECUTION_PROFILE ExecutionProfile;
 };
 
 // Library (always has preview enabled) sees:
 struct QUIC_REGISTRATION_CONFIG {  // 16 bytes (8 ptr + 4 enum + 4 flag)
     const char* AppName;
     QUIC_EXECUTION_PROFILE ExecutionProfile;
     QUIC_REGISTRATION_USE_XDP_MAP Flag;  // <-- extra 'Flag' field for XDP maps.
 };

// A non-preview caller does:
 QUIC_REGISTRATION_CONFIG config = { "myapp", QUIC_EXECUTION_PROFILE_LOW_LATENCY };
 MsQuic->RegistrationOpen(&config, &handle);
// The library receives a pointer to a 12-byte struct on the caller's stack. 
// If it reads Config->Flags, it's reading 4 bytes past the end of what the caller allocated
```

The main thing is the QUIC_REGISTRATION_CONFIG isn't versioned like how QUIC_SETTINGS is with the IsSet union.
 
We can add an IsSet field to version it... but it will change the struct layout. This is a hard ABI break. Every consumer of MsQuic would need to recompile. 
 
How do we want to handle this?

src/core/registration.c
ProjectsByJackHe: seems like we need this (static analysis on my machine does not get it)

src/core/registration.c
ProjectsByJackHe: swapped the order. alloc first, then if success update.

src/inc/msquic.h
mtfriesen: Well, if we don't have a concrete need right now to know this at registration time (and I don't see why we _need_ it), then I'm OK with leaving the lack of versioned struct for later improvements.

ProjectsByJackHe: Resolved offline.

src/test/lib/ApiTest.cpp
ProjectsByJackHe: I'd prefer to tackle that as a separate cleanup PR to avoid broadening the scope here.

src/core/library.c
```diff
@@ -1430,6 +1436,60 @@ QuicLibrarySetGlobalParam(
...
+            }
+        }
+
+        CxPlatLockAcquire(&MsQuicLib.Lock);
+
+        //
+        // Only allowed once, and only before any registration is opened.
+        //
+        if (MsQuicLib.XdpMapConfigs != NULL ||
+            !CxPlatListIsEmpty(&MsQuicLib.Registrations)) {
```
mtfriesen: Does this also protect the case where the final registration is closed? Does the stateless registration show up here?

src/inc/msquic.h
```diff
@@ -353,6 +353,26 @@ uint32_t
...
+
+//
+// XDP Maps configured for external XDP programs.
+// Passed via SetParam (QUIC_PARAM_GLOBAL_XDP_MAP_CONFIG) after
+// MsQuicOpenVersion but before opening any registrations.
+//
+#ifdef _WIN32
+typedef HANDLE QUIC_XDP_MAP_HANDLE;
+#else
+typedef void* QUIC_XDP_MAP_HANDLE;
```
mtfriesen: probably `typedef int QUIC_XDP_MAP_HANDLE` should be the cross-platform default/strawman.

.github/copilot-instructions.md
```diff
@@ -122,6 +122,7 @@ The `check-dotnet.yml` CI workflow validates the generated .NET interop files ar
 - Platform layer abstracts: TLS (`tls_schannel.c`, `tls_openssl.c`, `tls_quictls.c`), UDP sockets/datapath (`datapath_win.c`, `datapath_epoll.c`, `datapath_kqueue.c`), crypto (`crypt_bcrypt.c`, `crypt_openssl.c`).
 - Public API is in `src/inc/msquic.h`. C++ wrapper in `src/inc/msquic.hpp`.
 - Test infrastructure uses GoogleTest. Test cases in `src/test/lib/`, runner in `src/test/bin/`.
+- **`QUIC_API_ENABLE_PREVIEW_FEATURES`**: Always defined to `1` in `src/core/precomp.h`. Do **not** use `#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES` guards in `src/core/` or `src/platform/` code — the guard is only needed in `src/inc/msquic.h` to preserve ABI compatibility for consumers who don't opt into preview features.
```
guhetier: This isn't an architecture note.

> to preserve ABI compatibility

This has nothing to do with compatibility (ABI or other), it only exists to identify the part of the API that are preview and make it easy to not use them by mistake.

src/core/library.c
```diff
@@ -1430,6 +1436,60 @@ QuicLibrarySetGlobalParam(
         break;
     }
 
+    case QUIC_PARAM_GLOBAL_XDP_MAP_CONFIG: {
+        if (BufferLength == 0 || Buffer == NULL) {
+            return QUIC_STATUS_INVALID_PARAMETER;
```
guhetier: While generally an early return is what I would recommend, this function set a status and break. Please make it uniform (either break, or return everywhere)

src/core/library.c
```diff
@@ -1430,6 +1436,60 @@ QuicLibrarySetGlobalParam(
...
+        }
+
+        const uint32_t Count = BufferLength / sizeof(QUIC_XDP_MAP_CONFIG);
+        const QUIC_XDP_MAP_CONFIG* Configs = (const QUIC_XDP_MAP_CONFIG*)Buffer;
+
+        //
+        // Validate each config entry.
+        //
+        for (uint32_t i = 0; i < Count; i++) {
+#pragma warning(suppress: 6385) // False positive: Count derived from BufferLength / sizeof guarantees bounds.
```
guhetier: Is this really needed? I am surprised static analysis doesn't handle such a simple scenario.

src/core/library.c
```diff
@@ -1430,6 +1436,60 @@ QuicLibrarySetGlobalParam(
...
+                BufferLength);
+            return QUIC_STATUS_OUT_OF_MEMORY;
+        }
+
+        CxPlatCopyMemory(NewConfigs, Configs, BufferLength);
+        MsQuicLib.XdpMapConfigs = NewConfigs;
+        MsQuicLib.XdpMapConfigCount = Count;
+        CxPlatLockRelease(&MsQuicLib.Lock);
+
+        Status = QUIC_STATUS_SUCCESS;
```
guhetier: not needed, `Status` is initialized to sucess.

src/inc/msquic.h
guhetier: nit: Consider moving it to the per-platform files, like `QUIC_ADDR`.

src/test/lib/ApiTest.cpp
```diff
@@ -2983,6 +2983,154 @@ void QuicTestGlobalParam()
         TEST_TRUE(Length >= sizeof(Expected));
     }
 
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
```
guhetier, line 2989: Same as in previous iteration, lets avoid introducing more `QUIC_API_ENABLE_PREVIEW_FEATURES` in test code.
Please make a new test case instead, that we could skip in downlevel tests if we break the API.
We should not be shy creating test cases, granularity facilitates diagnostics when tests fail.

src/test/lib/ApiTest.cpp
```diff
@@ -2983,6 +2983,154 @@ void QuicTestGlobalParam()
...
+                    sizeof(QUIC_XDP_MAP_CONFIG),
+                    nullptr));
+        }
+
+        //
+        // Set with buffer length not a multiple of struct size should fail.
+        //
+        {
+            TestScopeLogger LogScope1("SetParam bad length");
+            QUIC_XDP_MAP_CONFIG Config = { 1, (QUIC_XDP_MAP_HANDLE)(uintptr_t)0x1234 };
```
guhetier: nit: For clarity and conciseness, define one or two QUIC_XDP_MAP_HANDLE at the start of your test as constants we can use.

src/test/lib/ApiTest.cpp
```diff
@@ -2983,6 +2983,154 @@ void QuicTestGlobalParam()
...
+                QUIC_STATUS_INVALID_PARAMETER,
+                MsQuic->SetParam(
+                    nullptr,
+                    QUIC_PARAM_GLOBAL_XDP_MAP_CONFIG,
+                    0,
+                    nullptr));
+        }
+
+        //
+        // Set with valid config should succeed.
```
guhetier: This modifies the state, it should be made very clear since it impacts every test that follows.

src/test/lib/ApiTest.cpp
```diff
@@ -2983,6 +2983,154 @@ void QuicTestGlobalParam()
...
+                    QUIC_PARAM_GLOBAL_XDP_MAP_CONFIG,
+                    &OutLength,
+                    OutConfigs));
+            TEST_EQUAL(OutLength, (uint32_t)sizeof(OutConfigs));
+            TEST_EQUAL(OutConfigs[0].InterfaceIndex, 3u);
+            TEST_EQUAL(OutConfigs[1].InterfaceIndex, 7u);
+        }
+
+        //
+        // Get with too-small buffer should return BUFFER_TOO_SMALL.
```
guhetier, line 3118: It would be good to validate what happens when the provided buffer is bigger than needed.

src/inc/msquic.h
```diff
@@ -973,6 +993,9 @@ void
 #define QUIC_PARAM_GLOBAL_STATELESS_RESET_KEY           0x0100000B  // uint8_t[] - Array size is QUIC_STATELESS_RESET_KEY_LENGTH
 #define QUIC_PARAM_GLOBAL_STATISTICS_V2_SIZES           0x0100000C  // uint32_t[] - Array of sizes for each QUIC_STATISTICS_V2 version. Get-only. Pass a buffer of uint32_t, output count is variable. See documentation for details.
 #define QUIC_PARAM_GLOBAL_STATELESS_RETRY_CONFIG        0x0100000D  // QUIC_STATELESS_RETRY_CONFIG
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+#define QUIC_PARAM_GLOBAL_XDP_MAP_CONFIG               0x0100000E  // QUIC_XDP_MAP_CONFIG[]
```
guhetier: nit: please align the value

src/core/library.c
```diff
@@ -759,6 +759,12 @@ MsQuicLibraryUninitialize(
         MsQuicLib.ExecutionConfig = NULL;
     }
 
+    if (MsQuicLib.XdpMapConfigs != NULL) {
+        CXPLAT_FREE((void*)MsQuicLib.XdpMapConfigs, QUIC_POOL_XDP_MAP_CONFIG);
```
guhetier: no need for a cast

src/core/library.c
```diff
@@ -1785,6 +1845,29 @@ QuicLibraryGetGlobalParam(
...
+            *BufferLength = 0;
+            Status = QUIC_STATUS_SUCCESS;
+            break;
+        }
+        if (*BufferLength < RequiredLength) {
+            *BufferLength = RequiredLength;
+            Status = QUIC_STATUS_BUFFER_TOO_SMALL;
+            break;
+        }
+        if (Buffer == NULL) {
```
guhetier: Consider treating this as a buffer too small case, this is what is done above and it is a pretty common pattern to call an API with a null buffer to query the size needed.

src/core/library.c
```diff
@@ -1785,6 +1845,29 @@ QuicLibraryGetGlobalParam(
 #endif // DEBUG
     }
 
+    case QUIC_PARAM_GLOBAL_XDP_MAP_CONFIG: {
+        const uint32_t RequiredLength =
+            MsQuicLib.XdpMapConfigCount * (uint32_t)sizeof(QUIC_XDP_MAP_CONFIG);
```
guhetier, line 1859: The reads should be protected by the lock too.

src/test/lib/ApiTest.cpp
```diff
@@ -2983,6 +2983,154 @@ void QuicTestGlobalParam()
...
+            QUIC_XDP_MAP_CONFIG Config = { 1, (QUIC_XDP_MAP_HANDLE)(uintptr_t)0x1234 };
+            TEST_QUIC_STATUS(
+                QUIC_STATUS_INVALID_STATE,
+                MsQuic->SetParam(
+                    nullptr,
+                    QUIC_PARAM_GLOBAL_XDP_MAP_CONFIG,
+                    sizeof(Config),
+                    &Config));
+        }
+    }
```
guhetier, line 3210: Should we add a test case to validate it can't be set after a registration is created?

src/core/library.c
ProjectsByJackHe: We need it. XdpMapConfigs is defined as a `const*`

src/core/library.c
ProjectsByJackHe: The pragma is needed. MSVC static analysis fires C6385 without it.

src/inc/msquic.h
ProjectsByJackHe: moved.

src/core/library.c
ProjectsByJackHe: Ahh, there is already a .LazyInitComplete flag. We should use that instead.

src/inc/msquic.h
ProjectsByJackHe: moved these to per-platform header files.

src/core/library.c
ProjectsByJackHe: good callout, added.

.github/copilot-instructions.md
ProjectsByJackHe: corrected the context, but I think having something here to tell copilot not to use the ifdef guards in core code is necessary.

src/test/lib/ApiTest.cpp
ProjectsByJackHe: Added.

src/test/lib/ApiTest.cpp
ProjectsByJackHe: Added N.B. comment.

src/core/library.c
ProjectsByJackHe: Uniformed

src/test/lib/ApiTest.cpp
ProjectsByJackHe: Added.

src/test/lib/ApiTest.cpp
ProjectsByJackHe: I get that during runtime we can skip these, but compile time needs these ifdef guards for downlevel: https://github.com/microsoft/msquic/actions/runs/25773121453/job/75700191362?pr=5983#logs

src/core/library.c
guhetier: `CXPLAT_FREE` already has a cast to deal with it this for the special situation of freeing memory. 
We don't want to start casting away `const`. Special case like freeing memory are ok, but the helpers take care of it. In general, casting const away is one of the biggest red flags.

src/core/library.c
guhetier: That's sad :(

src/test/lib/ApiTest.cpp
guhetier: Oh, right, because we have that useless logic that defines `QUIC_API_ENABLE_PREVIEW_FEATURES` conditionally for tests.
We need to clean that up. At least, making independent tests will make it easier.

src/core/library.c
```diff
@@ -1430,6 +1436,68 @@ QuicLibrarySetGlobalParam(
...
+        const uint32_t Count = BufferLength / sizeof(QUIC_XDP_MAP_CONFIG);
+        const QUIC_XDP_MAP_CONFIG* Configs = (const QUIC_XDP_MAP_CONFIG*)Buffer;
+
+        //
+        // Validate each config entry.
+        //
+        BOOLEAN ValidConfigs = TRUE;
+        for (uint32_t i = 0; i < Count; i++) {
+#pragma warning(suppress: 6385) // False positive: Count derived from BufferLength / sizeof guarantees bounds.
+            if (Configs[i].MapHandle == (QUIC_XDP_MAP_HANDLE)0) {
```
mtfriesen: IDK if MsQuic should have an opinion on what a valid map handle is? These are ultimately assigned by whatever the underlying map provider is, and 0 is a valid FD on Linux.

mtfriesen: And if we sidestep this pseudo-validation, we also eliminate the need for static analysis suppression.

src/core/library.c
mtfriesen: Weirdly, the `QUIC_PARAM_GLOBAL_EXECUTION_CONFIG` case above seems roughly similar in loop behavior, but isn't caught? If we just juggle the logic around a bit, can we satisfy the analyzer?

src/core/library.c
```diff
@@ -1430,6 +1436,68 @@ QuicLibrarySetGlobalParam(
...
+            Status = QUIC_STATUS_INVALID_PARAMETER;
+            break;
+        }
+
+        CxPlatLockAcquire(&MsQuicLib.Lock);
+
+        //
+        // Only allowed once, and only before the datapath is initialized.
+        //
+        if (MsQuicLib.XdpMapConfigs != NULL ||
```
ProjectsByJackHe: QQ: right now, this call can succeed at most once.
But do we want to allow apps to call this however many times they want before they open any registrations?

I vote no, agree?

src/core/library.c
ProjectsByJackHe: Agreed, lets remove this validation.

src/core/library.c
ProjectsByJackHe: got it, removed.

src/core/library.c
guhetier: What is the precedent for other similar parameters, like the execution config? Let's just do the same.
Being consistent and predictable is what matters.

ProjectsByJackHe: Got it. In that case, let's make it update-able prior to initializing the datapath.

src/test/bin/quic_gtest.cpp
```diff
@@ -289,6 +289,22 @@ TEST(ParameterValidation, ValidateGlobalParam) {
...
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+TEST(ParameterValidation, ValidateXdpMapConfigParam) {
+    //
+    // User-mode only: QUIC_PARAM_GLOBAL_XDP_MAP_CONFIG is set-once before the
+    // library's lazy initialization. In kernel mode the test driver shares one
+    // MsQuicLib across all tests in the run, so earlier tests have already
+    // triggered lazy init and this test cannot exercise the success paths.
+    //
+    if (TestingKernelMode) {
+        GTEST_SKIP() << "QuicTestXdpMapConfigParam is user-mode only.";
```
guhetier, line 302: Is `QuicTestXdpMapConfigParam` user-mode only, or is this test not working in KM because of what the comment explains?
We can skip the test in KM if the test harness can't support it, but let's be clear about the reason why: if it is only the test harness, we could fix it in the future.

ProjectsByJackHe: In KM, the driver stays attached. Whereas in UM, state is refreshed per test case. We can fix it in the test harness by refreshing the driver for each case.

ProjectsByJackHe: Added a further comment clarifying the reason.

.github/copilot-instructions.md
guhetier: I'm happy to have it in the instructions, but there is probably a better section than "Architecture".

ProjectsByJackHe: added a new 'coding conventions' section here.

guhetier (dismissed):

src/inc/msquic_posix.h
```diff
@@ -492,6 +492,8 @@ QuicAddrFromString(
 //
 // Represents an IP address and (optionally) port number as a string.
 //
+typedef int QUIC_XDP_MAP_HANDLE;
```
mtfriesen: The comment above this looks mismatched.

src/test/bin/quic_gtest.cpp
```diff
@@ -289,6 +289,23 @@ TEST(ParameterValidation, ValidateGlobalParam) {
...
 }
 
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+TEST(ParameterValidation, ValidateXdpMapConfigParam) {
+    //
+    // User-mode only: QUIC_PARAM_GLOBAL_XDP_MAP_CONFIG is set-once before the
+    // library's lazy initialization. In kernel mode the test driver shares one
+    // MsQuicLib across all tests in the run, so earlier tests have already
+    // triggered lazy init and this test cannot exercise the success paths.
+    // This is a current test harness limitation today.
```
mtfriesen: Super nit: "current" and "today" are redundant. I do this all the time.

src/test/lib/ApiTest.cpp
```diff
@@ -2986,6 +2986,194 @@ void QuicTestGlobalParam()
...
+                QUIC_STATUS_INVALID_PARAMETER,
+                MsQuic->SetParam(
+                    nullptr,
+                    QUIC_PARAM_GLOBAL_XDP_MAP_CONFIG,
+                    sizeof(QUIC_XDP_MAP_CONFIG) - 1,
+                    &Config));
+        }
+
+        //
+        // Set with zero-length should fail.
```
mtfriesen: I'm not sure why we allow this to be re-set (mutated) but don't allow it to be cleared? If this is just for consistency with other APIs, then I'm OK with it, but I'd rather it be set-once or allow the setting to be cleared, if never being set in the first place is also a legal state.

src/test/lib/ApiTest.cpp
```diff
@@ -2986,6 +2986,194 @@ void QuicTestGlobalParam()
...
+
+        //
+        // N.B. The following SetParam call modifies global library state.
+        // XDP map config can be updated multiple times before the first
+        // registration, so subsequent tests see this configured state.
+        //
+        {
+            TestScopeLogger LogScope1("SetParam valid");
+            QUIC_XDP_MAP_CONFIG Configs[2] = {
+                { 3, FakeHandle1 },
```
mtfriesen: Why did we define `FakeHandle1` but not `FakeIfIndex1`? I see 3 and 7 reused below.

src/test/lib/ApiTest.cpp
```diff
@@ -2986,6 +2986,194 @@ void QuicTestGlobalParam()
...
+                    nullptr,
+                    QUIC_PARAM_GLOBAL_XDP_MAP_CONFIG,
+                    &OutLength,
+                    &OutConfig));
+            TEST_EQUAL(OutLength, (uint32_t)sizeof(QUIC_XDP_MAP_CONFIG));
+            TEST_EQUAL(OutConfig.InterfaceIndex, 1u);
+        }
+
+        //
+        // Setting after a registration is created should fail.
```
mtfriesen, line 3179: Should test setting after a registration is closed should fail.

mtfriesen (changes requested):
A few nits but a few blockers.

src/test/lib/ApiTest.cpp
ProjectsByJackHe: Good callout, bufferlength == 0 is the signal for it to be cleared. Fixed.

src/test/lib/ApiTest.cpp
ProjectsByJackHe: Added FakeIndex

src/test/lib/ApiTest.cpp
ProjectsByJackHe: Good callout, added.

mtfriesen (approved):

PR #5995, opened 2026-05-13
 Fix race in TestStream::StartPing causing Reject0Rtt/v6/20000 flakes

### LINKED ISSUE #5894: [CI - FAILURE]:  / AppData_WithSend0RttArgs2.Reject0Rtt_7 (msquictest.exe)
https://github.com/microsoft/msquic/actions/runs/23551834552

gaurav2699:
## Description

  `AppData/WithSend0RttArgs2.Reject0Rtt/7` (`v6 / Length=20000`) intermittently fails with `MsQuic->StreamSend failed, 0x80070057` at `TestStream.cpp:165`. The failure is a race between `TestStream::StartPing` (main test thread) and `TestStream::HandleStreamSendComplete` (connection worker thread); both mutate `BytesToSend` non-atomically and can each decide they're sending the final chunk.

Make the "read remaining bytes + pick chunk size + commit subtraction" a single atomic claim via `InterlockedCompareExchange64` in both call sites. `Desired == 0` after the successful CAS becomes the unambiguous "I got the last chunk" signal used for FIN (no separate post-subtract check).

 If a thread loses the CAS, it loops; if it sees `Current <= 0` on a fresh read it exits cleanly. `BytesToSend` can no longer go negative, the 4 GB allocation path is unreachable, and the FIN flag is applied by exactly one sender.

Fixes: #5894 
## Testing

Tests Succeed

## Documentation

No


src/test/lib/TestStream.cpp
```diff
@@ -165,14 +179,14 @@ TestStream::StartPing(
             TEST_FAILURE("MsQuic->StreamSend failed, 0x%x.", Status);
             return false;
         }
-        if (resultingBytesLeft == 0) {
+        if (Desired == 0) {
             // On the finish packet if it succeeds, the instance
             // we are executing in will be deleted. Return
```
guhetier: .... wow. The person who wrote that code liked to live on the edge.

guhetier (commented):
I think we can simplify this.

If I read the help correctly, it is supposed to send a payload of N bytes, with a maximum of K send request in flight. If more than K requests are needed, the completion handler will queue more send requests.

We can just pre-compute how many requests this needs to send and how much data that will be. The member only ever needs to contain what the handler needs.

I'd also recommend to use a lock instead of atomic operation, if there are still any concurrency concerns. This is test code, simplicity and readabilty is much, much more important than an hypothetical performance gain.

src/test/lib/TestStream.cpp
```diff
@@ -127,15 +127,55 @@ TestStream::StartPing(
...
+    // Pre-compute the priming schedule (number of sends and the final
+    // chunk's size) so each iteration knows up front whether it is the
+    // last one. With the schedule fixed, the do/while loop is bounded by
+    // TotalSends and the race against HandleStreamSendComplete on the
+    // worker thread cannot drive iter counts past it. The remaining race
+    // is just on the InterlockedSubtract result: if it goes negative we
+    // know the worker beat us to the tail, so we roll back and stop.
+    //
+    long TotalSends;
+    uint32_t FinalSendBufferLength;
```
guhetier: Please always init variables in C++ code.

```suggestion
    long TotalSends{};
    uint32_t FinalSendBufferLength{};
```

src/test/lib/TestStream.cpp
```diff
@@ -127,15 +127,55 @@ TestStream::StartPing(
 {
     BytesToSend = PayloadLength / MaxSendBuffers;
```
guhetier: To be honest, I'm not sure I get this.
Is `BytesToSend` poorly named and actually not the number of bytes left to be sent?

src/test/lib/TestStream.cpp
```diff
@@ -127,15 +127,55 @@ TestStream::StartPing(
...
+        const int64_t resultingBytesLeft =
+            InterlockedSubtract64(&BytesToSend, SendBufferLength);
+        if (resultingBytesLeft < 0) {
+            //
+            // HandleStreamSendComplete on the worker thread already
+            // claimed the tail. Put the bytes back so BytesToSend can't
+            // stay negative (which would otherwise drive a bogus
+            // SendBufferLength on the next chained send) and stop here.
+            //
+            InterlockedExchangeAdd64(&BytesToSend, (int64_t)SendBufferLength);
```
guhetier: The goal is to not need to touch variables used by the callback more than once to init them, and never after we run the first send operation.
That way, they are not used concurrently anymore and life is simple.

This function should:
1) Determine how much work can be done in this function
2) Determine how much work needs to be done from the callback (if any)
3) Initialize some shared state so the callback knows what it needs to do
4) This function does all of its share of the work

 `OutstandingSendRequestCount` is fine to share, it is simply a counter.

Don't be afraid to refactor this function as much as needed.

src/test/lib/TestStream.cpp
gaurav2699: You're right, BytesToSend is per-buffer bytes remaining, not total. The real semantics is bytes still needed on each of the 2 buffers on the next StreamSend.

src/test/lib/TestStream.cpp
```diff
@@ -209,7 +213,7 @@ TestStream::StartPing(
             TEST_FAILURE("MsQuic->StreamSend failed, 0x%x.", Status);
             return false;
         }
-        if (resultingBytesLeft == 0) {
+        if (IsLast) {
```
guhetier: Is this still needed? If we are on the last packet, we are going to exit the loop anyway.

src/test/lib/TestStream.cpp
```diff
@@ -125,67 +125,71 @@ TestStream::StartPing(
...
-    // "stream forever" payload lengths (ServerDisconnect,
-    // StatelessResetKey) don't truncate to a 32-bit value.
+    // Pre-compute the full schedule so each priming iteration knows up
+    // front whether it is the final send (only the last one shrinks to
+    // LastSendBufferLength and carries FIN). 
     //
-    int64_t TotalSends;
-    uint32_t FinalSendBufferLength;
-    if (BytesToSend == 0) {
+    const uint64_t BytesPerBuffer = PayloadLength / MaxSendBuffers;
```
guhetier: Oh, I think I finally get what the variable represents, which doesn't make a lot of sense.
The number of bytes we need to send is divided by the number of buffers we use on every send call?

This variable and `BytesToSend` are pretty confusing still (already from before this PR), because they don't represent a clear concept (or at least, one that doesn't click in my head...)

It is pretty similar, but I think it would be clearer if:
`MaxBytesPerSendOperation = MaxSendBuffers * MaxSendLength`
then:
```
TotalSends = PayloadLength / MaxBytesPerSendOperation;
LastSend = PayloadLength % MaxBytesPerSendOperation;
if (LastSend != 0) {
    // Or we could just deal with the last send outside of the loop
    TotalSends++;
}
```

And then, each iteration of the loop send MaxSendBuffers of MaxSendLength, except for the last iteration that sends MaxSendBuffers of LastSend / MaxSendBuffers bytes.
Or we could have the last send just send a single QUIC_BUFFER, honnestly, and keep it simple.

The callback would have to be adapted in the same way so that `BytesToSend` is actually the number of bytes to send.

src/test/lib/TestStream.cpp
```diff
@@ -340,12 +337,22 @@ TestStream::HandleStreamSendComplete(
...
-            auto SendBufferLength = (uint32_t)CXPLAT_MIN(BytesToSend, (int64_t)MaxSendLength);
+            const int64_t BytesThisSend =
+                CXPLAT_MIN(BytesToSend, (int64_t)MaxBytesPerSend);
+            const uint32_t SendBufferLength =
+                (uint32_t)(BytesThisSend / SendBuffer->BufferCount);
             for (uint32_t i = 0; i < SendBuffer->BufferCount; ++i) {
                 SendBuffer->Buffers[i].Length = SendBufferLength;
             }
-            if (InterlockedSubtract64(&BytesToSend, SendBufferLength) == 0) {
+            if (InterlockedSubtract64(&BytesToSend, BytesThisSend) == 0) {
```
guhetier: nit: Doesn't need to be atomic anymore, we now ensure it is never accessed concurrently.

src/test/lib/TestStream.cpp
```diff
@@ -127,51 +127,56 @@ TestStream::StartPing(
...
+    // the callback could ever decrement that far anyway.
     //
     if (PrimingCount == TotalSends) {
         BytesToSend = 0;
     } else {
+        const uint64_t Remaining =
+            PayloadLength - (uint64_t)PrimingCount * MaxBytesPerSend;
         BytesToSend =
-            (int64_t)BytesPerBuffer - PrimingCount * (int64_t)MaxSendLength;
+            (Remaining > (uint64_t)INT64_MAX) ? INT64_MAX : (int64_t)Remaining;
```
guhetier: Simpler as `CXPLAT_MIN(INT64_MAX, Remaining);`?

guhetier (dismissed):
Looks good to me, a last few nitpicks.

There are a number of potential rounding errors that mean we could send slightly less bytes than requested, but I don't think it matters or happens with the values used for tests.

guhetier (approved):

PR #5996, opened 2026-05-14
create/update CI issues for Stress and RecvFuzz failures

### LINKED ISSUE #5993: Automatically create an issue when the stress test fails
When the BVT test fail, an issue is automatically created, pointing at the failing pipeline

(see `Create or Update Issue` in test.yml)

We should do the same for the stress test (and maybe other test workflows).
Ideally, this could be a flexible, re-usable step we can easily add to multiple pipelines.

> Thanks. We should add the same automation we have for BVT tests to automatically file bugs on stress test failures. 

 _Originally posted by @guhetier in [#5990](https://github.com/microsoft/msquic/issues/5990#issuecomment-4422689965)_

gaurav2699:
## Description

Stress and RecvFuzz failures were not being captured by the automatic issue creation path, which could delay visibility and triage for non-BVT regressions.  This PR ensures that we do. Fixes #5993 
## Testing
NA

## Documentation

NA


.github/workflows/test.yml
```diff
@@ -333,6 +333,22 @@ jobs:
         path: artifacts
         merge-multiple: false
 
+    - name: Download Stress Failures
+      uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c
+      continue-on-error: true
```
guhetier: Should continue on error also be added to `Download BVT Failures`?

.github/workflows/test.yml
guhetier: Did you confirm this function will find the stress failure? It looks specific to tests (with reference to `testcase`).

A good way to test is to have a dummy commit to:
- force a failure in the stress / fuzz test
- relax the condition `if: ${{ failure() && github.event_name == 'push' }}` so it runs on your PR

And you can check the issue is created like you expect with enough details to investigate it

.github/workflows/test.yml
gaurav2699: Yes

.github/workflows/test.yml
gaurav2699: testcase is a wrong variable name. The artifact logs for stress, and the other test logs are generic and should behave that way. Made the change

src/tools/spin/spinquic.cpp
guhetier, line 1: Is this a merge / rebase issue?
If not, please keep fixes to spinquic in a different PR.

.github/workflows/stress.yml
```diff
@@ -220,3 +222,142 @@ jobs:
       uses: re-actors/alls-green@05ac9388f0aebcb5727afa17fcccfecd6f8ec5fe
       with:
         jobs: ${{ toJSON(needs) }}
+
+  Create-Or-Update-Issue:
```
guhetier, line 226: question: is there a way we could factor this in a reusable workflow or a common script we only invoke here?
There is a lot of duplicated logic, and we would probably add similar bug-filing logic to more pipelines in the future, if that can be achieved easily enough.

.github/workflows/test.yml
guhetier: Now that the script is separated, does it still make sense to continue on failure if the download failed? There is nothing to inspect.

src/tools/spin/spinquic.cpp
gaurav2699: Nono was just testing if our issue is created since our spinquic changes was giving consistent failures. Will be reverted

.github/workflows/file-issue-on-test-failure.yml
guhetier, line 1: For the sake of clarity, lets rename the file "file-issue-on-test-failure" or something a bit more explicit.

.github/workflows/test.yml
```diff
@@ -323,130 +323,15 @@ jobs:
...
-              const created = await github.rest.issues.create({
-                owner, repo,
-                title: it.title,
-                body: it.description,
-                labels: ['bvt', 'ci-failure'] // tweak/remove labels as you wish
-              });
-              core.notice(`Created issue #${created.data.number}: ${it.title}`);
-            }
-          }
+    if: ${{ failure() && github.event_name != 'pull_request' }}
```
guhetier: Is the change from `== push` to `!= pull_request` intentional?

guhetier (dismissed):

.github/workflows/test.yml
gaurav2699: No, VS code auto suggestion did that. Will change to original for consistency sake.

guhetier (approved):

PR #5997, opened 2026-05-14
Fix SNI for IP literal server names

### LINKED ISSUE #3493: SNI is incorrectly set when dialing an IP address
### Describe the bug

When dialing a Quinn server (Rust based) I get a handshake failure on the rust side because it fails to parse the SNI field. The SNI field is set to the ip address of the server, which I don't believe is a valid domain name.

I think we're always setting the `RemoteServerName` field https://github.com/microsoft/msquic/blob/main/src/core/connection.c#L1980. Maybe we shouldn't in the case that it is an ip address?

### Affected OS

- [ ] Windows
- [X] Linux
- [ ] macOS
- [ ] Other (specify below)

### Additional OS information

_No response_

### MsQuic version

main

1. Have an MsQuic client dial a Quinn server.
2. Fail to handshake.

### Expected behavior

Handshake should succeed.

### Actual outcome

fails.

### Additional details

Not an experienced C programmer, but happy to attempt to tackle this with some pointers.

flcl42:
Fixes #3493.

Makes OpenSSL-backed TLS paths avoid sending SNI when the configured server name is an IP literal. Hostname SNI behavior is unchanged.

  Changes:
  - Detect IPv4 and IPv6 literal server names during TLS initialization.
  - Skip copying IP literals into the TLS SNI field for OpenSSL and quicTLS.

## Testing
  -  Added IPv4/IPv6 regression coverage for connecting with an IP literal server name while using an explicit remote address.
  - Confirmed it fixes "Illegal SNI hostname received" when connecting to a Rust quinn/rustls server

## Documentation

-

### DIFF
--- a/src/platform/tls_openssl.c
+++ b/src/platform/tls_openssl.c
@@ -150,6 +150,35 @@ typedef struct CXPLAT_TLS {
 
 } CXPLAT_TLS;
 
+static
+BOOLEAN
+CxPlatTlsServerNameIsIpLiteral(
+    _In_z_ const char* ServerName,
+    _In_ uint16_t ServerNameLength
+    )
+{
+    QUIC_ADDR Addr = {0};
+    if (QuicAddrFromString(ServerName, 0, &Addr)) {
+        return TRUE;
+    }
+
+    if (ServerNameLength >= 2 &&
+        ServerName[0] == '[' &&
+        ServerName[ServerNameLength - 1] == ']') {
+        char UnbracketedName[64];
+        size_t UnbracketedNameLength = ServerNameLength - 2;
+        if (UnbracketedNameLength >= sizeof(UnbracketedName)) {
+            return FALSE;
+        }
+
+        memcpy(UnbracketedName, ServerName + 1, UnbracketedNameLength);
+        UnbracketedName[UnbracketedNameLength] = '\0';
+        return QuicAddrFromString(UnbracketedName, 0, &Addr);
+    }
+
+    return FALSE;
+}
+
 //
 // @struct RECORD_ENTRY
 // @brief Represents a buffered SSL record in a linked list.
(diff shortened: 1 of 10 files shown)


flcl42:

src/platform/tls_openssl.c
```diff
@@ -150,6 +150,35 @@ typedef struct CXPLAT_TLS {
...
+{
+    QUIC_ADDR Addr = {0};
+    if (QuicAddrFromString(ServerName, 0, &Addr)) {
+        return TRUE;
+    }
+
+    if (ServerNameLength >= 2 &&
+        ServerName[0] == '[' &&
+        ServerName[ServerNameLength - 1] == ']') {
+        char UnbracketedName[64];
```
guhetier: If there is a reason to accept bracketed ip addresses, we should do it in `QuicAddrFromString` to be consistent in what we accept. Unless you have a specific need for bracketed ip addresses, I think it would be fine to ignore them for now..

guhetier (commented):
Thanks for opening this PR!

src/platform/tls_openssl.c
```diff
@@ -150,6 +150,35 @@ typedef struct CXPLAT_TLS {
...
 
+static
+BOOLEAN
+CxPlatTlsServerNameIsIpLiteral(
+    _In_z_ const char* ServerName,
+    _In_ uint16_t ServerNameLength
+    )
+{
+    QUIC_ADDR Addr = {0};
+    if (QuicAddrFromString(ServerName, 0, &Addr)) {
```
guhetier: At least on Windows, this uses `RtlIpv4StringToAddressExA` with `Strict` set to false, which could lead to interpreting as IP lots of strings that don't look like an IP, for instance "net.host" or "128.net.host".

https://learn.microsoft.com/en-us/windows/win32/api/ip2string/nf-ip2string-rtlipv4stringtoaddressexa

We probably need something stricter for this validation, we could add a `CxPlatIsIpLiteral` helper next to `QuicAddrFromString` that focuses on strict validation.

src/platform/tls_openssl.c
flcl42: Brackets are not considered now, added them to the test as non-IPs, not sure if they really need to be covered though

src/platform/tls_openssl.c
flcl42: Switched to the system API

anrossi (approved):
Looks good to me! Thanks for making this PR.

guhetier (approved):

PR #5999, opened 2026-05-15
[2/3] Add XDP map support to the sample tools

ProjectsByJackHe:
## Description
**Full E2E demo:** [DEMO](https://microsoft.sharepoint.com/:v:/t/OSPlatformNetworking/cQoKsihSVC3kQ63QHixCTQ7hEgUCS3dPKZahhfyXC9GFfx8OFQ?nav=eyJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJTdHJlYW1XZWJBcHAiLCJyZWZlcnJhbFZpZXciOiJTaGFyZURpYWxvZy1MaW5rIiwicmVmZXJyYWxBcHBQbGF0Zm9ybSI6IldlYiIsInJlZmVycmFsTW9kZSI6InZpZXcifX0=&)

Adds 2 new sample tools to support changes in  [PR ](https://github.com/microsoft/msquic/pull/6003):
- **quicxdpmapserver.exe**: A minimal server to demonstrate usage of XDP maps.
- **orchestrator.exe**: meant to replicate a trusted admin process that can create programs/rules. It gives a map handle to the consumer quicadvanced server.

## Testing

Local testing on a VM with XDP

## Documentation

No


guhetier:
It seems like a great idea to add those features in a sample. I have no problem with an incremental implementation.
One concern though: the MsQuic sample can be used as a test tool, but it is mainly documentation to help user getting started. It is already pretty complicated, with features that a basic use case doesn't need. 

I'd consider having multiple samples instead, focused on demonstrating one way of using MsQuic, and avoid putting everything in a single one.

A sample focused on XDP / CIBIR / XDP maps could be a nice unit.

src/tools/CMakeLists.txt
```diff
@@ -38,4 +38,6 @@ if(WIN32 AND (NOT QUIC_UWP_BUILD AND NOT QUIC_GAMECORE_BUILD))
     add_subdirectory(execution)
     add_subdirectory(etw)
     add_subdirectory(recvfuzz)
+    add_subdirectory(quicxskmapcreator)
+    add_subdirectory(quicadvanced)
```
guhetier: We can come up with a better name.
This sample is about using XDP with maps, and potentially CIBIR? Let's call it something with XDP in the name.

I'd also consider making a subfolder to contain both `quicxskmapcreator` and `quicadvanced`, since they are meant to work together.

src/tools/quicxskmapcreator/quicxskmapcreator.c
```diff
@@ -0,0 +1,345 @@
...
+    attaches an XDP program with QUIC-aware match rules once the consumer
+    signals it has inserted its XSKs.
+
+    This tool is the "creator" half of the XSKMAP creator/consumer pattern
+    for MsQuic. The consumer is typically quicsample running with
+    -xdp -xdp_map_ifindex (Phase 2) or the MsQuic datapath itself (Phase 3).
+
+    Usage:
+        1. Start the QUIC server:
+           quicsample -server -cert_hash:<hash> -xdp -xdp_map_ifindex:<N>
```
guhetier: Please update this doc

src/tools/xdpmap/orchestrator/orchestrator.c
```diff
@@ -0,0 +1,345 @@
...
+        printf("  UdpPort: %u, CidOffset: %u (SidLen=%u + PidLen=%u), CidLength: %u\n",
+            UdpPort, CidOffset, CidServerIdLength, QUIC_CID_PID_LENGTH, CibirIdLength);
+    } else {
+        printf("  Rule: XDP_MATCH_UDP_DST, Port: %u\n", UdpPort);
+    }
+    printf("  Target: XDP_REDIRECT_TARGET_TYPE_XSKMAP_BY_QUEUEID\n");
+    printf("\n");
+    printf("Press ENTER to detach the program and exit.\n");
+    fflush(stdout);
+    (void)getchar();
```
guhetier, line 315: Is there a reason to wait? If this process is closed, the map stays open as long as the target process runs?

src/tools/quicxdpsample/quicxdpsample.c
```diff
@@ -0,0 +1,1367 @@
+/*++
+
+    Copyright (c) Microsoft Corporation.
+    Licensed under the MIT License.
+
+Abstract:
+
+    Provides a very simple MsQuic API sample server and client application.
```
guhetier: Please update this comment

src/tools/quicxdpsample/quicxdpsample.c
```diff
@@ -0,0 +1,1367 @@
...
+        goto Error;
+    }
+
+    //
+    // If the server will use XDP map mode, configure the map before
+    // RegistrationOpen (which triggers lazy init and creates the datapath).
+    // QUIC_PARAM_GLOBAL_XDP_MAP_CONFIG must be set before LazyInitComplete.
+    //
+#ifdef _WIN32
+    if (GetFlag(argc, argv, "server") && GetFlag(argc, argv, "xdp")) {
```
guhetier: Make this a helper function.

src/tools/quicxdpsample/quicxdpsample.c
```diff
@@ -0,0 +1,1367 @@
...
+        printf("MsQuicOpen2 failed, 0x%x!\n", Status);
+        goto Error;
+    }
+
+    //
+    // If the server will use XDP map mode, configure the map before
+    // RegistrationOpen (which triggers lazy init and creates the datapath).
+    // QUIC_PARAM_GLOBAL_XDP_MAP_CONFIG must be set before LazyInitComplete.
+    //
+#ifdef _WIN32
```
guhetier: I think this sample focuses on WIN32 only features, we can make it WIN32 only.

src/tools/quicxdpsample/quicxdpsample.c
```diff
@@ -0,0 +1,1367 @@
...
+        goto Error;
+    }
+
+    if (GetFlag(argc, argv, "help") || GetFlag(argc, argv, "?")) {
+        PrintUsage();
+    } else if (GetFlag(argc, argv, "client")) {
+        RunClient(argc, argv);
+    } else if (GetFlag(argc, argv, "multiclient")) {
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+        RunMultiClient(argc, argv);
```
guhetier: Consider simplifying even more.
The basic sample shows this, this one can focus only on how to use XDP and maps.
Let's keep the code as focused as possible on what the sample teaches.

src/tools/quicxdpsample/quicxdpsample.c
```diff
@@ -0,0 +1,1367 @@
...
+        if (MapIfIndexStr != NULL) {
+            UINT32 MapIfIndex = (UINT32)atoi(MapIfIndexStr);
+            char InputBuf[64];
+            UINT_PTR HandleValue;
+
+            printf("=== XDP Map Mode ===\n");
+            printf("  PID: %u\n", (unsigned)GetCurrentProcessId());
+            printf("  IfIndex: %u\n", MapIfIndex);
+            printf("\n");
+            printf("Start quicxskmapcreator in another terminal:\n");
```
guhetier: Unless we provide extra instructions to setup permissions, demonstrate some IPC, etc... there is little value having it done in a different process. The map could be created in this sample, with a comment saying "in practice, do this in your privileged process and share provide the map handle to this process".
We could point to the XDP sample that demonstrates the handle duplication.

guhetier: If it helps with manual testing to have the map creation done independently, consider:

- bringing the map creation in this tool, so there is a single sample for the XDP map logic
- having an option to control creating the map (either create map + run server/client or create map only and share with another process)

It will of course be helpful to split the implementation in two files, no need to put everything in one giant one.

src/tools/quicxdpsample/quicxdpsample.c
```diff
@@ -0,0 +1,1367 @@
...
+    // Open a handle to the library and get the API function table.
+    //
+    if (QUIC_FAILED(Status = MsQuicOpen2(&MsQuic))) {
+        printf("MsQuicOpen2 failed, 0x%x!\n", Status);
+        goto Error;
+    }
+
+    //
+    // If the server will use XDP map mode, configure the map before
+    // RegistrationOpen (which triggers lazy init and creates the datapath).
```
guhetier: nit: In a sample, comments should focus on user concepts, not internal implementation details.

"Configure the map using `QUIC_PARAM_GLOBAL_XDP_MAP_CONFIG` it must be done before `RegistrationOpen`"

src/tools/quicxdpsample/quicxdpsample.c
mtfriesen, line 1: Generally LGTM but instead of `quicadvanced` this module should have a finer-grained name, similar to the clear name `quicxskmapcreator`.

ProjectsByJackHe: something like `quicxskmapconsumer` or `quicsamplexdpaware`?

ProjectsByJackHe: I like `quicsamplexdpcibiraware`

ProjectsByJackHe: Or just `quicxdpsample`

mtfriesen: It really depends on the intended long-term scope of the sample. Like Guillaume's comment above, I'd suggest making this sample pretty specific, and thus having a specific name. Samples that show how to do everything (e.g., https://learn.microsoft.com/en-us/samples/microsoft/windows-driver-samples/windows-filtering-platform-sample/) are unwieldy and better as test tools than demonstrations of how to use APIs.

mtfriesen: It looks like this is supposed to be a sample of various XDP usages, so the name `quicxdpsample` seems reasonable. That said, if that's the intended case, I think the first line of the abstract should mention XDP, and also suggest `xdpsample` as the ALPN unless it is supposed to interop with other QUIC samples.

ProjectsByJackHe: given that quicxdpsample is a superset of the quicsample already, we can update the ALPN and add more details about XDP in the abstract.

mtfriesen (dismissed):

src/tools/quicxdpsample/quicxdpsample.c
guhetier: > given that quicxdpsample is a superset of the quicsample already

I don't think it should be. Every option that is not essential to demonstrating xdp, maps and CIBIR should be removed, so that the boilerplate is out of the way.

src/tools/quicxdpsample/quicxdpsample.c
```diff
@@ -0,0 +1,1371 @@
...
+{
+    #define HEX_TO_CHAR(x) ((x) > 9 ? ('a' + ((x) - 10)) : '0' + (x))
+    for (uint8_t i = 0; i < BufferLen; i++) {
+        HexString[i*2]     = HEX_TO_CHAR(Buffer[i] >> 4);
+        HexString[i*2 + 1] = HEX_TO_CHAR(Buffer[i] & 0xf);
+    }
+}
+
+void
+WriteSslKeyLogFile(
```
guhetier: Let's remove this from this sample, this isn't the topic.

src/tools/quicxdpsample/quicxdpsample.c
```diff
@@ -0,0 +1,1371 @@
...
+    // Print XDP status info after the listener starts.
+    //
+#ifdef _WIN32
+    if (GetFlag(argc, argv, "xdp")) {
+        const char* CibirIdHex = GetValue(argc, argv, "cibir_id");
+        BOOLEAN MapMode = (GetValue(argc, argv, "xdp_map_ifindex") != NULL);
+
+        printf("\n");
+
+        if (MapMode) {
```
guhetier: Please extract this in a helper.
This is boilerplate code that is not helpful to a reader to understand what they need to do to implement a connection, let's not have it in the way.

src/tools/quicxdpsample/quicxdpsample.c
```diff
@@ -0,0 +1,1371 @@
...
+        //
+        // The connection has completed the shutdown process and is ready to be
+        // safely cleaned up.
+        //
+        printf("[conn][%p] All done\n", Connection);
+        if (!Event->SHUTDOWN_COMPLETE.AppCloseInProgress) {
+            MsQuic->ConnectionClose(Connection);
+        }
+        break;
+    case QUIC_CONNECTION_EVENT_RESUMPTION_TICKET_RECEIVED:
```
guhetier: We probably don't need this in the xdp sample.

src/tools/quicxdpsample/quicxdpsample.c
```diff
@@ -0,0 +1,1371 @@
...
+    }
+
+    //
+    // Optionally set the CIBIR ID on the connection for CID-based steering.
+    //
+    {
+        const char* CibirIdHex = GetValue(argc, argv, "cibir_id");
+        if (CibirIdHex != NULL) {
+            //
+            // CIBIR requires shared binding (so the connection uses source CIDs).
```
guhetier: Are you sure of that? Wouldn't a client connection put the CIBIR id on the destination CID?

src/tools/quicxdpsample/quicxdpsample.c
```diff
@@ -0,0 +1,1371 @@
...
+}
+
+
+//
+// Runs the multi client side of the protocol.
+//
+#ifdef QUIC_API_ENABLE_PREVIEW_FEATURES
+
+void
+RunMultiClient(
```
guhetier: Let's remove this, there is nothing XDP specific to it.

src/tools/CMakeLists.txt
```diff
@@ -38,4 +38,6 @@ if(WIN32 AND (NOT QUIC_UWP_BUILD AND NOT QUIC_GAMECORE_BUILD))
     add_subdirectory(execution)
     add_subdirectory(etw)
     add_subdirectory(recvfuzz)
+    add_subdirectory(quicxskmapcreator)
+    add_subdirectory(quicxdpsample)
```
guhetier: Because these two tools are meant to work together, I would:
- either put the two in a common subfolder (xdpmap/orchestrator and xdpmap/quicpeer for instance)
- or have a single exe to do everything based on a "-role" parameter

guhetier: it should also let you easily share some code like the hex to uint8_t array helpers

src/tools/quicxdpsample/quicxdpsample.c
```diff
@@ -0,0 +1,1371 @@
...
+
+    fflush(File);
+    fclose(File);
+}
+
+//
+// Allocates and sends some data over a QUIC stream.
+//
+void
+ServerSend(
```
guhetier: I'd honestly consider not keeping the stream send in this sample.
That's not the point, once the connection is established, everything there is to demonstrate about XDP map is demonstrated. The generic sample is already there for the basic MsQuic usage, lets keep this one focused and as short as possible so what matters is clear.

guhetier (commented):
I think this needs to be scoped down to only XDP map used, with everything else removed. Think of it as an implementation guide, not a convenient tool for manual tests.

A generic XDP sample is not needed: using XDP is nothing more than one setting. Everything else is already in the main sample.

There is a lot of value demonstrating how to use XDP maps, though, and focusing on it only (with minimal noise) will help users.

src/tools/xdpmap/orchestrator/orchestrator.c
ProjectsByJackHe: If this process closes, the rules get destroyed also.

src/tools/quicxdpsample/quicxdpsample.c
ProjectsByJackHe: removed.

src/tools/quicxdpsample/quicxdpsample.c
ProjectsByJackHe: removed.

src/tools/quicxdpsample/quicxdpsample.c
ProjectsByJackHe: removed.

src/tools/CMakeLists.txt
```diff
@@ -38,4 +38,5 @@ if(WIN32 AND (NOT QUIC_UWP_BUILD AND NOT QUIC_GAMECORE_BUILD))
     add_subdirectory(execution)
     add_subdirectory(etw)
     add_subdirectory(recvfuzz)
+    add_subdirectory(xdpmap)
```
guhetier, line 41: nit: xdpmapsample to make it easier to find as a sample?

src/tools/xdpmap/orchestrator/orchestrator.c
guhetier: ah, ok, the map stays alive but not the rules?
Might be worth mentioning in a short comment, that looks like an easy gotcha for users.

src/tools/xdpmap/quicxdpmapserver/quicxdpmapserver.c
```diff
@@ -0,0 +1,461 @@
...
+        PrintUsage();
+        return (int)QUIC_STATUS_INVALID_PARAMETER;
+    }
+
+    if (QUIC_FAILED(Status = MsQuicOpen2(&MsQuic))) {
+        printf("MsQuicOpen2 failed, 0x%x!\n", Status);
+        goto Error;
+    }
+
+    if (!ConfigureXdpMap(argc, argv)) {
```
guhetier: Consider splitting the argument parsing logic, getting the map handle and MsQuic logic.

- have a function to parse all arguments and populate a struct with the values (I guess globals are fine too for a C-style sample)
- have a function prompt for the map handle
- only then, open msquic, configure the map and run the server

It will separate the logic users don't care much about because they will have to do it differently from what you demonstrate.

src/tools/xdpmap/quicxdpmapserver/quicxdpmapserver.c
```diff
@@ -0,0 +1,461 @@
...
+    if (QUIC_FAILED(Status = MsQuic->RegistrationOpen(&RegConfig, &Registration))) {
+        printf("RegistrationOpen failed, 0x%x!\n", Status);
+        goto Error;
+    }
+
+    RunServer(argc, argv);
+
+Error:
+    if (MsQuic != NULL) {
+        if (Configuration != NULL) {
```
guhetier: The configuration doesn't need to be global, it can be scoped in `RunServer`.
Make it the context for your listener callback and close it right the listener is closed.

src/tools/CMakeLists.txt
ProjectsByJackHe: I like your earlier suggestion of "xdpmap"

src/tools/xdpmap/quicxdpmapserver/quicxdpmapserver.c
ProjectsByJackHe: 1. added a ParseArgs function
2. added a PromptMapHandle function
3. ConfigureMap + RunServer

src/tools/xdpmap/quicxdpmapserver/quicxdpmapserver.c
```diff
@@ -214,12 +258,13 @@ ServerListenerCallback(
 static
 BOOLEAN
 LoadServerConfiguration(
-    _In_ int argc,
-    _In_reads_(argc) _Null_terminated_ char* argv[]
+    _In_ const SERVER_ARGS* Args,
+    _Out_ HQUIC* ConfigurationOut
```
guhetier, line 256: nit: Set `ConfigurationOut` to NULL at the start of the function

src/tools/xdpmap/quicxdpmapserver/quicxdpmapserver.c
```diff
@@ -230,91 +275,72 @@ LoadServerConfiguration(
...
+            CredHelper.CredConfig.Type = QUIC_CREDENTIAL_TYPE_CERTIFICATE_FILE;
+            CredHelper.CredConfig.CertificateFile = &CredHelper.CertFile;
         }
-    } else {
-        printf("Must specify '-cert_hash' or '-cert_file' with '-key_file'.\n");
-        return FALSE;
     }
 
-    if (QUIC_FAILED(Status = MsQuic->ConfigurationOpen(Registration, &Alpn, 1, &Settings, sizeof(Settings), NULL, &Configuration))) {
+    if (QUIC_FAILED(Status = MsQuic->ConfigurationOpen(Registration, &Alpn, 1, &Settings, sizeof(Settings), NULL, &Config))) {
```
guhetier: nit: I recommend against this pattern, it is harder to read.
```
Status = ....
if (QUIC_FAILED(Status)) { ... }
```

src/tools/xdpmap/quicxdpmapserver/quicxdpmapserver.c
```diff
@@ -230,91 +275,72 @@ LoadServerConfiguration(
...
 }
 
 static
 BOOLEAN
-ConfigureXdpMap(
-    _In_ int argc,
-    _In_reads_(argc) _Null_terminated_ char* argv[]
+PromptForXdpMapHandle(
+    _In_ const SERVER_ARGS* Args,
+    _Out_ UINT_PTR* HandleValueOut
```
guhetier, line 325: Same here, init out param to NULL at the start of the function so they are always set on early return.

src/tools/xdpmap/quicxdpmapserver/quicxdpmapserver.c
guhetier, line 415: nit: Consider moving this to the argument parsing function

src/tools/xdpmap/quicxdpmapserver/quicxdpmapserver.c
```diff
@@ -0,0 +1,495 @@
...
+    QUIC_CREDENTIAL_CONFIG CredConfig;
+    union {
+        QUIC_CERTIFICATE_HASH CertHash;
+        QUIC_CERTIFICATE_HASH_STORE CertHashStore;
+        QUIC_CERTIFICATE_FILE CertFile;
+        QUIC_CERTIFICATE_FILE_PROTECTED CertFileProtected;
+    };
+} QUIC_CREDENTIAL_CONFIG_HELPER;
+
+_IRQL_requires_max_(DISPATCH_LEVEL)
```
guhetier: Let's avoid kernel mode related SAL annotation for a user-mode sample.

guhetier (dismissed):
A few nits it would be better to address, but seems good otherwise.

src/tools/xdpmap/quicxdpmapserver/quicxdpmapserver.c
```diff
@@ -0,0 +1,495 @@
...
+
+    switch (Event->Type) {
+    case QUIC_CONNECTION_EVENT_CONNECTED:
+        printf("[conn][%p] Connected\n", Connection);
+        break;
+    case QUIC_CONNECTION_EVENT_SHUTDOWN_INITIATED_BY_TRANSPORT:
+        printf("[conn][%p] Shutdown by transport, 0x%x\n", Connection, Event->SHUTDOWN_INITIATED_BY_TRANSPORT.Status);
+        break;
+    case QUIC_CONNECTION_EVENT_SHUTDOWN_INITIATED_BY_PEER:
+        printf("[conn][%p] Shutdown by peer, 0x%llu\n", Connection, (unsigned long long)Event->SHUTDOWN_INITIATED_BY_PEER.ErrorCode);
```
guhetier: Shouldn't this be `%llx`?

src/tools/xdpmap/orchestrator/orchestrator.c
```diff
@@ -0,0 +1,323 @@
...
+                LOGERR("Missing TargetPid value");
+                return FALSE;
+            }
+            TargetPid = (UINT32)atoi(ArgV[i]);
+        } else if (!_stricmp(ArgV[i], "-IfIndex")) {
+            if (++i >= ArgC) {
+                LOGERR("Missing IfIndex value");
+                return FALSE;
+            }
+            IfIndex = (UINT32)atoi(ArgV[i]);
```
guhetier, line 110: Consider rejecting `IfIndex == 0` here. `atoi` returns 0 for a missing or non-numeric argument, so a value like `-IfIndex foo` (or an explicit `0`) passes validation and is handed straight to `XdpCreateProgram`. Notably the consumer side rejects ifindex 0 explicitly (`quicxdpmapserver.c` `-xdp_map_ifindex` handling), so the two tools currently disagree on what a valid interface index is. For a reference sample it would be clearer to validate the parsed value (e.g. reject 0 and/or confirm the argument was numeric) right after parsing.

src/tools/xdpmap/xdpmap_common.h
```diff
@@ -0,0 +1,45 @@
...
+XdpMapDecodeHexChar(
+    char c
+    )
+{
+    if (c >= '0' && c <= '9') return (uint8_t)(c - '0');
+    if (c >= 'A' && c <= 'F') return (uint8_t)(10 + c - 'A');
+    if (c >= 'a' && c <= 'f') return (uint8_t)(10 + c - 'a');
+    return 0;
+}
+
```
guhetier: This hex decoder silently accepts invalid input. `XdpMapDecodeHexChar` returns `0` for any non-hex character, so a string like `-cert_hash:ZZ...` decodes to a wrong-but-plausible value with no error, and `XdpMapDecodeHexBuffer` only counts complete pairs so a trailing odd nibble is dropped silently. The length check in the caller catches wrong *counts* but not invalid digits, so a malformed cert hash can slip through as a valid-length-but-incorrect hash. For an exemplary sample, consider signaling an error on any non-hex digit (and on odd-length input) so bad arguments are rejected rather than silently mis-decoded.

guhetier (approved):

PR #6006, opened 2026-05-18
[2/3] Update XDP Submodules

ProjectsByJackHe:
## Description

Update XDP submodules to include headers for XSK maps

## Testing

N/A

## Documentation

N/A

### DIFF
--- a/src/platform/datapath_raw_xdp_win.c
+++ b/src/platform/datapath_raw_xdp_win.c
@@ -11,6 +11,8 @@ Abstract:
 
 #define _CRT_SECURE_NO_WARNINGS 1 // TODO - Remove
 #define XDP_API_VERSION 3
+#define XDP_MINIMUM_MAJOR_VER 1
+#define XDP_MINIMUM_MINOR_VER 1
 #define XDP_INCLUDE_WINCOMMON
 
 #include <xdp/wincommon.h>


src/platform/datapath_raw_xdp_win.c
```diff
@@ -11,6 +11,8 @@
 
 #define _CRT_SECURE_NO_WARNINGS 1 // TODO - Remove
 #define XDP_API_VERSION 3
+#define XDP_MINIMUM_MAJOR_VER 1
+#define XDP_MINIMUM_MINOR_VER 1
```
ProjectsByJackHe, line 15: this is needed because the newer XDP lib headers referenced by msquic calls into an older xdp driver (v1.1.0) deployed by the CI.

given that XDP hasn't released a version with the map features we want, lets add these ifdefs so that the header-only lib can understand and handle failures properly by falling back to an LKG.

guhetier: Before you add these, what is our minimum supported XDP version? Was it already v1.1.0, or do we happen to use v1.1.0 in the pipeline but would support an older one?

ProjectsByJackHe: There's no formally declared minimum supported XDP version. Historically, msquic has always compiled and tested against the same XDP version as the driver version defined in xdp.json. The xdp.json reflects what CI tests against, not a documented compatibility floor. Prior to this, the submodule was at v1.1.0 and CI installed v1.1.0.

guhetier:
Is this bumping to the head of XDP main? I can't remember what is XDP policy, but before we ship anything, we need to make sure we depend on a supported version, even if only for the header.

ProjectsByJackHe:
> Is this bumping to the head of XDP main? I can't remember what is XDP policy, but before we ship anything, we need to make sure we depend on a supported version, even if only for the header.

Yes this is bumping to XDP main. There is no other way to include the latest headers that has the map API. Maybe XDP can roll out a quick v1.3.1?

guhetier:
> > Is this bumping to the head of XDP main? I can't remember what is XDP policy, but before we ship anything, we need to make sure we depend on a supported version, even if only for the header.
> 
> Yes this is bumping to XDP main. There is no other way to include the latest headers that has the map API. Maybe XDP can roll out a quick v1.3.1?

Let's see with @mtfriesen what is xdp-for-windows policy for this.

src/platform/datapath_raw_xdp_win.c
guhetier: Ok. That sounds part of what we need to figure out, in the more general context of "what is the proper way of taking build time and runtime dependencies on XDP".

What are the consequences of using these defines for someone using XDP v1.0? We need to be intentional if it used to work and won't work anymore in the next release.

ProjectsByJackHe: Including these ifdefs are helpful at best and no-ops at worst, as it just compiles in the fallback code paths for when someone is using a newer api in the header not existing in an older driver.

ProjectsByJackHe: So, no harm to someone using XDP driver version v1.0. In fact, if we intend to go with upgrading to use the most recent XDP header library, we must have these ifdefs to fallback on.

guhetier: Ok, I didn't realize the maps were the first time we used this.
As long as we are using `_XdpOpen`, it shouldn't matter but this will become useful once we replace the other `_XdpOpen` calls with more specific ones I guess.

guhetier (approved):
Before we release v2.6, the submodule will need to be updated to a version xdp-for-windows is committing to support.
In general, we should clarify how we ingest XDP in MsQuic.

mtfriesen:
Thanks @ProjectsByJackHe and @guhetier. To converge a few threads and add my own thoughts:

1. The most agile solution for prototyping and testing pre-release features would be to use the XDP submodule to build a matching runtime, too. In this model, neither the SDK or runtime nuget packages would be used for testing pre-release XDP features in MsQuic.
2. MsQuic should probably have a test _matrix_ of XDP runtime versions, with at least the minimum supported official XDP runtime, and ideally also the most recent official XDP runtime version, with intermediate runtime versions having less test value. These should always use the official XDP-supported runtime nuget packages.
3. We need to ensure MsQuic production binaries (release/*) are built only against supported XDP code, whether via submodule or SDK nuget. This means we need to either converge all build-time dependencies onto an official XDP release prior to forking MsQuic for release, or have compile-time options to test prerelease features in CI while compiling a subset of features into production binaries.

As it stands, this PR looks good to get the submodule sync'd up in main.

